PR #945, opened 2025-05-24
[Bugfix] Adjust inputbatch to be compatible with latest vllm

MengqingCao:
Adjust inputbatch to be compatible with latest vllm, as kvcache group feature has been redo in https://github.com/vllm-project/vllm/pull/18593


wangxiyuan:
quick merge to unblock CI

PR #839, opened 2025-05-13
feat: support compile torchair graph while warming up

NeverRaR:
### What this PR does / why we need it?

feat: support compile torchair graph while warming up

PR #947, opened 2025-05-24
[perf]Support MOE Multi-stream in Deepseek

David9857:
Support MOE inner Multi-stream for Deepseek. 
This feature requires graph mode with mc2 enabled.


vllm_ascend/quantization/w8a8_dynamic.py
```diff
@@ -96,25 +115,39 @@ def apply_mlp(hidden_states_wrapper: List[torch.Tensor],
...
+                           topk_ids: torch.Tensor,
+                           top_k: int,
+                           expert_map: torch.Tensor = None,
+                           moe_all_to_all_group_name: str = "",
+                           **kwargs) -> torch.Tensor:
     global_bs = 0
     moe_expert_num = len(expert_map)
     # hidden_states = hidden_states.bfloat16()
-    kwargs = {
+    kwargs1 = {
```
wangxiyuan: rename to a readable name

vllm_ascend/envs.py
```diff
@@ -36,6 +36,8 @@
     lambda: bool(int(os.getenv("COMPILE_CUSTOM_KERNELS", "1"))),
     "VLLM_ENABLE_MC2":
     lambda: bool(int(os.getenv("VLLM_ENABLE_MC2", '0'))),
+    "VLLM_ENABLE_CV_PARALLEL":
```
wangxiyuan: use additional_config instead of env, since this change is only used for torchair GE mode. like https://github.com/vllm-project/vllm-ascend/pull/839 does, there are another 3 new config option coming. 

how about
```
{
   "additional_config": {
      "torchair_graph_config": {
         "enable": True,
         "enable_cv_parallet": True,
         "batch_sizes": "12345",
         "batch_sizes_init": True
      } 
  }
}

```
cc @zzzzwwjj

wangxiyuan:
And don't forget add e2e test. The model weight is here: https://www.modelscope.cn/models/vllm-ascend/DeepSeek-V2-Lite-W8A8 

take https://github.com/vllm-project/vllm-ascend/blob/main/tests/multicard/test_offline_inference_distributed.py#L49 as an example

vllm_ascend/models/deepseek_v2.py
```diff
@@ -179,6 +180,12 @@ def __init__(
         else:
             self.gate.e_score_correction_bias = None
 
+        self.enable_cv_parallel = False
+        additional_config = get_current_vllm_config().additional_config
```
wangxiyuan: please use ascend_config instead now. Note that doc should be updated at the same time.

PR #964, opened 2025-05-27
[Doc][0.7.3] Improve sleep documentation

Potabk:
Add `COMPILE_CUSTOM_KERNELS =1` note for sleep mode

### DIFF
--- a/docs/source/installation.md
+++ b/docs/source/installation.md
@@ -136,7 +136,8 @@ pip install vllm-ascend==|pip_vllm_ascend_version| --extra-index https://downloa
 ```
 
 :::{dropdown} Click here to see "Build from source code"
-or build from **source code**:
+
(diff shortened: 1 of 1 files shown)


Potabk:
@wangxiyuan cc

docs/source/installation.md
```diff
@@ -136,7 +136,8 @@ pip install vllm-ascend==|pip_vllm_ascend_version| --extra-index https://downloa
 ```
 
 :::{dropdown} Click here to see "Build from source code"
-or build from **source code**:
+
```
Yikun, line 139: https://github.com/vllm-project/vllm-ascend/blob/main/docs/source/installation.md?plain=1#L163-L170

I guess you could just refer this

Potabk: on v0.7.3 can we set this as an optional step rather than a note?

Yikun: it's okay to keep same to reduce mantain cost

PR #970, opened 2025-05-27
add optimze of dsv3

momo609:

PR #980, opened 2025-05-27
perf: speed up topk_topp_sampler

ZhengWG:

Under temperature > 0, the sampler logic's processing time increased to 8ms from 3ms for a batch size of 24 per dp rank, with the scatter operation accounting for 6ms. By optimizing the scatter operator with reference to the TPU implementation, we've reduced the total sampler time to 2ms.

No

Tested with vllm_ascend, verifying correct behavior in TopKTopPSampler using forward_npu.


wangxiyuan:
https://github.com/vllm-project/vllm-ascend/pull/970 is involved with the change and you have been added as co-author. Close this PR now.

PR #987, opened 2025-05-28
[Kernel] Remove cumsum in groupedmatmul

hahazhky:
 remove cumsum operator in MOE to improve performance

it should be tested on a case with mc2 operator and graph mode enabled

### DIFF
--- a/vllm_ascend/ops/fused_moe.py
+++ b/vllm_ascend/ops/fused_moe.py
@@ -89,15 +89,14 @@ def fused_experts_with_mc2(
         0:5]
 
     w1 = w1.transpose(1, 2)
-    expert_token_nums = torch.cumsum(expert_token_nums,
-                                     dim=0,
-                                     dtype=torch.int64)
+
     group_list = expert_token_nums.to(torch.int64)
     gate_up_out_list = torch_npu.npu_grouped_matmul(
         x=[expand_x],
         weight=[w1],
         split_item=2,
-        group_list_type=0,
+        # 1 means count mode, to avoid cumulative operation of the group list
+        group_list_type=1,
         group_type=0,
         group_list=group_list,
     )
@@ -111,7 +110,7 @@ def fused_experts_with_mc2(
         x=[gate_up_out],
         weight=[w2],
         split_item=2,
-        group_list_type=0,
+        group_list_type=1,
         group_type=0,
         group_list=group_list,
     )


vllm_ascend/ops/fused_moe.py
```diff
@@ -97,15 +104,17 @@ def fused_experts_with_mc2(
...
+        uniform_group_list = hidden_states.shape[0] * all_to_all_group_size * top_k // moe_expert_num
+        group_list = torch.Tensor([uniform_group_list] * w1.shape[0]).long().npu()
+    else:
+        group_list = expert_token_nums
     gate_up_out_list = torch_npu.npu_grouped_matmul(
         x=[expand_x],
         weight=[w1],
         split_item=2,
-        group_list_type=0,
+        group_list_type=1,
```
MengqingCao, line 99: Maybe we could add some comments to explain why `group_list_type=1` is used here, e.g., to avoid the cumulative calculation of the group list.

vllm_ascend/ops/fused_moe.py
```diff
@@ -97,15 +104,17 @@ def fused_experts_with_mc2(
...
 
     w1 = w1.transpose(1, 2)
-    expert_token_nums = torch.cumsum(expert_token_nums,
-                                     dim=0,
-                                     dtype=torch.int64)
-    group_list = expert_token_nums.to(torch.int64)
+
+    if VLLM_ENABLE_FIX_ROUTE:
+        uniform_group_list = hidden_states.shape[0] * all_to_all_group_size * top_k // moe_expert_num
+        group_list = torch.Tensor([uniform_group_list] * w1.shape[0]).long().npu()
```
MengqingCao:
```suggestion
        group_list = torch.Tensor([uniform_group_list] * w1.shape[0]).long().to(hidden_states.device)
```

vllm_ascend/ops/fused_moe.py
```diff
@@ -58,6 +59,14 @@ def fused_experts_with_mc2(
...
     global_bs = 0
     moe_expert_num = len(expert_map)
+
+    rank = torch.distributed.get_rank()
+    if VLLM_ENABLE_FIX_ROUTE:
+        step = hidden_states.shape[0] * top_k
+        uniform_topk_list = [
+            (i + rank) % moe_expert_num for i in range(rank * step, (rank + 1) * step)
+        ]
+        topk_ids = torch.Tensor(uniform_topk_list).int().view(hidden_states.shape[0], -1).npu()
```
MengqingCao:
```suggestion
        topk_ids = torch.Tensor(uniform_topk_list).int().view(hidden_states.shape[0], -1).to(hidden_states.device)
```

MengqingCao:
overall lgtm now, thanks!

vllm_ascend/ops/fused_moe.py
hahazhky: fixed

vllm_ascend/ops/fused_moe.py
hahazhky: fixed

vllm_ascend/ops/fused_moe.py
hahazhky: fixed

hahazhky:
@wangxiyuan @Yikun @ganyi1996ppo hello, please review this pr, @MengqingCao has replied lgtm

vllm_ascend/ops/fused_moe.py
```diff
@@ -50,6 +51,14 @@ def fused_experts_with_mc2(
 ) -> torch.Tensor:
     global_bs = 0
     moe_expert_num = len(expert_map)
+
+    rank = torch.distributed.get_rank()
+    if VLLM_ENABLE_FIX_ROUTE:
```
ganyi1996ppo: Performance test under the fix route scenario seems pointless, do we really need this env variable in the repo?

ganyi1996ppo:
Looks good except the fix routing part, have we really need to add this? Though the performance may gets better with this opened, but this performance data means nothing in the real production environment right? @hahazhky

hahazhky:
> Looks good except the fix routing part, have we really need to add this? Though the performance may gets better with this opened, but this performance data means nothing in the real production environment right? @hahazhky

yes，we cannot open this in real production environment，but we can use it as a performance debug option to check the upper limit，and with the training continues，time-cost will be close to this performance

vllm_ascend/envs.py
```diff
@@ -68,6 +68,9 @@
     lambda: os.getenv("VLLM_VERSION", None),
     "VLLM_ASCEND_TRACE_RECOMPILES":
     lambda: bool(int(os.getenv("VLLM_ASCEND_TRACE_RECOMPILES", '0'))),
+    # dispatch tokens to experts averagely for performance test
+    "VLLM_ENABLE_FIX_ROUTE":
+    lambda: bool(int(os.getenv("VLLM_ENABLE_FIX_ROUTE", '0'))),
```
Yikun:
```suggestion
    "VLLM_ASCEND_ENABLE_FIX_ROUTE":
    lambda: bool(int(os.getenv("VLLM_ASCEND_ENABLE_FIX_ROUTE", '0'))),
```

tests/multicard/test_offline_inference_distributed.py
```diff
@@ -61,3 +61,22 @@ def test_models_distributed_DeepSeek():
             distributed_executor_backend="mp",
     ) as vllm_model:
         vllm_model.generate_greedy(example_prompts, max_tokens)
+
+
+def test_models_distributed_fix_route_DeepSeek():
```
Yikun: we should unset ENV after test executed: `@patch.dict(os.environ, {"VLLM_ASCEND_ENABLE_TOPK_OPTIMZE": "1"})`

```suggestion
@patch.dict(os.environ, {"VLLM_ASCEND_ENABLE_TOPK_OPTIMZE": "1"})
def test_models_distributed_fix_route_DeepSeek():
```

[1] https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.dict

vllm_ascend/ops/fused_moe.py
Yikun: Please make sure `VLLM_ENABLE_FIX_ROUTE` is still needed or not after https://github.com/vllm-project/vllm-ascend/commit/afc4c0cd035ee22d51950d24b7f642abcaa0b176  (`enable_force_load_balance`)

huiyingCCCC: fix route is no longer needed, delete relevant code

tests/multicard/test_offline_inference_distributed.py
huiyingCCCC: fixed

vllm_ascend/envs.py
huiyingCCCC: deleted

tests/multicard/test_offline_inference_distributed.py
```diff
@@ -61,3 +61,21 @@ def test_models_distributed_DeepSeek():
...
+        "Briefly describe the major milestones in the development of artificial intelligence from 1950 to 2020.",
+        "Compare and contrast artificial intelligence with human intelligence in terms of processing information.",
+    ]
+    dtype = "half"
+    max_tokens = 5
+    with VllmRunner(
+            "deepseek-ai/DeepSeek-V2-Lite",
+            dtype=dtype,
+            tensor_parallel_size=8,
+            enable_expert_parallel=True,
```
Yikun: Could this change more general: `@pytest.mark.parametrize("enable_expert_parallel", [True, False])`?

Yikun (commented):
LGTM except inline comments

tests/multicard/test_offline_inference_distributed.py
```diff
@@ -61,3 +61,21 @@ def test_models_distributed_DeepSeek():
...
+        "vLLM is a high-throughput and memory-efficient inference and serving engine for LLMs.",
+        "Briefly describe the major milestones in the development of artificial intelligence from 1950 to 2020.",
+        "Compare and contrast artificial intelligence with human intelligence in terms of processing information.",
+    ]
+    dtype = "half"
+    max_tokens = 5
+    with VllmRunner(
+            "deepseek-ai/DeepSeek-V2-Lite",
+            dtype=dtype,
+            tensor_parallel_size=8,
```
Yikun: CI fail due to only 4 cards avaiable but you used tp8

wangxiyuan (approved):

ApsarasX (approved):

PR #969, opened 2025-05-27
[Bugfix][CI] Update guided decoding backend list

shen-shanshan:
### What this PR does / why we need it?

Update guided decoding backend list to make CI pass.

PR #989, opened 2025-05-28
[Bugfix][Worker] Clear NPU memory between test profiling

shen-shanshan:

Clear NPU memory between test profiling.

If we don't clear the memory this way, the CI may get `OOM` error between UTs.

This modification is adapted from `vllm/worker/worker.py`, find more details at https://github.com/vllm-project/vllm/blob/main/vllm/worker/worker.py#L184-L186.

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -15,7 +15,6 @@
 # This file is a part of the vllm-ascend project.
 #
 
-import gc
 import os
 from datetime import timedelta
 from typing import TYPE_CHECKING, Optional, Tuple
@@ -107,12 +106,6 @@ class NPUPlatform(Platform):
     @classmethod
     def mem_get_info(cls) -> Tuple[int, int]:
         return torch.npu.mem_get_info()
-
-    @classmethod
-    def clear_npu_memory(cls):
-        gc.collect()
-        torch.npu.empty_cache()
-        torch.npu.reset_peak_memory_stats()
 
     @classmethod
     def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
--- a/vllm_ascend/worker/worker.py
+++ b/vllm_ascend/worker/worker.py
@@ -23,7 +23,6 @@ from typing import Dict, List, Optional, Set, Tuple, Type, Union
 
 import msgpack  # type: ignore
 import torch
-import torch.distributed
 import zmq
 from torch import nn
 from vllm import envs
@@ -52,7 +51,7 @@ from vllm_ascend.device_allocator.camem import CaMemAllocator
 from vllm_ascend.distributed.parallel_state import init_ascend_model_parallel
 from vllm_ascend.platform import NPUPlatform
 from vllm_ascend.utils import (ACL_FORMAT_FRACTAL_ND, ACL_FORMAT_FRACTAL_NZ,
-                               is_310p, try_register_lib)
+                               is_310p, clear_npu_memory, try_register_lib)
 from vllm_ascend.worker.model_runner import NPUModelRunner
 from vllm_ascend.worker.pooling_model_runner import NPUPoolingModelRunner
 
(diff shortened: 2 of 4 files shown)


MengqingCao:
Thanks for this fixing! I think we can derectly move gc collection and peak mem status reset into `NPUPlatform.empty_cache()`

shen-shanshan:
> Thanks for this fixing! I think we can derectly move gc collection and peak mem status reset into `NPUPlatform.empty_cache()`

I have assembled the 3 methods into `clear_npu_memory()` in platform.

shen-shanshan:
@wangxiyuan The CI is passed, and this PR is may needed for https://github.com/vllm-project/vllm-ascend/pull/969.

MengqingCao:
> > Thanks for this fixing! I think we can derectly move gc collection and peak mem status reset into `NPUPlatform.empty_cache()`
> 
> I have assembled the 3 methods into `clear_npu_memory()` in platform.

Why not just do it in https://github.com/vllm-project/vllm-ascend/pull/989/files#diff-efe99563375ba645b1c1befb237632c2c91f85b21d2ab2d1095eef82aecd3999L106

shen-shanshan:
> > > Thanks for this fixing! I think we can derectly move gc collection and peak mem status reset into `NPUPlatform.empty_cache()`
> > 
> > 
> > I have assembled the 3 methods into `clear_npu_memory()` in platform.
> 
> Why not just do it in https://github.com/vllm-project/vllm-ascend/pull/989/files#diff-efe99563375ba645b1c1befb237632c2c91f85b21d2ab2d1095eef82aecd3999L106

Because maybe there are some other scenarios just need `empty_cache()` and don't need do the other 2 methods.

wangxiyuan:
please merge this pr and https://github.com/vllm-project/vllm-ascend/pull/969 into one

Potabk:
hope can merge, sleep mode v1 need this

shen-shanshan:
> hope can merge, sleep mode v1 need this

I will rebase soon.

shen-shanshan:
@wangxiyuan The CI is finally passed and this PR can be merged.

wangxiyuan (approved):

wangxiyuan:
@ganyi1996ppo @Yikun @jianzs  please doubel check. Thanks.

vllm_ascend/platform.py
```diff
@@ -307,3 +301,9 @@ def stateless_init_device_torch_dist_pg(
 
         pg._register_backend(device, backend_type, backend_class)
         return pg
+
+    @classmethod
+    def clear_npu_memory(cls):
```
ganyi1996ppo: This seems not a public interface inherited from vllm `Platform`, can we just put it into the files like utils.py?

shen-shanshan: @ganyi1996ppo I put this method here just because I think this method is highly related to the platform, and can be called using `current_platform` directly, which I think is easy to understand. 

If you think this method is inappropriate to put here and must move to `ustils.py`, I can also do this.

wangxiyuan: I agree with @ganyi1996ppo . the func in platform should be only from vllm

shen-shanshan: > I agree with @ganyi1996ppo . the func in platform should be only from vllm

OK, since it is so, I will modify it soon.

Yikun:
Does this has some impacts on RL case? Need to be confirmed otherwise LGTM

vllm_ascend/worker/worker.py
```diff
@@ -280,7 +278,7 @@ def determine_num_available_blocks(self) -> Tuple[int, int]:
         """
         # Profile the memory usage of the model and get the maximum number of
         # cache blocks that can be allocated with the remaining free memory.
-        NPUPlatform.empty_cache()
+        clear_npu_memory()
```
momo609, line 283: What is the difference between the two memory clear methods?

shen-shanshan: @momo609 `clear_npu_memory()` = `gc.collect()` + `empty_cache()` + `reset_peak_memory_stats()` (which will make the profile in dummy run more accurate).

shen-shanshan:
CC: @wangxiyuan

PR #997, opened 2025-05-29
Support multistream of shared experts in FusedMoE

sdmyzlp:
Contains on #1111 for completeness.

Implement multi-stream parallelism for MoE layers with shared experts, where computation of shared experts will be overlapped with expert token dispatch and combine. Also, when multi-stream is enabled, weights of shared experts will be force to replicate across all cards, regardless of any tensor parallelism configurations, to avoid AllReduce operations.

With the expected overlaping being:
```
| shared gate_up | shared act |              | shared down |
|    dispatch    | routed gate_up, act, down |   combine   |
```

No.

Tested on 1x16 910 node, with tailored 2 layer DSKv2.

### DIFF
--- a/vllm_ascend/ops/fused_moe.py
+++ b/vllm_ascend/ops/fused_moe.py
@@ -16,7 +16,7 @@
 # Adapted from vllm/tests/kernels/test_moe.py
 
 import os
-from typing import Any, Callable, List, Optional
+from typing import Any, Callable, List, Optional, Tuple, Union
 
 import torch
 import torch.distributed as dist
@@ -36,6 +36,7 @@ import vllm_ascend.envs as envs_ascend
 from vllm_ascend.ascend_config import get_ascend_config
 from vllm_ascend.distributed.parallel_state import get_ep_group, get_etp_group
 from vllm_ascend.ops.expert_load_balancer import ExpertLoadBalancer
+from vllm_ascend.utils import npu_stream_switch, npu_wait_tensor
 
 VLLM_ENABLE_MC2: bool = envs_ascend.VLLM_ENABLE_MC2
 USING_LCCL_COM: bool = envs_ascend.USING_LCCL_COM
@@ -106,15 +107,17 @@ def process_topk_ids(topk_ids: torch.Tensor, expert_num: int, ep_size: int,
     return topk_ids_pad, unpad_indices
 
 
-def fused_experts_with_mc2(hidden_states: torch.Tensor,
-                           w1: torch.Tensor,
-                           w2: torch.Tensor,
-                           topk_weights: torch.Tensor,
-                           topk_ids: torch.Tensor,
-                           top_k: int,
-                           expert_map: torch.Tensor = None,
-                           moe_all_to_all_group_name: Optional[str] = None,
-                           **kwargs) -> torch.Tensor:
+def fused_experts_with_mc2(
+    hidden_states: torch.Tensor,
+    w1: torch.Tensor,
+    w2: torch.Tensor,
+    topk_weights: torch.Tensor,
+    topk_ids: torch.Tensor,
+    top_k: int,
+    expert_map: torch.Tensor = None,
+    moe_all_to_all_group_name: Optional[str] = None,
+    shared_experts: Optional[Any] = None
+) -> Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
     global_bs = 0
     moe_expert_num = len(expert_map)
     kwargs_mc2 = {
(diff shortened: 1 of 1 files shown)


MengqingCao:
You can run `bash format.sh` locally to fix lint failures

sdmyzlp:
> You can run `bash format.sh` locally to fix lint failures

Done, with mypy error code `import-not-found` locally disabled for `vllm_ascend/utils.py`.

ganyi1996ppo:
@sdmyzlp Can you upload the profiling graph on this part? so we can have a more intuisive perspective on this PR.

vllm_ascend/ops/fused_moe.py
```diff
@@ -83,11 +85,20 @@ def fused_experts_with_mc2(
     }
     kwargs.update(stage1_kwargs)
 
+    if shared_experts is not None:
```
ganyi1996ppo: Have you tried to launch this after the dispatch op? If the dispatch will block enough time on the first stream, launch this after it seems can better overlap the execution part over host side right?

sdmyzlp: Fixed, with all secondary-stream operations launched after the corresponding main-stream operation. 

By the way, this patch only means to implement multi-stream shared experts for graph mode decode, operations will still be executed sequentially otherwise. One may extend `npu_switch_stream` / `npu_wait_tensor` in future to support eager mode multi-stream functionality.

sdmyzlp:
> @sdmyzlp Can you upload the profiling graph on this part? so we can have a more intuisive perspective on this PR.

Described expected overlap using ascii in commit message, I have trouble uploading screenshot of the profiling.

vllm_ascend/models/deepseek_v2.py
```diff
@@ -83,20 +83,35 @@ def __init__(
...
+                quant_config=quant_config,
+                prefix=f"{prefix}.gate_up_proj")
+            self.down_proj = RowParallelLinear(intermediate_size,
+                                               hidden_size,
+                                               bias=False,
+                                               quant_config=quant_config,
+                                               reduce_results=reduce_results,
+                                               prefix=f"{prefix}.down_proj")
+        else:
+            self.gate_up_proj = ReplicatedLinear(
```
ganyi1996ppo: Why do we need replicate linear here?

sdmyzlp: Otherwise, an `all_reduce` operation must follow `down` operation of the shared expert, forcing the overlapping to be:
```
| shared gate_up | shared act |              | shared down |  all_reduce  |
|    dispatch    | routed gate_up, act, down |   combine   |              |  add  |
```
By using `ReplicatedLinear`, no `all_reduce` is needed, while the (non-sharded) computation can still be overlapped by the dispatch / combine communications, achieving better performance.

ganyi1996ppo: Ok, sounds reasonable

ganyi1996ppo (approved):

vllm_ascend/models/deepseek_v2.py
```diff
@@ -167,6 +182,8 @@ def __init__(
         if config.hidden_act != "silu":
             raise ValueError(f"Unsupported activation: {config.hidden_act}. "
                              "Only silu is supported for now.")
+        self.enable_graph_mode = (get_current_vllm_config().additional_config
```
wangxiyuan: L238 has init `self.enable_graph_mode` already

sdmyzlp: Fixed, by removing the duplicated one.

vllm_ascend/ops/fused_moe.py
```diff
@@ -675,28 +695,36 @@ def apply(
...
+            )
         elif self.torchair_graph_enabled or get_ep_group().world_size == 1:
-            return fused_experts(hidden_states=x,
-                                 w1=layer.w13_weight,
-                                 w2=layer.w2_weight,
-                                 topk_weights=topk_weights,
-                                 topk_ids=topk_ids,
-                                 top_k=top_k,
-                                 expert_map=expert_map)
+            router_hidden_states = fused_experts(hidden_states=x,
```
ApsarasX: Why is it not supported in the `AscendW8A8DynamicFusedMoEMethod` class?

sdmyzlp: Fixed, combined with previously merged #947, both quantized and unquantized scenario are supported.

ApsarasX (approved):

PR #1003, opened 2025-05-29
remove old quantization model

22dimensions:
remove old quantization model, and new models will be added to testcase later.


Yikun:
@22dimensions Could we add new test first (`vllm-ascend/Qwen2.5-0.5B-Instruct-w8a8-new`) and then remove the old one.

22dimensions:
> @22dimensions Could we add new test first (`vllm-ascend/Qwen2.5-0.5B-Instruct-w8a8-new`) and then remove the old one.

The new quantization model's name is vllm-ascend/Qwen2.5-0.5B-Instruct-W8A8, but i can't create this model in modelscope, cause it is case insensitive, so i need to delete old model first. And the new PR will be merged toady.

Yikun (approved):
Could you estimate how long this test will be added? LGTM if CI passed

tests/singlecard/test_offline_inference.py
```diff
@@ -33,10 +34,14 @@
...
 MODELS = [
     "Qwen/Qwen2.5-0.5B-Instruct",
-    "vllm-ascend/Qwen2.5-0.5B-Instruct-w8a8",
     "Qwen/Qwen3-0.6B-Base",
 ]
 MULTIMODALITY_MODELS = ["Qwen/Qwen2.5-VL-3B-Instruct"]
+
+QUANTIZATION_MODELS = [
+    "vllm-ascend/Qwen2.5-0.5B-Instruct-W8A8-new",
+    "vllm-ascend/DeepSeek-V2-Lite-W8A8"
```
wangxiyuan: can DeepSeek-V2-Lite-W8A8 work with only one card?

22dimensions: It works on local machine with 32GB device memory, but it doesn't works on CI machine. So I remove it temporally. I will add this model later.

PR #1018, opened 2025-05-29
fix: ascend_scheduler adapt v0.9.0

zzzzwwjj:
fix ascend_scheduler with v0.9.0

If you don't want to open chunked-prefill and fallback to v0engine's scheduler, you can init LLM with param: `additional_config={'ascend_scheduler_config': {}}`.
> Now `ascend_scheduler` can't support with prefix_cache, so you need to add param: `enable_prefix_caching=False` to close it.
> `ascend_scheduler` can't support with PD Disaggreate and mtp.


wangxiyuan:
this pr can be closed now, it has been fixed here #839

PR #1036, opened 2025-05-30
mooncake connector v1

LCAIZJ:


david6666666:
Do we need to add code related to adapting vllm v0.9.0 v1 connector on vllm-ascend side, besides this connector file, for example in model_runner_v1.py? If you have done it, can you share the PR? That would be great.

david6666666:
@LCAIZJ

LCAIZJ:
I will submit a new PR

PR #1037, opened 2025-05-30
[V0.7.3][LoRA][Qwen3] Make v0.7.3 support Qwen3+LoRA

paulyu12:
Because the EOL vLLM-v0.7.3 lacks this PR(https://github.com/vllm-project/vllm/pull/13166), while launching Qwen3+LoRA on vllm-ascend0.7.3, the error **"Qwen3ForCausalLM" object has no attribute "embedding modules**" will be raised.

We modify qwen3.py to support Qwen3+LoRA on vllm-ascend v0.7.3 instead.

No.

### DIFF
--- a/vllm_ascend/models/qwen3.py
+++ b/vllm_ascend/models/qwen3.py
@@ -19,7 +19,7 @@
 # Adapted from vllm/model_executor/models/qwen3.py
 # This file is a part of the vllm-ascend project.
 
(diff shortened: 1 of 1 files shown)


wangxiyuan:
Thanks for the update! We can merge this first. While we don't have the detail plan for post2 release. Maybe it needs more time.

wangxiyuan (approved):

PR #1040, opened 2025-05-31
[WIP] accuracy report

Yikun:

PR #1041, opened 2025-06-01
Skip benchmarks/** in vllm ascend test

Yikun:
Skip benchmarks/** in vllm ascend test to reduce CI cost

No

CI passed


Yikun:
To reduce unused CI triggered, let's merge this first if https://github.com/vllm-project/vllm-ascend/pull/1040 has more refactor work.

PR #1054, opened 2025-06-03
kvcache nz

chenwaner:
Enable kvcache_nz for the decode process in the graph mode, which reduces the time consumed by FA in long sequences.
If need to enable kvcache_nz, should set the environment variable VLLM_ENABLE_KV_NZ: "1"
Tested in deepseek model, with batchsize 64 and seq_len 1k+3k:  61 layers FA total time 20.80ms -> 19.76ms


realliujiaxu:
torch_npu.npu_fused_infer_attention_score only support K V with format of ND(https://www.hiascend.com/document/detail/zh/Pytorch/700/apiref/apilist/ptaoplist_001232.html). Does this PR require newer vision of torch_npu and CANN?

PR #1071, opened 2025-06-05
[CI] Add workflow_dispatch and use main benchmarks directly

Potabk:

This is for the benchmark iteration, which will change the benchmark scripts while checkouting each commit. So we need ensure the benchmark scripts always available.


Yikun:
Add
```
on:
  workflow_dispatch:
```

Yikun (approved):

Yikun (approved):
LGTM

PR #1085, opened 2025-06-05
feat: support compile multiple batch graph

NeverRaR:

support compile multiple batch graph with different code object to avoid cache invalidation


NeverRaR:
@ganyi1996ppo @wangxiyuan Please take a look. Thanks.

wangxiyuan (approved):

PR #1092, opened 2025-06-05
[BugFix] Fix ascend config check

wangxiyuan:
Fix the ascend config check logic:
1. refactor check_ascend_config to make it clear:
    1. torchair graph should not work with enforce_eager=True
    2. aclgraph should not work with torchair graph
3. add refresh config for rlhf case
4. fix a typo in model runner
5. change expert_tensor_parallel_size default to 0 to keep the same as before

### DIFF
--- a/docs/source/user_guide/additional_config.md
+++ b/docs/source/user_guide/additional_config.md
@@ -28,7 +28,8 @@ The following table lists the additional configuration options available in vLLM
 | ---- | ---- | ------- | ----------- |
 | `torchair_graph_config` | dict | `{}` | The config options for torchair graph mode |
 | `ascend_scheduler_config` | dict | `{}` | The config options for ascend scheduler  |
-| `expert_tensor_parallel_size` | str | `1` | Expert tensor parallel size the model to use. |
+| `expert_tensor_parallel_size` | str | `0` | Expert tensor parallel size the model to use. |
+| `refresh` | bool | `false` | Whether to refresh global ascend config content. This value is usually used by rlhf case. |
 
 The details of each config option are as follows:
 
@@ -40,6 +41,7 @@ The details of each config option are as follows:
 | `use_cached_graph` | bool | `False` | Whether to use cached graph |
 | `graph_batch_sizes` | list[int] | `[]` | The batch size for torchair graph cache |
 | `graph_batch_sizes_init` | bool | `False` | Init graph batch size dynamically if `graph_batch_sizes` is empty |
+| `enable_multistream_shared_expert`| bool | `False` | Whether to enable multistream shared expert |
 
 **ascend_scheduler_config**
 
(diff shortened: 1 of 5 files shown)


Yikun (commented):
LGTM if CI passed

vllm_ascend/ascend_config.py
```diff
@@ -126,7 +129,7 @@ def check_ascend_config(vllm_config, enforce_eager):
                     "Torchair graph mode only works with deepseek model.")
 
     # for V1 Engine, aclgraph doesn't work with deepseek model and only qwen model is well tested.
-    if envs.VLLM_USE_V1 and vllm_config.model_config is not None and not enforce_eager:
+    if envs.VLLM_USE_V1 and vllm_config.model_config is not None and not enforce_eager and not ascend_config.torchair_graph_config.enabled:
```
NeverRaR: this check should be moved to platform.py:check_and_update_config

vllm_ascend/ascend_config.py
```diff
@@ -126,7 +129,7 @@ def check_ascend_config(vllm_config, enforce_eager):
                     "Torchair graph mode only works with deepseek model.")
```
NeverRaR: should be:
```
if ascend_config.torchair_graph_config.enabled:
        if envs.VLLM_MLA_DISABLE:
            logger.warning(
                "Torchair graph mode is still experimental and not supported for V1 without mla currently, "
                "it has been disabled automatically.")
            ascend_config.torchair_graph_config.enabled = False
        elif vllm_config.model_config:
            model_type = vllm_config.model_config.hf_config.model_type
            if "deepseek" not in model_type:
                raise NotImplementedError(
                    "Torchair graph mode only works with deepseek model.")
```
 ascend_config. ascend_scheduler_config.enabled = False -> ascend_config.torchair_graph_config.enabled = False
if vllm_config.model_config: -> elif vllm_config.model_config:

NeverRaR (changes requested):

wangxiyuan:
@NeverRaR Thanks for your review. I've updated the check logic to make it more clear. Please take a look at again.

docs/source/user_guide/additional_config.md
```diff
@@ -29,6 +29,7 @@ The following table lists the additional configuration options available in vLLM
 | `torchair_graph_config` | dict | `{}` | The config options for torchair graph mode |
 | `ascend_scheduler_config` | dict | `{}` | The config options for ascend scheduler  |
 | `expert_tensor_parallel_size` | str | `1` | Expert tensor parallel size the model to use. |
+| `refresh` | bool | `false` | Whether to refresh global ascend config content. This value is usually used by rlhf case. |
```
Yikun, line 32: QQ: Does vLLM has similar config?

NeverRaR (approved):

docs/source/user_guide/additional_config.md
wangxiyuan: Not, this value is only used for rlhf, for verl or someother framework, the case is:
1. verl load and update vllm config
2. verl start LLM with additional config in external_executor mode

in the fisrt step, the ascend config has been initialized, then in the second step, the additional config will be skipped.

To solve the problem, we should let verl pass `refresh`, the we can regenerate the config

PR #1097, opened 2025-06-06
[Misc] Adjust the default profiler configuration

ApsarasX:
When profiling, it is often necessary to disable the call stack to reduce profiling overhead, and adjust the profiler_level to level1 to obtain more detailed operator and communication information.

Therefore, it is recommended to modify the default profiling configuration.

No

No


ApsarasX:
> I'm OK with this, should we also change the configure for V0?

Yes, I have added the modifications on worker V0

wangxiyuan (approved):

wangxiyuan:
make it configuable if we hit other requirement in the future.

PR #1099, opened 2025-06-06
[CI][Benchmark] Add new model and v1 test to perf benchmarks

Potabk:

- Add qwen2.5-7b-instruct test
- Add v1 test

### DIFF
--- a/.github/workflows/nightly_benchmarks.yaml
+++ b/.github/workflows/nightly_benchmarks.yaml
@@ -193,7 +193,7 @@ jobs:
             --commit_title "$commit_title" \
             --created_at "$commit_time_no_tz" \
             --res_dir ./benchmarks/results \
-            --extra_feat '{"VLLM_USE_V1": "'${{ matrix.vllm_use_v1 }}'"}'
+            --extra_feat '{"VLLM_USE_V1": "${{ matrix.vllm_use_v1 }}"}'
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly_benchmarks.yaml
```diff
@@ -177,17 +183,17 @@ jobs:
...
-              --res_dir ./benchmarks/results 
-              rm -rf ./benchmarks/results
-            fi
+            escli add --vllm_branch ${{ matrix.vllm_branch }} \
+            --vllm_ascend_branch ${{ matrix.vllm_ascend_branch }} \
+            --commit_id $commit_id \
+            --commit_title "$commit_title" \
+            --created_at "$commit_time_no_tz" \
+            --res_dir ./benchmarks/results \
+            --extra_feat '{"VLLM_USE_V1": "${{ matrix.vllm_use_v1 }}"}'
```
Yikun, line 196: does this a separate col in es?

Is convenient to filter with it in frontend?

Potabk: yes, this is a series of key-value pairs for more new expansions feat in the feature, for example, `VLLM_USE_V1` can seen as using a feature, can convenient to be filter

Yikun (approved):

PR #1103, opened 2025-06-06
[CI/UT][Graph] Add ut for torchair graph mode

MengqingCao:
Add ut for torchair graph mode on DeepSeekV3

### DIFF
--- a/docs/source/user_guide/additional_config.md
+++ b/docs/source/user_guide/additional_config.md
@@ -53,7 +53,7 @@ The details of each config option are as follows:
 | ---- | ---- | ------- | ----------- |
 | `enabled` | bool | `False` | Whether to enable ascend scheduler for V1 engine|
 
-ascend_scheduler_config also support the options from [vllm scheduler config](https://docs.vllm.ai/en/stable/api/vllm/config.html#vllm.config.SchedulerConfig). For example, you  can add `chunked_prefill_enabled: true` to ascend_scheduler_config as well.
+ascend_scheduler_config also support the options from [vllm scheduler config](https://docs.vllm.ai/en/stable/api/vllm/config.html#vllm.config.SchedulerConfig). For example, you  can add `enable_chunked_prefill: true` to ascend_scheduler_config as well.
 
 ### Example
 
@@ -62,18 +62,18 @@ A full example of additional configuration is as follows:
 ```
 {
     "torchair_graph_config": {
-        "enabled": true,
-        "use_cached_graph": true,
+        "enabled": True,
+        "use_cached_graph": True,
         "graph_batch_sizes": [1, 2, 4, 8],
-        "graph_batch_sizes_init": false,
-        "enable_multistream_moe": false,
-        "enable_kv_nz": false
+        "graph_batch_sizes_init": False,
+        "enable_multistream_moe": False,
+        "enable_kv_nz": False
     },
     "ascend_scheduler_config": {
-        "enabled": true,
-        "chunked_prefill_enabled": true,
+        "enabled": True,
+        "enable_chunked_prefill": True,
     },
     "expert_tensor_parallel_size": 1,
-    "refresh": false,
+    "refresh": False,
 }
 ```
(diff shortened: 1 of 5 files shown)


vllm_ascend/ops/attention.py
```diff
@@ -154,20 +154,22 @@ def vanilla_chunked_prefill_mla(
     batch_size = block_tables.size(0)
     assert query_lens.size(0) == batch_size
     num_heads = query.size(1)
-    block_size = kv_cache.size(1)
-    latent_kv_dim = kv_cache.size(3) - rope_dim
+    block_size = kv_cache[0].size(1)
```
ganyi1996ppo: Torchair will get into this code path?

wangxiyuan: yes, with chunked prefill enabled.

but this is another bug. we should fix it later. @MengqingCao will add the test with ascend scheudler enabled.

MengqingCao: I have removed this fix now because there still exists oom issue on chunked prefill senario.
This pr will focus ut on ascend scheudler with chunked prefill disabled. cc @Yikun

MengqingCao: This pr is ready for review now. plz take a look, thx!

docs/source/user_guide/additional_config.md
```diff
@@ -53,7 +53,7 @@ The details of each config option are as follows:
 | ---- | ---- | ------- | ----------- |
 | `enabled` | bool | `False` | Whether to enable ascend scheduler for V1 engine|
 
-ascend_scheduler_config also support the options from [vllm scheduler config](https://docs.vllm.ai/en/stable/api/vllm/config.html#vllm.config.SchedulerConfig). For example, you  can add `chunked_prefill_enabled: true` to ascend_scheduler_config as well.
+ascend_scheduler_config also support the options from [vllm scheduler config](https://docs.vllm.ai/en/stable/api/vllm/config.html#vllm.config.SchedulerConfig). For example, you  can add `enable_chunked_prefill: true` to ascend_scheduler_config as well.
```
Yikun:
```suggestion
ascend_scheduler_config also support the options from [vllm scheduler config](https://docs.vllm.ai/en/stable/api/vllm/config.html#vllm.config.SchedulerConfig). For example, you  can add `enable_chunked_prefill: True` to ascend_scheduler_config as well.
```

tests/multicard/test_torchair_graph_mode.py
```diff
@@ -0,0 +1,75 @@
...
+                enforce_eager=False,
+        ) as vllm_model:
+            vllm_output = vllm_model.generate_greedy(example_prompts,
+                                                     max_tokens)
+        golden_results = [
+            'Hello, my name is feasibility伸 spazio debtor添',
+            'The president of the United States is begg"""\n杭州风和 bestimm',
+            'The capital of France is frequentlyশามalinkAllowed',
+            'The future of AI is deleting俯احت怎么样了حراف',
+        ]
```
Yikun, line 75: nits: Please add a note that the 

Quick question: is this deterministic?

docs/source/user_guide/additional_config.md
MengqingCao: done

tests/multicard/test_torchair_graph_mode.py
MengqingCao: Yes, we use greedy sampler to make sure the generated results are fix.

wangxiyuan (approved):

Yikun:
Let's merge this to protect DeepSeek graph mode on V1 (with `ascend_scheduler` enable) via `vllm-ascend/DeepSeek-V3-Pruning` e2e.

tests/multicard/test_torchair_graph_mode.py
```diff
@@ -0,0 +1,80 @@
...
+                tensor_parallel_size=4,
+                distributed_executor_backend="mp",
+                additional_config={
+                    "torchair_graph_config": {
+                        "enabled": True,
+                    },
+                    "ascend_scheduler_config": {
+                        "enabled": True,
+                    },
+                    "refresh": True,
```
Yikun, line 59: last question: why need refresh here?

Yikun: I merged this first feel free to open new PR to address this if needed

MengqingCao: `refresh` is used here because an initial setup of llm is already done in UTs, it need a refresh of additional config. And we don't need this in a normal usage

PR #1104, opened 2025-06-06
[CI][Benchmark] Add qwen2.5-7b test

Potabk:

- Add qwen2.5-7b performance benchmark, this is a sub pr of #1099, for v1 test, need more verify
- Fix get commit time after checkout


Potabk:
@Yikun This can merge first

PR #1107, opened 2025-06-06
[perf] optimize apply_penalties & topKtopP for V0&V1 Engine

linfeng-yuan:
Same as pull/525, this PR optimizes apply_penalties & topKtopP implementation in both V0/V1 Engine by avoiding using torch.scatter and matrix indexing operations.

No. 

This patch was tested with vllm v0.9.0, torch-2.5.1 & torch_npu-2.5.1 (both torch_npu in PyPI and newest internal beta version). At a concurrency of 58 and with post-processing parameters set to "temperature": 0.2, "top_k": 1000, "top_p": 0.92, the average sampling time in each decoding stage was reduced from 90ms to 8ms.

### DIFF
--- a/vllm_ascend/sample/ops/ascend_topk_topp_sampler.py
+++ b/vllm_ascend/sample/ops/ascend_topk_topp_sampler.py
@@ -1,0 +1,64 @@
+from typing import Dict, Optional
+
+import torch
+from vllm.v1.sample.ops.topk_topp_sampler import TopKTopPSampler, random_sample
+
+
+class AscendTopKTopPSampler(TopKTopPSampler):
+
+    def forward_native(
+        self,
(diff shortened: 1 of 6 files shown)


vllm_ascend/sample/ops/ascend_topk_topp_sampler.py
```diff
@@ -0,0 +1,64 @@
+from typing import Dict, Optional
+
+import torch
+from vllm.v1.sample.ops.topk_topp_sampler import TopKTopPSampler, random_sample
+
+
+class AscendTopKTopPSampler(TopKTopPSampler):
```
MengqingCao, line 7: I think this is duplicated with https://github.com/vllm-project/vllm-ascend/pull/970

PR #1117, opened 2025-06-07
vllm-ascend support chunked prefill

fems14:
vllm-ascend support chunked prefill for MLA


ttanzhiqiang:
How much profit is there before and after the modification?

Issue #734, opened 2025-04-30
[Guide]: Usage on Speculative Decoding and MTP


MengqingCao:
### How to use Speculative Decoding and MTP on vLLM Ascend

Please refer to [vLLM official doc on Speculative Decoding](https://docs.vllm.ai/en/latest/features/spec_decode.html) as a usage guide.

> [!NOTE]  
> When using Speculative Decoding and MTP, there are some limits on vllm-ascend compared with vllm:
> 
> 1. When request preemption is triggered, there exsists precision issue with Speculative Decoding, except for MTP.
> 2. Speculative Decoding with multi-step preparation on npu is not supported, only support replacing by circle `for` on cpu to simulate multi-step preparation.
> 3. Only BatchExpansionTop1Scorer is supported now, MQAScorer is not supported.

shen-shanshan:
Set `VLLM_ASCEND_ACL_OP_INIT_MODE` to `1` when using spec decode in V0 engine.

```bash
export VLLM_ASCEND_ACL_OP_INIT_MODE=1
```

PR #1123, opened 2025-06-09
[Bugfix] Set `ACL_OP_INIT_MODE` env var default to `0`

shen-shanshan:

Set `ACL_OP_INIT_MODE` env var default to `0`, since vllm-ascend may have problems in some scenarios when setting it to `1`.

Plus, the guide https://github.com/vllm-project/vllm-ascend/issues/734 has also been updated.

### DIFF
--- a/vllm_ascend/envs.py
+++ b/vllm_ascend/envs.py
@@ -127,7 +127,7 @@ env_variables: Dict[str, Callable[[], Any]] = {
     # We set this var default to `1` in vllm-ascend to avoid segment fault when
     # enable `pin_memory` while creating a tensor using `torch.tensor`.
     "VLLM_ASCEND_ACL_OP_INIT_MODE":
-    lambda: os.getenv("VLLM_ASCEND_ACL_OP_INIT_MODE", '1'),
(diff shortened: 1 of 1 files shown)


wangxiyuan (commented):
background: set this value to 1 will lead rlhf case error.

I guess it's becasue that in rlhf case, it calls torch-npu before vllm start. in this case, lazy_load may lead problem.

mengwei805:
set `VLLM_ASCEND_ACL_OP_INIT_MODE`(`ACL_OP_INIT_MODE`)  to 1 can fix v0 spec `segment fault` bug.
if set default to 0 to aviod rlhf case failed
it should be set to 1 in `platform.py` in v0 spec branch.
The user has no reason to be aware of `VLLM_ASCEND_ACL_OP_INIT_MODE`(`ACL_OP_INIT_MODE`).
We should be cautious about opening up environmental variables.
@shen-shanshan @wangxiyuan @Yikun 

BTW
Q: why CI passed when `VLLM_ASCEND_ACL_OP_INIT_MODE`(`ACL_OP_INIT_MODE`) to 0?
A: UT's architecture is different from full offline inference, so whether ACL_OP_INIT_MODE is set to 0 or 1 will not affect UT if no other bugs exist.

PR #1101, opened 2025-06-06
etp best a2

ttanzhiqiang:
### What this PR does / why we need it?

Single machine 16 cards deepseekr1 attention (tp8/dp2) / moe(etp) Best performance
截屏2025-06-06 11 32 37
vllm-ascend commit id:da9acfca6053352730fce75fb772e214755d0341
vllm commit id:b124e1085b1bf977e3dac96d99ffd9d8ddfdb6cc
+ https://github.com/vllm-project/vllm-ascend/pull/910 + [Reduce _npu_flash_attention mask to 128x128 for memory savings] https://github.com/vllm-project/vllm-ascend/pull/1100+ [Reduce memory usage by splitting tokens in fused_experts]

### How was this patch tested?

bash run_dp_attention_etp16.sh
bash run_dp_attention_etp16_benmark.sh &> output_etp16.log

PR #1098, opened 2025-06-06
Enable kvcache_nz for the decode process in torchair graph mode

chenwaner:
What this PR does / why we need it?
Enable kvcache_nz for the decode process in torchair graph mode, which reduces the time consumed by FA in long sequences.

Does this PR introduce any user-facing change?
If need to enable kvcache_nz, should set the additional_config.torchair_graph_config.enable_kv_nz=True

How was this patch tested?
1. Tested in deepseek model:
with batchsize 64 and seq_len 1k+3k, 61 layers FA total time improves 20.80ms -> 19.76ms
2. operator precision test: 
[aclnnFusedInferAttentionScoreV3_result.csv](https://github.com/user-attachments/files/20664138/aclnnFusedInferAttentionScoreV3_result.csv)
3. tpot test from @ttanzhiqiang, and curl one result is normal
https://github.com/vllm-project/vllm-ascend/pull/1098#issuecomment-2948542159
https://github.com/vllm-project/vllm-ascend/pull/1098#issuecomment-2954496588

PR #1131, opened 2025-06-09
Waiting for BMM NZ support(Improve TPOP 2ms performance) 

ttanzhiqiang:

W_UV/W_UK_T cannot be converted to nz, because this position will be fused into transposebatchmatmul, which does not support nz. The weights are actually converted back to nd in each run.

Use https://github.com/vllm-project/vllm-ascend/pull/1098 as the baseline
image
p90 TPOT 90.79ms->88.58ms, Improve TPOP 2ms

use https://github.com/vllm-project/vllm-ascend/pull/1101


ttanzhiqiang:
@wangxiyuan @Yikun

jianzs (approved):

PR #1132, opened 2025-06-09
[CI] Run e2e after pre check pass

wangxiyuan:
Make sure the lint test passed before start the e2e test to save compute resource.

Updated the patch doc to make sure the CI works as expect.


.github/workflows/vllm_ascend_test.yaml
```diff
@@ -15,7 +15,7 @@
 # This file is a part of the vllm-ascend project.
 #
 
-name: 'e2e test'
+name: 'Lint and Functional Test'
```
Yikun: This is too long many useful info are hide.

`basic test`

.github/workflows/vllm_ascend_test.yaml
```diff
@@ -39,7 +42,107 @@ defaults:
     shell: bash -el {0}
 
 jobs:
```
Yikun, line 44: Could we squash yapf, codespell, ruff into one?

wangxiyuan: I think leave mypy test to one is fine as well.

wangxiyuan:
@Yikun please take a look again. Merged the 4 lint check into one. https://github.com/vllm-project/vllm-ascend/actions/runs/15552048318?pr=1132

.github/workflows/vllm_ascend_test.yaml
```diff
@@ -15,7 +15,7 @@
 # This file is a part of the vllm-ascend project.
 #
 
-name: 'e2e test'
+name: 'Basic Test'
```
Yikun:
```suggestion
name: 'test'
```

.github/workflows/vllm_ascend_test.yaml
```diff
@@ -39,7 +42,65 @@ defaults:
     shell: bash -el {0}
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
```
Yikun, line 46:
```suggestion
    runs-on: ubuntu-latest
    name: lint
```

.github/workflows/vllm_ascend_test.yaml
```diff
@@ -39,7 +42,65 @@ defaults:
...
+        working-directory: vllm-empty
+        run: |
+          pip install -r requirements/build.txt --extra-index-url https://download.pytorch.org/whl/cpu
+          VLLM_TARGET_DEVICE=empty pip install .
+
+      - name: Mypy Check
+        run: |
+          echo "::add-matcher::.github/workflows/matchers/mypy.json"
+          tools/mypy.sh 1 ${{ matrix.python-version }}
+
```
Yikun, line 100: Any reason why only actionlint left?

https://github.com/vllm-project/vllm-ascend/blob/main/.github/workflows/actionlint.yml

Yikun (approved):
Looks good except comments inline.

BTW, I think we can do some thing `pre-commit` in future.
https://github.com/vllm-project/vllm/blob/main/.github/workflows/pre-commit.yml

.github/workflows/vllm_ascend_test.yaml
wangxiyuan: let's merge others in the next PR

PR #993, opened 2025-05-28
MLA layer eliminates redundant index operators

huiyingCCCC:
### What this PR does / why we need it?

During the autoregressive decoding process, the cos and sin values are exactly the same for each layer(such as 61 layers). Therefore, they only need to be calculated in the first layer, and subsequent layers can directly reuse them.

### Does this PR introduce _any_ user-facing change?
None

### How was this patch tested?
pytest tests/multicard/test_offline_inference_distributed.py

PR #1135, opened 2025-06-09
Support multistream of MLA vector operations

sdmyzlp:
Move all vector operations to a secondary stream, with the expected overlaping being:
```
              | q_rmsnorm |                  | kv_norm_rope_cache |       | q_rope |
| matmul W_DQ | matmul W_DKV | index | index |    matmul W_UQ     | split | matmul W_KV_T |
```

Currently, the `IndexByTensor` operators introduced by computation of `cos` and `sin` can't be offloaded to the secondary stream due to a known bug of graph fusion optimization pass. So we instead keep it in the main stream, only requires it be computed before `matmul W_UQ` to avoid hindering later overlapping. The problem may be solved by later optimization (#993), which hoists the computation of `cos` and `sin` up to the first layer.

Controlled by `torchair_graph_config.enable_multistream_mla`, defaulted to False.

Tested on 1x16 910 node, with tailored 2 layer DSKv2.

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -19,6 +19,7 @@ from vllm_ascend.multistream.base import MSAttentionMetadataSplitConfig
 from vllm_ascend.multistream.context import get_multistream_comm_context
 from vllm_ascend.multistream.ms_split import model_input_split_v1_mla_attn
 from vllm_ascend.ops.attention import vanilla_chunked_prefill_mla
+from vllm_ascend.utils import npu_stream_switch, npu_wait_tensor
 
 if TYPE_CHECKING:
     from vllm.v1.core.sched.output import SchedulerOutput
@@ -481,6 +482,9 @@ class AscendMLAImpl(MLAAttentionImpl):
         ascend_config = get_ascend_config()
         self.torchair_graph_enabled = ascend_config.torchair_graph_config.enabled
         self.enable_kv_nz = ascend_config.torchair_graph_config.enable_kv_nz
+        self.enable_multistream_mla = \
+            ascend_config.torchair_graph_config.enable_multistream_mla
+
         # Adapt torch air graph mode with spec decoding.
         speculative_config = get_current_vllm_config().speculative_config
         if speculative_config is not None:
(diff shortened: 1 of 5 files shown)


wangxiyuan:
there is a pr for refactor. I suggest to block this PR until that one is merged https://github.com/vllm-project/vllm-ascend/pull/1169

vllm_ascend/attention/mla_v1.py
```diff
@@ -867,23 +874,38 @@ def forward(
...
+                npu_wait_tensor(decode_hs_or_q_c,
+                                sin,
+                                enabled=self.enable_multistream_mla)
+            decode_ql_nope, decode_q_pe = \
+                self._q_proj_and_k_up_proj(decode_hs_or_q_c)
+            if self.running_in_graph:
                 decode_k_pe, decode_k_nope = self.exec_kv(
                     hidden_states_or_kv_c_normed, cos, sin, kv_cache,
                     attn_metadata.slot_mapping)
+                with npu_stream_switch("mla_secondary",
```
ganyi1996ppo, line 902: So, this `npu_stream_switch` can be used inside the torchair right? if that's so, can we further optimize the dbo path of deepseek to gain more performance boost.

vllm_ascend/models/deepseek_v2.py
```diff
@@ -505,7 +508,14 @@ def forward(
             attn_metadata: Optional[AttentionMetadata] = None) -> torch.Tensor:
         if self.q_lora_rank is not None:
             ckq = self.q_a_proj(hidden_states)[0]
-            hidden_states_or_q_c = self.q_a_layernorm(ckq)
+            use_multistream_mla = (self.enable_multistream_mla
+                                   and attn_metadata is not None
+                                   and attn_metadata.num_decodes > 0)
```
ganyi1996ppo, line 513: So, multistream mla will not be triggered if there are only prefill  requests?

sdmyzlp: Yes, current multistream utilities only support graph mode, i.e. `npu_stream_switch` returns `contextlib.nullcontext()` on non-graph mode, which is the case for prefill.

vllm_ascend/attention/mla_v1.py
sdmyzlp: Discussed with torchair colleagues, the `with torch.npu.stream(...):` scheme is not supported by torchair graph mode while `npu_stream_switch` only works under graph mode, so I recommend leaving this to later PRs:
1. to provide a joint version of `npu_stream_switch` encapsulating both cases, as well as,
2. adding graph mode support for dbo

sdmyzlp:
> there is a pr for refactor. I suggest to block this PR until that one is merged #1169

@zzzzwwjj Discussed with xiyuan, this PR seems to not have much confliction with 1169, please have a check, thanks~

PR #1146, opened 2025-06-10
[CI][0.9.0][0.9.1] Update CI job

wangxiyuan:
1. drop main check for 0.9.1-dev branch
2. drop daily build for 0.9.1-dev branch

### DIFF
--- a/.github/workflows/nightly_benchmarks.yaml
+++ b/.github/workflows/nightly_benchmarks.yaml
@@ -18,11 +18,7 @@
 name: 'Benchmarks / Performance'
 # This workflow runs nightly benchmarks for vllm-ascend.
 
-on:
-  schedule:
-    # Run at 02:00 everyday
-    - cron: '00 18 * * *'
-  
+on:  
   workflow_dispatch:
     # Allow manual triggering of the workflow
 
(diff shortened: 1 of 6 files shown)


wangxiyuan:
merge this once vllm 0.9.1 released

.github/workflows/nightly_benchmarks.yaml
```diff
@@ -18,11 +18,7 @@
 name: 'Benchmarks / Performance'
 # This workflow runs nightly benchmarks for vllm-ascend.
 
-on:
-  schedule:
-    # Run at 02:00 everyday
-    - cron: '00 18 * * *'
-  
+on:
```
Yikun, line 21: Why remove this on main?

btw, the schedule job will also only runs on default branch.

[1] https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

.github/workflows/vllm_ascend_test.yaml
```diff
@@ -44,7 +42,7 @@ jobs:
       max-parallel: 2
       matrix:
         os: [linux-arm64-npu-1, linux-arm64-npu-4]
-        vllm_version: [main, v0.9.0]
+        vllm_version: [v0.9.0, v0.9.1]
```
Yikun, line 45: It's ok

Yikun (changes requested):
Do you want to submit PR on `v0.9.1-dev` ?

wangxiyuan:
yeah, wrong target branch

PR #1165, opened 2025-06-11
[CI] Upgrade vllm to 0.9.1

wangxiyuan:
upgrade vllm to 0.9.1.  0.9.0 is not supported for main branch now.

Keep doc to 0.9.0 until we release the first 0.9.1 release.


Yikun:
Accuracy test here: https://github.com/vllm-project/vllm-ascend/actions/runs/15576341627

Issue #1147, opened 2025-06-10
[RFC]: Refactoring AscendFusedMoE


zzzzwwjj:
### Motivation.

This RFC proposes optimize AscendFusedMoE‘s implemention, change branch's condition, simplfy code and make it easier to maintain and develop.

### Current problem
1. Chaotic branch conditions
Currently there are three implementation for AscendFusedMoE: all_gather, all2all and mc2, and different implementation need to be matched with different communication-op. Some communication operators still have usage restrictions.
2. Relative code locations are not together
Some codes of fused_moe are in deepseek_v2.py, some are in fused_moe.py & w8a8_dynamic.py.
3. The optimal performance scenarios for A2 and A3 are different, and it has not been considered yet.

### Proposed Change.

1. Integrate branch conditions
We will implement a global function to choose AscendFusedMoE’s implemention, based on ep_size, with_prefill and soc_version.
2. Move all codes of fused_moe to fused_moe.py

### CC List.

@Yikun @wangxiyuan @ganyi1996ppo @jianzs 

### Any Other Things.

Recently I will push some PRs to resolve this issue.

zzzzwwjj:
The current branch selection algorithm is as follows:
![Image](https://github.com/user-attachments/assets/0ee36016-24ac-442a-9924-192e73bc3988)

And I will add param `soc_version` to choose algorithm later, ensure that we can achieve optimal performance on both A2 and A3.

jianzs:
@zzzzwwjj  Thanks for your work.  I'm wondering if there's a problem using communication operators in graph mode.  Some operators might not be compatible.  For example, I'm not sure if an "all-to-all" operator works.

PR #1169, opened 2025-06-11
[refactor] Refactoring AscendFusedMoE

zzzzwwjj:
This PR is used for resolved [issue 1147](https://github.com/vllm-project/vllm-ascend/issues/1147)
1. Move fused_moe code into one file `fused_moe.py`.
2. Integrate branch conditions into function `get_fused_moe_state`.

1. This PR has removed the env `VLLM_ENABLE_MC2`, because I think this env is useless, we can make judgments based on the current scenario without this env, it will only increase complexity.
2. This PR has removed the env `USING_LCCL_COM`, because this env has already expired.
3. `additional_config.expert_tensor_parallel_size` has already expired, and now we also use parameter `enable_expert_parallel`, consistent with the vLLM.

### DIFF
--- a/docs/source/user_guide/graph_mode.md
+++ b/docs/source/user_guide/graph_mode.md
@@ -23,7 +23,7 @@ import os
 
 from vllm import LLM
 
-os.environ["VLLM_USE_V1"] = 1
+os.environ["VLLM_USE_V1"] = "1"
 
 model = LLM(model="Qwen/Qwen2-7B-Instruct")
 outputs = model.generate("Hello, how are you?")
@@ -45,7 +45,7 @@ offline example:
 import os
 from vllm import LLM
 
-os.environ["VLLM_USE_V1"] = 1
+os.environ["VLLM_USE_V1"] = "1"
 
 # TorchAirGraph is only work without chunked-prefill now
 model = LLM(model="deepseek-ai/DeepSeek-R1-0528", additional_config={"torchair_graph_config": {"enabled": True},"ascend_scheduler_config": {"enabled": True,}})
@@ -70,7 +70,7 @@ offline example:
 import os
 from vllm import LLM
 
-os.environ["VLLM_USE_V1"] = 1
+os.environ["VLLM_USE_V1"] = "1"
 
 model = LLM(model="someother_model_weight", enforce_eager=True)
 outputs = model.generate("Hello, how are you?")


wangxiyuan:
I really like this change. Let's merge this first. @Yikun @ganyi1996ppo @jianzs  please take this as the high priory. Thanks.

vllm_ascend/utils.py
```diff
@@ -247,3 +248,21 @@ def npu_wait_tensor(self: torch.Tensor,
                     *,
                     enabled: bool = True):
     return _npu_wait_tensor(self, dependency) if enabled else self
+
+
+# TODO(zzzzwwjj): move this into forward_context
```
wangxiyuan: We can add a sub module names `fused_moe` in ops. Then move FusedMoEState and get_fused_moe_state  to that module's utils file.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1894,15 +1890,9 @@ def _generate_mtp_token_ids(
...
-        batch_size_step = 8
-        largest_batch_size = 1
 
-        if envs_ascend.VLLM_ENABLE_MC2:
-            batch_size_step = max(batch_size_step, tp_size)
-            largest_batch_size = batch_size_step
-        while (largest_batch_size < 8):
-            self.torchair_graph_batch_sizes.append(largest_batch_size)
-            largest_batch_size *= 2
+        batch_size_step = max(8, tp_size)
```
wangxiyuan: make `8` constant

wangxiyuan:
> 3. additional_config.expert_tensor_parallel_size has already expired, and now we also use parameter enable_expert_parallel, consistent with the vLLM.

Any change in this PR related to this commit message?

wangxiyuan:
`VLLM_ENABLE_MC2 ` in example files can be removed as well.

jianzs:
It's better to describe which communication kernel is chosen for different configurations.

wangxiyuan:
@jianzs it's described clear in RFC. @zzzzwwjj maybe you can add more note in the code logic

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -897,11 +894,11 @@ def _process_reqs(
         if self.dp_size > 1:
             max_num_tokens, with_prefill = self._get_forward_metadata_across_dp(
                 total_num_scheduled_tokens, with_prefill)
+            extra_builder_kwargs['max_num_tokens_across_dp'] = max_num_tokens
             extra_builder_kwargs['with_prefill_across_dp'] = with_prefill
 
         # Add graph_pad_size here
-        if envs_ascend.VLLM_ENABLE_MC2 or (self.torchair_graph_enabled
```
whx-sjtu: The current usage of MC2 kernel does not support non-uniform inputs, so padding is still required.

wangxiyuan:
please rebase to main to make sure the torchair CI passed

PR #1184, opened 2025-06-12
[fix] fix bug in 1p1d disaggregated_prefill example

wangyanhui-cmss:
fix  bug in 1p1d  disaggregated_prefill  example

No.

Tested with python find_device_ips.py and run disaggregated_prefill example


wangxiyuan:
Thanks

PR #1186, opened 2025-06-12
Fix static EPLB log2phy condition and improve unit test

songshanhu07:

1.add static EPLB unit test
2.fix bug: Tensor cannot be directly judged by if statements

Run the unit test.

### DIFF
--- a/tests/ut/ops/test_expert_load_balancer.py
+++ b/tests/ut/ops/test_expert_load_balancer.py
@@ -101,7 +101,7 @@ def test_generate_log2phy_expert_map(mock_expert_load_balancer):
 
 def test_get_rank_placement_map(mock_expert_load_balancer, mocker):
     mocker.patch("torch_npu.npu._lazy_init")
-    mocker.patch('torch.npu.current_device', return_value=0)
+    mocker.patch('torch.npu.current_device', return_value='cpu')
     layer_id = 0
     rank_id = 0
     rank_local_expert_num, rank_expert_map = mock_expert_load_balancer.get_rank_placement_map(


jianzs:
Please add e2e tests to ensure the feature continues to function correctly during future changes.

MengqingCao:
Overall ltgm, plz move it to proper path under `tests/ut/`

docs/requirements-test.txt
```diff
@@ -1,2 +1,2 @@
 pytest-asyncio
-
+pytest-mock
```
MengqingCao, line 2:
```suggestion
pytest >= 6.0
pytest-mock
```

requirements-dev.txt
```diff
@@ -4,6 +4,7 @@ modelscope
 openai
 pytest >= 6.0
 pytest-asyncio
+pytest-mock
```
MengqingCao, line 7: Let's add `-r requirements-test.txt` in line3 and move this to `requirements-test.txt` instead

MengqingCao: we could do this later in your e2e test pr

ganyi1996ppo (approved):

tests/ut/ops/test_expert_load_balancer.py
```diff
@@ -0,0 +1,146 @@
+# fused moe ops test will hit the infer_schema error, we need add the patch
```
wangxiyuan, line 1: for ut, I don't think use `assert` is not a good way. Please use `unittest` lib instead.

wangxiyuan: And ut should have the same code tree with vllm-ascend. this file should be moved to the right place

PR #1190, opened 2025-06-12
[Feature] Dynamic EPLB Zero-like-overhead Expert Movement

raindaywhu:


vllm_ascend/models/deepseek_v2.py
```diff
@@ -723,6 +723,33 @@ def forward(
...
+    def get_all_moe_loads(self, num_moe_layers) -> torch.Tensor:
+        """
+        output: [num_moe_layers, num_experts_per_layer]
+        """
+        all_loads = []
+        for layer_id in range(num_moe_layers):
+            load_tensor = self.get_moe_load(3+layer_id)  
+            all_loads.append(load_tensor)
+
+        return torch.stack(all_loads, dim=0)
```
jianzs, line 752: Store all layer moe loads in one large tensor.  Initialize individual moe load of each layer by extracting slices from this tensor.  Perform all-reduce directly on the large tensor, avoiding the need to stack the load distributions every time you need to aggregate them.

vllm_ascend/ops/fused_moe.py
```diff
@@ -1209,3 +1218,28 @@ def _forward_ms_fused_moe_comp(
...
+        return self.log2phy
+
+    def get_moe_load(self):
+        return self.moe_load
+
+    def calculate_moe_load(self):
+        if self.moe_load is None:
+            self.moe_load = torch.zeros(self.num_experts,
+                                        dtype=torch.int64,
+                                        device=self.topk_ids.device)
```
jianzs, line 1238: Don't initialize moe_load lazily.  Initialize all layer moe_loads after loading the model using slicing, within the worker.

vllm_ascend/eplb/eplb_updator.py
```diff
@@ -0,0 +1,150 @@
...
+            self.world_size = dist.get_world_size()
+        if dist.is_initialized():
+            device = local_load.device
+            if self._gather_buffer is None:
+                shape = (self.world_size, *local_load.shape)
+                self._gather_buffer = torch.empty(shape,
+                                                  dtype=local_load.dtype,
+                                                  device=device)
+
+            dist.all_gather_into_tensor(self._gather_buffer, local_load)
```
jianzs, line 132: Why use `all_gather` instead of `all_reduce`?

PR #1192, opened 2025-06-12
[Doc] Add Referer header for CANN package download url.

wonderful199082:
fix the CANN download url

no, do not have any user-facing change

run the **wget** command and cann package is rightly downloaded.


Yikun (approved):

Yikun:
According to feedback from wechat group, it works, let's just merge this!

@wonderful199082 Thanks!

PR #1196, opened 2025-06-12
[EPLB] support deepseek eplb strategy

ZhengWG:

This PR implements the DeepSeek Expert Parallel Load Balancing (EPLB) strategy to optimize expert distribution in vllm-ascend. The implementation:
- Adapts the expert-map format to work with vllm-ascend's architecture
- Provides DeepSeek-provided mechanism to balance expert workload across devices

This PR adds a new script that allows users to:
- Generate expert map configurations based on workload analysis
- Optimize expert distribution for their specific use case

To use this feature:
1. First collect expert heat information during model execution
2. Run the provided script to generate the expert map configuration
3. Apply the generated configuration to your vllm-ascend deployment

User example:

```bash
# expert_load_view.pt:  dumped expert heat info file
python3 examples/eplb/eplb_strategy.py --exp_name 'deepseek_demo' \
    --input_path expert_load_view.pt  --output_path examples/eplb/results/demo \
    --num_nodes 4
```

results:
截屏2025-06-12 21 17 13

### DIFF
--- a/examples/eplb/eplb_deepseek.py
+++ b/examples/eplb/eplb_deepseek.py
@@ -9,7 +9,8 @@ from typing import Tuple
 import torch
 
 
-def balanced_packing(weight: torch.Tensor, num_packs: int) -> Tuple[torch.Tensor, torch.Tensor]:
+def balanced_packing(weight: torch.Tensor,
+                     num_packs: int) -> Tuple[torch.Tensor, torch.Tensor]:
     """
     Pack n weighted objects to m packs, such that each bin contains exactly n/m objects and the weights of all packs
     are as balanced as possible.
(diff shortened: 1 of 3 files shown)


ZhengWG:
@Yikun @wangxiyuan  Can you help reivew it ~

wangxiyuan (dismissed):

requirements.txt
```diff
@@ -14,6 +14,7 @@ torch-npu==2.5.1
 torch>=2.5.1
 torchvision<0.21.0
 wheel
+matplotlib
```
wangxiyuan: no need to add this requirement. this is only used for example

examples/eplb/eplb_deepseek.py
```diff
@@ -0,0 +1,205 @@
...
+"""
+Expert parallelism load balancer (EPLB) for vLLM.
+The rearrangement algorithm is adapted from
+[DeepSeek EPLB](https://github.com/deepseek-ai/eplb).
+"""
+from typing import Tuple
+
+import torch
+
+
```
wangxiyuan, line 11: is there any way to just import `https://github.com/deepseek-ai/eplb` instead of copy the code?

examples/eplb/eplb_strategy.py
```diff
@@ -0,0 +1,183 @@
...
+def calculate_average(lst):
+    """calculate the average of a list"""
+    if not lst:
+        raise ValueError("list is empty")
+
+    total = 0.0
+    count = 0
+
+    for element in lst:
+        # 检查元素是否为数值类型
```
wangxiyuan: use English instead.

examples/eplb/eplb_deepseek.py
ZhengWG: Thanks for the review. As there's no PyPI package available for deepseek-ai/eplb, we're preserving the original code implementation, consistent with the practice in the sglang/vllm repository.

ZhengWG:
@wangxiyuan Can you help check the update?

wangxiyuan (approved):

examples/eplb/eplb_deepseek.py
wangxiyuan: Just check that vLLM copied the same code. I'm fine here. https://github.com/vllm-project/vllm/blob/eb0b2d2f08b622f4b93fb0a811a047ad987a46ca/vllm/distributed/eplb/rebalance_algo.py

PR #1197, opened 2025-06-12
[Fix] Revert "Add qwen2.5 vl multimodal feature for vllm-ascend v1 (#736)"

yiz-liu:
This reverts commit 20dedba5d1fc84b7ae8b49f9ce3e3649389e2193, which introduced a flawed modification to token-wise padding.

Alas, these changes have wrought havoc upon the vital workings of both ACL Graph and Torchair's domain, leaving their functions grievously impaired. Thus, we must, with firm resolve, turn back the clock on this unfortunate artifice, to restore order and rightful operation to our besieged codes.

@Yikun @wangxiyuan

### DIFF
--- a/tests/singlecard/test_offline_inference.py
+++ b/tests/singlecard/test_offline_inference.py
@@ -85,6 +85,8 @@ def test_quantization_models(model: str, max_tokens: int) -> None:
 
 
 @pytest.mark.parametrize("model", MULTIMODALITY_MODELS)
+@pytest.mark.skipif(os.getenv("VLLM_USE_V1") == "1",
+                    reason="qwen2.5_vl is not supported on v1")
 def test_multimodal(model, prompt_template, vllm_runner):
     image = ImageAsset("cherry_blossom") \
         .pil_image.convert("RGB")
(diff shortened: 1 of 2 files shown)


Yikun:
cc @ChenTaoyu-SJTU 

BTW, does this should also be reverted?
 https://github.com/vllm-project/vllm-ascend/commit/291c216898c989efb4b59a64dad818d2dd48a71c

ChenTaoyu-SJTU:
hello, could you please describe in detail which part of the code has problem?

Yikun:
@ChenTaoyu-SJTU Sorry for revert, thanks for quick reply, according to the discussion with @yiz-liu offline, after this PR the aclgraph and torchair cross multi cards (tp) have the problem (padding part), we don't want to break anything, so we plan to first recover feature and add e2e test: https://github.com/vllm-project/vllm-ascend/pull/1198

 But don't worry, after e2e test added, we can get VL support back soon.

wangxiyuan:
take https://github.com/vllm-project/vllm-ascend/pull/1198 instead. Let's focus on this PR

PR #1198, opened 2025-06-12
Revert break change

wangxiyuan:


wangxiyuan:
after some debug, torchair works well with ascend scheduler. The problem is for v1 scheduelr and chunked prefill. So it's no need to revert. Let's fix it by the new PR instead.

PR #1201, opened 2025-06-13
[CI] Add unit test framework

wangxiyuan:
This PR added the unit test framework to enable ut for vLLM Ascend.  Unit test runs on CPU machines. It'll be ran once lint check is passed the same as e2e test.

For unit test, this PR created a new folder called `ut` under `tests` module. All the test file in `ut` should keep the same with the code in `vllm-ascend`. The file name should be start with `test_` prefix. For example, in this PR. the `test_ascend_config.py` is added for `ascend_config.py` test.

A new fille `worker/test_worker_v1.py` is also added as the placeholder. This file should be the unit test for `vllm-ascend/worker/worker_v1.py`.

Additional, a new `fake_weight` folder is added, it contains the config.json from `facebook/opt-125m`, so that the test will not always visit huggingface.

TODO:
We should add all the unit test file one by one in the future.


wangxiyuan:
After the refactor, the test workflow will be like:

image

PR #1214, opened 2025-06-13
fix torchair execute issue on padding data, and mtp padding logic -- by#1160

henryxuxu0716:
…by#1160

- Fixes #
-->fix torchair execute issue on padding data, and mtp padding logic 

it has been tested and merged in main.


wangxiyuan:
Quick merge to 0.9.1-dev since this is just a cherry-pick from main

PR #1216, opened 2025-06-13
[Build] Speedup image build

Yikun:
1. Rename workflow name to show OS info
2. Speedup image build:
- PR: only arm64 build on openEuler arm64, only amd64 build on Ubuntu amd64
- Push/Tag: still keep origin logic use qemu on amd64

This PR actually drop the e2e image build per PR but I think it's fine consider it's stable enough, if we still meet some problem we can revert this PR

43-44mins ---> about 8-10 mins

No

CI passed

### DIFF
--- a/.github/workflows/image_openeuler.yml
+++ b/.github/workflows/image_openeuler.yml
@@ -1,4 +1,4 @@
-name: 'image'
+name: 'image / openEuler'
 # This is a docker build check and publish job:
 # 1. PR Triggered docker image build check
 #   - is for image build check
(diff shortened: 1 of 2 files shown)


Yikun:
@wangxiyuan Do you think it's a acceptable improvement? speed by droping e2e and build on arm64 / amd64 separately.

After this PR, time reduced then we can have time and chance to add a e2e image test per PR in future.

wangxiyuan:
after this PR:
1. with PR, CI will build arm64 image on openEuler and amd64 on Ubuntu without push
2. when the PR is merged. CI will build both arm64 and amd64 image on openEuler and Ubuntu

Some edge case is not covered, like build amd64 on openEuler. But I think it's fine. Let' merge this first. If we hit any problem later, we can revert it.

PR #1217, opened 2025-06-13
[CI] Make e2e test to be preemptible and simple

Yikun:
This PR make e2e test to be simple, even bring some repeat code between single card and multicard, but we will not struggle with across max-parallel, matrix and concurrency: 
1. This PR make e2e test to be preemptible and simple:
- lint ---> e2e (2 parallel) ---> e2e multi-card (1 parallel)
- Anytime you push another PR will cancel previous job, whatever the job is lint / e2e / multi-cards
2. Use Modelscope rather than hf-mirror
3. Resolve some error like `Canceling since a higher priority waiting request for pr-XXXX-limit-npu-4 exists`

No

CI passed
- lint ---> e2e (2 parallel) ---> e2e multi-card (1 parallel)
- e2e test will canceled by update patch

### DIFF
--- a/.github/workflows/vllm_ascend_test.yaml
+++ b/.github/workflows/vllm_ascend_test.yaml
@@ -118,8 +118,7 @@ jobs:
     needs: [lint]
     if: ${{ needs.lint.result == 'success' }}
     strategy:
-      # TODO(yikun): 1 ---> 2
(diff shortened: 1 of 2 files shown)


wangxiyuan:
Thanks for the improvement

PR #1219, opened 2025-06-14
support fused_moe_allgather_ep

lyj-jjj:
support fused_moe_allgather_ep

~~

~~

### DIFF
--- a/vllm_ascend/models/deepseek_v2.py
+++ b/vllm_ascend/models/deepseek_v2.py
@@ -67,7 +67,7 @@ from vllm.sequence import IntermediateTensors
 
 import vllm_ascend.envs as envs_ascend
 from vllm_ascend.ascend_config import get_ascend_config
-from vllm_ascend.distributed.parallel_state import get_ep_group
+from vllm_ascend.distributed.parallel_state import get_ep_group, get_etp_group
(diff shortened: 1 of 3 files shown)


vllm_ascend/models/deepseek_v2.py
```diff
@@ -317,10 +325,13 @@ def forward(
         use_separated_shared_experts = (self.shared_experts is not None
                                         and not self.enable_multistream_moe)
 
+        enable_alltoall_ep = self.ep_group.world_size > 1 and not self.fused_experts_allgather_ep_enabled
```
realliujiaxu, line 328: set enable_alltoall_ep to False when fused_experts_allgather_ep_enabled is True. Because torch_npu.npu_format_cast_(layer.w2_weight, 29) is not supported by torch_npu.npu_grouped_matmul in current release version of torch_npu.

code like:
```
enable_alltoall_ep = False
if not self.fused_experts_allgather_ep_enabled:
    if is_prefill:
        enable_alltoall_ep = self.ep_group.world_size > 1
    else:
            enable_alltoall_ep =  self.ep_group.world_size > 1 and (VLLM_ENABLE_MC2 or not self.torchair_graph_enabled)
```

Issue #1222, opened 2025-06-14
[Bug]: Load weight failed using EPLB when num_redundancy_expert > 0


ZhengWG:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
INFO 06-14 19:23:46 [__init__.py:39] Available plugins for group vllm.platform_plugins:
INFO 06-14 19:23:46 [__init__.py:41] - ascend -> vllm_ascend:register
INFO 06-14 19:23:46 [__init__.py:44] All plugins in this group will be loaded. Set `VLLM_PLUGINS` to control which plugins to load.
INFO 06-14 19:23:46 [__init__.py:235] Platform plugin ascend is activated
WARNING:root:Failed to import 'vllm_ascend.vllm_ascend_C': No module named 'vllm_ascend.vllm_ascend_C'. All custom ops will be disabled. 
WARNING 06-14 19:23:47 [_custom_ops.py:22] Failed to import from vllm._C with ModuleNotFoundError("No module named 'vllm._C'")
Collecting environment information...
PyTorch version: 2.5.1+cpu
Is debug build: False

OS: openEuler 24.03 (LTS) (x86_64)
GCC version: (GCC) 12.3.1 (openEuler 12.3.1-38.oe2403)
Clang version: Could not collect
CMake version: version 4.0.2
Libc version: glibc-2.38

Python version: 3.11.6 (main, Feb 19 2025, 17:40:52) [GCC 12.3.1 (openEuler 12.3.1-38.oe2403)] (64-bit runtime)
Python platform: Linux-5.10.112-100.alios7.x86_64-x86_64-with-glibc2.38

CPU:
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Address sizes:                   46 bits physical, 57 bits virtual
Byte Order:                      Little Endian
CPU(s):                          192
On-line CPU(s) list:             0-191
Vendor ID:                       GenuineIntel
BIOS Vendor ID:                  Intel(R) Corporation
Model name:                      Intel(R) Xeon(R) Platinum 8468
BIOS Model name:                 Intel(R) Xeon(R) Platinum 8468  CPU @ 2.1GHz
BIOS CPU family:                 179
CPU family:                      6
Model:                           143
Thread(s) per core:              2
Core(s) per socket:              48
Socket(s):                       2
Stepping:                        8
BogoMIPS:                        4200.00
[...]

Issue #122, opened 2025-02-20
DeepSeek-R1 on 0.7.1-dev  with `Torch not compiled with CUDA enabled`


ColdeZhang:
Running in docker build on 0.7.1-dev branch:

```
ctr container create  \
  --mount type=bind,src=/mnt/data2,dst=/models,options=rbind:rw \
  --net-host  \
  --device /dev/davinci_manager  \
  --device /dev/hisi_hdc  \
  --device /dev/devmm_svm  \
  --device=/dev/davinci0  \
  --device=/dev/davinci1  \
  --device=/dev/davinci2  \
  --device=/dev/davinci3  \
  --device=/dev/davinci4  \
  --device=/dev/davinci5  \
  --device=/dev/davinci6  \
  --device=/dev/davinci7  \
  --mount type=bind,src=/usr/local/dcmi,dst=/usr/local/dcmi,options=rbind:ro \
  --mount type=bind,src=/usr/local/bin/npu-smi,dst=/usr/local/bin/npu-smi,options=rbind:ro \
  --mount type=bind,src=/usr/local/Ascend/driver/lib64/,dst=/usr/local/Ascend/driver/lib64/,options=rbind:ro \
  --mount type=bind,src=/usr/local/Ascend/driver/version.info,dst=/usr/local/Ascend/driver/version.info,options=rbind:ro \
  --mount type=bind,src=/etc/ascend_install.info,dst=/etc/ascend_install.info,options=rbind:ro \
  docker.io/library/vllm-npu:0.7.1  \
  vllm-ds-dev
```

Serve command:

```
vllm serve /models/DeepSeek-R1/ --tensor_parallel_size 8 --max-model-len 26240 --enforce-eager --trust-remote-code
```

Errors:

```
ERROR 02-20 09:26:55 engine.py:387] 'model.layers.0.self_attn.q_a_proj.weight'
ERROR 02-20 09:26:55 engine.py:387] Traceback (most recent call last):
ERROR 02-20 09:26:55 engine.py:387]   File "/usr/local/python3.10/lib/python3.10/site-packages/vllm/engine/multiprocessing/engine.py", line 378, in run_mp_engine
ERROR 02-20 09:26:55 engine.py:387]     engine = MQLLMEngine.from_engine_args(engine_args=engine_args,
[...]

PR #1223, opened 2025-06-14
[EPLB]: Correct local expert number calculation with redundant experts

ZhengWG:

- Fixed issue: [#122](https://github.com/vllm-project/vllm-ascend/issues/1222)  , support for expert mapping with redundant experts

No

Test passed when passed a expert_map with redundant_experts == 16:

```bash
export VLLM_ENABLE_MC2=1
export VLLM_USE_V1=1
export TASK_QUEUE_ENABLE=1

export VLLM_VERSION=0.9.0
MODEL_PATH=DeepSeek-R1-W8A8-VLLM
python -m vllm.entrypoints.openai.api_server --model=$MODEL_PATH \
    --load-format=prefetch_auto \
    --quantization ascend \
    --served-model-name auto \
    --trust-remote-code \
    --port 8006 \
    -tp=8 \
    -dp=2 \
    --enable-expert-parallel \
    --max-num-seqs 24 \
    --max-model-len 2048 \
    --max-num-batched-tokens 2048 \
    --block-size 128 \
    --no-enable-prefix-caching \
    --additional-config '{"torchair_graph_config":{"enabled":true,"use_cached_graph":true,"graph_batch_sizes":[24]},"ascend_scheduler_config":{"enabled":true}, "expert_tensor_parallel_size":1, "expert_map_path": "delta_gsm8k_temp0.0_16_16.json"}' \
    --gpu-memory-utilization 0.90
```

### DIFF
--- a/vllm_ascend/ops/fused_moe.py
+++ b/vllm_ascend/ops/fused_moe.py
@@ -1092,8 +1092,11 @@ class AscendFusedMoE(FusedMoE):
 
         assert self.quant_method is not None
 
-        local_num_experts = torch.sum(self.expert_map != -1) \
-            if self.expert_map is not None else num_experts
+        if self.log2phy is not None:
+            local_num_experts = self.local_num_experts
+        else:
+            local_num_experts = torch.sum(self.expert_map != -1) \
+                if self.expert_map is not None else num_experts
 
         moe_quant_params = {
             "num_experts": local_num_experts,
--- a/vllm_ascend/ops/expert_load_balancer.py
+++ b/vllm_ascend/ops/expert_load_balancer.py
@@ -85,7 +85,10 @@ class ExpertLoadBalancer(object):
         layer_expert_map = expert_placement_map[layer_id]
         rank_expert_map = layer_expert_map[rank_id].to(
             torch.npu.current_device())
-        rank_local_expert_num = torch.sum(torch.ne(rank_expert_map, -1)).item()
+
+        # valid num in expert_map may not equal to
+        # rank_local_expert_num when num_redundant_experts > 0
+        rank_local_expert_num = len(self.expert_map_tensor[layer_id, rank_id])
         return rank_local_expert_num, rank_expert_map
 
     def get_rank_log2phy_map(self, layer_id, rank_id):


ZhengWG:
@wangxiyuan can you help review it～

jianzs (approved):
LGTM

vllm_ascend/ops/fused_moe.py
```diff
@@ -1092,8 +1092,11 @@ def __init__(
 
         assert self.quant_method is not None
 
-        local_num_experts = torch.sum(self.expert_map != -1) \
-            if self.expert_map is not None else num_experts
+        if self.log2phy is not None:
+            local_num_experts = self.local_num_experts
+        else:
+            local_num_experts = torch.sum(self.expert_map != -1) \
+                if self.expert_map is not None else num_experts
```
jianzs, line 1219: Why not just use the `self.local_num_experts` value when `log2phy` is None?  It's already set by `determine_expert_map`.

https://github.com/vllm-project/vllm-ascend/blob/fe0da59bd11c3c3567eeabfa0620a08e1c54f71b/vllm_ascend/ops/fused_moe.py#L1065-L1067

ZhengWG: Yes, it should return the same value. The current implementation intentionally preserves the original logic.

wangxiyuan:
You should add a e2e test for eplb case. I notice that there is PR for eplb test. https://github.com/vllm-project/vllm-ascend/pull/1186 can you combine it together to make sure the feature works as expect.

songshanhu07:
You need to check whether there is a rank with the same expert number in a JSON file. This may be the reason for your runtime error. The code changes you merged don't seem to make much sense.

ZhengWG:
> You need to check whether there is a rank with the same expert number in a JSON file. This may be the reason for your runtime error. The code changes you merged don't seem to make much sense.

Because when num_redundant_experts > 0, multiple experts with identical logic number might be loaded onto a single rank.

ZhengWG:
> You should add a e2e test for eplb case. I notice that there is PR for eplb test. #1186 can you combine it together to make sure the feature works as expect.

Ok，I will add it soon.

ZhengWG:
Hi @wangxiyuan,

I've added the E2E test and verified it locally. Could you please review the changes when you have time?

Let me know if you have any questions or suggestions.

Thanks in advance!

Yikun:
is it ready to go? Please do a rebase?

ZhengWG:
> is it ready to go? Please do a rebase?

It's ready now~ @Yikun @wangxiyuan

tests/e2e/run_eplb.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+set -eo errexit
+
+. $(dirname "$0")/common.sh
+
+export VLLM_ENABLE_MC2=1
+export VLLM_USE_V1=1
+export TASK_QUEUE_ENABLE=1
+export VLLM_VERSION=0.9.1
```
MengqingCao: please remove this
```suggestion

```

tests/e2e/eplb/test_eplb_e2e.py
```diff
@@ -0,0 +1,166 @@
...
+            if "vllm" in cmd or "proxy" in cmd or "engine_worker" in cmd:
+                kill_process_and_children(proc.pid)
+        except Exception:
+            continue
+
+
+def build_expert_map(expert_map_path,
+                     num_redundant_expert=0,
+                     num_layer=58,
+                     num_device=16,
```
MengqingCao: accutally there is only 4 cards on CI now, please reduce `num_device` to make it work

MengqingCao:
Please add `pytest -sv tests/e2e/eplb` at https://github.com/vllm-project/vllm-ascend/blob/main/.github/workflows/vllm_ascend_test.yaml#L372 to make it run

ZhengWG:
The same e2e test passes successfully in my local environment but fails on CI. The root cause appears to be a CANN version mismatch affecting EP parallel execution, @MengqingCao can you help check it, here is my local env info:   
image

wangxiyuan:
eplb will be refactor, let's close this now.

PR #1227, opened 2025-06-15
[Feature] Support offline recording expert load distribution

jianzs:
This PR implements APIs to record expert workload, allowing us to analyze and improve model performance.  It needs to be used with https://github.com/vllm-project/vllm/pull/19658.


wangxiyuan:
still working on this?

Issue #1170, opened 2025-06-11
[Bug]:DP Crash, After first request, process is crash, 'DPEngineCoreProc' object has no attribute 'dp_rank'


david6666666:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
Your output of above commands here
```

</details>

### 🐛 Describe the bug

vllm tag v0.9.1  
vllm-ascend main 7bdc606677705d072c1dc45f050a3c3471d6d379 

vllm serve DeepSeek-V2-Lite --trust-remote-code --max-model-len=4096 --gpu-memory-utilization=0.95 --data-parallel-size 2 --data-parallel-size-local 2 --block_size 128 --enforce-eager

vllm commit：https://github.com/vllm-project/vllm/pull/18502
#940 is not correct 

After first request, process is crash . Error:
```
....INFO:     127.0.0.1:51658 - "POST /v1/completions HTTP/1.1" 200 OK
INFO 06-11 11:46:37 [loggers.py:118] Engine 000: Avg prompt throughput: 1.9 tokens/s, Avg generation throughput: 10.3 tokens/s, Running: 0 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.0%, Prefix cache hit rate: 0.0%
INFO 06-11 11:46:37 [loggers.py:118] Engine 001: Avg prompt throughput: 1.5 tokens/s, Avg generation throughput: 10.0 tokens/s, Running: 0 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.0%, Prefix cache hit rate: 0.0%
(EngineCore_0 pid=2651) ERROR 06-11 11:46:39 [core.py:517] EngineCore encountered a fatal error.
(EngineCore_0 pid=2651) ERROR 06-11 11:46:39 [core.py:517] Traceback (most recent call last):
(EngineCore_0 pid=2651) ERROR 06-11 11:46:39 [core.py:517]   File "/home/dsv3/project/opensource/vllm/vllm/v1/engine/core.py", line 508, in run_engine_core
(EngineCore_1 pid=2655) ERROR 06-11 11:46:39 [core.py:517] EngineCore encountered a fatal error.
(EngineCore_0 pid=2651) ERROR 06-11 11:46:39 [core.py:517]     engine_core.run_busy_loop()
(EngineCore_1 pid=2655) ERROR 06-11 11:46:39 [core.py:517] Traceback (most recent call last):
[...]

PR #884, opened 2025-05-16
 Fix the device error when using ray as vllm-acend backend

zhuo97:
1. Remove RAY_EXPERIMENTAL_NOSET_ASCEND_RT_VISIBLE_DEVICES
2. Add lazy init for vllm_ascend_C

PR #1235, opened 2025-06-16
[DP][V1] Fix rank set in DP scenario

MengqingCao:
Fix rank set in DP scenario. The new poc version of torch-npu support setting `ASCEND_RT_VISIBLE_DEVICES` dynamically, thus we could use the rank set in `DPEngineCoreProc` directly instead of calculating local rank across dp by hand in the patched `_init_data_parallel`

Fixes https://github.com/vllm-project/vllm-ascend/issues/1170


MengqingCao:
This should be merged after https://github.com/vllm-project/vllm-ascend/pull/884

wangxiyuan:
I like the PR which remove the patch code. Let's merge this asap.

realliujiaxu (approved):

Yikun:
pd: https://github.com/vllm-project/vllm-ascend/actions/runs/15684668372/job/44184711624
accuracy: https://github.com/vllm-project/vllm-ascend/actions/runs/15684668412
long term: https://github.com/vllm-project/vllm-ascend/actions/runs/15684668386

All passed.

PR #1241, opened 2025-06-16
[CI] Move linting system to pre-commits hooks

Potabk:

- As the file `.pre-commit-config.yaml` you can see, we list all the lint check to the file, before submitting your commit, you can run `pip install -r requirements-lint.txt` to have pre-commit installed and  `pre-commit install`  to install the pre-commit hooks, and then linters will run automatically before each commit
- Made some clean code changes to keep the new CI passing


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,24 @@
+name: pre-commit
```
Yikun, line 1: Please move this to vllm_ascend_test.yaml as deps of e2e test

Yikun (commented):
Please keep `format.sh` content and lint type to help merge first versioon

.github/workflows/pre-commit.yml
Potabk: if pre-commits runs parallel with e2e test is batter?

Yikun: e2e resource is limited so do pre check then run e2e is better

PR #1251, opened 2025-06-17
Feat rope: enable npu_mrope by environment variables

Ronald1995:
npu_rotary_embedding will trigger gatherv3's error, enable npu_mrope to solve this issue.

export VLLM_ASCEND_ENABLE_NPU_MROPE=1 to enable npu_mrope

use ut to test this patch

### DIFF
--- a/vllm_ascend/ops/rotary_embedding.py
+++ b/vllm_ascend/ops/rotary_embedding.py
@@ -22,10 +22,8 @@ import torch
 from vllm.model_executor.layers.rotary_embedding import (
     DeepseekScalingRotaryEmbedding, RotaryEmbedding)
 
-
-from vllm_ascend.utils import enable_custom_op
 import vllm_ascend.envs as ascend_envs
-
+from vllm_ascend.utils import enable_custom_op
 
 
 def custom_rotary_embedding_enabled(query, neox_style, head_size):
--- a/tests/e2e/singlecard/ops/test_rotary_embedding.py
+++ b/tests/e2e/singlecard/ops/test_rotary_embedding.py
@@ -200,6 +200,7 @@ def test_rotary_embedding_quant_with_leading_dim(
                                atol=DEFAULT_ATOL,
                                rtol=DEFAULT_RTOL)
 
+
 # test rope with npu_mrope interface with leading dimension and merge seqlen and batch_size as num_tokens
 @pytest.mark.parametrize("is_neox_style", IS_NEOX_STYLE)
 @pytest.mark.parametrize("batch_size", BATCH_SIZES)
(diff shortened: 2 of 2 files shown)


jianzs:
Thank you for your PR. A brief and clear description of your changes would be helpful.

vllm_ascend/envs.py
```diff
@@ -133,6 +133,12 @@
     # value to False to disable the optimized model.
     "USE_OPTIMIZED_MODEL":
     lambda: bool(int(os.getenv('USE_OPTIMIZED_MODEL', '1'))),
+    # VLLM_ASCEND_ENABLE_NPU_MROPE:
+    #   0: using npu_rotary_embedding.
+    #   1: using npu_mrope.
+    # Just a temporary plan，will be removed after npu_mrope supports aclgraph mode.
+    "VLLM_ASCEND_ENABLE_NPU_MROPE":
```
Yikun, line 140: Why we need this ENV? Can we use the `npu_mrope` as default?

IMO, either it is stable enough that we merge it and set it as default. Or it is not stable enough and we are not ready to merge it yet.

cc @ganyi1996ppo

vllm_ascend/ops/rotary_embedding.py
```diff
@@ -20,13 +20,20 @@
 
 import torch
 from vllm.model_executor.layers.rotary_embedding import (
-    DeepseekScalingRotaryEmbedding, RotaryEmbedding)
+    DeepseekScalingRotaryEmbedding,
+    RotaryEmbedding,
+)
```
Yikun: Could we revert all unrelated lint changes? because this will make cherry pick as a diffcult thing.

vllm_ascend/envs.py
wangxiyuan: npu_mrope doesn't work with torchair.

ganyi1996ppo: If the `npu_rotary_embedding` have functionality issue, we should push torch_npu or the kernel-dev group to fix this bug rather than use a new op to replace the old one, right?

ganyi1996ppo: And it there any performance assessment on this kernel?

vllm_ascend/ops/rotary_embedding.py
Ronald1995: ok，i have reverted unrelated lint changes

wangxiyuan:
For reviewer: this pr replace `npu_rotary_embedding ` to  `npu_mrope` to solve gatherV3 bug. While a new problem is that `npu_mrope` doesn't work with aclgraph currently. So a new ENV is added in this PR. it's guaranteed that the error will be fixed in the next pta release. Then the ENV  can be removed. So I'm fine with this change for quick fix case.

wangxiyuan:
@Ronald1995  Just like @jianzs comment, please fill up your commit message. It should include the below info at least: 
1. why the `npu_rotary_embedding ` doesn't work.
2.  what case you fixed with PR.
3. what will be done in the next step

vllm_ascend/envs.py
Ronald1995: this pr replace npu_rotary_embedding  to npu_mrope to solve gatherV3 bug. While a new problem is that npu_mrope doesn't work with aclgraph currently. So a new ENV is added in this PR. it's guaranteed that the error will be fixed in the next pta release. Then the ENV can be removed.

Yikun:
Wait for @ganyi1996ppo decision, have a offline discussion with @ganyi1996ppo he had a simple way to resolve gather v3 error

Yikun:
> this pr replace npu_rotary_embedding to npu_mrope to solve gatherV3 bug. 

This is a wrong info.

Does this PR still needed?

I think if we think it's a good way to enhancement accuracy for RL case, we could introduce a ENV for RL, rather than separate standalone var for every features.

Yikun:
looks like it's no needed any more. feel free to reopen if you still need this.

PR #1252, opened 2025-06-17
[Bugfix] Remove cuda related lines and add additional pip mirror

Potabk:

- For npu environment, we should use  `PYTORCH_NPU_ALLOC_CONF ` rather than  `PYTORCH_CUDA_ALLOC_CONF`
- Add `PIP_EXTRA_INDEX_URL` to make nightly_benchmarks happy


Potabk:
more detils can see https://www.hiascend.com/document/detail/zh/Pytorch/700/comref/Envvariables/Envir_012.html

wangxiyuan (approved):

wangxiyuan:
It's good to add ut as well. It can be done in the future with sleep mode unit test together

PR #1229, opened 2025-06-15
[refactor] Refactoring AscendFusedMoE

zzzzwwjj:
### What this PR does / why we need it?
This PR is used for resolved [issue 1147](https://github.com/vllm-project/vllm-ascend/issues/1147)
1. Move fused_moe code into one file `fused_moe.py`.
2. Integrate branch conditions into function `get_fused_moe_state`.

### Does this PR introduce _any_ user-facing change?
1. This PR has removed the env `VLLM_ENABLE_MC2`, because I think this env is useless, we can make judgments based on the current scenario without this env, it will only increase complexity.
2. This PR has removed the env `USING_LCCL_COM`, because this env has already expired.
3. `additional_config.expert_tensor_parallel_size` has already expired, and now we also use parameter `enable_expert_parallel`, consistent with the vLLM.

PR #1264, opened 2025-06-17
[v0.9.1][refactor] Refactoring AscendFusedMoE (#1229)

ganyi1996ppo:
This PR is the cherry-pick of the PR https://github.com/vllm-project/vllm-ascend/pull/1229 which have already merged into the main branch. 

This PR is used for resolved [issue 1147](https://github.com/vllm-project/vllm-ascend/issues/1147)
1. Move fused_moe code into one file `fused_moe.py`.
2. Integrate branch conditions into function `get_fused_moe_state`.

1. This PR has removed the env `VLLM_ENABLE_MC2`, because I think this env is useless, we can make judgments based on the current scenario without this env, it will only increase complexity.
2. This PR has removed the env `USING_LCCL_COM`, because this env has already expired.
3. `additional_config.expert_tensor_parallel_size` has already expired, and now we also use parameter `enable_expert_parallel`, consistent with the vLLM.

CI passed

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -136,6 +136,7 @@ class AscendMLAMetadata:
     # For logging.
     num_input_tokens: int = 0  # Number of tokens including padding.
 
+    max_num_tokens_across_dp: int = 0
     with_prefill_across_dp: bool = False
 
     query_lens: Optional[list[int]] = None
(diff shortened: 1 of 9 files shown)


ganyi1996ppo:
@zzzzwwjj there is a conflict in `model_runner_v1.py`, please review this code change.

wangxiyuan (approved):

PR #1265, opened 2025-06-17
[DOC] add LoRA user guide

paulyu12:
Add LoRA user guide to DOC. The content refers to [LoRA Adapters](https://docs.vllm.ai/en/latest/features/lora.html).

No

No


wangxiyuan:
@paulyu12 Hi, thanks for the doc. I notice that most content is from vLLM doc. I think in vLLM Ascend doc, we don't need to re-write them again. For the most common part, we can add the link to vLLM directly. What we should write more is: what's the different usage on vLLM Ascend, for example, any model support list or limitation? Any feature + feature limitation? 2. How users use this feature with vLLM Ascend and so on.

paulyu12:
Actually, users can use LoRA with the user guide of vLLM doc. No extra limitation as I know. Because aclgraph mode doesn't support LoRA until now, maybe I'll add some tips about how to fallback to non-aclgraph mode.

paulyu12:
And the models that LoRA supports can be referred by [vLLM Supported Models](https://docs.vllm.ai/en/v0.6.5/models/supported_models.html)

image

wangxiyuan:
So the doc can be simple in vllm-ascend. Just add something like:
```
vLLM Ascend support lora the same as vLLM. More detail can be found in vLLM official document.

```

Then add some limitation, like: does lora in vllm-ascend support graph mode? what model has been tested and so on.

wangxiyuan (approved):

PR #1266, opened 2025-06-17
[BugFix] fix length of sin/cos cache in rope 

whx-sjtu:
This PR fixes the bug that constructs shorter sin/cos cache than model's max positional embedding.


wangxiyuan:
Please update the test as well or later

ganyi1996ppo (approved):

PR #1258, opened 2025-06-17
[Bugfix][Spec Decode] Enable `ACL_OP_INIT_MODE=1` directly only when using V0 spec decode

shen-shanshan:
### What this PR does / why we need it?

Enable `ACL_OP_INIT_MODE=1` directly only when using V0 spec decode.

Find more details at **mengwei805**'s comment in https://github.com/vllm-project/vllm-ascend/pull/1123.

### Does this PR introduce _any_ user-facing change?

The user will not be aware of `VLLM_ASCEND_ACL_OP_INIT_MODE` (`ACL_OP_INIT_MODE`).

### How was this patch tested?

Test scripts:

```python
from vllm import LLM, SamplingParams

prompts = [
    "The future of AI is",
]
sampling_params = SamplingParams(temperature=0.8, top_p=0.95)

llm = LLM(
    model="Qwen/Qwen2.5-1.5B-Instruct",
    tensor_parallel_size=1,
    speculative_config={
        "method": "ngram",
        "num_speculative_tokens": 5,
        "prompt_lookup_max": 4,
    },
)
outputs = llm.generate(prompts, sampling_params)

for output in outputs:
    prompt = output.prompt
    generated_text = output.outputs[0].text
    print(f"Prompt: {prompt!r}, Generated text: {generated_text!r}")
```

Results:

```
Adding requests: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 76.70it/s]
Processed prompts: 100%|███████████████████████████████████████████████████████████████| 1/1 [00:00<00:00,  1.33it/s, est. speed input: 6.64 toks/s, output: 21.26 toks/s]
Prompt: 'The future of AI is', Generated text: ' bright\n\n04/15/2020\n\nBy: James'
```

PR #1271, opened 2025-06-18
[Bugfix][Spec Decode][0.9.1] Enable `ACL_OP_INIT_MODE=1` directly only when using V0 spec decode

shen-shanshan:

Enable `ACL_OP_INIT_MODE=1` directly only when using V0 spec decode.

Cherry pick from https://github.com/vllm-project/vllm-ascend/pull/1258.


Yikun:
What's this PR do?  https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/envs.py#L114

Please keep same logic with latest main otherwise has some special note here?

shen-shanshan:
> What's this PR do? https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/envs.py#L114
> 
> Please keep same logic with latest main otherwise has some special note here?

Find more details at https://github.com/vllm-project/vllm-ascend/pull/1258.

wangxiyuan (approved):

PR #1273, opened 2025-06-18
[DP] Tiny fix of dp and update example

MengqingCao:
Add `max_num_tokens_across_dp` to AscendMetadata to fix dp

This pr fixes the bug introduced by https://github.com/vllm-project/vllm-ascend/pull/1229, which add an arg `max_num_tokens_across_dp` when dp_size > 1.


MengqingCao:
Not sure why CI time out now, try to increase timeout to fix

p.s. the same test sucessed in v0.9.1 already 
https://github.com/vllm-project/vllm-ascend/actions/runs/15729294008/job/44327332939?pr=1277#step:10:370

wangxiyuan (approved):

MengqingCao:
@wangxiyuan plz help merge this and then #1277 , thanks!

PR #1280, opened 2025-06-18
dLLM, short for distributed LLM, an easy-to-use tool for multi-node vllm deployment

zhouyeju:

An easy-to-use tool for multi-node vLLM instance deployment, backed by Ray Actors and Ray Serve. 

For example, with a Ray cluster already provisioned, to deploy a 1P1D disaggregated cluster with TP=4, DP=4 and EP=16, you just simply run
```
dllm deploy \
    -- num-prefill-instances=1 \
    --num-decode-instances=1 \
    -ptp=4 -dtp=4 -pdp=4 -ddp=4 \
    --prefill-startup-params="vllm serve llama2-13B --gpu-memory-utilization=0.8 --kv-transfer-config {kvconnector config here...} \"
    --decode-startup-params="vllm serve llama2-13B --gpu-memory-utilization=0.8 --kv-transfer-config {kvconnector config here...} \"
```
from a randomly chosen master node.

With Ray Dashboard turned on, it is easier to develop&debug with a web interface showing all instances' status and stdout/stderr logs.
- vLLM version: v0.9.1
- vLLM main: https://github.com/vllm-project/vllm/commit/2e610deb72dfc1e34b904d9b6b02c85eefa451d2

### DIFF
--- a/examples/dllm_tools/dllm/scripts.py
+++ b/examples/dllm_tools/dllm/scripts.py
@@ -22,7 +22,7 @@ def cli():
 
 @cli.command(name="deploy", context_settings={"show_default": True})
 @click.option("--head-ip", type=str, help='IP of Ray head node (e.g. "10.2.3.4")', default="auto")
-@click.option("--prefill-instances-num", type=int, help="the num of Prefill instances", default=1)
+@click.option("--prefill-instances-num", type=int, help="the num of Prefill instances", default=0)
 @click.option(
     "--prefill-startup-params",
     type=str,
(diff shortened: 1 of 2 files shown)


wangxiyuan:
dllm is a deployment tool rather than a feature inner vllm-ascend from my sight. 1. why add it into vllm-ascend? is it only work for ascend? 2. If it's a accept feature, I suggest to move to example folder @Yikun @ganyi1996ppo @jianzs

Yikun:
Frankly, I prefer to maintain this in a separate repo and use this as a separate pypi package and add tutorial in vllm ascend.

PR #1291, opened 2025-06-19
[0.9.1][Bugfix] fix dp error in dbo

shikang-hangzhou:

Fix running error in dbo when dp_size>1. Add conditional logic in `_get_forward_metadata_across_dp` to enable dbo. 

None
add test case  `test/e2e/multicard/test_data_parallel.py` at main branch after pr1277 merged
test dp2 inference in dbo model

### DIFF
--- a/vllm_ascend/models/deepseek_dbo.py
+++ b/vllm_ascend/models/deepseek_dbo.py
@@ -74,7 +74,6 @@ from vllm_ascend.multistream.layers import (MultiStreamPostTransformerLayer,
 from vllm_ascend.multistream.metadata import (MultiStreamConfig,
                                               MultiStreamStepMetadata,
                                               make_multistream_metadata_ds)
-from vllm_ascend.multistream.ms_split import compute_split_seq_index
 from vllm_ascend.ops.fused_moe import AscendFusedMoE
 from vllm_ascend.utils import dispose_tensor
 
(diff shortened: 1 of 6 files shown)


wangxiyuan (approved):

vllm_ascend/models/deepseek_dbo.py
```diff
@@ -881,22 +880,8 @@ def forward(
...
-            [token_index, seq_index
-             ] = compute_split_seq_index(attn_metadata.query_lens,
-                                         attn_metadata.attn_state,
-                                         attn_metadata.num_decode_tokens)
-            if token_index == 0 or seq_index == 0 or seq_index == len(
-                    attn_metadata.query_lens):
-                return False
-        # check whether the total tokens exceed the threshold
-        if self.multistream_config is None or attn_metadata.num_actual_tokens < self.multistream_config.min_total_tokens_to_split:
+        if attn_metadata is None or attn_metadata.num_prefills == 0 or not attn_metadata.enable_dbo_across_dp:
```
ganyi1996ppo, line 884: My suggestion is, if dbo already rewrite the modeling, then you are not necessary to adopt the attention backend to do mla, so you can overlap the calculation and communication better, do we have any plan on rewrite the attention impl in dbo path?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -569,16 +570,38 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
             self.input_batch.refresh_sampling_metadata()
 
     def _get_forward_metadata_across_dp(
-            self, total_num_scheduled_tokens: int,
-            with_prefill: bool) -> tuple[int, bool]:
+            self, total_num_scheduled_tokens: int, with_prefill: bool,
+            enable_dbo: bool) -> tuple[int, bool, bool]:
         forward_metadata = torch.tensor(
-            [total_num_scheduled_tokens, with_prefill],
+            [total_num_scheduled_tokens, with_prefill, not enable_dbo],
```
ganyi1996ppo, line 576: Why do we needs to allreduce this? if dbo only enabled in prefill, shouldn't we just use the `with_prefill` flag?

ganyi1996ppo (approved):

ganyi1996ppo (approved):

PR #1297, opened 2025-06-19
[v0.9.1][Fix] Fix block table shape

yiz-liu:

This fix the shape of block_table which was introduced by hybrid kv groups several weeks ago.

Error will be raised when enable prefix-cache (eager or not) and Ascend Scheduler at the same time, just send two identical requests and it will reproduce.


yiz-liu:
I have tested myself and observed no accuracy issues. However, I recommend adding accuracy tests for this scenario to catch any potential bugs. We should take a look into it @MengqingCao 

@ganyi1996ppo @wangxiyuan Please review and merge this at your earliest convenience, as this bug can be triggered easily and has a broad impact.

PR #1312, opened 2025-06-20
[CI] Update guided decoding ut

shen-shanshan:

Update guided decoding ut.


shen-shanshan:
CC: @wangxiyuan @Yikun

wangxiyuan (approved):

wangxiyuan:
Thanks for adding the test back.

PR #1319, opened 2025-06-20
[Benchmark] Set v1 engine as default

Potabk:

Considering v1 engine is becoming more and more important, and we are submitting large v1 related code , transform nightly performance benchmark run with v1 engine by default


Yikun (approved):

Yikun:
Replaced by: https://github.com/vllm-project/vllm-ascend/pull/1349

Yikun:
Sorry, I just saw this PR, feel free open PR if you think this is still need @Potabk

PR #1322, opened 2025-06-20
Handle with_prefill_across_dp for multistream mla

sdmyzlp:
After #1094, decode might be executed with non-compiled mode, despite of `torchair_graph_config.enabled`, causing multistream mla to fail, which assumes torchair compiled mode for decode when `torchair_graph_config.enabled == True`.
Augment that assumption to fix this.

No.

Tested both offline, and by graph mode mla e2e testcase.

### DIFF
--- a/vllm_ascend/models/deepseek_v2.py
+++ b/vllm_ascend/models/deepseek_v2.py
@@ -465,20 +465,21 @@ class CustomDeepseekV2MLAAttention(DeepseekV2MLAAttention):
             hidden_states: torch.Tensor,
             kv_cache: Optional[torch.Tensor] = None,
             attn_metadata: Optional[AttentionMetadata] = None) -> torch.Tensor:
+        enable_multistream_mla = (self.enable_multistream_mla
+                                  and attn_metadata is not None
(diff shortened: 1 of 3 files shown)


ApsarasX:
LGTM

ApsarasX (approved):

wangxiyuan (approved):

wangxiyuan (approved):

vllm_ascend/models/deepseek_v2.py
```diff
@@ -465,20 +465,21 @@ def forward(
             hidden_states: torch.Tensor,
             kv_cache: Optional[torch.Tensor] = None,
             attn_metadata: Optional[AttentionMetadata] = None) -> torch.Tensor:
+        enable_multistream_mla = (self.enable_multistream_mla
```
Yikun, line 468: nit: we'd better log here to notify users why enable_multistream_mla not enable here and add some code comment about this.

Yikun (approved):
some comments inline feel free to open new PR to address.

Yikun:
cc @ganyi1996ppo pls

PR #1109, opened 2025-06-07
Set default values to fix spec decode and fix multicard CI

Yikun:
### How was this patch tested?
CI passed, espcecially multicards CI

PR #913, opened 2025-05-21
[Fix] Fix update_aclgraph_sizes when running MoE models

yiz-liu:
### What this PR does / why we need it?

Fix update_aclgraph_sizes when running MoE models.

PR #1323, opened 2025-06-20
[v0.9.1-dev][CI/UT][bugfix]fix v0 spec decode

mengwei805:
1. [PR913](https://github.com/vllm-project/vllm-ascend/pull/913) introduced an error that caused V0's spec decode function to fail. [PR1109](https://github.com/vllm-project/vllm-ascend/pull/1109) wanted to fix this problem. Unfortunately, the fix broke the ngram function. I fixed the ngram function in this PR.  **PS**: Q: Why is there a problem when ngram is not found when pr1109 is merged? A: The newly introduced problem will only appear when tp>1, and the use cases on CI are all tp=1
2. In versions after 0.7.3, vllm-ascend deleted some spec decode UTs to avoid CI taking too long, including eagle speculative UTs, which made CI unable to take care of the eagle function. I added it(`test_eagle_correctness.py`) back in this PR
3. Because of the reason mentioned in 2, the current version of Eagle has a problem. I located and fixed this problem. It was because vllm's `draft_model_runner.py` was changed and vllm-ascend was not synchronized in time.
4. Currently, the UTs of v0 and v1 are mixed in the spec_decode directory. I split them into two directories: spec_decode_v0 and spec_decode_v1.
5. i found `vllm.spec_decode.multi_step_worker.MultiStepWorker.set_include_gpu_probs_tensor` and `vllm.spec_decode.multi_step_worker.MultiStepWorker.set_should_modify_greedy_probs_inplace` have changed in vllm, so i remove its patchs in this pr.
6. v1 mtp ut failed(https://github.com/vllm-project/vllm-ascend/actions/runs/15782006176/job/44489813330?pr=1323), 
I commented it out. @XWFAlone @JC-ut0 

This PR fixes the functions of ngram and eagle spec decode in the v0 engine

ngram and eagle were tested locally using an 800I A2 machine, using real weights instead of the random small weights used by UT, and using a scenario test with tp>1.
and other were tested by CI


MengqingCao:
Does the same issue exists on branch main?

PR #914, opened 2025-05-21
[WIP][Platform] Add support for Ascend 310P

farawayboat:
### What this PR does / why we need it?

- This PR introduces changes to adapt the `pos_encoding_kernels.cpp`, `utils.h`, `attention.py`, `layernorm.py`, `platform.py`, and `utils.py` files to support Ascend 310P devices. 
- Specifically, it adjusts the `loadSize` constant based on the Ascend AI Core version and adds conditional compilation directives for `bfloat16_t` support. 
- It also includes modifications to handle specific behaviors required for the Ascend 310P, such as tensor alignment and format casting.
- The purpose of these changes is to ensure compatibility and optimal performance on Ascend 310P devices.

### Does this PR introduce _any_ user-facing change?

- Yes, this PR introduces changes that affect the behavior of the library when running on Ascend 310P devices. 
- Users of Ascend 310P will see improved performance and compatibility due to the added support and optimizations.

### How was this patch tested?

The patch has been tested locally on Ascend 310P hardware to ensure that the changes do not break existing functionality and that the new features work as intended.

#### ENV information

```shell
npu-smi info
+--------------------------------------------------------------------------------------------------------+
| npu-smi 24.1.0.1                                 Version: 24.1.0.1                                     |
+-------------------------------+-----------------+------------------------------------------------------+
| NPU     Name                  | Health          | Power(W)     Temp(C)           Hugepages-Usage(page) |
| Chip    Device                | Bus-Id          | AICore(%)    Memory-Usage(MB)                        |
+===============================+=================+======================================================+
| 1536    310P3                 | OK              | NA           65                0     / 0             |
| 0       0                     | 0000:06:00.0    | 0            1524 / 44280                            |
+-------------------------------+-----------------+------------------------------------------------------+
| 1536    310P3                 | OK              | NA           64                17452 / 17452         |
| 1       1                     | 0000:06:00.0    | 0            36314/ 43693                            |
+===============================+=================+======================================================+
| 1792    310P3                 | OK              | NA           72                17636 / 17636         |
| 0       2                     | 0000:07:00.0    | 95           36982/ 44280                            |
+-------------------------------+-----------------+------------------------------------------------------+
| 1792    310P3                 | OK              | NA           68                0     / 0             |
| 1       3                     | 0000:07:00.0    | 0            1216 / 43693                            |
+===============================+=================+======================================================+
| 2048    310P3                 | OK              | NA           60                0     / 0             |
[...]

PR #1327, opened 2025-06-20
supoort 310p moe

Angazenn:
This PR is extension of #914 and #1204 , which support PanguProMoE on 310p platform. Currently `fused_experts_310p` is only implemented for PanguProMoE.


Yikun:
Many thanks for your contributions, I squashed all 310P related commits with your co-author: https://github.com/vllm-project/vllm-ascend/pull/1333

PR #1345, opened 2025-06-21
[CI] Enable merge trigger unit test and accuracy test schedule job

Yikun:
Enable merge trigger unit test and accuracy test schedule job

No

CI passed


.github/workflows/accuracy_test.yaml
```diff
@@ -181,7 +187,7 @@ jobs:
 
       - name: Install lm-eval, ray, and datasets
         run: |
-            pip install lm-eval
+            pip install lm-eval==0.4.8
```
Yikun, line 190: Pin to `0.4.8`, because qwen3 8b base gsm8k 0.4.9 results was rong

PR #1355, opened 2025-06-22
use npu_moe_gating_top_k_softmax

ttanzhiqiang:
The optimization solution for non-deepseek select_experts is to replace gating_topk_softmax with softmax+topk+to, which is optimized from 37us to 14us on bf16/fp16 of qwen3-235b
截屏2025-06-22 21 19 41
截屏2025-06-22 21 19 49

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/1a4f35e2eaa3ebdecb8ef9ff8302b01e289305c9


vllm_ascend/ops/common_fused_moe.py
```diff
@@ -44,19 +48,26 @@ def forward_oot(
     apply_router_weight_on_input: bool = False,
     activation: str = "silu",
 ) -> torch.Tensor:
-    topk_weights, topk_ids = select_experts(
-        global_num_experts=global_num_experts,
```
ApsarasX, line 58: Why is the `global_num_experts` variable no longer being used?

ttanzhiqiang: fixed it

ttanzhiqiang:
@wangxiyuan @ganyi1996ppo

ttanzhiqiang:
@wangxiyuan @Yikun

ttanzhiqiang:
@ganyi1996ppo Please help review

ganyi1996ppo:
Your screenshot seems only contains host time, can you paste the device time of this kernel too?

ttanzhiqiang:
> Your screenshot seems only contains host time, can you paste the device time of this kernel too?
The optimization solution for non-deepseek select_experts is to replace gating_topk_softmax with softmax+topk+to, which is optimized from 78us to 28us on bf16/fp16 of qwen3-235b
截屏2025-07-09 12 13 40
截屏2025-07-09 12 13 56
Both prefill and decode can be used

ganyi1996ppo (approved):

vllm_ascend/envs.py
```diff
@@ -121,6 +121,8 @@
     # value to False to disable the optimized model.
     "USE_OPTIMIZED_MODEL":
     lambda: bool(int(os.getenv('USE_OPTIMIZED_MODEL', '1'))),
+    "SELECT_GATING_TOPK_SOTFMAX_EXPERTS":
```
wangxiyuan, line 123: I dont' think we should add more config. Instead, how about check the model type to decide which function will be called?

ttanzhiqiang: select_gating_top_k_softmax_experts is theoretically better than select_experts in non-quantization

vllm_ascend/ops/fused_moe.py
```diff
@@ -822,6 +823,39 @@ def fused_experts(
...
+        topk_weights: Routing weights of shape (num_tokens, top_k).
+        topk_ids: Selected expert IDs of shape (num_tokens, top_k).
+
+    Raises:
+        ValueError: If an unsupported scoring function is provided.
+    """
+    topk_weights, topk_ids, row_idx = torch_npu.npu_moe_gating_top_k_softmax(
+        router_logits, None, k=top_k)
+
+    # # Required by npu_moe_init_routing
```
wangxiyuan, line 847: remove the comment code

vllm_ascend/envs.py
wangxiyuan: As discuss offline, we'll remove this env once the function is stable enough

wangxiyuan (approved):

PR #1361, opened 2025-06-23
[V0.9.1][Bugfix] Remove schedulre patch for disaggregated PD

ganyi1996ppo:
Remove the scheduler patch for disaggregated PD, since we found the patch can not really work on the online serving path. 

No

Ci test will guarantee this


jianzs:
@ganyi1996ppo  So, what's your solution?  Use AscendScheduler to fulfill the need?

ganyi1996ppo:
> So, what's your solution? Use AscendScheduler to fulfill the need? 

Yep, totally adopt the ascend scheduler for disaggregated pd, and with the method overriding on `update_waiting_for_remote_kv`, schedule with chunked prefill should also works fine.

examples/disaggregate_prefill_v1/README.md
```diff
@@ -66,7 +66,7 @@ vllm serve /data01/deepseek_r1_w8a8_zhw \
   "kv_connector_module_path": "vllm_ascend.distributed.llmdatadist_c_mgr_connector"
   }'  \
   --additional-config \
-  '{"torchair_graph_config": {"enabled": false, "enable_multistream_shared_expert": false}, "expert_tensor_parallel_size": 1}'
+  '{"torchair_graph_config": {"enabled": false, "enable_multistream_shared_expert": false}, “ascend_scheduler_config":{"enable_chunked_prefill":"false"}}'
```
wangxiyuan: enabled should be set to true as well for ascend_scheduler_config

ganyi1996ppo: Sure, changed

ganyi1996ppo:
Just noticed that the deepseek_dbo dose not support float-point test for now, just confirmed this with @zxdukki . I disabled this test in the ci workflow and will re-enable it after we verify this case in the future. Also I'm quite curious about how did this test get passed serveral time before, they should all be failed theoretically. cc @wangxiyuan @Yikun

PR #1376, opened 2025-06-23
[Bugfix] Fix sleep mode level 2

Potabk:

For sleep mode level 2, we discarded model both weights and kv_cache, but the problems is:  When we discard weights, we also discard some tensors representing the model state, such as:  `running_mean / running_var` in BatchNorm、rope cos-sin cache ...  when  we update weights, but forgot to update buffers as well, this will lead to some unknown issue

### DIFF
--- a/vllm_ascend/worker/worker_v1.py
+++ b/vllm_ascend/worker/worker_v1.py
@@ -91,10 +91,19 @@ class NPUWorker(WorkerBase):
             init_cached_hf_modules()
 
         self.profiler = self._init_profiler()
+        # Buffers saved before sleep
+        self._sleep_saved_buffers: dict[str, torch.Tensor] = {}
 
     def sleep(self, level: int = 1) -> None:
         NPUPlatform.set_device(self.device)
         free_bytes_before_sleep = NPUPlatform.mem_get_info()[0]
+        # Save the buffers before level 2 sleep
+        if level == 2:
+            model = self.model_runner.model
+            self._sleep_saved_buffers = {
+                name: buffer.cpu().clone()
+                for name, buffer in model.named_buffers()
+            }
         allocator = CaMemAllocator.get_instance()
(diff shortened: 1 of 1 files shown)


Yikun:
@MengqingCao Could help review this and sync with RL team have a test.

MengqingCao:
> @MengqingCao Could help review this and sync with RL team have a test.

Agree, we'd better to add an e2e test for level 2, after offline discussion, the RL team will add one later

vllm_ascend/worker/worker_v1.py
```diff
@@ -91,10 +91,19 @@ def __init__(
             init_cached_hf_modules()
 
         self.profiler = self._init_profiler()
+        # Buffers saved before sleep
+        self._sleep_saved_buffers: dict[str, torch.Tensor] = {}
```
wangxiyuan: please only init this value when sleep mode is enabled.

Potabk: fixed at [1861d3b](https://github.com/vllm-project/vllm-ascend/pull/1376/commits/1861d3b97a3d161e66c89631bf765135a1038806)

vllm_ascend/worker/worker_v1.py
```diff
@@ -103,13 +103,23 @@ def __init__(
...
     def sleep(self, level: int = 1) -> None:
         if not sleep_mode_enabled():
             raise ValueError(
                 "Sleep mode is not enabled. Please compile vllm-ascend with COMPILE_CUSTOM_KERNELS=1."
             )
         free_bytes_before_sleep = NPUPlatform.mem_get_info()[0]
+        # Save the buffers before level 2 sleep
+        if level == 2:
+            model = self.model_runner.model
+            self._sleep_saved_buffers = {
```
momo609, line 127: This operation only offloads to CPU, but freebytes remains unchanged. Is the sleepmode=2 setting taking effect?

wangxiyuan (approved):

PR #1380, opened 2025-06-23
[Bugfix] Fix memory-leak caused by dist._functional_collectives.reduce_scatter_tensor

ApsarasX:
In some cases, `dist._functional_collectives.reduce_scatter_tensor` can cause its input tensor not to be released immediately after the current layer ends. Instead, it will only be released when the GPU memory usage of the current process reaches a certain threshold (approximately every 15 layers each time).

**Before Fix**

截屏2025-06-24 01 26 13

**After Fix**
截屏2025-06-24 01 23 43

No

- vLLM version: v0.9.1
- vLLM main: https://github.com/vllm-project/vllm/commit/9ff2af6d2ba1757c6e59fe803225a73e61fe526f

### DIFF
--- a/vllm_ascend/ops/fused_moe.py
+++ b/vllm_ascend/ops/fused_moe.py
@@ -39,6 +39,8 @@ from vllm.model_executor.layers.quantization.base_config import \
 
 import vllm_ascend.envs as envs_ascend
 from vllm_ascend.ascend_config import get_ascend_config
+from vllm_ascend.distributed.communication_op import \
+    data_parallel_reduce_scatter
 from vllm_ascend.distributed.parallel_state import get_ep_group, get_etp_group
 from vllm_ascend.ops.expert_load_balancer import ExpertLoadBalancer
 from vllm_ascend.utils import (FusedMoEState, dispose_tensor,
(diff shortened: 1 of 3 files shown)


ApsarasX:
@realliujiaxu @linfeng-yuan @zzzzwwjj

vllm_ascend/ops/fused_moe.py
```diff
@@ -1297,11 +1297,16 @@ def forward(self,
                 final_hidden_states = final_hidden_states[:num_tokens]
             dispose_tensor(e_hidden_states)
         elif self.dp_size > 1 and fused_moe_state == FusedMoEState.AllGather:
-            final_hidden_states = dist._functional_collectives.reduce_scatter_tensor(
-                e_hidden_states,
-                "sum",
-                scatter_dim=0,
-                group=get_dp_group().device_group)
+            final_hidden_states_shape = (
```
realliujiaxu: Recommend extracting into a function: data_parallel_reduce_scatter(x: Tensor, op=SUM)
This communication operator can be placed in vllm_ascend/distributed/communication_op.py, following the style of vLLM.

vllm_ascend/ops/fused_moe.py
```diff
@@ -1219,7 +1219,7 @@ def forward(self,
         else:
             real_top_k = self.top_k
 
-        num_tokens, hidden_size = hidden_states.shape
+        num_tokens, _ = hidden_states.shape
```
realliujiaxu: Irrelevant modifications, recommend rolling back

vllm_ascend/ops/fused_moe.py
```diff
@@ -1297,11 +1297,16 @@ def forward(self,
                 final_hidden_states = final_hidden_states[:num_tokens]
             dispose_tensor(e_hidden_states)
         elif self.dp_size > 1 and fused_moe_state == FusedMoEState.AllGather:
-            final_hidden_states = dist._functional_collectives.reduce_scatter_tensor(
```
ganyi1996ppo: @zzzzwwjj why were we use this reduce_scattr_tensor before?

realliujiaxu: He forgot :)
In early version of torch, communication op under torch.distributed package cannot be captured by torch.compile. I guess that's why dist._functional_collectives.reduce_scatter_tensor is adopted

Yikun:
@ganyi1996ppo Please help take a look, thanks.

wangxiyuan (approved):

PR #1395, opened 2025-06-24
shared_experts+router_experts merge all_reduce(Improve TTOP 5ms)

ttanzhiqiang:
When all_reduce_merge is in progress, shared_experts does not do all_reduce in mlp, but waits until shared_experts+router_experts are completed before doing all_reduce
In prefill and decode, as long as shared_experts+router_experts are all_reduce, there will be benefits.
bash examples/run_dp_attention_etp16.sh
bash examples/run_dp_attention_etp16_benmark.sh
- vLLM version: v0.9.1
- vLLM main: https://github.com/vllm-project/vllm/commit/977180c912b1b07153decbeb62c2cef24032a701


ApsarasX:
LGTM

ttanzhiqiang:
@Yikun @wangxiyuan @ganyi1996ppo

jianzs:
@ttanzhiqiang  plz make ci happy.

examples/run_dp_attention_etp16.sh
```diff
@@ -4,9 +4,12 @@ export TASK_QUEUE_ENABLE=1
 source /usr/local/Ascend/ascend-toolkit/set_env.sh
 source /usr/local/Ascend/nnal/atb/set_env.sh
 export ASCEND_LAUNCH_BLOCKING=0
-export VLLM_VERSION=0.9.0
+export VLLM_VERSION=0.9.1
+export VLLM_ASCEND_SHARED_ROUTER_ALL_REDUCE_MERGE=1
 
 nohup python -m vllm.entrypoints.openai.api_server --model=/mnt/deepseek/DeepSeek-R1-W8A8-VLLM \
+    --load-format=prefetch_auto \
```
jianzs: Don't use features not in the community's scope.

vllm_ascend/ops/fused_moe.py
```diff
@@ -1135,8 +1137,10 @@ def __init__(
         self.e_score_correction_bias = e_score_correction_bias
         self.expert_map = None
         self.activation = activation
+        self.routed_scaling_factor = routed_scaling_factor
         self.log2phy = None
         self.global_redundant_expert_num = 0
+        self.all_reduce_merge = VLLM_ASCEND_SHARED_ROUTER_ALL_REDUCE_MERGE
```
jianzs: ditto

vllm_ascend/models/deepseek_v2.py
```diff
@@ -263,17 +265,20 @@ def __init__(
             topk_group=config.topk_group,
             prefix=f"{prefix}.experts",
             scoring_func=config.scoring_func,
-            e_score_correction_bias=self.gate.e_score_correction_bias)
+            e_score_correction_bias=self.gate.e_score_correction_bias,
+            routed_scaling_factor=self.routed_scaling_factor)
 
         if config.n_shared_experts is not None:
+            self.all_reduce_merge = VLLM_ASCEND_SHARED_ROUTER_ALL_REDUCE_MERGE
```
jianzs: Why not use the `envs_ascend .VLLM_ASCEND_SHARED_ROUTER_ALL_REDUCE_MERGE` here instead of creating a global one?

ttanzhiqiang:
> @ttanzhiqiang plz make ci happy.
The vllm interface has changed and needs to be re-adapted
https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/rotary_embedding.py#L1462C9-L1464
@jianzs @ganyi1996ppo @wangxiyuan

examples/run_dp_attention_etp16.sh
ttanzhiqiang: done

vllm_ascend/models/deepseek_v2.py
ttanzhiqiang: done

vllm_ascend/ops/fused_moe.py
ttanzhiqiang: done

ttanzhiqiang:
Theoretically, this solution is only applicable to AllGather and AllGatherEP. At the same time, a shared expert is required. This solution is to merge the all_reduce of the shared expert and the routing expert. Currently, the only model I have seen is deepseekv3, and qwen3-235b does not have a shared expert.

ttanzhiqiang:
@wangxiyuan @Yikun

ganyi1996ppo:
Looks good, this can help to prevent one unnecessary all_reduce. Maybe we can set it as default path and remove this env variable? @ttanzhiqiang

ttanzhiqiang:
> Looks good, this can help to prevent one unnecessary all_reduce. Maybe we can set it as default path and remove this env variable? @ttanzhiqiang

update @ganyi1996ppo

ganyi1996ppo (approved):

Issue #1324, opened 2025-06-20
[Bug]: qwen3 moe failed with aclgraph


MengqingCao:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
Your output of above commands here
```

</details>

### 🐛 Describe the bug

1. aclgraph disabling ep
```bash
  File "/usr/local/python3.11.10/lib/python3.11/site-packages/vllm/v1/executor/multiproc_executor.py", line 202, in get_response                                  

    raise RuntimeError(                                                                                                                                           

RuntimeError: Worker failed with error 'Constraints violated (L['input_ids'].size()[0])! For more information, run with TORCH_LOGS="+dynamic".                    

  - Not all values of RelaxedUnspecConstraint(L['input_ids'].size()[0]) are valid because L['input_ids'].size()[0] was inferred to be a constant (2048).   
```
https://github.com/vllm-project/vllm-ascend/issues/1115#issuecomment-2959368683

2. aclgraph + enable ep --> gatherV3 error
https://github.com/vllm-project/vllm-ascend/issues/1115#issuecomment-2965464610

ApsarasX:
PR https://github.com/vllm-project/vllm-ascend/pull/1381 may be helpful

MengqingCao:
closing as fixed by https://github.com/vllm-project/vllm-ascend/pull/1416

Issue #1038, opened 2025-05-30
[Bug]: Failed to complete vllm benchmark after enable VLLM_USE_V1=1 due to gather_v3 error


Yikun:
### Your current environment

vLLM v0.9.0
vLLM Ascend main v0.9.0rc1.dev0531
CANN:
package_name=Ascend-cann-toolkit
version=8.1.RC1

200 prompt only 28 success, then serve crash.

RuntimeError: ACL stream synchronize failed, error code:507035 
```
[rank0]:[W530 17:44:21.455843271 compiler_depend.ts:526] Warning: NPU warning, error code is 507035[Error]:
[Error]: The vector core execution is abnormal.
        Rectify the fault based on the error information in the ascend log.
EH9999: Inner Error!
        rtDeviceSynchronizeWithTimeout execute failed, reason=[vector core exception][FUNC:FuncErrorReason][FILE:error_message_manage.cc][LINE:53]
EH9999: [PID: 21632] 2025-05-30-17:44:21.283.810 wait for compute device to finish failed, runtime result = 507035.[FUNC:ReportCallError][FILE:log_inner.cpp][LINE:161]
        TraceBack (most recent call last):
 (function npuSynchronizeUsedDevices)
[W530 17:44:21.458925982 compiler_depend.ts:508] Warning: NPU warning, error code is 507035[Error]:
[Error]: The vector core execution is abnormal.
        Rectify the fault based on the error information in the ascend log.
EH9999: Inner Error!
        rtDeviceSynchronizeWithTimeout execute failed, reason=[vector core exception][FUNC:FuncErrorReason][FILE:error_message_manage.cc][LINE:53]
EH9999: [PID: 21632] 2025-05-30-17:44:21.287.450 wait for compute device to finish failed, runtime result = 507035.[FUNC:ReportCallError][FILE:log_inner.cpp][LINE:161]
        TraceBack (most recent call last):
 (function npuSynchronizeDevice)
[W530 17:44:21.460749068 compiler_depend.ts:151] Warning: NPU warning, error code is 507035[Error]:
[Error]: The vector core execution is abnormal.
        Rectify the fault based on the error information in the ascend log.
EH9999: Inner Error!
        rtDeviceSynchronizeWithTimeout execute failed, reason=[vector core exception][FUNC:FuncErrorReason][FILE:error_message_manage.cc][LINE:53]
[...]

PR #1397, opened 2025-06-24
[v0.9.1][Bugfix] Reset all unused positions to prevent out-of-bounds in GatherV3

yiz-liu:

Reset all unused positions in `NPUModelRunner` to prevent out-of-bounds asserts in the `GatherV3` operator.

Currently, in [`get_splitfuse_attn_mask`](https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/attention/attention.py#L124), the `position` tensor may contain values that exceed the dimensions of the attention mask, triggering a `GatherV3` boundary check failure. These invalid indices originate from stale “dirty” entries left over in `position` due to padding logic in the ACL graph. Specifically, in [`_process_reqs`](https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/worker/model_runner_v1.py#L989), the variable `num_input_tokens` is always greater than or equal to `total_num_scheduled_tokens`, so any positions not explicitly cleared from a previous batch will persist and cause this sporadic error.

BTW, in the original vLLM implementation, masks are constructed internally using other args, so these lingering values do not surface. However, on the Ascend platform—where split-fuse attention requires externally supplied masks—these residual indices become critical and lead to this elusive, hard-to-reproduce failure.

The fix is to explicitly reset or zero out all unused entries in the `position` tensor before passing it to `GatherV3`, ensuring that every index lies within the valid range of the attention mask.

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -870,6 +870,7 @@ class NPUModelRunner(LoRAModelRunnerMixin):
                 self.mrope_positions_cpu[:, :total_num_scheduled_tokens],
                 non_blocking=True)
 
+        self.positions[total_num_scheduled_tokens:num_input_tokens].zero_()
         self.positions[:total_num_scheduled_tokens].copy_(
             self.positions_cpu[:total_num_scheduled_tokens], non_blocking=True)
         positions = self.positions[:num_input_tokens]


yiz-liu:
This pull request addresses the following issues:
#1324 , #1115 , #1038

Yikun:
Please also submit the PR to main

ApsarasX:
LGTM

PR #1403, opened 2025-06-24
[Fix] Prevent Forced Stream Synchronization Triggered by Environment …

rjg-lyh:
…Variable

Prevent Forced Stream Synchronization Triggered by setting 'ASCEND_LAUNCH_BLOCKING' env to '1'. Additionally, expanded the conditions to set block_size to 128.

No.

CI passed with existing test.

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -180,6 +180,11 @@ class NPUPlatform(Platform):
             compilation_config.use_inductor = False
             compilation_config.splitting_ops.extend(
                 ["vllm.unified_ascend_attention_with_output"])
+            if os.getenv("ASCEND_LAUNCH_BLOCKING", '0') == '1':
+                logger.info(
+                    "ACL Graph mode does not allow to synchronize captured-stream, "
+                    "forcing asynchronize mode.")
+                os.environ["ASCEND_LAUNCH_BLOCKING"] = "0"
             update_aclgraph_sizes(vllm_config)
 
         if parallel_config and parallel_config.worker_cls == "auto":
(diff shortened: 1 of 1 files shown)


vllm_ascend/platform.py
```diff
@@ -180,6 +180,11 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
             compilation_config.use_inductor = False
             compilation_config.splitting_ops.extend(
                 ["vllm.unified_ascend_attention_with_output"])
+            if os.getenv("ASCEND_LAUNCH_BLOCKING", '0') == '1':
+                logger.info(
+                    "ACL Graph mode does not allow to synchronize captured-stream, "
+                    "forcing asynchronize mode.")
+                os.environ["ASCEND_LAUNCH_BLOCKING"] = "0"
```
ApsarasX, line 187: I think it's better to just fail outright, so users understand the consequences of their actions.

rjg-lyh: Thank you for your suggestion. Indeed, aborting outright would make the issue more apparent to users. However, from the perspective of framework robustness, I personally believe that manually intercepting and modifying the value is a more suitable approach.

vllm_ascend/platform.py
```diff
@@ -220,6 +216,16 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
                     ascend_config.ascend_scheduler_config)
                 vllm_config.scheduler_config = ascend_scheduler_config
 
+        if cache_config:
```
Yikun, line 219: quick question: Why move this?

vllm_ascend/platform.py
```diff
@@ -180,6 +180,11 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
             compilation_config.use_inductor = False
             compilation_config.splitting_ops.extend(
                 ["vllm.unified_ascend_attention_with_output"])
+            if os.getenv("ASCEND_LAUNCH_BLOCKING", '0') == '1':
+                logger.info(
```
Yikun, line 184: This should be warning level at least.

vllm_ascend/platform.py
Yikun: Just curious which case the ASCEND_LAUNCH_BLOCKING will be enabled? I assume that the users specified this explicitly rather than by default?

If so, I aggree with @ApsarasX this should be raised a error here, or at least a warning.

PR #1408, opened 2025-06-24
Add profiling multimodal model step and fix the OOM bug when profilin…

ChenTaoyu-SJTU:
This PR is to add the multimodal profile functionality. It also fixes the out-of-memory bug (using a solution that might be temporary).

The reason this approach is temporary is that on the CUDA platform in the upstream vLLM, the Qwen2.5 VL's ViT layers (i.e., the encoder layers) do not run out of memory even without adding `@torch.inference_mode()`. However, under the current PyTorch and Ascend hardware environment, **during the forward propagation through the multiple blocks of the ViT (over 30 layers in total), the memory increases by approximately 1.2 GB per layer**, which is unacceptable. Yet, after adding `@torch.inference_mode()`, **this issue disappears**, and the memory increase behavior becomes almost similar to the growth pattern observed on the CUDA platform. But the underlying cause currently remains unknown, need more research.

no

By pass the CI


Yikun:
please do a rebase

ChenTaoyu-SJTU:
need code review @Yikun @wangxiyuan @yiz-liu @shen-shanshan . By the way, why didn't the accuracy test trigger after I added the label?

Yikun:
You should first add `vl-accuracy-test` and then label `ready-for-test` otherwise will trigger twice.

The previous job link:
https://github.com/vllm-project/vllm-ascend/actions/runs/15869325461

shen-shanshan:
LGTM.

PR #1409, opened 2025-06-24
[v0.9.1][perf] add a switch for enabling NZ layout in weights and enable NZ for GMM.

linfeng-yuan:

1. add a switch for enabling NZ layout in weights
2. enable NZ for GMM
3. replace magic number of weights layout 

Users should set `enable_weight_nz_layout` to `true` in `--additional-config` when they wanna enable weights NZ layout.

1) CI passed. 
2) accuracy and performance comparison (only gsm8k-lite)

### DIFF
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -39,6 +39,8 @@ class AscendConfig:
         self.expert_map_path = additional_config.get("expert_map_path", None)
         self.chunked_prefill_for_mla = additional_config.get(
             "chunked_prefill_for_mla", False)
+        self.enable_weight_nz_layout = additional_config.get(
+            "enable_weight_nz_layout", False)
 
 
 class TorchairGraphConfig:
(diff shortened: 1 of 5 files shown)


vllm_ascend/ascend_config.py
```diff
@@ -41,6 +41,8 @@ def __init__(self, vllm_config):
         self.expert_map_path = additional_config.get("expert_map_path", None)
         self.chunked_prefill_for_mla = additional_config.get(
             "chunked_prefill_for_mla", False)
+        self.enable_weight_nz_layout = additional_config.get(
+            "enable_weight_nz_layout", False)
```
wangxiyuan, line 43: once when you sync the PR to main, please update additional config doc as well. And if this is a temp config, add TODO here is necessary as well.

wangxiyuan:
commit message should be updated

linfeng-yuan:
> commit message should be updated

Done.

vllm_ascend/ascend_config.py
linfeng-yuan: > once when you sync the PR to main, please update additional config doc as well. And if this is a temp config, add TODO here is necessary as well.

Sure, thanks for reminding me.

PR #1410, opened 2025-06-24
[Perf] Add a switch to enable NZ layout in weights

linfeng-yuan:


vllm_ascend/ascend_config.py
```diff
@@ -41,6 +41,8 @@ def __init__(self, vllm_config):
         self.expert_map_path = additional_config.get("expert_map_path", None)
         self.chunked_prefill_for_mla = additional_config.get(
             "chunked_prefill_for_mla", False)
+        self.enable_weight_nz_layout = additional_config.get(
```
Yikun, line 44: Could we enable this by default? Avoid to add many parameters

Yikun (commented):
@linfeng-yuan Please do a rebase.

@Angazenn Please also help to review this

vllm_ascend/ascend_config.py
Angazenn: As far as I know, NZ is not allowed in aclgraph mode at this moment. When running in eager mode or torchair graph mode, NZ seems to be preferred than ND in most cases (especially matmul). @linfeng-yuan @yiz-liu Have you encountered any other cases that NZ is not allowed or NZ worses performance compared to ND?

linfeng-yuan:
NZ is enabled by default on main branch, and this PR would be closed.

PR #1422, opened 2025-06-25
[refactor] Refactoring forward_context and model_runner_v1

zzzzwwjj:
A refactoring of `forward_context` and `model_runner_v1`, add some context which is necessary in model inference into `forward_context`, and refactor `dummy_run` logic, make it more reasonable.
Some details for this PR:
1. Fix acc bug when online + multi-DP + eager mode + all_gather mode;
2. Fix bug when online + multi-DP + eager mode + mc2 mode;
3. Fix bug when A2 + eager mode + mc2 mode;
4. enable different token_num on different chip when mc2 mode;
5. Update scripts in `examples` dir;

This PR remove `expert_tensor_parallel_size` in `additional_config`, we will use `enable_expert_parallel` to control whether expert_parallel is enable, which is consistent with vLLM.


vllm_ascend/utils.py
```diff
@@ -277,19 +278,27 @@ def npu_wait_tensor(self: torch.Tensor,
...
 
 
-# TODO(zzzzwwjj): move this into forward_context
-class FusedMoEState(Enum):
-    AllGather = 0
-    All2All = 1
-    MC2 = 2
+class AscendSocVersion(Enum):
+    A2 = 0
+    A3 = 1
```
whx-sjtu, line 283: Do we need to add another version here corresponding to 310P?

zzzzwwjj: We can do it in the future.

docs/source/user_guide/additional_config.md
```diff
@@ -28,7 +28,6 @@ The following table lists the additional configuration options available in vLLM
 |-------------------------------| ---- |------|-----------------------------------------------------------------------------------------------|
 | `torchair_graph_config`       | dict | `{}` | The config options for torchair graph mode                                                    |
 | `ascend_scheduler_config`     | dict | `{}` | The config options for ascend scheduler                                                       |
-| `expert_tensor_parallel_size` | str | `0`  | Expert tensor parallel size the model to use.                                                 |
```
MengqingCao, line 31: Don't we maintain etp any more?

wangxiyuan (approved):

docs/source/user_guide/additional_config.md
yiz-liu: Given the absence of relevant scenarios, employing EP or full TP is sufficient, for now. We may subsequently advocate implementing expert tensor parallelism in vLLM to support scenarios where the number of nodes exceeds the number of experts.

yiz-liu:
LGTM

whx-sjtu:
LGTM

weijinqian0:
This solution is not fully aligned with the current ETP solution. For example, EP and ETP cannot be supported at the same time.

docs/source/user_guide/additional_config.md
leograssroot: However, we do have customer scenarios that require such configurations. While DeepSeek models might not need this, there are use cases involving large-scale MoE (Mixture of Experts) models that require splitting across both Tensor Parallelism (TP) and Expert Parallelism (EP), or sometimes just TP alone. This is exactly the case with the current Jieyue Xingchen models

ApsarasX:
merge this PR to main?

PR #1434, opened 2025-06-25
add chunk mc2 for prefill

NNUCJ:
Add chunk mc2 in prefill 

1. In offline inference npu, dp=1, add some environments 
```
 os.environ["VLLM_ASCEND_ENABLE_CHUNK_MC2"]="1"
 os.environ["HCCL_BUFFSIZE"]="1024" 
 os.environ["VLLM_ASCEND_FUSED_MOE_MC2_CHUNK_SIZE"]="32"
```

2. One A3 Node, dp_size 2, tp_size 8
```
export VLLM_USE_V1=1
export HCCL_BUFFSIZE=1024
export VLLM_ASCEND_FUSED_MOE_MC2_CHUNK_SIZE=32
export VLLM_ASCEND_ENABLE_CHUNK_MC2=1
python data_parallel.py \
    --model="/mnt/nfs/weight/dsv3_w8a8" \
    --dp-size=2 \
    --tp-size=8 
```

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2101,7 +2101,9 @@ class NPUModelRunner(LoRAModelRunnerMixin):
         if self.parallel_config.enable_expert_parallel:
             new_graph_batch_sizes = []
             for graph_batch_size in self.torchair_graph_batch_sizes:
-                cur_graph_batch_size = graph_batch_size + tp_size - graph_batch_size % tp_size
-                if cur_graph_batch_size not in new_graph_batch_sizes:
+                cur_graph_batch_size = (graph_batch_size + tp_size -
+                                        1) // tp_size * tp_size
+                if cur_graph_batch_size not in new_graph_batch_sizes and \
+                    cur_graph_batch_size <= self.scheduler_config.max_num_batched_tokens:
                     new_graph_batch_sizes.append(cur_graph_batch_size)
             self.torchair_graph_batch_sizes = new_graph_batch_sizes


vllm_ascend/ascend_forward_context.py
```diff
@@ -7,22 +7,37 @@
...
-    elif ep_size < 16 or with_prefill:
-        return FusedMoEState.All2All
+    enable_chunk_mc2 = envs.VLLM_ASCEND_ENABLE_CHUNK_MC2
+    if enable_chunk_mc2:
+        if ep_size == 1:
+            return FusedMoEState.AllGather
+        # NOTE: mc2 need ep_size >= 16 & all2all can't use in torchair graph.
+        elif ep_size < 16 and with_prefill:
+            return FusedMoEState.All2All
+        elif ep_size >= 16 and with_prefill:
```
whx-sjtu: line 24 to 28 is duplicated with line 34 to 38, refactor here.

vllm_ascend/ascend_forward_context.py
```diff
@@ -7,22 +7,37 @@
 from vllm.distributed import get_dp_group
 from vllm.forward_context import get_forward_context, set_forward_context
 
+import vllm_ascend.envs as envs
+
 
 class FusedMoEState(Enum):
     AllGather = 0
     All2All = 1
```
ganyi1996ppo: Just wonder if we still need this all to all path.

vllm_ascend/ascend_forward_context.py
```diff
@@ -7,22 +7,37 @@
...
 
 
 # TODO(zzzzwwjj): add soc_version to choose branch
 def get_fused_moe_state(ep_size: int, with_prefill: bool):
-    if ep_size == 1:
-        return FusedMoEState.AllGather
-    # NOTE: mc2 need ep_size >= 16 & all2all can't use in torchair graph.
-    elif ep_size < 16 or with_prefill:
-        return FusedMoEState.All2All
+    enable_chunk_mc2 = envs.VLLM_ASCEND_ENABLE_CHUNK_MC2
```
ganyi1996ppo: if dispatch&combine naturally beats all to all, we can make it a default path for a3

vllm_ascend/quantization/w8a8_dynamic.py
```diff
@@ -249,6 +253,83 @@ def fused_experts_with_mc2(
...
+            hidden_states_chunk_list.append(prefill_expert_outputs)
+        else:
+            hidden_states_chunk_list.append(prefill_expert_outputs[0])
+            shared_chunk_list.append(prefill_expert_outputs[1])
+    if max_num_chunks > num_tokens:
+        hidden_states_chunk_list = hidden_states_chunk_list[:max_num_chunks -
+                                                            num_tokens]
+        if len(shared_chunk_list) > 0:
+            shared_chunk_list = shared_chunk_list[:max_num_chunks - num_tokens]
+    hidden_states = torch.cat(hidden_states_chunk_list, dim=0)
```
ganyi1996ppo: Can we allocate a output buffer at the beginning and index copy those output back to the buffer iterative.

vllm_ascend/quantization/w8a8_dynamic.py
```diff
@@ -215,6 +219,7 @@ def fused_experts_with_mc2(
     w2_scale_bias: torch.Tensor = None,
     quantized_x_for_share: Optional[Any] = None,
     dynamic_scale_for_share: Optional[Any] = None,
+    local_chunk_size: int = 0,
```
whx-sjtu: This `apply_mlp_decode` function uses DeqSwigluQuant which can only be used in decode phase due to memory problems. So I think you shouldn't modify this function because chunked MC2 should never goes into this func.

vllm_ascend/quantization/w8a8_dynamic.py
```diff
@@ -324,6 +332,113 @@ def fused_experts_with_mc2(
...
+                                             max_num_chunks,
+                                             dim=0)
+    topk_ids_chunks = torch.tensor_split(topk_ids, max_num_chunks, dim=0)
+    start_indx = 0
+    end_indx = 0
+    for i in range(len(hidden_states_chunks)):
+        hidden_states_chunk = hidden_states_chunks[i]
+        topk_weights_chunk = topk_weights_chunks[i]
+        topk_ids_chunk = topk_ids_chunks[i]
+        prefill_expert_outputs = fused_experts_with_mc2(
```
whx-sjtu: `fused_experts_with_mc2` calls `apply_mlp_decode` because we only goes into `fused_experts_with_mc2` in decode phase in the past. But now you also use this function to do prefill, so please check the implemention of `fused_experts_with_mc2` carefully.

vllm_ascend/quantization/w8a8_dynamic.py
NNUCJ: ok

PR #1446, opened 2025-06-26
[Core] Fix block table shape to make Prefix cache work with Ascend scheduler

yiz-liu:

This fix the shape of block_table which was introduced by hybrid kv groups several weeks ago.

Error will be raised when enable prefix-cache (eager or not) and Ascend Scheduler at the same time, just send two identical requests and it will reproduce.

v0.9.1: https://github.com/vllm-project/vllm-ascend/pull/1297

No

Test manually


wangxiyuan (approved):

MengqingCao:
Add e2e test in #1505

wangxiyuan (approved):

PR #1452, opened 2025-06-26
Optimize Patch developer guide

zeshengzong:

Fix some terms in the user guide.

No

Following instructions in [here](https://github.com/vllm-project/vllm-ascend/blob/main/docs/README.md), build and test locally.

image
image
image


wangxiyuan:
Thanks for the update

PR #1453, opened 2025-06-26
[CI] Fix nightly benchmark

Potabk:

Sometimes the performance benchmark workflow may fail. We hope to add a prompt when the operation fails and not upload the dirty data of the failed operation.

NO

### DIFF
--- a/.github/workflows/nightly_benchmarks.yaml
+++ b/.github/workflows/nightly_benchmarks.yaml
@@ -197,7 +197,7 @@ jobs:
             --commit_title "$commit_title" \
             --created_at "$commit_time_no_tz" \
             --res_dir ./benchmarks/results \
-            --error_msg $ERROR_MSG \
+            --error $ERROR_MSG \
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly_benchmarks.yaml
```diff
@@ -176,24 +176,28 @@ jobs:
...
             echo "vllm branch: ${{ matrix.vllm_branch }}"
             echo "vllm-ascend branch: ${{ matrix.vllm_ascend_branch }}"
             echo "------------------------"
-            
+
             cd /github/home
-            bash benchmarks/scripts/run-performance-benchmarks.sh
+            ERROR_MSG=""
+            if ! bash benchmarks/scripts/run-performance-benchmarks.sh; then
+              ERROR_MSG="Benchmark failed to run"
```
Yikun, line 191: qq: Failed to run but will filter on frontend, right?

Potabk: yes, will set the status of this commit as `error` and only save some necessary information(commit id, status..) can be filtered by the front end

PR #1468, opened 2025-06-26
【BUGFIXED】Fix the mtp inference error when the prompt is long or Chinese.

Irving11-BKN:
1Revert the code to support PD separation. The bug is fixed and model with mtp can output normally.


wangxiyuan (approved):

wangxiyuan:
CI failure due to transformers version problem

PR #1473, opened 2025-06-27
[0.9.1][bugfix] fix chunked_prefill_mla input for MTP

underfituu:

fix chunked_prefill_mla output for MTP

### DIFF
--- a/tests/long_term/spec_decode_v1/test_v1_mtp_correctness.py
+++ b/tests/long_term/spec_decode_v1/test_v1_mtp_correctness.py
@@ -63,7 +63,10 @@ def test_mtp_correctness(
     with monkeypatch.context() as m:
         m.setenv("VLLM_USE_V1", "1")
 
-        ref_llm = LLM(model=model_name, max_model_len=256, gpu_memory_utilization=0.8, enforce_eager=True)
+        ref_llm = LLM(model=model_name,
+                      max_model_len=256,
+                      gpu_memory_utilization=0.8,
+                      enforce_eager=True)
         ref_outputs = ref_llm.chat(test_prompts, sampling_config)
         del ref_llm
 


vllm_ascend/ops/attention.py
```diff
@@ -262,8 +262,8 @@ def vanilla_chunked_prefill_mla(
 
     attn_output = (attn_output[q_mask].view([-1, num_heads,
                                              v_head_dim]).to(output.dtype))
-    attn_output = attn_output.view_as(output)
-    output.copy_(attn_output)
+    output = output.view([-1, num_heads, v_head_dim])
+    output.copy_(attn_output[:query.size(0) - num_add_query])
```
ganyi1996ppo: Your PR will change the behaviour of the function, please note them at the begining of this function, such as the input shape and output shape. the meaning of the input and output

underfituu: Thanks for the suggestion. Now the logic of MTP has been essentially modified, and the chunked_prefill_mla function remains unchanged.

vllm_ascend/worker/mtp_proposer_v1.py
```diff
@@ -76,32 +77,39 @@ def prepare_inputs(
         # token_indices: [0, 1, ..., a - n1 - 1,
         #                 a, a + 1, ..., a + b - n2 - 1,
         #                 a + b, a + b + 1, ..., a + b + c - n3 - 1]
-
         # [0, a, a + b, a + b + c] -> [a, b, c]
         query_len_per_req = (cu_target_query_lens[1:] -
                              cu_target_query_lens[:-1])
         # [a, b, c] -> [a - n1, b - n2, c - n3]
         num_tokens_per_req = query_len_per_req - num_rejected_tokens
+        if force_one_token:
```
ganyi1996ppo, line 85: We are adding this `force_one_token` is aim for running the first token in consumer node with decode-only style rather than spec-decode style right?

Issue #1298, opened 2025-06-19
[RFC]: Unit test coverage improvement


MengqingCao:
Note: Setup test environment: https://vllm-ascend.readthedocs.io/en/latest/developer_guide/contribution/testing.html

### Motivation

This issue attempt to reduce the gap of unit tests to cover the code. There is a brief architecture of ut in `tests/ut/` already. We need to add more to cover all the code in vllm-ascend, and there are several principles to follow:

1. The overall file tree should be consistent with `vllm_ascend`
2. The file name should be the original file name with a prefix `test_`
3. Use `unittest` framework, make good use of mock
4. The UTs are all running on cpu node, mock the function related to device to host

Please refer to the official doc on [contributing](https://vllm-ascend.readthedocs.io/en/latest/developer_guide/contribution/index.html) and [testing](https://vllm-ascend.readthedocs.io/en/latest/developer_guide/contribution/testing.html) to develop, thanks!
### Unit tests need to add

- [x] |-- test__version.py
- [x] |-- test_ascend_config.py
- [ ] |-- test_attention
- [ ] |   |-- ~~test_attention.py~~ will drop later
- [x] |   |-- test_attention_v1.py https://github.com/vllm-project/vllm-ascend/pull/1529
- [ ] |   `-- test_mla_v1.py @SunnyLee151064
- [ ] |-- compilation
- [ ] |   `-- test_piecewise_backend.py @SunnyLee151064
- [ ] |-- core
- [ ] |   |-- test_schedule_config.py @nuclearwu
- [ ] |   `-- test_scheduler.py @SunnyLee151064
- [ ] |-- device_allocator
- [ ] |   `-- test_camem.py @1024daniel
- [ ] |-- distributed
- [ ] |   |-- test_communicator.py @FieeFlip
- [ ] |   |-- test_device_communicators @Agonixiaoxiao #1601
- [ ] |   |   |-- test_pyhccl.py @SunnyLee151064
- [ ] |   |   `-- test_pyhccl_wrapper.py @SunnyLee151064
- [x] |   |-- kv_transfer @Agonixiaoxiao 
- [x] |   |   |-- test_simple_buffer.py https://github.com/vllm-project/vllm-ascend/pull/1531
- [x] |   |   |-- test_simple_connector.py https://github.com/vllm-project/vllm-ascend/pull/1531
- [x] |   |   |-- test_simple_pipe.py https://github.com/vllm-project/vllm-ascend/pull/1531
[...]

PR #1476, opened 2025-06-27
[Test]Add unit test for platform.py

zhanghw0354:
According to issue #1298 , this pull request adds unit test code for platform.py.

No


zhanghw0354:
@MengqingCao , Please review when you have time. Thank you!

tests/ut/test_platform.py
```diff
@@ -0,0 +1,688 @@
...
+from torch.distributed import ProcessGroup
+from torch.distributed.distributed_c10d import PrefixStore
+from vllm.config import CompilationLevel
+from vllm.platforms import PlatformEnum
+
+from vllm_ascend.platform import NPUPlatform
+from vllm_ascend.utils import ASCEND_QUATIZATION_METHOD
+
+
+class TestNPUPlatform(unittest.TestCase):
```
Yikun: Better to inherit from base class to apply patch:
https://github.com/vllm-project/vllm-ascend/blob/a054f0f4cae72d94db2f1f56c9f5605eb436682e/tests/ut/base.py#L6

tests/ut/test_platform.py
```diff
@@ -0,0 +1,688 @@
...
+    @patch("vllm_ascend.ascend_config.check_ascend_config")
+    @patch("vllm_ascend.ascend_config.init_ascend_config")
+    def test_check_and_update_config_expert_parallel_enabled(
+            self, mock_init_ascend, mock_check_ascend, mock_is_310p):
+        mock_init_ascend.return_value = self.mock_ascend_config
+        self.mock_vllm_config.parallel_config.enable_expert_parallel = True
+        self.mock_vllm_config.parallel_config.tensor_parallel_size = 2
+        self.mock_vllm_config.parallel_config.world_size_across_dp = 4
+
+        from vllm_ascend import platform
```
Yikun, line 266: I guess all `importlib.reload` can be remove if we inherit base classe?

tests/ut/test_platform.py
zhanghw0354: The test class has been updated to inherit from TestBase. Kindly review the changes.

tests/ut/test_platform.py
zhanghw0354: After testing, even with the test class modified to inherit from TestBase, importlib.reload is still required; otherwise, the related unit test cases will continue to fail.

wangxiyuan (approved):

PR #1477, opened 2025-06-27
support pangumoe w8a8c8 and docs

GDzhu01:
support pangu moe w8a8c8

No


wangxiyuan (approved):
Since pangu weight is not avaliable, I'm OK with this PR. Please add test as well once pangu is avaiable.

PR #1478, opened 2025-06-27
[BugFix]Fix bugs when initializing communication groups with dp on 300I Duo

Angazenn:
This PR fixes a bug that use broadcast with cpu_group when running dp. The `broadcast310p` patch will take effects for both cpu_group and device group, but we only need it for device group. Hence a wrapper is added to allow cpu_group use native torch broadcast and it solves the bug. 

No.

With this PR, DP on 310p runs normally and generates reasonable answers.

### DIFF
--- a/vllm_ascend/patch/platform/patch_common/patch_distributed.py
+++ b/vllm_ascend/patch/platform/patch_common/patch_distributed.py
@@ -77,20 +77,28 @@ class NullHandle:
 
 def communication_adaptation_310p():
 
-    def broadcast310p(tensor, src, group=None, async_op=False):
-        rank = torch.distributed.get_rank(group)
-        world_size = torch.distributed.get_world_size(group)
-        tensor_list = [torch.empty_like(tensor) for _ in range(world_size)]
-        tensor_list[rank] = tensor
-        torch.distributed.all_gather(tensor_list, tensor, group=group)
-        tensor[...] = tensor_list[src]
-        if async_op:
-            return NullHandle()
-        else:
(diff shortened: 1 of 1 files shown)


ganyi1996ppo:
Please add more description on this PR, and unittest also if possible

vllm_ascend/patch/platform/patch_common/patch_distributed.py
```diff
@@ -77,20 +77,28 @@ def wait(self):
...
-        torch.distributed.all_gather(tensor_list, tensor, group=group)
-        tensor[...] = tensor_list[src]
-        if async_op:
-            return NullHandle()
-        else:
-            return None
-
-    torch.distributed.broadcast = broadcast310p
-    torch.distributed.distributed_c10d.broadcast = broadcast310p
+    def broadcast310p_wrapper(fn):
```
ganyi1996ppo, line 80: Why rewrite this broadcast for 310P? Shouldn't this been handled by the torch_npu or lower level? its not a wise move to wrap the function with soc name, its neither not general, nor smart.

ganyi1996ppo: Maybe you can consider rewrite the torch_npu's broadcast if you detect your chip version of 310P?

Angazenn: this is because hccl broadcast does not support 310p currently, so we have to patch it to avoid some bugs.

Issue #1470, opened 2025-06-27
TypeError: Qwen2_5_VLProcessor.__init__() got multiple values for argument 'image_processor'


Yikun:
### Your current environment

https://github.com/vllm-project/vllm-ascend/actions/runs/15909593074/job/44873297165
https://github.com/vllm-project/vllm-ascend/actions/runs/15909658129/job/44873510059

### 🐛 Describe the bug

```
  File "/usr/local/python3.10.17/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/local/python3.10.17/lib/python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/__w/vllm-ascend/vllm-ascend/vllm-empty/vllm/v1/engine/core.py", line 519, in run_engine_core
    raise e
  File "/__w/vllm-ascend/vllm-ascend/vllm-empty/vllm/v1/engine/core.py", line 506, in run_engine_core
    engine_core = EngineCoreProc(*args, **kwargs)
  File "/__w/vllm-ascend/vllm-ascend/vllm-empty/vllm/v1/engine/core.py", line 390, in __init__
    super().__init__(vllm_config, executor_class, log_stats,
  File "/__w/vllm-ascend/vllm-ascend/vllm-empty/vllm/v1/engine/core.py", line 76, in __init__
    self.model_executor = executor_class(vllm_config)
  File "/__w/vllm-ascend/vllm-ascend/vllm-empty/vllm/executor/executor_base.py", line 53, in __init__
    self._init_executor()
  File "/__w/vllm-ascend/vllm-ascend/vllm-empty/vllm/executor/uniproc_executor.py", line 47, in _init_executor
    self.collective_rpc("init_device")
  File "/__w/vllm-ascend/vllm-ascend/vllm-empty/vllm/executor/uniproc_executor.py", line 57, in collective_rpc
    answer = run_method(self.driver_worker, method, args, kwargs)
  File "/__w/vllm-ascend/vllm-ascend/vllm-empty/vllm/utils.py", line 2671, in run_method
    return func(*args, **kwargs)
  File "/__w/vllm-ascend/vllm-ascend/vllm-empty/vllm/worker/worker_base.py", line 606, in init_device
    self.worker.init_device()  # type: ignore
  File "/__w/vllm-ascend/vllm-ascend/vllm_ascend/worker/worker_v1.py", line 127, in init_device
[...]

PR #1482, opened 2025-06-27
[CI] Pin transformers<4.53.0 and fix EPLB load_weights to make CI passed

MengqingCao:
### What this PR does / why we need it?

- Fix vLLM EPLB break https://github.com/vllm-project/vllm/commit/e9fd658a736a4d30f7a367c317506c87ad7f5359 by recovering load_weights back to [v0.9.1 version](https://github.com/vllm-project/vllm/commit/07b8fae219b1fff51ef115c38c44b51395be5bb5) temporarily.

- Fix transformers>=4.53.0 image processor break
Related: https://github.com/vllm-project/vllm-ascend/issues/1470

- Mirror torch_npu requirements to pyproject.toml

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
CI passed

PR #1484, opened 2025-06-27
[CI] Pin transformers<4.53.0 

Potabk:

related: https://github.com/vllm-project/vllm-ascend/issues/1470


Yikun:
This will be fixed by https://github.com/vllm-project/vllm-ascend/pull/1482

PR #1491, opened 2025-06-28
disaggregate prefill support v1 schedule

liziyu179:

disaggregate prefill support v1 schedule

### DIFF
--- a/vllm_ascend/core/scheduler_v1.py
+++ b/vllm_ascend/core/scheduler_v1.py
@@ -1,0 +1,512 @@
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


ganyi1996ppo:
Can you point out what exactly the changes you made in this scheduler, that differ from vllm's original ones? There are lots of irrelavant code in this PR. And do we really need a v1 scheduler for v1 scenario?

vllm_ascend/core/scheduler_v1.py
```diff
@@ -0,0 +1,512 @@
...
+        if request.request_id not in self.finished_recving_kv_req_ids:
+            return False
+        assert len(self.kv_cache_config.kv_cache_groups
+                   ) == 1, "KV connector only supports one KV cache group now"
+        # Now that the blocks are ready, actually cache them.
+        # In order to make decode node always do the decode step, we transfer every block as long as it contains the
+        # data computed by prefill node.
+        num_computed_tokens = request.num_tokens
+        if num_computed_tokens == request.num_tokens:
+            num_computed_tokens -= 1
```
jianzs, line 501: Use the ascend scheduler if you only need this change. 

btw, I think the `num_computed_tokens` should be set to `request.num_tokens - 1`.

PR #1492, opened 2025-06-28
[BugFix] Fix accuray bug of prefix-caching.

whx-sjtu:
When use AscendScheduler with prefix-cache enabled and chunk-prefill disabled, there will be accuray problem because there is no branch in mla_v1 to process this scenario. This PR fixes it.


ganyi1996ppo (approved):

MengqingCao:
Thanks for this fix, please also backport on main

PR #1498, opened 2025-06-28
[BugFix] Fix prefix cache accuracy bug

whx-sjtu:
Port to main


Yikun:
Thanks for fixing it, how can we prevent this precision regresion break again? And please fulfill commit msg

MengqingCao:
Add e2e test in #1505

wangxiyuan (approved):

PR #1501, opened 2025-06-28
[Fix] Add warning for incompatible Ray backend with ACL Graph mode

yiz-liu:

Currently, Ray is not compatible with ACL Graph, so we need to fall back to eager mode when using the Ray backend.


Yikun (approved):
Please update commit msg and add UT

yiz-liu:
> Please update commit msg and add UT

Done and done, please review at your earliest convenience, thanks. @Yikun

PR #1504, opened 2025-06-28
[Doc] Refactor env vars module and change the way of doc generation

shen-shanshan:

Refactor env vars module and change the way of doc generation.

I have encapsulated each env var into a object, and then we can get the doc info from it.

Whenever we update the env vars dict, we just need to run the script directly to generate the latest doc.


wangxiyuan:

PR #1513, opened 2025-06-30
[CI] change to new ds model

wangxiyuan:
Previous, the DeepSeek V3 Pruning weight is not correct, the moe layer is not tested. We update a new Pruning model to enable moe layer compute. 

This PR fix the CI to address the new weight.

### DIFF
--- a/.github/workflows/vllm_ascend_test.yaml
+++ b/.github/workflows/vllm_ascend_test.yaml
@@ -66,238 +66,238 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  # lint:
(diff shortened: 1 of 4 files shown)


wangxiyuan:
test passed: https://github.com/vllm-project/vllm-ascend/actions/runs/15962737330/job/45017750566?pr=1513

I'll change to the new model directly in the next commit

wangxiyuan:
test should pass once the weight is updated in modelscope. I'll recheck later

Yikun (approved):
Recover CI first also cc @jianzs @ganyi1996ppo This is an exception case need quick merge.

PR #1518, opened 2025-06-30
add kv transfer unittest

Agonixiaoxiao:

test kv data transfer contains connect,pipe,buffer

N/A


xudongLi-cmss:
@Yikun please review my UT

PR #1524, opened 2025-06-30
[Benchmark] Refactor perf script to use benchmark cli

Potabk:

Since, `vllm bench` cli has optimized enough for production use(support more datasets), we are now do not need to copy vllm codes, now , with vllm installed, we can easily use the benchmark cli
No

CI passed

### DIFF
--- a/.github/workflows/nightly_benchmarks.yaml
+++ b/.github/workflows/nightly_benchmarks.yaml
@@ -115,6 +115,7 @@ jobs:
         env:
           PIP_EXTRA_INDEX_URL: https://mirrors.huaweicloud.com/ascend/repos/pypi
         run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/nightly_benchmarks.yaml
```diff
@@ -115,6 +115,7 @@ jobs:
         env:
           PIP_EXTRA_INDEX_URL: https://mirrors.huaweicloud.com/ascend/repos/pypi
         run: |
+          pip install "transformers<=4.52.4"
```
Yikun, line 118: This should be fixed in future

Yikun (approved):

Yikun:
let's do quick merge to recover benchmark CI

PR #1530, opened 2025-06-30
[bugfix] fix deepseek memory bug

zzzzwwjj:


vllm_ascend/ops/rotary_embedding.py
```diff
@@ -266,7 +266,7 @@ def deepseek_rope_init_func(
     super(DeepseekScalingRotaryEmbedding,
           self).__init__(head_size, rotary_dim, max_position_embeddings, base,
                          is_neox_style, dtype)
-    self.max_seq_len = max_position_embeddings
+    self.max_seq_len = max_position_embeddings * scaling_factor
     _set_cos_sin_cache(self,
                        max_position_embeddings,
```
ganyi1996ppo: There seems still have some problems, can you rewrite it here carefully.

vllm_ascend/ops/rotary_embedding.py
```diff
@@ -266,7 +266,7 @@ def deepseek_rope_init_func(
     super(DeepseekScalingRotaryEmbedding,
           self).__init__(head_size, rotary_dim, max_position_embeddings, base,
                          is_neox_style, dtype)
-    self.max_seq_len = max_position_embeddings
+    self.max_seq_len = max_position_embeddings * scaling_factor
```
whx-sjtu: What does this have to do with the memory problem?

zzzzwwjj: we check this param in `native_rope_deepseek_forward`, and when it is smaller than current seqlen, it will re-generate sin_cos_cache for one time, so it will be generated every step when long input scene.

PR #1536, opened 2025-06-30
[WIP] Add schedule_config.py test cases

zeshengzong:

### DIFF
--- a/tests/ut/core/test_schedule_config.py
+++ b/tests/ut/core/test_schedule_config.py
@@ -1,0 +1,35 @@
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


tests/ut/core/test_schedule_config.py
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# This file is a part of the vllm-ascend project.
+#
+
+import unittest
+from vllm_ascend.core.schedule_config import AscendSchedulerConfig
+from vllm.config import SchedulerConfig
+
+class TestAscendSchedulerConfig(unittest.TestCase):
```
Yikun: Better to inherit from base class to apply patch:
https://github.com/vllm-project/vllm-ascend/blob/a054f0f4cae72d94db2f1f56c9f5605eb436682e/tests/ut/base.py#L6

PR #1545, opened 2025-07-01
[V0.9.1] Use AddRmsNormQuant ops in the custom model to optimize Qwen3's performance

rjg-lyh:
Optimizes the performance of the Qwen3 quantization model by registering a custom model and adding the AddRmsNormQuant operation. Subsequent PRs will focus on performance optimizations based on this custom model.

No.

CI passed with existing test.

### DIFF
--- a/vllm_ascend/models/qwen3.py
+++ b/vllm_ascend/models/qwen3.py
@@ -1,0 +1,156 @@
+from collections.abc import Iterable
+from typing import Optional, Union
+
+import torch
+from torch import nn
+from transformers import Qwen3Config
+from vllm.compilation.decorators import support_torch_compile
+from vllm.config import CacheConfig, VllmConfig
+from vllm.distributed import get_pp_group
+from vllm.model_executor.layers.logits_processor import LogitsProcessor
+from vllm.model_executor.layers.quantization import QuantizationConfig
+from vllm.model_executor.layers.vocab_parallel_embedding import ParallelLMHead
+from vllm.model_executor.models.interfaces import SupportsLoRA, SupportsPP
+from vllm.model_executor.models.qwen2 import Qwen2Model
+from vllm.model_executor.models.qwen3 import Qwen3DecoderLayer
+from vllm.model_executor.models.utils import (AutoWeightsLoader,
+                                              PPMissingLayer, maybe_prefix)
+from vllm.model_executor.sampling_metadata import SamplingMetadata
+from vllm.sequence import IntermediateTensors
+
+from vllm_ascend.ops.layernorm import AddRMSNormQuant
+
+
+class CustomQwen3DecoderLayer(Qwen3DecoderLayer):
+
+    def __init__(
+        self,
+        config: Qwen3Config,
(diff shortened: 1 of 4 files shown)


vllm_ascend/quantization/w8a8.py
```diff
@@ -59,6 +60,7 @@ def get_pertensor_param(params_dtype: torch.dtype) -> Dict[str, Any]:
         params_dict = {}
         params_dict["input_scale"] = torch.empty(1, dtype=params_dtype)
         params_dict["input_offset"] = torch.empty(1, dtype=torch.int8)
+        AscendW8A8LinearMethod.params_dtype = params_dtype
```
ganyi1996ppo: What if there is a fp16 fallback？Then how dose that fallback linear do the calculation

ganyi1996ppo: I don't think its a good solution, can we write this dtype back to the param_dict and inject it into the layer eventually?

vllm_ascend/models/qwen3.py
```diff
@@ -0,0 +1,143 @@
...
+        prefix: str = "",
+    ) -> None:
+        super().__init__(config=config,
+                         cache_config=cache_config,
+                         quant_config=quant_config,
+                         prefix=prefix)
+        if quant_config is not None:
+            from vllm_ascend.quantization.quant_config import AscendQuantConfig
+            assert isinstance(quant_config, AscendQuantConfig)
+            self.input_layernorm = AddRMSNormQuant(config.hidden_size,
```
ganyi1996ppo: Just discussed with @realliujiaxu , this behaviour is not a general way to apply our optimization in modeling, can we try to leverage the compilation path in vllm to fuse ops in fx graph? cc @jianzs @Yikun @wangxiyuan

ganyi1996ppo: I'm fine with the changes in this PR, but I believe we also need an ultimate solution to handle this kind of problem for good.

vllm_ascend/ops/layernorm.py
```diff
@@ -21,6 +21,47 @@
...
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        residual: Optional[torch.Tensor] = None,
+    ) -> Union[torch.Tensor, tuple[torch.Tensor, torch.Tensor]]:
+        import torch_npu
+
+        if residual is not None:
+            x, _, residual = torch_npu.npu_add_rms_norm_quant(
```
realliujiaxu, line 51: does `torch_npu.npu_add_rms_norm_quant` require a newer version of torch_npu?

vllm_ascend/quantization/w8a8.py
rjg-lyh: I have fix it.

vllm_ascend/ops/layernorm.py
rjg-lyh: Now version of PTA has supported it.

ganyi1996ppo (approved):

PR #1100, opened 2025-06-06
[Bugfix] Reduce _npu_flash_attention mask to 128x128 for memory savings

ApsarasX:
### What this PR does / why we need it?
Avoid generating a special `attn_mask` (which theoretically could be as large as `bf16[max_model_len, max_model_len]`) when `chunked-prefill` is not enabled and the user input is extremely long.

However, this PR depends on [modifications to the `torch_npu._npu_flash_attention`](https://gitee.com/randler/op-plugin/compare/5056055ef832d2b0d9c5375d63bc9b14a5f5e297...f39454096ab5954f7da6fd2225688b67824eb7d8)

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
No
- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/ebf7605b0dd58ff5d572d1918e52ca732025eee0

PR #1551, opened 2025-07-01
[bugfix] fix deepseek memory bug

zzzzwwjj:

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -350,6 +350,7 @@ class NPUModelRunner(LoRAModelRunnerMixin):
         self.torchair_graph_enabled = ascend_config.torchair_graph_config.enabled and self.vllm_config.model_config.use_mla
         self.use_cached_npu_graph = ascend_config.torchair_graph_config.use_cached_graph
         self.torchair_graph_batch_sizes = ascend_config.torchair_graph_config.graph_batch_sizes
+        self.use_ring_mla = ascend_config.chunked_prefill_for_mla
 
         if ascend_config.torchair_graph_config.graph_batch_sizes_init:
             self.init_torchair_graph_batch_sizes()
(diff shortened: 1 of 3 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -908,11 +909,14 @@ def _process_reqs(
...
             attn_state = AscendAttentionState.PrefillCacheHit
 
-        attn_mask = self._make_attention_mask(seq_lens=seq_lens,
-                                              query_lens=num_scheduled_tokens,
-                                              position=positions,
-                                              attn_state=attn_state)
-        self.attn_mask = attn_mask
+        # NOTE: when use ring_mla, attn_mask don't need to generate here.
+        if not self.use_ring_mla or attn_state == AscendAttentionState.PrefillNoCache:
+            attn_mask = self._make_attention_mask(
```
ApsarasX, line 914: Can a compressed mask be used? If the sequence is too long, it might cause memory waste here.

Refer https://github.com/vllm-project/vllm-ascend/pull/1100

zzzzwwjj: ringmla's mask size is equal to chunksize, it won't cause memory waste.

ganyi1996ppo:
Looks good, please cherry-pick this change back to main

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -350,6 +350,7 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
         self.torchair_graph_enabled = ascend_config.torchair_graph_config.enabled and self.vllm_config.model_config.use_mla
         self.use_cached_npu_graph = ascend_config.torchair_graph_config.use_cached_graph
         self.torchair_graph_batch_sizes = ascend_config.torchair_graph_config.graph_batch_sizes
+        self.use_ring_mla = ascend_config.chunked_prefill_for_mla
```
jianzs, line 353: Should we consider which scheduler is be using?

```python
self.use_ring_mla = ascend_config.chunked_prefill_for_mla or \
            not ascend_config.ascend_scheduler_config.enabled
```

PR #1552, opened 2025-07-01
[BugFix] disaggregate prefill force producer dummy run with prefill

liziyu179:

Disaggregate prefill force producer dummy run with prefill

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -372,6 +372,11 @@ class NPUModelRunner(LoRAModelRunnerMixin):
         # NOTE: we need to use `in_profile_run` to determine whether `enable_force_load_balance` is True
         self.in_profile_run = False
 
+        # kv role
+        self.is_kv_producer = False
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -372,6 +372,12 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
         # NOTE: we need to use `in_profile_run` to determine whether `enable_force_load_balance` is True
         self.in_profile_run = False
 
+        # kv role
+        self.is_kv_producer = False
+        if vllm_config.kv_transfer_config is not None:
+            self.is_kv_producer = (
+                vllm_config.kv_transfer_config.kv_role == "kv_producer")
+
```
jianzs: Why not use the `vllm_config.kv_transfer_config.is_kv_producer` ?

liziyu179: Thank you for your suggestion. I have made changes based on your comments. If you have any other suggestions, please feel free to comment.

PR #1555, opened 2025-07-01
perf: use multicast to avoid padding decode request to prefill size

NeverRaR:

perf: use multicast to avoid padding decode request to prefill size

before:
image
after:
image


vllm_ascend/ops/fused_moe.py
```diff
@@ -1260,8 +1276,16 @@ def forward(self,
                         router_logits = nn.functional.pad(
                             router_logits,
                             (0, 0, 0, max_num_tokens_across_dp - num_tokens))
-            hidden_states = get_dp_group().all_gather(hidden_states, 0)
-            router_logits = get_dp_group().all_gather(router_logits, 0)
+            if is_prefill:
+                cu_tokens_across_dp_cpu = get_forward_context(
+                ).dp_metadata.cu_tokens_across_dp_cpu
+                hidden_states = self.naive_multicast(hidden_states,
```
jianzs: Do we need to change the selection in the `utils.py`?

vllm_ascend/ops/fused_moe.py
```diff
@@ -1300,13 +1324,22 @@ def forward(self,
...
+            if is_prefill:
+                start = 0 if self.dp_rank == 0 else cu_tokens_across_dp_cpu[
+                    self.dp_rank - 1]
+                end = cu_tokens_across_dp_cpu[self.dp_rank]
+                final_hidden_states = get_dp_group().all_reduce(
+                    e_hidden_states)
+                final_hidden_states = final_hidden_states[start:end, :]
+                dispose_tensor(e_hidden_states)
+            else:
+                final_hidden_states = dist._functional_collectives.reduce_scatter_tensor(
```
ApsarasX, line 1362: Do not use this `dist._functional_collectives.reduce_scatter_tensor`, as it may potentially cause memory leaks.

NeverRaR: I prefer to apply the fix in another PR

vllm_ascend/ops/fused_moe.py
NeverRaR: Added new state `FusedMoEState.NaiveMulticast`, but the current logic is a bit too complex. Perhaps we should consider better state naming in the future

vllm_ascend/ops/fused_moe.py
```diff
@@ -1205,6 +1207,22 @@ def __init__(
...
+        assert (len(x.shape) == 2)
+        buffer = torch.empty((cu_tokens_across_dp_cpu[-1], x.size(1)),
+                             device=x.device,
+                             dtype=x.dtype)
+        start = 0 if self.dp_rank == 0 else cu_tokens_across_dp_cpu[
+            self.dp_rank - 1]
+        end = cu_tokens_across_dp_cpu[self.dp_rank]
+        buffer[start:end, :].copy_(x)
+        for idx in range(self.dp_size):
+            start = 0 if idx == 0 else cu_tokens_across_dp_cpu[idx - 1]
```
jianzs, line 1241: Is using vLLM's `NaiveAll2AllManager` a better choice?

NeverRaR: Directly using `NaiveAll2AllManager` and rewriting the implementation of NPUCommuniator will lead to difficulties in optimizing  router_logits

ApsarasX (approved):

NeverRaR:
@jianzs please take a look

jianzs:
@NeverRaR  Since the upstream fuse moe  changed, please rebase to ensure the feature works correctly.

PR #502, opened 2025-04-10
[v0.7.3]support MTP in deepseek w8a8 quant model

mengwei805:
### What this PR does / why we need it?
Add support MTP in deepseek w8a8 quant model.

### Does this PR introduce _any_ user-facing change?
1. The quantized MTP layer of deepseek on the current NPU msmodelslim is not quantized, So the MTP layer in deepseek w8a8 quantization weight is still in bflaot16 format;
2. The description file generated by the current msmodelslim tool does not have MTP layer information. Please manually add it to `quantization_config` in `config.json` and set the value to `FLOAT`.

### How was this patch tested?
local tested

PR #940, opened 2025-05-23
[BugFix] Fix data parallel

yiz-liu:
### What this PR does / why we need it?

With this PR, we can migrate to the native `data_parallel.py` in vllm examples and remove the version in vllm-ascend.

At present, `ASCEND_RT_VISIBLE_DEVICES` introduces considerable difficulties; therefore, we must employ a temporary workaround and manually specify the device.

PR #1568, opened 2025-07-01
[P/D] Mooncake Connector for v1 distributed

LCAIZJ:
We implemented a v1 version of the connector for Mooncake, which utilizes Mooncake's TransferEngine with Ascend Transport（Mooncake[#502](https://github.com/kvcache-ai/Mooncake/pull/502)）.

No

Tested on DeepSeek-R1-W8A8


LCAIZJ:
### Change log
07/03: change transfer_sync_read to batch_transfer_sync_read 
07/05: solve some format and mypy problem
07/06: rename MooncakeConnectorV1

jianzs:
@LCAIZJ  plz make ci happy

zzy-ContiLearn:
> @LCAIZJ plz make ci happy

god bless we

LCAIZJ:
@jianzs CI is happy

LCAIZJ:
solve conflicts

jianzs:
@LCAIZJ  please rebase to latest commit, and test

btw, all your dependencies are ready?

LCAIZJ:
> @LCAIZJ please rebase to latest commit, and test
> 
> btw, all your dependencies are ready?

Yes

jianzs:
@LCAIZJ  A user guide for this connector, including installation for Mooncake Ascend, is preferable. Pengbo was supposed to create this tutorial - do you know how that's progressing?

LCAIZJ:
> @LCAIZJ A user guide for this connector, including installation for Mooncake Ascend, is preferable. Pengbo was supposed to create this tutorial - do you know how that's progressing?

We still have a few things to do:
- Tutorial is on track to be completed this week.
- Unit tests will be finalized by next Wednesday.
- The recently merged random port feature in Mooncake main has implications for our design. We're targeting completion of necessary adjustments by end of this week.

LCAIZJ:
> > @LCAIZJ A user guide for this connector, including installation for Mooncake Ascend, is preferable. Pengbo was supposed to create this tutorial - do you know how that's progressing?
> 
> We still have a few things to do:
> 
> * Tutorial is on track to be completed this week.
> * Unit tests will be finalized by next Wednesday.
> * The recently merged random port feature in Mooncake main has implications for our design. We're targeting completion of necessary adjustments by end of this week.

All work has been completed @jianzs

examples/disaggregated_prefill_v1/mooncake_connector_deployment_guide.md
```diff
@@ -0,0 +1,184 @@
+# Mooncake connector deployment Guide
+
+## Environmental Dependencies
+
+ *  Software:
+     *  Python >= 3.9, < 3.12
+     *  CANN >= 8.2.rc1
+     *  PyTorch >= 2.5.1, torch-npu >= 2.5.1.post1.dev20250619
```
wangxiyuan: torch-npu has been upgrade to 2.7.1 on main branch  https://github.com/vllm-project/vllm-ascend?tab=readme-ov-file#prerequisites

and the version of mooncake package should be added as well

examples/disaggregated_prefill_v1/mooncake_connector_deployment_guide.md
```diff
@@ -0,0 +1,184 @@
...
+     *  vLLM (same version as vllm-ascend)
+     *  mooncake-transfer-engine reference documentation: https://github.com/kvcache-ai/Mooncake/blob/main/doc/zh/ascend_transport.md
+
+The vllm version must be the same as the main branch of vllm-ascend, for example, 2025/07/30. The version is
+
+ *  vllm: v0.10.0
+ *  vllm-ascend: main branch ( as of 25/07/30)
```
wangxiyuan: this should be the next release tag

vllm: v0.10.1
vllm-ascend: v0.10.1rc1

examples/disaggregated_prefill_v1/mooncake_connector_deployment_guide.md
```diff
@@ -0,0 +1,184 @@
...
+
+```
+export HCCL_EXEC_TIMEOUT=204
+export HCCL_CONNECT_TIMEOUT=120
+export HCCL_IF_IP=localhost
+export GLOO_SOCKET_IFNAME="xxxxxx"
+export TP_SOCKET_IFNAME="xxxxxx"
+export HCCL_SOCKET_IFNAME="xxxxxx"
+export ASCEND_RT_VISIBLE_DEVICES=0,1,2,3
+export VLLM_USE_V1=1
```
wangxiyuan: `VLLM_USE_V1` is not need any more

examples/disaggregated_prefill_v1/mooncake_connector_deployment_guide.md
```diff
@@ -0,0 +1,184 @@
...
+             }
+      }
+  }'
+```
+
+`HCCL_EXEC_TIMEOUT`, `HCCL_CONNECT_TIMEOUT`, and `HCCL_IF_IP` are hccl-related configurations.<br>
+Set `GLOO_SOCKET_IFNAME`, `TP_SOCKET_IFNAME`, and `HCCL_SOCKET_IFNAME` to the corresponding NIC.<br>
+`ASCEND_RT_VISIBLE_DEVICES` specifies the cards on which the node run resides. The total number of cards equals `dp_size*tp_size`.<br>
+`BUDGET_CONFIG_PATH` specifies the path of the mooncake.json file.<br>
+Set `VLLM_USE_V1` to 1.<br>
```
wangxiyuan: ditto

examples/disaggregated_prefill_v1/mooncake_connector_deployment_guide.md
LCAIZJ: The version issue with the Mooncake package is mainly related to hardware-specific compilation configurations. Currently, Mooncake requires adjusting compile-time parameters based on different hardware platforms. For instance, when using Ascend 910B/910C NPUs, you need to enable the Ascend transport module by setting the USE_ASCEND option: `option(USE_ASCEND "Option to enable NPU support" ON)` . This configuration should be set in the CMake build system (refer to [common.cmake#L62](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-common/common.cmake#L62)).

The default compilation settings in the official Mooncake repository have USE_ASCEND disabled. Users targeting Ascend hardware must manually enable this option and build the package from source following the instructions in the [Ascend Transport Configuration Guide](https://github.com/kvcache-ai/Mooncake/blob/main/doc/zh/ascend_transport.md).

LCAIZJ:
@wangxiyuan Apologies, the code review comments were lost due to a force-push. Implemented the majority of your suggestions, and Mooncake version specifics need elaboration. 
The version issue with the Mooncake package is mainly related to hardware-specific compilation configurations. Currently, Mooncake requires adjusting compile-time parameters based on different hardware platforms. For instance, when using Ascend 910B/910C NPUs, you need to enable the Ascend transport module by setting the USE_ASCEND option: `option(USE_ASCEND "Option to enable NPU support" ON)` . This configuration should be set in the CMake build system (refer to [common.cmake#L62](https://github.com/kvcache-ai/Mooncake/blob/main/mooncake-common/common.cmake#L62)).
The default compilation settings in the official Mooncake repository have USE_ASCEND disabled. Users targeting Ascend hardware must manually enable this option and build the package from source following the instructions in the [Ascend Transport Configuration Guide](https://github.com/kvcache-ai/Mooncake/blob/main/doc/zh/ascend_transport.md).

vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -0,0 +1,1051 @@
...
+            "MooncakeConnector get_num_new_matched_tokens: "
+            "num_computed_tokens=%s, kv_transfer_params=%s",
+            num_computed_tokens, params)
+
+        if params is not None and params.get("do_remote_prefill"):
+            assert num_computed_tokens == 0, "Currently only support " \
+                                             "prefill with num_computed_tokens == 0."
+            # Assume that the request's KV cache is already fully prefilled and
+            # can be fetched entirely from the prefill node.
+            count = max(len(request.prompt_token_ids) - 1, 0)
```
ganyi1996ppo, line 631: This line looks strange, the external computed token should not include the local computed token? Or you just not support prefix cache on the decode instance?

vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -0,0 +1,1051 @@
...
+            "MooncakeConnector update_state_after_alloc: "
+            "num_external_tokens=%s, kv_transfer_params=%s",
+            num_external_tokens, params)
+
+        if params is not None and params.get("do_remote_prefill"):
+            if params.get("remote_block_ids"):
+                if all(p in params for p in ("remote_engine_id", "remote_host",
+                                             "remote_port")):
+                    local_block_ids = (blocks.get_unhashed_block_ids()
+                                       if num_external_tokens > 0 else [])
```
ganyi1996ppo, line 653: Just out of curiosity, is there any chance for disaggregated pd that a remote computed request dose not have `num_external_tokens`?

vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -0,0 +1,1051 @@
...
+            logger.info("Delaying free of %d blocks for request %s",
+                        len(computed_block_ids), request.request_id)
+        return delay_free_blocks, dict(
+            do_remote_prefill=True,
+            do_remote_decode=False,
+            remote_block_ids=computed_block_ids,
+            remote_engine_id=self.engine_id,
+            remote_host=self.side_channel_host,
+            remote_port=self.side_channel_port,
+            last_token_id=request.output_token_ids[-1],
```
ganyi1996ppo: So mooncake just support returning the first token from prefiller's output? But how dose the decode instance use this attributed? Seems no usage for this attr

jianzs: Thank you for raising this concern. This is on me. We've made internal modifications to the API server where the decoder appends the token ID to the original prompt token IDs upon receiving a request. We need to consider how to properly adapt this in the open-source version. Maybe, we could have the decoder directly return the first token?

vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -0,0 +1,1051 @@
...
+        remote_transfer_port = self.remote_te_port[remote_engine_id][
+            remote_handshake_port]
+        session_id = f"{remote_host}:{remote_transfer_port}"
+        src_list, dst_list, length_list = [], [], []
+        for k, (src_layer_base_addr, dst_layer_base_addr) in enumerate(
+                zip(local_kv_caches_base_addrs, remote_kv_caches_base_addrs)):
+            block_len = (self.block_len[k % 2]
+                         if self.use_mla else self.block_len[0])
+            for i, remote_block_id in enumerate(grouped_remote_block_ids):
+                local_block_ids = grouped_local_block_ids[i]
```
ganyi1996ppo, line 372: So you grouped the block ids to make sure you pull blocks in a locality friendly way? How much improvements it gets? Normally, the block ids should be already sequential in most scenario, just wonder if the cpu overhead of contiguous rearange can be actually beaten by its transfer benefits.

vllm_ascend/distributed/mooncake_connector.py
jianzs: Correct. For this implementation, we assume two things: decode node without prefix cache enabled, and prefill kv cache will not use block size truncation.

ganyi1996ppo: Why? prefix cache is almost some kind of free lunch, it should bring no pain in any scenario

vllm_ascend/distributed/mooncake_connector.py
jianzs: Although we haven't measured the specific benefits of block grouping, our experiments show that blocks are often non-sequential. Due to block sizes tending to be modest, it seems that the computational overhead for reordering operations should be negligible?

vllm_ascend/distributed/mooncake_connector.py
ganyi1996ppo: Sounds good, I think its worth to try

vllm_ascend/distributed/mooncake_connector.py
jianzs: Good suggestion. While we typically consider prefix cache and KV pool as a coupled solution, enabling GPU prefix cache alone could be a great option for decode nodes. We'll explore this approach.

vllm_ascend/distributed/mooncake_connector.py
LCAIZJ: You're correct. In our current implementation, decode node without prefix cache enabled, so the scenario where num_external_tokens = 0 does not occur. If remote_blocks and num_external_tokens = 0, we have a full prefix cache hit on the D worker. We need to call send_notif in _read_blocks to free the memory on the P.

Shichang-Zhang:
@LCAIZJ Hello, thanks for your contribution! Recently I'am trying to run vllm v1 + mooncake with mooncake connector on NPU. I built an image of your PR with vllm-ascend v0.10.1 and Ascend_transport. I tried to run it on K8s and assgined 4 of 8 910B NPU to the pod:
```
        resources:
          limits:
            huawei.com/Ascend910: "4"
          requests:
            huawei.com/Ascend910: "4"
```
Then I entered the container and exported ENV variables `ASCEND_RT_VISIBLE_DEVICES`=2,3,4,5 (K8s assigns NPU resource randomly).
When I start command to run vllm: 
```
JSON_CONTENT="{\"local_hostname\": \"$POD_IP\", \"metadata_server\": \"redis://10.250.39.62:6379\",\"protocol\": \"tcp\",\"device_name\": \"\",\"master_server_address\": \"10.250.39.46:50001\",\"protocol\":\"ascend\"}"
          echo "$JSON_CONTENT" > /app/mooncake.json
          MOONCAKE_CONFIG_PATH=/app/mooncake.json VLLM_USE_V1=1 python3 -m vllm.entrypoints.openai.api_server --model Qwen/Qwen2.5-7B-Instruct --port 8100 --tensor-parallel-size 2 --max-model-len 10000 --data-parallel-size 2 --data-parallel-address localhost --data-parallel-rpc-port 9100 --kv-transfer-config '{"kv_connector":"MooncakeConnectorV1","kv_role":"kv_producer","kv_buffer_device":"npu","kv_connector_module_path":"vllm_ascend.distributed.mooncake_connector","kv_parallel_size": 1,"kv_port": "20001","engine_id": "0","kv_rank": 0,"kv_connector_extra_config":{"prefill":{"tp_size":2,"dp_size":2},"decode":{"tp_size":2,"dp_size":2}}}'
```
I encountered the problem from torch_npu(v2.7.1): 
vllm-mooncake-torchnpu-bug1
I think in the containter torch_npu could only see the assigned NPU device with logicId ranged [0-4), but the possible physical id in ENV variables `ASCEND_RT_VISIBLE_DEVICES` ranged [0,8). 
But if ENV variables `ASCEND_RT_VISIBLE_DEVICES` is not assigend, in the mooncake connecotr initialization stage, the NPU device id will be automatically 0, which is the wrong physical device id, and cause the process try to connect an invisible NPU device. So I wonder should it to use ranktable to get the physical-logical id mapping instead of directly reading the ENV variables `ASCEND_RT_VISIBLE_DEVICES`.

zzy-ContiLearn:
> @LCAIZJ Hello, thanks for your contribution! Recently I'am trying to run vllm v1 + mooncake with mooncake connector on NPU. I built an image of your PR with vllm-ascend v0.10.1 and Ascend_transport. I tried to run it on K8s and assgined 4 of 8 910B NPU to the pod:
> 
> ```
>         resources:
>           limits:
>             huawei.com/Ascend910: "4"
>           requests:
>             huawei.com/Ascend910: "4"
> ```
> 
> Then I entered the container and exported ENV variables `ASCEND_RT_VISIBLE_DEVICES`=2,3,4,5 (K8s assigns NPU resource randomly). When I start command to run vllm:
> 
> ```
> JSON_CONTENT="{\"local_hostname\": \"$POD_IP\", \"metadata_server\": \"redis://10.250.39.62:6379\",\"protocol\": \"tcp\",\"device_name\": \"\",\"master_server_address\": \"10.250.39.46:50001\",\"protocol\":\"ascend\"}"
>           echo "$JSON_CONTENT" > /app/mooncake.json
>           MOONCAKE_CONFIG_PATH=/app/mooncake.json VLLM_USE_V1=1 python3 -m vllm.entrypoints.openai.api_server --model Qwen/Qwen2.5-7B-Instruct --port 8100 --tensor-parallel-size 2 --max-model-len 10000 --data-parallel-size 2 --data-parallel-address localhost --data-parallel-rpc-port 9100 --kv-transfer-config '{"kv_connector":"MooncakeConnectorV1","kv_role":"kv_producer","kv_buffer_device":"npu","kv_connector_module_path":"vllm_ascend.distributed.mooncake_connector","kv_parallel_size": 1,"kv_port": "20001","engine_id": "0","kv_rank": 0,"kv_connector_extra_config":{"prefill":{"tp_size":2,"dp_size":2},"decode":{"tp_size":2,"dp_size":2}}}'
> ```
> 
> I encountered the problem from torch_npu(v2.7.1): vllm-mooncake-torchnpu-bug1 I think in the containter torch_npu could only see the assigned NPU device with logicId ranged [0-4), but the possible physical id in ENV variables `ASCEND_RT_VISIBLE_DEVICES` ranged [0,8). But if ENV variables `ASCEND_RT_VISIBLE_DEVICES` is not assigend, in the mooncake connecotr initialization stage, the NPU device id will be automatically 0, which is the wrong physical device id, and cause the process try to connect an invisible NPU device. So I wonder should it to use ranktable to get the physical-logical id mapping instead of directly reading the ENV variables `ASCEND_RT_VISIBLE_DEVICES`.

I also encountered the same issue. The device_id assigned by Kubernetes might start counting from 0 by default. In our connector, we currently read the device_id from the OS environment variables, it may raise error when quering device_ip in hccn_conf file. A possible solution is to add a new environment variable in vLLM (e.g., VLLM_DEVICE_ID) and then retrieve the device_id from vllm.envs.

LCAIZJ:
> @LCAIZJ Hello, thanks for your contribution! Recently I'am trying to run vllm v1 + mooncake with mooncake connector on NPU. I built an image of your PR with vllm-ascend v0.10.1 and Ascend_transport. I tried to run it on K8s and assgined 4 of 8 910B NPU to the pod:
> 
> ```
>         resources:
>           limits:
>             huawei.com/Ascend910: "4"
>           requests:
>             huawei.com/Ascend910: "4"
> ```
> 
> Then I entered the container and exported ENV variables `ASCEND_RT_VISIBLE_DEVICES`=2,3,4,5 (K8s assigns NPU resource randomly). When I start command to run vllm:
> 
> ```
> JSON_CONTENT="{\"local_hostname\": \"$POD_IP\", \"metadata_server\": \"redis://10.250.39.62:6379\",\"protocol\": \"tcp\",\"device_name\": \"\",\"master_server_address\": \"10.250.39.46:50001\",\"protocol\":\"ascend\"}"
>           echo "$JSON_CONTENT" > /app/mooncake.json
>           MOONCAKE_CONFIG_PATH=/app/mooncake.json VLLM_USE_V1=1 python3 -m vllm.entrypoints.openai.api_server --model Qwen/Qwen2.5-7B-Instruct --port 8100 --tensor-parallel-size 2 --max-model-len 10000 --data-parallel-size 2 --data-parallel-address localhost --data-parallel-rpc-port 9100 --kv-transfer-config '{"kv_connector":"MooncakeConnectorV1","kv_role":"kv_producer","kv_buffer_device":"npu","kv_connector_module_path":"vllm_ascend.distributed.mooncake_connector","kv_parallel_size": 1,"kv_port": "20001","engine_id": "0","kv_rank": 0,"kv_connector_extra_config":{"prefill":{"tp_size":2,"dp_size":2},"decode":{"tp_size":2,"dp_size":2}}}'
> ```
> 
> I encountered the problem from torch_npu(v2.7.1): vllm-mooncake-torchnpu-bug1 I think in the containter torch_npu could only see the assigned NPU device with logicId ranged [0-4), but the possible physical id in ENV variables `ASCEND_RT_VISIBLE_DEVICES` ranged [0,8). But if ENV variables `ASCEND_RT_VISIBLE_DEVICES` is not assigend, in the mooncake connecotr initialization stage, the NPU device id will be automatically 0, which is the wrong physical device id, and cause the process try to connect an invisible NPU device. So I wonder should it to use ranktable to get the physical-logical id mapping instead of directly reading the ENV variables `ASCEND_RT_VISIBLE_DEVICES`.

Thank you for your question. In the Mooncake Connector, we identify NPUs through the ASCEND_RT_VISIBLE_DEVICES parameter. For example, when using NPUs 2, 3, 4, and 5, your Kubernetes YAML should set ASCEND_RT_VISIBLE_DEVICES="2,3,4,5". If this parameter is not configured, we assume the NPU indexing starts from 0 by default.
Our goal is to eliminate reliance on the ranktable. However, if the ranktable is still used, how would the physical NPU IDs be mapped in such a scenario?

wangxiyuan:
let's consider to build the mooncake package into docker image as well. @wxsIcey

wangxiyuan:
The CI failed due to other known issue. Let's merge this first. Feel free to add following-up PR if there is any other issue.

Shichang-Zhang:
@LCAIZJ @zzy-ContiLearn  Thank you for your advice!
Firstly, I think we could only assign the number of NPU devices such as 4 to the pod, but only aware the actual device assigned such as device 1,5,6,7 in the container? So I need to export the ENV variable ASCEND_RT_VISIBLE_DEVICES in the container instead of configuring it in the deployment yaml. 
Secondly, if the ENV variable ASCEND_RT_VISIBLE_DEVICES is used, the torch_npu will get the error shown previously at first. Torch_npu wants the device id in the range of [0,4). 
Currently I solve this problem by using a new ENV variable, ex. VLLM_VISIBLE_DEVICES, to avoid the error caused by torch_npu. Then I modify the processing logic of this ENV variable, since in multiple dp situation, every dp retrieves first several elements (number of tp rank) in the VLLM_VISIBLE_DEVICES. For example, if the ENV variable is 1,5,6,7 and config is dp2 tp2, dp0 tp0 gets device 1 and dp1 tp0 also gets device 1.
The patch file:
```
diff --git a/vllm_ascend/distributed/mooncake_connector.py b/vllm_ascend/distributed/mooncake_connector.py
index e223877..2d84098 100644
--- a/vllm_ascend/distributed/mooncake_connector.py
+++ b/vllm_ascend/distributed/mooncake_connector.py
@@ -758,13 +758,15 @@ class MooncakeConnectorWorker:
         # get tp device id
         # TODO(kw): https://github.com/vllm-project/vllm-ascend/pull/940
         # introducing some changes
-        device_ids_str = os.getenv("ASCEND_RT_VISIBLE_DEVICES", None)
+        device_ids_str = os.getenv("VLLM_VISIBLE_DEVICES", None)
         if device_ids_str is None:
             device_ids = list(
                 range(self.dp_rank * self.tp_size,
                       (self.dp_rank + 1) * self.tp_size))
         else:
-            device_ids = list(map(int, device_ids_str.split(',')))
+            #device_ids = list(map(int, device_ids_str.split(',')))
+            device_ids = list(map(int, device_ids_str.split(',')))[self.dp_rank * self.tp_size : (self.dp_rank + 1) * self.tp_size]
+        logger.info("device_ids: %s", device_ids)
         assert len(device_ids) > self.tp_rank  # type: ignore
         self.device_id = device_ids[self.tp_rank]  # type: ignore

```

LCAIZJ:
> @LCAIZJ @zzy-ContiLearn Thank you for your advice! Firstly, I think we could only assign the number of NPU devices such as 4 to the pod, but only aware the actual device assigned such as device 1,5,6,7 in the container? So I need to export the ENV variable ASCEND_RT_VISIBLE_DEVICES in the container instead of configuring it in the deployment yaml. Secondly, if the ENV variable ASCEND_RT_VISIBLE_DEVICES is used, the torch_npu will get the error shown previously at first. Torch_npu wants the device id in the range of [0,4). Currently I solve this problem by using a new ENV variable, ex. VLLM_VISIBLE_DEVICES, to avoid the error caused by torch_npu. Then I modify the processing logic of this ENV variable, since in multiple dp situation, every dp retrieves first several elements (number of tp rank) in the VLLM_VISIBLE_DEVICES. For example, if the ENV variable is 1,5,6,7 and config is dp2 tp2, dp0 tp0 gets device 1 and dp1 tp0 also gets device 1. The patch file:
> 
> ```
> diff --git a/vllm_ascend/distributed/mooncake_connector.py b/vllm_ascend/distributed/mooncake_connector.py
> index e223877..2d84098 100644
> --- a/vllm_ascend/distributed/mooncake_connector.py
> +++ b/vllm_ascend/distributed/mooncake_connector.py
> @@ -758,13 +758,15 @@ class MooncakeConnectorWorker:
>          # get tp device id
>          # TODO(kw): https://github.com/vllm-project/vllm-ascend/pull/940
>          # introducing some changes
> -        device_ids_str = os.getenv("ASCEND_RT_VISIBLE_DEVICES", None)
> +        device_ids_str = os.getenv("VLLM_VISIBLE_DEVICES", None)
>          if device_ids_str is None:
>              device_ids = list(
>                  range(self.dp_rank * self.tp_size,
>                        (self.dp_rank + 1) * self.tp_size))
>          else:
> -            device_ids = list(map(int, device_ids_str.split(',')))
> +            #device_ids = list(map(int, device_ids_str.split(',')))
> +            device_ids = list(map(int, device_ids_str.split(',')))[self.dp_rank * self.tp_size : (self.dp_rank + 1) * self.tp_size]
> +        logger.info("device_ids: %s", device_ids)
>          assert len(device_ids) > self.tp_rank  # type: ignore
>          self.device_id = device_ids[self.tp_rank]  # type: ignore
> ```

We ran some tests in our local environment using vLLM-Ascend v0.9.1 (we did not use v0.10.1 due to potential dependency compatibility concerns) and were unable to reproduce the issue you mentioned.
In our setup, when the device list is [0, 1, 2, 3], we observed that:
For DP0, ASCEND_RT_VISIBLE_DEVICES is configured as [0, 1].
For DP1, ASCEND_RT_VISIBLE_DEVICES is configured as [2, 3].
After reviewing the relevant code in vLLM, we believe the current implementation should correctly partition devices across different data parallelism (DP) groups. 
image

Shichang-Zhang:
@LCAIZJ I used v0.9.1 vllm-ascend also. In my condition, I use K8s on a physical machine, and run several vllm (P or D) pods on it. As shown in the first question I commented under this PR, firstly the ENV variable ASCEND_RT_VISIBLE_DEVICES  is not compatible with torch_npu. Because for each pod, the torch_npu in the container could only observe the mount NPU device, not all NPU device on the machine.  When I try to avoid using ASCEND_RT_VISIBLE_DEVICES by introducing a new ENV variable, I need to do extra operations to get the device for the current DP, as shown in the patch.

PR #1571, opened 2025-07-01
Only enable single version for wheel pr build

Yikun:
Only enable single version for wheel pr build

No

CI passed


wangxiyuan (approved):

Yikun:
Merge because it didn't depends on multi card test

PR #1575, opened 2025-07-02
Paralel test on Pr/1573

Yikun:


Yikun:
accuracy results: https://github.com/vllm-project/vllm-ascend/actions/runs/16015284839

PR #1136, opened 2025-06-09
[CI/UT][Refactor] move e2e spec decode and deepseek acc test to per pr

MengqingCao:
### What this PR does / why we need it?
1. run deepseek acc ut per pr --- multicard CI time increased by 9 min
2. run spec decode e2e test on v1 per pr --- singlecard CI time increased by 3 min (partly is disabled due to not work now)
~~3. align the output of whether dbo is enabled or not~~
    The generated results with and without dbo cannot be aligned.
    https://github.com/vllm-project/vllm-ascend/actions/runs/15822900528/job/44600029405?pr=1136
4. skip V0 mtp test due to failure in https://github.com/vllm-project/vllm-ascend/actions/runs/16012172833/job/45171988816
5. fix some version conflicts
### How was this patch tested?
CI passed with new added test.

PR #1580, opened 2025-07-02
[Bugfix] Add func `swap_states` to fix MLA attention

Potabk:

Notice that we discoupled `InputBatch` from vllm, but our mla attention still using the gpu_input_batch's attr:`swap_states`, which will lead to an error `AttributeError: 'InputBatch' object has no attribute 'swap_states'`


MengqingCao:
accuracy test in https://github.com/vllm-project/vllm-ascend/pull/1136 could cover this

Failure CI without changes in this pr: https://github.com/vllm-project/vllm-ascend/actions/runs/15971526516/job/45047293250

wangxiyuan:
@MengqingCao 's PR covered the test case, so it's ok to merge first to unblock mla function

vllm_ascend/attention/mla_v1.py
```diff
@@ -22,10 +22,10 @@
 from vllm_ascend.multistream.ms_split import model_input_split_v1_mla_attn
 from vllm_ascend.ops.attention import vanilla_chunked_prefill_mla
 from vllm_ascend.utils import npu_stream_switch, npu_wait_tensor
+from vllm_ascend.worker.npu_input_batch import InputBatch
```
wangxiyuan, line 25: meanless to move out from `TYPE_CHECKING`

PR #1584, opened 2025-07-02
[Feature] Enable inference support for Deepseekr1-w8a8-MTP

Irving11-BKN:

1. Support the inference of the Deepseekr1-w8a8-mtp model with statically-quantized shared_head in MTP layers.

Signed-off-by: curryliu <120010041@link.cuhk.edu.cn>

### DIFF
--- a/vllm_ascend/models/deepseek_mtp.py
+++ b/vllm_ascend/models/deepseek_mtp.py
@@ -28,8 +28,8 @@ from vllm.model_executor.layers.layernorm import RMSNorm
 from vllm.model_executor.layers.logits_processor import LogitsProcessor
 from vllm.model_executor.layers.quantization import QuantizationConfig
 from vllm.model_executor.layers.sampler import get_sampler
-from vllm.model_executor.layers.vocab_parallel_embedding import \
-    VocabParallelEmbedding
+from vllm.model_executor.layers.vocab_parallel_embedding import (
+    ParallelLMHead, VocabParallelEmbedding)
 from vllm.model_executor.models.deepseek_mtp import (
     DeepSeekMTP, DeepSeekMultiTokenPredictor, DeepSeekMultiTokenPredictorLayer,
     SharedHead)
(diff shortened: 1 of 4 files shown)


vllm_ascend/models/deepseek_mtp.py
```diff
@@ -40,6 +40,20 @@
 from .deepseek_v2 import CustomDeepseekV2DecoderLayer
 
 
+class CustomDeepSeekShareHead(SharedHead):
+
+    def __init__(self,
+                 config: PretrainedConfig,
+                 quant_config: Optional[QuantizationConfig] = None,
+                 prefix: str = "") -> None:
```
ganyi1996ppo, line 48: So, we are adding a Custom class implementation for this extra parameter passing through right?

vllm_ascend/quantization/quant_config.py
```diff
@@ -352,3 +360,20 @@ def apply(
...
+
+      Args:
+          quant_config: The Ascend quantization config.
+    """
+
+    def __init__(self, quant_config: AscendQuantConfig, prefix: str,
+                 packed_modules_mapping: Dict[str, Any]) -> None:
+        self.quantizer = AscendQuantizer.get_quantizer(
+            quant_config.quant_description, prefix, packed_modules_mapping)
+        self.quant_method = self.quantizer.build_shareHead_method()
```
ganyi1996ppo: Since this `build_sharedHead_method` returns Linear method, perhaps we can call build_linear_method directly instead?

vllm_ascend/models/deepseek_mtp.py
Jawonliu: Yes exactly, cause the vllm one doesn't include prefix args I need to use it.

vllm_ascend/quantization/quant_config.py
Jawonliu: Yes, but what if new sharehead quantization, like per token quantization or something else, we can modify the class directly

ganyi1996ppo: We can do that change then maybe? Too much interface seems unnecessary.

ganyi1996ppo: We can do that change then maybe? Too much interface seems unnecessary.

PR #1589, opened 2025-07-02
[0.9.1]DBO support EP parallel and optimize dual stream overlap

shikang-hangzhou:
1. DBO model support EP parallel
2. optimize dual stream overlap

max tokens:32784 input_len:1024  bs 32 dp2tp8ep16 
before open dbo
TTFT: 4017ms

before

after open dbo
TTFT: 3017ms
after

None

### DIFF
--- a/vllm_ascend/models/deepseek_dbo.py
+++ b/vllm_ascend/models/deepseek_dbo.py
@@ -630,7 +630,8 @@ class CustomDeepseekDBODecoderLayer(DeepseekV2DecoderLayer):
                 context.after_comm_event.record()
 
         for i in range(num_micro_batchs):
-            context.after_comm_event.wait()
+            ms_metadata.try_wait_event(layer_index, i,
+                                       MSEventKey.MOE_ALL_TO_ALL_FINISH)
             expert_tokens = torch_npu.npu_moe_compute_expert_tokens(
                 sorted_local_expert_idx[i], local_num_experts).to(torch.int64)
             group_list_type = 0
--- a/vllm_ascend/models/deepseek_dbo.py
+++ b/vllm_ascend/models/deepseek_dbo.py
@@ -630,8 +630,7 @@ class CustomDeepseekDBODecoderLayer(DeepseekV2DecoderLayer):
                 context.after_comm_event.record()
 
         for i in range(num_micro_batchs):
-            ms_metadata.try_wait_event(layer_index, i,
-                                       MSEventKey.MOE_ALL_TO_ALL_FINISH)
+            context.after_comm_event.wait()
             expert_tokens = torch_npu.npu_moe_compute_expert_tokens(
                 sorted_local_expert_idx[i], local_num_experts).to(torch.int64)
             group_list_type = 0
--- a/vllm_ascend/models/deepseek_dbo.py
+++ b/vllm_ascend/models/deepseek_dbo.py
@@ -25,40 +25,33 @@
 # # vllm-project/vllm/vllm/model_executor/models/deepseek_v2.py
 # """Inference-only DeepseekV2/DeepseekV3 model."""
 
-from typing import Any, Dict, List, Optional, Union
+from typing import List, Optional, Union
 
 import torch
 import torch.distributed as dist
 import torch_npu  # noqa: F401
-import vllm.envs as envs
 from torch import nn
 from transformers import PretrainedConfig
-from vllm.attention import Attention, AttentionMetadata
+from vllm.attention import AttentionMetadata
 from vllm.config import (CacheConfig, ModelConfig, VllmConfig,
                          get_current_vllm_config)
-from vllm.distributed import (get_pp_group,
+from vllm.distributed import (get_ep_group, get_pp_group,
+                              get_tensor_model_parallel_rank,
                               get_tensor_model_parallel_world_size,
                               get_tp_group, tensor_model_parallel_all_reduce)
 from vllm.distributed.parallel_state import get_dp_group
 from vllm.forward_context import get_forward_context
 from vllm.model_executor.layers.layernorm import RMSNorm
-from vllm.model_executor.layers.linear import (ColumnParallelLinear,
-                                               ReplicatedLinear,
-                                               RowParallelLinear,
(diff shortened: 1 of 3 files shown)


vllm_ascend/models/deepseek_dbo.py
```diff
@@ -446,10 +275,7 @@ def __init__(
         layer_idx = int(prefix.split(sep='.')[-1])
         self.layer_idx = layer_idx
         # TODO: enable mla in vllm-ascend
-        if model_config.use_mla:
-            attn_cls = CustomDeepseekDBOMLAAttention
-        else:
-            attn_cls = DeepseekV2Attention
+        attn_cls = CustomDeepseekV2MLAAttention
```
MengqingCao, line 278: why remove the branch when `use_mla` is False here?

vllm_ascend/models/deepseek_dbo.py
```diff
@@ -671,9 +531,138 @@ def _forward_ms_layer(
                     hidden_states[i], 0)
                 router_logits[i] = get_dp_group().all_gather(
                     router_logits[i], 0)
-            hidden_states[i] = self.mlp.experts._forward_ms_fused_moe_comp(
-                hidden_states[i], router_logits[i], is_prefill, real_top_k,
-                enable_force_load_balance)
+
+            if global_num_experts == 256:
```
MengqingCao, line 536: please add a comment that we use 256 here because the op `npu_moe_gating_top_k` only support this

vllm_ascend/models/deepseek_dbo.py
```diff
@@ -687,15 +676,22 @@ def _forward_ms_layer(
             context.before_comm_event.record()
             with torch.npu.stream(ms_metadata.communicate_stream):
                 context.before_comm_event.wait()
-                if self.dp_size > 1:
+                if (self.tp_size > 1
+                        and fused_moe_state != FusedMoEState.AllGather):
+                    dist.all_gather(list(chunk_hidden_states[i]),
```
MengqingCao, line 682: I recomand to use `tensor_model_parallel_all_gather` directly

shikang-hangzhou: here align with the deepseekv2 code

vllm_ascend/models/deepseek_dbo.py
shikang-hangzhou: dual stream overlap is a kind of optimized mode. deepseek-mha did not include in our application scenario, and have no improvements. So I think mha mode is useless.

vllm_ascend/models/deepseek_dbo.py
shikang-hangzhou: > please add a comment that we use 256 here because the op `npu_moe_gating_top_k` only support this

Thanks for your review,  we have add comments

vllm_ascend/models/deepseek_dbo.py
```diff
@@ -671,9 +531,139 @@ def _forward_ms_layer(
...
+
+            context = MultiStreamStepMetadata(
+                comm_stream=ms_metadata.communicate_stream,
+                before_comm_event=ms_metadata.ms_events[layer_index][i][
+                    MSEventKey.MOE_ALL_TO_ALL],
+                after_comm_event=ms_metadata.ms_events[layer_index][i][
+                    MSEventKey.MOE_ALL_TO_ALL_FINISH],
+            )
+            context.before_comm_event.record()
+            with torch.npu.stream(ms_metadata.communicate_stream):
```
ganyi1996ppo, line 601: This kind of stream control method seems can't be captured in torchair, so this is just a eager mode dual batch impl right?

vllm_ascend/models/deepseek_dbo.py
```diff
@@ -671,9 +531,139 @@ def _forward_ms_layer(
...
+
+                sorted_local_expert_idx_i, sorted_idx_i = torch.sort(
+                    local_expert_idx[i])
+                sorted_local_expert_idx.append(sorted_local_expert_idx_i)
+                sorted_idx.append(sorted_idx_i)
+                context.after_comm_event.record()
+
+        for i in range(num_micro_batchs):
+            ms_metadata.try_wait_event(layer_index, i,
+                                       MSEventKey.MOE_ALL_TO_ALL_FINISH)
```
ganyi1996ppo, line 634: What's the difference between `event.wait` and `try_wait_event` ?

vllm_ascend/models/deepseek_dbo.py
```diff
@@ -671,9 +531,139 @@ def _forward_ms_layer(
...
+            context.before_comm_event.record()
+            with torch.npu.stream(ms_metadata.communicate_stream):
+                context.before_comm_event.wait()
+                global_expert_tokens = torch.bincount(
+                    expanded_expert_idx[i], minlength=global_num_experts)
+                scatter_size = global_expert_tokens.view(
+                    ep_group.world_size, -1).sum(-1)
+                scatter_sizes.append(scatter_size)
+                gather_sizes = torch.empty_like(scatter_sizes[i])
+                dist.all_to_all_single(gather_sizes,
```
ganyi1996ppo, line 609: I wonder if my understand is correct, you are trying to overlap the all_to_all with gating_topk right, since the second stream launch needs to wait for the end of gating

ganyi1996ppo: And you overlap the combine phase of all to all with the calc of the shared expert.

vllm_ascend/models/deepseek_dbo.py
shikang-hangzhou: yes，so dual stream overlap only affect in prefill process.

vllm_ascend/models/deepseek_dbo.py
shikang-hangzhou: its same here, I will modify it.

shikang-hangzhou: > What's the difference between `event.wait` and `try_wait_event` ?

sorry, its diff between `event.wait` and `try_wait_event`. the last could assign num of microbatch which need wait

PR #950, opened 2025-05-26
Disaggregate prefill for kv cache register style

ganyi1996ppo:
### What this PR does / why we need it?
This PR adopt `LLMDataDist` for kv cache register and `pull_blocks` style disaggregate prefill implementation. The interface implementation mainly follows the design of NIXL PR https://github.com/vllm-project/vllm/pull/17751/files#diff-7eaad0b7dee0626bf29d10081b0f0c5e3ea15a4af97e7b182a4e0d35f8346953 .

This PR can be test with the following step:
- Generate the rank table for all machine.
- execute`toy_proxy.py` to launch the disaggregate prefill proxy server, specify the prefill ip, port and the decode ip, port
- Run the prefill server and decode server.
- send the request to the disaggregate prefill proxy

### How was this patch tested?

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/8d0a01a5f2b53794e4bc6b734d7b63cb8a9b7d7d

PR #1594, opened 2025-07-02
[Refactor] Use tuple as kv cache instead of tensor

lidenghui1110:

Previously, the graph mode stored KV cache as Tuple while eager mode used Tensor. To enable PD-separated zero-copy mode in later PR:  #950—where both P and D nodes must share the same storage format—this PR unifies the KV cache type to Tuple for both eager and graph modes.

No.

CI passed.

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -1020,8 +1020,7 @@ class AscendMLAImpl(MLAAttentionImpl):
                 num_kv_heads=self.num_kv_heads,
                 num_heads=self.num_heads,
                 scale_value=self.scale,
-                block_table=attn_metadata.decode.
-                block_table,  # type:ignore
+                block_table=attn_metadata.decode.block_table,  # type:ignore
                 context_lens=attn_metadata.decode.seq_lens,  # type:ignore
                 mla_vheadsize=self.kv_lora_rank,
                 out=attn_output)
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -659,12 +659,13 @@ class AscendMLAImpl(MLAAttentionImpl):
     def _compute_prefill_context(
         self,
         query: torch.Tensor,
-        kv_c_and_k_pe_cache: torch.Tensor,
+        kv_c_and_k_pe_cache: Tuple[torch.Tensor],
         rope_dim: int,
         attn_metadata: AscendMLAMetadata,
         prefix_output: torch.Tensor,
         prefix_lse: torch.Tensor,
     ):
+        assert len(kv_c_and_k_pe_cache) > 1
         prefill_metadata = attn_metadata.prefill
         if prefill_metadata is None or prefill_metadata.chunked_context is None:
             return prefix_output, prefix_lse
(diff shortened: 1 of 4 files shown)


vllm_ascend/attention/mla_v1.py
```diff
@@ -1008,13 +1007,21 @@ def _forward_decode(
                 actual_seq_lengths_kv=decode_meta.seq_lens_list,
             )
         else:
+            q = torch.cat([q_nope, q_pe], dim=-1)
+            attn_output = torch.empty(
+                [num_tokens, self.num_heads, self.kv_lora_rank],
+                dtype=q.dtype,
+                device=q.device)
+            k_cache = torch.cat(
```
MengqingCao, line 1017: I am worried about the extra NPU memory consumption this will bring

vllm_ascend/ops/attention.py
```diff
@@ -154,20 +154,21 @@ def vanilla_chunked_prefill_mla(
...
-        latent_kv_dim + rope_dim)[:, :max_context_len, :]
-    # get kv_c and k_pe
+    cache_kv_c = kv_c_and_k_pe_cache[0].squeeze()
+    cache_k_pe = kv_c_and_k_pe_cache[1].squeeze()
+    
     # cached_kv_c: [batch_size, max_context_len, latent_kv]
     # cached_k_pe: [batch_size, max_context_len, rope_dim]
-    cache_kv_c = cache_kv_c_pe[:, :, :latent_kv_dim]
-    cache_k_pe = cache_kv_c_pe[:, :, latent_kv_dim:]
+    batch_size = query_lens.size(0)
```
MengqingCao: redundant with L154
```suggestion

```

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2131,35 +2131,39 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
+                        num_blocks, block_size, num_kv_heads, head_dim = kv_cache_shape
+                        rope_dim = self.model_config.hf_text_config.qk_rope_head_dim
+                        nope_dim = self.model_config.hf_text_config.kv_lora_rank
+                        assert head_dim == rope_dim + nope_dim, \
+                            f"head_dim({head_dim}) != rope_dim({rope_dim}) + nope_dim({nope_dim})"
+                        nope_cache_shape = (num_blocks, block_size, 
+                                            num_kv_heads, nope_dim)
+                        rope_cache_shape = (num_blocks, block_size, 
+                                            num_kv_heads, rope_dim)
+                        nope_cache = torch.zeros(
```
MengqingCao: Let's keep the name unified, all `lantent_kv_cache` or `nope_cache`

wtr0504:
if i choose eager mode ,does this change matters ?

vllm_ascend/attention/mla_v1.py
lidenghui1110: Yes. It will use extra NPU memory in torch.cat, do you have any better suggestion? 
I noticed that in prefill, we alse use cat to do such things, like: [line:830](https://github.com/vllm-project/vllm-ascend/blob/70588531ab8b36fc04c0d067b5fd8d3ff20c3460/vllm_ascend/attention/mla_v1.py#L830), so the consumption should have been calcaulated in warmup stage.
And in deepseek-r1, the consumption should be per layer kv cache size, roughly about 15G/61 = 251M, maybe it is affordable.

vllm_ascend/ops/attention.py
lidenghui1110: done

vllm_ascend/worker/model_runner_v1.py
lidenghui1110: yes, modified to use kv_cache_nope and kv_cache_pe here. And in mla_v1.py, use kv_c and k_pe for short as before

lidenghui1110:
> if i choose eager mode ,does this change matters ?

no, it won't have any influence on any mode.

wtr0504:
> > if i choose eager mode ,does this change matters ?
> 
> no, it won't have any influence on any mode.
thank a lot

vllm_ascend/attention/mla_v1.py
MengqingCao: Acctualy I have no good suggestions on it, either. We seems must do this concat. But I think we can remove this until the ring attention could be enabled, right? If so, I think this change is acceptable. also cc @ganyi1996ppo

ganyi1996ppo: This line can be wipe out if ringmla is public accessible, and seems most of the change in this PR  already contains in the PR https://github.com/vllm-project/vllm-ascend/pull/950 , can you refactor it after that PR merge? Or there will be lots of conflict, which may bring more barrier for the 950 to merge....... cc @wangxiyuan @Yikun

lidenghui1110: @ganyi1996ppo It's OK. If all the changes in this PR has been done in [#950](https://github.com/vllm-project/vllm-ascend/pull/950), this PR can be closed. If so, hope it can be merged asap, other works rely on kv cache will rely on it.

PR #1634, opened 2025-07-06
[FOLLOWUP] Use base test to avoid patch everwhere

Yikun:
Use base test to avoid patch everwhere.

Followup here: https://github.com/vllm-project/vllm-ascend/pull/1566

No

ut ci passed


tests/ut/distributed/kv_transfer/test_simple_buffer.py
```diff
@@ -17,7 +17,7 @@ def __init__(self):
         self.deallocate_buffer = MagicMock()
 
 
-class TestSimpleBuffer(unittest.TestCase):
+class TestSimpleBuffer(TestBase):
 
     def setUp(self):
         self.pipe = MockSimplePipe()
```
wangxiyuan, line 23: I'm not sure if the supper.setup is needed here.

Yikun: will confirm soon, I think yes

PR #1647, opened 2025-07-07
[CustomOP][Refactor] Register CustomOP instead of overwrite forward_oot

MengqingCao:


wangxiyuan:
no update for long time. Close it now. Feel free to create a new one if it's still needed.

PR #1661, opened 2025-07-08
[Refactor] Refactor torchair

wangxiyuan:
Move tochair related code to a new module for better code maintainance

TODO: 
1. mla is still there, since it's only used for deepseek. We can refactor it later.
2. some ops are still there, we'll find a better way to register the ops with different case.

- vLLM version: v0.9.1
- vLLM main: https://github.com/vllm-project/vllm/commit/5358cce5ffbd4011f8fea2188995a249b43b8bfe


wangxiyuan:
Will createa new one instead.

PR #1668, opened 2025-07-08
[feat]add shared expert feature

sadatama:
添加静态共享专家外置特性


jianzs:
Thanks for your valuable contribution. Could you please share the current status, roadmap and any performance data for this feature?

Issue #1396, opened 2025-06-24
[Bug]: assert self.cpu_group is not None


iWasOmen:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```INFO 06-24 08:53:28 [importing.py:17] Triton not installed or not compatible; certain GPU-related functions will not be available.
WARNING 06-24 08:53:28 [importing.py:29] Triton is not installed. Using dummy decorators. Install it via `pip install triton` to enable kernel compilation.
INFO 06-24 08:53:31 [__init__.py:39] Available plugins for group vllm.platform_plugins:
INFO 06-24 08:53:31 [__init__.py:41] - ascend -> vllm_ascend:register
INFO 06-24 08:53:31 [__init__.py:44] All plugins in this group will be loaded. Set `VLLM_PLUGINS` to control which plugins to load.
INFO 06-24 08:53:31 [__init__.py:235] Platform plugin ascend is activated
WARNING 06-24 08:53:35 [_custom_ops.py:22] Failed to import from vllm._C with ImportError('/usr/local/python3.10.17/lib/python3.10/site-packages/vllm/_C.abi3.so: undefined symbol: _ZN3c108ListType3getERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEENS_4Type24SingletonOrSharedTypePtrIS9_EE')
Collecting environment information...
PyTorch version: 2.5.1
Is debug build: False

OS: Ubuntu 22.04.5 LTS (aarch64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version: Could not collect
CMake version: version 4.0.2
Libc version: glibc-2.35

Python version: 3.10.17 (main, May  8 2025, 07:18:04) [GCC 11.4.0] (64-bit runtime)
Python platform: Linux-5.10.0-136.12.0.86.r1526_92.hce2.aarch64-aarch64-with-glibc2.35

CPU:
Architecture:                       aarch64
CPU op-mode(s):                     64-bit
Byte Order:                         Little Endian
CPU(s):                             192
On-line CPU(s) list:                0-191
Vendor ID:                          HiSilicon
BIOS Vendor ID:                     HiSilicon
Model name:                         Kunpeng-920
BIOS Model name:                    HUAWEI Kunpeng 920 5250
Model:                              0
Thread(s) per core:                 1
Core(s) per socket:                 48
Socket(s):                          4
Stepping:                           0x1
[...]

Issue #1154, opened 2025-06-10
[Bug]: external_launcher下无法感知data_parallel_size 导致cpu_group计算异常 assert self.cpu_group is not None


tardis-key:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

PyTorch version: 2.5.1+cpu
Is debug build: False

OS: Debian GNU/Linux 10 (buster) (x86_64)
GCC version: (GCC) 9.3.0
Clang version: Could not collect
CMake version: version 4.0.1
Libc version: glibc-2.28

Python version: 3.10.14 (main, Jul  3 2024, 11:05:37) [GCC 8.3.0] (64-bit runtime)
Python platform: Linux-5.10.135.bsk.6-amd64-x86_64-with-glibc2.28

Versions of relevant libraries:
[pip3] gpytorch==1.14
[pip3] numpy==1.26.4
[pip3] pyzmq==26.4.0
[pip3] torch==2.5.1+cpu
[pip3] torch-npu==2.5.1.dev20250320
[pip3] torchaudio==2.5.1+cpu
[pip3] torchdata==0.11.0
[pip3] torchvision==0.20.1+cpu
[pip3] transformers==4.52.4
[conda] Could not collect
vLLM Version: 0.9.0
vLLM Ascend Version: 0.9.0rc2.dev0+g706de02.d20250610 (git sha: 706de02, date: 20250610)

CANN:
package_name=Ascend-cann-toolkit
version=8.1.RC1
innerversion=V100R001C21SPC001B203
compatible_version=[V100R001C15],[V100R001C18],[V100R001C19],[V100R001C20],[V100R001C21]
arch=x86_64
os=linux
path=/usr/local/Ascend/ascend-toolkit/8.1.RC1/x86_64-linux

</details>

### 🐛 Describe the bug

<details>
<summary>The output of `python collect_env.py`</summary>
`self.inference_engine = LLM(
            model=model_path,
            enable_sleep_mode=True,
            tensor_parallel_size=tensor_parallel_size,
            distributed_executor_backend="external_launcher", # 0.9.0推新的dp方案，ascend适配过程没到位,在el下没有感知dp
            data_parallel_size=8, # 手动修改，需要输入dp size
            dtype=config.dtype,
            enforce_eager=config.enforce_eager,
            gpu_memory_utilization=config.gpu_memory_utilization,
            disable_custom_all_reduce=True,
            disable_mm_preprocessor_cache=False,
[...]

PR #1681, opened 2025-07-09
[Dist][EP] Remove ETP/EP maintained in vllm-ascend

MengqingCao:
Remove ETP/EP maintained in branch main. We drop this as there is no relevant scenarios to use ETP now, and we may subsequently advocate implementing expert tensor parallelism in vLLM to support scenarios where the expert is needed to be sliced

This is a part of #1422 backport.

Fixes https://github.com/vllm-project/vllm-ascend/issues/1396 https://github.com/vllm-project/vllm-ascend/issues/1154

We'll not maintain etp/ep in vllm-ascend anymore, and use the tp/ep in vllm instead.

CI passed with new added and existing test.

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/fe8a2c544ad97119f4dafd316e5d9664521b73f9


tests/e2e/multicard/test_torchair_graph_mode.py
```diff
@@ -129,6 +129,7 @@ def _pangu_torchair_test_fixture(
             distributed_executor_backend="mp",
             enforce_eager=False,
             additional_config=additional_config,
+            enable_expert_parallel=True,
```
MengqingCao, line 126: We must set `enable_expert_parallel` to `True` when using pangu with the ep in vllm

MengqingCao: We should do this until disabling ep is supported in pangu. plz help review this, thanks! cc @Angazenn

tests/e2e/multicard/test_ep.py
```diff
@@ -0,0 +1,34 @@
+import os
+
+import pytest
+
+from tests.conftest import VllmRunner
+from tests.model_utils import check_outputs_equal
+
+
+@pytest.mark.skipif(os.getenv("VLLM_USE_V1") == "0",
```
wangxiyuan: we'll drop V0 support now. No need to add this `skipif`

tests/e2e/multicard/test_torchair_graph_mode.py
wangxiyuan: So maybe the related doc should be updated as well. For example https://vllm-ascend.readthedocs.io/en/latest/tutorials/multi_npu_moge.html

tests/e2e/multicard/test_ep.py
MengqingCao: Done

tests/e2e/multicard/test_torchair_graph_mode.py
MengqingCao: > So maybe the related doc should be updated as well. For example https://vllm-ascend.readthedocs.io/en/latest/tutorials/multi_npu_moge.html

Now the example is updated, thanks!

MengqingCao:
@jianzs @ttanzhiqiang could you help review this pr?

vllm_ascend/ops/fused_moe.py
```diff
@@ -142,22 +142,21 @@ def fused_experts_with_mc2(
...
-        "ep_world_size": all_to_all_group_size,
-        "ep_rank_id": local_rank,
-        # "group_tp": self.moe_rs_group_name,
+        "ep_world_size": ep_world_size,
+        "ep_rank_id": ep_rank_id,
         "group_tp": moe_all_to_all_group_name,
-        "tp_world_size": tp_size,
+        "tp_world_size": tp_world_size,
         "tp_rank_id": tp_rank,
     }
```
jianzs: There seems to be a mismatch in the TP size configuration - it's currently using attention TP size when it should be 1, if expert parallelism is enabled. Perhaps we should reference the TP size from `FusedMoeParallelConfig` instead?

https://github.com/vllm-project/vllm/blob/235bfd5dfe0975e42b115cfb910e73eff5c670d8/vllm/model_executor/layers/fused_moe/config.py#L259-L265

vllm_ascend/models/pangu_moe.py
```diff
@@ -144,8 +143,8 @@ def weight_loader(self, param: Parameter, loaded_weight: torch.Tensor,
 
         assert loaded_shard_id < len(self.output_sizes)
 
-        tp_rank = get_world_group().rank_in_group
-        tp_size = get_world_group().world_size
+        tp_rank = get_tp_group().rank_in_group
+        tp_size = get_tp_group().world_size
```
jianzs, line 147: ditto

vllm_ascend/models/pangu_moe.py
```diff
@@ -357,7 +356,7 @@ def pangu_group8_topk(
         num_tokens = scores.shape[0]
         router_scale = _ROUTER_SCALE.squeeze(  # type: ignore
         )
-
+        # TODO: support disable expert parallel
         ep_size = get_ep_group().world_size
```
ApsarasX, line 360: You should use `FusedMoEParallelConfig.make(...).ep_size` instead of `get_ep_group().world_size` as ep_size

ttanzhiqiang:
@MengqingCao 
1. If Prefill/decode uses AllGather or NaiveMulticast solution at the same time, this is ETP logic.
2. If Prefill/decode uses All2All/MC2 solution at the same time, this is EP logic.
3. Prefill uses AllGatherEP solution (using VLLM_ENABLE_FUSED_EXPERTS_ALLGATHER_EP switch), and Decode uses MC2 solution, this is AllGatherEP logic. .
4. In the PD separation scenario, the strategies used by P and D are separate.
After this pr, will the get_fused_moe_state function still have ep=1?

vllm_ascend/ops/fused_moe.py
MengqingCao: Good catch, will update soon

vllm_ascend/models/pangu_moe.py
MengqingCao: Thanks for this suggestion, will update in next commit

MengqingCao:
> @MengqingCao
> 
> 1. If Prefill/decode uses AllGather or NaiveMulticast solution at the same time, this is ETP logic.
> 2. If Prefill/decode uses All2All/MC2 solution at the same time, this is EP logic.
> 3. Prefill uses AllGatherEP solution (using VLLM_ENABLE_FUSED_EXPERTS_ALLGATHER_EP switch), and Decode uses MC2 solution, this is AllGatherEP logic. .

Thanks for this info, but I'm not formalir with the different fused moe state, maybe I need to read more code to understand the above 1, 2 and 3.

> 4. In the PD separation scenario, the strategies used by P and D are separate.
>    After this pr, will the get_fused_moe_state function still have ep=1?

I think we still have ep=1 when ep is disabled, you can refer to https://github.com/vllm-project/vllm/blob/235bfd5dfe0975e42b115cfb910e73eff5c670d8/vllm/model_executor/layers/fused_moe/config.py#L274-L281

vllm_ascend/models/pangu_moe.py
MengqingCao: I think it's fine to use tp group here? cause this is the weight loader for linear layer

vllm_ascend/models/pangu_moe.py
MengqingCao: Acctually we have no chance to get `moe_paralllel_config` in this `topk_warpper`, and I tried to make a new `FusedMoEParallelConfig` here, but it failed in torchair graph mode:
image

MengqingCao: I think the `get_ep_group().world_size` is always equal to `moe_parallel_config.ep_size` when ep is enabled, and pangu dose't support disabing ep even before this pr, thus I think it's fine to keep `get_ep_group().world_size` here.

ApsarasX: > I think the `get_tp_group().world_size` is always equal to `moe_parallel_config.ep_size` when ep is enabled, and pangu dose't support disabing ep even before this pr, thus I think it's fine to keep `get_tp_group().world_size` here.

Sorry, I misspoke. I meant to say that the flattened `tp_size` (`origin_tp_size * dp_size`) should be used as the MOE's `tp_size`. But I see you've already made the change, so that's fine.

MengqingCao: OK, thanks for your suggestion!

MengqingCao:
@ttanzhiqiang maybe we should remove the best example on A2 in https://github.com/vllm-project/vllm-ascend/pull/1101, WDYT?

ApsarasX:
> @ttanzhiqiang maybe we should remove the best example on A2 in #1101, WDYT?

I think you can remove these scripts, as we only need to maintain them internally on our side.

MengqingCao:
> > @ttanzhiqiang maybe we should remove the best example on A2 in #1101, WDYT?
> 
> I think you can remove these scripts, as we only need to maintain them internally on our side.

OK, I want to remove it mainly because it is a best example on ETP, which is removed here. I'll remove it then

MengqingCao:
Hi @jianzs @ttanzhiqiang @ApsarasX , your suggestions are addressed now, could you take a look again? thanks!

wangxiyuan (approved):

wangxiyuan:
Let's merge this first. Before next release, we should do a deep test about TP and EP

MingXiangL:
> Let's merge this first. Before next release, we should do a deep test about TP and EP

Do you have a timeline for the next release? Also, are there any temporary solutions for the bug described in https://github.com/vllm-project/vllm-ascend/issues/1396?

PR #1683, opened 2025-07-09
[V0 Deprecation] Remove prompt adapter in V1

shen-shanshan:

Remove V0 prompt adapter, find more details at https://github.com/vllm-project/vllm/pull/20588.

- vLLM version: v0.9.1
- vLLM main: https://github.com/vllm-project/vllm/commit/6db31e7a2735bb8132259bcbc21f046d62974325

### DIFF
--- a/vllm_ascend/worker/model_runner.py
+++ b/vllm_ascend/worker/model_runner.py
@@ -52,8 +52,6 @@ from vllm.model_executor.models.utils import set_cpu_offload_max_bytes
 from vllm.multimodal import (MULTIMODAL_REGISTRY, BatchedTensorInputs,
                              MultiModalKwargs, MultiModalPlaceholderMap,
                              MultiModalRegistry)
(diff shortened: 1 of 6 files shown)


vllm_ascend/worker/model_runner.py
```diff
@@ -52,8 +52,6 @@
 from vllm.multimodal import (MULTIMODAL_REGISTRY, BatchedTensorInputs,
```
wangxiyuan, line 52: For V0 code, this file can be removed totally. I think we can remove V0 first. Then remove the related `prompt adapter` code in V1

shen-shanshan: OK.

shen-shanshan:
Move to https://github.com/vllm-project/vllm-ascend/pull/1770, this is don't needed since prompt adapter doesn't exits in v1 code.

PR #1688, opened 2025-07-09
fix: use dist.reduce_scatter_tensor to avoid memory leak

NeverRaR:

fix: use dist.reduce_scatter_tensor to avoid memory leak

- vLLM version: v0.9.1
- vLLM main: https://github.com/vllm-project/vllm/commit/5358cce5ffbd4011f8fea2188995a249b43b8bfe


ApsarasX:
duplicate of PR #1380

PR #1562, opened 2025-07-01
Bump torch version to 2.7.1

leo-pony:
Signed-off-by: leo-pony <nengjunma@outlook.com>

### What this PR does / why we need it?
What this PR does / why we need it?
Bump torch version to 2.7.1
Cleanup https://github.com/vllm-project/vllm-ascend/commit/857f489 (https://github.com/vllm-project/vllm-ascend/pull/837)

### Does this PR introduce any user-facing change?
No

How was this patch tested?
CI passed

PR #1690, opened 2025-07-09
[CI] fix wheel build fail: "liberror_manager.so" could not be localed

leo-pony:
- Fixes # fix wheel build fail: "liberror_manager.so" could not be localed, fixing issue imformation:

### DIFF
--- a/.github/workflows/release_whl.yml
+++ b/.github/workflows/release_whl.yml
@@ -103,7 +103,8 @@ jobs:
           --exclude libc10.so \
           --exclude libc_sec.so \
           --exclude "libascend*.so" \
-          --exclude "libtorch*.so"
+          --exclude "libtorch*.so" \
+          --exclude "liberror_manager.so"
         done
         rm -f dist/*.whl
         mv dist/repaired/*.whl dist/


wxsIcey:
Compared with the main branch, the library files compiled by this branch directly depend on liberror_manager.so, so need to exclude this library file when using auditwheel repair.

This is the direct dependency of the vllm-ascend library file:
image
image

libstdc++.so.6, libgcc_s.so.1, libc.so.6, libm.so.6 are system libraries and are in the default exclusion list of auditwheel. No additional exclusion is required. For reference: https://github.com/pypa/auditwheel/blob/main/src/auditwheel/policy/manylinux-policy.json#L126. The other library files in the picture are from Ascend or torch need to be excluded.

Note that since adding new features may introduce direct dependencies on Ascend or other torch libraries, the library files that need to be excluded may be different for each vllm-ascend version.

wangxiyuan (approved):

leo-pony:
This pr will been included in PR:https://github.com/vllm-project/vllm-ascend/pull/1562

PR #1705, opened 2025-07-09
add qwen2.5_vl_without_padding's missing functions

zheliuyu:

To be supplemented

To be supplemented

To be supplemented


zheliuyu:
Ready for review. @wangxiyuan @leo-pony @as12138

vllm_ascend/models/qwen2_5_vl_without_padding.py
```diff
@@ -79,8 +79,9 @@ def forward(
 
         q, k, v = (rearrange(x, "s b ... -> b s ...").contiguous()
                    for x in (q, k, v))
-        q = torch_npu.npu_rotary_mul(q, cos, sin)
-        k = torch_npu.npu_rotary_mul(k, cos, sin)
+
+        q = torch_npu.npu_rotary_mul(q, cos.to(q.device), sin.to(q.device))
```
ganyi1996ppo: Why not put this `Tensor.to(device)` out of the model run?

vllm_ascend/models/qwen2_5_vl_without_padding.py
```diff
@@ -79,8 +79,9 @@ def forward(
...
+        k = torch_npu.npu_rotary_mul(k, cos.to(k.device), sin.to(k.device))
```
ganyi1996ppo: ditto

vllm_ascend/models/qwen2_5_vl_without_padding.py
zheliuyu: To fix RuntimeError:

```
Expected all tensors to be on the same device, 
but found at least two devices, npu:0 and cpu! (when checking argument for argument r1 in method wrapper__npu_rotary_mul)
```

ganyi1996ppo: Ok, but since you already customize the modeling, can you make the cos sin cache to be npu tensor to prevent the h2d operation layer by layer?

zheliuyu: > Ok, but since you already customize the modeling, can you make the cos sin cache to be npu tensor to prevent the h2d operation layer by layer?

Thanks for your advice. In order to make the PR more concise, this PR is only used to supplement the missing function of vl_without_padding.py. For Tensor.to(device), the PR will be submitted later after resuming the experimental results.

ganyi1996ppo:
I'm wondering how dose this PR been tested?  just reuse the ut of qwenvl?

zheliuyu:
> I'm wondering how dose this PR been tested? just reuse the ut of qwenvl?

Yes, reuse ut of qwenvl.

PR #1709, opened 2025-07-09
add qwen3-moe operation

shiyuan680:
origin qwen3_moe loss alltoall operation which result fault resultt, in this pr reuse some optimizations from deepseek.
no user-facing change

e2e & ut

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/b639327ad94b3aa16022ebea49f8e525660b736b

### DIFF
--- a/vllm_ascend/models/qwen3_moe.py
+++ b/vllm_ascend/models/qwen3_moe.py
@@ -16,7 +16,23 @@
 # Adapted from vllm/model_executor/models/qwen3_moe.py
 # This file is a part of the vllm-ascend project.
 
+from typing import Optional
+
+import torch
+import vllm
+from torch import nn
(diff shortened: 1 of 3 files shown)


vllm_ascend/models/qwen3_moe.py
```diff
@@ -33,3 +49,85 @@ class CustomQwen3MoeForCausalLM(Qwen3MoeForCausalLM):
         "experts":
         ["experts.0.gate_proj", "experts.0.up_proj", "experts.0.down_proj"],
     }
+
+
+class AscendQwen3MoeSparseMoeBlock(nn.Module):
```
ApsarasX, line 54: What is the issue that this PR aims to address?

PR #1719, opened 2025-07-10
[V0.9.1] optimize rope in qwen3

David9857:
use npu_apply_rotary_pos_emb when head_size is 128 and is noex_style

Optimize rope by extracting index_select from layers into model, which can reduce (layer_num -1) * 2 Gather ops in each prefill/decode stage.

NA

NA


ganyi1996ppo:
Please add unittest for this rope impl

David9857:
> Please add unittest for this rope impl

added

ganyi1996ppo (approved):

PR #1726, opened 2025-07-10
[V0.9.1] add support for flashcomm2 in qwen3

David9857:

Support FlashComm v2 in qwen3, which can reduce latency at prefill stage. set **VLLM_ENABLE_FC=1** and use **eager mode** to enbale this feature. 
**Note**: Enabling FlashComm in decoding stage may cause increased latency,  so it is recommended to use disaggregated prefilling and enbale this feature in the prefill instance only!!!

NA

NA

### DIFF
--- a/vllm_ascend/envs.py
+++ b/vllm_ascend/envs.py
@@ -143,6 +143,9 @@ env_variables: Dict[str, Callable[[], Any]] = {
     # Batch MC2 in prefill: The number of tokens in each batch
     "VLLM_ASCEND_FUSED_MOE_MC2_CHUNK_SIZE":
     lambda: int(os.getenv("VLLM_ASCEND_FUSED_MOE_MC2_CHUNK_SIZE", "128")),
+    # FlashComm optimization: Enable v1 and v2 by setting this flag to 1 or 2 respectively
+    "VLLM_ASCEND_ENABLE_FLASHCOMM":
+    lambda: int(os.getenv("VLLM_ASCEND_ENABLE_FLASHCOMM", '0'))
 }
 
 # end-env-vars-definition
--- a/vllm_ascend/models/qwen3.py
+++ b/vllm_ascend/models/qwen3.py
@@ -2,13 +2,19 @@ from collections.abc import Iterable
 from typing import Optional, Union
 
 import torch
+import torch.distributed as dist
+import torch.nn.functional as F
 from torch import nn
 from transformers import Qwen3Config
 from vllm.attention import AttentionType
 from vllm.compilation.decorators import support_torch_compile
 from vllm.config import CacheConfig, VllmConfig
-from vllm.distributed import get_pp_group
+from vllm.distributed import (get_pp_group, get_tensor_model_parallel_rank,
+                              get_tensor_model_parallel_world_size,
+                              get_tp_group, tensor_model_parallel_all_gather)
 from vllm.model_executor.layers.layernorm import RMSNorm
+from vllm.model_executor.layers.linear import (ReplicatedLinear,
+                                               RowParallelLinear)
 from vllm.model_executor.layers.logits_processor import LogitsProcessor
 from vllm.model_executor.layers.quantization import QuantizationConfig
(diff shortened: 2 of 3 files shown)


vllm_ascend/envs.py
```diff
@@ -143,6 +143,8 @@
     # Batch MC2 in prefill: The number of tokens in each batch
     "VLLM_ASCEND_FUSED_MOE_MC2_CHUNK_SIZE":
     lambda: int(os.getenv("VLLM_ASCEND_FUSED_MOE_MC2_CHUNK_SIZE", "128")),
+    "VLLM_ENABLE_FC":
+    lambda: int(os.getenv("VLLM_ENABLE_FC", 0))
```
rjg-lyh: I think this default value is better to be '0'?

vllm_ascend/models/qwen3.py
```diff
@@ -78,8 +170,19 @@ def forward(
...
                                skip_index_select=True)
         attn_output = self.attn(q, k, v)
+        pad_size = 0
+        if self.enable_fc:
+            # pad input because AllGather requires token_num to be divisible by tp_size
+            attn_output, pad_size = pad(attn_output, self.tp_size)
+            output = torch.empty(attn_output.shape,
+                                 dtype=attn_output.dtype,
+                                 device=attn_output.device)
+            dist.all_to_all_single(output, attn_output)
```
ganyi1996ppo: What is the comm_group of this operation? if you do this `all_to_all` in world_size, what if we have pipeline parallel ?

ganyi1996ppo: And what's the purpose of this `all_to_all` here? Why introduce this additional communication here?

ganyi1996ppo:
Do you have any performance statistics for this PR?

vllm_ascend/models/qwen3.py
David9857: > And what's the purpose of this `all_to_all` here? Why introduce this additional communication here?

Linear+Allreduce is replaced with All2All+Linear+AllGather in Flashcomm2.

ganyi1996ppo: What's the point for this change? dose it save the communication or computation? You just replicate two copy of o_proj on each device, which will increase the both bindwidth pressure and memory allocation. And the all2all dose not actually reduce the input data amount right? You just switch the data by all the tp rank. So what's the point of those change, does it really brings performance boost?

David9857: > What's the point for this change? dose it save the communication or computation? You just replicate two copy of o_proj on each device, which will increase the both bindwidth pressure and memory allocation. And the all2all dose not actually reduce the input data amount right? You just switch the data by all the tp rank. So what's the point of those change, does it really brings performance boost?

This change saves communication because input data size of all2all is 1/tp_size of allreduce. We can get performance benefits as long as the benefits of communication cover the increase of bindwidth pressure of linear.
refer this link for more details: https://gitcode.com/ascend-tribe/ascend-inference-cluster/blob/main/FlashComm/ascend-inference-cluster-flashcomm2.md

David9857:
> Do you have any performance statistics for this PR?

Here's the comparison of TTFT time when input_len=3000 and max_concurrency=20:
**origin:**
Mean TTFT (ms):                          1112.23
Median TTFT (ms):                        740.95
P99 TTFT (ms):                           3382.67
**with flashcomm2**
Mean TTFT (ms):                          1064.02
Median TTFT (ms):                        744.06
P99 TTFT (ms):                           3001.43

ganyi1996ppo:
Looks good, thanks for the explain !

> Here's the comparison of TTFT time when input_len=3000 and max_concurrency=20: **origin:** Mean TTFT (ms): 1112.23 Median TTFT (ms): 740.95 P99 TTFT (ms): 3382.67 **with flashcomm2** Mean TTFT (ms): 1064.02 Median TTFT (ms): 744.06 P99 TTFT (ms): 3001.43

PR #1732, opened 2025-07-11
[Perf] add patch to optimize apply_topk_topp

Pr0Wh1teGivee:
Performance optimization for apply_top_k_top_p
Use VLLM_ASCEND_ENABLE_TOPK_TOPP_OPTIMIZATION to enable this feature
e2e & ut

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/6a9e6b2abf88181f93a1959fe16291c3f1696329

### DIFF
--- a/tests/e2e/multicard/test_offline_inference_distributed.py
+++ b/tests/e2e/multicard/test_offline_inference_distributed.py
@@ -145,3 +145,25 @@ def test_models_distributed_pangu():
             distributed_executor_backend="mp",
     ) as vllm_model:
         vllm_model.generate_greedy(example_prompts, max_tokens)
+
+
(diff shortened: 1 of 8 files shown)


vllm_ascend/patch/worker/patch_0_9_2/patch_sampler.py
```diff
@@ -0,0 +1,83 @@
+#
```
wangxiyuan: move to worker/patch_common if this is a long-term patch.
and plase import this file in related  `__init__` file as well.

wangxiyuan:
This is a re-do PR. The test is passed before. Let's merge this in quick. I believe the CI will be happy.

PR #1748, opened 2025-07-11
torchair_graph bugfix when chunked_prefill is true

fems14:
when torchair_graph and chunked_prefill  are both true, save the decode kv_cache.

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -1200,7 +1200,7 @@ class AscendMLAImpl(MLAAttentionImpl):
                     prefill_hs, cos, sin, kv_cache,
                     attn_metadata.slot_mapping[num_decode_tokens:])
 
-                kv_c_normed = prefill_k_nope[:num_actual_toks, ...]
+                kv_c_normed_prefill = prefill_k_nope[:num_actual_toks, ...]
                 prefill_k_c_normed = prefill_k_nope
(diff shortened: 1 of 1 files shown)


vllm_ascend/attention/mla_v1.py
```diff
@@ -1218,12 +1218,23 @@ def forward(
...
                 # NOTE: Separate the kv cache in advance to avoid OOM or other issues
-                torch_npu._npu_reshape_and_cache(key=kv_c_normed.view(
+                torch_npu._npu_reshape_and_cache(key=kv_c_normed_prefill.view(
                     num_tokens, self.num_kv_heads, -1),
                                                  value=prefill_k_pe,
                                                  key_cache=kv_cache[0],
                                                  value_cache=kv_cache[1],
                                                  slot_indices=slots)
+
+            if kv_cache[0].numel(
```
ganyi1996ppo, line 1228: Can you solve this problem for good, and remove all those redundant `reshape_and_cache` operation, merge all the code path into one. All the rope_k and nope_k are supposed to be reshape and cached into the kv_cache right?

fems14: When the graph mode (without chunked prefill) is enabled, the reshapeAndCache path is not taken; instead, the fusion operator is used. The code for this part, including the entire file (mla_v1), has already been refactored by someone. It has been evaluated and is no longer on the current branch, but it will be merged into the main branch later.

ganyi1996ppo: fine

PR #1749, opened 2025-07-11
disable quantization in mindie_turbo

zouyida2052:
since the interface in vllm-ascend has changed so quickly, the quantization function in mindie_turbo is no longer needed, so it needs to be discarded.

no

through ci


wangxiyuan:
mindie-turbo doesn't contain quantization enhancement any more. Safe to remove relate code in vllm-ascend

PR #1759, opened 2025-07-12
[BugFix] Fix the bug when enabling both DBO and MoE multistream of deepseek.

whx-sjtu:
When enabling both DBO and MoE multistream of deepseek model, there exists problem. This PR fixes this situation and add e2e ut test for this scenario.


ganyi1996ppo:
Where is the ut test?

PR #1762, opened 2025-07-13
[CI] Remove benchmark patch and increase the scheduler frequency

Potabk:
This pr purpose to do the following things:
1. Remove `benchmark_datasets.py` patch
2. Increase the scheduler frequency to 2 times per day, due to the recent large number of daily submissions, we need to increase the default test time(6h)

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/247102f07f1dda117fef06493292e1925a5fcd31


wangxiyuan (approved):
Make sense

Yikun:
Merge to recover perf CI.

PR #1772, opened 2025-07-14
[Feat] chunkprefill mla support torchair graph

haojiangzheng:
chunkprefill mla only support eager mode now，we want to optimaze it by support torchair graph, the idea is simpler, when all the request is running in decode, use torchair graph to deal with it, else when chunkprefill or prefill only, use the eager mode

we can start the serve with chunkprefill mla torchair graph like this:

export VLLM_USE_V1=1
export VLLM_VERSION=0.9.0
python -m vllm.entrypoints.openai.api_server --model=DeepSeek-R1-W8A8-VLLM \
    --quantization ascend \
    --trust-remote-code \
    --port 8006 \
    -tp=8 \
    -dp=2 \
    --max-num-seqs 24 \
    --max-model-len 32768 \
    --block-size 128 \
    --additional-config '{"torchair_graph_config":{"enabled":true,"use_cached_graph":true,"graph_batch_sizes":[24]},"expert_tensor_parallel_size":16}' \
    --gpu-memory-utilization 0.96

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/586f286789a09f5616be74ee8bedde0a9f698a72

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,7 @@ RUN export PIP_EXTRA_INDEX_URL=https://mirrors.huaweicloud.com/ascend/repos/pypi
     python3 -m pip cache purge
 
 # Install modelscope (for fast download) and ray (for multinode)
-RUN python3 -m pip install modelscope ray && \
+RUN python3 -m pip install modelscope 'ray>=2.47.1' 'protobuf>3.20.0' && \
     python3 -m pip cache purge
 
 CMD ["/bin/bash"]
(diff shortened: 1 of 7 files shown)


ApsarasX:
Missing the fix code for ChunkPrefill OOM

zexplorerhj:
> Missing the fix code for ChunkPrefill OOM

fixed

jianzs (approved):

PR #1793, opened 2025-07-14
Move deepseek_v3 from deepseek_v2.py

xleoken:

Before patch, we can see `vllm_ascend.models.deepseek_v2:CustomDeepseekV3ForCausalLM`, it seems not friendly format.

```
WARNING 07-14 23:57:34 [registry.py:413] Model architecture DeepseekV2ForCausalLM is already registered, and will be overwritten by the new model class vllm_ascend.models.deepseek_v2:CustomDeepseekV2ForCausalLM.
WARNING 07-14 23:57:34 [registry.py:413] Model architecture DeepseekV3ForCausalLM is already registered, and will be overwritten by the new model class vllm_ascend.models.deepseek_v2:CustomDeepseekV3ForCausalLM.
WARNING 07-14 23:57:34 [registry.py:413] Model architecture Qwen3MoeForCausalLM is already registered, and will be overwritten by the new model class vllm_ascend.models.qwen3_moe:CustomQwen3MoeForCausalLM.
```

No.

Local Test.

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/bcdfb2a3308e14fbf46da6d6d41747f289af9300

### DIFF
--- a/vllm_ascend/models/deepseek_v2.py
+++ b/vllm_ascend/models/deepseek_v2.py
@@ -983,7 +983,3 @@ class CustomDeepseekV2ForCausalLM(DeepseekV2ForCausalLM):
                                    attn_metadata, intermediate_tensors,
                                    inputs_embeds)
         return hidden_states
-
-
-class CustomDeepseekV3ForCausalLM(CustomDeepseekV2ForCausalLM):
-    pass
--- a/vllm_ascend/models/__init__.py
+++ b/vllm_ascend/models/__init__.py
@@ -7,7 +7,7 @@ def register_model():
     from .deepseek_dbo import CustomDeepseekDBOForCausalLM  # noqa: F401
     from .deepseek_mtp import CustomDeepSeekMTP  # noqa: F401
     from .deepseek_v2 import CustomDeepseekV2ForCausalLM  # noqa: F401
-    from .deepseek_v2 import CustomDeepseekV3ForCausalLM  # noqa: F401
+    from .deepseek_v3 import CustomDeepseekV3ForCausalLM  # noqa: F401
     from .qwen2_5_vl import \
         AscendQwen2_5_VLForConditionalGeneration  # noqa: F401
     from .qwen2_vl import AscendQwen2VLForConditionalGeneration  # noqa: F401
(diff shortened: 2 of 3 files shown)


vllm_ascend/models/deepseek_v2.py
```diff
@@ -983,7 +983,3 @@ def forward(
                                    attn_metadata, intermediate_tensors,
```
wangxiyuan, line 983: how about rename this file to deepseek instead？

xleoken: hi @wangxiyuan, separate them would be more reasonable and suitable for expansion. We can refer to the VLLM project.

https://github.com/vllm-project/vllm/tree/main/vllm/model_executor/models

ApsarasX (approved):

vllm_ascend/models/deepseek_v2.py
wangxiyuan: https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/deepseek_v2.py#L928

So looks that there is no clear principle about the model file name and model arch.  add a new file with just one line is meanless IMO

xleoken: > https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/deepseek_v2.py#L928
> 
> So looks that there is no clear principle about the model file name and model arch. add a new file with just one line is meanless IMO


> https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/deepseek_v2.py#L928
> 
> So looks that there is no clear principle about the model file name and model arch. add a new file with just one line is meanless IMO

hi @wangxiyan you're right, from a code perspective, having only one line in a file seems meaningless. But the advantage of this is standardization, when adding support for new models, there is no need to modify the original code, but rather to add files, which makes the project more extensible.
For example, if we add deepseek_v4, v5, v6, and so on later, we only need to add files instead of modifying the original files. The advantage of this is to reduce the occurrence of bugs

wangxiyuan: This make sense

wangxiyuan (approved):

vllm_ascend/models/deepseek_v2.py
xleoken: Thanks for taking the time to review.

PR #1799, opened 2025-07-15
[CI] Fix changes CI

wangxiyuan:
Add `checkout` action before `dorny/paths-filter` to make it works with `push` case.
This is a known issue that `dorny/paths-filter` works without `checkout` in `pull_request` case but failed in `push` case. More detail is here:
https://github.com/dorny/paths-filter/issues/60#issuecomment-1464281021

The push CI works after this PR. The test result is here:
https://github.com/wangxiyuan/vllm-ascend/actions/runs/16285606468/job/45983607539
- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/d4d309409f2396e68e4b5a67ede194913502388b


Yikun (approved):

Yikun:
No need to run e2e, just merge

PR #1687, opened 2025-07-09
[WIP][Prefill Performance] Parallel Strategy Optimizations (VRAM-for-Speed Tradeoff)

SlightwindSec:
1. **Optimized MoE Expert All2All**: Rewritten for better throughput with increased memory cost.
2. **Shared Expert Sharding Strategy Update**: Switched from TP-aligned to pure DP for shared experts, enabling more efficient execution.
3. **O\_Proj AllReduce → ReduceScatter**: Reduced communication overhead by using ReduceScatter, made possible by pure DP sharding.
4. **AllGather Postponed**: Delayed to after QKV down projection to reduce synchronization impact during prefill.

PR #1802, opened 2025-07-15
[Prefill Perf] Parallel Strategy Optimizations (VRAM-for-Speed Tradeoff)

SlightwindSec:
1. **Shared Expert Sharding Strategy Update**: Switched from TP-aligned to pure DP for shared experts, enabling more efficient execution.
2. **O\_Proj AllReduce → ReduceScatter**: Reduced communication overhead by using ReduceScatter, made possible by pure DP sharding.
3. **AllGather Postponed**: Delayed to after QKV down projection to reduce synchronization impact during prefill.


ApsarasX:
same with PR #1687?

SlightwindSec:
> same with PR #1687?

Not exactly. PR #1687 also contains code that relies on an unreleased torch_npu build, so it can’t be merged yet. This new PR cherry‑picks only the backend‑agnostic parts of #1687 so they can land sooner; the torch_npu–specific pieces will stay in #1687 until that dependency becomes publicly available.

vllm_ascend/quantization/w8a8_dynamic.py
```diff
@@ -589,6 +661,115 @@ def fused_experts_with_all2all(hidden_states: torch.Tensor,
     return final_hidden_states, expert_tokens, group_list_type
 
 
+def fused_experts_with_all2all_v2(
```
ganyi1996ppo, line 664: Have you compare the performance gap between this PR the current 'token_dispatcher.py` implement

SlightwindSec: The current `token_dispatcher.py` implementation mainly supports **floating-point operations**. We previously attempted to adapt `token_dispatcher.py` directly for quantized scenarios but saw minimal to no performance gains.

However, we have indeed incorporated optimization ideas from `token_dispatcher.py` into a separate PR, **#1687**. This PR is still undergoing thorough validation and will be merged into the `main` branch once fully verified.

ganyi1996ppo: ok

PR #1806, opened 2025-07-15
[main] Use AddRmsNormQuant ops in the custom model to optimize Qwen3's performance

rjg-lyh:
Optimizes the performance of the Qwen3 quantization model by registering a custom model and adding the AddRmsNormQuant operation. Subsequent PRs will focus on performance optimizations based on this custom model.

No.

CI passed with existing test.

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/8d0a01a5f2b53794e4bc6b734d7b63cb8a9b7d7d

### DIFF
--- a/vllm_ascend/models/qwen3.py
+++ b/vllm_ascend/models/qwen3.py
@@ -1,0 +1,156 @@
+from collections.abc import Iterable
+from typing import Optional, Union
+
+import torch
+from torch import nn
+from transformers import Qwen3Config
+from vllm.compilation.decorators import support_torch_compile
+from vllm.config import CacheConfig, VllmConfig
+from vllm.distributed import get_pp_group
+from vllm.model_executor.layers.logits_processor import LogitsProcessor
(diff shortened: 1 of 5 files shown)


vllm_ascend/models/qwen3.py
```diff
@@ -0,0 +1,156 @@
+from collections.abc import Iterable
```
wangxiyuan, line 1: do not rewrite the model arch, if the change is only `AddRMSNormW8A8Quant`.

From 0.9.2, vllm support custom ops override, we can register our ops when setup vllm-ascend, take https://github.com/vllm-project/vllm-ascend/pull/1647 for example.

wangxiyuan:
As we talked offline, in the future, we should contribute the arch change to vLLM instead of maintain it inner vLLM Ascend. It's should be the common abiliity

wangxiyuan (approved):

wangxiyuan (approved):

Issue #1533, opened 2025-06-30
[Bug]: qwen3-1.7B using aclgraph report error


wuxi1117:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
Your output of above commands here
vllm 0.9.1
vllm-ascend 0.9.1rc1
CANN version=8.1.T18
```

</details>

### 🐛 Describe the bug

```python
# demo
from vllm import LLM, SamplingParams

import os
os.environ["VLLM_USE_V1"] = "1"

prompts = [
    "Hello, my name is",
]

# Create a sampling params object.
# sampling_params = SamplingParams(temperature=0.8, top_p=0.95)
# Create an LLM.
llm = LLM(model="Qwen3/Qwen3-1.7B")


# Generate texts from the prompts.
outputs = llm.generate(prompts)
```

```error 
(torch2_2) [ma-user wwwx]$python3 llm_test.py
INFO 06-30 19:21:33 [importing.py:17] Triton not installed or not compatible; certain GPU-related functions will not be available.
WARNING 06-30 19:21:33 [importing.py:29] Triton is not installed. Using dummy decorators. Install it via `pip install triton` to enable kernel compilation.
INFO 06-30 19:21:36 [__init__.py:39] Available plugins for group vllm.platform_plugins:
INFO 06-30 19:21:36 [__init__.py:41] - ascend -> vllm_ascend:register
INFO 06-30 19:21:36 [__init__.py:44] All plugins in this group will be loaded. Set `VLLM_PLUGINS` to control which plugins to load.
INFO 06-30 19:21:36 [__init__.py:235] Platform plugin ascend is activated
WARNING 06-30 19:21:40 [_custom_ops.py:22] Failed to import from vllm._C with ImportError('/home/ma-user/anaconda3/envs/torch2_2/lib/python3.10/site-packages/vllm/_C.abi3.so: undefined symbol: _ZN3c108ListType3getERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEENS_4Type24SingletonOrSharedTypePtrIS9_EE')
WARNING 06-30 19:21:46 [registry.py:401] Model architecture DeepSeekMTPModel is already registered, and will be overwritten by the new model class vllm_ascend.models.deepseek_mtp:CustomDeepSeekMTP.
[...]

PR #1582, opened 2025-07-02
Fix wheel glibc version incompatibility

wxsIcey:
### What this PR does / why we need it?
- Fixes https://github.com/vllm-project/vllm-ascend/issues/1533

### How was this patch tested?
1. Run the image
```
docker run \
    --name cann_container \
    --device /dev/davinci6 \
    --device /dev/davinci_manager \
    --device /dev/devmm_svm \
    --device /dev/hisi_hdc \
    -v /usr/local/dcmi:/usr/local/dcmi \
    -v /usr/local/bin/npu-smi:/usr/local/bin/npu-smi \
    -v /usr/local/Ascend/driver/lib64/:/usr/local/Ascend/driver/lib64/ \
    -v /usr/local/Ascend/driver/version.info:/usr/local/Ascend/driver/version.info \
    -v /etc/ascend_install.info:/etc/ascend_install.info \
    -it  quay.io/ascend/cann:8.1.rc1-910b-openeuler22.03-py3.11 bash
```

2. Install package
torch=2.5.1
torch-npu=2.5.1.post1.dev20250619 
vllm=0.9.1
vllm-ascend=vllm_ascend-0.1.dev1+g02ac443-cp311-cp311-manylinux2014_aarch64.manylinux_2_17_aarch64.whl
Artifact download URL: https://github.com/vllm-project/vllm-ascend/actions/runs/16039661265/artifacts/3454481370

3. Test offline script

```
from vllm import LLM, SamplingParams

import os
os.environ["VLLM_USE_V1"] = "1"

prompts = [
    "Hello, my name is",
]

llm = LLM(model="Qwen3/Qwen3-1.7B")

outputs = llm.generate(prompts)

for output in outputs:
    prompt = output.prompt
    generated_text = output.outputs[0].text
    print(f"Prompt: {prompt!r}, Generated text: {generated_text!r}")
```

4. Results
result

- vLLM version: v0.9.1
- vLLM main: https://github.com/vllm-project/vllm/commit/b942c094e3ab905aeb16f4136353f378e17159e8

PR #1808, opened 2025-07-15
[0.9.1] Fix wheel glibc version incompatibility

wxsIcey:

- Fixes https://github.com/vllm-project/vllm-ascend/issues/1533

Backported: https://github.com/vllm-project/vllm-ascend/pull/1582


Yikun:
Test passed on main, we need this in v0.9.1
https://github.com/vllm-project/vllm-ascend/actions/workflows/vllm_ascend_doctest.yaml?query=event%3Aschedule

cc @ganyi1996ppo

PR #1833, opened 2025-07-16
[1/N][CustomOp] Register RMSNorm instead of overwrite forward_oot

MengqingCao:
We'll refator `CustomOp` in vllm-ascend from this pr on. 

Use function `CustomOp.register_oot` to achieve the customop registery, taking `AscendRMSNorm` as an example:
```python
from vllm.model_executor.custom_op import CustomOp
CustomOp.register_oot(_decorated_op_cls=AscendRMSNorm, name="RMSNorm")
```
Part of https://github.com/vllm-project/vllm-ascend/pull/1647

N/A

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/d0dc4cfca48c2734da18ec42d6bba1346cbfc400

### DIFF
--- a/tests/ut/conftest.py
+++ b/tests/ut/conftest.py
@@ -16,7 +16,8 @@
 # This file is a part of the vllm-ascend project.
 #
 
-from vllm_ascend.utils import adapt_patch, register_ascend_customop  # noqa E402
+from vllm_ascend.utils import adapt_patch  # noqa E402
+from vllm_ascend.utils import register_ascend_customop
 
 adapt_patch()
 adapt_patch(True)
--- a/tests/ut/test_utils.py
+++ b/tests/ut/test_utils.py
@@ -301,6 +301,23 @@ class TestUtils(TestBase):
         self.assertFalse(utils.check_kv_cache_bytes_cache_exist(),
                          "Delete kv cache bytes cache dir failed")
 
+    @mock.patch("vllm.model_executor.custom_op.CustomOp")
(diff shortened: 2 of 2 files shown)


vllm_ascend/ops/layernorm.py
```diff
@@ -20,30 +20,30 @@
...
-        return x, residual
-
-    x, residual = torch_npu.npu_rms_norm(x, self.weight, self.variance_epsilon)
-    return x
-
-
-RMSNorm.forward_oot = forward_oot
+class AscendRMSNorm(RMSNorm):
+
+    def forward_oot(
```
wangxiyuan, line 26: This is a quick adapt for `CustomOp.register_oot` mechanism from vllm 0.9.2. For further step, we can remove  inherit from RMSNorm can write our own `RMSNorm` at all.

vllm_ascend/utils.py
```diff
@@ -561,3 +561,24 @@ def delete_torchair_cache_file():
     torch_air_abs_path = get_torchair_current_work_dir()
     if os.path.exists(torch_air_abs_path):
         shutil.rmtree(torch_air_abs_path)
+
+
+_ASCEND_CUSTOMOP_IS_REIGISTERED = False
+
+
+def register_ascend_customop():
```
wangxiyuan, line 569: should add ut for this func as well

MengqingCao: Thanks, now it is added

PR #1834, opened 2025-07-17
[Feature] Enable inference support for Deepseekr1-w8a8-MTP

Irving11-BKN:
1. Support the inference of the Deepseekr1-w8a8-mtp model with statically-quantized shared_head in MTP layers.

Signed-off-by: curryliu <120010041@link.cuhk.edu.cn>

### DIFF
--- a/vllm_ascend/models/deepseek_mtp.py
+++ b/vllm_ascend/models/deepseek_mtp.py
@@ -28,8 +28,8 @@ from vllm.model_executor.layers.layernorm import RMSNorm
 from vllm.model_executor.layers.logits_processor import LogitsProcessor
 from vllm.model_executor.layers.quantization import QuantizationConfig
 from vllm.model_executor.layers.sampler import get_sampler
-from vllm.model_executor.layers.vocab_parallel_embedding import \
-    VocabParallelEmbedding
+from vllm.model_executor.layers.vocab_parallel_embedding import (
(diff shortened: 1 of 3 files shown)


vllm_ascend/models/deepseek_mtp.py
```diff
@@ -40,6 +40,20 @@
...
+    def __init__(self,
+                 config: PretrainedConfig,
+                 quant_config: Optional[QuantizationConfig] = None,
+                 prefix: str = "") -> None:
+        nn.Module.__init__(self)
+        self.norm = RMSNorm(config.hidden_size, eps=config.rms_norm_eps)
+        self.head = ParallelLMHead(config.vocab_size,
+                                   config.hidden_size,
+                                   quant_config=quant_config,
+                                   prefix=maybe_prefix(prefix, "head"))
```
wangxiyuan, line 54: why only we need `prefix` here？The behavior of w8a8 weight is different with others?

Irving11-BKN: prefix is used for the function is_layer_skipped in quant_config.py file, the behavior is the same

wangxiyuan: if so, i think this should be done in vLLM instead. @MengqingCao can you take a look and create a PR for vLLM?

PR #1836, opened 2025-07-17
[Misc] Add `fusion_result.json` to `.gitignore`

shen-shanshan:
Add `fusion_result.json` to `.gitignore`.

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/76b494444fd864ffc53a623420668d1865c804b9


wangxiyuan:
Please add note that which case this file is genreated.

shen-shanshan:
> Please add note that which case this file is genreated.

I just directly run offline inference in my vscode, and this file is generated every time.

.gitignore
```diff
@@ -199,3 +199,5 @@ kernel_meta/
 # build info file generated by setup.py
 /vllm_ascend/_build_info.py
 /vllm_ascend/include/
+
+fusion_result.json
```
wangxiyuan, line 204:
```suggestion
# generated by CANN
fusion_result.json
```

shen-shanshan: Updated.

wangxiyuan (approved):

PR #1837, opened 2025-07-17
[Misc][V0 Deprecation] Add `__main__` guard to all offline examples

shen-shanshan:
Add `__main__` guard to all offline examples.

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/76b494444fd864ffc53a623420668d1865c804b9

### DIFF
--- a/examples/disaggregated_prefill/disaggregated_prefill_offline.py
+++ b/examples/disaggregated_prefill/disaggregated_prefill_offline.py
@@ -12,6 +12,9 @@ import os
 import time
 from multiprocessing import Event, Process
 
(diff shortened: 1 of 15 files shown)


wangxiyuan (commented):
yes, it's important, and what about env `VLLM_WORKER_MULTIPROC_METHOD=spawn`？ should it ba added as well?

shen-shanshan:
> yes, it's important, and what about env `VLLM_WORKER_MULTIPROC_METHOD=spawn`？ should it ba added as well?

Added.

wangxiyuan (approved):

PR #1839, opened 2025-07-17
[Misc] Remove transformers from vllm-ascend requirements

Potabk:

Remove transformers from vllm-ascend requirements, we do not need pin transformers to  a lower version for this [fix](https://github.com/vllm-project/vllm/pull/20244/files) in the upstream


.github/workflows/vllm_ascend_test.yaml
```diff
@@ -184,8 +184,8 @@ jobs:
         env:
           PIP_EXTRA_INDEX_URL: https://mirrors.huaweicloud.com/ascend/repos/pypi
         run: |
-          pip install -r requirements-dev.txt
           pip install -v -e .
+          pip install -r requirements-dev.txt
```
wangxiyuan, line 188: why change this order

Potabk: for verify the  the installation with vllm default transformers version without conflict

wangxiyuan: I still don't understand why. In this PR. the `transformers` has been removed from requirements. So install `requirement` first won't change anything IMO.

Potabk: https://github.com/vllm-project/vllm-ascend/blob/33ef5dc8130fb7132b23c14d13a5c9361118d912/requirements-dev.txt#L16C1-L16C3 this will install `transformers `, but for some users, installing `requirements-dev` is not necessary

wangxiyuan: but we should make sure `requirements.txt` is installed before install the package.

Potabk: `pip install -e .` will install req

wangxiyuan: it's not clear for developers. You can add a note at least

PR #1840, opened 2025-07-17
[0.9.1][MTP V1]MTP model adapt torchair

JC-ut0:


vllm_ascend/worker/mtp_proposer_v1.py
```diff
@@ -155,16 +163,19 @@ def propose(
...
         # Replace the last token with the next token.
         # E.g., [b1, b2, c1, c2, c3, c3] -> [a2, b2, b3, c2, c3, c4]
+        if token_indices is not None and self.runner.torchair_graph_enabled:
+            last_token_indices = token_indices
+            common_attn_metadata = self.runner.common_attn_metadata
+        else:
+            seq_lens = (target_positions[last_token_indices] + 1)
+            common_attn_metadata = CommonAttentionMetadata(
+                query_start_loc=cu_num_tokens, seq_lens=seq_lens)
+
```
NeverRaR, line 170: There is no need to process token_indices separately for the torchchair. You just need to set the padding length correctly in the runner

vllm_ascend/worker/mtp_proposer_v1.py
```diff
@@ -98,7 +99,13 @@ def prepare_inputs(
                              cu_target_query_lens[:-1])
         # [a, b, c] -> [a - n1, b - n2, c - n3]
         num_tokens_per_req = query_len_per_req - num_rejected_tokens
-        if force_one_token:
+        if is_torchair_graph:
```
NeverRaR, line 102: There is no need to process token_indices separately for the torchchair. You just need to set the padding length correctly in the runner

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2130,12 +2138,20 @@ def _generate_mtp_token_ids(
...
-            target_hidden_states = hidden_states[token_indices]
-            target_slot_mapping = attn_metadata.slot_mapping[token_indices]
+                force_one_token=False,
+                is_torchair_graph=self.torchair_graph_enabled)
+            if self.torchair_graph_enabled:
+                # the seq len of each bath is padded to 2, thus input is same as the main model
+                target_token_ids = self.input_ids[:num_scheduled_tokens]
+                target_positions = positions[:num_scheduled_tokens]
+                target_hidden_states = hidden_states[:num_scheduled_tokens]
+                target_slot_mapping = attn_metadata.slot_mapping[:
```
NeverRaR, line 2164: Using the same input as the main model will cause the proposer to use rejected tokens, which will significantly reduce the acceptance rate.The correct code can be:
```
            target_token_ids = self.input_ids[token_indices]
            target_positions = positions[token_indices]
            target_hidden_states = hidden_states[token_indices]
            target_slot_mapping = attn_metadata.slot_mapping[token_indices]

            if self.torchair_graph_enabled:
                self.num_token_pad_size += num_scheduled_tokens - len(token_indices)
```
Just set the padding size correctly and don't make any additional changes.

NeverRaR (changes requested):

vllm_ascend/worker/model_runner_v1.py
JC-ut0: The padding strategy involves padding each request to two tokens, regardless of whether it is accepted or rejected. If padding is not applied, there is a potential risk of operator errors, as the `torch_npu.npu_fused_infer_attention_score` function cannot handle scenarios where the single batch sequence length exceeds 16, which could potentially be triggered in padding scenarios.

PR #1850, opened 2025-07-17
[WIP][V0.9.1] add support for flashcomm2 in qwen2

David9857:

Same changes as #1726 but for qwen2.
Note: Enabling FlashComm in decoding stage may cause increased latency, so it is recommended to use disaggregated prefilling and enbale this feature in the prefill instance only!!!


wangxiyuan:
no update for long time. Close it now. Feel free to create a new one if it's still needed.

PR #1853, opened 2025-07-17
molink compatible vllm-ascend

oldcpple:


oldcpple:
Sorry we were trying to make vllm-ascend plugins compatible with our own project. Please ignore this PR, we can just use a local branch

PR #1872, opened 2025-07-18
[0.9.1][Perf]Remove NZ of kv_b_proj in Deepseek MLA.

whx-sjtu:
This PR removes NZ transformation of weights of kv_b_proj. This is because we find that this matmul weight is not quantized and will fall back to ND calculation in runtime (because currently float bmm nz is not supported in torchair graph), which causes two redundant transData operations (trans weight from NZ back to ND). Removing these two operations will provide an optimization of about 40us per layer.


ttanzhiqiang:
https://github.com/vllm-project/vllm-ascend/pull/1131 This PR does this

PR #1884, opened 2025-07-19
Add  Custom Kernels For LoRA Performance

taoxudonghaha:
Add two custom kernels(bgmv_shrink and bgmv expand) to solve the performance of LoRA
no user-facing change
we add Unit Test file to test the custom ascendc kernel. See vllm-ascend/tests/e2e/singlecard/ops/test_bgmv_expand.py and vllm-ascend/tests/e2e/singlecard/ops/test_bgmv_expand.py
Based on the actual test of the QWen2.5 7B model using vllm-ascend version v0.9.2.rc1, the TTFT, TPOT and throughput have increased by about 70%.
image

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/40d86ee412eeeca93e0c37432db6b96829cb64e2


csrc/torch_binding.cpp
```diff
@@ -228,6 +314,14 @@ TORCH_LIBRARY_EXPAND(_C, ops)
         "                         int added_vocab_start_index, "
         "                         int added_vocab_end_index) -> (Tensor masked_input, Tensor mask)");
     ops.impl("get_masked_input_and_mask", torch::kPrivateUse1, &vllm_ascend::get_masked_input_and_mask);
+
+    ops.def("bgmv_shrink(Tensor! x, Tensor! weight, Tensor! indices, Tensor! y, float scale) -> ()");
```
wangxiyuan, line 311: once the `bgmv_shrink` and `bgmv_expand` is exposed, I think we should use it to replace the common one as well. For example here: https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/lora/punica_wrapper/punica_npu.py#L51

taoxudonghaha: I have already replaced the relevant interface calls in the punica_npu.py

PR #1890, opened 2025-07-19
[Perf] Avoid performing index selection of sin/cos cache every layer

whx-sjtu:
This PR port PR #1614  and PR #1693 of v0.9.1-dev to main which optimize number of index selections of sin/cos cache.


ApsarasX:
LGTM

PR #1561, opened 2025-07-01
[Perf][MoE] Improve shared experts multi-stream for w8a8 dynamic.

whx-sjtu:
This PR designs the shared expert multi-stream parallelism of w8a8-dynamic-quantized MoE stage in more detail to achieve better performance.
Current multi-stream parallel for shared experts are shown in following pic:
image
Performance change：
Before：
image
After：
image

PR #1653, opened 2025-07-07
[CI/Build] Upgrade CANN to 8.2.RC1.alpha003

MengqingCao:
### What this PR does / why we need it?
Upgrade CANN to 8.2.RC1.alpha003

### Does this PR introduce _any_ user-facing change?
Need to upgrade CANN to 8.2.RC1.alpha003

### How was this patch tested?
CI passed with existing test.

PR #1891, opened 2025-07-19
[Perf][MoE] Improve MoE multistream parallel performace.

whx-sjtu:
This PR ports PR #1561 of v0.9.1-dev to main.

### DIFF
--- a/vllm_ascend/quantization/w8a8_dynamic.py
+++ b/vllm_ascend/quantization/w8a8_dynamic.py
@@ -30,6 +30,79 @@ from vllm_ascend.utils import (ACL_FORMAT_FRACTAL_NZ, FusedMoEState,
                                dispose_tensor, get_fused_moe_state,
                                npu_stream_switch, npu_wait_tensor)
 
+def apply_mlp_decode(hidden_states_wrapper: List[torch.Tensor],
+                     w1: torch.Tensor,
+                     w1_scale: torch.Tensor,
+                     w2: torch.Tensor,
+                     w2_scale: torch.Tensor,
+                     group_list: torch.Tensor,
+                     dynamic_scale: torch.Tensor = None,
+                     group_list_type: int = 1) -> torch.Tensor:
+    """
+    apply MLP: gate_up_proj -> swiglu -> down_proj
+    Args:
+        hidden_states_wrapper: wrapper of input hidden states with shape (num_tokens, hidden_size).
+        w1: expert weights1 with shape
+            (num_experts, hidden_size, intermediate_size * 2)
+        w1_scale: weights1 scale with shape (num_experts, intermediate_size * 2)
+        w2: expert weights2 with shape
+            (num_experts, intermediate_size, hidden_size)
+        w2_scale: weights2 scale with shape (num_experts, hidden_size)
+        group_list: number of tokens for each expert, follow cumsum mode, and
+            with shape (num_experts).
+        transpose_weight:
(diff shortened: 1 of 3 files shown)


leo-pony:
LGTM if CI passed

wangxiyuan (approved):

wangxiyuan:
1. This is a performance improve change. No new e2e test is needed.
2. the changed 3 files doesn't has ut test file. It's working by other developers. We can ignore it in this PR.

Yikun (approved):
also cc @jianzs @ApsarasX

wangxiyuan (approved):

vllm_ascend/quantization/w8a8_dynamic.py
```diff
@@ -31,6 +31,80 @@
                                dispose_tensor, get_fused_moe_state)
 
 
+def apply_mlp_decode(hidden_states_wrapper: List[torch.Tensor],
```
ApsarasX: For the `apply_mlp_decode` function, the first parameter no longer needs to be an array. You can refer to existing `apply_mlp` implementations.

ApsarasX: Why was a new `apply_mlp_decode` function added? It seems essentially identical to the existing `apply_mlp` function, except for the `split_item` and `output_dtype` fields.

vllm_ascend/quantization/w8a8_dynamic.py
```diff
@@ -708,6 +785,8 @@ def apply(
         log2phy: torch.Tensor = None,
         global_redundant_expert_num: int = 0,
         shared_experts: Optional[Any] = None,
+        quantized_x_for_share: Optional[Any] = None,
```
ApsarasX, line 827: Aren't the parameters for these two apply functions passed in at the outer level?

vllm_ascend/quantization/w8a8_dynamic.py
```diff
@@ -779,7 +866,9 @@ def apply(
                 moe_all_to_all_group_name=self.moe_all_to_all_group_name,
                 log2phy=log2phy,
                 global_redundant_expert_num=global_redundant_expert_num,
-                shared_experts=shared_experts)
+                shared_experts=shared_experts,
+                quantized_x_for_share=shared_gate_up,
```
ApsarasX: How about directly using `shared_gate_up` as the field name?

`quantized_x_for_share` doesn't quite fit the semantics in this context.

vllm_ascend/quantization/w8a8_dynamic.py
whx-sjtu: This function is added in order to introduce fused op `npu_dequant_swiglu_quant` which is only used in decode phase.

whx-sjtu: Thanks, I will change it.

ApsarasX: > This function is added in order to introduce fused op `npu_dequant_swiglu_quant` which is only used in decode phase.

We previously used `npu_dequant_swiglu_quant` in `apply_mlp`, triggered under certain conditions.

@realliujiaxu knows more specific details. Discuss with him?

ApsarasX (approved):

Yikun:
better rebase after https://github.com/vllm-project/vllm-ascend/pull/1653

PR #1900, opened 2025-07-21
[Bugfix]: Correct handling of cos_sin_cache length

jianzs:

This PR addresses the performance issue related to cos/sin cache handling:

1. The cos/sin cache is already initialized with the maximum context length during initialization. However, due to `max_seq_len_cache` being stored as `seq_len`, the condition check was incorrect, leading to unnecessary cache recreation.

2. Since the cos/sin cache is already initialized with maximum context length, it should not trigger recreation during the process.

3. Fixed variable naming: `max_seq_len_cache` was never used and should be `max_seq_len`. This also is the correct variable to check against the maximum context length.

No

CI pass.

### DIFF
--- a/vllm_ascend/ops/rotary_embedding.py
+++ b/vllm_ascend/ops/rotary_embedding.py
@@ -93,7 +93,9 @@ def native_rope_deepseek_forward(self,
                                  offsets: Optional[torch.Tensor] = None,
                                  max_seq_len: Optional[int] = None):
     if max_seq_len is not None and max_seq_len > self.max_seq_len:
-        _set_cos_sin_cache(self, max_seq_len, query.device, query.dtype)
+        raise ValueError(f"Unable to reach here, max_seq_len {max_seq_len} "
+                         "exceeds the maximum sequence length "
+                         f"{self.max_seq_len} for rotary embedding.")
     if len(key.shape) == 2:
         key = key[:, None, :]
     # Note: we implement the non neox_style method with shuffle the last dim and neox style
@@ -209,7 +211,7 @@ def apply_rotary_pos_emb(q, k, cos, sin, position_ids, unsqueeze_dim=1):
 
 
 def _set_cos_sin_cache(self, seq_len, device, dtype):
-    self.max_seq_len_cached = seq_len
+    self.max_seq_len = seq_len * self.scaling_factor
     dim = self.rotary_dim
 
     freq_extra = 1.0 / (self.base**(
@@ -276,7 +278,6 @@ def deepseek_rope_init_func(
     super(DeepseekScalingRotaryEmbedding,
           self).__init__(head_size, rotary_dim, max_position_embeddings, base,
                          is_neox_style, dtype)
-    self.max_seq_len = max_position_embeddings
     _set_cos_sin_cache(self,
                        max_position_embeddings,
                        dtype=dtype,


jianzs:
@whx-sjtu  PTAL

whx-sjtu (approved):
I reviewed codes of branch v0.9.1-dev and found that this problem has already been solved in that branch while hasn't been ported to main. Thanks for finding and fixing this. LGTM.

ApsarasX:
> I reviewed codes of branch v0.9.1-dev and found that this problem has already been solved in that branch while hasn't been ported to main. Thanks for finding and fixing this. LGTM.

@whx-sjtu Which PR fixed this issue in the 0.9.1-dev branch?

vllm_ascend/ops/rotary_embedding.py
```diff
@@ -209,7 +211,7 @@ def apply_rotary_pos_emb(q, k, cos, sin, position_ids, unsqueeze_dim=1):
 
 
 def _set_cos_sin_cache(self, seq_len, device, dtype):
-    self.max_seq_len_cached = seq_len
+    self.max_seq_len = seq_len * self.scaling_factor
```
weijinqian0: There is no problem in v0.9.1.  what happens.

jianzs:
> > I reviewed codes of branch v0.9.1-dev and found that this problem has already been solved in that branch while hasn't been ported to main. Thanks for finding and fixing this. LGTM.
> 
> @whx-sjtu Which PR fixed this issue in the 0.9.1-dev branch?

@ApsarasX   https://github.com/vllm-project/vllm-ascend/pull/1551

vllm_ascend/ops/rotary_embedding.py
jianzs: https://github.com/vllm-project/vllm-ascend/pull/1551 fixed this problem in v0.9.1-dev

wangxiyuan: So can we just cherry-pick this one?

jianzs: Ok, I've cherry-picked #1551 into the current pr.

jianzs:
@zzzzwwjj  PTAL

jianzs:
@wangxiyuan  @Yikun  @ganyi1996ppo  ready to merge.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1137,11 +1139,14 @@ def _process_reqs(
         else:
             attn_state = AscendAttentionState.PrefillCacheHit
 
-        self.attn_mask = self._make_attention_mask(
-            seq_lens=seq_lens,
-            query_lens=num_scheduled_tokens,
-            position=positions,
-            attn_state=attn_state)
+        # NOTE: when use ring_mla, attn_mask don't need to generate here.
+        if not self.use_ring_mla or attn_state == AscendAttentionState.PrefillNoCache:
```
ganyi1996ppo, line 1143: According to the comments, there should be and rather than or?

jianzs: Indeed, it looks as you described. This PR is a cherry-pick from https://github.com/vllm-project/vllm-ascend/pull/1551. @zzzzwwjj could you please verify this?

zzzzwwjj: > According to the comments, there should be and rather than or?

When use ring_mla and attn_state == PrefillNoCache, it will not use ringattn op and also need attn_mask. So `or` is right.

ApsarasX: should be considered with #2319

PR #1914, opened 2025-07-21
add tp optimze.

momo609:


ApsarasX:
How much memory can this PR save in a pure DP scenario? Will it affect performance?

Angazenn:
Hello, we have also test lmhead-tp for deepseek and it reduces 0.5~1ms on 2 A3 nodes with DP32. Therefore, can we make it more general so that it can be applied on both A2/A3, deepseek/qwen？

jianzs:
Please submit separate pull requests for changes related to the LM Head and pipeline parallelism features.

PR #1918, opened 2025-07-22
Adapt new model

Apocalypse990923-qshi:
Adapt new model

No

CI passed with new added/existing test

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/32142b3c62277ac7cb941f2036270decb6b514f4


vllm_ascend/models/__init__.py
```diff
@@ -60,3 +60,6 @@ def register_model():
     ModelRegistry.register_model(
         "PanguProMoEForCausalLM",
         "vllm_ascend.models.pangu_moe:PanguProMoEForCausalLM")
+
+    ModelRegistry.register_model(
+        "SophonForCausalLM", "vllm_ascend.models.pangu_moe:SophonForCausalLM")
```
Yikun, line 65: This is a little bit wired for me, the XXXForCausalLM seems not same as model name. Please do a double confirm.

Yikun (commented):
As offline discussed, considering that the model has only been rigorously tested on the Ascend, it makes sense be here. But I still think the modeling.py should be included in vLLM in the long term.

Yikun:
Close also with https://github.com/vllm-project/vllm-ascend/pull/2100

PR #1841, opened 2025-07-17
[1/N][CustomOp] Register activation customop instead of overwrite forward_oot

MengqingCao:
### What this PR does / why we need it?
We'll refator `CustomOp` in vllm-ascend from this pr on. 

Use function `CustomOp.register_oot` to achieve the customop registery, taking `AscendQuickGELU` as an example:
```python
from vllm_ascend.ops.activation import AscendQuickGELU
CustomOp.register_oot(_decorated_op_cls=AscendQuickGELU, name="QuickGELU")
```

This is a quick adapt for `CustomOp.register_oot` mechanism from vllm 0.9.2. For further step, we can remove inherit from `QuickGELU` can write our own `QuickGELU` at all.

Part of https://github.com/vllm-project/vllm-ascend/pull/1647

### How was this patch tested?

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/8dfb45ca3379b3a789ec529af4bf725daa07f10d

PR #1932, opened 2025-07-22
[2/N][CustomOp] Register RotaryEmbedding instead of overwrite forward_oot

MengqingCao:

This pr register `RotaryEmbedding` and `DeepseekScalingRotaryEmbedding` instead of overwriting its `__init__` and `forward_oot`

Follows #1841
Part of https://github.com/vllm-project/vllm-ascend/pull/1647

N/A

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/533db0935da051ac793e8b22afbcb9ae9fa4255b

### DIFF
--- a/vllm_ascend/ops/rotary_embedding.py
+++ b/vllm_ascend/ops/rotary_embedding.py
@@ -21,6 +21,7 @@ from typing import Optional, Tuple
 import torch
 from vllm.model_executor.layers.rotary_embedding import (
     DeepseekScalingRotaryEmbedding, RotaryEmbedding)
+from vllm.platforms import current_platform
 
 from vllm_ascend.ascend_config import get_ascend_config
 from vllm_ascend.utils import enable_custom_op, is_310p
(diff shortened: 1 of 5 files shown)


vllm_ascend/ops/rotary_embedding.py
```diff
@@ -86,207 +87,163 @@ def rope_forward_oot(
...
+            self._yarn_get_mscale(self.scaling_factor, float(mscale)) /
+            self._yarn_get_mscale(self.scaling_factor, float(mscale_all_dim)) *
+            attn_factor)
+        super(DeepseekScalingRotaryEmbedding,
+              self).__init__(head_size, rotary_dim, max_position_embeddings,
+                             base, is_neox_style, dtype)
+        self.max_seq_len = max_position_embeddings
+        self._set_cos_sin_cache(max_position_embeddings,
+                                dtype=dtype,
+                                device=current_platform.device_type)
```
wangxiyuan: no need to use `current_platform`, just use the target value. or `NPUPlatform.device_type`

Use current_platform in vLLM Ascend is dengerous, it'll lead circle imoprt in some unknown case.

vllm_ascend/utils.py
```diff
@@ -481,6 +481,13 @@ def register_ascend_customop():
     CustomOp.register_oot(_decorated_op_cls=AscendQuickGELU, name="QuickGELU")
     CustomOp.register_oot(_decorated_op_cls=AscendSiluAndMul,
                           name="SiluAndMul")
+    from vllm_ascend.ops.rotary_embedding import (
```
wangxiyuan: combine `import` at the same place, L480

vllm_ascend/ops/rotary_embedding.py
MengqingCao: Done

MengqingCao:
CI failed and will address it asap. And I found that it is mainly because patching `forward_oot` in spec-decode doesn't work, and after the registration method is implemented, the following error raised
https://github.com/vllm-project/vllm-ascend/actions/runs/16442779387/job/46470024153?pr=1932

PR #1948, opened 2025-07-23
[Test] Add ut for files in /ops

SunnyLee151064:
Add some uts for files in folder /ops
No

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/b77c7d327f2a463bb9ef8be36f30e920bc066502

### DIFF
--- a/tests/ut/ops/test_cache.py
+++ b/tests/ut/ops/test_cache.py
@@ -1,0 +1,32 @@
+import torch
+
+from tests.ut.base import TestBase
+from vllm_ascend.ops.cache import concat_and_cache_mla
+
+
+class TestCache(TestBase):
+
+    def setUp(self):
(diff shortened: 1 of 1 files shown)


tests/ut/ops/test_cache.py
```diff
@@ -0,0 +1,32 @@
+import torch
+
+from tests.ut.base import TestBase
+from vllm_ascend.ops.cache import concat_and_cache_mla
```
wangxiyuan, line 4: concat_and_cache_mla is not used anywhere in vLLM Ascend. We can remove this file totally instead of writing ut for it.

SunnyLee151064: OK

PR #1956, opened 2025-07-23
[0.9.1] Add LMhead TP communication groups.

Angazenn:
In pure dp scenarios (such as DP32)， LMHead comptuation takes 1~2ms. In this PR we customize the parallelism of LMHead，enabling the separate TP of LMHead. The computation flow is listed as follows:

```
get_lmhead_group().all_gather  # [num_tokens, hid_dim] -->  [num_tokens * lmhead_tp, hid_dim]
--> lmhead matmul  # [num_tokens * lmhead_tp, hid_dim] -->  [num_tokens * lmhead_tp, vocab_size //  lmhead_tp]
--> get_lmhead_group().all_to_all  # [num_tokens * lmhead_tp, vocab_size //  lmhead_tp] --> [num_tokens, vocab_size]
```

this can decrease 0.5~1ms for deepseek with 28BS on a single die、MTP.

In addition, this PR also fixes a bug that introduced by LMHead quantization. The OP `npu_quant_matmul` only accepts dim < 65536, while `vocab_size` is > 65536 if using TP 1. We can set lmhead tp size > 1 to avoid this bug.

Main version of this PR: #2309 .

Yes. We introduced another configurable options `lmhead_tp_size` in ascend_config. For example:
```
additional_config={
        "lmhead_tp_size": 16,
}
```
The default value is -1, and `lmhead_tp_size` is automatically set to `tensor_parallel_size` in this case. Besides,  it is suggested to use it when running full DP to avoid additional communication introduced by TP. Therefore, the parallel size of `lmhead` group will also be changed to `tensor_parallel_size` if TP > 1 so as to fall back to normally TP+DP case.

### DIFF
--- a/vllm_ascend/distributed/utils.py
+++ b/vllm_ascend/distributed/utils.py
@@ -25,6 +25,11 @@ def is_lmhead_tp():
     # We only activate optimization of lmhead communication
     # when tp_size == 1, dp_size > 1 and lmhead_tp_size > 1.
 
+    try:
+        get_lmhead_group()
+    except AssertionError:
+        return False
+
     tp_size = get_tensor_model_parallel_world_size()
     dp_size = get_dp_group().world_size
     lmhead_tp_size = get_lmhead_group().world_size


vllm_ascend/ascend_config.py
```diff
@@ -65,6 +65,7 @@ def __init__(self, vllm_config):
             False)  # Whether to enable DeepSeek models' prefill optimizations
         self.enable_cpu_binding = additional_config.get(  # Whether to enable the cpu binding
             "enable_cpu_binding", False)
+        self.lmhead_tp_size = additional_config.get("lmhead_tp_size", -1)
```
Ronald1995, line 68: it's better that the default value is 1

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1769,6 +1785,20 @@ def _dummy_run(
...
                         **model_kwargs)
+
+            if not self.in_profile_run and is_lmhead_tp():
+                # lmhead_tp introduces additional communication across
+                # dp when computing logits. Hence we need to add it
+                # in profile_run.
+                if not with_prefill:
+                    padded_num_indices = num_tokens
+                else:
+                    padded_num_indices = max_num_reqs
```
momo609, line 1796: Will padding here cause the time to be longer when DP has a serious uneven load?

Angazenn: Yes, there might be performance degradation. However, in some cases (you can see `_get_forward_metadata_across_dp` ) the `all_reduce` communication used for gathering metedata is skipped. Thus using true `num_tokens_across_dp` will incur another `all_reduce` communication in this case. Maybe we can have a better solution for this.

vllm_ascend/distributed/parallel_state.py
```diff
@@ -6,19 +6,26 @@
...
+_LMHEAD: Optional[GroupCoordinator] = None
 
 
 def get_mc2_group() -> GroupCoordinator:
     assert _MC2 is not None, ("mc2 group is not initialized")
     return _MC2
 
 
+def get_lmhead_group() -> GroupCoordinator:
+    assert _LMHEAD is not None, ("mc2 group is not initialized")
```
MengqingCao:
```suggestion
    assert _LMHEAD is not None, ("lmhead group is not initialized")
```

vllm_ascend/distributed/parallel_state.py
```diff
@@ -41,9 +48,23 @@ def init_ascend_model_parallel(
                                      backend,
                                      group_name="mc2")
 
+    all_ranks = torch.arange(world_size).reshape(-1, lm_head_tp_size)
```
wangxiyuan: TODO: please create this parallel only when runing deepseek

vllm_ascend/distributed/parallel_state.py
Angazenn: done

vllm_ascend/distributed/parallel_state.py
Angazenn: done

wangxiyuan (approved):

PR #1966, opened 2025-07-23
[main] Optimize rope in Qwen3

rjg-lyh:
Optimize rope by extracting index_select from layers into model, which can reduce (layer_num -1) * 2 Gather ops in each prefill/decode stage. Simultaneously, The original operator is replaced with the more efficient operator.

No.

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/ebf7605b0dd58ff5d572d1918e52ca732025eee0

### DIFF
--- a/vllm_ascend/models/qwen3.py
+++ b/vllm_ascend/models/qwen3.py
@@ -4,15 +4,18 @@ from typing import Optional, Union
 import torch
 from torch import nn
 from transformers import Qwen3Config
+from vllm.attention import AttentionType
 from vllm.compilation.decorators import support_torch_compile
 from vllm.config import CacheConfig, VllmConfig
 from vllm.distributed import get_pp_group
+from vllm.model_executor.layers.layernorm import RMSNorm
 from vllm.model_executor.layers.logits_processor import LogitsProcessor
 from vllm.model_executor.layers.quantization import QuantizationConfig
+from vllm.model_executor.layers.rotary_embedding import RotaryEmbedding
 from vllm.model_executor.layers.vocab_parallel_embedding import ParallelLMHead
 from vllm.model_executor.models.interfaces import SupportsLoRA, SupportsPP
 from vllm.model_executor.models.qwen2 import Qwen2Model
-from vllm.model_executor.models.qwen3 import Qwen3DecoderLayer
+from vllm.model_executor.models.qwen3 import Qwen3Attention, Qwen3MLP
 from vllm.model_executor.models.utils import (AutoWeightsLoader,
                                               PPMissingLayer, maybe_prefix)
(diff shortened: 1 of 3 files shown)


vllm_ascend/models/qwen3.py
```diff
@@ -21,7 +24,43 @@
...
+        # Add qk-norm
+        q_by_head = q.view(*q.shape[:-1], q.shape[-1] // self.head_dim,
+                           self.head_dim)
+        q_by_head = self.q_norm(q_by_head)
+        q = q_by_head.view(q.shape)
+        k_by_head = k.view(*k.shape[:-1], k.shape[-1] // self.head_dim,
+                           self.head_dim)
+        k_by_head = self.k_norm(k_by_head)
+        k = k_by_head.view(k.shape)
+        if type(self.rotary_emb) is RotaryEmbedding:
```
ApsarasX, line 47: Doesn't Qwen3's model all use RotaryEmbedding? So technically this 'if' condition has to be True, right?

vllm_ascend/models/qwen3.py
```diff
@@ -30,31 +69,99 @@ def __init__(
...
-                      AscendW8A8LinearMethod):
-            self.post_attention_layernorm = AddRMSNormW8A8Quant(
-                config.hidden_size,
-                layer=self.mlp.gate_up_proj,
-                eps=config.rms_norm_eps)
+    def forward(
+        self,
+        positions: torch.Tensor,
+        cos: torch.Tensor,
+        sin: torch.Tensor,
```
ApsarasX, line 144: Changing so many function signatures just for `cos` and `sin` seems like overkill. What if we put them in a global context instead, like we do with `forward_context`?

vllm_ascend/models/qwen3.py
```diff
@@ -77,6 +184,57 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
...
+                hidden_states = self.get_input_embeddings(input_ids)
+            residual = None
+        else:
+            assert intermediate_tensors is not None
+            hidden_states = intermediate_tensors["hidden_states"]
+            residual = intermediate_tensors["residual"]
+
+        # Generate cos and sin outside layers to avoid repeated calculation.
+        cos, sin = None, None
+        if self.cos_sin_cache is not None:
```
ApsarasX, line 212: Only when if `is_cos_sin_cached and neox_style and self.head_size == self.rotary_dim and self.head_size == 128` is True will the `cos` and `sin` here be used, so should we control the generation conditions of cos and sin to avoid redundant computation?

wangxiyuan:
we won't change model arch anymore. close this PR now.

PR #1969, opened 2025-07-23
[CI] Update CI models

Potabk:

1. Update `test_ilama_lora.py` to use `vllm-ascend` model
2. Update `test_aclgraph.py` to use modelscope models


Potabk:
this is only suitable for branch `main`, no need to update for v0.9.1, close it

PR #1973, opened 2025-07-23
[main] Optimize the performance of Qwen2.5 with AddRmsNormQuant ops and Rope optimization

rjg-lyh:
This PR optimizes the Qwen2.5 model based on the customModel with the following two improvements:

①add the AddRmsNormQuant operation to fuse AddRmsNorm and W8A8 quantization ops together.

②Optimize rope by extracting index_select from layers into model, which can reduce (layer_num -1) * 2 Gather ops in each prefill/decode stage. Simultaneously, The original operator is replaced with the more efficient operator.

No.

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/90eeea8f8501f22151d08e57d6f18f639c9faf2f


ApsarasX:
By the way, could you add the `AddRmsNormDyanamicQuant` operator in this PR?

wangxiyuan:
This can be closed now. We won't change model arch anymore.

PR #1981, opened 2025-07-24
[Doc] Fix running 'bash format.sh ci' on local machine "can not find vllm source path error"

leo-pony:
[Doc] Fix running 'bash format.sh ci' on local machine "can not find vllm source path error"

### DIFF
--- a/docs/source/developer_guide/contribution/index.md
+++ b/docs/source/developer_guide/contribution/index.md
@@ -44,6 +44,7 @@ git clone --branch |vllm_version| https://github.com/vllm-project/vllm.git
 cd vllm
 pip install -r requirements/build.txt
 VLLM_TARGET_DEVICE="empty" pip install .
+export PYTHONPATH="$(pwd):${PYTHONPATH}"
 cd ..
 
 # Install requirements


docs/source/developer_guide/contribution/index.md
```diff
@@ -44,6 +44,7 @@ git clone --branch |vllm_version| https://github.com/vllm-project/vllm.git
 cd vllm
 pip install -r requirements/build.txt
 VLLM_TARGET_DEVICE="empty" pip install .
+export PYTHONPATH="$(pwd):${PYTHONPATH}"
```
Yikun: maybe we can just create venv after vllm-project?

wangxiyuan: I think this is a specified problem, not the common error. it's clear that you installed vllm from source in a incorrect folder

leo-pony: @Yikun @wangxiyuan I changed the solution from adding code diretlly to the "normal path" to "specified use case" note. Could you help review once more?

docs/source/developer_guide/contribution/index.md
```diff
@@ -58,6 +58,17 @@ cat requirements.txt | grep -Ev '^#|^--|^$|^-r' | while read PACKAGE; do pip ins
 bash format.sh ci
 ```
 
+Note: If you did not create the Python .venv in the "Running lint Locally" section above,
```
wangxiyuan, line 61: It's quite strange. Even users run lint without .venv, mypy should find the requirement in system PYTHONPATH, why users need to add the pwd to PYTHONPATH?

PR #2008, opened 2025-07-25
[CI] Speed up CI

Potabk:

Currently our workflow run time takes about 3 hours in total, which seriously affects the developer experience, so I Increased concurrency of multi-card testing


MengqingCao:
LGTM, I think we can merge this firstly to speed up CI

PR #2023, opened 2025-07-25
[Bugfix] Add get_supported_tasks interface to fix broken CI

Potabk:
Added `get_supported_tasks` interface to adapt to vllm [changes](https://github.com/vllm-project/vllm/commit/46d81d69511ac11f27d19af80d19dd7b2cce8613#diff-80ee7e2a62f9dcfbb8a312dc4e3948557e97ef187290daebbcae1e28596bda29)

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/5ac3168ee342f4cae17b0b67375e647bd5dd9151

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -92,6 +92,10 @@ if vllm_version_is("0.9.2"):
     from vllm.model_executor.models.interfaces import has_step_pooler
     from vllm.v1.utils import bind_kv_cache
 else:
+    from vllm.model_executor.models.interfaces import supports_transcription
+    from vllm.model_executor.models.interfaces_base import \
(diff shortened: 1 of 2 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -706,6 +710,31 @@ def get_eagle_atten_dict(
     def get_model(self) -> nn.Module:
         return self.model
 
+    def get_supported_generation_tasks(self) -> "list[GenerationTask]":
```
wangxiyuan, line 713: we should inherit from base runner in vLLM in the future instead of copy the code again and again. I'll do the refactor later

Yikun:
Let's recover CI first.

PR #1800, opened 2025-07-15
[V1][PP] Support pp with ray backend in V1

MengqingCao:
### What this PR does / why we need it?
Support pipeline parallel with ray backend in V1Engine.

Fixes #1751

### Does this PR introduce _any_ user-facing change?
Users could specify ray as distributed backend when inferencing with pp

### How was this patch tested?
CI passed with new added test.

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/32142b3c62277ac7cb941f2036270decb6b514f4

PR #2027, opened 2025-07-26
[0.9.1][V1][PP] Support pp with ray backend in V1 (#1800)

MengqingCao:
  * Support pipeline parallel with ray backend in V1Engine.
This is a backport of #1800 

Users can use pp with ray backend with this pr


MengqingCao:
closing as not planned

PR #2028, opened 2025-07-26
[Build][Ray] Fix protobuf version in Dockerfile

MengqingCao:
Fix protobuf version in Dockerfile to resolve `AttributeError: 'str' object has no attribute 'DESCRIPTOR' when packaging message to dict` using protobuf. will remove version specification after https://github.com/ray-project/ray/pull/54910 is merged

N/A

CI passed with existing test.

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/0e36abf9931baa070609376debb4fb3772f4a3fe

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,7 +54,7 @@ RUN export PIP_EXTRA_INDEX_URL=https://mirrors.huaweicloud.com/ascend/repos/pypi
     python3 -m pip cache purge
 
 # Install modelscope (for fast download) and ray (for multinode)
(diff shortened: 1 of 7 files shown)


MengqingCao:
@Yikun plz take a look, thanks!

Yikun:
Should this PR backport to v0.9.1?

MengqingCao:
I found the same issue on V0 + ray, thus will backport this on v0.9.1-dev

PR #2029, opened 2025-07-26
Refactor accuracy test CI

wxsIcey:

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/2f6e6b33fb0acf5331982c8e1ea620005e3a19ba


Yikun:
Here is the initial framework we can have some discuss in next week: https://github.com/vllm-project/vllm-ascend/pull/2044

PR #2031, opened 2025-07-26
[Feature] Add cpu binding

SunnyLee151064:

Add Cpu Binding Feature

No

### DIFF
--- a/vllm_ascend/envs.py
+++ b/vllm_ascend/envs.py
@@ -106,6 +106,9 @@ env_variables: Dict[str, Callable[[], Any]] = {
     "VLLM_ASCEND_MODEL_EXECUTE_TIME_OBSERVE":
     lambda: bool(int(os.getenv("VLLM_ASCEND_MODEL_EXECUTE_TIME_OBSERVE", '0'))
                  ),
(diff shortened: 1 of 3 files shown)


jianzs:
When does binding a CPU offer benefits?  If it always helps, we should remove the extra environment variable and enable it by default.

ganyi1996ppo:
> When does binding a CPU offer benefits? If it always helps, we should remove the extra environment variable and enable it by default.

cpu binding usually used to prevent memory transition cross numa, helps alleviate the host speed unalignment issue between the different npu card

ganyi1996ppo:
@SunnyLee151064 Have you test the performance boost rate of this PR? We don't have to merge this in dev-0.9.1 if the performance boost is not that large

wangxiyuan:
does cpu bind aciton need root privilege? If yes, please write more note for the config option. And does this feature works in container as well?

SunnyLee151064:
> does cpu bind aciton need root privilege? If yes, please write more note for the config option. And does this feature works in container as well?

It also works well when not in root , I test this feature all in root privilege, and in a container.

PR #2032, opened 2025-07-26
[Feature] Add bind cpu

SunnyLee151064:

Add cpu binding feature

Not

### DIFF
--- a/vllm_ascend/worker/worker_v1.py
+++ b/vllm_ascend/worker/worker_v1.py
@@ -38,11 +38,13 @@ from vllm.v1.kv_cache_interface import KVCacheConfig, KVCacheSpec
 from vllm.v1.outputs import ModelRunnerOutput
 from vllm.v1.worker.worker_base import WorkerBase
 
+import vllm_ascend.envs as ascend_envs
 from vllm_ascend.ascend_config import init_ascend_config
 from vllm_ascend.device_allocator.camem import CaMemAllocator
 from vllm_ascend.platform import NPUPlatform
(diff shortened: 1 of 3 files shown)


jianzs:
When does binding a CPU offer benefits?  If it always helps, we should remove the extra environment variable and enable it by default.

vllm_ascend/worker/worker_v1.py
```diff
@@ -133,6 +135,16 @@ def init_device(self):
...
+        # Bind cpu
+        if ascend_envs.VLLM_ASCEND_CPU_BINDING:
+            try:
+                bind_cpus(self.local_rank, ratio=1.0)
+            except RuntimeError as e:
+                logger.error(f"{e} in {self.local_rank}")
+            except ValueError as e:
+                logger.error(f"{e} in {self.local_rank}")
+            except Exception:
+                logger.info("Skip binding cpu.")
```
jianzs: We should report the error, either by raising an exception or logging it, to inform the user.

MengqingCao:
CI seems failed due to an unrelated case, we're looking into it offline. This is an occasional problem on the new CI machine, not on the original one.

vllm_ascend/worker/worker_v1.py
SunnyLee151064: OK，I will modify it.

SunnyLee151064:
> When does binding a CPU offer benefits? If it always helps, we should remove the extra environment variable and enable it by default.

This feature currently offers limited benefits, and after review, this activation method will be retained for the time being.

PR #2037, opened 2025-07-26
add ut for device allocator/camem and mutistream/layers

1024daniel:

test device allocator/camem and mutistream/layers contains resource allocation and stream ops

N/A
How was this patch tested?

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/2836dd73f13015ee386c544760ca0d16888203f3

### DIFF
--- a/tests/ut/device_allocator/test_camem.py
+++ b/tests/ut/device_allocator/test_camem.py
@@ -1,0 +1,188 @@
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


tests/ut/device_allocator/test_camem.py
```diff
@@ -0,0 +1,191 @@
...
+import pytest
+import torch
+
+from vllm_ascend.device_allocator.camem import (AllocationData, CaMemAllocator,
+                                                create_and_map,
+                                                find_loaded_library,
+                                                get_pluggable_allocator,
+                                                unmap_and_release)
+
+
```
wangxiyuan, line 28: Please add a test wrapper class like other ut does to limit the scope. Thanks

wangxiyuan (approved):

PR #684, opened 2025-04-27
[Feature][1/2] Impl the connector based on the llmdatadist for v1

jianzs:
This PR implements the connector functionality for NPU based on LLMDataDist, building upon the connector API merged in vLLM v1. (https://github.com/vllm-project/vllm/pull/15960) We've successfully tested various scenarios in offline environments:

- Single-machine: Verified 2P2D testing with dense models (Llama) and MoE models (DeepSeek v2 Lite)
- Two-machine: Completed 1P1D testing with DeepSeek R1 W8A8

Key implementation aspects include:

**Cross-machine PD**: LLMDataDist requires NPU device IP for connection establishment. Our approach utilizes a global rank table (JSON) on each machine containing:
- Unique server IDs
- IP addresses and device IDs for each card
- Server ID specification in connector extra config at startup for instance information retrieval

**nPmD**: Given that the community's nPmD design, particularly the router component API, is still evolving, we've implemented a solution using a meta server component (to be provided separately) that:
- Records prefill completion details (device and dp rank information)
- Responds to decode node queries with prefill node locations
- Enables decode nodes to retrieve data from appropriate prefill nodes

We propose initially merging the 1P1D implementation, where the global rank table contains information for two nodes, allowing direct prefill node identification. The nPmD implementation can be refined and merged following community discussion.

Todo:
- [x] Implement 1P1D (one prefill, one decode) configuration support
- [ ] Add sample script for automatic global rank table generation
- [ ] Document global rank table format specifications
- [ ] Provide user guide for PD (Prefill-Decode) functionality
- [ ] Add test cases

re #448 

---

Note:
A minor modification to vLLM's codebase is required to run this example successfully. The patch enables the scheduler process to locate the appropriate connector class by importing the necessary module.

The change should be made in `vllm/v1/core/sched/scheduler.py`, adding an import statement for `vllm_ascend.distributed`.

**This is a temporary solution, and we need to implement a more elegant module discovery mechanism.**
```diff
diff --git a/vllm/v1/core/sched/scheduler.py b/vllm/v1/core/sched/scheduler.py
index 69e7cc8ee..b15525971 100644
--- a/vllm/v1/core/sched/scheduler.py
+++ b/vllm/v1/core/sched/scheduler.py
@@ -31,6 +31,8 @@ from vllm.v1.structured_output import StructuredOutputManager
 
 logger = init_logger(__name__)
 
+# TODO(jianzs): Find the suitable place to put this.
+import vllm_ascend.distributed
 
[...]

PR #2039, opened 2025-07-26
Added support for KV connector v1

marcobarlo:
- This PR adds the support for the KV connector interface in the V1 architecture, in the same way as vllm. Vllm-ascend currently lacks of this support, required to support also layerwise management of KV caches.

- The connector interface allows using external tools and integrate them with vllm

### Notes:
We are aware of Issue #684 , however that issue does not modify the attention classes as necessary to perform layerwise management of KV caches required for connectors like LMCache.

The PR does not modify current APIs, but it extends the behavior of current worker runner and attention classes to save and load KV caches. In absence of connectors, the behavior should stay untouched.

- No unit test implemented yet for the worker.

- Tested together with LMCache using https://github.com/LMCache/LMCache/blob/dev/examples/kv_cache_reuse/local_backends/offload.py with the following models:
1 Deepseek-R1-Distill-Qwen-1.5B
2 Qwen3-30B-A3B
3 Deepseek-v2-lite
4 Llama-3.1-8B
LMCache used in both layerwise and non-layerwise mode.

- Performed LMEval on LMCache integrated with vllm-ascend.
Results without LMCache on Qwen3-8B:
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8400|±  |0.0101|
|     |       |strict-match    |     5|exact_match|↑  |0.8355|±  |0.0102|
 
 
Results with LMCache Layerwise:
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8385|±  |0.0101|
|     |       |strict-match    |     5|exact_match|↑  |0.8332|±  |0.0103|

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -28,6 +28,9 @@ from vllm.forward_context import ForwardContext, get_forward_context
 from vllm.utils import direct_register_custom_op
 from vllm.v1.core.sched.output import SchedulerOutput
 from vllm.v1.worker.gpu_input_batch import InputBatch
+from vllm.distributed.kv_transfer import (get_kv_transfer_group,
+                                          has_kv_transfer_group,
+                                          is_v1_kv_transfer_group)
 
 from vllm_ascend.ops.attention import vanilla_chunked_prefill
 from vllm_ascend.utils import (ACL_FORMAT_FRACTAL_NZ, aligned_16, is_310p,
@@ -200,6 +203,36 @@ class AscendAttentionMetadataBuilder:
             with_prefill_across_dp=with_prefill_across_dp)
         return attn_metadata
 
+def wait_for_kv_layer_from_connector(layer_name: str):
+    if not has_kv_transfer_group() or not is_v1_kv_transfer_group():
+        return
+
+    connector = get_kv_transfer_group()
+
+    forward_context: ForwardContext = get_forward_context()
+    attn_metadata = forward_context.attn_metadata
+    if attn_metadata is None:
+        return
+    # TODO: assert ascendMetadata
+    connector.wait_for_layer_load(layer_name)
+
+
+def maybe_save_kv_layer_to_connector(
+    layer_name: str,
+    kv_cache_layer: List[torch.Tensor],
+):
+    if not has_kv_transfer_group() or not is_v1_kv_transfer_group():
+        return
+
+    connector = get_kv_transfer_group()
+
+    forward_context: ForwardContext = get_forward_context()
+    attn_metadata = forward_context.attn_metadata
+    if attn_metadata is None:
+        return
+    # TODO: assert ascendMetadata
+    connector.save_kv_layer(layer_name, kv_cache_layer,
+                            attn_metadata)
 
 class AscendAttentionBackendImpl(AttentionImpl):
 
@@ -467,6 +500,7 @@ def unified_ascend_attention_with_output(
     output: torch.Tensor,
     layer_name: str,
 ) -> None:
+    wait_for_kv_layer_from_connector(layer_name)
     forward_context: ForwardContext = get_forward_context()
     attn_metadata = forward_context.attn_metadata
     self = forward_context.no_compile_layers[layer_name]
@@ -479,6 +513,7 @@ def unified_ascend_attention_with_output(
                       attn_metadata,
                       output,
                       trace_flag=False)
+    maybe_save_kv_layer_to_connector(layer_name, kv_cache)
     return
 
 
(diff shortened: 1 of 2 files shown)


jianzs:
Please ensure the modification works with torchair graph mode.

LCAIZJ:
Great work on this! I have two follow-up questions to clarify:
1. What do you mean by "vanilla vllm"?
2. LMCache support KvCache transfer in Ascend?
Could you please address these when you have a chance? Thanks!

marcobarlo:
@LCAIZJ 
1. Sorry, by "vanilla vllm" I meant the upstream vllm code for GPUs and other accelerators, without vllm-ascend ... specifically here -> https://github.com/vllm-project/vllm/blob/main/vllm/attention/layer.py 
2. We are working on it, we have a working version, upstreaming soon: https://github.com/LMCache/LMCache/issues/807 
This PR was required to create the upcoming LMC PR.

@jianzs  My bad, I rebased from previous version and didn't see the torchair file. Looks like it does not support Deepseek lite. I am looking for a way to test without running a full deepseek. Any advice for small models?

wangxiyuan:
Thanks for the contribution. @Potabk please test this PR with LMcache. Thanks.

Potabk:
why don't we directly use the vllm interface

marcobarlo:
@Potabk the vllm implementations are in vllm/attention/layer.py, and makes assumptions on the attention objects. For example, at the current status, the vllm function asserts that attention metadata is a dict and then accesses the metadata of specific layer, which does not hold for vllm-ascend.
I don't think is wise to rely on something internally implemented to the vllm attention module. 

@wangxiyuan Please notice that we worked on 0.9.2: the LMCache PR works for 0.9.2, when kvcaches was a list of tensors and not a list of tuples. This PR was developed before the commit df0ec55162339c08b0bfcb79a2dcfac89f8c6e33 (https://github.com/marcobarlo/vllm-ascend/commit/df0ec55162339c08b0bfcb79a2dcfac89f8c6e33). I had to modify accordingly this PR to avoid conflicts with that commit, which was integrating the KV connector for the model_runner_v1.py file. However the same commit also changes the kvcache from list of tensors to list of tuples. If you want to test this PR together with LMCache, I would suggest to cherrypick this PR commit on the 0.9.2rc, and integrate the required modifications in the model_runner_v1.py file, which are part of the df0ec55162339c08b0bfcb79a2dcfac89f8c6e33 commit.
In particular, the required pieces of code from https://github.com/marcobarlo/vllm-ascend/commit/df0ec55162339c08b0bfcb79a2dcfac89f8c6e33 are:
L20; L42 to L44; L47; L1158; L1189 to L1191; L1454 to L1460; L1644 to L1686.
Essentially, all the code related to KV connectors without disaggregation.

We will try to port LMCache to 0.10 and deal with list of tuples in the next couple of weeks.
  
@jianzs Still working on the torchair attention file, graph compilation is causing some issues. I'll get back as soon as possible. Any suggestion is welcome

Potabk:
@wangxiyuan I think this will not affect the normal logic of LLMDataDistCMgrConnector that we have currently implemented
@marcobarlo if I want test this with llmcache, what should I do. We haven't even implemented an llmcache-based connector yet

marcobarlo:
@Potabk I would suggest one of the two ways:
- Apply the commit of this PR to 0.9.2 and add the kv connector calls in the model_runner_v1.py file as I specified in the previous message (this is what I did); or
- Use the code of this PR but revert the kvcaches to be list of tensors (currently should be list of tuples). 
Either way you prefer, but LMCache at the moment assumes the kvcaches are list of tensors. 

Feel free to contact me in pvt if you need support

wangxiyuan:
This is not needed anymore, right?

marcobarlo:
@wangxiyuan LMCache-Ascend still requires it. You can look into https://github.com/LMCache/LMCache-Ascend/blob/main/docker/Dockerfile.a2.openEuler in the PR https://github.com/LMCache/LMCache-Ascend/pull/1 how it is built. Why do you think is not needed?

wangxiyuan:
OK. plesase let the CI pass. Then I think we can merge this.

marcobarlo:
@wangxiyuan is the CI ok? I rebased the commit on the latest main, but the tests look like failing and not due to the modifications of this PR. I saw a commit today called "Fix broken CI"

wangxiyuan:
@marcobarlo we noticed this error, it's caused by https://github.com/vllm-project/vllm-ascend/issues/2546 feel free to ignore it. We'll fix it asap

wangxiyuan (approved):

matthewygf:
@wangxiyuan Hi, once this is ready, we aim to start our initial support for v0.10.x

PR #2044, opened 2025-07-26
new acc test

Yikun:


.github/workflows/accuracy_test_new.yaml
```diff
@@ -0,0 +1,154 @@
...
+    if:  >-
+      ${{
+      contains(github.event.pull_request.labels.*.name, 'accuracy-new-test') &&
+      contains(github.event.pull_request.labels.*.name, 'ready-for-test') ||
+      github.event_name == 'workflow_dispatch' || github.event_name == 'schedule'
+      }}
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        include:
```
Yikun, line 69: After new added model we just need to
- add matrix.include to make sure the model<->runner mapping
- Add xxx.yaml and test_xxx.py

PR #2045, opened 2025-07-27
Update version info in conf.y to fix doctest in v0.9.1

Yikun:
doctest passed

CI passed, verfify manually


Yikun:
Emm:

v0.9.1rc1 also has the problem: https://github.com/vllm-project/vllm-ascend/blob/v0.9.1rc1/requirements.txt

But already fix in v0.9.1-dev, so we can only resolve the issue in v0.9.1rc2:
https://github.com/vllm-project/vllm-ascend/blob/03b2910364c4e40d968f6d66992ea49119d80dee/requirements.txt#L30

Yikun:
Let's merge this after v0.9.1rc2 bump

Yikun:
replaced by https://github.com/vllm-project/vllm-ascend/pull/2233

PR #2048, opened 2025-07-27
Pin transformers to fix v0.9.1 doctest

Yikun:
Pin transformers to fix v0.9.1 doctest

No

doctest passed

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/c6573698414c88990c3ddcd5386d26fb5911c59d

### DIFF
--- a/tests/e2e/doctests/002-pip-binary-installation-test.sh
+++ b/tests/e2e/doctests/002-pip-binary-installation-test.sh
@@ -49,6 +49,12 @@ function install_binary_test() {
     pip install vllm=="$(get_version pip_vllm_version)"
     pip install vllm-ascend=="$(get_version pip_vllm_ascend_version)"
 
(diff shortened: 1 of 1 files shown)


wangxiyuan:
commit to main, then pick to 0.9.1-dev?

Yikun:
@wangxiyuan No, only main doctest will run, because all doctest will be replaced with main code.

https://github.com/vllm-project/vllm-ascend/blob/e3a2443c3a61f0f8376e75df4083a1554595af68/.github/workflows/vllm_ascend_doctest.yaml#L74-L80

I update this to avoid 0.9.1-dev doctest be break by other unexpected change.

Yikun:
This is just for test easyily, we can just release v0.9.1rc2, before release we can trigger doctest manually

Yikun:
Only effect on doctest so just merge this.

PR #2050, opened 2025-07-28
[Refactor]Refactor sampler

wangxiyuan:
Refactor Sampler implementation from patch way to inherit from vLLM Sampler interface.

Next step: Make the op `TopKTopPSampler` in vLLM support custom ops register mechanism 

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/61a6905ab036fd00eafdb1b0ca130d5feccfe686


wangxiyuan:
@Pr0Wh1teGivee please help review. Thanks.

Pr0Wh1teGivee:
LGTM

tests/ut/sample/test_sampler.py
```diff
@@ -0,0 +1,32 @@
+import unittest
+from unittest import mock
+
+import torch
+
+from vllm_ascend.sample.sampler import AscendSampler, AscendTopKTopPSampler
+
+
+class TestAscendSampler(unittest.TestCase):
```
Yikun: Better to use testbase to avoid testcase being impacted by other future patches

vllm_ascend/sample/sampler.py
```diff
@@ -0,0 +1,59 @@
...
+        logits: torch.Tensor,
+        k: torch.Tensor,
+        p: torch.Tensor,
+    ) -> torch.Tensor:
+        if p is not None and k is not None:
+            # npu_top_k_top_p's parameter order is (logits, p, k), not (logits, k, p)
+            return torch_npu.npu_top_k_top_p(logits, p, k)
+
+        probs = logits.softmax(dim=-1)
+        probs_sort, _ = probs.sort(dim=-1, descending=False)
```
Yikun, line 31: Unrelated: it seems `probs` and `probs_sort` compution are ununsed `if p is None and k is None`, the logits can return directly.

tests/ut/sample/test_sampler.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def test_init_with_raw_logprobs(self):
+        sampler = AscendSampler(logprobs_mode="raw_logprobs")
+        self.assertEqual(sampler.logprobs_mode, "raw_logprobs")
+        self.assertTrue(hasattr(sampler, 'topk_topp_sampler'))
+        self.assertIsInstance(sampler.topk_topp_sampler, AscendTopKTopPSampler)
+
+
+class TestAscendTopKTopPSampler(unittest.TestCase):
```
Yikun: same

Yikun (approved):

vllm_ascend/sample/sampler.py
Yikun: cc @Pr0Wh1teGivee

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -168,7 +168,15 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
...
-        self.sampler = Sampler()
+        if envs.VLLM_ASCEND_ENABLE_TOPK_TOPP_OPTIMIZATION:
+            # TODO: drop the env config to use ascend sampler by default
+            from vllm_ascend.sample.sampler import AscendSampler
+
+            self.sampler = AscendSampler()
+        else:
+            from vllm.v1.sample.sampler import Sampler
+
+            self.sampler = Sampler()
```
weijinqian0, line 176: Can it take effect directly by removing the flag?

wangxiyuan: yes, I hope so. Let's have a fully test to ensure ascend sampler is stable enough. Then we can remove the flag

vllm_ascend/sample/sampler.py
wangxiyuan: Done

vllm_ascend/worker/model_runner_v1.py
wangxiyuan: I set the value to True to enable ascend sampler by default now.

Yikun (approved):

PR #2058, opened 2025-07-28
Br gloo fix hccl

zhaowei1936:


wangxiyuan:
take https://github.com/vllm-project/vllm/pull/22243 instead

PR #2059, opened 2025-07-28
[bugfix] fix prefix cache in qwen

lidenghui1110:

We find that performance become worse after prefix cache enabled in qwen3-235B model, and figure out that it is caused by attn_state = ChunkedPrefill, which makes it using torch_npu._npu_paged_attention_splitfuse rather than torch_npu._npu_flash_attention_qlens in attention when prefix cache hit.

So we should let attn_state = PrefillCacheHit when all the requests are prefill in current batch. In this case，_npu_flash_attention_qlens performance is better than _npu_paged_attention_splitfuse.

performance data:
before this fix
image
after this fix
image


wangxiyuan:

PR #2062, opened 2025-07-28
[Misc] Refractor forward metadata retrieval across DP nodes to reduce redundant padding.

yx0716:
Before refactoring cross-DP decoding metadata aggregation, clean up the token‐padding logic .
### What this PR does：

1. First checks whether any DP instance is in the prefill phase.

2. If in the `decode` phase and `torchair_graph_enabled `is true, pads each DP instance’s token count up to the global maximum.

3. If in the `prefill` phase, or in decode phase with graph mode **disabled**, returns each DP instance’s original token count without padding.

This reordering removes the previous two‐step padding/unpadding flow and ensures padding only occurs when strictly necessary.


ApsarasX:
I think the third scenario should be considered, in the `decode` phase and `torchair_graph_enabled` is true, but `with_prefill` is true

yx0716:
> I think the third scenario should be considered, in the `decode` phase and `torchair_graph_enabled` is true, but `with_prefill` is true

- In a single DP instance, TorchAir’s graph mode and chunk-prefill mode are not enabled at the same time, thus prefill and decode will not exist simultaneously. 
- Across multiple DP instances, if even one instance enters the prefill stage, all instances automatically revert to non-graph mode. 

Given these two scenarios, it seems unlikely that a third case would arise at this point.

wangxiyuan (approved):

wangxiyuan:
@ApsarasX I'm working on torchair model runner refacor, let's do more work there https://github.com/vllm-project/vllm-ascend/pull/2204

ApsarasX:
> @ApsarasX I'm working on torchair model runner refacor, let's do more work there #2204

OK

ApsarasX (approved):

PR #2065, opened 2025-07-28
[CI] Enable linux-aarch64-a2 (64GB) and tp2 * 2 max-parallel to speed up CI

Potabk:
Currently our workflow run time takes about 3 hours in total, which seriously affects the developer experience, so it is urgent to have a optimization, after this pr, It is expected that the running time of the full CI can be shortened to 1h40min.

- Enable linux-aarch64-a2 (64GB) to replace linux-arm64-npu (32GB)
- Change TP4 ---> TP2 * 2 max-parallel
- Move DeepSeek-V2-Lite-W8A8 to single card test

No

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/a2480251ec92ba2a849464dde48db8a2b7f6ef81


Yikun:
cc @wangxiyuan @ganyi1996ppo @jianzs @ApsarasX @zzzzwwjj @yiz-liu @whx-sjtu @Angazenn @mengwei805 

FYI, after this PR we will use A2 (64GB) in CI

Yikun:
https://github.com/vllm-project/vllm-ascend/blob/main/benchmarks/scripts/run_accuracy.py
https://github.com/vllm-project/vllm-ascend/blob/main/tests/e2e/multicard/test_offline_inference_distributed.py

Seems also need change?

And many func should be added on yaml, this can be done in new PR

```
def test_models_distributed_pangu():
def test_models_distributed_topk() -> None:
def test_models_distributed_Qwen3_W8A8():
```
Should be included

Issue #1980, opened 2025-07-24
[Feature]: Bump default Python version to 3.11 series


Yikun:
### 🚀 The feature, motivation and pitch

- [ ] Upgrade cann image to `8.2.rc1-xxx-xxx-py3.11`
- [ ] tests, ut / e2e / 
- [ ] docs
- [ ] https://github.com/search?q=repo%3Avllm-project%2Fvllm-ascend%203.10&type=code

Yikun:
We can do it after
- [x] CANN release  https://github.com/vllm-project/vllm-ascend/pull/1653
- [x] vllm ascend v0.10.0 release tagged https://github.com/vllm-project/vllm-ascend/pull/1927
- [ ] torch npu v2.7.1 upgrade https://github.com/vllm-project/vllm-ascend/pull/1562

Yikun:
Close by: https://github.com/vllm-project/vllm-ascend/pull/2072

PR #2072, opened 2025-07-28
bump default python version to 3.11

ChenTaoyu-SJTU:
Bump default python version to 3.11, see #1980 

no
pass CI


.pre-commit-config.yaml
```diff
@@ -71,9 +71,9 @@ repos:
     language: system
     types: [python]
     stages: [manual] # Only run in CI
-  - id: mypy-3.10 # TODO: Use https://github.com/pre-commit/mirrors-mypy when mypy setup is less awkward
-    name: Run mypy for Python 3.10
-    entry: tools/mypy.sh 1 "3.10"
+  - id: mypy-3.11 # TODO: Use https://github.com/pre-commit/mirrors-mypy when mypy setup is less awkward
```
Yikun: This shouldn't be changed

ChenTaoyu-SJTU: Should I keep the original code like this:
```
  - id: mypy-3.10 # TODO: Use https://github.com/pre-commit/mirrors-mypy when mypy setup is less awkward
    name: Run mypy for Python 3.10
    entry: tools/mypy.sh 1 "3.10"
```
Or just remove the `# TODO` comment and just keep:
```
  - id: mypy-3.11
    name: Run mypy for Python 3.11
    entry: tools/mypy.sh 1 "3.11"
```

Yikun: keep the original code, this PR only make 3.11 as default, but not drop 3.10 support

PR #2074, opened 2025-07-28
Fix for tp execute model output

marcobarlo:
- Fix for model running in tensor parallelism: without this modification, with vllm tag v0.10.0 and current vllm_ascend main branch, models can return None, which is then accessed in "update_finished_set(output.finished_sending". This causes an exception.

- Fixes: Ported code from gpu_worker in vllm, which checks for kv_transfer_group.

No function signature modified. The execute_model function has a modified behavior to check for finished transfers of KV.

### DIFF
--- a/vllm_ascend/worker/worker_v1.py
+++ b/vllm_ascend/worker/worker_v1.py
@@ -19,6 +19,7 @@
 
 from typing import Optional
 
+import copy
 import torch
 import torch.nn as nn
 import torch_npu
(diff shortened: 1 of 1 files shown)


wangxiyuan:

weijinqian0:
Does this problem still exits?

wangxiyuan:
no update for long time. close it now. Feel free to create a new one if it's still necessary

PR #2079, opened 2025-07-29
Modify the obtain methods of card id and hccn_tool_path

yangqinghao-cmss:

### DIFF
--- a/examples/disaggregated_prefill_v1/gen_ranktable.py
+++ b/examples/disaggregated_prefill_v1/gen_ranktable.py
@@ -26,6 +26,9 @@ print("enter py")
 
 hccn_tool_path = os.environ.get("HCCN_TOOL_PATH",
                                 "/usr/local/Ascend/driver/tools/hccn_tool")
+if not os.path.exists("/usr/local/Ascend/driver/tools/hccn_tool"):
+    hccn_tool_path = "/usr/bin/hccn_tool"
+
 master_addr = os.environ.get("MASTER_ADDR")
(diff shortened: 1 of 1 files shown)


wangxiyuan (dismissed):

examples/disaggregated_prefill_v1/gen_ranktable.py
```diff
@@ -26,6 +26,9 @@
 
 hccn_tool_path = os.environ.get("HCCN_TOOL_PATH",
                                 "/usr/local/Ascend/driver/tools/hccn_tool")
+if not os.path.exists("/usr/local/Ascend/driver/tools/hccn_tool"):
```
wangxiyuan: This logic is not correct. If user set `hccn_tool_path` to a new place and there is no `/usr/local/Ascend/driver/tools/hccn_tool`, you will force rewrite to `"/usr/bin/hccn_tool"
`
```suggestion
DEFAULT_PATHS = ["/usr/local/Ascend/driver/tools/hccn_tool", "/usr/bin/hccn_tool"]
hccn_tool_path = os.environ.get("HCCN_TOOL_PATH")
if not hccn_tool_path:
    for default_path in  DEFAULT_PATHS:
        if os.path.exists(default_path):
            hccn_tool_path = default_path
            break
    else:
        raise Exception("can't get correct hccn_tool_path")
        
```

yangqinghao-cmss: Sure, you're right. I've fixed it now.

PR #2080, opened 2025-07-29
[Bugfix][PD] Make multiple Ps and Ds work on a single machine

CaveNightingale:
(cherry picked from commit 816375e0c1071d0696dfab1a1ce35674f9f37aa0)

Suppose that you want to start a prefiller instance with npus `2,3` only. So you start the instance with `ASCEND_RT_VISIBLE_DEVICES=2,3`. The current programming will start two workers, whose ranks are `0` and `1` respectedly. And they will pick the first and second ip addresses of npus in the ranktable instead of the thirdth and forth ones. But actually they are using card `2,3` and therefore they can not link with remote instances when they attempt to transfer the KVCache.

Hence, at most 1 prefiller instance and at most 1 decoder instance can work  on a single machine since they always pick the first npu ip address in the ranktable currently.

This pull request is proposed to fix the problem. This fix pick ips of only those devices that are in `ASCEND_RT_VISIBLE_DEVICES` from the ranktable.

If the user use ranktable generated by `gen_ranktable.sh`, they should not face any change.

Qwen-0.6B 1P 1D, dp=2, `ASCEND_RT_VISIBLE_DEVICES=2,3` for prefiller and `ASCEND_RT_VISIBLE_DEVICES=4,5` for decoder.

- vLLM version: v0.10.0

### DIFF
--- a/tests/ut/kv_connector/test_llmdatadist_connector.py
+++ b/tests/ut/kv_connector/test_llmdatadist_connector.py
@@ -2,10 +2,13 @@
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 # Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
 
+import os
+import types
+
 from tests.ut.kv_connector.utils import (create_request, create_scheduler,
                                          create_vllm_config)
-from vllm_ascend.distributed.llmdatadist_c_mgr_connector import \
-    LLMDataDistCMgrConnectorMetadata
+from vllm_ascend.distributed.llmdatadist_c_mgr_connector import (
+    LLMDataDistCMgrConnectorMetadata, LLMDataDistCMgrConnectorWorker, LLMRole)
 
 
 def test_basic_inferface():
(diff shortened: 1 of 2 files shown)


tests/ut/kv_connector/test_llmdatadist_connector.py
```diff
@@ -40,3 +43,54 @@ def test_basic_inferface():
...
+        worker.tp_rank = worker_tp_rank
+        worker.llm_datadist_role = LLMRole.PROMPT
+        os.environ["ASCEND_RT_VISIBLE_DEVICES"] = worker_visible_devices
+        agent_metadata = LLMDataDistCMgrConnectorWorker.read_agent_metadata(
+            worker, rank_table)
+        os.environ["ASCEND_RT_VISIBLE_DEVICES"] = old_visible_devices
+        return agent_metadata.device_ip
+
+    assert get_device_ip("192.168.1.1", 0, "0") == "10.30.0.1"
+    assert get_device_ip("192.168.1.1", 0, "1") == "10.30.0.2"
```
Ronald1995, line 90: you could use pytest.mark.parameterize to simplify the assert case.

wangxiyuan (approved):

PR #2084, opened 2025-07-29
[Test] add rejection sampler ut

momo609:
add rejection sampler ut.

No

UT passed

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/586f286789a09f5616be74ee8bedde0a9f698a72

### DIFF
--- a/tests/ut/sample/test_rejection_sampler.py
+++ b/tests/ut/sample/test_rejection_sampler.py
@@ -1,0 +1,201 @@
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


tests/ut/sampler/test_rejection_sampler.py
```diff
@@ -0,0 +1,200 @@
...
+    expand_batch_to_tokens, expand_pytorch, rejection_greedy_sample_pytorch,
+    rejection_random_sample_pytorch, sample_recovered_tokens_pytorch)
+
+# Global constants
+PLACEHOLDER_TOKEN_ID = -1
+GREEDY_TEMPERATURE = 0.0
+MAX_SPEC_LEN = 8  # Used as MAX_NUM_TOKENS in expand_batch_to_tokens
+
+
+class TestRejectionSampler:
```
Yikun: https://github.com/vllm-project/vllm-ascend/blob/27c2b5c14599881895a086c51bfa7bcdc680d5a8/tests/ut/sample/test_sampler.py#L9

refer this to use base to make sure patch is [applied](https://github.com/vllm-project/vllm-ascend/blob/main/tests/ut/base.py#L27-L29) and it protect your case will not be impacted with other patches.

Yikun (commented):
1. The path should be `tests/ut/sample/test_rejection_sampler.py`, same with code, other wise LGTM
2. Also pleae fullfill  the PR description

PR #2100, opened 2025-07-30
Support new model

yzy1996:

Support new model

Yes, new model supported

Test locally

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/02f82fe4386b3e84eb0f06bfaf7744c5b4fdba4f


Yikun:
As discussed offline, the modeling py should be in vLLM repo or model repo (trustremote), we should also avoid code redundancy.

So, close now, feel free to open new PR if needed.

PR #1857, opened 2025-07-17
[Feature] Optimize forward metadata collection across dp ranks

jianzs:
Resubmit #1593

### What this PR does / why we need it?

This PR introduces two optimizations for cases where data parallel size > 1:
~~1. Eliminates DP communication in `set_forward_context`~~
2. Implements HCCL for DP metadata communication, resulting in significant performance improvements for large DP configurations
   - Achieves ~20ms latency reduction with DP size of 64

### Does this PR introduce _any_ user-facing change?
no

### How was this patch tested?
CI passed.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/4e4d017b6f70c729e7c78f74e4328a4ebca7b8ec

PR #2102, opened 2025-07-30
fix gllo allreduce to hccl

zhaowei1936:
Implements HCCL for DP has_unfinished_dp communication, resulting in significant performance improvements for large DP
Achieves ~10ms latency reduction with A3 DP size of 128
image

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/6807af8f46acd184f99342ff38f2a1359f693b10

### DIFF
--- a/vllm_ascend/patch/platform/patch_common/patch_distributed.py
+++ b/vllm_ascend/patch/platform/patch_common/patch_distributed.py
@@ -19,8 +19,12 @@
 
 import torch
 import vllm.envs as envs
+from torch.distributed import ProcessGroup, ReduceOp
 from vllm.config import ParallelConfig
+from vllm.distributed.utils import \
+    stateless_init_torch_distributed_process_group
 
+from vllm_ascend import envs as ascend_envs
 from vllm_ascend.utils import is_310p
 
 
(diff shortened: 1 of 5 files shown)


jianzs:
https://github.com/vllm-project/vllm-ascend/pull/1857 I submitted a pull request with this feature, and eliminated the communication overhead for DP metadata when constructing the forward context.  Now, I'm waiting for @wangxiyuan to refactor the model runner.

zhaowei1936:
> #1857 I submitted a pull request with this feature, and eliminated the communication overhead for DP metadata when constructing the forward context. Now, I'm waiting for @wangxiyuan to refactor the model runner.

ok， I abandon modifying this function **_get_forward_metadata_across_dp**， what time   @wangxiyuan to refactor the model runner？ I modifying this funciton **has_unfinished_dp**  first

jianzs:
As far as I know, `has_unfinished_dp` runs inside the **Scheduler** process, not the **Worker** process. In that case, why does optimizing this function improve performance?

vllm_ascend/patch/platform/patch_common/patch_distributed.py
```diff
@@ -41,7 +44,32 @@ def parallel_config_get_dp_port(self) -> int:
     return port
 
 
+def ascend_stateless_init_dp_group(self) -> "ProcessGroup":
```
momo609, line 48: This will result in increased memory usage. Has an impact analysis been performed on this aspect?

zhaowei1936: increased memory every device is only  sizeof(torch.int32),  is small

zhaowei1936:
> As far as I know, `has_unfinished_dp` runs inside the **Scheduler** process, not the **Worker** process. In that case, why does optimizing this function improve performance?
image
The specific benefits of a single case are derived as follows:
1) Gloo was initially designed for small-scale clusters (such as single machines with multiple cards), using relatively simple communication protocols. In multi-machine cross-node scenarios, the protocol overhead (such as handshakes, message headers) significantly increases, leading to a decrease in actual data transmission efficiency.
2) Gloo does not support acceleration from dedicated network hardware (such as RDMA, NCCL) and relies on the standard TCP/IP protocol. In multi-machine environments, TCP's congestion control and retransmission mechanisms introduce additional delays, especially in high-bandwidth networks where the hardware potential cannot be fully utilized.
3) The core communication logic of Gloo is single-threaded, unable to fully utilize the multi-core CPU resources in a multi-machine environment. As the number of nodes increases, single-threaded processing becomes a significant bottleneck.
The advantages of HCCL communication based on A3 super nodes are mainly as follows:
1) HCCL is a communication library specifically designed by Huawei for the Ascend series of NPUs, directly invoking the communication engine built into the chip, avoiding the overhead of the general TCP/IP protocol stack that Gloo relies on.
2) HCCL supports the RoCE (RDMA over Converged Ethernet) protocol, utilizing Ethernet to achieve RDMA-level high-performance communication, with latency reduced to the microsecond level, whereas Gloo's TCP-based latency is typically in the millisecond range.
3) HCCL can automatically recognize the cluster network topology (such as switch hierarchy, link bandwidth) and dynamically select the optimal communication path. For example, in multi-machine multi-node scenarios, HCCL adjusts the partitioning of the Ring algorithm for all_reduce based on the network distance between nodes, whereas Gloo uses a fixed strategy and cannot adapt to complex topologies.
4) For large-scale clusters (such as those with more than 64 nodes), HCCL employs hierarchical aggregation (Hierarchical AllReduce): first, nodes within a cabinet quickly aggregate, then cross-cabinet aggregation is performed, significantly reducing the number of cross-switch communications. Gloo lacks this intelligent hierarchical mechanism. 

报错

wangxiyuan:
please take a look at this one. https://github.com/vllm-project/vllm/pull/22243

PR #426, opened 2025-03-28
support aclgraph

fakeYan:
### What this PR does / why we need it?

This PR supports the access of vllm-acend to the piecewise_graph feature provided by the v1 engine.

1. register unifiled_ascend_attention_with_output for piecewise_graph to split graph.
2. support NPUGraph to accelerate kernel launch.

### Does this PR introduce _any_ user-facing change?

support npugraph to default， Users can disenable the npugraph feature by configuring enforce_eager.

This has corresponding requirements for the versions of torch_npu and CANN, and they need to support graph capture.

### How was this patch tested?

it turn to default

PR #2113, opened 2025-07-30
[core] Support capture custom ops into aclgraph

ganyi1996ppo:
Thanks to the PR https://github.com/vllm-project/vllm-ascend/pull/426 make vllm-ascend support the aclgraph inference to reduce the host overhead. However, the capability of aclgraph strongly relies on the functionality provided by `torch.compile`, which is the key feature supported in torch 2.x . Therefore, capture custom op into aclgraph is only possible when it can be recognize and captured by `torch.compile`. 

In this PR, we register the meta implementation of current custom ops to enable the fx graph capture. And by doing that, insert those custom ops into aclgraph become a natural thing to the ascend runtime.

No user face change.

Tested in unittest, we will integrate the `rotary_embedding` op into a small custom model and use `torch.compile` and aclgraph to capture and replay it to verify its functionality.

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/1b9902806915040ac9b3029f2ab7522ec505afc3

### DIFF
--- a/tests/e2e/singlecard/ops/test_rotary_embedding.py
+++ b/tests/e2e/singlecard/ops/test_rotary_embedding.py
@@ -22,6 +22,7 @@ ROTARY_DIMS = [None, 32]  # None means rotary dim == head size
 NUM_HEADS = [17]  # Arbitrary values for testing
 BATCH_SIZES = [5]  # Arbitrary values for testing
 SEQ_LENS = [11, 4096]  # Arbitrary values for testing
+NUM_TOKENS = [10, 21]
 SEEDS = [0]
 DEVICES = [f"npu:{0}"]
 # Set tolerance to 1 for quant ops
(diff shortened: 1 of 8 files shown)


ganyi1996ppo:
@ttanzhiqiang Please review this PR also

ganyi1996ppo:
@yiz-liu Please review this PR also

tests/e2e/singlecard/ops/test_rotary_embedding.py
```diff
@@ -198,3 +199,118 @@ def test_rotary_embedding_quant_with_leading_dim(
...
+            static_output = compiled_model(static_positions,
+                                           static_hidden_states,
+                                           offsets=None)
+    stream.wait_stream(torch.npu.current_stream())
+
+    aclgraph = torch.npu.NPUGraph()
+
+    with torch.npu.graph(aclgraph):
+        # Capture the model in aclgraph.
+        static_output = compiled_model(static_positions, static_hidden_states)
```
ttanzhiqiang, line 322: This place is a static shape. If the shape of static_positions, static_hidden_states has changed, does meta need to go through again?

ttanzhiqiang:
Can custom ops be specially written to handle the problem of aclgraph without adding an operator to a meta at a time

tests/e2e/singlecard/ops/test_rotary_embedding.py
ganyi1996ppo: yes

ganyi1996ppo:
> Can custom ops be specially written to handle the problem of aclgraph without adding an operator to a meta at a time

No for now. We can try to write a macro or template to automatically generate the meta implementation, but no plan for this yet. We can consider this after we have enough number of custom ops.

wangxiyuan:
LGTM, Please make the CI happy, I think we should merge this in high priority

ganyi1996ppo:
> LGTM, Please make the CI happy, I think we should merge this in high priority

The aclgraph path seems have unaligned accuracy compared with eager path on CI, I can't reproduce it on my local environment.

MengqingCao:
It seems the first failed case is acc test case due to the model download issue. I think you can rebase your code, as acc test will not run per pr now.
image

MengqingCao:
> It seems the first failed case is acc test case due to the model download issue. I think you can rebase your code, as acc test will not run per pr now. image

ganyi1996ppo:
> It seems the first failed case is acc test case due to the model download issue. I think you can rebase your code, as acc test will not run per pr now.

Got, I'll give it a try

wangxiyuan (approved):

PR #1856, opened 2025-07-17
[MoE][Dist] Fix Qwen MoE accuracy bug in DP scenario

MengqingCao:
### What this PR does / why we need it?
Fix Qwen MoE accuracy bug in DP scenario.

Now the implentment of `FusedMoE` in vLLM use `All2AllManager` to manager different all2all algorithm branch. And the default branch use `Multicast` in `dispatch` phase and `all_reduce` in `combine` phase, which are not implented in vLLM-Ascend. This leading to invoking into a default implentment in `base_communicator`, with empty `dispatch` and `combine` operations, thus causing the accuracy issue on it.

This pr is a temporary workaround, refacting all2all in vLLM-Ascend could be a better way.

### How was this patch tested?

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/ad57f23f6a528ab01066998b41796a44340fd43d

PR #2118, opened 2025-07-30
[BugFix] Fix accuracy problem in dp situation.

whx-sjtu:
This PR is a quick fix of the accuracy problem mentioned in PR #1856.
- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/004203e95330ac9a878df8192619570b0770667e

### DIFF
--- a/vllm_ascend/distributed/communicator.py
+++ b/vllm_ascend/distributed/communicator.py
@@ -20,6 +20,8 @@ import torch
 import torch.distributed as dist
 from vllm.distributed.device_communicators.base_device_communicator import \
     DeviceCommunicatorBase
+from vllm.forward_context import get_forward_context
+from vllm.distributed.parallel_state import get_dp_group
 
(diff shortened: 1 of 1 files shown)


MengqingCao:
It seems the implement of `dispatch` and `combine` is same as that in `NaiveAll2AllManager`, I think we'd better to call these communication ops through `NaiveAll2AllManager`, like #1856 does, instead of coping them here.

wangxiyuan:
yes, if nothing changed with `NaiveAll2AllManager`, we should import from vllm directly instead of copy the code.

whx-sjtu:
Got it, I will change this soon.

MengqingCao:
> Got it, I will change this soon.

To avoid duplicated work, I think #1856 is enough, and I'm working on adding accuracy test on dp + moe.

PR #2120, opened 2025-07-31
add mlp tp optimze

momo609:
For dense models, by not applying tensor parallelism (TP) to the attention module and applying TP to the MLP module, the allreduce operations in the attention module can be eliminated, thereby reducing computational overhead. However, this approach increases memory usage, so the environment variable VLLM_ASCEND_ENABLE_MLP_OPTIMZE is used to control this optimization.
image

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/b17109beeafbf9577c319ab61530810943a7fc4b

### DIFF
--- a/tests/ut/ops/test_linear.py
+++ b/tests/ut/ops/test_linear.py
@@ -1,0 +1,363 @@
+import os
+import unittest
+from unittest import mock
+
+import torch
+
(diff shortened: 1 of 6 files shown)


weijinqian0:
approved

wangxiyuan (approved):

PR #2122, opened 2025-07-31
[Bugfix] Adopt the new changes on disaggregated pd from vllm main branch

ganyi1996ppo:
We notice that vllm's main branch merged the PR https://github.com/vllm-project/vllm/pull/21072 and https://github.com/vllm-project/vllm/pull/21473 to support ray backend and fix some rebase bug from previous change. Those changes makes the disaggregate pd in vllm ascend breaks in some scenario. 

In this PR, we adopt those changes to make sure the `llmdatddist_c_mgr_connector` works fine on the newest vllm main branch.

No user face change.

relevant ut will be added to make sure the functionality of those changes.

- vLLM version: v0.10.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -1445,6 +1445,8 @@ class NPUModelRunner(LoRAModelRunnerMixin):
         hidden_states: torch.Tensor,
         num_scheduled_tokens: int,
         num_scheduled_tokens_np: np.ndarray,
+        finished_sending: Optional[set[str]],
+        finished_receiving: Optional[set[str]],
     ) -> ModelRunnerOutput:
         assert self.input_batch.num_reqs ==\
             len(self.input_batch.pooling_params), \
@@ -1479,6 +1481,8 @@ class NPUModelRunner(LoRAModelRunnerMixin):
             logprobs=None,
             prompt_logprobs_dict={},
             pooler_output=pooler_output,
+            finished_sending=finished_sending,
+            finished_recving=finished_receiving
         )
 
     @torch.inference_mode()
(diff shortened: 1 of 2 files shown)


Potabk:
looks good if CI passed

vllm_ascend/worker/worker_v1.py
```diff
@@ -204,9 +206,18 @@ def execute_model(
...
+
+            new_output = EMPTY_MODEL_RUNNER_OUTPUT
+            if output.finished_sending or output.finished_recving:
+                new_output = copy.copy(new_output)
+                new_output.finished_sending = output.finished_sending
+                new_output.finished_recving = output.finished_recving
+            output = new_output
+
         assert isinstance(output, ModelRunnerOutput)
-        return output if self.is_driver_worker else None
```
wangxiyuan, line 209: can this be removed? I suppose this is usefull in spmd case

ganyi1996ppo: vllm's main branch removed this in https://github.com/vllm-project/vllm/pull/21473/files , now every tp return its own result, so the tp scenario will encounter the None input case with our previous code

wangxiyuan: emm, but it may break PP function I guess. @MengqingCao can you take a check as well？

wangxiyuan (approved):

vllm_ascend/worker/worker_v1.py
MengqingCao: I think this is is ok with pp, LGTM

wangxiyuan: Double checked with contributor for PP and RL case.  It's OK to do the change

wangxiyuan (approved):

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1445,6 +1445,8 @@ def _pool(
         hidden_states: torch.Tensor,
         num_scheduled_tokens: int,
         num_scheduled_tokens_np: np.ndarray,
+        finished_sending: Optional[set[str]],
```
wangxiyuan, line 1448: vLLM now use `kv_connector_metadata` to store finished_sending and finished_receiving, @Potabk will fix it in a new PR

PR #2124, opened 2025-07-31
Implementation of simple load balance routing proxy server (#1953)

ganyi1996ppo:
The PR is the cherry-pick from v0.9.1 https://github.com/vllm-project/vllm-ascend/pull/1953

This PR introduce a new load balance proxy server example implementation for disaggregated pd, which support simple token&kv_cache aware load balance routing strategy for the disaggregated pd system compared with origin round robin toy_proxy.

No
tested on real workload and unittest

- vLLM version: v0.10.0


wangxiyuan:
does this example rely on DP patch?

ganyi1996ppo:
> does this example rely on DP patch?

No, this is an totally independant proxy .

wangxiyuan:
OK, make sense for me.

wangxiyuan (approved):

examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
```diff
@@ -0,0 +1,435 @@
+# Adapted from https://github.com/vllm-project/vllm/tests/v1/kv_connector/nixl_integration/toy_proxy_server.py
+
+# SPDX-License-Identifier: Apache-2.0
+
```
wangxiyuan, line 87: it's good to add some usage guide here to let users know how to run this exmaple in quick

jianzs:
Do we need both proxy server implementations in the example folder? We could keep either toy_proxy_server or the one from this PR?

examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
ganyi1996ppo: Good advise, I'll add some example in the comments

ganyi1996ppo:
> Do we need both proxy server implementations in the example folder? We could keep either toy_proxy_server or the one from this PR?

Sounds fair, I'll remove the old one

wangxiyuan (approved):

PR #2126, opened 2025-07-31
[v0.9.1][bugfix] fix dtype of topk_ids

loukong33:

This PR fixes the dtype of topk_ids when dtype of router_logits is float32.

No.

e2e


wangxiyuan:
looks like we don't need this change? it's int32 by default?

loukong33:
> looks like we don't need this change? it's int32 by default?

Yes. I have confirmed it. And I will close this pr.

PR #2129, opened 2025-07-31
[bugfix] add with_prefill cpu allreduce to handle D-node recomputatio…

liziyu179:
…n situations

Add with-prefill CPU AllReduce to handle D-node recomputation situations.


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -633,12 +633,19 @@ def _get_forward_metadata_across_dp(
...
-            max_num_decode_tokens = self.torchair_graph_batch_sizes[0]
-            num_tokens_across_dp = torch.tensor([max_num_decode_tokens] *
-                                                self.dp_size,
-                                                device="cpu",
-                                                dtype=torch.int32)
-            return max_num_decode_tokens, num_tokens_across_dp, False, enable_dbo
+            with_prefill_tensor = torch.tensor([with_prefill],
+                                               device="cpu",
+                                               dtype=torch.bool)
+            dist.all_reduce(with_prefill_tensor,
```
momo609, line 639: Has the performance impact of this been tested? Can it be recovered using the allreduce communication in num_tokens_across_dp?

PR #2002, opened 2025-07-25
[v0.9.1][RejectSampler][Perf] Optimize greedy reject sampler with vectorization.

whx-sjtu:
This PR rewrites `rejection_greedy_sample_pytorch` function in vectorized form instead of for loop to improve performance, especially in large batch size scenario. Further more I write a clean function `rejection_greedy_sample_spec_len_1_pytorch` as simple as possible which is specially used in all-greedy and `spec_len==1` situation.

PR #2137, opened 2025-07-31
[Perf][MTP] Optimize reject sampler in greedy situation.

whx-sjtu:
This PR port optimization in PR #2002 to main and makes it cleaner.

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/afa5b7ca0b417abadfa85e32f28969b72e58a885

### DIFF
--- a/vllm_ascend/sample/rejection_sampler.py
+++ b/vllm_ascend/sample/rejection_sampler.py
@@ -147,16 +147,26 @@ def rejection_sample(
     if not sampling_metadata.all_random:
         # Rejection sampling for greedy sampling requests.
         target_argmax = target_probs.argmax(dim=-1)
-        rejection_greedy_sample_pytorch(
-            output_token_ids,
-            cu_num_draft_tokens,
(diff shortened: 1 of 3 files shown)


wangxiyuan (approved):

vllm_ascend/sample/rejection_sampler.py
```diff
@@ -147,16 +147,25 @@ def rejection_sample(
...
-            cu_num_draft_tokens,
-            draft_token_ids,
-            target_argmax,
-            bonus_token_ids,
-            is_greedy,
-            max_spec_len,
-            # num_warps=1,
-        )
+        if min(num_draft_tokens) == 1 and max(
+                num_draft_tokens) == 1 and sampling_metadata.all_greedy:
```
Ronald1995, line 151: min and max will do the sort action by twice,  it think `if all(num == 1 for num in num_draft_tokens)` will make the logic，but it will get a better preformance.

PR #2150, opened 2025-07-31
refactor fused_moe.py

shiyuan680:

### DIFF
--- a/vllm_ascend/ops/moe_layer/select_experts.py
+++ b/vllm_ascend/ops/moe_layer/select_experts.py
@@ -1,0 +1,79 @@
+from abc import ABC, abstractmethod
+import torch_npu
+from vllm_ascend.ops.fused_moe import select_experts
+import vllm_ascend.envs as envs_ascend
+SELECT_GATING_TOPK_SOTFMAX_EXPERTS: bool = envs_ascend.SELECT_GATING_TOPK_SOTFMAX_EXPERTS
+
+
+class BaseSelectExperts(ABC):
+
+    def __init__(self):
+        need_param = SelectExpertConfig.get_config
+        self.top_k = need_param["top_k"]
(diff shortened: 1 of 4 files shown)


ApsarasX:
I think this PR helps improve code modularity, but it would be better to use a clearer title and pass CI.

vllm_ascend/ops/moe_layer/select_experts.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+class QuantizedExpertsSelector(BaseExpertsSelector):
+    def __init__(self):
+        super().__init__()
+
+    def apply(self, router_logits: torch.Tensor, x: torch.Tensor):
+
+        return super().apply(router_logits, x)
+
```
weijinqian0: Do not estimate requirements in advance; it may be necessary to remove both quantitative and non-quantitative implementations.

shiyuan680: i will merge implementations  of quantitative and non-quantitative, just has one class

vllm_ascend/ops/moe_layer/select_experts.py
```diff
@@ -0,0 +1,226 @@
...
+                    topk_weights, renormalize)
+                return topk_weights, topk_ids
+
+            topk_weights = router_logits.softmax(dim=-1)
+        elif scoring_func == "sigmoid":
+            topk_weights = router_logits.sigmoid()
+        else:
+            raise ValueError(f"Unsupported scoring function: {scoring_func}")
+
+        if use_grouped_topk:
```
weijinqian0: It is necessary to determine whether to extract an independent function.

shiyuan680: done

wangxiyuan (approved):

PR #2156, opened 2025-08-01
[Fix] Adjust use_aclgraph logic

yiz-liu:
Updates the FusedMoE method to determine whether to use ACL Graph based on the `torchair_graph_config`

This is equivalent to #2154 on v0.9.1-dev.

None.

None needed.

- vLLM version: v0.10.0

### DIFF
--- a/vllm_ascend/ops/common_fused_moe.py
+++ b/vllm_ascend/ops/common_fused_moe.py
@@ -22,6 +22,7 @@ from vllm.config import CompilationLevel, get_current_vllm_config
 from vllm.model_executor.layers.fused_moe.layer import \
     UnquantizedFusedMoEMethod
 
+from vllm_ascend.ascend_config import get_ascend_config
 from vllm_ascend.ops.fused_moe import (fused_experts, fused_experts_moge,
                                        select_experts)
 from vllm_ascend.utils import is_310p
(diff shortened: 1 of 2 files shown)


vllm_ascend/ops/common_fused_moe.py
```diff
@@ -33,7 +34,15 @@ def unquantized_fused_moe_init_func(self, *args, **kwargs):
     original_unquantized_fused_moe_init_func(self, *args, **kwargs)
     vllm_config = get_current_vllm_config()
     self.max_num_batched_tokens = vllm_config.scheduler_config.max_num_batched_tokens
-    self.use_aclgraph = vllm_config.compilation_config.level == CompilationLevel.PIECEWISE and not vllm_config.model_config.enforce_eager
+
+    ascend_config = get_ascend_config()
```
ApsarasX, line 38: I suggest referring to #1841 and adding a new class called `AscendUnquantizedFusedMoEMethod`, incorporating the ACLGraph switch into this class's `__init__` method, while also placing the new `forward_oot` method within this class.

ApsarasX (changes requested):

vllm_ascend/ops/common_fused_moe.py
wangxiyuan: @ApsarasX it's on the todo list. we'll first combine common_fused_moe and fused_moe into one. Then refactor to Custom OP register way.

ApsarasX: > @ApsarasX it's on the todo list. we'll first combine common_fused_moe and fused_moe into one. Then refactor to Custom OP register way.

OK

wangxiyuan (approved):

wangxiyuan:
let's merge this first to fix the known bug. For the next step, we should refactor the fused moe totally.

PR #2157, opened 2025-08-01
add super kernel for decode moe

NNUCJ:
&emsp;&emsp;Using  super kernel feature to fuse some operators in the Moe stage, reducing scheduling overhead on devices. enable_super_kernel is valid only when Torchair graph mode and enable_multistream_moe is enabled.

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/59f3b936365afd200e474ddc9d1f5aa33f05b634

### DIFF
--- a/vllm_ascend/models/deepseek_v2.py
+++ b/vllm_ascend/models/deepseek_v2.py
@@ -323,15 +323,23 @@ class CustomDeepseekV2MoE(nn.Module):
         self.enable_multistream_moe = \
             ascend_config.torchair_graph_config.enable_multistream_moe and \
             self.torchair_graph_enabled
+        self.enable_super_kernel = self.enable_multistream_moe and self.tp_size == 1
+        self.params_dtype = torch.float32 if self.enable_super_kernel else \
+            torch.get_default_dtype()
 
+        # Converting gate weight to fp32 is to adapt to the super kernel feature.
+        # Super kernel feature currently cannot fuse operators such as cast, stridedslice, and add.
+        # In the moe stage, Cast will interrupt the fusion of the super kernel. To avoid this problem,
+        # modifications will be made in the initialization stage.
         self.gate = ReplicatedLinear(config.hidden_size,
                                      config.n_routed_experts,
                                      bias=False,
                                      quant_config=None,
+                                     params_dtype=self.params_dtype,
                                      prefix=f"{prefix}.gate")
         if config.topk_method == "noaux_tc":
             self.gate.e_score_correction_bias = nn.Parameter(
-                torch.empty(config.n_routed_experts))
+                torch.empty(config.n_routed_experts, dtype=self.params_dtype))
         else:
             self.gate.e_score_correction_bias = None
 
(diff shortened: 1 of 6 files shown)


vllm_ascend/models/deepseek_v2.py
```diff
@@ -315,15 +315,22 @@ def __init__(
...
-                                     bias=False,
-                                     quant_config=None,
-                                     prefix=f"{prefix}.gate")
+        self.gate = ReplicatedLinear(
+            config.hidden_size,
+            config.n_routed_experts,
+            bias=False,
+            quant_config=None,
+            params_dtype=torch.float32
+            if self.enable_super_kernel else self.params_dtype,
```
realliujiaxu: Why is the gate weight cast to fp32? Please add a comment here.

realliujiaxu:
Any benchmark result or profiling timeline？

vllm_ascend/models/deepseek_v2.py
NNUCJ: &emsp;&emsp;The super kernel feature currently cannot fuse operators such as cast, stridedslice, and add. In the moe stage, some type conversions are performed by default in the torchair, as shown in the following figure. This operation will interrupt the fusion of the super kernel. To avoid this problem, modifications will be made in the initialization stage.
image

NNUCJ:
> Any benchmark result or profiling timeline？
1. baseline 
baseline
2. add super_kernel 
spuer_kernel

vllm_ascend/models/deepseek_v2.py
realliujiaxu: please leave a comment here

vllm_ascend/models/deepseek_v2.py
```diff
@@ -315,15 +315,26 @@ def __init__(
         self.enable_multistream_moe = \
             ascend_config.torchair_graph_config.enable_multistream_moe and \
             self.torchair_graph_enabled
+        self.enable_super_kernel = self.enable_multistream_moe and self.tp_size == 1
```
ApsarasX, line 326: Why can only TP1 use the super_kernel?

vllm_ascend/models/deepseek_v2.py
```diff
@@ -315,15 +315,26 @@ def __init__(
...
+        self.params_dtype = torch.get_default_dtype()
```
ApsarasX: use the following suggestion to avoid duplicated code?
```suggestion
        self.params_dtype = torch.float32 if self.enable_super_kernel else torch.get_default_dtype()
```

vllm_ascend/torchair/utils.py
```diff
@@ -96,3 +97,7 @@ def npu_wait_tensor(self: torch.Tensor,
                     *,
                     enabled: bool = True):
     return _npu_wait_tensor(self, dependency) if enabled else self
+
+
+def super_kernel(prefix: str, stream: str, enabled: bool = True):
```
ApsarasX:
```suggestion
def super_kernel(prefix: str, options: str, enabled: bool = True):
```

vllm_ascend/models/deepseek_v2.py
NNUCJ: Thank you for your suggestion.

vllm_ascend/models/deepseek_v2.py
NNUCJ: When the tp size is greater than 1, some stripdslice operators will be introduced in fusid_moe, which will interrupt the fusion of the super kernel

ApsarasX (approved):

wangxiyuan:

PR #2073, opened 2025-07-28
Enable pytest and yaml style accuracy test

wxsIcey:
### What this PR does / why we need it?

This PR enabled pytest and yaml style accuracy test, users now can enable accuracy test by running:

```bash
cd ~/vllm-ascend
pytest -sv ./tests/e2e/singlecard/models/test_lm_eval_correctness.py \
          --config ./tests/e2e/singlecard/models/configs/Qwen3-8B-Base.yaml \
          --report_output ./benchmarks/accuracy/Qwen3-8B-Base.md

pytest -sv ./tests/e2e/singlecard/models/test_lm_eval_correctness.py \
          --config-list-file ./tests/e2e/singlecard/models/configs/accuracy.txt
```

Closes: https://github.com/vllm-project/vllm-ascend/issues/1970

### Does this PR introduce _any_ user-facing change?
no

### How was this patch tested?

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/2836dd73f13015ee386c544760ca0d16888203f3

PR #2163, opened 2025-08-01
Fix accuracy test config --config-list-file

wxsIcey:
This PR fix accuracy test related to https://github.com/vllm-project/vllm-ascend/pull/2073, users can now perform accuracy tests on multiple models simultaneously and generate different report files by running:

```bash
cd ~/vllm-ascend
pytest -sv ./tests/e2e/singlecard/models/test_lm_eval_correctness.py \
          --config-list-file ./tests/e2e/singlecard/models/configs/accuracy.txt
```

no

image

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/61dcc280faf305778c0c44597e823f40063aaed6

### DIFF
--- a/tests/e2e/singlecard/models/conftest.py
+++ b/tests/e2e/singlecard/models/conftest.py
@@ -62,7 +62,7 @@ def report_output(pytestconfig, config_filename):
         report_path = Path(report_dir) / f"{model_name}.md"
         report_path.parent.mkdir(parents=True, exist_ok=True)
         return report_path
-    
+
     output_path = pytestconfig.getoption("--report_output")
     if output_path:
         output_path = Path(output_path)
@@ -84,6 +84,5 @@ def pytest_generate_tests(metafunc):
             return
         single_config = metafunc.config.getoption("--config")
         config_path = Path(single_config).resolve()
-        
+
         metafunc.parametrize("config_filename", [config_path])
-        


tests/e2e/singlecard/models/conftest.py
```diff
@@ -48,26 +54,35 @@ def config(pytestconfig):
...
 
 
 def pytest_generate_tests(metafunc):
     if "config_filename" in metafunc.fixturenames:
-        # If config specified, use the --config directly
-        single_config = metafunc.config.getoption("--config")
-        if single_config:
-            metafunc.parametrize("config_filename",
-                                 [Path(single_config).resolve()])
+        if metafunc.config.getoption("--config-list-file"):
```
Yikun, line 74: Why reorder this?

the original thoughts is that `--config` has higer priority than `--config-list-file`

tests/e2e/singlecard/models/conftest.py
```diff
@@ -30,6 +30,12 @@ def pytest_addoption(parser):
         default="./benchmarks/accuracy/Qwen3-8B-Base.md",
         help="Path to the report output file",
     )
+    parser.addoption(
+        "--report-dir",
+        action="store",
+        default="./benchmarks/accuracy",
```
Yikun, line 36: it's better to keep current path, `./`, otherwise the `./benchmarks/accuracy` is required.

tests/e2e/singlecard/models/conftest.py
wxsIcey: Because the --config option has default settings, if --config is checked first, the logic for --config-list-file will definitely not be executed.

tests/e2e/singlecard/models/conftest.py
wxsIcey: The code has ensured the creation of the ./benchmarks/accuracy directory: https://github.com/vllm-project/vllm-ascend/blob/main/tests/e2e/singlecard/models/test_lm_eval_correctness.py#L88

PR #2165, opened 2025-08-01
[Doc][PD] Restore the default configuration items in examples/disaggregate_prefill_v1/README.md 

underfituu:
- In the D node, the max-num-batched-tokens parameter can be set to a smaller value since the D node processes at most max-num-seqs batches concurrently. As the profile_run only needs to handle max-num-seqs sequences at a time, we can safely set max-num-batched-tokens equal to max-num-seqs. This optimization will help reduce activation memory consumption.
- Restore the default configuration items for PD separation.

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/61dcc280faf305778c0c44597e823f40063aaed6

### DIFF
--- a/examples/disaggregated_prefill_v1/README.md
+++ b/examples/disaggregated_prefill_v1/README.md
@@ -6,7 +6,7 @@ This demo document provides instructions for running a disaggregated vLLM-ascend
 ## Prerequisites
 - Ascend NPU environment with vLLM 0.9.1 installed
 - Network interfaces configured for distributed communication (eg: eth0)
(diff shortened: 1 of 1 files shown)


ganyi1996ppo:
Perhaps we can also update the 800I A3's best practice in the future ?

underfituu:
> Perhaps we can also update the 800I A3's best practice in the future ?
- Thanks for highlighting this! We’ll definitely include the 800I A3 best practices in a future update. As the main branch evolves, we’ll keep the README synchronized with optimized configurations and performance benchmarks. Stay tuned for improvements!

PR #2167, opened 2025-08-01
[feat]: oproj tensor parallelism in  pure DP and graph-mode scenarios.

lidenghui1110:
This PR introduces Oproj matrix tensor model parallel to achieve decreasing of memory consumption. It only support graph mode in pure DP scenario.

In deepseek r1 w8a8 PD disagregated Decode instance, using pure DP, with oproj_tensor_parallel_size = 8, we have 1 ms TPOT increasing, saved 5.8 GB NPU memory per RANK. We got best performance when  oproj_tensor_parallel_size=4 without TPOT increasing.

performance data:
image

This PR introduces one new config in `additional_config`.
| Name                          | Effect                                   | Required | Type | Constraints        |
| :---------------------------- | :--------------------------------------- | :------- | :--- | :----------------- |
| oproj_tensor_parallel_size | Split the o_proj matrix along the row dimension (head num * head dim) into oproj_tensor_parallel_size pieces.  | No       | int  | default value is None, once this value is set, the feature will be enabled, head num * head dim must be divisible by this value. |

example

`--additional_config={"oproj_tensor_parallel_size": 8}`

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/eddaafc1c77b0690194cbd1b73747d572793838c


vllm_ascend/ops/linear.py
```diff
@@ -0,0 +1,242 @@
...
+            self.bias = Parameter(
+                torch.empty(self.output_size, dtype=params_dtype))
+            set_weight_attrs(self.bias, {
+                "output_dim": 0,
+                "weight_loader": self.weight_loader,
+            })
+        else:
+            self.register_parameter("bias", None)
+
+    def weight_loader(self, param: Parameter, loaded_weight: torch.Tensor):
```
sdmyzlp: This function seems to be identical with that of `RowParallelLinear`, why do we need to rewrite it here?

zzhx1: in origin weight_load, 
```
tp_rank = get_tensor_model_parallel_rank()
tp_size = get_tensor_model_parallel_world_size()
```
we need replace it into custom comm group
```
tp_rank = self.tp_rank
tp_size = self.tp_size
```

It seems that the latest vllm does not have this problem.

sdmyzlp: Understood, thanks

vllm_ascend/quantization/quant_config.py
```diff
@@ -223,9 +225,14 @@ def apply(
...
         if isinstance(layer, RowParallelLinear):
-            tp_rank = get_tensor_model_parallel_rank()
-            return self.quant_method.apply(layer, x, bias, tp_rank)
-        return self.quant_method.apply(layer, x, bias)
+            if get_ascend_config().oproj_tensor_parallel_size is not None:
+                tp_rank = get_otp_group().rank_in_group
+            else:
+                tp_rank = get_tensor_model_parallel_rank()
+        else:
+            tp_rank = 0
```
ApsarasX, line 233: What means `tp_rank = 0`?

zzhx1: This origin code here
```python
if isinstance(layer, RowParallelLinear):
            tp_rank = get_tensor_model_parallel_rank()
            return self.quant_method.apply(layer, x, bias, tp_rank)
        return self.quant_method.apply(layer, x, bias)
```
The default situation is not passing tp, which is tp=0

vllm_ascend/ops/linear.py
```diff
@@ -0,0 +1,157 @@
...
+from vllm.distributed import divide, split_tensor_along_last_dim
+from vllm.distributed.parallel_state import GroupCoordinator
+from vllm.model_executor.layers.linear import (WEIGHT_LOADER_V2_SUPPORTED,
+                                               LinearBase, RowParallelLinear,
+                                               set_weight_attrs)
+from vllm.model_executor.layers.quantization.base_config import \
+    QuantizationConfig
+
+
+class CustomRowParallelLinear(RowParallelLinear):
```
wangxiyuan: register custom ops like other does https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/utils.py#L466

zzhx1: Completed, after registering custom ops, AscendRowParallelLinear is now effective globally for vllm-ascend. Although it does not affect the overall logic, but I am not sure if this is reasonable. The current rewrite of RowParallelLinear is just to change the original TP comm_group in __init__() and rewrite forward() so that it can adapt to OTP features.
Please review the latest modifications.
https://github.com/vllm-project/vllm-ascend/pull/2167/commits/58c70b5b7111c56427caacd3f4fefdbe5eaf2513

zzhx1:
After merging this pull request, I will refactor the `linear.py` file to support the custom communication parallel groups `RowParallelLinear` and `ColumnParallelLinear`, and align with #2120 by abstracting a unified interface to minimize redundant code.

tests/ut/distributed/test_parallel_state.py
```diff
@@ -0,0 +1,42 @@
+from unittest.mock import MagicMock, patch
+
+import pytest
+from vllm.config import ParallelConfig
+
+from vllm_ascend.distributed.parallel_state import (
+    _MC2, _OTP, destroy_ascend_model_parallel, get_mc2_group, get_otp_group,
```
ApsarasX: Better not import global variables directly from other files.

tests/ut/distributed/test_parallel_state.py
```diff
@@ -0,0 +1,42 @@
...
+        parallel_config.oproj_tensor_parallel_size = 2
+        init_ascend_model_parallel(parallel_config)
+
+        mc2_group = get_mc2_group()
+        assert mc2_group is not None
+        otp_group = get_otp_group()
+        assert otp_group is not None
+
+        destroy_ascend_model_parallel()
+        assert _MC2 is None
```
ApsarasX, line 46: ditto

tests/ut/distributed/test_parallel_state.py
zzhx1: Fixed completed

tests/ut/distributed/test_parallel_state.py
zzhx1: Fixed completed

ApsarasX:
Plz review @momo609 @jianzs @wangxiyuan @Angazenn

mypy.ini
```diff
@@ -3,6 +3,9 @@
 warn_unused_configs = True
 
 ; Suppress all missing import errors from torch_npu for mypy.
+[mypy-numpy.*]
+ignore_missing_imports = True
+
```
jianzs: Why do we need to update these configurations? If this is a bug in the repo, I suggest creating a separate PR to fix it.

vllm_ascend/torchair/models/torchair_deepseek_v2.py
```diff
@@ -514,11 +514,18 @@ def __init__(
...
+                                            quant_config=quant_config,
+                                            prefix=f"{prefix}.o_proj")
+        elif (config.n_routed_experts is not None
+              and self.debug_layer_idx >= config.first_k_dense_replace
+              and self.debug_layer_idx % config.moe_layer_freq == 0
+              and (ascend_config.torchair_graph_config.enable_multistream_moe
+                   or self.enable_shared_expert_dp)):
             self.o_proj = TorchairDeepseekV2RowParallelLinearReplaceAllreduce(
                 self.num_heads * self.v_head_dim,
                 self.hidden_size,
```
jianzs, line 531: Is it still not possible to eliminate these if-else branches even with CustomOp? @wangxiyuan @Yikun

vllm_ascend/ops/linear.py
```diff
@@ -134,14 +135,21 @@ def __init__(
...
+        elif prefix.find("o_proj") != -1 and oproj_tp_enable():
+            comm_group = get_otp_group()
+            self.forward_type = "oproj_tp"
         else:
-            self.tp_size = get_tensor_model_parallel_world_size()
-            self.tp_rank = get_tensor_model_parallel_rank()
-            self.enable_mlp_optimze = False
+            comm_group = get_tp_group()
+            self.forward_type = "normal"
+        self.comm_group = comm_group
```
jianzs: Is adding more if-else conditions the way to extend support for new models?

mypy.ini
zzhx1: Alright, I will remove this here. Here is just some local CI checks.

vllm_ascend/ops/linear.py
```diff
@@ -187,58 +195,106 @@ def __init__(
...
-            # dispose_tensor(output_parallel)
+
+        if self.input_is_parallel:
+            input_parallel = input_
+        else:
+            splitted_input = split_tensor_along_last_dim(
+                input_, num_partitions=self.tp_size)
+            input_parallel = splitted_input[self.tp_rank].contiguous()
+        assert self.quant_method is not None
+        # Choose different forward function according to the type of TP group
```
momo609: This part use dict format may be more extensible. The same logic applies as mentioned above.

vllm_ascend/utils.py
```diff
@@ -484,27 +485,25 @@ def register_ascend_customop():
...
+                                        AscendMergedColumnParallelLinear,
+                                        AscendRowParallelLinear)
     from vllm_ascend.ops.rotary_embedding import (
         AscendDeepseekScalingRotaryEmbedding, AscendRotaryEmbedding)
     CustomOp.register_oot(_decorated_op_cls=AscendQuickGELU, name="QuickGELU")
     CustomOp.register_oot(_decorated_op_cls=AscendSiluAndMul,
                           name="SiluAndMul")
     CustomOp.register_oot(_decorated_op_cls=AscendRotaryEmbedding,
                           name="RotaryEmbedding")
+    CustomOp.register_oot(_decorated_op_cls=AscendColumnParallelLinear,
```
momo609, line 507: If this component is enabled by default, modifications to the original vLLM repository will require ongoing maintenance and updates. What is the long-term maintenance strategy for this?

zzhx1: I think there won't be many changes here, we just need to focus on the maintenance of the `__init__` method in the follow-up.

vllm_ascend/ops/linear.py
zzhx1: I tried to modify it and found it not very intuitive, but I made some changed it to super.forward().

wangxiyuan:
according to the comment https://github.com/vllm-project/vllm-ascend/pull/2678#issuecomment-3241805082  please remove the patch_linear as well

wangxiyuan:
for the e2e lora error, please patch more like other does here https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/patch/worker/patch_common/patch_lora_embedding.py

zzhx1:
@wangxiyuan  This PR is ready，and also fixed the bug related to linearBase.

vllm_ascend/patch/worker/patch_common/patch_lora.py
```diff
@@ -0,0 +1,15 @@
+import vllm
```
wangxiyuan, line 1: looks that these 3 file can merged into one

PR #2183, opened 2025-08-04
[BugFix] Fix the bug that qwen3 moe doesn't work with aclgraph

wangxiyuan:
What's the PR does:
 1. Move AscendSparseMoeBlock to qwen3 model, since it's only used by qwen3 model.
 2. Disable AscendSparseMoeBlock if aclgraph  is enabled, AscendSparseMoeBlock doesn't work with aclgraph currently.

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/cdfd6871a5c4f125c9b3707ec5c1260db54f4b03


wangxiyuan:
depends on #2181

Yikun (approved):

PR #2186, opened 2025-08-04
SwiftBalancer Zero OverHead Expert Movement

raindaywhu:

Dynamic Experts load balance for MoE LLM Models, #1943 from v0.9.1-dev into main

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/c2e75b3c11047eec0f184577ce134879ce993f77


Yikun:

PR #2192, opened 2025-08-04
enable mm allreduce test

Ronald1995:
This PR is to add e2e test for using npu_mm_all_reduce_base fusion kernel.
no

not involved


ApsarasX:
This fusion operator `npu_mm_all_reduce_base` only accelerates the prefill phase, so it's best not to use it during the decode phase.

However, currently, as long as the environment variable `VLLM_ASCEND_ENABLE_MATMUL_ALLREDUCE `is enabled, this fusion operator will be applied in both prefill and decode stages. I believe we need a solution to address this issue.

Ronald1995:
> This fusion operator `npu_mm_all_reduce_base` only accelerates the prefill phase, so it's best not to use it during the decode phase.
> 
> However, currently, as long as the environment variable `VLLM_ASCEND_ENABLE_MATMUL_ALLREDUCE `is enabled, this fusion operator will be applied in both prefill and decode stages. I believe we need a solution to address this issue.

in some case such as the batch size is big enough, npu_mm_all_reduce_base will also gain performance improvement in decode phase.

ApsarasX:
> > This fusion operator `npu_mm_all_reduce_base` only accelerates the prefill phase, so it's best not to use it during the decode phase.
> > However, currently, as long as the environment variable `VLLM_ASCEND_ENABLE_MATMUL_ALLREDUCE `is enabled, this fusion operator will be applied in both prefill and decode stages. I believe we need a solution to address this issue.
> 
> in some case such as the batch size is big enough, npu_mm_all_reduce_base will also gain performance improvement in decode phase.

But often the decode phase cannot achieve enough batch size. I believe enabling `npu_mm_all_reduce_base` only during prefill is necessary.

Ronald1995:
> > > This fusion operator `npu_mm_all_reduce_base` only accelerates the prefill phase, so it's best not to use it during the decode phase.
> > > However, currently, as long as the environment variable `VLLM_ASCEND_ENABLE_MATMUL_ALLREDUCE `is enabled, this fusion operator will be applied in both prefill and decode stages. I believe we need a solution to address this issue.
> > 
> > 
> > in some case such as the batch size is big enough, npu_mm_all_reduce_base will also gain performance improvement in decode phase.
> 
> But often the decode phase cannot achieve enough batch size. I believe enabling `npu_mm_all_reduce_base` only during prefill is necessary.

if you only need to accelerate prefill phase ,you can use flashcomm(or sp), which is imcompatible with mm_allreduce

PR #2199, opened 2025-08-04
[V0.9.1][BugFix] Fix the bug in decoraotor patch

rjg-lyh:
Fix the bug in the decorator patch, which would have a negative impact on performance with mini batch. Now fix it by enabling this patch just when engine_v1 and flashcomm_v1 enabled together.

No.


Yikun (approved):

Yikun:
As offline discussed, we should remove `patch_decorator` and refactor according aclgraph capture size before v0.9.1 release

PR #2213, opened 2025-08-05
support bf16 model static eplb

shiyuan680:
support bf16 model static eplb

test in qwen3-moe-235b has 3% optimziation

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/1b9902806915040ac9b3029f2ab7522ec505afc3


shiyuan680:
we will add other ut(alltoallv ...) in refactor work

PR #2224, opened 2025-08-05
[main][Bugfix] Fix `AscendFusedMoE` not working with AclGraph

zhoux77899:

Parameter `active_num` in `torch_npu.npu_moe_init_routing` read from `hidden_states.shape` broken `AscendFusedMoE` working with AclGraph.
Register vllm-ascend's `AscendFusedMoE.forward` as a torch ops like vllm's `FusedMoE.forward`.

None

`tests/e2e/multicard/test_qwen3_moe.py` tests `AscendFusedMoE` working with AclGraph.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/51d5e9be7dbf4d914374447548dd01f9bfb68f89


vllm_ascend/models/qwen3_moe.py
```diff
@@ -108,7 +105,7 @@ def forward(
             router_logits=router_logits,
             is_prefill=is_prefill,
             top_k=self.top_k,
-            enable_force_load_balance=enable_force_load_balance,
+            enable_force_load_balance=False,
```
weijinqian0, line 118: It will lead to more accurate estimations?

zhoux77899: Directly set `enable_force_load_balance=False` because randomizing experts is not working with AclGraph.
Especially the experts randomization only occurs in vllm initialization.

PR #2268, opened 2025-08-07
[CI][Quickfix] Fix AscendFusedMoE init error

MengqingCao:
Fix AscendFusedMoE init error. Use `super().__init__()` instead of `super(FusedMoE, self).__init__()` to ensure the member variables in base class could be called by the children class

N/A
CI passed with new existing test.

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/766bc8162cb37ad32605eee051d4f049ec325926

### DIFF
--- a/tests/ut/ops/test_fused_ops.py
+++ b/tests/ut/ops/test_fused_ops.py
@@ -20,6 +20,7 @@ import torch
 import torch.nn as nn
 import torch_npu
 from pytest_mock import MockerFixture
+from vllm.model_executor.layers.fused_moe import FusedMoEMethodBase
 
 from vllm_ascend.ascend_forward_context import _get_fused_moe_state
 from vllm_ascend.ops.fused_moe import (AscendFusedMoE,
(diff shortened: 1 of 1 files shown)


vllm_ascend/ops/fused_moe.py
```diff
@@ -1181,8 +1181,27 @@ def __init__(
...
+            topk_group=topk_group,
+            quant_config=quant_config,
+            tp_size=tp_size,
+            ep_size=ep_size,
+            dp_size=dp_size,
+            prefix=prefix,
+            custom_routing_function=custom_routing_function,
+            scoring_func=scoring_func,
+            e_score_correction_bias=e_score_correction_bias,
+            activation=activation,
```
ApsarasX, line 1203: missing `apply_router_weight_on_input`?

MengqingCao: actually there is no `apply_router_weight_on_input` in `FusedMoE`:
https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/layer.py#L681-L705

ganyi1996ppo:
We need this pr merge asap to fix the ci issue

vllm_ascend/ops/fused_moe.py
ApsarasX: > actually there is no `apply_router_weight_on_input` in `FusedMoE`: https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/layer.py#L681-L705

image

MengqingCao: Thanks for this reminder, I'll fix it soon

PR #2274, opened 2025-08-08
Fix accuracy test create PR

wxsIcey:

Fix create PR of accuracy test 

no

Local testing: https://github.com/nv-action/vllm-benchmarks/pull/87

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/099c0464637f330f8ea38b07fe0694717c16d815

### DIFF
--- a/.github/workflows/accuracy_test.yaml
+++ b/.github/workflows/accuracy_test.yaml
@@ -313,7 +313,7 @@ jobs:
               head: `vllm-ascend-ci:${{ env.BRANCH_NAME }}`,
               base: 'main',
               title: `[Doc] Update accuracy reports for ${{ needs.accuracy_tests.outputs.vllm_ascend_version }}`,
-              body: `The accuracy results running on NPU Altlas A2 have changed, updating reports for: All models (Qwen/Qwen3-30B-A3B, Qwen2.5-VL-7B-Instruct, Qwen3-8B-Base)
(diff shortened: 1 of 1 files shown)


.github/workflows/accuracy_test.yaml
```diff
@@ -258,7 +258,7 @@ jobs:
           TIMESTAMP=$(date +%Y%m%d%H%M%S)
           BRANCH_NAME="auto-pr/accuracy-report-${TIMESTAMP}"
           echo "BRANCH_NAME=${BRANCH_NAME}" >> $GITHUB_ENV
-          git checkout -B "${BRANCH_NAME}" upstream/${{ github.event.inputs.vllm-ascend-version }}
+          git checkout -B "${BRANCH_NAME}" upstream/main
```
Yikun, line 261: That means currently we only support submit report in main branch:

image

We should think how to make it more general for all branches, first select the branch, and use the branch to run vllm-ascend branch.

Yikun (approved):
Let's merge this first to make main release work and then support dev branch

.github/workflows/accuracy_test.yaml
wxsIcey: Thanks. I think it can be implemented in this way. For example, when the next latest version is v0.10.0rc2, first check whether v0.10.0-dev exists in the repository. If so, submit a report PR to v0.10.0-dev. Otherwise, submit a report PR to main.

PR #2288, opened 2025-08-08
[Bugfix] fix bug in AllgatherEP

realliujiaxu:
When the FusedMoEState is set to AllGatherEP, there are two bugs:

1. DP>1 is not supported, communication for DP is missing.
2. After switching to CANN version 8.2.RC1, the behavior of npu_moe_init_routing_v2 was modified, causing the returned value expanded_x_idx to become random. This may lead to out-of-bounds errors when executing sorted_topk_weight= torch.index_select(topk_weights.view(-1), 0, expanded_x_idx).

No

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/f48a9af8924ea617a964b1158acc142b64843edb


vllm_ascend/ascend_forward_context.py
```diff
@@ -19,6 +19,7 @@ class FusedMoEState(Enum):
     AllGatherEP = 3
     NaiveMulticast = 4
     All2AllSeq = 5
+    AllGatherEPNaiveMulticast = 6
```
ApsarasX: Now we've added yet another state – isn't this getting a bit too complex? 😂

cc @zzzzwwjj

realliujiaxu:
@ApsarasX @zzzzwwjj please review

PR #2307, opened 2025-08-11
Optimize rope in Qwen3Moe

liyu119:

Optimize rope by extracting index_select from layers into model, which can reduce (layer_num -1) * 2 Gather ops in each prefill/decode stage. 

No.

### DIFF
--- a/vllm_ascend/models/qwen3_moe.py
+++ b/vllm_ascend/models/qwen3_moe.py
@@ -33,6 +33,7 @@ from vllm.model_executor.layers.layernorm import RMSNorm
 from vllm.model_executor.layers.linear import ReplicatedLinear
 from vllm.model_executor.layers.logits_processor import LogitsProcessor
 from vllm.model_executor.layers.quantization import QuantizationConfig
+from vllm.model_executor.layers.rotary_embedding import RotaryEmbedding
 from vllm.model_executor.layers.vocab_parallel_embedding import (
     ParallelLMHead, VocabParallelEmbedding)
 from vllm.model_executor.models.interfaces import (MixtureOfExperts,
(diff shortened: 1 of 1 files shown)


vllm_ascend/models/qwen3_moe.py
```diff
@@ -51,6 +52,40 @@
...
+        # Add qk-norm
+        q_by_head = q.view(*q.shape[:-1], q.shape[-1] // self.head_dim,
+                           self.head_dim)
+        q_by_head = self.q_norm(q_by_head)
+        q = q_by_head.view(q.shape)
+        k_by_head = k.view(*k.shape[:-1], k.shape[-1] // self.head_dim,
+                           self.head_dim)
+        k_by_head = self.k_norm(k_by_head)
+        k = k_by_head.view(k.shape)
+        if type(self.rotary_emb) is RotaryEmbedding:
```
ApsarasX, line 75: Doesn't Qwen3's MoE model all use RotaryEmbedding? So technically this 'if' condition has to be True, right?

vllm_ascend/models/qwen3_moe.py
```diff
@@ -51,6 +52,40 @@
...
+            # We optimized RotaryEmbedding by moving index_select of cos & sin outside.
+            # if cos & sin are provided, set is_cos_sin_cached to True to skip index_select.
+            q, k = self.rotary_emb(positions,
+                                   q,
+                                   k,
+                                   cos=cos,
+                                   sin=sin,
+                                   is_cos_sin_cached=True)
```
ApsarasX, line 83: Is this `is_cos_sin_cached` field actually being used anywhere?

liyu119: Yes, it's used in rope_forward_oot(vllm-ascend/ops/rotary_embedding.py). Pls refer to PR 1966 for more information.

vllm_ascend/models/qwen3_moe.py
```diff
@@ -294,10 +338,23 @@ def forward(
             assert intermediate_tensors is not None
             hidden_states = intermediate_tensors["hidden_states"]
             residual = intermediate_tensors["residual"]
+
+        # Generate cos and sin outside layers to avoid repeated calculation.
+        cos, sin = None, None
+        if self.cos_sin_cache is not None:
```
momo609, line 344: we can use the original rope in the first layer and reuse the precomputed cos_sin_cache in subsequent layers to eliminate these small operators logic here.

wangxiyuan:
we won't change model file any more

PR #2317, opened 2025-08-11
ut: add ci guard for ut coverage

Ronald1995:
add ci guard for ut coverage, if ut coverage of patch pr is below 80%, the ci will failed/

not involved

not involved

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/458e74eb907f96069e6d8a4f3c9f457001fef2ea

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -17,7 +17,7 @@
 
 coverage:
   status:
-     # Patch coverage is mandatory and must be >= 80%
+    # Patch coverage is mandatory and must be >= 80%
     patch:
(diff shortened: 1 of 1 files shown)


codecov.yml
```diff
@@ -20,9 +20,7 @@ coverage:
     # non-voting, new code must be fully tested
```
Yikun:
```suggestion
     # Patch coverage is mandatory and must be >= 80%
```

PR #1678, opened 2025-07-08
[Bugfix] Fix accuracy problem caused by mask pollution

ApsarasX:
### What this PR does / why we need it?
If a small batch of short requests is sent first, forming a chunk with a length <128, it will corrupt the `attn_mask_cache`, causing subsequent requests that do not form a chunk to have accuracy issues.

The root cause of this problem is the use of in-place multiplication. Modifying it to use out-of-place multiplication will resolve the accuracy problem.

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
Yes.

- vLLM version: v0.9.2
- vLLM main: https://github.com/vllm-project/vllm/commit/ad6c2e1a0b56c29065c7d70ff2e736e4f2fb03af

PR #2326, opened 2025-08-11
[V0.9.1][BugFix] Fix bugs and refactor cached mask generation logic

rjg-lyh:
This PR fix bugs and refactor cached mask generation logic. Now just pre-construct and use the cached mask on cpu instead of device on npu.

No.

### DIFF
--- a/vllm_ascend/attention/attention.py
+++ b/vllm_ascend/attention/attention.py
@@ -65,7 +65,6 @@ class AttentionMaskBuilder:
     def __init__(self, attn_mask: torch.Tensor):
         self._seq_len_cached = attn_mask.shape[0]
         self.attn_mask_cache = attn_mask
-        self.splitfuse_mask_value = -10000
 
     @classmethod
     def initialize_from_len(cls,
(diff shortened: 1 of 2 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -351,17 +351,8 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
...
-        # that is too small will require dynamic concatenation during inference,
-        # leading to performance degradation.
-        # Therefore, an environment variable is added here to dynamically set
-        # the size of the pre-constructed mask matrix based on requirements.
-        mask_len = os.getenv("PAGED_ATTENTION_MASK_LEN", 10000)
-        self.attn_mask_len = min(self.model_config.max_model_len,
-                                 int(mask_len))
         self.attn_mask_builder = AttentionMaskBuilder.initialize_from_len(
-            self.attn_mask_len, self.dtype)
+            self.model_config.max_model_len, self.dtype)
```
ApsarasX, line 354: This may cause memory waste.

Please refer to #1100

vllm_ascend/attention/attention.py
```diff
@@ -65,7 +65,8 @@ class AttentionMaskBuilder:
     def __init__(self, attn_mask: torch.Tensor):
         self._seq_len_cached = attn_mask.shape[0]
         self.attn_mask_cache = attn_mask
-        self.splitfuse_mask_value = -10000
+        self.bf16_mask_scale_factor = -10000
+        self.fp16_mask_scale_factor = 1
```
whx-sjtu: It's better to use a dict here with torch.dtype as key and padding number as value.

rjg-lyh: Yes, you are right. I have fix it by a staticmethod.

vllm_ascend/attention/attention.py
```diff
@@ -94,53 +100,28 @@ def get_decode_attn_mask(
...
-
-        return attn_mask.to(device, non_blocking=True)
+        self.update_attn_cache(max_seq_len, dtype)
+        # FIXME: Currently the mask value of chunked-prefill situation and Prefill-Only situation
+        # is not the same. Fix this in the future when kernel is ready.
+        mask_scale_factor = AttentionMaskBuilder.get_mask_scale_factor(dtype)
+        attn_mask = torch.index_select(self.attn_mask_cache,
+                                       dim=0,
+                                       index=position)[:, :max_seq_len]
+        attn_mask *= mask_scale_factor
```
ApsarasX, line 123: Will this place produce mask pollution fixed by #1678 ?

rjg-lyh: The mask is independently constructed, not a shallow copy from the cached mask, so i think that it won't cause any pollution.

PR #2337, opened 2025-08-12
Torchair graph812 cov

NicholasTao:


weijinqian0:
The related changes for torchair need to be placed under the torchair folder.

wangxiyuan:
torchair is removed from vllm-ascend now.

PR #2362, opened 2025-08-14
[Qwen-moe] use npu_add_rms_norm_quant operator

s30076806:
This PR  the npu_add_rms_norm_quant operator is enabled

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/00e3f9da462b31f271d9d9fdb526f148572609a9

### DIFF
--- a/vllm_ascend/models/qwen3_moe.py
+++ b/vllm_ascend/models/qwen3_moe.py
@@ -50,7 +50,9 @@ from vllm.sequence import IntermediateTensors
 from vllm_ascend.ops.fused_moe import AscendFusedMoE
 from vllm_ascend.ops.sequence_parallel import (MetadataForPadding,
                                                init_metadata_for_sp)
-
+from vllm_ascend.ops.layernorm import AddRMSNormW8A8Quant
+from vllm_ascend.quantization.w8a8 import AscendW8A8LinearMethod
+from vllm_ascend.quantization.quant_config import AscendQuantConfig
 
 class CustomSparseMoeBlock(Qwen3MoeSparseMoeBlock):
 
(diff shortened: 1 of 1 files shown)


vllm_ascend/envs.py
```diff
@@ -159,6 +159,11 @@
     #   1: enable moe all2all seq.
     "VLLM_ASCEND_ENABLE_MOE_ALL2ALL_SEQ":
     lambda: bool(int(os.getenv('VLLM_ASCEND_ENABLE_MOE_ALL2ALL_SEQ', '0'))),
+    # Whether to enable layernorm.py torch_npu.npu_add_rms_norm_quant
+    # 0: default
+    # 1: enable
+    "USE_ADD_RMSNORM_QUANT":
```
ApsarasX: Why is this switch necessary when using quantitative model weights?

In other words, are there any disadvantages to using `AddRMSNormW8A8Quant`?

If there are no disadvantages, I recommend not adding any more environment variables.

s-jiayang: OK, there should be no disadvantages.

vllm_ascend/models/qwen3_moe.py
```diff
@@ -183,8 +185,18 @@ def __init__(
...
                                    quant_config=quant_config,
                                    prefix=f"{prefix}.mlp")
-        self.input_layernorm = RMSNorm(config.hidden_size,
-                                       eps=config.rms_norm_eps)
+
+        assert isinstance(quant_config, AscendQuantConfig), \
+            "Expected quant_config to be an instance of AscendQuantConfig"
+        if isinstance(self.self_attn.qkv_proj.quant_method.quant_method,
+                      AscendW8A8LinearMethod):
+            self.input_layernorm = AddRMSNormW8A8Quant(
```
ApsarasX, line 193: We will use the `torch.fx` rewriter to solve this kind of problem.

Please refer #2389

PR #2365, opened 2025-08-14
[Bugfix] Update `transformers` version to `4.55.0`

shen-shanshan:

Fix error:

```bash
ImportError: cannot import name 'DeepseekV2Config' from 'transformers' (/home/sss/miniconda3/envs/vllm/lib/python3.10/site-packages/transformers/__init__.py)
```

Follow the version of `transformers` in vllm upstream, which is `4.55.0`.


wangxiyuan:
it's required by vllm already https://github.com/vllm-project/vllm/blob/main/requirements/common.txt#L10

PR #2373, opened 2025-08-14
[Qwen-moe] Remove the minor operation arange

s30076806:
Integrate the arange operator to reduce the time spent and improve performance

No

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/56dcf4e7e965e34043acf20ca4e4aceda21d41ec


s-jiayang:
I mentioned this pr very early, and has been verified locally, and no error is reported. In order to resolve the conflict of the previous integration, the pr was re-mentioned.

tests/e2e/singlecard/ops/test_fused_moe.py
```diff
@@ -148,7 +148,7 @@ def test_select_experts(
         mock_native_grouped_topk.side_effect = lambda x, num_groups, k: torch.randn_like(
             x)
 
-        topk_weights, topk_ids = select_experts(
+        topk_weights, topk_ids, _ = select_experts(
```
ApsarasX: Could add a test case to cover the `fused_experts` function with the `row_idx` parameter passed in.

vllm_ascend/ops/layers/experts_selector.py
```diff
@@ -83,7 +94,8 @@ def select_experts(hidden_states: torch.Tensor,
             e_score_correction_bias=e_score_correction_bias,
             global_num_experts=global_num_experts,
         )
-    return topk_weights, topk_ids
+        row_idx = return_row_idx(hidden_states, top_k)
```
ApsarasX, line 99: Seems all layers use the same `row_idx`, can we just construct it once?

wangxiyuan (approved):

PR #2384, opened 2025-08-14
[1/N][refactor] torchair deepseek modeling  refactor

linfeng-yuan:

Move torchair related model arch into torchair moduel to make the code clear. Next step we'll remove all torchair related code outside of torchair moduel.

No.

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/08d5f7113a024818b2867782c2539794b7aa162b

### DIFF
--- a/vllm_ascend/torchair/torchair_model_runner.py
+++ b/vllm_ascend/torchair/torchair_model_runner.py
@@ -27,6 +27,7 @@ from vllm.logger import logger
 
 from vllm_ascend.platform import NPUPlatform
 from vllm_ascend.torchair.utils import (check_torchair_cache_exist,
+                                        register_torchair_model,
                                         write_kv_cache_bytes_to_file)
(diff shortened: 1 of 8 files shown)


vllm_ascend/torchair/torchair_model_runner.py
```diff
@@ -169,3 +173,37 @@ def _capture_model(self):
         if self.new_kv_cache_bytes > 0:
             write_kv_cache_bytes_to_file(torch.distributed.get_rank(),
                                          self.new_kv_cache_bytes)
+
+    def load_model(self) -> None:
```
wangxiyuan: how about register the model in `__init__`, so that we don't need to override this func?

linfeng-yuan: > how about register the model in `__init__`, so that we don't need to override this func?

It works. I'll remove this block and register the torchair_models in __init__ func.

wangxiyuan:
merge this to unblock other refactor work. The CI failure related to another known issue.

PR #2388, opened 2025-08-15
Fix the accuracy issues caused by the mrope operator

zhihaofang1017:
Fix precision and inference length issues with MRoPE operator in multi-image long sequences for Qwen2.5-VL-7B

This PR addresses two critical issues observed with the MRoPE operator:
1. Precision degradation in V1 model outputs
2. Inference length limitations (only generating a few characters) when processing multi-image long sequences

Root cause analysis revealed that most CANN operators internally perform contiguous() operations to ensure accessing contiguous data during computations. However, the **”positions“** tensor was missing this crucial step, leading to incorrect memory access and corrupted values during operator calculations.

The fix adds a contiguous() operation on the positions tensor at the Python level, ensuring proper memory layout consistency with other tensor operations.

no

- vLLM version: v0.9.1
- vLLM main: https://github.com/vllm-project/vllm/commit/83fc60bdab2a4187799ece9b8a0f84695858975b

Comparing the accuracy of V0 and V1, the error meets the standard

### DIFF
--- a/csrc/kernels/pos_encoding_kernels.cpp
+++ b/csrc/kernels/pos_encoding_kernels.cpp
@@ -15,11 +15,6 @@
  */
 
 #include "kernel_operator.h"
-#include "kernel_tpipe_impl.h"
-#include "kernel_tensor_impl.h"
-#include "kernel_type.h"
-#include "kernel_operator_intf.h"
-#include "inner_interface/inner_kernel_operator_intf.h"
 #include <stdio.h>
 #include "types.h"
 #include "utils.h"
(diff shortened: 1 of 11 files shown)


zhihaofang1017:
image
Because of other people's spelling mistake, I couldn't pass all the tests

wangxiyuan:
you can fix it with this PR

zhihaofang1017:
Because positions don't have .contiguous() enabled, there are accuracy issues. Before merging the code, I always have a test that fails. It's about ensuring the output is identical when using and not using TP. The specific test script is here **(vllm-ascend/tests/multicard/test_multimodal_context_parallel.py).**
I run several experiments.
1. Commit: **Fix the accuracy issues caused by the mrope operator**. When I added the following line after positions (query, key = torch_npu.npu_mrope(positions.clone().detach().contiguous(),) as normal, the test failed.
2. commit: **Check the test**. I added an irrelevant line before torch_npu.npu_mrope (positions = positions
query, key = torch_npu.npu_mrope(positions,)). This time, the TP verification test didn't even get to the point.
3. commit: Check the test2. I only added the clone information (query, key = torch_npu.npu_mrope(positions.clone())), but the test still failed.

Based on Experiment 3, I suspect there's something wrong with the script for the test case that checks whether TP is enabled.

Yikun:

wangxiyuan:
0.9.1 is released. Feel free move the change to main

Issue #2386, opened 2025-08-15
[RFC]: Automatic Kernel Fusion via torch.fx.graph and graph rewriter for vLLM-Ascend


ganyi1996ppo:
### Motivation.

Currently, the **vLLM** project’s high-performance execution path relies heavily on **PyTorch Inductor** to perform automatic kernel fusion. However, for **vLLM-Ascend**, we face two major limitations:

1. **No Inductor Support in `torch_npu`**  
   - The `torch_npu` backend does not implement Inductor becaused of the limited support on triton, meaning we cannot leverage vLLM’s existing kernel fusion mechanisms out-of-the-box.

2. **High Maintenance Burden from Model Duplication**  
   - To achieve equivalent fusion behavior, we have been rewriting and maintaining large portions of model code directly inside **`vllm-ascend`**, like Qwen3, Qwen3Moe, Qwen2VL etc. Some of them only have very few changes compared with the community one like Qwen3 only add a simple fusion kernel on `add_rms_norm` and `quantize`.
   - This results in:
     - Code duplication across `vllm` and `vllm-ascend`.
     - Divergence over time, making upgrades from upstream `vllm` difficult.
     - Reduced maintainability and increased development cost.

Without an automatic fusion mechanism, **Ascend NPU execution suffers from suboptimal kernel launch patterns**, leading to:
- Higher latency due to excessive kernel dispatch.
- Poor hardware utilization.
- Increased complexity in managing performance-critical paths.

We need a solution that:
- **Keeps all model definitions inside `vllm`**.
- **Implements NPU-specific fusions in `vllm-ascend`** without forking or rewriting model code.
- Is **maintainable and extensible** for future optimizations.

### Proposed Change.

We propose implementing an **Automatic Kernel Fusion Compiler Interface** for Ascend NPUs that:
- Operates on **`torch.fx.graph`** extracted from original `vllm` models.
- Uses **PyTorch’s Graph Rewriter API** (`torch.fx.subgraph_rewriter`) to:
  - Match known computational subgraphs that can be fused.
  - Replace them with fused kernels implemented in `torch_npu`.
- Runs **before** the model is executed, ensuring the runtime graph is already optimized for Ascend NPUs.

#### Design Overview

Our approach builds on the existing `vllm` execution pipeline while replacing only the Inductor-specific fusion logic with an NPU-compatible implementation. The key changes are:

1. **Leverage vLLM’s Existing `VllmBackend` Path**  
[...]

PR #2389, opened 2025-08-15
[core] Adopt graph rewriter on fx.graph to enable automatic kernel fusion

ganyi1996ppo:
The main goal of this PR to alleviate the high maintenance burden from model duplication when we are going to do the model optimization. Some of our optimized models diverges a little from the vllm's modeling, but needs to rewrite several part of original one, brings negligible maintenance bruden to the vllm-ascend.

In order to solve that, we propose to leverage `torch.compile`, and automatically fuse the pattern we want to merge. For more details can refer to the RFC https://github.com/vllm-project/vllm-ascend/issues/2386

Yes, we add a new `additional_config` 

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/a11adafdcab61c059d2a76d952367a722e1b71d5

### DIFF
--- a/vllm_ascend/compilation/quant_fusion_pass.py
+++ b/vllm_ascend/compilation/quant_fusion_pass.py
@@ -16,7 +16,7 @@
 # limitations under the License.
 #
 
-from typing import List, Tuple, Callable
+from typing import Callable, List, Tuple
 
 import torch
 from torch.fx.subgraph_rewriter import replace_pattern
--- a/tests/e2e/singlecard/test_graph_rewriter.py
+++ b/tests/e2e/singlecard/test_graph_rewriter.py
@@ -18,7 +18,7 @@
 
 import copy
 import re
-
+from typing import Dict, Any
 import pytest
 import torch
 import torch.nn as nn
@@ -61,7 +61,7 @@ class ModelWithRMSNormQuant(nn.Module):
 
 class CustomizeCompilationInterface:
 
-    def __init__(self, vllm_config, checking_fusion_pass: str = "torch.ops"):
+    def __init__(self, vllm_config, checking_fusion_pass: Dict[str, tuple[Any, Any]]):
         self.vllm_config = vllm_config
         self.graph_rewriter_manager = GraphRewritePassManager()
         self.graph_rewriter_manager.configure(vllm_config)
@@ -96,6 +96,7 @@ class TestGraphRewriter(PytestBase):
     @pytest.mark.parametrize("num_tokens", NUM_TOKENS)
     @pytest.mark.parametrize("hidden_size", HIDDEN_SIZES)
     def test_quant_fusion_pass(
+        self,
         num_tokens: int,
         hidden_size: int,
     ) -> None:


ApsarasX:
👍🏻👍🏻👍🏻This's a super smart idea – using compiler tricks to make model code way simpler.

vllm_ascend/ascend_config.py
```diff
@@ -105,6 +108,19 @@ def __init__(self, torchair_graph_config):
                     "enable_kv_nz is valid only when Torchair graph mode is enabled"
                 )
 
+class AscendCompilationConfig:
+    """
+    Configuration Object for ascend_compilation_config from additional_config
+    """
+
+    def __init__(self, ascend_compilation_config: dict):
```
jgong5: Suggest to explicitly name the options as args (e.g. `enable_graph_rewriter = True` ...) and if you want things extensible, I guess **kwargs is more pythonic.

docs/source/user_guide/configuration/additional_config.md
```diff
@@ -49,6 +50,13 @@ The details of each config option are as follows:
 | `graph_batch_sizes_init` | bool | `False` | Init graph batch size dynamically if `graph_batch_sizes` is empty |
 | `enable_kv_nz`| bool | `False` | Whether to enable kvcache NZ layout. This option only takes effects on models using MLA (e.g., DeepSeek). |
 
+**ascend_compilation_config**
+| Name | Type | Default | Description |
+| ---- | ---- | ------- | ----------- |
+| `enable_graph_rewrite` | bool | `True` | Whether to enable the graph rewriter to rewrite the fx graph generated by torch.compile |
```
jgong5: Shall we highlight that this option is a primary flag that could turn off all the compilation and make the other compiler options ignored?

docs/source/user_guide/configuration/additional_config.md
```diff
@@ -49,6 +50,13 @@ The details of each config option are as follows:
...
+| `enable_quantization_fusion` | bool | `True` | Whether to enable the fusion pass on op + quantize, this should remain open by default to benefit all users for performance boost |
```
jgong5, line 58: Not sure if we really want to expose this as an official configuration flag here. I feel this complicate the UI for normal users. Some considerations:
1. What is the granularity of such configurations? For example, is it better to make each fusion configurable, e.g., naming it `enable_rmsnorm_quant_fusion`?
2. Shall we expose this to normal users? Or, is it better to make it a private option first, e.g., `_enable_quantization_fusion` so that we make it as a private and debugging-only flag?

vllm_ascend/patch/__init__.py
```diff
@@ -102,3 +102,24 @@
...
+#    Why:
+#       We need to use the `GraphRewriterPassManager` at the compiler interface to actually modify the graph. Therefore we 
+#       need to implement our own compiler interface to do our customized operations.
+#    How:
+#       We implement our own compiler interface `AscendAdaptor` to fetch the `GraphRewriterPassManager` and use it to rewrite the 
+#       piecewise graph cutting by vllm's own backend. This function will just return the `AscendAdaptor` to the  `VllmBackend``.
+#    Related PR (if no, explain why):
+#       - We might add PR to make vllm support custom compiler interface. But its not sure yet.
+#    Future Plan:
+#       We might push the customized compiler interface to the vllm main repo, and leave the backend selection to the platform itself.
```
jgong5, line 117: Do I understand correctly that vLLM hard-coded "inductor" as the compiler backend for piece-wise graphs? Is there a way to plugin the custom compiler backend instead of "inductor"?

vllm_ascend/compilation/quant_fusion_pass.py
```diff
@@ -0,0 +1,77 @@
...
+
+      def replace(rms_norm_input, residual, rms_norm_weight, scale, offset):
+          """
+          Replacement for the AddRMSNormQuant fusion.
+          """
+          output = torch.ops.npu.npu_add_rms_norm_quant(
+              rms_norm_input, 
+              residual, 
+              rms_norm_weight, 
+              1. / scale,
```
jgong5: I guess this worth some comments? :)

vllm_ascend/compilation/quant_fusion_pass.py
```diff
@@ -0,0 +1,77 @@
...
+class AscendQuantFusionPass(VllmInductorPass):
+    """
+    A pass for fusing AddRMSNorm and W8A8 quantization operations on Ascend.
+    """
+
+    def __init__(self, vllm_config):
+        super().__init__(vllm_config)
+        self.patterns = []
+        # Register the AddRMSNormQuant fusion pattern into the graph rewriter pattern list
+        AddRMSNormQuantPattern(vllm_config).register(self.patterns)
```
jgong5, line 73: Perhaps it is better to use some decorator to register new patterns, following open/closed principle...

vllm_ascend/ascend_config.py
ganyi1996ppo: Good suggestion, I'll change this

vllm_ascend/compilation/quant_fusion_pass.py
ganyi1996ppo: Sure

vllm_ascend/patch/__init__.py
ganyi1996ppo: No, vllm actually rewrite its own backend called 'VllmBackend', and inside of that vllm will do its own graph break and pattern register routine to make sure the optimization and compatibility to other repos and packages

docs/source/user_guide/configuration/additional_config.md
ganyi1996ppo: Got, I'll emphasis this one

docs/source/user_guide/configuration/additional_config.md
ganyi1996ppo: This is mainly refer to the vllm's official code. Its true that the granularity here is quite obscure in this PR, but this flag is more like a safety trigger here to make sure that if anything goes wrong, we can quickly guide our customer to bypass some issue.

ganyi1996ppo:
Just notice the `piecewise_graph_backend.py` was removed last week, I port the changes on that file to the `aclgraph.py` to make sure the fx_graph fusion on eager mode works as expected. Please review if the changes work on our current graph solution @MengqingCao

vllm_ascend/compilation/compiler_interface.py
```diff
@@ -0,0 +1,75 @@
...
+        arg_dtypes = get_dtype_from_args(example_inputs)
+        arg_shapes = get_shapes_from_args(example_inputs)
+        kwargs = {
+            "runtime_shape": runtime_shape,
+            "arg_shapes": arg_shapes,
+            "arg_dtypes": arg_dtypes
+        }
+        graph = graph_rewriter_manager(graph, **kwargs)
+
+        compilation_counter.num_eager_compiles += 1
```
MengqingCao, line 72: This pr only support rewrite graph in eager mode, is my understanding correct?

vllm_ascend/compilation/acl_graph.py
```diff
@@ -102,13 +104,16 @@ def __call__(self, *args, **kwargs):
         aclgraph_runtime_mode = forward_context.cudagraph_runtime_mode
 
         if aclgraph_runtime_mode == CUDAGraphMode.NONE or \
-                            aclgraph_runtime_mode != self.runtime_mode:
+                            aclgraph_runtime_mode != self.runtime_mode or \
+                                self.fx_graph_eager:
```
MengqingCao, line 110: I think this is enough to make `ACLGraphWrapper` return an eager graph when rewriting fx graphs

vllm_ascend/compilation/quant_fusion_pass.py
```diff
@@ -0,0 +1,94 @@
...
+    def __init__(self, vllm_config):
+        self.vllm_config = vllm_config
+
+    def register(self, patterns: List[Tuple[callable]]):
+
+        def pattern(rms_norm_input, residual, rms_norm_weight, scale, offset):
+            """
+          Pattern for AddRMSNormQuant fusion.
+          """
+            output = torch.ops.npu.npu_add_rms_norm(rms_norm_input, residual,
```
MengqingCao, line 37: could a similar patten `add -> rms_norm -> quant` to be matched, either? Or we still have to implentment a new `XXXPattern` to achieve this?

vllm_ascend/compilation/quant_fusion_pass.py
```diff
@@ -0,0 +1,94 @@
...
+    def __init__(self, vllm_config):
+        super().__init__(vllm_config)
+        self.patterns = []
+        # Register the AddRMSNormQuant fusion pattern into the graph rewriter pattern list
+        AddRMSNormQuantPattern(vllm_config).register(self.patterns)
+
+    def __call__(self, graph: torch.fx.Graph):
+        self.begin()
+        for pattern, replace in self.patterns:
+            replace_pattern(graph, pattern, replace)
```
MengqingCao, line 78: QQ: Does this replacement happen after the whole fx graph is captured?

ganyi1996ppo: Yes, its taking effect inside the torch.compile

vllm_ascend/compilation/quant_fusion_pass.py
ganyi1996ppo: New pattern is required for that. But we don't need them since we already have the add_rms_norm fusion in vllm.

vllm_ascend/compilation/compiler_interface.py
ganyi1996ppo: No, aclgraph can also take advantage of this, if you `fx_graph_eager` is disabled.

Yikun:
Replaced by: https://github.com/vllm-project/vllm-ascend/pull/4168 @ganyi1996ppo Thanks for your initial work, I also add you as coauthor in #4168

ganyi1996ppo:
> Replaced by: #4168 @ganyi1996ppo Thanks for your initial work, I also add you as coauthor in #4168

Congrats! New PR looks amazing.

PR #2394, opened 2025-08-15
[bugfix] ascend schedule encountered an incorrect req block length in…

liziyu179:
… the check_watermark_for_prefill function

ascend schedule encountered an incorrect req block length in the check_watermark_for_prefill function，under the current writing method, it will always be 1. 

before：
http://image.huawei.com/tiny-lts/v1/images/mdstorm/c6cff7cf33d500a3833f5f80352df373_1183x377.png
after：
http://image.huawei.com/tiny-lts/v1/images/mdstorm/57207a490d8ac0a70fc87dd08d02dee6_1470x954.png


whx-sjtu:
LGTM

PR #2400, opened 2025-08-15
[0.9.1][bugfix] Unify MoE routing init with standard torch_npu operator

SlightwindSec:
### Description
This PR fixes the dependency on a POC version of `torch_npu` for the MoE routing initialization feature.

**Before:**
To get the best performance, users needed a `torch_npu` version containing the `npu_moe_init_routing_quant` operator. Official versions would trigger a slower, pure PyTorch fallback.

**After:**
The code is updated to use the `npu_moe_init_routing_v2` operator, which is included in the official `torch_npu` releases and provides equivalent performance. This change unifies the implementation, removes the fallback logic, and makes the high-performance path accessible to all users without requiring a special library version.


Yikun:

SlightwindSec:
> v0.9.1-dev are code freezing, can you make sure is this still needed? or just move to main branch. Thanks.

OK

PR #2409, opened 2025-08-18
[Bugfix] Fix custom op register issue

ganyi1996ppo:
Our current code register the custom ops inside the platform intialization phase. however, when a new process started by creating a worker, the former patch will lose it effect on the custom ops and lead to fallback to the native pass wrote in vllm. This PR move the patch code to the worker to make sure the custom op patch worker as our expected.

No

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/8ea0c2753a273e24957ab4587c200a3254ebe970


MengqingCao:
LGTM

PR #2434, opened 2025-08-19
[Scheduler] validate max_num_batched_tokens and max_model_len in AscendSchedulerConfig

linfeng-yuan:
Add configuration check logic for ascend scheduler: if chunked_prefill is disabled, max_num_batched_tokens couldn't be less than max_model_len, following vLLM; 

users cannot set max_num_batched_tokens smaller than max_model_len with ascend scheduler
CI and vllm serving passed

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/f77a0802b758a32c5b9f7bc04e9498d77e8d99e0


wangxiyuan:
please update the commit message. Thanks

PR #2436, opened 2025-08-19
Add gpt oss

ChenTaoyu-SJTU:
This PR is for verifying the reason causing the GPT-OSS accuracy problem. After completing the following adaptation, the accuracy should works well:

1. moe bias
2. `swiglu_oai` activation function
3. sink attention
4. sliding window attention

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/0c6e40bbaa4707528286a1e7bf17c90c88a1d920


ChenTaoyu-SJTU:
This PR is move to #2572

PR #2437, opened 2025-08-19
[1/N][Draft][Refactor]torchair pangu_moe modeling refactor

Angazenn:

1. Similar to #2384 , this PR add a torchair-specific modeling for pangu.
2. Fixes a bug introduced by routed_scaling_factor in #2675 .
3. remove eager test case for pangu since there has already been a torchair test case.

No.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/6997a25ac65ed6cc3c2be6d09ca45f633a345f63

### DIFF
--- a/tests/e2e/multicard/test_torchair_graph_mode.py
+++ b/tests/e2e/multicard/test_torchair_graph_mode.py
@@ -22,8 +22,6 @@ Run `pytest tests/multicard/test_torchair_graph_mode.py`.
 import os
 from typing import Dict
 
-import pytest
-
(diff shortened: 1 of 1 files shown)


vllm_ascend/models/pangu_moe.py
```diff
@@ -57,7 +57,6 @@
 from vllm.model_executor.utils import set_weight_attrs
```
wangxiyuan, line 57: I think we can delete this file

wangxiyuan (approved):

PR #2438, opened 2025-08-19
[1/N][refactor] torchair fused_moe refactor

hust17yixuan:
Move torchair related fused_moe section into torchair_fused_moe to make the code clear. Next step we'll remove all torchair related code outside of torchair_fused_moe .

No
vLLM version: v0.10.0
vLLM main: https://github.com/vllm-project/vllm/commit/08d5f7113a024818b2867782c2539794b7aa162b


vllm_ascend/torchair/models/torchair_deepseek_v2.py
```diff
@@ -335,7 +335,7 @@ def __init__(
         else:
             self.gate.e_score_correction_bias = None
 
-        self.experts = AscendFusedMoE(
+        self.experts = TorchairAscendFusedMoE(
```
wangxiyuan, line 338: @wxsIcey please note that fused moe in torchair module is changed to this. So for your Fused Moe custom ops refactor  work, you can just ignore torchair case.

PR #2448, opened 2025-08-20
[misc] remove uesless envs

wangxiyuan:
Remove the env used for v0 pd feature.


MengqingCao (approved):
This change is reasonable as the removed env variables are not needed in V1

PR #2453, opened 2025-08-20
feat: add mtp ut and fix some bugs

ZhaoJiangJiang:
Fix mtp mode ut

Nothing

This can be tested in the same way as a unit test.

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/de7b67a0232e35ae8e8ecd944aeddfc8cbc02631


tests/e2e/singlecard/spec_decode_v1/test_v1_mtp_correctness.py
```diff
@@ -61,28 +59,43 @@ def test_mtp_correctness(
...
 
-    spec_llm = LLM(model=model_name,
-                   trust_remote_code=True,
-                   speculative_config={
-                       "method": "deepseek_mtp",
-                       "num_speculative_tokens": 1,
-                   },
-                   max_model_len=256,
-                   enforce_eager=True)
+    spec_llm = LLM(
```
wangxiyuan: use VLLMRunner, it can clean the memory automtic

vllm_ascend/models/deepseek_mtp.py
```diff
@@ -216,3 +220,79 @@ def forward(
                                    attn_metadata, previous_hidden_states,
                                    inputs_embeds, spec_step_idx)
         return hidden_states
+
+    def load_weights(self, weights: Iterable[tuple[str,
```
wangxiyuan: please chekc that if v0.10.1 and main for vLLM has the problem, if not, please remove this override

linfeng-yuan: If this function is needed, please add this function also in ./torchair/models/torchair_deepseek_mtp.py

PR #2462, opened 2025-08-20
[Do not merge][0.9.1][feat] use tensor_list for gmm

linfeng-yuan:


Yikun:
This PR depends on CANN Q3 release we should revisit this in post1 or main

PR #2469, opened 2025-08-21
[2/N][Feat] Add MC2 communication method for MoE layers

yiz-liu:
This method replaces the previous all-gather approach for small numbers of tokens.

The key changes include:
- A new `AscendFusedMoE` layer that handles token splitting, local computation, and final aggregation via all-gather.
- Logic in the model runner to dynamically select between the new MC2 method and the existing all-gather method based on the number of input tokens.
- Sharding the MoE communication mask across tensor-parallel ranks.

None.

Test case fixed.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/b00e69f8ca55f4a82847d39466f57ceb748324c1

### DIFF
--- a/vllm_ascend/distributed/parallel_state.py
+++ b/vllm_ascend/distributed/parallel_state.py
@@ -12,6 +12,12 @@ _MC2: Optional[GroupCoordinator] = None
 _MLP_TP: Optional[GroupCoordinator] = None
 
 
+def get_mc2_comm_name(rank: int) -> str:
+    assert _MC2 is not None, ("mc2 group is not initialized")
+    return _MC2.device_group._get_backend(
+        torch.device("npu")).get_hccl_comm_name(rank)
+
+
 def get_mc2_group() -> GroupCoordinator:
     assert _MC2 is not None, ("mc2 group is not initialized")
     return _MC2
--- a/vllm_ascend/distributed/parallel_state.py
+++ b/vllm_ascend/distributed/parallel_state.py
@@ -12,12 +12,6 @@ _MC2: Optional[GroupCoordinator] = None
 _MLP_TP: Optional[GroupCoordinator] = None
 
 
-def get_mc2_comm_name(rank: int) -> str:
-    assert _MC2 is not None, ("mc2 group is not initialized")
-    return _MC2.device_group._get_backend(
-        torch.device("npu")).get_hccl_comm_name(rank)
-
-
 def get_mc2_group() -> GroupCoordinator:
     assert _MC2 is not None, ("mc2 group is not initialized")
     return _MC2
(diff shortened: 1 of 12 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1557,6 +1559,9 @@ def execute_model(
              intermediate_tensors) = (self._prepare_inputs(
                  scheduler_output, intermediate_tensors))
 
+        moe_comm_method = (self.moe_comm_method if num_input_tokens
```
weijinqian0: Does the `moe_comm_method` get initialized every time `forward` is executed?

yiz-liu: Once every step.

yiz-liu: Now only init once in `AscendFusedMoE`.

vllm_ascend/distributed/parallel_state.py
```diff
@@ -12,6 +12,12 @@
 _MLP_TP: Optional[GroupCoordinator] = None
 
 
+def get_mc2_comm_name(rank: int) -> str:
+    assert _MC2 is not None, ("mc2 group is not initialized")
+    return _MC2.device_group._get_backend(
+        torch.device("npu")).get_hccl_comm_name(rank)
+
```
weijinqian0: If it is only used within the Mc2 method, there is no need to extract a separate function.

vllm_ascend/utils.py
```diff
@@ -500,6 +500,9 @@ def register_ascend_customop():
     from vllm_ascend.ops.layernorm import AscendRMSNorm
     CustomOp.register_oot(_decorated_op_cls=AscendRMSNorm, name="RMSNorm")
 
+    from vllm_ascend.ops.common_fused_moe import AscendFusedMoE
+    CustomOp.register_oot(_decorated_op_cls=AscendFusedMoE, name="FusedMoE")
```
momo609, line 513: Will the fused MoE also be affected in eager execution mode?

vllm_ascend/distributed/parallel_state.py
yiz-liu: Will revert it.

yiz-liu:
Look into the memory problem with `DummyCommImpl`.

vllm_ascend/utils.py
yiz-liu: Not affected.

yiz-liu:
1. Fix UT;
2. Delete `NaiveMulticast` related codes.

yiz-liu:
> 1. Fix UT;
> 2. Delete `NaiveMulticast` related codes.

Fixed, please merge it at your earliest convenience, @wangxiyuan

wangxiyuan (approved):

PR #2472, opened 2025-08-21
[Fix] fix resources limit error when apply speculative decoding and aclgraph

WithHades:
When both speculative decoding and aclgraph are applied, and cudagraph_capture_sizes uses the default value, it will report that the stream resources are insufficient.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/9c99e4871f12aa57072b3dfd14157018d30e327b


wangxiyuan (approved):

wangxiyuan:
please rebase to main to make CI happy

yiz-liu:
I checked the CI failures, should be irrelevant to this PR, maybe try rebase main and see if any of those happens again? @WithHades

PR #2478, opened 2025-08-21
[Bugfix] Fix the bug that qwen3 moe doesn't work with aclgraph

shen-shanshan:

What's the PR does:

1. Move AscendSparseMoeBlock to qwen3 model, since it's only used by qwen3 model.
2. Disable AscendSparseMoeBlock if aclgraph is enabled, AscendSparseMoeBlock doesn't work with aclgraph currently.

### DIFF
--- a/tests/multicard/test_qwen3_moe.py
+++ b/tests/multicard/test_qwen3_moe.py
@@ -1,0 +1,55 @@
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+# Copyright 2023 The vLLM team.
+#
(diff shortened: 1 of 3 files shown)


tests/multicard/test_qwen3_moe.py
```diff
@@ -0,0 +1,55 @@
...
+"""
+
+from tests.conftest import VllmRunner
+
+
+def test_models_distributed_Qwen3_MOE_TP2():
+    example_prompts = [
+        "Hello, my name is",
+    ]
+    dtype = "half"
```
wangxiyuan, line 31: vllm runner actually runs in eager mode by default.

PR #2479, opened 2025-08-21
[Bugfix] Fix the bug of incorrect precision

weiguihua2:
Fix the bug of incorrect precision

- vLLM version: v0.10.0
- vLLM main: https://github.com/vllm-project/vllm/commit/53415653ff24be03e7c90f5b42ef9cb3f72aad71

### DIFF
--- a/vllm_ascend/quantization/quantizer.py
+++ b/vllm_ascend/quantization/quantizer.py
@@ -75,8 +75,8 @@ class VLLMAscendQuantizer:
                     "vllm.model_executor.layers.layernorm.RMSNorm", "__init__",
                     [wrapper_rmsnorm_init])
                 VLLMAscendQuantizer.apply_patch(
-                    "vllm_ascend.ops.layernorm.AscendRMSNorm",
-                    "forward_oot", [wrapper_rmsnorm_forward_oot])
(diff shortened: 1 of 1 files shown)


vllm_ascend/quantization/quantizer.py
```diff
@@ -75,8 +75,8 @@ def __init__(self, quant_description):
                     "vllm.model_executor.layers.layernorm.RMSNorm", "__init__",
```
wangxiyuan, line 75: does other patch has the simliar problem as well?

MengqingCao: No, as we only override the `forward_oot` of `RMSNorm`

MengqingCao (approved):
LGTM

vllm_ascend/quantization/quantizer.py
jgong5: RMSNorm is a `CustomOp` from vllm, why do we have to do monkey patch to override things?

wangxiyuan: @jgong5 this is a hack way for quantization. this is really bad and ugly, we'll refactor this later to avoid patch!

PR #2483, opened 2025-08-22
[Embedding] Recover embedding function

wangxiyuan:
Fix broken embedding function. It's broken by http://github.com/vllm-project/vllm/pull/23162

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/efc88cf64a399f5459cd6256223e99672c13614d


Yikun:
codecov/patch failed due to we add branch to support diff version, so It's ok to go to recover pool feature

PR #2486, opened 2025-08-22
Prefetching MoE weights to enhance the performance of Qwen3-235B.

coder-fny:
Co-authored-by: zhaoning <zhaon14@126.com>


wangxiyuan:
we'll remove model files from vllm-ascend. Please find a better way which doesn't need change model file

wangxiyuan:
we won't change model files any more.

PR #2489, opened 2025-08-22
add logger.warning to avoid possible oom conditions

zouyida2052:
after capturing graph, the available memory on each NPU would be scarce, so we introduce this warning log to remind users.

no

by ci.


wangxiyuan:
nice improvment. Please let CI pass

PR #2493, opened 2025-08-22
[Feat] Enable SP across DP for DeepSeekV2 MLA

clrs97:
The DeepSeekV2 MLA computation is facing two problems under multi-way DP configuration:
- Incoming requests often have varying input sequence lengths, leading to the unbalance among each DP.
- When there aren't many requests, some DP shards will just do dummy run and become useless.

This PR introduces a staged hybrid parallelism strategy optimized for MLA computation during the prefill phase:
<div align="center">

pic

Picture from https://arxiv.org/pdf/2506.12708 Fig. 16
</div>

- Stage 1: All the input tokens among all DP shards will be evenly rearranged to each rank. (SP)
- Stage 2: For each DP, All-Gather to get its own tokens and apply TP to run attention. After solving the KV Cache problem, this stage will be further optimized such that each DP is able to handle all the global tokens instead.
- Stage 3: All-to-All to run SP for `o_proj`. Here the weight of `o_proj` matrix will be completed. To save the memory, this PR introduces a shard approach. The weights of `o_proj` are kept at different ranks, and broadcast asynchronously to get the complete matrix in `FRACTAL_NZ` format for current layer during the computation.

This PR introduces three new configs in `additional_config`:
| Name | Effect | Required | Type | Constraints |
|---------|---------|---------|---------|---------|
| `enable_mla_sp` | Whether to enable SP for DeepSeekV2 MLA. | No | bool | Must be enabled with `o_proj` shard feature. It is temporary disabled when some other features are enabled. |
| `o_shard_parallel_size` | How to divide the rows of the `o_proj` matrix evenly across ranks for shard. | No | int | It should be at least 2 and should be a divisor of `world_size`. When it is missing, the shard feature is disabled. |
| `o_shard_full_layers` | The first complete `o_shard_full_layers` layers will be kept among all ranks without shard. When computing the `k`-th layer, the `o_proj` weight of the `k + o_shard_full_layers`-th layer will be broadcast asynchronously. | No | int | Default value is `0`. |

Example:
`--additional-config '{"enable_mla_sp": true, "o_shard_parallel_size": 8, "o_shard_full_layers": 5}'`

vLLM main: https://github.com/vllm-project/vllm/commit/a4454e9401c9c76150dc3a4d23d10377926b2fb6

- For DP 4 TP 4 configuration, start script:
```
export VLLM_USE_V1=1
export TASK_QUEUE_ENABLE=1

python -m vllm.entrypoints.openai.api_server --model=DeepSeek-R1-W8A8-VLLM \
 --served-model-name auto \
 --load-format=auto \
 --trust-remote-code \
 --enforce_eager \
 -tp=4 \
 -dp=4 \
 --quantization ascend \
 --additional-config '{"chunked_prefill_for_mla": true, "enable_mla_sp": true, "o_shard_parallel_size": 8, "o_shard_full_layers": 5}' \
 --port 8006 \
 --max-num-seqs 24 \
 --max-model-len 8192 \
 --block-size 128 \
 --gpu-memory-utilization 0.96 \
 --no-enable-prefix-caching
```
| max-concurrency | #requests | input tokens distribution | SP Throughput | Baseline Throughput |
|---------|---------|---------|---------|---------|
| 1 | 200 | Fixed 4096 | **0.77** | 0.64 |
| 96 | 400 | Fixed 4096 | **1.87** | 1.30 |
| 96 | 400 | Randomly from [1 .. 4096] | **3.67** | 2.44 |

- For DP 1 TP 16 configuration, the `max-model-len` and `max-num-batched-tokens` are `32768`.

| max-concurrency | #requests | input tokens distribution | SP Throughput | Baseline Throughput |
|---------|---------|---------|---------|---------|
| 1 | 200 | Fixed 4096 | 1.35 | **1.52** |
| 24 | 200 | Fixed 4096 | **2.29** | 2.21 |

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/c07a73317d202c2dad67f12893fcddb6d3664950

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -24,6 +24,12 @@ from vllm_ascend.ops.attention import vanilla_chunked_prefill_mla
 from vllm_ascend.utils import npu_prefetch
 from vllm_ascend.worker.npu_input_batch import InputBatch
 
+import torch.distributed as dist
+from vllm.distributed.parallel_state import get_dp_group
+from vllm_ascend.distributed.parallel_state import get_mla_sp_world_group
+from vllm_ascend.mla_sp_context import get_sp_context
+from vllm_ascend.ops.shard import RowShardLinear
+
 if TYPE_CHECKING:
     from vllm.v1.core.sched.output import SchedulerOutput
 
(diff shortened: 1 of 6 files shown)


vllm_ascend/attention/mla_v1.py
```diff
@@ -1094,6 +1102,61 @@ def _forward_decode(
...
+            dist.all_to_all_single(
+                output=sp_output,
+                input=sp_send,
+                output_split_sizes=sp_context.output_split_sizes,
+                input_split_sizes=sp_context.input_split_sizes,
+                group=sp_world_group.device_group,
+            )
+            sp_output = sp_output.reshape(num_sp_tokens, tp_size * self.num_heads * self.v_head_dim)
+        from vllm_ascend.utils import dispose_tensor
+        dispose_tensor(sp_send)
```
ApsarasX: Don't just use `dispose_tensor` casually. It hurts performance. Only use it where NPU memory spikes happen.

vllm_ascend/mla_sp_context.py
```diff
@@ -0,0 +1,168 @@
...
+    global_tokens: torch.Tensor
+    dp_sp_start_token: list[int]
+    dp_sp_end_token: list[int]
+    rank_sp_start_token: list[int]
+    rank_sp_end_token: list[int]
+    my_dp: int
+    my_rank: int
+    my_rank_sp_start_token_within_dp: int
+    my_rank_sp_end_token_within_dp: int
+    num_my_dp_sp_tokens: int
```
weijinqian0: The variable name is incorrect. 'my' may be replaced by 'local'

vllm_ascend/mla_sp_context.py
```diff
@@ -0,0 +1,168 @@
...
+            receive_len = max(0, receive_end - receive_start)
+            output_split_sizes.append(receive_len)
+
+    forward_context.max_tokens_across_dp = num_tokens_per_dp
+    forward_context.padded_num_tokens = calc_div_ceil(num_tokens_per_dp, tp_size) * tp_size
+    from vllm_ascend.ascend_forward_context import FusedMoEState
+    if forward_context.fused_moe_state == FusedMoEState.NaiveMulticast:
+        forward_context.fused_moe_state = FusedMoEState.AllGather
+    dp_metadata = forward_context.dp_metadata
+    if dp_metadata is not None:
```
weijinqian0: FusedMoEState will be removed this week.

weijinqian0:
The solution may not be sufficiently versatile.

vllm_ascend/attention/mla_v1.py
clrs97: Fixed.

vllm_ascend/attention/mla_v1.py
```diff
@@ -1094,6 +1102,61 @@ def _forward_decode(
...
+                device=sp_send.device
+            )
+            dist.all_to_all_single(
+                output=sp_output,
+                input=sp_send,
+                output_split_sizes=sp_context.output_split_sizes,
+                input_split_sizes=sp_context.input_split_sizes,
+                group=sp_world_group.device_group,
+            )
+            sp_output = sp_output.reshape(num_sp_tokens, tp_size * self.num_heads * self.v_head_dim)
```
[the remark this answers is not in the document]
clrs97: This step is designed based on both dp and tp groups, so it is comfirmed.

vllm_ascend/attention/mla_v1.py
```diff
@@ -1292,6 +1368,9 @@ def forward(
             else:
                 o_proj_input[:num_decode_tokens] = output_decode
 
+        if isinstance(self.o_proj, RowShardLinear):
+            o_proj_input = get_tp_group().all_gather(o_proj_input, dim=-1)
```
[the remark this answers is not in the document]
clrs97: It is to make sure the second dimension is complete to match o_proj.

vllm_ascend/mla_sp_context.py
clrs97: Fixed.

vllm_ascend/models/deepseek_v2.py
```diff
@@ -508,10 +512,16 @@ def __init__(
             bias=False,
             quant_config=quant_config,
             prefix=f"{prefix}.kv_b_proj")
-        if (config.n_routed_experts is not None
-                and self.debug_layer_idx >= config.first_k_dense_replace
-                and self.debug_layer_idx % config.moe_layer_freq == 0
-                and self.enable_shared_expert_dp):
+        if ascend_config.enable_o_shard:
```
weijinqian0: Variable naming needs to specify whether it enables prefill or decode. Currently, it might be confused with the O matrix TP split. Additionally, the location can be referenced in ops/linear.py.

wangxiyuan:
replace by #2961 ?

clrs97:
> replace by #2961 ?

This PR is focus on DP load rebalancing. The current feature name `SP` is confusing, it should be renamed as a new more proper name.

vllm_ascend/models/deepseek_v2.py
clrs97: Changed into `enable_mla_prefill_dp_rebalancing`.

wangxiyuan:
torchair is removed already.

No update for long time. close it now. Feel free to create a new one if it's still necessary

PR #2495, opened 2025-08-22
[5/N][Refactor] Make `AscendAttentionTorchairBackendImpl` extend from `AscendAttentionBackendImpl`

shen-shanshan:

Refactor `AscendAttentionTorchairBackendImpl`, making it extend from `AscendAttentionBackendImpl` and so removing reduntant codes.

Workflow in `forward()`:

1. `_additional_process_before_forward()` --> get `ForwardAdditionalProcessResult`
2. Dispatch forward to 4 kinds of specific `_forward_xxx()`

Thus, we can remove the "huge" `forward()` method in `AscendAttentionTorchairBackendImpl`. All we need to do is costomizing (overwriting) the process logic in `_additional_process_before_forward()` and the 4 kings of `_forward_xxx()`.

- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/8e5cdcda4e5a55ff49d92d37139042dda44b6b3c

### DIFF
--- a/vllm_ascend/torchair/torchair_attention.py
+++ b/vllm_ascend/torchair/torchair_attention.py
@@ -405,7 +405,7 @@ class AscendAttentionTorchairBackendImpl(AscendAttentionBackendImpl):
     def _forward_decode_only(
         self,
         query: torch.Tensor,
-        attn_metadata: "AscendMetadata",
+        attn_metadata,
         output: Optional[torch.Tensor] = None,
         **kwargs,
     ) -> torch.Tensor:


vllm_ascend/attention/attention_v1.py
```diff
@@ -490,76 +553,62 @@ def forward(
...
                                  dtype=query.dtype,
                                  device=query.device)
         ori_output = output
-        if trace_flag:
-            torch.ops.vllm.unified_ascend_attention_with_output(
-                query=query,
-                key=key,
-                value=value,
-                output=output,
-                layer_name=layer.layer_name)
```
weijinqian0, line 508: why remove it?

shen-shanshan: > why remove it?

@weijinqian0 Some check logic (only base class have, and subclass is different) have been moved to https://github.com/vllm-project/vllm-ascend/pull/2495/files?diff=split&w=0#diff-c06b611c1540fb1685713a92ab626d7594a52aca56ef56112b90cfcef30d4bf2R327-R337.

Thus, all the code in the backbone of `forward()` is common logic, and the subclass can remove this `forward()` while just need to overwrite some `xxx_forward()` method.

PR #2496, opened 2025-08-22
[Bugfix] Fix wrong torchair model register key

shen-shanshan:

Fix wrong torchair model register key.


weijinqian0:
approved

wangxiyuan:
please fix the ut failure

PR #2508, opened 2025-08-23
[bugfix] ascend schedule encountered an incorrect req block length in the check_watermark_for_prefill function

liziyu179:
bugfix ascend schedule encountered an incorrect req block length in the check_watermark_for_prefill function

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/426cc8629f7e630e1c5a0b96fe2db737a170a06d

### DIFF
--- a/vllm_ascend/core/scheduler.py
+++ b/vllm_ascend/core/scheduler.py
@@ -465,7 +465,7 @@ class AscendScheduler(Scheduler):
                                    self.block_size)
         req_blocks = self.kv_cache_manager.coordinator.get_blocks(
             request.request_id)
-        num_new_blocks = (num_required_blocks - len(req_blocks) -
+        num_new_blocks = (num_required_blocks - len(req_blocks[0]) -
                           len(computed_blocks))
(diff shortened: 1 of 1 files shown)


vllm_ascend/core/scheduler.py
```diff
@@ -465,7 +465,7 @@ def _check_watermark_for_prefill(self,
                                    self.block_size)
         req_blocks = self.kv_cache_manager.coordinator.get_blocks(
             request.request_id)
-        num_new_blocks = (num_required_blocks - len(req_blocks) -
+        num_new_blocks = (num_required_blocks - len(req_blocks[0]) -
```
MengqingCao, line 468: ~~I think this should be `req_blocks[request.request_id]`?~~

Just ignore this, I'll check if we should use the first kv group

MengqingCao: As we only support one block-size in vllm-ascend now, using the first kv cache group seems good in this pr

MengqingCao (approved):

PR #2511, opened 2025-08-25
[Bugfix]Support Qwen3-MOE on aclgraph mode in sizes capture and add new ut

lilinsiman:
[Bugfix]Support Qwen3-MOE on aclgraph mode in sizes capture and add new ut

This PR solves the problem of sizes capture and stream error caused by using ACLgraph on the Qwen3-30B MOE model.
Add new ut.

no

How was this patch tested?
ut

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/6fad29b11b3680c44782cd6e5fe555779d620d6c

### DIFF
--- a/tests/e2e/multicard/test_qwen3_moe.py
+++ b/tests/e2e/multicard/test_qwen3_moe.py
@@ -21,6 +21,8 @@
 Run `pytest tests/e2e/multicard/test_qwen3_moe.py`.
 """
 
+import os
+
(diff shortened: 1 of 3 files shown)


wangxiyuan:
e2e test failure doesn't relate to this change. I'm working on fixing it.

wangxiyuan (approved):

PR #2515, opened 2025-08-25
[CustomOp] Register VocabParallelEmbedding instead of overwrite forward

wxsIcey:
Register VocabParallelEmbedding instead of overwrite forward

N/A

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/644d57d53191b94d9e50a4765891c498790d924b


wangxiyuan:
similar like this one https://github.com/vllm-project/vllm-ascend/pull/2479  maybe the quantization patch should be updated as well?

wxsIcey:
done

wxsIcey:
> similar like this one #2479 maybe the quantization patch should be updated as well?

done

PR #2519, opened 2025-08-25
[Perf] Remove unnecessary dispose_tensor

ApsarasX:
In #966, I added the `dispose_tensor` function to reduce NPU memory spikes, but it has a slight impact on performance. So I only used it in the MOE w8a8_dynamic part. However, over the past few months, this function has been overused. So this PR gets rid of the unnecessary uses.

No.

Existing test cases.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/d0a4a3f645dd57364a0562e01ee8a04e4afa63ba

### DIFF
--- a/vllm_ascend/models/deepseek_dbo.py
+++ b/vllm_ascend/models/deepseek_dbo.py
@@ -76,7 +76,6 @@ from vllm_ascend.multistream.metadata import (MultiStreamConfig,
                                               MultiStreamStepMetadata,
                                               make_multistream_metadata_ds)
 from vllm_ascend.ops.fused_moe import AscendFusedMoE
-from vllm_ascend.utils import dispose_tensor
 
 VLLM_ASCEND_ENABLE_DBO: bool = envs_ascend.VLLM_ASCEND_ENABLE_DBO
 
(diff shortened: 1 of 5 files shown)


vllm_ascend/models/deepseek_dbo.py
```diff
@@ -76,7 +76,6 @@
                                               MultiStreamStepMetadata,
                                               make_multistream_metadata_ds)
 from vllm_ascend.ops.fused_moe import AscendFusedMoE
-from vllm_ascend.utils import dispose_tensor
```
wangxiyuan, line 79: this function can be removed from utils as well?

ApsarasX: Not for now. It hasn't been fully decided which `dispose_tensor` calls will be removed. I'll use performance data to justify the necessity of each removal. If it turns out that some `dispose_tensor` calls are still needed, then we can't remove the function from utils.

wangxiyuan:
feel free to merge this once it's fully tested.

wangxiyuan:
@ApsarasX

PR #2524, opened 2025-08-25
[Structured Output] Replace `apply_grammar_bitmask()` method with that in vllm to avoid maintenance

shen-shanshan:

Replace `apply_grammar_bitmask()` method with that in vllm to avoid maintenance.

> [!NOTE]
> **Do not merge** this before https://github.com/vllm-project/vllm/pull/21999.


shen-shanshan:
CC @wangxiyuan

PR #2527, opened 2025-08-25
[CI] Upgrade vllm in accuracy and performance CI

MengqingCao:
Upgrade vllm in accuracy and performance CI

CI passed with existing test.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/5c4b6e66fec0a59f702e32ec5c476cdd8a800566


Yikun:
Let's merge this to recover accuracy test and track perf test failure on https://github.com/vllm-project/vllm-ascend/issues/2533

PR #2529, opened 2025-08-25
[main][QuickFix] change `free_encoder_input_ids` into `free_encoder_mm_hashes`

zhoux77899:

Change `free_encoder_input_ids` into `free_encoder_mm_hashes` due to https://github.com/vllm-project/vllm/pull/22711

### DIFF
--- a/tests/ut/core/test_scheduler.py
+++ b/tests/ut/core/test_scheduler.py
@@ -309,7 +309,7 @@ class TestAscendScheduler(TestBase):
                                            },
                                            num_common_prefix_blocks=0,
                                            finished_req_ids=set(),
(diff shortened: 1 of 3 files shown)


tests/ut/core/test_scheduler.py
```diff
@@ -309,7 +309,7 @@ def test_stop_via_update_from_output(self):
                                            },
```
wangxiyuan: you should use `vllm_version_is` to make the backward compatibility. Please take look at other related code for examle.

wangxiyuan (approved):

wangxiyuan:
this need more change, see: https://github.com/vllm-project/vllm-ascend/pull/2530

PR #2532, opened 2025-08-25
[bugfix] fix torchair runtime error caused by configuration mismtaches and file missing

linfeng-yuan:
This PR ports #2312 #2506 #2531 to main branch.

Original implementation of torchair caching forces users to make everything prepared, fix all the configuration and enable `use_cached_npu_graph`, and it might cause some problems confusing to understand and tackle for users. It is better to compile the graph twice instead of reusing the old kvcaches and cached torchair graph. And the extra duration time is acceptable. Additionally, this pr fixes a recompilation problem of torchair graph mode caused by `running_in_graph` variable in `AscendMLATorchairImpl`.

If users want to enabling torchair.cache_compile with high compilation speed, it is recommended to enable both `use_cached_kv_cache_bytes` and `use_cached_graph` in `torchair_graph_config`. Without `use_cached_kv_cache_bytes`, we'll compile torchair computation graph twice to avoid runtime error caused by configuration mismtaches (the second compilation will be much faster). Additionally, we've made a change to how the TORCHAIR_CACHE_HOME enviroment variable is utilized to enhance safety and prevent accidental file deletion by adding a suffix directory.

CI and e2e vllm serving pass.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/70549c1245c3eeb3706e3c09a9e18d702fbf705f

### DIFF
--- a/vllm_ascend/torchair/torchair_mla.py
+++ b/vllm_ascend/torchair/torchair_mla.py
@@ -628,6 +628,7 @@ class AscendMLATorchairImpl(MLAAttentionImpl):
         self.torchair_graph_enabled = ascend_config.torchair_graph_config.enabled
         self.enable_kv_nz = ascend_config.torchair_graph_config.enable_kv_nz
         self.enable_shared_expert_dp = ascend_config.enable_shared_expert_dp
+        self.running_in_graph = False
 
         # Adapt torch air graph mode with spec decoding.
         speculative_config = get_current_vllm_config().speculative_config


vllm_ascend/torchair/torchair_worker.py
```diff
@@ -33,7 +34,9 @@ def determine_available_memory(self) -> int:
 
         available_kv_cache_memory = super().determine_available_memory()
 
-        if check_torchair_cache_exist() and check_kv_cache_bytes_cache_exist():
+        if get_ascend_config(
```
ApsarasX, line 36: The logic here is a bit complex. Can we refactor it—maybe use an Enum to represent the torchair cache state, similar to how `FusedMOEState` works?

vllm_ascend/platform.py
```diff
@@ -168,6 +170,14 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
             compilation_config.level = CompilationLevel.NO_COMPILATION
             compilation_config.cudagraph_mode = CUDAGraphMode.NONE
+            # Note: We delete the torchair cache folder here to prevent runtime issues caused by dimension
+            # mismatches or configuration inconsistencies when users reuse cached computation graphs. Though
+            # this will increase graph compilation duration, it significantly enhances robustness and decreases
+            # graph launching time during inference. In order to decrease torchair graph compilation time, users
+            # can enable both `use_cached_graph` and `use_cached_kv_cache_bytes` in torchair_graph_config.
+            if check_torchair_cache_exist(
+            ) and not ascend_config.torchair_graph_config.use_cached_kv_cache_bytes:
+                delete_torchair_cache_file()
```
ApsarasX, line 186: Add a `logger.warning` log?

linfeng-yuan: Done.

vllm_ascend/torchair/torchair_worker.py
linfeng-yuan: > The logic here is a bit complex. Can we refactor it—maybe use an Enum to represent the torchair cache state, similar to how `FusedMOEState` works?

It is not necessary to add an enum class for torchair cache state. And I find it equivalent to only check kvcache_bytes here.

wangxiyuan (approved):

PR #2560, opened 2025-08-26
[Bugfix][LoRA][Patch] Fix the LoRA inference bug after upstream vLLM codebase changed

paulyu12:
The mergence of the upstream https://github.com/vllm-project/vllm/pull/22592 caused a vllm-ascend LoRA inference bug. The details are following:

According to [torch_npu/npu/_stream_check.py](https://github.com/Ascend/pytorch/blob/863b9071cbdf47023c12c246e3efa9c6e2285fc6/torch_npu/npu/_stream_check.py#L74), NPU device type tensors have attributes is_cuda=True and is_npu=True. This causes that vLLM's apply_repetition_penalties function will run into the branch of "if logits.is_cuda and logits.is_contiguous()" and call the custom op implemented in CUDA, which is not compatible with NPU.

No.

pytest -sv tests/e2e/singlecard/test_ilama_lora.py
pytest -sv tests/e2e/multicard/test_ilama_lora_tp2.py

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/fe8d7b6f03e7d8a36ffb6931397fc81ee594dd64

### DIFF
--- a/.github/workflows/vllm_ascend_test.yaml
+++ b/.github/workflows/vllm_ascend_test.yaml
@@ -192,7 +192,7 @@ jobs:
           VLLM_USE_MODELSCOPE: True
         run: |
           pytest -sv tests/e2e/singlecard/test_offline_inference.py
-          # pytest -sv tests/e2e/singlecard/test_ilama_lora.py
+          pytest -sv tests/e2e/singlecard/test_ilama_lora.py
           pytest -sv tests/e2e/singlecard/test_guided_decoding.py
           pytest -sv tests/e2e/singlecard/test_camem.py
           pytest -sv tests/e2e/singlecard/test_embedding.py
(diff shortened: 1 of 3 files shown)


.github/workflows/vllm_ascend_test.yaml
```diff
@@ -284,7 +283,6 @@ jobs:
           pytest -sv tests/e2e/multicard/test_offline_inference_distributed.py::test_models_distributed_DeepSeek_W4A8DYNAMIC
           pytest -sv tests/e2e/multicard/test_offline_inference_distributed.py::test_sp_for_qwen3_moe
           pytest -sv tests/e2e/multicard/test_data_parallel.py
-          pytest -sv tests/e2e/multicard/ --ignore=tests/e2e/multicard/test_ilama_lora_tp2.py \
```
wangxiyuan: this should be kept

MengqingCao: LGTM if the above comment is addressed

paulyu12: Done.
For a quick fix, I have commented both of the e2e tests about LoRA. We can merge it first.
LoRA offline inference test still has an accuracy issue, and I'm working on it.

wangxiyuan (approved):

leo-pony:
@paulyu12 Bug reproduced in vllm v0.11.0 + vllm-ascend v0.11.0rc1. The code in this PR is not in vllm-ascend v0.11.0rc1? Could your help check ?

paulyu12:
Ok, I'll check it.

PR #2573, opened 2025-08-27
[CI] Add e2e ci test for A3

zhangxinyuehfad:
Add e2e ci test for A3

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/11a7fafaa8807bfeea4b60466c576ec6a7031bfd

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -11,3 +11,7 @@ self-hosted-runner:
     - linux-aarch64-310p-2
     - linux-aarch64-310p-4
     - ubuntu-24.04-arm
+    - linux-aarch64-a3-1
+    - linux-aarch64-a3-2
+    - linux-aarch64-a3-4
+    - linux-aarch64-a3-8
--- a/.github/workflows/vllm_ascend_dist.yaml
+++ b/.github/workflows/vllm_ascend_dist.yaml
@@ -1,0 +1,127 @@
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+#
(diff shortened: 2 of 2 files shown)


.github/workflows/vllm_ascend_dist.yaml
```diff
@@ -0,0 +1,128 @@
...
+      matrix:
+        os: [linux-aarch64-a3-8]
+        vllm_version: [v0.10.1.1, main]
+    name: vLLM Ascend test
+    runs-on: ${{ matrix.os }}
+    container:
+      image: m.daocloud.io/quay.io/ascend/cann:8.2.rc1-a3-ubuntu22.04-py3.11
+      env:
+        DEBIAN_FRONTEND: noninteractive
+        COMPILE_CUSTOM_KERNELS: 1
```
MengqingCao: This is defautly set to 1 already

zhangxinyuehfad: deleted it

.github/actionlint.yaml
```diff
@@ -11,3 +11,7 @@ self-hosted-runner:
     - linux-aarch64-310p-2
     - linux-aarch64-310p-4
     - ubuntu-24.04-arm
+    - linux-aarch64-a3-1
+    - linux-aarch64-a3-2
```
MengqingCao, line 15: I think these two runner could also add to the e2e test, but we can do it in next pr, WDYT? @wangxiyuan @Yikun

MengqingCao:
Let's merge this first to unblock CI for more ST cases

PR #2577, opened 2025-08-27
[Bugfix] fix aclgraph not enabled by default

MengqingCao:
As vllm will set `cudagraph_mode` to `NONE` before `check_and_update_config` in post init of `VllmConfig` (https://github.com/vllm-project/vllm/blob/5da4f5d857933329aaca779e3a81f1385c84e34a/vllm/config/__init__.py#L3630), we always have `cudagraph_mode` isn't `None`, thus we must remove this check and add it when the related adaption in vllm is done.

Test pass with the new added e2e test

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/5da4f5d857933329aaca779e3a81f1385c84e34a

### DIFF
--- a/tests/e2e/singlecard/test_aclgraph.py
+++ b/tests/e2e/singlecard/test_aclgraph.py
@@ -33,23 +33,30 @@ MODELS = [
     # "vllm-ascend/Qwen3-30B-A3B-Puring"
 ]
 
+PROMPTS = [
+    "Hello, my name is", "The president of the United States is",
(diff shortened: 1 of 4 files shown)


MengqingCao:
@yiz-liu @wangxiyuan plz review this and hope this could be merge quickly to fix main

MengqingCao:
closing as done in #2590

PR #2578, opened 2025-08-27
[feat] Adapting Qwen3-32B to Eagle3 mode to resolve head dimension (headdim) mismatch issues

coder-fny:

The adapted weights from https://www.modelscope.cn/models/AngelSlim/Qwen3-32B_eagle3/ can perform Eagle3 inference, but the head dimension (80) of these weights is inconsistent with that of the target model (128), causing an error; therefore, adaptation is required.


ApsarasX:
@CSlearnerZM FYI

vllm_ascend/attention/attention_mask.py
```diff
@@ -81,7 +81,7 @@ def get_splitfuse_attn_mask(
         mask_scale_factor = AttentionMaskBuilder.get_mask_scale_factor(dtype)
         attn_mask = torch.index_select(self.attn_mask_cache,
                                        dim=0,
-                                       index=position)[:, :max_seq_len]
+                                       index=position.cpu())[:, :max_seq_len]
```
momo609, line 84: This operation may cause a device mismatch between the two inputs. Please add a validation check for this.

vllm_ascend/models/qwen3.py
```diff
@@ -154,3 +154,10 @@ def load_weights(self, weights: Iterable[tuple[str,
                            if self.config.tie_word_embeddings else None),
```
wangxiyuan, line 154: can we find another way that won't modify the model file? we'll remove all the model file from vllm-ascend later

wangxiyuan:

PR #2582, opened 2025-08-27
[Fix] Fix DP-related padding logic

yiz-liu:
The determination of attention state, padding, and other forward metadata has been moved to an earlier stage within the input preparation process. This change enables us to utilize a single all-reduce operation, maximizing synchronization efficiency as early as possible.

The logic for synchronizing metadata—such as the number of tokens, prefill status, and DBO status—across data parallel (DP) ranks has now been unified and simplified.

For performance improvements, the all-reduce operation has been switched from the `gloo` backend to the `npu` backend, which results in an reduction of several milliseconds per step (**approximately 10% performance gain for TPOT!**).

Additionally, the multi-DP server hang issue has been resolved, ensuring no more hangs occur when `num_requests < dp_size`. This is a significant improvement!

Finally, the miscalculated memory usage issue has been addressed by removing the unnecessary `DummyCommImpl`, allowing the system to use the real communication method when determining available memory.

None.

Maybe we should add an test case for multi-DP online server? @MengqingCao


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -595,32 +590,43 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
...
+        global_with_prefill = bool(synced_flags[0])
+        global_enable_dbo = not bool(synced_flags[1])
+
+        # Create a tensor for num_tokens_after_padding
+        num_tokens_after_padding = torch.tensor([max_tokens_across_dp] *
+                                                self.dp_size,
+                                                device="npu",
+                                                dtype=torch.int32)
+
+        return max_tokens_across_dp, num_tokens_after_padding, global_with_prefill, global_enable_dbo
```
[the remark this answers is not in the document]
yiz-liu: The purpose of this change in the code is to enhance clarity and extensibility. This performance-related suggestion has been duly noted, and if there are any other viewpoints, I will reconsider the approach accordingly.

Angazenn:
I have test this PR with Qwen3-30B on DP32 with 2 A3 nodes. The server can generate reponse to requests correctlly. Changing cpu all_reduce to device all_reduce also reduce  ~10ms in this scenario.

vllm_ascend/distributed/moe_comm_method.py
```diff
@@ -94,41 +94,41 @@ def unpermute(self, mlp_output: torch.Tensor,
...
-                                    dtype=torch.int64,
-                                    device=hidden_states.device)
-        group_list_type = 0
-        return permuted_hidden_states, expert_tokens, group_list_type
-
-    def unpermute(self, mlp_output: torch.Tensor,
-                  hidden_states: torch.Tensor) -> None:
-        """Dummy implementation that does nothing."""
-        pass
+# class DummyCommImpl(MoECommMethod):
```
wangxiyuan: please remove it

yiz-liu: Yeah right, my mistake, working on it.

yiz-liu: Removed.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -593,32 +588,43 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
...
-        num_tokens_across_dp = num_tokens_across_dp[:-2]
-        return num_tokens_across_dp, with_prefill, enable_dbo
-
-    def _get_forward_metadata_across_dp_and_pad(
+    def _sync_metadata_across_dp(
             self, num_tokens: int, with_prefill: bool, enable_dbo: bool
     ) -> tuple[int, Optional[torch.Tensor], bool, bool]:
-        if self.dp_size == 1:
+        if self.dp_size == 1 or self.vllm_config.model_config.enforce_eager:
             return num_tokens, None, with_prefill, enable_dbo
```
JC-ut0, line 595: when `self.vllm_config.model_config.enforce_eager` is enabled, `dummy_run` would always return `with_pretill=Fasle`, which changes the original logic, and will cause `_get_fused_moe_state` returning a different value from the original one.

PR #2583, opened 2025-08-28
test support v0.10.1

Yikun:

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/321938e9ac4000e0cb37e328359a7fd3026bc672


wangxiyuan:
Thanks for the test, it works as expect

PR #2584, opened 2025-08-28
Support v0.10.1

Yikun:
This patch also supports v0.10.1

No

- CI passed
- test 0.10.1: https://github.com/vllm-project/vllm-ascend/pull/2583
- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/321938e9ac4000e0cb37e328359a7fd3026bc672

### DIFF
--- a/tests/ut/core/test_scheduler.py
+++ b/tests/ut/core/test_scheduler.py
@@ -21,7 +21,7 @@ from tests.ut.base import TestBase
 from vllm_ascend.core.scheduler import AscendScheduler
 from vllm_ascend.utils import vllm_version_is
 
-if not vllm_version_is("0.10.1.1"):
+if not (vllm_version_is("0.10.1.1") or vllm_version_is("0.10.1")):
     from vllm.v1.outputs import DraftTokenIds
 else:
     DraftTokenIds = None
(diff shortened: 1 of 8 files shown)


tests/ut/core/test_scheduler.py
```diff
@@ -21,7 +21,7 @@
 from vllm_ascend.core.scheduler import AscendScheduler
 from vllm_ascend.utils import vllm_version_is
 
-if not vllm_version_is("0.10.1.1"):
+if not (vllm_version_is("0.10.1.1") or vllm_version_is("0.10.1")):
```
ApsarasX, line 24: Maybe we can think of both 0.10.1 and 0.10.1.1 as like 0.10.1.x?

Yikun: Yes, this change mainly help vllm-ascend work with vLLM 0.10.1 and 0.10.1.1.

And avoid apply them to 0.10.1.devxxx, so I use the strict match.

Yikun:
Just add for version, so we can ignore `codecov/patch`.

ApsarasX (approved):

PR #2025, opened 2025-07-26
[Refactor] Pre-transpose MoE weights for improved performance

yiz-liu:
### What this PR does / why we need it?
Pre-transpose MoE weights for improved performance

### Does this PR introduce _any_ user-facing change?
None.

### How was this patch tested?
No further test needed.

PR #2614, opened 2025-08-28
[3/N][Feat][Graph] Support `all-to-all` and quantized models with ACL Graph

yiz-liu:
* **Unify execution paths:** Consolidates the quantized and non-quantized execution paths into a single `fused_experts` function, removing duplicated logic and making the control flow clearer and easier to maintain.
* **W8A8 dynamic quantization:** Adds support for W8A8 dynamic quantization inside the unified MoE kernel. Communication routines are updated to correctly handle dynamic quantization scales for activations.
* **Weight pre-processing:** Prae-transpose the `w13` and `w2` weight matrices (as implemented in PR #2025) so that quantized and non-quantized models follow the same code path for the MoE gating, up-projection, and down-projection operations.
* **All-to-all communication:** Adds an `all-to-all` collective communication pattern. For large token counts on modern hardware, `all-to-all` is more efficient than the previous `all-gather` strategy. However, `all-to-all` is not really captured and replayed due to multiple D2H operations which will trigger synchronization, and thus raise error when capture graphs. We only use `all-to-all` when fallback to `compiled_graph_for_general_shape`.
* **Dynamic communication selection:** The model runner now selects the optimal MoE communication method (`mc2`, `allgather`, or `alltoall`) at runtime based on token count and the Ascend SoC version.
* **Limitation:** `all-gather` is not yet supported for quantized models, which means there is still something left to do on A2.

None.

No further test cases needed.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/d660c98c1b59580af97d6c7dd162c7c8894d40ed


vllm_ascend/ops/common_fused_moe.py
```diff
@@ -224,12 +287,19 @@ def __init__(
             has_bias,
         )
 
+        with_quant = quant_config is not None
+        setup_token_dispatchers(self.moe_config.ep_size,
+                                top_k=self.top_k,
+                                num_experts=self.global_num_experts,
+                                num_local_experts=self.local_num_experts,
+                                with_quant=with_quant)
```
Pr0Wh1teGivee: with_quant will not be passed in __init__ phase.

vllm_ascend/distributed/moe_comm_method.py
```diff
@@ -457,3 +461,91 @@ def unpermute(self, mlp_output: torch.Tensor,
...
+        topk_weights: torch.Tensor,
+        expert_map: torch.Tensor,
+        num_experts: int,
+        use_a8: bool,
+    ) -> tuple[torch.Tensor, torch.Tensor, Optional[torch.Tensor], int]:
+        results = self.token_dispatcher.token_dispatch(hidden_states,
+                                                       topk_weights,
+                                                       topk_ids,
+                                                       None,
+                                                       log2phy=None)
```
Pr0Wh1teGivee: add with_quant here

wangxiyuan (approved):

vllm_ascend/distributed/moe_comm_method.py
```diff
@@ -457,3 +464,93 @@ def unpermute(self, mlp_output: torch.Tensor,
...
+            dist.all_gather(list(self.split_hidden_states), hidden_states,
+                            self.moe_config.tp_group.device_group)
+            hidden_states = torch.cat(self.split_hidden_states, dim=0)
+
+        if self.num_tokens < hidden_states.shape[0]:
+            hidden_states = hidden_states[:self.num_tokens]
+
+        return hidden_states
+
+    def permute(
```
momo609, line 535: How is the CPU logic in alltoall kept within the computation graph here? I only see the registration of fake functions, but no explicit mechanism to detach the graph.

PR #2617, opened 2025-08-28
 support nz embedding

WithHades:
support nz embedding

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/b5ee1e3261d9edf94d76ba8b437ebdef7ac599ea

### DIFF
--- a/vllm_ascend/ops/vocab_parallel_embedding.py
+++ b/vllm_ascend/ops/vocab_parallel_embedding.py
@@ -18,8 +18,11 @@
 from typing import Optional, Tuple
 
 import torch
+import torch_npu
 from torch import nn
 from torch.nn.parameter import Parameter
+from vllm.config import get_current_vllm_config
+from vllm.config.compilation import CUDAGraphMode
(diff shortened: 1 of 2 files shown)


vllm_ascend/ops/vocab_parallel_embedding.py
```diff
@@ -31,7 +34,20 @@
...
+
+
+class AscendUnquantizedEmbeddingMethod(UnquantizedEmbeddingMethod):
+
+    def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
+        super().process_weights_after_loading(layer)
+
+        cudagraph_mode = get_current_vllm_config(
+        ).compilation_config.cudagraph_mode
+        if torch.version.cann.startswith("8.3") and (cudagraph_mode
```
momo609, line 47: Why was the CUDA Graph check added here, NZ can be enabled in both ACLGraph and eager modes?

weijinqian0:
What is the optimization benefit?

PR #2627, opened 2025-08-29
[main] [bugfix] Fix misjudging quantized/unquantized scenarios

Pr0Wh1teGivee:
In a mixed-precision scenario, quant_config is not None, but MoE needs to perform unquantized computation; however, quantized computation is currently being used. Therefore, we put the with_quant logic into forward, avoid misjudging in mix-precision scenarios.
no
e2e & ut

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/98ac0cb32d9462e50bd998f9f2eb6e4c09232c95


Pr0Wh1teGivee:
Error from e2e test is not introduced by this pr.

wangxiyuan (approved):

Issue #676, opened 2025-04-27
[Installation]: 安装 vllm-ascend v0.8.4rc1 编译失败


rfy48:
### Your current environment

```text
The output of `python collect_env.py`
```
vllm                                     0.8.4+empty
torch                                   2.5.1+cpu
torch-npu                           2.5.1.dev20250320

### How you are installing vllm and vllm-ascend

```sh
# Install vLLM Ascend
git clone  --depth 1 --branch v0.8.4rc1 https://github.com/vllm-project/vllm-ascend.git
cd vllm-ascend
python setup.py develop
cd ..
```

gmake[3]: *** [CMakeFiles/vllm_ascend_C.dir/build.make:90: CMakeFiles/vllm_ascend_C.dir/csrc/torch_binding.cpp.o] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:337: CMakeFiles/vllm_ascend_C.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:344: CMakeFiles/vllm_ascend_C.dir/rule] Error 2
gmake: *** [Makefile:286: vllm_ascend_C] Error 2
Traceback (most recent call last):
  File "/mnt/deepseek/renfeiyang/vllm-test-250427/vllm-ascend-250427/setup.py", line 331, in <module>
    setup(
  File "/usr/lib/python3.11/site-packages/setuptools/__init__.py", line 107, in setup
    return distutils.core.setup(**attrs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/setuptools/_distutils/core.py", line 185, in setup
    return run_commands(dist)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/setuptools/_distutils/core.py", line 201, in run_commands
    dist.run_commands()
  File "/usr/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 969, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3.11/site-packages/setuptools/dist.py", line 1234, in run_command
    super().run_command(command)
  File "/usr/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 988, in run_command
    cmd_obj.run()
  File "/usr/lib/python3.11/site-packages/setuptools/command/develop.py", line 34, in run
    self.install_for_development()
[...]

PR #2631, opened 2025-08-29
[Image] Upgrade openEuler to 24.03

wxsIcey:
Upgrade openEuler to 24.03

N/A

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/4071c76cf3cff46e14630f1e66cbf006b6eb51d3

### DIFF
--- a/Dockerfile.310p.openEuler
+++ b/Dockerfile.310p.openEuler
@@ -15,7 +15,7 @@
 # This file is a part of the vllm-ascend project.
 #
 
-FROM quay.io/ascend/cann:8.2.rc1-310p-openeuler22.03-py3.11
+FROM quay.io/ascend/cann:8.2.rc1-310p-openeuler24.03-py3.11
 
 ARG PIP_INDEX_URL="https://mirrors.tuna.tsinghua.edu.cn/pypi/web/simple"
 ARG COMPILE_CUSTOM_KERNELS=1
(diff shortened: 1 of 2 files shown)


Dockerfile.openEuler
```diff
@@ -47,6 +47,7 @@ RUN export PIP_EXTRA_INDEX_URL=https://mirrors.huaweicloud.com/ascend/repos/pypi
     source /usr/local/Ascend/ascend-toolkit/set_env.sh && \
     source /usr/local/Ascend/nnal/atb/set_env.sh && \
     export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/Ascend/ascend-toolkit/latest/`uname -i`-linux/devlib && \
+    export CPLUS_INCLUDE_PATH=$CPLUS_INCLUDE_PATH:/usr/include/c++/12:/usr/include/c++/12/`uname -i`-openEuler-linux && \
```
Yikun, line 50: This seems just a workaround to specify c++ version manually, just curious Why there are two gcc version here?

wxsIcey: **error action:** https://github.com/vllm-project/vllm-ascend/actions/runs/17316412493/job/49160001312
**root cause:** fatal error: 'cstdint' file not found
**problem analysis:** Taking aarch64 as an example, the c++ header file directory of openEuler22.03 version is in aarch64-linux-gnu (Ubuntu is also in this directory), but 24.03 is changed to aarch64-openEuler-linux, which causes the header file not to be found. Specify additional C++ header file search directories CPLUS_INCLUDE_PATH can solve this problem.

openEuler24.03 only has gcc12, there are no two gcc versions.
related to: https://github.com/vllm-project/vllm-ascend/issues/676

wangxiyuan:
please update all the openEuler dockerfile. Thanks.

wxsIcey:
> please update all the openEuler dockerfile. Thanks.

done

PR #2659, opened 2025-08-30
[v0.9.1][bugfix] disable the chunked prefill feature in Non-MLA LLMs

rjg-lyh:
This PR enforces the forcible disabling of the chunked prefill feature in Non-MLA models, as the performance of operators supporting this functionality is currently suboptimal. 
At the same time, in engine v1 mode, the ascend scheduler is forcibly enabled, and the `enable_chunked_prefill` specified by the user in additional_config is disabled.

No.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -86,8 +86,8 @@ templates_path = ['_templates']
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-locale_dirs = ['locale/']   
-gettext_compact = False   
+locale_dirs = ['locale/']
+gettext_compact = False
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
--- a/docs/source/tutorials/large_scale_ep.md
+++ b/docs/source/tutorials/large_scale_ep.md
@@ -345,7 +345,7 @@ for process in processes:
 
 :::::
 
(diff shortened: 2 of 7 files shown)


Yikun:
Current 0.9.1-dev lastest commit: https://github.com/vllm-project/vllm-ascend/commit/5926225f9911c89a9c919c9cbd8d97241408dda2
---
Case 1 git revert 9dc23b6583a40ccdc83a2cc4043e0d5e84ec2624, test
Case 2 apply #2659, test (green part latency +0.5 ms)

image

There are still stable 0.5 ms perf regression compare with your PR and revert #2326 

```
# Start vLLM V1
export MODEL=Qwen/Qwen3-8B
VLLM_USE_V1=1 VLLM_USE_MODELSCOPE=true python3 -m vllm.entrypoints.openai.api_server --model $MODEL \
         --tensor-parallel-size 1 --swap-space 16 --disable-log-stats \
         --disable-log-requests  --load-format dummy

# Benchmark
docker exec -it  yikun-091 bash
export MODEL=Qwen/Qwen3-8B
export VLLM_USE_MODELSCOPE=true
pip config set global.index-url https://mirrors.tuna.tsinghua.edu.cn/pypi/web/simple
pip install -r /vllm-workspace/vllm-ascend/benchmarks/requirements-bench.txt
python3 /vllm-workspace/vllm/benchmarks/benchmark_serving.py --model $MODEL --dataset-name random \
         --random-input-len 200 --num-prompts 200 --request-rate 1 \
         --save-result --result-dir ./
```

PR #2660, opened 2025-08-30
[Bugfix] Keep prefill node using prefill code path only in disaggregation mode

jianzs:

When running in prefill-decode disaggregation mode, we're not completely confident that the prefill node can handle decode phase requests properly. To ensure reliability, this PR continues to use the prefill code path in the prefill node at all times.

No

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/e9b92dcd89e8d05f162a8fdaa3d5d60012615514

### DIFF
--- a/tests/ut/attention/test_mla_v1.py
+++ b/tests/ut/attention/test_mla_v1.py
@@ -205,6 +205,7 @@ class TestAscendMLAMetadataBuilder(TestBase):
         mock_vllm_config.cache_config.block_size = 16
         mock_vllm_config.scheduler_config.max_num_seqs = 4
         mock_vllm_config.scheduler_config.chunked_prefill_enabled = False
+        mock_vllm_config.kv_transfer_config = None
         mock_device = 'cpu'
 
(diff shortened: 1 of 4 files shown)


wangxiyuan (approved):

zzzzwwjj:
I think this restriction is only necessary in the torchair scenario. How about adding it to the torchair folder?

jianzs:
> I think this restriction is only necessary in the torchair scenario. How about adding it to the torchair folder?

Actually, this isn't specific to graph mode. We had a case where a request with 129 tokens (with 128 tokens hitting prefix cache) entered the MC2 code path and ended up crashing.

zzzzwwjj:
> > I think this restriction is only necessary in the torchair scenario. How about adding it to the torchair folder?
> 
> Actually, this isn't specific to graph mode. We had a case where a request with 129 tokens (with 128 tokens hitting prefix cache) entered the MC2 code path and ended up crashing.

This sounds like a bug😂, this case shouldn't cause error theoretically, we can try to solve it.

PR #2678, opened 2025-09-01
Register AscendRowParallelLinear instead of overwrite forward_oot

wxsIcey:
Register AscendRowParallelLinear instead of overwrite forward_oot

N/A


zzhx1:
I will rewrite the code inside linear to increase its versatility and avoid overwrite forward_oot of AscendRowParallelLinear.
https://github.com/vllm-project/vllm-ascend/pull/2167/files#diff-8d0ab7de90fcda34624c34fdbe26a2106be218a33c6be6dc6750321e2e105fc0

PR #2680, opened 2025-09-01
[1/N][Refactor][Quantization] remove redundant quantizer class

22dimensions:


wangxiyuan:
please fix the UT, otherwise LGTM

22dimensions:
> please fix the UT, otherwise LGTM

ut is fixed, since quantization patch will be removed by next pr, so these three function apply_quantization_patch, apply_patch, parse_path is not covered by ut.

PR #2683, opened 2025-09-01
[7/N][refactor]fix torchair rope ops

hust17yixuan:
Due to the registration mechanism, torchair ops can not take effect， so have to patch the Ascend ops to adapt torchair

No

vLLM version: main
vLLM main: https://github.com/vllm-project/vllm/commit/7ea22e42d5f666a26b3ce4117724dadfdb4d3887

- vLLM version: main
- vLLM main: https://github.com/vllm-project/vllm/commit/7ea22e42d5f666a26b3ce4117724dadfdb4d3887

### DIFF
--- a/vllm_ascend/torchair/utils.py
+++ b/vllm_ascend/torchair/utils.py
@@ -185,15 +185,14 @@ def torchair_quant_method_register():
 
 
 def torchair_ops_patch():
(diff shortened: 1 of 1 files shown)


wangxiyuan:
rewrite custom ops in torchair will lead much more work which is not suggested. So i'm fine with this change.

PR #2685, opened 2025-09-01
[DP] External dp server starter

whx-sjtu:
This PR re-implements external-dp starter based on vllm's support for external dp.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/f38035c123b32f239f746585e197e7250694a1ca

### DIFF
--- a/examples/external_online_dp/run_dp_template.sh
+++ b/examples/external_online_dp/run_dp_template.sh
@@ -10,8 +10,6 @@ export PYTORCH_NPU_ALLOC_CONF=expandable_segments:True
 export HCCL_DETERMINISTIC=True
 export HCCL_BUFFSIZE=1024
 export TASK_QUEUE_ENABLE=1
-# Spawn the process inside the vllm maybe cause the circular import issue, using fork here is necessary
-export VLLM_WORKER_MULTIPROC_METHOD="fork"
 
 export VLLM_USE_V1=1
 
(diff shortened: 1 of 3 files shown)


examples/external_online_dp/run_dp_template.sh
```diff
@@ -0,0 +1,48 @@
...
+export HCCL_SOCKET_IFNAME=your_socket_ifname_here
+export DISAGGREGATED_PREFILL_RANK_TABLE_PATH=your_rank_table_path_here
+export VLLM_LOGGING_LEVEL="info"
+export OMP_PROC_BIND=false
+export OMP_NUM_THREADS=10
+export PYTORCH_NPU_ALLOC_CONF=expandable_segments:True
+export HCCL_DETERMINISTIC=True
+export HCCL_BUFFER_SIZE=1024
+export TASK_QUEUE_ENABLE=1
+# Spawn the process inside the vllm maybe cause the circular import issue, using fork here is necessary
```
wangxiyuan: what circular import issue will raise?

wangxiyuan (commented):
Is there any guide how to use the example?

examples/external_online_dp/run_dp_template.sh
whx-sjtu: Ops, this might be an old problem in 091, I have tested and found that there is no such problems in main, so I will remove this.

whx-sjtu: Later I will add a readme as usage guidance.

wangxiyuan (approved):

PR #2703, opened 2025-09-02
Add mtp multi cards ST

JC-ut0:

Add mtp multi cards ST


MengqingCao:
Hi @JC-ut0 is there any update on this pr? could you check the failed case locally?

JC-ut0:
reopen

PR #2704, opened 2025-09-02
[BugFix][MLA] Fix attn_mask bug for ring mla

whx-sjtu:
This PR fix a bug related to attention mask used in ring mla. Current ring mla has supported compressed mask, so we can directly use a 512 * 512 attention mask.


vllm_ascend/attention/mla_v1.py
```diff
@@ -483,10 +483,16 @@ def __init__(
         self.enable_kv_nz = ascend_config.torchair_graph_config.enable_kv_nz
         self.chunked_prefill_for_mla = ascend_config.chunked_prefill_for_mla
 
-        self.prefill_mask = None
+        vllm_config = get_current_vllm_config()
+        self.prefill_mask = torch.triu(
+            torch.ones(512,
+                       512,
+                       device="npu",
```
MengqingCao: will this always create mask on rank 0? is there any way to get current device?

whx-sjtu: This is called in a process that has already set_device, in this situation 'npu' will be automatically set to corresponding rank in set_device

wangxiyuan:
please fix the UT asap

wangxiyuan:
the new commit is only for ut fix. it's good to go

vllm_ascend/attention/mla_v1.py
```diff
@@ -483,10 +483,12 @@ def __init__(
         self.enable_kv_nz = ascend_config.torchair_graph_config.enable_kv_nz
         self.chunked_prefill_for_mla = ascend_config.chunked_prefill_for_mla
 
+        vllm_config = get_current_vllm_config()
+        self.ring_mla_mask_size = 512
         self.prefill_mask = None
 
         # Adapt torch air graph mode with spec decoding.
-        speculative_config = get_current_vllm_config().speculative_config
+        speculative_config = vllm_config.speculative_config
```
momo609, line 491: does 091dev have the same bug?

PR #2706, opened 2025-09-02
[main] [refactor] refactor common_fused_moe.py

Pr0Wh1teGivee:
1. Move prepare/finalize operation from moe_comm_method to /ops/moe/fused_moe_prepare_and_finalize
2. Adapt to token_dispatcher in moe_comm_method
3. Move moe_comm_method/experts_selector/token_dispatcher/fused_moe_prepare_and_finalize to /ops/moe
no
e2e & ut

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/f4962a6d55a340ebb569d377c842deff7611d8f7


wangxiyuan (approved):

wangxiyuan:
it's very clear now.

vllm_ascend/ops/moe/moe_comm_method.py
```diff
@@ -0,0 +1,298 @@
...
+from vllm_ascend.ops.moe.fused_moe_prepare_and_finalize import (
+    FusedMoEPrepareAndFinalizeWithAll2All,
+    FusedMoEPrepareAndFinalizeWithAllGather, FusedMoEPrepareAndFinalizeWithMC2)
+from vllm_ascend.ops.moe.moe_mlp import unified_apply_mlp
+from vllm_ascend.ops.moe.token_dispatcher import (TokenDispatcherWithAll2AllV,
+                                                  TokenDispatcherWithAllGather,
+                                                  TokenDispatcherWithMC2)
+
+
+class MoECommMethod(ABC):
```
MengqingCao, line 33: nits: I think an abstract class should just contain the unified api, without the detailed implementation, and make more comments like the original `MoECommMethod` class. Maybe we could add a class `MoECommMethodBase` to inherit form `MoECommMethod`, which contains the detailed prepare and other member functions.  We could do this in next pr

vllm_ascend/ops/moe/fused_moe_prepare_and_finalize.py
```diff
@@ -0,0 +1,240 @@
...
+from vllm.distributed.parallel_state import (
+    get_dp_group, get_tensor_model_parallel_rank,
+    get_tensor_model_parallel_world_size)
+from vllm.forward_context import get_forward_context
+from vllm.model_executor.layers.fused_moe import FusedMoEConfig
+
+
+class FusedMoEPrepareAndFinalize(ABC):
+
+    def __init__(self, moe_config: FusedMoEConfig):
```
MengqingCao, line 32: Let's make more comments of `FusedMoEPrepareAndFinalize` and all the member functions of it

vllm_ascend/ops/moe/moe_comm_method.py
Pr0Wh1teGivee: good point, will be fixed in my next pr.

vllm_ascend/ops/moe/fused_moe_prepare_and_finalize.py
Pr0Wh1teGivee: same, comments will be added in my next pr.

PR #2708, opened 2025-09-02
[FEATURE][MTP] Support MTP > 1

1092626063:
image


wangxiyuan (approved):

jianzs:
@1092626063  When `lmhead_tp` is enabled, setting MTP greater than 1 will cause an error.

PR #2718, opened 2025-09-03
LLMdatadist connector adapt the distributed KV aggregation for the main branch. 

liziyu179:
LLMdatadist connector adapt the distributed KV aggregation for the main branch. Change the P node from returning "finish sending" only when TP0 responds to returning "finish sending" as soon as each NPU receives it. For the MLA model, the "finish sending" count for the P node is the TP size of the D node. For the MHA and GQA models, it is the TP size of the P node.


wangxiyuan (approved):

MengqingCao:
UT failed due to the removed cases in main. But please take a look at the failure of test on pd aggregation

PR #2740, opened 2025-09-04
add gatherep select.

momo609:
add gatherep select.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/e599e2c65ee32abcc986733ab0a55becea158bb4

### DIFF
--- a/vllm_ascend/ascend_forward_context.py
+++ b/vllm_ascend/ascend_forward_context.py
@@ -45,13 +45,12 @@ def _get_fused_moe_state(ep_size: int, with_prefill: bool,
 def get_dispatcher_name(ep_size: int, with_prefill: bool) -> str:
     if ep_size == 1:
         return "TokenDispatcherWithAllGather"
-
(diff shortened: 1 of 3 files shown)


vllm_ascend/ascend_forward_context.py
```diff
@@ -45,13 +45,12 @@ def _get_fused_moe_state(ep_size: int, with_prefill: bool,
 def get_dispatcher_name(ep_size: int, with_prefill: bool) -> str:
     if ep_size == 1:
         return "TokenDispatcherWithAllGather"
-
-    if ep_size < 16:
-        return "TokenDispatcherWithAll2AllV"
-
-    if with_prefill:
+    elif envs_ascend.VLLM_ENABLE_FUSED_EXPERTS_ALLGATHER_EP and ep_size > 1:
```
momo609, line 48: TODO: this logic will be consolidated with moe_common_method, without relying on environment variable checks.

PR #2741, opened 2025-09-04
[long_seq_optim] support cp&sp

LookAround0301:

- vLLM version: v0.10.2


vllm_ascend/core/scheduler.py
```diff
@@ -155,6 +162,26 @@ def skip_cur_request():
                 num_new_local_computed_tokens = 0
                 num_computed_tokens = request.num_computed_tokens
 
+            if self.cp_size * self.sp_size > 1:
+                # block_size align
+                num_total_blocks = cdiv((request.num_tokens - num_computed_tokens), self.block_size)
+
+                num_blocks_of_cp_sp = np.array([num_total_blocks // (self.cp_size * self.sp_size)] * self.cp_size * self.sp_size)
```
momo609: record the result of (self.cp_size * self.sp_size) to avoid repeated calculation.

vllm_ascend/models/qwen3_moe.py
```diff
@@ -386,4 +386,4 @@ def forward(
             input_ids, self.enable_sequence_parallelism)
         hidden_states = self.model(input_ids, positions, intermediate_tensors,
                                    inputs_embeds, _metadata_for_padding)
-        return hidden_states
+        return hidden_states
```
momo609, line 389: the sp logic in moe may be integrated with the flashcomm1 code in the dense model?

LookAround0301: FlashComm1 has not yet been adapted for Qwen3-MoE.

vllm_ascend/core/scheduler.py
LookAround0301: Okay, it has been modified.

wangxiyuan:
vllm-ascend 0.12.0 has support cp/sp already.

PR #2772, opened 2025-09-05
[main] addrmsnorm + quant fusion optim in Dense Models

rjg-lyh:
This PR fused addrmsnorm op and w8a8 quant op to get better perf.

No.

- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/0faf3cc3e84a83bac56492d297f2c9909fec7de9


Angazenn:
Can this PR handle the situation where RMSNorm has an extra bias that is introduced by anti-outlier algorithms?

MengqingCao (approved):

PR #2776, opened 2025-09-05
Refactor the Spec decode module to merge MTP non-torchair and eagle modes into one file, separating torchair and non-torchair modes #2773

weisirui-eng:
Refactor the Spec decode module to merge MTP non-torchair and eagle modes into one file, separating torchair and non-torchair modes

No.

How was this patch tested?


vllm_ascend/torchair/eagle_torchair_proposer.py
```diff
@@ -1,4 +1,5 @@
 import types
```
wangxiyuan, line 1: why move mtp file to here and rename to eagle?

vllm_ascend/spec_decode/eagle_proposer.py
```diff
@@ -1,27 +1,35 @@
 # SPDX-License-Identifier: Apache-2.0
```
wangxiyuan, line 1: any reason to merge mtp and eagle into one?

weisirui-eng: Currently, the vLLM community manages MTP and Eagle in a unified manner, which ensures alignment with the vLLM community's standards

wangxiyuan: OK. It makes sense. @JC-ut0 any idea about the change?

PR #2777, opened 2025-09-05
support qwen25 vl w8a8 quantization

wenba0:


wangxiyuan:

Issue #2742, opened 2025-09-04
[Bug]: doctest failed due to rotary_embedding


Yikun:
### Your current environment

https://github.com/vllm-project/vllm-ascend/actions/runs/17454967332/job/49566622600

### 🐛 Describe the bug

```
terminate called after throwing an instance of 'c10::Error'
  what():  
Mismatch in kernel C++ signatures
  operator: _C::rotary_embedding(Tensor positions, Tensor($0! -> ) query, Tensor($1! -> )? key, int head_size, Tensor cos_sin_cache, bool is_neox) -> ()
    registered at /tmp/pip-install-vnbqqk9u/vllm_cd6a66d211f542769373702664426970/csrc/cpu/torch_bindings.cpp:74
  kernel 1: void (at::Tensor&, at::Tensor&, std::optional<at::Tensor>, long, at::Tensor&, bool)
    dispatch key: CPU
    registered at /tmp/pip-install-vnbqqk9u/vllm_cd6a66d211f542769373702664426970/csrc/cpu/torch_bindings.cpp:74
  kernel 2: std::tuple<at::Tensor, at::Tensor> (at::Tensor&, at::Tensor&, at::Tensor&, long, at::Tensor&, bool)
    dispatch key: Meta
    registered at /workspace/vllm-ascend/csrc/torch_binding_meta.cpp:91

Exception raised from registerKernel at /pytorch/aten/src/ATen/core/dispatch/OperatorEntry.cpp:121 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0xd4 (0xffff757c3ea4 in /tmp/vllm_venv/lib/python3.11/site-packages/torch/lib/libc10.so)
frame #1: c10::detail::torchCheckFail(char const*, char const*, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 0xe4 (0xffff75763e44 in /tmp/vllm_venv/lib/python3.11/site-packages/torch/lib/libc10.so)
[...]

PR #2786, opened 2025-09-05
Fix the bugs about operator registration by PyTorch Dispatcher

FFFrog:
**Background:**

There are two principles about operator registration in PyTorch
- The same namespace can be only registered once by `TORCH_LIBRARY`
- The operator signatures can be only registered once by `def`

Considering that all custom operators defined in the current repo are only used by Ascend, instead of defining a common operator schema by vLLM, all accelerators then follow this operator schema and complete the implementation based on their respective hardware, which is conducive to functional abstraction.

Therefore, we can rename the operator registration namespace to an Ascend-specific namespace(**_C_ascend**).

Related ISSUE: https://github.com/vllm-project/vllm-ascend/issues/2742

- vLLM version: main
- vLLM main: https://github.com/vllm-project/vllm/commit/f592b3174b39a7aeac52432d66d76e89ff0a80b4

### DIFF
--- a/csrc/torch_binding.cpp
+++ b/csrc/torch_binding.cpp
@@ -141,7 +141,7 @@ std::tuple<at::Tensor, at::Tensor> get_masked_input_and_mask(
     TP2, rank 1:
                             |< -----------BASE----------- >|< -BASE PADDING- >|< -----------LORA PADDING----------- >|
     corresponding token_id: | 512 | 513 | 514 | ... | 1009 | -1  | ...  | -1  |  -1  | ... |  -1  | -1  | ... |   -1 |
-                     index: |  0  |  1  |  2  | ... | 497  | 498 | ...  | 511 | 512  | ... | 519  | 520 | ... |  543 | 
+                     index: |  0  |  1  |  2  | ... | 497  | 498 | ...  | 511 | 512  | ... | 519  | 520 | ... |  543 |
     Parameters:
         org_vocab_start_index //base embeddings start
         org_vocab_end_index //base embeddings end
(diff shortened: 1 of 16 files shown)


fffrog:
@Yikun The draft is ready, please help to take a look at it.

csrc/torch_binding.cpp
```diff
@@ -104,7 +105,9 @@ std::tuple<at::Tensor, at::Tensor> rotary_embedding(at::Tensor &positions, at::T
         return 0;
     });
     cmd.Run();
-    return {query_dst, key_dst};
+
+    query.copy_(query_dst);
```
ganyi1996ppo: There are some inputs for this kernel. We Implement this rotary embedding originally is aiming for prevent additional memory reorder triggered by contiguous. Although the changes in this PR aligns the torch schema with the vllm's impl, but it may bring huge regression on e2e scenario( stride tensor -> contiguous tensor -> stride tensor).

ganyi1996ppo: Those changes will bring no performance advantage compared with `torch_npu._npu_rotary_embedding`, If we are looking for adopt this in real workload, I do not suggest this changes.

ganyi1996ppo:
> **Background:**
> 
> There are two principles about operator registration in PyTorch
> 
> * The same namespace can be only registered once by `TORCH_LIBRARY`
> * The operator signatures can be only registered once by `def`
> 
> **How to fix:**
> 
> * for the first problem, we can use `TORCH_LIBRARY_FRAGMEN` to expand operators within the same NAMESPACE.
> * for the second problem, the best way to fix it is to define all the general operator schemas in vLLM insteal of in every plugin repo.
> 
> Related ISSUE: #2742
> 
> * vLLM version: v0.10.1.1
> * vLLM main: [vllm-project/vllm@006e7a3](https://github.com/vllm-project/vllm/commit/006e7a34aeb3e905ca4131a3251fe079f0511e2f)

I remember we won't compile the cpu version of vllm right @Yikun @wangxiyuan ? If we don't compile the cpu version, this shouldn't happen. And if we have to compile a cpu version of vllm, I suggest we just adopt another naming for this kernel, or just add a overload version of rope.

Yikun:
> I remember we won't compile the cpu version of vllm right @Yikun @wangxiyuan ? If we don't compile the cpu version, this shouldn't happen. 

We are recommand to install the vllm cpu version in docs: https://vllm-ascend.readthedocs.io/en/latest/installation.html#setup-vllm-and-vllm-ascend.

csrc/torch_binding.cpp
fffrog: Got it, thank you.

Restore all the changes and keep it as it is

fffrog:
> I remember we won't compile the cpu version of vllm right @Yikun @wangxiyuan ? If we don't compile the cpu version, this shouldn't happen. And if we have to compile a cpu version of vllm, I suggest we just adopt another naming for this kernel, or just add a overload version of rope.

Thank you for your helpful advices and have followed it to update this pr, please take a look at it, thank you.

wangxiyuan:
https://github.com/vllm-project/vllm-ascend/blob/eab3635850ba351af81d76a7b4b3db46ffb7f697/vllm_ascend/compilation/acl_graph.py#L18  weak_ref_tensors is called from vLLM which use torch.ops._C. we should use our ops instead.

ganyi1996ppo:
LGTM

PR #2788, opened 2025-09-05
[Disagg][Perf] Use CUDA event sync instead of blocking tolist to avoid unintentional copy ops blocking across different CUDA streams, improving disagg TTIT/TTFT

jesse996:
This PR is based on top of https://github.com/vllm-project/vllm/pull/22760
When we copy the sampled valid token ids from device to host, avoid using tolist which would trigger a CUDA wise stream sync if the source is on device. We change it to use non-blocking copy followed by an explicit CUDA event sync.

Bring up vLLM server
```bash
VLLM_USE_V1=1 vllm serve Qwen/Qwen2.5-14B-Instruct --disable-l
og-requests -tp 8 --max-num-seqs 64 --no-enable-prefix-caching --max_num_batched_tokens=8000
```


wangxiyuan:
nice work, can you print the benchmark result with/without this PR to make sure it works as expect?

jesse996:
> nice work, can you print the benchmark result with/without this PR to make sure it works as expect?

added  to the beginning

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3054,3 +3061,18 @@ def get_supported_pooling_tasks(self):
 
     def _build_drafter_prepare_inputs_torchair_param(self):
         return False
+
+    def _to_list(self, sampled_token_ids: torch.Tensor) -> list[list[int]]:
+        # This is a short term mitigation for issue mentioned in
```
wangxiyuan, line 3066: can you rewrite the comment to ascend case?

jesse996: updated

wangxiyuan (approved):

PR #2790, opened 2025-09-05
Deepseek Mtp model uses the lm_head and embedding from the main model

zzhx1:
In the Deepseek technical report, it is mentioned that the embedding and lmhead layers of the MTP layer are shared with the main model, but the current implementation independently loads the complete embedding and lmhead. In the Deepseek-R1 model, their weight sizes are 129280*7168 in fp16 format, which is 1.72G.

This PR fixes the MTP layer to use the lmhead and embedding of the main model, saving 3.45G of GPU memory in the pure DP scenario.

The current process will first create temporary spaces for the embedding and lmhead in the mtp layer, then I will call torch.equal to determine if the two matrices are the same. If they are the same, they will be reused, and the previous tensor will be released.

image

- vLLM version: v0.12.0


vllm_ascend/torchair/models/torchair_deepseek_mtp.py
```diff
@@ -216,3 +212,95 @@ def forward(
...
+            ckpt_up_proj_name="up_proj",
+            num_experts=self.config.n_routed_experts)
+
+        params_dict = dict(self.named_parameters())
+        loaded_params: set[str] = set()
+        for name, loaded_weight in weights:
+            if "rotary_emb.inv_freq" in name:
+                continue
+            # The mtp model uses the lm_head and embedding from the main model.
+            if "shared_head" in name:
```
zzhx1: Added code to skip loading of lmhead and embed, can only rewrite load_weight

vllm_ascend/spec_decode/mtp_proposer.py
```diff
@@ -100,6 +100,12 @@ def load_model(self, model) -> None:
         process_weights_after_loading(self.model, draft_model_config,
                                       target_device)
 
+        main_embed_tokens = main_model.model.embed_tokens
```
momo609: The absence of redundant lm_head invocations in the dummy run it may lead to deadlocks during DP load imbalance scenarios?

zzhx1:
@ApsarasX please add a ready-to-test lable

vllm_ascend/spec_decode/mtp_proposer.py
```diff
@@ -105,6 +106,11 @@ def load_model(self, model) -> None:
         process_weights_after_loading(self.model, draft_model_config,
                                       target_device)
 
+        # use main model's embedding and LMhead
```
wangxiyuan: let' wait https://github.com/vllm-project/vllm-ascend/pull/3528 merge first. Then mtp change can be in torchair module as well.

vllm_ascend/spec_decode/mtp_proposer.py
```diff
@@ -105,6 +106,11 @@ def load_model(self, model) -> None:
...
+        self.model.model.embed_tokens = main_model.model.embed_tokens
+        for layer_module in self.model.model.layers.values():
+            layer_module.shared_head.head = main_model.lm_head
```
linfeng-yuan: In the quantized DeepSeek weights provided by the Ascend community that address function call accuracy, the weight values of lm_head and shared_head differ numerically. When using these new quantized weights, this code line will result in precision degradation.

Please have a check whether there is an elegant option to recognize these two weights. @wangxiyuan

zzhx1:
```
对比 lm_head.weight (文件: quant_model_weight_w8a8_dynamic-00158-of-00162.safetensors) vs model.layers.61.shared_head.head.weight (文件: quant_model_weight_w8a8_dynamic-00161-of-00162.safetensors):
    最大差值=0.000000, 百分比误差=0.0000%, 是否在1e-4范围内=True
对比 model.embed_tokens.weight (文件: quant_model_weight_w8a8_dynamic-00001-of-00162.safetensors) vs model.layers.61.embed_tokens.weight (文件: quant_model_weight_w8a8_dynamic-00162-of-00162.safetensors):
    最大差值=0.000000, 百分比误差=0.0000%, 是否在1e-4范围内=True
```

image

I tested the deepseek model files myself and found that the weights of lmhead and shared_head are exactly the same, and the embedding weights are also the same.
And it has been tested without affecting the acceptance rate of the MTP layer

linfeng-yuan: https://modelers.cn/models/Modelers_Park/DeepSeek-V3.1-w8a8-function_call
Modelslim has updated new quantization option `--rot` to fix function calling accuracy for quantized deepseek-like models. And re-using the lm_head and shared_head would lead to acceptance rate degradation for deepseek_r1 (~2% which can be ignored). The developers of ModelSlim do not recommend the re-usage behaviour becuase of robustness issues.  Could you please add a verification here to check the numerical value of lm_head and shared_head weights?

linfeng-yuan: As far as I know, MindIE adds a weight validation here to determine which we can re-use the lm_head of non-mtp part.

linfeng-yuan (approved):

PR #2792, opened 2025-09-06
replace npu_incre_flash_attention with npu_fused_infer_attention_score

panchao-hub:
The npu_incre_flash_attention  interface is no longer maintained and is replaced by npu_fused_infer_attention_score
No

- vLLM version: main
- vLLM main: https://github.com/vllm-project/vllm/commit/e93f4cc9e37484009f74e15d3111a1f335c532a5


Angazenn:
please note that FIA might not support 300I Duo platform, so we need a branch here.

wangxiyuan:
Torchair is removed already. Please use aclgraph instread

PR #2801, opened 2025-09-08
[CI] Reduce CI time

wangxiyuan:
Only run light e2e test before the PR is approved to reduce CI time.


.github/workflows/vllm_ascend_test_full.yaml
```diff
@@ -0,0 +1,230 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# This file is a part of the vllm-ascend project.
+#
+name: 'test-full'
+
+on:
+  pull_request_review:
```
Yikun: Every approved will trigger this repeatly, and the previous job will cancel and retrigger

PR #2820, opened 2025-09-09
Dynamic Expert Load Balance with Zero-like-overhead

offline893:
### Motivation
Currently dynamically experts balancing would stop-the-world. Asynchronously expert load balancing would be better without flowing problems:

Host-bound latency:
There are many cpu operations during EPLB such as eplb-algorithm、creating p2p ops、and log2phy expert converting would spend long cpu time, as ~1s.
Communication latency: The transfer time would cost much in the situation without nvlink. As the weight of an expert maybe transfer to multiple new positions, thus N times send/recv for one expert, with result long latency. We had tested that batch_isend_irecv cost more 100ms for 16 experts weight transmission in A2 server of ascend.

SwiftBalancer would not stop-the-world anymore, in out test on NPU 1~2ms cost for each layer while benefit 5ms-8ms decode latency with ep_size = 64.
The following updates have been made:
1、expert distribution recording with lower cost.
2、async cpu computing for eplb algo and other python operator.
3、new eplb algo with less expert rebalancing while almost the same effect.
### Proposed Change
We will gradually migrate the EPLB logic to the VLLM community and implement a generalized design. Relevant RFC: https://github.com/vllm-project/vllm/issues/22246
The overall workflow involves:
474430541-23b06f58-23bc-44a3-a1be-00f268aeb15c
1. Record experts distribution during forward. We using expert_token_num after disptach instead of topk_ids, thus we got much smaller tensor shape to reduce cost of hbm recording and add-operator.
2. Do all-gather for experts distribution. Using all-gather instead of all-reduce as less traffic volume.
3. Wake up eplb worker process with experts distribution when num_iterations comes. Run eplb algorithm in eplb worker.
4. Generate p2p send/recv ops and other operator such as log2phy would cost long cpu time.
5. Lanch ibatch_send_recv in async_stream before forward.
6. After forward, wait for the ibatch_send_recv finish, then do uapte expert map and expert weights.
### Co-author
Co-authored-by: raindaywhu raindaywhu@raindaywhu@ 163.con
Co-authored-by: njuyuan yuanjl19@smail.nju.edu.cn
Co-authored-by: qmkakaxi wjh1594260677@qq.com

- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/0af3ce13558056b3637dd4661c89f4b037faa85a


vllm_ascend/ascend_config.py
```diff
@@ -45,6 +45,11 @@ def __init__(self, vllm_config):
             ascend_scheduler_config)
 
         self.expert_map_path = additional_config.get("expert_map_path", None)
+        # Eplb config
```
wangxiyuan: Add more note here:

`# TODO: once xxx pr is merged in vLLM. Remove this config`

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1877,6 +1900,9 @@ def _dummy_run(
                                                        num_reqs,
                                                        skip_attn=True)
 
+        if not is_torchair_compile and not self.in_profile_run and self.dynamic_eplb:
```
wangxiyuan: remove torchair related code

vllm_ascend/ops/common_fused_moe.py
```diff
@@ -368,6 +373,25 @@ def __init__(
         self.moe_config.dp_group = get_dp_group()
         self.moe_config.ep_group = get_ep_group()
         self.moe_config.mc2_group = get_mc2_group()
+        ascend_config = get_ascend_config()
+        self.dynamic_eplb = ascend_config.dynamic_eplb
+        self.expert_map_path = ascend_config.expert_map_path
+        if self.expert_map_path and os.path.exists(self.expert_map_path):
```
momo609: Add validation for os.path.exists(self.expert_map_path) to ensure path security and prevent unauthorized access.

vllm_ascend/worker/model_runner_v1.py
offline893: I have modified it according to the review comments.

vllm_ascend/ops/common_fused_moe.py
offline893: I have modified it according to the review comments.

vllm_ascend/ascend_config.py
offline893: I have modified it according to the review comments.

vllm_ascend/ops/common_fused_moe.py
```diff
@@ -289,12 +296,55 @@ def __init__(
         self.moe_config.dp_group = get_dp_group()
         self.moe_config.ep_group = get_ep_group()
         self.moe_config.mc2_group = get_mc2_group()
+        ascend_config = get_ascend_config()
+        self.dynamic_eplb = ascend_config.dynamic_eplb
+        self.expert_map_path = ascend_config.expert_map_path
```
wangxiyuan: I think all the code L301-L326 should be under `if self.dynamic_eplb:`

offline893: I have modified it according to the review comments.

PR #2825, opened 2025-09-09
[Bugfix] Fix broken CI

Potabk:
1. Initial support disable tp for integrating with [vllm-commit](https://github.com/vllm-project/vllm/pull/23024)
2. [vllm@commit](https://github.com/vllm-project/vllm/pull/23673) now use `bytes` to save the `BlockHash` to reduce GC overhead, this pr add the integration

- vLLM version: main
- vLLM main: https://github.com/vllm-project/vllm/commit/e40827280b225bf0a0797cc9842fc3cdfea8ebdf

### DIFF
--- a/.github/workflows/vllm_ascend_test.yaml
+++ b/.github/workflows/vllm_ascend_test.yaml
@@ -71,10 +71,10 @@ jobs:
               - 'tests/ut/**'
 
   ut:
-    needs: [changes]
+    needs: [lint, changes]
     name: unit test
     # only trigger unit test after lint passed and the change is e2e and ut related.
-    if: ${{ needs.changes.outputs.e2e_tracker == 'true' || needs.changes.outputs.ut_tracker == 'true' }}
+    if: ${{ needs.lint.result == 'success' && (needs.changes.outputs.e2e_tracker == 'true' || needs.changes.outputs.ut_tracker == 'true') }}
     runs-on: ubuntu-latest
     container:
       image: quay.io/ascend/cann:8.2.rc1-910b-ubuntu22.04-py3.11
@@ -83,7 +83,7 @@ jobs:
         VLLM_USE_MODELSCOPE: True
     strategy:
       matrix:
-        vllm_version: [v0.10.1.1, main]
+        vllm_version: [main]
     steps:
       - name: Install packages
         run: |
@@ -139,7 +139,7 @@ jobs:
       max-parallel: 2
       matrix:
         os: [linux-aarch64-a2-1]
-        vllm_version: [v0.10.1.1, main]
+        vllm_version: [main]
     name: singlecard e2e test - light
     runs-on: ${{ matrix.os }}
     container:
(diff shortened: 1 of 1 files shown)


.github/workflows/vllm_ascend_test.yaml
```diff
@@ -71,10 +71,10 @@ jobs:
               - 'tests/ut/**'
 
   ut:
-    needs: [lint, changes]
```
MengqingCao: plz revert this change

.github/workflows/vllm_ascend_test.yaml
```diff
@@ -71,10 +71,10 @@ jobs:
...
+    needs: [changes]
     name: unit test
     # only trigger unit test after lint passed and the change is e2e and ut related.
-    if: ${{ needs.lint.result == 'success' && (needs.changes.outputs.e2e_tracker == 'true' || needs.changes.outputs.ut_tracker == 'true') }}
```
MengqingCao: ditto

tools/mypy.sh
```diff
@@ -35,6 +35,6 @@ run_mypy() {
     mypy --check-untyped-defs --follow-imports skip --python-version "${PYTHON_VERSION}" "$@"
 }
 
-run_mypy vllm_ascend
```
MengqingCao: ditto

vllm_ascend/ops/linear.py
```diff
@@ -62,15 +62,16 @@ def __init__(
...
     ):
         self.comm_group = None
         if prefix.find("gate_up_proj") != -1 and mlp_tp_enable():
             self.comm_group = get_mlp_tp_group()
         else:
             self.comm_group = get_tp_group()
 
-        self.tp_size = self.comm_group.world_size
-        self.tp_rank = self.comm_group.rank_in_group
+        self.tp_size = self.comm_group.world_size if not disable_tp else 1
```
MengqingCao: could the `mlp_tp` be controled by `disable_tp`, either? @momo609 PTAL

.github/workflows/vllm_ascend_test.yaml
Potabk: I think if we should keep lint and e2e run in parallel, cause the lint spent 9min and e2e have 13min(they are almost the same cost)

vllm_ascend/ops/linear.py
momo609: The optimizations of mlptp and o_proj_tp both require TP to be enabled. disable_tp directly will impact performance.

tools/mypy.sh
MengqingCao: Let's recover this in next pr to fix ci quickly

vllm_ascend/ops/linear.py
MengqingCao: OK, we'll only do API input args alignment in this CI recover pr and discuss ViT DP and the customized tp in vllm-ascend later

.github/workflows/vllm_ascend_test.yaml
wangxiyuan: no, we should not waste CI resouce if the lint is not passed

MengqingCao: as the e2e test and ut test are all passed, plz revert this and will merge this pr after this

Potabk: if lint not passed, e2e will cancelled(in parallel scenario), so with all factor considered, the benefits outweigh the wasted time.

PR #2827, opened 2025-09-09
[Bugfix] Fix aclgraph sizes capture error in Qwen3 Moe model case

lilinsiman:
1. Solved the problem that in the Qwen3 Moe model case, opening DP would use an extra stream, causing ACLgraph sizes capture error
2. After experimentation, it was found that in many cases, some operators would occupy more streams than expected. Therefore, the buffer area for streams in ACLgraph was not large enough. After discussion, extra 120 streams were added as buffer.

no

ut

- vLLM version: main
- vLLM main: https://github.com/vllm-project/vllm/commit/0ae43dbf8cb28a299ae724fc742b0c5bcddea868

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -40,12 +40,13 @@ if TYPE_CHECKING:
 else:
     VllmConfig = None
 
-# NOTE: Currently, we can only capture 1920 graphs at most,
+# NOTE: Currently, we can only capture 1800 graphs at most,
 # due to the limitation of ACL graph. This number is bounded by
-# the number of streams, which is 2048, we save 128 streams
+# the number of streams, which is 2048, we save 248 streams
 # as a buffer.
 # Maximum number of graphs that can be captured by ACL Graph
-MAX_CAPTURE_SIZE = 1920
+# TODO: Find out whether we need to solve allreduce function
+MAX_CAPTURE_SIZE = 1800
 
 ASCEND_QUANTIZATION_METHOD = "ascend"
 SOC_VERSION_INFERENCE_SERIES = ["Ascend310P3"]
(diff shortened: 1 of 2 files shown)


vllm_ascend/utils.py
```diff
@@ -320,6 +321,8 @@ def update_aclgraph_sizes(vllm_config: VllmConfig) -> None:
         # TODO: Find out whether we need to take into account the pp_size
         parallel_factor = 1 + num_comm_groups + int(
             parallel_config.enable_expert_parallel)
+        if is_moe_model(vllm_config):
```
MengqingCao, line 324: we still need to check if is moe model right?

yiz-liu: Yeah, for MoE, we'll have to do `prepare` and `finalize` around `quant_method.apply`, which will have extra communication.

MengqingCao:
ut failed due to some known issues, which are unrelated to this pr. Let's merge this after the full e2e test passed

Yikun:
Please rewrite the PR title as a meaningful one. I didn't get any human readable info from original title.

lilinsiman:
> Please rewrite the PR title as a meaningful one. I didn't get any human readable info from original title.

Already done

vllm_ascend/utils.py
```diff
@@ -320,6 +321,8 @@ def update_aclgraph_sizes(vllm_config: VllmConfig) -> None:
...
+            parallel_factor += (parallel_config.data_parallel_size > 1)
         # Calculate maximum supported batch sizes considering model architecture on the A2 Hardware Device
         # Assume the following case:
         # MAX_CAPTURE_SIZE = 1920, num_hidden_layers = 48, data_parallel_size is 1, tensor_parallel_size is 4,
```
momo609, line 328: Also update the note.

PR #2829, opened 2025-09-09
GLM4.5 W8A8 Quantization Adaptation

liuhy1213-cell:
GLM4.5 W8A8 Quantization Adaptation, Adapted from vllm/model_executor/models/glm4_moe.py

- vLLM version: main
- vLLM main: https://github.com/vllm-project/vllm/commit/e40827280b225bf0a0797cc9842fc3cdfea8ebdf


wangxiyuan:

PR #2841, opened 2025-09-09
[Feat]support dynamic quantization in allgather

WithHades:

[Feat]support dynamic quantization in allgather


yiz-liu:
@MengqingCao Please add `ready` label for this PR to enable full test cases.

PR #2856, opened 2025-09-10
[Quantization] register AscendQuantRMSNorm for quantization

22dimensions:

modelslim will generate self.bias for rms norm in quantization, since RMSNorm in vllm has no this parameter, so its nesscesary
to create a AscendQuantRmsNorm.

No

test by DeepSeek R1


vllm_ascend/utils.py
```diff
@@ -521,7 +527,7 @@ def register_ascend_customop():
         "VocabParallelEmbedding": AscendVocabParallelEmbedding,
         "ParallelLMHead": AscendParallelLMHead,
         "LogitsProcessor": AscendLogitsProcessor,
-        "RMSNorm": AscendRMSNorm,
+        "RMSNorm": rms_norm_class,
         "FusedMoE": AscendFusedMoE,
         "MultiHeadLatentAttention": AscendMultiHeadLatentAttention,
     }
```
wangxiyuan, line 527:
```
    if vllm_config is not None and \
        vllm_config.quant_config is not None and \
        any("norm.bias" in name for name in vllm_config.quant_config.quant_description.keys()):
        REGISTERED_ASCEND_OPS["RMSNorm"]= AscendQuantRMSNorm
```

Angazenn:
what's the difference between this op and `AddRMSNormW8A8Quant` in layernorm.py ? which forward_oot takes effects if we run a w8a8 model?

wangxiyuan:
@Angazenn it contains bias here. Let's merge them into one in the future

vllm_ascend/ops/layernorm.py
```diff
@@ -97,3 +97,28 @@ def forward_oot(
         x, residual = torch_npu.npu_rms_norm(x, self.weight,
                                              self.variance_epsilon)
         return x
+
+
+class AscendQuantRMSNorm(AscendRMSNorm):
```
wangxiyuan, line 102: we shoud merge AscendQuantRMSNorm , AscendRMSNorm and AddRMSNormW8A8Quant  into one

PR #2866, opened 2025-09-10
[main] rm old version DBO patch

shikang-hangzhou:


vllm_ascend/models/__init__.py
```diff
@@ -23,22 +23,13 @@ def register_model():
             "vllm_ascend.models.qwen2_5_vl_without_padding:AscendQwen2_5_VLForConditionalGeneration_Without_Padding"
         )
 
-    if envs_ascend.VLLM_ASCEND_ENABLE_DBO:
```
wangxiyuan, line 26: 1. VLLM_ASCEND_ENABLE_DBO this env should be removed as well
2. similar dbo code in model runner should be removed 
3.  enable_dbo_accros_dp should be removde as well?

PR #2883, opened 2025-09-12
[Ops] Fix bug in register_custom_ops without forward_context

rjg-lyh:
This PR fixed the bug in register_custom_ops without forward_context. We set try-except to consider this situation.

No.

- vLLM version: main
- vLLM main: https://github.com/vllm-project/vllm/commit/7920de0a2af7c663efae29856f1bac061a553b58

### DIFF
--- a/vllm_ascend/ops/register_custom_ops.py
+++ b/vllm_ascend/ops/register_custom_ops.py
@@ -7,6 +7,7 @@ from vllm.distributed import (get_tensor_model_parallel_rank,
                               tensor_model_parallel_all_reduce,
                               tensor_model_parallel_reduce_scatter)
 from vllm.forward_context import get_forward_context
+from vllm.logger import logger
 from vllm.utils import direct_register_custom_op
 
 import vllm_ascend.envs as envs_ascend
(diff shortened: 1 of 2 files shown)


wxsIcey:
Need to add e2e test, I will submit another PR.

vllm_ascend/ops/register_custom_ops.py
```diff
@@ -31,19 +38,31 @@ def _maybe_chunk_residual_impl(x: torch.Tensor,
 
 def _maybe_all_gather_and_maybe_unpad_impl(x: torch.Tensor,
                                            label: bool) -> torch.Tensor:
-    flashcomm_v1_enabled = get_forward_context().flashcomm_v1_enabled
+    try:
+        forward_context = get_forward_context()
+    except AssertionError:
+        logger.info("Forward context is None, skipping the operation.")
+        return x
```
momo609, line 45: If returning x directly without pad here may cause issues with the flashcomm_v1 functionality，check also be added in linear.py ?

rjg-lyh: This bug happens when vlm directly runs `get_multimodal_embeddings`. In this situation, pad action also can't happen, due to no forward_context. Therefore, your concerns will not arise, and you can rest assured.

PR #2893, opened 2025-09-12
[Revert]Revert unquantized linear nz support

anon189Ty:
If we want a performance improvement by cast the format of weights from ND to NZ in unquantized linear, we need to transpose the weight shape too. But will case some errors in current code when running some models such as  DeepSeek-V3-w4a8. Currently temporarily roll back.

- vLLM version: main
- vLLM main: https://github.com/vllm-project/vllm/commit/64d90c3e4fe2a0e4395b8c94344dcdf78fa4cd22

### DIFF
--- a/tests/ut/ops/test_linear.py
+++ b/tests/ut/ops/test_linear.py
@@ -5,13 +5,11 @@ from unittest.mock import MagicMock, patch
 
 import torch
 
-from tests.ut.base import TestBase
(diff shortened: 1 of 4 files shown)


wangxiyuan:
https://github.com/vllm-project/vllm-ascend/commit/756b8a1946aa9396d5bc7b9c67547fcb93fad630 merged by this already.

PR #2894, opened 2025-09-12
[Core] Disable the chunked prefill feature in Non-MLA LLMs

rjg-lyh:
This PR enforces the forcible disabling of the chunked prefill feature in Non-MLA models, as the performance of operators supporting this functionality is currently suboptimal. Unless the user has enabled chunked prefill in the ascend_scheduler_config, we would allow this feature.

No.

Related: https://github.com/vllm-project/vllm-ascend/pull/2659

- vLLM version: main
- vLLM main: https://github.com/vllm-project/vllm/commit/d21a36f5f9569949dd6c313deed609d77e393850


Yikun:
image

The round2 perf has perf regression which is very wired, but compare to v0.10.1rc1 it's OK, so merge this first.

PR #2895, opened 2025-09-12
Support Longcat flash Model

zhanghw0354:
Support Longcat flash Model

No

- vLLM version: v0.10.2


wangxiyuan:
Thanks for the contribution. It's great. While maybe you can find that vllm-ascend is refactored a lot recently, since we'd like to remove model stracuture mantainance and reuse the model in vLLM totally. So you should do the model support in the new way. I.e. just register oot ops to replace the ascend implement dynamicly like others do now.

PR #2906, opened 2025-09-13
Enable push trigger for image job

Yikun:
Enable push trigger for image job

No

CI passed

Followup on https://github.com/vllm-project/vllm-ascend/pull/2864


Yikun:
image

Confirm this worked as expected.

PR #2910, opened 2025-09-13
Bugfix_091dev_OOM

wangxiaoteng888:
Resolving the issue of the chunked work size being too large, which leads to D node out-of-memory (OOM) errors.

no

by ci

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -189,6 +189,10 @@ class AscendMLAMetadataBuilder:
                            self.block_size - 1) // self.block_size
         self.chunked_prefill_enabled = scheduler_config.chunked_prefill_enabled
         if self.chunked_prefill_enabled:
+            if vllm_config.kv_transfer_config.is_kv_consumer:
+                max_chunked_size = scheduler_config.max_num_seqs * self.block_size
+            else:
+                max_chunked_size = 128 * 1024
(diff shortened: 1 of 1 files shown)


vllm_ascend/attention/mla_v1.py
```diff
@@ -189,6 +189,10 @@ def __init__(self,
                            self.block_size - 1) // self.block_size
         self.chunked_prefill_enabled = scheduler_config.chunked_prefill_enabled
         if self.chunked_prefill_enabled:
+            if vllm_config.kv_transfer_config.is_kv_consumer:
+                max_chunked_size = scheduler_config.max_num_seqs * self.block_size
+            else:
+                max_chunked_size = 128 * 1024
```
momo609, line 196: const variables should be defined at the front of file with bigger case, and don't forget to note it also.

vllm_ascend/attention/mla_v1.py
```diff
@@ -189,6 +189,10 @@ def __init__(self,
                            self.block_size - 1) // self.block_size
         self.chunked_prefill_enabled = scheduler_config.chunked_prefill_enabled
         if self.chunked_prefill_enabled:
+            if vllm_config.kv_transfer_config.is_kv_consumer:
+                max_chunked_size = scheduler_config.max_num_seqs * self.block_size
```
weijinqian0, line 194: Is the OOM caused by the batch size being greater than 1024?

PR #2920, opened 2025-09-14
[Doc] eval-type not support service but server, and test the result b…

L4-1024:
…y online inference


zhangxinyuehfad:
In the latest version of evalscope, the parameter `--eval-type service` is invalid and needs to be changed to `--eval-type server`.

PR #2931, opened 2025-09-15
[1/N][Feat] Cut down memory usage for o_proj in DeepSeek

clrs97:
To cut down the memory usage of large weight matrices, we often rely on various linear operations:
- `ReplicatedLinear`: Stores the entire matrix, consuming excessive memory.
- `RowParallelLinear`: Requires an `all_reduce` to merge answer, introducing additional communication overhead and potential accuracy loss. Each token is handled across multiple devices rather than a single device, which is undesirable in SP scenario.
- ...

Furthermore, in multi-way Data Parallelism (DP) configurations, layers typically store redundant weight copies.

This PR introduces a shared-weight plugin for layers inheriting from `LinearBase`. It offers the following advantages:
- It evenly distributes a set of layers with identical structures across devices. Each layer retains its complete weights, eliminating redundant memory usage.
- It supports asynchronous broadcasting to prefetch weights for upcoming layers.
- It preserves the custom `process_weights_after_loading()` method to make keeping NZ format possible.
- It is compatible with any linear class that inherits from `LinearBase`, thereby preserving all the features of the original linear implementation.

No

vLLM main: https://github.com/vllm-project/vllm/commit/f4a948f33f8766c843e0d59d9ec1ab44f7b2bfcf

- vLLM version: v0.10.2


vllm_ascend/ops/layer_shard_linear.py
```diff
@@ -0,0 +1,313 @@
...
+                    dispose_tensor(layer.weight)
+
+        dispose_tensor(self.dummy_weight)
+
+    def get_shared_window(self, layer_idx: int):
+        assert self.shared_windows
+        return self.shared_windows[self.layers[layer_idx -
+                                               self.start_layer].window_idx]
+
+    def reach_layer(self, layer_idx: int):
```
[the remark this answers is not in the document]
clrs97: Single layer is not suggested to use this ops.

vllm_ascend/ops/layer_shard_linear.py
```diff
@@ -0,0 +1,313 @@
...
+    end_layer: int,
+    prefetch_step: int,
+    layer_idx: int,
+    layer: LinearBase,
+) -> ClusterMetadata:
+    global _cluster_dict
+    if name not in _cluster_dict:
+        num_layers = end_layer - start_layer
+        assert num_layers > 0
+        assert prefetch_step >= 0 and prefetch_step <= num_layers - 2
```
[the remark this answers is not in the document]
clrs97: When prefetch_step == num_layers - 1, it is rather better to use ReplicatedLinear.

vllm_ascend/ops/layer_shard_linear.py
```diff
@@ -0,0 +1,313 @@
...
+    cluster = _cluster_dict[name]
+    assert layer.quant_method is not None
+    cluster.layers[layer_idx - start_layer] = LayerMetadata(
+        layer=layer,
+        post_method=layer.quant_method.process_weights_after_loading,
+        weight=layer.weight,
+        window_idx=-1,
+    )
+    # Discard the original `process_weights_after_loading` method such that it won't be called by others.
+    layer.quant_method.process_weights_after_loading = lambda layer: None
```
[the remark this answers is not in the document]
clrs97: It will be called later.

vllm_ascend/ops/layer_shard_linear.py
```diff
@@ -0,0 +1,313 @@
...
+class SharedWindowMetadata:
+    """Metadata for a shared window.
+    """
+    weight: torch.Tensor  # The weight tensor to be shared by layers.
+    data_layer_idx: int  # The index of the layer this window's weight is equal to.
+    work: Optional[torch.distributed.Work]  # The asynchronous broadcast work.
+
+
+@dataclass
+class ClusterMetadata:
```
CalvinXKY: Use "Instance" in place of "Cluster"？

vllm_ascend/ops/layer_shard_linear.py
```diff
@@ -0,0 +1,313 @@
...
+
+@CustomOp.register("layer_shard_linear")
+class LayerShardLinear(LinearBase):
+    """Linear layer with sharding storage.
+
+    Each device in the parallel group evenly stores a set of disjoint layers. All layers must have the same structure. Assuming there are n devices, the weight matrix of the i-th layer will be stored on the (i % n)-th device.
+
+    After loading the model, you must call `post_process_after_loading_for_cluster()` to complete the initialization.
+
+    Each time a new layer is reached, you must call `reach_layer()` to prefetch the weights.
```
CalvinXKY: Maybe we should add an example of this calling relationship to the tests.

vllm_ascend/ops/layer_shard_linear.py
```diff
@@ -0,0 +1,313 @@
...
+        bias: bool = True,
+        skip_bias_add: bool = False,
+        params_dtype: Optional[torch.dtype] = None,
+        quant_config: Optional[QuantizationConfig] = None,
+        prefix: str = "",
+        *,
+        return_bias: bool = True,
+        cluster_name: str,
+        group: GroupCoordinator,
+        start_layer: int,
```
CalvinXKY: Default values are used for start_layer and end_layer if not set.

vllm_ascend/ops/layer_shard_linear.py
```diff
@@ -0,0 +1,313 @@
...
+        end_layer: int,
+        layer_idx: int,
+        prefetch_step: int = 0,
```
CalvinXKY: The "prefetch_step" setting needs a clearer description for users. Using tests?

vllm_ascend/ops/layer_shard_linear.py
```diff
@@ -0,0 +1,313 @@
...
+            f"to a parameter of size {param.size()}")
+        param.data.copy_(loaded_weight)
+
+    def post_process_after_loading_for_cluster(self):
+        self.cluster.post_process_after_loading()
+
+    def reach_layer(self):
+        self.cluster.reach_layer(self.layer_idx)
+
+    def forward(
```
CalvinXKY: TODO: 
It should be considered to implement a default mechanism, or implicit invocation, for these two functions: `post_process_after_loading_for_cluster()`  and  `reach_layer()

vllm_ascend/ops/layer_shard_linear.py
clrs97: What about "stack"?" The model comprises a stack of 100 identical linear layers B (16384x7168), each sandwiched between computations A and C within a block."

clrs97: Finally changed to "Series".

vllm_ascend/ops/layer_shard_linear.py
clrs97: Updated.

vllm_ascend/ops/layer_shard_linear.py
clrs97: It is risky. For example, assume in dummy run, there might or not be forward of this ops, which will cause communication error.

PR #2944, opened 2025-09-16
add test_patch_distributed.py

yangqinghao-cmss:


wangxiyuan:
can you rebase to main to let the CI pass? Thanks.

wangxiyuan:
patch file is not need for test. Since this is just a copy from vLLM and we'll remove it in the future.

PR #2681, opened 2025-09-01
[bugfix][torchair] fix multistream_moe problems in torchair graph mode

linfeng-yuan:
### What this PR does / why we need it?
This pr fixes two problems while `multistream_moe` enabled in torchair graph mode:
1. check `TorchairAscendW8A8DynamicFusedMoEMethod` instead of  incorrect `AscendW8A8DynamicFusedMoEMethod` 
2. mc2_mask should be chunked no matter `replace_allreduce` is True or False in forward function of `TorchairAscendFusedMoE`

### Does this PR introduce _any_ user-facing change?
No.

### How was this patch tested?
e2e vllm serving with `multistream_moe` (DP32TP1EP32)

- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/0fb2551c238c7ccbcf6f25ef4646ce6c92f684d1

PR #2946, opened 2025-09-16
[Feature] Support moe multi-stream for aclgraph.

whx-sjtu:
This PR puts the calculation of shared experts into a separate stream, overlaping with routing experts.


wangxiyuan:
1. Add e2e test
2. set the default value to True

whx-sjtu:
> 2. set the default value to True

This will be done in PR #2681

PR #2962, opened 2025-09-16
[Perf] Add new npu_fused_infer_attention_score op to improve perfomance in splitfuse cases and resolve long-seq mask problems

qyqc731:
Add new npu_fused_infer_attention_score op to improve perfomance in splitfuse cases and resolve long-seq mask problems .

1. The original op's performance is suboptimal in certain scenarios, necessitating optimization through the _new op_ (npu_fused_infer_attention_score)。
2. For ultra-long sequences (128k), the original operator will allocate a large attn_mask, which consumes excessive CPU memory. In contrast, the _new op_ supports a fixed-size compressed mask, effectively resolving this issue.

NOTE1: The current PR retains the original logic and uses a version check of the CANN package to determine whether the _new op_ can be enabled. This ensures no impact on existing users. In future versions, this version check and the original logic will be deprecated, and the _new op_ scheduling will be uniformly adopted.
NOTE2: This pr relies on future CANN version, which is not available now.
NOTE3: To enable the new op in chunked prefill, the parameter additional_config should be set like `--additional-config '{"ascend_scheduler_config": {"enabled":true,"enable_chunked_prefill":true}}' \` at least.

No
CI passed

- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/6c5f82e5aa87cd73ce03ce10fc44138f75ee1aea

### DIFF
--- a/vllm_ascend/attention/attention_mask.py
+++ b/vllm_ascend/attention/attention_mask.py
@@ -82,7 +82,7 @@ class AttentionMaskBuilder:
         dtype: torch.dtype = None,
         device: torch.device = None,
     ) -> torch.Tensor:
-        if self.device:
+        if torch.version.cann.startswith("8.3"):
             return self.chunked_prefill_attn_mask
         else:
             if dtype not in [torch.float16, torch.bfloat16]:
--- a/vllm_ascend/attention/attention_mask.py
+++ b/vllm_ascend/attention/attention_mask.py
@@ -39,12 +39,13 @@ class AttentionMaskBuilder:
         self,
         max_seq_len: int,
         dtype: torch.dtype,
+        device: torch.device,
     ):
         attn_mask = _generate_attn_mask(max_seq_len, dtype)
 
         self._seq_len_cached = attn_mask.shape[0]
         self.attn_mask_cache = attn_mask
-        self.chunked_prefill_attn_mask = torch.triu(torch.ones(2048, 2048), diagonal=1).to(torch.int8)
+        self.chunked_prefill_attn_mask = torch.triu(torch.ones(2048, 2048), diagonal=1).to(torch.int8).to(device)
 
     @staticmethod
     def get_mask_scale_factor(dtype: torch.dtype = torch.float16):
@@ -67,9 +68,8 @@ class AttentionMaskBuilder:
 
     def get_splitfuse_attn_mask(
         self,
-        device: torch.device,
     ) -> torch.Tensor:
-        return self.chunked_prefill_attn_mask.to(device)
+        return self.chunked_prefill_attn_mask
 
     def _update_attn_cache(self, seqlen: int, dtype: torch.dtype):
         if seqlen > self._seq_len_cached:
--- a/vllm_ascend/attention/attention_mask.py
+++ b/vllm_ascend/attention/attention_mask.py
@@ -41,7 +41,7 @@ class AttentionMaskBuilder:
         dtype: torch.dtype,
         device: torch.device = None,
     ):
-        # NOTE: The device argument specifies the target NPU 
+        # NOTE: The device argument specifies the target NPU
         # to be used for the newly added FIA operator.
         # Only pass this parameter when using the new FIA operator.
 
@@ -52,8 +52,9 @@ class AttentionMaskBuilder:
         self.device = device
         if torch.version.cann.startswith("8.3"):
             assigned_mask_dim = 2048
-            self.chunked_prefill_attn_mask = torch.triu(torch.ones(assigned_mask_dim, assigned_mask_dim), diagonal=1
-            ).to(torch.int8).to(device)
+            self.chunked_prefill_attn_mask = torch.triu(
+                torch.ones(assigned_mask_dim, assigned_mask_dim),
+                diagonal=1).to(torch.int8).to(device)
 
     @staticmethod
     def get_mask_scale_factor(dtype: torch.dtype = torch.float16):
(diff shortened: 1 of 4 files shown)


rjg-lyh:
Could you integrate other scenarios, such as full FlashAttention, using the FIA interface as well, and provide the performance test results?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -293,7 +293,7 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
         )
 
         self.attn_mask_builder = AttentionMaskBuilder(
-            self.model_config.max_model_len, self.dtype)
+            self.model_config.max_model_len, self.dtype, self.device)
```
Angazenn: If we pass `self.model_config.max_model_len` here we still have an attention mask of shape [max_model_len, max_model_len]. We may have 2 choices:
1. Passing `self.scheduler_config.max_num_batched_tokens` instead of `self.model_config.max_model_len`.
2. Also using new attention op for full prefill case.

vllm_ascend/attention/attention_v1.py
```diff
@@ -455,18 +455,25 @@ def _forward_v1_style(
...
+            key=key,
+            value=value,
+            atten_mask=attn_metadata.attn_mask.to(device=query.device),
             block_table=attn_metadata.block_tables,
-            seq_len=attn_metadata.query_lens,
-            context_lens=attn_metadata.seq_lens,
-            num_kv_heads=self.num_kv_heads,
+            input_layout="TND",
+            block_size=block_size,
+            actual_seq_lengths=attn_metadata.query_start_loc[1:],
```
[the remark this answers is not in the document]
Angazenn: plz consider use `attn_metadata.query_lens`

vllm_ascend/worker/model_runner_v1.py
Angazenn: for choice 1:
```suggestion
            self.scheduler_config.max_num_batched_tokens, self.dtype, self.device)
```

Angazenn:
Also notice another 2 points:
1. compatibility with difference torch_npu/cann version.
2. fix bug when padding is introduced in dp (len of query exceeds attn_metadata.query_lens[-1]).

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -797,8 +797,7 @@ def _make_attention_mask(self, seq_lens, position,
                              attn_state) -> torch.Tensor:
         # Chunk Prefill situation.
         if attn_state == AscendAttentionState.ChunkedPrefill and not self.vllm_config.model_config.use_mla:
-            return self.attn_mask_builder.get_splitfuse_attn_mask(
-                seq_lens, position, self.dtype, self.device)
+            return torch.triu(torch.ones(2048, 2048), diagonal=1).to(torch.int8)
```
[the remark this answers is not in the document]
Angazenn: this is because new op support compressed mask

vllm_ascend/attention/attention_v1.py
```diff
@@ -455,18 +458,39 @@ def _forward_v1_style(
...
-            value_cache=self.value_cache,
-            mask=attn_metadata.attn_mask,
-            block_table=attn_metadata.block_tables,
-            seq_len=attn_metadata.query_lens,
-            context_lens=attn_metadata.seq_lens,
-            num_kv_heads=self.num_kv_heads,
-            num_heads=self.num_heads,
-            scale_value=self.scale,
-            out=output)
+        if self.compressed_mask:
```
qyqc731: add TODO: This op will be used in more situation in future.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -797,8 +804,11 @@ def _make_attention_mask(self, seq_lens, position,
                              attn_state) -> torch.Tensor:
         # Chunk Prefill situation.
         if attn_state == AscendAttentionState.ChunkedPrefill and not self.vllm_config.model_config.use_mla:
-            return self.attn_mask_builder.get_splitfuse_attn_mask(
-                seq_lens, position, self.dtype, self.device)
+            if selkf.compressed_mas:
+                return self.attn_mask_builder.get_splitfuse_attn_mask()
```
weijinqian0, line 827: code bug

vllm_ascend/utils.py
```diff
@@ -617,3 +617,28 @@ def weak_ref_tensors(
     if isinstance(tensors, tuple):
         return tuple(weak_ref_tensor(t) for t in tensors)
     raise ValueError("Invalid type for tensors")
+
+
+def verify_torch_npu_version(
```
qyqc731: The cann version needs to be verfied. using package.version

qyqc731: add todo for temp

vllm_ascend/attention/attention_mask.py
```diff
@@ -39,11 +39,19 @@ def __init__(
...
 
         self._seq_len_cached = attn_mask.shape[0]
         self.attn_mask_cache = attn_mask
+        self.device = device
+        if self.device:
+            #NOTE: New compressed mask needs to be sent to certain device, 
+            # so device needs to be passed here.
+            assigned_mask_dim = 2048
+            self.chunked_prefill_attn_mask = torch.triu(torch.ones(assigned_mask_dim, assigned_mask_dim), diagonal=1
+            ).to(torch.int8).to(device)
```
qyqc731: check type again

vllm_ascend/attention/attention_mask.py
```diff
@@ -39,11 +39,19 @@ def __init__(
         self,
         max_seq_len: int,
         dtype: torch.dtype,
+        device: torch.device = None,
```
MengqingCao, line 42: Let's make some comments to describe what the `device` mean, and make clear that we should only pass by this arg when using the new attention op

vllm_ascend/attention/attention_mask.py
Angazenn: this op currently only supports int8 mask

vllm_ascend/worker/model_runner_v1.py
Angazenn: fixed

vllm_ascend/attention/attention_mask.py
qyqc731: Added

PR #2964, opened 2025-09-16
[bugfix] fix shared expert dp with hybrid kvcache

linfeng-yuan:
https://github.com/vllm-project/vllm-ascend/pull/2849 moves the implementation of `shared_expert_dp` to torchair deepseek_modeling. However, the calling of  `set_forward_context` with `enforce_eager` and `shared_expert_dp` falls back to the implementation of model_runner_v1.py and set the global attn_metadata as a dictionary. It leads to a RuntimerError when attn_metadata is got from the forward context and used in torchair_deepseek_v2.py. This PR fixes this problem by introducing the transformation of attn_metadata in this file. 

Note that current E2E testing lacks the case of deepseek with `shared_expert_dp`. We need to add an ST with `shared_expert_dp` in testing workflow.

No.
e2e vllm serving with `enable_shared_expert_dp: true` passed.

- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/de3e53a75ba9f31f446926911b7c44561af3b2ee

### DIFF
--- a/vllm_ascend/torchair/models/torchair_deepseek_v2.py
+++ b/vllm_ascend/torchair/models/torchair_deepseek_v2.py
@@ -813,6 +813,8 @@ class TorchairDeepseekV2DecoderLayer(DeepseekV2DecoderLayer):
             residual = get_tp_group().all_gather(residual, 0)
 
             attn_metadata = get_forward_context().attn_metadata
+            if attn_metadata is not None and isinstance(attn_metadata, dict):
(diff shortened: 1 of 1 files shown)


wangxiyuan:
please update the commit message to explain why e2e test passed and should we update the e2e as well?

linfeng-yuan:
> please update the commit message to explain why e2e test passed and should we update the e2e as well?

I've updated the commit message and plan to add this ST before this weekend~

wangxiyuan (approved):

PR #2972, opened 2025-09-17
[Bugfix] ngram spec decode attention error and repeat add sampled tok…

wxsIcey:
- Fixes  ngram spec decode attention error(https://github.com/vllm-project/vllm-ascend/issues/2971) and repeat add sampled token ids

N/A


wangxiyuan:
it's good to add a e2e test as well.

wxsIcey:
N-gram spec decode currently has accuracy issues

PR #2995, opened 2025-09-17
add a case

jiangyunfan1:


zhangxinyuehfad:
You'd better add a change module so that when your YAML or test file is modified, testing can be conducted。

tests/smoke_test/test_qwen3_int8.py
```diff
@@ -0,0 +1,48 @@
...
+    "model": "Qwen3",
+    "prompt": "San Francisco is a",
+    "max_tokens": 10,
+    "temperature": 0
+}
+
+
+def test_vllm_aclgraph_qwen3_32b_server_A2():
+    script_path = "tests/smoke_test/qwen3_32b/run_dp_server_qwen3_32B_aclgraph.sh"
+    output_file = "qwen3_32b_int8_output.txt"
```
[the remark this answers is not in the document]
jiangyunfan1: It is unique.

tests/smoke_test/qwen3_32b/run_dp_server_qwen3_32B_aclgraph.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+set -euo pipefail
+
+weight_dir="slanyer/Qwen3_32B_INT8"
```
Yikun, line 4: All models should be accessible, not local models at private addresses

tests/smoke_test/test_qwen3_int8.py
```diff
@@ -0,0 +1,48 @@
...
+DATA = {
+    "model": "Qwen3",
+    "prompt": "San Francisco is a",
+    "max_tokens": 10,
+    "temperature": 0
+}
+
+
+def test_vllm_aclgraph_qwen3_32b_server_A2():
+    script_path = "tests/smoke_test/qwen3_32b/run_dp_server_qwen3_32B_aclgraph.sh"
```
Yikun, line 17: I don't think this is an accept way to have test.

Try to take a look on this:
[1] online test: https://github.com/vllm-project/vllm/blob/36429096171ff8785645c40c662d859dddedd829/tests/utils.py#L81
[2] offline test: https://github.com/vllm-project/vllm-ascend/tree/main/tests/e2e/models/configs

tests/smoke_test/test_qwen3_int8.py
```diff
@@ -0,0 +1,48 @@
...
+HEADERS = {"Content-Type": "application/json"}
+DATA = {
+    "model": "Qwen3",
+    "prompt": "San Francisco is a",
+    "max_tokens": 10,
+    "temperature": 0
+}
+
+
+def test_vllm_aclgraph_qwen3_32b_server_A2():
```
Yikun, line 16: Theoretically, A2 and A3 should be reusable, so we should increase their reusability as much as possible. If you still think there is a problem, please see markers

[1] https://docs.pytest.org/en/stable/example/markers.html#working-with-custom-markers

tests/smoke_test/qwen3_32b/run_dp_server_qwen3_32B_aclgraph.sh
```diff
@@ -0,0 +1,27 @@
...
+
+export TASK_QUEUE_ENABLE=1
+export VLLM_USE_V1=1
+export OMP_PROC_BIND=false
+#export ASCEND_RT_VISIBLE_DEVICES=0,1
+# AIV
+export HCCL_OP_EXPANSION_MODE="AIV"
+# MASK
+#export PAGED_ATTENTION_MASK_LEN= 5500
+#图模式--aclgraph
```
Yikun, line 14: Please use english

tests/smoke_test/test_qwen3_int8.py
```diff
@@ -0,0 +1,48 @@
...
+    script_path = "tests/smoke_test/qwen3_32b/run_dp_server_qwen3_32B_aclgraph.sh"
+    output_file = "qwen3_32b_int8_output.txt"
+    server_proc = None
+    output_fp = None
+    try:
```
Yikun, line 21: Please try to make this call reusable. rather than repeat any where.

Yikun:
Thanks for your contributions. We strongly encourage contributions to testing, but I think the code framework still has a lot of room for improvement, rather than `emm, this test works`.

Because once the test code is merged, all community developers will follow your implementation. The first version of the framework code bears a great responsibility for the maintainability of subsequent code. So please forgive our strictness.

PR #3007, opened 2025-09-18
[HybridKV] Fix prefill disaggregation kvcache addr alignment & use hybrid kv cache only when running qwen3_next

MengqingCao:
This pr fixes a few issues on prefill disaggregation:
1. Fix prefill disaggregation kvcache addr alignment issue, llmdatadist needs the addr of tensors to be aligned with 2M
2. Fix prefill disaggregation kvcache shape error, llmdatadist requires k/v tensors with shape [num_blocks, ...], however the implentment before this pr is  [2, num_blocks, ...], which will break prefill disaggregation
3. Use hybrid kv cache only when running qwen3_next to fix accuracy issue on prefill disaggregation.

N/A

Tested locally by @liziyu179 

- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/4f02b77de4e794a0d417ed98a26884208f75e043

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2635,7 +2635,7 @@ class NPUModelRunner(LoRAModelRunnerMixin):
                 # TODO: remove this after the OOM issue is located and fixed, otherwise, some model may
                 # encounter OOM issue
                 if isinstance(kv_cache_spec, FullAttentionSpec):
-                    raw_k_tensor, raw_v_tensor = kv_cache_raw_tensors[
+                    raw_k_tensor, raw_v_tensor = kv_cache_raw_tensors[  # type: ignore
                         layer_name]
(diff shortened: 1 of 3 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2713,7 +2765,9 @@ def may_reinitialize_input_batch(self,
                     attn_groups = self.attn_groups[kv_cache_group_id]
                 except IndexError:
                     attn_groups = None
-                if attn_groups:
+                # TODO: Refactor the hard code of qwen3_next when we could get `use_hybrid_blocks`
+                # from vllm_config or anywhere others before this
+                if attn_groups and self.model_config.hf_config.model_type == "qwen3_next":
```
MengqingCao: check if use hybrid_block through `len(attn_groups)`

PR #3016, opened 2025-09-18
[Fixbug] Fix accuracy for DeepSeek-V2-Lite

zhangxinyuehfad:

Fix accuracy for DeepSeek-V2-Lite


Yikun (approved):

Yikun:
Let's merge to recover accuracy test first

PR #3021, opened 2025-09-19
[3/N][Refactor][Quantization]remove packed_modules_mapping from models

22dimensions:
### What this PR does / why we need it?

Some custom models in vllm-ascend define packed_modules_mapping, which prevent keeping same model class with vllm community. So move these custom packed_modules_mapping to quant utils.py. After this pr, some custom models can be removed.

### Does this PR introduce _any_ user-facing change?

tested by CI

### How was this patch tested?

tested by CI

- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/5089fd749cbe4233a29f29ce706d56c47464c117

PR #3038, opened 2025-09-19
Support W8A8_dynamic on Step3 Model

yifeilin1015:
###What this PR does / why we need it?
Support w8a8_dynamic for Step3 model

###Does this PR introduce any user-facing change?
No user-facing change

###How was this patch tested?
This patch is tested with following datasets:
MMMU: 73.11， MMSTAR: 70.68, AIME25: 80.00， GPQA-diamond: 72.73

###Modify points:
- add step3_text.py file with new load_weights method
- load_weights method only change the way of loading MoE experts compared to the original method

vLLM version: v0.10.2

### DIFF
--- a/vllm_ascend/models/step3_text.py
+++ b/vllm_ascend/models/step3_text.py
@@ -1,0 +1,153 @@
+from typing import Dict, Iterable, Tuple
+from vllm.model_executor.models.step3_text import Step3TextForCausalLM
+import torch
+from vllm.model_executor.layers.fused_moe import FusedMoE
+from vllm.model_executor.models.utils import is_pp_missing_parameter
+from vllm.model_executor.model_loader.weight_utils import default_weight_loader
+from vllm.config import VllmConfig
+
+class CustomStep3TextForCausalLM(Step3TextForCausalLM):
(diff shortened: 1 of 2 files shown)


vllm_ascend/models/step3_text.py
```diff
@@ -0,0 +1,153 @@
...
+import torch
+from vllm.model_executor.layers.fused_moe import FusedMoE
+from vllm.model_executor.models.utils import is_pp_missing_parameter
+from vllm.model_executor.model_loader.weight_utils import default_weight_loader
+from vllm.config import VllmConfig
+
+class CustomStep3TextForCausalLM(Step3TextForCausalLM):
+    experts_ = [f"experts.{i}.{proj}" for i in range(48) for proj in ("down_proj", "gate_proj", "up_proj")]
+    
+    packed_modules_mapping = {
```
realliujiaxu, line 12: this should be moved this to [quant_config.py](https://github.com/vllm-project/vllm-ascend/pull/3021/files#diff-1d9cd3d7a19e647d95492fe714f07f1b582c2903ad5df49377300d6abadb1b21), after https://github.com/vllm-project/vllm-ascend/pull/3021/files is merged

wangxiyuan:

PR #3042, opened 2025-09-19
FlashLB algorithm

Mercykid-bash:
# Feature: FlashLB algorithm

## Purpose

This Pull Request enhances the EPLB (Expert Parallelism Load Balancing) system by introducing a novel load balancing algorithm: FlashLB.
1. The default algorithm adopts two separate sub-procedures to optimize expert replication and placement independently:

    a.  **Expert Replica Allotment Sub-procedure** : Determines the number of replicas for all experts. At each step, it greedily adds one more replica to the expert with the highest per-replica load, aiming to minimize load skew at the expert replica granularity (Min Max Replica, MMR).

    b.  **Expert Replica Placement Sub-procedure** : Distributes all replicas across devices. First, it sorts the generated replicas in descending order of hotness, then iteratively places the currently hottest replica onto the device with the lowest cumulative load and available slots.
    However, this simplistic combination of two separate procedures lacks synergy and often leads to sub-optimal load balancing. For example, in the simple scenario illustrated below: Given 8 logical experts with hotness values [600, 560, 120, 120, 20, 10, 10, 10], and 2 replicas allocated per device across 8 devices, the default EPLB algorithm results in a maximum per-device hotness of 232 (peak-average load ratio 1.28), while our proposed FlashLB algorithm reduces this value to 205 (peak-average load ratio 1.13).

      <figure>

2. The default algorithm simply aggregates hotness measurements across the entire profiling window. While this provides a coarse approximation of the hotness distribution, it fails to capture the time-phased variations and temporal correlations in expert hotness (both within and between experts) across iterations—phenomena that have been observed in real-world scenarios. Such single-point hotness estimation degrades the solution quality of the load balancing algorithm.

3. The default algorithm regularly recalculates updated expert placement results for all layers without discrimination. Considering that excessive expert updates can impact Service Level Objectives (SLOs), such full-scale redeployment leads to excessively high adjustment overhead, which negatively affects end-to-end performance.

### 1. Joint Optimization of Replica Allotment and Placement

FlashLB achieves joint optimization of replica allotment and placement through a novel tree search approach, combined with carefully designed e Fl fficient pruning and lightweight look-ahead estimation. We partition all experts into several subsets, and for each subset, hierarchically determine the optimal replica count and placement. Leveraging efficient pruning and lightweight look-ahead estimation, the process consistently aims to optimize the globally expected inter-device load balance degree (considering both deployed and unexplored experts) while ensuring sufficient computational efficiency. Additionally, precompilation techniques are employed for acceleration, delivering load balancing that is both high-quality and practically efficient.
### 2. Multi-Episode Enhancement

Instead of performing full-duration averaging like the default algorithm, FlashLB partitions each profiling interval (e.g., 1024 iterations) into multiple consecutive smaller episodes (e.g., 16 iterations). This preserves hotness fluctuation and correlation information. It then constructs a multi-objective optimization problem to co-optimize these episodes simultaneously, enabling adaptability to interleaved hotness patterns and improving statistical robustness.

### 3. Layer-wise Cherry-Picking Redeployment

To reduce the overhead of frequent expert redeployment, FlashLB introduces a cherry-picking redeployment scheme. During each algorithmic decision cycle, it real-time tracks load balance degree of all layers and triggers expert placement updates only for those layers whose peak-average ratio exceeds a predefined threshold. This avoids unnecessary redeployment for stable layers, significantly reducing adjustment overhead and thereby improving end-to-end performance gains.

## Co-author:

- vLLM version: v0.10.2


vllm_ascend/eplb/core/policy/policy_flashlb.py
```diff
@@ -0,0 +1,611 @@
...
+    exam_config.num_die_per_host = 16
+    algo = FlashLB(exam_config)
+    # Generate target tensor
+    expert_tensor = generate_layered_experts(num_layers=58, layer_shape=(32, 9))
+
+    algo.rebalance_experts(expert_tensor, torch.randint(1, 1000, (58, 32, 9)))
+    FlashLB.par_history = defaultdict(float)
+    FlashLB.hotness_window = {}
+
+warm_up()
```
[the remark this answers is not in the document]
Mercykid-bash: The warm-up call at the module level is intentionally designed to trigger JIT compilation during import rather than at inference time. This approach ensures that numba's just-in-time compilation overhead is incurred upfront during module loading, preventing unexpected latency spikes during actual inference operations. 

While module imports may take slightly longer, this trade-off is necessary to guarantee consistent performance during runtime, where low and predictable latency is critical for the load balancing system's effectiveness. Explicitly deferring this warm-up to application initialization could lead to missed compilation steps and subsequent runtime delays if the initialization sequence is not strictly followed.

The current implementation prioritizes inference-time performance predictability over import speed, which aligns with the requirements of latency-sensitive load balancing operations.

vllm_ascend/eplb/core/policy/policy_flashlb.py
```diff
@@ -0,0 +1,654 @@
...
+    exam_config.num_die_per_host = 16
+    algo = FlashLB(exam_config)
+    # Generate target tensor
+    expert_tensor = generate_layered_experts(num_layers=58,
+                                             layer_shape=(32, 9))
+
+    algo.rebalance_experts(expert_tensor, torch.randint(1, 1000, (58, 32, 9)))
+
+
+warm_up()
```
wangxiyuan: Thanks for the contribution. Only one question: `warm_up` should be called when import the module? is there any way to let it only runs when the policy is set?

Mercykid-bash: Another approach is to incorporate a check within `policy_factory.py` that triggers the `warm_up` operation **exclusively** when the `RealTimeLB` policy is selected.  

This way, the `warm_up` process will only run if and when the `RealTimeLB` policy is actually instantiated (i.e., when this specific policy is set), avoiding unnecessary execution for other policy types.  

Would this align with your intended implementation? Please let me know if you need further adjustments.

class PolicyFactory:

    @staticmethod
    def generate_policy(policy_type: int, config: DynamicConfig) -> EplbPolicy:
        policy = {
            # Constraint applying Dynamic EPLB policy V2:
            # If there exists redundant expert:
            # only one redundant expert can be placed in one NPU and its physical expert index must be 0

            # Applying greedy d2d expert weight update composing
            0:
            RandomLoadBalance,  # RandomLoadBalance: shuffle last physical expert on NPU 1 and 3
            1:
            DynamicEplb,  # Dynamic EPLB policy: overall expert replacement based on current moe load
            2:
            DynamicEplbV2,  # Dynamic EPLB policy V2:  expert replacement with constrained number of expert shuffle
            3:
            FlashLB,  # FlashLB EPLB policy: expert replacement based on Joint Optimization, Multi-Shot Enhancement and Incremental Adjustment
        }
        policy_class = policy.get(policy_type, RandomLoadBalance)
        policy_instance = policy_class(config)
        if policy_type == 3:
            policy_instance.warm_up()
        return policy_instance

PR #3048, opened 2025-09-20
[Feature] Reduce host memory usage for attention mask generation

jianzs:

Previously, the mask construction process created multiple tensors of size (max_model_len, max_model_len). When max_model_len reached 128k, single GPU host memory usage exceeded hundreds of GB, causing process OOM crashes. This update optimizes the mask generation to significantly reduce memory consumption.

No.

CI pass.

### DIFF
--- a/vllm_ascend/attention/attention_mask.py
+++ b/vllm_ascend/attention/attention_mask.py
@@ -17,9 +17,9 @@ import torch
 
 def _generate_attn_mask(max_seq_len, dtype):
     # Construct lower triangle matrix.
-    mask_flag = torch.tril(
-        torch.ones((max_seq_len, max_seq_len),
-                   dtype=torch.bool)).view(max_seq_len, max_seq_len)
+    mask_flag = torch.ones(
+        (max_seq_len, max_seq_len),
+        dtype=torch.bool).tril_().view(max_seq_len, max_seq_len)
     # Create upper triangle matrix used to mark mask positions.
     mask_flag = ~mask_flag
     # Currently for fp16 dtype, the mask value should be set to -inf.
@@ -28,8 +28,8 @@ def _generate_attn_mask(max_seq_len, dtype):
         mask_value = torch.finfo(torch.float32).min
     else:
         mask_value = 1
-    attn_mask = torch.masked_fill(torch.zeros(size=(max_seq_len, max_seq_len)),
-                                  mask_flag, mask_value).to(dtype)
+    attn_mask = torch.zeros(size=(max_seq_len, max_seq_len), dtype=dtype) \
+        .masked_fill_(mask_flag, mask_value)
     return attn_mask
 
 


wangxiyuan:
@rjg-lyh

rjg-lyh:
Thank you for your PR. However, I have two questions:

First, this PR still caches the (max_seq_len, max_seq_len) mask, which theoretically still occupies the same amount of host memory and does not resolve the OOM issue.

The second point is that during actual inference, the main cause of performance degradation is the repetitive host-to-device operations during the construction of the splitfuse mask, which are too time-consuming.

I would greatly appreciate it if you could take a look at these points. @jianzs

jianzs:
> Thank you for your PR. However, I have two questions:
> 
> First, this PR still caches the (max_seq_len, max_seq_len) mask, which theoretically still occupies the same amount of host memory and does not resolve the OOM issue.
> 
> The second point is that during actual inference, the main cause of performance degradation is the repetitive host-to-device operations during the construction of the splitfuse mask, which are too time-consuming.
> 
> I would greatly appreciate it if you could take a look at these points. @jianzs

This PR isn't focused on runtime performance. 

It's all about fixing the high GPU memory spike during attention mask generating. The issue was that we weren't reusing GPU memory efficiently, which caused the memory usage to balloon to several times what the final mask actually needs. We've fixed that by implementing proper memory reuse during the generation process.

rjg-lyh:
> > Thank you for your PR. However, I have two questions:
> > First, this PR still caches the (max_seq_len, max_seq_len) mask, which theoretically still occupies the same amount of host memory and does not resolve the OOM issue.
> > The second point is that during actual inference, the main cause of performance degradation is the repetitive host-to-device operations during the construction of the splitfuse mask, which are too time-consuming.
> > I would greatly appreciate it if you could take a look at these points. @jianzs
> 
> This PR isn't focused on runtime performance.
> 
> It's all about fixing the high GPU memory spike during attention mask generating. The issue was that we weren't reusing GPU memory efficiently, which caused the memory usage to balloon to several times what the final mask actually needs. We've fixed that by implementing proper memory reuse during the generation process.

But this cached mask is created on CPU, i can't understand this pr how to resolve the OOM issue.

jianzs:
> > > Thank you for your PR. However, I have two questions:
> > > First, this PR still caches the (max_seq_len, max_seq_len) mask, which theoretically still occupies the same amount of host memory and does not resolve the OOM issue.
> > > The second point is that during actual inference, the main cause of performance degradation is the repetitive host-to-device operations during the construction of the splitfuse mask, which are too time-consuming.
> > > I would greatly appreciate it if you could take a look at these points. @jianzs
> > 
> > 
> > This PR isn't focused on runtime performance.
> > It's all about fixing the high GPU memory spike during attention mask generating. The issue was that we weren't reusing GPU memory efficiently, which caused the memory usage to balloon to several times what the final mask actually needs. We've fixed that by implementing proper memory reuse during the generation process.
> 
> But this cached mask is created on CPU, i can't understand this pr how to resolve the OOM issue.

This is definitely a host memory OOM issue.

PR #3054, opened 2025-09-20
[Bugfix] fix e2e test

loukong33:
fix oom

use e2e tests

- vLLM version: v0.10.2

### DIFF
--- a/examples/offline_external_launcher.py
+++ b/examples/offline_external_launcher.py
@@ -125,6 +125,12 @@ def parse_args():
                         type=float,
                         default=None,
                         help="Model weight memory usage in GiB (e.g., 1.0 for 0.5B model).")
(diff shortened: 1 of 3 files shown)


examples/offline_external_launcher.py
```diff
@@ -125,6 +125,11 @@ def parse_args():
                         type=float,
                         default=None,
                         help="Model weight memory usage in GiB (e.g., 1.0 for 0.5B model).")
+    parser.add_argument("--gpu-memory-utilization",
+                        type=float,
```
momo609, line 130: Please add notes detailing the rationale behind this modification to assist in future bug fix.

loukong33: added

PR #3056, opened 2025-09-20
[Feature] qwen3_moe qk norm support multi_stream.

weijinqian0:
[Feature] qwen3_moe qk norm support multi_stream.
- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/6c5f82e5aa87cd73ce03ce10fc44138f75ee1aea

### DIFF
--- a/vllm_ascend/models/qwen3_moe.py
+++ b/vllm_ascend/models/qwen3_moe.py
@@ -17,7 +17,7 @@
 # Adapted from vllm/model_executor/models/qwen3_moe.py
 # This file is a part of the vllm-ascend project.
 
-from typing import Optional, Union
+from typing import Any, Optional, Union
 
 import torch
 from torch import nn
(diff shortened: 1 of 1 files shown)


vllm_ascend/models/qwen3_moe.py
```diff
@@ -125,6 +126,67 @@ def forward(
...
+            rope_scaling,
+            max_position_embeddings,
+            head_dim,
+            rms_norm_eps,
+            qkv_bias,
+            cache_config,
+            quant_config,
+            prefix,
+            dual_chunk_attention_config)
+        self.alt_stream = torch.npu.Stream()
```
ApsarasX, line 160: Each layer will create a stream, which will cause waste.

Can a stream pool be used like torchair to obtain streams?

PR #3069, opened 2025-09-20
[Misc] Refactor platform

wangxiyuan:
This pr refactor platform file to make it more clear.

1. Move useless import to TYPING CHECK
2. Update check_and_update_config to make it more readable.
3. Fixed a bug that if `enable_shared_expert_dp` is enable, the worker is forced set to torchair. torchair should be set only when mla is used as well.

- vLLM version: v0.10.2

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -22,7 +22,6 @@ from typing import TYPE_CHECKING, Optional, Tuple
 import torch
 import vllm.envs as envs_vllm
 from torch.distributed import ProcessGroup
-from torch.distributed.distributed_c10d import PrefixStore
 from vllm.logger import logger
 from vllm.platforms import Platform, PlatformEnum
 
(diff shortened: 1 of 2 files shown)


vllm_ascend/platform.py
```diff
@@ -239,27 +193,59 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
+        if compilation_config.pass_config.enable_sequence_parallelism:
+            if not parallel_config.enable_expert_parallel or vllm_config.model_config.hf_config.model_type != "qwen3_moe":
+                raise NotImplementedError(
+                    "Sequence Parallel only works on qwen moe models with expert parallel enabled currently."
+                )
 
+        # 4. Override worker cls
         if parallel_config and parallel_config.worker_cls == "auto":
-            if ascend_config.torchair_graph_config.enabled or ascend_config.enable_shared_expert_dp:
+            if ascend_config.torchair_graph_config.enabled or (
```
wangxiyuan, line 211: @linfeng-yuan  please review this change. we should consider mla case when set worker cls

linfeng-yuan: > @linfeng-yuan please review this change. we should consider mla case when set worker cls

I think that is good. `enable_shared_expert_dp` only support deepseek and kimi models, so this change will not affect current function as well as make this flag flexible and expandable.

PR #3072, opened 2025-09-21
[Feature] Remove restriction on kv nz usage without torchair

jianzs:

Previously, kv nz could only be used when torchair was enabled. This commit removes that limitation, allowing users to utilize kv nz functionality regardless of torchair is enabled or not.

Move the `enable_kv_nz` configuration option out of `torchair_graph_config` and place it at a higher level.

CI pass.

### DIFF
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -44,6 +44,8 @@ class AscendConfig:
         self.ascend_scheduler_config = AscendSchedulerConfig(
             ascend_scheduler_config)
         # Todo: Once https://github.com/vllm-project/vllm/issues/22246 is merged in vllm. Remove this config
+
+        self.enable_kv_nz = additional_config.get("enable_kv_nz", False)
         self.expert_map_path = additional_config.get("expert_map_path", None)
         self.expert_map_record_path = additional_config.get(
             "expert_map_record_path",
@@ -116,7 +118,6 @@ class TorchairGraphConfig:
             "enable_view_optimize", True)
         self.enable_frozen_parameter = torchair_graph_config.get(
             "enable_frozen_parameter", True)
-        self.enable_kv_nz = torchair_graph_config.get("enable_kv_nz", False)
 
         if not isinstance(self.graph_batch_sizes, list):
             raise TypeError("graph_batch_sizes must be list[int]")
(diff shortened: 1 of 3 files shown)


zzzzwwjj:
Please test the case without torchair and post the performance data, thks!

jianzs:
> Please test the case without torchair and post the performance data, thks!

Sorry for the dely. I tested on the latest code version using the DeepSeek V3 model with a TP1/DP32 configuration. The max_concurrency was around 768, and TPOT decreased from 87.24 to 85.05. MTP was not enabled during the test.

enable kv_nz
image

disable kv_nz
image

jianzs:
gsm8k_gen_250702: 96.66

jianzs:
@wangxiyuan  @ApsarasX  @zzzzwwjj  PTAL, ready to merge

jianzs:
I tested various configuration combinations offline, and here are the results:

image

vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -489,97 +491,116 @@ def _transfer_kv_cache(self, req_meta: dict[str, Any]):
...
-            (1, block_size)) + block_ids_tensor.reshape(
-                (num_blocks, 1)) * block_size)
-        slot_mapping = slot_mapping.flatten().to(device=device)
+            (1, self.block_size)) + block_ids_tensor.reshape(
+                (num_blocks, 1)) * self.block_size).flatten()
+
+        # FIXME: Right now, if we skip synchronization at this point, the system
+        # will crash in GQA scenarios. However, we still haven't identified the
+        # root cause.
+        torch.npu.synchronize()
```
realliujiaxu, line 550: why add this while this PR  is adding kv nz to mla?

jianzs: The original code had a lot of H2D operations, which looked messy, so I refactored it. However, it was later discovered that removing H2D caused the engine to crash in GQA scenarios when the tensor parallel sizes of prefill and decode nodes were inconsistent. To resolve this, I added a synchronization here.

jianzs: This code segment is shared by both cat_cache and nz_format.

PR #3073, opened 2025-09-21
[feat] support customized and separated hccl_buffer_size for process group initialization

linfeng-yuan:


jianzs:
As far as I'm aware, MC2 can only support the HCCL_BUFSIZE environment variable, not through API parameters. In which version was this restriction removed?

linfeng-yuan:
> As far as I'm aware, MC2 can only support the HCCL_BUFSIZE environment variable, not through API parameters. In which version was this restriction removed?

Yes, we also contacted to developers for HCCL and mc2 operators, as well as conducting some experiments. The result shows that setting buffer size for mc2 group through pg_options would lead to VRAM misalignment ​unless the `HCCL_BUFFSIZE` environment variable is set to the same value as specified in pg_options. Therefore, our current approach is to not set pg_options for mc2 group, and instead continue to rely on the user-configured `HCCL_BUFFSIZE` to control its buffer size. For other process groups (e.g., dp, tp, ep, etc.), we will continue to use pg_options for buffer size configuration.​

linfeng-yuan:
> > As far as I'm aware, MC2 can only support the HCCL_BUFSIZE environment variable, not through API parameters. In which version was this restriction removed?
> 
> Yes, we also contacted to developers for HCCL and mc2 operators, as well as conducting some experiments. The result shows that setting buffer size for mc2 group through pg_options would lead to VRAM misalignment ​unless the `HCCL_BUFFSIZE` environment variable is set to the same value as specified in pg_options. Therefore, our current approach is to not set pg_options for mc2 group, and instead continue to rely on the user-configured `HCCL_BUFFSIZE` to control its buffer size. For other process groups (e.g., dp, tp, ep, etc.), we will continue to use pg_options for buffer size configuration.​

@jianzs @wangxiyuan Please have a look of my latest commit and updated commit message.

PR #3078, opened 2025-09-22
moe weight prefetch in Qwen MOE Models

yuzhup:
This PR prefetchs the weight of moe layers in Qwen MOE Models to optimize the performance in Decode phase mainly.

No.

- vLLM version: v0.10.2


ApsarasX:
same to #2486

PR #3064, opened 2025-09-20
Refactor ci to reuse base workflow and re-enable ut coverage

Yikun:
### What this PR does / why we need it?
1. Refactor ci to reuse base workflow and enable main 2 hours trigger job:
- Extract e2e test in to _e2e_test.yaml
- Reuse _e2e_test in light / full job
- Enable main 2 hours trigger job

2. Rename e2e test to ascend test to make sure action display label 
3. Re-enable ut coverage which was failed since https://github.com/vllm-project/vllm-ascend/commit/5bcb4c1528311b8f20c91a9239ac38172f002fcb  and disable on https://github.com/vllm-project/vllm-ascend/commit/6d8bc38c7b0faba0b666585a09b7523c7714b9cc

### Does this PR introduce _any_ user-facing change?
Only developer behavior changes:
- Every job trigger full test with vllm release and hash
- Run full job per 2 hours with vllm main
- e2e light test (30 mins): `lint` (6mins) ---> ut (10mins) ---> `v0.10.2 + main / 4 jobs` (15mins)
- e2e full test (1.5h): `ready label` ---> `v0.10.2 + main / 4 jobs`, about 1.5h 
- schedule test: 2hours ---> `v0.10.2 + main / 4 jobs`, about 1.5h 

### How was this patch tested?
CI passed

- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/c60e6137f0bf2034853919b3a9d705d7e06b93cf

PR #3079, opened 2025-09-22
[CI] Enable main based lint check and light ci matrix

wangxiyuan:
Followup on https://github.com/vllm-project/vllm-ascend/pull/3064
1. should limit vllm version to the same hash with mypy
2. fix the vllm version bug for e2e light test.
no
CI passed

- vLLM version: v0.10.2

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -2,6 +2,10 @@ name: pre-commit
 
 on:
     workflow_call:
+      inputs:
+        vllm:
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -2,6 +2,10 @@ name: pre-commit
 
 on:
     workflow_call:
+    inputs:
```
Yikun: Indentation

PR #3085, opened 2025-09-22
[Refactor] [SP]The sequence parallelism characteristics in the MoE and Dense models are integrated into a single solution.

weijinqian0:

there are two sets of sp implementations for moe and dense models. One is called sequence_parallelism, and the other is flashcomm_v1.
We did the following things：

Merge two sets of code with the same implementation into one.
Remove the implementation of sequence_parallelism, as this solution cannot support aclgraph.

No

How was this patch tested?

e2e&ut

- vLLM version: v0.10.2


vllm_ascend/ops/linear_op.py
```diff
@@ -139,7 +139,7 @@ def apply(
         return output, output_bias
 
 
-class DenseOptimMergedColumnParallelOp(CustomColumnParallelOp):
+class SequenceMergedColumnParallelOp(CustomColumnParallelOp):
```
rjg-lyh, line 156: Since these ops include many optimizations that are independent of SP feature, the function name could be improved to better reflect its core functionality and general purpose, rather than associating it with a particular optimization or feature.

weijinqian0: Remove the prefetch feature to meet the single responsibility principle.

vllm_ascend/ops/linear_op.py
```diff
@@ -83,9 +85,17 @@ def update_attrs(self):
...
         self.quant_method = self.layer.quant_method
 
+    def apply_impl(self, input_):
+        raise NotImplementedError
+
     # Replace layer.forward to customize the layer computation process.
     def apply(self, input_):
-        raise NotImplementedError
+        output, output_bias = self.apply_impl(input_)
+        if dense_optim_enable():
```
weijinqian0: This variable will be modified by Yuhui.

vllm_ascend/ops/linear_op.py
```diff
@@ -82,10 +83,19 @@ def update_attrs(self):
...
+
+    def apply_impl(self, input_):
+        raise NotImplementedError
 
     # Replace layer.forward to customize the layer computation process.
     def apply(self, input_):
-        raise NotImplementedError
+        output, output_bias = self.apply_impl(input_)
+        if dense_optim_enable():
+            torch.ops.vllm.maybe_prefetch_mlp_gate_up_proj(output, self.prefix)
```
rjg-lyh: Weight prefetch should be enabled just in RowParallelLinear, so i think these code need to be moved to CustomRowParallelOp.

vllm_ascend/ops/linear_op.py
```diff
@@ -113,6 +121,14 @@ def update_attrs(self):
         self.reduce_results = self.layer.reduce_results
         self.input_size_per_partition = self.layer.input_size_per_partition
 
+    def apply(self, input_):
+        output, output_bias = self.apply_impl(input_)
+        if dense_optim_enable():
```
weijinqian0, line 126: This varible will be modified by yuhui.

vllm_ascend/ops/linear_op.py
weijinqian0: good idea.

wangxiyuan (approved):

PR #3095, opened 2025-09-22
[Bugfix][LoRA] Fix bug introduced by upstream vllm#25249

paulyu12:
Fix the impact to LoRA that https://github.com/vllm-project/vllm/pull/25249 brought.

No.

pytest -sv tests/e2e/singlecard/test_ilama_lora.py
pytest -sv tests/e2e/multicard/test_ilama_lora_tp2.py

- vLLM version: v0.10.2

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -124,7 +124,6 @@ else:
 import torch_npu
 
 import vllm_ascend.envs as envs_ascend
-from vllm_ascend.utils import vllm_version_is
 
 # if true, allow tensor initialization and casting with internal format (e.g., NZ)
 torch.npu.config.allow_internal_format = True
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2525,10 +2525,7 @@ def load_model(self) -> None:
                         self.model.get_eagle3_aux_hidden_state_layers())
 
             if self.lora_config:
-                self.model = self.load_lora_model(self.model,
-                                                  self.model_config,
-                                                  self.scheduler_config,
-                                                  self.lora_config,
+                self.model = self.load_lora_model(self.model, self.vllm_config,
```
Yikun:
```

if vllm_version_is("0.10.2"):
    // old
else:
    // new
```

We should also keep 0.10.2 support, otherwise CI will fail

Yikun:
Thanks for your contributions! See above comments to address ci failure

vllm_ascend/worker/model_runner_v1.py
paulyu12: Done.

wangxiyuan (approved):

PR #3116, opened 2025-09-23
fix wrong --num-gpus parameter requirements, and avoid ambiguity

MaoJianwei:
fix the problem of https://github.com/vllm-project/vllm-ascend/issues/3114
- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/5aeb9254521023f97aca292b3478aa7ff485ffb2


MaoJianwei:
Happy to see Ascend better :)

Thanks,
Mao

PR #3122, opened 2025-09-23
[Model] Support pooling models

### LINKED ISSUE #1960: [Bug]: BAAI/bge-reranker-v2-m3 failed to start in graph and eager mode due to Text-only XLMRobertaForSequenceClassification not be supported
### Your current environment

vllm version: v0.9.2rc1
```
VLLM_USE_MODELSCOPE=True vllm  serve BAAI/bge-reranker-v2-m3 --trust_remote_code --enforce-eager &

VLLM_USE_MODELSCOPE=True vllm  serve BAAI/bge-reranker-v2-m3  --trust_remote_code &
```

### 🐛 Describe the bug

bug:
```
INFO 07-22 07:01:33 [config.py:1472] Using max model len 8192
WARNING 07-22 07:01:33 [arg_utils.py:1735] ['XLMRobertaForSequenceClassification'] is not supported by the V1 Engine. Falling back to V0.
INFO 07-22 07:01:33 [config.py:4601] Only "last" pooling supports chunked prefill and prefix caching; disabling both.
  File "/usr/local/python3.10.17/bin/vllm", line 8, in <module>
    sys.exit(main())
  File "/__w/vllm-benchmarks/vllm-benchmarks/vllm-empty/vllm/entrypoints/cli/main.py", line 65, in main
    args.dispatch_function(args)
  File "/__w/vllm-benchmarks/vllm-benchmarks/vllm-empty/vllm/entrypoints/cli/serve.py", line 55, in cmd
    uvloop.run(run_server(args))
  File "/usr/local/python3.10.17/lib/python3.10/site-packages/uvloop/__init__.py", line 82, in run
    return loop.run_until_complete(wrapper())
  File "uvloop/loop.pyx", line 1518, in uvloop.loop.Loop.run_until_complete
  File "/usr/local/python3.10.17/lib/python3.10/site-packages/uvloop/__init__.py", line 61, in wrapper
    return await main
  File "/__w/vllm-benchmarks/vllm-benchmarks/vllm-empty/vllm/entrypoints/openai/api_server.py", line 1431, in run_server
    await run_server_worker(listen_address, sock, args, **uvicorn_kwargs)
  File "/__w/vllm-benchmarks/vllm-benchmarks/vllm-empty/vllm/entrypoints/openai/api_server.py", line 1451, in run_server_worker
    async with build_async_engine_client(args, client_config) as engine_client:
  File "/usr/local/python3.10.17/lib/python3.10/contextlib.py", line 199, in __aenter__
    return await anext(self.gen)
  File "/__w/vllm-benchmarks/vllm-benchmarks/vllm-empty/vllm/entrypoints/openai/api_server.py", line 158, in build_async_engine_client
    async with build_async_engine_client_from_engine_args(
  File "/usr/local/python3.10.17/lib/python3.10/contextlib.py", line 199, in __aenter__
    return await anext(self.gen)
  File "/__w/vllm-benchmarks/vllm-benchmarks/vllm-empty/vllm/entrypoints/openai/api_server.py", line 180, in build_async_engine_client_from_engine_args
    vllm_config = engine_args.create_engine_config(usage_context=usage_context)
  File "/__w/vllm-benchmarks/vllm-benchmarks/vllm-empty/vllm/engine/arg_utils.py", line 1286, in create_engine_config
    config = VllmConfig(
  File "/usr/local/python3.10.17/lib/python3.10/site-packages/pydantic/_internal/_dataclasses.py", line 123, in __init__
    s.__pydantic_validator__.validate_python(ArgsKwargs(args, kwargs), self_instance=s)
pydantic_core._pydantic_core.ValidationError: 1 validation error for VllmConfig
  Value error, vLLM Ascend does not support V0 engine. [type=value_error, input_value=ArgsKwargs((), {'model_co...additional_config': {}}), input_type=ArgsKwargs]
    For further information visit https://errors.pydantic.dev/2.11/v/value_error
[ERROR] 2025-07-22-07:01:33 (PID:16357, Device:-1, RankID:-1) ERR99999 UNKNOWN applicaiton exception

```

lianyiibo:

Support pooling models (like `bge-reranker-v2-m3`) in vllm-ascend, this pr covered the three model types of embed (cls_token, mean_token, lasttoken).

After this [commit](https://github.com/vllm-project/vllm/commit/17373dcd93ca60554d72cef4e159e70abbfd15af), vllm has provided support for adapting pooling models on the v1 engine. This PR includes corresponding adaptations on the vllm-ascend side.

Fixes #1960

No.

Taking `BAAI/bge-reranker-v2-m3` as an example.

```python
from argparse import Namespace
import os

from vllm import LLM, EngineArgs
from vllm.utils import FlexibleArgumentParser

def parse_args():
    parser = FlexibleArgumentParser()
    parser = EngineArgs.add_cli_args(parser)
    # Set example specific arguments
    parser.set_defaults(
        model="BAAI/bge-reranker-v2-m3", task="score", enforce_eager=True
    )
    return parser.parse_args()

def main(args: Namespace):
    # Sample prompts.
    text_1 = "What is the capital of France?"
    texts_2 = [
        "The capital of Brazil is Brasilia.",
        "The capital of France is Paris.",
    ]

    # Create an LLM.
    # You should pass task="score" for cross-encoder models
    llm = LLM(**vars(args))

    # Generate scores. The output is a list of ScoringRequestOutputs.
    outputs = llm.score(text_1, texts_2)

    # Print the outputs.
    print("\nGenerated Outputs:\n" + "-" * 60)
    for text_2, output in zip(texts_2, outputs):
        score = output.outputs.score
        print(f"Pair: {[text_1, text_2]!r} \nScore: {score}")
        print("-" * 60)

if __name__ == "__main__":
    args = parse_args()
    main(args)
```

#### Scripts
```bash
ASCEND_RT_VISIBLE_DEVICES=1 VLLM_USE_MODELSCOPE=true python3 -m vllm.entrypoints.openai.api_server --model BAAI/bge-m3
vllm bench serve --model BAAI/bge-m3 --backend vllm-rerank --endpoint /v1/rerank --dataset-name random-rerank --tokenizer BAAI/bge-m3 --random-input-len 512
```
#### Results
**Before this pr**
```bash
============ Serving Benchmark Result ============
Successful requests:                     1000      
Failed requests:                         0         
Benchmark duration (s):                  6.08      
Total input tokens:                      0         
Request throughput (req/s):              164.43    
Total Token throughput (tok/s):          0.00      
----------------End-to-end Latency----------------
Mean E2EL (ms):                          4029.80   
Median E2EL (ms):                        4024.80   
P99 E2EL (ms):                           5717.67   
==================================================
```
**After this pr**
```bash
============ Serving Benchmark Result ============
Successful requests:                     1000      
Failed requests:                         0         
Benchmark duration (s):                  6.37      
Total input tokens:                      0         
Request throughput (req/s):              157.04    
Total Token throughput (tok/s):          0.00      
----------------End-to-end Latency----------------
Mean E2EL (ms):                          3990.05   
Median E2EL (ms):                        3894.46   
P99 E2EL (ms):                           5869.24   
==================================================
```

- vLLM version: v0.12.0

### DIFF
--- a/tests/e2e/multicard/test_data_parallel.py
+++ b/tests/e2e/multicard/test_data_parallel.py
@@ -28,7 +28,10 @@ from unittest.mock import patch
 import pytest
 
 MODELS = [
-    "Qwen/Qwen3-0.6B", "Qwen/Qwen3-30B-A3B", "vllm-ascend/Qwen3-30B-A3B-W8A8"
+    "Qwen/Qwen3-0.6B",
+    "Qwen/Qwen3-30B-A3B",
+    # FIXME(Potabk): Skip this case for now
+    # "vllm-ascend/Qwen3-30B-A3B-W8A8"
 ]
 
 
--- a/tests/e2e/singlecard/test_async_scheduling.py
+++ b/tests/e2e/singlecard/test_async_scheduling.py
@@ -13,6 +13,7 @@ from tests.e2e.conftest import VllmRunner
 from tests.e2e.model_utils import check_outputs_equal
 
 MODEL = "Qwen/Qwen3-0.6B"
+MTP_MODEL = "wemaster/deepseek_mtp_main_random_bf16"
 
 first_prompt = ("The following numbers of the sequence " +
                 ", ".join(str(i) for i in range(10)) + " are:")
@@ -42,6 +43,27 @@ def test_without_spec_decoding(monkeypatch: pytest.MonkeyPatch, ):
     ]
 
     run_tests(monkeypatch, MODEL, test_configs, test_sampling_params)
+
+
+def test_with_spec_decoding(monkeypatch: pytest.MonkeyPatch):
+    """Test consistency and acceptance rates with some different combos of
+    preemption, executor, async scheduling, prefill chunking,
+    spec decoding model length.
+    """
+
+    spec_config = {
+        "method": "mtp",
+        "num_speculative_tokens": 2,
+    }
+
+    # test_preemption, executor, async_scheduling,
+    # spec_config, test_prefill_chunking
+    test_configs = [
+        (False, "mp", True, spec_config, False),
+        (False, "mp", False, spec_config, False),
+    ]
+
+    run_tests(monkeypatch, MTP_MODEL, test_configs, [{}])
 
 
 @dynamo_config.patch(cache_size_limit=16)


Potabk:
can we add some e2e and example for the model

lianyiibo:
> can we add some e2e and example for the model

OK, but since I have no relevant experience before, I would like to ask which test scenarios this feature needs to cover. At present, I think it is necessary to write test cases in the `e2e/singlecard` directory. Is this correct?

Potabk:
can refer to https://github.com/vllm-project/vllm-ascend/blob/main/tests/e2e/singlecard/test_embedding.py

lianyiibo:
> can refer to https://github.com/vllm-project/vllm-ascend/blob/main/tests/e2e/singlecard/test_embedding.py

@Potabk I've finished writing the e2e test case and refined the logic of the code to support more pooling models, PTAL.

vllm_ascend/attention/attention_v1.py
```diff
@@ -123,6 +123,7 @@ class AscendAttentionState(Enum):
     DecodeOnly = 2
     ChunkedPrefill = 3
     SpecDecoding = 4
+    EncoderOnly = 5
```
MengqingCao: I think `EncoderOnly` is an `AttentionType` instead of `AscendAttentionState`, why we put it here?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2604,6 +2640,33 @@ def _convert_torch_format(self, tensor):
         tensor = torch_npu.npu_format_cast(tensor, ACL_FORMAT)
         return tensor
 
+    def may_add_encoder_only_layers_to_kv_cache_config(self) -> None:
```
MengqingCao: let's make attention spec in https://github.com/lianyiibo/vllm-ascend/blob/551fce8ee2737c2e2aeacddc6d8b3c49632370ab/vllm_ascend/worker/model_runner_v1.py#L3287

vllm_ascend/attention/attention_v1.py
lianyiibo: This is to ensure that the value of AscendAttentionState is assigned as expected; for most pooling models, the other four states of AscendAttentionState are not representative of the inference procedure at this point in time, and in the absence of the addition of this extra Type, since the pooling model does not use the AscendScheduler, in some cases it will go into https://github.com/vllm-project/vllm-ascend/blob/cd69385dab402a75d27800ec003cb73dcf0b44fa/vllm_ascend/worker/model_runner_v1.py#L1550-L1551 

At the same time, I think this can be decoupled from the inference procedure of the generate model, avoiding unanticipated errors caused by future additions of certain logical judgments. :)

MengqingCao: I guess you want to make sure `forward_encoder` will be stepped into when the attn type is `ENCODE_ONLY`. However I think it has no relation to `AscendAttentionState`, which mainly refers to the requests state (whether it is chunked) in current batch. I recomand to include `forward_encoder` into each `_forward_xxx`(`_forward_prefill_no_cache`, `_forward_prefill_cache_hit`, `_forward_decode_only` and `_forward_v1_style`), so that it could be compatible with different schedulers and batch states

MengqingCao: > This is to ensure that the value of AscendAttentionState is assigned as expected; for most pooling models, the other four states of AscendAttentionState are not representative of the inference procedure at this point in time, and in the absence of the addition of this extra Type, since the pooling model does not use the AscendScheduler, in some cases it will go into
> 
> [vllm-ascend/vllm_ascend/worker/model_runner_v1.py](https://github.com/vllm-project/vllm-ascend/blob/cd69385dab402a75d27800ec003cb73dcf0b44fa/vllm_ascend/worker/model_runner_v1.py#L1550-L1551)
> 
> Lines 1550 to 1551 in [cd69385](/vllm-project/vllm-ascend/commit/cd69385dab402a75d27800ec003cb73dcf0b44fa)
> 
>  elif not ascend_config.ascend_scheduler_config.enabled or self.chunked_prefill_enabled: 
>      attn_state = AscendAttentionState.ChunkedPrefill 


A little confused, if pooling model could not use `AscendScheduler`, why it won't use chunked prefill? this is the  default state in v1 scheduler

lianyiibo: > A little confused, if pooling model could not use `AscendScheduler`, why it won't use chunked prefill? this is the default state in v1 scheduler

Yes, but I think the problem is that the pooling model doesn't have the prefill/decode phase of the generate model, and for the most part no kv_cache exists, which goes against the meaning of the Chunked Prefills themselves (as well as the rest of the AscendAttentionState), and thus I constructed another execution logic.

Of course, if the newly added `AscendAttentionState` is still considered less appropriate, I could merge the encoder forward logic into `_forward_prefill_no_cache` to remove the current `_forward_encoder` method and EncoderOnly state, which might also be expected to run.

vllm_ascend/worker/model_runner_v1.py
lianyiibo: This code reference is from this section in vllm 

https://github.com/vllm-project/vllm/blob/98f30b8cba5bb7e16421ab4c75fc200357744310/vllm/v1/worker/gpu_model_runner.py#L4551-L4575 

and can be found in the corresponding get_kv_cache_spec which uses a similar continue method of skipping 

https://github.com/vllm-project/vllm/blob/98f30b8cba5bb7e16421ab4c75fc200357744310/vllm/v1/worker/gpu_model_runner.py#L4642-L4647

I'm not sure if the modification would cause unintended problems.

vllm_ascend/attention/attention_v1.py
MengqingCao: I get your point, but I think it still not quite like a state. It is unique that no decode phase for encoder-only model, which has no autoregression process.

The `AscendAttentionState` actually just used to adapt `AscendScheduler`, a v0 style scheduler, which process prefill and decode req seperately. If you ensure this pr not support `AscendScheduler`, I think we should move the current `_forward_encoder` into `_forward_v1_style`

vllm_ascend/worker/model_runner_v1.py
MengqingCao: Okay, then let's keep aligh with vllm first

vllm_ascend/attention/attention_v1.py
lianyiibo: Removal of `AscendAttentionState.EncoderOnly` related code has been completed, PTAL.❤️

MengqingCao:
I'm sorry that I just noticed another pr to support pooling models, https://github.com/vllm-project/vllm-ascend/pull/3171, which post a performance improvement and is almost ready to merge. Could you test the performance of this pr to make a compare?

lianyiibo:
> I'm sorry that I just noticed another pr to support pooling models, #3171, which post a performance improvement and is almost ready to merge. Could you test the performance of this pr to make a compare?

I just read this PR and I don't think there seems to be a big conflict functionally? His PR only adapts bge-m3 in the pooling model and adds aclgraph support, for the current PR there is a wider variety of pooling models supported, a better idea would be to combine the code of the two PR's to get the most out of it?

Also I don't seem to have any performance test code for the pooling model in my current environment :( , I'll try to test it later if I need to.❤️

lianyiibo:
@MengqingCao Hi. Since I have no previous experience writing e2e test cases, I would like to know why the classify and score cases I wrote are not running correctly. The errors reported in them don't seem to be a problem that occurs when the model is running. And I can't reproduce the errors locally (running the cases using `pytest -sv tests/e2e/singlecard/test_pooling.py` passes them all), and I've tried a number of solutions today but can't solve them all😢

https://github.com/vllm-project/vllm-ascend/actions/runs/18497242697/job/52704528701#step:11:1160

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -314,13 +315,17 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
             self.block_size,
             use_mla=self.model_config.use_mla,
             use_sfa=self.ascend_config.use_sfa)
+        pooler_config = self.model_config.pooler_config
+        tril = self.model_config.runner_type == "generate" or (
```
MengqingCao: this logic will make an all-zeros matrix for the other type models, too. I think it is a bug?

MengqingCao:
> I just read this PR and I don't think there seems to be a big conflict functionally? His PR only adapts bge-m3 in the pooling model and adds aclgraph support, for the current PR there is a wider variety of pooling models supported, a better idea would be to combine the code of the two PR's to get the most out of it?

IMO, I think #3171 and this pr both aims to support encoder-only models, maybe you could make some examples that the wider variety of pooling models supported you mentioned refers to?

> Also I don't seem to have any performance test code for the pooling model in my current environment :( , I'll try to test it later if I need to.❤️

That's fine of this

MengqingCao:
> @MengqingCao Hi. Since I have no previous experience writing e2e test cases, I would like to know why the classify and score cases I wrote are not running correctly. The errors reported in them don't seem to be a problem that occurs when the model is running. And I can't reproduce the errors locally (running the cases using `pytest -sv tests/e2e/singlecard/test_pooling.py` passes them all), and I've tried a number of solutions today but can't solve them all😢
> 
> https://github.com/vllm-project/vllm-ascend/actions/runs/18497242697/job/52704528701#step:11:1160

Sorry for late relpy, I'll help to check if this turn of CI is failed and this pr indeed brings new features compared with #3171

lianyiibo:
> IMO, I think #3171 and this pr both aims to support encoder-only models, maybe you could make some examples that the wider variety of pooling models supported you mentioned refers to?

There are many of these, for example four different embed types can be found [here](https://www.sbert.net/docs/package_reference/sentence_transformer/models.html#main-modules) regarding the embed model, in addition to the embed task, different pooling models have different applicable tasks, in addition to embed, vllm provides support for classify, score, and reward tasks, which can be seen [here](https://github.com/vllm-project/vllm/blob/44c85556216bb24ab5ccf410dbb2bcfc25c85dd5/vllm/config/model.py#L93).

In the test cases currently given, I covered the three model types of embed (cls_token, mean_token, lasttoken), and provided use cases for two common tasks, classify and score (the local test passed, ci failed). Example models corresponding to these tasks can be found in the written test cases. In addition, since the current [conftest.py](https://github.com/vllm-project/vllm-ascend/blob/main/tests/e2e/conftest.py) file has outdated parts compared to vllm, and does not provide a pooling task interface, I rewrote the conftest file content with reference to vllm.

For scenarios that do not use aclgraph, #3171 also has limitations in adapting to all pooling models, e.g., there are accuracy issues when running the Qwen2.5-1.5B-apeach classify task, and it does not support the embed task for non-cls_token and non-lasttoken (for example intfloat/multilingual-e5-small).

lianyiibo:
> Sorry for late relpy, I'll help to check if this turn of CI is failed and this pr indeed brings new features compared with #3171

Thank you! Once I deleted the classify and score test cases (retention of embed test cases), the ci was able to work fine, see [here ](https://github.com/vllm-project/vllm-ascend/actions/runs/18547375253/job/52867907143), and I'm currently testing a single classify test case for passability.

---

The test results are in, and my newly added `test_classification.py` use case works fine locally, but it still doesn't pass the ci😢. This model is [jason9693/Qwen2.5-1.5B-apeach](https://modelscope.cn/models/Howeee/Qwen2.5-1.5B-apeach) in hugging face and [Howeee/Qwen2.5-1.5B-apeach](https://huggingface.co/jason9693/Qwen2.5-1.5B-apeach) in modelscope, while similar classify use cases can be found in vllm [here](https://github.com/vllm-project/vllm/blob/main/tests/models/language/pooling/test_classification.py).

lianyiibo:
@MengqingCao Hi! I recently refactored the e2e test code and added doc support. At the same time, I added broader aclgraph support based on #3171. Now aclgraph can also support almost all BERT/RoBERTa type pooling models. The code adaptation on the framework side has been basically completed. Since the current code is significantly different from the original code modification, it may be necessary to review the current code modification again.

Meanwhile, I still can't do anything about the error in ci test, the two models `Qwen2.5-1.5B-apeach` and `bge-reranker-v2-m3` are able to run normally in my local environment. I can't confirm whether it's due to some environment difference that I can't reproduce ci's error, and it may need you to analyze it further. Thanks you!❤️

MengqingCao:
> @MengqingCao Hi! I recently refactored the e2e test code and added doc support. At the same time, I added broader aclgraph support based on #3171. Now aclgraph can also support almost all BERT/RoBERTa type pooling models. The code adaptation on the framework side has been basically completed. Since the current code is significantly different from the original code modification, it may be necessary to review the current code modification again.
> 
> Meanwhile, I still can't do anything about the error in ci test, the two models `Qwen2.5-1.5B-apeach` and `bge-reranker-v2-m3` are able to run normally in my local environment. I can't confirm whether it's due to some environment difference that I can't reproduce ci's error, and it may need you to analyze it further. Thanks you!❤️

Sorry for the delay, I'll take a look at this pr and the failed CI later time at today.

MengqingCao:
also cc @JC-ut0 to take a look, thanks!

lianyiibo:
It looks like the `Qwen2.5-1.5B-apeach` problem was solved automatically without me modifying the code, so far the only model left that won't pass is `bge-reranker-v2-m3`.

vllm_ascend/attention/attention_mask.py
```diff
@@ -72,14 +74,6 @@ def get_attn_mask(self, max_seq_len: int, dtype: torch.dtype,
...
         ).to(device, non_blocking=True)
 
-    def get_pooling_mask(self, device):
-        if self.pooling_mask is None:
-            # the compressed attention mask for npu_fusion_attention sparse mode 4
-            self.pooling_mask = torch.triu(torch.ones(
-                2048, 2048), diagonal=1).to(torch.bool).to(device,
-                                                           non_blocking=True)
-        return self.pooling_mask
-
```
JC-ut0: have you compared the performance of the compressed attention mask for npu_fusion_attention which is in shape of [2048,2048], and the attention mask you constructed which is in shape of [max_seq_len,max_seq_len]?

Potabk:
can you list all the enc-dec models you tested? I'd like to help check the weight

lianyiibo:
> can you list all the enc-dec models you tested? I'd like to help check the weight

embedding: "Qwen/Qwen3-Embedding-0.6B", "BAAI/bge-small-en-v1.5", "intfloat/multilingual-e5-small", "sentence-transformers/all-MiniLM-L12-v2"
classification: "Howeee/Qwen2.5-1.5B-apeach"
scoring: "dengcao/ms-marco-MiniLM-L6-v2", "BAAI/bge-reranker-v2-m3"

They are all model names on modelscope.

tests/e2e/singlecard/pooling/test_embedding.py
```diff
@@ -16,22 +16,30 @@
...
+import pytest
 from modelscope import snapshot_download  # type: ignore[import-untyped]
 
 from tests.e2e.conftest import HfRunner, VllmRunner
 from tests.e2e.utils import check_embeddings_close
 
+MODELS = [
+    "Qwen/Qwen3-Embedding-0.6B",  # lasttoken
+    "BAAI/bge-small-en-v1.5",  # cls_token
+    "intfloat/multilingual-e5-small"  # mean_tokens
```
MengqingCao, line 28: CI failed may due to the download conflict while downloading. Plz help to take a look @Potabk

vllm_ascend/attention/attention_mask.py
```diff
@@ -72,14 +74,6 @@ def get_attn_mask(self, max_seq_len: int, dtype: torch.dtype,
         return self.attn_mask_cache[:max_seq_len, :max_seq_len].contiguous(
         ).to(device, non_blocking=True)
 
-    def get_pooling_mask(self, device):
```
MengqingCao: This couldn't be removed as this is required by `torch_npu.npu_fusion_attention`

vllm_ascend/patch/worker/patch_roberta.py
```diff
@@ -1,88 +0,0 @@
...
-def _decode_token_type_ids(input_ids: torch.Tensor) -> torch.Tensor:
-
-    token_type_ids = input_ids // TOKEN_TYPE_MULTIPLIER
-
-    input_ids.bitwise_and_(TOKEN_MASK)
-
-    return token_type_ids
-
-
-def roberta_for_sequence_classification_forward(
```
MengqingCao, line 49: Thanks for this cleanup!

vllm_ascend/patch/__init__.py
```diff
@@ -128,16 +128,7 @@
 #    Future Plan:
 #       Fix this bug in torch-npu, bump torch-npu version and remove this patch.
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-#   1. `vllm.model_executor.models.roberta.RobertaEmbedding.forward`
-#    Why:
```
MengqingCao, line 110: plz don't remove the reason and plan of this patch, thus we could remove it in the future

vllm_ascend/attention/attention_mask.py
```diff
@@ -15,7 +15,9 @@
 import torch
 
 
-def _generate_attn_mask(max_seq_len, dtype):
+def _generate_attn_mask(max_seq_len, dtype, tril):
+    if not tril:
+        return torch.zeros(size=(max_seq_len, max_seq_len)).to(dtype)
```
MengqingCao: why we need a full-zeros mask? this seems weird

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1672,7 +1711,14 @@ def _generate_process_reqs_hidden_states(self, attn_metadata, with_prefill,
...
                           num_valid_tokens):
         ascend_config = get_ascend_config()
-        if np.array_equal(self.seq_lens_np[:num_reqs], num_scheduled_tokens):
+        if self.model_config.runner_type == "pooling":
+            if isinstance(
+                    self.kv_cache_config.kv_cache_groups[0].kv_cache_spec,
+                    EncoderOnlyAttentionSpec):
+                attn_state = AscendAttentionState.PrefillNoCache
+            else:
+                attn_state = AscendAttentionState.PrefillCacheHit
```
MengqingCao, line 2070: Actually I'm confused about this change. Alough there is only prefill in encoder-only scenario, it is not recommanded to set attn_state accordding to model type

vllm_ascend/attention/attention_v1.py
```diff
@@ -1047,25 +1049,8 @@ def forward(
             if self.pcp_size * self.dcp_size > 1:
                 output = self._forward_pcp_dcp(query, key, value,
                                                attn_metadata, output)
-
-            elif attn_type == AttentionType.ENCODER_ONLY:
-                cum_seq_len = attn_metadata.query_start_loc[1:].tolist()
-                attn_out = torch_npu.npu_fusion_attention(
```
MengqingCao: Removing this seems a bug, this will make no attention operators could use for encoder-only

vllm_ascend/attention/attention_mask.py
JC-ut0: I guess this mask is used for encoder models, not sure about the performance impact

vllm_ascend/patch/__init__.py
lianyiibo: In this PR, I have updated the original `patch_roberta.py` file to `patch_bert.py`, and no longer patch the above two methods. I don't quite understand why there is no need to change the comment?

vllm_ascend/attention/attention_mask.py
lianyiibo: Yes, I removed the calling site of `torch_npu.npu_fusion_attention` together

vllm_ascend/worker/model_runner_v1.py
lianyiibo: This problem has similar reasons to many of the following problems, and I will explain them here:

1. In vllm, there are only two runner_type types: generate and pooling. For the generate model, masks are mostly tril matrices without considering sparse attention. As for the pooling model, it is divided into cls, lasttoken, max, mean, etc. according to the pooling_type. See this [link](https://www.sbert.net/docs/package_reference/sentence_transformer/models.html#main-modules) for details. Except for the lasttoken type of pooling model, all other models require an all-zero mask matrix (the tril here can be regarded as the more widely used is_causal parameter, maybe I can change the tril variable to is_causal?), so that each query can consider the keys and values ​​of all tokens in a single attention calculation.
2. In the actual reasoning process of vllm, the lasttoken type pooling model is often regarded as a generate-like model, including opening kv_cache, not using the AttentionType.ENCODER_ONLY mark, and kv_cache_spec not being set to Enc oderOnlyAttentionSpec, etc., the main reason is that most of their underlying models are not the two typical encoder models of bert/roberta, such as Qwen/Qwen3-Embedding-0.6B, which is basically consistent with the prefill stage reasoning of Qwen3-0.6B
3. Regarding the issue of AscendAttentionState, it has also been mentioned before. Under the non-lasttoken type of pooling model, it can be regarded as the equivalent of PrefillNoCache. Under the lasttoken type of pooling model, these models have been able to run correctly before this PR (such as Qwen/Qwen3-Embedding-0.6B), in order to ensure the consistency of its front and rear paths, it is artificially set to PrefillCacheHit (before this PR, requests for this type of model will fall into the else branch, but because the pooling model does not use AscendScheduler, it will be assigned ChunkedPrefill if it is not specified)

    > In fact, I found that if it is set to ChunkedPrefill, accuracy problems will occur on the Howeee/Qwen2.5-1.5B-apeach model, resulting in full nan values. Perhaps it is an interface problem with attention? I'm not sure.

Jeaniowang:
hello, is this method performance better than torch_npu.npu_fusion_attention?

lianyiibo:
> hello, is this method performance better than torch_npu.npu_fusion_attention?

I think yes, the performance of the atb operator is better than the acl operator in many scenarios.

vllm_ascend/attention/attention_mask.py
lianyiibo: In vllm, there are only two runner_type types: generate and pooling. For the generate model, masks are mostly tril matrices without considering sparse attention. As for the pooling model, it is divided into cls, lasttoken, max, mean, etc. according to the pooling_type. See this [link](https://www.sbert.net/docs/package_reference/sentence_transformer/models.html#main-modules) for details. Except for the lasttoken type of pooling model, all other models require an all-zero mask matrix (the tril here can be regarded as the more widely used is_causal parameter, maybe I can change the tril variable to is_causal?), so that each query can consider the keys and values ​​of all tokens in a single attention calculation.

---

11.14: In the current code, `tril` has been modified to `is_causal`

vllm_ascend/worker/model_runner_v1.py
lianyiibo: Under the non-lasttoken type of pooling model, it can be regarded as the equivalent of PrefillNoCache. Under the lasttoken type of pooling model, these models have been able to run correctly before this PR (such as Qwen/Qwen3-Embedding-0.6B), in order to ensure the consistency of its front and rear paths, it is artificially set to PrefillCacheHit (before this PR, requests for this type of model will fall into the else branch, but because the pooling model does not use AscendScheduler, it will be assigned ChunkedPrefill if it is not specified)

> In fact, I found that if it is set to ChunkedPrefill, accuracy problems will occur on the Howeee/Qwen2.5-1.5B-apeach model, resulting in full nan values. Perhaps it is an interface problem with attention? I'm not sure.

vllm_ascend/attention/attention_v1.py
lianyiibo: The pooling model of lasttoken will also not be marked as `AttentionType.ENCODER_ONLY`. This judgment cannot include all pooling models, and the `npu_fusion_attention` interface forces `mask.shape=(2048,2048)`. As an implementation of attention, it is consistent with the forward logic of other branches, I think there is no need to add additional judgment logic.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -4130,6 +4176,15 @@ def _capture_aclgraphs(self, compilation_cases: list[int],
...
                     aclgraph_runtime_mode.name))
+
+        force_attention = (aclgraph_runtime_mode == CUDAGraphMode.FULL)
+        # When the kv cache spec is empty, PiecewiseBackend is not initialized, and
+        # compilation_case=1 will cause the dynamic shape position to be incorrectly derived.
+        if not self.get_kv_cache_spec():
+            self._dummy_run(2,
+                            aclgraph_runtime_mode=CUDAGraphMode.NONE,
+                            force_attention=force_attention,
+                            uniform_decode=uniform_decode)
```
lianyiibo, line 4106: @MengqingCao Hello, I'm sorry that I didn't have time to track the progress of the community and maintain the resolution of PR conflicts in the past two weeks. Yesterday, I provided the latest version of the pooling model aclgraph support, which focused on adding the above code snippet. At the same time, I think the ci problem seems to have been solved. If the current code look good, I will adjust the order of pooling-related tests to other test items in `_e2e_test.yaml`, thank you!

MengqingCao:
Thanks for the update, I just wonder if there is any performance improvement or regression on this pr? Because acctually you changed the attention operator in this pr. What's more, I think it is better to keep a branch for encoder-only attention, as I tested locally with this pr, the attention operator in `_forward_v1_style` could not handle the encoder-only scenario.

Jeaniowang:
I test your pr， the performance is almost same with torch_npu.npu_fusion_attention

lianyiibo:
> Thanks for the update, I just wonder if there is any performance improvement or regression on this pr? Because acctually you changed the attention operator in this pr. What's more, I think it is better to keep a branch for encoder-only attention, as I tested locally with this pr, the attention operator in `_forward_v1_style` could not handle the encoder-only scenario.

Okay, I will recently separate the encoder-only attention into a separate branch for processing.

> I test your pr， the performance is almost same with torch_npu.npu_fusion_attention

Thanks for the test, I was wondering if it would be convenient to provide the relevant test scripts and parameters? Much appreciated!

lianyiibo:
> Thanks for the update, I just wonder if there is any performance improvement or regression on this pr? Because acctually you changed the attention operator in this pr. What's more, I think it is better to keep a branch for encoder-only attention, as I tested locally with this pr, the attention operator in `_forward_v1_style` could not handle the encoder-only scenario.

I have added a separate branch code in the attention section for the pooling model to process. In addition, I also restored the attention operator to `npu_fusion_attention`. You can check the modifications from the latest commit diff, PTAL. @MengqingCao

MengqingCao:
> > Thanks for the update, I just wonder if there is any performance improvement or regression on this pr? Because acctually you changed the attention operator in this pr. What's more, I think it is better to keep a branch for encoder-only attention, as I tested locally with this pr, the attention operator in `_forward_v1_style` could not handle the encoder-only scenario.
> 
> I have added a separate branch code in the attention section for the pooling model to process. In addition, I also restored the attention operator to `npu_fusion_attention`. You can check the modifications from the latest commit diff, PTAL. @MengqingCao

Really sorry for the delay, I'll take a look at this weekend as I'm busy on the other issues.

lianyiibo:
> > > Thanks for the update, I just wonder if there is any performance improvement or regression on this pr? Because acctually you changed the attention operator in this pr. What's more, I think it is better to keep a branch for encoder-only attention, as I tested locally with this pr, the attention operator in `_forward_v1_style` could not handle the encoder-only scenario.
> > 
> > 
> > I have added a separate branch code in the attention section for the pooling model to process. In addition, I also restored the attention operator to `npu_fusion_attention`. You can check the modifications from the latest commit diff, PTAL. @MengqingCao
> 
> Really sorry for the delay, I'll take a look at this weekend as I'm busy on the other issues.

I am very happy that you have not forgotten this PR. If you have any suggestions for modifications to the current refactored code, I hope you can continue to make them. I will continue to improve it in my free time to make it more suitable.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3937,6 +3976,15 @@ def _capture_aclgraphs(self, compilation_cases: list[int],
                 desc="Capturing ACL graphs ({}, {})".format(
                     "decode" if uniform_decode else "mixed prefill-decode",
                     aclgraph_runtime_mode.name))
+
+        force_attention = (aclgraph_runtime_mode == CUDAGraphMode.FULL)
+        # When the kv cache spec is empty, PiecewiseBackend is not initialized, and
+        # compilation_case=1 will cause the dynamic shape position to be incorrectly derived.
+        if not self.get_kv_cache_spec():
+            self._dummy_run(2,
```
MengqingCao, line 4103: QQ: why specifying `num_tokens` to 2 here?

lianyiibo: > QQ: why specifying `num_tokens` to 2 here?




I have written the brief reasons above through comments. Specifically, the following code 
https://github.com/vllm-project/vllm-ascend/blob/e8e20c0bbf03420bb4dc6a953777a1b324527ced/vllm_ascend/worker/model_runner_v1.py#L3976
will arrange `aclgraph_batch_sizes` from small to large. Generally speaking, the minimum value of `aclgraph_batch_sizes` is 1. From [this document](https://docs.pytorch.org/docs/stable/torch.compiler_dynamo_deepdive.html#are-always-specialized), you can find that for a compile with an input dimension of 1, dynamo will always regard it as a static variable, and the dynamic mark will not have any effect. The position of 2 here can be replaced with any number greater than 1 to correctly initialize the dynamo graph.

In addition, in the adapted version of the previous version of aclgraph, `compilation_cases` will be sorted in reverse order. I am not sure why it was changed to positive order. In order to cause other possible problems, I separately judged the situation of the pooling model.

MengqingCao: Thanks for the details, I got your point now

MengqingCao:
I'll help to do the rebase, a force-push will be needed after it

docs/source/user_guide/support_matrix/supported_models.md
```diff
@@ -48,7 +48,8 @@ Get the latest info here: https://github.com/vllm-project/vllm-ascend/issues/160
 |-------------------------------|-----------|----------------------------------------------------------------------|------|--------------------|------|-----------------|------------------------|------|----------------------|------------------|-----------------|-------------------|-----------------|---------------|-------------------------------|--------------------|--------------------|---------------|---------------------|-----|
 | Qwen3-Embedding               | ✅        |                                                                      |||||||||||||||||||
 | Molmo                         | ✅        | [1942](https://github.com/vllm-project/vllm-ascend/issues/1942)      |||||||||||||||||||
-| XLM-RoBERTa-based             | ❌        | [1960](https://github.com/vllm-project/vllm-ascend/issues/1960)      |||||||||||||||||||
+| XLM-RoBERTa-based             | ✅        | [1960](https://github.com/vllm-project/vllm-ascend/issues/1960)      |||||||||||||||||||
+| Bert                          | ✅        | [1960](https://github.com/vllm-project/vllm-ascend/issues/1960)      |||||||||||||||||||
```
MengqingCao: we need to remove the issue link here, let's do this after CI passes

lianyiibo: > we need to remove the issue link here, let's do this after CI passes

OK, thank you very much for your help!

MengqingCao:
Update: After multiple tests, the throughput and end-to-end performance fluctuated, and the fluctuation range before and after this PR was almost the same, so I think it's best to keep it as it is.

@lianyiibo Thanks for the great work! I will update the graph batch sizes in the new adding test cases to speed up CI. 
I make a perf test locally before and after this pr, ensuring that there is not a performance regression on `bge-m3`. ~~The performance on throughput and P99 E2EL is deteriorated but the Mean E2EL and Median E2EL is better than before. I'll try to revert the fusion attention changes and test again.~~

I think this pr could be merge today or tomorrow


### Perf test
#### Scripts
```bash
ASCEND_RT_VISIBLE_DEVICES=1 VLLM_USE_MODELSCOPE=true python3 -m vllm.entrypoints.openai.api_server --model BAAI/bge-m3
vllm bench serve --model BAAI/bge-m3 --backend vllm-rerank --endpoint /v1/rerank --dataset-name random-rerank --tokenizer BAAI/bge-m3 --random-input-len 512
```
#### Results
**Before this pr**
```bash
============ Serving Benchmark Result ============
Successful requests:                     1000      
Failed requests:                         0         
Benchmark duration (s):                  6.08      
Total input tokens:                      0         
Request throughput (req/s):              164.43    
Total Token throughput (tok/s):          0.00      
----------------End-to-end Latency----------------
Mean E2EL (ms):                          4029.80   
Median E2EL (ms):                        4024.80   
P99 E2EL (ms):                           5717.67   
==================================================
```
**After this pr**
```bash
============ Serving Benchmark Result ============
Successful requests:                     1000      
Failed requests:                         0         
Benchmark duration (s):                  6.37      
Total input tokens:                      0         
Request throughput (req/s):              157.04    
Total Token throughput (tok/s):          0.00      
----------------End-to-end Latency----------------
Mean E2EL (ms):                          3990.05   
Median E2EL (ms):                        3894.46   
P99 E2EL (ms):                           5869.24   
==================================================
```

MengqingCao:
TODO: add a lightly CI for pooling model

MengqingCao:
All CI passed in https://github.com/vllm-project/vllm-ascend/actions/runs/20052554215/job/57534470268?pr=3122

MengqingCao:
Great thanks for the hard work! @lianyiibo

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2022,7 +2061,14 @@ def _generate_process_reqs_hidden_states(self, attn_metadata, with_prefill,
...
                           num_valid_tokens):
-        if np.array_equal(self.seq_lens_np[:num_reqs], num_scheduled_tokens):
+        if self.model_config.runner_type == "pooling":
+            if isinstance(
+                    self.kv_cache_config.kv_cache_groups[0].kv_cache_spec,
+                    EncoderOnlyAttentionSpec):
+                attn_state = AscendAttentionState.PrefillNoCache
+            else:
+                attn_state = AscendAttentionState.PrefillCacheHit
+        elif np.array_equal(self.seq_lens_np[:num_reqs], num_scheduled_tokens):
```
lianyiibo, line 2071: @MengqingCao Hi, when I was checking the latest code today, I found that this judgment logic has become redundant code after using `attn_metadata.is_causal_pooling`. Consider removing them in later releases.

MengqingCao: Okay, pr welcome!

PR #3102, opened 2025-09-22
add pagedattention to support FULL_DECODE_ONLY.

momo609:
### What this PR does / why we need it?
Calculate in advance the workspace memory size needed for the PagedAttention operator to avoid deadlocks during resource cleanup. This PR requires torch_npu version 0920 or newer.
### How was this patch tested?

- vLLM version: v0.11.0rc3
- vLLM main: https://github.com/vllm-project/vllm/commit/v0.11.0

PR #3125, opened 2025-09-23
[Feat][Graph]Support FULL_DECEDE_ONLY mode for MLA models

panchao-hub:
Adds support for capturing the Multi-Layer Attention (MLA) decode operation into an ACL graph. This improves performance by compiling the attention kernel for single-token decoding.

Key changes include:
- Implementing the graph capture logic for the MLA kernel, including workspace management and parameter updates.
- Modifying the rotary embedding (RoPE) handling to use pre-allocated tensors, which is a requirement for graph capture.
- Adding a `build_for_graph_capture` method to the MLA metadata builder to create dummy metadata during the graph compilation phase.

Known issues:
- Currently, MTP is not supported in FULL_DECEDE_ONLY mode -- we're working on a fix
- We are preparing to remove update_mla_attn_params with auto_dispatch_capture

compilation_config={
    "cudagraph_mode": "FULL_DECODE_ONLY",
},

- vLLM version: v0.11.0rc3

### DIFF
--- a/tests/e2e/singlecard/test_aclgraph.py
+++ b/tests/e2e/singlecard/test_aclgraph.py
@@ -21,6 +21,7 @@ Run `pytest tests/compile/test_aclgraph.py`.
 """
 
 import os
+
 import pytest
 from vllm import SamplingParams
 
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2252,7 +2252,8 @@ class NPUModelRunner(LoRAModelRunnerMixin):
                                    inputs_embeds=inputs_embeds)
         forward_context = get_forward_context()
         assert forward_context is not None
-        if forward_context.cudagraph_runtime_mode == CUDAGraphMode.FULL:
+        if forward_context.cudagraph_runtime_mode == CUDAGraphMode.FULL and \
+            not forward_context.capturing:
             if self.vllm_config.model_config.use_mla:
                 # FIXME: Try using `auto_dispatch_capture=True`
                 update_mla_attn_params(self.update_stream, forward_context,


vllm_ascend/attention/mla_v1.py
```diff
@@ -408,8 +415,8 @@ def build(
                 max_seq_lens=max_seq_lens,
                 attn_mask=common_attn_metadata.spec_attn_mask,
                 actual_seq_lengths_q=actual_seq_lengths_q,
-                sin=sin,
-                cos=cos)
+                sin=sin[:num_decodes, ...],
+                cos=cos[:num_decodes, ...])
```
momo609: add fullgraph check avoid piecewisegraph fail.

panchao-hub: has been fixed

vllm_ascend/attention/mla_v1.py
```diff
@@ -834,24 +880,63 @@ def _forward_decode(
...
+            event = torch.npu.ExternalEvent()
+            event.wait(stream)
+            event.reset(stream)
+            graph_params.events[num_tokens].append(event)
+
+            workspace = graph_params.workspaces.get(num_tokens)
+            if workspace is None:
+                workspace = torch_npu._npu_fused_infer_attention_score_get_max_workspace(
+                    q_nope, k_nope, k_nope, **common_kwargs)
+                graph_params.workspaces[num_tokens] = workspace
```
yiz-liu, line 915: Please merge this after #3102 , so that we can use the `update_graph_params_workspaces` instead of set value directly. @wangxiyuan @panchao-hub

panchao-hub: Do not need to change it. We will replace it with auto_dispatch_capture later.

wangxiyuan (approved):

PR #3127, opened 2025-09-23
fix error async_scheduler can't be enabled

Ronald1995:
PR #2894 make ascend_scheduler_config.enabled always be `True` for non-mla models，when `ascend_scheduler_config.enabled=True `, it will always initialize `AscendScheduler` which is a subclass of `Scheduler`, but when we enbale async_scheduling,we need to initialize `AsyncScheduler` in vllm,  this will make async_scheduling can't be enabled.

not-related

when user set `async_scheduling`, it means user don't want to use `AscendScheduler`, so we shouldn't set `ascend_scheduler_config.enabled = True`

- vLLM version: v0.10.2

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -135,7 +135,8 @@ class NPUPlatform(Platform):
         else:
             structured_outputs_config = vllm_config.structured_outputs_config
 
-        if model_config is not None and not model_config.use_mla:
+        if (model_config is not None and not model_config.use_mla
+                and not scheduler_config.async_scheduling):
             logger.info(
                 "Non-MLA LLMs forcibly disable the chunked prefill feature,"
                 "as the performance of operators supporting this feature "


vllm_ascend/platform.py
```diff
@@ -134,6 +134,13 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
             structured_outputs_config = vllm_config.decoding_config
         else:
             structured_outputs_config = vllm_config.structured_outputs_config
+        chunked_prefill_enabled_in_ascend_scheduler = getattr(
+            ascend_scheduler_config, "enable_chunked_prefill", False)
```
rjg-lyh: I suggest that it's unnecessary to write so redundantly here. Once asynchronous scheduling is enabled, the Ascend scheduler can be directly disabled. The specific operation can involve adding this condition to the subsequent if statement.

wangxiyuan (approved):

MengqingCao:
LGTM, let's modify the `test_async_scheduling` to check if the accuracy is correct

wangxiyuan (approved):

PR #3132, opened 2025-09-23
[BugFix] Fix dummy_run memory explosion in eager mode

dragondream-chen:

It is a quick bugfix for the memory explosion issue that requires further refactoring.
The dummy_run in eager mode may lead to OOM and the reason is that `hidden_states` were not released in time. 
The PR temporarily resolves the issue by manually clearing the cache, and further refactoring will be conducted subsequently. 

Before the modification, the dummy_run's memory showed an accumulation issue. 
image

After modification, it can be observed that the memory is released promptly. 
image

And it was verified that the model responded normally after a single data input. 

- vLLM version: v0.10.2
- vLLM main: https://github.com/vllm-project/vllm/commit/b1068903fdca26cf6b4a1a51a32c3365ce3ac636

### DIFF
--- a/vllm_ascend/ops/moe/fused_moe_prepare_and_finalize.py
+++ b/vllm_ascend/ops/moe/fused_moe_prepare_and_finalize.py
@@ -186,10 +186,10 @@ class FusedMoEPrepareAndFinalizeWithMC2(FusedMoEPrepareAndFinalize):
                                 self.moe_config.tp_group.device_group)
                 hidden_states = torch.cat(self.split_hidden_states, dim=0)
 
-                # TODO: It is a quick bugfix for the single-operator memory explosion issue
-                # that requires further restructuring. 
-                # If the cache is not cleared after `self.split_hidden_states` is created, 
-                # it can lead to the single-operator memory explosion.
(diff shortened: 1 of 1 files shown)


yiz-liu:
A friendly reminder is that you can use "eager mode" instead of "single-operator" (which is not a thing, kinda like google translation lol).

vllm_ascend/ops/moe/fused_moe_prepare_and_finalize.py
```diff
@@ -186,6 +186,12 @@ def finalize(self, hidden_states: torch.Tensor,
                                 self.moe_config.tp_group.device_group)
                 hidden_states = torch.cat(self.split_hidden_states, dim=0)
 
+                # TODO: It is a quick bugfix for the memory explosion issue in eager mode
+                # that requires further restructuring.
+                # If the cache is not cleared after `self.split_hidden_states` is created,
+                # it can lead to the memory explosion in eager mode.
+                del self.split_hidden_states
```
realliujiaxu, line 192: Consider not assigning `self.split_hidden_states` during the prepare phase. Instead, use `get_tp_group().allgather()` in the finalize phase, which internally employs `allgather_into_tensor` to eliminate tensor move overhead.

PR #2816, opened 2025-09-08
[main] mlp weight prefetch in Qwen Dense Models

rjg-lyh:
### What this PR does / why we need it?
This PR prefetchs the weight of mlp layers in Qwen Dense Models to optimize the performance in Decode phase mainly.

### Does this PR introduce _any_ user-facing change?
 No.

### How was this patch tested?
CI passed with new added/existing test.

- vLLM version: main
- vLLM main: https://github.com/vllm-project/vllm/commit/a1213fae5f8fe624c4b954768d23a86e384abdb1

PR #3146, opened 2025-09-24
[WIP][1/N][Feat] Attention and MoE weight prefetch in `Qwen3MoE` models

zhoux77899:

- Prefetch `qkv_proj.weight` and `o_proj.weight` in Attention blocks, `gate_up_proj.weight` in MoE blocks
- Prefetching these weights before matmul-like operators executed imporves performance due to L2

If not enabling weight prefetch, add a new config in `--additional-config`:
```json
{
    "weight_prefetch_config": {
        "attn": {
            "enabled": "false"
        },
        "moe": {
            "enabled": "false"
        }
    }
}
```


vllm_ascend/ops/weight_prefetch.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+
+    def __init__(self, weight_prefetch_config: WeightPrefetchConfig):
+        super().__init__(weight_prefetch_config)
+
+    @classmethod
+    def create(cls, weight_prefetch_config: WeightPrefetchConfig, vllm_config: VllmConfig) -> "AttentionWeightPrefetchMethod":
+        return cls(weight_prefetch_config)
+
+    def maybe_weight_prefetch_preprocess(self, weight: torch.Tensor, start_flag: torch.Tensor):
```
momo609: Can this prefetching be integrated with the prefetching in the dense layer?

vllm_ascend/ops/weight_prefetch.py
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @classmethod
+    def init_forward_prefetch(cls, vllm_config: VllmConfig):
+        weight_prefetch_config = get_ascend_config().weight_prefetch_config
+        for prefetch_cls in PrefetchManager.prefetch_map.keys():
+            PrefetchManager.prefetch_map[prefetch_cls] = prefetch_cls.create(weight_prefetch_config, vllm_config)
+
+    @classmethod
+    def get_prefetch_obj(cls, prefetch_cls: type[WeightPrefetchMethod]) -> WeightPrefetchMethod:
```
wangxiyuan: get_prefetch_method

vllm_ascend/quantization/w8a8.py
```diff
@@ -96,12 +97,19 @@ def apply(
         bias: Optional[torch.Tensor] = None,
         tp_rank: Optional[int] = 0,
     ) -> torch.Tensor:
+        if layer.__class__.__name__ in ["AscendQLVParallelLinear", "AscendRowParallelLinear"]:
```
wangxiyuan: add a support_prefecth api for linear ops is better.

vllm_ascend/quantization/w8a8.py
```diff
@@ -96,12 +97,19 @@ def apply(
...
+            PrefetchManager.get_prefetch_obj(AttentionWeightPrefetchMethod).maybe_weight_prefetch_preprocess(layer.weight, x)
```
wangxiyuan: move `PrefetchManager.get_prefetch_obj(AttentionWeightPrefetchMethod)` to  init func?

vllm_ascend/ops/weight_prefetch.py
```diff
@@ -0,0 +1,180 @@
...
+        return PrefetchManager.prefetch_map.get(prefetch_cls)
+
+
+def npu_prefetch(inputs: torch.Tensor,
```
momo609: util.py has the same func.

vllm_ascend/ops/weight_prefetch.py
```diff
@@ -0,0 +1,180 @@
...
+        else:
+            self.activated = False
+            return
+
+    def maybe_weight_prefetch_preprocess(self, weight: torch.Tensor, start_flag: torch.Tensor):
+        if not self.activated or self.prefetch_instruction_delivered:
+            return
+        forward_context = get_forward_context()
+        self.prefetch_stream.wait_stream(self.main_stream)
+        with torch.npu.stream(self.prefetch_stream):
```
whx-sjtu: Please use the `npu_stream_switch` api in `vllm_ascend/utils.py`.

vllm_ascend/ops/weight_prefetch.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @classmethod
+    def init_forward_prefetch(cls, vllm_config: VllmConfig):
+        weight_prefetch_config = get_ascend_config().weight_prefetch_config
+        for prefetch_cls in PrefetchManager.prefetch_map.keys():
+            PrefetchManager.prefetch_map[prefetch_cls] = prefetch_cls.create(weight_prefetch_config, vllm_config)
+
+    @classmethod
+    def get_prefetch_obj(cls, prefetch_cls: type[WeightPrefetchMethod]) -> WeightPrefetchMethod:
+        return PrefetchManager.prefetch_map.get(prefetch_cls)
```
weijinqian0: may should be deleted.

vllm_ascend/ops/weight_prefetch.py
zhoux77899: yes, it can be integrated in the new unified method and @rjg-lyh will refactor it

vllm_ascend/ops/weight_prefetch.py
zhoux77899: this func has a `TODO` comment that wants to move it to `ops` module and I did it, thanks.

vllm_ascend/ops/weight_prefetch.py
zhoux77899: used it, thanks.

vllm_ascend/ops/weight_prefetch.py
zhoux77899: deleted it, thanks.

vllm_ascend/ascend_config.py
```diff
@@ -174,6 +179,24 @@ def __init__(self, ascend_scheduler_config: dict):
...
+
+    prefetch_ratio: dict = {
+        "attn": {
+            "qkv": 1.0,
+            "o": 1.0,
+        },
+    }
+
+    def __init__(self, weight_prefetch_config: dict):
+        self.enabled = weight_prefetch_config.get("enabled", True)
```
rjg-lyh: This value is better set to `False`

vllm_ascend/ops/weight_prefetch.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class WeightPrefetchMethod:
+    """
+    Unified weight prefetch method.
+    """
+
+    def __init__(self, weight_prefetch_config: WeightPrefetchConfig) -> None:
+        self.calculation_stream = current_stream()
+        self.prefetch_stream = prefetch_stream()
```
rjg-lyh: The extra creation of prefetch stream should not be done, when weight_prefetch disabled.

vllm_ascend/ops/weight_prefetch.py
```diff
@@ -0,0 +1,73 @@
...
+
+        self.attn = ModuleWeightPrefetchConfig(
```
rjg-lyh, line 37: as same as above comment, this initalization is better done when weight_prefetch enabled.

vllm_ascend/ops/weight_prefetch.py
```diff
@@ -0,0 +1,73 @@
...
+            self.enable = self.enable and any(self.prefetch_ratio.values()) > 0
+
+
+class WeightPrefetchMethod:
+    """
+    Unified weight prefetch method.
+    """
+
+    def __init__(self, weight_prefetch_config: WeightPrefetchConfig) -> None:
+        self.calculation_stream = current_stream()
```
rjg-lyh: The current stream here has no use and i recommend to delete it.

vllm_ascend/ops/weight_prefetch.py
```diff
@@ -0,0 +1,73 @@
...
+                                  max_weight_size=int(weight_size))
+
+    def maybe_prefetch_attn_weight_postprocess(self) -> None:
+        if self.attn.enable and self.prefetch_stream is not None:
+            self.calculation_stream.wait_stream(self.prefetch_stream)
+
+    def weight_prefetch_impl(self, weight: torch.Tensor,
+                             start_flag: torch.Tensor,
+                             max_weight_size: int) -> None:
+        self.prefetch_stream.wait_stream(self.calculation_stream)
```
rjg-lyh: when we use current_stream, we get it.

vllm_ascend/quantization/w8a8.py
```diff
@@ -96,12 +98,15 @@ def apply(
...
         tp_rank: Optional[int] = 0,
     ) -> torch.Tensor:
-        if x.dtype != torch.int8:
-            x = quant_per_tensor(
-                x,
-                layer.aclnn_input_scale_reciprocal,
-                layer.aclnn_input_offset,
-            )
+        # TODO(zhoux77899): integrate prefetch qkvo_proj.weight with unquantized layers
+        x = torch.ops.vllm.maybe_prefetch_with_quant(
```
rjg-lyh: I strongly disagree with this approach, as the quant operation itself is decoupled from the behavior of prefetching weights. My recommended approach would be to treat the prefetching weight operations as a custom operation, rather than involving it in the higher-level modules that utilize this feature.
You can refer to [PR 2816](https://github.com/vllm-project/vllm-ascend/pull/2816).

vllm_ascend/ascend_config.py
zhoux77899: done

vllm_ascend/ops/weight_prefetch.py
zhoux77899: all variables of npu streams are deleted and to create/get the npu stream if needed

vllm_ascend/ops/weight_prefetch.py
zhoux77899: all variables of npu streams are deleted and to create/get the npu stream if needed

vllm_ascend/ops/weight_prefetch.py
zhoux77899: all variables of npu streams are deleted and to create/get the npu stream if needed

vllm_ascend/quantization/w8a8.py
zhoux77899: weight prefetching and quant is decoupled now

vllm_ascend/ops/weight_prefetch.py
zhoux77899: I believe it would be better to initialize the configuration for each module (`self.attn`, etc.) directly here, because this way each module can control whether to enable weight prefetching using only its own switch, without having to check whether a module’s configuration exists or whether the master switch is turned on. This approach makes the `if-else` condition checks more concise and elegant.

vllm_ascend/quantization/w8a8.py
```diff
@@ -97,11 +98,32 @@ def apply(
         tp_rank: Optional[int] = 0,
     ) -> torch.Tensor:
         if x.dtype != torch.int8:
+            attn_weight_map = {
+                "AscendQKVParallelLinear": "qkv",
+                "AscendRowParallelLinear": "o",
```
wangxiyuan, line 103: this two key is hard-code which is not good. Please try another way, for example `AscendQKVParallelLinear.__name__`?
And move `attn_weight_map` to utils is better.

vllm_ascend/ops/weight_prefetch.py
```diff
@@ -0,0 +1,75 @@
+from dataclasses import dataclass, field
+
+import torch
+import torch_npu
+
+from vllm_ascend.ascend_config import WeightPrefetchConfig
+
+SUPPORTED_MODULES = ["attn", "mlp", "moe"]
```
wangxiyuan, line 8: SUPPORTED_LAYERS？

vllm_ascend/quantization/w8a8.py
zhoux77899: removed the hard-code key and moved `attn_weight_map` as a config

vllm_ascend/ops/weight_prefetch.py
zhoux77899: I understand that ‘Layer’ may readily evoke associations with ‘DecoderLayer’, which typically encompasses components like Attention/MLA+MLP/MoE. However, this const should refer to a more narrowly defined components?

PR #3149, opened 2025-09-24
[Perf] Integrate new split_with_sizes_copy op to enhance inference performace

xintongli1001:
Integrate new split_with_sizes_copy op to improve perfomance in noraml inference cases and resolve long-seq mask problems .

The original op's performance is suboptimal in certain scenarios, necessitating optimization through the new op (npu_fused_infer_attention_score)。
For ultra-long sequences (128k), the original operator will allocate a large attn_mask, which consumes excessive CPU memory. In contrast, the new op supports a fixed-size compressed mask, effectively resolving this issue.
NOTE1: The current PR retains the original logic and uses a version check of the CANN package to determine whether the new op can be enabled. This ensures no impact on existing users. In future versions, this version check and the original logic will be deprecated, and the new op scheduling will be uniformly adopted.
NOTE2: This pr relies on future CANN version, which is not available now.
NOTE3: To enable the new op in chunked prefill, the parameter additional_config should be set like --additional-config '{"ascend_scheduler_config": {"enabled":true,"enable_chunked_prefill":true}}' \ at least.

No

How was this patch tested?
CI passed
- vLLM version: v0.10.2

### DIFF
--- a/vllm_ascend/models/qwen3_moe.py
+++ b/vllm_ascend/models/qwen3_moe.py
@@ -17,7 +17,7 @@
 # Adapted from vllm/model_executor/models/qwen3_moe.py
 # This file is a part of the vllm-ascend project.
 
-from typing import Optional
+from typing import Optional, Any
 
 import torch
 from torch import nn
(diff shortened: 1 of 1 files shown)


wangxiyuan:

PR #3157, opened 2025-09-24
[CI] Fix `multistream_overlap_shared_expert` test

shen-shanshan:

Fix `multistream_overlap_shared_expert` test.

Main changes:

1. Use model with shared experts for test.
2. enable ep.
3. Move this test to multicard dir.

### DIFF
--- a/tests/e2e/multicard/test_multistream_overlap_shared_expert.py
+++ b/tests/e2e/multicard/test_multistream_overlap_shared_expert.py
@@ -28,7 +28,7 @@ from tests.e2e.conftest import VllmRunner
 from tests.e2e.model_utils import check_outputs_equal
 
 MODELS = [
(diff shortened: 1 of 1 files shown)


tests/e2e/multicard/test_multistream_overlap_shared_expert.py
```diff
@@ -39,15 +39,19 @@ def test_models_with_multistream_overlap_shared_expert(
...
+        "The capital of France is",
+        "The future of AI is",
     ]
 
     sampling_params = SamplingParams(max_tokens=max_tokens, temperature=0.0)
     with VllmRunner(
             model,
             max_model_len=1024,
             enforce_eager=True,
+            data_parallel_size=2,
```
wangxiyuan, line 53: dp works in this case?

PR #3158, opened 2025-09-24
 [bugfix] fix oom in aclgraph

WithHades:
fix oom in aclgraph.

1. In the current token dispatch implementation, tensors are mounted on class instances to facilitate parameter passing between different methods. This approach prevents automatic recycling of these tensors. In some cases, it may lead to out-of-memory error. To address this issue, we manually set these tensors to None to release corresponding memory. 

2. The `profile_run` method is designed to accurately estimate the maximum NPU memory usage during vLLM inference. However, in certain scenarios, MoE models perform inference via MC2, which includes communication and consumes additional NPU memory. This leads to inaccurate estimation by the profile run. We address this by actively triggering the MC2 during profile run for initialization.```.

- vLLM version: v0.10.2

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2499,6 +2499,14 @@ class NPUModelRunner(LoRAModelRunnerMixin):
         with self.set_in_profile_run():
             hidden_states = self._dummy_run(self.max_num_tokens,
                                             with_prefill=True)
+            # MC2 will consume additional NPU memory.
+            # Therefore, we need to run the MC2 path once here to complete its initialization,
+            # allowing vLLM to correctly estimate the maximum memory required.
+            if self._select_moe_comm_method(
+                    self.mc2_tokens_capacity,
+                    with_prefill=True) == MoECommType.MC2:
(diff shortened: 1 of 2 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1917,6 +1917,12 @@ def set_in_profile_run(self):
     def profile_run(self) -> None:
         # Trigger compilation for general shape.
         with self.set_in_profile_run():
+            soc_version = get_ascend_soc_version()
+            if self.parallel_config.enable_expert_parallel and (
+                    soc_version in {
+                        AscendSocVersion.A2, AscendSocVersion.A3
+                    }) and (self.max_num_tokens > self.mc2_tokens_capacity):
+                self._dummy_run(self.mc2_tokens_capacity - 1)
```
weijinqian0: Why need this?

weijinqian0 (approved):

wangxiyuan (approved):
please update the commit message to explain more what this PR fixed.

weijinqian0 (approved):

wangxiyuan (approved):

PR #3188, opened 2025-09-25
[CI] Enable multi thread load weight to speed up

Potabk:
Enable multi thread load weight to speed up test term

- vLLM version: v0.10.2


Potabk:
This mechanism has little effect, closed

PR #3189, opened 2025-09-25
[Model][1/N] Delete deepseek v2/v3 modeling codes.

whx-sjtu:
This PR deletes model codes of deepseek_v2 and deepseek_v3 to reuse the model file from vLLM.

vLLM Ascend now uses custom ops register way instead of model file hard-coding.

- vLLM version: v0.11.0rc3

### DIFF
--- a/tests/ut/models/conftest.py
+++ b/tests/ut/models/conftest.py
@@ -90,7 +90,7 @@ def mock_distributed():
     mock_vllm_config.scheduler_config = Mock(max_num_seqs=256)
     mock_vllm_config.model_config = Mock(max_model_len=2048, quant_config=None)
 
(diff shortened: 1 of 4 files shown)


LCAIZJ:
Why is it possible to delete these two model files now?

whx-sjtu:
> Why is it possible to delete these two model files now?

What do you think is the reason why it cannot be deleted? If you refer to the CUDA hard-coding issue of DeepSeek 3.2, I will circumvent it through a patch later. @LCAIZJ

whx-sjtu:
Deleting all the model scripts is currently the most urgent work. If it blocks the development of some features that rely on modifying model scripts, I think these features should consider submitting a pull request to the vllm community rather than modifying our current model scripts. Our ultimate goal must be to reuse vllm's model scripts rather than maintaining our own.

wangxiyuan:
Nice work

PR #2802, opened 2025-09-08
[main] flashcomm_v1 optim in Qwen Dense Models

rjg-lyh:
### What this PR does / why we need it?
Flashcomm_v1 optim in Qwen Dense Models.

### Does this PR introduce _any_ user-facing change?
No.

### How was this patch tested?
CI passed with new added/existing test.

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/5e537f45b453096c489d06f0c47776f9436ef99b

PR #3193, opened 2025-09-25
[Bugfix] fix bug when tp=1

realliujiaxu:
Addresses a bug in DenseOptimRowParallelOp that occurs when tensor parallelism is not used
No

- vLLM version: v0.10.2


wangxiyuan (approved):

wangxiyuan:
seems that there is no e2e test for the optima dense case?

weijinqian0 (approved):

realliujiaxu:
> seems that there is no e2e test for the optima dense case?

Dense optim has end-to-end test cases (https://github.com/vllm-project/vllm-ascend/pull/2802), but TP=1 is not the target scenario and requires supplementation.

PR #3204, opened 2025-09-26
[BugFix] Fix ACLgraph bug in Qwen3_32b_int8 case

lilinsiman:
1. Solved the issue where sizes capture failed for the Qwen3-32b-int8 model when aclgraph, dp1, and tp4 were enabled.
2. Added the exception thrown when sizes capture fails and provided a solution
3. Add this common problem to the FAQ doc
no

ut

- vLLM version: v0.10.2


wangxiyuan:
This PR fixed capture graph bug, it's safe to merge and I'm sure it won't break any other function.

PR #3209, opened 2025-09-26
[Disagg][Perf] Use NPU event sync instead of blocking tolist to avoid unintentional copy ops blocking across different NPU streams, improving disagg TTIT/TTFT

jesse996:
This PR is based on top of https://github.com/vllm-project/vllm/pull/22760
When we copy the sampled valid token ids from device to host, avoid using tolist which would trigger a CUDA wise stream sync if the source is on device. We change it to use non-blocking copy followed by an explicit CUDA event sync.

Bring up vLLM server
```bash
VLLM_USE_V1=1 vllm serve Qwen/Qwen2.5-14B-Instruct --disable-l
og-requests -tp 8 --max-num-seqs 64 --no-enable-prefix-caching --max_num_batched_tokens=8000
```
## Before：
76218085a0cde9b2a73214e35fb7fc08
## After
6c2111136673332244d3ce11060f4048

As shown in the figure, the TTFT decreased

- vLLM version: v0.12.0


wangxiyuan:
this change doesn't work with CANN8.3, we're working on it.

wangxiyuan (approved):

PR #3220, opened 2025-09-28
[Bugfix]Fix quant_config input parameter bug in qwenvl series

booker123456:
Fix quant_config input parameter bug in qwenvl series. Currently, non-instantiated variables should be passed.
None

- vLLM version: v0.10.2


wangxiyuan (approved):
This is a quick fix for qwen vl only. I'll merge this in quick.

PR #3227, opened 2025-09-28
[Doc]Add qwen3_vl series guide

booker123456:

### DIFF
--- a/docs/source/tutorials/single_npu_qwen3vl.md
+++ b/docs/source/tutorials/single_npu_qwen3vl.md
@@ -1,0 +1,185 @@
+# Single NPU (Qwen3-VL 4B)
+
+## Run vllm-ascend on Single NPU
+
+### Offline Inference on Single NPU
+
+Run docker container:
+
+```{code-block} bash
+   :substitutions:
+# Update the vllm-ascend image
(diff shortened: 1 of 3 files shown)


docs/source/tutorials/single_npu_qwen3vl.md
```diff
@@ -0,0 +1,186 @@
+# Single NPU (Qwen3-VL 4B)
```
Yikun, line 1: This should also be removed

Yikun, line 1: https://github.com/vllm-project/vllm/blob/main/examples/offline_inference/vision_language.py#L1459

OK, Although Qwen3 VL 4B has not been released yet, even the vLLM upstream has already include this model name, we'd better do a post verification after model published

wangxiyuan: model is here https://huggingface.co/Qwen/Qwen3-VL-235B-A22B-Instruct . the problem now is it's rely on a new release of transformers. And 4B doesn't affect this PR IMO

wangxiyuan:
Let's merge this first. If vllm release without new transformers. We can add a note in doc later.

PR #3238, opened 2025-09-28
[bugfix] Fix warning bug: model config is None.

weijinqian0:
Cleanup wrong warning log error: model config is None

- vLLM version: v0.10.2

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -592,10 +592,13 @@ def dense_optim_enable() -> bool:
 
 def enable_sp(vllm_config=None) -> bool:
     from vllm.config import get_current_vllm_config
+    enable_sequence_parallelism = False
     if vllm_config is None:
         vllm_config = get_current_vllm_config()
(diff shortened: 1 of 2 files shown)


vllm_ascend/utils.py
```diff
@@ -590,11 +590,12 @@ def dense_optim_enable() -> bool:
     return envs_ascend.VLLM_ASCEND_ENABLE_DENSE_OPTIMIZE
 
 
-def enable_sp() -> bool:
-    from vllm.config import get_cached_compilation_config
-
+def enable_sp(vllm_config=None) -> bool:
+    from vllm.config import get_current_vllm_config
+    if vllm_config is None:
+        vllm_config = get_current_vllm_config()
```
MengqingCao: Let's only import when vllm_config is None
```suggestion
    if vllm_config is None:
        from vllm.config import get_current_vllm_config
        vllm_config = get_current_vllm_config()
```

weijinqian0: fixed

PR #3246, opened 2025-09-28
[Quickfix] Fix dp+ep+tp error when sp chunked the hidden_states

MengqingCao:
Fix dp+ep+tp inplace copy error when sp chunked the `hidden_states`.

test locally with the following scripts
```bash
python examples/offline_data_parallel.py \
        --model="Qwen/Qwen3-30B-A3B" \
        --dp-size=2 \
        --tp-size=2 \
        --enable-expert-parallel
```

### DIFF
--- a/vllm_ascend/ops/fused_moe.py
+++ b/vllm_ascend/ops/fused_moe.py
@@ -295,6 +295,7 @@ class AscendFusedMoE(FusedMoE):
                 in_dtype=params_dtype,
             )
         self.moe_config = moe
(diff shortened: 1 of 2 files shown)


MengqingCao:
CI failed due to eagle `tests/e2e/singlecard/spec_decode_v1/test_v1_spec_decode.py::test_eagle_correctness`, which is fixed by https://github.com/vllm-project/vllm-ascend/pull/3252

PR #3256, opened 2025-09-29
[Core]Append padding logic for Attention

Angazenn:

This PR aims to add padding logic to seq_lens、block_tables when running in full decode scenario. Before this PR, the number of input tokens with padding might exceeds corresponding seq_lens. For example, when running in full decode scenario:

```
input_ids : [1, 3, 0, 0]
seq_lens: [2, 1]
query_start_loc: [0, 1, 2]
```
Here, `input_ids` is padded by 2 tokens while `seq_lens`/`query_start_loc` are not. The mismatch between `input_ids` and `seq_lens`/`query_start_loc` might cause some potential bugs. This PR would change it into :

```
input_ids : [1, 3, 0, 0]
seq_lens: [2, 1, 1, 1]
query_start_loc: [0, 1, 2, 3, 4]
```

No.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -217,7 +217,8 @@ class AscendAttentionMetadataBuilder:
                                                                        num_reqs
                                                                        + 1]
 
-        if attn_state == AscendAttentionState.DecodeOnly:
+        if attn_state == AscendAttentionState.DecodeOnly and \
+            common_attn_metadata.num_input_tokens > num_actual_tokens:
             padded_num_tokens = common_attn_metadata.num_input_tokens - num_actual_tokens
             seq_lens = torch.cat([
                 seq_lens,
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2261,7 +2261,6 @@ class NPUModelRunner(LoRAModelRunnerMixin):
                     seq_lens=self.seq_lens_cpu[:num_reqs],
                     num_reqs=num_reqs,
                     num_actual_tokens=num_tokens,
-                    num_input_tokens=num_tokens,
                     actual_seq_lengths_q=self.actual_seq_lengths_q,
                     block_table_tensor=block_table_tensor[:num_reqs],
                     slot_mapping=self.slot_mapping,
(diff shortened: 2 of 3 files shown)


vllm_ascend/attention/attention_v1.py
```diff
@@ -207,6 +207,28 @@ def build(
...
+                device=block_table.device)
+            block_table = torch.cat([block_table, block_table_padding], dim=0)
+            query_start_loc_cpu = torch.cat([
+                query_start_loc_cpu,
+                torch.arange(query_start_loc_cpu[-1] + 1,
+                             query_start_loc_cpu[-1] + padded_num_tokens,
+                             dtype=query_start_loc_cpu.dtype,
+                             device=query_start_loc_cpu.device)
+            ])
+
```
yiz-liu, line 241: Why don't we handle this in `prepare_inputs` — make everything in `common_attn_metadata` final so it doesn't need further modification? I'm worried this version will add maintenance overhead, since there are other attention implementations.

Angazenn: I move this to `prepare_inputs` function. But I'm not sure whether these padding can generalize to different attention backend. More test is needed.

Angazenn: It seems that moving these padding to `prepare_inputs` is in conflicts with mtp and there might be other potential bugs with different attention backends. So I decide to move it back to `build`.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1583,7 +1583,10 @@ def _build_attn_state(self, num_reqs, num_scheduled_tokens,
         return attn_state
 
     def _update_graph_pad_size(self, with_prefill, graph_pad_size):
-        self.graph_pad_size = -1
+        if not with_prefill:
+            self.graph_pad_size = graph_pad_size
+        else:
+            self.graph_pad_size = -1
```
yiz-liu: Please use attn_state instead of modifying this method.

yiz-liu (changes requested):

vllm_ascend/worker/model_runner_v1.py
Angazenn: done

yiz-liu (approved):

vllm_ascend/attention/attention_v1.py
```diff
@@ -216,6 +216,29 @@ def build(
         query_start_loc_cpu = common_attn_metadata.query_start_loc_cpu[:
                                                                        num_reqs
                                                                        + 1]
+
+        if attn_state == AscendAttentionState.DecodeOnly and \
```
momo609, line 220: why this only affect decodeonly state.

PR #3267, opened 2025-09-29
[v0.9.1] Igore unrelated check

Yikun:
Igore unrelated check for examples and benchmark

No

CI passed


Yikun:
This PR include yaml change so trigger lint is expeceted, test on https://github.com/vllm-project/vllm-ascend/pull/3200

PR #3279, opened 2025-09-30
[bugfix] Fix moe bug: allgather error.

weijinqian0:
It will crash when deepseek model executed in A2.

- vLLM version: v0.11.0rc3


wangxiyuan:
let's wait for CI pass.

PR #3282, opened 2025-09-30
[bugfix]change log2phy map to npu

offline893:

Resolved the issue of EPLB failure caused by changes in the log2phy map due to device type modifications when using MTP rotation position encoding. 

https://github.com/vllm-project/vllm/commit/releases/v0.11.0

- vLLM version: v0.11.0rc3


wangxiyuan:
No update for long time. close it now. Feel free to create a new one if it's still necessary

PR #3304, opened 2025-09-30
add model_type deepseek_v32 support

kkrazy:
Add model_type deepseek_v32 support with proper model registration and AutoConfig support

No.

Image was built and fully tested with various seq length.


Yikun:
Please commit with `-s` on future to signed the DCO, now I set it manually.

See more in:
https://github.com/vllm-project/vllm-ascend/runs/51595967455

Yikun (approved):

PR #3309, opened 2025-10-02
faster chunk gated delta rule by torch ops

xyz2606:
A faster chunked gated delta rule implementation by basic torch operators.
(1) faster inverse of a tril matrix
(2) faster computation of attn score and recurrent state for each chunk

- Fixes #
num_qk_heads and sequence_length were reversed
wrong initialization of last_recurrent_state when initial_state is None (# q heads were used instread of # v heads)
-->

No

tested by random input of batch_size = 1 & seq_len=8192

- vLLM version: v0.11.0rc3


wangxiyuan:

PR #3321, opened 2025-10-09
[Perf] Add FIA interface in FA case

ZYang6263:

Add new npu_fused_infer_attention_score op to improve perfomance in flash attention case.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -265,9 +265,7 @@ class AscendAttentionMetadataBuilder:
             attn_mask=attn_mask,
             attn_state=attn_state,
             enable_dbo_across_dp=common_attn_metadata.enable_dbo_across_dp)
-
         return attn_metadata
-
 
     def build_for_graph_capture(
         self,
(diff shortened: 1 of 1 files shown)


vllm_ascend/attention/attention_v1.py
```diff
@@ -312,15 +312,29 @@ def _forward_prefill_no_cache(
...
-                                       num_kv_heads=self.num_kv_heads,
-                                       out=output)
+        if torch.version.cann.startswith("8.3") and self.head_size != 256:
+            output, _ = torch_npu.npu_fused_infer_attention_score(
+                query=query,
+                key=key,
+                value=value,
+                atten_mask=mask,
+                input_layout="TND",
+                actual_seq_lengths=attn_metadata.query_start_loc[1:],
```
rjg-lyh: The `query_start_loc` in `attn_metadata` is passed in as a tensor, which causes an internal type conversion within PTA, converting it to a list. This results in significant memory overhead on the device side. I recommend trying to pass the data in a list format directly from the outside and then conducting performance tests.

PR #3337, opened 2025-10-09
[CI] fix the vllm run-npu-test.sh break: cannot import _Backend from vllm.platforms

leo-pony:
 Adapt to vllm changes: ["Move Backend enum into registry #25893" ](https://github.com/vllm-project/vllm/pull/25893)

No

- vLLM version: v0.11.0rc3


wangxiyuan:
we'd like to remove this patch file once ds3.2 is updated to vLLM way.

PR #3343, opened 2025-10-09
[Bugfix] fix the assert error in thedeepseek + torchair + MTP scenario

WithHades:

On the A2 server, when deploying PD separation using four machines, in the scenario of deepseek+torch+MTP and setting `max_num_batched_tokens`, the service failed to start up and reported an error during `dummy_run`. The reason for the problem is that the `mc2_tokens_capacity` was not checked during `profile_run`.

- vLLM version: v0.11.0rc3


wangxiyuan:

PR #3356, opened 2025-10-10
[Feat]Unquantized Linear to nz and control all nz-cast

anon189Ty:
Currently, when executing to the Linear layer of models in vLLM-Ascend, the weights format is ND in unquantized case and skipped ascend case.
This PR supplements the execution logic for Linear layer. We use a new global variable: VLLM_ASCEND_ENABLE_NZ. When VLLM_ASCEND_ENABLE_NZ=1 and CANN version is 8.3, the weights of the Linear layer will be converted to FRACTAL_NZ, in both unquantized case and skipped ascend case. We also use VLLM_ASCEND_ENABLE_NZ to control the existing NZ conversion, such as w8a8-quantized case.

Add a new global variable VLLM_ASCEND_ENABLE_NZ. If you want to use NZ format, you should set VLLM_ASCEND_ENABLE_NZ=1.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/ops/linear.py
+++ b/vllm_ascend/ops/linear.py
@@ -24,17 +24,29 @@ from typing import Optional, Union
 
 import torch
 import torch.nn as nn
+import torch_npu
 from torch.nn.parameter import Parameter
 from vllm.distributed import divide
 from vllm.model_executor.layers.linear import (  # noqa
     WEIGHT_LOADER_V2_SUPPORTED, ColumnParallelLinear, LinearBase,
     MergedColumnParallelLinear, QKVParallelLinear, QuantizeMethodBase,
-    RowParallelLinear, UnquantizedLinearMethod)
+    ReplicatedLinear, RowParallelLinear, UnquantizedLinearMethod)
 from vllm.model_executor.layers.quantization.base_config import \
     QuantizationConfig
 from vllm.model_executor.utils import set_weight_attrs
 
-from vllm_ascend.ops.linear_op import get_parallel_op
+from vllm_ascend.ops.linear_op import get_parallel_op, get_replicated_op
+from vllm_ascend.utils import ACL_FORMAT_FRACTAL_NZ, is_enable_nz
+
+
+class AscendUnquantizedLinearMethod(UnquantizedLinearMethod):
+    """Linear method without quantization"""
+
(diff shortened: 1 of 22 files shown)


vllm_ascend/quantization/w8a8.py
```diff
@@ -158,8 +137,9 @@ def process_weights_after_loading(self, layer):
             requires_grad=False).to(layer.aclnn_input_scale.dtype)
         if self.transpose_weight:
             layer.weight.data = layer.weight.data.transpose(0, 1).contiguous()
-        layer.weight.data = torch_npu.npu_format_cast(layer.weight.data,
-                                                      ACL_FORMAT_FRACTAL_NZ)
+        if envs_ascend.VLLM_ASCEND_ENABLE_NZ:
```
momo609: This check can be extracted into a common function.

vllm_ascend/ops/linear.py
```diff
@@ -24,18 +24,33 @@
...
+class AscendUnquantizedLinearMethod(UnquantizedLinearMethod):
+    """Linear method without quantization"""
+
+    def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
+        super().process_weights_after_loading(layer)
+        if envs_ascend.VLLM_ASCEND_ENABLE_NZ and torch.version.cann.startswith(
+                "8,3"):
+            layer.weight.data = torch_npu.npu_format_cast(
+                layer.weight.data, ACL_FORMAT_FRACTAL_NZ)
+            self.unquant_to_nz = False
```
weijinqian0: may can remove this parameter

realliujiaxu:
why not set NZ by default, but instead add an environment variable for control?

vllm_ascend/ops/linear_op.py
weijinqian0, line 110: dont need inherit this base class

vllm_ascend/quantization/w8a8.py
anon189Ty: Has been extracted into common function 'is_enable_nz'

vllm_ascend/ops/linear.py
anon189Ty: This parameter has been removed.

vllm_ascend/ops/linear_op.py
anon189Ty: Resolved.

PR #3364, opened 2025-10-10
[BugFix]Fix eplb problems when using dynamic eplb.

offline893:
When using dynamic eplb,it will be blocking by nz tensor.We fix these prolems by clone src tensor and recv tensor.

Qwen3_moe in A3.

- vLLM version: v0.11.0rc3

### DIFF
--- a/tests/ut/eplb/core/test_eplb_device_transfer_loader.py
+++ b/tests/ut/eplb/core/test_eplb_device_transfer_loader.py
@@ -49,8 +49,6 @@ def test_generate_task_and_state_flow(mock_adaptor):
         loader_obj.generate_expert_d2d_transfer_task([], [], {}, 0)
         assert loader_obj.comm_op_list is None
         assert loader_obj.state == loader.ExpertWeightUpdateState.WAITING
-        assert loader_obj.comm_op_list
-        assert loader_obj.recv_expert_list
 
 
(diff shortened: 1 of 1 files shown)


vllm_ascend/ops/common_fused_moe.py
```diff
@@ -227,6 +238,7 @@ def __init__(self, *args, **kwargs):
         if (self.quant_method.__class__.__name__
                 in ("GPTQMarlinMoEMethod", "CompressedTensorsWNA16MoEMethod")):
             moe_quant_params["intermediate_size_full"] = intermediate_size
+        self.quant_method.create_weights(layer=self, **moe_quant_params)
```
wangxiyuan, line 241: if this only affect eplb case, can we add the check ?
```
if eplb_enabled:
  self.quant_method.create_weights(layer=self, **moe_quant_params)
```

offline893: Creating weights is a necessary step for the proper initialization of the entire fusedmoe, whether it is torch air fused moe or common fused moe. This operation has been confirmed with relevant colleagues.

wangxiyuan: we should add a note here to explain why do the create_weights again.

PR #3366, opened 2025-10-10
[Bugfix][P/D] TP size larger than KV cache head causes accuracy issues

wangxiaoteng888:
Resolve the issue where, in the case of unequal TP (Tensor Parallelism), the TP size  is larger than the number of model attention kvcache heads, causing the KV cache to generate duplicates, which leads to transmission errors in the original code.
No
By ci
- vLLM version: v0.11.0rc3


MengqingCao:
plz update the pr message to describe the details on the fixed bug and the fix method

PR #3372, opened 2025-10-10
[Bugfix] Fix race condition in d2h transfer

jianzs:

Using non-blocking operations for device-to-host transfers can lead to data corruption in later steps. The CPU tensor is accessed right after the transfer is triggered, but the transfer might not be complete yet. As a result, the data could be wrong. This problem was seen in the A3 environment during `profile_run`.

No

CI pass.
- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/torchair/quantization/torchair_w8a8_dynamic.py
+++ b/vllm_ascend/torchair/quantization/torchair_w8a8_dynamic.py
@@ -444,7 +444,7 @@ def torchair_fused_experts_with_all2all(
         token_counts_combined = token_counts_combined.view(
             2, ep_group.world_size, -1).sum(dim=2)
         token_counts_combined_cpu = token_counts_combined.to(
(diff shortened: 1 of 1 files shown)


jianzs:
When non_blocking is set to False:
image

When non_blocking is set to True:
image

jianzs:
@wangxiyuan  @MengqingCao @Yikun  @ApsarasX  @yiz-liu  Ready to merge. I believe this fix is important and should be included in the upcoming release.

MengqingCao:
This fix is lgtm. However, I have no idea of why this non_blocking operation couldn't keep order by the stream? Maybe it is related to torch-npu?

jianzs:
I'm not sure whether PyTorch ensures that data reading always occurs after the copy operation when `non_blocking` is set to True.

MengqingCao (approved):

PR #3381, opened 2025-10-11
[BugFix]Fix moe load problems in torchair when using dynamic eplb

dsxsteven:

When using dynamic eplb, moe load is not imported. We fix this problem by modifying the return value of hidden states in torchair.

DeepseekV3 in A3.

- vLLM version: v0.11.0rc3


wangxiyuan:
please fix the CI problem

dsxsteven:
> please fix the CI problem

done

PR #2330, opened 2025-08-12
[CI] Add accuracy CI 

zhangxinyuehfad:
### What this PR does / why we need it?
Add accuracy CI
1. test  all  models accuarcy by schedule;
2. add new models accuracy yaml: Qwen3-8B, Qwen2-VL-7B-Instruct, Qwen2-7B, Qwen2-Audio-7B-Instruct

### How was this patch tested?

new models test ok: https://github.com/vllm-project/vllm-ascend/actions/runs/18067663489?pr=2330

- vLLM version: v0.11.0rc3
- vLLM main: https://github.com/vllm-project/vllm/commit/v0.11.0

PR #3394, opened 2025-10-12
Add models test and add serval new models yaml

Yikun:
This PR added Add accuracy CI for servals new models
- `ascend test / accuracy` is for PR triggered check popluar models accuracy
- `ascedn test / models` is for accuracy report, full models test, nightly model test
- Add Qwen2-Audio-7B-Instruct, Qwen2-VL-7B-Instruct, Qwen3-8B, Qwen3-VL-30B-A3B-Instruct

No

CI passed

Closes: https://github.com/vllm-project/vllm-ascend/pull/2330
Closes: https://github.com/vllm-project/vllm-ascend/pull/3362

- vLLM version: v0.11.0rc3

### DIFF
--- a/.github/workflows/_accuracy_test.yaml
+++ b/.github/workflows/_accuracy_test.yaml
@@ -66,7 +66,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: vllm-project/vllm
-          ref: v0.11.0
+          ref: ${{ inputs.vllm }}
           path: ./vllm-empty
 
       - name: Install vllm-project/vllm from source


tests/e2e/models/test_lm_eval_correctness.py
```diff
@@ -48,7 +48,7 @@ def build_model_args(eval_config, tp_size):
     }
     for s in [
             "max_images", "gpu_memory_utilization", "enable_expert_parallel",
-            "tensor_parallel_size", "enforce_eager"
+            "tensor_parallel_size", "enforce_eager", "enable_thinking"
```
MengqingCao, line 51: maybe we could remove "enforce_eager" to test aclgraph by default?

.github/workflows/_accuracy_test.yaml
```diff
@@ -0,0 +1,175 @@
...
+          INFO_FILE="/usr/local/Ascend/ascend-toolkit/${TOOLKIT_DIR}/$(uname -i)-linux/ascend_toolkit_install.info"
+          GHA_CANN_VERSION=$(grep "version=" "$INFO_FILE" \
+                           | head -n1 \
+                           | cut -d'=' -f2 \
+                           | tr -d '"')
+          {
+            echo "GHA_CANN_VERSION=$GHA_CANN_VERSION"
+            pip show torch | grep "Version:" | awk '{print "GHA_TORCH_VERSION="$2}'
+            pip show torch_npu | grep "Version:" | awk '{print "GHA_TORCH_NPU_VERSION="$2}'
+            pip show vllm | grep "Version:" | awk '{print "GHA_VLLM_VERSION="$2}' | sed 's/+.*//'
```
MengqingCao, line 139: maybe we can set `GHA_VLLM_VERSION` by `${{ inputs.vllm }}`

tests/e2e/models/test_lm_eval_correctness.py
Yikun: This is just try to map yaml to dict:
https://github.com/vllm-project/vllm-ascend/tree/main/tests%2Fe2e%2Fmodels%2Fconfigs

If we don't specify config explicitly, this will not be specified, and use aclgraph by default.

MengqingCao: Got it, that's okay then

.github/workflows/_accuracy_test.yaml
Yikun: This just get the real vllm in env, I think it's ok.
inputs.vllm should be used in https://github.com/vllm-project/vllm-ascend/pull/3394/commits/b5be0e3132b213a1e149ff6744806a0f55359520 

Done

PR #3396, opened 2025-10-12
[Feature] Enhanced MooncakeConnector to support prefix caching on the decode node

jianzs:

Previously, in Mooncake Connector, we assumed that prefix caching was not enabled during decoding, and pulling the cache always started from the first block—which was unnecessary. Enabling prefix caching during decoding helps reduce data transfer. This PR removes that assumption and handles cases where data isn't fully transmitted.

No

CI pass.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/distributed/mooncake_connector.py
+++ b/vllm_ascend/distributed/mooncake_connector.py
@@ -478,11 +478,12 @@ class MooncakeConnectorMetadata(KVConnectorMetadata):
         # Filter out the remote block ids that correspond to the unhashed local
         # block ids. We assume that the order of remote_block_ids matches the
         # order of local_block_ids.
-        remote_block_ids = []
-        for block_idx, remote_block_id in enumerate(
-                kv_transfer_params["remote_block_ids"]):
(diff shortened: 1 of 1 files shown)


jianzs:
@LCAIZJ  PTAL

vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -644,11 +655,20 @@ def update_state_after_alloc(self, request: "Request",
...
+                    num_pulled_blocks = math.ceil(num_external_tokens /
+                                                  self.block_size)
+                    assert num_pulled_blocks == len(unhashed_block_idxs), (
+                        f"The number of blocks to pull ({num_pulled_blocks}) "
+                        "does not match the number of unhashed blocks "
+                        f"({len(unhashed_block_idxs)}) for request "
+                        f"{request.request_id}.")
                     self._reqs_need_recv[request.request_id] = (
-                        request, local_block_ids)
+                        request, unhashed_block_idxs, block_ids)
```
LCAIZJ, line 671: why you need block_ids?

PR #3397, opened 2025-10-12
adapt the mla_v1 with the `mla_preprocess` kernel

kiscad:

This pull request integrates a new `mla_preprocess` kernel to create an optimized path for MLA (Multi-Layer Attention) decode operations on Ascend hardware, controlled by an environment flag. The changes include new utility functions for weight transformation, a method to prepare weights for the fused kernel, and logic to route decode-only batches to this new path. My review identified a critical bug in the `transdata` utility function where padding dimensions are swapped, which will lead to incorrect tensor shapes and kernel failures. Additionally, I've pointed out a high-severity maintainability issue in the trans_rope_weight function, which modifies its input in-place, and I have provided a pure-function alternative.

No user-facing changes by default. User can enable the `mla_preprocess` kernel in model by enable the env-var `VLLM_ASCEND_ENABLE_MLAPO`.

Dedicated Ascend kernels are not covered by our CI yet, so no extra automated tests were added. Future MLA-focused regression runs will cover this path.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -16,11 +16,13 @@ from vllm.model_executor.layers.linear import (LinearBase,
 from vllm.utils import cdiv, round_down
 from vllm.v1.attention.backends.utils import AttentionCGSupport
 
+from vllm_ascend import envs
 from vllm_ascend.ascend_config import get_ascend_config
 from vllm_ascend.attention.attention_v1 import AscendAttentionState
 from vllm_ascend.attention.utils import (AscendCommonAttentionMetadata,
                                          maybe_save_kv_layer_to_connector,
                                          split_decodes_and_prefills,
+                                         trans_rope_weight, transdata,
                                          wait_for_kv_layer_from_connector)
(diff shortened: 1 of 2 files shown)


kiscad:
Hi @Yikun @wangxiyuan
This PR is ready for review. Please take a look when you have time.
Let me know if any clarification is needed. Thanks!

whx-sjtu:
Why can't we use this kernel for decode phase in default? Are there any restrictions on this kernel now?

kiscad:
> Why can't we use this kernel for decode phase in default? Are there any restrictions on this kernel now?

The plan is to enable this kernel by default after it goes through more extensive validation in real-world scenarios. Once its stability and generalization are fully verified, it should be safe to use as the default for the decode phase.

wangxiyuan (approved):
please add a e2e test with `VLLM_ASCEND_ENABLE_MLAPO=true` to make sure this works as expect in a new PR. Thanks

kiscad:
> please add a e2e test with `VLLM_ASCEND_ENABLE_MLAPO=true` to make sure this works as expect in a new PR. Thanks

Sure. Thanks.

PR #3400, opened 2025-10-13
[Bugfix][P/D] multi_node_pd_disaggregation_mooncake.md update

wangxiaoteng888:
multi_node_pd_disaggregation_mooncake.md update. Fix issues encountered during service startup.
No
By ci

- vLLM version: v0.11.0rc3

### DIFF
--- a/docs/source/tutorials/multi_node_pd_disaggregation_mooncake.md
+++ b/docs/source/tutorials/multi_node_pd_disaggregation_mooncake.md
@@ -58,10 +58,11 @@ git clone -b pooling_async_memecpy_v1 https://github.com/AscendTransport/Mooncak
 Update and install Python
 
 ```shell
+apt-get update
 apt-get install python3
 ```
 
-Install the relevant dependencies.
+Install the relevant dependencies. The installation of Go is not required.
 
 ```shell
 cd Mooncake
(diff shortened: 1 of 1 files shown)


docs/source/tutorials/multi_node_pd_disaggregation_mooncake.md
```diff
@@ -118,6 +118,8 @@ export ASCEND_AGGREGATE_ENABLE=1
 export ASCEND_TRANSPORT_PRINT=0
 export ACL_OP_INIT_MODE=1
 export ASCEND_A3_ENABLE=1
+export LD_LIBRARY_PATH=/usr/local/Ascend/ascend-toolkit/latest/python/site-packages:$LD_LIBRARY_PATH
```
MengqingCao, line 121: why we should add this? If this is required, I think it's better to add it in the image cc @Yikun @wxsIcey

wangxiaoteng888: Because we need the libascend_transport_mem.so and libtransfer_engine.so packages from mooncake.

MengqingCao: Okay, then it's reasonable to add this in doc

Issue #3368, opened 2025-10-10
[Bug]: Qwen3-next: Out of Memory due to repeated allocation of shared hybrid cache


QilaiZhang:
### Your current environment

Runtime environment matches the setup at: https://vllm-ascend.readthedocs.io/zh-cn/latest/tutorials/multi_npu_qwen3_next.html

### 🐛 Describe the bug

When launching vLLM server for the Qwen3-Next-80B model with a ```--gpu-memory-utilization``` of 0.7, the actual memory usage reaches 0.85, exceeding the set limit. Increasing the limit ```--gpu-memory-utilization``` to 0.9 results in an "NPU out of memory" error.

The issue has been primarily traced to the ```initialize_kv_cache_tensors``` function in ```vllm_ascend/worker/model_runner_v1.py```. This function repeatedly allocates KV cache tensors within the same group, which deviates from the behavior in standard vLLM.
```python
def initialize_kv_cache_tensors(
            self, kv_cache_config: KVCacheConfig) -> dict[str, torch.Tensor]:
        """
        Initialize the memory buffer for KV cache.

        Args:
            kv_cache_config: The KV cache config
        Returns:
            Dict[str, torch.Tensor]: A map between layer names to their
            corresponding memory buffer for KV cache.
        """
        # init kv cache tensors
        kv_cache_raw_tensors: dict[str, Union[torch.Tensor,
                                              Optional[torch.Tensor]]] = {}
        # llmdatadist need the addr of cache tensor be aligned with 2M
        alignment = 2 * 1024 * 1024
        for kv_cache_tensor in kv_cache_config.kv_cache_tensors:
            # TODO: REFACTOR ME to sharing hybrid cache
            for idx in range(len(kv_cache_tensor.shared_by)):
                layer_name = kv_cache_tensor.shared_by[idx]
                if "linear_attn" in layer_name:
                    # for mamba linear attention
                    for layer_name_inner in kv_cache_tensor.shared_by:
                        if ("attn" in layer_name_inner and "linear_attn" not in layer_name_inner) or \
                            layer_name_inner in kv_cache_raw_tensors.keys():
                            continue
                        if self.vllm_config.kv_transfer_config is None:
                            tensor = torch.zeros(kv_cache_tensor.size,
                                                 dtype=torch.int8,
                                                 device=self.device)
                        else:
                            cache_size_aligned = kv_cache_tensor.size + alignment
                            tensor = torch.zeros(cache_size_aligned,
[...]

PR #3404, opened 2025-10-13
[Bugfix] Fix duplicated KV cache allocation in qwen3-Next

QilaiZhang:
Fix the bug described in https://github.com/vllm-project/vllm-ascend/issues/3368: Duplicated Allocation of Shared Hybrid Cache in Qwen3-Next

No
The model qwen3-next has been tested.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2874,7 +2874,7 @@ class NPUModelRunner(LoRAModelRunnerMixin):
                 layer_name = kv_cache_tensor.shared_by[idx]
                 if "linear_attn" in layer_name:
                     # for mamba linear attention
-                   kv_cache_raw_tensors[layer_name] = tensor
+                    kv_cache_raw_tensors[layer_name] = tensor
                 elif "attn" in layer_name:
                     # for other attentions, e.g., self_attn, sliding window attn
                     kv_cache_raw_tensors[layer_name] = (k_tensor, v_tensor)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2850,49 +2850,33 @@ def initialize_kv_cache_tensors(
...
-                                                 device=self.device)
-                        else:
-                            cache_size_aligned = kv_cache_tensor.size + alignment
-                            tensor = torch.zeros(cache_size_aligned,
-                                                 dtype=torch.int8,
-                                                 device=self.device)
-                            tensor = self._align_memory(
-                                tensor, alignment)[:kv_cache_tensor.size]
-                        kv_cache_raw_tensors[layer_name_inner] = tensor
+                    kv_cache_raw_tensors[layer_name] = tensor
```
MengqingCao, line 2877: Sharing kv cache memory between `linear_attn` and `self_attn` will cause accuracy issue in Qwen3-Next. Thus we should allocate memory for `linear_attn` and `self_attn` seperately.
Actually I also have a PR on fixing this issue and refactor the kvcache initialization logic, but it is not ready for review now. Feel free to change the bug fix logic in this pr according to #3106

MengqingCao:
@QilaiZhang plz take a look at https://github.com/vllm-project/vllm-ascend/pull/3760, which shared kvcache in the same attention type, before fixing accuracy issue on sharing buffer between `linear_attn` and `self_attn`, I prefer to waste some memory to guarantee the accuracy

QilaiZhang:
> @QilaiZhang plz take a look at #3760, which shared kvcache in the same attention type, before fixing accuracy issue on sharing buffer between `linear_attn` and `self_attn`, I prefer to waste some memory to guarantee the accuracy

@MengqingCao Thanks for flagging this. Acknowledged - we'll preserve the current memory usage to guarantee accuracy. We'll look into #3760 and revisit buffer sharing once the accuracy issues are fixed.

PR #3411, opened 2025-10-13
[Fix] Fix mc2_tokens_capacity-related issues

yiz-liu:
Replaces the hardcoded `mc2_tokens_capacity` with the max graph capture size for a more accurate allocation.

This change ensures the capacity is correctly sized relative to the graph capture configuration, removing a magic number and making the setup more robust.

This PR fixes two issues:

1. <del>MC2 op restrictions differ between SoCs.</del> @Angazenn This requires an overhaul, hence removed from this PR, please commit another PR.
2. The hardcoded value `512` allocates too much buffer for large models.

None.

Tested in daily checks.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -461,10 +461,13 @@ class NPUModelRunner(LoRAModelRunnerMixin):
         # NOTE: To be clear, we need to make sure that during graph capture, the number of
         # tokens is less than or equal to mc2_tokens_capacity. According to _set_cudagraph_sizes,
         # the max number of tokens in graph is min(max_num_seqs * 2, 512).
-        max_graph_size = self.compilation_config.cudagraph_capture_sizes[0]
+        if self.compilation_config.cudagraph_capture_sizes:
+            max_num_tokens = self.compilation_config.cudagraph_capture_sizes[0]
+        else:
+            max_num_tokens = self.max_num_reqs * self.uniform_decode_query_len
         tp_size = self.parallel_config.tensor_parallel_size
         # Use integer arithmetic for ceiling division.
-        num_tokens_per_tp_rank = (max_graph_size + tp_size - 1) // tp_size
+        num_tokens_per_tp_rank = (max_num_tokens + tp_size - 1) // tp_size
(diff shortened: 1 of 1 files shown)


yiz-liu:
@Angazenn @linfeng-yuan Please take a look at this, does this resolve your questions?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -458,10 +458,12 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
...
 
-        # NOTE: Technically, MC2 can have 512 tokens each rank, but this will consume too much memory. The formula is:
-        # ((maxBs * tokenNeedSizeDispatch * ep_worldsize * localMoeExpertNum) + (maxBs * tokenNeedSizeCombine * (k + sharedExpertNum))) * 2
-        # so we have to limit the MC2 tokens to save memory, should fix this in the future.
-        self.mc2_tokens_capacity = 512
+        # NOTE: To be clear, we need to make sure that during graph capture, the number
+        # of tokens is less than or equal to mc2_tokens_capacity. According to
+        # VllmConfig._set_cudagraph_sizes, the max graph size is min(max_num_seqs * 2, 512),
+        # so we set mc2_tokens_capacity to max_num_seqs // tp_size.
+        self.mc2_tokens_capacity = self.scheduler_config.max_num_seqs // \
```
linfeng-yuan: We don't need to divide by tp_size, but instead we need to use `--max-num-seqs` or `max_graph_size as` the mc2_tokens_capacity, rather than 512. Additionally, `mc2_tokens_capacity` should be a multiple of `tp_size` and multiplied by `(1 + num_spec_tokens)`.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -458,10 +458,12 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
...
+            self.parallel_config.tensor_parallel_size
```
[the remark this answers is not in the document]
yiz-liu: Fixed.

vllm_ascend/worker/model_runner_v1.py
yiz-liu: Got it.

PR #3419, opened 2025-10-13
bugfix for mtp with multistream_moe

zouyida2052:
when infer deepseek mtp layer with multistream_moe, we should pass a boolean to evaluate this feature and fix bugs when we are  in mtp layer

no

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/attention/utils.py
+++ b/vllm_ascend/attention/utils.py
@@ -101,8 +101,6 @@ def split_decodes_and_prefills(
         return num_reqs, 0, num_tokens, 0
 
     first_prefill = is_prefill.int().argmax(dim=-1).item()
-    assert torch.all(query_lens[first_prefill:] >= decode_threshold)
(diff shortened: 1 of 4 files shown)


vllm_ascend/attention/utils.py
```diff
@@ -101,7 +101,7 @@ def split_decodes_and_prefills(
         return num_reqs, 0, num_tokens, 0
 
     first_prefill = is_prefill.int().argmax(dim=-1).item()
-    assert torch.all(query_lens[first_prefill:] > decode_threshold)
+    assert torch.all(query_lens[first_prefill:] >= decode_threshold)
```
zzzzwwjj: This assert is unnecessary, and in some corner case, it will lead to asserterror.

hust17yixuan:
approval，this pr can fix the moe_multistream bugs in mtp

PR #3421, opened 2025-10-13
Fix warning msg print

SunnyLee151064:
Avoid printing some warning msg as below :
UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach ...
No

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -75,7 +75,7 @@ class AscendMLAPrefillMetadata:
         chunk_seq_lens: torch.Tensor
 
     attn_mask: torch.Tensor
-    query_lens: list[int]
+    query_lens: torch.Tensor
     seq_lens: list[int]
     context_lens: torch.Tensor
     input_positions: torch.Tensor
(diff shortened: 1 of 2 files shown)


vllm_ascend/attention/mla_v1.py
```diff
@@ -742,9 +742,8 @@ def _forward_prefill(
                                    k_rope=k_pe,
                                    value=value,
                                    mask=self.prefill_mask,
-                                   seqlen=torch.tensor(
-                                       attn_metadata.prefill.query_lens,
-                                       dtype=torch.int32),
+                                   seqlen=attn_metadata.prefill.query_lens.to(
+                                       torch.int32),
```
whx-sjtu: Perform this `to(torch.int32)` operation also in `build` of attention metadata for performance consideration.

vllm_ascend/attention/mla_v1.py
```diff
@@ -742,9 +742,7 @@ def _forward_prefill(
                                    k_rope=k_pe,
                                    value=value,
                                    mask=self.prefill_mask,
-                                   seqlen=torch.tensor(
-                                       attn_metadata.prefill.query_lens,
-                                       dtype=torch.int32),
+                                   seqlen=attn_metadata.prefill.query_lens.to(torch.int32),
```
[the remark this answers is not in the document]
whx-sjtu: Also, don't forget to change the data type definition of `query_lens` for readablity. It should be torch.Tensor instead of List[int]

vllm_ascend/attention/mla_v1.py
SunnyLee151064: Modified yet

yiz-liu:
@zzzzwwjj Please take a look at this, also related to attention metadata.

PR #3429, opened 2025-10-13
[BugFix] Add addtional check for mc2 on different hardwares.

Angazenn:
This if an extension based on #3411. Hardware-specific restriction for mc2 is added. Besides, I also limit the scenarios where this check take effects to only moe models that might uses mc2.

For situations that mc2 is not supported with excessive input tokens, we throws an error earlier in vLLM-Ascend.

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -664,7 +664,7 @@ class NPUModelRunner(LoRAModelRunnerMixin):
         # tokens is less than or equal to mc2_tokens_capacity. According to _set_cudagraph_sizes,
         # the max number of tokens in graph is min(max_num_seqs * uniform_decode_query_len, 512).
         if self.compilation_config.cudagraph_capture_sizes:
-            max_num_tokens = self.compilation_config.cudagraph_capture_sizes[0]
+            max_num_tokens = self.compilation_config.max_cudagraph_capture_size
         else:
             # NOTE: To save memory, we cap the max number of tokens to 512.
             max_num_tokens = min(
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -433,7 +433,11 @@ def _is_default_capture_sizes(vllm_config: VllmConfig) -> bool:
             cudagraph_capture_sizes += list(
                 range(256, max_cudagraph_capture_size + 1, 16))
 
-    if sorted(cudagraph_capture_sizes, reverse=True) == \
+    if vllm_version_is("0.11.0"):
+        target_cudagraph_capture_sizes = sorted(cudagraph_capture_sizes, reverse=True)
+    else:
+        target_cudagraph_capture_sizes = sorted(cudagraph_capture_sizes)
+    if target_cudagraph_capture_sizes == \
             vllm_config.compilation_config.cudagraph_capture_sizes:
         return True
 


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -458,15 +459,43 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
...
+
+            soc_version = get_ascend_soc_version()
+            limit = None
+            if soc_version in {AscendSocVersion.A3}:
+                limit = 512
+            elif soc_version in {AscendSocVersion.A2}:
+                limit = 256
+
+            if limit is not None and num_tokens_per_tp_rank > limit:
+                raise ValueError(
```
whx-sjtu: Perhaps in this scenario, it would be better to fall back to other moe methods that support current token size (such as all2all) rather than just reporting an error? I think it is weired to force users to use smaller `max_num_seqs` due to kernel restriction.

Angazenn: It seems that we can make a better selection on this. But the problem is that aclgraph mode requires a unified communication schema. So we need to ensure that each communication ops in a model keep unchanged with respect to different capture sizes. cc @yiz-liu

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1779,6 +1768,52 @@ def _pool(
             kv_connector_output=kv_connector_output,
         )
 
+    def _is_moe_model(self):
```
wangxiyuan: use `vllm_ascend.utils.is_moe_model`

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2465,11 +2500,14 @@ def profile_run(self) -> None:
             # MC2 will consume additional NPU memory.
             # Therefore, we need to run the MC2 path once here to complete its initialization,
             # allowing vLLM to correctly estimate the maximum memory required.
-            if self.max_num_tokens > self.mc2_tokens_capacity and \
-                self._select_moe_comm_method(
-                    self.mc2_tokens_capacity,
-                    with_prefill=True) == MoECommType.MC2:
-                self._dummy_run(self.mc2_tokens_capacity, with_prefill=True)
+            if self._is_moe_model():
```
weijinqian0: may replace this with model_config.get_num_experts()

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -458,22 +459,10 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
...
-        self.mc2_tokens_capacity = num_tokens_per_tp_rank * tp_size
-        self.reserved_mc2_mask = torch.zeros(
-            self.mc2_tokens_capacity,
-            dtype=torch.bool,
-            device=self.device,
-        )
+        # Since not all models have moe modules and requires mc2, we leave
+        # the initialization of mc2 related parameters later.
+        self.mc2_tokens_capacity = 0
+        self.reserved_mc2_mask = None
```
wangxiyuan: nice change

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2465,11 +2500,14 @@ def profile_run(self) -> None:
...
             # allowing vLLM to correctly estimate the maximum memory required.
-            if self.max_num_tokens > self.mc2_tokens_capacity and \
-                self._select_moe_comm_method(
-                    self.mc2_tokens_capacity,
-                    with_prefill=True) == MoECommType.MC2:
-                self._dummy_run(self.mc2_tokens_capacity, with_prefill=True)
+            if self._is_moe_model():
+                self._initialize_mc2()
+                if self.max_num_tokens > self.mc2_tokens_capacity and \
+                    self._select_moe_comm_method(
```
wangxiyuan: `_select_moe_comm_method` has been called in `_initialize_mc2` and is called in `_dummy_run` . You should consider to merge them into one.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -543,8 +551,46 @@ def _init_mc2_tokens_capacity(self):
...
+        # different hardwares.
+        if self._select_moe_comm_method(self.mc2_tokens_capacity,
+                                        with_prefill=False) == MoECommType.MC2:
+
+            # MC2 will be applied in runtime. Therefore we check whether
+            # the number of input tokens exceed the limit of MC2.
+
+            soc_version = get_ascend_soc_version()
+            limit = None
+            if soc_version in {AscendSocVersion.A3}:
```
wangxiyuan: Add a TODO here

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -532,7 +528,19 @@ def _may_pad_kv_consumer_num_seq(self):
         # exceeding a sequence length limit (16 tokens) in npu_fused_infer_attention_score operation
         pass
 
-    def _init_mc2_tokens_capacity(self):
+    def _init_mc2(self):
+        """Initialization of MC2-related parameters and verify the validity."""
+
```
JC-ut0: Need to change `_init_mc2_tokens_capacity` in `torchair_model_runner.py` as well.

Angazenn: sure, I will revert this change.

Issue #3241, opened 2025-09-28
[RFC]: native kvcache offloading


HF-001:
### Motivation.

Currently, in vLLM v1 there is  in-house solution for offloading KV cache data from the GPU memory to other medium (in particular, CPU memory).But it is not yet available in vllm-ascend, so I hope to add this feature to vllm -ascend.

There is a proposed RFC ([#19854](https://github.com/vllm-project/vllm/issues/19854)) from vllm，and PRs([#19848](https://github.com/vllm-project/vllm/pull/19848)，[#20075](https://github.com/vllm-project/vllm/pull/19848)， [#21448](https://github.com/vllm-project/vllm/pull/21448)， [#22595](https://github.com/vllm-project/vllm/pull/22595) and [#24251](https://github.com/vllm-project/vllm/pull/24251))

### Proposed Change.

To enable this feature，we need add a CPUOffloadingSpec compatible with vllm-ascend，and add a CpuNpuOffloadingHandler for kvcache offloading and loading(cpu->npu and npu-> cpu).

jianzs:
I'm also interested in this. I'll look into how to implement it, and PRs are welcome too.

HF-001:
@jianzs hi, the pr is here:https://github.com/vllm-project/vllm-ascend/pull/3433

PR #3433, opened 2025-10-14
[Feat] add native kvcache offload

HF-001:
This pr is for https://github.com/vllm-project/vllm-ascend/issues/3241 , which is in-house solution for offloading KV cache data from the GPU memory to other medium (in particular, CPU memory)。Previous solutions required reliance on third-party components, which had issues with compatibility between different versions.

test results: TTFT -40.84%( 595.7ms -> 352.61ms)  TPOT -34.57%(85ms->56.06ms)

 main(without kvcache offload):
image

this pr(with kvcache offload):
image

run script of main(without kvcache offload):
export CUDA_VISIBLE_DEVICES=0
export TP=1
export MODEL_PATH=/model/Qwen3-14B
export MODEL_NAME=Qwen3-14B
export PORT=10000
#export ASCEND_LAUNCH_BLOCKING=1
#export ASCEND_SLOG_PRINT_TO_STDOUT=1
python3 -m vllm.entrypoints.openai.api_server  --host 0.0.0.0 --port ${PORT} --dtype bfloat16 --model ${MODEL_PATH} --served-model-name ${MODEL_NAME} --tensor-parallel-size ${TP} --gpu-memory-utilization 0.7   --max-model-len  32768 --trust-remote-code --disable-log-requests

run script of this pr:
export CUDA_VISIBLE_DEVICES=0
export TP=1
export MODEL_PATH=/model/Qwen3-14B
export MODEL_NAME=Qwen3-14B
export PORT=10000
#export ASCEND_LAUNCH_BLOCKING=1
#export ASCEND_SLOG_PRINT_TO_STDOUT=1
python3 -m vllm.entrypoints.openai.api_server  --host 0.0.0.0 --port ${PORT} --dtype bfloat16 --model ${MODEL_PATH} --served-model-name ${MODEL_NAME} --tensor-parallel-size ${TP} --gpu-memory-utilization 0.7   --max-model-len  32768 --trust-remote-code --disable-log-requests \
    --block-size 128 \
    --kv-transfer-config '{"kv_connector":"OffloadingConnector","kv_role":"kv_both","kv_connector_extra_config":{"block_size": 128, "num_cpu_blocks": 1000, "spec_name":"NPUOffloadingSpec", "spec_module_path": "vllm_ascend.kv_offload.npu"}}'

test script:
export MODEL_NAME=/model/Qwen3-14B/

python /model/vllm/benchmarks/multi_turn/benchmark_serving_multi_turn.py --url http://127.0.0.1:10000 --model $MODEL_NAME --served-model-name Qwen3-14B --seed 1234 --input-file /model/vllm/benchmarks/multi_turn/generate_multi_turn.json \
--num-clients 16 --max-active-conversations 48

- vLLM version: v0.11.0rc3

### DIFF
--- a/csrc/torch_binding.cpp
+++ b/csrc/torch_binding.cpp
@@ -48,9 +48,7 @@ void swap_blocks_impl(torch::Tensor& src, torch::Tensor& dst,
   } else if (src_device.is_cpu() && (!dst_device.is_cpu())) {
     memcpy_type = ACL_MEMCPY_HOST_TO_DEVICE;
   } else {
-    TORCH_WARN("src tensor device: ", src_device);
-    TORCH_WARN("dst tensor device: ", dst_device);
-    TORCH_CHECK(false, "Invalid device combination");
+    TORCH_CHECK(false, "Invalid device combination, src tensor device: ", src_device, ", dst tensor device: ", dst_device);
   }
 
   TORCH_CHECK(block_mapping.device().is_cpu(), "block_mapping must be on CPU");
--- a/csrc/torch_binding.cpp
+++ b/csrc/torch_binding.cpp
@@ -17,15 +17,79 @@
 #include <torch/extension.h>
 #include <torch/library.h>
 #include <torch/version.h>
+#include <torch/torch.h>
 #include <torch_npu/csrc/core/npu/NPUStream.h>
 #include <torch_npu/csrc/framework/OpCommand.h>
+#include "torch_npu/csrc/core/npu/NPUGuard.h"
 #include <torch_npu/csrc/npu/Module.h>
 #include "acl/acl.h"
+#include "acl/acl_rt.h"
 #include "ops.h"
 #include "utils.h"
 #include "mla_preprocess/op_host/mla_preprocess.h"
 
+#include <c10/core/Device.h>
+#include <c10/util/Exception.h>
+#include <c10/util/Logging.h>
+
 namespace vllm_ascend {
+void swap_blocks_impl(torch::Tensor& src, torch::Tensor& dst,
+                 const torch::Tensor& block_mapping, aclrtStream stream) {
+  torch::Device src_device = src.device();
+  torch::Device dst_device = dst.device();
+  aclrtMemcpyKind memcpy_type;
+
+  if ((!src_device.is_cpu()) && (!dst_device.is_cpu())) {
+    TORCH_CHECK(src_device.index() == dst_device.index(),
+                "src and dst must be on the same npu");
+    memcpy_type = ACL_MEMCPY_DEVICE_TO_DEVICE;
+  } else if ((!src_device.is_cpu()) && dst_device.is_cpu()) {
+    memcpy_type = ACL_MEMCPY_DEVICE_TO_HOST;
+  } else if (src_device.is_cpu() && (!dst_device.is_cpu())) {
+    memcpy_type = ACL_MEMCPY_HOST_TO_DEVICE;
+  } else {
+    TORCH_WARN("src tensor device: ", src_device);
+    TORCH_WARN("dst tensor device: ", dst_device);
+    TORCH_CHECK(false, "Invalid device combination");
+  }
+
+  TORCH_CHECK(block_mapping.device().is_cpu(), "block_mapping must be on CPU");
+
+  char* src_ptr = static_cast<char*>(src.data_ptr());
+  char* dst_ptr = static_cast<char*>(dst.data_ptr());
+
+  const int64_t block_size_in_bytes = src.element_size() * src.stride(0);
+  
+  const int64_t num_blocks = block_mapping.size(0);
+  const int64_t max_src_block = src.size(0);
+  const int64_t max_dst_block = dst.size(0);
+  for (size_t i = 0; i < num_blocks; i++) {
+    int64_t src_block_number = block_mapping[i][0].item<int64_t>();
+    int64_t dst_block_number = block_mapping[i][1].item<int64_t>();
+    TORCH_CHECK(src_block_number >= 0 && src_block_number <= max_src_block,
+                    "src block index ", src_block_number, " out of range (max: ", max_src_block, ")");
+    TORCH_CHECK(dst_block_number >= 0 && dst_block_number <= max_dst_block,
+                    "dst block index ", dst_block_number, " out of range (max: ", max_dst_block, ")");
+    
+    int64_t src_offset = src_block_number * block_size_in_bytes;
+    int64_t dst_offset = dst_block_number * block_size_in_bytes;
+
+    aclrtMemcpyAsync(dst_ptr + dst_offset, block_size_in_bytes,
+                          src_ptr + src_offset, block_size_in_bytes,
+                          memcpy_type, stream);
+  }
+}
+
+void swap_blocks(torch::Tensor &x, torch::Tensor &y, const torch::Tensor &z)
+{    
+  
+  const c10_npu::OptionalNPUGuard npuGuard(
+        (!x.device().is_cpu()) ? x.device() : y.device()
+    );
+  aclrtStream stream = c10_npu::getCurrentNPUStream().stream();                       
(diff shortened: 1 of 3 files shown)


jianzs:
In my opinion, `NPUOffloadingSpec` is unnecessary. Our goal is also to offload kv blocks to the CPU, so the objective of `CPUOffloadingSpec` aligns with ours. The key is to enable `CPUOffloadingSpec` to support more handlers for various mediums. Therefore, we may need to encourage vLLM to make code adjustments.

At the development stage, I think we can temporarily modify `CPUOffloadingSpec` by replacing `CpuGpuOffloadingHandler` with `CpuNpuOffloadingHandler` for debugging purposes.

image

HF-001:
@jianzs If the CPUOffloadingSpec of vllm can be adjusted more flexibly, NPUOffloadingSpec is not needed.

csrc/torch_binding.cpp
```diff
@@ -17,15 +17,77 @@
...
+  if ((!src_device.is_cpu()) && (!dst_device.is_cpu())) {
+    TORCH_CHECK(src_device.index() == dst_device.index(),
+                "src and dst must be on the same npu");
+    memcpy_type = ACL_MEMCPY_DEVICE_TO_DEVICE;
+  } else if ((!src_device.is_cpu()) && dst_device.is_cpu()) {
+    memcpy_type = ACL_MEMCPY_DEVICE_TO_HOST;
+  } else if (src_device.is_cpu() && (!dst_device.is_cpu())) {
+    memcpy_type = ACL_MEMCPY_HOST_TO_DEVICE;
+  } else {
+    TORCH_CHECK(false, "Invalid device combination");
```
realliujiaxu: print `src_device` and `dst_device` maybe better when check fails

csrc/torch_binding.cpp
```diff
@@ -17,15 +17,77 @@
...
+    int64_t dst_block_number = block_mapping[i][1].item<int64_t>();
+    TORCH_CHECK(src_block_number >= 0 && src_block_number <= max_src_block,
+                    "src block index ", src_block_number, " out of range (max: ", max_src_block, ")");
+    TORCH_CHECK(dst_block_number >= 0 && dst_block_number <= max_dst_block,
+                    "dst block index ", dst_block_number, " out of range (max: ", max_dst_block, ")");
+    
+    int64_t src_offset = src_block_number * block_size_in_bytes;
+    int64_t dst_offset = dst_block_number * block_size_in_bytes;
+
+    aclrtMemcpyAsync(dst_ptr + dst_offset, block_size_in_bytes,
```
realliujiaxu, line 75: func name is swap_blocks_npu, but there is only copy from src to dst?

HF-001: @realliujiaxu yes，this func is copy from src to dst， the name is refer to vllm, which is swap_blocks. do you think swap_blocks is better?

csrc/torch_binding.cpp
HF-001: @realliujiaxu ok i will add print here

HF-001:
@realliujiaxu hi, based on your comments, I have completed the modifications. Can we merge now?

csrc/torch_binding.cpp
realliujiaxu: yes, swap_blocks is better

HF-001:
@realliujiaxu hi，can you help merge this pr? or is there anything else that needs to be modified?

realliujiaxu:
> @realliujiaxu hi，can you help merge this pr? or is there anything else that needs to be modified?

Sorry, I'm not a maintainer. @jianzs @wangxiyuan please take a look.

jianzs:
@HF-001  Have you tested the full functionality? This includes the entire process of offloading, request eviction, and KV cache loading.

HF-001:
@jianzs yes, I have conducted comprehensive functional and performance testing. The functional testing not only verified the features/processes you mentioned, but also recorded detailed logs of intermediate processes for validation. Additionally, in the performance testing, enabling KVCache offload reduced TTFT by 30% to 45%.

jianzs:
> @jianzs yes, I have conducted comprehensive functional and performance testing. The functional testing not only verified the features/processes you mentioned, but also recorded detailed logs of intermediate processes for validation. Additionally, in the performance testing, enabling KVCache offload reduced TTFT by 30% to 45%

Could you share your testing approach and results in this pr?

HF-001:
@jianzs firstly, the functional testing has two ways: 1) when running model, set log level to debug, will see store and load log from vllm; 2) add detail log in cpu_npu.py and torch_bind.cpp. secondly, the performance testing is by vllm multi_turn berchmark.

jianzs:
@HF-001  I'm testing a scenario where vLLM evicts certain requests when the KV cache runs out of space. With CPU offloading enabled, I expected that upon resuming scheduling, it would simply load the previously computed KV cache from DRAM and only calculate the KV cache for a small number of tokens. But my tests indicate that it's still doing a full recomputation.

HF-001:
@jianzs Perhaps this is as expected, as the request only undergoes kvcache loading during the first scheduling. do you think so?

jianzs:
@HF-001  When I set num_cpu_blocks to a large value, such as 131072, the engine fails to start.

jianzs:
> @jianzs Perhaps this is as expected, as the request only undergoes kvcache loading during the first scheduling. do you think so?

When resuming scheduling after eviction, the scheduler will also attempt to contact the connector to check for any matching blocks—this is the primary scenario your feature is designed to handle.

HF-001:
> @HF-001 When I set num_cpu_blocks to a large value, such as 131072, the engine fails to start.
@jianzs I have never set such a large number before. I have tested 1000 and 20000 on VLLM and A800, and there is no difference in TTFT

HF-001:
> > @jianzs Perhaps this is as expected, as the request only undergoes kvcache loading during the first scheduling. do you think so?
> 
> When resuming scheduling after eviction, the scheduler will also attempt to contact the connector to check for any matching blocks—this is the primary scenario your feature is designed to handle.

@jianzs I haven't tested this scenario yet, I'm considering how to validate it. Do you think there are any other tests that need to be supplemented

jianzs:
> > @HF-001 When I set num_cpu_blocks to a large value, such as 131072, the engine fails to start.
> > @jianzs I have never set such a large number before. I have tested 1000 and 20000 on VLLM and A800, and there is no difference in TTFT

I've identified the cause of the startup failure: the requested host memory exceeds the memory limit.

jianzs:
> When resuming scheduling after eviction, the scheduler will also attempt to contact the connector to check for any matching blocks—this is the primary scenario your feature is designed to handle.

@HF-001  My apologies, that was my mistake. This feature doesn't support that particular scenario. However, I did test it with the prefix caching scenario, and it works.

vllm_ascend/kv_offload/cpu_npu.py
```diff
@@ -0,0 +1,168 @@
+import numpy as np
```
wangxiyuan, line 1: how about moving this module to distrubuted/cpu_offload_manager?

HF-001: > how about moving this module to distrubuted/cpu_offload_manager?

@wangxiyuan this position and naming convention are designed to align with vllm.If we don't consider consistency with VLLM, I think the way you said would be better.

wangxiyuan: I didn't find a module named kv_offload in vLLM. can you give me any link? Thanks.

HF-001: @wangxiyuan yes，This directory is only available in vllm after 0.10.2。the link:  https://github.com/vllm-project/vllm/tree/main/vllm/v1/kv_offload

wangxiyuan (approved):

vllm_ascend/kv_offload/cpu_npu.py
wangxiyuan: Thanks, it make senses

PR #3432, opened 2025-10-14
[Refactor] Adapt deepseek-v3.2 to vllm 0.11.0

MengqingCao:
### What this PR does / why we need it?
Adapt deepseek-v3.2 to vllm 0.11.0, removing the useless patch.

The final goal is to remove all the patches and align the code arch to vllm, thus we need to do the following work in next prs.
TODO:
- [x] remove patch on attention spec
- [ ] refactor the kvcache creation logic

### Does this PR introduce _any_ user-facing change?
N/A

### How was this patch tested?
1. CI passed with existing test.
2. Test pass with deepseek-v3.2-exp

- vLLM version: v0.11.0rc3
- vLLM main: https://github.com/vllm-project/vllm/commit/v0.11.0

PR #3438, opened 2025-10-14
[KVCache] Refactor KVCache as page_size_bytes is ineffective

MengqingCao:
Refactor KVCache as page_size_bytes is ineffective.

1. Currently the `AttentionSpec` is patched, but the `page_size_bytes` is still using that in vLLM in runtime, thus the patch is not working actually. Thus this pr removes the patch on `AttentionSpec`, and will do the final fix in vLLM.
2. Use `MLAAttentionSpec` instead of `FullAttentionSpec` to reduce `page_size_bytes` of spec, so that num_blocks in spec could double

- vLLM version: v0.11.0rc3


wxsIcey:
Qwen3-next test passed.
image

MengqingCao:
DeepSeek-V3.2-Exp-W8A8 works fine with this pr + #3432 + vllm 0.11.0
image

PR #3448, opened 2025-10-14
[Bugfix] Route requests requiring KVC recomputation from the decode instance to the P instance

Shirley125:
This PR is aimed to fix the recomputing out of memory bug in decode instance. When recomputing happens in decode, kv cache usage may exceed the pre-allocated memory, and it will cause OOM.

So we propose a new scheduling strategy,  when decode instance cannot allocate new block for running requests, we will stop the request that will be preempted. These stopped request will be recognied by proxy, and they will be send to prefill instance again to calculate kvc and then direct to decode instance. 

This is a temporary plan to fix the bug. The long-term stratege is to use CPU offload in decode instance.

An extra ascend configuration option **-- recompute_scheduler_enable = True** is added to enable this strategy. The default value is False

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/core/recompute_schedule_config.py
+++ b/vllm_ascend/core/recompute_schedule_config.py
@@ -1,0 +1,39 @@
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


wangxiyuan (approved):

vllm_ascend/ascend_config.py
```diff
@@ -72,6 +72,8 @@ def __init__(self, vllm_config):
         ) and not self.torchair_graph_config.enabled and vllm_config.parallel_config.enable_expert_parallel
         self.multistream_overlap_shared_expert = additional_config.get(
             "multistream_overlap_shared_expert", False)
+        self.recompute_scheduler_enable = additional_config.get(
+            "recompute_scheduler_enable", True)
```
wangxiyuan: the default vlaue should be false

Shirley125: OK, thanks

~
```diff
@@ -0,0 +1,51 @@
+# This is a combination of 3 commits.
```
wangxiyuan: useless file. please remove it

Shirley125: already removed

PR #3456, opened 2025-10-14
add new e2e tests for aclgraph

lilinsiman:
Add new e2e tests for aclgraph:
1. Memory test during graph capture
2. Aclgraph auto enable test in init, capture and reply process

no

ut

- vLLM version: v0.11.0rc3

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -89,6 +89,7 @@ jobs:
           # the test separately.
 
           pytest -sv tests/e2e/singlecard/test_aclgraph.py
+          pytest -sv tests/e2e/singlecard/test_aclgraph_mem.py
           pytest -sv tests/e2e/singlecard/test_ascend_scheduler.py
           pytest -sv tests/e2e/singlecard/test_camem.py
           pytest -sv tests/e2e/singlecard/test_chunked.py
--- a/tests/e2e/singlecard/test_aclgraph_mem.py
+++ b/tests/e2e/singlecard/test_aclgraph_mem.py
@@ -1,0 +1,99 @@
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+# Copyright 2025 The vLLM team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import multiprocessing
+import os
+from unittest.mock import patch
+
+import pytest
+import torch
+from modelscope import snapshot_download  # type: ignore
+from vllm import LLM, SamplingParams
+
+from vllm_ascend.worker.model_runner_v1 import NPUModelRunner
+
+MODELS = ["Qwen/Qwen3-0.6B", "vllm-ascend/DeepSeek-V2-Lite-W8A8"]
+
+
+@pytest.mark.skipif(os.getenv("VLLM_USE_V1") == "0",
+                    reason="aclgraph only support on v1")
+@pytest.mark.parametrize("model", MODELS)
+@pytest.mark.parametrize("max_tokens", [4])
+@patch.dict(os.environ, {"ASCEND_RT_VISIBLE_DEVICES": "0,1"})
+def test_aclgraph_mem_use(model: str, max_tokens: int) -> None:
+    del os.environ["VLLM_WORKER_MULTIPROC_METHOD"]
+    capture_called = multiprocessing.Value("i", 0)  # int, 0 or 1
+    capture_mem_before = multiprocessing.Value("q", -1)  # long long (64-bit)
+    capture_mem_after = multiprocessing.Value("q", -1)  # long long
+
+    def capture_model_wrapper(original_method):
+
+        def wrapped(self):
+            mem_before = torch.npu.mem_get_info()[0]  # free memory
+            result = original_method(self)
+            mem_after = torch.npu.mem_get_info()[0]
+            with capture_called.get_lock():
+                capture_called.value = 1
+                capture_mem_before.value = mem_before
+                capture_mem_after.value = mem_after
+            return result
+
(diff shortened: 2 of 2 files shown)


tests/e2e/multicard/test_aclgraph_enable_mem.py
```diff
@@ -0,0 +1,184 @@
...
+
+    assert capture_called.value == 1, "_capture_model was not called during test"
+    assert capture_mem_before.value != -1, "capture_mem_before not set"
+    assert capture_mem_after.value != -1, "capture_mem_after not set"
+
+    print("Main process: capture_mem_before =", capture_mem_before.value)
+    print("Main process: capture_mem_after =", capture_mem_after.value)
+
+    mem_used_by_capture = capture_mem_before.value - capture_mem_after.value
+    max_mem_expected = 0.33 * (1024**3) * 1.3
```
yiz-liu: Why do we have this assumption? Could you please add some notes?

tests/e2e/multicard/test_aclgraph_enable_mem.py
```diff
@@ -0,0 +1,184 @@
...
+    actual_replay = replay_counter.value
+
+    num_hidden_layers = num_hidden_layers_shared.value
+    num_acl_graphs = num_hidden_layers + 1  # per rank
+    num_inference_steps = max_tokens + 1    # first token + max_tokens
+
+    expected_replay_per_rank = num_acl_graphs * num_inference_steps
+    expected_total_replay = expected_replay_per_rank * dp_size
+
+    assert actual_capture > 0, "No graph captured!"
```
yiz-liu: We can actually validate the number of captured graphs right? Please add this.

tests/e2e/multicard/test_aclgraph_replay_capture_mem.py
```diff
@@ -0,0 +1,195 @@
...
+import os
+from unittest.mock import patch
+
+import pytest
+import torch
+from vllm import LLM, SamplingParams
+from vllm_ascend.utils import MAX_CAPTURE_SIZE
+from vllm_ascend.worker.model_runner_v1 import NPUModelRunner
+
+MODELS = ["Qwen/Qwen3-8B"]
```
wangxiyuan: use 0.6B to speed up test

tests/e2e/multicard/test_aclgraph_replay_capture.py
```diff
@@ -0,0 +1,195 @@
...
+
+MODELS = ["Qwen/Qwen3-8B"]
+
+
+@pytest.mark.skipif(os.getenv("VLLM_USE_V1") == "0",
+                    reason="aclgraph only support on v1")
+@pytest.mark.parametrize("model", MODELS)
+@pytest.mark.parametrize("max_tokens", [4])
+@patch.dict(os.environ, {"ASCEND_RT_VISIBLE_DEVICES": "0,1"})
+def test_acl_graph_capture_replay_dp2(
```
MengqingCao: Let's use `examples/offline_data_parallel.py` to write this case, you can refer to [dp case](https://github.com/vllm-project/vllm-ascend/blob/main/tests/e2e/multicard/test_data_parallel.py)

tests/e2e/multicard/test_aclgraph_replay_capture.py
```diff
@@ -0,0 +1,195 @@
+#
```
wangxiyuan: make suer this test is in .github workflow  https://github.com/vllm-project/vllm-ascend/blob/main/.github/workflows/_e2e_test.yaml

tests/e2e/multicard/test_aclgraph_replay_capture_mem.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+@pytest.mark.skipif(os.getenv("VLLM_USE_V1") == "0",
```
MengqingCao: we can remove this, as v0 is not maintained anymore

tests/e2e/multicard/test_aclgraph_replay_capture.py
MengqingCao: It seems no changes to patch `torch.npu.NpuGraph` if reuse dp example. cc @wangxiyuan

tests/e2e/multicard/test_aclgraph_replay_capture_mem.py
xuyutao2000: Agreed!

tests/e2e/multicard/test_aclgraph_replay_capture.py
xuyutao2000: Thanks! I’ll make this change

tests/e2e/multicard/test_aclgraph_replay_capture_mem.py
xuyutao2000: Thanks! I’ll make this change

tests/e2e/multicard/test_aclgraph_replay_capture.py
xuyutao2000: @wangxiyuan  Here’s the latest status .I’d appreciate your suggestions, thanks very much!
1、 As for the case for tracking the number of capture and replay operations, I’ve discussed this with @MengqingCao . She mentioned that the DP example she provided doesn’t support wrapping within it, making it impossible to retrieve the capture and replay counts from the main test process by simply invoking the test case. Therefore, the current change may be the only viable solution.
2、As for capturing memory changes, it appears that the main test process cannot directly capture logs printed by the runner process. At this point, I see two viable options:
Option 1: Keep the current patch logic without further changes.
Option 2: Have the logging output written to a temporary file, and then read the required information from that file. I’ve tested this approach, and it works reliably

tests/e2e/multicard/test_aclgraph_replay_capture.py
```diff
@@ -0,0 +1,142 @@
...
+        os.environ["VLLM_DP_SIZE"] = str(dp_size)
+        os.environ["VLLM_DP_MASTER_IP"] = dp_master_ip
+        os.environ["VLLM_DP_MASTER_PORT"] = str(dp_master_port)
+
+        original_replay = torch.npu.NPUGraph.replay
+
+        def replay_wrapper(self):
+            with replay_counter.get_lock():
+                replay_counter.value += 1
+            return original_replay(self)
```
MengqingCao: Using this to detect whether it is the forward of `execute_model`
```python
import inspect
inspect.stack()
```

MengqingCao: dummy run `num_execute_model + (32 - num_execute_model) * dp_size`

PR #3458, opened 2025-10-14
[wip] Adopt fx graph fusion and define quantization fusion pass

wxsIcey:
Adopt fx graph fusion and define quantization fusion pass
refer to https://github.com/vllm-project/vllm/issues/23612
needs https://github.com/vllm-project/vllm/pull/26804

This PR reference https://github.com/vllm-project/vllm-ascend/pull/2389

N/A

local test passed

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/ops/layernorm.py
+++ b/vllm_ascend/ops/layernorm.py
@@ -32,8 +32,6 @@ class AscendRMSNorm(RMSNorm):
 
         from vllm_ascend.utils import is_310p
         if residual is not None:
-            residual = torch.ops.vllm.maybe_chunk_residual(x, residual)
-            assert x.size(0) == residual.size(0)
             if is_310p():
                 orig_dtype = residual.dtype
                 x = x + residual.to(x.dtype)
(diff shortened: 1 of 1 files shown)


vllm_ascend/compilation/quant_fusion_pass.py
```diff
@@ -0,0 +1,94 @@
...
+    def __init__(self, vllm_config):
+        super().__init__(vllm_config)
+        self.patterns: List[Tuple[Callable, Callable]] = []
+        # Register the AddRMSNormQuant fusion pattern into the graph rewriter pattern list
+        AddRMSNormQuantPattern(vllm_config).register(self.patterns)
+
+    def __call__(self, graph: torch.fx.Graph):
+        self.begin()
+        for pattern, replace in self.patterns:
+            replace_pattern(graph, pattern, replace)
```
whx-sjtu, line 78: Use replace_pattern of subgraph_rewriter is a good idea to perform pattern match, but maybe it's better to utilize `torch._inductor.pattern_matcher` to perform pattern match which is also supported in Ascend environment. It has several advantages over subgraph_rewriter:

1. It can easiliy utilize aot_autograd to perform normalization to fx graph, like decompositions. With graph normalization, the difficulty of writing patterns that can be successfully merged is reduced.
2. It is closer to the graph fusion strategy used by the vllm community, so we can directly reuse some pattern passes that vllm have written.

PR #3459, opened 2025-10-14
[bugfix] fix pipeline parallel for mla & sfa attention backend

linfeng-yuan:
Fix pipeline parallel break for mla & sfa attention backend caused by a magic number in metadata builder. The error report:
`AttributeError: 'PPMissingLayer' object has no attribute 'self_attn'`

No.

This PR was tested with "mp" backend (PP2TP8 on an A3 node) as well as "ray" backend (PP2TP8 on two A2 nodes).

- vLLM version: v0.11.0rc3

### DIFF
--- a/tests/e2e/multicard/test_pipeline_parallel.py
+++ b/tests/e2e/multicard/test_pipeline_parallel.py
@@ -20,6 +20,7 @@ from tests.e2e.conftest import VllmRunner
 
 MODELS = [
     "Qwen/Qwen3-0.6B",
+    "deepseek-ai/DeepSeek-V2-Lite-Chat",
 ]
 
 TENSOR_PARALLELS = [1]


Yikun:
How was this PR tested? Which E2E tests should be added to ensure that similar errors won't occur next time?

whx-sjtu (changes requested):
This is really my previous oversight of not taking pp into account when doing rope optimization, thus introducing bugs. I think we need to add deepseek-v3-pruning to the model list of pipeline parallel e2e test.
https://github.com/vllm-project/vllm-ascend/blob/15b2e5c9959a22e326f132b6964e10f9d3c7e26a/tests/e2e/multicard/test_pipeline_parallel.py#L22

linfeng-yuan:
> How was this PR tested? Which E2E tests should be added to ensure that similar errors won't occur next time?

This PR was tested with "mp" backend (PP2TP8 in an A3 node) as well as "ray" backend (PP2TP8 in two A2 nodes). Additionally, an e2e test case for pruned deepseek model with pp=2 tp=1 has been added. CI is retriggered to run this case.

linfeng-yuan:
> This is really my previous oversight of not taking pp into account when doing rope optimization, thus introducing bugs. I think we need to add deepseek-v3-pruning to the model list of pipeline parallel e2e test.
> 
> https://github.com/vllm-project/vllm-ascend/blob/15b2e5c9959a22e326f132b6964e10f9d3c7e26a/tests/e2e/multicard/test_pipeline_parallel.py#L22

thx. This case has been added and let's wait for multicard CI passed.

linfeng-yuan:
Deepseek model initialization in vLLM requires each pp_rank contains at least one moe layer, so I replaced the pruned deepseek model with a deepseek-v2-lite.

whx-sjtu (approved):

PR #3466, opened 2025-10-15
[BugFix] fix qwenVL quant assertion error

elilzhu:
This PR fixes issues:
1. Solve the problem that multimodal scene cannot do weight prefetching and throw an assertion error exception.
2. Standardize the grid_thw data type of qwen2VL to torch.int32.

None.

- ci & e2e

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/models/qwen2_vl.py
+++ b/vllm_ascend/models/qwen2_vl.py
@@ -314,6 +314,7 @@ class AscendQwen2VisionTransformer(Qwen2VisionTransformer):
         x: torch.Tensor,
         grid_thw: torch.Tensor,
     ) -> torch.Tensor:
+        grid_thw = torch.tensor(grid_thw, dtype=torch.int32)
         # compute cu_seqlens and avoid cumsum to fit operator unpadFA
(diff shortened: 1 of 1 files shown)


vllm_ascend/quantization/w8a8.py
```diff
@@ -99,8 +99,11 @@ def apply(
     ) -> torch.Tensor:
         if x.dtype != torch.int8:
             layer_cls_name = layer.__class__.__name__
-            weight_prefetch_method = get_forward_context(
-            ).weight_prefetch_method
+            try:
+                weight_prefetch_method = get_forward_context(
```
wangxiyuan, line 103: `weight_prefetch_method = getattr(get_forward_context(), "weight_prefetch_method", None)`

wangxiyuan:
please add a e2e for vl quantization case 

tests/e2e/singlecard/test_vlm.py  or tests/e2e/singlecard/test_quantization.py

vllm_ascend/quantization/w8a8.py
elilzhu: Considering that getattr may not be able to capture the assertion exception thrown in the parameter generation process, try is still used.

wangxiyuan (approved):

PR #3469, opened 2025-10-15
[Bugfix] [Attention] Remove performance affecting debug variable

sdenis1995:
There is ori_output variable in the forward function of AscendAttentionBackendImpl, which heavily affects the performance (up to 30%) of the attention. This variable is not used anywhere in vllm and vllm-ascend, and looks like just like a debug variable left unnoticed and unremoved.
Profiling with ori_output variable:
with_ori_output_v2

Profiling without ori_output variable:
without_ori_output_v2

No

e2e
models used: Qwen3-14B, Qwen3-30B-A3B, Qwen3-235B-A22B
vLLM version: v0.11.0rc3
- vLLM version: v0.11.0rc3


wangxiyuan:
this change look is out of data now. Do we still need this change?

PR #3474, opened 2025-10-15
[TEST] Add initial aisbench support and Qwen3 32B acc/perf test

jiangyunfan1:
This PR adds the first aisbench case for nightly test, it lays a foundation for following performance and accuracy tests in nightly test.

No

By running the test

- vLLM version: v0.11.0rc3

### DIFF
--- a/tools/aisbench.py
+++ b/tools/aisbench.py
@@ -1,0 +1,220 @@
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+# Copyright 2023 The vLLM team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# This file is a part of the vllm-ascend project.
+#
+import json
+import os
+import re
+import subprocess
+
+import pandas as pd
+from modelscope import snapshot_download  # type: ignore
+
+DATASET_CONF_DIR = "benchmark/ais_bench/benchmark/configs/datasets"
+REQUEST_CONF_DIR = "benchmark/ais_bench/benchmark/configs/models/vllm_api"
+DATASET_DIR = "benchmark/ais_bench/datasets"
+
+
+class AisbenchRunner:
+    RESULT_MSG = {
+        "performance": "Performance Result files locate in ",
+        "accuracy": "write csv to "
+    }
+    DATASET_RENAME = {
+        "aime2024": "aime",
+        "gsm8k-lite": "gsm8k",
+        "textvqa-lite": "textvqa"
+    }
+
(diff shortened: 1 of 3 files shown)


tests/e2e/nightly/models/test_qwen2_5_vl_7b.py
```diff
@@ -0,0 +1,99 @@
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
```
Yikun: Please remove this file, and add this file with job enable later.

[1] https://github.com/vllm-project/vllm-ascend/blob/main/.github/workflows/vllm_ascend_test_nightly.yaml#L46

tests/e2e/nightly/models/test_qwen3_32b_int8.py
```diff
@@ -0,0 +1,99 @@
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
```
Yikun: Please remove this file, and add this file with job enable later.

[1] https://github.com/vllm-project/vllm-ascend/blob/main/.github/workflows/vllm_ascend_test_nightly.yaml#L46

tools/send_mm_request.py
```diff
@@ -0,0 +1,49 @@
+import base64
```
Yikun: Please remove this file, and add this file with vl job enable later.

tools/aisbench.py
```diff
@@ -0,0 +1,220 @@
...
+            if self.task_type == "performance":
+                self.threshold = aisbench_config.get("threshold", 0.97)
+                self._performance_verify()
+
+    def _init_dataset_conf(self):
+        if self.task_type == "accuracy":
+            dataset_name = os.path.basename(self.dataset_path)
+            dataset_rename = self.DATASET_RENAME.get(dataset_name, "")
+            dst_dir = os.path.join(DATASET_DIR, dataset_rename)
+            command = ["cp", "-r", self.dataset_path, dst_dir]
```
Yikun, line 99: This will cause redundant dataset copy, not sure why we can't use modelscope one directly? @jiangyunfan1 You should address this later

tools/aisbench.py
```diff
@@ -0,0 +1,200 @@
...
+                return
+            if "ERROR" in line:
+                raise RuntimeError(
+                    "Some errors happen to Aisbench task.") from None
+
+    def _get_result_performance(self):
+        result_dir = re.search(r'Performance Result files locate in (.*)', 
+                               self.result_line).group(1)[:-1]
+        result_csv_file = os.path.join(result_dir, "gsm8kdataset.csv")
+        result_json_file = os.path.join(result_dir, "gsm8kdataset.json")
```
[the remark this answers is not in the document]
Yikun: This is validate review, pls resolve this. @jiangyunfan1 You should address this later

Yikun (commented):
Please see comments inline, remove unrelated unready file.

Yikun:
LGTM if @jiangyunfan1 you confirm below two comments should be addressed later:

1. https://github.com/vllm-project/vllm-ascend/pull/3474#discussion_r2442711844
2. https://github.com/vllm-project/vllm-ascend/pull/3474#discussion_r2442710722

tools/aisbench.py
jiangyunfan1: The dataset is downloaded from modelscope first to /root/.cache, and then here is copied to the dir which aisbench can directly find and run. It should be discarded with the container when the task is completed.

PR #3477, opened 2025-10-15
[ModelRunner][Qwen3-Next] Fix attn_group initialization timing

MengqingCao:
Fix attn_group initialization timing so that fix qwen3-next model

- vLLM version: v0.11.0rc3


Jeaniowang:
LGTM

PR #3482, opened 2025-10-15
[Fix] Clears unused slot mappings and fix accuracy issue with MLA models when enabling `FULL_DECODE_ONLY`

yiz-liu:
MLA and GQA use different computation logic: MLA slice batches and only compute on the actually valid tokens. That means outer padding must be handled carefully — the accuracy issue this PR fixes was caused by stale data in `slot_mapping` being reused by subsequent inference steps.

So we zeros out the portion of the slot mapping tensor that is not used by the currently scheduled tokens.

None.

Working on it.
- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -1452,6 +1452,7 @@ class NPUModelRunner(LoRAModelRunnerMixin):
                     slot_mapping[:total_num_scheduled_tokens],
                     non_blocking=True,
                 )
+                self.slot_mapping[total_num_scheduled_tokens:].fill_(0)
 
             # Make AscendCommonAttentionMetadata
             common_attn_metadata = AscendCommonAttentionMetadata(


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1452,6 +1452,7 @@ def _prepare_inputs(
                     slot_mapping[:total_num_scheduled_tokens],
                     non_blocking=True,
                 )
+                self.slot_mapping[total_num_scheduled_tokens:].fill_(0)
```
whx-sjtu, line 1455: I found out that this is actually a known problem on the vllm side. Maybe we can go through full-graph related on the vllm side to see if there are more known problems that we can avoid.
https://github.com/vllm-project/vllm/blob/d2740fafbf2527f4026863854bf1aeaa383c1a8e/vllm/v1/worker/gpu_model_runner.py#L1323

yiz-liu: Will do.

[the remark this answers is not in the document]
yiz-liu: Need to make sure the buffer is fixed, cannot clone.

PR #1916, opened 2025-07-21
add super kernel in decode moe

NNUCJ:

PR #3485, opened 2025-10-15
[feat][torchair] support super kernel feat for quantized dsr1

linfeng-yuan:
Port #1916 and #2157 to master branch to fuse operators in deepseek moe layers, which can reduce scheduling overhead on devices. Note that this feature is valid only when `tp_size = 1`  and `multistream_overlap_shared_expert` is enabled with torchair graph mode. 

Users can enable this feature with `--additional-config '{"torchair_graph_config":{"enabled":true, "enable_super_kernel":true}, "multistream_overlap_shared_expert":true}'`.

E2E deepseek serving with 2P1D disaggregated prefill scenarios.

- vLLM version: v0.11.0rc3

### DIFF
--- a/docs/source/user_guide/configuration/additional_config.md
+++ b/docs/source/user_guide/configuration/additional_config.md
@@ -58,6 +58,7 @@ The details of each config option are as follows:
 | `graph_batch_sizes` | list[int] | `[]` | The batch size for torchair graph cache |
 | `graph_batch_sizes_init` | bool | `False` | Init graph batch size dynamically if `graph_batch_sizes` is empty |
 | `enable_kv_nz`| bool | `False` | Whether to enable kvcache NZ layout. This option only takes effects on models using MLA (e.g., DeepSeek). |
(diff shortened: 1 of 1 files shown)


vllm_ascend/torchair/ops/torchair_fused_moe.py
```diff
@@ -1094,6 +1096,7 @@ def __init__(
         self.multistream_overlap_shared_expert = \
             ascend_config.multistream_overlap_shared_expert and \
             self.torchair_graph_enabled
+        self.enable_super_kernel = ascend_config.torchair_graph_config.enable_super_kernel and self.multistream_overlap_shared_expert and tp_size == 1
```
whx-sjtu: Why is `tp_size == 1` necessary here?

linfeng-yuan: > Why is `tp_size == 1` necessary here?

As mentioned in #2157, `StridedSlice` operators will be introduced in moe layers, which interrupts the fusion of the super kernel.

linfeng-yuan:
Full CI is passed except for a3-test because of resource occupation issues. Once the documentation is updated, this pr can be merged to main.

Issue #3328, opened 2025-10-09
[RFC]: Dynamic Batch


KyrieDrewWang:
### Motivation.

Chunked Prefills is a critical feature that improves inference performance in vllm. This feature splits the requests of prefilling into several chunks and batches them with the decoding requests. The parameter `chunk_size` is utilized to control the size of the grouped batch.  

Although grouped in one batch, the resource requirements of the prefilling and decoding requests are quite the opposite. However, the `chunk_size` in current vllm-ascend is a preset fixed value, and fails to be tunned adaptively based on the resource requirements per inference. This fixed batch leads to failure in maximizing the effective throughput (throughput meets the SLO).  Therefore, we propose a **dynamic batch** feature to adaptively adjust the `chunk_size` (batch size) to improve the inference performance on Ascend NPUs.

### Proposed Change.

1. Propose a cost model regarding the `chunk_size`, `prefill_length`, `decode_number`, and `decode_length`. The cost refers to the latency. This cost model can generate the best  `chunk_size` according to the `SLO_limits` during each inference. So far, **dynamic batch** only supports 910 B3 NPU. The cost model is precomputed on 910 B3 and loaded as a lookup table by `BudgetRefiner` before inference.
2. Add a new parameter `SLO_limits_for_dynamic_batch` in additional config for this feature.
3. Add a new schedule function `schedule_with_dynamic_batch` in AscendScheduler for this feature. This schedule function serves with dynamic batch, decode-first chunked prefill and FCFS strategy.

### Any Other Things.

NOTE: Only support 910 B3 so far.

PR #3490, opened 2025-10-16
[Feat] Dynamic Batch Feature

KyrieDrewWang:
[RFC](https://github.com/vllm-project/vllm-ascend/issues/3328) for more details.
Add dynamic batch feature in chunked prefilling strategy, the token budget can be refined to achieve better effective throughput and TPOT. 

!!! NOTE: only 910B3 is supported till now, we are working on further improvements.
Additional file for lookup table is required.

- vLLM version: v0.11.0rc3

### DIFF
--- a/.github/workflows/vllm_ascend_test_nightly.yaml
+++ b/.github/workflows/vllm_ascend_test_nightly.yaml
@@ -56,6 +56,25 @@ jobs:
       vllm: v0.11.0
       runner: ${{ matrix.os }}
       tests: tests/e2e/nightly/models/test_qwen3_32b.py
+  qwen3-32b-in8-a3:
+    strategy:
+      matrix:
+        os: [ linux-aarch64-a3-4 ]
+    uses: ./.github/workflows/_e2e_nightly.yaml
+    with:
+      vllm: v0.11.0
+      runner: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


whx-sjtu (changes requested):
Please add ut for `BudgetRefiner` and `SchedulerDynamicBatch`. You can refer to the ut of `AscendScheduler` as an example:
https://github.com/vllm-project/vllm-ascend/blob/bf87606932fa66c92a444c2e0187b60be7b576de/tests/ut/core/test_scheduler.py#L177
You can modity this original ut to apply to both `AscendScheduler` and `SchedulerDynamicBatch`, or create a new file for SchedulerDynamicBatch, I recommend the former.

requirements.txt
```diff
@@ -8,11 +8,13 @@ pip
 pybind11
 pyyaml
 scipy
+pandas
```
wangxiyuan, line 11: let's only require it when running dynamic batch. Pands should not be a common requirement.

vllm_ascend/ascend_config.py
```diff
@@ -129,6 +129,8 @@ def __init__(self, vllm_config):
             if self.pd_tp_ratio == 0:
                 raise AssertionError(
                     "Only support P node tp size lagger then D node tp size")
+        self.SLO_limits_for_dynamic_batch = additional_config.get(
```
wangxiyuan, line 132: addintional_config doc should be update as well.

wangxiyuan (approved):

PR #3495, opened 2025-10-16
[Feat] Shared expert dp for deepseek and deepseek_mtp

zhaozx-cn:
shared expert dp for deepseek and deepseek_mtp, could be combined with sp to improve performance.

- vLLM version: v0.11.0rc3

### DIFF
--- a/tests/ut/models/test_deepseek_mtp.py
+++ b/tests/ut/models/test_deepseek_mtp.py
@@ -58,7 +58,7 @@ class TestCustomDeepSeekMultiTokenPredictorLayer(PytestBase):
                             return_value=torch.randn(2, 3, 768))
         mocker.patch("torch.cat", return_value=torch.randn(2, 3, 768))
         mocker.patch("torch.ops.vllm.maybe_all_gather_and_maybe_unpad",
-                     lambda x: x)
+                     lambda x, label: x)
         mtp_layer.mtp_block.return_value = (torch.randn(2, 3, 768),
                                             torch.randn(2, 3, 768))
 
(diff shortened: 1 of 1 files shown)


vllm_ascend/models/layers/mla.py
```diff
@@ -122,8 +122,20 @@ def forward(
             hidden_states: torch.Tensor,
             kv_cache: Optional[torch.Tensor] = None,
             attn_metadata: Optional[AttentionMetadata] = None) -> torch.Tensor:
-        need_gather_q_kv = get_forward_context().sp_enabled
-        output_shape = hidden_states.shape
+        try:
+            forward_context = get_forward_context()
+            sp_enabled = forward_context.sp_enabled
+            if sp_enabled and self.debug_layer_idx < self.layers:
```
weijinqian0: debug_layer_idx should be named as mtp_layer_idx

vllm_ascend/utils.py
```diff
@@ -613,7 +613,8 @@ def enable_sp(vllm_config=None) -> bool:
         or envs_ascend.VLLM_ASCEND_ENABLE_FLASHCOMM1
         # Flash comm 1 should be enabled by env VLLM_ASCEND_ENABLE_FLASHCOMM1
         # We retain the env VLLM_ASCEND_ENABLE_FLASHCOMM here for backward compatibility.
-        or bool(int(os.getenv("VLLM_ASCEND_ENABLE_FLASHCOMM", '0'))))
+        or bool(int(os.getenv("VLLM_ASCEND_ENABLE_FLASHCOMM", '0'))) or
+        get_ascend_config().enable_shared_expert_dp)
```
weijinqian0: it may comfused users. when we set enable_shared_expert_dp, it will lead to set sp=1. But these two settings' meanings are different.

vllm_ascend/models/deepseek_mtp.py
```diff
@@ -88,6 +88,9 @@ def forward(
         spec_step_index: int = 0,
     ) -> torch.Tensor:
         assert inputs_embeds is not None
+        if inputs_embeds.shape[0] != positions.shape[0]:
```
realliujiaxu: delete `if inputs_embeds.shape[0] != positions.shape[0]:`. If this is False, `inputs_embeds = torch.ops.vllm.maybe_all_gather_and_maybe_unpad(...)` will not be captured by dynamo

MengqingCao:
plz update pr message

PR #3497, opened 2025-10-16
[Fix] Refactor dummy attention metadata creation

yiz-liu:
The `force_attention` parameter is designed for flash infer kernel warmup, we don't actually need it on Ascend device (at least for now).And it tends to make things more complicated. So we replace the `force_attention` parameter with `aclgraph_runtime_mode` in the attention metadata creation logic.

This change makes the control flow more explicit by directly using the graph runtime mode to determine how to build attention metadata, rather than relying on an intermediate boolean flag. This simplification removes redundant logic and clarifies the conditions for building attention metadata for full decode graph mode.

None.

DP + `FULL_DECODE_ONLY` + online serving.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2330,7 +2330,7 @@ class NPUModelRunner(LoRAModelRunnerMixin):
         forward_context = get_forward_context()
         assert forward_context is not None
         if forward_context.cudagraph_runtime_mode == CUDAGraphMode.FULL and \
-            not forward_context.capturing:
+            not forward_context.capturing and forward_context.attn_metadata is not None:
             if self.vllm_config.model_config.use_mla:
                 # FIXME: Try using `auto_dispatch_capture=True`
                 update_mla_attn_params(self.update_stream, forward_context,


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2412,12 +2418,14 @@ def _dummy_run(
         if self.is_kv_producer and not self.is_kv_consumer:
             with_prefill = True
 
-        # TODO(cmq): check if with_prefill is reasonable
-        attn_metadata = self._build_attention_metadata(
+        # TODO(Mengqing): Set create_mixed_batch to False since it's only used in FI warmup
+        # and not supported in ASCEND now. We could remove it in the future.
+        attn_metadata = self._build_dummy_attn_metadata(
             False,
```
MengqingCao, line 2424: This should check if it is reasonable that `execute_model` in `with_prefill` state and `dummy_run` in `decode_only` state. Let's do this in next prs

yiz-liu: Sure, will look into it.

yiz-liu:
UT failed because of irrelevant PR #3545 .

PR #3334, opened 2025-10-09
[Feat] Flash comm allgher ep

realliujiaxu:
### What this PR does / why we need it?
Support flash comm v1(Sequence Parallelism) for Allgather EP.

- [x] Deepseek R1(MLA model) eager and aclgraph
- [x] GLM 4.5 MoE (GQA model) eager and aclgraph

### How was this patch tested?

set `export VLLM_ASCEND_ENABLE_FLASHCOMM=1` on A2
I test Deepseek-R1 on GSM 8k:
image

- vLLM version: v0.11.0rc3
- vLLM main: https://github.com/vllm-project/vllm/commit/v0.11.0

PR #3505, opened 2025-10-16
[Bugfix] fix logging and d2h bug for flash comm1

realliujiaxu:

Fix 3 bugs in flash comm1 of Allgather EP(https://github.com/vllm-project/vllm-ascend/pull/3334):
1. call `enable_sp()` with argument `vllm_config` trigger a lot of warning log, this PR caches its return value.
2. `num_tokens_after_padding` should be cpu tensor as it will used as `num_tokens_across_dp_cpu` in `DPMetadata`. It will causes may d2h copy when running model.
3. In PD, model runner will execute `kv_connector_no_forward`，where `num_tokens` is None

No

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/ascend_forward_context.py
+++ b/vllm_ascend/ascend_forward_context.py
@@ -114,7 +114,7 @@ def set_ascend_forward_context(
         # the performance may degrade due to the switching of communication methods.
         if is_moe_model(vllm_config):
             sp_enabled = enable_sp(vllm_config) and \
-                tp_world_size > 1
+                tp_world_size > 1 and num_tokens is not None
         else:
             sp_enabled = enable_sp(vllm_config) and \
                 tp_world_size > 1 and \


vllm_ascend/utils.py
```diff
@@ -605,15 +606,20 @@ def dense_optim_enable() -> bool:
...
+            vllm_config = get_current_vllm_config()
+        _ENABLE_SP = (
+            vllm_config.compilation_config.pass_config.
+            enable_sequence_parallelism
+            or envs_ascend.VLLM_ASCEND_ENABLE_FLASHCOMM1
+            # Flash comm 1 should be enabled by env VLLM_ASCEND_ENABLE_FLASHCOMM1
+            # We retain the env VLLM_ASCEND_ENABLE_FLASHCOMM here for backward compatibility.
+            or bool(int(os.getenv("VLLM_ASCEND_ENABLE_FLASHCOMM", '0'))))
+
+    return _ENABLE_SP
```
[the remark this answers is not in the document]
realliujiaxu: vllm_config should be consistant after server started

vllm_ascend/utils.py
```diff
@@ -605,15 +606,20 @@ def dense_optim_enable() -> bool:
...
-    if vllm_config is None:
-        from vllm.config import get_current_vllm_config
-        vllm_config = get_current_vllm_config()
-    return (
-        vllm_config.compilation_config.pass_config.enable_sequence_parallelism
-        or envs_ascend.VLLM_ASCEND_ENABLE_FLASHCOMM1
-        # Flash comm 1 should be enabled by env VLLM_ASCEND_ENABLE_FLASHCOMM1
-        # We retain the env VLLM_ASCEND_ENABLE_FLASHCOMM here for backward compatibility.
-        or bool(int(os.getenv("VLLM_ASCEND_ENABLE_FLASHCOMM", '0'))))
+    global _ENABLE_SP
```
MengqingCao, line 609: caching `_ENABLE_SP` is a good idea, but it seems will also print a lot of logs when both `_ENABLE_SP` and `vllm_config` are `None`, maybe we chould `assert vllm_config is not None` when `_ENABLE_SP` is `None` to remind developers to pass in this parameter

realliujiaxu: The first call of `enable_sp()` is in `linear_op` when initializing model, which is in `set_current_vllm_config` context. So 
vllm_config can only can only be obtained from `get_current_vllm_config` when `_ENABLE_SP` is ` None`
image

vllm_ascend/ascend_forward_context.py
jianzs, line 121:
```suggestion
        sp_enabled = (enable_sp(vllm_config) and tp_world_size > 1 and
                      num_tokens is not None) and (is_moe_model(vllm_config)
                                                   or num_tokens > 1000)
```

PR #3506, opened 2025-10-16
[BUGFIX] Mtp torchair pd fix

JC-ut0:

In memory of https://github.com/vllm-project/vllm-ascend/pull/2610 and #3449 Fix Mtp torchair pd bug.

In the pd Disaggregation scenario, the first token of the inference after the d node receives the kv follows the eager mode.

Fixes:
Running with MTP torchair graph mode with Prefilling Decoding Disaggregation , if all requests processed by the D node are requests just transmitted from the P node, it will break the torchair graph.

Reason: During PD Disaggregation , the P node only transmits the KV cache and prompt to the D node, not the actual tokens inferred (neither the main model tokens nor the MTP tokens are transmitted). Therefore, the D node will treat this request as one without MTP tokens for inference (seq_len=1).
The community does not have graph mode issues because the community's attention has a seq_len=1 for each batch during the decode phase.
We have issues because the graph mode pads according to processing 2 tokens per request. When there are some seq_len=1 and some seq_len=2, padding is done at the end. If all requests received by the D node are seq_len=1, padding cannot be performed normally according to the attention's fia operator constraints.

Solution:

The kv consumer uses extra torchair graph padding to avoid breaking FIA graph constrains (The one this PR implemented).

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/torchair/torchair_model_runner.py
+++ b/vllm_ascend/torchair/torchair_model_runner.py
@@ -84,7 +84,7 @@ class NPUTorchairModelRunner(NPUModelRunner):
 
     def _may_pad_kv_consumer_num_seq(self):
         # pd disaggregation scenario need redundant_batch_sizes to avoid each batch's seq_len exceed 16 tokens
-        # self.max_num_reqs here is greather than the actual maximum request number
+        # self.max_num_reqs here is greater than the actual maximum request number
         if self.is_kv_consumer:
             FIA_SEQ_LEN_LIMIT = 16
             new_max_num_reqs = self.max_num_reqs + math.ceil(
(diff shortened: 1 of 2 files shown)


linfeng-yuan (commented):
This pr was validated with 2P1D disaggregated prefill scenarios (P:DP2TP8EP16, D: DP32EP32).

jianzs:
@JC-ut0  @yiz-liu  @linfeng-yuan  Isn't there any other way to fix this? Changing the max num seqs has a big impact on performance.

JC-ut0:
> @JC-ut0 @yiz-liu @linfeng-yuan Isn't there any other way to fix this? Changing the max num seqs has a big impact on performance.

You can try to patch the vllm scheduler, to add a padding spec token for the request got from the prefill node.
So that our graph mode constraints are not broken, and all logic is the same as the PD mixed deployment .

jianzs:
> > @JC-ut0 @yiz-liu @linfeng-yuan Isn't there any other way to fix this? Changing the max num seqs has a big impact on performance.
> 
> You can try to patch the vllm scheduler, to add a padding spec token for the request got from the prefill node. So that our graph mode constraints are not broken, and all logic is the same as the PD mixed deployment .

Thx. I just deleted the `_may_pad_kv_consumer_num_seq` function, and the engine runs fine without any issues. So, what exactly triggers the problem this PR is addressing, and what are the symptoms?

PR #3510, opened 2025-10-16
[Perf] Add fused matmul/reduce-scatter kernel for performance optimization.

ZYang6263:
This PR boosts performance by introducing a fused kernel for the matrix matmul and reduce scatter operations. It supports both unquantized (e.g., BFloat16) and W8A8 quantized models.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/ops/register_custom_ops.py
+++ b/vllm_ascend/ops/register_custom_ops.py
@@ -1,3 +1,4 @@
+from typing import Optional
 import torch
 import torch.nn.functional as F
 import torch_npu
--- a/vllm_ascend/ops/linear_op.py
+++ b/vllm_ascend/ops/linear_op.py
@@ -375,13 +375,94 @@ class SequenceRowParallelOp(CustomRowParallelOp):
     def matmul_and_reduce(self, input_parallel: torch.Tensor,
                           bias_: Optional[Parameter]) -> torch.Tensor:
         assert self.quant_method is not None
-        output_parallel = self.quant_method.apply(self.layer,
-                                                  input_parallel,
-                                                  bias=bias_)
-        from vllm_ascend.ops.register_custom_ops import \
-            _maybe_pad_and_reduce_impl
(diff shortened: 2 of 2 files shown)


vllm_ascend/ops/register_custom_ops.py
```diff
@@ -15,6 +17,103 @@
...
+        input_parallel: torch.Tensor) -> torch.Tensor:
+    model_instance = get_forward_context().model_instance
+    layer_idx = int(prefix.split('.')[2])
+    layer_name = prefix.split('.')[-1]
+    if layer_name == "o_proj":
+        layer = model_instance.model.layers[layer_idx].self_attn.o_proj
+    if layer_name == "down_proj":
+        layer = model_instance.model.layers[layer_idx].mlp.down_proj
+    output_size_per_partition = layer.output_size_per_partition
+    pad_size = output_size_per_partition - input_parallel.size(1)
```
rjg-lyh: The scenario here is not fully considered, as there may be cases where `output_size_per_partition` is greater than `input_parallel.size(1)`. I suggest adding an if condition to handle different scenarios, using either `pad` or `slice` operations. Alternatively, you can construct a tensor with the feature dimension equal to `output_size_per_partition` using `torch.empty` to return from the fake function.

ZYang6263: I have done this.

vllm_ascend/ops/register_custom_ops.py
```diff
@@ -16,6 +17,108 @@
...
+    if pad_size > 0:
+        x = F.pad(x, (0, 0, 0, pad_size))
+
+    from vllm.model_executor.layers.linear import UnquantizedLinearMethod
+
+    from vllm_ascend.quantization.w8a8 import (AscendW8A8LinearMethod,
+                                               quant_per_tensor)
+
+    # no quant
+    if isinstance(layer.quant_method, UnquantizedLinearMethod):
```
weijinqian0: Need to think about how to decouple the code.

PR #3511, opened 2025-10-16
[Patch]patch of v1 executor when enable eplb.

offline893:
when using dynamic eplb, patch v1 executor to avoid create child process failed.

deepseek in v3.

- vLLM version: v0.11.0rc3

### DIFF
--- a/docs/source/user_guide/feature_guide/eplb_swift_balancer.md
+++ b/docs/source/user_guide/feature_guide/eplb_swift_balancer.md
@@ -16,7 +16,7 @@ Expert balancing for MoE models in LLM serving is essential for optimal performa
 
 ### Dynamic EPLB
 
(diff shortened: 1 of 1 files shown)


vllm_ascend/patch/platform/patch_common/__init__.py
```diff
@@ -14,7 +14,27 @@
...
+
+
+def patch_v1_executor():
+    try:
+        dynamic_eplb = os.getenv("DYNAMIC_EPLB", False) or os.getenv(
+            "EXPERT_MAP_RECORD", False)
+        if dynamic_eplb:
+            import vllm_ascend.patch.platform.patch_common.patch_multiproc_executor  # noqa
+        else:
+            logger.warning("Do not patch v1 executor.")
```
zzzzwwjj, line 33: I think it's only necessary to print warning logs when patching.

PR #3515, opened 2025-10-17
[Fix] Ensure num_tokens is not None when enabling sp in set_ascend_forward_context

jianzs:

When running DeepSeek inference, after enabling SP, the engine crashed during the execution of the kv connector's "no_forward" logic. The crash occurred because "num_tokens" was set to None during "no_forward," leading to an error when attempting to set the forward context.

No

CI pass.

- vLLM version: v0.11.0rc3


jianzs:
Duplicate with 3505, closing.

Issue #3513, opened 2025-10-17
[Bug]: vllm:EngineCore process coredump while testing TextVQA dataset for both Qwen3-VL-30B-A3B-Instruct and Qwen2.5-VL-7B-Instruct


leijie2015:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>
[root@4257864fdaf1 workspace]# python collect_env.py
Collecting environment information...
PyTorch version: 2.7.1+cpu
Is debug build: False

OS: openEuler 24.03 (LTS-SP2) (aarch64)
GCC version: (GCC) 12.3.1 (openEuler 12.3.1-98.oe2403sp2)
Clang version: Could not collect
CMake version: version 4.1.0
Libc version: glibc-2.38

Python version: 3.11.13 (main, Aug 29 2025, 04:31:33) [GCC 12.3.1 (openEuler 12.3.1-97.oe2403sp2)] (64-bit runtime)
Python platform: Linux-5.10.0-153.56.0.134.oe2203sp2.aarch64-aarch64-with-glibc2.38

CPU:
Architecture:                       aarch64
CPU op-mode(s):                     64-bit
Byte Order:                         Little Endian
CPU(s):                             192
On-line CPU(s) list:                0-191
Vendor ID:                          HiSilicon
BIOS Vendor ID:                     HiSilicon
Model name:                         Kunpeng-920
BIOS Model name:                    HUAWEI Kunpeng 920 5250 To be filled by O.E.M. CPU @ 2.6GHz
BIOS CPU family:                    280
Model:                              0
Thread(s) per core:                 1
Core(s) per socket:                 48
Socket(s):                          4
Stepping:                           0x1
Frequency boost:                    disabled
CPU(s) scaling MHz:                 100%
CPU max MHz:                        2600.0000
CPU min MHz:                        200.0000
BogoMIPS:                           200.00
Flags:                              fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma dcpop asimddp asimdfhm ssbs
L1d cache:                          12 MiB (192 instances)
L1i cache:                          12 MiB (192 instances)
L2 cache:                           96 MiB (192 instances)
L3 cache:                           192 MiB (8 instances)
NUMA node(s):                       8
NUMA node0 CPU(s):                  0-23
NUMA node1 CPU(s):                  24-47
NUMA node2 CPU(s):                  48-71
[...]

PR #3519, opened 2025-10-17
[Doc] Add --shm-size option to Docker command for qwen3 vl 235B

leijie-ww:
Added shared memory size option to Docker run command.If shm-size is not specified, docker will use 64MB by default. In this case, vllm:EngineCore process may coredump if workload is high.
No
Done

Closes: https://github.com/vllm-project/vllm-ascend/issues/3513

- vLLM version: v0.11.0rc3

### DIFF
--- a/docs/source/tutorials/multi_node_qwen3vl.md
+++ b/docs/source/tutorials/multi_node_qwen3vl.md
@@ -37,7 +37,7 @@ docker run --rm \
 --device /dev/davinci_manager \
 --device /dev/devmm_svm \
 --device /dev/hisi_hdc \
---shm-size=1000g # Adjust based on workload.  Consider using a smaller value if memory is constrained.
+--shm-size 256g \
 -v /usr/local/dcmi:/usr/local/dcmi \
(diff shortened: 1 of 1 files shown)


docs/source/tutorials/multi_node_qwen3vl.md
```diff
@@ -37,6 +37,7 @@ docker run --rm \
 --device /dev/davinci_manager \
 --device /dev/devmm_svm \
 --device /dev/hisi_hdc \
+--shm-size=1000g # Adjust based on workload.  Consider using a smaller value if memory is constrained.
```
Yikun: 1. This cmd can't be excute after this fix.
```suggestion
--shm-size=4g \
```

2. What's the recommanded value of `shm-size` and setting rule? just saw [1] has many different examples.

[1] https://docs.vllm.ai/en/latest/cli/run-batch.html?q=shm-size

leijie-ww: We assume this configuration depends on workloads, in `Qwen3-VL-235B-A22B-Instruct and multiple nodes scenario, suggest set shm-size start from 256g, with 8g or 16g incremental step

Yikun (approved):

PR #3533, opened 2025-10-18
fix vl float model not support NZ format weight error

shaopeng-666:
fix vl float model not support nz mm op

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/models/qwen2_5_vl.py
+++ b/vllm_ascend/models/qwen2_5_vl.py
@@ -42,7 +42,7 @@ from vllm.model_executor.models.qwen2_5_vl import (
 from vllm.model_executor.models.utils import maybe_prefix
 from vllm.multimodal import MULTIMODAL_REGISTRY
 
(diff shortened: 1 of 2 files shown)


vllm_ascend/models/qwen2_5_vl.py
```diff
@@ -281,15 +283,24 @@ def pad_qkv_weight(self, data):
             [qkv_weight_first_half_padded, qkv_weight_second_half_padded],
             dim=2)
         qkv_weight_final = qkv_weight_padded.reshape(-1, self.hidden_size)
-        return qkv_weight_final
+
+        qkv_weight_final_copy = torch.empty_like(qkv_weight_final).copy_(
```
wangxiyuan: add `if enable_nz` check?

shaopeng-666: I have added this judge, the bug can be truely solved when the performance of the FA operator is not affected when dim != 128.

PR #3535, opened 2025-10-19
bugfix for mooncake

zouyida2052:
bugfix for mooncake, remove useless judgement.

no

by ci

- vLLM version: v0.11.0rc3


MengqingCao (approved):
LGTM, checking if is deepseek mla is enough for both deepseek-v3/r1/v3.2

PR #3559, opened 2025-10-20
[Test]Add accuracy test for model Qwen2.5-Omni-7B

MrZ20:
Using non-blocking operations for device-to-host transfers can lead to data corruption in later steps. The CPU tensor is accessed right after the transfer is triggered, but the transfer might not be complete yet. As a result, the data could be wrong. This problem was seen in the A3 environment during `profile_run`.

CI pass.

Add accuracy test for model Qwen2.5-Omni-7B

- vLLM version: v0.11.0rc3


MrZ20:
@MengqingCao Review required

PR #3566, opened 2025-10-20
[BugFix]Fix mtp torchair bug caused by #2719

JC-ut0:
Fix mtp tochair bug cuased by #2719
Since FIA need extra space for padding, we need to enforce `self.max_num_seqs > self.scheduler_config.max_num_seqs` in KV consumer + MTP
This means that, `self.max_num_seqs` **>** the actual maximum requests  (`self.scheduler_config.max_num_seqs`)

- vLLM version: v0.11.0rc3


realliujiaxu:
add tag [BugFix]

zzzzwwjj (approved):

jianzs:
Is there another way to solve this issue? The approach used in this pull request significantly impacts performance and could confuse users.

If the FIA requires additional space, we should reserve that space during the `determine_available_memory` process.

PR #3570, opened 2025-10-21
[CI] Fix test_mla_v1

Potabk:
Remove test cases containing CPU incompatible operators

- vLLM version: v0.11.0rc3

### DIFF
--- a/tests/ut/attention/test_mla_v1.py
+++ b/tests/ut/attention/test_mla_v1.py
@@ -343,21 +343,6 @@ class TestAscendMLAImpl(TestBase):
         self.assertEqual(self.impl.num_queries_per_kv, 32)
         self.assertEqual(self.impl.tp_size, 2)
 
-    def test_v_up_proj(self):
-        batch_size = 4
-        x = torch.randn(batch_size, self.impl.num_heads,
-                        self.impl.kv_lora_rank)
-
-        if not hasattr(self.impl, 'W_UV') or self.impl.W_UV is None:
-            self.impl.W_UV = torch.randn(self.impl.num_heads,
-                                         self.impl.kv_lora_rank,
-                                         self.impl.v_head_dim)
-        result = self.impl._v_up_proj(x)
-
-        self.assertEqual(result.shape[0], batch_size)
-        self.assertEqual(result.shape[1],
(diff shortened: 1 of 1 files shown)


vllm_ascend/attention/mla_v1.py
```diff
@@ -559,14 +559,19 @@ def __init__(
...
-        x = torch_npu.npu_transpose_batchmatmul(x,
-                                                self.W_UV,
-                                                perm_x1=[1, 0, 2],
-                                                perm_x2=[0, 1, 2],
-                                                perm_y=[1, 0, 2])
-        x = x.reshape(-1, self.num_heads * self.v_head_dim)
-        return x
+        if x.device.type == "npu":
+            x = x.view(-1, self.num_heads, self.kv_lora_rank)
+            x = torch_npu.npu_transpose_batchmatmul(x,
```
MengqingCao: We could mock this in ut instead of adding cpu branch in this operator

wangxiyuan:
it should be fixed in ut code

Potabk:
the test case:
```python
    def test_v_up_proj(self):
        batch_size = 4
        x = torch.randn(batch_size, self.impl.num_heads,
                        self.impl.kv_lora_rank)

        if not hasattr(self.impl, 'W_UV') or self.impl.W_UV is None:
            self.impl.W_UV = torch.randn(self.impl.num_heads,
                                         self.impl.kv_lora_rank,
                                         self.impl.v_head_dim)
        result = self.impl._v_up_proj(x)

        self.assertEqual(result.shape[0], batch_size)
        self.assertEqual(result.shape[1],
                         self.impl.num_heads * self.impl.v_head_dim)
```
the origin case:

```python
    def _v_up_proj(self, x):
        x = x.view(-1, self.num_heads, self.kv_lora_rank)
        x = torch_npu.npu_transpose_batchmatmul(x,
                                                self.W_UV,
                                                perm_x1=[1, 0, 2],
                                                perm_x2=[0, 1, 2],
                                                perm_y=[1, 0, 2])
        x = x.reshape(-1, self.num_heads * self.v_head_dim)
        return x
```
if mock the op `torch_npu.npu_transpose_batchmatmul`, I think the test is meanless, so I suggest remove it

PR #3574, opened 2025-10-21
[Chore] Prevents use of ASCEND_LAUNCH_BLOCKING with ACL Graph

yiz-liu:
Adds a validation check to prevent running with an incompatible configuration.

The `ASCEND_LAUNCH_BLOCKING=1` environment variable, used for debugging, enforces synchronous execution which is incompatible with ACL Graph.

This change raises an explicit error to inform the user about the conflict and how to resolve it, preventing a more obscure failure later.

None.

None needed.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -270,11 +270,16 @@ class NPUPlatform(Platform):
             compilation_config.cudagraph_mode = CUDAGraphMode.NONE
             compilation_config.level = CompilationLevel.NO_COMPILATION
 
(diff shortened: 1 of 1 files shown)


wangxiyuan (approved):
Make sense

yiz-liu:
All tests have passed, adding new error message.

wangxiyuan:
https://github.com/vllm-project/vllm-ascend/actions/runs/18675581780/job/53244990091?pr=3574

nalinaly:
Torch_npu would add a feature to handle compatibility between aclgraph and non_blocking. The basic consideration is avoiding synchronization during capture.

PR #3582, opened 2025-10-21
[MoE][Multistream] Avoid performing communication in extra stream.

whx-sjtu:
This PR moves the communication operation of shared experts out of extra stream because I found that this might cause rtMemcpy related errors when running shared experts multistream with aclgraph.

Furthermore, I utilize a global variable as extra stream object to avoid allocating streams for each layer in full-graph mode.

- vLLM version: v0.11.0rc3


yiz-liu (approved):
Nit: maybe rename `calculation` to `computation` later.

PR #3594, opened 2025-10-21
[BugFix] Fix kv_no_split not contiguous

zhaozx-cn:
allgather need contiguous data, split operation return uncontiguous data.

- vLLM version: v0.11.0rc3


vllm_ascend/attention/mla_v1.py
```diff
@@ -1156,6 +1156,8 @@ def _mla_preprocess(self, layer_name, hidden_states, kv_cache,
                 dim=-1,
             )
             q_c = self.q_a_layernorm(q_c)
+            # allgather need contiguous data
+            kv_no_split = kv_no_split.contiguous()
```
whx-sjtu, line 1255: `qkv_lora.split` is applied on the last dim, which might also cause `q_c` to be discontinuous. Thus I think `q_c` also needs `contiguous()` here.

whx-sjtu (changes requested):

vllm_ascend/attention/mla_v1.py
zhaozx-cn: q_c is contiguous after layernorm,

whx-sjtu: LGTM.

jianzs (approved):
LGTM

PR #3597, opened 2025-10-21
[Bugfix][MTP] Fix performance degradation when mtp>1

slippersss:
This PR aims to fix performance degradation when mtp>1. Since mtp>1 may result in more tokens (i.e. larger batch size) than acl graph maximum batch size, this will cause draft model to run in eager mode.

N/A

by ci

- vLLM version: v0.11.0rc3

### DIFF
--- a/tests/ut/test_utils.py
+++ b/tests/ut/test_utils.py
@@ -272,6 +272,7 @@ class TestUtils(TestBase):
             len(test_vllm_config.compilation_config.cudagraph_capture_sizes))
 
         test_vllm_config.speculative_config = mock.MagicMock()
+        test_vllm_config.speculative_config.num_speculative_tokens = 2
         test_vllm_config.speculative_config.draft_model_config = mock.MagicMock(
         )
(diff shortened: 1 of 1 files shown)


vllm_ascend/utils.py
```diff
@@ -410,6 +410,26 @@ def update_aclgraph_sizes(vllm_config: VllmConfig) -> None:
...
+        if original_sizes[0] < (num_speculative_tokens + 1) * max_num_seqs:
+            enlarged_sizes = [(num_speculative_tokens + 1) * size
+                              for size in original_sizes]
+            compilation_config.init_with_cudagraph_sizes(enlarged_sizes)
+            logger.info(
+                "Adjusted ACL graphs: %s → %s for speculative decoding",
+                original_sizes, enlarged_sizes)
+        else:
+            compilation_config.cudagraph_capture_sizes = original_sizes
+
```
yiz-liu, line 432: The best approach is to open an issue in vLLM and ask why it's hardcoded to `max_num_seqs * 2`

wangxiyuan (approved):

PR #3332, opened 2025-10-09
[Bugfix]fix_mulit_connector_bug

baxingpiaochong:
### What this PR does / why we need it?
When using multi connector, the multi connector does not define get_finished_count, which will cause the kv cache to be released
### How was this patch tested?

- vLLM version: v0.11.0rc3
- vLLM main: https://github.com/vllm-project/vllm/commit/83f478bb19489b41e9d208b47b4bb5a95ac171ac

PR #3607, opened 2025-10-22
[P/D] [Bugfix] fix delay free prefill req & D node support prefix cache

liziyu179:
Fix mooncake connector. In scenarios where TP is not equal, when the prefill TP size is less than the number of key-value heads, _get_remote_tp_ranks_for_req will return a list of np.arrays. Performing an operation like int in list of np.arrays will cause an error. Converting the list of np.arrays into a single np.array resolves this issue.

qwen235B
P tp16, D tp1
P tp8, D tp1
P tp4, D tp1
P tp8, D tp2

- vLLM version: v0.11.0rc3


lidenghui1110:
@liziyu179 this bug has been fixed in #3332 , before the PR merged, you can refer to the fix by flatten list[list[int]] into list[int] to fix it.

PR #3616, opened 2025-10-22
[main][refactor] refactor SequenceRowParallelOp forward

rjg-lyh:
This PR refactors SequenceRowParallelOp forward. In order to further expand the operator inclusion scope in dynamic judgment scenarios, this PR customizes the entire matmul computation and communication as a custom operator masking. With this refactor, it will support directly writing code such as common operation fusion into the `SequenceRowParallelOp` class's member function `matmul_and_reduce`, without the need to register more redundant custom masking operators.

No.

Perf comparison:
before:
image

after:
image

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/ops/linear.py
+++ b/vllm_ascend/ops/linear.py
@@ -26,6 +26,7 @@ import torch
 import torch.nn as nn
 import torch_npu
 from torch.nn.parameter import Parameter
+from vllm.config import get_current_vllm_config
 from vllm.distributed import divide
 from vllm.model_executor.layers.linear import (  # noqa
     WEIGHT_LOADER_V2_SUPPORTED, ColumnParallelLinear, LinearBase,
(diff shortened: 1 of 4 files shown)


vllm_ascend/ops/linear.py
```diff
@@ -234,6 +235,13 @@ def __init__(
         return_bias: bool = True,
         disable_tp: bool = False,
     ):
+        compilation_config = get_current_vllm_config().compilation_config
+        # TODO(shaopeng-666): Remove the visual check after the mm model reconstruction is complete.
+        if prefix in compilation_config.static_forward_context and \
+            "visual" not in prefix:
```
MengqingCao, line 241: Adding the check on visual is mainly due to the replicated initialization of self_attn layer in custom modeling of qwen vl. And currently sp is not supported in VLM sceanrio, thus this is okay for me

rjg-lyh: Yes，Just sp supported in mm sceanrio in the future, this check would not Impact basic functionality. But better to remove this check, like the description in todo.

PR #3530, opened 2025-10-18
remove redundant params in mla_preprocess kernel

kiscad:
### What this PR does / why we need it?

This pull request removes the redundant parameters `gamma1` and `beta1` (also named `gamma0`/`beta0` in some places) from the `mla_preprocess` kernel and its calling hierarchy. The changes are consistent across C++ kernel code, bindings, and Python call sites. The parameters were unused in the lower-level functions, so their removal is a good cleanup.

### Does this PR introduce _any_ user-facing change?

The python interface of the kernel is affected, and the params of `gamma0` and `beta0` are not needed.

### How was this patch tested?

The unit-test of the kernel is adapted accordingly.

- vLLM version: v0.11.0rc3
- vLLM main: https://github.com/vllm-project/vllm/commit/v0.11.0

PR #3618, opened 2025-10-22
[Fix] Fixes attribute error in MLA implementation

yiz-liu:
Corrects the attribute access for retrieving the device from `q_a_proj` to `q_proj`. This prevents an `AttributeError` as `q_a_proj` does not exist on the class instance.

None.

Need MLAPO tests.

- vLLM version: v0.11.0rc3


yiz-liu:
@kiscad Because the model weights are bfloat16, the test cases in #3530 are invalid and won’t trigger the MLAPO operator. Per the reviewers’ decision, I’ve removed those two cases, please submit a new PR with test cases that actually exercise the operator.

PR #3632, opened 2025-10-22
[v0.11.0][Perf] Eliminating the zerolike operator through patch

ZYang6263:
There is a zero-like operator before the attention operation in each decoding stage. After analysis, this operator can be eliminated. The purpose of this PR is to remove this operator and improve performance.

### DIFF
--- a/vllm_ascend/patch/worker/__init__.py
+++ b/vllm_ascend/patch/worker/__init__.py
@@ -27,4 +27,5 @@ import vllm_ascend.patch.worker.patch_roberta  # noqa
 import vllm_ascend.patch.worker.patch_weight_loader  # noqa
 import vllm_ascend.patch.worker.patch_multimodal_merge  # noqa
 import vllm_ascend.patch.worker.patch_minicpm  # noqa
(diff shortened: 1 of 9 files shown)


vllm_ascend/patch/worker/__init__.py
```diff
@@ -27,4 +27,5 @@
 import vllm_ascend.patch.worker.patch_weight_loader  # noqa
 import vllm_ascend.patch.worker.patch_multimodal_merge  # noqa
 import vllm_ascend.patch.worker.patch_minicpm  # noqa
+import vllm_ascend.patch.worker.patch_attention_layer  # noqa
```
whx-sjtu: To add a new patch you need to add the detailed description of this patch in the following file:
https://github.com/vllm-project/vllm-ascend/blob/3f5bbde4360ac4d53e75dbcab45d12bf67bd71d0/vllm_ascend/patch/__init__.py#L26

wangxiyuan (approved):

whx-sjtu (approved):

wangxiyuan:
This is a performance patch picked from vLLM main, so it's only needed for 0.11.0-dev branch

PR #3641, opened 2025-10-22
[BugFix][P/D] Modify the recalculation logic to prevent waiting requests from filling up the D node KVCache 

underfituu:
Modify the recalculation logic to prevent waiting requests from filling up the D node KVCache 
None

- vLLM version: v0.11.0rc3
- vLLM main: https://github.com/vllm-project/vllm/commit/17c540a993af88204ad1b78345c8a865cf58ce44

### DIFF
--- a/vllm_ascend/core/recompute_scheduler.py
+++ b/vllm_ascend/core/recompute_scheduler.py
@@ -362,7 +362,7 @@ class RecomputeScheduler(SchedulerInterface):
         skipped_waiting_requests = create_request_queue(self.policy)
 
         # Next, schedule the WAITING requests.
-        if not preempted_reqs:
+        if not preempted_reqs and not recomputed_reqs:
             while self.waiting and token_budget > 0:
                 if len(self.running) == self.max_num_running_reqs:
                     break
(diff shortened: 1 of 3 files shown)


vllm_ascend/core/recompute_scheduler.py
```diff
@@ -361,7 +361,7 @@ def schedule(self) -> RecomputeSchedulerOutput:
         skipped_waiting_requests = create_request_queue(self.policy)
 
         # Next, schedule the WAITING requests.
-        if not preempted_reqs:
+        if not preempted_reqs or not recomputed_reqs:
```
whx-sjtu: The `preempted_reqs` is used for producer and `recomputed_reqs` is used for consumer. LGTM.

underfituu: I found that in the hybrid deployment scenario, `recomputed_reqs` is always empty, and the P node has no recalculation logic, so `recomputed_reqs` is also always empty. Therefore, the logic here `or` needs to be changed to `and`;

PR #3645, opened 2025-10-23
[Doc] Upgrade docker run command

Potabk:
Update the docker run command, specifically: add --shm-size=1g
users/developers using docker to pull vllm-ascend, the shared memory of the container  will be increased from the default 64MB to 1G

- vLLM version: v0.11.0rc3


jiangyunfan1:
LGTM

PR #3648, opened 2025-10-23
【main】patch sched_yield

fems14:
On Arm systems, os.sched_yield() does not take effect, causing the GIL (Global Interpreter Lock) to remain unrelinquished and resulting in CPU bound issues. This PR applies a patch to sched_yield in vLLM, making the process execute time.sleep(0) instead to release the GIL.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/patch/platform/patch_sched_yield.py
+++ b/vllm_ascend/patch/platform/patch_sched_yield.py
@@ -1,4 +1,5 @@
 import sys
+
 import vllm.distributed.utils
 from vllm.platforms import CpuArchEnum, Platform
 


jianzs:
@fems14  Explain the reasoning behind your code changes in both comments and commit messages.

vllm_ascend/patch/platform/patch_sched_yield.py
```diff
@@ -0,0 +1,3 @@
+import vllm.distributed.utils
+
+vllm.distributed.utils.USE_SCHED_YIELD = False
```
wangxiyuan: add os.arch check? I assume  this patch should only work on aarch64?

wangxiyuan (approved):

PR #3656, opened 2025-10-23
Optimizing PD-Proxy for First Token Early Return (Better & More Stable TTFT)

h225yang:
This PR is solving this [issue](https://github.com/vllm-project/vllm-ascend/issues/3099)

Under PD disaggreation, the code logic for PD-Proxy is as follows: After the P (Prefill) node completes inference, it needs to wait for the D (Decoding) node to complete one iteration, and then the D node returns the first token. Compared to the PD hybrid deployment scenario, this mechanism inherently leads to a higher Time To First Token (TTFT).

Test on 2P1D setup (64 920B2C, 16 npus for one P, 32 npus for one D), 0.9.1-dev image.

Before:
aisbench_after

After:

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ By using vLLM Ascend plugin, popular open-source models, including Transformer-l
 - Software:
   * Python >= 3.9, < 3.12
   * CANN >= 8.2.RC1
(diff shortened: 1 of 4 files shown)


wangxiyuan:
how about push this PR to main branch?

wangxiyuan:

PR #3660, opened 2025-10-23
[Misc] Limit ray version

wangxiyuan:
We notice that with ray>2.48.0, the npu card count is not correct from ray. This is a known bug. Let's limit ray version to <2.48.0 now.

- vLLM version: v0.11.0rc3


Yikun:
Any upstream fix in ray? I noticed there is a security issue on [1].

[1] https://github.com/vllm-project/vllm-ascend/security/dependabot/2

wangxiyuan:
@Yikun this issue is found by some users that the NPU carc number is not correctly when ray start. currently, there is related issue in Ray project from my sight.

PR #3672, opened 2025-10-23
[Feature] Remove stream synchronization during ring_mla

jianzs:

This optimization existed before but appears to have been removed at some point. This pull request restores it to remove the stream synchronization overhead caused by the MLA chunk context. As a result, the DS R1 Prefill performance increased from 4.15 to 4.20.

No.

CI pass.

- vLLM version: v0.11.0

### DIFF
--- a/tests/ut/attention/test_mla_v1.py
+++ b/tests/ut/attention/test_mla_v1.py
@@ -69,7 +69,8 @@ class TestAscendMLAPrefillMetadata(TestBase):
         self.assertIsNone(metadata.chunked_context)
 
     @patch("torch_npu.npu._lazy_init")
-    def test_ascend_mla_prefill_metadata_with_chunked_context(self):
+    def test_ascend_mla_prefill_metadata_with_chunked_context(
+            self, mock_lazy_init):
         cu_seq_lens = torch.tensor([0, 2, 4])
         starts = torch.tensor([0, 2])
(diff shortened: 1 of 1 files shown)


vllm_ascend/attention/mla_v1.py
```diff
@@ -370,6 +371,7 @@ def build(
                     seq_tot=chunk_seq_lens.sum(dim=1).tolist(),
                     max_seq_lens=chunk_seq_lens.max(dim=1).values.tolist(),
                     chunk_seq_lens=chunk_seq_lens,
+                    chunk_seq_lens_npu=chunk_seq_lens.to(device, non_blocking=True),
```
whx-sjtu: `chunk_seq_lens` is not defined to use `pin_memory`, will this `non_blocking` take effect?

vllm_ascend/attention/mla_v1.py
```diff
@@ -790,7 +794,7 @@ def _compute_prefill_context(
                 cache_kv_c,
                 cache_k_pe,
                 prefill_metadata.block_table,
-                seq_len2.to(q_nope.device),
+                seq_len2_npu,
```
whx-sjtu, line 920: Ops, removing this optimization might be a bug introduced by MLA refactoring. Thanks for the fix. LGTM.

PR #3693, opened 2025-10-24
[Main][Perf] Add fused matmul/reduce-scatter kernel for performance optimization.

ZYang6263:
This PR boosts performance by introducing a fused kernel for the matrix matmul and reduce scatter operations. It supports both unquantized (e.g., BFloat16) and W8A8 quantized models.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/ops/linear_op.py
+++ b/vllm_ascend/ops/linear_op.py
@@ -39,11 +39,15 @@ from typing import Optional, Union
 
 import torch
 import torch.distributed as dist
+import torch.nn.functional as F
 import torch_npu
 from torch.distributed import ProcessGroup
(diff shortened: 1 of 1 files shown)


vllm_ascend/ops/linear_op.py
```diff
@@ -375,12 +379,79 @@ def apply_impl(
...
+                self.layer.weight,
+                hcom_name,
+                world_size,
+                reduce_op="sum",
+                bias=None,
+                comm_turn=0,
+                x2_scale=deq_scale,
+                output_dtype=output_dtype,
+                comm_mode=comm_mode)
+            output.add(
```
rjg-lyh: add_

PR #3703, opened 2025-10-24
[Bugfix] The server fails to locate the request, leading to the server hanging.

luoxiaolin712:
fix bug: In the mooncake pooling scenario, when the client closes the request, the server fails to locate the request, leading to the server hanging.oling scenario, when the client closes the request, the server fails to locate the request, leading to the server hanging.

Pull up the PD separated pooling service, send requests using aisbench, press CTRL+C twice, and check if the vllm_ascend service exit.

- vLLM version: v0.11.0rc3
- vLLM main: https://github.com/vllm-project/vllm/commit/17c540a993af88204ad1b78345c8a865cf58ce44


whx-sjtu (commented):
This PR solves the bug of running async-scheduling with kv cache pooling. I have argued with relevant responsible person

wangxiyuan (approved):

PR #3708, opened 2025-10-24
[Feat] Add mrope fusion op

shaopeng-666:
Add mrope fusion op for qwen2.5-vl. This mrope operator dosen't support Qwen3-VL currently. Thus could only take affect in qwen2.5-vl

- vLLM version: v0.11.0rc3


yiz-liu (approved):
Please see #3680 .

MengqingCao (approved):

PR #3715, opened 2025-10-24
[bugfix] mlapo adaptation

kiscad:

This pull request fixes a potential bug in the `mlapo` adaptation within the `_mla_decode_preprocess` method. The change ensures that tensors for query projection are allocated with the correct size by using `bsz` (the number of decode tokens) directly, instead of relying on `hidden_states.shape[0]`. This improves the code's robustness and prevents potential runtime errors that could arise from using a padded tensor's shape. The fix is correct and enhances the reliability of the attention mechanism.

No.

- vLLM version: v0.11.0rc3


kiscad:
Hi, @yiz-liu 
This PR is related with aclgraph in specific scenario. Please take a look when you have time.
Let me know if any clarification is needed. Thanks!

PR #3768, opened 2025-10-25
[P/D] force with_prefill true after allreduce in kv producer

liziyu179:
force with_prefill true after allreduce in kv producer

- vLLM version: v0.11.0rc3


yiz-liu (approved):
We should remove this `with_prefill = True` once `attn_state` is refactored or removed.

linfeng-yuan:
LGTM. It doesn't affect torchair scenario since we always expect `with_prefill` to be `True` for P node.

PR #3770, opened 2025-10-25
add mtp doc

zouyida2052:
add mtp develop doc

no

- vLLM version: v0.11.0


docs/source/developer_guide/feature_guide/Multi_Token_Prediction.md
```diff
@@ -0,0 +1,112 @@
...
+├── sample
+│   ├── rejection_sample.py
+├── spec_decode
+│   ├── mtp_proposer.py
+└───────────
+```
+
+**1. sample**
+
+- *rejection_sample.py*: During decoding, the main model processes the previous round’s output token and the predicted token together (computing 1+k tokens simultaneously). The first token is always correct, while the second token—referred to as the **bonus token**—is uncertain since it is derived from speculative prediction, thus We employ **Greedy Strategy** and **Rejection Sampling Strategy** to determine whether the bonus token should be accepted. The module structure consists of an `AscendRejectionSampler` class with a forward method that implements the specific sampling logic.
```
JC-ut0: “**bonus token**—is” Typo？

PR #3771, opened 2025-10-25
[v0.11.0][P/D]Set adxl as default backend and update readme

nwpu-zxr:
Set adxl engine as the default Mooncake backend, because Ascend Transport is no longer maintained.
Update README to include instructions for installing the adxl backend Mooncake.

Users need to compile and install the mooncake backend for adxl according to the revised README instructions.

By CI.

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -89,6 +89,7 @@ jobs:
           # the test separately.
 
           pytest -sv tests/e2e/singlecard/test_aclgraph.py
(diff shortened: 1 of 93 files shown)


liziyu179:
> It may be necessary to inform users which Mooncake version onward uses ADXL as the default, since the ADXL transport was merged into Mooncake after the Ascend transport.

Community has updated the version dependency documentation, requiring CANN ≥ 8.3.RC1. The installation documentation for Mooncake should be updated accordingly along with the community documentation.
https://vllm-ascend.readthedocs.io/en/latest/installation.html

Issue #3342, opened 2025-10-09
[Bug]: Custom kernels fail to compile without raising any errors


yiz-liu:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

None.

</details>

### 🐛 Describe the bug

# Description

This is a long-standing problem: in some scenarios compilation fails but the installation still succeeds. This causes several hard-to-explain issues:

1. Graph mode becomes unusable — without `weak_ref_tensor` we can't free the intermediate tensors created during graph capture.
```
AttributeError: '_OpNamespace' '_C_ascend' object has no attribute 'weak_ref_tensor'
```
3. LoRA is affected too, because its `bgmv` kernels are custom kernels.
4. Other potential problems.

# Validation

```python
import torch
import vllm
import vllm_ascend.vllm_ascend_C

torch.ops._C_ascend.weak_ref_tensor
```

If custom kernels weren't compiled, this will raise errors.

# Workaround
 Double check the installation process and make sure you follow the guide, validate your installation with the methods mentioned above.

# Reproduction

There are multiple ways to trigger this. For example, manually changing the `torch_npu` version in `requirements.txt` can cause it.

# Suggestion

@wangxiyuan , I suggest we take a look and make sure compilation failures produce visible errors. Could you assign someone to investigate and arrange a fix?

PR #3791, opened 2025-10-27
[Build] Force torch version

wangxiyuan:
We notice that sometimes user build vllm-ascend with incorrect torch version. In this case, the build is passed, but when running the code, the error `AttributeError: '_OpNamespace' '_C_ascend' object has no attribute 'weak_ref_tensor'` is raised. Let's force the torch version to 2.7.1 and check the torch version when build from source to fix the issue.

closes: #3342

- vLLM version: v0.11.0rc3


ApsarasX:
The torch version I am using is `2.7.1+cpu`, but I encountered the following error during compilation.
image

image

Issue #3543, opened 2025-10-20
[Misc]: Qwen3-VL-30B-A3B-Instruct部署报错


Sunxiaohu0406:
### Anything you want to discuss about vllm on ascend.

镜像：quay.io/ascend/vllm-ascend:v0.11.0rc0

容器：
docker run --name vllm-ascend -it -d --ipc=host \
--privileged=true \
-w /data \
--device=/dev/davinci0 \
--device=/dev/davinci1 \
--device=/dev/davinci2 \
--device=/dev/davinci3 \
--device=/dev/davinci4 \
--device=/dev/davinci5 \
--device=/dev/davinci6 \
--device=/dev/davinci7 \
--device=/dev/davinci_manager \
--device=/dev/hisi_hdc \
--device=/dev/devmm_svm \
--entrypoint=bash \
-v /usr/local/Ascend/driver:/usr/local/Ascend/driver \
-v /usr/local/dcmi:/usr/local/dcmi \
-v /usr/local/bin/npu-smi:/usr/local/bin/npu-smi \
-v /usr/local/sbin:/usr/local/sbin \
-v /etc/ascend_install.info:/etc/ascend_install.info \
-v /data:/data \
-v /tmp:/tmp \
-v /usr/share/zoneinfo/Asia/Shanghai:/etc/localtime \
-p 50066:50066 \
-e http_proxy=$http_proxy \
-e https_proxy=$https_proxy \
quay.io/ascend/vllm-ascend:v0.11.0rc0

服务：
vllm serve /data/test/Qwen3-VL-30B-A3B-Instruct \
--host 0.0.0.0 \
--port 50066 \
--served-model-name qwen3vl \
--tensor-parallel-size 4 \
--max-num-batched-tokens 4096 \
--trust-remote-code \
--no-enable-prefix-caching \
--gpu-memory-utilization 0.85

报错：
al_subscribe_addr='ipc:///tmp/1cf29284-4fd8-4ba3-bed1-539848833442', remote_subscribe_addr=None, remote_addr_ipv6=False)
INFO 10-20 11:34:15 [parallel_state.py:1208] rank 0 in world size 4 is assigned as DP rank 0, PP rank 0, TP rank 0, EP rank 0
INFO 10-20 11:34:15 [parallel_state.py:1208] rank 1 in world size 4 is assigned as DP rank 0, PP rank 0, TP rank 1, EP rank 1
[...]

PR #3795, opened 2025-10-27
[Platform][Bugfix] Add platform interface to check `max_model_len`

shen-shanshan:
When we launch vllm without specifying `--max-model-len`, it will use the default value derived from the model config, e.g., `max_position_embeddings`.

```bash
vllm serve /root/.cache/modelscope/hub/models/Qwen/Qwen3-VL-30B-A3B-Instruct \
--tensor-parallel-size 4 \
--enable-expert-parallel \
--enforce-eager
```

With regard to `Qwen3-VL-30B-A3B`, its default `max_model_len` is `262144`, which is too large and will make the engine stuck in initializing distributed communication.

```bash
...
Using max model len 262144
...
INFO 10-25 02:18:40 [parallel_state.py:1208] rank 0 in world size 4 is assigned as DP rank 0, PP rank 0, TP rank 0, EP rank 0
INFO 10-25 02:18:40 [parallel_state.py:1208] rank 1 in world size 4 is assigned as DP rank 0, PP rank 0, TP rank 1, EP rank 1
INFO 10-25 02:18:40 [parallel_state.py:1208] rank 3 in world size 4 is assigned as DP rank 0, PP rank 0, TP rank 3, EP rank 3
INFO 10-25 02:18:40 [parallel_state.py:1208] rank 2 in world size 4 is assigned as DP rank 0, PP rank 0, TP rank 2, EP rank 2
# Get stuck in here
```

Thus, we need to set a limited default value when `--max-model-len` is not specified. After testing, I find that the engine will get stuck when `max_model_len` > **65536** on Ascend A2 devices. Thus, we set it to **65536** when the default value is too large.

> [!NOTE]
> Find more details at https://github.com/vllm-project/vllm/pull/27556 and https://github.com/vllm-project/vllm-ascend/issues/3543.
> Part of https://github.com/vllm-project/vllm-ascend/issues/3508.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -43,6 +43,8 @@ else:
     VllmConfig = None
     FlexibleArgumentParser = None
 
+ASCEND_MAX_MODEL_LEN = 65535
+
 
 class NPUPlatform(Platform):
 
@@ -476,13 +478,14 @@ class NPUPlatform(Platform):
         """
         Check max_model_len for the current platform.
         """
-        if max_model_len > 65536:
+        if max_model_len > ASCEND_MAX_MODEL_LEN:
             logger.warning(
                 "max_model_len is not specified and the default value"
                 "derived from the model config is %d, which is too large"
                 "and will make the engine stuck in initializing distributed"
-                "communication. Set max_model_len to 65536.",
+                "communication. Set max_model_len to %d.",
                 max_model_len,
+                ASCEND_MAX_MODEL_LEN,
             )
-            max_model_len = 65535
+            max_model_len = ASCEND_MAX_MODEL_LEN
         return max_model_len


shen-shanshan:
CC @wangxiyuan 

Upstream has merged this https://github.com/vllm-project/vllm/pull/27556

vllm_ascend/platform.py
```diff
@@ -43,6 +43,8 @@
     VllmConfig = None
     FlexibleArgumentParser = None
 
+ASCEND_MAX_MODEL_LEN = 65535
```
wangxiyuan, line 46: how this value is from, why 65535?

shen-shanshan: I will further explore how to set this value later.

Issue #3785, opened 2025-10-27
[Bug]: kimi-k2 start bug, weight load ERROR


Levi-JQ:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
Init 115487 task_queue_enable = 1
Collecting environment information...
PyTorch version: 2.7.1+cu126
Is debug build: False

OS: Alibaba Group Enterprise Linux Server 7.2 (Paladin) (x86_64)
GCC version: (GCC) 12.3.1 (openEuler 12.3.1-38.oe2403)
Clang version: Could not collect
CMake version: version 4.1.0
Libc version: glibc-2.38

Python version: 3.11.6 (main, Sep 23 2025, 18:31:41) [GCC 12.3.1 (openEuler 12.3.1-38.oe2403)] (64-bit runtime)
Python platform: Linux-5.10.112-100.alios7.x86_64-x86_64-with-glibc2.38

CPU:
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Address sizes:                   46 bits physical, 57 bits virtual
Byte Order:                      Little Endian
CPU(s):                          192
On-line CPU(s) list:             0-191
Vendor ID:                       GenuineIntel
BIOS Vendor ID:                  Intel(R) Corporation
Model name:                      Intel(R) Xeon(R) Platinum 8468
BIOS Model name:                 Intel(R) Xeon(R) Platinum 8468  CPU @ 2.1GHz
BIOS CPU family:                 179
CPU family:                      6
Model:                           143
Thread(s) per core:              2
Core(s) per socket:              48
Socket(s):                       2
Stepping:                        8
BogoMIPS:                        4200.00
[...]

PR #3798, opened 2025-10-27
【Bugfix】bugfix for weight load of  kimi-k2

Levi-JQ:
Signed-off-by: Levi-JQ <yujinqi2@huawei.com>

Fix kimi-k2 start bug, weight load ERROR：https://github.com/vllm-project/vllm-ascend/issues/3785

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/quantization/quant_config.py
+++ b/vllm_ascend/quantization/quant_config.py
@@ -193,6 +193,11 @@ packed_modules_model_mapping = {
         ["experts.0.gate_proj", "experts.0.up_proj", "experts.0.down_proj"],
         "fused_qkv_a_proj": ["q_a_proj", "kv_a_proj_with_mqa"]
     },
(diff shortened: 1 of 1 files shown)


Potabk:
use `git commit -sm` to add a signature

vllm_ascend/quantization/quant_config.py
```diff
@@ -193,6 +193,11 @@ def get_scaled_act_names(self) -> List[str]:
         ["experts.0.gate_proj", "experts.0.up_proj", "experts.0.down_proj"],
         "fused_qkv_a_proj": ["q_a_proj", "kv_a_proj_with_mqa"]
     },
+    "kimi_k2": {
```
wangxiyuan, line 196: it's good to add all moe model to support the mapping in the future.

PR #3532, opened 2025-10-18
add `dispatch_gmm_combine` kernel

kiscad:
### What this PR does / why we need it?

This PR introduces the Ascend implementation of the `dispatch_ffn_combine` kernel and wires it into the vLLM-Ascend runtime, together with follow‑up fixes to ensure the kernel builds and runs correctly in CI.

- Add full host and device implementation of the `dispatch_ffn_combine` kernel under `csrc/dispatch_ffn_combine`, including tiling logic, MOE routing helpers, and kernel utilities for quantized FFN dispatch.
- Integrate the new kernel with the PyTorch binding (csrc/torch_binding.cpp, csrc/torch_binding_meta.cpp) and the Ascend runtime (vllm_ascend/ascend_forward_context.py, vllm_ascend/worker/model_runner_v1.py).
- Extend fused MoE communication and token dispatch support in `vllm_ascend/ops/fused_moe`, adding methods/utilities needed by the new dispatch path.
- Update quantization logic in vllm_ascend/quantization/w8a8_dynamic.py to support the new FFN dispatch flow.
- Fix kernel build issues by adjusting `csrc/build_aclnn.sh`, CMake configuration, and include/namespace usage in the new kernel files.
- Add an end‑to‑end nightly test `tests/e2e/nightly/ops/test_dispatch_ffn_combine.py` and helper utilities in `vllm_ascend/utils.py` to validate the new kernel.

### How was this patch tested?

- vLLM version: v0.12.0
- vLLM main: https://github.com/vllm-project/vllm/commit/ad32e3e19ccf0526cb6744a5fed09a138a5fb2f9

PR #3226, opened 2025-09-28
add mla_preprocess kernel

kiscad:

PR #3804, opened 2025-10-27
[Kernel] add custom op GmmSwigluQuantWeightNzTensorList

ChenxiQ:

This PR introduces support for adding custom CANN `aclnn` ops to `vllm-ascend`, allowing users to define and use their own custom operators.

Key changes include:
- Building and installing custom ops into the `vllm-ascend`-specified directory
- Binding the `aclnn` op interface to the `torch.ops._C_ascend` module
- Enabling invocation of these ops within `vllm-ascend`

This PR includes a sample custom op: `aclnnGroupedMatmulSwigluQuantWeightNzTensorList`, which is adapted from the CANN operator [`aclnnGroupedMatmulSwigluQuantWeightNZ`](https://www.hiascend.com/document/detail/zh/canncommercial/83RC1/API/aolapi/context/aclnnGroupedMatmulSwigluQuantWeightNZ.md).  
Its input parameters `weight` and `weight_scale` now accept `list[torch.Tensor]` (i.e., `at::TensorList`).

No.

- vLLM version: v0.11.2


csrc/cmake/config.cmake
```diff
@@ -0,0 +1,257 @@
...
+# This file is a part of the CANN Open Software.
+# Licensed under CANN Open Software License Agreement Version 1.0 (the "License").
+# Please refer to the License for details. You may not use this file except in compliance with the License.
+# THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+# INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+# See LICENSE in the root of the software repository for the full text of the License.
+# ======================================================================================================================
+
+########################################################################################################################
+# 环境检查
```
SlightwindSec: Please use English for all comments in the code.

csrc/torch_binding.cpp
```diff
@@ -520,6 +522,41 @@ at::Tensor sgmv_expand(at::Tensor &x, at::Tensor &weight, at::Tensor &lora_indic
...
+    auto x_size = x.sizes();
+    int n = weight[0].sizes()[1];
+    int m = x_size[0];
+    int k = x_size[1];
+
+    at::Tensor output = at::zeros({m, n/2}, x.options().dtype(at::kChar));
+    at::Tensor output_scale = at::zeros({m}, x.options().dtype(at::kFloat));
+    at::Tensor output_offset = at::zeros({m}, x.options().dtype(at::kFloat));
+
+    EXEC_NPU_CMD(
```
SlightwindSec, line 574: * I noticed this PR introduces the `EXEC_NPU_CMD` macro and its dependencies, which adds a significant amount of code (~10k+ lines). Could we instead follow the pattern from **PR #3226** for registering the custom op? This should help keep the adaptation layer much smaller.
* Looking ahead, if we want to use helper macros like this from op-plugin within `vllm-ascend`, the best long-term path would be to promote `torch_npu` to expose these interfaces publicly. That way, we can call them directly instead of vendoring all the dependency code.

csrc/build_aclnn.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# build custom ops
+cd csrc/cann_ops_custom
+bash build.sh -n grouped_matmul_swiglu_quant -c ascend910b --disable-check-compatible
+
+# install custom ops
+./output/CANN-custom_ops--linux.x86_64.run
```
SlightwindSec: I see the new custom operator is compiled into a separate `.run` installer. This approach seems to complicate the build process and will likely cause problems for users who want to build a Python wheel (`.whl`) package, as the operator won't be included.

**Suggestion:** Could we compile this operator directly into the main `vllm-ascend` shared library?

This would simplify the build, fix the packaging issue, and align with how other operators are handled (e.g., in PR #3226).

ChenxiQ: This PR takes #3532 as an example, introducing a new path to integrate custom ops into `vllm-ascend`. The ops in both PRs follows the standard invocation of `aclnn` ops, which currently is different from all other custom ops on `vllm-ascend`.

The standard two-step invocation of `aclnn` ops requires the invocation of `aclnnXXXGetWorkspace` and `aclnnXXX`, which is originally implemented in `op-plugin`. Current custom ops on `vllm-ascend` looks like they are either ATB ops or avoiding these by invoking ops directly.

csrc/build_aclnn.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+ROOT_DIR=$1
+SOC_VERSION=$2
+
+case "$SOC_VERSION" in
```
wangxiyuan: SOC_VERSION: Enum("310", "910b", "910c", "950")

csrc/cmake/config.cmake
ChenxiQ: Fixed.

PR #3805, opened 2025-10-27
[CI] Add multi-node test case for a2

Potabk:
This patch add multi-node test case for a2

- vLLM version: v0.11.0rc3


Potabk:
This should merged after #3764

PR #3824, opened 2025-10-28
[Bugfix] [MoE] fix error in deepseek when using allgather

Pr0Wh1teGivee:
After refactoring vllm_ascend/models and FusedMoE, we are unable to pass `gate` from deepseekv2.py to `AscendFusedMoE.forward`, which will result in error when running deepseek v3/r1 with allgather.
Hence, this pr removes `gate` related computations from FusedMoE module in eager/aclgraph mode.
`rm_router_logits` is deprecated in eager/aclgraph.
e2e & ut

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/ops/fused_moe/prepare_finalize.py
+++ b/vllm_ascend/ops/fused_moe/prepare_finalize.py
@@ -27,7 +27,7 @@ from vllm.distributed.parallel_state import (
 from vllm.forward_context import get_forward_context
 from vllm.model_executor.layers.fused_moe import FusedMoEConfig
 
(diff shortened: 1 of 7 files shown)


vllm_ascend/ops/fused_moe/prepare_finalize.py
```diff
@@ -27,7 +27,7 @@
 from vllm.forward_context import get_forward_context
 from vllm.model_executor.layers.fused_moe import FusedMoEConfig
 
-from vllm_ascend.utils import enable_sp, get_rm_router_logits_state
```
Pr0Wh1teGivee, line 30: move to torchair/utils

PR #3833, opened 2025-10-28
[Model] Add LongCat-Flash

chuyuelin:
Add LongCat-Flash support.
N/A
CI passed

- vLLM version: v0.12.0


vllm_ascend/ops/fused_moe/experts_selector.py
```diff
@@ -195,7 +204,8 @@ def _select_experts_with_fusion_ops(
             # y2_flag=False, # old api; should the third output be output
             routed_scaling_factor=1,
             eps=float(1e-20))
-    if not use_grouped_topk and custom_routing_function is None and scoring_func == "softmax":
+    is_longcat_flash = zero_expert_num != 0 and zero_expert_type is not None
+    if not use_grouped_topk and custom_routing_function is None and scoring_func == "softmax" and not is_longcat_flash:
```
zzzzwwjj: `npu_moe_gating_top_k` op can support longcat, you can give it a try.

vllm_ascend/models/layers/mla.py
```diff
@@ -82,7 +82,11 @@ def __init__(
         hf_config = get_current_vllm_config().model_config.hf_config
         self.enable_shared_expert_dp = get_ascend_config(
         ).enable_shared_expert_dp
-        self.debug_layer_idx = int(self.prefix.split(".")[-2])
+        if hf_config.model_type == "longcat_flash":
+            self.debug_layer_idx = int(self.prefix.split(".")[2])
+            hf_config.first_k_dense_replace = 0
```
zzzzwwjj: use `getattr()` or `.get()` instead of it.

zzzzwwjj (changes requested):

vllm_ascend/ops/rotary_embedding.py
```diff
@@ -127,6 +127,16 @@ def __init__(
         self.sin = None
         super().__init__(head_size, rotary_dim, max_position_embeddings, base,
                          is_neox_style, dtype)
+        self._set_cos_sin_cache(dtype=dtype)
+
+    def _set_cos_sin_cache(self, dtype):
```
whx-sjtu: why we need this?

vllm_ascend/quantization/w8a8_dynamic.py
```diff
@@ -197,6 +198,7 @@ def apply(
         num_expert_group: Optional[int] = None,
         custom_routing_function: Optional[Callable] = None,
         scoring_func: str = "softmax",
+        routed_scaling_factor: float = 1.0,
```
whx-sjtu, line 187: Is `routed_scaling_factor` passed from vllm model script?

chuyuelin: Yes, the routed_scaling_factor is indeed passed from the vllm model script

vllm_ascend/models/layers/mla.py
chuyuelin: Since there's no need to distinguish between model types here, I have removed it.

vllm_ascend/ops/fused_moe/experts_selector.py
chuyuelin: I have already used the npu_moe_gating_top_k op.

vllm_ascend/ops/rotary_embedding.py
chuyuelin: The AscendMLAMetadataBuilder class requires both cos_cached and sin_cached. However, the longcat model uses RotaryEmbedding instead of YaRNRotaryEmbedding, which necessitates setting up the cos_cached and sin_cached.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3269,9 +3269,10 @@ def initialize_kv_cache_tensors(
         kv_caches = self._reshape_kv_cache_tensors(kv_cache_config,
                                                    kv_cache_raw_tensors)
 
+        num_attn_module = 2 if self.model_config.hf_config.model_type == "longcat_flash" else 1
```
whx-sjtu, line 2311: Is there any other params in model config that indicates the number of attention modules? Don't hard code model_type unless imperative.

whx-sjtu (changes requested):

vllm_ascend/worker/model_runner_v1.py
chuyuelin: Currently, only LongCat model has two attention modules per layer, and the model_type parameter is the only way to distinguish this. This is also how it's implemented in the vLLM codebase.

whx-sjtu: OK, if that's true, please add a TODO note here and deletes this in the future. Does vllm has plans to refactor this? If not, maybe we can get involved.

chuyuelin: TODO note has been added. Currently, vLLM doesn't have any plans for refactoring this. We can keep an eye on this part and potentially get involved in future improvements.

wangxiyuan (approved):

vllm_ascend/attention/mla_v1.py
```diff
@@ -441,10 +441,18 @@ def build(
...
+                    nn.ModuleList):
+                self.cos_cache = model.model.layers[
+                    model.model.start_layer].self_attn[0].rotary_emb.cos_cached
+                self.sin_cache = model.model.layers[
+                    model.model.start_layer].self_attn[0].rotary_emb.sin_cached
+            else:
+                self.cos_cache = model.model.layers[
+                    model.model.start_layer].self_attn.rotary_emb.cos_cached
+                self.sin_cache = model.model.layers[
+                    model.model.start_layer].self_attn.rotary_emb.sin_cached
```
yiz-liu: Please take a look at #4764 , we might need to unify this cos/sin init method.

wangxiyuan (approved):

yiz-liu (approved):

jianzhnie:
Longcat 模型的TPx EP 支持已经向官方仓库提交了PR   https://github.com/vllm-project/vllm-ascend/pull/13241

PR #3834, opened 2025-10-28
[Main][Bugfix]Avoid using the fusion operator in the MOE model

ZYang6263:
The current MatmulReduceScatter operator experiences performance degradation in small-shape scenarios, so it determines whether to use this operator by judging the size of the shape.

- vLLM version: v0.11.0rc3

### DIFF
--- a/vllm_ascend/ascend_forward_context.py
+++ b/vllm_ascend/ascend_forward_context.py
@@ -109,26 +109,25 @@ def set_ascend_forward_context(
         # due to multiple warmups before actual capturing
         forward_context.capturing = False
 
-        # set for sequence parallelism, 1024 is the batch size concurrency threshold for enabling the flashcomm_v1 or sequence_parallelism feature.
+        # set for sequence parallelism, 1000 is the batch size concurrency threshold for enabling the flashcomm_v1 or sequence_parallelism feature.
         # Currently, it is an empirical value. In normal scenarios, if the concurrency exceeds this threshold,
         # the performance benefits can be maximized. Conversely, if the concurrency is below the threshold,
         # the performance may degrade due to the switching of communication methods.
+        mmrs_fusion = True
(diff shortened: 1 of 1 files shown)


vllm_ascend/ascend_forward_context.py
```diff
@@ -125,6 +125,7 @@ def set_ascend_forward_context(
             pad_size = (tp_world_size -
                         (num_tokens % tp_world_size)) % tp_world_size
             forward_context.pad_size = pad_size
+            forward_context.comm_comp_fusion = num_tokens is not None and num_tokens > 1024
```
rjg-lyh: I suggest making this variable name more descriptive, for example, mm_rs_fusion.

vllm_ascend/ops/linear_op.py
```diff
@@ -410,7 +411,7 @@ def matmul_and_reduce(self, input_parallel: torch.Tensor,
 
         # For unquant
         if isinstance(self.layer.quant_method, UnquantizedLinearMethod
-                      ) and torch.version.cann.startswith("8.3"):
+                      ) and comm_comp_fusion and torch.version.cann.startswith("8.3"):
```
rjg-lyh: This check can be moved to the very beginning to reduce the overhead caused by redundant checks.

vllm_ascend/ops/linear_op.py
```diff
@@ -382,6 +382,7 @@ def matmul_and_reduce(self, input_parallel: torch.Tensor,
         try:
             forward_context = get_forward_context()
             sp_enabled = forward_context.sp_enabled
+            comm_comp_fusion = forward_context.comm_comp_fusion
         except AssertionError:
```
rjg-lyh, line 386: I suggest keeping the behavior consistent with sp_enabled by setting a default value when an error is caught.

PR #3835, opened 2025-10-28
[v0.11.0] [P/D] force with_prefill true after allreduce in kv producer

liziyu179:
force with_prefill true after allreduce in kv producer


MengqingCao:
This is a backport of #3768 and #3849

PR #3848, opened 2025-10-29
[Perf] Move attention update stream out of loop to optimize performance

momo609:
In the `update_*attn_params` functions, the `torch.npu.stream(update_stream)` context manager was previously located inside the for-loop that updates parameters for each layer. This resulted in redundant stream initiations for every layer, adding unnecessary overhead.

This commit refactors the code by moving the stream context manager to wrap the entire for-loop. This ensures that the update stream is initiated only once per function call, rather than for each layer. This change reduces 90us in each decode model.
update stream in every layer:
image

remove update stream in every layer:
image

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/compilation/acl_graph.py
+++ b/vllm_ascend/compilation/acl_graph.py
@@ -194,26 +194,25 @@ def update_attn_params(update_stream, forward_context, runtime_shape):
     graph_params = get_graph_params()
     # FIXME: Behold! We are using a temporary hack here to update the args
     # for each layer's attention op in the graph.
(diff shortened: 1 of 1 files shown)


yiz-liu (changes requested):
Please also refactor `update_mla_attn_dcp_pcp_params` and `update_attn_dcp_pcp_params`.

momo609:
> Please also refactor `update_mla_attn_dcp_pcp_params` and `update_attn_dcp_pcp_params`.
dcp，pcp cannot test now, will add in other pr.

PR #3870, opened 2025-10-29
Add FAQ for docker pull error on Kylin OS

Liwx1014:
Added instructions for resolving 'invalid tar header' error on Kylin OS (麒麟操作系统) with an ARM64 architecture on Atlas300I hardware during docker pull, including steps for offline loading of docker images.

---

The primary motivation for this PR is to address a critical `docker pull` failure that occurs on specific, yet important, enterprise environments. Specifically, when operating on **Kylin OS (麒麟操作系统) with an ARM64 architecture on Atlas300I hardware**, users frequently encounter an `archive/tar: invalid tar header` error, which completely blocks the setup process. This issue has been consistently reproduced, with multiple retries failing with the same error, confirming that it is a persistent environmental problem rather than a transient network issue.

image

This guide provides a robust, step-by-step workaround using an offline-loading method (`docker save` on a host machine and `docker load` on the target machine). This solution is crucial for enabling users on this platform to use vLLM.

This contribution does not directly fix an existing issue number, but it proactively solves a significant environmental and usability problem for a growing user base.

No.It does not alter any code, APIs, interfaces, or existing behavior of the vLLM project.

The instructions and troubleshooting steps in this guide were validated through a real-world, end-to-end test case on the my hardware and OS.

The testing process was as follows:

1.  **Problem Reproduction**: An attempt was made to directly `docker pull` the `vllm-ascend:v0.10.0rc1-310p` image on a target machine running Kylin OS (ARM64). The `invalid tar header` failure was successfully and consistently reproduced, confirming the existence of the problem.
2.  **Solution Implementation**: The workaround detailed in the guide was executed:
    *   On a separate host machine (Ubuntu x86_64), the image was successfully pulled using the `--platform linux/arm64` flag.
    *   The image was then saved to a `.tar` archive using `docker save`.
    *   The `.tar` archive was transferred to the target Kylin OS machine.
    *   The image was successfully loaded from the archive using `docker load -i ...`.
3.  **End-to-End Validation**: After loading the image, the vLLM container was launched on the target machine following the instructions in the guide. Both online inference (via `curl` to the API server) and offline inference (via the Python script) were executed successfully, confirming that the entire workflow described in the document is accurate and effective.

Since this is a documentation-only change based on a validated workflow, no new unit or integration tests were added to the codebase.

- vLLM version: v0.11.0rc3

### DIFF
--- a/docs/source/faqs.md
+++ b/docs/source/faqs.md
@@ -215,7 +215,7 @@ torch-npu will be overridden  when installing vllm-ascend. If you need to instal
 
 ### 20. On certain systems (e.g., Kylin OS), `docker pull` may fail with an `invalid tar header` error
 
-On certain operating systems, such as Kylin OS (麒麟操作系统), you may encounter an `invalid tar header` error during the `docker pull` process:
+On certain operating systems, such as Kylin OS , you may encounter an `invalid tar header` error during the `docker pull` process:
 
 ```text
 failed to register layer: ApplyLayer exit status 1 stdout: stderr: archive/tar: invalid tar header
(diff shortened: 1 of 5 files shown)


wangxiyuan:
Can you fix the lint error? Otherwise LGTM

Liwx1014:
> Can you fix the lint error? Otherwise LGTM

Thanks for the review! I've pushed the fix for the lint error. It was indeed caused by an extra trailing space. The CI should pass now.

PR #3876, opened 2025-10-29
deepseek torchair adapt for torch_npu version

hust17yixuan:
To adapt the torch_npu version to avoid the precision problem of torchair deepseek. The torch_npu version may result in the different branches in the ops register, the rms_norm ops has two branches according to the verson_check, this pr unify the rms_norm in torchair by patch method. #3862 

No


MengqingCao (approved):
LGTM, this is a backport of https://github.com/vllm-project/vllm-ascend/commit/870a3f21cb56b344217908fbed764400df8087a8. pre-commit is failed due to the unrelated docs

PR #3879, opened 2025-10-29
add new e2e tests case for aclgraph memory

lilinsiman:
add new e2e tests case for aclgraph memory

no

ut

- vLLM version: v0.11.0rc3

### DIFF
--- a/tests/e2e/singlecard/test_aclgraph_mem.py
+++ b/tests/e2e/singlecard/test_aclgraph_mem.py
@@ -1,0 +1,100 @@
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+# Copyright 2023 The vLLM team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


tests/e2e/singlecard/test_aclgraph_mem.py
```diff
@@ -0,0 +1,100 @@
...
+
+    assert capture_called.value == 1, "_capture_model was not called during test"
+    assert capture_mem_before.value != -1, "capture_mem_before not set"
+    assert capture_mem_after.value != -1, "capture_mem_after not set"
+
+    print("capture_mem_before =", capture_mem_before.value)
+    print("capture_mem_after =", capture_mem_after.value)
+
+    mem_used_by_capture = capture_mem_before.value - capture_mem_after.value
+    # Empirical observation: capturing ACL graphs for Qwen3-0.6B uses ~0.20 GiB of NPU memory.
```
yiz-liu, line 85: Maybe we need to update this observation from time to time.

yiz-liu (approved):

tests/e2e/singlecard/test_aclgraph_mem.py
lilinsiman: > Maybe we need to update this observation from time to time.

After extensive testing, multiplying the threshold by 1.3 and 1.5 has provided a significant margin of allowable variation.

lilinsiman: > Maybe we need to update this observation from time to time.

If subsequent code causes the memory usage to exceed this value, it indicates that an abnormal value has been reached, which poses a risk of OOM. In this case, I suggest re-evaluating the code to be merged.

PR #3892, opened 2025-10-30
[Feat] Support MTP to running in full graph mode

anon189Ty:
Currently, the MTP model still runs in eager in full graph mode. This PR adapts the MTP with the full graph capture and execution. When the graph mode is set to "FULL_DECODE_ONLY", the MTP will run in full-graph to improve the performance.

The change in both disable_padded_drafter_batch is True and False case include:

1. Add _mtp_graph_params in acl_graph.py to isolate the data of main model and the data of MTP.
2. Padding some metadata in mla_v1.py when in fullgraph mode.
3. Fixed the essential data address that will be used in model.forward.
4. Adapted according to the aclgraph capture framwork:
    1). Rebuild MTP model with ACLGraphWrapper.
    2). Add common attn metadata when start capture in MTP dummy_run.
    3). Add common attn metadata update in MTP.
    4). Addapted data update when num_speculative_tokens > 1.
5. Add a patch of MTP to adapt vllm v0.11.0.

Existing Issues:
  1. When disable_padded_drafter_batch=True and running in FullGraph mode, the data of the first-round requests in MTP is abnormal. We need to identify the cause subsequently.
  2. When disable_padded_drafter_batch=False and running in FullGraph mode, the acceptance rate of the second and third tokens will decrease (For example, if we set the num_speculative_tokens=3, the acceptance rate of first token is 90%, the second is only 50% lower than 60%, the third is only 20% lower than 30%). The reason is that the data processed after the model runs does not match. This is a problem from another PR. It works fine in eager and PIECEWISE mode, but has problem in FullGraph mode. Once we have a solution, we will submit a bugfix.

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -434,11 +434,13 @@ class AscendMLAMetadataBuilder:
 
         # If graph_pad_size > -1, mean is running in fullgraph mode.
         graph_pad_size = common_attn_metadata.graph_pad_size
+        # NOTE: Maybe this block_table change can be removed when graph_pad_size > 1.
         if graph_pad_size > num_reqs and self.speculative_config.disable_padded_drafter_batch:
             block_table = (
                 common_attn_metadata.block_table_tensor[:graph_pad_size])
         else:
             block_table = (common_attn_metadata.block_table_tensor[:num_reqs])
+        # NOTE: Currently, MTP-fullgraph is incompatibility pcp
         if self.pcp_size > 1:
             num_decodes_flatten = num_decodes * self.decode_threshold
             block_table = common_attn_metadata.block_table_tensor[:
@@ -448,6 +450,7 @@ class AscendMLAMetadataBuilder:
         if num_actual_tokens_pcp_padded is None:
             num_actual_tokens_pcp_padded = num_actual_tokens
 
+        # NOTE: Currently, MTP-fullgraph is incompatibility pcp
         slot_mapping = common_attn_metadata.slot_mapping[:
                                                          num_actual_tokens_pcp_padded]
         input_positions = common_attn_metadata.positions[:
@@ -640,8 +643,10 @@ class AscendMLAMetadataBuilder:
                 block_table = block_table[:num_decodes_flatten, ...]
             else:
                 block_table = block_table[:num_decodes, ...]
-            # NOTE: Currently, MTP-fullgraph is incompatibility pcp 
-            if graph_pad_size > num_decodes and self.speculative_config.disable_padded_drafter_batch:
+            # NOTE: Currently, MTP-fullgraph is incompatibility pcp
+            # NOTE: Maybe this block_table change can be removed when graph_pad_size > 1.
+            if graph_pad_size > num_decodes and \
+                    self.speculative_config.disable_padded_drafter_batch:
                 block_table = block_table[:graph_pad_size, ...]
             seq_lens_list = seq_lens.tolist()
 
@@ -663,7 +668,7 @@ class AscendMLAMetadataBuilder:
                 cp_seq_len = torch.where(cp_seq_len == 0, 1, cp_seq_len)
             else:
                 cp_seq_len, batch_seq_mask = None, None
-            
+
             if graph_pad_size > num_reqs:
                 if self.speculative_config.disable_padded_drafter_batch:
                     num_reqs_pad_size = graph_pad_size - num_reqs
@@ -677,8 +682,8 @@ class AscendMLAMetadataBuilder:
                             (num_block_pad_size, ) + block_table.shape[1:],
                             dtype=block_table.dtype,
                             device=block_table.device)
-                        block_table = torch.cat([block_table, block_table_padding],
-                                                dim=0)
+                        block_table = torch.cat(
+                            [block_table, block_table_padding], dim=0)
                 else:
                     num_token_pad_size = graph_pad_size - num_decode_tokens
                     num_reqs_pad_size = (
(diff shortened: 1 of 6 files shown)


vllm_ascend/ascend_forward_context.py
```diff
@@ -157,6 +158,7 @@ def set_ascend_forward_context(
         forward_context.prefetch_mlp_enabled = prefetch_mlp_enabled
         forward_context.model_instance = model_instance
         forward_context.weight_prefetch_method = weight_prefetch_method
+        forward_context.is_mtp_model = is_mtp_model
```
yiz-liu, line 162: We should try to remove this flag in the future.

vllm_ascend/attention/mla_v1.py
```diff
@@ -1447,8 +1573,11 @@ def _forward_decode(
             "actual_seq_lengths": actual_seq_lengths,
             "actual_seq_lengths_kv": decode_meta.seq_lens_list,
         }
-        graph_params = get_graph_params()
         forward_context: ForwardContext = get_forward_context()
+        if forward_context.is_mtp_model:
+            graph_params = get_mtp_graph_params()
+        else:
+            graph_params = get_graph_params()
```
yiz-liu, line 1403: Graph params is also deviating from its initial design and having more and more if-else.

vllm_ascend/compilation/acl_graph.py
```diff
@@ -278,6 +282,12 @@ def update_mla_attn_params(update_stream, forward_context, runtime_shape,
                     spec_multiple * (i + 1)
                     for i in range(runtime_shape // spec_multiple)
                 ]
+            elif forward_context.is_mtp_model:
+                actual_seq_lengths = forward_context.attn_metadata[
+                    key].decode.actual_seq_lengths_q
+                block_table = forward_context.attn_metadata[key].decode.block_table
+                seq_lens_list = seq_lens_list + [0] * (len(actual_seq_lengths) -
+                                                       len(seq_lens_list))
```
yiz-liu: If we can delete this, then `is_mtp_model` is no longer necessary right?

vllm_ascend/patch/worker/patch_deepseek_mtp.py
```diff
@@ -51,4 +67,161 @@ def predictor_init(self, vllm_config: VllmConfig, prefix: str) -> None:
...
+
+    hidden_states, residual = self.mtp_block(positions=positions,
+                                             hidden_states=hidden_states,
+                                             residual=None)
+    hidden_states = residual + hidden_states
+    return hidden_states
+
+
+@support_torch_compile
+class CustomDeepSeekMTP(nn.Module, SupportsPP):
```
yiz-liu: Maybe try slim this.

vllm_ascend/spec_decode/mtp_proposer.py
```diff
@@ -751,12 +874,19 @@ def _propose(
...
+                    attn_metadata_i.decode.seq_lens_list = decode_seq_lens_list + [
+                        0
+                    ] * (graph_pad_size - len(decode_seq_lens_list))
                 attn_metadata_i.decode.input_positions = self.positions[:
                                                                         num_input_tokens]
                 attn_metadata_i.decode.max_seq_lens += 1
                 attn_metadata_i.decode.max_seq_lens = min(
                     attn_metadata_i.decode.max_seq_lens,
                     self.runner.model_config.max_model_len)
+            torch.npu.synchronize()
```
yiz-liu, line 889: Remove this line after #4233 being merged.

vllm_ascend/attention/mla_v1.py
```diff
@@ -343,6 +345,74 @@ def reorder_batch(self, input_batch: "InputBatch",
...
+            assert len(interpolated) == len(padding_seq_len_q)
+            actual_seq_lengths_q = actual_seq_lengths_q + interpolated
+        else:
+            actual_seq_lengths_q = actual_seq_lengths_q + common_attn_metadata.actual_seq_lengths_q[
+                num_reqs:num_reqs + num_reqs_pad_size]
+
+        return actual_seq_lengths_q
+
+    def pad_actual_seq_len_q_mtp_disable_pad(self, num_reqs_pad_size, num_reqs,
+                                             actual_seq_lengths_q):
```
zouyida2052: In this case, when we enable num_mtp_tokens=3, the acceptance rate in positions 2 and 3 is 0 in first request.

vllm_ascend/attention/mla_v1.py
```diff
@@ -343,6 +345,74 @@ def reorder_batch(self, input_batch: "InputBatch",
         # better way of doing this
         return modified_batch
 
+    def pad_actual_seq_len_q_mtp_enable_pad(self, num_reqs_pad_size, num_reqs,
+                                            actual_seq_lengths_q,
+                                            common_attn_metadata):
```
zouyida2052: In this case, mean acceptance length seems to be a little lower than piecewise mode or eager mode.

vllm_ascend/attention/mla_v1.py
```diff
@@ -343,6 +345,74 @@ def reorder_batch(self, input_batch: "InputBatch",
...
+            assert interpolated[-1] == end_val
+            assert len(interpolated) == len(padding_seq_len_q)
+            actual_seq_lengths_q = actual_seq_lengths_q + interpolated
+        else:
+            actual_seq_lengths_q = actual_seq_lengths_q + common_attn_metadata.actual_seq_lengths_q[
+                num_reqs:num_reqs + num_reqs_pad_size]
+
+        return actual_seq_lengths_q
+
+    def pad_actual_seq_len_q_mtp_disable_pad(self, num_reqs_pad_size, num_reqs,
```
whx-sjtu, line 385: The function name `pad_actual_seq_len_q_mtp_disable_pad` looks confusing to me. If the function is used to pad `actual_seq_len_q`, why the name ended with diable_pad?

vllm_ascend/attention/mla_v1.py
```diff
@@ -343,6 +345,74 @@ def reorder_batch(self, input_batch: "InputBatch",
         # better way of doing this
         return modified_batch
 
+    def pad_actual_seq_len_q_mtp_enable_pad(self, num_reqs_pad_size, num_reqs,
```
whx-sjtu, line 342: ditto

vllm_ascend/attention/mla_v1.py
anon189Ty: The "disable_pad" in function name means "disable_padded_drafter_batch=True", but it's too long. Similarly, the "enable_pad" means "disable_padded_drafter_batch=False".

PR #3912, opened 2025-10-30
[0.11.0][MTP][Aclgraph] Fix the support aclgraph with MTP

MengqingCao:
Fix 2 breaks of aclgraph with MTP:
1. deepseekmtp in vllm 0.11.0 does not support aclgraph and lack the `support_torch_compile` decorator
2. There is a d2h synchornization in the original forward of mtp predictor. The fix pr in vllm https://github.com/vllm-project/vllm/pull/27643

As we'll fix it in vllm main, this fix pr is only needed in branch v0.11.0-dev

The profling shows that MTP replays in aclgraph now:
a7d7f04155df4ed454b7eb20a92b2e2a

### DIFF
--- a/vllm_ascend/patch/worker/patch_deepseek_mtp.py
+++ b/vllm_ascend/patch/worker/patch_deepseek_mtp.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 import torch
 import torch.nn as nn
 import vllm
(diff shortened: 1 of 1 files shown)


vllm_ascend/patch/platform/patch_mtp_predictor.py
```diff
@@ -0,0 +1,45 @@
+# SPDX-License-Identifier: Apache-2.0
```
MengqingCao: move to exsited patch mtp in worker

MengqingCao:
@wangxiyuan @zzzzwwjj This pr is ready for merge, plz take a look, thanks!

PR #3929, opened 2025-10-31
[v0.11.0-dev][bugfix] fix valueError in static_forward_context when prefix is empty

rjg-lyh:
This PR temporarily bypasses the scenario where some models in vLLM trigger a `ValueError` during the process of storing values in `static_forward_context` when no `prefix` is specified for the linear layers, which is a bug in some models in vLLM. The official fix will be addressed by submitting a PR to the vLLM community that specifies a prefix for the linear layers in each model.

- vLLM version: v0.11.0

No.


MengqingCao:
This is a cherry-pick of https://github.com/vllm-project/vllm-ascend/commit/c1a6aeab460fc2977535070a399c33a614f9ac0f

MengqingCao (approved):

PR #3944, opened 2025-11-02
Quality enhancement: Immediately interrupt execution when allocate NPU memory OOM

leo-pony:
Protect the scene where the first problem occurs. The execution should be interrupted when the video memory application fails, rather than waiting until an illegal address is accessed.
Test result:
1) Exception throws normally:
d0ce7a04-4102-46cc-a981-f44421f388e6

Corresponding code:
2c58c5fa-1882-435e-8d53-7c536846945b
dc9b4dc0-76ae-4a6b-90b3-4fca9ad63ec0

2) HBM Memory malloc test case successfully:
9766fb23-1830-43c5-b29f-fb6973a78000

3) extern launcher test with sleep and weakup: 
d1f31a03-fe07-455a-bf64-7420937d62b3

NA

NA
- vLLM version: v0.11.0

### DIFF
--- a/csrc/camem_allocator.cpp
+++ b/csrc/camem_allocator.cpp
@@ -15,6 +15,8 @@
  */
 
 #include <iostream>
+#include <stdexcept>
+#include <string>
 
 extern "C" {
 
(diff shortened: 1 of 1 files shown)


leo-pony:
After discussing with @weijinqian0  it was determined that this PR only strengthens the quality under conditions of memory OOM (Out of Memory) and does not affect the normal operation of the path. It is recommended to merge it into version v0.11.0-dev.
@weijinqian0 Could you help review this PR?

PR #3981, opened 2025-11-04
[P/D]Make kv-transfer env variable take effect & Fix load-balance proxy

nwpu-zxr:
Make kv-transfer env variable take effect and Fix load-balance proxy.

No.

By CI.

- vLLM version: v0.11.0

### DIFF
--- a/tests/ut/kv_connector/test_mooncake_connector.py
+++ b/tests/ut/kv_connector/test_mooncake_connector.py
@@ -24,15 +24,6 @@ fake_engine = types.ModuleType("mooncake.engine")
 fake_engine.TransferEngine = MagicMock()  # type: ignore[attr-defined]
 sys.modules["mooncake.engine"] = fake_engine
 
-envs_mod = types.ModuleType("vllm_ascend.envs")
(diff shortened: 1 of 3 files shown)


vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -855,6 +857,8 @@ class MooncakeConnectorWorker:
 
     def __init__(self, vllm_config: VllmConfig, engine_id: str):
         self._get_prefill_decode_size(vllm_config)
+        os.environ["ASCEND_TRANSFER_TIMEOUT"] = str(
+            get_transfer_timeout_value())
```
LCAIZJ, line 861: Users can specify the timeout by setting it in their launch script, for example: `export ASCEND_TRANSFER_TIMEOUT=10000`. With the current change, users can no longer configure this value and are forced to use the fixed value returned by get_transfer_timeout_value().

nwpu-zxr: OK, we will fix this problem.

leo-pony:
Please correct the title to "make kv-transfer env take effect", not currently add a new env variable.

wangxiyuan (approved):

PR #3996, opened 2025-11-05
【main】optimization of kimi-k2 in cann8.3

Levi-JQ:
In cann8.3， npu_moe_gating_top_k operator can support expert nums with 384, so kimi can use the operator to get better preformance.

- vLLM version: v0.11.0


wangxiyuan:

PR #4002, opened 2025-11-05
[KVPOOL]Redundancy removal when prefill node saves KV cache

baxingpiaochong:

In the current kvpool, each TP saves its corresponding kv cache when determining the kv cache, which is too cumbersome for the MLA. The MLA should only save the kv cache result of a single tp.

This PR aims to store only the kv cache of a single tp when determining the MLA, without affecting the normal operation of other attention mechanisms.

- vLLM version: v0.11.0


LCAIZJ:
Since only one copy of the data is stored, multiple TP ranks may fetch the same data simultaneously during get(). We need to test whether this causes network pressure and will add performance benchmark results later.

wangxiyuan:

PR #4015, opened 2025-11-06
[Doc] Update structured output doc with upstream link

shen-shanshan:
Currently, the usage of structured output feature in vllm-ascend is totally the same as that in vllm.

Thus, IMO, it's better to remove this doc directly to avoid some case that there are some changes in the upstream doc and we don't update our doc in time, which can be misleading to users.

- vLLM version: v0.12.0


wangxiyuan:
you can update the content to link to vllm. And only include some note. Like lora doc does

shen-shanshan:
> you can update the content to link to vllm. And only include some note. Like lora doc does

@wangxiyuan  Done.

PR #4019, opened 2025-11-06
[Bugfix] fix sleepmode level2 e2e test

wangx700:

enable sleepmode level2 e2e test and add the check logic to ensure the nz is not enabled.

no

use e2e tests

- vLLM version: v0.11.0

### DIFF
--- a/tests/e2e/multicard/test_external_launcher.py
+++ b/tests/e2e/multicard/test_external_launcher.py
@@ -108,6 +108,7 @@ def test_moe_external_launcher(model):
     assert proc.returncode == 0
 
 
+@patch.dict(os.environ, {"VLLM_ASCEND_ENABLE_NZ": "0"})
 def test_external_launcher_and_sleepmode():
     script = Path(
         __file__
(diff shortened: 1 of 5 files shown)


tests/e2e/multicard/test_external_launcher.py
```diff
@@ -154,6 +154,7 @@ def test_external_launcher_and_sleepmode():
     assert proc.returncode == 0
 
 
+@patch.dict(os.environ, {"VLLM_ASCEND_ENABLE_NZ": "0"})
 def test_external_launcher_and_sleepmode_level2():
```
leo-pony: To make external luancher test cases take effect please remove the skip fo this file. as following:
#pytest -sv tests/e2e/multicard/test_external_launcher.py in file .github/workflows/_e2e_test.yaml.

leo-pony:
It look goods to me.
This is a test case for the RL scenario. In the RL scenario, NZ is not used, so there is no issue with the code. Here, we just need to ensure that the RL scenario is properly handled.

PR #4021, opened 2025-11-06
[0.11.0][BugFix] Improve the performance of prefixcache features

underfituu:

The code bug caused an empty bubble. When the npu_paged_cache_load operator was called, it forcibly transferred seq_len2 to the device, which triggered synchronization and interrupted the CPU operator's launch stream.


underfituu:
cherry pick：https://github.com/vllm-project/vllm-ascend/pull/4022 （main）

PR #4022, opened 2025-11-06
[BugFix] Improve the performance of prefixcache features

underfituu:
The code bug caused an empty bubble. When the npu_paged_cache_load operator was called, it forcibly transferred seq_len2 to the device, which triggered synchronization and interrupted the CPU operator's launch stream.

- vLLM version: v0.11.0


underfituu:
nightly CI PASS
CI-PASS-prefix-cache

underfituu:
Prefix cache CI has passed, re-trigger CI.

PR #4029, opened 2025-11-06
[bugfix] fix rmsnorm redundant memory usage

hwhaokun:
The rmsnorm operator was not releasing the allocated memorysharply. This PR fixes this issue, saving 448MB of memory.
-->

before：
before
after：
after

by ci

- vLLM version: v0.11.0


vllm_ascend/ops/layernorm.py
```diff
@@ -110,9 +110,13 @@ def forward_oot(
 
         if residual is not None:
             assert x.size(0) == residual.size(0)
+            previous_x = x
+            previous_residual = residual
             x, residual = _addrmsnorm_forward_oot(
                 self, x, residual, self.next_need_quant_fusion_linear,
                 self.bias)
+            dispose_tensor(previous_x)
+            dispose_tensor(previous_residual)
```
yiz-liu: `dispose_tensor` should be deprecated a long time ago, maybe we can fix this in another way. For example, we could make `_addrmsnorm_forward_oot` an inplace op which modifies the input `x` and `residual` directly.

PR #4045, opened 2025-11-07
[Feature] Integrate Suffix Spec Decoding

fluctlux:
This PR integrate suffix decoding (https://arxiv.org/abs/2411.04975) from vllm (https://github.com/vllm-project/vllm/pull/25784) 

#
Suffix Decoding is a dynamic n-gram matching method that:

1. Uses suffix trees to generate speculative tokens quickly using branch frequency counts.
2. Can keep a history of prior model responses, which tends to work very well with repetitive agentic use cases.
3. Can be dynamically updated with newly generated tokens, and FIFO eviction of older requests.
#
This feature should be implemented as opt-in and remain seamless for users who do not require suffix speculative decoding.

For users who wish to enable it, they must first install arctic-inference:
`pip install arctic-inference
`

After installation, the suffix speculative decoding feature can be enabled using the following speculative config:
`--speculative_config '{"method": "suffix", "num_speculative_tokens": 5}'
`

This PR is currently being tested on vLLM main:https://github.com/vllm-project/vllm/commit/83f478bb19489b41e9d208b47b4bb5a95ac171ac
 with PR https://github.com/vllm-project/vllm/pull/25784

In our previous testing, suffix decoding achieved a 13%-30% throughput improvement over n-gram on the sonnet dataset, tested on vllm-ascend v0.9.1 with concurrency ranging from 2 to 40.

- vLLM version: v0.11.2

### DIFF
--- a/vllm_ascend/spec_decode/suffix_proposer.py
+++ b/vllm_ascend/spec_decode/suffix_proposer.py
@@ -26,7 +26,8 @@ class SuffixDecodingProposer(VllmSuffixDecodingProposer, Proposer):
                   num_reqs=None,
                   num_tokens_across_dp=None,
                   aclgraph_runtime_mode: CUDAGraphMode = CUDAGraphMode.NONE,
-                  batch_descriptor=None):
+                  batch_descriptor=None,
+                  dummy_compute_logits=lambda hidden_states: None):
         pass
 
     def generate_token_ids(self,


zhangguinan:
In our previous testing, suffix decoding achieved a 13%-30% throughput improvement over n-gram on the sonnet dataset, tested on vllm-ascend v0.9.1 with concurrency ranging from 2 to 40.
---May I ask about the configuration of this test: how many cards were deployed, which model was used, and was it in BF16 precision format?

PR #4071, opened 2025-11-08
[P/D]Add a heartbeat mechanism to PD separation

wangxiaoteng888:
Add a heartbeat mechanism to PD separation and clean up historical link information based on heartbeat status.

No

By ci

- vLLM version: v0.11.0


LCAIZJ:
I think in the PD-separation scenario, heartbeat management should be handled by the upper-layer proxy. The proxy should use heartbeats as a basis when routing requests to either P or D. So, is the reason we're making P and D aware of heartbeats here because, due to layer-wise execution, their connection establishment takes relatively long, and thus they need to be heartbeat-aware?

wangxiaoteng888:
> I think in the PD-separation scenario, heartbeat management should be handled by the upper-layer proxy. The proxy should use heartbeats as a basis when routing requests to either P or D. So, is the reason we're making P and D aware of heartbeats here because, due to layer-wise execution, their connection establishment takes relatively long, and thus they need to be heartbeat-aware?

The main reason we do this is to clean up the connector’s historical link information. In the connector, the base_address and port information are persisted. By adding heartbeats here, we can detect disconnections and then conveniently clean up the above persisted information afterward.

wangxiyuan:

wangxiaoteng888:
> Any progress? If this PR is still alive, please rebase to main and make CI happy. Thanks
we do not need heartbeat now. I will close this PR.

PR #4097, opened 2025-11-10
[main][bugfix] Change seq_lens in dummy attn_metadata to max_query_len

Angazenn:
Currently, we set `seq_lens` in dummy attn_metadata to be `max_model_len` to get max workspace for attention during capturing.
However, setting it consistently to be `max_model_len` causing dummy_run to execute a long attention when running actual inference. For example, if there is a single req with `seqs_lens` as [8] but `max_model_len` is 131072, the whole process will be slow down by dummy_run as it execute a fake long-seq attention. Therefore, we instead set it to max_query_len, which is also consistent with vLLM gpu implementation.

No.

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2665,12 +2665,7 @@ class NPUModelRunner(LoRAModelRunnerMixin):
 
             attn_metadata = {}
 
-            # When force_attention == True, the model runs in capturing so we
-            # need seq_lens as max_model_len to get max workspace for attention op.
-            # However, when force_attention == False, the model might be running
-            # normal inference. If dp_size > 1, we only need dummy_run
-            # to execute a short attention with seq_lens as 1.
-            seq_lens = self.model_config.max_model_len if force_attention else 1
+            seq_lens = max_query_len
             self.seq_lens_np[:num_reqs] = seq_lens
             self.seq_lens_np[num_reqs:] = 0
 


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2665,7 +2665,12 @@ def _build_dummy_attn_metadata(
 
             attn_metadata = {}
 
-            seq_lens = self.model_config.max_model_len
+            # When force_attention == True, the model runs in capturing so we
+            # need seq_lens as max_model_len to get max workspace for attention op.
+            # However, when force_attention == False, the model might be running
+            # normal inference. If dp_size > 1, we only need dummy_run
+            # to execute a short attention with seq_lens as 1.
+            seq_lens = self.model_config.max_model_len if force_attention else 1
```
whx-sjtu: When `force_attention == False`, the dummy attn metadata is None meaning skipping attention part, so I don't quite understand why we need this.

Angazenn: The condition is `force_attention or aclgraph_runtime_mode == CUDAGraphMode.FULL`. When capturing, `force_attention = (aclgraph_runtime_mode == CUDAGraphMode.FULL)`. However, when not capturing, force_attention in `_dummy_run` is always `False`. So `_dummy_run` in actual inference executes dummy attention when using full graph mode.

yiz-liu (commented):
Please leave the error message when we remove the `max_model_len` completely. @Angazenn

Angazenn:
> Please leave the error message when we remove the `max_model_len` completely. @Angazenn

The bug before should be resolved. I think we can change to`max_query_len` here.

PR #4109, opened 2025-11-11
[Doc] Recover installation doc to use pip install

Yikun:
Use pip installation in installation doc and change related doctest to validate.

No, doc only

Doctest related CI passed
- vLLM version: v0.11.0


Yikun:
unit test is unrelated let's merge this first. @wangxiyuan

PR #4113, opened 2025-11-11
[Perf] fix async copy for async scheduling

realliujiaxu:
Only CPU tensors with `pin_memory=True` can be asynchronously copied to the device. Currently, there are two instances where non-pinned CPU tensors are being copied to the device, which will trigger synchronous operations, reducing the expected benefits of asynchronous scheduling.

I test this PR on Qwen3 8B.

### before 
TOPT 17.91  ms
image

### after
TOPT 17.02 ms
image

- vLLM version: v0.11.0


vllm_ascend/worker/block_table.py
```diff
@@ -94,7 +94,7 @@ def __init__(self,
         self.slot_mapping_cpu = torch.zeros(
             self.max_num_batched_tokens +
             2 * self.pcp_world_size * self.max_num_reqs,
-            dtype=torch.int64,
+            dtype=torch.int32,
```
Ronald1995, line 97: why change the dtype?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1849,11 +1849,10 @@ def _prepare_inputs(
                 logits_indices = torch.from_numpy(
                     cu_num_tokens
                 ) * self.pcp_size - self.num_pcp_pads[:num_reqs] - 1
-                logits_indices = logits_indices.to(self.device,
-                                                   non_blocking=True)
-            else:
-                logits_indices = torch.from_numpy(cu_num_tokens - 1).to(
+                logits_indices = logits_indices.pin_memory().to(
```
Ronald1995, line 1852: you call pin_memory in model_runner, but `pin_memory` has performance issue, i will be good to call it in init method, https://docs.pytorch.ac.cn/tutorials/intermediate/pinmem_nonblock.html#pin-memory. i don't acctually know why it synchronize here without `pin_memory`

PR #4117, opened 2025-11-11
[Bugfix][SHM] Use writer lock by default and remove redundant env

slippersss:

This PR aims to remove env introduced by #3988 and use lock by default. As described in https://github.com/vllm-project/vllm/issues/27858, we have tested the writer lock method in various scenarios and the performance is almost unaffected. Therefore, we believe that it would be safe to enable the lock by default and remove the redundant env `SHM_BARRIER` now.

After discussion, we decide to preserve env and set it as true by default.

`SHM_BARRIER` is set as true by default.

by ci

### DIFF
--- a/vllm_ascend/patch/platform/__init__.py
+++ b/vllm_ascend/patch/platform/__init__.py
@@ -17,12 +17,14 @@
 import os
 
 import vllm_ascend.patch.platform.patch_config  # noqa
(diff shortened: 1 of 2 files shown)


weijinqian0 (approved):

weijinqian0:
This will lead to port occupancy when test in customer's scenario.

PR #4119, opened 2025-11-11
[OPS] support triton causal_conv1d_fn ops

QilaiZhang:
Support triton causal_conv1d_fn ops.

No.

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/ops/triton/mamba/causal_conv1d.py
+++ b/vllm_ascend/ops/triton/mamba/causal_conv1d.py
@@ -7,7 +7,7 @@
 # and https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/mamba/ops/causal_conv1d.py
 # mypy: ignore-errors
 
-from typing import Optional, Union
+from typing import Any, Optional, Union
 
 import torch
 import triton
@@ -271,6 +271,7 @@ def causal_conv1d_fn(x: torch.Tensor,
                      has_initial_state: Optional[torch.Tensor] = None,
                      activation: Optional[str] = "silu",
                      pad_slot_id: int = PAD_SLOT_ID,
+                     metadata: Optional[Any] = None,
                      validate_data=False):
     """support varlen + continuous batching when x is 2D tensor
     x: (dim,cu_seq_len)


weijinqian0:
Some performance tests need to be conducted.

vllm_ascend/ops/triton/mamba/causal_conv1d.py
```diff
@@ -89,46 +302,143 @@ def causal_conv1d_fn(
...
+
+    # tracking BLOCK_M-based index in the sequence the Triton program is handling
+    max_blocks = seq_blocks.max().item() if len(seq_blocks) > 0 else 0
+    arange = torch.arange(max_blocks, device=x.device)
+    mask = arange.unsqueeze(0) < seq_blocks.unsqueeze(1)
+    token_chunk_offset_ptr = arange.repeat(len(seq_blocks),
+                                           1)[mask].to(torch.int32)
+
+    BLOCK_N = 256
+    grid = (total_seq_blocks, triton.cdiv(dim, BLOCK_N))
```
weijinqian0, line 398: This parameter may need to be tuned to achieve better performance.

QilaiZhang:
@weijinqian0 @wxsIcey Regarding parameter tuning: Since the `batch_ptr` and `token_chunk_offset_ptr` of this operator need to be calculated based on `BLOCK_M` and `BLOCK_N`, it is not suitable for `triton.autotune`. I manually tested the performance under different inputs and parameters using the msprof op tool, and the results are shown in the figure below.
triton-optimized

For an input length of 64, the optimal combination is (32, 256); for a length of 256, the optimal combination is (64, 256); and for a length of 8192, the optimal combination is (128, 512). However, when `BLOCK_M` is greater than or equal to 128, the compilation time is long and the kernel often hangs or crashes during testing, so this option is not considered.

Taking both performance and stability into account, I decided not to use branch statements to dynamically select parameters. Instead, the parameters are directly set to **(64, 256)**. This rationale is consistent with the implementation of this operator on NVIDIA GPUs, which also hardcodes its parameters to (8, 256). We can revisit this for further optimization in the future when the triton ascend is more stable.

QilaiZhang:
Regarding performance testing: Since the operator executes during the prefill stage, I set the concurrency to 40 with an input size of 64 and output size of 10 to better demonstrate the performance gains. The test script used is as follows.
```
python3 benchmarks/benchmark_serving.py --backend openai-chat --base-url ${BASE_URL} \
--endpoint=/v1/chat/completions --model /model/Qwen3-next-80B-A3B-Thinking/ --served-model-name Qwen3-next-80B-A3B-Thinking \
--dataset-name sonnet --dataset-path benchmarks/sonnet.txt --max-concurrency 40 \
--sonnet-input-len 64 --sonnet-output-len 10 --sonnet-prefix-len 0 --num-prompts 400 \
--ignore-eos --percentile-metrics "ttft,tpot,itl,e2el"
```
The results are summarized below:
- **PyTorch implementation**
微信图片_20251113104738

- **Triton kernel (before tuning) with (8, 256)**
微信图片_20251113104744

- **Triton kernel (after tuning) with (64, 256)**
微信图片_20251113104728

QilaiZhang:
@wxsIcey The current tests have passed. We are ready for the next test when convenient.

QilaiZhang:
Adapt MTP support, rebased and reopened PR

vllm_ascend/ops/triton/mamba/causal_conv1d.py
```diff
@@ -0,0 +1,848 @@
...
+    stride_state_indices = (conv_state_indices.stride(0)
+                            if conv_state_indices is not None else 0)
+    state_len = width - 1 + (seqlen - 1)  # effective state_len needed
+    np2_statelen = triton.next_power_of_2(state_len)
+
+    def grid(META):
+        return (
+            batch,
+            triton.cdiv(dim, META["BLOCK_N"]),
+        )
```
OsirisDuan, line 1285: The grid logic needs optimization.

QilaiZhang: @OsirisDuan 
Actually, I only added support for the triton `causal_conv1d_fn` this time, and did not modify the `causal_conv1d_update_npu` Triton kernel. The changes were solely due to fixing a typo in the filename `causal_conv1d.py`, which caused the entire file to appear as newly added in the diff. I sincerely apologize for the inconvenience this may have caused during your review.

If necessary, I would be more than happy to open a separate PR in the future to discuss and work on optimizing the `update` kernel.

vllm_ascend/ops/triton/mamba/causal_conv1d.py
```diff
@@ -0,0 +1,848 @@
...
+
+def causal_conv1d_fn(x: torch.Tensor,
+                     weight: torch.Tensor,
+                     bias: Union[torch.Tensor, None],
+                     conv_states: torch.Tensor,
+                     query_start_loc: torch.Tensor,
+                     cache_indices: Optional[torch.Tensor] = None,
+                     has_initial_state: Optional[torch.Tensor] = None,
+                     activation: Optional[str] = "silu",
+                     pad_slot_id: int = PAD_SLOT_ID,
```
jingshenghang, line 273: Here should add an optional `metadata` parameter:
```python
metadata: Optional[Any] = None,
```

Also need to add the imports at the top of the file:
```python
from typing import Any, Optional, Union
```

jingshenghang (commented):
metadata issue

wangxiyuan (approved):

PR #3579, opened 2025-10-21
[P/D] [Bugfix] mooncake connector support external dp & update readme

liziyu179:
### What this PR does / why we need it?

mooncake connector support external dp & update readme

### How was this patch tested?

- vLLM version: v0.11.0rc3
- vLLM main: https://github.com/vllm-project/vllm/commit/v0.11.0

PR #4125, opened 2025-11-11
[P/D] [Bugfix] mooncake connector device_ids out of index

LCAIZJ:
In https://github.com/vllm-project/vllm-ascend/pull/3579, the local DP rank was changed to the global DP rank, causing users to encounter an error when using the `export PHYSICAL_DEVICES=$(ls /dev/davinci* 2>/dev/null | grep -o '[0-9]\+' | sort -n | paste -sd',' -)`.
@liziyu179 

- vLLM version: v0.11.0


wangxiyuan:

PR #4152, opened 2025-11-12
[CI] Use eager mode to load weight

Potabk:
This path mainly do the following things:
1. Change the default mode of safetensors' loading, which using `safetensors_load_strategy="eager"` to speed up the loading term
2. Set the global env `TRANSFORMERS_OFFLINE: 1`, which will avoid downloading the file and return the path to the
local cached file if it exists when using modelscope's `snapshot_download` api
**Note**: for the developers: If you try to download files in the CI system for some new test cases, the default env `TRANSFORMERS_OFFLINE: 1` means  only the cached files will be accessed. If no cache file is detected, an error is raised This is useful in case your network is slow and you don’t care about having the latest version of a file. So, if you want to get some files from network, please set the env: `HF_HUB_OFFLINE` or `TRANSFORMERS_OFFLINE` as false to get the access of network

- vLLM version: v0.11.0

### DIFF
--- a/tests/e2e/nightly/multi_node/config/models/DeepSeek-V3.yaml
+++ b/tests/e2e/nightly/multi_node/config/models/DeepSeek-V3.yaml
@@ -19,6 +19,7 @@ env_common:
   OMP_NUM_THREADS: 100
   HCCL_BUFFSIZE: 1024
   SERVER_PORT: 8080
+  NUMEXPR_MAX_THREADS: 128
 disaggregated_prefill:
   enabled: true
   prefiller_host_index: [0]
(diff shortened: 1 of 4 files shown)


Potabk:
CI failed due to modelscope httperror 400, set to offline mode

Yikun:
seems e2e full not triggered...but not sure why:
https://github.com/vllm-project/vllm-ascend/blob/main/.github/workflows/vllm_ascend_test_full.yaml#L54

Potabk:
Have some tiny fix for nightly-ci, by the way, trigger the full test :)

wangxiyuan:

PR #4154, opened 2025-11-12
[perf][bugfix] improve performance of rejection sampler and eliminate HD synchronize in TopKTopPSampler

linfeng-yuan:
1. Use optimized apply_top_k_top_p for NPU platfrom in rejection sampler; (avoid scatter elements which can reduce ~26ms TPOT with bs=24 per DP)
2. <del>Avoid D2H Synchronization before calling npu_top_k_top_p introduced by parameter validation which improves inference speed with `async_scheduling` enabled;</del> In order to elminate the D2H synchronization introduced by parameter validation before calling `npu_top_k_top_p`, we directly drop this fused operator since the performance improvement is not significant compared to async_scheduling and may bring potential accuracy problem.
3. Refactor the implementation of AscendTopKTopPSampler to align that of vLLM.

No.
E2E serving test with combinations of `k=500` and `p=0.95` with async_scheduling in single node and wide-EP scenarios.

- vLLM version: v0.11.0

### DIFF
--- a/.github/workflows/_e2e_nightly_single_node.yaml
+++ b/.github/workflows/_e2e_nightly_single_node.yaml
@@ -148,5 +148,10 @@ jobs:
         run: |
           # ignore test_dispatch_ffn_combine until the test is fixed
           pytest -sv ${{ inputs.tests }} \
-          --ignore=tests/e2e/nightly/ops/test_dispatch_ffn_combine.py 
+          --ignore=tests/e2e/nightly/ops/test_dispatch_ffn_combine.py \
+          --ignore=tests/e2e/nightly/ops/test_fused_moe.py \
+          --ignore=tests/e2e/nightly/ops/test_rotary_embedding.py \
+          --ignore=tests/e2e/nightly/ops/test_matmul_allreduce_add_rmsnorm.py
+
+
 
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -117,6 +117,7 @@ jobs:
           pytest -sv --durations=0 tests/e2e/singlecard/pooling/
           pytest -sv --durations=0 tests/e2e/singlecard/compile/test_norm_quant_fusion.py
           pytest -sv --durations=0 tests/e2e/singlecard/test_cross_layer_attn_model.py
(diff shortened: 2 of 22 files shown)


vllm_ascend/sample/sampler.py
```diff
@@ -130,3 +98,54 @@ def forward_native(self, logits, generators, k, p):
...
+    k: torch.Tensor,
+    p: torch.Tensor,
+) -> torch.Tensor:
+    """Use torch_npu.npu_top_k_top_p if possible."""
+    if p is None and k is None:
+        return logits
+    # npu_top_k_top_p requires parameter k ranged from 1 to 1024 if given
+    if k is None or 1 <= int(k.max()) <= 1024:
+        # npu_top_k_top_p's parameter order is (logits, p, k), not (logits, k, p)
+        return torch_npu.npu_top_k_top_p(logits, p, k)
```
realliujiaxu: in the original rejection sampler, there is `top_p.to(logits.dtype)`, please make sure removing `.to(logits.dtype)` is ok.

vllm_ascend/sample/sampler.py
```diff
@@ -65,59 +66,26 @@ def do_async_exponential(self, b_s, head_dim, generators):
...
 class AscendTopKTopPSampler(TopKTopPSampler):
 
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        is_async_scheduling = get_current_vllm_config(
+        ).scheduler_config.async_scheduling
+        is_ascend_310p = get_ascend_device_type() == AscendDeviceType._310P
+        # Currently, we do not call fused top_k_top_p op when async scheduling is enabled
+        # since it must cause D2H synchronize according to operation constraints.
+        self.use_fused_top_k_top_p_op = not (is_async_scheduling
```
realliujiaxu: maybe a temp variable `use_fused_top_k_top_p_op` is enough.

linfeng-yuan: Done.

linfeng-yuan:
All the E2E CI passed (https://github.com/vllm-project/vllm-ascend/actions/runs/20461117976/job/58793951915?pr=4154). I would remove `npu_top_k_top_p` op calling in all cases

PR #4155, opened 2025-11-12
[Feat] Encoder separation for Encode-Prefill-Decode Disaggregation

jesse996:
Disaggregated Encoder
A disaggregated encoder runs the vision-encoder stage of a multimodal LLM in a process that is separate from the prefill / decoder stage. Deploying these two stages in independent vLLM instances brings three practical benefits:

 - Independent, fine-grained scaling
 - Lower time-to-first-token (TTFT)
 - Cross-process reuse and caching of encoder outputs

- vLLM version: v0.11.0


jesse996:
Will reopen later

PR #4162, opened 2025-11-13
[Feat] Adds a utility for printing from within ACL graphs

yiz-liu:
Introduces the `acl_graph_print` function to enable printing debug information from code running inside an ACL graph, such as custom operators.

This works by launching a host function on a dedicated stream, bypassing the limitations of standard `print` within compiled graph execution. The implementation handles the necessary stream subscriptions and ensures they are properly unregistered upon exit.

None.

None.

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -60,6 +60,65 @@ _IS_MOE_MODEL = None
 _ENABLE_SP = None
 _HAS_LAYER_IDX = None
 _ENABLE_NZ = None
+_SUBSCRIBED_COMPUTE_STREAMS = set()
+_GRAPH_PRINT_STREAM = None
+_GRAPH_PRINT_STREAM_LOCK = Lock()
+
+
+def _print_callback_on_stream(*args):
+    """Callback function to print arguments on the dedicated print stream."""
+    global _GRAPH_PRINT_STREAM
+    with torch_npu.npu.stream(_GRAPH_PRINT_STREAM):
+        print(*args, flush=True)
+
+
+def acl_graph_print(*args):
+    """
+    Prints arguments from within an ACL graph.
+
+    This function is provided for developers to print debug information when encountering
+    issues within an ACL graph, pretty handy for dumping input/output tensor values, or
+    resolving unexpected hangs. Usage:
+    ```python
+    from vllm_ascend.utils import acl_graph_print
+    ...
+    acl_graph_print("Debug info")
+    ```
+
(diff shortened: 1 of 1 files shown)


yiz-liu:
@wangxiyuan This is pretty important, could you please review this ASAP?

vllm_ascend/utils.py
```diff
@@ -60,6 +60,53 @@
...
+    global _GRAPH_PRINT_STREAM
+    with torch_npu.npu.stream(_GRAPH_PRINT_STREAM):
+        print(*args, flush=True)
+
+
+def acl_graph_print(*args):
+    """
+    Prints arguments from within an ACL graph.
+
+    This function launches a host function on the current compute stream to print
```
wangxiyuan, line 88: add more note here like:
```
this function is provided for developers who want to debug code when runing vLLM Ascend under aclgraph mode. 
You can call this function at the code place where you want to dig into. For example, xxxxxxx
```

vllm_ascend/utils.py
```diff
@@ -60,6 +60,53 @@
...
+        _SUBSCRIBED_COMPUTE_STREAMS.append(current_compute_stream)
+
+    torch_npu.npu._launch_host_func(current_compute_stream,
+                                    _print_callback_on_stream, args)
+
+
+def _unregister_print_streams_on_exit():
+    """Unsubscribe all compute streams used for printing at exit."""
+    global _SUBSCRIBED_COMPUTE_STREAMS
+    for stream in _SUBSCRIBED_COMPUTE_STREAMS:
```
wangxiyuan:
```suggestion
    if _SUBSCRIBED_COMPUTE_STREAMS is not None:
        for stream in _SUBSCRIBED_COMPUTE_STREAMS:
```

vllm_ascend/utils.py
yiz-liu: Sure.

vllm_ascend/utils.py
yiz-liu: No need.

vllm_ascend/utils.py
```diff
@@ -60,6 +60,53 @@
...
+        torch_npu.npu._unsubscribe_report(stream)
+
+
+atexit.register(_unregister_print_streams_on_exit)
```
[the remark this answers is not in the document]
yiz-liu: Good catch.

wangxiyuan (approved):
Thanks for this helpful improvement

PR #4165, opened 2025-11-13
qwen3-vl Vit module enable sp

Edkamoni:
Enable Qwen3-VL vit sp parallel and mrope npu fusion op

No

Test Qwen3-VL 30B model accuracy on textVQA with aisbench

- vLLM version: v0.11.2


wangxiyuan:

PR #4168, opened 2025-11-13
[Fusion] Adopt inductor fusion and define quantization fusion pass

wxsIcey:

Part of: https://github.com/vllm-project/vllm-ascend/issues/4239

The main goal of this PR to alleviate the high maintenance burden from model duplication when we are going to do the model optimization. Some of our optimized models diverges a little from the vllm's modeling, but needs to rewrite several part of original one, brings negligible maintenance bruden to the vllm-ascend.In order to solve that, we propose to leverage `torch.compile` and `inductor pattern matcher`, automatically fuse the pattern we want to merge. For more details can refer to the RFC https://github.com/vllm-project/vllm-ascend/issues/4239

This pr integrates `AddRMSNorm` and the `Quant` operator, which can improve the inference speed of models using `w8a8 `quantization.

Performance improvement results:
image

image

Yes, add new additional_config

```python
def main():
    prompts = [
        "The president of the United States is Mr.",
    ]

    sampling_params = SamplingParams(max_tokens=100, temperature=0.6, top_k=40, top_p=0.95)
    # Create an LLM.
    llm = LLM(
        model="/root/.cache/modelscope/hub/models/vllm-ascend/Qwen3-8B-W8A8",
              # enforce_eager=True,
              tensor_parallel_size=1,
              trust_remote_code=True,
              gpu_memory_utilization=0.7,
              quantization="ascend",
              )

    for output in outputs:
        prompt = output.prompt
```

```text
Prompt: 'The president of the United States is Mr.', Generated text: ' Trump. The president of the United States is Mr. Biden. Which of the following statements is correct? \n\nA. Mr. Trump is Mr. Biden.  \nB. Mr. Trump is not Mr. Biden.  \nC. The president of the United States is not Mr. Trump.  \nD. The president of the United States is not Mr. Biden.\n\nThe question presents a contradiction: it states that "The president of the United States is Mr. Trump" and "The president of'
```

Co-authored-by: @ganyi1996ppo insipred by https://github.com/vllm-project/vllm-ascend/pull/2389

- vLLM version: 86e178f7c4d8c3b0eaf3c8e3f810a83f63b90e24
- vLLM main: https://github.com/vllm-project/vllm/commit/86e178f7c4d8c3b0eaf3c8e3f810a83f63b90e24

### DIFF
--- a/tests/e2e/singlecard/test_quant_fusion.py
+++ b/tests/e2e/singlecard/test_quant_fusion.py
@@ -186,7 +186,8 @@ class TestBackend:
 @pytest.mark.parametrize("hidden_size", [64])
 @pytest.mark.parametrize("num_tokens", [257])
 @pytest.mark.parametrize("eps", [1e-5, 1e-6])
-def test_rmsnorm_quant_fusion(dtype, hidden_size, num_tokens, eps):
+def test_rmsnorm_quant_fusion(dtype: torch.dtype, hidden_size: int,
+                              num_tokens: int, eps: float):
     """
     End-to-end test for AddRMSNorm+Quantize fusion.
     Compares: Operator presence/absence before and after graph transformation
--- a/vllm_ascend/compilation/passes/quant_fusion_pass.py
+++ b/vllm_ascend/compilation/passes/quant_fusion_pass.py
@@ -21,11 +21,12 @@ import torch
 import torch._inductor.pattern_matcher as pm
 from torch._inductor.pattern_matcher import PatternMatcherPass
 from vllm.compilation.vllm_inductor_pass import VllmInductorPass
+from vllm.config import VllmConfig
 
 
 class AddRMSNormQuantPattern:
 
-    def __init__(self, vllm_config, eps=1e-6):
+    def __init__(self, vllm_config: VllmConfig, eps: float = 1e-6):
         self.vllm_config = vllm_config
         self.eps = eps
 
@@ -83,7 +84,7 @@ class AddRMSNormQuantFusionPass(VllmInductorPass):
     A pass for fusing AddRMSNorm and W8A8 quantization operations on Ascend.
     """
 
-    def __init__(self, vllm_config):
+    def __init__(self, vllm_config: VllmConfig):
         super().__init__(vllm_config)
         self.pattern_match_passes: PatternMatcherPass = PatternMatcherPass(
             pass_name="rmsnorm_quant_fusion_pass")
@@ -105,7 +106,7 @@ class AddRMSNormQuantFusionPass(VllmInductorPass):
         logging.debug("Replaced %s patterns", self.matched_count)
         self.end_and_log()
 
-    def is_applicable(self, runtime_shape):
+    def is_applicable(self, runtime_shape: int | None = None) -> bool:
         """
         Check if the pass is applicable for the current configuration.
         """
--- a/tests/ut/test_ascend_config.py
+++ b/tests/ut/test_ascend_config.py
@@ -59,9 +59,6 @@ class TestAscendConfig(TestBase):
         ascend_compilation_config = ascend_config.ascend_compilation_config
         self.assertTrue(ascend_compilation_config.enable_quantization_fusion)
 
-        ascend_scheduler_config = ascend_config.ascend_scheduler_config
-        self.assertFalse(ascend_scheduler_config.enabled)
-
     @_clean_up_ascend_config
     def test_init_ascend_config_with_additional_config(self):
         test_vllm_config = VllmConfig()


wxsIcey:
Although the fusion of rmsnorm and quantizate has been achieved, there are still accuracy issues that need to be addressed.

wxsIcey:
Currently, operator fusion has been achieved through pattern matching using inductors. But it has been found that using aot-autograd causes accuracy issues. @whx-sjtu Would you be willing to review it?

vllm_ascend/compilation/compiler_interface.py
```diff
@@ -0,0 +1,72 @@
...
+    Extract the shapes from the kwargs dictionary.
+    """
+    shape_list = []
+    for value in args:
+        if isinstance(value, torch.Tensor):
+            shape_list.append(value.shape)
+    return shape_list
+
+
+class AscendAdaptor(CompilerInterface):
```
whx-sjtu: The name `AscendAdaptor` is too vague; I suggest a more specific one like `AscendCompiler`.

vllm_ascend/compilation/quant_fusion_pass.py
```diff
@@ -0,0 +1,104 @@
...
+        return [rms_norm_input, residual, rms_norm_weight, scale, offset]
+
+    def register(self, pm_pass: PatternMatcherPass):
+
+        def pattern(rms_norm_input, residual, rms_norm_weight, scale, offset):
+            """
+          Pattern for AddRMSNormQuant fusion.
+          """
+            output = torch.ops.npu.npu_add_rms_norm(rms_norm_input, residual,
+                                                    rms_norm_weight, 1e-6)
```
whx-sjtu: Instead of  fixed to `1e-6`, the `eps` should be defined as a static variable of `AddRMSNormQuantPattern`, with different values of eps corresponding to different pattern objects. Some models might use different eps like `1e-5`.

vllm_ascend/compilation/quant_fusion_pass.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class AscendQuantFusionPass(VllmInductorPass):
+    """
+    A pass for fusing AddRMSNorm and W8A8 quantization operations on Ascend.
+    """
+
+    def __init__(self, vllm_config):
+        super().__init__(vllm_config)
+        self.patterns: PatternMatcherPass = PatternMatcherPass(
```
whx-sjtu: The name of `self.patterns` is a bit confusing here. It should be named as something like `self.pattern_match_pass`.

vllm_ascend/compilation/quant_fusion_pass.py
```diff
@@ -0,0 +1,104 @@
...
+        arg_dtypes = kwargs.get("arg_dtypes", None)
+        if arg_dtypes is None:
+            return False
+        # We assume the first tensor's dtype is the data type of this model, update this solution when there is
+        # better solution.
+        dtype = arg_dtypes[0] if isinstance(
+            arg_dtypes, list) and len(arg_dtypes) > 0 else arg_dtypes
+        # We found that the kernel npu_add_rms_norm_quant accept varying data format for different dtypes, therefore, we only
+        # provide the solution on bfloat16 here.
+        return dtype in (torch.bfloat16, )
```
whx-sjtu: I don't quiet understand here. Does the format of data also influence pattern matching? Maybe we can define patterns separately for bf16 and fp16 to support them both?

whx-sjtu:
This feature is very important for vllm-ascend. I also hope @jgong5 can take some time to review this PR. Thanks.

wxsIcey:
> I have another question here. With current proposal can we reuse the ready-made fusion passes defined in vLLM, like the SequenceParallel Fusion Pass. Because I'm not very familiar with the stack of the current Fusion pass in vLLM, I'm confirming it here. Reusability is what we expect.

Thank you for your reply. The current PR aims to define our own compiler backend to implement custom fusion. Reusing fusion passes in VLLM is my next goal. I will submit an RFC once the solution is finalized.

vllm_ascend/ascend_config.py
```diff
@@ -128,6 +133,22 @@ def __init__(self, vllm_config):
                     "Only support P node tp size lagger then D node tp size")
 
 
+class AscendCompilationConfig:
+    """
+    Configuration Object for ascend_compilation_config from additional_config
```
jgong5: This comment doesn't bring extra info about this class. In fact, we can get that from the class name. If you want to explain anything meaningful here, you can consider to add why we need this configuration here and what are the rules to add more configurations under it etc.

vllm_ascend/ascend_config.py
```diff
@@ -128,6 +133,22 @@ def __init__(self, vllm_config):
...
+    """
+
+    def __init__(self,
+                 enable_graph_fusion: bool = True,
+                 fx_graph_eager: bool = False,
+                 enable_quantization_fusion: bool = True,
+                 **kwargs):
+        self.enable_graph_fusion = enable_graph_fusion
+        self.fx_graph_eager = fx_graph_eager
+        self.enable_quantization_fusion = enable_quantization_fusion
```
jgong5: Add the meaning as the code doc for each field.

vllm_ascend/ascend_config.py
```diff
@@ -279,6 +300,17 @@ def check_ascend_config(vllm_config, enforce_eager):
...
+            # This graph fusion can actually works on eager mode.
+            if ascend_config.ascend_compilation_config.enable_graph_fusion:
+                logger.info(
+                    "graph fusion enabled! Automatic kernel fusion is expected."
+                )
+
+                if ascend_config.ascend_compilation_config.enable_quantization_fusion:
+                    logger.info(
+                        "Quantization fusion enabled! op fusion on quantization are expected. "
+                    )
```
jgong5: Take care of your grammar.

vllm_ascend/ops/layernorm.py
```diff
@@ -63,49 +30,23 @@ def forward_oot(
...
+            if is_310p():
+                orig_dtype = residual.dtype
+                x = x + residual.to(x.dtype)
+                residual = x.to(orig_dtype)
+                x, _ = torch_npu.npu_rms_norm(x, self.weight,
+                                              self.variance_epsilon)
+            else:
+                x, _, residual = torch_npu.npu_add_rms_norm(
+                    x, residual, self.weight, self.variance_epsilon)
             return x, residual
```
jgong5: I don't quite follow the logic here. Why do we need such a check here?

vllm_ascend/compilation/quant_fusion_pass.py
jgong5: Right, we usually don't decide the application of graph passes based on the concrete input. If we really have to do so, we have to add "guards" to make sure that the graph is recompiled when the input changes.

vllm_ascend/compilation/compiler_interface.py
```diff
@@ -0,0 +1,72 @@
...
+    name = "AscendAdaptor"
+
+    def compile(
+        self,
+        graph: fx.GraphModule,
```
jgong5, line 52: Is the graph processed by AoT dispatcher before being passed here to the compiler backend?

vllm_ascend/compilation/passes/quant_fusion_pass.py
```diff
@@ -0,0 +1,104 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import torch
+import torch._inductor.pattern_matcher as pm
+from torch._inductor.pattern_matcher import PatternMatcherPass
+from vllm.compilation.vllm_inductor_pass import VllmInductorPass
+
+
+class AddRMSNormQuantPattern:
```
ApsarasX, line 27: Can we add a directory called `passes` or `fx_masses` specifically to store these passes?

vllm_ascend/platform.py
```diff
@@ -59,6 +59,19 @@ class NPUPlatform(Platform):
 
     def is_sleep_mode_available(self) -> bool:
         return True
+    
+    @property
+    def pass_key(self) -> str:
+        return "graph_fusion_manager"
+    
+    @classmethod
+    def get_pass_manager_cls(cls) -> str:
```
ApsarasX, line 79: Does this interface have any requirements for the vllm version?

vllm_ascend/platform.py
```diff
@@ -59,6 +59,19 @@ class NPUPlatform(Platform):
...
+        return "vllm_ascend.compilation.graph_fusion_pass_manager.GraphFusionPassManager"
+    
+    @classmethod
+    def get_compile_backend(self) -> str:
```
ApsarasX, line 87: ditto

vllm_ascend/ascend_config.py
wxsIcey: Thanks. I have added it.

vllm_ascend/compilation/compiler_interface.py
wxsIcey: Yes, I used aot-autograd.

vllm_ascend/compilation/passes/quant_fusion_pass.py
wxsIcey: Of course, I've already added it.

vllm_ascend/platform.py
wxsIcey: I'm trying to understand what you mean. We're defining our own pass manager and compiler backend here, which should be independent of the vllm version.

vllm_ascend/compilation/quant_fusion_pass.py
wxsIcey: Thank you for your suggestion. I have revised it.

vllm_ascend/compilation/quant_fusion_pass.py
wxsIcey: done

vllm_ascend/compilation/quant_fusion_pass.py
wxsIcey: Thanks. I have removed this judgment. Currently, the fusion operator supports float16 and bfloat16, so no special processing is required.

vllm_ascend/ascend_config.py
wxsIcey: done

vllm_ascend/ascend_config.py
wxsIcey: done

vllm_ascend/platform.py
wxsIcey: Please see the explanation above.

vllm_ascend/ops/layernorm.py
wxsIcey: The check on 310p is to maintain the original logic, see https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/ops/layernorm.py#L71. But I do not know why the 310p needs special processing.

wxsIcey:
The operators have been correctly fused, and the functionality and accuracy are normal. Could you please take another look? @whx-sjtu @jgong5

vllm_ascend/platform.py
wangxiyuan: vllm 0.12.0 and later.

vllm_ascend/platform.py
wangxiyuan: ditto

vllm_ascend/platform.py
```diff
@@ -61,6 +61,19 @@ class NPUPlatform(Platform):
...
+        return "graph_fusion_manager"
+
+    @classmethod
+    def get_pass_manager_cls(cls) -> str:
+        return "vllm_ascend.compilation.graph_fusion_pass_manager.GraphFusionPassManager"
+
+    @classmethod
+    def get_compile_backend(self) -> str:
+        from vllm_ascend.compilation.compiler_interface import AscendAdaptor
+        return AscendAdaptor.__module__ + "." + AscendAdaptor.__name__
```
wangxiyuan: use string instead like others to make the coe more clear

vllm_ascend/compilation/compiler_interface.py
wxsIcey: I have changed to `AscendCompiler`, it's definitely a better fit.

tests/e2e/singlecard/test_quant_fusion.py
```diff
@@ -0,0 +1,219 @@
+#
```
wangxiyuan, line 1: this should be added to .github workflow to enable test by CI

wxsIcey: Thanks. I think this pr can be merged first, I will enable it in next fusion pr.

vllm_ascend/platform.py
wxsIcey: Thanks. I will change it in next pr.

paulyu12:
This PR seems break LoRA e2e test. The error log is here: https://github.com/vllm-project/vllm-ascend/actions/runs/20450707034/job/58763094223?pr=4075

Does anybody know why this occurs or give any advice?

Reproduce:
- vllm v0.12.0 + vllm-ascend commit 178ca1607e77454297bea8193cd1e44aff012663, run "pytest -sv tests/e2e/singlecard/test_ilama_lora.py" can reproduce the bug.
- but vllm-ascend commit c4a71fc6d56af0ae9d8e84bbeb91092790162ae0 works well. (just 1 commit ahead of that)

PR #4170, opened 2025-11-13
[Bugfix] fix cannot import name get_mp_context

weiguihua2:
fix bug cannot import name get_mp_context

- vLLM version: v0.11.0

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,13 +68,6 @@ repos:
   #   language: system
   #   types: [python]
   #   stages: [pre-commit] # Don't run in CI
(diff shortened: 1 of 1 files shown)


MengqingCao:
Thanks for this fix, could you help also fix this https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/distributed/mooncake/mooncake_store_connector_v1.py#L12, plz refer to https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/distributed/cpu_offload_manager/metadata.py#L19-L22

wangxiyuan:
add some check like others `vllm_version_is`

PR #2917, opened 2025-09-14
[New model] Qwen3-next support

wangxiyuan:
Add qwen3-next support

PR #4221, opened 2025-11-17
[bugfix] pcp + mtp acl graph bugfix

zhangsicheng5:
Fix pcp + mtp bug while using acl graph.
While using pcp + mtp, we need to flatten block_table to avoid irregular attn mask shape, this was done in mla attn_metadata builder, but we found out that this influences block_table address and leads to incorrect results while enable acl graph.
To fix this, we enlarge block_table buffer size and flatten block_table in model_runner prepare_inputs, so this will not influence block_table address.

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/worker/block_table.py
+++ b/vllm_ascend/worker/block_table.py
@@ -27,13 +27,29 @@ class BlockTable:
                  pin_memory: bool,
                  device: torch.device,
                  kernel_sizes: Union[list[int], None] = None,
-                 cp_kv_cache_interleave_size: int = 1):
+                 cp_kv_cache_interleave_size: int = 1,
+                 num_speculative_tokens: int = 0):
         self.max_num_reqs = max_num_reqs
         self.max_num_blocks_per_req = max_num_blocks_per_req
         self.max_num_batched_tokens = max_num_batched_tokens
         self.pin_memory = pin_memory
         self.device = device
         self.physical_block_size = block_size
+
+        try:
+            self.pcp_world_size = get_pcp_group(
+            ).world_size if prefill_context_parallel_enable() else 1
+            self.pcp_rank = get_pcp_group(
+            ).rank_in_group if self.pcp_world_size > 1 else 0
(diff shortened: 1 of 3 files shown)


vllm_ascend/worker/block_table.py
```diff
@@ -27,13 +27,29 @@ def __init__(self,
...
         self.physical_block_size = block_size
+
+        try:
+            self.pcp_world_size = get_pcp_group(
+            ).world_size if prefill_context_parallel_enable() else 1
+            self.pcp_rank = get_pcp_group(
+            ).rank_in_group if self.pcp_world_size > 1 else 0
+            self.dcp_world_size = get_dcp_group().world_size
+            self.dcp_rank = get_dcp_group().rank_in_group
+        except AssertionError:
```
wangxiyuan, line 46: do not `expect` error for test case only. It's really hack and ugly.

zhangsicheng5: This pr just moved pcp/dcp related initialization from line 86 to line 39, since we need to use pcp_world_size before the original initialization place.
This try-except pattern is inherited from the original vllm-ascend block_table implementation from this pr: [[New model] Qwen3-next support](https://github.com/vllm-project/vllm-ascend/pull/2917), and vllm community also use the same implementation (refer to [block_table.py](https://github.com/vllm-project/vllm/blob/main/vllm/v1/worker/block_table.py)).
Since in some test case pcp/dcp might not be initialized, and there's no better way to know whether pcp/dcp is set, so we also agree that it's the best way to except an error here and set cp params properly if there's no cp group.

yiz-liu (approved):

vllm_ascend/worker/block_table.py
wangxiyuan: Yes, I notice that there are a lot `AssertionError` usage in vllm-ascend. Let's find a better way to deal this kind of problem in the future.

PR #4223, opened 2025-11-17
[Bugfix] fix nightly multi-node EPLB tests' "DYNAMIC_EPLB=true" environment not working

845473182:
fix nightly multi-node EPLB tests by adjusting vllm_ascend\eplb\core\eplb_utils.py dynamic_eplb gate checking
no

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/patch/platform/__init__.py
+++ b/vllm_ascend/patch/platform/__init__.py
@@ -16,11 +16,12 @@
 
 import os
 
+import vllm_ascend.envs as envs_ascend
 import vllm_ascend.patch.platform.patch_config  # noqa
 import vllm_ascend.patch.platform.patch_distributed  # noqa
 import vllm_ascend.patch.platform.patch_mamba_config  # noqa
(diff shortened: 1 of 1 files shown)


Potabk:
maybe we could refer to https://github.com/vllm-project/vllm-ascend/blob/e38ef2c4344e11130a8fae5260b9dad6962d1fd1/vllm_ascend/envs.py#L104, Create a lambda which perform some preprocessing on environment variables and return a callable object. and maintaining the current PR unchanged as a temporary solution is acceptable.

845473182:
> maybe we could refer to
> 
> https://github.com/vllm-project/vllm-ascend/blob/e38ef2c4344e11130a8fae5260b9dad6962d1fd1/vllm_ascend/envs.py#L104
> 
> , Create a lambda which perform some preprocessing on environment variables and return a callable object. and maintaining the current PR unchanged as a temporary solution is acceptable.

done

PR #4224, opened 2025-11-17
[Refactor] remove moe type of multicast.

weijinqian0:
The main purposes of this PR are as follows: 
1. Remove the multicast-related code; 

Reason:
1. In the scenario like a2 Dual-System Back-to-Back Networking，the performance is worse than all_gather.  Before the modification, in e2e test, it was 3 tps; after the modification, it is 10 tps.
2. At the same time, we usually enable the SP feature，it is consistent with the current logic.
3. The advantage of broadcast communication lies in the fact that it does not suffer from uneven DP load and does not require the prefill ACL graph to be enabled. But we support prefill Acl graph recently.

So we think there is no need to maintain the multicast as one choice in moe communication.

Performance benefits are as follows:
When not enable_flashcomm1, TTFT remains relatively stable at around 43000ms, which is approximately 15000ms faster than before the modification.

When enable_flashcomm1, there is no diffenence, TTFT remains relatively stable at around 29000ms.

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/spec_decode/mtp_proposer.py
+++ b/vllm_ascend/spec_decode/mtp_proposer.py
@@ -688,8 +688,7 @@ class MtpProposer(Proposer):
          with_prefill) = self.runner._sync_metadata_across_dp(
              num_input_tokens, self.runner.with_prefill)
 
-        moe_comm_type = self.runner._select_moe_comm_method(
-            num_input_tokens)
+        moe_comm_type = self.runner._select_moe_comm_method(num_input_tokens)
 
         if scheduler_output:
             max_query_len = common_attn_metadata.max_query_len


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2351,12 +2351,6 @@ def _select_moe_comm_method(self, num_tokens: int,
         else:
             raise ValueError(f"Unsupported soc_version: {soc_version}")
 
-        if moe_comm_type == MoECommType.ALLGATHER and with_prefill:
```
Pr0Wh1teGivee, line 2388: The parameter "with_prefill" is useless after removing multicast

vllm_ascend/torchair/torchair_fused_moe_state.py
```diff
@@ -0,0 +1,31 @@
...
+                        is_deepseek_v3_r1: bool):
+    # the fusion operator torch_npu.npu_grouped_matmul_finalize_routing called by allgather ep
+    # only supports deepseek v3/r1
+    if (envs_ascend.VLLM_ENABLE_FUSED_EXPERTS_ALLGATHER_EP and ep_size > 1
+            and is_deepseek_v3_r1):
+        return FusedMoEState.AllGatherEP
+    elif ep_size == 1:
+        if with_prefill:
+            return FusedMoEState.NaiveMulticast
+        else:
```
Ronald1995: it's good to remove the else statement.

vllm_ascend/torchair/torchair_fused_moe_state.py
```diff
@@ -0,0 +1,31 @@
...
+            return FusedMoEState.AllGather
+    # NOTE: mc2 need ep_size >= 16 & all2all can't use in torchair graph.
+    elif ep_size < 16 or with_prefill:
+        return FusedMoEState.All2All
+    else:
```
Ronald1995: it's good to remove else statement.

vllm_ascend/ascend_forward_context.py
```diff
@@ -11,6 +11,7 @@
                                   set_forward_context)
 
 import vllm_ascend.envs as envs_ascend
+from vllm_ascend.torchair.torchair_fused_moe_state import get_fused_moe_state
```
wangxiyuan: avoid to import torchair outside torchair module

wangxiyuan (changes requested):
please refresh the commit message to describe why the `NaiveMulticastCommImpl` is removed

realliujiaxu:
Please provide a performance comparison before and after refactoring in a standalone scenario, for example:

```
REQ_RATE=4
MAX_CON=10
NUM_PROMPT=200
vllm bench serve \
--backend openai-chat \
--dataset-name random \
--trust-remote-code \
--model /path/to/DeepSeek-R1-0528-W8A8 \
--served-model-name auto \
--random-input-len 4096 \
--random-output-len 1536 \
--num-prompts ${NUM_PROMPT} \
--max-concurrency ${MAX_CON} \
--request-rate ${REQ_RATE} \
--ignore-eos \
--metric-percentiles "50,90,99" \
--base-url http://localhost:8009 \
--endpoint /v1/chat/completions
```

vllm_ascend/ascend_forward_context.py
weijinqian0: ok

realliujiaxu (approved):

Pr0Wh1teGivee:
approve

MengqingCao:
Do we still need `FusedMoEState.NaiveMulticast` now? if not, let's also remove it in this pr. https://github.com/vllm-project/vllm-ascend/blob/97999347c8e89e4d94c5ae2ce9cc803327e3d163/vllm_ascend/ascend_forward_context.py#L28

weijinqian0:
> Do we still need `FusedMoEState.NaiveMulticast` now? if not, let's also remove it in this pr.
> 
> https://github.com/vllm-project/vllm-ascend/blob/97999347c8e89e4d94c5ae2ce9cc803327e3d163/vllm_ascend/ascend_forward_context.py#L28

This variable is used in torchair.

MengqingCao:
> > Do we still need `FusedMoEState.NaiveMulticast` now? if not, let's also remove it in this pr.
> > https://github.com/vllm-project/vllm-ascend/blob/97999347c8e89e4d94c5ae2ce9cc803327e3d163/vllm_ascend/ascend_forward_context.py#L28
> 
> This variable is used in torchair.

Okay, then lgtm

MengqingCao (approved):

weijinqian0:
> please refresh the commit message to describe why the `NaiveMulticastCommImpl` is removed

It is already supplemented

wangxiyuan (approved):

PR #4230, opened 2025-11-17
fix: sorts aclgraph batch sizes in ascending order

yiz-liu:
Sorts aclgraph batch sizes in ascending order, corresponding to vLLM [#26016](https://github.com/vllm-project/vllm/pull/26016)

Ensures batch sizes for aclgraph are sorted ascending when aclgraph mode is enabled, improving consistency and compatibility with later logic that may depend on order.

None.

Waiting for #3886 

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -521,8 +521,12 @@ class NPUModelRunner(LoRAModelRunnerMixin):
             if self.speculative_config else 0)
 
         self.use_aclgraph = self._use_aclgraph()
-        self.aclgraph_batch_sizes = list(
-            reversed(self.compilation_config.cudagraph_capture_sizes))
+
+        # self.aclgraph_batch_sizes sorts in ascending order.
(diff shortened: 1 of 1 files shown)


yiz-liu:
@MengqingCao Please take a look at this.

MengqingCao:
LGTM, thanks for this fixing!
BTW, I thought of a question unrelated to this PR. I believe that since you added the `support_static_graph_mode` API to `platform`, `_set_cudagraph_sizes` will be called by default logic during the post-init process of `VllmConfig`, so there's no need to add this logic to our `NPUPlatform`, right? If so, we could remove it in another pr

yiz-liu:
> LGTM, thanks for this fixing! BTW, I thought of a question unrelated to this PR. I believe that since you added the `support_static_graph_mode` API to `platform`, `_set_cudagraph_sizes` will be called by default logic during the post-init process of `VllmConfig`, so there's no need to add this logic to our `NPUPlatform`, right? If so, we could remove it in another pr

I'll take a look at it later.

PR #4248, opened 2025-11-18
[Bugfix] Fix out-of-bounds access to token_id due to uninitialized logprobs

ghphotoframe:
The logprobs_tensor was not initialized before accessing its token_id
member, leading to a crash when tokenizer.decode() is called by passing a  negative token_id
No
Constructed an inference request with two prompts and set SamplingParams(prompt_logprobs=<non-None value>) (e.g., prompt_logprobs=1).
After applying the fix (proper initialization of logprobs_tensor), the same request completed successfully without errors, and the returned logprobs matched expected values.

- vLLM version: v0.12.0


MengqingCao (approved):

MengqingCao:
Thanks for this fixing!

PR #4254, opened 2025-11-18
[Bugfix] Resolve MTP > 1 issue when lm head tp > 1

jianzs:

Previously, the dummy run executed compute_logits only once, regardless of num_speculative_tokens. This caused execute_model to hang on compute_logits when lm head tensor parallelism exceeded 1. The fix ensures compute_logits executes correctly during dummy run, matching num_speculative_tokens.

I set the `non_blocking` argument to False when moving `exceeds_max_model_len` to the CPU. From what I understand, using `non_blocking=True` and immediately accessing the tensor on the CPU can cause accuracy problems. However, this issue doesn't happen when transferring data to a device. ref: https://discuss.pytorch.org/t/should-we-set-non-blocking-to-true/38234/18

No.

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/spec_decode/ngram_proposer.py
+++ b/vllm_ascend/spec_decode/ngram_proposer.py
@@ -26,7 +26,8 @@ class NgramProposer(VllmNgramProposer, Proposer):
                   num_reqs=None,
                   num_tokens_across_dp=None,
                   aclgraph_runtime_mode: CUDAGraphMode = CUDAGraphMode.NONE,
-                  batch_descriptor=None):
+                  batch_descriptor=None,
+                  dummy_compute_logits=lambda hidden_states: None):
         pass
 
     def generate_token_ids(self,


whx-sjtu (approved):
We firstly fixed hanging issue of running MTP=1 with llm head tp in PR #3915. This PR refactors it to run `dummy_compute_logits` in drafter's `dummy_run` and further fixes MTP > 1 scenario. LGTM.

zouyida2052:
I've tested it on deepseek and it proves to be useful, please make ci happy

vllm_ascend/spec_decode/mtp_proposer.py
```diff
@@ -821,7 +824,7 @@ def _propose(
             # For the requests that exceed the max model length, we set the
             # sequence length to 1 to minimize their overheads in attention.
             exceeds_max_model_len_cpu = exceeds_max_model_len.to(
-                attn_metadata_i.seq_lens.device, non_blocking=True)
+                attn_metadata_i.seq_lens.device, non_blocking=False)
```
zouyida2052: What does this chang fix? Is it necessary?

MengqingCao, line 827: @jianzs I noticed you explain the reason why we disable non-blocking here. But IMO, the stream will keep the right order of data copy and the following operations in the same stream. I don't get the point on why there is an accuracy issue of this, is this a bug of torch-npu?

MengqingCao: In discuss offline, @jianzs mentioned that it is fine with h2d non-blocking copy, and the accuracy issue occurs with d2h copy. I think it might be a bug of torch-npu. Thus I'm fine with this change here as a workround, and we'll report this to torch-npu to finally fix it.

PR #4257, opened 2025-11-18
[Feat] Add support for MLP tensor splitting in the DeepSeek DenseFFN layer

yuweivvv:
#### Motivation
In the DeepSeek-R1 architecture, the first 3 blocks utilize Dense Feed-Forward Network (DenseFFN) layers, while the subsequent 58 blocks are MoE layers. The Gate_Up_Proj and Down_Proj matrices in these DenseFFN layers have large dimensions ([7168, 36864] and [18432, 7168] respectively).

This results in significant memory overhead:
FP16/BF16: ~1.48 GB per layer × 3 layers = 4.44 GB
W8A8 Quantization: ~1.11 GB total

During the decoding phase, using a pure Data Parallel (DP) strategy requires each GPU to store the complete model weights for these non-MoE layers. This redundancy causes excessive VRAM usage on each device.

#### Proposed Changes
This PR implements partial Tensor Parallelism (TP) splitting for the memory-intensive DenseFFN weights. Instead of replicating the full weights across all devices (DP), we split these specific tensors across the available GPUs.

#### Impact Analysis
**Memory Efficiency:** significantly reduces the memory footprint of weights on individual cards.
**Performance:** The performance gain from parallelizing the large matrix multiplications outweighs the introduced communication overhead.
**Conclusion:** We achieve lower VRAM usage without compromising TPOT (Time Per Output Token).
image
This PR introduces a new parameter in additional_config.

image

`--additional_config={"denseffn_tensor_parallel_size": 8}`

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -1037,8 +1037,9 @@ def is_first_k_dense(prefix: str) -> bool:
         )
     config = vllm_config.model_config.hf_config
 
-    n_routed_experts= getattr(config, 'n_routed_experts', 0)
-    first_k_dense_replace = getattr(config, 'first_k_dense_replace', float('inf'))
+    n_routed_experts = getattr(config, 'n_routed_experts', 0)
+    first_k_dense_replace = getattr(config, 'first_k_dense_replace',
+                                    float('inf'))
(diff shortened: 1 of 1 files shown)


wangxiyuan:

PR #4294, opened 2025-11-20
[chore] add triton-ascend to requirements.txt

wxh571001500:
Previously, when adapting vllm-ascend to models such as Qwen-Next, users needed to install the triton_ascend dependency separately. Therefore, to improve usability, the dependency is now included in requirements.txt and installed along with vllm-ascend.

nope

After installing vllm-ascend, we can use “pip list” to observe that triton-ascend is also installed synchronously.

- vLLM version: v0.11.0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ requires = [
     "quart",
     "numba",
     "opencv-python-headless<=4.11.0.86", # Required to avoid numpy version conflict with vllm
-    "triton-ascend==3.2.0rc4",
+    "triton-ascend",
 ]
 build-backend = "setuptools.build_meta"
 


realliujiaxu:
title should be [core]xxx ?

wxh571001500:
> title should be [core]xxx ?

"Chore" means the activity of building processes or supporting tools. Or would my change be more suitable for other keywords?

realliujiaxu:
> > title should be [core]xxx ?
> 
> "Chore" means the activity of building processes or supporting tools. Or would my change be more suitable for other keywords?

My mistake, it's ok to use chore.

wxh571001500:
Since there is currently no stable version of triton-ascend, it will be temporarily shut down.

PR #4036, opened 2025-11-06
[Quantization] Support compressed tensors w8a8 static and w8a8 dynamic weight

LHXuuu:
### What this PR does / why we need it?

While using the LLM Compressor quantization tool from the VLLM community to generate quantized weights, the VLLM Ascend engine needs to be adapted to support the compressed tensors quantization format.

1. Add AscendCompressedTensorsConfig to replace CompressedTensorsConfig in vllm.
2. Support CompressedTensorsW8A8 static weight.
   - weight: per-channel, int8, symmetric; activation: per-tensor, int8, symmetric.
4. Support CompressedTensorsW8A8Dynamic weight.
   - weight: per-channel, int8, symmetric; activation: per-token, int8, symmetric, dynamic.
5. Modify the override_quantization_method in AscendQuantConfig.

Co-authored-by: taoqun110 taoqun@huawei.com
Co-authored-by: chenxi-hh chen464822955@163.com

### Does this PR introduce _any_ user-facing change?

No

### How was this patch tested?

- vLLM version: v0.11.2
- vLLM main: https://github.com/vllm-project/vllm/commit/v0.11.2

PR #4316, opened 2025-11-20
[Quantization][Feature] Add AWQ quantization in vllm-ascend.

menogrey:
Add AWQ quantization in vllm-ascend. Most of the code refer to [existing implement](https://github.com/sgl-project/sglang/pull/10158) , and new quantization adaptation refer to compressed tensor: https://github.com/vllm-project/vllm-ascend/pull/4036 .
Known issue: 
- Use export VLLM_ASCEND_ENABLE_NZ=0 to avoid error when processing weights.
- expert parallel is not supported now.
- Some models have accuracy issue, Qwen3-235B-A22B-Instruct-AWQ Qwen3-VL-235B-A22B-Instruct-AWQ and Qwen2.5-VL-32B-Instruct-AWQ. More infomation at #4378 

- vLLM version: v0.12.0


MengqingCao:
@paulyu12 this pr implement AWQ quantization, and now it is under testing, just at you to take a look

menogrey:
Validation on this issue https://github.com/vllm-project/vllm-ascend/issues/4378

wangxiyuan:
Please rebase to main now.

vllm_ascend/quantization/awq/awq.py
```diff
@@ -0,0 +1,469 @@
...
+            return AWQMoEAscendMethod(self)
+        return None
+
+
+class AWQLinearAscendMethod(AWQLinearMethod):
+
+    def __init__(self, quant_config: AWQQuantConfig):
+        self.quant_config = quant_config
+
+    def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
```
SlightwindSec, line 213: **This is a solid solution for unifying weight formats.**
By preprocessing the weights during the loading phase, you effectively resolve the packing layout discrepancy between GPU and NPU, which is critical for compatibility.

**Optimization Suggestions:**
However, for larger models, the current `for` loop might introduce noticeable latency due to repeated kernel launches. Additionally, bitwise shift operations (`>>`, `<<`) can be less efficient than arithmetic operations on NPU hardware.

I suggest refactoring this implementation to **vectorize the logic** (eliminating the loop) and **replace the bitwise shifts with equivalent multiplication and integer division**. This should significantly improve the loading performance.

menogrey: Thanks for your suggestions. I try to remove the `for` loop but an OOM problem occurred, and the fact is if you want to execute the unpack operation, you have to alloc a big tensor (8x times memory for one tensor), and need more temporary variable if not using a in-place operation (like the bitwise operation). I think it is a trade-off between the memory usage and execution efficiency, but for a quantization, the memory usage is more important, so i won't change it, unless you have some better suggestions. @SlightwindSec

SlightwindSec: Fair point. I agree that avoiding OOM is the top priority here.

ZhongsJie:
Any progress on this?

menogrey:
> Any progress on this?

@ZhongsJie I will resume work on this PR following the completion of the quantization module refactoring.

ZhongsJie:
> > Any progress on this?
> 
> @ZhongsJie I will resume work on this PR following the completion of the quantization module refactoring.

@menogrey I’m reworking this part in https://github.com/vllm-project/vllm-ascend/pull/7672 — if you have time, could you take a look?

menogrey:
> > > Any progress on this?
> > 
> > 
> > @ZhongsJie I will resume work on this PR following the completion of the quantization module refactoring.
> 
> @menogrey I’m reworking this part in #7672 — if you have time, could you take a look?

That's great! In addition there are two reason that this work is pending
* I want to refactor the quantization module to align with the vllm upstream cause the upstream use better kernel abstraction. Refer to https://github.com/vllm-project/vllm-ascend/pull/6644
* The current w4a16 implement have some issues on some models as i showed above. And you should refer to the modelslim w4a16 implement.https://github.com/vllm-project/vllm-ascend/pull/4516

PR #4327, opened 2025-11-21
[Doc][Model] Add principles of modeling files in vllm-ascend

shen-shanshan:
Add principles of modeling files in vllm-ascend.

- vLLM version: v0.11.0


shen-shanshan:
CC @Yikun

PR #4351, opened 2025-11-21
[Bugfix] fix patch typo

845473182:
Fix a bug caused by this pr: https://github.com/vllm-project/vllm-ascend/pull/4223
The bug makes vllm-ascend/vllm_ascend/patch/platform/patch_multiproc_executor.py patch in a wrong way

no
Tested in a single node. When the environment DYNAMIC_EPLB is set to true, the patch works correctly. When it's set to false, the patch do not patch
- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/patch/platform/__init__.py
+++ b/vllm_ascend/patch/platform/__init__.py
@@ -16,12 +16,11 @@
 
 import os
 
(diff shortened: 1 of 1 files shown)


845473182:
> thanks for fixing! plz update your pr message

Updated

PR #4354, opened 2025-11-22
[Bugfix] use module-level import for patched function in Qwen3Next

zjchenn:

**Problem**: The Qwen3Next model implementation currently imports chunk_gated_delta_rule directly using `from ... import ...`

In frameworks like `verl`, the model file is often imported before `vllm-ascend` initializes and applies its patches. This causes the model to permanently hold a reference to the original (unpatched) vLLM kernel, resulting in execution errors on Ascend devices even if the patch runs later.

**Solution**: Changed the import style to `from vllm...ops import chunk` and call `chunk.chunk_gated_delta_rule().`

This ensures that the function lookup happens at runtime (dynamic dispatch), allowing the model to correctly pick up the patched function regardless of import order.

No. This is an internal bug fix to resolve import reference issues.

- vLLM version: v0.11.0


zjchenn:
@wangxiyuan hi, please help me have a review

PR #4359, opened 2025-11-22
[refact] unified soc_version code

zzzzwwjj:

Currently, there are two paths to judge the chip type in code, `get_ascend_soc_version` use `get_soc_version` api in torch_npu, and `is_310p` `use _build_info.__soc_version__`, which generate when install. We need to unify the two paths.

We need to unify these codes based on the following points:

1. We need to ensure consistency in chip type judgment between compiling and running states;
2. In compiling state, we need chip type to complete op's compilation, but in running state, we only need device type(910B/910_93/310P/910_95/etc) to make code branch judgement;
3. In compiling state, torch_npu may not have been installed yet, so we can't use torch_npu's api.

Based on the above points, we have made the following changes:

1. When user set env `SOC_VERSION`, use it; when not set, query soc_version by `npu-smi`;
2. generate device_type based on soc_version when compiling, and write `__device_type__` instead of `__soc_version__` in `_build_info.py`;
3. In running state, use `__device_type__` to judge code branch.

When not set env `SOC_VERSION`, it will not be `ASCEND910B1` by default, we will query soc_version by `npu-smi`. And env `SOC_VERSION` must be in the list `soc_to_device` in `setup.py`.

- vLLM version: v0.11.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,25 +65,103 @@ def check_or_set_default_env(cmake_args,
     return cmake_args
 
 
+def get_value_from_lines(lines: List[str], key: str) -> str:
+    for line in lines:
+        line = ' '.join(line.split())
+        if key in line:
+            return line.split(':')[-1].strip()
+    return ""
+
+
+def get_chip_info() -> str:
+    try:
+        npu_info_lines = subprocess.check_output(
+            ['npu-smi', 'info', '-l']).decode().strip().split('\n')
+        npu_id = int(get_value_from_lines(npu_info_lines, 'NPU ID'))
+        chip_info_lines = subprocess.check_output(
+            ['npu-smi', 'info', '-t', 'board', '-i',
+             str(npu_id), '-c', '0']).decode().strip().split('\n')
+        chip_name = get_value_from_lines(chip_info_lines, 'Chip Name')
+        chip_type = get_value_from_lines(chip_info_lines, 'Chip Type')
+        npu_name = get_value_from_lines(chip_info_lines, 'NPU Name')
+
+        if "310" in chip_name:
+            # 310P case
+            assert chip_type
+            return (chip_type + chip_name).lower()
+        elif "910" in chip_name:
+            if chip_type:
+                # A2 case
+                assert not npu_name
+                return (chip_type + chip_name).lower()
+            else:
+                # A3 case
+                assert npu_name
+                return (chip_name + '_' + npu_name).lower()
+        else:
+            # TODO(zzzzwwjj): Currently, A5's chip name has not determined yet.
+            raise ValueError(
+                f"Unable to recognize chip name: {chip_name}, please manually set env SOC_VERSION"
+            )
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(f"Get chip info failed: {e}")
+    except FileNotFoundError:
+        # cpu envir, release code case, return `ascend910b1` by default
+        return "ascend910b1"
+
+
 envs = load_module_from_path("envs",
                              os.path.join(ROOT_DIR, "vllm_ascend", "envs.py"))
+
+soc_version = get_chip_info()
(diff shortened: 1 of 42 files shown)


vllm_ascend/utils.py
```diff
@@ -732,31 +723,29 @@ def register_ascend_customop(vllm_config: Optional[VllmConfig] = None):
     _ASCEND_CUSTOMOP_IS_REIGISTERED = True
 
 
-# TODO(zzzzwwjj): Currently there is no clear SOC_VERSION policy for A2 and A3 in CANN.
-# So we get the version dynamically. In the future, we should get the version info from _build_info like 310p does.
 class AscendSocVersion(Enum):
     A2 = 0
     A3 = 1
-    UNDEFINED = 2
```
whx-sjtu, line 711: A variable name starting with `_` generally indicates private usage, but the `_310P` here is actually directly used by other modules. Why use `_310P` instead of `310P` here?

zzzzwwjj: cannot start with digit.

setup.py
```diff
@@ -65,25 +65,67 @@ def check_or_set_default_env(cmake_args,
...
+    except subprocess.CalledProcessError as e:
+        # cpu envir, release code case, return `ascend910b1` by default
+        return "ascend910b1"
+
+
 envs = load_module_from_path("envs",
                              os.path.join(ROOT_DIR, "vllm_ascend", "envs.py"))
 
+if not envs.SOC_VERSION:
+    envs.SOC_VERSION = get_chip_info()
```
whx-sjtu: Add a warning here if user setting of `SOC_VERSION` doesn't match with results of `get_chip_info()`

setup.py
```diff
@@ -65,25 +65,69 @@ def check_or_set_default_env(cmake_args,
...
+        "ascend310p1": "310P",
+        "ascend910b": "A2",
+        "ascend910b1": "A3",
+    }
+
     package_dir = os.path.join(ROOT_DIR, "vllm_ascend", "_build_info.py")
     with open(package_dir, "w+") as f:
         f.write('# Auto-generated file\n')
-        f.write(f"__soc_version__ = '{soc_version}'\n")
+        f.write(f"__soc_version__ = '{soc_version_dict[soc_version]}'\n")
```
whx-sjtu: What if user sets a value of `SOC_VERSION` with wrong format that doesn't match with any keys in `soc_version_dict`?

setup.py
```diff
@@ -65,25 +65,108 @@ def check_or_set_default_env(cmake_args,
...
+        "ascend910b3": "A2",
+        "ascend910b4": "A2",
+        "ascend910b4-1": "A2",
+        "ascend910_9391": "A3",
+        "ascend910_9381": "A3",
+        "ascend910_9372": "A3",
+        "ascend910_9392": "A3",
+        "ascend910_9382": "A3",
+        "ascend910_9362": "A3",
+        "ascend310p1": "300I",
```
whx-sjtu: Why use `300I` instead of `310P` here? I think `300I` is confused because there are machines like `300I A2`.

vllm_ascend/utils.py
```diff
@@ -703,32 +694,55 @@ def register_ascend_customop(vllm_config: Optional[VllmConfig] = None):
...
 
 
-# TODO(zzzzwwjj): Currently there is no clear SOC_VERSION policy for A2 and A3 in CANN.
-# So we get the version dynamically. In the future, we should get the version info from _build_info like 310p does.
-class AscendSocVersion(Enum):
+class AscendDeviceType(Enum):
     A2 = 0
     A3 = 1
-    UNDEFINED = 2
+    _300I = 2
```
whx-sjtu: ditto

MengqingCao (approved):

PR #4365, opened 2025-11-23
[Doc] Upgrade multi-node doc

Potabk:
When we are using `Ascend scheduler`, the param `max_num_batched_tokens` should be larger than `max_model_len`, otherwise, will encountered the follow error:
```shell
Value error, Ascend scheduler is enabled without chunked prefill feature. Argument max_num_batched_tokens (4096) is smaller than max_model_len (32768). This effectively limits the maximum sequence length to max_num_batched_tokens and makes vLLM reject longer sequences. Please increase max_num_batched_tokens or decrease max_model_len. [type=value_error, input_value=ArgsKwargs((), {'model_co...g': {'enabled': True}}}), input_type=ArgsKwargs]
```

Users/Developers who running the model according to the [tutorial](https://docs.vllm.ai/projects/ascend/en/latest/tutorials/multi_node.html), the parameters can be specified correctly.

- vLLM version: v0.11.0

### DIFF
--- a/.github/workflows/_e2e_nightly_multi_node.yaml
+++ b/.github/workflows/_e2e_nightly_multi_node.yaml
@@ -216,7 +216,7 @@ jobs:
 
               # 1) check follower pods
               ALL_FOLLOWERS_READY=true
-              for ((i=1; i<${SIZE}; i++)); do
+              for ((i=1; i<SIZE; i++)); do
                 POD="${POD_PREFIX}-${i}"
                 PHASE=$(kubectl get pod "$POD" -n "$NAMESPACE" -o jsonpath='{.status.phase}' 2>/dev/null || echo "NotFound")
                 READY=$(kubectl get pod "$POD" -n "$NAMESPACE" -o jsonpath='{.status.containerStatuses[*].ready}' 2>/dev/null)


docs/source/tutorials/multi_node.md
```diff
@@ -131,9 +131,9 @@ vllm serve vllm-ascend/DeepSeek-V3.1-W8A8 \
 --served-model-name deepseek_v3.1 \
 --enable-expert-parallel \
 --max-num-seqs 16 \
---max-model-len 32768 \
+--max-model-len 8192 \
 --quantization ascend \
---max-num-batched-tokens 4096 \
+--max-num-batched-tokens 8192 \
```
MengqingCao, line 136: `max-num-batched-tokens` will be set equal to `max-model-len` defaultly, and we don't need to specify it

MengqingCao: Actually I'm curious that this constraint exists before your last pr, why it is triggered now?

wangxiyuan: it's a limit by ascend schduler 
https://github.com/vllm-project/vllm-ascend/blob/b34f195cc8760ead9a704662cd3f6211054bb553/vllm_ascend/core/schedule_config.py#L66-L75

PR #4377, opened 2025-11-24
Drop 0.11.0 support

wangxiyuan:
There is  a lot hack code for v0.11.0, which makes the code hard to upgrade to newer vLLM version. Since v0.11.0 will release soon. Let's drop v0.11.0 support first. Then we'll upgrade to v0.11.2 soon.

- vLLM version: v0.11.0


wangxiyuan:
E2E passed here: https://github.com/vllm-project/vllm-ascend/actions/runs/19623142380/job/56187141663?pr=4377  the error test is  known issue which is not stable enough

PR #4384, opened 2025-11-24
[MM][Bugfix] Minor fix for VL model verification

shen-shanshan:

To fix ops test, where `model_config` has been set to `None` and doesn't has `hf_config` attribute, we have added a check for `model_config` to guarantee it is not `None_Type`.

- vLLM version: v0.11.0


shen-shanshan:
CC @wangxiyuan

shen-shanshan:
CC @wangxiyuan

PR #4385, opened 2025-11-24
[MM][Bugfix] Minor fix for VL model verification

shen-shanshan:

To fix ops test, where `model_config` has been set to `None` and doesn't has `hf_config` attribute, we have added a check for `model_config` to guarantee it is not `None_Type`.

cherry-pick from main: https://github.com/vllm-project/vllm-ascend/pull/4384.


shen-shanshan:
CC @wangxiyuan

PR #4388, opened 2025-11-24
[Performance] Pre-issued exponential distribution operator.

weijinqian0:
Pre-issued exponential distribution operator.

We will add the performance data later.

- vLLM version: v0.11.0

### DIFF
--- a/vllm_ascend/sample/sampler.py
+++ b/vllm_ascend/sample/sampler.py
@@ -17,16 +17,17 @@ def random_sample(
     We use this function instead of torch.multinomial because torch.multinomial
     causes CPU-GPU synchronization.
     """
-    q = torch.empty_like(probs)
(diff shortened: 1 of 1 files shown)


vllm_ascend/utils.py
```diff
@@ -312,6 +313,15 @@ def current_stream() -> torch.npu.Stream:
     return _CURRENT_STREAM
 
 
+def global_stream() -> torch.npu.Stream:
```
jgong5, line 316: The name "global" feels not very explicit here. Can you explain when one is supposed to use the "global" stream, not the "current" stream?

weijinqian0: it is for the multi-stream cover-up.

weijinqian0:
recommit in #4908.

PR #4394, opened 2025-11-24
[Kernel] add triton kernels for sampling

swy20190:
Replace pyorch implement of sampling with triton kernels

No

- vLLM version: v0.11.2


vllm_ascend/sample/rejection_sampler.py
```diff
@@ -149,25 +158,15 @@ def rejection_sample(
...
-                output_token_ids,
-                cu_num_draft_tokens,
-                draft_token_ids,
-                target_argmax,
-                bonus_token_ids,
-                num_draft_tokens,
-                max_spec_len,
-                is_greedy,
-            )
+        rejection_greedy_sample_kernel[(batch_size,)](
```
whx-sjtu: After communicating with @wangxiyuan, you need to use `HAS_TRITON` check here. If user hasn't installed triton, fall back to original implementation to ensure functionality. An example is here:
https://github.com/vllm-project/vllm-ascend/blob/a3225c4682adc4e7a5b2c8421a14d893f2cf4b36/vllm_ascend/attention/sfa_v1.py#L505

whx-sjtu (approved):

vllm_ascend/sample/rejection_sampler.py
```diff
@@ -10,6 +10,15 @@
...
 from vllm.v1.spec_decode.metadata import SpecDecodeMetadata
 
+from vllm_ascend.utils import vllm_version_is
+
+if vllm_version_is("0.11.0"):
+    from vllm.v1.sample.rejection_sampler import compute_probs
+else:
+    from vllm.v1.sample.rejection_sampler import apply_sampling_constraints
+
+import triton.language as tl
```
whx-sjtu: Use `from vllm.triton_utils import tl, triton` here, same reason.

whx-sjtu:
In conclusion, currently we need to make sure that original functionality won't be broken by new triton optimization in environments without triton.

lio1226:
> In conclusion, currently we need to make sure that original functionality won't be broken by new triton optimization in environments without triton.

I'd like to ask which type of machine you tested these operators on. When I tested the `sample_recovered_token_kernel` on a 910B3, I encountered an UB memory overflow issue. Perhaps we should optimize this kernelto adapt to all machines.

vllm_ascend/sample/rejection_sampler.py
jgong5: Also, it would be helpful (e.g. for accuracy checking purpose or UT) if we allow running the original implementation even with triton installed, perhaps via some configurations.

vllm_ascend/sample/rejection_sampler.py
```diff
@@ -282,226 +280,198 @@ def sample_recovered_tokens(
...
-                else:
-                    rejected = True
-                    token_id = recovered_token_ids[start_idx + pos].item()
-
-                output_token_ids[req_idx, pos] = token_id
-
+    req_idx = tl.program_id(0)
+    is_greedy = tl.load(is_greedy_ptr + req_idx)
+    if is_greedy:
+        # Early exost for greedy sampling requests
```
jgong5: nit: exost -> exit

wangxiyuan:

PR #4396, opened 2025-11-24
prefillnocache branch use fia op.

momo609:
prefillnocache branch use fia op.


wangxiyuan:
replace by https://github.com/vllm-project/vllm-ascend/pull/4455

PR #4413, opened 2025-11-24
[Ops][Triton] Add a triton kernel supporting partial rope.

whx-sjtu:
This PR adds a triton rope kernel witch supports scenarios of `rope_dim != head_dim`. This can save the split op before rope and the concat op after rope. Profiling shows improvement.

Original Implementation(2 split+2 rope+ 2 slice +2 concat):
image
Because currently we only support piecewise aclgraph for DS 3.2, so there are plenty of free bubbles. You can see the computing time of all rope related kernels: **35us**
image

New Triton Rope: **12us**
image

image

None
I will add related ut after ci integrated with triton.

- vLLM version: v0.11.2

### DIFF
--- a/tests/e2e/nightly/ops/triton/test_rope.py
+++ b/tests/e2e/nightly/ops/triton/test_rope.py
@@ -1,5 +1,4 @@
 import gc
-from typing import Optional
 
 import pytest
 import torch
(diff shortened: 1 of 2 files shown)


whx-sjtu:
For prefill stage with larger bs, the performance gains remains substantial. Here is an example of `bs=1001` with DS V3.2:

Without triton rope: **109us**
image

With triton rope: **38us**
image

PR #4416, opened 2025-11-25
[bugfix] dep ineffective

shenchuxiaofugui:
The expert mapping table and weights of the dynamic EPLB were not updated, causing the accuracy to be correct but not effective. This bug has now been fixed.

NO


MengqingCao:
This is a backport of https://github.com/vllm-project/vllm-ascend/pull/4417

PR #4432, opened 2025-11-25
Change comment location

herizhen:
When running 'python example.py',connection issues often occur.The solution is to comment out the first line the code.
Complete the specific names of machines A2 and A3.
Standardize document format,a space should be added after the colon.
no
ut

- vLLM version: v0.11.2

### DIFF
--- a/docs/source/tutorials/multi_node_kimi.md
+++ b/docs/source/tutorials/multi_node_kimi.md
@@ -5,7 +5,7 @@
 Refer to [multi_node.md](https://vllm-ascend.readthedocs.io/en/latest/tutorials/multi_node.html#verification-process).
 
 ## Run with Docker
-Assume you have two Atlas 800I A3 (64G*16)  or four Atlas 800I A2 nodes, and want to deploy the `Kimi-K2-Instruct-W8A8` quantitative model across multiple nodes.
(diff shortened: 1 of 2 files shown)


docs/source/tutorials/multi_node_kimi.md
```diff
@@ -5,7 +5,7 @@
 Refer to [multi_node.md](https://vllm-ascend.readthedocs.io/en/latest/tutorials/multi_node.html#verification-process).
 
 ## Run with Docker
-Assume you have two Atlas 800 A3 (64G*16)  or four A2 nodes, and want to deploy the `Kimi-K2-Instruct-W8A8` quantitative model across multiple nodes.
+Assume you have two Atlas 800I A3 (64G*16)  or four Atlas 800I A2 nodes, and want to deploy the `Kimi-K2-Instruct-W8A8` quantitative model across multiple nodes.
```
wangxiyuan: I think we support both 800I and 800T here. So we just leave `800` with no suffix.

herizhen: Alright,after discussion, this part will not be modified and it has been re-committed.

wangxiyuan (approved):

PR #4437, opened 2025-11-25
[v0.11.0-dev][bugfix] Add branch for stream up-lifting in `update_attn_params`

Angazenn:
#3985 move stream context initialization before for-loops to improve performance. However, we find that this might cause potential accuracy drop when used with pd disaggregation. Thus we partly revert this change when using pd disaggregation, and we shall fix this bug in th future. 

No.


Wwwzff:
@Angazenn Hi I'm curious about this bug and could you please share more information about the accuracy issues with pd disaggregation?

PR #4465, opened 2025-11-26
[Feat] MTP support DeepSeekV3.2

ZYang6263:
Currently, MTP does not support the DeepSeekV3.2 model. In this PR, we have enabled this feature.

- vLLM version: v0.11.2


rjg-lyh:
This PR supports the MTP feature under the DSV3.2 scenario and significantly reduces inference latency in the E2E performance tests. The specific performance results are shown in the graph below:
Before:
image
After，MTP=2:
image

PR #4475, opened 2025-11-26
unregister mooncake layerwise connector

liziyu179:
unregister mooncake layerwise connector


wangxiyuan:
let's keep as it is. we'll announce that this feature is experimental for v0.11.0 release

PR #4478, opened 2025-11-26
Modify the position of npu sync ops in layerwise connector for better TTFT performance

anysources:
When testing with **mooncake_layerwise_connector**, it was found that in the case of long input sequences and high concurrency, the masking effect between layer-wise kv tranfer and prefill computing is very poor. Most layer-wise kv transfer are exposed, resulting in unsatisfactory TTFT performance. 
After analysis, by modifying the NPU synchronization operation before the **batch_transfer_sync_write** function and placing it in the **wait_for_layer_load** function, the layer-wise kv transfer and prefill computing can be perfectly masked, and the TTFT performance is improved.

vllm version: v0.11.0

- vLLM version: v0.11.2


wangxiyuan:

PR #4485, opened 2025-11-27
chip type judgement code optimization

zzzzwwjj:
| | cpu envir | npu envir |
|---|---|---|
| set `SOC_VERSION` | check if `SOC_VERSION` is in dict `soc_to_device`, if not, raise an error that can not support current chip type. | print a warning log when `SOC_VERSION` is not equal to chip type from `npu-smi`, same as left for others. |
| not set `SOC_VERSION` | raise an error that `SOC_VERSION` is necessary when compiling in a cpu envir. | use chip type from `npu-smi` to compile vllm-ascend. |

Now we must set env `SOC_VERSION` when compiling in cpu envir. 

- vLLM version: v0.11.2

### DIFF
--- a/.github/Dockerfile.buildwheel
+++ b/.github/Dockerfile.buildwheel
@@ -18,10 +18,12 @@ ARG PY_VERSION=3.11
 FROM quay.io/ascend/manylinux:8.3.rc1-910b-manylinux_2_28-py${PY_VERSION}
 
 ARG COMPILE_CUSTOM_KERNELS=1
+ARG SOC_VERSION
 
 # Define environments
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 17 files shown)


Yikun:
```
raise an error that SOC_VERSION is necessary when compiling in a cpu envir.
```

1. This break current user and dev experience, it's UNACCEPTABLE. Previously, user don't need to know it.
2. The user is very diffcult to know which chip (such as `ascend910_9391`) are used and they only know they have a A2/A3/310P.

I suggested to:
1. Add default var in Dockerfile rather than workflow.
2. Consider the right var, I don't think let vLLM user know  `ascend910_9391` and `ascend310p1` is NOT a good choice.

PR #4511, opened 2025-11-27
support async mtp

Ronald1995:
this pr aims to support async_scheduling for mtp, which refer to vllm pr https://github.com/vllm-project/vllm/pull/24799.
and this pr fix some synchronize problem in vllm-ascend.

- vLLM version: v0.12.0


vllm_ascend/attention/attention_v1.py
```diff
@@ -348,8 +348,8 @@ def build(
                              device=query_start_loc_cpu.device)
             ])
 
-        query_start_loc = query_start_loc_cpu.to(self.device,
-                                                 non_blocking=True)
+        query_start_loc = query_start_loc_cpu.pin_memory().to(
```
realliujiaxu, line 320: Would it be better to mimic the implementation of gpu_model_runner?
```
def __init__(...):
       self.query_start_loc = self._make_buffer(
            self.max_num_reqs + 1, dtype=torch.int32
        )

def _prepare_inputs(...):
     ....
    self.query_start_loc.copy_to_gpu()
```

Ronald1995: query_start_loc  is a field of common_attn_metadata, it's not defined in __init__ method

weijinqian0 (approved):

PR #4512, opened 2025-11-27
[bugfix] the hccl_buffsize config for EP

kiscad:

The hccl_buffsize for EP process group is set by the `_DEFAULT_BUFFER_SIZE`, which is not enough for fused_moe kernel, especially for large max-num-batched-tokens. So we need to calculate the hccl_buffer for EP from the vllm_config.

- vLLM version: v0.11.2

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -739,6 +739,9 @@ def get_hccl_config_for_pg_options(group_name: str) -> Optional[dict]:
         "dp": {
             "hccl_buffer_size": calculate_dp_buffer_size()
         },
+        "ep": {
+            "hccl_buffer_size": calculate_ep_buffer_size()
+        },
     }
     return hccl_config_map.get(group_name, get_default_buffer_config())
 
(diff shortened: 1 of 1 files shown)


vllm_ascend/utils.py
```diff
@@ -730,21 +729,27 @@ def get_hccl_config_for_pg_options(group_name: str) -> Optional[dict]:
...
-    # result in memory misalignment problems.
-    if group_name and "mc2" in group_name:
-        return None
     hccl_config_map = {
         "dp": {
             "hccl_buffer_size": calculate_dp_buffer_size()
         },
     }
-    return hccl_config_map.get(group_name, get_default_buffer_config())
+    return hccl_config_map.get(group_name, get_buffer_config())
```
linfeng-yuan: The aim of this design is to only set mc2 VRAM buffer with `HCCL_BUFFSIZE` and keep other operator buffers as default, decreasing wasted VRAM allocation. If you wanna adjust buffer allocation of specific process group(e.g., `ep_group`), it is recommended to calculate and set it in `hccl_config_map` in L732.

vllm_ascend/utils.py
```diff
@@ -760,6 +763,25 @@ def calculate_dp_buffer_size() -> int:
...
+    vllm_config = get_current_vllm_config()
+    hf_config = vllm_config.model_config.hf_config
+
+    hidden_size = hf_config.hidden_size
+    topk = getattr(hf_config, "num_experts_per_token", 1)
+    batch_size = vllm_config.scheduler_config.max_num_batched_tokens
+    int8_size = torch.iinfo(torch.int8).bits // 8
+    bf16_size = torch.finfo(torch.bfloat16).bits // 8
+    ep_buffer_size = math.ceil((batch_size * hidden_size * topk *
+                                (int8_size * 2 + bf16_size)) / (1024 * 1024))
```
Sparkheart, line 781: 1024 is recommended to be extracted as a constant.

PR #4518, opened 2025-11-28
[Fix] Fix FIA `query` and `query_start_loc` shape mismatch error

momo609:
Due to the requirement of the FIA operator that the **query.shape[0]** must match **actual_seq_len[-1]**, in graph mode and multi-DP scenarios, the query is padded to the size of **num_input_token**. This leads to validation errors during tiling in the operator. However, since the padding is applied at the end of the query, it does not affect the actual execution result of the operator, and the precision remains unaffected.
image
Our modification padding both **actual_seq_len** and **actual_seq_len_kv** to resolve the validation issue in the operator.

- vLLM version: v0.11.2

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -183,7 +183,6 @@ class AscendMetadataForDecode:
 class AscendMetadata:
     # **************************** Basic Properties ************************** #
     attn_mask: Optional[torch.Tensor] = None
-    fia_attn_mask: Optional[torch.Tensor] = None
     # Current state of this attention run.
     attn_state: AscendAttentionState = AscendAttentionState.ChunkedPrefill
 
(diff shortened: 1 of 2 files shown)


Angazenn:
Please check whether padding block_table is necessary and whether it causes potential excessive memory consumption. cc @yiz-liu

yiz-liu:
> Please check whether padding block_table is necessary and whether it causes potential excessive memory consumption. cc @yiz-liu

Why did we pad this in the first place? @Angazenn

yiz-liu:
> > Please check whether padding block_table is necessary and whether it causes potential excessive memory consumption. cc @yiz-liu
> 
> Why did we pad this in the first place? @Angazenn

I think it's OK to just remove it, better check it with op developer.

yiz-liu (approved):

PR #4533, opened 2025-11-28
feat(sfa): Integrate fused mla_preprocess operator

yiz-liu:
Replaces several discrete operations in the SFA forward pass with a single call to the fused `mla_preprocess` custom operator. This operator combines Q/K/V projection, RoPE application, and KV cache updates into one kernel.

A new weight processing method is added to transform and pre-process weights into the specific layout required by the fused operator. This change aims to improve performance by reducing kernel launch overhead.

Additionally, the condition for allocating RoPE caches is relaxed to support MLA in modes other than just full decode.

None

None

- vLLM version: v0.11.2


yiz-liu:
It's included in #4753 , closing this one.

PR #4536, opened 2025-11-28
fix qwen3vl ci

shaopeng-666:
Follow the interface changes from the upstream vllm repository ‘
[Core][MM] Add mechanism to configure multimodal fields which should stay on CPU (https://github.com/vllm-project/vllm/pull/28168)

- vLLM version: v0.11.2


wangxiyuan:
@shen-shanshan

shen-shanshan:
LGTM.
This PR just sync the changes in https://github.com/vllm-project/vllm/pull/28168 to keep interface compatible.
But the while Qwen3-VL ViT may be removed directly to avoid maintaince.

wangxiyuan:

PR #4498, opened 2025-11-27
drop ascend scheduler

wangxiyuan:
Ascend scheduler was added for non chunk prefill case before, since that the npu ops didn't work well with chunked prefill.

Now the ops with chunked prefill work better, it's time to remove the ascend scheduler to use vLLM default scheduler.

- vLLM version: v0.11.2
- vLLM main: https://github.com/vllm-project/vllm/commit/v0.11.2

PR #4400, opened 2025-11-24
upgrade to vllm 0.11.2

wangxiyuan:
Bump vLLM version to v0.11.2

What's broken and changed by vLLM:
1. structured_output is broken by https://github.com/vllm-project/vllm/pull/26866
2. get_mrope_input_positions is broken by https://github.com/vllm-project/vllm/pull/28399
3. graph mode is broken by https://github.com/vllm-project/vllm/pull/25110 we'll upgrade torch to 2.8 to fix the problem later
4. embedding is broken by https://github.com/vllm-project/vllm/pull/27583
5. `get_attn_backend_cls` and attention backend is broken are broken by https://github.com/vllm-project/vllm/pull/28534
6. spec decode is broken by https://github.com/vllm-project/vllm/pull/28771
7. sp feature is broken by https://github.com/vllm-project/vllm/pull/27126
8. mtp is broken by https://github.com/vllm-project/vllm/pull/27922
9. lora is broken by https://github.com/vllm-project/vllm/pull/21068
10. execute_model is broken by https://github.com/vllm-project/vllm/pull/26866
11. `VLLM_DISABLE_SHARED_EXPERTS_STREAM` env is broken by https://github.com/vllm-project/vllm/pull/28159
12. kv cahe is broken by https://github.com/vllm-project/vllm/pull/27753
13. dp is broken by https://github.com/vllm-project/vllm/pull/25110

 
What's broken and changed by ourself:
1. qwen vl is broken by https://github.com/vllm-project/vllm/pull/28455 We'll remove model files in the future to avoid this kind of error
2. Engine core is broken by https://github.com/vllm-project/vllm/pull/23691 We'll remove the patch file in the future.
3. Ascend scheduler is broken by https://github.com/vllm-project/vllm/pull/28733 We'll remove ascend scheudler later.
4. qwen3-next is broken by https://github.com/vllm-project/vllm/pull/28083 We'll remove model files in the future to avoid this kind of error
5. qwen vl is broken by https://github.com/vllm-project/vllm/pull/27764. We'll remove model files in the future

Known issue:
1. ray doesn't work 
2. the accuracy of qwen3-next is not correct
3. qwen3-vl is broken
4. prefix cache+ ascend scheduler + deepseek v2 lite is broken.

Co-authored-by: MengqingCao <cmq0113@163.com>
Co-authored-by: hfadzxy <starmoon_zhang@163.com>
Co-authored-by: leo-pony <nengjunma@outlook.com>
[...]

PR #4551, opened 2025-11-28
[FixCI] Enable chunked prefill for auto-prefix-caching test

MengqingCao:
Enable chunked prefill for auto-prefix-caching test.

After upgrading the vLLM version to v0.11.2 in #4400, the accuracy on auto-prefix-caching + non-chunked-prefill is broken. Due to chunked prefill is enabled by default, we change the apc test case firstly to recover CI.

After https://github.com/vllm-project/vllm/pull/28833, non-Pooling models won't be able to disable chunked prefill any more. 
Maybe we should fix the chunked prefill configuration issue in https://github.com/vllm-project/vllm-ascend/pull/4498 together
CI passed with existing test.

- vLLM version: v0.11.2

### DIFF
--- a/tests/e2e/multicard/test_prefix_caching.py
+++ b/tests/e2e/multicard/test_prefix_caching.py
@@ -58,7 +58,6 @@ INPUT_PROMPTS = [
 ]
 
 
-@pytest.mark.skip(reason="Fix me, the accuracy is not correct")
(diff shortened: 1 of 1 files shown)


MengqingCao:
@wangxiyuan PTAL

MengqingCao:
closing as done by https://github.com/vllm-project/vllm-ascend/pull/4569

PR #4562, opened 2025-11-29
[bugfix] Repair the problem of moe model accuracy caused by version upgrade.

weijinqian0:
Repair the problem of moe model accuracy caused by version upgrade.

Reason:
The new version adds the "reduce_output" operation after "forward_impl".

Then we have fully taken over the implementation of the FusedMoe module.

- vLLM version: v0.11.2


vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -292,7 +292,7 @@ def maybe_all_reduce_tensor_model_parallel(
         return torch.ops.vllm.maybe_all_reduce_tensor_model_parallel(
             final_hidden_states)
 
-    def forward_impl(self, hidden_states: torch.Tensor,
+    def forward_native(self, hidden_states: torch.Tensor,
```
MengqingCao: if we just want to replace the whole forward pipeline of FusedMoE, I think we can just overwirte the `forward_oot` to make it clearly.
```suggestion
    def forward_oot(self, hidden_states: torch.Tensor,
```

MengqingCao:
LGTM, plz fix the lint issue and we'll merge it as soon as the CI is passed

vllm_ascend/ops/fused_moe/fused_moe.py
weijinqian0: fixed

wangxiyuan: maybe override `forward` is better

MengqingCao (approved):

PR #4590, opened 2025-12-01
【fix】: add attn_groups judge

wucong25:


weijinqian0:
This PR is not common in RL scenarios.

PR #4593, opened 2025-12-01
[Bugfix] fix custom op GmmSwigluQuantWeightNzTensorList

ChenxiQ:

1. Fixes the environment path used to locate custom op shared libraries.
2. Uses empty tensor initialization for op outputs instead of zero-initialization for better efficiency.

- vLLM version: v0.11.2

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -247,6 +247,19 @@ def enable_custom_op():
     Ensure that ASCEND_RT_VISIBLE_DEVICES can be dynamically modified before torch.npu.set_device().
     """
     global _CUSTOM_OP_ENABLED
+
+    # set custom ops path
+    CUR_DIR = os.path.dirname(os.path.realpath(__file__))
+    CUSTOM_OPP_PATH = os.path.join(CUR_DIR, "_cann_ops_custom", "vendors",
+                                   "vllm-ascend")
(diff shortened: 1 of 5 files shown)


vllm_ascend/utils.py
```diff
@@ -247,6 +247,19 @@ def enable_custom_op():
     Ensure that ASCEND_RT_VISIBLE_DEVICES can be dynamically modified before torch.npu.set_device().
     """
     global _CUSTOM_OP_ENABLED
+
+    # set custom ops path
+    CUR_DIR = os.path.dirname(os.path.realpath(__file__))
```
hukongyi, line 252: @wangxiyuan This line causes the test tests/e2e/multicard/test_offline_inference_distributed.py::test_models_distributed_QwQ to fail.
os.path.realpath involves system calls that are not supported by torch.compile (Dynamo) during graph capture. This triggers a runtime crash when graph mode is enabled.
Maybe move the CUR_DIR calculation to the module level (global scope) to avoid this trace error and unnecessary re-calculation.

wangxiyuan: yes, I found it. Thanks for reminding

wangxiyuan: @hukongyi https://github.com/vllm-project/vllm-ascend/pull/4675 I think this one should fix the CI

PR #4599, opened 2025-12-01
[main][bugfix] bugfix for qwen3 moe quantization

kunpengW-code:
Fix the issue where the qwen3 moe service cannot be started due to upgrading the vllm version

Error info:
AttributeError: 'AscendFusedMoE' object has no attribute 'use dp chunking'

no

- vLLM version: v0.11.2


vllm_ascend/quantization/quant_config.py
```diff
@@ -408,11 +408,10 @@ class AscendFusedMoEMethod(FusedMoEMethodBase):
...
     """
 
-    def __init__(self,
-                 quant_config: AscendQuantConfig,
-                 prefix: str,
-                 packed_modules_mapping: Dict[str, Any],
-                 layer: torch.nn.Module = None):
+    def __init__(self, quant_config: AscendQuantConfig, prefix: str,
+                 packed_modules_mapping: Dict[str,
+                                              Any], layer: torch.nn.Module):
```
Sparkheart, line 413: Suggest formatting it.

PR #4612, opened 2025-12-02
[Doc]clean up ascend scheduler config from doc

wangxiyuan:
clean up ascend scheduler config from doc

- vLLM version: v0.11.2


MengqingCao:
Do these cases changed need to be test locally?

wangxiyuan:
@MengqingCao I think no, since we have drop the config from test already and the test looks fine.

MengqingCao (approved):

MengqingCao:
> @MengqingCao I think no, since we have drop the config from test already and the test looks fine.

Okay, make sense

mitseng:
Was ascend scheduler deprecated then?

PR #4617, opened 2025-12-02
fix transformer version dependency

shaopeng-666:

- vLLM version: v0.11.2


wangxiyuan:
out of date

PR #4630, opened 2025-12-02
Optimization for NPU small-batch padding in fused_moe.prepare_finalize.py

HJ92-hub:
Optimization for NPU small-batch padding in fused_moe.prepare_finalize.py

* Problem:
In small batch scenarios (specifically decoding phase with BatchSize=1 and TP=8), the forward_context.padded_num_tokens is usually fixed at 8 while the actual num_tokens is 1.
The original implementation uses torch.nn.functional.pad to pad the tensor. On Ascend NPU (910C), this specific pad operation for small shapes introduces unexpected high latency (>50us), which is significant in the decoding loop. This overhead is likely due to the pad operator falling back to AICPU execution or triggering unnecessary format conversions.
* Solution:
This PR optimizes the padding logic for short sequences (Decoding phase) by introducing a buffer + copy mechanism:
Threshold Control: A PAD_CACHE_THRESHOLD = 128 is added. The optimization only applies when target_pad_length <= 128 to target the decoding bottleneck without affecting large-shape prefill performance.
Efficient Ops: Replaces nn.functional.pad with torch.zeros (allocation) + tensor.copy_ (data transfer).
It creates a zero-filled tensor of the target shape.
Copies the valid tokens into the buffer using copy_.
* Benefits:
Performance: copy_ and zeros are highly optimized AICore operations on Ascend, bypassing the AICPU overhead associated with pad.
Latency: Reduces the padding latency from ~55us to <10us for BS=1, TP=8 scenarios.

NA

vLLM version: v0.11.0
vLLM main: https://github.com/vllm-project/vllm/commit/v0.11.0

**TestCase**
* Scenario: A3, Decoding phase, TP=8, DP=2, EP=16, Batch Size=1 (num_tokens=1, padded_num_tokens=8).
* Model: GLM 4.5 W8A8
* Performance Result: End-to-end decoding latency (TPOT) for MoE models (TP=8, BS=1) improved from 51ms to 46ms(~10% improvement).

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/ops/fused_moe/prepare_finalize.py
+++ b/vllm_ascend/ops/fused_moe/prepare_finalize.py
@@ -36,7 +36,10 @@ from vllm_ascend.utils import enable_sp, prefill_context_parallel_enable
 PAD_CACHE_THRESHOLD = 128  
 
 
-def fast_pad_with_buffer(buffer_ref: torch.Tensor | None, target_pad_len: int, data_ref: torch.Tensor):
+def fast_pad_with_buffer(buffer_ref: torch.Tensor | None, 
+                         target_pad_len: int, 
+                         data_ref: torch.Tensor,
+                         last_num_tokens: int):
     """
     Optimization for NPU small-batch padding:
     We cache a small buffer to avoid overhead from nn.functional.pad (~50us)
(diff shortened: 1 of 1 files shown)


wangxiyuan:
Please rebase and fix the merge conflicts

wangxiyuan:

PR #4638, opened 2025-12-02
[Bugfix] Resolve the interface compatibility issue of get_input_embeddings in MM

Levi-JQ:

Resolve the interface compatibility issue of get_input_embeddings in MM， because the get_input_embeddings func of other model does not have the is_multimodal parameter

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -19,7 +19,6 @@
 
 import copy
 import gc
-import inspect
 import itertools
 import math
 import re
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1394,11 +1395,19 @@ def _prepare_inputs(
             # embeddings), we always use embeddings (rather than token ids)
             # as input to the multimodal model, even when the input is text.
             input_ids = self.input_ids[:total_num_scheduled_tokens]
-            inputs_embeds = self.model.get_input_embeddings(
-                input_ids,
-                multimodal_embeddings=mm_embeds,
-                is_multimodal=is_mm_embed,
-            )
+            has_is_multimodal = 'is_multimodal' in inspect.signature(self.model.get_input_embeddings).parameters
```
MengqingCao: I have checked that `is_multimodal` is not included in vLLM v0.11.0, thus only the custom Qwen2_5_vl in vllm-ascend have this arg. Maybe we can just check the `model_type`

Levi-JQ: ok

MengqingCao:
I think this pr is okay to merge as the broken test case tests/e2e/singlecard/test_vlm.py::test_multimodal_audio has already passed. https://github.com/vllm-project/vllm-ascend/actions/runs/19857925293/job/56900838462?pr=4638

PR #4654, opened 2025-12-03
[Ops][Triton]Add qk_rmsnorm_triton for qwen3-next

InSec:
- Add the implementation of qk_rmsnorm_triton. The qk_rmsnorm_triton_kernel adapts to common cases and the qk_rmsnorm_gate_triton_kernel adapts to the qwen3-next model which has a gating mechanism in its attention part. 
- This PR will apply a patch to modify the forward method of Qwen3NextAttention.

N/A

- vLLM version: v0.12.0


vllm_ascend/patch/worker/patch_qwen3_next.py
```diff
@@ -0,0 +1,54 @@
...
+    attn_output = self.attn(q, k, v)
+
+    if self.attn_output_gate:
+        gate = torch.sigmoid(gate)
+        attn_output = attn_output * gate
+
+    output[:], _ = self.o_proj(attn_output)
+
+
+Qwen3NextAttention.forward = forward
```
SlightwindSec, line 54: Please refrain from monkey patching (`Qwen3NextAttention.forward = ...`) as it makes the code fragile and hard to track. We should likely use inheritance here. Also, nit: please add a trailing newline at the end of the file.

PR #4655, opened 2025-12-03
[Bugfix] fix the incorrect use of python's sum on tensors.

wangx700:
Fix the incorrect use of python's sum function on PyTorch tensors.
1. Using Python's sum() function on a tensor self.num_pcp_pads resulted in 6ms execution time
Optimization: replacing with PyTorch's torch.sum() reduced execution time to 474µs
2. scheduler_output.scheduled_spec_decode_tokens undergoes repeated loop processing even when speculative decoding is not used

Optimization: added conditional logic to skip processing loops when speculative decoding is disabled, eliminating unnecessary computational overhead.

- vLLM version: 86e178f7c4d8c3b0eaf3c8e3f810a83f63b90e24
- vLLM main: https://github.com/vllm-project/vllm/commit/86e178f7c4d8c3b0eaf3c8e3f810a83f63b90e24


weijinqian0 (approved):

zhenwenqi2024:
I think it is ok,

wangxiyuan:
let's wait for CI.

PR #4673, opened 2025-12-03
[BugFix][Triton] Fix ub overflow bug of sample_recover_tokens_kernel

whx-sjtu:
Original `sample_recover_tokens_kernel` of reject sampler didn't tile the vocab size dim, whitch will cause ub overflow problem for models with big vocab size like deepseek. This PR adds tiling to the vocab size dim to avoid this problem.

Note that currently we just use a emperical `SUB_BLOCK_SIZE` of `4*1024` for functionality. If in the future this kernel becomes performance bottle neck, we can use triton autotune to optimize this. What's more, we have to disable multibuffer of this kernel due to some accuracy issues. 

None

I will add new ut later.

- vLLM version: v0.12.0


vllm_ascend/sample/rejection_sampler.py
```diff
@@ -698,49 +701,84 @@ def sample_recovered_tokens_kernel(
...
     PADDED_VOCAB_SIZE: tl.constexpr,
     NO_DRAFT_PROBS: tl.constexpr,
+    SUB_BLOCK: tl.constexpr,
 ):
     req_idx = tl.program_id(0)
-    start_idx = 0 if req_idx == 0 else tl.load(cu_num_draft_tokens_ptr +
-                                               req_idx - 1)
+    if req_idx == 0:
+        start_idx = 0
+    else:
```
Sparkheart: Ternary expressions merge branches, eliminate redundant branches, and improve readability.

whx-sjtu: done

MengqingCao (approved):
LGTM

MengqingCao:
All CI passed in https://github.com/vllm-project/vllm-ascend/actions/runs/19927948151 already

PR #4676, opened 2025-12-03
[main][preformance] optimization of kimi-k2 in torchair

Levi-JQ:

In cann8.3， npu_moe_gating_top_k operator can support expert nums with 384, so kimi can use the operator to get better preformance.

- vLLM version: v0.12.0


wangxiyuan:

PR #4690, opened 2025-12-04
[Misc] Upgrade vllm vllm commit to 2025_12_04

Potabk:
As title shows, upgrade vllm commit hash to `ad32e3e`

- vLLM version: v0.12.0
- vLLM main: https://github.com/vllm-project/vllm/commit/v0.12.0


wangxiyuan:
please update version policy doc as well

Potabk:
yes

PR #4708, opened 2025-12-04
fix synchronize error of exceeds_max_model_len d2h copy

Ronald1995:
there is d2h copy blocking cpu operations in mtp propose method, which make host bound issue. this pr refactor it and use cpu tensor to implement it.

no

vllm main f5d3d93c40417c296c20dc301100e55708a17f3f

- vLLM version: v0.12.0

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -103,6 +103,7 @@ jobs:
           pytest -sv tests/e2e/singlecard/test_sampler.py
           pytest -sv tests/e2e/singlecard/test_vlm.py
           pytest -sv tests/e2e/singlecard/multi-modal/test_internvl.py
+          pytest -sv tests/e2e/singlecard/test_xlite.py
 
           # ------------------------------------ v1 spec decode test ------------------------------------ #
           pytest -sv tests/e2e/singlecard/spec_decode_v1/test_v1_mtp_correctness.py
--- a/docs/source/tutorials/DeepSeek-V3.1.md
+++ b/docs/source/tutorials/DeepSeek-V3.1.md
@@ -430,7 +430,6 @@ vllm serve /weights/DeepSeek-V3.1_w8a8mix_mtp \
   "engine_id": "0",
   "kv_connector_module_path": "vllm_ascend.distributed.mooncake_connector",
   "kv_connector_extra_config": {
-            "use_ascend_direct": true,
             "prefill": {
                     "dp_size": 2,
                     "tp_size": 8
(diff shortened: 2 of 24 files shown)


vllm_ascend/spec_decode/mtp_proposer.py
```diff
@@ -854,10 +854,9 @@ def _propose(
             attn_metadata_i.seq_lens[:batch_size] += 1
             # For the requests that exceed the max model length, we set the
             # sequence length to 1 to minimize their overheads in attention.
-            exceeds_max_model_len_cpu = exceeds_max_model_len.to(
-                attn_metadata_i.seq_lens.device, non_blocking=False)
-            attn_metadata_i.seq_lens[:batch_size].masked_fill_(
-                exceeds_max_model_len_cpu, 1)
+            exceeds_mask = attn_metadata_i.seq_lens[:batch_size] > \
+                self.runner.model_config.max_model_len
```
jianzs, line 889: Why not just use 'exceeds_max_model_len'? In my opinion, your change replaces `exceeds_max_model_len = positions[:batch_size] >= self.runner.model_config.max_model_len` with `exceeds_mask = attn_metadata_i.seq_lens[:batch_size] > self.runner.model_config.max_model_len`. Are these two interchangeable?

Ronald1995: yes, you are right, they actually have same logic.but if we use exceeds_max_model_len, we need to make a d2h copy and incur cpu blocked, this will make a hostbound problem. if we enable async_scheduling, the cpu will be blocked here and lead to a performance degrade

jianzs: Why does `exceeds_max_model_len` need CPU synchronization? It's not a cpu tensor. Also, are the results of `exceeds_max_model_len` and `exceeds_mask` always identical? They seem to be constructed differently.

Ronald1995: `exceeds_max_model_len_cpu = exceeds_max_model_len.to(
                attn_metadata_i.seq_lens.device, non_blocking=False)` non_blocking is set to Fasle. exceeds_max_model_len and exceeds_mask  are both to handle the edge case that seq_len is bigger than max_model_len.

jianzs: Why can't we simply use `exceeds_max_model_len` instead of `exceeds_max_model_len_cpu`?

Ronald1995: because attn_metadata_i.seq_lens is a cpu tensor, it can't use a device tensor to mask_fill

jianzs: Oh, I see. Great.

PR #4712, opened 2025-12-04
[Performence] Optimize TTFT in Prefill-Decode Disaggregation for first token return from prefill node

gao12312:

In order solve the issue https://github.com/vllm-project/vllm-ascend/issues/3099 and  optimize the PR https://github.com/vllm-project/vllm-ascend/pull/3656 , we can get the first token from prefill node for early return . After the P node completes the prefill, the first token should be returned directly by the P node without waiting for the D node.

No

With 2P1D (64 910B2, 16 npus each for  P, 32 npus for one D) on deepseek-R1 ，the following result tesed on evalscope:

TTFT: 3 s -> 1.9 s
vLLM version: v0.11.0
vLLM Ascend: v0.11.0rc2

```bash
evalscope perf \
  --url "http://127.0.0.1:8000/v1/chat/completions" \
  --parallel 8 \
  --number 40 \
  --model "deepseek_r1" \
  --log-every-n-query 100 \
  --read-timeout=1200 \
  --max-tokens 2000 \
  --dataset random \
  --max-prompt-length 7000 \
  --min-prompt-length 7000 \
  --tokenizer-path /workspace/DeepSeek-R1-W8A8 \
  --api openai \
  --extra-args '{"ignore_eos": true}' \
  --seed 1042 \
  --stream
```

**baseline：**
3dde

**optimized:**

- vLLM version: v0.12.0
- vLLM main: https://github.com/vllm-project/vllm/commit/v0.12.0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@
 FROM quay.io/ascend/cann:8.5.0-910b-ubuntu22.04-py3.11
 
 ARG PIP_INDEX_URL="https://mirrors.tuna.tsinghua.edu.cn/pypi/web/simple"
-ARG MOONCAKE_TAG="v0.3.8.post1"
+ARG MOONCAKE_TAG="v0.3.7.post2"
 ARG SOC_VERSION="ascend910b1"
 
 # Define environments
--- a/Dockerfile.a3
+++ b/Dockerfile.a3
@@ -18,7 +18,7 @@
 FROM quay.io/ascend/cann:8.5.0-a3-ubuntu22.04-py3.11
 
 ARG PIP_INDEX_URL="https://mirrors.tuna.tsinghua.edu.cn/pypi/web/simple"
-ARG MOONCAKE_TAG=v0.3.8.post1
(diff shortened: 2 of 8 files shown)


LCAIZJ:
"How is the improvement from 3s to 1.9s in TTFT achieved? The transmission of KV cache between PDs takes about 100ms, plus the TPOT time—there shouldn't theoretically be a 1.1s reduction."

wangxiyuan:
@gao12312 can you reply @LCAIZJ 's review? Thanks

examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
```diff
@@ -626,7 +668,7 @@ async def generate_stream():
                                         instance_info.decoder_score)
 
         return StreamingResponse(generate_stream(),
-                                 media_type="application/json")
+                                 media_type="text/event-stream")
```
zzzzwwjj: Why we need to modify this param?

PR #4745, opened 2025-12-05
[Image] Correcting the vllm tag of the openeuler image on the A2 device.

Potabk:

Corrected the vllm tag, which should have been in v0.11.0


Potabk:
Sorry for everybody, my fault!

PR #4751, opened 2025-12-05
[Bugfix] Disable the dispatch_ffn_combine kernel in MTP path

kiscad:

This PR is to fix a smoking test failure. Adjust mtp_proposer and model_runner_v1 to route MTP decoding through the non‑fused MoE implementation while keeping the overall inference flow unchanged.

This PR will be tested in smoking tests.

- vLLM version: v0.12.0


zzzzwwjj:
1. moe_comm_type judgement conditions need to consider whether is quant case;
2. dispatch_ffn_combine op need support EP>16 case;
This two problems need to finish afterwards.

PR #4754, opened 2025-12-05
[Bugfix] Prevent engine hang during KVCacheSendingThread startup

jianzs:

Previously, if the KVCacheSendingThread couldn't create a socket because of port conflicts or other problems, the main thread would wait endlessly for the ready_event signal, causing the entire engine initialization to freeze. This update fixes the issue by adding timeouts for thread startup and handling unexpected thread exits, so the initialization process no longer gets stuck indefinitely.

No

- vLLM version: v0.12.0


wangxiyuan (approved):
@liziyu179 @LCAIZJ

liziyu179:
> @liziyu179 @LCAIZJ

Looks good to me

PR #4755, opened 2025-12-05
[CI] Fix ngram & suffix test oom

fluctlux:
Avoid oom during CI by using `with VllmRunner` instead of `LLM()`, and enable `test_ngram_correctness`

Before:
image
After:
image

CI passed.

- vLLM version: v0.12.0


wangxiyuan (approved):
Hope the test can works now

wangxiyuan:
really nice change. Thanks very much！！

PR #4761, opened 2025-12-06
[KVPOOl]Support pp

baxingpiaochong:
Support pp for kv pool

- vLLM version: v0.12.0


LCAIZJ:
When pipeline parallelism (PP) is enabled, `self.maybe_wait_for_kv_save()` cannot remain in its current position and must be moved back to its previous location. Additionally, we need to synchronize with the ADXL version that includes the fix for the hang issue.

LCAIZJ:
The ADXL-related fixes will be included in 8.5.RC1.

PR #4762, opened 2025-12-06
Adapt dispatch_ffn_combine for decoding

kiscad:

- Extend the `dispatch_ffn_combine` host and kernel implementations to support the decoding phase, including updated tiling, epilogue, and quantized MoE routing handling.
- Introduce and refine HCCL shared‑memory utilities (`HcclShmem`) for cross‑rank synchronization and safe shared‑memory access in multi‑rank decode.
- Wire the new `dispatch_ffn_combine` decode path into the Ascend runtime (forward context, fused MoE communication, W8A8 quantization, MTP proposer, and model_runner_v1).
- Fix rank initialization and shared‑memory size/offset handling in the FFN dispatch headers to avoid crashes and incorrect memory access during decoding.

- vLLM version: v0.12.0


wangxiyuan:

PR #4765, opened 2025-12-07
[Attention] Temporarily add back pa for small batch sizes.

whx-sjtu:
This PR adds back pa in scenarios of small batch sizes due to performance consideration. Will remove pa once fia performs better than pa in all scenarios.

No

CI passed with existing test.

- vLLM version: v0.12.0


vllm_ascend/attention/attention_v1.py
```diff
@@ -660,9 +722,14 @@ def forward_impl(
                 output = self._forward_prefill(query, key, value,
                                                attn_metadata, output)
         else:
-            attn_output, num_tokens = self.full_graph_attention(
-                query, key, value, attn_metadata, output)
-            output[:num_tokens] = attn_output[:num_tokens]
+            num_tokens = query.shape[0]
+            if num_tokens in PAGED_ATTENTION_LIST:
```
weijinqian0: The code that determines whether to use PA or FIA based on the BS size can only take effect in full graph mode. Is this as expected?

whx-sjtu: Yes. After all this will be deleted in the near future. So I only add back pa for full graph mode to improve performance.

vllm_ascend/attention/attention_v1.py
```diff
@@ -660,9 +722,14 @@ def forward_impl(
                 output = self._forward_prefill(query, key, value,
                                                attn_metadata, output)
         else:
-            attn_output, num_tokens = self.full_graph_attention(
-                query, key, value, attn_metadata, output)
-            output[:num_tokens] = attn_output[:num_tokens]
+            num_tokens = query.shape[0]
```
yiz-liu, line 766: Use `num_input_tokens` instead, otherwise multimodal models may encounter problems.

vllm_ascend/compilation/acl_graph.py
```diff
@@ -193,7 +194,65 @@ def __call__(self, *args, **kwargs):
...
+                key_cache,
+                value_cache,
+                num_kv_heads,
+                num_heads,
+                scale,
+                block_table,
+                seq_lens,
+                output,
+            ) = param
+            seq_lens = forward_context.attn_metadata[key].seq_lens
```
yiz-liu, line 220: You sure this is `seq_lens` rather than `seq_lens_cpu` or `seq_lens_list` right?

yiz-liu (commented):
One last question, when will we finally remove this PA?

whx-sjtu:
> One last question, when will we finally remove this PA?

After 1230, fia will support flash decoding. Then all GQA models will do performance tests of different scenarios using new fia. If results show no performances problem, I will remove pa finally.

vllm_ascend/attention/attention_v1.py
whx-sjtu: Can't obtain `num_input_tokens` in attention forward. What exactly are you worried about?

vllm_ascend/compilation/acl_graph.py
whx-sjtu: The `seq_lens` here is a cpu tensor, what pa needed.

vllm_ascend/attention/attention_v1.py
yiz-liu: I thought about it, it's probably OK here.

yiz-liu (approved):

PR #4820, opened 2025-12-09
[CI] Skip `test_suffix_correctness`

Potabk:
Currently, suffix decoding has known correctness issue see https://github.com/vllm-project/vllm-ascend/actions/runs/20033509824/job/57457565620?pr=4781"

- vLLM version: v0.12.0

### DIFF
--- a/tests/e2e/singlecard/spec_decode_v1/test_v1_spec_decode.py
+++ b/tests/e2e/singlecard/spec_decode_v1/test_v1_spec_decode.py
@@ -151,6 +151,8 @@ def test_eagle_correctness(
     assert matches > int(0.66 * len(ref_outputs))
 
 
+@pytest.mark.skip(
+    "Fix me, suffix decoding now exists some known accuracy issue, skip it")
 def test_suffix_correctness(
     test_prompts: list[list[dict[str, Any]]],
     sampling_config: SamplingParams,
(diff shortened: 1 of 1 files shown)


.github/workflows/_e2e_test.yaml
```diff
@@ -108,7 +109,6 @@ jobs:
           # ------------------------------------ v1 spec decode test ------------------------------------ #
           pytest -sv tests/e2e/singlecard/spec_decode_v1/test_v1_mtp_correctness.py
           pytest -sv tests/e2e/singlecard/spec_decode_v1/test_v1_mtp_torchair_correctness.py
-          pytest -sv tests/e2e/singlecard/spec_decode_v1/test_v1_spec_decode.py
```
MengqingCao, line 111: why move this here? I think it's better to keep it under `v1 spec decode test` header

vllm_ascend/utils.py
```diff
@@ -487,9 +487,10 @@ def update_aclgraph_sizes(vllm_config: VllmConfig) -> None:
 
     # Calculate maximum supported batch sizes considering model architecture
     resources_per_graph = num_hidden_layers + 1
-    if vllm_config.speculative_config is not None:
-        draft_model_hf_config = vllm_config.speculative_config.draft_model_config.hf_config
-        resources_per_graph += draft_model_hf_config.num_hidden_layers + 1
+    # For suffix decoding, use the suffix path when no draft_model_config is provided.
+    if (spec := vllm_config.speculative_config) and \
```
MengqingCao: I think we can do this fix in another pr

.github/workflows/_e2e_test.yaml
Potabk: I think this approach is reasonable. From the perspective of timely problem exposure, we need to conduct diverse tests as early as possible in the CI testing phase to expose existing problems as quickly as possible, thereby reducing CI waiting time.

PR #4843, opened 2025-12-09
[Fix] Add extra warmup run count for MC2 on specific SoC version

yiz-liu:
We didn’t account for this earlier because we didn’t have A3 in CI, but now that we do, this test case needs a few extra tweaks — please take a look at `profile_run`.

None.

None.

- vLLM version: v0.12.0


yiz-liu:
https://github.com/vllm-project/vllm-ascend/actions/runs/20064208752?pr=4844, This PR is validated, please merge it. @wangxiyuan @MengqingCao

PR #4845, opened 2025-12-09
[Feature] Training-Inferencing ops matching

YuhanBai:
To matching up operators with training side, we replace the following operators with operators on right-hand side:

- AddRmsNorm --> add + rmsnorm
- MoeGatingTopKSoftmax  -> softmax + topk + etc.
- UnpadFlashAttentionBF16NdKernel -> FA

Since this replacement might cause potential performance drop, we introduce a switch `TRAIN_INFER_MATCHING` to control whether we want to enable this replacement. The switch is default set to **OFF**

**YES**, added a switch `TRAIN_INFER_MATCHING` (default is **OFF**) to control whether we want to enable the training-inferencing operator matching.

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -514,7 +514,8 @@ class AscendAttentionBackendImpl(AttentionImpl):
                 shape_order = "TND"
                 scale = self.scale
                 mask = attn_metadata.attn_mask
-                mask = mask.to(torch.uint8)
+                if mask is not None:
+                    mask = mask.to(torch.uint8)
                 attn_output = torch_npu.npu_fusion_attention(
                     query=query,
                     key=key,


vllm_ascend/attention/attention_v1.py
```diff
@@ -504,6 +505,36 @@ def _forward_prefill(self, query: torch.Tensor, key: torch.Tensor,
...
+                    scale=scale,
+                    inner_precise=0,
+                    sparse_mode=1,
+                    actual_seq_qlen=attn_metadata.actual_seq_lengths_q,
+                    actual_seq_kvlen=attn_metadata.actual_seq_lengths_q,
+                )[0]
+                attn_output = attn_output.view(num_tokens, self.num_heads,
+                                               self.head_size)
+                output[:num_tokens] = attn_output[:num_tokens]
+                assert output is not None
```
Sparkheart: It is recommended to provide explanations after an assertion failure to enhance understandability.

vllm_ascend/ops/layernorm.py
```diff
@@ -21,6 +21,8 @@
 from vllm.config import get_current_vllm_config
 from vllm.model_executor.layers.layernorm import GemmaRMSNorm, RMSNorm
 
+import vllm_ascend.envs as envs_ascend
```
Sparkheart, line 24: It is recommended to place import before from.

vllm_ascend/attention/attention_v1.py
YuhanBai: Will add some explanations in next bug-fix commit

vllm_ascend/ops/layernorm.py
YuhanBai: I think it should be ok, since we are importing things from vllm_ascend after import form vllm

wangxiyuan:

PR #4864, opened 2025-12-10
Remove COMPILE_CUSTOM_KERNELS env

wangxiyuan:
With more and more custom ops merged, disable `COMPILE_CUSTOM_KERNELS ` for vllm ascend seems useless now. Let's enable csrc compile by default.

- vLLM version: v0.12.0


Yikun:
It's better to mention this removal  and note for how to enable sleep mode now in release note anf commit msg.

PR #4868, opened 2025-12-10
[feat] apply flashcomm1 on bailing

hwhaokun:
This PR adjusts the layer prefix matching rules for tensor parallelism (column/row parallel ops) to fit Bailing model's naming conventions (adding "query_key_value" for column parallel and "attention.dense" for row parallel), enabling flashcomm1 to work properly on the Bailing model.

No

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/ops/linear_op.py
+++ b/vllm_ascend/ops/linear_op.py
@@ -610,12 +610,16 @@ def _get_column_parallel_op(
     if enable_sp():
         if "shared_expert" in prefix:
             return None
-        if "gate_up_proj" in prefix:
-            return SequenceColumnParallelOp(layer)
-        if "in_proj" in prefix:
-            return SequenceColumnParallelOp(layer)
-        if "qkv_proj" in prefix or "conv1d" in prefix:
-            return SequenceColumnParallelOp(layer)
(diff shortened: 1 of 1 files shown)


vllm_ascend/ops/linear_op.py
```diff
@@ -614,7 +614,7 @@ def _get_column_parallel_op(
             return SequenceColumnParallelOp(layer)
         if "in_proj" in prefix:
             return SequenceColumnParallelOp(layer)
-        if "qkv_proj" in prefix or "conv1d" in prefix:
+        if "qkv_proj" in prefix or "conv1d" in prefix or "query_key_value" in prefix:
```
realliujiaxu: could you refactor code here, and add comment for each prefix? for example:
```python
sp_prefix = [
    "gate_up_proj", # MLP of mostLLM
    "in_proj", # gated delta net of qwen3 next
    "qkv_proj", # qkv linear of most LLM
    "conv1d", # gated delta net of qwen3 next
    "query_key_value", # qkv linear of Bailing
]
for a_prefix in sp_prefix:
    if a_prefix in prefix:
        return SequenceColumnParallelOp(layer)
```

PR #4878, opened 2025-12-10
[refactor] refactor weight trans nz and transpose

zzzzwwjj:

Now `VLLM_ASCEND_ENABLE_NZ` will have three options:
0: disable nz;
1: only quant case enable nz;
2: enable nz as long as possible;

And `VLLM_ASCEND_ENABLE_NZ`=1 by default.

All cases are shown in the table below:

|  | W4A4 | W4A8 | W8A8 | fp16/bf16 | fp32 |
|---|---|---|---|---|---|
| trans nz | can't support nz | trans nz by default | trans nz by default | trans nz when VLLM_ASCEND_ENABLE_NZ is 2 | can't support nz |
| transpose | only support not transpose case | only support transpose case | only support transpose case | linear: only support not transpose case<br>gmm: only support transpose case | same to fp16/bf16 |

Some exceptional cases:
1. MLAPO op need to do some additional processing on the weights, including trans nz. If use MLAPO op, some weight will be transformed to nz forcely;
2. MLA/SFA's weight `W_UV` will be used by op `torch.ops._C_ascend.batch_matmul_transpose`, and this op can't support nz currently;

Now fp16/bf16 weight will not trans nz by default.

- vLLM version: v0.12.0

### DIFF
--- a/.github/workflows/_e2e_nightly_single_node.yaml
+++ b/.github/workflows/_e2e_nightly_single_node.yaml
@@ -54,6 +54,7 @@ jobs:
   e2e-nightly:
     name: ${{ inputs.tests }}
     runs-on: ${{ inputs.runner }}
+    timeout-minutes: 600
     container:
       image: ${{ inputs.image }}
     steps:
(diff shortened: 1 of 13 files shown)


yiz-liu:
Maybe we should change it to `VLLM_ASCEND_FP16_ENABLE_NZ` and default it to 0. Also leave it for training environment to disable all "cast tensor format to NZ" operations.

vllm_ascend/quantization/w8a8_dynamic.py
```diff
@@ -268,14 +263,12 @@ def apply(
...
-                1, 2).contiguous()
-            layer.w2_weight.data = layer.w2_weight.data.transpose(
-                1, 2).contiguous()
-        if is_enable_nz():
-            torch_npu.npu_format_cast_(layer.w13_weight, ACL_FORMAT_FRACTAL_NZ)
-            torch_npu.npu_format_cast_(layer.w2_weight, ACL_FORMAT_FRACTAL_NZ)
+        layer.w13_weight.data = layer.w13_weight.data.transpose(
+            1, 2).contiguous()
+        layer.w2_weight.data = layer.w2_weight.data.transpose(1,
+                                                              2).contiguous()
```
wangxiyuan, line 271: https://github.com/vllm-project/vllm-ascend/commit/01a13a9b770093f54095b7aa425b7d21bd8bbaa5

PR #4893, opened 2025-12-10
[Feat] Support async_scheduler and disable_padded_drafter_batch in eagle

anon189Ty:
We refactored the eagle_proposer.py to adapt the framework of eagle.py in vllm-v0.12.0, to support the logit of padded drafter batch and async-scheduler.

- vLLM version: v0.12.0


vllm_ascend/attention/utils.py
```diff
@@ -107,6 +107,36 @@ class AscendCommonAttentionMetadata:
...
+            slot_mapping=self.slot_mapping[:num_actual_tokens],
+            actual_seq_lengths_q=self.actual_seq_lengths_q[:num_actual_tokens],
+            positions=self.positions[:num_actual_tokens],
+            attn_mask=self.attn_mask,
+            spec_attn_mask=self.spec_attn_mask,
+            attn_state=self.attn_state,
+            is_only_prefill=self.is_only_prefill,
+            graph_pad_size=-1,  # It should be -1 when not run in fullgraph mode.
+            num_input_tokens=num_actual_tokens,
+            cos=self.cos,
```
whx-sjtu: check whether cos/sin needed to be sliced.

anon189Ty: Checked. In the current scene, cos/sin will only be None, so won't cause any errors. Slicing has been added for subsequent scenes.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1824,7 +1830,7 @@ def propose_draft_token_ids(sampled_token_ids):
         with ProfileExecuteDuration().capture_async("Draft"):
             if self.speculative_config:
                 use_padded_batch_for_eagle = self.speculative_config and \
-                    self.speculative_config.method == "mtp" and \
+                    self.speculative_config.method in ("mtp", "eagle", "eagle3") and \
```
realliujiaxu: use `self.speculative_config.use_eagle()`

anon189Ty: Done.

wangxiyuan (approved):

tests/e2e/singlecard/spec_decode_v1/test_v1_spec_decode.py
```diff
@@ -115,41 +116,65 @@ def test_eagle_correctness(
...
+    ]
+
+    sampling_params = SamplingParams(
+        max_tokens=300,
+        temperature=0.0,
+        ignore_eos=False,
+    )
+
+    # Create an LLM.
+    llm = LLM(
```
MengqingCao, line 153: I just noticed you're using LLM in a test case, we must clean up npu hbm when not using `VllmRunner`. plz modify this case to use `VllmRunner` or clean up by hand if keeping it as is.
plz refer to https://github.com/vllm-project/vllm-ascend/blob/main/tests/e2e/conftest.py#L81

PR #203, opened 2025-02-28
[v0.7.3] Recover vllm-ascend dev image

Yikun:
### What this PR does / why we need it?
Recover vllm-ascend dev image by soft revert:
- https://github.com/vllm-project/vllm-ascend/commit/701a2870469d8849a50378f9450dc3e851c8af20
- https://github.com/vllm-project/vllm-ascend/commit/aba0b839b96cd453450904508bc245c6670cfbbe

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?
CI passed

PR #4894, opened 2025-12-10
[Performance]: adapt community's dbo to vllm-ascend

zxdukki:

We noted that vLLM has supported dbo in pr : [vLLM #23693](https://github.com/vllm-project/vllm/pull/23693).
However, it is mainly designed for gpu platform. Besides, it is only supported with DeepEP and DP+EP deployments, which cannot directly used in vllm-ascend.
In this PR, we attempt to adapt and extend the capability of the dbo feature to vllm-ascend.

Compared with the previous version of dbo in vllm-ascend and the current version in vLLM, some major changes are listed below:

- move dbo from model layer to model_runner layer to improve generality
- support both token-level and request-level microbatch splitting
- support multiple parallelism policies including TP/DP/PP + EP
- support multiple moe comm methods including AllGather and AlltoAll
- support multiple models including deepseek and qwen3 dense/moe
- support modifying aic/aiv core num occupied by comm/comp kernels (using `with create_core_control_context(aic_core, aiv_core)`, or we can set core_limit num in `dbo_wait_current_stream_and_yield` and `dbo_record_current_stream` using `VLLM_ASCEND_DBO_COMM_AIC_NUM` and `VLLM_ASCEND_DBO_COMM_AIV_NUM`. We found that the vector core assigned to HCCL should greater than 16 or HCCL kernel may trigger exceptions.
- （new）add agent skills to automatically generate dbo templates for new models. We have tested it on glm4.7 and qwen3 235B / 32B, it can achieve good overlap performance. Also, we can use profiling analysis skill to analyze npu profiling traces to find the opportunity for overlapping.

We also optimize the overlap strategy for comp/comm kernels to achieve better performance. Performance improvements in some scenarios are listed below:

Note that some results below maybe out of date since most of them are tested under v0.15.0. But at the current version(v0.18.0), we can still observe performance gains for prefill phase in models including deepseek and glm 4.7/5.0.

Prefill qps for **Qwen3 moe** under 1000 prompts(4k) + inf rate on 910B: 

- Allgather + flashcomm1 + tp16: **4.92-> 6.07**

image

image

- Allgather + flashcomm1 + flashcomm2 + tp16: **5.53-> 6.37**

> We use the core limit api to use aic core num = 4 for the communication block and aic core num = 20 for the computation block here. We dont set aiv core in this scenario since HCCL does not use AIV kernel in tp16.

image

| aic core num for comm block  | aic core num for comp block  | prefill qps(qwen3 dbo + fc1 + fc2) |
| ------------- | ------------- | ------------- |
| 1  | 23 |5.09  |
| 2  | 22  |5.97  |
| 3  | 21  |6.33  |
| 4  |20  | 6.37  |
| 6  | 18  |6.34  |
| 8  | 16  |5.89  |
| 16  | 8 |4.69  |
| *(not limit core num)  | *  |6.01  |

- Allgather + flashcomm1 + tp8: **4.53->5.03**

Prefill qps for **DeepSeek** under 1000 prompts(4k) + inf rate on 910B: 

> Update at 12/11 based on the commit: 1a443f277 

without dbo / with dbo
- Allgather  + flashcomm1  + tp16:  **4.41 ->5.43**
image

image

- Allgather + flashcomm1 + flashcomm2 + tp16 : **4.78-> 5.63**

image

image

- Allgather + flashcomm1 + dp2tp8 ： **5.20->6.29**
image

image

- Allgather + flashcomm1 + pp2tp8 :   **6.84->7.49**

According to the profiling data, dbo can overlap 60%+ communication time while adding negligible overhead(\i.e., splitting microbatches and executing kernels for two microbatches)

image

Prefill qps for DeepSeek under 1000 prompts(4k) + inf rate on 910C: 
- AlltoAll:

(We will add more performance results for dbo in other scenarios, \i.e., other models, PP parallelism, AlltoAll comm method )

> We observe that enabling `HCCL_OP_EXPANSION_MODE=AIV` results in poor overlap performance  in certain cases. In this mode, the HCCL communication kernels occupy all available AI Vector kernels(?), resulting in significant performance degradation for concurrently executing computation kernels ( resource contention for ai vector ). But dbo can achieve better performance even using `HCCL_OP_EXPANSION_MODE=AI_CPU`. Besides, we also attempt to limit the use of ai vector resources using APIs like ``torch.npu.set_stream_limit()''

Overlap Design:

- Allgather comm method:

> Although there may exist different overlap strategies, we find that the strategy below can achieve good performance in most of the scenarios when using Allgather as the moe comm method.

image

No, users can add ``--enable_dbo'' to the model config to allow using dbo in vllm-ascend as same as in vllm.
Note that the assertion located at `vllm/config/vllm.py:870-876` presents a current blocker for enabling DBO (Distributed Batch Offloading) in vllm-ascend. To proceed with our implementation, we need to temporarily comment out this assert. This is a necessary workaround until a permanent fix can be integrated upstream.

It can work with vllm v0.18.0 now.

TODO list:

- [x] try to set the num vector kernels using by HCCL
- [x] some modifications based on vllm v0.18.0

Suggestions and comments are welcome.

- vllm-ascend: main

- vLLM version: v0.21.0


wangxiyuan:

zxdukki:
> Any progress? If this PR is still alive, please rebase to main and make CI happy, otherwise you can close it. Thanks

Thanks for your interests. This PR is still alive and i will solve the CI errors ASAP.

jiangkuaixue123:
We're working on AFD and have implemented ubatch on vllm-ascend. Since there's some overlap with your team, would you mind sharing a contact? We'd like to set up a quick sync to align on the details. @zxdukki

zxdukki:
> We're working on AFD and have implemented ubatch on vllm-ascend. Since there's some overlap with your team, would you mind sharing a contact? We'd like to set up a quick sync to align on the details. @zxdukki

We also noticed the related AFD work implemented by your team and are glad to sync up the details!  You can reach me at zxdu1997@gmail.com, maybe we can schedule an online meeting to discuss the related PR. @jiangkuaixue123

vllm_ascend/worker/ubatching.py
```diff
@@ -0,0 +1,261 @@
...
+    def _signal_comm_done(self, event=UBatchEventKey.DEFAULT):
+        self.gpu_comm_done_event[event].record(self.comm_stream)
+
+    def _signal_compute_done(self, event=UBatchEventKey.DEFAULT):
+        self.gpu_compute_done_event[event].record(self.compute_stream)
+
+    def _wait_compute_done(self, event=UBatchEventKey.DEFAULT):
+        self.comm_stream.wait_event(self.gpu_compute_done_event[event])
+
+    def _wait_comm_done(self, event=UBatchEventKey.DEFAULT):
```
tangtiangu, line 104: For functions like these that are not directly related to Ascend, could we consider reusing the UBatchContext implementation from vllm?

zxdukki: > For functions like these that are not directly related to Ascend, could we consider reusing the UBatchContext implementation from vllm?

Here we slightly modify the implementation of `_signal/wait_comm/compute_done` in vllm-ascend by adding different events to indicate the phase of overlap, i.e., the communication before/after attention/moe, each use a different npu event. We use a dict storing the related events here.

Normally, dbo should work like this:
image

However, we consider that if a kernel in stream 2 consume extremely long time, dbo thread may repeatly record/wait the same event(if we use the same event here) , which may lead to incorrect comp-comm overlap for dbo.  For example, the first event record will be deprecated if the second event record occurs before the first event wait? 

So we would like to use different events here to avoid the corner case. 
image

wangxiyuan:
thanks for keeping update. we'll review soon.

wangxiyuan:
@zzzzwwjj

jiangkuaixue123:
Thanks for the DBO adaptation work!

I noticed that `split_attn_metadata()` creates per-ubatch `AscendCommonAttentionMetadata`, but it does not seem to preserve or rebuild `prefill_context_parallel_metadata`. Since PCP/DCP attention builders rely on this metadata, I’m wondering whether `enable_dbo` is currently expected to support PCP scenarios.

If PCP is not supported yet, could we add an explicit guard or document the limitation? Also, do you have a plan for supporting `DBO + PCP` in the future, e.g. by slicing/rebuilding `PrefillContextParallelMetadata` per ubatch?

zxdukki:
> I noticed that `split_attn_metadata()` creates per-ubatch `AscendCommonAttentionMetadata`, but it does not seem to preserve or rebuild `prefill_context_parallel_metadata`. Since PCP/DCP attention builders rely on this metadata, I’m wondering whether `enable_dbo` is currently expected to support PCP scenarios.  If PCP is not supported yet, could we add an explicit guard or document the limitation?

Thanks for your comments. As you mentioned above, currently DBO does not split the PCP-related metadata so it cannot support PCP scenarios. I think it's a good idea to add an assert which can prevent users from enabling dbo and PCP simultaneously and I will update it soon.


> Also, do you have a plan for supporting `DBO + PCP` in the future, e.g. by slicing/rebuilding `PrefillContextParallelMetadata` per ubatch?

I have thought about the approach to enable DBO and PCP at the same time before.
I think it is safe to enable `request-level dbo(split microbatch at the edge of requests)` with PCP at the same time. However, the scenario of enabling token-level dbo` and pcp is more complicated. In my opinion, PCP need allgather kv across all the chunks from a request; but dbo usually do communications across the same microbatch's scope(i.e., chunk1+chunk2, or chunk3+chunk4).  This should be carefully considered / modified to avoid accuracy issues.
image

Please let me know if I've misunderstood the background of the relationship between dbo and PCP .

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3249,10 +3294,19 @@ def _build_attn_group_metadata(
...
-                    kv_cache_gid, attn_gid, cm, num_reqs_actual,
-                    prefill_ratio_to_sas_metadata, decode_ratio_to_sas_metadata,
-                    common_ratio_to_sas_metadata)
+                # TODO(zxdu): check new params for dbo logic
+                if ubatch_slices is not None:
+                    for ubid, _cm in enumerate(split_attn_metadata(ubatch_slices, cm)):
+                        _build_attn_group_metadata(
+                            kv_cache_gid, attn_gid, _cm, num_reqs_actual,
+                            prefill_ratio_to_sas_metadata, decode_ratio_to_sas_metadata,
+                            common_ratio_to_sas_metadata, ubid)
```
ShwStone, line 3303: Could we verify metadata ownership across ubatches here? Metadata built for ubatch 0 must remain unchanged after building ubatch 1.

For example, the DSA ratio/SAS cache dictionaries are currently shared by both ubatches, although they contain token-layout-dependent positions, sequence lengths, RoPE tensors, and SAS metadata. This can make ubatch 1 reuse or overwrite data belonging to ubatch 0.

Please audit every attention backend supported by DBO for similar shared mutable caches, module/class-level runtime buffers, or views into reusable storage. We could either:

1. isolate/materialize such data per ubatch in the DBO build path; or
2. refactor the backends so step-dependent mutable state is owned by each builder/ubatch.

It would also be useful to add an invariant test that building ubatch 1 does not mutate metadata previously returned for ubatch 0.

Related cases: [vLLM-Ascend #10825](https://github.com/vllm-project/vllm-ascend/pull/10825#discussion_r3457664346), [AFD Plugin #203](https://github.com/vllm-project/afd-plugin/pull/203#issuecomment-5251516666).

PR #4899, opened 2025-12-11
【Bugfix】bugfix_for_bmm_transpose

xlan-huawei:
The bmm_transpose operator in version 3.2 is only used in the decoding stage due to shape limitations.

- vLLM version: v0.12.0


rjg-lyh:
This PR restricts the operator’s usage scenarios to ensure functionality, performance, and its ability to enter the aclgraph full-decode-only mode, as shown in the diagram below:
before fusion:
image
after fusion:
image

wangxiyuan (approved):

PR #4901, opened 2025-12-11
[CI] speed up ut

wangxiyuan:
avoid model download to speed up ut test. 

- vLLM version: v0.12.0

### DIFF
--- a/tests/e2e/singlecard/test_aclgraph_accuracy.py
+++ b/tests/e2e/singlecard/test_aclgraph_accuracy.py
@@ -192,3 +192,22 @@ def test_output_between_eager_and_full_decode_only(
         name_0="vllm_eager_outputs",
         name_1="vllm_aclgraph_outputs",
     )
+
+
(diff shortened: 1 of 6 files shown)


tests/e2e/singlecard/test_aclgraph_accuracy.py
```diff
@@ -192,3 +195,31 @@ def test_output_between_eager_and_full_decode_only(
         name_0="vllm_eager_outputs",
         name_1="vllm_aclgraph_outputs",
     )
+
+
+def test_aclgraph_enable(self):
```
MengqingCao: not sure how it works, should this be wrapped in a testcase class?

wangxiyuan: yes, my fault here. Not sure how CI passed. LOL

wangxiyuan: I've tested it locally, it works now.

PR #4908, opened 2025-12-11
[Performance] Pre-issued exponential distribution operator.

weijinqian0:
Pre-issued exponential distribution operator.

Result:
Single inference saves 200-300 microseconds.
before：

After：

image

- vLLM version: v0.12.0


vllm_ascend/sample/sampler.py
```diff
@@ -1,13 +1,40 @@
...
 
 
+def random_sample(
+    probs: torch.Tensor,
+    generators: dict[int, torch.Generator],
+) -> torch.Tensor:
+    """Randomly sample from the probabilities.
+
+    We use this function instead of torch.multinomial because torch.multinomial
+    causes CPU-GPU synchronization.
```
whx-sjtu: fix note here: "CPU-GPU"

PR #4916, opened 2025-12-11
[0.11.0][Bugfix] Remove the ZMQ communication setup on the D node

SlightwindSec:
In the PD separation scenario, the D node does not need to perform get operations, and therefore does not need to create ZeroMQ (ZMQ) communication.


wangxiyuan:
please update the commit message and point that which PR is picked from main.

SlightwindSec:
> please update the commit message and point that which PR is picked from main.

done

PR #4925, opened 2025-12-11
[CI] Cancel whl build when submitting a new commit

Potabk:
From a resource-saving perspective, canceling old jobs when submitting new commits can reduce github_hosted in queue

- vLLM version: v0.12.0


wangxiyuan:
This is important, can you rebase and fix it in quick?

Potabk:
yes

PR #4956, opened 2025-12-12
[feat] apply flashcomm1 on multimodal_model

hwhaokun:

- vLLM version: v0.12.0


vllm_ascend/utils.py
```diff
@@ -943,6 +930,17 @@ def is_hierarchical_communication_enabled():
             and os.getenv("HCCL_INTRA_PCIE_ENABLE", "") == "1")
 
 
+def parse_layer_idx(prefix: str) -> Optional[int]:
+    pattern = r'layers\.(\d+)'
+    match = re.search(pattern, prefix)
+    if match:
+        layer_idx = int(match.group(1))
```
zzhx1: You do not need to create a new function `parse layer idx`, you can directly use the function `extract_layer_index` provided by vllm.

```
from vllm.model_executor.models.utils import extract_layer_index
```

vllm_ascend/ops/register_custom_ops.py
```diff
@@ -40,11 +40,14 @@ def _maybe_chunk_residual_impl(x: torch.Tensor,
 def _maybe_all_gather_and_maybe_unpad_impl(
         x: torch.Tensor,
         label: bool,
-        is_ep_comm: bool = False) -> torch.Tensor:
+        is_ep_comm: bool = False,
+        is_first_allgather: bool = None) -> torch.Tensor:
```
zzhx1:
```
is_first_allgather: bool = False) -> torch.Tensor:
```

PR #4963, opened 2025-12-12
[Fix] Fixes issues in MTP with async scheduling and ACL graph

yiz-liu:
Corrects attention metadata size for MTP when both asynchronous scheduling and full ACL graph mode are enabled. This prevents potential size mismatches during execution.

Additionally, improves the robustness of calculating token sample indices by explicitly aligning tensor shapes.

Finally, prevents padding when the number of input tokens exceeds the maximum ACL graph batch size to avoid out-of-bounds errors.

None.

Need to add corresponding test case ASAP.
- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -1019,7 +1019,7 @@ class NPUModelRunner(GPUModelRunner):
             # TODO: We should make this official ASAP. Also note that if we pad here,
             # the builders won’t need to add any extra padding.
             if self.compilation_config.cudagraph_mode.decode_mode() == CUDAGraphMode.FULL and \
-                uniform_decode and num_input_tokens <= self.aclgraph_batch_sizes[-1]:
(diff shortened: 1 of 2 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -992,7 +992,7 @@ def _prepare_inputs(
             # TODO: We should make this official ASAP. Also note that if we pad here,
             # the builders won’t need to add any extra padding.
             if self.compilation_config.cudagraph_mode.decode_mode() == CUDAGraphMode.FULL and \
-                uniform_decode:
+                uniform_decode and num_input_tokens <= self.aclgraph_batch_sizes[-1]:
```
jianzs:
```suggestion
                uniform_decode and num_input_tokens <= self.cudagraph_batch_sizes[-1]:
```

yiz-liu: OK, seen that refactor PR.

PR #4967, opened 2025-12-12
[perfermance] Eliminate the D2H synchronization operations in post-processing to resolve the Eagle partial fast/slow card issue

coder-fny:
When temperature > 0 and both top_p and top_k are present, the post-processing logic enters a branch that checks whether to use the fused npu_top_k_top_p operator.
image
 In this check, the value of top_k originally resides on the NPU, and evaluating whether it lies within the range [1, 1024] triggers an implicit device-to-host (D2H) synchronization, which is highly time-consuming. To avoid this overhead, the check should instead use the CPU-side copy of top_k directly, thereby 
1765522679445_CCEAFA06-A600-4da9-90F5-F85A2AE19409

This causes the operator dispatch in the Eagle component to be significantly slowed down, leading to a severe fast/slow card issue.
image
After the change：
image

No

- vLLM version: v0.12.0


wangxiyuan:

PR #4973, opened 2025-12-12
[Bugfix] fix pipeline parallelism bug introduced by async-scheduling refactor work

lidenghui1110:
Currently, when using pipeline parallel and pd disaggregate, model_runner will return None on non-last-pp-rank stages in `sample_tokens`, which will cause assert error in vllm KVOutputAggregator on [this line](https://github.com/vllm-project/vllm/blob/main/vllm/distributed/kv_transfer/kv_connector/utils.py#L84).

In fact, all pp workers should return a model_runner_output which contains kv_connector_output to do aggregate in Enginecore scheduler process to ensure all kv transfer is finished for kv cache releasing later. 

To fix this issue, this PR follows gpu_model_runner in vllm, passing kv_connector_output in `sample_tokens` to make sure all ranks will return a ModelRunnerOutput, in non-last-pp-rank workers, it will return EMPTY_MODEL_RUNNER_OUTPUT with kv_connector_output.

No

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -402,7 +402,6 @@ class NPUModelRunner(GPUModelRunner):
         self.reorder_batch_threshold: int | None = None
         self.query_start_loc = self._make_buffer(self.max_num_reqs + 1,
                                                  dtype=torch.int32)
-        self.kv_connector_output: KVConnectorOutput | None = None
 
     def _init_device_properties(self) -> None:
         self.num_sms = None


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -403,6 +402,7 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
         self.reorder_batch_threshold: int | None = None
         self.query_start_loc = self._make_buffer(self.max_num_reqs + 1,
                                                  dtype=torch.int32)
+        self.kv_connector_output: KVConnectorOutput | None = None
```
wangxiyuan: gpu model runner has initial this property

lidenghui1110: > gpu model runner has initial this property

remove done.

PR #4975, opened 2025-12-12
[Feat] Refactor rejection sampler

realliujiaxu:

Currently, we are using `AscendRejctionSampler` that extends from `RejctionSampler` in spec decoding. `AscendRejctionSampler` override `forward` of `RejctionSampler`, only aming to replace `rejection_sample` func. This   
 causes a lot of code of `RejctionSampler` cannot be reused, for example: 
- https://github.com/vllm-project/vllm/pull/19482
- https://github.com/vllm-project/vllm/pull/26060
- https://github.com/vllm-project/vllm/pull/29223

#### Proposed Change:
- Delete `AscendRejctionSampler` and use `RejctionSampler` directly in model runner.
- Patch `RejctionSampler.expand_batch_to_tokens` and `RejctionSampler.rejection_sample`, maybe a better way is to make them as custom ops.
- Modify `NPUModelRunner` following https://github.com/vllm-project/vllm/pull/26060

No

- [x] test logits processor for spec decoding
- [x] test logprobs for spec decoding
- [x] test logprobs for spec decoding + async shcheduling (test with https://github.com/vllm-project/vllm-ascend/pull/4893/)

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/ops/fused_moe/moe_mlp.py
+++ b/vllm_ascend/ops/fused_moe/moe_mlp.py
@@ -46,7 +46,7 @@ def cumsum_group_list(group_list: torch.Tensor,
         return group_list.cumsum(dim=0)
     if src_list_type == 0 and dst_list_type == 1:
         group_diff = torch.diff(group_list)
-        new_group = torch.cat([group_diff[0].unsqueeze(0), group_diff], dim=0)
+        new_group = torch.cat([group_list[0].unsqueeze(0), group_diff], dim=0)
         return new_group
     if src_list_type == 2 and dst_list_type == 0:
         experts = pad(group_list[:, 0], (1, 0))


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1901,13 +1785,164 @@ def propose_draft_token_ids(sampled_token_ids):
...
+    ) -> tuple[
+            LogprobsLists | None,
+            list[list[int]],
+            dict[str, LogprobsTensors | None],
+            list[str],
+            dict[str, int],
+            list[int],
+    ]:
+        num_reqs = self.input_batch.num_reqs
+        discard_sampled_tokens_req_indices = np.nonzero(
```
zhenwenqi2024: can some attrs like discard_requests_mask be deleted in prepare_inputs?

realliujiaxu: we need to implement https://github.com/vllm-project/vllm/pull/28597 first.

vllm_ascend/patch/worker/__init__.py
```diff
@@ -33,3 +33,4 @@
 import vllm_ascend.patch.worker.patch_qwen3_vl  # noqa
```
wangxiyuan, line 33: update this file please https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/patch/__init__.py

PR #4987, opened 2025-12-13
[Misc]modify rope related logic

Angazenn:

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/ops/rotary_embedding.py
+++ b/vllm_ascend/ops/rotary_embedding.py
@@ -42,7 +42,6 @@ from vllm_ascend.utils import (AscendDeviceType, enable_custom_op,
 # by different approaches.
 _cos_mla: Optional[torch.Tensor] = None
 _sin_mla: Optional[torch.Tensor] = None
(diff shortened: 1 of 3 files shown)


vllm_ascend/ops/rotary_embedding.py
```diff
@@ -105,29 +104,25 @@ def get_cos_and_sin_mla():
...
 
-    if _cos_sin_cache is None or \
-        _cos is None or \
-        _sin is None:
+    if _cos is None or _sin is None or not hasattr(model, "model"):
         return
 
+    model_type = vllm_config.model_config.hf_config.model_type
+    if model_type in ["qwen3_next"]:
+        rope_layer = 3
```
lidenghui1110: as [qwen3-next layer definition](https://github.com/vllm-project/vllm/blob/main/vllm/transformers_utils/configs/qwen3_next.py#L253) shows that every four layers is a linear_attention, otherwise it is a full_attention, if rope_layer = 3 here, will full_attention share same cos/sin cache with linear_attention?

wangxiyuan:

PR #4998, opened 2025-12-14
[CI][Bugfix] Fix scheduleroutput has no attr get error in prompt logprobs

MengqingCao:
Fix scheduleroutput has no attr get error in prompt logprobs

Fix https://github.com/vllm-project/vllm-ascend/actions/runs/20194753373/job/57977131870

CI passed with existing test.

- vLLM version: v0.12.0


zhenwenqi2024:
approve

PR #4999, opened 2025-12-14
[Feat] Support MLP_TP feature, exclude MOE layer

zzhx1:
#4257 This PR implements the dense_ffn TP of the first three layers of the deepseek model, I have refactored this PR and used very little code to support the implementation of this feature.
This PR adds a function `is_moe_layer` to mlp_tp, which supports MLP TP in models with both mlp and moe, such as deepseek or chat GLM.

- vLLM version: v0.12.0


vllm_ascend/ops/linear_op.py
```diff
@@ -681,3 +683,21 @@ def get_replicated_op(disable_tp, prefix,
         return None
 
     return CustomReplicatedOp(layer)
+
+
+def is_moe_layer(prefix: str) -> bool:
```
wangxiyuan, line 689: add lru cache

weijinqian0 (approved):

vllm_ascend/ops/linear_op.py
zzhx1: fix done

jianzs (approved):

PR #5001, opened 2025-12-14
Pcp dcp refactor

zhenwenqi2024:

- vLLM version: v0.12.0


vllm_ascend/worker/pcp_utils.py
```diff
@@ -0,0 +1,592 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+from typing import List
```
LookAround0301: 不单独创建新的pcp_utils文件，采用和GPU的commn_utils方式，创建公共的工具文件

zhenwenqi2024: ok,i will chanage it

jianzs:
Is it possible to refactor the logic of pcp_dcp in attention so that both pcp and non-pcp paths share as much common processing as possible, minimizing major branching differences?

PR #5007, opened 2025-12-15
[KVPool]Fix PP get bug

baxingpiaochong:

When kv caches are evicted from the key-value pool, it's possible that the kv cache for pp0 is still active, but the kv cache for pp1 has already been evicted. Therefore, a unified check is needed during the get operation.

- vLLM version: v0.12.0


LCAIZJ:
For the PP (Pipeline Parallelism) scenario, consider the query pattern of each PP stage.

PR #5038, opened 2025-12-15
[Bugfix] Fix the attn_metadata is None

dragondream-chen:
Fix the bug " TypeError: 'NoneType' object is not iterable'  " in vllm_ascend/compilation/acl_graph.py
The reason of that is the attn_metadata is none in the dummy_run of MTP.

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/spec_decode/eagle_proposer.py
+++ b/vllm_ascend/spec_decode/eagle_proposer.py
@@ -117,7 +117,7 @@ class EagleProposer(Proposer):
     def dummy_run(self,
                   num_tokens: int,
                   with_prefill: bool = False,
-                  skip_attn: bool = False,
+                  in_graph_capturing: bool = False,
                   num_reqs: int = 0,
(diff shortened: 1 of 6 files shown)


yiz-liu (changes requested):
Please update the name of `skip_attn`, as its current name no longer accurately reflects its semantics. It would be more appropriate to rename it to `in_graph_capturing` or a similarly descriptive term.

dragondream-chen:
> Please update the name of `skip_attn`, as its current name no longer accurately reflects its semantics. It would be more appropriate to rename it to `in_graph_capturing` or a similarly descriptive term.

DONE. The old name is indeed not very easy to understand.

yiz-liu:
I’m afraid there is a misunderstanding here: `skip_attn` does not correspond to `in_graph_capturing`, in fact, it represents the opposite behavior. Please take a closer look at the underlying logic and revise it accordingly as soon as possible.
 @dragondream-chen

PR #5046, opened 2025-12-15
[Bugfix]fix dsv3.1 FIA err in async_scheduling with mtp

### LINKED ISSUE #4986: [Bug]: Server hangs when running FULL_DECODE_ONLY mode  and enabling MTP with DeepSeek V3.1
### Your current environment

``

### 🐛 Describe the bug

```
vllm serve /path/to/model --api-server-count 4 --port 8200 --trust-remote-code --served-model-name auto --distributed-executor-backend mp --model-loader-extra-config '{"enable_multithread_load":true,"num_threads":8}' --enable-prompt-tokens-details --max-num-seqs 42 --max-model-len 131072 --max-num-batched-tokens 84 --block-size 128 --gpu-memory-utilization 0.90 --no-enable-prefix-caching --data-parallel-size 32 --data-parallel-size-local 16 --data-parallel-address 33.182.142.47 --data-parallel-rpc-port 14435 --data-parallel-start-rank 0 --tensor-parallel-size 1 --enable-expert-parallel --quantization ascend --speculative-config '{"num_speculative_tokens":1, "method": "deepseek_mtp"}' --compilation-config '{"cudagraph_capture_sizes":[2,4,8,16,24,32,40,42], "cudagraph_mode": "FULL_DECODE_ONLY"}' --additional-config '{"enable_cpu_binding":true,"enable_kv_nz":true,"multistream_overlap_shared_expert":true,"finegrained_tp_config":{"embedding_tensor_parallel_size":8,"lmhead_tensor_parallel_size":8}}' --kv-transfer-config '{
        "kv_connector": "MooncakeConnector",
        "kv_connector_module_path": "vllm_ascend.distributed.mooncake_connector",
        "kv_buffer_device": "npu",
        "kv_role": "kv_consumer",
        "kv_parallel_size": 2,
        "kv_port": "20002",
        "engine_id": "decode-0",
        "kv_rank": 1,
            "use_ascend_direct": true,
            "prefill": {
                    "dp_size": 2,
                    "tp_size": 8,
                    "pp_size": 1,
                    "pp_layer_partition": null
             },
             "decode": {
                    "dp_size": 32,
                    "tp_size": 1
             }
        }
    }'
```

In the prefill-decode disaggregation scenario, the decode engine freezes when it receives requests.

The plog shows:

Image

hust17yixuan:
When enable the async_scheduling, in large scale EP scene, mtp module goes to eagler mode, which results in the mismatch of seq_lens_list、block_table. So adapt the judgement before the draft model forward.

fix #4986 

No

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/spec_decode/mtp_proposer.py
+++ b/vllm_ascend/spec_decode/mtp_proposer.py
@@ -725,7 +725,6 @@ class MtpProposer(Proposer):
         has_lora = len(self.runner.input_batch.lora_id_to_lora_request) > 0
         aclgraph_runtime_mode, batch_descriptor = \
             self.runner.cudagraph_dispatcher.dispatch(num_tokens=num_input_tokens, uniform_decode=uniform_decode, has_lora=has_lora)
-        original_aclgraph_runtime_mode = aclgraph_runtime_mode
         if self.use_async_scheduling:
             # there is synchronization between mtp steps when enabling aclgraph,
             # disable aclgraph when use async scheduling to avoid the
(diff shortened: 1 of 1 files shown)


yiz-liu:
I'll admit this padding logic is confusing, reluctantly, we enforce slicing so the shapes match. I know we're all eager for a formal fix, but for now I think the best option is to wait for the Eagle/MTP unification, it may cause some initial performance regression, but it will greatly improve maintainability.

PR #5051, opened 2025-12-15
[Feature] implement model runner v2 basic framework

Ronald1995:
This PR aim to implement model runner v2 basic framework in vllm-ascend, the e2e function is not guaranteed by this pr.
 
use envs.VLLM_USE_V2_MODEL_RUNNER to decide if choose model_runenr_v2.

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/worker/v2/model_runner.py
+++ b/vllm_ascend/worker/v2/model_runner.py
@@ -6,7 +6,6 @@ import torch
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
 from vllm.v1.core.sched.output import GrammarOutput, SchedulerOutput
-from vllm.v1.kv_cache_interface import EncoderOnlyAttentionSpec
 from vllm.v1.worker.gpu.input_batch import (InputBatch,
                                             combine_sampled_and_draft_tokens,
                                             prepare_pos_seq_lens,
--- a/vllm_ascend/worker/v2/model_runner.py
+++ b/vllm_ascend/worker/v2/model_runner.py
@@ -5,8 +5,7 @@ import numpy as np
 import torch
 from vllm.config import VllmConfig
 from vllm.logger import init_logger
-from vllm.v1.worker.gpu.sample.output import SamplerOutput
-from vllm.v1.core.sched.output import SchedulerOutput, GrammarOutput
+from vllm.v1.core.sched.output import GrammarOutput, SchedulerOutput
 from vllm.v1.kv_cache_interface import EncoderOnlyAttentionSpec
 from vllm.v1.worker.gpu.input_batch import (InputBatch,
                                             combine_sampled_and_draft_tokens,
@@ -14,13 +13,12 @@ from vllm.v1.worker.gpu.input_batch import (InputBatch,
                                             prepare_prefill_inputs)
 from vllm.v1.worker.gpu.model_runner import GPUModelRunner
 from vllm.v1.worker.gpu.sample.metadata import SamplingMetadata
+from vllm.v1.worker.gpu.sample.output import SamplerOutput
 
 from vllm_ascend.worker.v2.aclgraph_utils import AclGraphManager
-from vllm_ascend.worker.v2.attn_utils import (
-    build_attn_metadata,
-    build_attn_state,
-    make_attention_mask,
-)
+from vllm_ascend.worker.v2.attn_utils import (build_attn_metadata,
+                                              build_attn_state,
+                                              make_attention_mask)
 from vllm_ascend.worker.v2.input_batch import AscendInputBuffers
 from vllm_ascend.worker.v2.states import AscendRequestState
 from vllm_ascend.worker.v2.utils import torch_cuda_wrapper
@@ -42,7 +40,10 @@ class NPUModelRunner(GPUModelRunner):
         del self.input_buffers
 
         # NPU specific initializations can be added below.
-        self.cudagraph_manager = AclGraphManager(vllm_config, device)
+        self.cudagraph_manager: AclGraphManager = AclGraphManager(
+            vllm_config,
+            device,
+        )
         # AscendRequestState has extra `num_computed_tokens_cpu` attribute.
         # so reinitialize req_states here.
         self.req_states: AscendRequestState = AscendRequestState(
@@ -96,7 +97,7 @@ class NPUModelRunner(GPUModelRunner):
         num_tokens_after_padding: int,
     ) -> InputBatch:
         """Override GPUModelRunner.prepare_inputs for Ascend NPUs.
-        npu attention bakcends need seq_lens_cpu to work.
+        npu attention backends need seq_lens_cpu to work.
         so we need to prepare seq_lens_cpu here.
         """
         num_tokens = scheduler_output.total_num_scheduled_tokens
@@ -308,6 +309,8 @@ class NPUModelRunner(GPUModelRunner):
             # when doing speculative decoding with async_scheduling,
             # we need to copy num_rejected_tokens back to cpu.
             default_stream = torch.cuda.current_stream()
+            assert self.num_rejected_token_stream is not None
+            assert self.num_rejectd_tokens_cpu is not None
             with torch.npu.stream(self.num_rejected_token_stream):
                 self.num_rejected_token_stream.wait_stream(default_stream)
                 self.num_rejectd_tokens_cpu.copy_(
--- a/vllm_ascend/worker/v2/model_runner.py
+++ b/vllm_ascend/worker/v2/model_runner.py
@@ -4,20 +4,23 @@
 import numpy as np
 import torch
 from vllm.config import VllmConfig
-from vllm.distributed.parallel_state import get_dcp_group, get_pcp_group
 from vllm.logger import init_logger
-from vllm.v1.core.sched.output import SchedulerOutput
+from vllm.v1.worker.gpu.sample.output import SamplerOutput
+from vllm.v1.core.sched.output import SchedulerOutput, GrammarOutput
 from vllm.v1.kv_cache_interface import EncoderOnlyAttentionSpec
 from vllm.v1.worker.gpu.input_batch import (InputBatch,
                                             combine_sampled_and_draft_tokens,
                                             prepare_pos_seq_lens,
                                             prepare_prefill_inputs)
 from vllm.v1.worker.gpu.model_runner import GPUModelRunner
+from vllm.v1.worker.gpu.sample.metadata import SamplingMetadata
 
-from vllm_ascend.attention.attention_mask import AttentionMaskBuilder
-from vllm_ascend.attention.attention_v1 import AscendAttentionState
 from vllm_ascend.worker.v2.aclgraph_utils import AclGraphManager
-from vllm_ascend.worker.v2.attn_utils import build_attn_metadata
+from vllm_ascend.worker.v2.attn_utils import (
+    build_attn_metadata,
+    build_attn_state,
+    make_attention_mask,
+)
 from vllm_ascend.worker.v2.input_batch import AscendInputBuffers
 from vllm_ascend.worker.v2.states import AscendRequestState
 from vllm_ascend.worker.v2.utils import torch_cuda_wrapper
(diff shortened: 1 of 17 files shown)


vllm_ascend/worker/worker_v1.py
wangxiyuan, line 95: rename to worker.py

vllm_ascend/worker/worker.py
```diff
@@ -228,7 +230,12 @@ def init_device(self):
         # for more details
         self.device = self._init_device()
         # Init ModelRunner here, so that we have access to self.device.
-        self.model_runner = NPUModelRunner(self.vllm_config, self.device)
+        if self.use_v2_model_runner:
```
wangxiyuan, line 235: check env.VLLM_USE_V2_MODEL_RUNNER directly instead of set `self.use_v2_model_runner`

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+    """Model runner for Ascend NPUs."""
+
+    def __init__(self, vllm_config: VllmConfig, device: torch.device):
+        with torch_cuda_wrapper():
+            super().__init__(vllm_config, device)
+
+        # release original object.
+        del self.cudagraph_manager
+        del self.req_states
+        del self.input_buffers
```
wangxiyuan, line 39: why `del` them?

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+            max_num_reqs=self.max_num_reqs,
+            max_num_tokens=self.max_num_tokens,
+            inputs_embeds_size=self.inputs_embeds_size,
+            vocab_size=self.vocab_size,
+            dtype=self.dtype,
+            device=self.device,
+            pin_memory=self.pin_memory,
+        )
+
+        self.dcp_size = get_dcp_group().world_size
```
wangxiyuan: I think we don't need introduce this in the first PR. It's good to contribute to vLLM instead.

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+        self.req_ids: list[str] = []
+        if self.use_async_scheduling and self.do_spec_decode:
+            self.num_rejected_tokens_event = torch.npu.Event()
+            self.num_rejected_token_stream = torch.npu.Stream()
+            self.num_rejectd_tokens_cpu = torch.empty(
+                self.max_num_reqs,
+                dtype=torch.int64,
+                device="cpu",
+                pin_memory=self.pin_memory,
+            )
```
wangxiyuan, line 91: please make sure all property above are needed. We should contribute to vLLM as much as we can. If they're still need, please add note one by one to describe why。

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+        self,
+        scheduler_output,
+        num_tokens_after_padding,
+    ) -> InputBatch:
+        """Override GPUModelRunner.prepare_inputs for Ascend NPUs.
+        npu attention bakcends need seq_lens_cpu to work.
+        so we need to prepare seq_lens_cpu here.
+        """
+        num_tokens = scheduler_output.total_num_scheduled_tokens
+        assert num_tokens > 0
```
wangxiyuan, line 103: I don't think `assert` is uesful totally,  In model runner V2, can we drop `assert` at all? Instead, raise useful error is better.

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+
+        # Decode first, then prefill.
+        # batch_idx -> req_id
+        req_ids = sorted(
+            scheduler_output.num_scheduled_tokens.keys(),
+            key=lambda k: scheduler_output.num_scheduled_tokens[k],
+        )
+
+        # special handling for npu.
+        self.req_ids = req_ids
```
wangxiyuan: we should avoid set property value in functions. It make the code very complex and hard for maintainer in the future. We should consider a better way to pass it. If we have to, please add more note.

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+
+    def prepare_inputs(
+        self,
+        scheduler_output,
+        num_tokens_after_padding,
+    ) -> InputBatch:
+        """Override GPUModelRunner.prepare_inputs for Ascend NPUs.
+        npu attention bakcends need seq_lens_cpu to work.
+        so we need to prepare seq_lens_cpu here.
+        """
```
wangxiyuan, line 101: add more docstring to explain input and output

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+        # Decode first, then prefill.
+        # batch_idx -> req_id
+        req_ids = sorted(
+            scheduler_output.num_scheduled_tokens.keys(),
+            key=lambda k: scheduler_output.num_scheduled_tokens[k],
+        )
+
+        # special handling for npu.
+        self.req_ids = req_ids
+        self._update_seq_lens_cpu(scheduler_output, req_ids)
```
wangxiyuan, line 113: since `req_ids` is set to `self.req_ids`, no need to pass it. Or just pass it and remove `self.req_ids`

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+        num_scheduled_tokens = np.array(
```
wangxiyuan, line 115: clear the type of the variable, e.g. `num_scheduled_tokens: np.array = xxx`  And the same for all other variable in this PR.

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+            [scheduler_output.num_scheduled_tokens[i] for i in req_ids],
+            dtype=np.int32)
+        num_valid_tokens = num_scheduled_tokens
+        if scheduler_output.scheduled_spec_decode_tokens:
+            num_valid_tokens = np.array([
+                num_tokens -
+                len(scheduler_output.scheduled_spec_decode_tokens.get(i, []))
+                for num_tokens, i in zip(num_scheduled_tokens, req_ids)
+            ],
+                                        dtype=np.int32)
```
wangxiyuan: `num_valid_tokens  = np.array(xxx) if scheduler_output.scheduled_spec_decode_tokens else num_scheduled_tokens`

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+        self.attn_state = self._build_attn_state(
```
wangxiyuan: you set `self.attn_state` here, but it looks that it's only used in this funtion. how about just `attn_state`?

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+            num_reqs,
+            num_scheduled_tokens,
+            num_valid_tokens,
+        )
+        self.attn_mask = self._make_attention_mask(self.attn_state)
```
wangxiyuan: ditto

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+
+        idx_mapping_list = [
```
wangxiyuan, line 143: we should drop useless variable define IMO. for exmaple, we can set  `idx_mapping.np[:num_reqs] = [self.req_states.req_id_to_index[req_id] for req_id in req_ids]`, unless the name `idx_mapping_list` here is useful?

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+        return output
+
+    def sample(
+        self,
+        hidden_states,
+        input_batch,
+        sampling_metadata,
+        grammar_output,
+    ):
+        sampler_output, num_sampled, num_rejected = super().sample(
```
wangxiyuan, line 300: ditto

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -0,0 +1,383 @@
...
+            seq_lens=seq_lens,
+            seq_lens_np=seq_lens_np,
+            input_ids=input_ids,
+            positions=positions,
+            attn_metadata=attn_metadata,
+            logits_indices=logits_indices,
+            cu_num_logits=cu_num_logits,
+        )
+
+    def sample_tokens(self, grammar_output):
```
wangxiyuan: like `prepare_input`, please add docstring to descrite  that this is a overrdie func and what's the input and output

vllm_ascend/worker/worker_v1.py
Ronald1995: ok, i will fix it.

vllm_ascend/worker/worker.py
Ronald1995: this is just the same as vllm

vllm_ascend/worker/v2/model_runner.py
Ronald1995: the purpose is to make sure python gc collect these object, i will make more specific comment for this.

vllm_ascend/worker/v2/model_runner.py
Ronald1995: ok, i will delete these attributes

vllm_ascend/worker/v2/model_runner.py
Ronald1995: i will make more specific comments.

vllm_ascend/worker/v2/model_runner.py
Ronald1995: this is just the same as vllm

vllm_ascend/worker/v2/model_runner.py
Ronald1995: this is just the same as vllm, i think they define idx_mapping_list is for code format.

vllm_ascend/worker/v2/model_runner.py
Ronald1995: The purpose i upack the output here is to make it clear what results is returned and why use a new stream to copy num_rejected here.

vllm_ascend/worker/v2/model_runner.py
Ronald1995: ok，i will add docstring.

vllm_ascend/worker/v2/model_runner.py
Ronald1995: i just remove it for now

vllm_ascend/worker/v2/model_runner.py
Ronald1995: i think it's not necessary to add typehint for this kind of variable, because the return type of np.array is clear, all ides can infer the type of the variable. it's necessary to add typehint for those variables that mypy or ide can't infer the type.

vllm_ascend/worker/v2/model_runner.py
Ronald1995: else statement is too long if we use if ... else ..., i think it's better for format to only use if statement.

vllm_ascend/worker/v2/model_runner.py
Ronald1995: i remove it.

vllm_ascend/worker/v2/model_runner.py
Ronald1995: variable

vllm_ascend/worker/v2/model_runner.py
wangxiyuan: Nice update

vllm_ascend/worker/worker.py
```diff
@@ -230,7 +232,15 @@ def init_device(self):
...
+            logger.error(
```
wangxiyuan, line 236: logger.info

wangxiyuan (approved):

wangxiyuan:
this is a good start to support model runner v2 on ascend. I'm fine this change.

yiz-liu (approved):

PR #5058, opened 2025-12-16
[Doc] Update structured output doc with upstream link

shen-shanshan:

Cherry-pick from main https://github.com/vllm-project/vllm-ascend/pull/4015.

Currently, the usage of structured output feature in vllm-ascend is totally the same as that in vllm.

Thus, IMO, it's better to remove this doc directly to avoid some case that there are some changes in the upstream doc and we don't update our doc in time, which can be misleading to users.


shen-shanshan:
CC @MengqingCao

wangxiyuan (approved):

PR #5059, opened 2025-12-16
[BugFix]Fix FIA input err in DSv3.1

hust17yixuan:
When use mtp, full decdoe only and async_scheduling together, finding a input err for FIA ops due to the non-increasing input
of the 'actual_seq_lengths'. This bug is caused by the filling the variable ‘query_start_loc’. We need to fill the query_start_loc' s end by the 'cu_num_tokens' instead of '-1'

No

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -767,6 +767,7 @@ class NPUModelRunner(GPUModelRunner):
 
         self.query_start_loc.np[0] = 0
         self.query_start_loc.np[1:num_reqs + 1] = cu_num_tokens
+        self.query_start_loc.np[num_reqs + 1:].fill(cu_num_tokens[-1])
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -775,7 +776,6 @@ def _prepare_inputs(
         self.seq_lens.copy_to_gpu()
 
         # Fill unused with -1. Needed for reshape_and_cache
```
yiz-liu: Delete this comment.

yiz-liu (changes requested):

yiz-liu:
We overwrite possible stale values from previous batches in this PR. Without this fix, these dirty values could cause errors in subsequent operators that process the tensor. Also we move this before copying to NPU, which makes more senses.

wangxiyuan (approved):

PR #5077, opened 2025-12-16
[Graph][Fusion]Add new pattern for AddRmsnormQuant with SP.

Angazenn:
1. In addition to [#4168](https://github.com/vllm-project/vllm-ascend/pull/4168), [#5011](https://github.com/vllm-project/vllm-ascend/pull/5011)， this PR adds two more pattern for AddRmsnormQuant with SP enabled. The key difference is to insert an additional `maybe_all_gather_and_maybe_unpad` between `addrmsnorm` and `quantize`.
2. This PR also introduce another api `torch.ops.vllm.quantize`, so that we pass `input_scale` and `input_scale_reciprocal` at the same time. This is because `npu_add_rms_norm_quant` and `npu_quantize` requires different `div_mode`. To avoid introducing additional reciprocal calculation in runtime, we have to pass both of them to quantize api.
3. Removes redundant `AscendQuantRmsnorm`.

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/ops/linear_op.py
+++ b/vllm_ascend/ops/linear_op.py
@@ -543,8 +543,7 @@ class SequenceRowParallelOp(CustomRowParallelOp):
         from vllm.model_executor.layers.linear import UnquantizedLinearMethod
 
         from vllm_ascend.quantization.quant_config import AscendLinearMethod
-        from vllm_ascend.quantization.w8a8 import (AscendW8A8LinearMethod,
-                                                   quant_per_tensor)
+        from vllm_ascend.quantization.w8a8 import AscendW8A8LinearMethod
 
         # For unquant
         if mmrs_fusion and isinstance(self.layer.quant_method,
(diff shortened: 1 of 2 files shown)


Angazenn:
Please take a look on this @wxsIcey @whx-sjtu .

wxsIcey:
I think replacing the allreduce+rmsnorm+quant operation with reducescatter+rmsnorm+quant+allgather is better, like https://github.com/vllm-project/vllm/blob/main/vllm/compilation/sequence_parallelism.py#L160. This can be a futher work.

vllm_ascend/compilation/passes/norm_quant_fusion_pass.py
```diff
@@ -83,25 +83,145 @@ class AddRMSNormQuantPatternWithBias:
...
+            out1 = output[2]
+            quantized_output = torch.ops.vllm.maybe_all_gather_and_maybe_unpad(
+                quantized_output, True)
+            return quantized_output, out1
+
+        pm.register_replacement(pattern, replacement, self.get_inputs(),
+                                pm.fwd_only, pm_pass)
+
+
+class AddRMSNormQuantSPPatternWithBias:
```
wxsIcey, line 212: Please write a separate pass for the fusion of sp.

Angazenn: Using fusion of sp as vLLM is another issue, since it is contradictory to flashcomm in vLLM-Ascend. At this moment, I think it is ok to fuse AddRMSNormQuant with sp here. For further discussion with sp implementation, cc @weijinqian0 @realliujiaxu .

wxsIcey: I mean that the `AddRMSNormQuantSPPattern` should be added to a new Pass, like `SequenceParallelPass`, this can make the design more reasonale.

PR #5118, opened 2025-12-17
[Feat] Support to use fullgraph with eagle

anon189Ty:
    
We support to use full graph with eagle. See #5459 .

Change list:
    1. Distinguish between processing graph_params and draft_graph_params in attention_v1.
    2. Adapt the full-graph mode in eagle_proposer, include:
        1). If use full graph, make Fullgraph Wrapper when load model.
        2). Build a new meatadata, set running mode in FULL and mark attention update in dummy_run when in Fullgraph mode.
        3). Fixed and fill any attn_metadata, such as attn_metadata.slot_mapping.
        4). Add a descriptor.
        5). Set running mode and triggered update metadata.
    3. Trans is_mtp_model to is_draft_model, and add the update of workspace.

NOTE:
    When set async_scheduling=True, the draft model will enforce execution in eager mode.

- vLLM version: v0.12.0


vllm_ascend/spec_decode/eagle_proposer.py
```diff
@@ -147,16 +173,71 @@ def dummy_run(self,
...
+                    in_profile_run=True,
+                    batch_descriptor=batch_descriptor,
+                    aclgraph_runtime_mode=aclgraph_runtime_mode,
+                    is_draft_model=True):
+                forward_context = get_forward_context()
+                self.model(
+                    input_ids=self.input_ids[:num_tokens],
+                    positions=self.positions[:num_tokens],
+                    hidden_states=self.hidden_states[:num_tokens],
+                )
```
yiz-liu, line 258: Why is it different from `propose` that seperates `i=0` and `i>0`?

yiz-liu: Also it should not be `not in_graph_capturing`, see #5072 .

vllm_ascend/compilation/acl_graph.py
```diff
@@ -196,6 +196,7 @@ def __call__(self, *args, **kwargs):
 
 
 def _update_attn_pa_params(update_stream, forward_context, runtime_shape):
+    # TODO: Is pa should be adapted with draft_graph_params?
```
yiz-liu: Please take a look at `full_graph_pa` in model runner.

anon189Ty: Checked. Model will not run paged attention when set draft model because of the function 'using_paged_attention' in 'vllm_ascend/attention/utils.py'.

vllm_ascend/spec_decode/eagle_proposer.py
anon189Ty: 1. In propose, the num_tokens is different between with 'i=0' and 'i>0'. We can use a smaller graph when i>0. Using 'i=0' and 'i>0' is also to maintain consistency with the code structure of the 'propose' in vLLM.
2. Fixed.

yiz-liu (approved):

PR #5129, opened 2025-12-17
Qwen3-Next：Update the gpu-memory-utilization parameter to 0.7

ming1212:
Update the gpu-memory-utilization parameter to 0.7

- vLLM version: v0.12.0


wangxiyuan:
make sense, we'll fix the oom in the next few release.

PR #5133, opened 2025-12-17
[Perf] Autotune and cache causal_conv1d_fwd launch parameters

kiscad:

- add Triton autotune configs (BLOCK_N 64–1024) to `_causal_conv1d_fwd_kernel` so the forward path can automatically pick the best width for the current `dim`/`state_len`
- choose `BLOCK_M` dynamically from the median sequence length (capped by a new `DEFAULT_BLOCK_M`) and reuse metadata-provided scheduling info when available
- fall back to a CPU-side precomputation path that resizes and fills reusable `batch_ptr`/`token_chunk_offset_ptr` buffers before moving them to the NPU, cutting repeated allocations and copies
- clean up assorted helpers/comments around the causal conv kernels while wiring the new launch grid logic

- vLLM version: v0.12.0

### DIFF
--- a/tests/e2e/nightly/ops/triton/test_causal_conv1d.py
+++ b/tests/e2e/nightly/ops/triton/test_causal_conv1d.py
@@ -1,8 +1,10 @@
+from types import SimpleNamespace
 from typing import Optional
 
 import pytest
 import torch
 import torch.nn.functional as F
+from vllm.v1.attention.backends.utils import compute_causal_conv1d_metadata
 
 from vllm_ascend.ops.triton.mamba.causal_conv1d import (PAD_SLOT_ID,
                                                         causal_conv1d_fn)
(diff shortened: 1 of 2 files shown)


wangxiyuan:

kiscad:
> Any progress? If this PR is still alive, please rebase to main and make CI happy, otherwise you can close it. Thanks

This PR can improve the performance of `causal_conv1d` kernel by 8 times in Qwen Next model. But the triton-ascend compiler seems have some bug which could make other aclnn operator crash (only in the scenario of full-graph), which is also reported by other triton kernels' integration.

So this PR will wait triton-ascend compiler to fix that bug.

wangxiyuan:
we upgraded triton-ascend to 3.2.1 now. Can this PR keep going on now?

PR #5138, opened 2025-12-17
deepseek量化适配

ElleElleWu:

- vLLM version: v0.12.0


wangxiyuan:

PR #5142, opened 2025-12-17
fix: use batch_matmul_transpose operator in MLA _v_up_proj for better performance

LICO1314:

This PR fixes a bug in the `AscendMLAImpl._v_up_proj` method where the optimized `batch_matmul_transpose` operator was not being utilized.

**Changes:**
- Modified `_v_up_proj` method to use `torch.ops._C_ascend.batch_matmul_transpose` operator for FP16/BF16 dtypes when available
- Added fallback path using the original `torch.bmm` implementation for other cases
- This avoids unnecessary transpose operations and improves performance

**Why needed:**
- The previous implementation only used `torch.bmm` with multiple transpose operations, which is less efficient
- The Ascend backend provides an optimized `batch_matmul_transpose` operator that can handle the computation more efficiently
- This fix improves inference performance for MLA (Multi-head Latent Attention) models on Ascend NPU

No. This is a performance optimization that maintains the same functionality and output. Users will experience faster inference for MLA-based models, but no API or interface changes are introduced.

The changes maintain backward compatibility with the fallback path, ensuring correct behavior when the operator is not available or for unsupported dtypes.
- vLLM version: v0.12.0


weijinqian0 (approved):

realliujiaxu (approved):

Yikun:
Thanks for your first contributions! Your awesome first PR has been included in vLLM Ascend v0.13.0rc1 release.

[1] https://github.com/vllm-project/vllm-ascend/releases/tag/v0.13.0rc1
[2] https://mp.weixin.qq.com/s/3Psz3mYFTLktgSEDGqM9wQ

PR #5149, opened 2025-12-18
[E2E] add E2E for Prefix Caching cp & Chunked Prefill cp

ader47:
Add E2E for Prefix Caching cp & Chunked Prefill cp 
No.

- vLLM version: v0.12.0

### DIFF
--- a/.github/workflows/_e2e_nightly_multi_node.yaml
+++ b/.github/workflows/_e2e_nightly_multi_node.yaml
@@ -32,7 +32,7 @@ on:
         description: how many pods will be pulled up via lws.yaml, indicates number of nodes we need
       vllm_version:
         required: false
-        default: "v0.14.1"
+        default: "v0.15.0"
         type: string
         description: vllm version to use
       vllm_ascend_remote_url:
@@ -78,7 +78,7 @@ jobs:
         - name: Decode kubeconfig from secrets
           run: |
             # Decode and save kubeconfig
-            echo "${{ secrets.KUBECONFIG_B64 }}" | base64 -d > $KUBECONFIG
+            echo "${{ secrets.KUBECONFIG_B64 }}" | base64 -d > "$KUBECONFIG"
 
         - name: Checkout code
           uses: actions/checkout@v6
@@ -133,7 +133,7 @@ jobs:
             image="${{ inputs.image }}"
             config_file_path="${{ inputs.config_file_path }}"
             fail_tag=FAIL_TAG_"${{ inputs.config_file_path }}"
-            echo "FAIL_TAG=${fail_tag}" >> $GITHUB_ENV
+            echo "FAIL_TAG=${fail_tag}" >> "$GITHUB_ENV"
 
             required_params=("size" "replicas" "image" "config_file_path")
             for param in "${required_params[@]}"; do
@@ -264,5 +264,5 @@ jobs:
         - name: Post process
           if: always()
           run: |
-            kubectl get pods -n $NAMESPACE --ignore-not-found=true
+            kubectl get pods -n "$NAMESPACE" --ignore-not-found=true
             kubectl delete -f ./lws.yaml --ignore-not-found=true || true
--- a/.github/workflows/bot_pr_create.yaml
+++ b/.github/workflows/bot_pr_create.yaml
@@ -37,8 +37,8 @@ jobs:
     steps:
       - name: Get vLLM version
         run: |
-          VLLM_COMMIT=dc917cceb877dfd13f98c538c4c96158047d98bd
-          echo "VLLM_COMMIT=https://github.com/vllm-project/vllm/commit/$VLLM_COMMIT" >> $GITHUB_ENV
+          VLLM_COMMIT=v0.15.0
+          echo "VLLM_COMMIT=https://github.com/vllm-project/vllm/commit/$VLLM_COMMIT" >> "$GITHUB_ENV"
 
       - name: Checkout repository
         uses: actions/checkout@0c366fd6a839edf440554fa01a7085ccba70ac98 # v4.2.2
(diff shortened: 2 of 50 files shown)


tests/e2e/multicard/long_sequence/test_chunked_prefill_cp.py
```diff
@@ -0,0 +1,154 @@
...
+    "deepseek-ai/DeepSeek-V2-Lite-Chat": {
+        "TP": 2,
+        "PCP": 2,
+        "DCP": 2,
+        "quantization": None,
+    }
+}
+
+# A prompt containing a large markdown table. The table is randomly generated by GPT-4.
+LONG_PROMPT = "You are a helpful assistant in recognizes the content of tables in markdown format. Here is a table as follows.\n# Table\n" + """
```
MengqingCao, line 18: let's make an unify defination of `LONG_PROMPT` to avoid redundant code, you can refer to https://github.com/vllm-project/vllm/blob/8cef137689170338fdd573838414fc7f9e3ec84a/tests/conftest.py#L82

tests/e2e/multicard/long_sequence/test_chunked_prefill_cp.py
```diff
@@ -0,0 +1,154 @@
...
+            prefill_context_parallel_size=SETTINGS[model]['PCP'],
+            decode_context_parallel_size=SETTINGS[model]['DCP']) as vllm_model:
+        vllm_model.generate_greedy(INPUT_PROMPTS, max_tokens)
+
+
+@pytest.mark.skipif(vllm_version_is('0.12.0'),
+                    reason="0.12.0 is not supported for context sequence.")
+@pytest.mark.parametrize("model", MODELS)
+@pytest.mark.parametrize("max_tokens", [50])
+def test_models_prefix_cache_with_cp_piece_wise(model: str,
```
MengqingCao:
```suggestion
def test_models_prefix_cache_with_cp_piecewise(model: str,
```

tests/e2e/multicard/long_sequence/test_chunked_prefill_cp.py
```diff
@@ -0,0 +1,154 @@
...
+            model,
+            max_model_len=4096,
+            enforce_eager=True,
+            max_num_batched_tokens=128,
+            enable_expert_parallel=True,
+            tensor_parallel_size=SETTINGS[model]['TP'],
+            quantization=SETTINGS[model]["quantization"],
+            prefill_context_parallel_size=SETTINGS[model]['PCP'],
+            decode_context_parallel_size=SETTINGS[model]['DCP']) as vllm_model:
+        vllm_model.generate_greedy(INPUT_PROMPTS, max_tokens)
```
MengqingCao: I think we could merge the 3 cases, and compare the output with each others' to make sure the accuracy is correct

tests/e2e/multicard/long_sequence/test_prefix_caching_cp.py
```diff
@@ -0,0 +1,151 @@
...
+]
+
+
+@pytest.mark.skipif(vllm_version_is('0.12.0'),
+                    reason="0.12.0 is not supported for context sequence.")
+@pytest.mark.parametrize("model", MODELS)
+@pytest.mark.parametrize("max_tokens", [50])
+def test_models_prefix_cache_with_cp_basic(model: str,
+                                           max_tokens: int) -> None:
+    with VllmRunner(
```
MengqingCao, line 68: ditto

tests/e2e/multicard/long_sequence/test_prefix_caching_cp.py
```diff
@@ -0,0 +1,151 @@
...
+            model,
+            max_model_len=4096,
+            enforce_eager=True,
+            enable_expert_parallel=True,
+            tensor_parallel_size=SETTINGS[model]['TP'],
+            quantization=SETTINGS[model]["quantization"],
+            prefill_context_parallel_size=SETTINGS[model]['PCP'],
+            decode_context_parallel_size=SETTINGS[model]['DCP']) as vllm_model:
```
MengqingCao, line 79: It seems the prefix caching is always enabled, what's the different with `tests/e2e/multicard/long_sequence/test_prefix_caching_cp.py` and `tests/e2e/multicard/long_sequence/test_chunked_prefill_cp.py`

tests/e2e/multicard/long_sequence/test_chunked_prefill_cp.py
ader47: I created a unified definition of long prompt in `tests/e2e/prompts/long_prompt.txt` for both tests to reference

tests/e2e/multicard/long_sequence/test_prefix_caching_cp.py
ader47: Thank you for pointing this out. You're correct that both tests were previously enabling prefix caching. The key difference was that test_chunked_prefill_cp.py sets max_num_batched_tokens to make prefill work in chunked manner.
However, as you noticed, both tests had prefix caching enabled, which created overlapping functionality. To have clearer separation of concerns, I've now:
- Disabled prefix caching in the chunked prefill test
- Disabled chunked prefill in the prefix caching test

Each test now focuses on testing only one specific feature independently.

tests/e2e/multicard/long_sequence/test_chunked_prefill_cp.py
ader47: Thank you for the suggestion. While comparing outputs between different configurations would be valuable for accuracy verification, this PR is focused on functional testing - specifically verifying that each feature (chunked prefill, prefix caching) works correctly when enabled.

MengqingCao: I think we can do the accuracy verification togather in this pr, I think it won't be too complex, WDYT?

MengqingCao:
I checked the time increased by this pr is about 40 min, this cost is huge

wangxiyuan:

PR #5154, opened 2025-12-18
[BugFix]Fix wrong _cos, _sin instantiation

Angazenn:
This PR add additional check on creating global `_cos` and `_sin`, avoid creating them when using `mrope` or encoder-decoder model.

- vLLM version: v0.12.0


vllm_ascend/ops/rotary_embedding.py
```diff
@@ -78,7 +78,9 @@ def set_cos_and_sin(vllm_config, max_num_reqs, decode_token_per_req, dtype,
                                rope_dim,
                                dtype=dtype,
                                device=device)
-    elif not is_vl_model(vllm_config) and not vllm_config.model_config.use_mla:
+    elif not vllm_config.model_config.is_encoder_decoder and \
+        "mrope_section" not in model_config.hf_text_config.rope_parameters and \
+        not vllm_config.model_config.use_mla:
         rope_dim = model_config.get_head_size()
```
weijinqian0, line 83: If there are three or more conditions, you can abstract a function.

Angazenn: yes, this can be improved

PR #5172, opened 2025-12-18
[Doc][P/D] Fix MooncakeConnector's name

nwpu-zxr:
vLLM community has integrated their MooncakeConnector. The original scripts will now find this MooncakeConnector instead of the one from vLLM-Ascend. All scripts that involve using the MooncakeConnector need to be modified to another name.

Yes, users need to use a new name to load vLLM-Ascend MooncakeConnector.

By CI.

- vLLM version: v0.12.0


LCAIZJ:
Indeed, using MooncakeConnector now will use the connector from vLLM itself. This wasn't an issue before because vLLM didn't have a MooncakeConnector at that time, and vLLM-Ascend relied on the module_path to locate the specific connector implementation rather than connector name.

PR #5175, opened 2025-12-18
[Image] Refactor image build

Potabk:

In the past time, we used a hybrid architecture cross-compilation approach for image building. This method had a problem: cross-compilation performance was very poor, leading to extremely long build times(abort 4h) and even a probability of failure(see https://github.com/vllm-project/vllm-ascend/actions/runs/20152861650/job/57849208186). Therefore, I recommend using a separate architecture build followed by manifest merging, which significantly reduces image build time(20min).

- vLLM version: v0.12.0

### DIFF
--- a/.github/workflows/_pr_image_build.yaml
+++ b/.github/workflows/_pr_image_build.yaml
@@ -6,6 +6,11 @@ on:
         description: 'The tag subfix to use'
         required: true
         type: string
+      should_push:
+        description: 'Whether to push the image'
+        required: false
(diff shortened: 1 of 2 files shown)


Potabk:
https://github.com/nv-action/vllm-benchmarks/actions/runs/20358096054/job/58497645337 test it in my repo

Potabk:
alse cc @Yikun

wangxiyuan (approved):

PR #5187, opened 2025-12-19
[CI] Use offline mode for nightly test

Potabk:
For single node test, the lack of a retry mechanism for accessing ModelScope resulted in an HTTP 400 error sometimes. I recommend using a local offline cache instead.

- vLLM version: v0.12.0


.github/workflows/_e2e_nightly_single_node.yaml
```diff
@@ -57,6 +57,9 @@ jobs:
     timeout-minutes: 600
     container:
       image: ${{ inputs.image }}
+    env:
+      TRANSFORMERS_OFFLINE: 1
```
wangxiyuan, line 61: HF_HUB_OFFLINE?

Potabk: both ok

wangxiyuan: keep the same in the whole repo

Potabk: yes, we have keep the same with https://github.com/vllm-project/vllm-ascend/blob/636265be6dcb80a8d1c13f8b8a1865893f19cf7e/.github/workflows/_e2e_test.yaml#L28

.github/workflows/pr_tag_image_build_and_push.yaml
```diff
@@ -26,7 +26,7 @@ on:
       - 'cmake/**'
       - 'CMakeLists.txt'
       - 'csrc/**'
-    types: [ labeled, synchronize ]
+    types: [ labeled ]
```
MengqingCao, line 29: QQ: why we remove sync here?

PR #5192, opened 2025-12-19
[perf] Fix MLAPO weight disposal for KV-consumer MLA in PD-mix deploy...

kiscad:

- Problem: In MLA+MLAPO, KV-consumer deployments keep fused_qkv_a_proj/q_proj weights and quant params even though MLAPO uses the prepacked buffers, increasing memory footprint on decode nodes.
- Fix: Conditionally drop those tensors only when `kv_transfer_config.is_kv_consumer` to reclaim memory (consistent with the SFA behavior #4774 ).

- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -909,6 +909,21 @@ class AscendMLAImpl(MLAAttentionImpl):
         self.ctkv_scale = torch.tensor([1], dtype=act_dtype, device=device)
         self.q_nope_scale = torch.tensor([1], dtype=act_dtype, device=device)
 
+        # On KV consumers (decode-only) MLAPO uses the transformed weights built above;
+        # the original fused_qkv_a_proj/q_proj weights and quant params are no longer
+        # referenced, so drop them to save memory.
+        ascend_config = get_ascend_config()
+        if self.vllm_config.kv_transfer_config is not None and \
+                self.vllm_config.kv_transfer_config.is_kv_consumer and \
+                ascend_config.recompute_scheduler_enable:
+            self.fused_qkv_a_proj.weight = None
+            self.fused_qkv_a_proj.deq_scale = None
+            self.fused_qkv_a_proj.quant_bias = None
+            self.q_proj.weight = None
+            self.q_proj.deq_scale = None
+            self.q_proj.quant_bias = None
+            torch.npu.empty_cache()
+
     def get_context_seq_len_npu(self, index: int,
                                 attn_metadata: AscendMLAMetadata):
         prefill_metadata = attn_metadata.prefill
--- a/vllm_ascend/attention/sfa_v1.py
+++ b/vllm_ascend/attention/sfa_v1.py
@@ -371,7 +371,7 @@ class AscendSFAImpl(MLAAttentionImpl):
         if self.enable_sfa_cp:
             self.local_num_heads = self.num_heads * self.tp_size
 
-            #TODO: Temporarily adapt sfa-cp, remove after adapting near PCP. --clrs97
+            # TODO: Temporarily adapt sfa-cp, remove after adapting near PCP. --clrs97
             self._replace_linear_class_for_sfa_cp()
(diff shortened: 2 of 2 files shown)


jianzs:
If this step is taken, it means the decoding node must skip the prefill stage. How is this currently guaranteed? What happens if the request is preempted and needs to be computed?

kiscad:
> If this step is taken, it means the decoding node must skip the prefill stage. How is this currently guaranteed? What happens if the request is preempted and needs to be computed?

Decoding node may never enter the prefilling stage.

ZYang6263:
Similar to SFA, this PR releases unused memory.

jianzs:
> > If this step is taken, it means the decoding node must skip the prefill stage. How is this currently guaranteed? What happens if the request is preempted and needs to be computed?
> 
> Decoding node may never enter the prefilling stage.

How can we ensure this assumption holds? What if some requests get preempted due to insufficient KV cache?

vllm_ascend/utils.py
```diff
@@ -939,9 +939,6 @@ def get_hccl_config_for_pg_options(group_name: str) -> Optional[dict]:
         "dp": {
             "hccl_buffer_size": calculate_dp_buffer_size()
         },
-        "ep": {
-            "hccl_buffer_size": calculate_ep_buffer_size()
-        },
```
jianzs: What does this code have to do with the current PR?

vllm_ascend/attention/mla_v1.py
```diff
@@ -924,6 +924,16 @@ def _process_weights_for_fused_mlapo(self, act_dtype: torch.dtype):
         self.ctkv_scale = torch.tensor([1], dtype=act_dtype, device=device)
         self.q_nope_scale = torch.tensor([1], dtype=act_dtype, device=device)
 
+        if self.vllm_config.kv_transfer_config is not None and \
```
zzzzwwjj, line 916: Please add note about why we can do this?

vllm_ascend/utils.py
kiscad: Actually not. I will put it in another PR.

vllm_ascend/attention/mla_v1.py
kiscad: done

kiscad:
> > > If this step is taken, it means the decoding node must skip the prefill stage. How is this currently guaranteed? What happens if the request is preempted and needs to be computed?
> > 
> > 
> > Decoding node may never enter the prefilling stage.
> 
> How can we ensure this assumption holds? What if some requests get preempted due to insufficient KV cache?

For KV-cache insufficiency/preemption: on consumer nodes we rely on `recompute_scheduler_enable`. `RecomputeScheduler` handles allocation failures by dropping the preempted request back to the PD proxy (`stop_reason="recomputed"`) instead of recomputing locally, so the prompt/prefill gets recomputed on a producer and KV is reloaded before decode resumes.

jianzs:
> > > > If this step is taken, it means the decoding node must skip the prefill stage. How is this currently guaranteed? What happens if the request is preempted and needs to be computed?
> > > 
> > > 
> > > Decoding node may never enter the prefilling stage.
> > 
> > 
> > How can we ensure this assumption holds? What if some requests get preempted due to insufficient KV cache?
> 
> For KV-cache insufficiency/preemption: on consumer nodes we rely on `recompute_scheduler_enable`. `RecomputeScheduler` handles allocation failures by dropping the preempted request back to the PD proxy (`stop_reason="recomputed"`) instead of recomputing locally, so the prompt/prefill gets recomputed on a producer and KV is reloaded before decode resumes.

The `recomputed` feature is currently only available in the vllm-ascend proxy's examples folder and is not yet a stable solution, so it's not ready for production use.

zzzzwwjj (approved):

jianzs:
It would be better to adjust the condition for #4774 as well.

kiscad:
> It would be better to adjust the condition for #4774 as well.

Yea, good suggestion.

Issue #4084, opened 2025-11-10
[RFC]: Remove VL Modeling Files


shen-shanshan:
### Motivation.

To avoid maintaining a variety of models, we propose to remove all modeling files in vllm-ascend. To reach this, there are some refactors need to be done for multi-modal models in both vllm and vllm-ascend.

### Proposed Change.

**vllm:**

- [ ] Extract ViT backend base class @shen-shanshan
- [ ] Extract PatchEmbed layer base class @shen-shanshan
- [ ] Refactor `Qwen2_5_VisionRotaryEmbedding` and `Qwen2VisionRotaryEmbedding` to use `RotaryEmbedding` as their base class @gcanlin
- [ ] Remove redundant TP logic in split_qkv @gcanlin https://github.com/vllm-project/vllm/pull/28271
- [ ] Make mamba backend pluggable @shen-shanshan https://github.com/vllm-project/vllm/pull/26487

**vllm-ascend:**

- [ ] Implement ascend ViT backend and related interface in platform @shen-shanshan
- [ ] Move FA padding logic into ViT forward @shen-shanshan
- [ ] Implement ascend PatchEmbed custom op and register it @shen-shanshan

### CC List.

@Yikun @wangxiyuan @gcanlin

Yikun:
@shen-shanshan Thanks for your effort and Many thanks for your help @Isotr0py .

PR #5196, opened 2025-12-19
[Bugfix] Implement multimodal_cpu_fields in model runner

zhangxinyuehfad:
Related to https://github.com/vllm-project/vllm-ascend/issues/4084 Implement multimodal_cpu_fields in model runner

- vLLM version: v0.12.0


shen-shanshan:
Please delete the whole file `patch_qwen3_vl.py`.

shen-shanshan:
According to https://github.com/vllm-project/vllm/pull/28168, I suppose this `multimodal_cpu_fields` is different for every model, we should not just use one `multimodal_cpu_fields` for all MM models.

Besides, we should avoid letting users to manually configure this by passing an `additional_config`, which could lead to an awful user experience.

shen-shanshan:
Currently, this `multimodal_cpu_fields` is only used by Qwen series VL models, and the `multimodal_cpu_fields` of these models are  all the same. Thus, your resolution could work temporarily, but I suppose this is not a good choice to implement the function like this for long term.

shen-shanshan:
LGTM if all CI passed.

Note that we need to remove this logic after this field is deprecated in vllm.

zhangxinyuehfad:
> Please delete the whole file `patch_qwen3_vl.py`.

done

zhangxinyuehfad:
> According to [vllm-project/vllm#28168](https://github.com/vllm-project/vllm/pull/28168), I suppose this `multimodal_cpu_fields` is different for every model, we should not just use one `multimodal_cpu_fields` for all MM models.
> 
> Besides, we should avoid letting users to manually configure this by passing an `additional_config`, which could lead to an awful user experience.

Yes, I deleted multimodal_cpu_fields from additional_config.

shen-shanshan:
CC @wangxiyuan All CI passed.

PR #5203, opened 2025-12-19
[Refactor] move the metadata from attention_v1 to util(ready for extract common_cp) & realize Ascendmetadata inherit from the parent class.

weijinqian0:
RFC: https://github.com/vllm-project/vllm-ascend/issues/4629

1. Remove the pcp-related code from attention_v1.
2. Establish the inheritance relationship of CommonAttentionMetadata.

TODO
1. extract common_cp
2. move cp metadata to common_cp.
3. remove commonAttentionMetadata for aclgraph.

- vLLM version: v0.12.0


vllm_ascend/attention/utils.py
```diff
@@ -69,45 +115,20 @@ class AscendPrefillContextParallelMetadata:
...
-    decode_token_per_req: int
+    decode_token_per_req: int = 1
     """decode token number per request"""
 
-    block_table_tensor: torch.Tensor
-
-    slot_mapping: torch.Tensor
-
-    actual_seq_lengths_q: list[int]
+    actual_seq_lengths_q: list[int] = field(default_factory=list)
```
yiz-liu, line 137: We should at least leave some comments here to explain why we need these extra params?

weijinqian0: it will be added next pr with other cp metadata.

PR #5205, opened 2025-12-19
[Bugfix] Use hf_text_config instead of hf_config to support multimodal PD-Disaggregated

ApsarasX:
In code files such as`mooncake_connector.py`, `vllm_config.model_config.hf_config` is used to get the LLM configs. This approach works for LLMs, but not for multi-modal models. For multi-modal models, `vllm_config.model_config.hf_text_config` must be used instead to get the LLM configs.

No.

Existing UT
- vLLM version: v0.12.0

### DIFF
--- a/vllm_ascend/distributed/mooncake_connector.py
+++ b/vllm_ascend/distributed/mooncake_connector.py
@@ -506,7 +506,8 @@ class KVCacheRecvingThread(threading.Thread):
         head_dim = self.model_config.hf_text_config.head_dim
         block_size = self.vllm_config.cache_config.block_size
         num_kv_head = max(
-            self.model_config.hf_text_config.num_key_value_heads // self.tp_size, 1)
+            self.model_config.hf_text_config.num_key_value_heads //
+            self.tp_size, 1)
 
         flat_block_ids = [item for sublist in block_ids for item in sublist]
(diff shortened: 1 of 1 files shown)


ApsarasX:
BTW, I found that in many other places in vllm-ascend, `hf_config` rather than `hf_text_config` is used, but since those code sections were not executed, they didn't cause any errors.

So, I suggest that @shen-shanshan  replace all  `hf_config` in the vllm‑ascend with `hf_text_config`.

cc @wangxiyuan

shen-shanshan:
LGTM.

CC @wangxiyuan Find more details at https://github.com/vllm-project/vllm/blob/560ae9638c00e761d6f1ca33882249dbfebbe8aa/vllm/config/model.py#L475 and https://github.com/vllm-project/vllm/blob/560ae9638c00e761d6f1ca33882249dbfebbe8aa/vllm/transformers_utils/config.py#L966-L980.

shen-shanshan:
> BTW, I found that in many other places in vllm-ascend, `hf_config` rather than `hf_text_config` is used, but since those code sections were not executed, they didn't cause any errors.
> 
> So, I suggest that @shen-shanshan replace all `hf_config` in the vllm‑ascend with `hf_text_config`.
> 
> cc @wangxiyuan

OK, I will do that next week.

LCAIZJ:
LGTM

PR #5094, opened 2025-12-16
[main][Refactor] Remove `with_prefill` parameter from `set_ascend_forward_context`

SlightwindSec:
Removes the redundant `with_prefill` parameter from `set_ascend_forward_context` to align the interface with vLLM's `set_forward_context` for future refactoring.
- vLLM version: v0.12.0
- vLLM main: https://github.com/vllm-project/vllm/commit/ad32e3e19ccf0526cb6744a5fed09a138a5fb2f9

PR #5241, opened 2025-12-22
[Refactor] Remove `with_prefill` flag

yiz-liu:
Eliminates the global prefill status flag from the forward context, favoring local determination based on attention metadata. This change improves the reliability of prefill detection across different attention implementation paths by relying on per-request metadata rather than global state.

Also integrates a specialized Eagle proposer for speculative decoding to improve support for specific hardware configurations.

None.

None.


yiz-liu:
This is a duplicate PR of #5094 and should be closed.

PR #5247, opened 2025-12-22
[Bugfix] fix pcp 128K break

weiguihua2:
In the pcp scenario, when the profile is running, the max token should be divided by the pcp size to reduce the memory usage; otherwise, the pcp 128K does not work.

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -193,6 +193,7 @@ class NPUModelRunner(GPUModelRunner):
     def __init__(self, vllm_config: VllmConfig, device: torch.device):
         with _torch_cuda_wrapper():
             super().__init__(vllm_config, device)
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2191,7 +2180,11 @@ def profile_run(self) -> None:
             self._dummy_run(mc2_tokens_capacity,
                             with_prefill=True,
                             is_profile=True)
+        origin_max_num_tokens = self.max_num_tokens
+        if self.pcp_size > 1:
+            self.max_num_tokens = self.max_num_tokens // self.pcp_size
         super().profile_run()
+        self.max_num_tokens = origin_max_num_tokens
```
wangxiyuan, line 2187: I'm not very understand why the code logic is so complex. Is it OK to just like:

```
if self.pcp_size > 1:
    self.max_num_tokens = self.max_num_tokens // self.pcp_size
super().profile_run()
```
The other two lines can be removed?

PR #5249, opened 2025-12-22
[Bugfix] Fix mm_merge

Potabk:
We should transfer the mm_embed to the dtype of input_embed before performing the in-place assignment

### DIFF
--- a/vllm_ascend/patch/worker/patch_multimodal_merge.py
+++ b/vllm_ascend/patch/worker/patch_multimodal_merge.py
@@ -37,8 +37,9 @@ def _merge_multimodal_embeddings(
         This updates ``inputs_embeds`` in place.
     """
     flattened = _flatten_embeddings(multimodal_embeddings)
+    input_dtype = inputs_embeds.dtype
     try:
-        inputs_embeds[is_multimodal] = flattened
+        inputs_embeds[is_multimodal] = flattened.to(dtype=input_dtype)
     except RuntimeError as e:
         num_expected_tokens = is_multimodal.sum().item()
         assert isinstance(num_expected_tokens, int)


ApsarasX:
Will there be any accuracy problems without this PR?

So is this a Bugfix or a Misc?

BTW, typo in PR title: `Msic` -> `Misc`

Potabk:
@ApsarasX For some models, eg: `bagel`, the mm_embeds's dtype are different from input_embeds,  so without this pr, there exist functional issue

Potabk:
also cc @booker123456 @gcanlin @shen-shanshan

shen-shanshan:
Are there any issues had recorded the error without this PR? I'm not sure about the background of this PR.

Potabk:
test locally with the script https://github.com/vllm-project/vllm/blob/main/examples/offline_inference/vision_language.py
`python examples/offline_inference/vision_language.py --model-type bagel` not work

Potabk:
The root cause is in this `bagel` modeling implementation, the dtype of `mm_embeds` and `input_embed` not match

gcanlin (approved):
LGTM if CI can pass, vllm also has this dtype convert.

booker123456:
LGTM. Compatible with data types is reasonable.

vllm_ascend/patch/worker/patch_multimodal_merge.py
```diff
@@ -37,8 +37,9 @@ def _merge_multimodal_embeddings(
         This updates ``inputs_embeds`` in place.
```
wangxiyuan, line 37: any plan to remove this patch?

gcanlin: I think that we should request for torch_npu/CANN team to support `torch.Tensor.masked_scatter_` then we can remove this patch.

Potabk: After communicating offline with the author of this patch, I learned that it was added for performance reasons. The original `mask_scatter` operator has no functional issues. Therefore, we may need to push for the addition of a new ascend branch upstream.

gcanlin: After testing on NPU, it really doesn't have functional issues. @booker123456 is there any performance test for this patch change?

gcanlin: I suggest we consider directly removing this patch to reduce the maintaining cost. It seems that it can't take much performance. @booker123456 WDYT?

ApsarasX: I think this patch is still necessary until torch_npu's `masked_scatter_` performance catches up with index put.

Potabk:
So can we merge it? @wangxiyuan

wangxiyuan (approved):

PR #5250, opened 2025-12-22
fix transformer version to 4.57.3

shaopeng-666:
In certain scenarios (such as smoke testing), the source code is used to update the vllm-ascend version for running updated models (such as Qwen3-VL). However, vllm and vllm-ascend themselves have no restrictions on the transformer version, and the transformer will not be updated, resulting in errors when launching the model.
No

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,5 +30,5 @@ numba
 #--extra-index-url https://mirrors.huaweicloud.com/ascend/repos/pypi
 torch-npu==2.8.0
 
-transformers==4.57.3
+transformers>=4.57.3
 fastapi<0.124.0


weijinqian0:
ready

Yikun:
1. Please fullfiil the commit msg to show why upgrade this

2. should we just removed version limited?

shaopeng-666:
> 1. Please fullfiil the commit msg to show why upgrade this
> 2. should we just removed version limited?
PR description has been updated.

requirements.txt
```diff
@@ -30,5 +30,5 @@ numba
 #--extra-index-url https://mirrors.huaweicloud.com/ascend/repos/pypi
 torch-npu==2.8.0
 
-transformers<=4.57.1
+transformers==4.57.3
```
Yikun: Could we just remove this line and just follow vllm default version?

https://github.com/vllm-project/vllm/blob/b57b967386e1962955c93b7cb39828448b68789b/requirements/common.txt#L10

shaopeng-666: Following vllm could not resolve the problem I noticed in commit msg.It still could not update the transformer version when i update my vllm using source code.

PR #5266, opened 2025-12-23
[Bugfix] fix pcp 128K break

weiguihua2:
[Bugfix] Fixing the issue where 128K context does not work in long sequence scenarios. 

This issue is caused by not splitting num_token according to pcp_size during profile_run.
During `profile_run`, a warm-up is performed based on `self.max_num_tokens`. When PCP is enabled, each PCP group will only schedule up to `self.max_num_tokens / pcp_size`. After `profile_run` is completed, the original scheduling size needs to be restored. 

This is a temporary workaround; once https://github.com/vllm-project/vllm/pull/28988/files is implemented, this part can be removed.


zzzzwwjj (approved):

zzzzwwjj:
Same as https://github.com/vllm-project/vllm-ascend/pull/5247
cc @wangxiyuan

wangxiyuan (approved):

MengqingCao:
plz make more description on what the bug this pr fix, and why does this pr fixes it. We should make this pr more clear to everyone checks this.

weiguihua2:
I'm not very understand why the code logic is so complex. Is it OK to just like:

if self.pcp_size > 1:
    self.max_num_tokens = self.max_num_tokens // self.pcp_size
super().profile_run()
The other two lines can be removed?

--- -------
The semantics of max_num_tokens cannot be changed. Only during profile_run, max_num_tokens needs to be evenly distributed across each card's computing resources.
@wangxiyuan

weiguihua2:
> plz make more description on what the bug this pr fix, and why does this pr fixes it. We should make this pr more clear to everyone checks this.

done

PR #5286, opened 2025-12-23
[TEST]Add sending request with and without chat

jiangyunfan1:
This PR adds the method for sending chat and non-chat request, we need it to test much folloing cases.

No

by running the test

### DIFF
--- a/tools/send_request.py
+++ b/tools/send_request.py
@@ -1,19 +1,6 @@
-import os
 from typing import Any
 
 import requests
-from modelscope import snapshot_download  # type: ignore
-
-
-def get_prompt_from_dataset(dataset_path, file_name):
-    if os.path.isabs(dataset_path):
-        dataset_dir = dataset_path
-    else:
-        dataset_dir = snapshot_download(dataset_path, repo_type='dataset')
-    file_path = os.path.join(dataset_dir, file_name)
-    with open(file_path, "r", encoding="utf-8") as f:
-        content = f.read()
-    return content
 
 
 def send_v1_completions(prompt, model, server, request_args=None):


tools/send_request.py
```diff
@@ -1,23 +1,47 @@
...
 
-data: dict[str, Any] = {
-    "messages": [{
-        "role": "user",
-        "content": "",
-    }],
-}
 
+def get_prompt_from_dataset(dataset_path, file_name):
+    dataset_dir = snapshot_download(dataset_path, repo_type='dataset')
```
MengqingCao: @Potabk please take a look at this

tools/send_request.py
```diff
@@ -1,23 +1,50 @@
...
 
+def get_prompt_from_dataset(dataset_path, file_name):
+    if os.path.isabs(dataset_path):
+        dataset_dir = dataset_path
+    else:
+        dataset_dir = snapshot_download(dataset_path, repo_type='dataset')
+    file_path = os.path.join(dataset_dir, file_name)
+    with open(file_path, "r", encoding="utf-8") as f:
+        content = f.read()
+    return content
```
Potabk:
```python
    def load_data(self) -> None:
        if self.dataset_path is None:
            raise ValueError("dataset_path must be provided for loading data.")

            for _, row in jsonl_data.iterrows():
                self.data.append(row.to_dict())
        else:
            raise NotImplementedError(
                "Only JSONL format is supported for CustomDataset."
            )

        random.seed(self.random_seed)
        if not getattr(self, "disable_shuffle", False):
            random.shuffle(self.data)
```

Potabk: there may have some reference: https://github.com/vllm-project/vllm/blob/ddfac7034e3c30518a0588f568094c8f7a4b80a4/vllm/benchmarks/datasets.py#L1942

tools/send_request.py
Potabk: This is ok for the first time download from internet

tools/send_request.py
Potabk: For robustness and production-ready testing, perhaps we should add a `shuffle` to simulate a production prompt.

jiangyunfan1: This is not expected to load a dataset with more than one prompt, as for now, we've decided a better way to test single 128K without semantics, and this method will be removed.

PR #5294, opened 2025-12-23
[E2E] Optimize the E2E test time.

menogrey:
Add cudagraph_capture_sizes for E2E CI test.

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -115,7 +115,8 @@ jobs:
           pytest -sv --durations=0 tests/e2e/singlecard/test_sampler.py
           pytest -sv --durations=0 tests/e2e/singlecard/test_vlm.py
           pytest -sv --durations=0 tests/e2e/singlecard/test_xlite.py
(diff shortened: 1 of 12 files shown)


tests/e2e/multicard/long_sequence/test_basic.py
wangxiyuan, line 130: is it possible to add the default value for VllmRunner directlly?

menogrey: I tried it, but meet some error, one problem is that you have to deal with all possible situation in the `VllmRunner`, such as 
- `cudagraph_capture_sizes` and `compilation.cudagraph_capture_sizes` must be consistent
- `max_cudagraph_capture_sizes` must be the maximum of the `cudagraph_capture_sizes`
- User input of `compilation` argument, compatible with `dict` and `CompilationConfig` class

And change the default value may cause user's confusion. In some test case such as `test_aclgraph_mem.py`, i have to use the original default sizes from 1 to 512.

It seems the `cudagraph_capture_sizes` change will also cause some accuracy issue in some test cases, like two output not equal.

So I decided to use one conservative plan to do it.

PR #5298, opened 2025-12-23
[bugfix] solve dp scenario Host-Device sync

hwhaokun:
In the speculative decoding scenario, the original code performs Host-Device synchronization, which slows down the main model's execution speed. 

### DIFF
--- a/docs/source/tutorials/long_sequence_context_parallel_multi_node.md
+++ b/docs/source/tutorials/long_sequence_context_parallel_multi_node.md
@@ -2,6 +2,10 @@
 
 ## Getting Start
 
+:::{note}
+Context parallel feature currently is only supported on Atlas A3 device, and will be supported on Atlas A2 in the future.
+:::
+
 vLLM-Ascend now supports long sequence with context parallel options. This guide takes one-by-one steps to verify these features with constrained resources.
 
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1856,10 +1856,11 @@ def _build_dummy_attn_metadata(
                 # QUESTION: Why do we separately set query_start_loc for spec in the first place?
                 # While in _prepare_inputs we don't?
                 if self.speculative_config:
-                    self.query_start_loc.gpu[:num_reqs + 1] = torch.tensor(
+                    self.query_start_loc.cpu[:num_reqs + 1] = torch.tensor(
                         [0] + self.actual_seq_lengths_q[:num_reqs],
-                        device=self.device,
+                        device="cpu",
                         dtype=torch.int32)
+                self.query_start_loc.copy_to_gpu()
```
realliujiaxu: move this line under `if self.speculative_config:`

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1856,10 +1856,11 @@ def _build_dummy_attn_metadata(
                 # QUESTION: Why do we separately set query_start_loc for spec in the first place?
                 # While in _prepare_inputs we don't?
                 if self.speculative_config:
-                    self.query_start_loc.gpu[:num_reqs + 1] = torch.tensor(
+                    self.query_start_loc.cpu[:num_reqs + 1] = torch.tensor(
                         [0] + self.actual_seq_lengths_q[:num_reqs],
-                        device=self.device,
+                        device="cpu",
                         dtype=torch.int32)
+                    self.query_start_loc.copy_to_gpu()
```
yiz-liu, line 1870: Modify `self.query_start_loc.np` instead, it should be more efficient.

yiz-liu: Also, this will break the `pin_memory` beween `np` and `cpu`, I think we should avoid it.

yiz-liu (approved):
After discussion, this `actual_seq_lengths_q` should be removed later @weijinqian0 , and hopefully this condition can be removed too by then. For now we'll just leave it be.

PR #5168, opened 2025-12-18
[KVPOOL]decode save kvcache

baxingpiaochong:
### What this PR does / why we need it?

kvpool decode save kvcache
now only support mla

### How was this patch tested?

- vLLM version: v0.12.0
- vLLM main: https://github.com/vllm-project/vllm/commit/ad32e3e19ccf0526cb6744a5fed09a138a5fb2f9

PR #5316, opened 2025-12-24
[P/D] Support decode rank store its kvcache to mooncake store so that prefill rank can use

cocacolafan:
1.  support decode rank store it's kvcache to mooncake store so that prefill rank can use
2.  clean code for mooncake_connector.py and ascend_store_connector part

### DIFF
--- a/docs/source/user_guide/support_matrix/supported_models.md
+++ b/docs/source/user_guide/support_matrix/supported_models.md
@@ -6,51 +6,52 @@ Get the latest info here: https://github.com/vllm-project/vllm-ascend/issues/160
 
 ### Generative Models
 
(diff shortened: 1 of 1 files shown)


LCAIZJ:
There's already a PR addressing this: https://github.com/vllm-project/vllm-ascend/pull/5168, you can compare the differences.

cocacolafan:
> There's already a PR addressing this: #5168, you can compare the differences.
1. this pr will enable decode store to kvpool under the condition both GQA and MLA, same with prefill rank can do.
2. due to prefill rank can store its kvcache into the pool，there is no need additional operations, because whether current rank is p or d, they use the same code, just need a little change, the decode rank can achieve store function too.

LCAIZJ:
> 2\. additional

Could you please fix the formatting issues first? It looks like many of the code diffs are due to formatting changes rather than actual content changes.

PR #4138, opened 2025-11-12
[KV-Sharing] Support KV-Sharing feature in CLA models

MengqingCao:
### What this PR does / why we need it?
Support KV-Sharing feature in CLA (cross layer attention) models, which sharing kv cache in some layers.

### How was this patch tested?
test with Gemma3n
```python
import os

os.environ["VLLM_USE_MODELSCOPE"] = "True"
os.environ["VLLM_WORKER_MULTIPROC_METHOD"] = "spawn"
os.environ["VLLM_VERSION"] = "0.12.0"

from vllm import LLM, SamplingParams
def main():
    prompts = [
        "窗前明月光，",
        "The president of the United States is Mr.",
        "The capital of France is",
        "The future of AI is",
        "感时花溅泪，",
        "家书抵万金啥意思？",
        "plz tell me a story: ",
    ]

    # Create a sampling params object.
    sampling_params = SamplingParams(max_tokens=100, temperature=0.6, top_k=40, top_p=0.95)

    # Create an LLM.
    llm = LLM(
        model="/home/cache/modelscope/models/google/gemma-3n-E2B-it",
              tensor_parallel_size=4,
              trust_remote_code=True,
              max_model_len=1024,
              gpu_memory_utilization=0.8,
              cuda_graph_sizes=[16]
              )

    # Generate texts from the prompts.
    outputs = llm.generate(prompts, sampling_params)
    for output in outputs:
        prompt = output.prompt
        generated_text = output.outputs[0].text
        print(f"Prompt: {prompt!r}, Generated text: {generated_text!r}")

if __name__ == "__main__":
    main()

```

e9a942464e4663c79d86b44b421b99d2

- vLLM version: v0.12.0
- vLLM main: https://github.com/vllm-project/vllm/commit/ad32e3e19ccf0526cb6744a5fed09a138a5fb2f9

PR #5325, opened 2025-12-24
Redo "[KV-Sharing] Support KV-Sharing feature in CLA models (#4138)"

MengqingCao:
Support KV-Sharing feature in CLA (cross layer attention) models, which sharing kv cache in some layers.
This pr fixes shared_kv_cache_layers check. The original check of `shared_kv_cache_layers`, the default value of which is `{}`, will always be `True`, and always stepping into chunked prefill is not what we expected. This pr fixes it.

CI passed with existing test.

### DIFF
--- a/tests/e2e/singlecard/test_cross_layer_attn_model.py
+++ b/tests/e2e/singlecard/test_cross_layer_attn_model.py
@@ -50,7 +50,7 @@ def test_models_with_aclgraph(
             model,
             max_model_len=1024,
             enforce_eager=False,
-            cuda_graph_sizes=[4],
+            cudagraph_capture_sizes=[4],
     ) as vllm_model:
         vllm_aclgraph_outputs = vllm_model.generate_greedy(prompts, max_tokens)
 


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1195,7 +1195,7 @@ def _generate_process_reqs_hidden_states(self, maybe_padded_num_tokens,
 
     def _build_attn_state(self, num_reqs, num_scheduled_tokens,
                           num_valid_tokens):
-        if self.shared_kv_cache_layers is not None:
+        if self.shared_kv_cache_layers != {}:
```
zhenwenqi2024: We can directly determine whether a dictionary is empty. 
if self.shared_kv_cache_layers will be better. {} in python is False

MengqingCao: Fixed, that make sense, thx!

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1195,7 +1195,7 @@ def _generate_process_reqs_hidden_states(self, maybe_padded_num_tokens,
 
     def _build_attn_state(self, num_reqs, num_scheduled_tokens,
                           num_valid_tokens):
-        if self.shared_kv_cache_layers is not None:
+        if not self.shared_kv_cache_layers:
```
zhenwenqi2024: It is wrong here. here should not be chunkprefill for all scenes

wangxiyuan:

PR #5327, opened 2025-12-24
[CI] Add cache for CI `pip install`, optimize the run time.

menogrey:
Add cache for CI `pip install`, optimize the run time.


.github/workflows/_e2e_nightly_single_node_models.yaml
```diff
@@ -92,17 +92,27 @@ jobs:
           ref: ${{ inputs.vllm }}
           path: ./vllm-empty
 
+      - name: Cache pip
```
Yikun, line 95: I guess this will not work if no surprise because donwload cache from github remote cache to infra machine will cost more time.

menogrey: OK, i think may be we should try other way to do it.

menogrey:
From the log, i can find that downloading cache take about 10min for 50% of all, it is useless if the downloading and uploading time greater than the real pip install time. For the self-hosted runner, may be we should try other way to cache for CI. So this PR will be closed.

PR #5329, opened 2025-12-24
[Feature] Enhance all-reduce skipping logic for MoE models in NPUModelRunner

jianzs:

Besides enabling `recompute_scheduler_enable`, we can skip all_reduce when max_num_batched_tokens is below mc2's requirement.

No.


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -399,24 +399,48 @@ def _use_aclgraph(self) -> bool:
...
+        # For prefill, use the scheduler's max_num_batched_tokens for a single
+        # batch. NOTE: if tensor parallelism is enabled, each TP rank only holds
+        # a fraction of the tokens, but the current MoE communication-selection
+        # logic does not account for TP. Therefore we intentionally use the
+        # total token count here.
+        # tp_size = self.vllm_config.parallel_config.tensor_parallel_size
+        # max_batched_tokens_per_rank = (
+        #      self.vllm_config.scheduler_config.max_num_batched_tokens // tp_size
+        # )
+        prefill_must_use_mc2 = needs_mc2(
```
realliujiaxu, line 428: should `prefill_must_use_mc2` be decided by `num_tokens` dynamically ?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -399,24 +399,48 @@ def _use_aclgraph(self) -> bool:
...
-        return self.is_kv_consumer and self.ascend_config.recompute_scheduler_enable and select_moe_comm_method(
-            potential_max_num_tokens,
-            self.vllm_config) in {MoECommType.MC2, MoECommType.FUSED_MC2}
+            potential_max_tokens = self.max_num_reqs * self.uniform_decode_query_len
+        decode_must_use_mc2 = needs_mc2(potential_max_tokens)
+
+        # For prefill, use the scheduler's max_num_batched_tokens for a single
+        # batch. NOTE: if tensor parallelism is enabled, each TP rank only holds
+        # a fraction of the tokens, but the current MoE communication-selection
+        # logic does not account for TP. Therefore we intentionally use the
```
linfeng-yuan: I think current implementation without performing division is totally correct. Since set_mc2_tokens_capacity in ascend_forward_context.py has already considered `tp_size`. So we do not need to  consider the `tp_size` while calling `needs_mc2` and `select_moe_comm_method`. Could you please check the logic mentioned above and change the comments here?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -399,24 +399,48 @@ def _use_aclgraph(self) -> bool:
...
+        # max_batched_tokens_per_rank = (
+        #      self.vllm_config.scheduler_config.max_num_batched_tokens // tp_size
+        # )
+        prefill_must_use_mc2 = needs_mc2(
+            self.vllm_config.scheduler_config.max_num_batched_tokens)
+
+        # Skip all-reduce if decode requires MC2 and either prefill also
+        # requires MC2 or recompute-based scheduler is enabled.
+        return decode_must_use_mc2 and (
+            prefill_must_use_mc2
```
linfeng-yuan, line 434: LGTM. I think we only have to change the comments in L427 since `moe_comm_method` has considered `tp_size` and `num_tokens` for each DPEnginceCore. Just out of curiosity, what's the background behind implementation of `prefill_must_use_mc2`? I think this can enable prefill requests execution when `max_num_batched_tokens` < `vllm_config.compilation_config.max_cudagraph_capture_size` but not sure about the necessity and background.

vllm_ascend/worker/model_runner_v1.py
linfeng-yuan: > should `prefill_must_use_mc2` be decided by `num_tokens` dynamically ?

I think it's hard to use runtime attribute `num_tokens` here since we cannot ensure num_tokens among all ranks can perform same collective communication. So we can only use `max_num_batced_tokens` here.

vllm_ascend/worker/model_runner_v1.py
jianzs: Skipping all reduce operations can significantly boost performance, but the current implementation relies on a development feature called `recompute_scheduler_enable`, which only works with the proxy located in the examples folder. Since recompute is the production-ready method in this version, we can achieve similar performance gains in production by reducing the `max_num_batched_tokens` setting. If we assume preemption is rare, this would be a practical approach.

vllm_ascend/worker/model_runner_v1.py
jianzs: You're correct. I'll update these comments. But I have a question about the `set_mc2_tokens_capacity` function - it seems that when graph mode is enabled, the 512-token limit can be bypassed. Is that right?

vllm_ascend/worker/model_runner_v1.py
linfeng-yuan: Make sense. It's my fault that ignoring possibilities of other proxy options while designing the bypass mechanism of `all_reduce`. I think this pr can overcome this problem functionally in your production envrionment. And as you said, this would be a good practice without changing your customized proxy if preemption is rare. 

Beyond this PR, I also recommend to add simple handling logic of recomputation in your proxy since prefill execution with small `num_batched_tokens` on D Nodes would increase both TTFT and TPOT, or cause OOM issue because of `atb_ring_mla` calling.  That's the reason we modifed vLLM's original scheduler. And I believe the final solution is stable CPU KVCache offloading mechanism which is coming soon. Thanks for your addition~

vllm_ascend/worker/model_runner_v1.py
linfeng-yuan: > You're correct. I'll update these comments. But I have a question about the `set_mc2_tokens_capacity` function - it seems that when graph mode is enabled, the 512-token limit can be bypassed. Is that right?

As far as I know, max_cudagraph_capture_size is set to min(self.scheduler_config.max_num_seqs * decode_query_len * 2, 512) by default in _set_cudagraph_sizes function in VllmConfig if not initialized. I am not sure whether it would be set larger than 512 with abnormal user configuration or model setting. @yiz-liu Could you please have a check here? And 512-token-limit-per-rank of MC2 operators still exists.

vllm_ascend/worker/model_runner_v1.py
jianzs: Thank you for your suggestion. We're eager to see and use the stable CPU KVCache offloading mechanism. Do you know the latest progress?

PR #5360, opened 2025-12-25
[CI] Exclude broken nodes in nightly test

Potabk:
Exclude broken nodes in nightly test


Yikun (changes requested):
The internal IP shouldn't be here.

PR #5365, opened 2025-12-25
[Doc] Upgarde deepseek-v3.2 to use tokenizer mode

Potabk:
Upgarde deepseek-v3.2 to use tokenizer mode


Potabk:
@zhangxinyuehfad also cc

PR #5376, opened 2025-12-25
[EPLB]Eplb Config Renaming

shenchuxiaofugui:
1. Rename num_iterations_eplb_update to expert_heat_collection_interval
2. Rename num_wait_worker_iterations
3. Rename init_redundancy_expert to num_redundant_experts because the variable with the same meaning in vLLM is named this way.
4. Delete gate_eplb because we don't need this feature.
5. qwen3-30B-w8a8 e2e test use aclgraph
#### test dsr1 mtp eplb num_redundant_experts=16
{"id":"chatcmpl-b11629a35f62667c","object":"chat.completion","created":1766716638,"model":"dsr1","choices":[{"index":0,"message":{"role":"assistant","content":"<think>\nOkay, the user asked \"What is deeplearning?\" Hmm, seems like a straightforward definition request, but I should be careful not to assume their knowledge level. Maybe they're a complete beginner, or perhaps they have some ML background and just need clarity on deep learning specifically.\n\nFirst, I recall that deep learning is a subfield of machine learning. But I shouldn't start with jargon. Need to connect it to something familiar - like how humans learn from experience, since that's relatable. The \"artificial neural networks\" part is crucial though - that's what makes it \"deep\". \n\nWait, should I mention the \"deep\" refers to layers? Yeah, that's important context. Three types of layers: input, hidden, output. The hidden layers do the heavy lifting. And the \"deep\" just means many hidden layers. \n\nOh right, automatic feature learning! That's the superpower that distinguishes it from traditional ML. Traditional methods need hand-crafted features, but deep learning figures them out by itself. Huge advantage for complex data like images or speech. \n\nApplications... should I list examples? Yes, because it shows why it matters. Computer vision, NLP - those are the big ones. Maybe mention self-driving cars and voice assistants to make it concrete. \n\nThe user didn't specify their field, so I'll keep it broad. No need to dive into convolutional networks or transformers unless they ask. Just the core idea. \n\n...After writing the","refusal":null,"annotations":null,"audio":null,"function_call":null,"tool_calls":[],"reasoning":null,"reasoning_content":null},"logprobs":null,"finish_reason":"length","stop_reason":null,"token_ids":null}],"service_tier":null,"system_fingerprint":null,"usage":{"prompt_tokens":8,"total_tokens":308,"completion_tokens":300,"prompt_tokens_details":null},"prompt_logprobs":null,"prompt_token_ids":null,"kv_transfer_params":null

### DIFF
--- a/docs/source/developer_guide/feature_guide/eplb_swift_balancer.md
+++ b/docs/source/developer_guide/feature_guide/eplb_swift_balancer.md
@@ -160,7 +160,7 @@ class RandomLoadBalance(EplbPolicy):
 ## DFX
 ### Parameter Validation
 #### Integer Parameters
-All integer input parameters must explicitly specify their maximum and minimum values and be subject to valid value validation. For example, `num_iterations_eplb_update` must be greater than 0:
+All integer input parameters must explicitly specify their maximum and minimum values and be subject to valid value validation. For example, `expert_heat_collection_interval` must be greater than 0:
 
 ```python
     @staticmethod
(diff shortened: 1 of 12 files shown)


wangxiyuan:
the eplb config refactor should be done in one PR as we talked offline and a RFC is required

shenchuxiaofugui:
> the eplb config refactor should be done in one PR as we talked offline and a RFC is required

This PR is a bit messy after rebasing. I will close this PR and make the changes on PR5533.

PR #5377, opened 2025-12-25
[Bugfix] fix resource are insufficient when pcp and piecewise

weiguihua2:
Resolving the issue of insufficient resources during service operation when PCP is enabled in a piecewise scenario.

When enabling PCP and executing in piecewise mode, the curl request fails due to insufficient resources, resulting in the error message "The resources are insufficient." Through profiling analysis, it was found that the PCP communication domain also occupies streams and consumes resources. Therefore, when updating aclgraph sizes, the PCP communication domain needs to be taken into account.

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -482,6 +482,9 @@ def update_aclgraph_sizes(vllm_config: VllmConfig) -> None:
     # TODO: Find out whether we need to solve allreduce function
     MAX_CAPTURE_SIZE = 1800
 
+    # enable pcp or dcp will add new communication and consume additional approximately less than 100 streams
+    CP_ADDITIONAL_STREAM_NUM = 100
+
     # Store original configuration and temporarily clear it
     compilation_config = vllm_config.compilation_config
     original_sizes, compilation_config.cudagraph_capture_sizes = \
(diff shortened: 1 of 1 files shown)


yiz-liu:
@weiguihua2 Can we have some test cases covering this please?

vllm_ascend/utils.py
```diff
@@ -513,6 +513,7 @@ def update_aclgraph_sizes(vllm_config: VllmConfig) -> None:
     num_comm_groups = sum(size > 1 for size in [
         parallel_config.data_parallel_size,
         parallel_config.tensor_parallel_size,
+        parallel_config.prefill_context_parallel_size,
```
zzzzwwjj: When piecewise, attention is out of graph and will not cost one more stream for each subgraph. It is unreasonable.

weiguihua2: Determine how much additional flow will be occupied based on the collected profiling values.

vllm_ascend/utils.py
```diff
@@ -538,6 +538,12 @@ def update_aclgraph_sizes(vllm_config: VllmConfig) -> None:
             "Calculated maximum supported batch sizes for ACL graph: %s",
             max_num_batch_sizes)
     else:
+        # enable pcp or dcp will add new communication and consume additional approximately less than 100 streams
+        if parallel_config.prefill_context_parallel_size > 1:
+            MAX_CAPTURE_SIZE = MAX_CAPTURE_SIZE - 100
```
whx-sjtu: Don't use magic number 100. Instead, use a global constant to replace it.

weiguihua2: done

weiguihua2:
> @weiguihua2 Can we have some test cases covering this please?

This issue only occurs when the model has a large number of layers. Currently, CI uses lightweight models or runs test cases with reduced layers, so it's not possible to run the full-featured model.

zzzzwwjj (approved):

PR #5390, opened 2025-12-26
[Feature] Remove the transpose step after attention and switch to transpose_batchmatmul

jianzs:

1. The `npu_fused_infer_attention_score` kernel supports specifying the output layout. By selecting the appropriate layout, we can avoid the transpose operation typically required after the attention.
2. The `transpose_batchmatmul` function allows us to control whether the output tensor is transposed. If we configure `perm_y`, an additional transpose after executing `v_up` becomes unnecessary.

No.


wangxiyuan (approved):

vllm_ascend/attention/mla_v1.py
```diff
@@ -1169,23 +1159,30 @@ def _forward_decode(
...
+            # operation after attention.
+            input_layout = "TND_NTD"
             # TODO: If the driver is upgraded later, the contiguous function can be deleted.
             q_nope = q_nope.view(num_tokens, self.num_heads, -1).contiguous()
             q_pe = q_pe.view(num_tokens, self.num_heads, -1)
             sparse_mode = 3
             spec_attn_mask = attn_metadata.decode.attn_mask  # type:ignore
             actual_seq_lengths = decode_meta.actual_seq_lengths_q
         else:
+            # Input shape: [num_tokens, num_heads, seq_len, dim]
```
realliujiaxu: shape should be `[num_reqs, num_heads, seq_len, dim]`

PR #5397, opened 2025-12-26
[bugfix] Fix MHA model runtime error in aclgraph mode

kunpengW-code:
Currently, MHA models (eg: minicpm-2b, Baichuan-7b) will encounter errors when running in piecewise graph mode, with error messages similar to:
```
(E89999):  When layout is TND and PA not enabled, keyT(8) and valueT(8) must be equal to the last element of actualSeqenceLengthKV(5)[FUNC:CheckInputShapeWhenLayoutIsTND][FILE:prompt_flash_attention_tiling.cpp][LINE:3618]
```
The error occurs because the qkv in the Prefill stage is also padded, causing the shape to be inconsistent with actual_seq_lengths.
Add unpadding logic for kv.
no


weijinqian0 (approved):

vllm_ascend/attention/attention_v1.py
```diff
@@ -454,6 +454,9 @@ def _get_fia_params(self, key: torch.Tensor, value: torch.Tensor,
             block_size = 128
             block_table = None
             actual_seq_lengths_kv = attn_metadata.actual_seq_lengths_q
+            num_tokens_kv = actual_seq_lengths_kv[-1]
+            key = key[:num_tokens_kv]
+            value = value[:num_tokens_kv]
```
weijinqian0: move this to eager or piecewise branch.

kunpengW-code: fix

weijinqian0 (approved):

PR #5402, opened 2025-12-26
[Build] Add installation script of fused_infer_attention_score kernel with flash decoding 

MengqingCao:
Add installation script of `fused_infer_attention_score` kernel with flash decoding

### Userface changes
Users can install the kernel `fused_infer_attention_score` with flash decoding feature by `bash tools/install_flash_infer_attention_score_ops_a2.sh` or `bash tools/install_flash_infer_attention_score_ops_a3.sh`

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -115,8 +115,7 @@ jobs:
           pytest -sv --durations=0 tests/e2e/singlecard/test_sampler.py
           pytest -sv --durations=0 tests/e2e/singlecard/test_vlm.py
           pytest -sv --durations=0 tests/e2e/singlecard/test_xlite.py
-          # TODO: revert me when minicpm is fixed
-          # pytest -sv --durations=0 tests/e2e/singlecard/test_models.py
+          pytest -sv --durations=0 tests/e2e/singlecard/test_models.py
           pytest -sv --durations=0 tests/e2e/singlecard/pooling/
           pytest -sv --durations=0 tests/e2e/singlecard/compile/test_norm_quant_fusion.py
           pytest -sv --durations=0 tests/e2e/singlecard/test_multistream_overlap_shared_expert.py
--- a/.github/workflows/bot_pr_create.yaml
+++ b/.github/workflows/bot_pr_create.yaml
@@ -34,7 +34,7 @@ jobs:
     steps:
       - name: Get vLLM version
         run: |
-          VLLM_COMMIT=254f6b986720c92ddf97fbb1a6a6465da8e87e29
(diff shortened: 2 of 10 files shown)


MengqingCao:
@Yikun @wangxiyuan plz take a look, thx!

tools/install_flash_infer_attention_score_ops_a3.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# replace fused_infer_attention_score operation files
+cd $ASCEND_TOOLKIT_HOME/opp/built-in/op_impl/ai_core/tbe/kernel/ascend910_93
+rm -rf fused_infer_attention_score
+cp -r /vllm-workspace/fused_infer_attention_score_a3_$(uname -i)/fused_infer_attention_score .
+
+# replace related so
+cd $ASCEND_TOOLKIT_HOME/opp/built-in/op_impl/ai_core/tbe/op_tiling/lib/linux/$(uname -i)
+rm libopmaster_ct.so libopmaster_rt2.0.so liboptiling.so
+cp /vllm-workspace/fused_infer_attention_score_a3_$(uname -i)/*.so .
```
Yikun, line 34: cd /workspace?

tools/install_flash_infer_attention_score_ops_a3.sh
```diff
@@ -0,0 +1,34 @@
...
+tar -zxvf ./fused_infer_attention_score_a3_$(uname -i).tar.gz
+
+# replace fused_infer_attention_score operation files
+cd $ASCEND_TOOLKIT_HOME/opp/built-in/op_impl/ai_core/tbe/kernel/ascend910_93
+rm -rf fused_infer_attention_score
+cp -r /vllm-workspace/fused_infer_attention_score_a3_$(uname -i)/fused_infer_attention_score .
+
+# replace related so
+cd $ASCEND_TOOLKIT_HOME/opp/built-in/op_impl/ai_core/tbe/op_tiling/lib/linux/$(uname -i)
+rm libopmaster_ct.so libopmaster_rt2.0.so liboptiling.so
```
Yikun, line 33: pls make sure does this trigger y confirm

tools/install_flash_infer_attention_score_ops_a3.sh
MengqingCao: when exiting the bash script, it always fall back to the path where you're running it. in this case, it means `/vllm-workspace/vllm-ascend`. Thus I think it is fine because users choose to step into this path when they're installing this kernel.

tools/install_flash_infer_attention_score_ops_a3.sh
MengqingCao: After testing, it won't trigger y confirm in the image

PR #5419, opened 2025-12-27
[Doc] delete environment variable HCCL_OP_EXPANSION_MODE in DeepSeekV3.1/R1

GDzhu01:
Currently, HCCL_OP_EXPANSION_MODE="AIV" is causing some freezing issues on A2.so we have temporarily removed it from the documentation.


Yikun:
https://github.com/search?q=repo%3Avllm-project%2Fvllm-ascend+AIV+language%3AMarkdown+&type=code

What's the policy to address other AIV related doc?

PR #5417, opened 2025-12-27
[Bugfix] fix greedy temperature detection

realliujiaxu:
### What this PR does / why we need it?
fix greedy temperature detection from https://github.com/vllm-project/vllm/pull/27077

### Does this PR introduce _any_ user-facing change?
No
### How was this patch tested?

- vLLM version: release/v0.13.0
- vLLM main: https://github.com/vllm-project/vllm/commit/81786c87748b0177111dfdc07af5351d8389baa1

PR #5423, opened 2025-12-27
[BugFix] fix greedy temperature in eagle

bnbryan:

Updated the GREEDY_TEMPERATURE value in rejection_sampler.py to 0. 

Commit https://github.com/vllm-project/vllm/commit/a676e668ee10585099816b60a440ec089fd391c4 changed greedy temperature from -1 to 0 in gpu_input_batch.py. So, previously setting this constant to -1 in rejection_sampler.py will compromised the correctness of the is_greedy flag, which will cause a crash if a batch contains a mix of sequences with temperature > 0 and temperature = 0.

No.

vLLM version: release/v0.13.0
vLLM main: https://github.com/vllm-project/vllm/commit/be2a947521414272fa1ed17fd5345d45ddf10774


jianzs:
The issue has already been resolved in #5417

PR #5431, opened 2025-12-27
[P/D] bugfix for p node force free requset

liziyu179:
Fix the bug where the P-node's schedule dead after it force-frees a request due to timeout and then receives the completed kv cache pulled by the D-node again. By add list to recode all requests.

- vLLM main: https://github.com/vllm-project/vllm/commit/81786c87748b0177111dfdc07af5351d8389baa1

### DIFF
--- a/vllm_ascend/distributed/mooncake_connector.py
+++ b/vllm_ascend/distributed/mooncake_connector.py
@@ -118,6 +118,7 @@ class KVCacheTaskTracker:
     def add_not_transfer_request(self, request_id: str):
         with self.done_task_lock:
             self.finished_requests.add(request_id)
+            self.reqs_to_process.discard(request_id)
 
     def update_done_task_count(self, request_id: str):
         with self.done_task_lock:
@@ -982,7 +983,8 @@ class MooncakeConnectorScheduler:
             "num_external_tokens=%s, kv_transfer_params=%s",
             num_external_tokens, params)
 
-        if params.get("do_remote_decode"):
+        if params is not None and (params.get("do_remote_prefill", False)
+                                   or params.get("do_remote_decode", False)):
             self._reqs_in_batch.add(request.request_id)
         if params is not None and params.get("do_remote_prefill"):
             if params.get("remote_block_ids"):
@@ -1027,6 +1029,7 @@ class MooncakeConnectorScheduler:
         meta.requests_to_send = self._reqs_need_send
         self._reqs_need_send = {}
         meta.reqs_in_batch = self._reqs_in_batch
+        self._reqs_in_batch = set()
 
         return meta
 


vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -106,20 +106,21 @@ def __init__(self):
...
         self.delayed_free_requests: OrderedDict[str, float] = OrderedDict()
+        self.is_kv_producer = vllm_config.kv_transfer_config.is_kv_producer
 
     def add_not_transfer_request(self, request_id: str):
         with self.done_task_lock:
             self.finished_requests.add(request_id)
 
     def update_done_task_count(self, request_id: str):
         with self.done_task_lock:
-            self.finished_requests.add(request_id)
```
LCAIZJ, line 121: Will placing this line of code back in its original position cause any issues?

liziyu179:
> Could you provide more details about the issue you're trying to resolve?

When a request is forced_free by a P node, if it happens to be completed by being pulled by a D node, this request will enter the get_finished interface of MooncakeConnectorWorker twice, and will also enter the _update_from_kv_xfer_finished function of the Scheduler twice. The second time will cause an assertion failure because req_id is not in self.requests.

vllm_ascend/distributed/mooncake_connector.py
liziyu179: > Will placing this line of code back in its original position cause any issues?

For a P node, if the request has been forced_free, then it will not be in delayed_free_requests. This indicates it was previously marked as finished and does not need to be marked again.

vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -137,10 +138,7 @@ def get_and_clear_finished_requests(self) -> set[str]:
     def add_delayed_request(self, request_id: str, delay_start_time: float):
         """Add a delayed free request."""
         with self.done_task_lock:
-            if request_id not in self.record_finished_requests:
```
baxingpiaochong, line 143: `add_delayed_request` occurs in the next forward. Therefore, it's possible that `DONE_RECVING_MSG` is received before `add_delayed_request`. In this case, this modification would force normally released requests to be forcibly released after timeout.

LCAIZJ:
> > Could you provide more details about the issue you're trying to resolve?
> 
> When a request is forced_free by a P node, if it happens to be completed by being pulled by a D node, this request will enter the get_finished interface of MooncakeConnectorWorker twice, and will also enter the _update_from_kv_xfer_finished function of the Scheduler twice. The second time will cause an assertion failure because req_id is not in self.requests.

If the timeout for forced release is longer than the timeout for aborting a request, this issue should not occur. That said, resolving it properly in the code logic would also be a good approach.

vllm_ascend/distributed/mooncake_connector.py
LCAIZJ: cc @liziyu179

liziyu179: The add_delayed_request operation occurs within the execute_model function following the current scheduling round. Requests are forwarded to the D node only after the P node has completed its execution of execute_model. Consequently, any request received by the D node is necessarily present in the P node’s delayed_requests collection. Therefore, the situation you described is deemed not to exist.

LCAIZJ: @liziyu179 We've previously considered the scenario described in this PR https://github.com/vllm-project/vllm-ascend/pull/2899, so please check whether the current changes are compatible with it.

baxingpiaochong (approved):

vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -111,18 +111,22 @@ def __init__(self):
...
-                self._remove_delayed_requests(request_id)
+            if request_id not in self.forced_free_requests:
+                self.finished_requests.add(request_id)
+                if request_id in self.delayed_free_requests:
+                    self._remove_delayed_requests(request_id)
+                else:
+                    self.record_finished_requests.add(request_id)
             else:
-                self.record_finished_requests.add(request_id)
+                self.forced_free_requests.discard(request_id)
```
LCAIZJ: Here, you only add forced free requests but never remove them. Line 129 only removes requests that became abnormal after force free, which will lead to memory leaks over time.

vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -109,20 +109,22 @@ def __init__(self):
...
         with self.done_task_lock:
-            self.finished_requests.add(request_id)
-            if request_id in self.delayed_free_requests:
-                self._remove_delayed_requests(request_id)
-            else:
-                self.record_finished_requests.add(request_id)
+            if request_id in self.reqs_to_process:
+                self.finished_requests.add(request_id)
+                self.reqs_to_process.discard(request_id)
+                self.delayed_free_requests.pop(request_id, None)
```
baxingpiaochong, line 128: You can add an else branch to report the log error. An exception occurs when the `req_id` received by `update_done_task_count` is not in the process, indicating a precision issue.

vllm_ascend/distributed/mooncake_connector.py
```diff
@@ -1601,6 +1597,10 @@ def start_load_kv(self, metadata: MooncakeConnectorMetadata):
                         all_task_done=(i == self.tp_num_need_pulls *
                                        self._prefill_pp_size - 1))
 
+        if self.kv_send_thread is not None:
+            for req_id, meta in metadata.requests.items():
```
baxingpiaochong: The prefill node might lack meta information? If you need the req_id, you need to bring it down from the scheduler side. I didn't see this operation in this PR.

liziyu179: > The prefill node might lack meta information? If you need the req_id, you need to bring it down from the scheduler side. I didn't see this operation in this PR.

You're right, we added a req_in_batch variable to pass the request from the scheduler to the worker.

vllm_ascend/distributed/mooncake_connector.py
liziyu179: > You can add an else branch to report the log error. An exception occurs when the `req_id` received by `update_done_task_count` is not in the process, indicating a precision issue.

Yes, we need to remind users about this here.

wangxiyuan:
@LCAIZJ Please merge if this change is fine

wangxiyuan:
@jianzs cc

LCAIZJ:
> @LCAIZJ Please merge if this change is fine
LGTM! Merging now.

LCAIZJ (approved):

wangxiyuan:
force mege. The CI failure doesn't related to this PR.

PR #5444, opened 2025-12-27
[Bugfix] Fix chunk prefill bug for long_sequence feature

LookAround0301:
Fix chunk prefill bug for long_sequence feature

When there are two requests with chunk prefill enabled in the long-sequence scenario, if one request has only 1 token during scheduling, it will be identified as a decode request and trigger an error. This PR fixes the issue.

Yes, fix the bug

Closes: https://github.com/vllm-project/vllm-ascend/issues/5445

- vLLM main: https://github.com/vllm-project/vllm/commit/81786c87748b0177111dfdc07af5351d8389baa1

### DIFF
--- a/tests/e2e/multicard/long_sequence/test_chunked_prefill.py
+++ b/tests/e2e/multicard/long_sequence/test_chunked_prefill.py
@@ -22,9 +22,9 @@ Run `pytest tests/e2e/multicard/test_qwen3_moe.py`.
 """
 
 import os
-from unittest.mock import patch
 import random
 import string
+from unittest.mock import patch
 
 from vllm import SamplingParams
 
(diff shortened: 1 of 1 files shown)


Yikun:
@LookAround0301 Would mind adding a e2e test to prevent the regression happened again?

LookAround0301:
> @LookAround0301 Would mind adding a e2e test to prevent the regression happened again?

Okay, I have added e2e test

tests/e2e/multicard/long_sequence/test_chunked_prefill.py
```diff
@@ -0,0 +1,79 @@
...
+def generate_prompts(input_len, batchsize):
+    prompts = [
+        " ".join([
+            f"{random.choice(string.ascii_letters)}" for _ in range(input_len)
+        ]) for _ in range(batchsize)
+    ]
+    return prompts
+
+
+@pytest.mark.skipif(vllm_version_is('0.12.0'),
```
MengqingCao: we can just remove this decorator, as 0.12.0 is not maintained anymore

tests/e2e/multicard/long_sequence/test_chunked_prefill.py
```diff
@@ -0,0 +1,79 @@
...
+
+import pytest
+from vllm import SamplingParams
+
+from tests.e2e.conftest import VllmRunner
+from vllm_ascend.utils import vllm_version_is
+
+os.environ["HCCL_BUFFSIZE"] = "768"
+os.environ["VLLM_ASCEND_ENABLE_FLASHCOMM1"] = "1"
+os.environ["VLLM_ALLOW_LONG_MAX_MODEL_LEN"] = "1"
```
MengqingCao: let's use patch instead of this, to ensure the unexpected env set in other test cases. plz refer to https://github.com/vllm-project/vllm-ascend/blob/350b95efcf6ebedaae7b87ba621c38c021b471a8/tests/ut/ops/test_linear.py#L65

PR #2753, opened 2025-09-04
[main] add pd transfer for ascend scheduler

Liccol:
### What this PR does / why we need it?
For offline scenarios, adjust the scheduling process to prioritize the prefill phase of all requests, then process the decode phase of all requests.
### How was this patch tested?

```
max_num_seqs=24,
additional_config={
    "ascend_scheduler_config":{
        "enabled": True,
        "enable_pd_transfer": True,
        "decode_max_num_seqs": 24,
        "enable_chunked_prefill": False
    }
},
```
| input | output | num prompts | max_num_seqs | dp | tp | scheduler      | tps           |
| ------ | ------ | ---------- | ---------------- | ---- | ---- | ---------------- | --------------- |
| dapo-math-17K | 2K   |      384 |             24 |  2 |  1 | v1             |        234.06 |
| dapo-math-17K | 2K   |      384 |             24 |  2 |  1 | pd transfer        | 239.59(+2.4%) |
| dapo-math-17K| 2K   |      384 |             24 |  4 |  1 | v1             |        222.85 |
| dapo-math-17K| 2K   |      384 |             24 |  4 |  1 | pd transfer         |         225.81(+1.3%) |

- vLLM version: v0.10.1.1
- vLLM main: https://github.com/vllm-project/vllm/commit/6fb27881634d89c2e70e9e5fbad1b918c0d916cf

PR #5469, opened 2025-12-29
[Refactor] Import global var form vllm instead of overwirte it

MengqingCao:
Import global var form vllm instead of overwirte it, so that we could use the correct global variant value

N/A

CI passed with existing test.

- vLLM version: v0.13.0


vllm_ascend/sample/logits_processor/__init__.py
```diff
@@ -18,16 +18,7 @@
...
 
-# Error message when the user tries to initialize vLLM with a pooling model
-# and custom logitsproces
-STR_POOLING_REJECTS_LOGITSPROCS = ("Pooling models do not support custom"
-                                   " logits processors.")
-
-BUILTIN_LOGITS_PROCESSORS: list[type[LogitsProcessor]] = [
-    MinTokensLogitsProcessor,
-    LogitBiasLogitsProcessor,
-    AscendMinPLogitsProcessor,
```
MengqingCao, line 29: check if this is indeed

weijinqian0 (approved):

vllm_ascend/sample/logits_processor/builtin.py
```diff
@@ -1,52 +0,0 @@
-import torch
```
MengqingCao, line 1: There are 2 reasons for overwriting `MinPLogitsProcessor`:
1. support pd-transfer in ascend scheduler scenario, introduced by #2753. As ascend scheduler is drop now, we don't need to overwirte the func `__init__` anymore
2. avoid indexing operation, which is already done in upstream vllm in https://github.com/vllm-project/vllm/pull/29826

Thus we could drop this safely

PR #5472, opened 2025-12-29
[BugFix] Support ALL D-Nodes in fullgraph when running MTP in PD

dragondream-chen:
**BUG Problem**
When using prefill-decode disaggregation + MTP + full graph +asynchronous scheduling, the KV cache pulled by decode nodes from prefill decodes does not include spec tokens. As a result, the total_num_scheduled_tokens obtained by decode nodes from the scheduler lacks spec tokens. When determining whether to enqueue the full graph on decode nodes, the condition for uniform_decode ` scheduler_output.total_num_scheduled_tokens == self.input_batch.num_reqs * max_query_len` is not met, leading to the current instance not being enqueued into the full graph. 

The above situation leads to both full graph and eagle mode instances coexisting in the decode instances. Due to the synchronization wait of MoeDispatch, the decode instances in full graph are significantly slowed down by the instance in eagle mode.

**Solution**
The scenario is PD separation + MTP + Full Graph + asynchronous scheduling.
On the decode nodes, the spec tokens of the request with KV cache from P need be padded. Then, the padded spec tokens will be rejected by sampling. This operation ensures that the uniform_decode condition is satisfied when determining whether decode nodes are included in the full graph, thereby guaranteeing that all decode instances are present in the full graph and avoiding synchronous waiting for MoeDispatch.

- vLLM version: v0.15.0

### DIFF
--- a/vllm_ascend/core/recompute_scheduler.py
+++ b/vllm_ascend/core/recompute_scheduler.py
@@ -40,6 +40,7 @@ from vllm.v1.engine import EngineCoreEventType, EngineCoreOutput, EngineCoreOutp
 from vllm.v1.metrics.perf import PerfStats
 from vllm.v1.outputs import ModelRunnerOutput
 from vllm.v1.request import Request, RequestStatus
+from vllm.v1.sample.rejection_sampler import PLACEHOLDER_TOKEN_ID
 from vllm.v1.spec_decode.metrics import SpecDecodingStats
 from vllm.v1.utils import ConstantList, record_function_or_nullcontext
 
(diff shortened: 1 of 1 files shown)


realliujiaxu:
@jianzs please take a look, this bugfix depends on recompute scheduler

realliujiaxu (changes requested):
please discribe why need patch and how to eliminate it in `vllm_ascend/patch/__init__.py`

dragondream-chen:
> please discribe why need patch and how to eliminate it in `vllm_ascend/patch/__init__.py`
In this PR, we need patch `scheduler.py` and the `_update_state` in `gpu_model_runner.py` of vllm, and the reason is in the instruction of this PR.
In `platform.py`, we indicate that the patch can only be enabled if recomputation (PD separation + MTP) is enabled.

vllm_ascend/patch/worker/patch_model_runner.py
```diff
@@ -0,0 +1,292 @@
...
+    # Allow attention backend to reorder the batch, potentially
+    self._may_reorder_batch(scheduler_output)
+    # Refresh batch metadata with any pending updates.
+    self.input_batch.refresh_metadata()
+
+
+if vllm_version_is('0.13.0'):
+    GPUModelRunner._update_states = _update_states
+else:
+    raise ValueError(
```
whx-sjtu: For main branch, it is not properly to just raise error here. Maybe just log a warnning message here?

vllm_ascend/patch/worker/patch_model_runner.py
```diff
@@ -0,0 +1,292 @@
...
+from vllm.model_executor.models.interfaces_base import VllmModelForPooling
+from vllm.sampling_params import SamplingType
+from vllm.v1.core.sched.output import SchedulerOutput
+from vllm.v1.worker.gpu_input_batch import CachedRequestState
+from vllm.v1.worker.gpu_model_runner import GPUModelRunner
+
+from vllm_ascend.utils import vllm_version_is
+
+
+def _update_states(self, scheduler_output: "SchedulerOutput") -> None:
```
JC-ut0: Maybe use some special comments like `====` to state the difference?

vllm_ascend/patch/worker/patch_model_runner.py
dragondream-chen: If the patch takes effect and the corresponding code in the vLLM community is updated, causing a bug, users may not be aware of it, making it difficult to identify.

vllm_ascend/patch/worker/patch_model_runner.py
dragondream-chen: Sounds good. I intended to do it this way, but I couldn't find a reference sample. I'll update it right away.

whx-sjtu (approved):

wangxiyuan:
for main branch, we should try to fix it in vLLM first.

PR #5481, opened 2025-12-29
[Refactor] Formatting output types related to FuseMoE

jianzs:
### What this PR does / why we need it?

Currently in the Fused MoE module, functions of classes like MoECommMethod and MoETokenDispatcher output data in dictionary or tuple format, which hampers code maintainability, readability, and extensibility. This PR introduces dataclasses for these key output types to address these issues.

### Does this PR introduce _any_ user-facing change?

No

### How was this patch tested?

- vLLM version: v0.13.0
- vLLM main: https://github.com/vllm-project/vllm/commit/5326c89803566a131c928f7fdd2100b75c981a42

PR #5482, opened 2025-12-29
[Feature] Support fine-grained shared expert overlap

jianzs:

Fine-grained control over shared expert overlap to prevent resource contention.

Depends on https://github.com/vllm-project/vllm-ascend/pull/5481

No

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/ops/fused_moe/fused_moe.py
+++ b/vllm_ascend/ops/fused_moe/fused_moe.py
@@ -50,8 +50,7 @@ from vllm_ascend.quantization.w8a8_dynamic import \
 from vllm_ascend.utils import (AscendDeviceType, enable_sp,
                                get_ascend_device_type, maybe_trans_nz,
                                npu_stream_switch, shared_expert_dp_enabled,
-                               shared_experts_calculation_stream)
-
+                               shared_experts_calculation_stream, vllm_version_is)
 
 @dataclass
 class FusedMoEResult:
@@ -455,9 +454,13 @@ class AscendSharedFusedMoE(SharedFusedMoE, AscendFusedMoE):
         # Qwen3-Next specific gating mechanism
         if hasattr(self._shared_experts, "expert_gate") and \
             self._shared_experts.expert_gate is not None:
-            shared_out = F.sigmoid(
-                self._shared_experts.expert_gate(hidden_states)  # type: ignore
-            ) * shared_out
+            if vllm_version_is('0.13.0'):
+                # TODO(jianzs): remove this branch after vLLM new version is
+                # released
+                gate_out = self._shared_experts.expert_gate(hidden_states)  # type: ignore
+            else:
+                gate_out, _ = self._shared_experts.expert_gate(hidden_states)  # type: ignore
+            shared_out = F.sigmoid(gate_out) * shared_out
         return shared_out
 
     def _validate_shared_expert_consistency(self):


vllm_ascend/ops/fused_moe/token_dispatcher.py
```diff
@@ -223,6 +248,17 @@ def token_dispatch(self,
         expand_x, dynamic_scale, assist_info_for_combine, expert_token_nums, \
             ep_recv_counts, tp_recv_counts, expand_scales = output[0:7]
 
+        # Handle shared experts (store intermediate results in local vars, not self)
+        shared_act = None
+        if shared_experts is not None:
```
realliujiaxu: how about we move wait_event and calculation into AscendSharedFusedMoE, let dispatcher focus on dispatching tokens.

jianzs: Great idea. Please you take another look.

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -422,44 +433,70 @@ def forward(
...
+                               fused_moe_evts: FusedMoEEvents):
+
+        def maybe_wait_event(evt: torch.npu.Event | None):
+            if evt is not None:
+                torch.npu.current_stream().wait_event(evt)
+
+        with npu_stream_switch(shared_experts_calculation_stream(),
+                               enabled=self.multistream_overlap_shared_expert):
+            # Ensure the shared experts wait for hidden_states to be ready.
+            torch.npu.current_stream().wait_event(
```
lidenghui1110, line 533: maybe we can use `maybe_wait_event` function to make it unified？

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -422,44 +433,70 @@ def forward(
...
+                fused_moe_evts.before_routed_experts)
+            # Execute the gate projection and activation concurrently with the
+            # dispatch communication.
+            maybe_wait_event(fused_moe_evts.before_dispatch)
```
lidenghui1110, line 537: is waiting for before_routed_experts and before_dispatch events duplicated here？maybe we just need wait for before_dispatch？

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -422,44 +433,70 @@ def forward(
...
+            # Execute the down projection concurrently with the combine
+            # communication.
+            maybe_wait_event(fused_moe_evts.before_combine)
+            shared_out, _ = self._shared_experts.down_proj(
+                shared_act)  # type: ignore
+
+        # Make sure the default stream waits for the shared experts stream to
+        # finish.
+        if self.multistream_overlap_shared_expert:
+            torch.npu.current_stream().wait_stream(
```
lidenghui1110, line 547: use `maybe_wait_event` function too.

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -422,44 +433,70 @@ def forward(
...
-                shared_out = tensor_model_parallel_all_reduce(shared_out)
-        else:
+        if self.multistream_overlap_gate:
             fc3_context = get_flash_common3_context()
             assert fc3_context is not None
             shared_out = fc3_context.shared_out
-
-        return shared_out, fused_output
+        else:
+            shared_out = self._forward_shared_experts(
```
lidenghui1110, line 578: I'm not sure this fine-grained overlap will work for all scenario, as my experiments for shared-expert overlap, the computation of shared-expert will contend vector core with AIV kernel in communication which may cause the matmul ops performance reduced about ten times.

And in pure TP scenario, will this overlap still in use? If there are no communication in dispatch, it may make performance worse, am I right?

vllm_ascend/ops/fused_moe/fused_moe.py
jianzs: The `shared_experts` can only run after this event is ready, so `maybe_wait_event` is not used.

vllm_ascend/ops/fused_moe/fused_moe.py
jianzs: The `before_dispatch` value might be None. There is routing logic between `before_dispatch` and `before_routed_experts`.

vllm_ascend/ops/fused_moe/fused_moe.py
jianzs: This is `wait_stream`, not `wait_event`.

vllm_ascend/ops/fused_moe/fused_moe.py
jianzs: Users can turn off this feature by setting `multistream_overlap_shared_expert` to false.

jianzs:
> Looking good, can we have a RFC about this?

Please take a look #5708

PR #5495, opened 2025-12-30
[Main2Main] Upgrade vllm commit to 1230

wjunLu:

Upgrade vllm commit to 1230

Affected by https://github.com/vllm-project/vllm/pull/27614 (and the core PR https://github.com/vllm-project/vllm/pull/26866), we have to make the following changes:

1. Modify `tests/e2e/multicard/test_aclgraph_capture_replay.py` to keep compatible with both vllm version of `v0.13.0` and latest main commitID, while vllm enables async scheduling by default
2. Skip `test_guided_decoding.py` due to xgrammar errors (https://github.com/vllm-project/vllm-ascend/issues/5524)

- vLLM version: v0.13.0


wangxiyuan:
CI passed here: https://github.com/vllm-project/vllm-ascend/actions/runs/20599218105?pr=5495

PR #5519, opened 2025-12-30
[BugFix] Support setting tp=1 for the Eagle draft model to take effect

zhaomingyu13:
According to the official documentation, the parameter "draft_tensor_parallel_size": 1 is supposed to be applied to the Eagle3 model. However, based on actual debugging, it was found that the number of tensor parallelisms (tp) of the Eagle model is consistent with that of the target model. The setting of tp for the draft model did not take effect as expected.

**Note:** This feature has not been superimposed and tested with `sp` and `dp`. It will be adapted later
No
```python

def main():
    prompts = [
        "The future of AI is",
    ]

    sampling_params = SamplingParams(temperature=0.8, top_p=0.95)
    # Create an LLM.
    llm = LLM(
            model="meta-llama/Llama-3.1-8B-Instruct",
            tensor_parallel_size=4,
            gpu_memory_utilization=0.9,
            enforce_eager=True,
            speculative_config={
                "method": "eagle3",
                "model": "yuhuili/EAGLE3-LLaMA3.1-Instruct-8B"
                "draft_tensor_parallel_size": 1,
                "num_speculative_tokens": 3,
            },
        )

    print(f"Outputs: {outputs}")
    for output in outputs:
        prompt = output.prompt
```
- vLLM version: v0.13.0

Fixes vllm-project/vllm#31345

### DIFF
--- a/vllm_ascend/spec_decode/eagle_proposer.py
+++ b/vllm_ascend/spec_decode/eagle_proposer.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
+from contextlib import nullcontext
 from typing import Optional
 
 import numpy as np
@@ -7,7 +8,9 @@ import torch.nn as nn
 import torch.nn.functional as F
 from vllm.config import (CompilationMode, CUDAGraphMode, VllmConfig,
                          get_layers_from_vllm_config)
-from vllm.distributed.parallel_state import get_pp_group
+from vllm.distributed.parallel_state import (get_pp_group, get_world_group,
+                                             init_model_parallel_group,
+                                             patch_tensor_parallel_group)
 from vllm.forward_context import get_forward_context
 from vllm.logger import logger
 from vllm.model_executor.layers.attention_layer_base import AttentionLayerBase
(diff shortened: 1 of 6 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2232,7 +2234,14 @@ def load_model(self) -> None:
                 model_register(self.model, self.model_config)
             if self.drafter:
                 logger.info("Loading drafter model...")
-                self.drafter.load_model(self.model)
+                if self.drafter.speculative_config.draft_tensor_parallel_size == 1:
```
pjgao: why only tp=1 need this patch? Maybe we can add more comments here to hint others?

zhaomingyu13: Comments have been added

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2232,7 +2234,26 @@ def load_model(self) -> None:
                 model_register(self.model, self.model_config)
             if self.drafter:
                 logger.info("Loading drafter model...")
-                self.drafter.load_model(self.model)
+                # NOTE:
+                # `draft_tensor_parallel_size` does not take effect for Eagle3:
+                # the draft model uses the same TP size as the target model in practice.
+                # so we applied this patch.
+                if self.drafter.speculative_config.draft_tensor_parallel_size == 1:
```
drslark: The initialization of `draft_tp_group` may be better to move to `eagle3.__init__()`.
Beacause `draft_tp_group` is a attribute of `eagle3`.  `draft_tp_group` be `self.draft_tp_group` is more semantically accurate.

drslark: LGTM

zhaomingyu13: have modified

vllm_ascend/spec_decode/eagle_proposer.py
```diff
@@ -85,6 +88,21 @@ def __init__(self,
 
         self.use_sparse = hasattr(vllm_config.model_config.hf_config,
                                   "index_topk")
+        # NOTE:
+        # `draft_tensor_parallel_size` does not take effect for Eagle3:
+        # the draft model uses the same TP size as the target model in practice.
+        # so we applied this patch.
+        if self.speculative_config.draft_tensor_parallel_size == 1:
+            self.tp_group = init_model_parallel_group(
```
wangxiyuan: `self.tp_group` to `tp_group`

wangxiyuan (approved):

zzzzwwjj (approved):

vllm_ascend/spec_decode/eagle_proposer.py
zhaomingyu13: have modified

PR #5522, opened 2025-12-30
[Bugfix] fix the precision issues that may raise from the inter-layer reuse of the workspace in certain scenarios

WithHades:

In the current process of implementing attention updates, the FIA operator shares a single workspace among different layers within the same computation graph. To enable memory reuse, we adopt the weak_ref_tensor mechanism. However, this approach may lead to precision anomalies in certain scenarios. To address this issue, different layers in the same computation graph are assigned independent workspaces.

- vLLM version: v0.13.0


yiz-liu:
> Please fix other attn impl as well like `full_graph_fia` or `mla`. @WithHades

Turns out that PA `weak_ref_tensors` twice before, other impl don't need to be fixed.

yiz-liu (approved):

wangxiyuan (approved):

PR #5546, opened 2025-12-31
[Main2Main] Upgrade vllm commit to 0102

wjunLu:
Upgrade vllm commit to 0102

1. Remove `maybe_padded_num_tokens` arg in `model_runner_v1.py` due to https://github.com/vllm-project/vllm/pull/31517
2. Remove `Qwen/Qwen3-0.6B` in `tests/e2e/multicard/test_aclgraph_capture_replay.py` because that [ Offline data parallel mode will be not supported/useful for dense models](https://github.com/vllm-project/vllm/commit/bd877162ebd7a2eb36961a8e49824422dd35bcdc#diff-c1550d0a38469d039370567d8981969530cbfffc7302cd1778e7c2c8a9322deaL548)

- vLLM version: v0.13.0

### DIFF
--- a/.github/workflows/nightly_test_a3.yaml
+++ b/.github/workflows/nightly_test_a3.yaml
@@ -68,6 +68,12 @@ jobs:
           - name: multi-node-qwenw8a8-2node-eplb
             config_file_path: Qwen3-235B-W8A8-EPLB.yaml
             size: 2
+          - name: multi-node-deepseek-r1-w8a8-longseq
+            config_file_path: DeepSeek-R1-W8A8-longseq.yaml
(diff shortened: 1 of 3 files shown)


jianzs:
Can we use GitHub Actions to automatically sync the latest commits in vllm?

wjunLu:
> Can we use GitHub Actions to automatically sync the latest commits in vllm?

We have this workflow https://github.com/vllm-project/vllm-ascend/actions/workflows/schedule_test_vllm_main.yaml to automatically sync and verfiy, but there are always some error so that we have to process them manually

PR #5553, opened 2025-12-31
[Bugfix] Fix the graph capture failure issue in the eagle3+full scenario.

WithHades:
【Bugfix】When launching the service in the scenario where the cudagraph_mode is set to FULL and Eagle3 acceleration is enabled for inference, an error in fia will cause graph capture to fail. This PR fixes the issue.

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/spec_decode/eagle_proposer.py
+++ b/vllm_ascend/spec_decode/eagle_proposer.py
@@ -65,7 +65,8 @@ class EagleProposer(VllmEagleProposer):
 
         self.pcp_size = self.runner.pcp_size
         self.decode_threshold = 1 + self.num_speculative_tokens
-
+        self.query_start_loc = self.runner._make_buffer(
+            self.runner.max_num_reqs + 1, dtype=torch.int32)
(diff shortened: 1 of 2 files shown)


yiz-liu (commented):
I'm afraid we cannot merge this, maybe we can try refactor this to proposer instead of just deleting it.

vllm_ascend/spec_decode/eagle_proposer.py
```diff
@@ -65,7 +65,8 @@ def __init__(self,
 
         self.pcp_size = self.runner.pcp_size
         self.decode_threshold = 1 + self.num_speculative_tokens
-
+        self.query_start_loc = self.runner._make_buffer(
+            self.runner.max_num_reqs + 1, dtype=torch.int32)
```
yiz-liu, line 69: A bit weird, but OK for a temporary fix. @weijinqian0 Please make sure you don't miss this PR when `actual_seq_lengths_q` is removed.

yiz-liu (approved):

PR #5574, opened 2026-01-04
[P/D]Remove mooncake kvpool unused parameter `local_hostname`

LCAIZJ:
In mooncake kvpool, `local_hostname` is not used. Instead, the local IP is obtained directly via `get_ip()`. Therefore, remove this parameter to avoid confusion.

- vLLM version: v0.13.0


LCAIZJ:
cc @fems14 @Pz1116

wangxiyuan (approved):

PR #5582, opened 2026-01-04
[bugfix]update bishengir source envs

Meihan-chen:
Due to the update of the Bisheng version's installation path, the corresponding source path in the environment variables needs to be updated.

- vLLM version: v0.13.0

### DIFF
--- a/docs/source/tutorials/Qwen3-Next.md
+++ b/docs/source/tutorials/Qwen3-Next.md
@@ -72,12 +72,6 @@ pip install triton_ascend-3.2.0.dev20251229-cp311-cp311-manylinux_2_27_$(uname -
 
 ### Inference
 
-Please make sure you have already executed the command:
-
-```bash
-source /usr/local/Ascend/ascend-toolkit/8.3.RC2/bisheng_toolkit/set_env.sh
-```
-
 :::::{tab-set}
 ::::{tab-item} Online Inference
 


.github/workflows/pr_test_light.yaml
```diff
@@ -133,7 +133,9 @@ jobs:
           BISHENG_NAME="Ascend-BiSheng-toolkit_aarch64_20251225.run"
           BISHENG_URL="https://vllm-ascend.obs.cn-north-4.myhuaweicloud.com/vllm-ascend/${BISHENG_NAME}"
           wget -O "${BISHENG_NAME}" "${BISHENG_URL}" && chmod a+x "${BISHENG_NAME}" && "./${BISHENG_NAME}" --install && rm "${BISHENG_NAME}"
-          source /usr/local/Ascend/8.5.0/bisheng_toolkit/set_env.sh
+          which bishengir-compile
+          export PATH=/usr/local/Ascend/tools/bishengir/bin:$PATH
+          which bishengir-compile
```
MengqingCao: plz remove `which bishengir-compile` after the right bisheng compiler is confimed

Meihan-chen: OK, this command is used to test whether Bisheng has taken effect and will be removed after the test is completed.

PR #5319, opened 2025-12-24
[MM][Bugfix] Update `hf_config` to `hf_text_config`

shen-shanshan:
### What this PR does / why we need it?

Following https://github.com/vllm-project/vllm-ascend/pull/5205, update `hf_config` to `hf_text_config`.

Find more details at https://github.com/vllm-project/vllm-ascend/pull/5205#issuecomment-3675417534 and https://github.com/vllm-project/vllm-ascend/pull/5205#issuecomment-3677920872.

### How was this patch tested?

- vLLM version: release/v0.13.0
- vLLM main: https://github.com/vllm-project/vllm/commit/5fbfa8d9ef15948599631baeb91e8220b2ee9bcc

PR #5600, opened 2026-01-05
[Bugfix] Make pd more generalization

Potabk:
Currently, in pd disaggregated scenario, when the prefiller's `tp_size` > decoder's `tp_size`, we need to get `num_key_value_heads` from model_config, but for different models, the `config.json` are different(eg: for multi-modal models, there exist a `text_config` section), we cannot directly obtain this field through hf_config. Instead, we need to use special preprocessing in transformers to perform different processing on different models to obtain a unified configuration.

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -94,16 +94,18 @@ class AscendConfig:
             self.pd_tp_ratio = prefill_tp_size // decode_tp_size
             if self.pd_tp_ratio > 1:
                 try:
-                    # only support Qwen model now
-                    # TODO: use a more robust method to get kv_head_num
-                    num_kv_head = vllm_config.model_config.hf_config.num_key_value_heads
+                    num_kv_head = vllm_config.model_config.hf_text_config.num_key_value_heads
                     self.num_head_replica = prefill_tp_size // num_kv_head if prefill_tp_size >= num_kv_head else 1
                     prefill_tp_size = min(prefill_tp_size, num_kv_head)
                     decode_tp_size = min(decode_tp_size, num_kv_head)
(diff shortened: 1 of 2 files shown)


vllm_ascend/ascend_config.py
```diff
@@ -94,16 +94,18 @@ def __init__(self, vllm_config: "VllmConfig"):
...
                     self.pd_head_ratio = prefill_tp_size // decode_tp_size
                 except Exception:
-                    raise AssertionError(
-                        "Can not get num_key_value_heads from model_config")
+                    raise ValueError(
+                        "The text_config extracted from the model config does not have "
+                        "`num_key_value_heads` attribute. This indicates a mismatch "
+                        "between the model config and vLLM's expectations. Please "
+                        "ensure that the model config is compatible with vLLM."
+                    )
```
wangxiyuan, line 108: vLLM Ascend

And please descript why it's mismatch

Potabk: In fact, regardless of whether the model's `config.json` is nested or tiled, we want to retrieve the parameters from `text_config`. Therefore, we should need a function to handle these cases. (See reference https://github.com/huggingface/transformers/blob/3c2517727ce28a30f5044e01663ee204deb1cdbe/src/transformers/configuration_utils.py#L1129)

Potabk: https://github.com/vllm-project/vllm-ascend/pull/5205 also cc @ApsarasX

Potabk: Sorry, I didn't notice before. We may use https://github.com/vllm-project/vllm-ascend/pull/5319; it looks more comprehensive. also cc @shen-shanshan

PR #5613, opened 2026-01-05
[BugFix] Fix Smoke Testing Bug for DSR1 longseq

dsxsteven:
Fix Smoke Testing Bug for DSR1 longseq
We need to make this change because the daily smoke test case is throwing an error: "max_tokens or max_completion_tokens is too large: 32768.This model's maximum context length is 32768 tokens and your request has 128 input tokens". We encounter this error due to max-out-len equals to max-model-len. We can fix this error by increasing max-model-len argument in the script.

- vLLM version: v0.13.0


wangxiyuan:
please add more info to explain why

dsxsteven:
> please add more info to explain why

done

PR #5620, opened 2026-01-06
Improve the performance of Mooncake as a backend.

heijian123:
Call the `batch_put_from_multi_buffers` interface and set `prefer_alloc_in_same_node` to true to prioritize storing kvcache on the same node.

Performance comparison：
| Scenario| Sequence Length | TTFT （s）|
| --- | --- | --- |
| `prefer_alloc_in_same_node` is False | 35450 | 1.199 |
| `prefer_alloc_in_same_node` is True| 35450 | 0.634 |

- vLLM version: v0.13.0
- vLLM main: https://github.com/vllm-project/vllm/commit/8be6432bdaf6275664d857b1e5e9bf8ed1ce299e

### DIFF
--- a/vllm_ascend/distributed/kvpool/backend/mooncake_backend.py
+++ b/vllm_ascend/distributed/kvpool/backend/mooncake_backend.py
@@ -21,13 +21,15 @@ class MooncakeBackend(Backend):
 
     def __init__(self, parallel_config: ParallelConfig):
         try:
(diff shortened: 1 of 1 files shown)


LCAIZJ:
`preferred_segment` and `prefer_alloc_in_same_node` can be added, but they may not be suitable for all scenarios. Therefore, these options can be exposed as configurable parameters in MooncakeStoreConfig, allowing users to enable them selectively.

PR #5624, opened 2026-01-06
adapt to minimax_m2

Feng-xiaosuo:
This PR fixes Minimax model loading in vLLM Ascend backend by:

Adding model type check for "minimax" and "minimax_m2" to replace "mlp" prefix with "block_sparse_moe"
Implementing special handling for Minimax expert layer naming conventions
Adding Minimax configuration to packed_modules_model_mapping for proper qkv_proj and experts module handling
Without these changes, Minimax models fail to load on Ascend devices due to incompatible layer naming and module packing.

Yes. Users can now successfully load and run Minimax models on Ascend hardware with vLLM. This enables inference capabilities for this model family on Ascend devices.

Local Testing:
Verified model loading for minimax-xxx and minimax_m2-xxx model variants on Atlas 800I A2 hardware
Tested inference with sample prompts using vLLM's OpenAI-compatible API server

Benchmark Validation:
Compared throughput and latency metrics against GPU baseline
Verified memory usage stays within expected limits for different batch sizes
Tested multi-card inference scenarios with tensor parallelism

- vLLM version: v0.13.0
- vLLM main: https://github.com/vllm-project/vllm/commit/8be6432bdaf6275664d857b1e5e9bf8ed1ce299e

### DIFF
--- a/vllm_ascend/quantization/quant_config.py
+++ b/vllm_ascend/quantization/quant_config.py
@@ -121,13 +121,13 @@ class AscendQuantConfig(QuantizationConfig):
         if model_type in ["minimax", "minimax_m2"]:
             prefix = prefix.replace("mlp", "block_sparse_moe")
 
-        #To adapt to minimax, modify the prefix of the model layer name
-        parts = prefix.split('.')
-        if "experts" in parts and len(parts) > 2:
-            exp_idx = parts.index("experts")
-            if exp_idx + 1 < len(parts) and parts[exp_idx + 1].isdigit():
-                parts = parts[:exp_idx + 1]
-                prefix = ".".join(parts)
+            #To adapt to minimax, modify the prefix of the model layer name
+            parts = prefix.split('.')
+            if "experts" in parts and len(parts) > 2:
+                exp_idx = parts.index("experts")
+                if exp_idx + 1 < len(parts) and parts[exp_idx + 1].isdigit():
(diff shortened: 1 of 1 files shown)


realliujiaxu:
Thanks for your contribution! Could you explain why replacing "mlp" prefix with "block_sparse_moe"?

Feng-xiaosuo:
> Thanks for your contribution! Could you explain why replacing "mlp" prefix with "block_sparse_moe"?

In the vLLM Ascend backend, we use a unified quantization loading mechanism that expects specific module naming patterns. For most models (like Llama, Qwen), the expert layers use "mlp" as the prefix (e.g., model.layers.0.mlp.gate_proj).

However, Minimax models have a different architecture:

They use a block-sparse mixture-of-experts (MoE) implementation
Their weight files contain layers named with "block_sparse_moe" prefix (e.g., model.layers.0.block_sparse_moe.gate)
The standard "mlp" prefix doesn't exist in their actual weight files
Without this replacement:

During quantization loading, the system tries to find weights with "mlp" prefix
These weights don't exist in Minimax model files
The quant_config.py module fails to map weight names to modules
Results in KeyError when loading quantized models
This replacement ensures the weight loading path correctly matches Minimax's actual layer naming convention. It's not a generic transformation but a model-specific adaptation required for compatibility with the Ascend backend's quantization infrastructure.

wangxiyuan (approved):

PR #5642, opened 2026-01-06
[Nightly] Move ops to the correct path

Potabk:
Move ops to the correct path where they belong

- vLLM version: v0.13.0


.github/workflows/_e2e_test.yaml
```diff
@@ -99,7 +99,6 @@ jobs:
           # We found that if running aclgraph tests in batch, it will cause AclmdlRICaptureBegin error. So we run
           # the test separately.
 
-          pytest -sv --durations=0 tests/e2e/nightly/single_node/ops/singlecard_ops/triton/
```
MengqingCao: why remove this here?

Potabk: term is too long, we may consider test a small part of them to make per pr CI lightweight and fast if must

Potabk:
also cc @Trunrain

Potabk:
@wangxiyuan let's merge it to recover nightly ops tests

PR #5612, opened 2026-01-05
[CI] Specify the version of xlite

changdawei1:
### What this PR does / why we need it?
Pin the xlite version to avoid CI failures during its upgrade.
### How was this patch tested?

- vLLM version: v0.13.0
- vLLM main: https://github.com/vllm-project/vllm/commit/7157596103666ee7ccb7008acee8bff8a8ff1731

PR #5646, opened 2026-01-06
[CI] Add workflow to cancel running workflows on PR close

pacoxu:
Example: https://github.com/vllm-project/vllm-ascend/actions/runs/20735955959/job/59533181655 is still running after https://github.com/vllm-project/vllm-ascend/pull/5612 is closed.

And the action will be running for more than 2 hours, which needs to be cleanup.

It seems that the Github Aciton will not cancel it automatically, so I add this to cannel those PR related actions once it is closed.

Tested in https://github.com/pacoxu/pacoxu/actions/runs/20743173119.
- vLLM version: v0.13.0

### DIFF
--- a/.github/workflows/pr_close_cancel_job.yaml
+++ b/.github/workflows/pr_close_cancel_job.yaml
@@ -1,0 +1,46 @@
+name: Cancel runs on PR close
+on:
+  pull_request:
+    types: [closed]
+
+permissions:
+  actions: write
+  contents: read
+
+jobs:
+  cancel:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v7
+        with:
+          github-token: ${{ github.token }}
+          script: |
+            const { owner, repo } = context.repo;
+            const branch = context.payload.pull_request.head.ref;
+
+            const statuses = ["in_progress", "queued", "waiting", "pending", "requested"];
+            for (const status of statuses) {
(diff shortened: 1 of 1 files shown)


pacoxu:
cc @MengqingCao @wangxiyuan

.github/workflows/pr_close_cancel_job.yaml
```diff
@@ -0,0 +1,46 @@
+name: Cancel runs on PR close
```
wangxiyuan, line 1: rename this file to `pr_close_cancel_job` to keep the same style like others

.github/workflows/pr_close_cancel_job.yaml
```diff
@@ -0,0 +1,46 @@
...
+on:
+  pull_request:
+    types: [closed]
```
wangxiyuan, line 4: if the pr is merged, does this job still work?

.github/workflows/pr_close_cancel_job.yaml
pacoxu: Updated.

.github/workflows/pr_close_cancel_job.yaml
pacoxu: Yes. `closed` includes `merged`.

wangxiyuan:
Thanks

PR #5641, opened 2026-01-06
[Feat] Integrate FIA operator in mla_cp._forward_decode

845473182:
### What this PR does / why we need it?
Replace the npu_multi_head_latent_attention with FIA operator in mla_cp.py _forward_decode.
Adjust mla_attn_dpc_pcp in acl_graph.py
### Does this PR introduce _any_ user-facing change?
no
### How was this patch tested?

- vLLM version: v0.13.0
- vLLM main: https://github.com/vllm-project/vllm/commit/2f4e6548efec402b913ffddc8726230d9311948d

PR #5659, opened 2026-01-06
[Feat] Remove Redundant Variables after Integrate FIA operator in mla_cp._forward_decode

dsxsteven:
PCP/DCP splits the kv-cache onto different cards. After introducing the parameter `cp-kv-cache-interleave-size`, the first `size` tokens will be cached at Card 0, and so on.
However, if there are too few tokens, some cards will not store the key-value pairs, resulting in values ​​of 0, corrupted values, and precision issues. Currently, additional operations are introduced to avoid this precision problem. 

After we integrate FIA operator in mla_cp._forward_decode, we now can **remove** these additional operations.

No

- vLLM version: v0.13.0


dsxsteven:
This PR should be merged after #5641

PR #5669, opened 2026-01-06
VL model enable flashcomm V1

shaopeng-666:

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/ops/layernorm.py
+++ b/vllm_ascend/ops/layernorm.py
@@ -18,8 +18,12 @@
 from typing import Optional, Tuple, Union
 
 import torch
+import torch.nn.functional as F
 from vllm.config import get_current_vllm_config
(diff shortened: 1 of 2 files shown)


vllm_ascend/patch/worker/patch_qwen3vl.py
```diff
@@ -0,0 +1,41 @@
...
+                k: v.chunk(tp_size)[tp_rank]
+                for k, v in deepstack_input_embeds.tensors.items()
+            }
+        return deepstack_input_embeds
+
+    return wrap
+
+
+Qwen3VLForConditionalGeneration._get_deepstack_input_embeds = tensor_parallel_wrap(
+    Qwen3VLForConditionalGeneration._get_deepstack_input_embeds)
```
gcanlin, line 41: Could you please add the plan to remove the patch in future and explain its necessity? thx!

PR #5692, opened 2026-01-07
[main][bugfix] Fix fullgraph padding bug in mtp eagle refactor

lilinsiman:
The condition for determining padding in the fullgraph overlay with MTP and PCP has been modified to accommodate corner cases where the shape capture size is manually specified.

no

ut and tests

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -963,23 +963,27 @@ class NPUModelRunner(GPUModelRunner):
 
             # TODO: We should make this official ASAP. Also note that if we pad here,
             # the builders won’t need to add any extra padding.
-            max_decode_tokens = min(self.scheduler_config.max_num_seqs * self.uniform_decode_query_len, self.cudagraph_batch_sizes[-1])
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -965,7 +965,7 @@ def _prepare_inputs(
             # the builders won’t need to add any extra padding.
             max_decode_tokens = self.scheduler_config.max_num_seqs * self.uniform_decode_query_len
             if self.compilation_config.cudagraph_mode.decode_mode() == CUDAGraphMode.FULL and \
-                uniform_decode and self.uniform_decode_query_len <= num_input_tokens <= max_decode_tokens:
+                uniform_decode and self.uniform_decode_query_len <= num_input_tokens <= min(max_decode_tokens, self.cudagraph_batch_sizes[-1]):
```
yiz-liu: I suggest
```python
max_decode_tokens = min(self.scheduler_config.max_num_seqs * self.uniform_decode_query_len, self.cudagraph_batch_sizes[-1])
```
It's more resonable.

lilinsiman: done

yiz-liu (approved):

PR #5711, opened 2026-01-08
[BugFix] Xlite: Bypass the padding of the graph mode in non-MTP cases to obtain the correct decode num.

wwwumr:
This PR fixes a bug in Xlite backend(https://atomgit.com/openeuler/GVirt/issues/1), The direct cause of the problem is that the XModel::PrepareAttn function obtained an illegal number of tokens to be inferred, -540. This illegal value is due to the padding feature of inference in graph mode and the residual state across steps. This issue is triggered when a prefill request is newly added in a step and a decode ends simultaneously. It is first fixed using num_decode_tokens instead of attn_metadata.num_decodes.
1. In graph mode, vllm_ascend has padding characteristics. In the _prepare_inputs function, if the number of tokens to be inferred is less than the set threshold (8 in this case), the attn_metadata.num_decode array will be expanded to 8. 
2. Meanwhile, vllm_ascend uses the class variable self.query_start_loc of NPUModelRunner to record the tokens to be inferred. Due to poor coordination with the graph mode padding mechanism when crossing steps, in some cases (such as when a decode request is completed in a certain step and a new prefill request is added at the same time), negative values may be calculated for attn_metadata.query_lens.
3. After type conversion, the negative values in query_lens cause an overflow. Xlite detects that the number of tokens to be inferred for the decode request is too large and triggers a "decode len too long" alert.

No

Same with https://atomgit.com/openeuler/GVirt/issues/1
- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/xlite/xlite.py
+++ b/vllm_ascend/xlite/xlite.py
@@ -255,7 +255,14 @@ class XliteWrapper:
         ]
 
         if not with_prefill or self.full_mode:
-            batch = attn_metadata.num_prefills + attn_metadata.num_decodes
+            # TODO: When vllm_ascend enables graph mode, attn_metadata.num_decodes
+            # will be padded in decode requests. Therefore, it is first fixed using
+            # num_decode_tokens. However, in the future, when MTP is enabled, there
+            # may be cases where a single request involves multiple tokens, which
+            # will need to be solved.
+            num_decodes = attn_metadata.num_decode_tokens
+            num_prefills = attn_metadata.num_prefills
+            batch = num_prefills + num_decodes
             seq_lens = attn_metadata.seq_lens[:batch]
             seq_tensor = torch.cat([
                 torch.tensor([0]),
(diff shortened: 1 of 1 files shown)


vllm_ascend/xlite/xlite.py
```diff
@@ -267,6 +283,12 @@ def __call__(
             cached_lens = seq_lens - query_lens
 
             xlite_attn_metadata = ModelAttnMeta()
+            query_list = query_lens.tolist()
+            if not is_query_list_valid(query_list,
+                                       attn_metadata.num_decode_tokens):
+                logger.error(
+                    f"query_list format error, attn_metadata: {attn_metadata}")
+                raise Exception(f"query_list format error: {query_list}")
             xlite_attn_metadata.lens = query_lens.tolist()
```
lulina: use query_list, and test performance

weijinqian0 (approved):

PR #5726, opened 2026-01-08
[P/D] fix forced free bug for mooncake connector

liziyu179:
Fix the bug where the P-node's schedule dead after it force-frees a request due to timeout and then receives the completed kv cache pulled by the D-node again. Ignore this request in _update_from_kv_xfer_finished function.

- vLLM version: v0.13.0


LCAIZJ:
If the community encounters the same issue, I suggest opening an issue in the vLLM community to discuss potential solutions together.

PR #5731, opened 2026-01-08
[CI] Speed up e2e test

Potabk:
e2e_2card and e2e_4card run with parallel, which is expected to reduce run e2e time 1h

- vLLM version: v0.13.0

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -239,9 +239,7 @@ jobs:
           pytest -sv --durations=0 tests/e2e/multicard/2-cards/test_single_request_aclgraph.py
 
   e2e-4-cards:
-    name: multicard-4
-    needs: [e2e, e2e-2-cards]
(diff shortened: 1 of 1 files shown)


.github/workflows/_e2e_test.yaml
```diff
@@ -239,9 +239,7 @@ jobs:
           pytest -sv --durations=0 tests/e2e/multicard/2-cards/test_single_request_aclgraph.py
 
   e2e-4-cards:
-    name: multicard-4
-    needs: [e2e, e2e-2-cards]
-    if: ${{ needs.e2e.result == 'success' && needs.e2e-2-cards.result == 'success' && inputs.type == 'full' }}
+    needs: e2e
```
Yikun, line 242: seem no improment, because e2e 2 cards time less than e2e 1 cards....

Potabk: This is not very reasonable; will lead to the in queue times for most pr;

PR #5779, opened 2026-01-11
[0.13.0][Bugfix] Fix memory inconsistency in cross-process shared memory

slippersss:

This PR aims to fix potential memory inconsistency in current cross-process shared memory. We use `memory_fence()` to ensure the order of the buffer (i.e. `buf`) and flag (i.e. `metadata_buffer[0]`) writes.

`SHM_BARRIER` env is introduced for controlling this fix. `SHM_BARRIER=true` means using this fix while `SHM_BARRIER=false` means not using it. Currently, `true` is the default value.

by ci


wangxiyuan:
rebase to enable CI. Thanks

Issue #5476, opened 2025-12-29
[RFC]: Support fused operator DispatchGmmCombineDecode


puxiaokun3721:
### Motivation.

Currently, vllm-ascend uses the MC2 operators Dispatch and Combine to transfer tokens across experts. These operators are executed serially with the experts' GMM computations, causing the token transfer latency to be fully exposed and potentially accounting for 10%–30% of the end-to-end latency.
By fusing Dispatch, Combine, and GMM into a single operator, the token transfer latency can be hidden at a fine granularity by the GMM computation, thereby reducing end-to-end latency and improving throughput.

### Proposed Change.

The fused operator DispatchGmmCombineDecode is designed to replace the original MC2 path composed of multiple operators (Dispatch → GMM → SwiGLU → GMM → Combine).  The scenarios that are already supported and planned to be supported include:

- [ ] hardware
  - [x] A3
  - [ ] A5
- [ ] quant type
  - [x] W8A8 https://github.com/vllm-project/vllm-ascend/pull/5040
  - [ ] W4A8 
  - [ ] W4A16
  - [ ] W16A16
- [ ] support dynamic EPLB
- [ ] support coexistence with MC2 operators (e.g., the main model uses W8A8 with the fused operator, while MTP uses W16A16 with MC2 operators)

The functionality of the operator can be referenced in this test case
https://github.com/vllm-project/vllm-ascend/blob/main/tests/e2e/nightly/multicard_ops/test_dispatch_gmm_combine_decode.py

### CC List.

@Yikun @wangxiyuan @MengqingCao @ApsarasX

PR #5806, opened 2026-01-12
[Bugfix] Fix acc bug when enbale dispatch_gmm_combine_decode and eplb

wangqiankun13:

Fix acc bug when enbale dispatch_gmm_combine_decode and eplb.

After eplb, expert table may change, so mapping is needed, while fused_mc2 miss the mapping.

More info about this operator, please refer to RFC: issue https://github.com/vllm-project/vllm-ascend/issues/5476

No

without this pr, qwen3-235b eplb with dispatch_gmm_combine_decode get acc 3.33% on aime2024.

with this pr,

test qwen3-235b eplb on a single A3 node(ep16)
without dispatch_gmm_combine_decode
| aime2024 | 604a78 | accuracy | gen | 86.67 |

with dispatch_gmm_combine_decode
| aime2024 | 604a78 | accuracy | gen | 86.67 |

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/ops/fused_moe/moe_comm_method.py
+++ b/vllm_ascend/ops/fused_moe/moe_comm_method.py
@@ -300,6 +300,11 @@ class FusedMC2CommImpl(MoECommMethod):
 
         assert isinstance(self.token_dispatcher, TokenDispatcherWithMC2), \
             "token_dispatcher must be an instance of TokenDispatcherWithMC2."
+
+        # Apply log2phy if needed
+        if log2phy is not None:
+            topk_ids = log2phy[topk_ids]
+
         group_list_type = None
         expert_tokens = None
         if envs_ascend.VLLM_ASCEND_ENABLE_FUSED_MC2 == 1:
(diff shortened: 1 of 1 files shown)


vllm_ascend/ops/fused_moe/moe_comm_method.py
```diff
@@ -331,7 +336,7 @@ def fused_experts(
                 group_ep=self.token_dispatcher.moe_all_to_all_group_name,
                 ep_rank_size=self.token_dispatcher.ep_world_size,
                 ep_rank_id=self.token_dispatcher.ep_rank_id,
-                moe_expert_num=len(expert_map),
+                moe_expert_num=self.moe_config.num_experts,
                 global_bs=self.token_dispatcher.fused_global_bs)
```
weijinqian0, line 340: why use this, this may cause error in top9 scenario.

shenchuxiaofugui: 1. Top9 has not been committed to the repository yet.
2. len(expert_map) is the number of logic experts. self.moe_config.num_experts is the number of physical which is needed by fused op.
3. The only issue is that self.moe_config only retains the last layer's settings. For example, when the number of physical experts in the MTP layer is different from that in other layers, the MTP layer will be applied to other layers, which may cause a bug. However, this can be avoided through configuration.

PR #5819, opened 2026-01-12
3vl model enbale flashcommV1

shaopeng-666:

### DIFF
--- a/vllm_ascend/ops/layernorm.py
+++ b/vllm_ascend/ops/layernorm.py
@@ -20,10 +20,10 @@ from typing import Optional, Tuple, Union
 import torch
 import torch.nn.functional as F
 from vllm.config import get_current_vllm_config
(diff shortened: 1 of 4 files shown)


vllm_ascend/ops/layernorm.py
```diff
@@ -50,6 +54,17 @@ def forward_oot(
 
         from vllm_ascend.utils import AscendDeviceType, get_ascend_device_type
         if residual is not None:
+            if x.size(0) != residual.size(0):
```
realliujiaxu, line 57: replace this with torch custom op `maybe_chunk_residule`. op from `torch.ops` **will not** be traced inside by torch Dynamo, so `if...else...` in it can be chosen dynamically when capturing aclgraph.

shaopeng-666: Have fixed, new code link is https://github.com/vllm-project/vllm-ascend/pull/5848

PR #5822, opened 2026-01-12
[Feature] Mooncake connector get remote ptp size

yuxinshan:
To support elastic scaling when using mooncake connector, we should support to **configure different tp sizes for different nodes**.
As a result, we transfer the prefill node information, such as tp size, through **the request's kv_transfer_params**.
The decode nodes **get the prefill tp size** through the request's kv_transfer_params, instead of getting it from the configuration of the mooncake connector .

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
@@ -87,6 +87,7 @@ class ReqMeta:
     remote_request_id: str
     remote_pcp_size: int
     remote_dcp_size: int
+    remote_ptp_size: int | None
     remote_multi_nodes_meta_mapping: dict[str, dict[str, Any]]
     num_prompt_blocks: int
 
@@ -773,6 +774,7 @@ class MooncakeConnectorMetadata(KVConnectorMetadata):
             remote_port=kv_transfer_params["remote_port"],
             remote_pcp_size=kv_transfer_params.get("remote_pcp_size", 1),
             remote_dcp_size=kv_transfer_params.get("remote_dcp_size", 1),
+            remote_ptp_size=kv_transfer_params.get("remote_ptp_size"),
             remote_multi_nodes_meta_mapping=kv_transfer_params.get("remote_multi_nodes_meta_mapping", {}),
             num_prompt_blocks=kv_transfer_params.get("num_prompt_blocks", 0),
         )
@@ -890,6 +892,7 @@ class MooncakeConnectorScheduler:
         self.side_channel_host = get_ip()
         self.pcp_size = vllm_config.parallel_config.prefill_context_parallel_size
         self.dcp_size = vllm_config.parallel_config.decode_context_parallel_size
+        self.tp_size = vllm_config.parallel_config.tensor_parallel_size
         self.max_device_id = (
             vllm_config.parallel_config.tensor_parallel_size
             * vllm_config.parallel_config.data_parallel_size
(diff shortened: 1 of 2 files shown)


weijinqian0 (approved):

LCAIZJ:
Could you please resolve the conflicts?

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
```diff
@@ -1569,12 +1574,17 @@ def start_load_kv(self, metadata: MooncakeConnectorMetadata):
                 "Num local_block_ids: %s. Num remote_block_ids: %s. ", req_id,
                 meta.remote_engine_id, len(meta.local_block_ids),
                 len(meta.remote_block_ids))
+
+            prefill_tp_size = meta.remote_ptp_size if getattr(
```
liziyu179: Under what circumstances would the remote_ptp_size in req meta be None?

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
```diff
@@ -1348,8 +1351,9 @@ def _get_kv_split_metadata(
         In cp/dcp scenario, kv_cache may be split, so we need to pull multiple blocks from multiple remote P node.
         Use this function to calculate remote port and remote block number of each remote P node that we need to pull.
         """
+        prefill_tp_size = meta.remote_ptp_size if getattr(meta, "remote_ptp_size", None) else self._prefill_tp_size
```
liziyu179, line 1331: Perhaps we shouldn't obtain the prefill tp size and decode tp size from the kv config.

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
yuxinshan: In normal scenarios, the remote_ptp_size will not be None. But when we only test the part of the decode, the meta will not has the attribute of remote ptp size.

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
yuxinshan: To support different tp size of prefill nodes, we get the prefill tp size from the kv transfer params refered to nixl.

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
```diff
@@ -1556,8 +1560,9 @@ def get_remote_port_send_num(local_remote_block_port_mappings):
                                      num_blocks_to_pull])
             local_block_offset += num_blocks_to_pull
 
-        assert self.tp_num_need_pulls == len(remote_handshake_port_list[0]), \
-        f"tp_num_need_pulls: {self.tp_num_need_pulls}, remote_handshake_port_list: {remote_handshake_port_list}"
+        tp_num_need_pulls = self._get_tp_num_need_pulls(prefill_tp_size)
```
LCAIZJ: `self. tp_num_need_pulls=self._get_tp_num_need_pulls(prefill_tp_size)` If we make this change, will it be compatible with the subsequent logic, thereby reducing the need for further modifications later?

LCAIZJ:
Could we modify the `_get_prefill_decode_size` function to obtain the value of `self._prefill_tp_size` from `kv_transfer_params`, so that we can minimize modifications elsewhere?

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
yuxinshan, line 1526: tp_num_need_pulls will be recalculated when the prefill tp size is different from the self._prefill_tp_size.

LCAIZJ:
I feel this feature is unrelated to elastic scaling; it primarily addresses the heterogeneity among multiple Prefill nodes in a PD-separated architecture.

PR #5673, opened 2026-01-06
Ensure that the qwen3-vl model uses the mrope operator(fix issue#5670)

Xiangyanglikecode:
### What this PR does / why we need it?
Resolve iusse #5670 , currently the mrope operator only supports the mrope section of qwen2.5-vl, but the shape of qwen3-vl is different, and the inability to use the mrope operator will result in rollback. Therefore, add the shape of qwen3-vl so that the qwen3-vl model can also use the mrope operator
- Fixes #5670  
### Does this PR introduce _any_ user-facing change?
no

#### code
Collect profiling for testing using the offline inference script test_outline.py
```
from transformers import AutoProcessor
from vllm import LLM, SamplingParams
from qwen_vl_utils import process_vision_info
import os
from torch_npu.profiler.profiler import analyse
MODEL_PATH = "Qwen3-VL-8B-Instruct"

#profiling_path="./vllm_ascend_profile_outline"
#os.environ["VLLM_TORCH_PROFILER_DIR"]=profiling_path
os.environ["HCCL_BUFFSIZE"]='1024'
os.environ["VLLM_USE_V1"]='1'
os.environ["OMP_NUM_THREADS"]='10'
os.environ["CPU_AFFINITY_CONF"]='1'
os.environ["TASK_QUEUE_ENABLE"]='1'

os.environ["VLL_ASCEND_ENABLE_TOPK_OPTIMIZE"]='1'
os.environ["MIES_USE_MB_SWAPPER"]='1'

llm = LLM(
    model=MODEL_PATH,
    max_model_len=16384,
    limit_mm_per_prompt={"image": 10},
    compilation_config ={"cudagraph_mode":"FULL_DECODE_ONLY", "cudagraph_capture_sizes":[1, 2, 4, 8, 16, 24]},
    enable_prefix_caching=False,
    mm_processor_cache_gb = 0,
)

sampling_params = SamplingParams(
    max_tokens=10
)

image_messages = [
    {"role": "system", "content": "You are a helpful assistant."},
    {
        "role": "user",
        "content": [
            {
                "type": "image",
                "image": "https://modelscope.oss-cn-beijing.aliyuncs.com/resource/qwen.png",
                "min_pixels": 224 * 224,
                "max_pixels": 1280 * 28 * 28,
            },
            {"type": "text", "text": "Please provide a detailed description of this image"},
        ],
    },
]

messages = image_messages

processor = AutoProcessor.from_pretrained(MODEL_PATH)
prompt = processor.apply_chat_template(
[...]

PR #5827, opened 2026-01-13
[Performance]use triton mrope for Qwen3-VL

ichaoren:

The performance can be optimized by use triton mrope for Qwen3-VL.

1. Qwen3-VL-235B-A22B-Instruct-W8A8

  | accuracy | TTFT | TPOT | TPS
-- | -- | -- | -- | --
base | 83.76 | 4.8771s | 0.1472 | 49
test | 83.59 | 4.3273 | 0.0615 | 105

2. Qwen3-VL-8B-Instruct

  | accuracy | TTFT | TPOT | TPS
-- | -- | -- | -- | --
base | 80.65 | 4.1744 | 0.0499 | 125
test | 80.86 | 3.1858 | 0.0245 | 227

Name: triton-ascend
Version: 3.2.0.dev20260105

vllm serve $MODEL_DIR \
 --served-model-name $MODEL_NAME \
 --quantization ascend \
 --tensor-parallel-size 8 \
 --data-parallel-size 1 \
 --enable-expert-parallel \
 --max-model-len 20000 \
 --allowed-local-media-path /vllm-workspace/vllm/tests/multimodal/assets/ \
 --enforce-eager \
 --max-num-batched-tokens 5000 \
 --gpu-memory-utilization 0.8 \
 --trust-remote-code \
 --port 8800 \
 --no-enable-prefix-caching \
 --mm_processor_cache_type="shm" \
 --async-scheduling

- performance test
evalscope perf   --parallel 8  --model qwen3vl  --url http://ip:port/v1/chat/completions   --api openai   --dataset random_vl   --min-tokens 300  --max-tokens 300  --prefix-length 0   --min-prompt-length 1024   --max-prompt-length 1024   --image-width 2048  --image-height 2048   --image-format RGB   --image-num 1  --number 32   --tokenizer-path /path/Qwen3-VL-235B-A22B-Instruct-W8A8/

- Accuracy Test
ais_bench --models vllm_api_stream_chat --datasets textvqa_gen --debug


vllm_ascend/ops/rotary_embedding.py
```diff
@@ -533,6 +533,10 @@ def forward_oot(
         query: torch.Tensor,
         key: torch.Tensor,
     ):
+        # use triton mrope for Qwen3-VL
+        if self.mrope_section == [24, 20, 20]:
+            return super().forward_cuda(positions, query, key)
```
gcanlin: `forward_cuda` is specific for CUDA platform instead of NPU platform. We should invoke triton kernel in `forward_oot` if you want to enable triton. FYI, there is the similar work: https://github.com/vllm-project/vllm-ascend/pull/5673.

PR #5830, opened 2026-01-13
[P/D] fix forced free bug for mooncake connector 

wangxiaoteng888:
Fix the bug where the P-node's schedule dead after it force-frees a request due to timeout and then receives the completed kv cache pulled by the D-node again. Ignore this request in _update_from_kv_xfer_finished function.

No

By ci


LCAIZJ:
Is this the same as this PR https://github.com/vllm-project/vllm-ascend/pull/5726?

PR #5833, opened 2026-01-13
[BugFix]converting pa get_workspace back to capturing

Angazenn:

This helps to fix a bug in for pa get_workspace. In earlier implementation, we use `_npu_paged_attention_get_workspace` in `_update_pa_attn_params`. However, this might cause some potential memory problems as it dynamically allocate new memory for workspace when calling this api. Therefor, we move this back to capturing, and use a fixed `SEQ_LEN_WITH_MAX_PA_WORKSPACE` to get max workspace.

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -76,7 +76,7 @@ from vllm.v1.worker.utils import AttentionGroup
 
 from vllm_ascend.ascend_config import get_ascend_config
 from vllm_ascend.attention.attention_v1 import AscendAttentionState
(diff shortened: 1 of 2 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1916,7 +1919,7 @@ def _build_dummy_attn_metadata(
 
             attn_metadata = {}
 
-            seq_lens = max_query_len
+            seq_lens = SEQ_LEN_WITH_MAX_PA_WORKSPACE if force_attention else max_query_len
```
yiz-liu: Yeah, this seems fine but we need to add some comments here.

PR #5848, opened 2026-01-13
enbale qwen3-vl model fc1 feature

brandneway:
enble fc1 feature for qwen3-vl models

no

running vl dense and moe model with fc1 feature, i curl a request,it returns correctly


vllm_ascend/utils.py
```diff
@@ -1189,3 +1189,14 @@ def enable_dsa_cp_with_layer_shard() -> bool:
     vllm_config = get_current_vllm_config()
     is_prefill_instance = vllm_config.kv_transfer_config is not None and vllm_config.kv_transfer_config.is_kv_producer
     return is_prefill_instance
+
+
+def parse_layer_idx(prefix: str) -> Optional[int]:
```
Levi-JQ, line 1206: Can we use extract_layer_index in vllm.model_executor.models.utils?

zzhx1:
It seems that your PR has a lot of logic similar to #5669, please evaluate it and simplify your code.

shaopeng-666:
> It seems that your PR has a lot of logic similar to #5669, please evaluate it and simplify your code.

Same author..This pr fix precision and function problem on dense model based pr #5669 .#5669 will be closed  once this pr is merge into 0.13.0 release. And this pr will be cherry pick to main branch later.

vllm_ascend/utils.py
shaopeng-666: Using extract_layer_index  will cause Linear func UT failed. The reason is "extract_layer_index" force you to pass a string include a number, if not, it will be asserted. And vllm-ascend UT test gives a string not satisfied this func's requirement.

PR #5852, opened 2026-01-13
[doc]index display by category

herizhen:
upgrade tutorial doc index display by category

no

ut
- vLLM version: v0.13.0


herizhen:
**To the maintainers:** An inconsistency has been found in the file reference formats within tutorials/index.md. For example, the file ray.md physically has the .md extension in the directory, but must be referenced as ray (without the extension) in the index to pass the CI check. It is suggested to consider documenting this special rule or unifying the reference format in the conf.py configuration to prevent contributors from encountering similar CI check confusion in the future. @wangxiyuan

PR #5865, opened 2026-01-13
use  Vit Sp for Qwen3-VL

TaDa-aaa:

The performance can be optimized about 4% by use vit sp for Qwen3-VL.

Name: Vit Sp

vllm serve $MODEL_DIR \
--host 0.0.0.0 \
--port 8000 \
--served-model-name $MODEL_NAME \
--tensor-parallel-size 4 \
--max-model-len 5000 \
--max-num-batched-tokens 5000 \
--trust-remote-code \
--enable-prefix-caching \
--gpu-memory-utilization 0.8 \
--allowed-local-media-path / \
--media-io-kwargs '{"video":{"num_frames":5}}' \
--additional-config='{"enable_cpu_binding":true}' \
--compilation-config '{"cudagraph_mode":"FULL_DECODE_ONLY","cudagraph_capture_sizes":[1,2,4,8,16,32,64,80,96,128]}' \
--async-scheduling \
--mm-processor-cache-gb 0 \
--quantization ascend

- performance test

evalscope perf   --parallel 128   --model $MODEL_NAME   --url http://ip:port/v1/chat/completions   --api openai   --dataset random_vl   --min-tokens 256   --max-tokens 256   --prefix-length 0   --min-prompt-length 8   --max-prompt-length 8   --image-width 1080   --image-height 1920   --image-format RGB   --image-num 1  --number 512   --tokenizer-path  $MODEL_DIR

- Accuracy Test

ais_bench --models vllm_api_stream_chat --datasets textvqa_gen --debug

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -215,6 +215,7 @@ jobs:
           pytest -sv --durations=0 tests/e2e/multicard/2-cards/test_expert_parallel.py
           pytest -sv --durations=0 tests/e2e/multicard/2-cards/test_external_launcher.py
           pytest -sv --durations=0 tests/e2e/multicard/2-cards/test_full_graph_mode.py
+          pytest -sv --durations=0 tests/e2e/multicard/2-cards/spec_decode/test_spec_decode.py
           # torch 2.8 doesn't work with lora, fix me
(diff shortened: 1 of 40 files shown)


wangxiyuan:
please consider to contribute to main branch instead.

Issue #5467, opened 2025-12-29
[RFC]: Refactor and unify eagle_proposer and mtp_proposer


slippersss:
### Motivation.

To catch up with vllm-eagle, we are planning to refactor and unify eagle_proposer and mtp_proposer in vllm-ascend.

Considering significant amount of code duplication between ascend-eagle and ascend-mtp, together with the requirement to ultimately converge to vllm-eagle, we would like to build inheritance chain among these three classes, i.e., "vllm-eagle <----- ascend-eagle <----- ascend-mtp". In this way, we can gradually delete ascend-mtp and enable ascend-eagle to converge to vllm-eagle, while not breaking the current existing features.

Once the above step is finished, we will try to refactor eagle at a deeper level, enabling several ascend-specific optimizations to be located in a suitable place, involving full graph, expert parallel dp, lmhead tp, pcp dcp, etc.

### Proposed Change.

CI:
1. e2e acceptance test for eagle/eagle3/mtp
#5366 (eagle/eagle3)
#5326 (qwen3_next_mtp)
#5337 (deepseek_mtp)
2. e2e correctness test for eagle/eagle3/mtp (unified)
#5326
3. ut completion
coming soon...

Refactor:
1. delete `init` method in mtp_proposer, build inheritance chain for the upcoming PRs
#5176
2. delete `load_model` and `generate_token_ids` method in mtp_proposer, align `propose` interface called by model runner to that in vllm-eagle
#5437
3. delete the rest duplicate methods in mtp_proposer, leave only `propose` and `dummy_run`
#5420
4. refactor and unify `propose` `dummy_run` in eagle_proposer and mtp_proposer, delete mtp_proposer
coming soon...
5. correct and validate acceptance of eagle_proposer
coming soon...
6. delete similar methods in eagle_proposer, compared with vllm-eagle (need triton-ascend and AscendCommonMetadata-refactoring support)
coming soon...
7. continue refactoring eagle...
...

### CC List.

@wangxiyuan @MengqingCao @weijinqian0 @yiz-liu @zzzzwwjj @realliujiaxu @Yikun
@zhenwenqi2024 @zouyida2052 @lilinsiman

lilinsiman:
PR 5496 devided to parts, the first one: https://github.com/vllm-project/vllm-ascend/pull/5869

PR #5870, opened 2026-01-13
[Refactor][EAGLE] 4/N extract common methods from eagle and mtp

slippersss:
This PR aims to extract common methods from eagle_proposer and mtp_proposer. This is a small step towards merging eagle and mtp.

RFC: #5467

N/A

by ci

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/spec_decode/eagle_proposer.py
+++ b/vllm_ascend/spec_decode/eagle_proposer.py
@@ -384,11 +384,17 @@ class EagleProposer(VllmEagleProposer):
             model_hidden_states, model_positions = self.maybe_pad_and_reduce(
                 model_hidden_states, model_positions)
 
-            last_hidden_states, hidden_states = self._model_forward(
+            ret_hidden_states = self.model(
                 input_ids=model_input_ids,
                 positions=model_positions,
                 hidden_states=model_hidden_states,
             )
+            if self.method == "mtp":
+                last_hidden_states = ret_hidden_states
+                hidden_states = last_hidden_states
+            else:
+                last_hidden_states, hidden_states = ret_hidden_states
+
             forward_context = get_forward_context()
             if forward_context.cudagraph_runtime_mode == CUDAGraphMode.FULL:
                 self._update_full_graph_params(forward_context,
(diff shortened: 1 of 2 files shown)


wangxiyuan (approved):

vllm_ascend/spec_decode/eagle_proposer.py
```diff
@@ -969,3 +923,77 @@ def _update_full_graph_params(self, forward_context, num_tokens):
...
+            positions=positions,
+            hidden_states=hidden_states,
+            inputs_embeds=inputs_embeds,
+        )
+        if self.method == "mtp":
+            last_hidden_states = ret_hidden_states
+            hidden_states = last_hidden_states
+        else:
+            last_hidden_states, hidden_states = ret_hidden_states
+        return last_hidden_states, hidden_states
```
yiz-liu: Maybe we don't need to extract this to a independent function.

vllm_ascend/spec_decode/eagle_proposer.py
```diff
@@ -969,3 +923,77 @@ def _update_full_graph_params(self, forward_context, num_tokens):
...
+    def maybe_unpad_attn_metadata(self, attn_metadata: dict) -> None:
+        for layer_name in self.attn_layer_names:
+            decode_metadata = getattr(attn_metadata[layer_name], "decode",
+                                      None)
+            if self.use_async_scheduling and decode_metadata is not None:
+                actual_size = len(decode_metadata.actual_seq_lengths_q)
+                decode_metadata.seq_lens_list = (
+                    decode_metadata.seq_lens_list[:actual_size])
+                decode_metadata.block_table = (
+                    decode_metadata.block_table[:actual_size])
```
yiz-liu: This should be replaced by `unpadded`, please remove this later.

vllm_ascend/spec_decode/eagle_proposer.py
slippersss: Done.

vllm_ascend/spec_decode/eagle_proposer.py
slippersss: This branch will be triggered only when using async scheduling. Since original common attention metadata has been unpadded in current logic, so it is reasonable to remove this method. I have tested it with an offline script where main model is in eager/piecewise/full_decode_only mode, and it appears to be ok.

yiz-liu (approved):

PR #5896, opened 2026-01-14
[nightly] Remove node tolerations for hk cluster

Potabk:
Since we have upgrade all the nodes' `cann` HDK version to `25.3rc1`, we should not limit nightly schedule to the specific nodes

- vLLM version: v0.13.0


Potabk:
@wangxiyuan are you tend to approve to make CI happy quickly;

Issue #4842, opened 2025-12-09
[RFC]: Proposal for new PD Disaggregation scheme via Request Forwarding to D Node then P Node


liziyu179:
### Motivation.

In large-scale distributed deep learning systems, the need to handle high-throughput requests efficiently becomes critical, especially in cases where data processing is distributed across different nodes, such as D nodes (Data nodes) and P nodes (Processing nodes).

Currently, the system architecture sends requests directly from the Proxy to the Prefill (P) node, which can lead to inefficiencies, particularly in scenarios where coordination and optimal load balancing are needed to ensure smooth data transfer and computation. This request flow is suboptimal for environments with large request volumes or heterogeneous node configurations, where the need to delegate and manage requests properly is paramount for maintaining high throughput and low latency.

This proposal introduces a new architecture that improves request routing by incorporating a Coordinator (C) node as the first point of contact, followed by forwarding the request to the D node (Data node) and eventually to the P node (Processing node). This solution aims to improve coordination, optimize resource allocation, and reduce latency during the request processing lifecycle.

### Proposed Change.

To optimize the request routing, we propose the following architecture:

Initial Request Forwarding to the Coordinator (C):
The request arrives at the Proxy, which forwards it to a designated Coordinator node. The Coordinator’s responsibility is to balance load across the system and manage routing decisions to minimize congestion and optimize resource utilization.

Forwarding to the D Node:
The Coordinator, upon receiving the request, evaluates the status of the D nodes. It selects a Data Node (D) based on the current load or available resources. The Coordinator then forwards the request to the selected D node, which holds the necessary data for further processing.

Data Processing by the P Node:
Once the D node has processed the necessary data, it forwards the request to the corresponding P node (Processing node). The P node performs the actual computation based on the data received from the D node.

Request Flow

Request Arrival:

A request arrives at the Proxy, which routes it to the Coordinator (C) based on load-balancing policies.

The request format at this stage includes essential metadata such as the request_id, remote_engine_id, and any other relevant identifiers required for processing.

Coordinator to D Node:

The Coordinator evaluates the current load and resource status across the available D nodes.

Based on the assessment, the request is forwarded to the selected D node.

The request format sent to the D node contains the necessary parameters, such as:

{
  "request_id": "xxx",
  "dp_rank": 1,
  "remote_host": "xxx",
  "remote_port": "xxx"
}

D Node to P Node:

Upon receiving the request, the D node processes the data and forwards it to the corresponding P node.

[...]

PR #5900, opened 2026-01-14
[P/D] layerwise connector support recompute scheduler

liziyu179:
layerwise connector support recompute scheduler. 

NOTE：
Triggering recompute will invoke the tokenizer again, which may lead to precision fluctuations.

[RFC]: CDCP Scheduling for Disaggregated Prefilling with KV Cache Layerwise Push Support https://github.com/vllm-project/vllm-ascend/issues/4842

- vLLM version: v0.13.0

### DIFF
--- a/examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
+++ b/examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
@@ -444,7 +444,7 @@ async def _handle_completions(api: str, request: Request):
             origin_prompt = ""
         # refer to vLLM sampling_params: max_token default value
         origin_max_tokens = req_data.get("max_tokens", 16)
-        
+
         async def generate_stream():
             nonlocal released_kv
             generated_token = ""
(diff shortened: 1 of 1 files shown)


LCAIZJ:
Could you please provide a more detailed explanation of the background and design of this PR?

examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
```diff
@@ -451,7 +525,10 @@ async def generate_stream():
             # After streaming done, release tokens
             proxy_state.release_decoder(decoder_idx, decoder_score)
```
winson-00178005, line 526: 如果抛出异常的时候，是不是也要调用这个来释放？

examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
```diff
@@ -428,19 +430,91 @@ async def _handle_completions(api: str, request: Request):
...
+                        if retry_count > 0 and not stream_flag:
+                            if chat_flag:
+                                choice["message"]["content"] = generated_token
+                            else:
+                                choice["text"] = generated_token
+                            chunk = json.dumps(chunk_json).encode("utf-8")
+                        yield chunk
             except Exception as e:
                 logger.error(
                     f"Error during streaming from decoder {decoder.url}: {str(e)} "
```
winson-00178005, line 520: 可以把request_id/retry_count加入日志中，方便问题定位

examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
liziyu179: Running to this line of the function signifies the end of stream return, releasing the key-value cache records of node D.

examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
liziyu179: Thanks for reviewing the code. The number of retries will be printed in stream_service_response_with_retry, and the request_id will be printed on the next line.

PR #5912, opened 2026-01-15
[Lint]Style: Convert `vllm-ascend/` to ruff format(Batch #1)

MrZ20:
Convert `vllm-ascend/compilation` to ruff format.

During this migration, we encountered some **errors** in our CI and testing environments, such as:
```
vllm_ascend/utils.py:653: in <module>
    def register_ascend_customop(vllm_config: VllmConfig | None = None):
                                              ^^^^^^^^^^^^^^^^^
E   TypeError: unsupported operand type(s) for |: 'NoneType' and 'NoneType'
```

**1. Root Cause Analysis:**
The project uses a common pattern to break circular dependencies:
```python
if TYPE_CHECKING:
    from vllm.config import VllmConfig
else:
    VllmConfig = None  # Placeholder assigned at runtime
```
When Python parses the function definition `def register_ascend_customop(vllm_config: VllmConfig | None)`, it attempts to evaluate the expression `VllmConfig | None`.
Since `VllmConfig` is assigned `None` at runtime, the expression effectively becomes `None | None`. In Python, `None` is an instance of `NoneType`. While the `|` operator is implemented for Type objects (classes), it is not supported for `NoneType` instances, leading to the `TypeError` shown above.

**2. Solution:**
To maintain the modern `|` syntax required by our new linting standards while preserving our dependency management strategy, I have introduced:
```python
from __future__ import annotations
```
at the top of the affected files. This enables **Postponed Evaluation of Annotations (PEP 563)**.

**3. Impact and Benefits:**
- By enabling `annotations`, Python no longer executes the `VllmConfig | None` operation during module load. Instead, it stores the annotation as a string literal, completely avoiding the `None | None` calculation.
- We can keep the `VllmConfig = None` placeholders. This ensures that other modules can still import these symbols without triggering an `ImportError`, maintaining a stable dependency graph.
- IDEs and static type checkers (MyPy/Pyright) continue to resolve the types correctly. This allows us to use modern syntax without sacrificing type safety or runtime stability.
- The only side effect is that `__annotations__` will now return strings instead of type objects. Since this module does not use runtime type enforcement or reflection, this change has zero negative impact on existing functionality.

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -15,6 +15,8 @@
 # This file is a part of the vllm-ascend project.
 #
 
+from __future__ import annotations
+
 import math
 import os
 from typing import TYPE_CHECKING, Any
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -17,6 +17,8 @@
 # Adapted from vllm-project/vllm/vllm/worker/worker.py
 #
 
+from __future__ import annotations
+
 import atexit
 import functools
 import math


Potabk:
LGTM if full CI passed

PR #5935, opened 2026-01-15
[Misc] Use upstream `mem_utils` for profiling

shen-shanshan:
Following https://github.com/vllm-project/vllm/pull/32322, use the `memory_profiling` context manager in vllm for profiling.

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/worker/worker.py
+++ b/vllm_ascend/worker/worker.py
@@ -211,6 +211,14 @@ class NPUWorker(WorkerBase):
         gc.collect()
         torch.npu.empty_cache()
 
+        GiB = lambda b: round(b / GiB_bytes, 2)
+        torch_memory = torch.npu.memory_reserved()
+        free_memory, total_memory = torch.npu.mem_get_info()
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/worker.py
```diff
@@ -260,48 +280,47 @@ def init_device(self):
 
     @torch.inference_mode()
     def determine_available_memory(self) -> int:
-        # Profile the memory usage of the model and get the maximum number of
-        # cache blocks that can be allocated with the remaining free memory.
-        gc.collect()
-        torch.npu.empty_cache()
-        torch.npu.reset_peak_memory_stats()
+        """Profiles the peak memory usage of the model to determine how much
```
gcanlin, line 300: Why not directly use vLLM's `determine_available_memory`?

shen-shanshan: `determine_available_memory` located in `GPUWorker`, but the base class of our `NPUWorker` is `Worker`, which doesn't have  `determine_available_memory` method.

PR #5938, opened 2026-01-15
[0.11.0][Doc] Upgrade outdated ut doc

Potabk:

For cpu env, we should set `SOC_VERSION` to mock different NPU chips for different compilation paths


wangxiyuan:
please push to 0.13.0 branch as well.

PR #5940, opened 2026-01-15
[Feat] Merge the multi eagle graphs to one graph

anon189Ty:
This PR merge all steps of draft model in fullgraph mode, to avoid the synchronize between each graph, reduce the bubble time.

#### Key ideas:
- The "model forward" of the step 0 (first step) and remaining steps are captured together as a "Callable", rather than capturing each model individually.
- "update_attn_params" is moved outside the entire graph, meaning that all "attn_metadata" required by all steps are constructed before "replay", and the "attn_params" of all steps are updated at once.
- Remove synchronization between the main model graph and draft model graph.

#### Key params/functions:
- params: draft_attn_metadatas, attn_metadata_multi_steps, slot_mapping_group
- functions: _run_merged_draft, attn_update_stack_num_spec_norm, update_attn_params, _propose, dummy_run

- vLLM version: v0.13.0


yiz-liu:
Maybe try `set_additional_forward_context` later.

yiz-liu (approved):

PR #5945, opened 2026-01-16
[Bugfix] Fix setting of `speculative_config.enforce_eager` for dsv32

slippersss:
This PR aims to fix setting of `speculative_config.enforce_eager` in deepseek v3.2 mtp. The point is that, vllm sets `speculative_config.enforce_eager` as True if using deepseek_v32 with mtp. Since we support graph mode, we simply ignore it here. However, this fix will also implicitly ignore user setting of `speculative_config.enforce_eager`, we need to take care and remove it once vllm supports this feature.

N/A

by ci

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -412,6 +412,21 @@ class NPUPlatform(Platform):
             os.environ["PYTORCH_NPU_ALLOC_CONF"] = npu_alloc_configs
             logger.info("Set PYTORCH_NPU_ALLOC_CONF=%s", npu_alloc_configs)
 
+        # NOTE: vllm sets `speculative_config.enforce_eager` as True if using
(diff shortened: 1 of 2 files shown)


wangxiyuan (approved):

slippersss:
> Like the comments said, this will replace the flag set by users, which might cause some confusion. I wonder if we could log something, as a reminder, here. And also we need to disclose this change in the docs.

It seems that we have no way to judge whether `speculative_config.enforce_eager = True` is set by user or modified by `speculative_config.__post_init__`, so logging something may be a bit difficult. Perhaps we should log runtime configuration after `platform.check_and_update_config` to remind user, as suggested by @wangxiyuan. As for the docs, let me find out where I should add it later.

PR #5946, opened 2026-01-16
[P/D] Remove restrictions on mooncake for IPv6

liziyu179:
Remove restrictions on mooncake for IPv6
Dependencies: cann8.5、mooncake v0.3.8.post1

- vLLM version: v0.13.0


wangxiyuan:
please upgrade version in all dockerfile and CI as well. For example, https://github.com/vllm-project/vllm-ascend/blob/main/Dockerfile#L21 as well.

PR #5956, opened 2026-01-16
[CI][Lint] Show lint diff on failure

Potabk:
Currently, some of lint checks default automatic code correction but only shows which files were modified (without specifying the changes); in a CI environment, we can make a small optimization to show which lines were modified to give the developers some specifying hint.

- vLLM version: v0.13.0


wangxiyuan (approved):

wangxiyuan:
https://github.com/vllm-project/vllm-ascend/blob/main/format.sh should be updated as well

Potabk:
For the local hooks, this is not work very well I think; The developers may have plenty uncommitted changes, when they run the script `format.sh`, the result can be confusing.

PR #5761, opened 2026-01-09
[bugfix]limit graph replay sync

Jeaniowang:
### What this PR does / why we need it?
when graph mode is picewise，replay by synchronize will be effect performance, sync almost cost 250us

### Does this PR introduce _any_ user-facing change?
only sync when graph mode contain full mode
### How was this patch tested?

- vLLM version: v0.13.0
- vLLM main: https://github.com/vllm-project/vllm/commit/2f4e6548efec402b913ffddc8726230d9311948d

PR #5965, opened 2026-01-16
Revert "[bugfix]limit graph replay sync (#5761)"

Angazenn:
reverts #5761 to fix accuracy issues when using piecewise graph mode.

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/compilation/acl_graph.py
+++ b/vllm_ascend/compilation/acl_graph.py
@@ -186,13 +186,12 @@ class ACLGraphWrapper:
             )
 
         logger.info_once("Replaying aclgraph")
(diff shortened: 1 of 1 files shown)


caozuoba:
@Angazenn @yiz-liu @wangxiyuan Hi , regarding [#5761](https://github.com/vllm-project/vllm-ascend/pull/5761) and the revert in [#5965](https://github.com/vllm-project/vllm-ascend/pull/5965), could you share what exact accuracy issue was observed after removing the PIECEWISE replay sync?

Was it output corruption / gibberish, nondeterminism, or a measurable benchmark accuracy drop? Was the root cause identified?

PR #5971, opened 2026-01-17
[Refactor] Clean up maybe_padded_num_tokens

gcanlin:

After torchair's removal, we don't need maybe_padded_num_tokens anymore. This PR cleans up it and didn't introduce other changes.

NO.

Will be tested by CI.

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -487,10 +487,11 @@ class NPUModelRunner(GPUModelRunner):
         self,
         scheduler_output: "SchedulerOutput",
         intermediate_tensors: Optional[IntermediateTensors] = None,
-    ) -> tuple[dict[str, Any], torch.Tensor, np.ndarray, int, torch.Tensor,
-               int, torch.Tensor, SpecDecodeMetadata, Optional[torch.Tensor],
-               Optional[torch.Tensor], Optional[torch.Tensor], int, dict[str,
-                                                                         Any]]:
+    ) -> tuple[dict[str, Any], torch.Tensor, np.ndarray, int,
+               Optional[torch.Tensor], torch.Tensor,
+               Optional[SpecDecodeMetadata], Optional[torch.Tensor],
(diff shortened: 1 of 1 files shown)


gcanlin:
cc @yiz-liu

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -589,15 +590,10 @@ def _prepare_inputs(
         self.query_lens = torch.from_numpy(num_scheduled_tokens)
 
         # Get info across DP ranks.
-        # NOTE: maybe_padded_num_tokens is only used when using TorchAir with DP,
-        # Otherwise, it's just max_tokens_across_dp_cpu
-        (maybe_padded_num_tokens, num_tokens_across_dp,
+        (num_input_tokens, num_tokens_across_dp,
          with_prefill) = self._sync_metadata_across_dp(num_input_tokens,
```
wangxiyuan, line 594: if `maybe_padded_num_tokens` is useless, why not update `_sync_metadata_across_dp` as well?

gcanlin: IIUC, `_sync_metadata_across_dp` doesn't have additional logic to handle `maybe_padded_num_tokens`. Could you give some guidances for it?

gcanlin: `maybe_padded_num_tokens` is added for `NPUTorchairModelRunner`, which overrides `_sync_metadata_across_dp` and returns `maybe_padded_num_tokens`. But after some torchair related logic had been removed, `maybe_padded_num_tokens` is totally same as `num_input_tokens` for now. So we still need to make `num_input_tokens` to accept the return value of `_sync_metadata_across_dp`.

Reference:
[1] https://github.com/vllm-project/vllm-ascend/pull/2582
[2] https://github.com/yiz-liu/vllm-ascend/blob/005b9af81676cb7b103d51bee3840861d3e70c96/vllm_ascend/torchair/torchair_model_runner.py#L73

Issue #5304, opened 2025-12-24
[RFC]: Env and additionalConfig cleanup


wangxiyuan:
### Motivation.

More and more env and additional config are added to vLLM Ascend. It makes vLLM Ascend hard to be used. The performance by default is not good as well. Users need to open/add complex tag to speed up inference, it's not friendly for users. We should have a plan to cleanup the env and additional config to make vLLM Ascend more clear and friendly for users.

### Proposed Change.

1

yiz-liu:
We plan to deprecate `ProfileExecuteDuration` (see [feature description here](https://docs.vllm.ai/projects/ascend/en/v0.11.0/developer_guide/performance/profile_execute_duration.html)) for several reasons:

1. It introduces redundant environment variables and configuration options.
2. vLLM already offers a more comprehensive feature via `ObservabilityConfig`.
3. Keeping it hinders our efforts to fully align with the upstream community code.
4. Most importantly, our usage data shows that no one is currently using this feature.

xuchi-0808:
PR #15617 removes two sub-keys of `finegrained_tp_config`: `mlp_tensor_parallel_size` and `olora_tensor_parallel_size`. Neither has any in-tree consumer — no e2e config, CI workflow, doc example, or test uses either key, and `FinegrainedTPConfig` has `extra="forbid"`, so any deployment still passing them now fails fast at startup. This executes part of the cleanup tracked in this table. Could you help record it in the `finegrained_tp_config` row's "Related PR" column? Note the config itself survives — `oproj`/`lmhead`/`embedding_tensor_parallel_size` are unaffected — so the row probably shouldn't be struck through entirely.

PR #5976, opened 2026-01-17
[Patch] Remove the patch of ECExampleConnector

gcanlin:

Part of #5304.

https://github.com/vllm-project/vllm/pull/30225 has been merged now. We don't need this patch anymore.

- vLLM version: v0.13.0


wangxiyuan:
https://github.com/vllm-project/vllm/pull/30225 doesn't work with v0.13.0. Let's block this PR until we drop 0.13.0 on main.

PR #5979, opened 2026-01-18
[Bugfix] Add `synced_ag_mode` to limit mixed graph modes in dp ranks

slippersss:
This PR aims to fix different graph modes running in different dp ranks. Since communications between inside graph and outside graph may cause problems, we add `synced_ag_mode` to obtain minimum graph mode across dp ranks and adjust them accordingly.

N/A

by ci

- vLLM version: v0.13.0

### DIFF
--- a/tests/ut/spec_decode/test_mtp_proposer.py
+++ b/tests/ut/spec_decode/test_mtp_proposer.py
@@ -114,7 +114,10 @@ class TestMtpProposer:
         proposer = MtpProposer(vllm_config, torch.device("cpu"), runner)
         proposer.model = MagicMock()
         proposer.enable_shared_expert_dp = False
-        runner._sync_metadata_across_dp.return_value = (8, 8, False)
+        runner._sync_metadata_across_dp.return_value = (8, 8, False, 0)
+        runner.cudagraph_dispatcher.dispatch.return_value = (
+            CUDAGraphMode.NONE, MagicMock()
+        )
 
         mock_get_forward_context = MagicMock()
         mock_get_forward_context.cudagraph_runtime_mode = None
@@ -141,7 +144,10 @@ class TestMtpProposer:
         proposer = MtpProposer(vllm_config, torch.device("cpu"), runner)
         proposer.enable_shared_expert_dp = False
         proposer.model = MagicMock()
-        runner._sync_metadata_across_dp.return_value = (8, 8, False)
+        runner._sync_metadata_across_dp.return_value = (8, 8, False, 0)
+        runner.cudagraph_dispatcher.dispatch.return_value = (
+            CUDAGraphMode.FULL, MagicMock()
+        )
         runner.attn_groups = []
 
         mock_get_forward_context = MagicMock()


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -430,22 +430,23 @@ def needs_mc2(num_tokens: int) -> bool:
...
-            return num_tokens, None, with_prefill
+            return num_tokens, None, with_prefill, ag_mode
 
         if self._skip_all_reduce_across_dp_group():
             num_tokens_after_padding = torch.tensor([num_tokens] *
                                                     self.dp_size,
                                                     device="cpu",
                                                     dtype=torch.int32)
-            return num_tokens, num_tokens_after_padding, with_prefill
+            return num_tokens, num_tokens_after_padding, with_prefill, ag_mode
```
jianzs, line 449: If we skip the `all_reduce` step, it still seems to trigger the issue you're concerned about.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -570,30 +581,29 @@ def _prepare_inputs(
...
-            # In this case, when sequence parallelism is enabled, we need to pad tokens to align
-            # with tp_size because pad_size cannot be captured by the FX graph
+        uniform_decode = (max_num_scheduled_tokens == self.uniform_decode_query_len) \
+            and (total_num_scheduled_tokens == max_num_scheduled_tokens * num_reqs)
+        has_lora = len(self.input_batch.lora_id_to_lora_request) > 0
+        if self.use_aclgraph and enable_sp(self.vllm_config):
             tp_size = self.vllm_config.parallel_config.tensor_parallel_size
-            num_input_tokens = math.ceil(
+            num_tokens_padded = math.ceil(
                 total_num_scheduled_tokens / tp_size) * tp_size
```
yiz-liu, line 590: Corresponding to `_pad_for_sequence_parallelism` except we only do this when `enforce_eager=False` while vLLM does it anyway.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1501,7 +1511,9 @@ def execute_model(
             == self.input_batch.num_reqs * max_query_len)
         has_lora = len(self.input_batch.lora_id_to_lora_request) > 0
         aclgraph_runtime_mode, batch_descriptor = \
-            self.cudagraph_dispatcher.dispatch(num_tokens=num_input_tokens, uniform_decode=uniform_decode, has_lora=has_lora)
+            self.cudagraph_dispatcher.dispatch(num_tokens=num_input_tokens, uniform_decode=uniform_decode, has_lora=has_lora,
+                                               disable_full=synced_ag_mode <= CUDAGraphMode.PIECEWISE.value)
```
yiz-liu, line 1515: The key change.

vllm_ascend/worker/model_runner_v1.py
yiz-liu: The original issue is "A2 + AIV" will hang due to HCCL not supporting graph/eager mode communication. And as for your concern, yes, it would seem that "A2 + AIV + recompute_scheduler + P/D Disaggregate" will still hang. So it's recommended not enable recompute_scheduler, so no skipping, in this case.

slippersss:
Since upcoming model_runner refactoring contains similar fix in this PR, we move this fix in #6011 for v0.13.0.

PR #5980, opened 2026-01-18
[Tests] move qwen3 performance test from nightly to e2e

wxsIcey:
Move the qwen3 performance test from nightly to e2e to intercept performance degradation.

- vLLM version: v0.13.0


Yikun:
+1, I think it's Ok to have a 201.89s test to avoid potential performance regression. This happens in almost every version.

wxsIcey:
The previous nightly test failed because I was using a3 for local testing and a2 for nightly testing, resulting in inconsistent performance results. Therefore, I moved the test to a3.

wangxiyuan (approved):

PR #6011, opened 2026-01-19
[0.13.0][Bugfix] Add `synced_cudagraph_mode` to limit mixed graph modes in dp ranks

slippersss:
This PR aims to fix the issue that using A2 + AIV will hang due to the fact that HCCL does not support eager/graph mode communication. To handle it, following https://github.com/vllm-project/vllm/pull/30173, we introduce `synced_cudagraph_mode` to enable all ranks to know the minimum mode across ranks. Main changes are described below:
1. `execute_model` now performs "dispatch -> sync -> re-dispatch" just as `_dummy_run`
2. `_sync_metadata_across_dp` now receives `cudagraph_mode` from all ranks and returns `synced_cudagraph_mode` to all ranks
3. Re-dispatch steps in both `execute_model` and `_dummy_run` include `disable_full=synced_cudagraph_mode <= CUDAGraphMode.PIECEWISE.value` so that when it is true, no FULL will be dispatched

N/A

by ci

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -426,9 +426,11 @@ class NPUModelRunner(GPUModelRunner):
             or self.ascend_config.recompute_scheduler_enable)
 
     def _sync_metadata_across_dp(
-            self, num_tokens: int,
-            with_prefill: bool,
-            cudagraph_mode: int = 0) -> tuple[int, Optional[torch.Tensor], bool, int]:
+        self,
+        num_tokens: int,
+        with_prefill: bool,
+        cudagraph_mode: int = 0,
+    ) -> tuple[int, Optional[torch.Tensor], bool, int]:
         # TODO: In vLLM, the only thing that needs to be synced is num_tokens, but in
         # our case, we still need to sync the other two flags as well. So we need to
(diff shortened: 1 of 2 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -426,23 +426,26 @@ def needs_mc2(num_tokens: int) -> bool:
...
-            return num_tokens, None, with_prefill
+            return num_tokens, None, with_prefill, cudagraph_mode
 
         if self._skip_all_reduce_across_dp_group():
             num_tokens_after_padding = torch.tensor([num_tokens] *
                                                     self.dp_size,
                                                     device="cpu",
                                                     dtype=torch.int32)
-            return num_tokens, num_tokens_after_padding, with_prefill
+            return num_tokens, num_tokens_after_padding, with_prefill, cudagraph_mode
```
slippersss, line 448: @jianzs Hi, as what you mentioned in #5979, when we skip `all_reduce`, there is still a possibility that different dp ranks may run different graph modes. Since we have met the issue that A2 + AIV will hang, so we have to ensure that there is no prefill when entering this `_skip_all_reduce_across_dp_group` branch.

PR #6013, opened 2026-01-19
[Feat] Remove CP Redundant Variables after FIA operator enables for CANN 8.5 

dsxsteven:
### What this PR does / why we need it?
PCP/DCP splits the kv-cache onto different cards. After introducing the parameter cp-kv-cache-interleave-size, the first size tokens will be cached at Card 0, and so on.
However, if there are too few tokens, some cards will not store the key-value pairs, resulting in values ​​of 0, corrupted values, and precision issues. Currently, additional operations are introduced to avoid this precision problem.

After we integrate FIA operator in mla_cp._forward_decode and CANN updates to 8.5.0, we now can remove these additional operations.
### How was this patch tested?
passed all CI by CANN 8.5.0
- vLLM version: v0.13.0
- vLLM main: https://github.com/vllm-project/vllm/commit/2c24bc6996cb165fce92f780b388a5e39b3f4060

PR #6039, opened 2026-01-20
[0.13.0][cherry-pick][CP&SP] Remove CP Redundant Variables after FIA operator enables for CANN 8.5 

dsxsteven:

PCP/DCP splits the kv-cache onto different cards. After introducing the parameter cp-kv-cache-interleave-size, the first size tokens will be cached at Card 0, and so on.
However, if there are too few tokens, some cards will not store the key-value pairs, resulting in values ​​of 0, corrupted values, and precision issues. Currently, additional operations are introduced to avoid this precision problem.

After we integrate FIA operator in mla_cp._forward_decode and CANN updates to 8.5.0, we now can remove these additional operations.
pick-from: https://github.com/vllm-project/vllm-ascend/pull/6013

passed all CI by CANN 8.5.0


dsxsteven:
This PR should be merged after https://github.com/vllm-project/vllm-ascend/pull/6046

PR #6089, opened 2026-01-21
Remove the default dual-stream implementation of random sampling due to potential precision issues, and enable enable_async_exponential by default.

coder-fny:
Here, there are two parallel implementations for computing exponentially distributed random numbers in random sampling; however, by default, the original random sampling suffers from precision issues in scenarios with high concurrency at the eager mode.

- vLLM version: v0.13.0


wangxiyuan:

PR #6106, opened 2026-01-21
[kv_cache] support multi_block_pool

LookAround0301:

[Long Seq Feature] Multi-Block-Pool can manage block-pools with different numbers of blocks to implement more fine-grained block allocation.
Multi-Block-Pool is supported to alleviate the issue of mixed long and short sequences in the Dynamic-Context-Parallel scenario and the issue of block resource waste caused by inconsistent page sizes at different layers in the hybrid spec scenario.

### DIFF
--- a/vllm_ascend/patch/platform/__init__.py
+++ b/vllm_ascend/patch/platform/__init__.py
@@ -25,6 +25,11 @@ import vllm_ascend.patch.platform.patch_set_cudagraph_sizes  # noqa
 from vllm_ascend import envs
 from vllm_ascend.utils import vllm_version_is
 
+USE_MULTI_BLOCK_POOL = False
+
+if USE_MULTI_BLOCK_POOL:
+    import vllm_ascend.patch.platform.patch_kv_cache_coordinator  # noqa
+
 if os.getenv("DYNAMIC_EPLB", "false").lower() in ("true", "1") or os.getenv(
         "EXPERT_MAP_RECORD", "false") == "true":
     import vllm_ascend.patch.platform.patch_multiproc_executor  # noqa


vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
```diff
@@ -0,0 +1,142 @@
...
+    max_model_len: int,
+    use_eagle: bool,
+    enable_caching: bool,
+    enable_kv_cache_events: bool,
+    dcp_world_size: int,
+    pcp_world_size: int,
+    hash_block_size: int,
+    metrics_collector: KVCacheMetricsCollector | None = None,
+) -> KVCacheCoordinator:
+    if USE_MULTI_BLOCK_POOL:
```
MengqingCao, line 94: This will be enabled after dynamic cp is supported

vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
```diff
@@ -0,0 +1,142 @@
...
+                                               UnitaryKVCacheCoordinator)
+from vllm.v1.core.kv_cache_metrics import KVCacheMetricsCollector
+from vllm.v1.core.kv_cache_utils import BlockHash, KVCacheBlock
+from vllm.v1.core.single_type_kv_cache_manager import \
+    get_manager_for_kv_cache_spec
+from vllm.v1.kv_cache_interface import KVCacheConfig
+
+from vllm_ascend.core.multi_block_pool import MultiBlockPool
+
+USE_MULTI_BLOCK_POOL = False
```
MengqingCao, line 17: let's move this to `__init__.py`, and only patch when this is enabled

PR #6113, opened 2026-01-22
[Bugfix]KV pool rank 0 consumes more HBM

baxingpiaochong:

before add_set_deivce
image
after
image

- vLLM version: v0.13.0


LCAIZJ:
All other GPUs default to set device 0, causing processes from other GPUs to run on GPU 0, which further results in GPU 0 consuming more HBM than the other GPUs.

LCAIZJ (approved):

Pz1116:
Thanks for the contribution, in the long term I think we should have a discussion with the torch_npu team about the default behavior of set_device.

PR #6138, opened 2026-01-22
[Bugfix] Fix the issue of the acceptance rate decline for Qwen3-30B-A3B-EAGLE3

zhaomingyu13:
Due to the long-term lack of synchronization with the upstream code, a problem that led to a decrease in the acceptance rate of the Qwen3-30B-A3B-EAGLE3 draft model was introduced when fixing the bug（#5967）. Now, synchronize with the upstream and fix this bug
No
```python

def main():
    prompts = [
        "The future of AI is",
    ]

    sampling_params = SamplingParams(temperature=0.8, top_p=0.95)
    # Create an LLM.
    llm = LLM(
            model="Qwen/Qwen3-30B-A3B",
            tensor_parallel_size=4,
            gpu_memory_utilization=0.9,
            enforce_eager=True,
            speculative_config={
                "method": "eagle3",
                "model": "AngelSlim/Qwen3-a3B_eagle3"
                "num_speculative_tokens": 3,
            },
        )

    print(f"Outputs: {outputs}")
    for output in outputs:
        prompt = output.prompt
```
- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/spec_decode/eagle_proposer.py
+++ b/vllm_ascend/spec_decode/eagle_proposer.py
@@ -200,19 +200,49 @@ class EagleProposer(VllmEagleProposer):
                 )
             # If pp>1, the weights of mtp and the main model's embedding are not on the same device.
             # check if mtp model use main model's embedding and LMhead
-            if hasattr(model, "model") and hasattr(model.model, "embed_tokens") and \
-                    torch.equal(self.model.model.embed_tokens.weight,
-                                model.model.embed_tokens.weight):
-                logger.info(
-                    "The EAGLE head shares the same vocab embedding" \
-                    " with the target model."
-                )
-                self.model.model.embed_tokens = target_embed_tokens
+            share_embeddings = False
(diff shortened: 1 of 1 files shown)


vllm_ascend/spec_decode/eagle_proposer.py
```diff
@@ -200,19 +200,49 @@ def load_model(self, model: nn.Module) -> None:
...
+                        " model."
+                    )
+                else:
+                    logger.info(
+                        "Detected EAGLE model with distinct embed_tokens weights. "
+                        "Keeping separate embedding weights from the target model."
+                    )
             else:
+                # MTP model
+                share_embeddings = True
```
drslark, line 236: Please follow the original logic.

drslark: Oops, it's my mistake.

drslark: LGTM.

PR #6141, opened 2026-01-22
[Refactor] profiler config optimze

mengchengTang:
This PR optimizes the torch_npu profiler configuration to significantly reduce overhead and trace file size. The key changes include:
**Enable Data Simplification**: Explicitly sets data_simplification=True in _ExperimentalConfig. This filters out unnecessary intermediate data during profiling, drastically reducing the memory footprint and I/O overhead.
**Use Lightweight Stack Tracing**: Replaces with_stack with with_modules when torch_profiler_with_stack is enabled. In torch_npu, with_stack introduces heavy latency. with_modules provides equivalent semantic information with much lower overhead.
**Code Simplification:** Removes redundant parameter configurations in _ExperimentalConfig by utilizing default values, making the codebase cleaner and easier to maintain.

**Test setup:**
 max length = 50, profiler + stack enabled

**Before optimization:**
Profiler data size: 651 MB
Generate time: 3 seconds

**After optimization:**
Profiler data size: 156 MB (≈76% reduction)
Generate time: <1 second

No API changes. Users profiling on Ascend will experience faster profiling execution and smaller trace files when stack tracing is enabled.
Manually verified on Ascend NPU by running vLLM with the profiler enabled. Confirmed that trace files are generated correctly containing necessary stack/module info, while showing the reported reduction in size and time.

- vLLM version: v0.13.0

### DIFF
--- a/vllm_ascend/worker/worker.py
+++ b/vllm_ascend/worker/worker.py
@@ -561,7 +561,7 @@ class NPUWorker(WorkerBase):
                 aic_metrics=torch_npu.profiler.AiCMetrics.AiCoreNone,
                 l2_cache=False,
                 op_attr=False,
-                data_simplification=False,
+                data_simplification=True,
                 record_op_args=False,
                 gc_detect_threshold=None,
             )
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/worker.py
```diff
@@ -504,23 +504,20 @@ def _init_profiler(self):
...
                 profiler_level=torch_npu.profiler.ProfilerLevel.Level1,
-                msprof_tx=False,
                 aic_metrics=torch_npu.profiler.AiCMetrics.AiCoreNone,
-                l2_cache=False,
-                op_attr=False,
-                data_simplification=False,
-                record_op_args=False,
-                gc_detect_threshold=None,
+                data_simplification=True,
             )
```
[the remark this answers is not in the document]
mengchengTang: The default values for these parameters are stable and will not change. We prefer to keep the code concise by relying on these defaults."

vllm_ascend/worker/worker.py
```diff
@@ -504,23 +504,20 @@ def _init_profiler(self):
             experimental_config = torch_npu.profiler._ExperimentalConfig(
                 export_type=torch_npu.profiler.ExportType.Text,
                 profiler_level=torch_npu.profiler.ProfilerLevel.Level1,
-                msprof_tx=False,
                 aic_metrics=torch_npu.profiler.AiCMetrics.AiCoreNone,
-                l2_cache=False,
```
whx-sjtu, line 562: I think it's better to keep these default values in case user wants to enable one of these parameters but don't known concrete param name.

mengchengTang: Thanks for the suggestion. I agree that explicitly listing these parameters makes it easier for users to discover and enable them if needed. I have restored the parameters with their default values.

PR #6159, opened 2026-01-22
[v0.13.0][bugfix] fix capture shape in sp_eagle_fullgraph

lilinsiman:
fix capture shape error in sp+eagle+fullgraph feature

Overall idea: Implement a solution from the vLLM community to handle the capture shape error when sp, fullgraph, and eagle are enabled simultaneously.

Implementation: Add a context manager to the modelrunner file to pass the value of enable_sp to the vLLM function _check_and_update_cudagraph_mode, so that it can detect whether sp is enabled, thereby achieving the reuse of the vLLM community function. This vllm function now includes checks for tp and num_spec, and then raises an error to resolve this issue.

no

ut and tests

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -359,9 +359,6 @@ class NPUModelRunner(GPUModelRunner):
         self.reorder_batch_threshold: int | None = None
         self.long_seq_metadata = None
 
-        # TODO: remove it when flash_common1 is removed
-        self.sp_context = _set_sp_context(self)
-
     def _init_device_properties(self) -> None:
         self.num_sms = None
 
@@ -3019,9 +3016,9 @@ class NPUModelRunner(GPUModelRunner):
         attention_backends: list[set[type[AttentionBackend]]],
         kv_cache_groups: list[KVCacheGroupSpec],
     ) -> None:
-        with self.sp_context:
+        with update_pass_config(self):
             super()._check_and_update_cudagraph_mode(attention_backends,
-                                                    kv_cache_groups)
+                                                     kv_cache_groups)
 
         # NOTE: Since aclgraph_batch_sizes cannot be determined until here,
         # we set the graph params right before initializing the keys.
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3120,3 +3124,13 @@ def _replace_gpu_model_runner_function_wrapper(target_module_name):
...
+
+# TODO: remove it when flash_common1 is removed
+@contextmanager
+def _set_sp_context(model_runner):
+    try:
+        original_pass_config_sp = model_runner.compilation_config.pass_config.enable_sp
+        model_runner.compilation_config.pass_config.enable_sp = enable_sp(model_runner.vllm_config)
+        yield
+    finally:
+        model_runner.compilation_config.pass_config.enable_sp = original_pass_config_sp
```
yiz-liu, line 3135: Do we have any ideas when can we support the native `pass_config.enable_sp`? @wxsIcey @realliujiaxu

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -359,6 +359,9 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
         self.reorder_batch_threshold: int | None = None
         self.long_seq_metadata = None
 
+        # TODO: remove it when flash_common1 is removed
+        self.sp_context = _set_sp_context(self)
+
```
yiz-liu: This is unnecessary while inappopriate at the same time.

This object cannot be called twice, since you have `yield` in it, and we sure don't use it anywhere else. So I don't see why we need to init it as a `self.sp_context`.

And we can also rename `_set_sp_context` to `update_pass_config` in case there are other temporary updates need to be done in the future.

Please change it to
```python
        with update_pass_config(self):
            super()._check_and_update_cudagraph_mode(attention_backends,
                                                    kv_cache_groups)
```

yiz-liu:
Please update PR title and description to address why this change can fix the original issue.

PR #6178, opened 2026-01-23
[0.13.0][Feat] Merge the multi eagle graphs to one graph

slippersss:
This PR is mainly cherry-picked fron #5940.

Major difference from #5940 is:
1. remove `piece_all_attn_layer_name`, which is never used in #5940
2. remain `attn_metadata` update process in `attn_update_stack_num_spec_norm` instead of `common_attn_metadata` update and rebuild `attn_metadata` process in #5940

Minor change in this PR is:
1. `self.cudagraph_batch_sizes` is replaced with `self.runner.cudagraph_batch_sizes`, since the former one is not correct (self.cudagraph_batch_sizes[-1] is not guarantee to be valid)

N/A

by ci


yiz-liu:
> 1. remove `piece_all_attn_layer_name`, which is never used in [[Feat] Merge the multi eagle graphs to one graph #5940](https://github.com/vllm-project/vllm-ascend/pull/5940)

@anon189Ty Please take a look at this.

PR #6182, opened 2026-01-23
BugFix:  Fix moe_load accumulation error in ACL graph mode

Mercykid-bash:
This PR fixes the numerical error in moe_load accumulation under ACL graph mode on NPU: using += for NPU tensors in graph mode does not throw errors but leads to incorrect values, so we replace it with the in-place add_() method to ensure accurate calculation.


yiz-liu:
I don't quite get it, what will happen in graph mode? You mean there may be error being silenced? What error?

wangxiyuan:
I still want to get the root reason for this change.

PR #6193, opened 2026-01-23
[Worker] Implement update max_model_len interface for NPUWorker

Potabk:
This patch purpose to add the `update_max_model_len` interface.

- vLLM version: v0.14.0


vllm_ascend/worker/worker.py
```diff
@@ -426,6 +426,19 @@ def get_kv_connector_handshake_metadata(self) -> Optional[dict]:
     def get_kv_cache_spec(self) -> dict[str, KVCacheSpec]:
         return self.model_runner.get_kv_cache_spec()
 
+    def update_max_model_len(self, max_model_len: int) -> None:
```
jianzs, line 429: Maybe we could put this function in `WorkerBase` instead of adding it here?

Potabk: According to the doc string of the class `WorkerBase`: `Worker interface that allows vLLM to cleanly separate implementations for
    different hardware. Also abstracts control plane communication, e.g., to
    communicate request metadata to other workers.`,  it is a abstract control plane, we may add an abstract method interface to the workerbase, but the concrete implementation should be in the specific worker.

PR #6203, opened 2026-01-23
[Feature]mooncke_connector add handle tranfer failed

baxingpiaochong:
Based on the VLLM logic, an adaptation was added to address the issue of failed KV transmission to Mooncake.

- vLLM version: v0.14.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/kv_transfer.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/kv_transfer.py
@@ -257,21 +257,6 @@ class KVCacheStoreLayerSendingThread(KVTransferThread):
                          name="KVCacheStoreLayerSendingThread")
         self.final_layer_id = num_layers - 1
         self.put_step = put_step
-        self.stored_requests = defaultdict[str, int](int)
-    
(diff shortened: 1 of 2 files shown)


Pz1116:
How does this PR handle the case where TP is enabled but each rank has different invalid block id (as we need to find the minimum valid prefix across all TP ranks)? I don't see any "cross-rank aggregation" logic in `_update_requests_with_invalid_blocks` in vllm scheduler, so I assume we are expected to do this in the KV  connectors?

baxingpiaochong:
> How does this PR handle the case where TP is enabled but each rank has different invalid block id (as we need to find the minimum valid prefix across all TP ranks)? I don't see any "cross-rank aggregation" logic in `_update_requests_with_invalid_blocks` in vllm scheduler, so I assume we are expected to do this in the KV connectors?

In p2p transfers, since the transfer task is distributed all at once via a batch interface, I'm unsure if the returned ret value includes information on failed blocks. This modification assumes all blocks fail by default.

In kvpool transfers, each transfer (TP) simply returns the ID of its failed block. A union operation is then performed, and the result of the union is the maximum number of failed blocks.

SlightwindSec:

wangxiyuan:

PR #6229, opened 2026-01-25
[Bugfix]The service fails to be started when the memcache pool is enabled

luoxiaolin712:
The service fails to be started when the memcache pool is enabled without configuring the mooncake path.

NA

```
#memcache
echo 200000 > /proc/sys/vm/nr_hugepages
source /usr/local/memfabric_hybrid/set_env.sh
source /usr/local/memcache_hybrid/set_env.sh
export MMC_LOCAL_CONFIG_PATH=/usr/local/memcache_hybrid/latest/config/mmc-local.conf

vllm serve /mnt/weight/DeepSeek-V3.2-Exp-W8A8 \
  --host $local_ip \
  --port 8002 \
  --served-model-name model \
  --data-parallel-size 2 \
  --tensor-parallel-size 8 \
  --enable-expert-parallel \
  --no-enable-prefix-caching \
  --no-enable-chunked-prefill \
  --max-num-seqs 4 \
  --max-model-len 8192 \
  --trust-remote-code \
  --enforce-eager \
  --quantization ascend \
  --additional_config '{"ascend_scheduler_config":{"enabled":false}}' \
  --kv-transfer-config \
    '{
            "kv_connector": "AscendStoreConnector",
            "kv_role": "kv_both",
	            "backend": "memcache",
                "lookup_rpc_port":"0"
            }
    }'
```

- vLLM version: v0.14.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
@@ -130,8 +130,10 @@ class KVPoolWorker:
         self.token_database = ChunkedTokenDatabase(self.metadata, self.block_size, self.use_mla, partitions)
 
         backend = backend_map.get(self.backend.lower())
+        assert backend is not None
         backend_path = backend.get("path")
         backend_name = backend.get("name")
+        assert backend_path is not None and backend_name is not None
         backend_module = importlib.import_module(backend_path)
         real_backend = getattr(backend_module, backend_name)
 


luoxiaolin712:
> I think it's better to move `from mooncake.store import ReplicateConfig` into `__init__` of MooncakeBackend, right under `from mooncake.store import MooncakeDistributedStore `, so if any exception is caught, we can handle it as an Import Error.
> 
> Right now if an exception is caught the log would say "failed to put key", which is confusing.

If move `from mooncake.store import ReplicateConfig` into `__init__` of MooncakeBackend，put function can't use ReplicateConfig. Add a log for `from mooncake.store import ReplicateConfig` ?

Pz1116:
> If move `from mooncake.store import ReplicateConfig` into `__init__` of MooncakeBackend，put function can't use ReplicateConfig. Add a log for `from mooncake.store import ReplicateConfig` ?

Can we instead define a new `self.replicate_config` in `__init__' so that it can be used by put function?

luoxiaolin712:
> Can we instead define a new `self.replicate_config` in `**init**' so that it can be used by put function?

Thanks, I also think this is better.

luoxiaolin712:
The pool_worker.py file is modified to solve the problem at the root cause.

luoxiaolin712:
> The pool_worker.py file is modified to solve the problem at the root cause.

Modified

PR #6234, opened 2026-01-25
[Feature] Support to handle shared-expert wrappers in SharedFusedMoE

gcanlin:

- vLLM version: v0.14.1

### DIFF
--- a/vllm_ascend/ops/fused_moe/fused_moe.py
+++ b/vllm_ascend/ops/fused_moe/fused_moe.py
@@ -448,7 +448,7 @@ class AscendSharedFusedMoE(SharedFusedMoE, AscendFusedMoE):
         shared_act = shared_expert.act_fn(shared_gate_up)  # type: ignore
         shared_out, _ = shared_expert.down_proj(shared_act)  # type: ignore
 
-        # Shared expert gating mechanism (e.g., Qwen2/3-Next, Qwen3-Omni).
+        # Shared expert gating mechanism (e.g., Qwen3-Next, Qwen3-Omni).
         expert_gate = self._get_shared_expert_gate()
         if expert_gate is not None:
             gate_out = expert_gate(hidden_states)


vllm_ascend/ops/fused_moe/fused_moe.py
wangxiyuan, line 412: wht not init share_experts here?

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -437,24 +437,49 @@ def wrapped_process_weights(*args, **kwargs):
...
             shared_out = F.sigmoid(gate_out) * shared_out
         return shared_out
 
+    def _get_shared_expert_mlp(self) -> torch.nn.Module:
+        if self._shared_experts is None:
+            raise AttributeError("Shared experts are not set.")
+        if hasattr(self._shared_experts, "gate_up_proj"):
+            return self._shared_experts
+        if hasattr(self._shared_experts, "_shared_expert"):
+            return self._shared_experts._shared_expert  # type: ignore[attr-defined]
```
jianzs, line 466: Please add comments to explain which model will trigger this case.

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -437,24 +437,49 @@ def wrapped_process_weights(*args, **kwargs):
...
+            "_shared_expert with those attributes. Got: "
+            f"{type(self._shared_experts).__name__}")
+
+    def _get_shared_expert_gate(self):
+        if self._shared_experts is None:
+            return None
+        if hasattr(self._shared_experts, "expert_gate"):
+            expert_gate = self._shared_experts.expert_gate
+            if expert_gate is not None:
+                return expert_gate
```
jianzs, line 478: Returning expert_gate directly has no side effects.

vllm_ascend/ops/fused_moe/fused_moe.py
jianzs: Good idea.

gcanlin:
@jianzs Hey, thanks for reviewing! This PR was for Qwen3-Omni in vLLM-Omni. I finally decided to make downstream to expose the interfaces in the wrapper. So we would close this PR.

PR #6284, opened 2026-01-26
[0.13.0][Bugfix] Fix FIA operator validation error in Eagle scenario with CANN 8.5

SlightwindSec:

Fixes FIA operator validation error in Eagle speculative decoding after CANN 8.5 upgrade. The new validation requires `queryT` to equal the last element of `actualSequenceLengthQ` when using TND layout, which fails in Eagle scenarios.

**Changes**:
- Added `_pad_attention_seq_params` function to pad `actual_seq_lengths_q` and `seq_lens` so the last element of `actual_seq_lengths_q` matches `runtime_shape` (queryT)
- Updated `_update_attn_fia_params` to accept `vllm_config` parameter and apply padding for draft model attention metadata
- Adjusted `uniform_decode` calculation logic in Eagle proposer
- Fixed `query_start_loc` calculation to handle CUDA graph runtime mode correctly
- Re-enabled previously skipped tests

No.

## How was this patch tested?

Re-enabled previously skipped tests that were failing with CANN 8.5.

**Co-authored-by: @yiz-liu**

### DIFF
--- a/vllm_ascend/compilation/acl_graph.py
+++ b/vllm_ascend/compilation/acl_graph.py
@@ -14,7 +14,7 @@ import vllm.envs as envs
 from vllm.compilation.counter import compilation_counter
 from vllm.compilation.cuda_graph import CUDAGraphOptions
 from vllm.compilation.monitor import validate_cudagraph_capturing_enabled
-from vllm.config import CUDAGraphMode, VllmConfig, get_current_vllm_config
+from vllm.config import CUDAGraphMode, VllmConfig
 from vllm.forward_context import BatchDescriptor, get_forward_context
 from vllm.logger import logger
 from vllm.platforms import current_platform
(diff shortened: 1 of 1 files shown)


vllm_ascend/compilation/acl_graph.py
```diff
@@ -260,10 +260,43 @@ def _update_attn_pa_params(update_stream, forward_context, runtime_shape):
...
+                interpolated = [runtime_shape]
+            padded_actual_seq_lengths_q = actual_seq_lengths_q + interpolated
+
+    target_len = len(padded_actual_seq_lengths_q)
+    if len(seq_lens) >= target_len:
+        padded_seq_lens = seq_lens
+    else:
+        padded_seq_lens = seq_lens + [0] * (target_len - len(seq_lens))
+
+    return padded_actual_seq_lengths_q, padded_seq_lens
```
yiz-liu, line 292: Such a pain in the ass, I hate to admit it but it seems now we have no other choices since the padded `query_start_loc` from runner got `unpadded` mid-way.

yiz-liu: This is hacky as hell, but it must do for now, hope we can solve this later.

yiz-liu (approved):

PR #6291, opened 2026-01-27
[Misc] Drop patch_rope.py

gcanlin:

Part of #5304.

We have align with vLLM's latest change for `RotaryEmbeddingBase`. Don't need this patch anymore.

- vLLM version: v0.14.1


gcanlin:
cc @wangxiyuan

PR #6298, opened 2026-01-27
[Misc] Print triton info in collect_env.py

menogrey:
[Misc] Print triton info in collect_env.py, help us to collect more info when user create an issue.

- vLLM version: v0.14.1


collect_env.py
```diff
@@ -67,6 +67,7 @@
     "transformers",
     "zmq",
     "pynvml",
+    "triton",
```
wangxiyuan: should it be triton-ascend?

menogrey: done

PR #6326, opened 2026-01-27
[Refactor][npugraph_ex]disable npigraph_ex when peicewise mode

ChenCangtao:
When using piecewise graph mode, there are still issues when using the npugraph_ex backend with the static kernel enabled.
So when the graph mode is a piecewise graph, we set the enable configuration of npugraph_ex to false.
Previously, we handled this logic in check_and_update_config() in platform.py, but the ascend_config might be reinitialized during worker initialization, so in this PR we are migrating the disable logic.
At the same time, we also added UT test cases for this scenario.

- vLLM version: v0.14.1

### DIFF
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -280,9 +280,7 @@ class NpugraphExConfig:
 
         # npugraph only support FULL_DECODE_ONLY or FULL graph mode now
         compilation_config = vllm_config.compilation_config
-        if compilation_config.cudagraph_mode in (CUDAGraphMode.FULL_DECODE_ONLY, CUDAGraphMode.FULL):
-            self.enable = self.enable and True
-        else:
+        if compilation_config.cudagraph_mode not in (CUDAGraphMode.FULL_DECODE_ONLY, CUDAGraphMode.FULL):
             self.enable = False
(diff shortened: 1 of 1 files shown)


vllm_ascend/ascend_config.py
```diff
@@ -268,12 +270,21 @@ def __init__(
...
         self.enable = enable
         self.enable_static_kernel = enable_static_kernel
         self.fuse_norm_quant = fuse_norm_quant
         self.fuse_qknorm_rope = fuse_qknorm_rope
         self.fuse_allreduce_rms = fuse_allreduce_rms
 
+        # npugraph only support FULL_DECODE_ONLY or FULL graph mode now
+        compilation_config = vllm_config.compilation_config
+        if compilation_config.cudagraph_mode in (CUDAGraphMode.FULL_DECODE_ONLY, CUDAGraphMode.FULL):
+            self.enable = self.enable and True
```
wangxiyuan: `and True` is useless

ChenCangtao: I have already modified the code here, and the current code is the same as the changes suggested by the gemini-code-assist.

wangxiyuan:

PR #6350, opened 2026-01-28
Fix the prefetch weight invalidation issue

ZYang6263:
Fix the prefetch weight invalidation issue

- vLLM version: v0.14.1

### DIFF
--- a/vllm_ascend/ops/linear_op.py
+++ b/vllm_ascend/ops/linear_op.py
@@ -234,7 +234,6 @@ class OProjRowParallelOp(CustomRowParallelOp):
                 input_, num_partitions=self.tp_size)
             input_parallel = splitted_input[self.tp_rank].contiguous()
 
-        # Prepare tensors for all-to-all communication
         local_batch_size = input_parallel.size(0)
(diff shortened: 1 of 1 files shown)


juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **1 point that might be worth attention** (could be false positives, please use your judgment):

1. <strong>Potential schema mismatch for custom operator return types</strong>
   - [Location](https://github.com/vllm-project/vllm-ascend/blob/157308400bfdba2d5982093b635b34c52343600b/vllm_ascend/ops/register_custom_ops.py#L113-L114): `vllm_ascend/ops/register_custom_ops.py:L113-L114`
   - Impact: If the registered schema still specifies a `void` return, it may cause runtime `AttributeError` or `TypeError` when the calling code expects a `Tensor`.
   - Suggestion: Consider updating the operator schema definitions (e.g., in `torch.library.define`) to reflect the new `Tensor` return type.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

PR #6355, opened 2026-01-28
[CI] Nightly tests use `releases/v0.13.0`

Potabk:
The pre-requirement pr is https://github.com/vllm-project/vllm-ascend/pull/6353, this patch aims to transfer nightly tests to `releases/v0.13.0`, what we need to do is just use the branch built image for nightly

- vLLM version: v0.14.1

### DIFF
--- a/.github/Dockerfile.nightly.a3
+++ b/.github/Dockerfile.nightly.a3
@@ -15,7 +15,7 @@
 # This file is a part of the vllm-ascend project.
 #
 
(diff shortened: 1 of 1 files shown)


Yikun:
Before merge this, we need to think the potential impact carefully.

Potabk:
The workflow we are using is main, and the test case folder is for version v0.13.0, which causes an issue: a test that exists in the main branch but doesn't have a corresponding test case in v0.13.0 will fail.

PR #2947, opened 2025-09-16
[WIP][Feature] mooncake connector support QGA transport

lidenghui1110:
### What this PR does / why we need it?

TODO:

PR #6366, opened 2026-01-28
[Kernel] Add AscendC fused op transpose_kv_cache_by_block to speed up GQA transfer

lidenghui1110:
As #2947 describe, we need to transpose kv cache layout after GQA kv transfer when prefill and decode tensor parallel size are heterogeneous, in the previous implementation, we use `npu_paged_cache_load ` + `tranpose` + `_npu_reshape_and_cache` to do this work.

But obviously, it is not an efficient plan, the ops above need to be called for each layer, which introduces 3 * layer_num kernel launch, and 6 * layer_num data movement between L1 Cache and HBM for one request on decode node. Usually, decode node uses graph mode, so these op kernels will be called between decode forward launched by an async thread in mooncacke connector, this kernels maybe last for several decode forward and TTFT will increase by 3~4 decode forward time.

 In this PR, we implement an AscendC fused op `transpose_kv_cache_by_block` to do this with only once kernel launch and move data between L1 Cache and HBM only once.

After using this fused op, the time cost in transpose kv cacke layout  can  be decreased to 0.24ms from 7ms in UT on 910C, and in PD disaggregation scenario, TTFT can decrease about 90 ~ 110 ms in qwen3-235B.

| request_num | original | fused_op|
|:----------------------:|:---------------:|:-------------------:|
|           1            |      643 ms      |        578 ms        |
|          128           |     1480 ms      |       1368 ms        |

Use fused op by default, incase the op has bug in any scenario, provide fallback choice using env to disable it.

**DISABLE fused op by add following env**
`export VLLM_ASCEND_FUSION_OP_TRANSPOSE_KV_CACHE_BY_BLOCK=0`

- vLLM version: v0.14.1

### DIFF
--- a/csrc/transpose_kv_cache_by_block/op_kernel/common.h
+++ b/csrc/transpose_kv_cache_by_block/op_kernel/common.h
@@ -7,7 +7,8 @@ using namespace AscendC;
 template <typename T>
 __aicore__ inline __gm__ T* GetTensorAddr(uint16_t index, GM_ADDR tensorPtr) {
     __gm__ uint64_t* dataAddr = reinterpret_cast<__gm__ uint64_t*>(tensorPtr);
-    uint64_t tensorPtrOffset = *dataAddr;  // The offset of the data address from the first address.
+    // The offset of the data address from the first address.
+    uint64_t tensorPtrOffset = *dataAddr;
     // Moving 3 bits to the right means dividing by sizeof(uint64 t).
     __gm__ uint64_t* retPtr = dataAddr + (tensorPtrOffset >> 3);
     return reinterpret_cast<__gm__ T*>(*(retPtr + index));


tests/e2e/nightly/single_node/ops/singlecard_ops/test_transpose_kv_cache_by_block.py
```diff
@@ -0,0 +1,141 @@
...
+                relative_diff,
+                rtol,
+                f"Relative error too large: {relative_diff} > {rtol}. Max difference: {max_diff}",
+            )
+
+        self.assertLessEqual(max_diff, atol,
+                             f"Absolute error too large: {max_diff} > {atol}")
+
+
+if __name__ == "__main__":
```
wangxiyuan: remove these two lines

vllm_ascend/ascend_config.py
```diff
@@ -214,17 +214,23 @@ class AscendFusionConfig:
...
             fusion_ops_gmmswigluquant (bool): Whether to use a fused operator gmmswigluquant.
                 When set to True, the system will use a fused operator gmmswigluquant.
                 Default: True
+            fusion_ops_transpose_kvcache_by_block (bool): Whether to use a fused operator transpose_kvcache_by_block.
+                When set to True, the system will use a fused operator transpose_kvcache_by_block.
+                Default: True
             **kwargs: Additional optional parameters for forward compatibility and configuration extension.
         """
         self.fusion_ops_gmmswigluquant = fusion_ops_gmmswigluquant
+        self.fusion_ops_transpose_kvcache_by_block = fusion_ops_transpose_kvcache_by_block
```
wangxiyuan: move it to additional config or use ENV instead. `AscendCompilationConfig` is used for  compilation feature only.

lidenghui1110: done.

tests/e2e/nightly/single_node/ops/singlecard_ops/test_transpose_kv_cache_by_block.py
lidenghui1110: done.

jianzs (approved):

PR #6375, opened 2026-01-29
[Misc] support --kv-cached-memory-bytes option

NyaMisty:

This PR fixes `--kv-cached-memory-bytes` option for vllm-ascend. Before this fix, the corresponsding options will not have any effect.

In this PR, I manually cherry-picked the `--kv-cached-memory-bytes` related changes from vLLM upstream.

No

This PR is relatively simple itself, and this options is actually an advanced options that are only used when `--gpu-memory-utilization` is not working, thus should not introduce additional unit tests. I've manually tested this change in Qwen3-Next model
- vLLM version: v0.14.1


wangxiyuan:

PR #6388, opened 2026-01-29
[ST]Add e2e test for Npugraphex_pass

ForBetterCodeNine:
We found the custom passes of NPUGraphEX have implemented fusion operator features, which still require E2E test case validation and guard. This PR implements E2E test cases for the AddRMSNormQuant and SplitQKVNormRope operator fusions under NPUGraphEX that are already in the codebase.
NO

- vLLM version: v0.14.1


.github/workflows/scripts/config.yaml
```diff
@@ -56,6 +60,10 @@ e2e-singlecard:
     is_skipped: true
 
 e2e-singlecard-light:
+  - name: tests/e2e/singlecard/compile/test_graphex_norm_quant_fusion.py
```
wangxiyuan: don't need to add it for light test

.github/workflows/scripts/config.yaml
```diff
@@ -1,4 +1,8 @@
 e2e-singlecard:
+  - name: tests/e2e/singlecard/compile/test_graphex_norm_quant_fusion.py
+    estimated_time: 300
```
wangxiyuan: set to 80

.github/workflows/scripts/config.yaml
```diff
@@ -1,4 +1,8 @@
...
+  - name: tests/e2e/singlecard/compile/test_graphex_qknorm_rope_fusion.py
+    estimated_time: 300
```
wangxiyuan: ditto

PR #6432, opened 2026-01-30
[Bugfix] Synchronize only the current stream to avoid device sync

IWantFight:

Following [PR #4233](https://github.com/vllm-project/vllm-ascend/pull/4233), a synchronization mechanism was introduced between steps in asynchronous scheduling with ACL Graph to address a hanging issue. However, full device-level synchronization is unnecessary—only the operations on the current stream need to be synchronized. Otherwise, if other background operations (such as send and recv) are running concurrently, they may negatively impact inference performance for the instance.

hang problem
c4bbfac9a9088acec0ad335b4c2af437

Synchronizing only the current stream can also resolve the hang issue.
9c4feb3200abb407845eb38b3eaf81d6

No

- vLLM version: v0.14.1


realliujiaxu:
LGTM, thanks for your contribution!

PR #6440, opened 2026-01-30
【work around】reset the None reference count to prevent it from droppi…

1092626063:
…ng to zero and causing a crash.

work around: reset the None reference count to prevent it from dropping to zero and causing a crash.

- vLLM version: v0.14.1

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -17,6 +17,7 @@
 # Adapted from vllm-project/vllm/vllm/worker/gpu_model_runner.py
 #
 
+import ctypes
 import math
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -136,6 +137,9 @@
 
 SEQ_LEN_WITH_MAX_PA_WORKSPACE = 6144
 
+# TODO: remove this after python update to 3.12
+NONE_REF_COUNT_THRESHOLDS = 500000
```
winson-00178005, line 141: 可以把这个值设的再大一些

wangxiyuan:

PR #6451, opened 2026-01-31
【main】【bugfix】fix: restrict default MLAPO activation to Decode nodes only

fems14:
There is an issue with the current default logic for MLAPO (MLA Policy Optimization). By design, MLAPO should only be enabled by default on Decode (D) nodes. However, in hybrid (collocated prefill and decode) scenarios, the strategy is erroneously activated during the Prefill stage.
This PR corrects the default behavior to ensure that MLAPO is exclusively enabled for the Decoding phase. This prevents unexpected policy interference during Prefill and ensures optimal performance in hybrid deployment environments.

- vLLM version: v0.14.1

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -1,5 +1,14 @@
 # Pull Request Summary Style Guide
 
+## Output Instructions
+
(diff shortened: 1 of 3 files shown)


leo-pony (approved):
LGTM

wangxiyuan (approved):

PR #6453, opened 2026-01-31
[CI] support build wheel and docker image by workflow

wangxiyuan:
Make image and wheel build CI job work with workflow_dispatch way

- vLLM version: v0.14.1

### DIFF
--- a/.github/workflows/schedule_release_code_and_wheel.yml
+++ b/.github/workflows/schedule_release_code_and_wheel.yml
@@ -25,6 +25,17 @@ on:
     tags:
       - 'v*'
   workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Docker tag for build results'
+        default: main
+        required: true
+        type: choice
+        options:
+          - main
+          - v0.15.0rc1
+          - v0.14.0rc1
+          - v0.13.0rc2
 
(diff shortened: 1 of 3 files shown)


.github/workflows/schedule_release_code_and_wheel.yml
```diff
@@ -25,6 +25,12 @@ on:
     tags:
       - 'v*'
   workflow_dispatch:
+    inputs:
+      tag:
+        description: 'vllm-ascend tag to build'
+        type: string
+        default: main
```
Yikun, line 31: I suggest to add this as choice  like [1] to prevent unexpected unaccident. You can't assume that only you will use this feature:

[1] https://github.com/apache/spark-docker/blob/master/.github/workflows/publish-java21.yaml#L30

Do a simple review and merge before trigger publish.

Yikun: https://github.com/vllm-project/vllm-ascend/blame/main/.github/workflows/schedule_image_build_and_push.yaml#L25

Should also consider to add this

wangxiyuan: Done

PR #6469, opened 2026-02-01
[Bugfix] Fix the incorrect use of the output parameter in _forward_fia_slidingwindow

acat-rw:
Fix the incorrect use of the `output` parameter in `_forward_fia_slidingwindow`:
```
# Original (incorrect)
output, _ = torch_npu.npu_fused_infer_attention_score(...)
output= output.view(batch_size, self.num_heads, self.head_size)
```

In the original writing, the `output `parameter was directly assigned a new value, which is inconsistent with the interface definition, resulting in the inability to directly update `output `when calling externally.

```
attn_output, _ = torch_npu.npu_fused_infer_attention_score(...)
attn_output = attn_output.view(batch_size, self.num_heads, self.head_size)
output[:batch_size] = attn_output[:batch_size]
```

No change.

Co-authored-by: GoCHug<gch59135228@163.com>

vLLM ascend version: v0.13.0rc1


weijinqian0:
Thanks @GoCHug for same fixed pr.  https://github.com/vllm-project/vllm-ascend/pull/6558

PR #6491, opened 2026-02-02
[bugfix]Fix accuracy issue in PCP/DCP with speculative decoding

kunpengW-code:

This PR fixes an accuracy issue that occurs when using Prefill/Decode Context Parallelism (PCP/DCP) in conjunction with speculative decoding (MTP). The issue is caused by an irregular attention mask shape when both features are enabled.

The fix involves flattening the `block_table` for speculative decoding requests under PCP/DCP to ensure a regular attention mask. This PR also introduces a `use_cp` property for cleaner code and updates dummy runs to handle this scenario correctly.

No. This is a bug fix that improves accuracy and should not have user-facing API changes.

- vLLM version: v0.15.0

### DIFF
--- a/vllm_ascend/worker/pcp_utils.py
+++ b/vllm_ascend/worker/pcp_utils.py
@@ -521,7 +521,7 @@ class PCPManager:
         total_num_scheduled_tokens: int,
         query_lens: torch.Tensor,
         input_batch: "NPUInputBatch",
-        num_scheduled_tokens: np.ndarray,
+        num_scheduled_tokens: np.ndarray | None,
         block_table_tensor: torch.Tensor,
         num_reqs_padded: int,
         num_reqs: int,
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1869,6 +1869,34 @@ def _get_block_table_and_slot_mapping(kv_cache_gid: int):
             prefill_context_parallel_metadata=self.long_seq_metadata,
         )
 
+        if self.speculative_config and self.use_cp:
```
pisceskkk: Maybe we should extract this code snippet into a function of PCPManager?

kunpengW-code: Yes, this is more appropriate; we can integrate this logic with the generate_pcp_metadata method.

kunpengW-code:
deepseekv3.1 w8a8  tp8 pcp2 dcp8 mtp1 
gsm8k accuracy:
| dataset | version | metric | mode | vllm-api-general |
| gsm8kdataset | - | accuracy | gen | 96.29 |

whx-sjtu (approved):

PR #6519, opened 2026-02-03
[bugfix] Resolve operator name collision for DeepSeekV3.2 in RL scena…

Mind-s:
 
When using DeepSeek-V3.2 with the VERL framework, the ops names in vllm_ascend are identical to those in CANN, this causes the training process to invoke the ACLNN operators from vllm_ascend instead of the intended CANN operators, which ultimately leads to a segmentation fault.

Fixes:
We rename sparse_flash_attention and lightning_indexer to sparse_flash_attention_custom and lightning_indexer_custom to avoid the naming collision and the resulting segmentation fault in training.

No: The ops in sfa_v1.py have already been suffixed with custom, and this will not affect the usage of sfa_v1.py.

The ops are suffixed with custom, with no change to their usage in testing.


SlightwindSec:

PR #6535, opened 2026-02-04
[Bugfix] Fix FIA parameters in full-graph update for EAGLE speculative decoding

LICO1314:

This PR fixes two issues in the FIA (Fused Infer Attention) update logic for full-graph mode that caused `RuntimeError: error code 561002` during EAGLE speculative decoding:

**Issue 1: num_kv_heads mismatch for MLA models**

For MLA models (e.g., DeepSeek), key/value tensors are expanded from latent compression, so their actual head count equals `num_heads` rather than `num_kv_heads`. The code was passing `self.num_kv_heads` directly, causing the ratio of `num_heads / num_key_value_heads` to exceed the allowed limit.

**Fix:** Use actual key tensor shape (`key.shape[1]`) to determine `num_kv_heads`.

**Issue 2: actual_seq_lengths_q mismatch in graph replay**

In FIA with TND layout, query tensor size must match `actual_seq_lengths_q[-1]`. During graph replay, `actual_seq_lengths_q` was incorrectly updated from runtime metadata while the query tensor remained fixed from the capturing phase, causing: `When layout is TND, queryT(72) must be equal to the last element of actualSequenceLengthQ(65)`

**Fix:** Keep using the stored `actual_seq_lengths_q` from capturing phase instead of runtime metadata.

No.

CI is expected to pass. This fix resolves the runtime error for models running with EAGLE speculative decoding + full-graph (aclgraph) mode.

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -472,8 +472,8 @@ class AscendAttentionBackendImpl(AttentionImpl):
                         block_tables,
                         attn_mask,
                         block_size,
-                        seq_lens,
-                        query_start_loc,
+                        seq_lens_kv,
+                        actual_seq_lengths_q,
                         num_kv_heads,
                         num_heads,
                         scale,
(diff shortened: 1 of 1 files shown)


vllm_ascend/attention/attention_v1.py
```diff
@@ -540,6 +538,9 @@ def full_graph_fia(
         # Get workspace from cache or calculate it if not present.
         workspace = graph_params.workspaces.get(num_tokens)
         softmax_lse = torch.empty(1, dtype=query.dtype, device=query.device)
+        # For MLA models, key/value are expanded from latent, so their head count
+        # equals num_heads rather than num_kv_heads. Use actual tensor shape.
+        actual_num_kv_heads = key.shape[1]
         if workspace is None:
```
weijinqian0: why mla model use attention_v1

vllm_ascend/attention/attention_v1.py
```diff
@@ -540,6 +538,9 @@ def full_graph_fia(
         # Get workspace from cache or calculate it if not present.
         workspace = graph_params.workspaces.get(num_tokens)
         softmax_lse = torch.empty(1, dtype=query.dtype, device=query.device)
+        # For MLA models, key/value are expanded from latent, so their head count
```
whx-sjtu, line 541: MLA models won't go into attention_v1.

PR #6543, opened 2026-02-04
[EPLB][Nightly] Refactor UT

Spicy-Stick:
The basic configs are extracted and reused for eplb UT. This is done so that if the basic configs are changed later, eplb UT does not need to be modified repeatedly.

- vLLM version: v0.15.0

### DIFF
--- a/tests/e2e/nightly/single_node/models/test_deepseek_r1_0528_w8a8.py
+++ b/tests/e2e/nightly/single_node/models/test_deepseek_r1_0528_w8a8.py
@@ -63,10 +63,7 @@ aisbench_cases = [{
 }]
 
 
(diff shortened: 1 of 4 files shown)


tests/e2e/nightly/single_node/models/test_deepseek_r1_0528_w8a8.py
```diff
@@ -62,29 +62,29 @@
     "threshold": 0.97
 }]
 
+port = get_open_port()
+env_dict = {
```
shenchuxiaofugui: 这些可以用一个函数返回

zzzzwwjj (approved):

PR #6558, opened 2026-02-05
[BugFix]  Fixed an accuracy issue caused by the fact that the attention calculation result was not saved to the output when the FIA used the sliding window method.

GoCHug:
Issue: The model using _forward_fia_slidingwindow returns incorrect answers during the decode phase.
image

image

Fixed an accuracy issue caused by the fact that the attention calculation result was not saved to the output when the FIA used the sliding window method on branch main.

No change.

After the fix, the accuracy of the responses is correct.
image

- vLLM version: v0.15.0


weijinqian0 (approved):

weijinqian0:
The same fixed PR. https://github.com/vllm-project/vllm-ascend/pull/6469

PR #6592, opened 2026-02-06
[Ops] Update fused_sigmoid_gating_delta_rule_update_kernel

AyiStar:
Update fused_sigmoid_gating_delta_rule_update_kernel for performance on NPU.
The original implementation migrated from GPU will launch too many logical grids, which results in host-side cost especially on large batch (for bs=64, tp=4, 1024 grids).
So instead we bind grids physically for NPU, and move tiling logic into triton kernel.

For bs=64, dim=2048 input, the kernel latency is reduced by ~15-20%.

No

- vLLM version: v0.15.0


SlightwindSec:

PR #6614, opened 2026-02-08
[CI] Improve build speed

tfhddd:

- vLLM version: v0.15.0

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -15,7 +15,16 @@ on:
       contains_310:
         required: true
         type: boolean
-
+env:
+  UV_INDEX_URL: http://cache-service.nginx-pypi-cache.svc.cluster.local/pypi/simple
+  UV_EXTRA_INDEX_URL: https://mirrors.huaweicloud.com/ascend/repos/pypi
+  UV_INDEX_STRATEGY: unsafe-best-match
+  UV_NO_CACHE: 1
+  UV_SYSTEM_PYTHON: 1
+  SCCACHE_BUCKET: obs-guiiyang1-ascend-test
+  SCCACHE_ENDPOINT: https://obs.cn-southwest-2.myhuaweicloud.com
+  SCCACHE_REGION: cn-southwest-2
+  SCCACHE_S3_ENABLE_VIRTUAL_HOST_STYLE: true
 jobs:
   e2e-light:
     name: singlecard-light
@@ -31,11 +40,6 @@ jobs:
         VLLM_LOGGING_LEVEL: ERROR
         VLLM_USE_MODELSCOPE: True
         HF_HUB_OFFLINE: 1
-        # uv
-        UV_INDEX_URL: http://cache-service.nginx-pypi-cache.svc.cluster.local/pypi/simple
-        UV_EXTRA_INDEX_URL: https://mirrors.huaweicloud.com/ascend/repos/pypi
-        UV_INDEX_STRATEGY: unsafe-best-match
-        UV_NO_CACHE: 1
     steps:
       - name: Checkout vllm-project/vllm-ascend repo
         uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


SlightwindSec:

.github/workflows/_e2e_test.yaml
```diff
@@ -66,14 +72,15 @@ jobs:
...
-          VLLM_TARGET_DEVICE=empty pip install -e .
+          VLLM_TARGET_DEVICE=empty uv pip install --system -e .
 
       - name: Install vllm-project/vllm-ascend
         env:
           PIP_EXTRA_INDEX_URL: https://mirrors.huaweicloud.com/ascend/repos/pypi
         run: |
-          pip install -r requirements-dev.txt
-          pip install -v -e .
+          pip install uc-manager
```
Yikun, line 96: What's this? It seems a wrong deps.

.github/workflows/_e2e_test.yaml
```diff
@@ -31,6 +31,11 @@ jobs:
         VLLM_LOGGING_LEVEL: ERROR
         VLLM_USE_MODELSCOPE: True
         HF_HUB_OFFLINE: 1
+        # uv
+        UV_INDEX_URL: http://cache-service.nginx-pypi-cache.svc.cluster.local/pypi/simple
+        UV_EXTRA_INDEX_URL: https://mirrors.huaweicloud.com/ascend/repos/pypi
```
Yikun: We don't add any extra index before, this could contaminate environmental variables, causing us to overlook some dependency issues. I don't think this should be added.

.github/workflows/_e2e_test.yaml
```diff
@@ -132,14 +145,16 @@ jobs:
       - name: Install vllm-project/vllm from source
         working-directory: ./vllm-empty
         run: |
-          VLLM_TARGET_DEVICE=empty pip install -e .
+          VLLM_TARGET_DEVICE=empty uv pip install --system -e .
```
Yikun, line 170: why we need a system level deps?

.github/workflows/_e2e_test.yaml
```diff
@@ -226,6 +249,11 @@ jobs:
         VLLM_USE_MODELSCOPE: True
         HCCL_BUFFSIZE: 1024
         HF_HUB_OFFLINE: 1
+        # uv
+        UV_INDEX_URL: http://cache-service.nginx-pypi-cache.svc.cluster.local/pypi/simple
+        UV_EXTRA_INDEX_URL: https://mirrors.huaweicloud.com/ascend/repos/pypi
```
Yikun: same

.github/workflows/_unit_test.yaml
```diff
@@ -47,18 +54,18 @@ jobs:
...
+          VLLM_TARGET_DEVICE=empty uv pip install --system . --extra-index-url https://download.pytorch.org/whl/cpu/
+          uv pip uninstall --system triton
 
       - name: Checkout vllm-project/vllm-ascend repo
         uses: actions/checkout@v6
 
       - name: Install vllm-project/vllm-ascend
         run: |
-          export PIP_EXTRA_INDEX_URL=https://mirrors.huaweicloud.com/ascend/repos/pypi
+          pip install uc-manager
```
Yikun, line 65: ?

.github/workflows/_e2e_test.yaml
tfhddd: The current requirements-dev.txt file requires uc-manager to be installed, but installing uc-manager using `uv pip install` results in an error. I have submitted an issue: https://github.com/ModelEngine-Group/unified-cache-management/issues/736

.github/workflows/_e2e_test.yaml
tfhddd: 1. Based on the current workflow, the virtual environment is not activated, so directly using `uv pip install` will fail.

2. The role of `--system`: When you add the `--system` flag, you are essentially telling `uv`:

"I know what I'm doing. Please skip the virtual environment check and install the package directly to the global system path (or the currently selected system Python metric)."

.github/workflows/_e2e_test.yaml
tfhddd: After the functionality is verified, I will try to switch to a more elegant coding style.

PR #6625, opened 2026-02-09
[Refactor][Bugfix] Use upstream `mem_utils` for profiling and correct non-torch memory recorded during profiling

shen-shanshan:

1. Following https://github.com/vllm-project/vllm/pull/32322, use the `memory_profiling` context manager from vllm for profiling.
2. Fix wrong non-torch memory value recorded during profiling, which is not its peak during inference.

---
**More details about point 2:**

After profling, the non-torch memory value we recorded is lower than that in real inference. This is mainly because of the different memory management behaviour between `torch.cuda.empty_cache()` and `torch.npu.empty_cache()`.

With regard to `torch.cuda.empty_cache()`, it only recycle the unused memory in pytorch memory pool (i.e., memory managed by pytorch caching allocator), **with no affect to non-torch memory**. However, as for `torch.npu.empty_cache()`, it has a totally different memory management mechanism, i.e., it may call `aclrtSynchronize` and **enable Ascend runtime to free up non-torch memory**.

Thus, the non-torch memory value we recorded after `torch.npu.empty_cache()` is much lower than its peak during profling.

Resolution:

We record the peak non-torch memory value (`non_torch_memory_before_empty_cache`) after profiling, but before `torch.npu.empty_cache()`. Then, we add the diff (`non_torch_memory_cleared_by_empty_cache = non_torch_memory_before_empty_cache - self.non_torch_memory`) to non-torch memory when calculating available KV cache memory, which will lead to less KV cache memory (i.e., it's safer to avoid OOM issues).

---
> [!NOTE]
> This PR needs to wait for main2main aligning to latest vllm commit before merging.

no.

Before this PR, the non-torch memory we used to calculate available KV cache memory is **0.90 G**, whereas its peak during real inference is **1.08 G**, diff: **182.00 M**.

After this PR, we add this diff to non-torch memory after profiling and thus make the profiling results more accurate.
- vLLM version: v0.15.0

### DIFF
--- a/vllm_ascend/worker/worker.py
+++ b/vllm_ascend/worker/worker.py
@@ -52,7 +52,6 @@ from vllm_ascend.cpu_binding import bind_cpus
 from vllm_ascend.device_allocator.camem import CaMemAllocator
 from vllm_ascend.distributed.parallel_state import init_ascend_model_parallel
 from vllm_ascend.ops.triton.triton_utils import init_device_properties_triton
-from vllm_ascend.platform import NPUPlatform
 from vllm_ascend.utils import (
     AscendDeviceType,
     check_ascend_device_type,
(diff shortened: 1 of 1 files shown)


vllm_ascend/platform.py
```diff
@@ -505,6 +505,13 @@ def get_current_memory_usage(cls, device: torch.types.Device | None = None) -> f
         torch.npu.reset_peak_memory_stats(device)
         return torch.npu.max_memory_allocated(device)
 
+    @classmethod
+    def get_non_torch_memory_usage(cls, device: torch.types.Device | None = None) -> float:
```
wangxiyuan: if this is not a interface from vLLM, do not add it here.

shen-shanshan: ok.

PR #6659, opened 2026-02-10
[fix]: resolve DispatchFFNCombineBF16 operator errors and refactor redundant code for DispatchFFNCombine/DispatchFFNCombineBF16

guanguan0308:
1. Resolve the issue of inaccurate precision for the DispatchFFNCombineBF16 operator in decode scenarios
2. Eliminate unnecessary synchronization operations within the DispatchFFNCombine/DispatchFFNCombineBF16 operator
3. Resolve compilation errors that occur when building versions subsequent to b020：
Root Cause
During operator compilation, we previously modified the names of structs HcclOpResParam and HcclRankRelationResV2 in the moe_distribute_base.h file. After version b020, moe_distribute_base.h was updated with additional code that references these two structs. This resulted in compilation errors, as renaming the structs alone broke the newly added references to them.
Solution
To resolve the issue, we abandoned the approach of renaming HcclOpResParam and HcclRankRelationResV2 in moe_distribute_base.h. Instead, we adjusted the operator implementation to be fully compatible with the original struct names (as defined in the post-b020 version of moe_distribute_base.h).

- vLLM version: v0.15.0
- vLLM main: https://github.com/vllm-project/vllm/commit/13397841ab469cecf1ed425c3f52a9ffc38139b5


SlightwindSec:

PR #6670, opened 2026-02-10
add mxfp8 moe quantization

Eric-dot:
support mxfp8 quantization (Qwen MOE )
Using adaptor to make the hardware-specific behavior clearer and more maintainable

- vLLM version: v0.15.0
- vLLM main: https://github.com/vllm-project/vllm/commit/13397841ab469cecf1ed425c3f52a9ffc38139b5

### DIFF
--- a/vllm_ascend/device/device_op.py
+++ b/vllm_ascend/device/device_op.py
@@ -358,11 +358,11 @@ class A5DeviceAdaptor(BaseDeviceAdaptor):
         )[0]
 
 
-def get_device_adaptor():
+def get_device_adaptor() -> type["BaseDeviceAdaptor"]:
     ascend_device_type = get_ascend_device_type()
     if ascend_device_type == AscendDeviceType.A5:
         return A5DeviceAdaptor
     return BaseDeviceAdaptor
 
 
-DeviceOperator: type["BaseDeviceAdaptor"] | None = get_device_adaptor()
+DeviceOperator: type["BaseDeviceAdaptor"] = get_device_adaptor()


vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -35,7 +35,7 @@
 from vllm_ascend.flash_common3_context import get_flash_common3_context, set_flash_common3_context
 from vllm_ascend.ops.fused_moe.experts_selector import select_experts, zero_experts_compute
 from vllm_ascend.ops.fused_moe.moe_comm_method import AllGatherCommImpl, FusedExpertsResult, setup_moe_comm_method
-from vllm_ascend.quantization.methods.base import QuantType
+from vllm_ascend.ops.fused_moe.prepare_finalize import QuantType
```
SlightwindSec: There's no need to import `QuantType` from `prepare_finalize`, as it is already being imported there from `vllm_ascend.quantization.methods.base`. It's better to import it directly from the `base` module to avoid an indirect import.

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -235,13 +235,22 @@ def __init__(self, *args, **kwargs):
...
+        if hasattr(method, "quant_type"):
+            from vllm_ascend.quantization.methods.base import QuantType as SchemeQuantType
+
+            scheme_quant_type = method.quant_type
+            if scheme_quant_type == SchemeQuantType.W8A8:
+                return QuantType.W8A8
+            elif scheme_quant_type == SchemeQuantType.W4A8:
+                return QuantType.W4A8
+
+        return QuantType.NONE
```
SlightwindSec: Since `QuantType` only contains `NONE`, `W8A8`, and `W4A8`, the original implementation of `_get_quant_type` is logically equivalent and much more concise. Let's stick with the original implementation to keep the code clean.

vllm_ascend/ops/fused_moe/moe_comm_method.py
```diff
@@ -138,19 +142,43 @@ def fused_experts(
...
+            dispatch_results = self.token_dispatcher.token_dispatch_with_A5_quant(
+                hidden_states=hidden_states,
+                topk_weights=topk_weights,
+                topk_ids=topk_ids,
+                expert_map=expert_map,
+                global_redundant_expert_num=self.moe_config.
+                global_redundant_expert_num,
+                # shared_experts=shared_experts,
+                # quantized_x_for_share=quantized_x_for_share,
+                # dynamic_scale_for_share=dynamic_scale_for_share,
```
SlightwindSec: Please just delete the commented-out code.

vllm_ascend/ops/fused_moe/moe_mlp.py
```diff
@@ -64,6 +64,82 @@ def cumsum_group_list(
...
+
+    if dynamic_scale is None:
+        unquantized_hidden_states = hidden_states
+        hidden_states, pertoken_scale = torch_npu.npu_dynamic_mx_quant(
+            hidden_states, dst_type=act_quant_type)
+        # Dispose the original unquantized hidden states
+        # to save npu memory because they're no longer used.
+        dispose_tensor(unquantized_hidden_states)
+    else:
+        if dynamic_scale.ndim == 2: # w8a8mx gmm量化需要输入3维scale,拆分尾轴2出来
```
SlightwindSec: Please use English for all comments.

vllm_ascend/ops/fused_moe/fused_moe.py
daveth3t3chg33k: The var is already Initialized hence no need to call it

daveth3t3chg33k: The var is already Initialized hence no need to call it

SlightwindSec:
It looks like the **DCO check failed**. Please follow this guide to fix it: https://docs.vllm.ai/projects/ascend/en/latest/developer_guide/contribution/index.html

vllm_ascend/ops/fused_moe/moe_mlp.py
```diff
@@ -81,15 +95,49 @@ def quant_apply_mlp(
     w2_offset: torch.Tensor | None = None,
     fusion: bool = False,
     dynamic_eplb: bool = False,
+    **kwargs,
 ) -> torch.Tensor:
+    adaptor_cls = DeviceOperator
```
wangxiyuan: use DeviceOperator directly.

linfeng-yuan: We used it at the beginning but lead to mypy lint error. I would fix it at once.

linfeng-yuan: @wangxiyuan We've narrowed DeviceOperator type to pass mypy. Please have a quick check for the update~

linfeng-yuan (approved):

whx-sjtu (approved):

PR #6707, opened 2026-02-11
[Bugfix][DispatchFFNCombine] resolve vec error caused by unaligned UB access

serlar:
1. Fix a vec error caused by unaligned UB accesss in the DispatchFFNCombine;
2. Fix expert_token_nums tensor defined on host instead of NPU in moe_comm_method.py
3. Fix multi-core copy issue of expert_token_nums in dispatchffnCombine op (single aiv copy is sufficient)

No, this PR does not introduce any user-facing changes. The fix only addresses internal memory access logic and does not modify any public APIs, interfaces, or user-visible behaviors.

`export VLLM_ASCEND_ENABLE_FUSED_MC2=1`

vLLM version: v0.15.0

- vLLM version: v0.15.0
- vLLM main: https://github.com/vllm-project/vllm/commit/9562912cead1f11e8540fb91306c5cbda66f0007


vllm_ascend/ops/fused_moe/moe_comm_method.py
```diff
@@ -277,6 +277,11 @@ class FusedMC2CommImpl(MoECommMethod):
     Communication and Computation parallelism on Ascend devices.
     """
 
+    def __init__(self, moe_config):
+        super().__init__(moe_config)
+        self.expert_token_nums = torch.zeros([self.moe_config.num_local_experts], dtype=torch.int32, device="npu")
+        return
```
weijinqian0: remove this return

wangxiyuan (approved):

PR #6708, opened 2026-02-11
[CI] Refactor to speedup image building and CI Installation

wjunLu:
1. Refactor  image workflow using cache-from to speedup builds
build

Simultaneously refactored all Dockerfiles by placing layers that rarely change before those that change frequently, improving build cache hit rate.

2. Refactor E2E test using vllm-ascend container images, to skip C compile while no C code are changed
e2e

In this case, the job will only replace the source code of vllm-ascend and install `requirements-dev.txt`, saving about 10min before tests

- vLLM version: v0.15.0
- vLLM main: https://github.com/vllm-project/vllm/commit/9562912cead1f11e8540fb91306c5cbda66f0007

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -17,16 +17,71 @@ on:
         type: boolean
 
 jobs:
+  select-image:
+    runs-on: linux-aarch64-a2b3-0
+    outputs:
+      image: ${{ steps.select.outputs.image }}
+      image_a3: ${{ steps.select.outputs.image_a3 }}
+      image_310p: ${{ steps.select.outputs.image_310p }}
+    steps:
+      - name: Select image based on base branch
+        id: select
+        env:
+          IMAGE_NAMESPACE: swr.cn-southwest-2.myhuaweicloud.com/base_image/ascend-ci/vllm-ascend
+        run: |
+          BRANCH="${{ github.base_ref }}"
+          BRANCH_TAG="${BRANCH//\//-}"
+          MAIN_IMAGE="${{ inputs.image }}"
+          MAIN_IMAGE_A3="${{ inputs.image }}-a3"
+          MAIN_IMAGE_310P="${{ inputs.image }}-310p"
+          if [ "$BRANCH_TAG" = "main" ]; then
+            echo "Target branch is main, using main images: ${MAIN_IMAGE} / ${MAIN_IMAGE_A3} / ${MAIN_IMAGE_310P}"
+            echo "image=${MAIN_IMAGE}" >> $GITHUB_OUTPUT
+            echo "image_a3=${MAIN_IMAGE_A3}" >> $GITHUB_OUTPUT
+            echo "image_310p=${MAIN_IMAGE_310P}" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          # while target branch is not main
+          BRANCH_IMAGE="${IMAGE_NAMESPACE}:${BRANCH_TAG}"
+          BRANCH_IMAGE_A3="${IMAGE_NAMESPACE}:${BRANCH_TAG}-a3"
+          BRANCH_IMAGE_310P="${IMAGE_NAMESPACE}:${BRANCH_TAG}-310p"
+          # Check if branch-specific A2 image exists via IMAGE_NAMESPACE, fallback to main if not
+          if docker manifest inspect "${BRANCH_IMAGE}" > /dev/null 2>&1; then
+            echo "Using branch image: ${BRANCH_IMAGE}"
+            echo "image=${BRANCH_IMAGE}" >> $GITHUB_OUTPUT
+          else
+            echo "Branch image not found, falling back to ${MAIN_IMAGE}"
+            echo "image=${MAIN_IMAGE}" >> $GITHUB_OUTPUT
+          fi
+          # Check if branch-specific A3 image exists via IMAGE_NAMESPACE, fallback to main if not
+          if docker manifest inspect "${BRANCH_IMAGE_A3}" > /dev/null 2>&1; then
+            echo "Using branch A3 image: ${BRANCH_IMAGE_A3}"
+            echo "image_a3=${BRANCH_IMAGE_A3}" >> $GITHUB_OUTPUT
+          else
+            echo "Branch A3 image not found, falling back to ${MAIN_IMAGE_A3}"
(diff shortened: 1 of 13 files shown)


.github/workflows/scripts/install_vllm_ascend.sh
```diff
@@ -0,0 +1,30 @@
...
+if [ -n "$IMAGE_SHA" ] && git cat-file -e "${IMAGE_SHA}" 2>/dev/null; then
+  C_CHANGES=$(git diff "${IMAGE_SHA}"..HEAD --name-only -- \
+    csrc/ cmake/ CMakeLists.txt setup.py requirements.txt requirements-dev.txt)
+else
+  C_CHANGES="yes"
+fi
+
+if [ -n "$C_CHANGES" ]; then
+  echo "C code / build changes detected, reinstalling vllm-ascend..."
+  pip install -r requirements-dev.txt
```
Potabk:
```suggestion
  pip install -r ${DEST}/requirements-dev.txt
```

wjunLu: got

.github/workflows/scripts/install_vllm_ascend.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
Yikun: I don't like this separate file to install vllm-ascend. This will increase the cost to understand build workflow

.github/workflows/scripts/install_vllm_ascend.sh
```diff
@@ -0,0 +1,35 @@
...
+# Install vllm-ascend from the current checkout into /vllm-workspace/vllm-ascend.
+# Only recompiles C extensions if relevant files changed relative to the image.
+# Must be run from the vllm-ascend checkout directory.
+# Requires PIP_EXTRA_INDEX_URL to be set in the environment.
+set -e
+
+DEST="/vllm-workspace/vllm-ascend"
+
+# Read the image's vllm-ascend commit SHA before overwriting
```
Yikun: What if someone didn't do a rebase?

I don't think we should prebuild the vllm-ascend now.

.github/workflows/_e2e_test.yaml
```diff
@@ -9,6 +9,9 @@ on:
       image:
         required: true
         type: string
+      image_a3:
+        required: true
+        type: string
```
Yikun: this is not a good design.

the base workflow only have image, and child workflow fullfill the real image

Yikun (changes requested):
I need take a another look before merge.

.github/workflows/pr_test_full.yaml
```diff
@@ -81,6 +81,7 @@ jobs:
     uses: ./.github/workflows/_e2e_test.yaml
     with:
       vllm: ${{ matrix.vllm_version }}
-      image: swr.cn-southwest-2.myhuaweicloud.com/base_image/ascend-ci/cann:8.5.0-910b-ubuntu22.04-py3.11
+      image: quay.nju.edu.cn/ascend/vllm-ascend:main
```
Yikun: This should also use huaweicloud to speed ci

.github/workflows/pr_test_light.yaml
```diff
@@ -107,6 +107,7 @@ jobs:
     uses: ./.github/workflows/_e2e_test.yaml
     with:
       vllm: ${{ matrix.vllm_version }}
-      image: swr.cn-southwest-2.myhuaweicloud.com/base_image/ascend-ci/cann:8.5.0-910b-ubuntu22.04-py3.11
+      image: quay.nju.edu.cn/ascend/vllm-ascend:main
```
Yikun: I suggested this should be `base`

.github/workflows/scripts/install_vllm_ascend.sh
wjunLu: The key point is that there is always a gap between the code in image and the newest code in repo. 

Even though, the code in image is always the **real ENV** (like using CANN as ENV currently) compare to PR's code, in this case, and also the only way, we have to recompile all C code (like using CANN as base).

At least, this will not lead to failure  while holds the same time cost as current.

After the image build cost reduces, we can shorten the scheduled interval for image builds.

.github/workflows/scripts/install_vllm_ascend.sh
wjunLu: No problem, this can be placed in the job steps

.github/workflows/pr_test_full.yaml
wjunLu: OK, This will be fixed

.github/workflows/_e2e_test.yaml
wjunLu: get

.github/workflows/scripts/install_vllm_ascend.sh
wjunLu: We can add auto rebase here

wjunLu: done

Yikun:
It seems this PR didn't follow the workflow we discussed before, I just wann sure below case:

1. What if CANN upgrade case for e2e test case, seems you need to use the old CANN?
2. What if Mooncake upgrade case for e2e test case, seems we will use the old main image....

The workflow we discussed before is that:
- Step 1 **Build job image**: build `vllm-ascend-ci-image:branch-workflow-id` from cache `vllm-ascend-ci-image-cache:branch`
    do online e2e build for image from cache (as much as posible), if CANN/Mooncake and other deps upgrade, we should do *online build* (It's definately need more build time, it's acceptable). This step will **guarantee the job image will consider all code change (Dockerfile/code/deps)**
- Step 2 **Push job image** Push `vllm-ascend-ci-image:branch-workflow-id`
    (will only push delta layer, if no change will only do a re-tag action): `push to the image as tmp tag in registry vllm-ascend-ci-image:branch-workflow-id`
- Step 3 **Use job image** (will only pull delta layer, if no dockerfile change will only do a tag pull action): Use `vllm-ascend-ci-image:branch-workflow-id` to run all tests. 
- Step4 approved and merge
- Step5 **Refresh code** that means all devs will use the new dockerfile/deps if rebase
- (0-15mins)
- Step6 **Refresh base image** Push `vllm-ascend-ci-image:branch` from cache and `vllm-ascend-ci-image-cache:branch` from cache

if some PR submit between step5 and step6 will do build but no cache avalable (code is change but no base image yet).

PR #6715, opened 2026-02-12
[Feat]Xlite Qwen3 MoE Support Data Parallel

banxiaduhuo:
This patch adds support for the Qwen3-MoE data parallel in Xlite. For more details about Xlite, please refer to the following link:[https://atomgit.com/openeuler/GVirt/blob/master/xlite/README.md](https://atomgit.com/openeuler/GVirt/blob/master/xlite/README.md).

online server config:
```shell
port=$1
log=$2
export VLLM_USE_V1=1
export TASK_QUEUE_ENABLE=1
export HCCL_BUFFSIZE=512
export OMP_PROC_BIND=false
export VLLM_ASCEND_ENABLE_NZ=0
sysctl -w vm.swappiness=0
sysctl -w kernel.numa_balancing=0
sysctl kernel.sched_migration_cost_ns=50000
ip=127.0.0.1
python -m vllm.entrypoints.openai.api_server \
        --model /mnt/nvme1n1/wy/models/Qwen3-30B-A3B  \
        --tensor-parallel-size 2 \
        --enable-expert-parallel \
        --data-parallel-size 4 \
        --max-num-batched-tokens 32768 \
        --data-parallel-size-local 4 \
        --max-num-seqs=200 \
        --block-size 128 \
        --max-model-len 6656 \
        --trust-remote-code \
        --disable-log-requests \
        --served-model-name qwen \
        --no-enable-prefix-caching \
	--additional-config '{"xlite_graph_config": {"enabled": true, "full_mode": true}, "enable_cpu_binding": true}' \
	--compilation-config '{"cudagraph_capture_sizes":[1, 16, 32, 48, 64, 100, 150, 200], "cudagraph_mode": "FULL_DECODE_ONLY"}' \
	--async-scheduling \
	--host ${ip} \
	--port ${port} > ${log} 2>&1 &
``` 
test_config:
```shell
vllm bench serve \
    --max-concurrency ${maxconcurrency} \
    --num-prompts ${num_prompts} \
    --host ${HOST} \
    --port ${PORT} \
    --model ${MODEL_NAME} \
    --dataset-name random \
    --backend openai-chat \
    --random-input-len 512 \
    --random-output-len 512  \
    --random-range-ratio 0.2 \
    --temperature 0.6 \
    --metric-percentiles "50,90,99" \
    --tokenizer ${TOKENIZER_PATH} \
    --endpoint /v1/chat/completions \
    --ignore-eos
``` 

No

- vLLM version: v0.16.0
- vLLM main: https://github.com/vllm-project/vllm/commit/c86cdcbcd2d49c4d4cd38339315bacb1d8b2a1c0


vllm_ascend/xlite/xlite.py
```diff
@@ -299,7 +303,11 @@ def __call__(
             xlite_attn_metadata.is_prefills = [False] * num_decodes + [True] * num_prefills
             xlite_attn_metadata.block_tables = attn_metadata.block_tables.cpu().tolist()
 
-            h = self.hidden_states[: attn_metadata.num_actual_tokens]
+            # - Non-DP: batch_tokens == current_tokens
+            # - DP: batch_tokens >= current_tokens
+            batch_tokens = forward_context.batch_descriptor.num_tokens
```
MengqingCao: The naming seems a little wired. BTW, I think the pad in `CudagraphDispatcher` does not only occurs in DP scenario, but also when the batch size is not included in the capture sizes. @yiz-liu plz also take a look.

banxiaduhuo: xlite used num_actual_tokens to truncate hidden_states in non-DP (Data Parallel) scenarios. However, this caused issues in DP scenarios because collective communication requires all ranks within the DP domain to have the same tensor size. 
Therefore, batch_descriptor.num_tokens is now used to ensure compatibility between both non-DP and DP scenarios.

PR #6802, opened 2026-02-25
[Patch][Misc] Cleanup and update patches

wangxiyuan:

This PR performs a cleanup and update of the patch mechanism in `vllm-ascend`.

- Removes several obsolete patches: `patch_deepseek.py`.
- Updates the central patch documentation in `vllm_ascend/patch/__init__.py` to reflect these removals and additions, re-numbering and re-organizing the patch list for better clarity.

No. These are internal changes to the patching mechanism and should not affect users.

- vLLM version: v0.15.0


vllm_ascend/patch/worker/patch_rope.py
```diff
@@ -1,31 +0,0 @@
-#
```
wangxiyuan: this patch is useless now, right? @gcanlin

gcanlin: Yes. We don't need it now.

PR #6841, opened 2026-02-27
[Fix] Routing replay remove patch

Li-Yongwen:

[Fix] Routing replay remove patch

- vLLM version: v0.15.0


wangxiyuan:
do not remove the patch until we drop 0.16.0 support. 0.16 still need it IMO

PR #6844, opened 2026-02-27
[Perf] Optimize MTP execution by reordering state update operation

SlightwindSec:
## Summary
- Move `_update_states_after_model_execute` call from after main model sampling to after draft model execution
- This reordering reduces pipeline bubbles between main model and draft model execution
- No accuracy impact - the state update operation is independent of draft token proposal

## Performance Impact
Reduces idle time between main model and draft model execution stages, improving overall MTP (Multi-Token Prediction) performance.
- vLLM version: v0.15.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -115,6 +115,7 @@ from vllm_ascend.utils import (
     check_gdn_layer,
     enable_sp,
     enable_sp_by_pass,
+    global_stream,
     is_drafter_moe_model,
     is_moe_model,
     lmhead_tp_enable,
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1447,7 +1449,12 @@ def sample_tokens(
             sampler_output = self._sample(logits, spec_decode_metadata)
 
         if self.need_accepted_tokens:
-            self._update_states_after_model_execute(sampler_output.sampled_token_ids, scheduler_output)
+            if self.state_update_stream is None:
```
weijinqian0: this may use global_stream in utils.py when there is not conflict.

PR #6877, opened 2026-02-28
[EPLB] Display the expert hotness comparison before and after eplb.

shenchuxiaofugui:
To intuitively show the effect of the eplb algorithm, we print the expert heat before and after eplb.

Snipaste_2026-02-28_17-23-42

- vLLM version: v0.16.0

### DIFF
--- a/vllm_ascend/eplb/core/eplb_worker.py
+++ b/vllm_ascend/eplb/core/eplb_worker.py
@@ -17,6 +17,7 @@
 from multiprocessing import Process, Queue
 from typing import Any
 
+import numpy as np
 import torch
 import torch.distributed as dist
 from vllm.logger import logger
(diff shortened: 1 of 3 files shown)


vllm_ascend/eplb/core/eplb_worker.py
```diff
@@ -60,6 +61,15 @@ def do_update(self):
         old_placement = self.global2local(self.old_expert_maps, self.num_local_experts)
         _, _, new_placement = self.calculate_rebalance_experts(load_info, old_placement)
 
+        if self.rank_id == 0:
+            hotness = self._calculate_hotness(old_placement, load_info)
+            current_mean, current_max = self._compute_imbalance(old_placement, hotness)
+            update_mean, update_max = self._compute_imbalance(new_placement, hotness)
+            logger.info(
+                f"[Expert Hotness] Current: mean={current_mean:.3f}, max={current_max:.3f}, "
+                f"Updated: mean={update_mean:.3f}, max={update_max:.3f}"
```
Ronald1995: don't use f-string in logger.

PR #6885, opened 2026-03-01
update doc ui

wangxiyuan:

- vLLM version: v0.16.0

### DIFF
--- a/docs/source/_static/custom.css
+++ b/docs/source/_static/custom.css
@@ -1,0 +1,214 @@
+/**
+ * vLLM Ascend Documentation - Custom Styles
+ */
+
+/* CSS Variables */
+:root {
+  --primary: #0073e6;
+  --accent: #f04a00;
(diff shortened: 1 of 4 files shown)


Yikun:
Frankly, I don't really like the new style:
1. Even though it looks better visually, the information density is lower. Less information is displayed on a single screen, at the same time, it increases the cost of AI (more css means more token cost).
2. The visual style increases costs for both users and developers. It would be more reasonable to maintain the current or switch to a vLLM style, rather than creating a completely new style.
3. From a GEO's (Generative Engine Optimization) perspective, it's said that AI search tends to exclude self-generated data.

For other part, it's good (such as markdown and doc cmd improvement)

Yikun:
We need more something like `llms.txt` and `llms-full.txt` work to help AI reading.
(later)
https://github.com/vllm-project/vllm-ascend/pull/6886

wangxiyuan:
@Yikun Thanks for your suggestion. this is just a draft PR. I'll mark ready once it's done

wangxiyuan:
@gemini-code-assist please help me review this pull request. Let me know if there is any useless change.

PR #6887, opened 2026-03-01
[KVCache]Qwen3.5 supports contiguous tensor hybrid-attn kv-cache

nwpu-zxr:
Supports contiguous tensor hybrid-attn kv-cache on fullattn-mamba hybrid model, such as Qwen3Next and Qwen3.5.
Due to the restrictions of Ascend operators, all KV tensors, conv tensors, and SSM tensors must be contiguous. Therefore, this PR uses the following solution to generate the KV cache:
tensor1: [(kv_padding), conv                      , ...]
tensor2: [k                   , ssm                       , ...]
tensor3: [v                   , (mamba_padding), ...]
Under this scheme, although some waste may occur, the tensors of all caches are guaranteed to be contiguous.

No.

By CI.

- vLLM version: v0.16.0

### DIFF
--- a/vllm_ascend/patch/platform/patch_mamba_config.py
+++ b/vllm_ascend/patch/platform/patch_mamba_config.py
@@ -56,9 +56,7 @@ def verify_and_update_config(cls, vllm_config) -> None:
     # NOTE(zxr): because of the limit of Ascend Hardware, we need to keep
     # all cache tensors contiguous, so we align the page size of ssm_block
     # and single attn_block
-    attn_block_size = kernel_block_size * cdiv(
-        ssm_block_page_size, kernel_block_size * attn_single_token_k_page_size
-    )
+    attn_block_size = kernel_block_size * cdiv(ssm_block_page_size, kernel_block_size * attn_single_token_k_page_size)
     assert attn_single_token_k_page_size * attn_block_size == ssm_block_page_size, (
         "Cannot align ssm_page_size and attn_page_size."
     )


vllm_ascend/patch/platform/patch_mamba_config.py
```diff
@@ -33,33 +34,42 @@ def verify_and_update_config(cls, vllm_config) -> None:
...
-        block_size=1,
-        num_kv_heads=model_config.get_num_kv_heads(parallel_config),
-        head_size=model_config.get_head_size(),
-        dtype=kv_cache_dtype,
-    ).page_size_bytes
+    block_alignment_bytes = 128
+    # get attention block size
+    attn_num_kv_heads = model_config.get_num_kv_heads(parallel_config)
+    attn_head_size = model_config.get_head_size()
+    attn_single_token_page_size = attn_head_size * attn_num_kv_heads * get_dtype_size(kv_cache_dtype)
```
MengqingCao: ssm is aligned with k, let's change the name to make it more clear
```suggestion
    attn_single_token_k_page_size = attn_head_size * attn_num_kv_heads * get_dtype_size(kv_cache_dtype)
```

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2514,12 +2515,32 @@ def _allocate_kv_cache_tensors(self, kv_cache_config: KVCacheConfig) -> dict[str
         kv_cache_raw_tensors: dict[str, torch.Tensor | torch.Tensor | None] = {}
         # prefill disaggregation need the addr of cache tensor be aligned with 2M
         alignment = 2 * 1024 * 1024
+        layer_kv_cache_spec: dict[str, KVCacheSpec] = {}
+        for group_kv_cache_spec in kv_cache_config.kv_cache_groups:
+            for layer_name in group_kv_cache_spec.layer_names:
+                if layer_name not in layer_kv_cache_spec:
+                    layer_kv_cache_spec[layer_name] = []
+                layer_kv_cache_spec[layer_name].append(group_kv_cache_spec.kv_cache_spec)
+                assert len(layer_kv_cache_spec[layer_name]) <= 1, (
```
MengqingCao: If a layer has no kvcache, it won't appears in `group_kv_cache_spec.layer_names`, thus this assertion should be like this:
```suggestion
                assert len(layer_kv_cache_spec[layer_name]) == 1, (
```

vllm_ascend/patch/platform/patch_mamba_config.py
```diff
@@ -33,33 +34,42 @@ def verify_and_update_config(cls, vllm_config) -> None:
...
-    # with mamba layers, use FlashInfer instead).
-    attn_block_size = block_alignment_bytes * cdiv(mamba_page_size, block_alignment_bytes * attn_page_size_1_token)
+    )
+    ssm_shape, ssm_dtype = None, None
+    conv_shape, conv_dtype = None, None
+    for shape, dtype in zip(dummy_mamba_spec.shapes, dummy_mamba_spec.dtypes):
+        if len(shape) == 3:
+            ssm_shape, ssm_dtype = shape, dtype
+        else:
+            conv_shape, conv_dtype = shape, dtype
```
MengqingCao: Maybe a more general method is, calculating the physical mem size directly, and choosing the larger one to aligned with k, and another one as a pad block. we can make some comments to declare that in models with linear-full-hybrid attention, the larger one refers to ssm, and the smaller one refers to conv.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2514,12 +2515,32 @@ def _allocate_kv_cache_tensors(self, kv_cache_config: KVCacheConfig) -> dict[str
...
+                    "vLLM-Ascned does not support multi kv_cache_spec on one layer now."
```
MengqingCao:
```suggestion
                    "vLLM-Ascend does not support multiple `kv_cache_spec`s on one layer now."
```

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2514,12 +2515,32 @@ def _allocate_kv_cache_tensors(self, kv_cache_config: KVCacheConfig) -> dict[str
...
         for kv_cache_tensor in kv_cache_config.kv_cache_tensors:
-            # TODO: REFACTOR ME to sharing hybrid cache
+            use_mamba, use_attn = False, False
+            for layer_name in kv_cache_tensor.shared_by:
+                if isinstance(layer_kv_cache_spec[layer_name][0], MambaSpec):
+                    use_mamba = True
+                if isinstance(layer_kv_cache_spec[layer_name][0], AttentionSpec):
+                    use_attn = True
+            if use_mamba and use_attn:
+                use_attn_linear_hybrid = True
```
MengqingCao:
```suggestion
            use_attn_linear_hybrid = (use_mamba and use_attn)
```

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2514,12 +2515,32 @@ def _allocate_kv_cache_tensors(self, kv_cache_config: KVCacheConfig) -> dict[str
...
+        # have only linear or attention layers, for example, the mtp layer.
+        use_attn_linear_hybrid = False
         for kv_cache_tensor in kv_cache_config.kv_cache_tensors:
-            # TODO: REFACTOR ME to sharing hybrid cache
+            use_mamba, use_attn = False, False
+            for layer_name in kv_cache_tensor.shared_by:
+                if isinstance(layer_kv_cache_spec[layer_name][0], MambaSpec):
+                    use_mamba = True
+                if isinstance(layer_kv_cache_spec[layer_name][0], AttentionSpec):
+                    use_attn = True
```
MengqingCao: Let's end the loop earlier if both `use_mamba` and `use_attn` are `True`
```suggestion
                elif isinstance(layer_kv_cache_spec[layer_name][0], AttentionSpec):
                    use_attn = True
                if use_attn and use_mamba:
                    break
```

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2619,6 +2639,20 @@ def _reshape_kv_cache_tensors(
...
+            for layer_name in group.layer_names:
+                layer_kv_cache_spec[layer_name] = group.kv_cache_spec
+        attn_linear_hybrid_layers = []
+        for kv_cache_tensor in kv_cache_config.kv_cache_tensors:
+            has_mamba, has_attn = False, False
+            for layer_name in kv_cache_tensor.shared_by:
+                if isinstance(layer_kv_cache_spec[layer_name], MambaSpec):
+                    has_mamba = True
+                elif isinstance(layer_kv_cache_spec[layer_name], AttentionSpec):
+                    has_attn = True
```
MengqingCao: This code is written twice, maybe this should be extracted as a tool function in utils, and maybe a `self.hybrid_with_attn_and_mamba` variable is helpful.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2514,12 +2515,32 @@ def _allocate_kv_cache_tensors(self, kv_cache_config: KVCacheConfig) -> dict[str
...
         alignment = 2 * 1024 * 1024
+        layer_kv_cache_spec: dict[str, KVCacheSpec] = {}
+        for group_kv_cache_spec in kv_cache_config.kv_cache_groups:
+            for layer_name in group_kv_cache_spec.layer_names:
+                if layer_name not in layer_kv_cache_spec:
+                    layer_kv_cache_spec[layer_name] = []
+                layer_kv_cache_spec[layer_name].append(group_kv_cache_spec.kv_cache_spec)
+                assert len(layer_kv_cache_spec[layer_name]) <= 1, (
+                    "vLLM-Ascned does not support multi kv_cache_spec on one layer now."
+                )
```
MengqingCao: Does the meaning of this checking if there is only one spec in each layer? If so, I think it is redundant, cause all the functions that grouping kvcache in vLLM side allow this and will raise error if anything breaks it.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2636,6 +2670,18 @@ def _reshape_kv_cache_tensors(
                         ]
                         assert raw_dsa_k_tensor is not None
                         sum_page_size_bytes = raw_k_tensor.numel() + raw_v_tensor.numel() + raw_dsa_k_tensor.numel()
+                    elif self.use_hybrid_blocks and layer_name in attn_linear_hybrid_layers:
+                        raw_k_tensor, raw_v_tensor = kv_cache_raw_tensors[layer_name], kv_cache_raw_tensors[layer_name]
```
Angazenn: this seems to be redundant? since we only need one raw_tensor here, and it is sliced later.

nwpu-zxr: There are two assertions to check whether the KV tensor is empty. Therefore, the code needs to be written as follows.

vllm_ascend/worker/model_runner_v1.py
nwpu-zxr: OK, I will delete this.

vllm_ascend/patch/platform/patch_mamba_config.py
nwpu-zxr: OK, the variable name has been changed.

vllm_ascend/patch/platform/patch_mamba_config.py
nwpu-zxr: Okay, now the code will determine the shape and data type of conv and ssm based on the block size.

vllm_ascend/worker/model_runner_v1.py
nwpu-zxr: Fixed.

vllm_ascend/worker/model_runner_v1.py
nwpu-zxr: All code related to "multiple kvcache specs may exist at one layer" has been discarded.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2665,6 +2705,14 @@ def _reshape_kv_cache_tensors(
...
                             kv_cache_spec.head_size,
                         )
+                        if len(attn_linear_hybrid_layers) > 0:
+                            attn_tensor_page_size = int(np.prod(kv_cache_shape[1:])) * get_dtype_size(
+                                kv_cache_spec.dtype
+                            )
+                            conv_block_padding_size = raw_k_tensor.numel() - attn_tensor_page_size * 2
+                            raw_kv_tensor = raw_k_tensor[conv_block_padding_size:]
+                            raw_k_tensor = raw_kv_tensor[:attn_tensor_page_size]
+                            raw_v_tensor = raw_kv_tensor[attn_tensor_page_size : attn_tensor_page_size * 2]
```
MengqingCao:
```suggestion
                            raw_v_tensor = raw_kv_tensor[attn_tensor_page_size :]
```

MengqingCao (approved):
As the non-contigous triton operation is broken, there is no way to get the performance data of plan that k-ssm, v-conv share the same buffers seperately in a short time. Thus I'm okay with this pr to merge first, and support non-contigous kvcache in the ascendc operators later.

PR #6888, opened 2026-03-01
[CI] Revert pr-6708 to keep the old workflow that using CANN as the base container

wjunLu:
Revert pr-6708 to keep the old workflow that using CANN as the base container
- Related PR: https://github.com/vllm-project/vllm-ascend/pull/6708

- vLLM version: v0.16.0

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -17,71 +17,16 @@ on:
         type: boolean
 
 jobs:
-  select-image:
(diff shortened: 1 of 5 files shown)


Yikun:
I suggested revert all related PRs consider need more works. See https://github.com/vllm-project/vllm-ascend/pull/6891

Yikun:
Replaced by https://github.com/vllm-project/vllm-ascend/pull/6891

PR #6904, opened 2026-03-02
[P/D]add ut for cp

BossH2:

- vLLM version: v0.16.0

### DIFF
--- a/tests/ut/kv_connector/test_utils.py
+++ b/tests/ut/kv_connector/test_utils.py
@@ -1,73 +1,200 @@
 import unittest
+import math
 
-from vllm_ascend.distributed.kv_transfer.utils.utils import get_cp_group, get_tp_rank_head_mapping, parallel_info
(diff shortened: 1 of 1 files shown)


LCAIZJ:
When our input conditions (e.g., decode with tp=2, pcp=1, dcp=2, kv_head=8) result in dcp_repeat_num=0, I observed a KeyError in local_remote_block_port_mappings during testing. Could we enhance the unit test to cover this case?

wangxiaoteng888:
> When our input conditions (e.g., decode with tp=2, pcp=1, dcp=2, kv_head=8) result in dcp_repeat_num=0, I observed a KeyError in local_remote_block_port_mappings during testing. Could we enhance the unit test to cover this case?

Thank you for your suggestion. We do not support enabling DCP in this case because it does not have repeated heads.We will add more test cases and include additional validations.

wangxiyuan:
no update for long time. Feel free to reopen a new one if you're still working on this.

PR #6921, opened 2026-03-02
fix(metrics): ensure num_computed_tokens is monotonic to prevent nega…

xueliangyang-oeuler:
…tive counter increments

The AsyncLLM output_handler fails when Prometheus counters are incremented by negative amounts. This can happen if num_computed_tokens decreases between iterations (e.g., due to speculative decoding rejection or prefix cache adjustments). This commit ensures num_computed_tokens is monotonic in AscendScheduler.update_from_output to prevent such crashes.

- vLLM version: v0.16.0


xueliangyang-oeuler:
closed.

PR #6928, opened 2026-03-02
[BugFix] Fix muls_add fusion not working for GLM5 models

liuchenbing2026:
fix: support model-specific routed_scaling_factor in muls_add fusion
Previously, MulsAddFusionPass used a hardcoded scale=1.0, which failed
to match the x * routed_scaling_factor + y pattern in models like GLM5
that use routed_scaling_factor=2.5. This caused the muls_add fusion to
be skipped, leaving unoptimized mul+add operations.

This fix reads routed_scaling_factor from model config (defaulting to 1.0
for backward compatibility) and uses it as the pattern scale, enabling
correct fusion for GLM5 and other models with custom scaling factors.

Fixes: Unoptimized mul+add in GLM5 attention blocks
Tested: GLM5-W8A8 with routed_scaling_factor=2.5

- vLLM version: v0.16.0

### DIFF
--- a/vllm_ascend/compilation/passes/muls_add_pass.py
+++ b/vllm_ascend/compilation/passes/muls_add_pass.py
@@ -95,10 +95,8 @@ class MulsAddFusionPass(VllmInductorPass):
             logger.debug("MulsAdd fusion not enabled: unsupported dtype %s", dtype)
             return
 
-        # Currently we only register a single pattern instance with a fixed
-        # scalar scale value. If needed, multiple instances with different
-        # scales can be added here in the future.
-        MulsAddPattern(vllm_config, scale=1.0).register(self.pattern_match_passes)
+        routed_scaling_factor = getattr(vllm_config.model_config.hf_text_config, "routed_scaling_factor", 1.0)
+        MulsAddPattern(vllm_config, scale=routed_scaling_factor).register(self.pattern_match_passes)
 
     def __call__(self, graph: torch.fx.Graph) -> None:  # type: ignore[override]
         self.begin()


vllm_ascend/compilation/passes/muls_add_pass.py
```diff
@@ -95,11 +121,14 @@ def __init__(self, vllm_config: VllmConfig):
...
-        # scalar scale value. If needed, multiple instances with different
-        # scales can be added here in the future.
+        # Register default scale=1.0 pattern
         MulsAddPattern(vllm_config, scale=1.0).register(self.pattern_match_passes)
 
+        # Register routed_scaling_factor from model config (e.g. GLM5 uses 2.5)
+        routed_scaling_factor = getattr(vllm_config.model_config.hf_text_config, "routed_scaling_factor", None)
+        if routed_scaling_factor is not None and routed_scaling_factor != 1.0:
+            MulsAddPattern(vllm_config, scale=routed_scaling_factor).register(self.pattern_match_passes)
+
```
whx-sjtu:
```suggestion
routed_scaling_factor = getattr(vllm_config.model_config.hf_text_config, "routed_scaling_factor", 1.0)
MulsAddPattern(vllm_config, scale=routed_scaling_factor).register(self.pattern_match_passes)
```

whx-sjtu (changes requested):

vllm_ascend/compilation/passes/muls_add_pass.py
```diff
@@ -77,6 +77,32 @@ def replacement(x: torch.Tensor, y: torch.Tensor):
...
+    def register(self, pm_pass):
+        """Override base register to use scale-based pattern_id instead of
+        eps-based, so multiple MulsAddPattern instances with different scales
+        can coexist."""
+        import torch._inductor.pattern_matcher as pm
+        import torchair
+
+        pattern_id = f"{self.__class__.__name__}_scale_{self.scale}"
+        if pattern_id in _registered_patterns:
+            return
```
whx-sjtu: Let's just change the interface of `register` and directly pass `pattern_id` from outer calling. Then you can remove the `eps` definition in the `BasePattern` class.

whx-sjtu (approved):

PR #6936, opened 2026-03-02
[csrc][bugfix] Add compile-time Ascend950/910_95 compatibility for custom ops between CANN8.5 and 9.0

linfeng-yuan:
Remove hardcoded ASCEND910_95 usage in csrc custom-op host/tiling code and
select the SoC target at CMake configure time.

- Probe CANN headers with check_cxx_source_compiles:
  prefer platform_ascendc::SocVersion::ASCEND950, fallback to ASCEND910_95.
- Export the selected enum/config string via shared compile definitions (VLLM_ASCEND_950_SOC_ENUM / VLLM_ASCEND_950_SOC_CONFIG).
- Apply the shared macros to affected paths (moe_gating_top_k, add_rms_norm_bias) to avoid per-file hardcoding.
- Keep behavior unchanged; this is an internal build-compatibility fix for CANN 8.5 and 9.x.

No.

- vLLM version: v0.16.0


linfeng-yuan:
This solution cannot cover the cross-version case of “build with CANN 9.0 + run with CANN 8.5,” and that scenario itself is high-risk.

Why:

The current selection logic is decided once at build time (config.cmake probes headers and fixes macros); it is not re-evaluated at runtime.
If built with 9.0, it will be fixed to the ASCEND950/ascend950 path, which may not correctly match 910_95 semantics when running with 8.5.
More importantly, binary compatibility is a concern: custom ops/tiling/op host usually depend on the CANN headers and libraries used at build time, and ABI/behavior compatibility is not guaranteed across major runtime versions.

wangxiyuan (approved):

PR #6956, opened 2026-03-03
[Bugfix] fix the wrong use of extract_slice and insert_slice

wangx700:
fix the wrong use of extract_slice and insert_slice. In the TritonAscend main branch, extract_slice and insert_slice are located in the triton.language.extra.cann.extension library, rather than in the triton.language library

- vLLM version: v0.16.0

### DIFF
--- a/vllm_ascend/ops/triton/fla/solve_tril.py
+++ b/vllm_ascend/ops/triton/fla/solve_tril.py
@@ -15,7 +15,7 @@ from vllm.triton_utils import tl, triton
 from .utils import prepare_chunk_indices
 
 try:
-    import triton.language.extra.cann.extension as extension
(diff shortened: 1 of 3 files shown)


vllm_ascend/ops/triton/fla/solve_tril.py
```diff
@@ -14,6 +14,15 @@
 
 from .utils import prepare_chunk_indices
 
+try:
+    import triton.language.extra.cann.extension as extension  # type: ignore
+
+    insert_slice = extension.insert_slice
+    extract_slice = extension.extract_slice
+except ImportError:
```
Ronald1995, line 22: you could move the same logic to triton_utils.py

PR #7002, opened 2026-03-05
[Bugfix] Add `profile_prefix` param to worker `profile()` method

shen-shanshan:

When profiling, I get an error:

```bash
typeerror: npuworker.profile() takes from 1 to 2 positional arguments but 3 were given
```

To fix this bug, we need to receive the param `profile_prefix` passed from vllm.

- vLLM version: v0.16.0


wangxiyuan:
SEE https://github.com/vllm-project/vllm-ascend/pull/6968

PR #7009, opened 2026-03-05
[Feature] Remove block_size = 128 limitation

underfituu:
Improve prefix cache hit rate

Set `--block-size 16` in the configuration.

None
- vLLM version: v0.16.0


wangxiyuan:
Ascend doesn't perform good if block size!=128, have you tested the performance when set block size to 16 or some else?

PR #7020, opened 2026-03-05
[Bugfix] fallback to NPU RoPE ops for GLM5 in SFA graph mode

liuchenbing2026:

- vLLM version: v0.16.0

### DIFF
--- a/vllm_ascend/attention/sfa_v1.py
+++ b/vllm_ascend/attention/sfa_v1.py
@@ -421,6 +421,8 @@ class AscendSFAImpl(MLAAttentionImpl):
             self.is_rope_neox_style = False
             self.use_torch_npu_lightning_indexer = True
 
+        self._use_triton_rope = HAS_TRITON and self.vllm_config.model_config.hf_config.model_type not in ["glm_moe_dsa"]
+
         # Effective in SFA when FlashComm is enabled.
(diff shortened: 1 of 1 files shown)


vllm_ascend/attention/sfa_v1.py
```diff
@@ -421,6 +421,8 @@ def __init__(
             self.is_rope_neox_style = False
             self.use_torch_npu_lightning_indexer = True
 
+        self._use_triton_rope = HAS_TRITON and self.vllm_config.model_config.hf_config.model_type not in ["glm_moe_dsa"]
```
whx-sjtu: Please use `self.n_head` and `self.head_dim` to check whether to use triton rope.

vllm_ascend/attention/sfa_v1.py
```diff
@@ -421,6 +421,13 @@ def __init__(
             self.is_rope_neox_style = False
             self.use_torch_npu_lightning_indexer = True
 
+        # triton rope kernel does not tile the head dimension, skip when
+        # the tile (n_head * rope_dim // 2) would overflow the UB.
+        _max_triton_rope_heads = 16
```
rjg-lyh: If you write it this way, dsv3.2 will not be compatible with the Triton operator.

PR #7032, opened 2026-03-06
[P/D][KVPool]Mooncake Layerwise Connector supports kv_pool

nwpu-zxr:
This PR creates and registers `ascend_multi_connector`, which allows the `mooncake_layerwise_connector` to use the kv_pooling feature.
We unregister the original vllm's `MultiConnector` and replace it with `AscendMultiConnector` when registering the connectors.

No. User can use `MultiConnector` to initialize `AscendMultiConnector`.

By CI.

- vLLM version: v0.16.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/__init__.py
+++ b/vllm_ascend/distributed/kv_transfer/__init__.py
@@ -19,8 +19,11 @@ from vllm.distributed.kv_transfer.kv_connector.factory import KVConnectorFactory
 
 
 def register_connector():
+    # override multi_connector as ascend_multi_connector
+    if "MultiConnector" in KVConnectorFactory._registry:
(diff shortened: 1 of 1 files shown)


liziyu179:
@LCAIZJ Have any comments on this PR?

LCAIZJ:
> @LCAIZJ Have any comments on this PR?

1. The changes look good to me. This should resolve the issue we're currently facing https://github.com/vllm-project/vllm/issues/31145.
2. There is one edge case this doesn't cover yet. Currently, only the Prefill (P) stage uses the kvpool Connector. If the Decode (D) stage also needs to use it, this solution might not suffice. However, we can address this in a follow-up.
3. Could you clarify the rationale behind relocating ensure_kv_transfer_initialized(self.vllm_config, kv_cache_config)? It seems this change is unrelated to the current issue?

PR #7035, opened 2026-03-06
[CI]Upgrade niglty multi-node-tests max-parallel to 2

zhangxinyuehfad:

1. Increase nightly multi-node test max-parallel from 1 to 2, and fix resource conflicts that arise when tests run concurrently.
2. Fix parse-trigger job: Add an if condition so it only runs on schedule, workflow_dispatch, or PRs labeled nightly-test
3. Adjust nightly schedule: Shift trigger time from 24:00 to 23:45 (UTC+8)

- vLLM version: v0.16.0

### DIFF
--- a/.github/workflows/schedule_nightly_test_a2.yaml
+++ b/.github/workflows/schedule_nightly_test_a2.yaml
@@ -21,8 +21,8 @@ name: Nightly-A2
 
 on:
   schedule:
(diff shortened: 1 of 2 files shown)


zhangxinyuehfad:
/nightly deepseek-r1-0528-w8a8 deepseek-r1-0528-w8a8-prefix-cache

PR #7040, opened 2026-03-06
[CI] Image build from cache

Potabk:
## What

Introduce Docker registry cache (`cache-from`/`cache-to`) backed by
GitHub Container Registry (ghcr.io) to accelerate scheduled and
manual image builds.

### Workflow changes
- Add `cache_repo` input (default: `ghcr.io/vllm-project/vllm-ascend`)
  to both `_schedule_image_build.yaml` and
  `schedule_image_build_and_push.yaml`, letting callers choose where
  cache layers are stored.
- Add GHCR login step and `packages: write` permission so the runner
  can push cache manifests.
- Derive a per-arch, per-suffix cache tag at build time
  (e.g. `buildcache-openeuler-arm64`) so caches never collide across
  matrix variants.

### Dockerfile layer-order optimisation
Reorder layers in all Dockerfiles so the slow, rarely-changing steps
(system packages, Mooncake, vLLM core) appear before the fast-changing
`COPY . /vllm-workspace/vllm-ascend/` step. This maximises cache-hit
rate: rebuilding vllm-ascend from source no longer invalidates the
upstream dependency layers.

### Additional Dockerfile tweaks:
- Pre-install `clang-15` (required by triton-ascend) in the same
  apt layer as other system deps.
- Pre-install `modelscope` and `ray` before the vllm-ascend install
  so those layers are also cached.
- Switch vLLM clone from `--depth 1` to `--filter=blob:none` to avoid
  re-downloading on branch checkouts.
- Move `ARG SOC_VERSION` / `ENV` declarations immediately before
  `COPY .` to avoid early cache invalidation.

## How does this patch tested
see https://github.com/Potabk/vllm-ascend-fork/actions/runs/23038468201 in my personal repo

## Next steps

Extend `workflow_dispatch` inputs to support optional ARM + Ubuntu
build variants for nightly testing, while reusing the cache mechanism
introduced here (same `cache_repo` / cache-tag convention).

### DIFF
--- a/.github/workflows/_schedule_image_build.yaml
+++ b/.github/workflows/_schedule_image_build.yaml
@@ -28,10 +28,18 @@ on:
             description: 'Quay password for pushing images'
             required: false
 
+env:
+  QUAY_REPO: quay.io/ascend/vllm-ascend
+  CACHE_REPO: ghcr.io/vllm-project/vllm-ascend
+
 jobs:
   build-push-digest:
     name: build
     runs-on: ${{ matrix.runner }}
+    permissions:
+      packages: write
+      contents: read
+
     strategy:
       matrix:
         include:
@@ -42,6 +50,7 @@ jobs:
             runner: ubuntu-22.04-arm
             tag: arm64
     steps:
+
     - uses: actions/checkout@v6
       if: ${{ github.event_name != 'workflow_dispatch' }}
       with:
@@ -76,6 +85,22 @@ jobs:
         driver: docker-container
         use: true
 
+    - name: Login to GitHub Container Registry (for cache)
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Compute cache tag
+      id: cache-tag
+      run: |
+        if [ -n "${{ inputs.suffix }}" ]; then
+          echo "value=buildcache-${{ inputs.suffix }}-${{ matrix.tag }}" >> $GITHUB_OUTPUT
+        else
+          echo "value=buildcache-${{ matrix.tag }}" >> $GITHUB_OUTPUT
+        fi
+
     - name: Build and push
       uses: docker/build-push-action@v7
       id: build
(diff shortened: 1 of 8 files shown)


.github/workflows/_schedule_image_build.yaml
```diff
@@ -76,6 +83,22 @@ jobs:
         driver: docker-container
         use: true
 
+    - name: Login to GitHub Container Registry (for cache)
```
Yikun, line 88: Why we need introduce the `ghcr` as cache registry rather than `quay`?

wjunLu:
Maybe we should disable `cache-to` while releasing to save time

.github/workflows/_schedule_image_build.yaml
```diff
@@ -86,10 +109,12 @@ jobs:
         file: ${{ inputs.dockerfile || 'Dockerfile' }}
         # only trigger when tag, branch/main push
         push: ${{ inputs.should_push }}
-        outputs: type=image,name=quay.io/ascend/vllm-ascend,push-by-digest=true,name-canonical=true,push=${{ inputs.should_push }}
+        outputs: type=image,name=${{ env.QUAY_REPO }},push-by-digest=true,name-canonical=true,push=${{ inputs.should_push }}
         build-args: |
           PIP_INDEX_URL=https://pypi.org/simple
         provenance: false
+        cache-from: type=registry,ref=${{ env.CACHE_REPO }}:${{ steps.cache-tag.outputs.value }}
+        cache-to: ${{ inputs.should_push && format('type=registry,ref=ghcr.io/{0}:{1},mode=max', github.repository, steps.cache-tag.outputs.value) || '' }}
```
Yikun: Add some note for this two lines

.github/workflows/_schedule_image_build.yaml
```diff
@@ -76,6 +83,22 @@ jobs:
...
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Compute cache tag
+      id: cache-tag
+      run: |
+        if [ -n "${{ inputs.suffix }}" ]; then
```
Yikun, line 98: Also add note for this

.github/workflows/_schedule_image_build.yaml
```diff
@@ -32,6 +32,13 @@ jobs:
   build-push-digest:
     name: build
     runs-on: ${{ matrix.runner }}
+    permissions:
+      packages: write
+      contents: read
+    env:
+      QUAY_REPO: quay.io/ascend/vllm-ascend
+      CACHE_REPO: ghcr.io/${{ github.repository }}
```
Yikun: Why not use the static repo dirctly?

.github/workflows/_schedule_image_build.yaml
```diff
@@ -86,10 +109,12 @@ jobs:
         file: ${{ inputs.dockerfile || 'Dockerfile' }}
         # only trigger when tag, branch/main push
         push: ${{ inputs.should_push }}
-        outputs: type=image,name=quay.io/ascend/vllm-ascend,push-by-digest=true,name-canonical=true,push=${{ inputs.should_push }}
+        outputs: type=image,name=${{ env.QUAY_REPO }},push-by-digest=true,name-canonical=true,push=${{ inputs.should_push }}
```
Yikun, line 114: Why `branch` info are not in here?

Potabk: I'm wondering if we need branch info for different ref. Unless there exists a scenario where different branches have different versions of cann and Mooncake

.github/workflows/_schedule_image_build.yaml
Potabk: both support, but default is ghcr

.github/workflows/_schedule_image_build.yaml
Potabk: yes

PR #7050, opened 2026-03-06
Refactor quantization layer name mapping to leverage vLLM built-in mappers

Feng-xiaosuo:
…the quantization layer name

This PR modifies the loading logic for layer name prefixes in quantized models. The goal is to reduce or eliminate the need for point-to-point (hardcoded) modifications by leveraging the built-in mapper mechanism already provided in vLLM's model code. For models that do not yet have a corresponding mapper, the original point-to-point modification approach has been retained to ensure backward compatibility.

No.

The changes were validated using an offline deployment script to launch and verify multiple multimodal models. Testing confirmed that the updated loading logic correctly handles layer name prefixes across different model architectures, with no regression in model initialization or inference behavior.
- vLLM version: v0.16.0

### DIFF
--- a/.github/workflows/_unit_test.yaml
+++ b/.github/workflows/_unit_test.yaml
@@ -73,7 +73,8 @@ jobs:
             --ignore tests/ut/core/test_scheduler_dynamic_batch.py \
             --ignore tests/ut/kv_connector/test_mooncake_connector.py \
             --ignore tests/ut/worker/test_worker_v1.py \
-            --ignore tests/ut/spec_decode/test_mtp_proposer.py
+            --ignore tests/ut/spec_decode/test_mtp_proposer.py \
+            --ignore tests/ut/kv_connector/test_mooncake_layerwise_connector.py
 
       - name: Upload coverage to Codecov
         # only upload coverage when commits merged
--- a/.github/workflows/schedule_image_build_and_push.yaml
+++ b/.github/workflows/schedule_image_build_and_push.yaml
@@ -30,6 +30,7 @@ on:
         type: choice
         options:
           - main
+          - v0.16.0rc1
           - v0.15.0rc1
           - v0.14.0rc1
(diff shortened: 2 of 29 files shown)


vllm_ascend/quantization/modelslim_config.py
```diff
@@ -452,12 +436,75 @@ def override_quantization_method(cls, hf_quant_cfg, user_quant) -> str | None:
...
+            manual_mapping = QUANT_MODEL_PREFIX_MAPPINGS[self.model_type]
+            # Manual mapping is already in vLLM to HF direction, no need to reverse
+            combined_mapping.update(manual_mapping)
+
+        # Create and store the reverse WeightsMapper instance
+        self.vllm_to_hf_mapper = WeightsMapper(orig_to_new_prefix=combined_mapping)
+
+        # Debug info
+        if not new_to_orig_prefix:
+            logger.warning("No reverse mapping found for WeightsMapper. Using manual mappings if available.")
```
kunpengW-code: Since `orig_to_new_prefix` is not mandatory, this should not be logged as a warning.

vllm_ascend/quantization/modelslim_config.py
```diff
@@ -452,12 +436,75 @@ def override_quantization_method(cls, hf_quant_cfg, user_quant) -> str | None:
...
+
+        # Store the original mapper
+        self.hf_to_vllm_mapper = hf_to_vllm_mapper
+
+        # Create a reverse mapper (vLLM to HF)
+        new_to_orig_prefix = {}
+
+        # Try different ways to get the mapping based on WeightsMapper implementation
+        # Method 1: Check for public or private attribute with possible names
+        mapping_attrs = ["orig_to_new_prefix", "_orig_to_new_prefix"]
```
kunpengW-code: `_orig_to_new_prefix` does not exist in vllm, right?

PR #7087, opened 2026-03-09
[Bugfix] Resolve weight loading error with Ascend W4A8 quantization

ZhongsJie:
Resolve weight loading error with Ascend W4A8 quantization
```python
RuntimeError: The expanded size of the tensor (4) must match the existing size (16) at non-singleton dimension 1
```

- vLLM version: v0.15.0rc1

### DIFF
--- a/vllm_ascend/quantization/method_adapters.py
+++ b/vllm_ascend/quantization/method_adapters.py
@@ -220,6 +220,15 @@ class AscendFusedMoEMethod(FusedMoEMethodBase):
             set_weight_attrs(param, extra_weight_attrs)
 
         extra_weight_attrs.update({"quant_method": FusedMoeWeightScaleSupported.CHANNEL.value})
+
+        quant_method_name = self.quant_method.__class__.__name__
+        always_group_param = []
+        # w2_scale_bias must use GROUP only for w4a8 with new_quant_version
+        # so that _load_w2 narrows the checkpoint tensor [H, 16] to [H, 16 // tp_size] per TP rank.
+        if quant_method_name == "AscendW4A8DynamicFusedMoEMethod":
+            is_new_quant_version = getattr(self.quant_method, "new_quant_version", False)
+            if is_new_quant_version:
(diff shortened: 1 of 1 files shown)


ZhongsJie:
Fixes #7088

kunpengW-code:
This is probably a bug in constructing the `per_group_param` parameter.A simpler modification method is available at https://github.com/vllm-project/vllm-ascend/pull/7090

ZhongsJie:
> This is probably a bug in constructing the `per_group_param` parameter.A simpler modification method is available at #7090

@kunpengW-code As far as I know, this error only occurs in Ascend W4A8–quantized models.

In the implementation, new_quant_version is already considered:
```
if self.new_quant_version:
    param_dict["w13_scale_bias"] = torch.empty(
        num_experts, 2 * intermediate_size_per_partition, 1, dtype=torch.float32
    )
    param_dict["w2_scale_bias"] = torch.empty(
        num_experts, hidden_sizes, 16 // self.tp_size, dtype=torch.float32
    )
```
In the mentioned pr, the `scale_bias` tensors were all processed as group, and additionally `weight_scale_second` and `weight_offset_second` were also handled.

I’m wondering:
- Is it necessary to handle `scale_bias` entirely in a grouped way?
- What is the rationale for additionally processing `weight_scale_second` and `weight_offset_second` in this context?

kunpengW-code:
Both of your questions are related to the weight format generated by `Modelslim`.
1. `scale_bias` exists only in w4a8 weights, and it is generated considering TP, specifically generated for TP16. If EP is not enabled, the last dimension of `scale_bias` needs to be split according to the number of TPs, just like the weights. Therefore, even if the w4a8 weights are per-channel, they still need to load the actual weights in a per-group manner so that the actual weights can be correctly split.
2. `weight_scale_second` and `weight_offset_second` are bound to per-group quantization and are not affected by other conditions.

These three values should always be loaded in a per-group manner and should not be influenced by other conditions.

PR #7090, opened 2026-03-09
[bugdix] The problem that the w4a8 weight fails to be loaded when the EP is not enabled is resolved.

shaopeng-666:
This is a bug fix to resolve the issue where the MOE model fails to load quantized weights in w4a8 format when EP is not enabled.The parameters ["weight_scale_second", "weight_offset_second", "scale_bias"] shall be parsed in per-group mode, regardless of other conditions.
No

- vLLM version: v0.16.0


kunpengW-code:
This is a bug fix to resolve the issue where the MOE model fails to load quantized weights in w4a8 format when EP is not enabled.

PR #7114, opened 2026-03-10
[BugFix] Fix compilation errors for operators dispatch_gmm_combine_decode/moe_combine_normal/moe_dispatch_normal

wangyibo1005:
**What this PR does / why we need it?**
Fix compilation errors encountered when building versions later than b020 for the following operators:
dispatch_gmm_combine_decode, moe_combine_normal, moe_dispatch_normal
**Root Cause**
After the b020 version update, the original moe_distribute_base.h file was updated and its definitions changed, which caused compilation failures for the above three operators that depend on this file.
**Solution**
We have added a dedicated copy of moe_distribute_base.h into the implementation of these three operators, ensuring stable compilation independent of framework version updates.

**Does this PR introduce any user-facing change?**
No. There are no user-facing changes; this fix only resolves compilation issues without affecting functionality or user behavior.

**How was this patch tested?**
vLLM version: v0.16.0
vLLM main: https://github.com/vllm-project/vllm/commit/4034c3d32e30d01639459edd3ab486f56993876d
- vLLM version: v0.16.0

### DIFF
--- a/csrc/build_aclnn.sh
+++ b/csrc/build_aclnn.sh
@@ -39,12 +39,6 @@ elif [[ "$SOC_VERSION" =~ ^ascend910_93 ]]; then
             exit 1
         fi
     fi
-    # dependency: cann-toolkit file moe_distribute_base.h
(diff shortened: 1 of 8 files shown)


weijinqian0:
Please use only one of the same header files.

PR #7130, opened 2026-03-10
Add Qwen3_5 to model list

ppppeng:
The pr aims to add new models like Qwen3.5-35B-A3B/Qwen3.5-27B to model list for testing.

- vLLM version: v0.16.0


MengqingCao:
plz write more information in your pr message

PR #7146, opened 2026-03-11
[KV Pool]get_num_new_matched_tokens return 0 if token length < block_size

Pz1116:
Currently, we call lookup_client for looking up token hit in KV Pool, however, when token length < block size, the key will be empty and there is no point to lookup in KV Pool backend since there will never be a hit. 
Hence, add early return in `get_num_new_matched_tokens` when `token_len` < `block_size`

- vLLM version: v0.16.0

### DIFF
--- a/.github/workflows/schedule_image_build_and_push.yaml
+++ b/.github/workflows/schedule_image_build_and_push.yaml
@@ -30,6 +30,7 @@ on:
         type: choice
         options:
           - main
+          - v0.16.0rc1
           - v0.15.0rc1
           - v0.14.0rc1
           - v0.13.0rc3


LL1024LL:
Hi, I'm experiencing an issue after upgrading to v0.17.0-rc1.
                                                                                                                                               
  The API response's `prompt_tokens_details.cached_tokens` field is now returning `null` for all requests, whereas in the previous version it
  showed the number of cached tokens from the KV pool.                                                                                         
                  
  After investigation, I found that this PR (#7146) added an early return when `token_len < block_size`, which prevents the `load_spec` from   
  being created for short sequences. Since `load_spec` is the only mechanism that tracks `kvpool_cached_tokens`, this causes the metrics to be
  lost.                                                                                                                                        
                  
  Is this the intended behavior? Should the metrics tracking be decoupled from the KV pool lookup optimization?                                
  
  For context:                                                                                                                                 
  - Most of my requests are shorter than `block_size` (128 tokens)
  - The KV pool is still functioning (I can see transfer activity in logs)                                                                     
  - Only the metrics reporting is missing                                                                                                      
                                                                                                                                               
  Would it be possible to create a `load_spec` with `kvpool_cached_tokens=0` for short sequences to preserve metrics tracking while keeping the
   lookup optimization?

Pz1116:
> Hi, I'm experiencing an issue after upgrading to v0.17.0-rc1.
> 
> The API response's `prompt_tokens_details.cached_tokens` field is now returning `null` for all requests, whereas in the previous version it showed the number of cached tokens from the KV pool.
> 
> After investigation, I found that this PR (#7146) added an early return when `token_len < block_size`, which prevents the `load_spec` from being created for short sequences. Since `load_spec` is the only mechanism that tracks `kvpool_cached_tokens`, this causes the metrics to be lost.
> 
> Is this the intended behavior? Should the metrics tracking be decoupled from the KV pool lookup optimization?
> 
> For context:
> 
> * Most of my requests are shorter than `block_size` (128 tokens)
> * The KV pool is still functioning (I can see transfer activity in logs)
> * Only the metrics reporting is missing
> 
> Would it be possible to create a `load_spec` with `kvpool_cached_tokens=0` for short sequences to preserve metrics tracking while keeping the lookup optimization?

`load_spec` with `kvpool_cached_tokens=0` is doable, but I'm curious that where does `prompt_tokens_details.cached_tokens` take value from? Since LoadSpec is a class defined internally by AscendstoreConnector while `prompt_tokens_details.cached_tokens` seems like a metric from vLLM.

PR #7169, opened 2026-03-11
feat: fuse RMSNorm and bias add into single npu_add_rms_norm_bias kernel

liuchenbing2026:

- vLLM version: v0.16.0

### DIFF
--- a/vllm_ascend/ops/layernorm.py
+++ b/vllm_ascend/ops/layernorm.py
@@ -61,9 +61,15 @@ class AscendRMSNorm(RMSNorm):
                     x.add_(self.bias)
             return x, residual
 
-        x, residual = torch_npu.npu_rms_norm(x, self.weight, self.variance_epsilon)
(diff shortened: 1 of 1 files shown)


vllm_ascend/ops/layernorm.py
```diff
@@ -61,9 +61,15 @@ def forward_oot(
                     x.add_(self.bias)
             return x, residual
 
-        x, residual = torch_npu.npu_rms_norm(x, self.weight, self.variance_epsilon)
-        if self.bias is not None:
-            x.add_(self.bias)
+        if self.bias is not None and enable_custom_op():
+            x, _, _ = torch.ops._C_ascend.npu_add_rms_norm_bias(
+                x, torch.zeros_like(x), self.weight, self.bias,
```
whx-sjtu, line 66: Personally I prefer to make kernel support input without residual instead of performing `torch.zeros_like` here

PR #7170, opened 2026-03-11
docs: remove unsupported full_cuda_graph from Qwen2.5-Omni tutorial

stdjhs:
Clarify that full graph capture is not supported on Ascend NPU for Qwen2.5-Omni and remove the stale full_cuda_graph examples from single-node and multi-node commands.

- vLLM version: v0.16.0


wangxiyuan:
qwen2.5-omni's doc has been removed already

PR #7186, opened 2026-03-12
refactor: add a check before layer_sharding logging

pisceskkk:
We should only display this log message when layer_sharding is enabled.
- vLLM version: v0.16.0


MengqingCao (approved):
Thanks for this improvement!

PR #7226, opened 2026-03-13
[Perf] Optimize bias handling in AscendRMSNorm

rjg-lyh:
This PR optimizes bias handling in `AscendRMSNorm` without changing the intended
functional behavior.

In the current implementation, bias may be initialized for `AscendRMSNorm` based
on configuration-level detection, even though some norm layers never actually
load a bias weight. This can cause the inference path to enter the bias branch
and execute an unnecessary `add_` operator.

To improve this, this PR introduces a loader-based flag to record whether the
bias has actually been loaded. The bias addition is then executed only when the
bias is truly present.

This optimization reduces redundant computation in inference and makes the bias
application logic better aligned with the actual model weights.

No.

Before:
rmsnorm+add
image

After:
just rmsnorm
image

- vLLM version: v0.17.0

### DIFF
--- a/vllm_ascend/ops/layernorm.py
+++ b/vllm_ascend/ops/layernorm.py
@@ -37,11 +37,28 @@ class AscendRMSNorm(RMSNorm):
         super().__init__(hidden_size, eps, var_hidden_size, has_weight, dtype)
         vllm_config = get_current_vllm_config()
         self.bias = None
+        self.bias_loaded = False
+
         # quantization with anti_method m4 will generate none-zero norm bias
         if vllm_config.quant_config is not None and any(
             "norm.bias" in name for name in vllm_config.quant_config.quant_description
         ):
             self.bias = torch.nn.Parameter(torch.zeros(hidden_size), requires_grad=False)
+            self.bias.weight_loader = self._bias_weight_loader
+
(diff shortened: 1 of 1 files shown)


vllm_ascend/ops/layernorm.py
```diff
@@ -37,11 +37,28 @@ def __init__(
...
+        self.bias_loaded = False
+
         # quantization with anti_method m4 will generate none-zero norm bias
         if vllm_config.quant_config is not None and any(
             "norm.bias" in name for name in vllm_config.quant_config.quant_description
         ):
             self.bias = torch.nn.Parameter(torch.zeros(hidden_size), requires_grad=False)
+            self.bias.weight_loader = self._bias_weight_loader
+
+    def _bias_weight_loader(self, param: torch.nn.Parameter, loaded_weight: torch.Tensor) -> None:
```
clrs97, line 49: It is better to wrap the original weight loader such that we don't need to implement the details.

MengqingCao: I think the custom op has no weight loader function? I don't get your point. If you still have any question on this, plz feel free to open a new pr. This pr is ready for merge now.

PR #7230, opened 2026-03-13
[SpecDecode] Fix Draft model proposer

MengqingCao:
This pr fix the Unified draft parallel feature. 
1. In Draft model proposer, there are exceed 1 attention layers in target model,  thus removing the assertion on layer number.
2. we should get block size through `draft_attn_groups` instead of `attn_metadata_builder` after 0.17.0.
3. `attn_update_stack_num_spec_norm` shouldn't be done when unified draft parallel is enabled

Test pass with `tests/e2e/singlecard/spec_decode/test_v1_spec_decode.py::test_parallel_drafting_acceptance`, which is already included in CI

- vLLM version: v0.17.0


MengqingCao:
CI failed due to unrelated reasons, let's merge quickly to unblock CI
https://github.com/vllm-project/vllm-ascend/actions/runs/23078413556/job/67055332044?pr=7230

PR #7238, opened 2026-03-14
[BugFix][310P] remove triton-ascend's installation of 310p devices

Tflowers-0129:
The triton-ascend dependency has been removed from the pyproject.toml build requirements to prevent its installation on 310p, because the triton-ascend cannot use by 310p, when it install by default cause some error in spec and fusion pass.
NO
the ci and local tests
- vLLM version: v0.18.0


huangmengwei-cmss:
310p的dockerfile文件中的triton ascend也应该移除吧

Tflowers-0129:
> 310p的dockerfile文件中的triton ascend也应该移除吧

感谢，我去看看

Tflowers-0129:
no need

PR #7302, opened 2026-03-16
add qwen c8 kv cache fake-quantization

yesyue-w:

- vLLM version: v0.17.0


vllm_ascend/attention/attention_v1.py
```diff
@@ -829,6 +876,8 @@ def forward_fused_infer_attention(
                 num_heads=self.num_heads,
                 scale=self.scale,
                 sparse_mode=3,
+                key_antiquant_scale=key_scale,
+                value_antiquant_scale=value_scale,
```
kunpengW-code, line 880: Does the FIA operator support `chunked_prefill` and `prefix_cache` in the GQA fake quantization scenario on A5?

yesyue-w: support, but performance degradation, FIA is still optimizing.

PR #7312, opened 2026-03-16
[CI] Add multi-hardware wheel build and release workflow

zhangxinyuehfad:
Adds a scheduled CI workflow (schedule_release_code_and_wheel.yml) to automatically build and release vllm-ascend source packages and binary wheels for multiple Ascend hardware targets.

Key features:

1. Source release: Builds tar.gz sdist and uploads to PyPI on version tag push
2. Multi-hardware wheel builds: Supports three hardware targets in parallel:
2.1 A2 (Ascend 910B): x86_64 + ARM64, Python 3.10 / 3.11
2.2 A3 (Ascend 910C): x86_64 + ARM64, Python 3.10 / 3.11
2.3 310P: x86_64 + ARM64, Python 3.10 / 3.11
3. Wheel repair: Uses auditwheel to produce manylinux-compatible wheels, excluding Ascend NPU runtime libs (libascend*.so, libtorch*.so, etc.) that must be provided by the runtime environment
4. Variant wheels: Generates hardware-variant wheels via variantlib for hardware-specific distribution
5. OBS upload: Aggregates all variant wheels and a combined index JSON, then uploads to Huawei OBS for hosting

Yes. Users will be able to install hardware-specific vllm-ascend wheels from PyPI or the OBS variant index, eliminating the need to build from source.

1. CI verification only — workflow syntax and job dependency logic reviewed manually
2. Wheel build steps validated against existing Dockerfiles (Dockerfile.buildwheel.a2/a3/310p)
3. auditwheel exclusion list verified against known Ascend runtime shared libraries

- vLLM version: v0.17.0

### DIFF
--- a/.github/workflows/scripts/wheel/make_variant.py
+++ b/.github/workflows/scripts/wheel/make_variant.py
@@ -8,14 +8,12 @@ import shlex
 import subprocess
 import sys
 import zipfile
-from concurrent.futures import ThreadPoolExecutor
-from concurrent.futures import as_completed
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
-from variantlib.constants import VALIDATION_WHEEL_NAME_REGEX
-from variantlib.constants import VARIANT_DIST_INFO_FILENAME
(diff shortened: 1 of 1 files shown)


MengqingCao:
Let's trigger the wheel build manually later. Thanks for this great pr!

PR #7342, opened 2026-03-16
[Feat] Support separate attention backend for target and draft model.

SidaoY:
This PR enables separate attention backend configuration for target and draft models in speculative decoding, decoupling the previously bound attention backend settings between the two models.

It solves the compatibility issue where some draft models do not support the attention backend used by the target model, and allows users to select the optimal attention backend for each model individually to maximize inference performance. The change is fully backward compatible.

- Full backward compatibility: existing global attention backend configurations work without any modifications.
- New feature: support specifying a dedicated attention backend for the draft model for fine-grained configuration in speculative decoding.

- Tested on vLLM v0.17.0 (upstream commit `4034c3d`)
- Backward compatibility test: verified no regression with original global backend configuration
- Functional test: validated speculative decoding works correctly with different attention backend combinations for target and draft models
- Model compatibility test: covered mainstream LLM + draft model combinations
- End-to-end test: confirmed correct inference output and performance improvement with optimal backend selection


vllm_ascend/spec_decode/eagle_proposer.py
```diff
@@ -388,7 +390,10 @@ def dummy_run(
                     : num_reqs * self.decode_threshold
                 ]
 
-            builder = self.runner.attn_groups[0][0].get_metadata_builder()
+            if hasattr(self, "draft_attn_groups") and len(self.draft_attn_groups) > 0:
```
MengqingCao: Let's use `vllm_version_is` to separate 0.17.0 and other versions, that will make version upgrade more smoothly. And you can add an assertion on the length of `self.draft_attn_groups` in the 0.17.0 branch

PR #7347, opened 2026-03-16
Fix manual mapping registration and kimi_k2 layer name mapping

Feng-xiaosuo:

This PR fixes the layer name mapping logic in `AscendModelSlimConfig` for quantization config loading.
1. **kimi_k2 model layer name mapping issue**: The `kimi_k2` model has a unique layer naming convention that differs from the standard `hf_to_vllm` mapping. One layer was defined in the mapper but was not being correctly applied, causing quantization config lookup failures.
2. **Manual mapping registration timing issue**: The manual mapping check in `apply_vllm_mapper` was executed before `vllm_config` was initialized, causing `model_type` to be unavailable. This prevented some models with manual mappings from being correctly registered.
NO
Tested with `kimi_k2` model to verify the special layer name mapping works correctly. Also tested with other models that have manual mappings defined in `QUANT_MODEL_PREFIX_MAPPINGS` to ensure the registration timing fix works properly.
- vLLM version: v0.17.0

### DIFF
--- a/vllm_ascend/_310p/attention/attention_v1.py
+++ b/vllm_ascend/_310p/attention/attention_v1.py
@@ -78,6 +78,10 @@ class AscendAttentionBackend310(AscendAttentionBackend):
         """
         return AscendAttentionMetadataBuilder310
 
+    @staticmethod
+    def get_supported_kernel_block_sizes() -> list[int]:
+        return [128, 64]
+
 
 class AscendAttentionBackendImpl310(AscendAttentionBackendImpl):
     """
--- a/vllm_ascend/_310p/model_runner_310p.py
+++ b/vllm_ascend/_310p/model_runner_310p.py
@@ -17,6 +17,7 @@
 
 from __future__ import annotations
 
+import math
 from contextlib import contextmanager, nullcontext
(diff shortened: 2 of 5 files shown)


vllm_ascend/quantization/modelslim_config.py
```diff
@@ -478,12 +467,72 @@ def override_quantization_method(cls, hf_quant_cfg, user_quant) -> str | None:
                 return ASCEND_QUANTIZATION_METHOD
         return None
 
+    def apply_vllm_mapper(self, hf_to_vllm_mapper: "WeightsMapper"):
```
kunpengW-code, line 485: It is recommended to resolve the issue by modifying the key values in `self.quant_description` instead of flipping the community's `hf_to_vllm_mapper`.

kunpengW-code, line 485:
```suggestion
    # TODO: Modify the key values in self.quant_description instead of flipping the hf_to_vllm_mapper
    def apply_vllm_mapper(self, hf_to_vllm_mapper: "WeightsMapper"):
```

PR #7348, opened 2026-03-16
[Bugfix] Fix padding logic in eagle proposer for kimi25

slippersss:
This PR aims to fix padding logic in eagle proposer for kimi25. Main changes involve:
1. modify the way to obtain draft model attention builder and backend
2. add block table padding & related tensor slicing in common metadata when `draft_step>1` for solving fia verifying error
3. replace block table in `update_graph_params` for solving fia verifying error

N/A

by ci

- vLLM version: v0.17.0

### DIFF
--- a/vllm_ascend/spec_decode/eagle_proposer.py
+++ b/vllm_ascend/spec_decode/eagle_proposer.py
@@ -17,7 +17,7 @@ from vllm.distributed.parallel_state import (
     init_model_parallel_group,
     patch_tensor_parallel_group,
 )
-from vllm.forward_context import get_forward_context
+from vllm.forward_context import BatchDescriptor, get_forward_context
 from vllm.logger import logger
 from vllm.model_executor.layers.attention_layer_base import AttentionLayerBase
 from vllm.model_executor.model_loader import get_model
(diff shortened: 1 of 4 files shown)


slippersss:
> My suggestion is to delay `pad_query_start_loc_for_fia` after syncing across DP groups.

@yiz-liu I have changed it accordingly, and `target_model_batch_desc` is also added to ensure correct `uniform_decode` in eagle proposer. However, we should notice that `block_table_tensor` is fully passed from model runner to eagle proposer, so in some cases, we will obtain redundant table rows and thus fia ops will incur verifying error. Thus, we are forced to have no choice but to do slicing in `_pad_tensor`. We will take it seriously and find some way to fix it soon.

vllm_ascend/spec_decode/eagle_proposer.py
```diff
@@ -1546,10 +1559,14 @@ def _update_full_graph_params(self, forward_context, num_tokens, draft_attn_meta
...
-        padded_tensor = F.pad(tensor, pad, mode="constant", value=0)
-        return padded_tensor
+    def _pad_tensor(self, tensor, desired_size):
+        pad_size = desired_size - tensor.shape[0]
+        if pad_size > 0:
+            pad = [0] * (2 * tensor.dim() - 1) + [pad_size]
+            tensor = F.pad(tensor, pad, mode="constant", value=0)
+        else:
+            tensor = tensor[:desired_size]
+        return tensor
```
MengqingCao, line 1569: Maybe we could avoid to use `_pad_tensor` through adding a member `block_table` to modelrunner or eagleproposer. This could be done in a follow-up pr.

PR #7363, opened 2026-03-17
[310P][Feat] Support pooling runner for embedding and reranking models [Atlas Duo 300l]

kcramp858:
Remove the NotImplementedError that blocked runner_type='pooling' on 310P and let pooling models use the existing causal mask (matching 910 behavior). Override _forward_encoder_attention to use _npu_flash_attention since 310P does not support npu_fusion_attention with internal format tensors. Add a Meta tensor guard to maybe_trans_nz() to prevent npu_format_cast crashes during VL model profiling.

The 310P attention backend blocks runner_type='pooling' with a NotImplementedError stub, preventing embedding and reranking models (e.g. Qwen3-VL-Embedding-8B, Qwen3-VL-Reranker-8B) from serving on 310P hardware. The 910 backend already has full pooling support via _forward_encoder_attention().

This PR removes the blocker and fixes two additional issues discovered during integration testing on 310P3 hardware:

get_attention_mask() raises NotImplementedError for pooling models (attention_mask.py):
Removed the 3-line guard block. Pooling models now fall through to _get_causal_mask(), which is the same behavior as the 910 backend (see AscendAttentionBackendImpl._forward_encoder_attention which uses sparse_mode=0, meaning the mask content is irrelevant — only its existence in the metadata struct matters).

npu_fusion_attention rejects internal format tensors on 310P (attention_v1.py):
The parent class's _forward_encoder_attention() calls npu_fusion_attention with sparse_mode=3, which fails on 310P with RuntimeError: Current operator npu_fusion_attention do not support internal format. Added a 310P override that uses _npu_flash_attention instead — this is the same operator all other 310P attention paths (prefill, decode, chunked prefill) already use.

npu_format_cast crashes on Meta tensors during VL model profiling (utils.py):
During model profiling, vLLM creates Meta tensors (shape-only, no device data) to determine memory layout. On 310P, maybe_trans_nz() passes these to npu_format_cast, which crashes with "npu::npu_format_cast: attempted to run this operator with Meta tensors". Added if weight.is_meta: return weight at the top of maybe_trans_nz() as a single fix point — this protects all 7+ callers across the codebase (linear layers, quantization methods, MoE, embeddings, etc.) without needing scattered guards.

Yes. Users can now serve pooling models on 310P hardware using --runner pooling, which was previously blocked with NotImplementedError. No API changes — the existing /v1/embeddings, /v1/score, and /v1/rerank endpoints work as expected.

How was this patch tested?
Unit tests (3 test files updated):

test_attention_mask_310.py: Changed pooling test from assertRaises(NotImplementedError) to asserting a causal mask is returned with correct shape and -inf values.
test_attention_v1_310.py: Added test_forward_encoder_attention_310 verifying _npu_flash_attention is called with correct num_heads, num_kv_heads, and scale_value.
test_utils.py: Added test case 7 verifying Meta tensors are returned unchanged with no npu_format_cast call.
Integration tests on 310P3 hardware (two NPU devices, davinci0 + davinci1):

Qwen3-VL-Embedding-8B with --runner pooling on davinci1:

Returns 4096-dim embeddings with norm=1.0000, deterministic across 10 identical requests
Warm avg latency: ~121ms/request
Cosine similarity vs existing --task embed code path on davinci0: 0.999999 (max abs diff 0.00047, expected from float16 rounding in different code paths)
Qwen3-VL-Reranker-8B with --runner pooling on davinci1:

Relevant pair ("What is a cat?" / "A cat is a small domesticated carnivorous mammal."): score 0.96
Irrelevant pair ("What is a cat?" / "The stock market crashed yesterday..."): score 0.59
Existing embedding service on davinci0 unaffected during all testing
- vLLM version: v0.17.0

### DIFF
--- a/vllm_ascend/_310p/attention/attention_v1.py
+++ b/vllm_ascend/_310p/attention/attention_v1.py
@@ -34,6 +34,9 @@ from vllm_ascend.attention.attention_v1 import (
     AscendMetadata,
 )
 
+# ATB mask_type for full causal mask (MASK_TYPE_NORM)
+_MASK_TYPE_NORM = 1
+
 
 @register_backend(AttentionBackendEnum.CUSTOM, "ASCEND")
 class AscendAttentionBackend310(AscendAttentionBackend):
@@ -86,6 +89,19 @@ class AscendAttentionBackendImpl310(AscendAttentionBackendImpl):
     optimized for the Ascend 310P architecture.
     """
 
+    _use_v2: bool | None = None
+    # Shared mask builder instance, set on first pooling forward call.
+    _pooling_mask_builder: AttentionMaskBuilder310 | None = None
+
+    @classmethod
+    def _check_v2_available(cls) -> bool:
+        """Lazily detect whether _npu_flash_attention_v2 is available."""
+        if cls._use_v2 is None:
+            cls._use_v2 = (hasattr(torch_npu, 'atb')
+                           and hasattr(torch_npu.atb,
+                                       '_npu_flash_attention_v2'))
+        return cls._use_v2
+
     def _forward_encoder_attention(
         self,
         query: torch.Tensor,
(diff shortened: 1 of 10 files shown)


vllm_ascend/_310p/attention/attention_v1.py
```diff
@@ -85,6 +86,34 @@ class AscendAttentionBackendImpl310(AscendAttentionBackendImpl):
     optimized for the Ascend 310P architecture.
     """
 
+    def _forward_encoder_attention(
```
pu-zhe, line 105: This function is never be called in this framework. Actually, forward_prefill_310 will be called and work well in pooling model.
So, codes in line 89 - 116 are not necessary.

kcramp858: Are you sure? I think the confusion might be that you're looking at the forward_impl() dispatch
  path — where forward_prefill_310() handles the PrefillNoCache state — and noticing that pooling models     would go through there. But actually, pooling models never reach forward_impl() at all.               
                                                                                                             The parent class AscendAttentionBackendImpl.forward() (line 980) checks attn_metadata.model_runner_type 
  == "pooling" and returns early via self._forward_encoder_attention(...) before forward_impl() is called.   Since AscendAttentionBackendImpl310 doesn't override forward(), the parent's version runs, and Python's
  MRO dispatches self._forward_encoder_attention() to the 310P override.                                                                                               
  So the path is:
  - Pooling models: forward() → _forward_encoder_attention() (310P override, returns at line 983)
  - Causal models: forward() → reshape_and_cache() → forward_impl() → forward_prefill_310()                                                                                                           
  Is that what you were thinking — that pooling would go through forward_impl() → forward_prefill_310()?

PR #7388, opened 2026-03-17
[Nightly] Nightly pre-build image

Potabk:
This pull request refactor nightly image build and simplify the logic of multi workflows.
1. Nightly image build become the prerequisite when the test are triggered by `schedule` or `workflow_dispatch`
2. Simplify the pull request select case logic
3. Next step: Implement replaceable nightly tests. Specifically, if nightly tests are manually triggered, they can accept any optional docker image to meet the needs of different commits(Which means the image is customizable).

- vLLM version: v0.17.0

### DIFF
--- a/.github/workflows/nightly_image_build.yaml
+++ b/.github/workflows/nightly_image_build.yaml
@@ -22,6 +22,7 @@ name: Nightly Image Build Schedule
 
 on:
   workflow_dispatch:
+    # Next step: Add more inputs here if needed, e.g. vllm version, vllm-ascend version, image tag, etc.
 
 jobs:
   build-a2:


.github/workflows/nightly_image_build.yaml
```diff
@@ -1,66 +1,43 @@
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
```
wjunLu, line 2: No schedule anymore, rename the yaml

Potabk: yes, looks reasonable and fixed

PR #7390, opened 2026-03-17
[Feature] Support Flash Comm V1 for VL models (with MLA)

SlightwindSec:
## Summary

Flash Comm V1 (flashcomm1) was previously blocked for all VL models.

**Root cause:** For VL models, `inputs_embeds` at layer 0 originates from the vision encoder as a full `[N, H]` tensor — it has **not** been reduce-scattered across TP ranks. The original MLA forward path assumed inputs were already scattered, producing wrong output shapes under TP > 1.

**Fix:**
- Detect at init time (statically, not via runtime shape checks) whether a layer is the first layer of a VL model (`is_vl_first_layer`) so dynamo treats the branch as a constant.
- In `AscendMultiHeadLatentAttention.forward`, when `flashcomm1 + TP > 1 + is_vl_first_layer`, set `need_gather_q_kv=False` and pre-allocate output as `[N//tp_size, H]`.
- Remove the platform-level assertion that prevented VL models from enabling Flash Comm V1.

**Other improvements:**
- `is_vl_model()` now uses vllm's canonical detection (`hf_config is not hf_text_config`) instead of fragile key-name checks, with the old checks kept as fallback.
- Added `parse_layer_idx(prefix)` utility.
- Added `maybe_chunk_residual` call in `AscendRMSNorm` before the add-rms-norm op.
- Removed unnecessary CPU/fp32 round-trip in
  `AscendLearnable2DInterpPosEmbDivided_fixed.forward()`.

___

Huge thanks to @realliujiaxu for his guidance and contribution to this feature! (Missed the co-author tag in the commit, dropping a mention here).

- vLLM version: v0.17.0

### DIFF
--- a/vllm_ascend/patch/worker/patch_kimi_k25.py
+++ b/vllm_ascend/patch/worker/patch_kimi_k25.py
@@ -42,15 +42,20 @@ class AscendLearnable2DInterpPosEmbDivided_fixed(nn.Module):
     def forward(self, x: torch.Tensor, grid_thws: torch.Tensor) -> torch.Tensor:
         pos_embs = []
         for t, h, w in grid_thws.tolist():
+            x_device = x.device
+            x_dtype = x.dtype
             assert t <= self.num_frames, f"t:{t} > self.num_frames:{self.num_frames}"
             if (h, w) == self.weight.shape[:-1]:
                 pos_emb_2d = self.weight.flatten(end_dim=1)
             else:
+                weight_fp32 = self.weight.to(dtype=torch.float32)
+                weight_cpu = weight_fp32.to("cpu")
                 pos_emb_2d = get_rope_shape(
-                    self.weight,
+                    weight_cpu,
                     interpolation_mode=self.interpolation_mode,
                     shape=(h, w),
                 )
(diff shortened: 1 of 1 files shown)


vllm_ascend/utils.py
```diff
@@ -1244,3 +1259,9 @@ def trans_nd_to_nz(cache_tensor: torch.Tensor):
     cache_tensor = cache_tensor.reshape(nz_shape[:-4] + [m1, m0, n1, n0])
     cache_tensor = cache_tensor.permute(*array_trans)
     return cache_tensor
+
+
+def parse_layer_idx(prefix: str) -> int | None:
```
MengqingCao, line 1264: I think this is redundant with `extract_layer_index`?

SlightwindSec:
Huge thanks to @realliujiaxu for his guidance and contribution to this feature! (Missed the co-author tag in the commit, dropping a mention here).

vllm_ascend/utils.py
SlightwindSec: Thanks for catching this! I've switched to `extract_layer_index` and removed `parse_layer_idx` in #7565.

PR #7418, opened 2026-03-18
[Feat]skip lightning indexer for the first 2048 preceding tokens

1024daniel:
Skip first 2048 tokens for lighting indexer to avoid redundant computation

No

- vLLM version: v0.17.0
- vLLM main: https://github.com/vllm-project/vllm/commit/4497431df654e46fb1fb5e64bf8611e762ae5d87
### precision test with DeepSeek V3.2
The markdown format results is as below:
| gsm8k | 7cd45e | accuracy | gen | 95.75 |

### DIFF
--- a/vllm_ascend/attention/sfa_v1.py
+++ b/vllm_ascend/attention/sfa_v1.py
@@ -30,6 +30,8 @@ from vllm_ascend.attention.utils import (
     AscendCommonAttentionMetadata,
     ascend_chunked_prefill_workspace_size,
     enable_cp,
+    get_index_of_skipped_queries_numpy,
+    get_li_skip_indices,
     maybe_save_kv_layer_to_connector,
     trans_rope_weight,
     transdata,
@@ -53,6 +55,7 @@ from vllm_ascend.utils import (
     enable_dsa_cp,
     enable_dsa_cp_with_layer_shard,
     enable_dsa_cp_with_o_proj_tp,
+    enable_lightning_indexer_skip,
     get_weight_prefetch_method,
     maybe_trans_nz,
 )
@@ -63,6 +66,8 @@ if TYPE_CHECKING:
 
 # token count limits within bmm_transpose operator
 BMM_TRANS_MAX_SUPPORTED_TOKENS = 1024
+
+SPARSE_COUNT = 2048
 
 
 class AscendSFABackend(AttentionBackend):
@@ -144,11 +149,17 @@ class AscendSFAMetadata:
     # chunked prefill by default if no attn_states passed
     attn_state: AscendAttentionState = AscendAttentionState.ChunkedPrefill
     dsa_cp_context: DSACPContext | None = None
+    li_reorder_indices: torch.Tensor | None = None
     reshape_cache_event: torch.npu.Event = None
     sfa_cp_metadata: AscendPCPMetadata | None = None
+    num_actual_seqs: int = 0
+    top_k_indices_skip_li_query: torch.Tensor | None = None
     num_decodes: int = 0
     num_decode_tokens: int = 0
     num_prefills: int = 0
+
+    num_of_non_skip_tokens: int = 1
+    skip: bool = False
 
 
 M = TypeVar("M", bound=AscendSFAMetadata)
@@ -200,6 +211,10 @@ class AscendSFAMetadataBuilder(MLACommonMetadataBuilder[AscendSFAMetadata]):
         self.actual_seq_lengths_query = torch.zeros(max_num_reqs + 1, dtype=torch.int32, device=device)
         self.actual_seq_lengths_key = torch.empty_like(self.actual_seq_lengths_query)
 
+        self.enable_lightning_indexer_skip = enable_lightning_indexer_skip()
+        self.index_of_skipped_queries = None
+        self.num_actual_seqs = max_num_reqs
+
     @staticmethod
     def determine_chunked_prefill_workspace_size(vllm_config: VllmConfig) -> int:
         return ascend_chunked_prefill_workspace_size(vllm_config)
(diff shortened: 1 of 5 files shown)


vllm_ascend/attention/utils.py
```diff
@@ -331,3 +343,135 @@ def enabling_mlapo(vllm_config: VllmConfig) -> bool:
...
     return bool(envs.VLLM_ASCEND_ENABLE_MLAPO and is_decode_instance)
+
+
+def to_numpy(x):
+    if isinstance(x, torch.Tensor):
+        return x.cpu().numpy()
+    return x
+
+
+def get_sfa_skip_indices(num_comptuted_tokens, query_lens):
```
rjg-lyh: I think better to rename this func name, like "get_li_skip_indices)".

vllm_ascend/attention/sfa_v1.py
```diff
@@ -910,53 +977,106 @@ def indexer_select_post_process(
...
-                q_li, cos, sin, rope_dim=self.qk_rope_head_dim, is_neox_style=self.is_rope_neox_style
-            )
+        sparse_count = 2048
+
+        # =========================
+        # step1: determine token range
+        # =========================
+        if attn_metadata.skip:
+            num_tokens = attn_metadata.non_skip_num_actual_tokens
+            if num_tokens > 0:
```
rjg-lyh: I suggest that you could refactor this logic after capacity finish, in order to separate skip and non-skip token sequences for function invocation.

vllm_ascend/attention/sfa_v1.py
```diff
@@ -1169,7 +1289,7 @@ def forward(
 
             k_li = self._get_full_kv(k_li, attn_metadata)
 
-        if kv_cache is not None:
+        if kv_cache is not None and (not attn_metadata.skip or attn_metadata.non_skip_num_actual_tokens > 0):
```
rjg-lyh: This condition may lead to precision issues. The skip indices are introduced to reduce matmul and LI operator computation, but `k_li` still needs to be stored globally for use in subsequent scheduling batches.

1024daniel: has fix this branch condition

vllm_ascend/attention/utils.py
1024daniel: has change the func name

vllm_ascend/attention/utils.py
```diff
@@ -331,3 +341,133 @@ def enabling_mlapo(vllm_config: VllmConfig) -> bool:
...
+    indices = np.concatenate([q_cal_part_indices, q_skip_indices]).astype(np.int32)
+
+    return indices, li_cum_query_lens, li_seq_lens, li_skipped_mask, num_of_non_skip_tokens
+
+
+def get_index_of_skipped_queries_numpy(actual_seq_lengths_query, actual_seq_lengths_key, num_actual_seqs, sparse_count):
+    actual_seq_lengths_query = to_numpy(actual_seq_lengths_query)
+    actual_seq_lengths_key = to_numpy(actual_seq_lengths_key)
+    num_actual_seqs = to_numpy(num_actual_seqs)
+    import numpy as np
```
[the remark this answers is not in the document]
1024daniel: has put import expression to the top

vllm_ascend/attention/utils.py
```diff
@@ -162,6 +168,8 @@ class AscendCommonAttentionMetadata(CommonAttentionMetadata):
     # Metadata for Prefill Context Parallelism (PCP) operations.
     prefill_context_parallel_metadata: AscendPrefillContextParallelMetadata | None = None
 
+    lightning_indexer_metadata: AscendLightningIndexerMetadata | None = None
```
YzTongNiar: We can refactor 'lightning_indexer_metadata' as 'lightning_indexer_context' just like 'dsa_cp_context' in the sfa metadata builder.

vllm_ascend/attention/sfa_v1.py
```diff
@@ -234,15 +248,84 @@ def build(
...
+                    torch.from_numpy(li_seq_lens)
+                    .pin_memory()
+                    .to(dtype=torch.int32, device=self.device, non_blocking=True)
+                )
+                li_skipped_query_mask = (
+                    torch.from_numpy(li_skipped_query_mask)
+                    .pin_memory()
+                    .to(dtype=torch.bool, device=self.device, non_blocking=True)
+                )
+                common_attn_metadata.lightning_indexer_metadata = AscendLightningIndexerMetadata(
```
YzTongNiar: Refactor 'AscendLightningIndexerMetadata' as 'lightling_indexer_context' (just like 'dsa_cp_context') to avoid building an extra metadata.

vllm_ascend/attention/sfa_v1.py
```diff
@@ -234,15 +248,84 @@ def build(
         cum_query_lens = common_attn_metadata.query_start_loc[1 : num_reqs + 1]
         seq_lens = common_attn_metadata.seq_lens[:num_reqs]
 
+        query_start_loc = common_attn_metadata.query_start_loc[: num_reqs + 1]
+        tokens = query_start_loc[1:] - query_start_loc[:-1]
```
YzTongNiar, line 255: should rename 'tokens' as 'num_computed_tokens' to clarify its usage.

vllm_ascend/attention/sfa_v1.py
```diff
@@ -234,15 +248,84 @@ def build(
...
+        # Split tokens evenly across TP devices, recompute per-device
+        # actual_seq_lengths_query and actual_seq_lengths_key.
+        # When combined with lightning_indexer_skip, operates on the
+        # already-reordered cum_query_lens/seq_lens and uses skip-aware
+        # key_lens computation.
+        # ====================================================================
         dsa_cp_context = None
         if self.enable_dsa_cp:
+            num_of_non_skip_tokens = 0
+            num_segs_for_cp = cum_query_lens.shape[0]
```
YzTongNiar: should rename 'num_segs_for_cp' as 'num_segs' since it does not depend on cp actually.

weiguihua2:
Have the DCP and PCP scenarios been verified? Code modifications for DCP and PCP should also be involved in this. @1024daniel

vllm_ascend/attention/sfa_v1.py
```diff
@@ -276,10 +359,15 @@ def build(
...
-            actual_seq_lengths_query = self.actual_seq_lengths_query
-            actual_seq_lengths_key = self.actual_seq_lengths_key
+            if skip:
+                actual_seq_lengths_query = cum_query_lens
+                actual_seq_lengths_key = seq_lens
+            else:
+                actual_seq_lengths_query = self.actual_seq_lengths_query
+                actual_seq_lengths_key = self.actual_seq_lengths_key
+            seg_is_skip = np.zeros(num_segs, dtype=bool)
+            seg_is_skip[num_reqs:] = True
```
clrs97: The skip tensor is not necessary since it is equal to `i < num_reqs`.

vllm_ascend/attention/sfa_v1.py
```diff
@@ -314,6 +405,17 @@ def build(
                 actual_seq_lengths_query=actual_seq_lengths_query,
                 actual_seq_lengths_key=actual_seq_lengths_key,
             )
+        if self.enable_lightning_indexer_skip and self.enable_dsa_cp:
```
clrs97, line 404: We only need to do a slice here instead of construction.

vllm_ascend/attention/sfa_v1.py
```diff
@@ -915,37 +1022,71 @@ def indexer_select_post_process(
...
-        if HAS_TRITON:
-            q_li = rope_forward_triton_siso(
-                q_li, cos, sin, rope_dim=self.qk_rope_head_dim, is_neox_style=self.is_rope_neox_style
-            )
+        sparse_count = 2048
+
+        # =========================
+        # step1: determine token range
+        # =========================
+        if attn_metadata.skip:
```
clrs97, line 1042: Multiple branches can be merged by slicing.

vllm_ascend/attention/utils.py
1024daniel: has refactor the code to remove the lightning_indexer_metadata

vllm_ascend/attention/sfa_v1.py
1024daniel: has refactor the code to remove the lightning_indexer_metadata

vllm_ascend/attention/sfa_v1.py
1024daniel: has refactor the code to just compute topk indices here

vllm_ascend/attention/sfa_v1.py
1024daniel: when num_of_non_skip_tokens is 0, we cannot just slice

1024daniel:
> Have the DCP and PCP scenarios been verified? Code modifications for DCP and PCP should also be involved in this. @1024daniel
have not test with DCP and PCP. These test will be test with future work

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1385,9 +1389,24 @@ def execute_model(
...
+
+            li_skip = li_reorder_indices is not None
+            if li_skip:
+                input_ids, positions = maybe_pad_and_reorder_inputs(
+                    input_ids, positions, li_reorder_indices
+                )
             hidden_states = self._model_forward(
                 num_tokens_padded, input_ids, positions, intermediate_tensors, inputs_embeds, **model_kwargs
             )
+            if li_skip:
```
weijinqian0, line 1407: The naming is not OK.

yiz-liu (commented):
I don’t think model_runner_v1 is the right place for this logic. It leaks SFA/lightning-indexer-specific behavior into the generic runner path, adds brittle coupling to per-layer metadata, and makes the optimization harder to reason about and validate across other execution modes. Could we keep the runner untouched and handle this inside `sfa_v1` instead?

PR #7422, opened 2026-03-18
[main][Attention][Feature] Add support for Qwen3-next hybrid attention in piecewise & full_decode_only

845473182:
This PR adds support for the Qwen3-next model, which utilizes a hybrid attention mechanism. The changes primarily focus on adapting the context parallelism (PCP) and graph mode execution paths to handle this new attention type. Key modifications include:
- Passing new metadata (`max_num_tokens_across_pcp`, `total_num_scheduled_tokens`) for handling padded tokens in flash attention.
- Refactoring the query, key, and value tensor preparation in `attention_cp.py` to support hybrid attention logic.
- Adding support for multiple KV cache groups in `model_runner_v1.py` and `pcp_utils.py`, which is required for hybrid attention models.
- Updating PCP utilities to correctly handle token splitting and metadata generation for the new model.

These changes are necessary to enable efficient inference for the Qwen3-next model on Ascend hardware with context parallelism and graph mode enabled.

No. This is a backend change to support a new model.


yiz-liu (approved):
I checked `attention_cp.py` and I suspect it conflicts with `FULL` mode fundamentally, separate code paths for prefill and decode, just like MLA. However, I think it's fine if we are only using `PIECEWISE` and `FULL_DECODE_ONLY` for PCP.

PR #7436, opened 2026-03-18
[release] Add GLM5 known issue for 2-node PD mixed deployment

MrZ20:
Documented an issue in the 2-node PD mixed deployment scenario where inference may hang when concurrency exceeds 8.(GLM5)

Noted that the issue has been fixed in PR:
- #7235 
- #7290.

- vLLM version: v0.17.0
- vLLM main: https://github.com/vllm-project/vllm/commit/4497431df654e46fb1fb5e64bf8611e762ae5d87

### DIFF
--- a/docs/source/user_guide/release_notes.md
+++ b/docs/source/user_guide/release_notes.md
@@ -51,7 +51,7 @@ This is the first release candidate of v0.17.0 for vLLM Ascend. Please follow th
 
 - GLM5 requires transformers==5.2.0, and this will resolved by [vllm-project/vllm#30566](https://github.com/vllm-project/vllm/pull/30566), will not included in v0.17.0.
 - There is a precision issue with Qwen3-Next due to the changed tp weight split method. Will fix it in next release.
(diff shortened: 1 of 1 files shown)


docs/source/user_guide/release_notes.md
```diff
@@ -52,6 +52,7 @@ This is the first release candidate of v0.17.0 for vLLM Ascend. Please follow th
 - GLM5 requires transformers==5.2.0, and this will resolved by [vllm-project/vllm#30566](https://github.com/vllm-project/vllm/pull/30566), will not included in v0.17.0.
 - There is a precision issue with Qwen3-Next due to the changed tp weight split method. Will fix it in next release.
 - The minimum number of tokens of prefix cache hit in hybrid model is 2k now
```
MengqingCao:
```suggestion
- The minimum number of tokens of prefix cache hit in hybrid model is large now. The exact number is related to tp size, e.g., with tp 2, the block_size is adjusted to 2048, which means that any prefix shorter than 2048 will never be cached.
```

PR #7456, opened 2026-03-19
[Doc][KV Pool]Revision KV Pool User Guide [2/2]

Pz1116:
Revise the KV Pool user guide:
4. Revise parameters for Memcache for better clarity, at notification that currently heterogeneous protocol setting is not supported (e.g. enable `device_rdma` and `device_sdma` at the same time, a example scenario would be data transfer by memcache across different super pods)
5. Modify the condition for Mooncakestore warmup, warmup is now needed only when `ASCEND_BUFFER_POOL` is enabled.

- vLLM version: v0.17.0
- vLLM main: https://github.com/vllm-project/vllm/commit/8a680463fab3bc9e6760417cd5c0a6aa58283065

### DIFF
--- a/docs/source/user_guide/feature_guide/kv_pool.md
+++ b/docs/source/user_guide/feature_guide/kv_pool.md
@@ -470,6 +470,7 @@ ock.mmc.config_store.tls.decrypter.path =
 ```
 
 **Key Focuses：**
+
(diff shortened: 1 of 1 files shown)


LCAIZJ (approved):
LGTM

PR #7458, opened 2026-03-19
[Feat] Support MXFP8 rollout for veRL on Ascend 950

zhijie-os:

Enables MXFP8 rollout in veRL v0.8.0 when using Ascend 950 devices (DV100 and DV120).

**Key Changes:**

1. verl Integration: Implements the `restore_weights_for_rl_loading` method (required by veRL but previously missing in the Ascend backend).  Additional `_mxfp8_transformed` flag is attached, allowing for the correct sequential execution of `process_weights_after_loading` under the W8A8 MXFP8 config.
2. Attribute Protection: Adds quantization configuration checks to prevent resetting attributes of `w13` and `w2` weights when quantization is enabled.

No additional user-facing change

Tested by running Qwen3 Dense and MoE models on DV120.

Dense: Qwen3-06B, Qwen3-32B

MoE: Qwen3-30B-A3B

`export PYTORCH_NPU_ALLOC_CONF="expandable_segments:False` because expandable_segments are not supported in memory pool when invoking vllm using verl.

- vLLM version: v0.17.0
- vLLM main: https://github.com/vllm-project/vllm/commit/8a680463fab3bc9e6760417cd5c0a6aa58283065

### DIFF
--- a/vllm_ascend/quantization/methods/w8a8_mxfp8.py
+++ b/vllm_ascend/quantization/methods/w8a8_mxfp8.py
@@ -89,10 +89,80 @@ class AscendW8A8MXFP8DynamicLinearMethod(AscendLinearScheme):
         return output
 
     def process_weights_after_loading(self, layer):
+        """Process weights after loading for MXFP8 inference.
+
+        This method transforms weights for NPU MXFP8 computation:
+        - weight: (output_size, input_size) -> (input_size, output_size)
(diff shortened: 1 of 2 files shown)


weijinqian0:
1. The update_weights API of the community 31943 is reused, and the transpose logic is added. (The transpose flag determines whether to perform transpose.)
2. The transpose flag is added to the process_after_loading function of the related module.

PR #7459, opened 2026-03-19
[Bugfix] Fix the bug of MTP1 crashing in multiple concurrent scenarios.

PHOEBEMOON0802:
The triton operator does not perform boundary checks on the global position within the loop, leading to the memory overflow in scenarios with multiple concurrency + 1-step  MTP launch.

Solution: Add a check that global_pos < vec_len, and strictly limit the boundaries of all memory accesses to avoid out-of-bounds writes.

No.

- vLLM version: v0.17.0

Signed-off-by: chenyue1122 <oyoy7102@163.com>

### DIFF
--- a/vllm_ascend/ops/triton/reject_sample.py
+++ b/vllm_ascend/ops/triton/reject_sample.py
@@ -56,7 +56,6 @@ def rejection_greedy_sample_spec_len_1_triton(
 
     draft_token_id = tl.load(draft_token_ids_ptr + offset, mask)
     target_argmax_id = tl.load(target_argmax_ptr + offset, mask)
-    # Use batch_mask when writing to output_token_ids to avoid out-of-bounds errors
     tl.store(output_token_ids_ptr + offset * 2, target_argmax_id, mask)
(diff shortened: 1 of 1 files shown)


vllm_ascend/ops/triton/reject_sample.py
```diff
@@ -48,26 +48,33 @@ def rejection_greedy_sample_spec_len_1_triton(
...
     bonus_token_ids_ptr,
     vec_len,
+    batch_size,  # adapt: add batch_size parameter for validating the boundaries of output_token_ids
     BLOCK_SIZE: tl.constexpr,
 ):
     block_idx = tl.program_id(0)
     offset = block_idx * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
-    mask = offset < vec_len
+    # adapt: Double mask: simultaneously check num_tokens and batch_size (for output_token_ids)
+    mask_token = offset < vec_len
```
AlbertCheese: vec_len 实际就是batch

PR #7486, opened 2026-03-19
Qwen3.5 MoE supports flash comm

wangbj127:
Multimodal models like Qwen3.5 MoE does embedding in model_runner, so when flash comm is enabled, the first `AllGather` operation should be skipped.

No.

- vLLM version: v0.18.0

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -196,9 +196,6 @@ class NPUPlatform(Platform):
     @classmethod
     def apply_config_platform_defaults(cls, vllm_config: VllmConfig) -> None:
         """Apply Ascend-specific defaults. Set sp_min_token_num=1 when enable_sp and not set."""
-        default_max_cg_capture_size = cls._get_default_max_cudagraph_capture_size(vllm_config)
-        if default_max_cg_capture_size is not None:
-            vllm_config.compilation_config.max_cudagraph_capture_size = default_max_cg_capture_size
         pass_config = vllm_config.compilation_config.pass_config
         if pass_config.enable_sp and pass_config.sp_min_token_num is None:
             from vllm_ascend.compilation.passes.sequence_parallelism import get_sp_min_token_num


vllm_ascend/ops/register_custom_ops.py
```diff
@@ -37,13 +37,17 @@ def _maybe_chunk_residual_impl(x: torch.Tensor, residual: torch.Tensor) -> torch
     return residual
 
 
-def _maybe_all_gather_and_maybe_unpad_impl(x: torch.Tensor, label: bool, is_ep_comm: bool = False) -> torch.Tensor:
+def _maybe_all_gather_and_maybe_unpad_impl(
+    x: torch.Tensor, label: bool, is_ep_comm: bool = False, is_first_allgather: bool = False
```
realliujiaxu: Allgather will be skipped when `label` is True. 
Please use `label` instead of `is_first_allgather`.

Note: The name `label` may be unclear. Consider using a more descriptive name, such as `skip_allgather` or `should_skip_allgather`, for better readability.

vllm_ascend/ops/register_custom_ops.py
```diff
@@ -97,7 +103,12 @@ def _maybe_pad_and_reduce_impl(x: torch.Tensor, is_ep_comm: bool = False) -> tor
         return get_ep_group().reduce_scatter(padded_x.view(-1, *x.shape[1:]), 0)
 
 
-def _maybe_all_gather_and_maybe_unpad_fake(x: torch.Tensor, label: bool, is_ep_comm: bool = False) -> torch.Tensor:
+def _maybe_all_gather_and_maybe_unpad_fake(
+    x: torch.Tensor, label: bool, is_ep_comm: bool = False, is_first_allgather: bool = False
```
realliujiaxu: Ditto.

vllm_ascend/ops/linear_op.py
```diff
@@ -430,8 +432,8 @@ def apply_impl(self, input_: torch.Tensor) -> torch.Tensor | tuple[torch.Tensor,
 
         # Matrix multiply.
         assert self.quant_method is not None
-
-        input_ = torch.ops.vllm.maybe_all_gather_and_maybe_unpad(input_, True)
+        need_all_gather = not (extract_layer_index(self.layer.prefix) == 0 and is_vl_model())
```
realliujiaxu: Will this cause the allgather before MLP to be skipped as well?

realliujiaxu (approved):

vllm_ascend/ops/linear_op.py
wangbj127: Yes, so we need to add a condition. Thanks!

PR #7487, opened 2026-03-20
[Feature]  Optimize Qwen3.5/Qwen3Next GDN prefill by prebuilding chunk metadata

maoxx241:

This PR optimizes the Qwen3.5 and Qwen3Next GDN prefill path on Ascend by reducing host/device synchronization overhead.

The current implementation of the `chunk_gated_delta_rule` path for variable-length sequences prepares chunk metadata during the forward pass. This approach triggers frequent CPU intervention and host/device round-trips. When running prefill-heavy workloads with asynchronous scheduling enabled, these synchronizations result in execution "bubbles" and prefill stalling (stuttering). **Note that this does not cause asynchronous scheduling to fail; rather, it prevents the system from reaching its theoretical throughput due to these unnecessary stalls.**

To resolve this, the patch moves metadata preparation out of the hot path:
- **Prebuilt Metadata:** All non-speculative varlen chunk metadata for GDN is now prebuilt on the CPU.
- **Asynchronous Transfer:** Staging buffers are kept in pinned memory and transferred to the NPU asynchronously.
- **Integration:** The prebuilt bundle is attached to GDN attention metadata via `patch_gdn_attn.py` and passed into Triton wrappers.
- **Backward Compatibility:** Triton wrappers fall back to the legacy preparation path if no prebuilt metadata is provided.

The scope of these changes is limited to the Qwen3.5 and Qwen3Next GDN paths on Ascend.

There are no API changes.

This is a pure performance optimization for the Ascend backend. For Qwen3.5/Qwen3Next workloads utilizing the GDN non-spec prefill path—particularly those using asynchronous scheduling and speculative decoding—this PR improves efficiency by removing H2D/D2H interactions from the critical path.

The output remains numerically equivalent to the legacy path. Internal validation shows that the speculative acceptance rate remains stable and consistent with baseline results.

**Unit Tests:**
- `python3 -m pytest -q tests/ut/patch/worker/patch_common/test_patch_gdn_attn.py -vv`

**Performance Benchmarks:**
- Models: `Qwen3.5-0.8B`, `Qwen3.5-35B-A3B`, `Qwen3.5-397B-A17B`
- Configuration:
  - `async_scheduling=True`
  - `compilation_config={"cudagraph_mode":"FULL_DECODE_ONLY"}`
  - `speculative_config='{"method": "qwen3_5_mtp", "num_speculative_tokens": 3, "enforce_eager": true}'`

**Validated vLLM Version:**
- Commit: `b31e9326a7d9394aab8c767f8ebe225c65594b60`

**Observed Impact:**
- Significant performance gains in prefill-heavy and high-batch-size (BS) scenarios.
- Improved throughput and reduced TTFT across all tested Qwen3.5 models.
- Speculative acceptance rate remained stable.

---

- vLLM version: v0.17.0

### DIFF
--- a/vllm_ascend/patch/worker/patch_gdn_attn.py
+++ b/vllm_ascend/patch/worker/patch_gdn_attn.py
@@ -17,6 +17,12 @@ from dataclasses import dataclass
 
 import torch
 import vllm.v1.attention.backends.gdn_attn as gdn_attn
+from vllm.v1.utils import CpuGpuBuffer
+
+_GDN_CHUNK_SIZE = 64
+# Keep this aligned with solve_tril.LARGE_BLOCK_T in ops/triton/fla/solve_tril.py.
+_GDN_SOLVE_TRIL_LARGE_BLOCK_SIZE = 608 * 2
+_GDN_CUMSUM_WORKING_SET = 2**18
 
 _IS_PATCHED = False
 _ORIGINAL_BUILD = gdn_attn.GDNAttentionMetadataBuilder.build
@@ -24,10 +30,10 @@ _ORIGINAL_BUILD = gdn_attn.GDNAttentionMetadataBuilder.build
 
 @dataclass
 class GDNChunkedPrefillMetadata:
-    chunk_indices_64: torch.Tensor
-    chunk_offsets_64: torch.Tensor
-    update_chunk_offsets_64: torch.Tensor
-    final_chunk_indices_64: torch.Tensor
+    chunk_indices_chunk64: torch.Tensor
+    chunk_offsets_chunk64: torch.Tensor
+    update_chunk_offsets_chunk64: torch.Tensor
+    final_chunk_indices_chunk64: torch.Tensor
     chunk_indices_large_block: torch.Tensor
     block_indices_cumsum: torch.Tensor
     _buffer_slot: object | None = None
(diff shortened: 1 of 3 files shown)


vllm_ascend/patch/worker/patch_gdn_attn.py
```diff
@@ -0,0 +1,317 @@
...
+import torch
+import vllm.v1.attention.backends.gdn_attn as gdn_attn
+
+_IS_PATCHED = False
+_ORIGINAL_BUILD = gdn_attn.GDNAttentionMetadataBuilder.build
+
+
+@dataclass
+class GDNChunkedPrefillMetadata:
+    chunk_indices_64: torch.Tensor
```
MengqingCao: Does all the 64 refers to int64 or chunck size?

vllm_ascend/patch/worker/patch_gdn_attn.py
```diff
@@ -0,0 +1,317 @@
...
+    update_chunk_offsets_64: torch.Tensor
+    final_chunk_indices_64: torch.Tensor
+    chunk_indices_large_block: torch.Tensor
+    block_indices_cumsum: torch.Tensor
+    _buffer_slot: object | None = None
+
+
+@dataclass
+class _GDNChunkedPrefillBufferSlot:
+    chunk_indices_64_cpu: torch.Tensor
```
MengqingCao: I think it is better to use `CpuGpuBuffer` instead of creating cpu and device tensor seperately?

vllm_ascend/patch/worker/patch_gdn_attn.py
```diff
@@ -0,0 +1,317 @@
...
+    )
+
+
+def _ensure_chunk_meta_state(builder, device: torch.device) -> None:
+    if getattr(builder, "_ascend_gdn_chunk_meta_initialized", False):
+        return
+    builder._ascend_gdn_chunk_meta_initialized = True
+    builder._ascend_gdn_chunk_meta_device = device
+    builder._ascend_gdn_chunk_size = 64
+    builder._ascend_gdn_large_block_size = 608 * 2
```
MengqingCao: what's the meaning of `608*2` here?

vllm_ascend/patch/worker/patch_gdn_attn.py
maoxx241: This suffix referred to the 64-token chunk size, not int64. I renamed the metadata fields to *_chunk64 to make that explicit.

vllm_ascend/patch/worker/patch_gdn_attn.py
maoxx241: Replaced the raw literal with _GDN_SOLVE_TRIL_LARGE_BLOCK_SIZE and added a comment tying it to solve_tril.LARGE_BLOCK_T.

vllm_ascend/patch/worker/patch_gdn_attn.py
maoxx241: I think so. That's much cleaner. I've refactored the implementation to use CpuGpuBuffer in the latest commit. Thanks!

Issue #5915, opened 2026-01-15
[Bug]: Transposing during wake_up causing MoE shape mismatch in RL scenarios.


Shangwei-Li:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
Collecting environment information...
PyTorch version: 2.8.0+cpu
Is debug build: False

OS: Debian GNU/Linux 12 (bookworm) (x86_64)
GCC version: (Debian 12.2.0-14+deb12u1) 12.2.0
Clang version: Could not collect
CMake version: version 3.25.1
Libc version: glibc-2.36

Python version: 3.11.9 (main, Dec 26 2024, 11:08:39) [GCC 12.2.0] (64-bit runtime)
Python platform: Linux-5.10.135.bsk.6-amd64-x86_64-with-glibc2.36

CPU:
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Address sizes:                   46 bits physical, 57 bits virtual
Byte Order:                      Little Endian
CPU(s):                          180
On-line CPU(s) list:             0-179
Vendor ID:                       GenuineIntel
Model name:                      Intel(R) Xeon(R) Platinum 8457C
CPU family:                      6
Model:                           143
Thread(s) per core:              2
Core(s) per socket:              45
Socket(s):                       2
Stepping:                        8
BogoMIPS:                        5199.63
[...]

PR #7498, opened 2026-03-20
[Fix][RL] Refactor weight transpose for RL.

Shangwei-Li:

Solving https://github.com/vllm-project/vllm-ascend/issues/5915. This patch moves `transpose` out of `wake_up` which is not promised to be called for reloading of MoE models, and provides a universal solution for transposed weight reloading.

Tested with verl 0.7.0 and vllm-ascend 0.14.rc1 on Qwen3-30B-Base
image


weijinqian0:
1. The update_weights API of the community 31943 is reused, and the transpose logic is added. (The transpose flag determines whether to perform transpose.)
2. The transpose flag is added to the process_after_loading function of the related module.

weijinqian0 (approved):

vllm_ascend/worker/worker.py
```diff
@@ -214,26 +214,7 @@ def wake_up(self, tags: list[str] | None = None) -> None:
...
-        hidden_size = self.vllm_config.model_config.hf_text_config.hidden_size
         model = self.model_runner.model
-        if tags is None or "weights" in tags:
-            for name, param in model.named_parameters():
-                if "w2_weight" in name and param.shape[2] == hidden_size:
-                    parts = name.split(".")
-                    param_name = parts[-1]
-                    parent_module = model.get_submodule(".".join(parts[:-1]))
-
-                    w2_data = param.transpose(1, 2)
```
momo609, line 226: w2_data is a uncontinguous tensor?

Shangwei-Li: `w2_data` and `w13_data` is uncontinguous during weight update. After weight updating, they are contiguous again after `process_weights_after_loading`.

PR #7506, opened 2026-03-20
[BugFix][P/D] fix padding error on FullGraph mode && fix layerwise connector mamba accuracy

nwpu-zxr:
1. When the FullGraph mode is used, the branches in the Triton operator are compiled and fixed during the graph capture process, causing the branch condition in the `fused_recurrent_gated_delta_rule` operator, which checks whether `ssm_state_indices >= 0` before writing to the SSM cache, to become invalid. Now, the write operation is performed regardless of the value. This results in the operator performing address offset calculations and writing to the SSM cache based on the -1 offset after -1 is used for padding in vLLM GDN backend. Since the conv cache and SSM cache in vLLM Ascend implementation are actually a single continuous tensor divided into two parts, this leads to data overwriting and the generation of NaN values.
This PR addresses two cases where padding -1 is required in the GDN metadata builder. The same logic is used to replace the padding with 0 to avoid the problem of memory overwriting, because block 0 is a reserved block.
2. Fix layerwise connector bug for mamba cache sending on heterogeneous TP.

No.

Start vLLM qwen3.5 service, preferably in the pd disaggregation mode with MTP >= 1, enable the FullGraph mode, and compile only the largest graph to perform the GSM8k precision test.

- vLLM version: v0.17.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2152,13 +2152,6 @@ class NPUModelRunner(GPUModelRunner):
             if for_cudagraph_capture:
                 attn_metadata_i = builder.build_for_cudagraph_capture(common_attn_metadata)
             else:
-                if isinstance(builder, GDNAttentionMetadataBuilder):
-                    # TODO(zxr): The fill_0 operation below is intended to prevent unexpected kvcache write
-                    #  operations of the Mamba Triton operator in FullGraph mode. However, this issue seems
-                    #  to have been resolved in vLLM 0.18.0. These operations should be verified and removed
-                    #  once it is confirmed that no unexpected kvcache writes will occur anymore.
-                    builder.spec_state_indices_tensor.fill_(0)
-                    builder.non_spec_state_indices_tensor.fill_(0)
                 attn_metadata_i = builder.build(
                     common_prefix_len=cascade_attn_prefix_len,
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2157,6 +2157,12 @@ def _build_attn_group_metadata(
                     common_attn_metadata=common_attn_metadata,
                     **extra_attn_metadata_args,
                 )
+                # NOTE(zxr): Due to the Triton operator does not deal with -1 padding in FullGraph mode,
+                # the padding needs to be changed from -1 to 0 to avoid writing invalid mamba block.
+                if self.vllm_config.compilation_config.cudagraph_mode.has_full_cudagraphs() \
+                    and isinstance(builder, GDNAttentionMetadataBuilder) and attn_metadata_i.num_prefills == 0:
+                    if attn_metadata_i.num_decodes == 0 and attn_metadata_i.num_spec_decodes > 0:
+                        attn_metadata_i.spec_state_indices_tensor[attn_metadata_i.num_spec_decodes:].fill_(0)
```
MengqingCao, line 2165: This is just a workround. We'll fix it in `fused_recurrent_gated_delta_rule_fwd_kernel` in the future, just like the kernel in vLLM, we need to handle the pad value in the kernel.

PR #7511, opened 2026-03-20
[Lint]Add lint hooks for clang-format, shellcheck, forbidden imports, and boolean context manager checks

MrZ20:
This PR introduces several upstream `vllm`-aligned lint hooks into `vllm-ascend` and makes them part of the actual `pre-commit` flow.

Main changes in this PR:
- add `check-boolean-context-manager` to catch boolean expressions in `with` statements
- add `check-forbidden-imports` to forbid direct `re` imports and disallowed direct `triton` imports
- enable shell script linting through `tools/shellcheck.sh`
- add root `.clang-format` aligned with upstream `vllm`, enable `clang-format` in `pre-commit`, temporarily **exclude all `csrc/**`** from `clang-format` to avoid bringing a large native code reformat into this PR

This PR focuses on landing the smaller and immediately useful lint alignment first, without mixing in the larger requirements-management migration.

No.

This PR only updates repository lint configuration, static checks, and internal import/style enforcement. It does not change runtime behavior or public interfaces.

Tested locally in the project virtual environment.

Commands used:
```bash
bash format.sh
```
Verified checks passed:
``` bash
ruff check...............................................................Passed
ruff format..............................................................Passed
codespell................................................................Passed
typos....................................................................Passed
clang-format.............................................................Passed
Lint GitHub Actions workflow files.......................................Passed
Lint shell scripts.......................................................Passed
Lint PNG exports from excalidraw.........................................Passed
Check for spaces in all filenames........................................Passed
Enforce __init__.py in Python packages...................................Passed
Check for forbidden imports..............................................Passed
Check for boolean ops in with-statements.................................Passed
Suggestion...............................................................Passed
- hook id: suggestion
- duration: 0s

To bypass pre-commit hooks, add --no-verify to git commit.
```
**note:**
clang-format is enabled but currently excludes all csrc/**

- vLLM version: v0.17.0

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,7 +41,7 @@ repos:
   hooks:
   - id: clang-format
     exclude: ^csrc/.*
-    types_or: [c++, cuda]
+    types_or: [c++]
     args: [--style=file, --verbose]
 
 - repo: https://github.com/igorshubovych/markdownlint-cli
--- a/tools/check_boolean_context_manager.py
+++ b/tools/check_boolean_context_manager.py
@@ -17,7 +17,19 @@
 # This file is a part of the vllm-ascend project.
 # Adapted from https://github.com/vllm-project/vllm/tree/main/tools
 #
-"""Lint: detect `with a() and b():` or `with a() or b():`."""
+"""Lint: detect `with a() and b():` (boolean op in with-statement context).
+
+Using `and`/`or` to combine context managers is almost always a bug:
+
+    with ctx_a() and ctx_b():   # BUG: only ctx_b is entered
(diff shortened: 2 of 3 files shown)


.clang-format
```diff
@@ -0,0 +1,26 @@
+BasedOnStyle: Google
+UseTab: Never
+IndentWidth: 2
```
menogrey, line 3: Do we really use this clang-format config? Maybe you should check with relative people who work on C++ code. The key point is that we don't have to follow the upstream config cause we have our code format rules.

MrZ20: This is part of the plan. Currently, I'm only providing the interface and have temporarily blocked the `crsc` folder. More specialized personnel will later modify the rules and C code.

PR #7514, opened 2026-03-20
[P/D] [Bugfix] fix mooncake layerconnector dead when update_decoder_info fail

liziyu179:
Fix mooncake layerconnector dead when update_decoder_info fail. For the scenario where node D is dead, node P failing to update_decoder_info should not cause node P to become dead.

by CI

- vLLM version: v0.17.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_layerwise_connector.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_layerwise_connector.py
@@ -1628,7 +1628,7 @@ class MooncakeLayerwiseConnectorWorker:
                 try:
                     req_meta_update = self.update_decoder_info(req_id, req_meta)
                 except Exception as e:
-                    logger.error(
+                    logger.warning(
                         f"MooncakeLayerwiseConnector transfer fail for req_id {req_id} in layer_idx "
                         f"{self.current_layer}, update_decoder_info with error: {e}"
                     )


vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_layerwise_connector.py
```diff
@@ -1625,7 +1625,14 @@ def save_kv_layer(
             for req_id, req_meta in connector_metadata.requests.items():
                 if len(req_meta.local_block_ids[layer_group_idx]) == 0:
                     continue
-                req_meta_update = self.update_decoder_info(req_id, req_meta)
+                try:
+                    req_meta_update = self.update_decoder_info(req_id, req_meta)
+                except Exception as e:
+                    logger.error(
```
wangxiyuan: do not use `error` level if the error is ignored.

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_layerwise_connector.py
```diff
@@ -1625,7 +1625,14 @@ def save_kv_layer(
             for req_id, req_meta in connector_metadata.requests.items():
                 if len(req_meta.local_block_ids[layer_group_idx]) == 0:
                     continue
-                req_meta_update = self.update_decoder_info(req_id, req_meta)
+                try:
+                    req_meta_update = self.update_decoder_info(req_id, req_meta)
+                except Exception as e:
+                    logger.warning(
```
zzzzwwjj, line 1631: Pls add an annoation afterwards on why this error needn't to throw an exception.

PR #7528, opened 2026-03-21
[Platform][BugFix] Preserve hybrid block size on Ascend

maoxx241:

This PR fixes a startup regression for Ascend hybrid attention + mamba models after upgrading to vLLM `0.18.0`.

The failure shows up during engine initialization with errors like:

```text
ValueError: None of the kernel sizes [128] can divide physical block size 16 evenly
```

The affected path is the hybrid KV cache setup used by Qwen3.5 models on Ascend.

### Root cause

For Ascend hybrid attention + mamba models, we already have model-specific block size logic in `patch_mamba_config.py`.

That logic computes a larger attention block size so that the attention page size is compatible with the mamba page size. For example:

- Qwen3.5-0.8B: attention block size is raised to `512`
- Qwen3.5-35B-A3B: attention block size is raised to `1024`

However, after the vLLM `0.18.0` upgrade, worker initialization still calls the generic platform hook:

- `current_platform.update_block_size_for_backend(vllm_config)`

In the upstream base implementation, the hybrid branch does this:

- if `user_specified_block_size == True`, keep the current value
- otherwise, reset `cache_config.block_size` to the default block size `16`

That means the Ascend-specific hybrid block size computed earlier can be overwritten later in worker init.

Once it is reset to `16`, the downstream KV cache path uses that value to build `FullAttentionSpec`, and Ascend eventually reaches `BlockTable` with:

- physical block size = `16`
- supported attention kernel block sizes = `[128]`

Since `16` is not divisible by `128`, startup fails.

### Why CI did not catch it

This bug is real, but our current CI setup accidentally masks it.

The key detail is that the Ascend E2E helper `VllmRunner` has this default:

- `block_size=16`

So even when the test does not explicitly care about block size, the helper passes `block_size=16` into `LLM(...)`.

That has an important side effect in vLLM:

- `CacheConfig` marks `user_specified_block_size = True`

Once that flag is set, the generic `update_block_size_for_backend()` path returns early and does **not** overwrite the current block size again.

So the CI path effectively becomes:

1. hybrid model-specific logic computes the Ascend-compatible block size
2. test harness sets `user_specified_block_size=True`
3. generic backend block-size update is skipped
4. startup succeeds

But real startup paths that do **not** explicitly pass `block_size` do not get that protection, so they still fail.

In other words, CI was green because the test harness was unintentionally hiding the bug.

This is also why the issue initially looked subtle:
- it was reproducible in ad-hoc startup
- it was not reproduced by the existing helper-based E2E flow
- and it was unrelated to `speculative_config.enforce_eager=True`

### How this PR fixes it

This PR keeps the fix strictly inside `vllm-ascend`.

It adds an Ascend override for `NPUPlatform.update_block_size_for_backend()`:

- for hybrid models, do not run the generic upstream block-size fallback
- preserve the block size that was already computed by the hybrid model-specific config logic
- for non-hybrid models, keep the original upstream behavior unchanged

This is the smallest safe fix because:

- it does not change upstream vLLM
- it does not special-case a single model name
- it does not change `BlockTable`
- it does not widen Ascend attention kernel support
- it only prevents the hybrid block size from being clobbered after it was already computed correctly

### Tests / validation

Added UT coverage for the exact regression:

- verify that `NPUPlatform.update_block_size_for_backend()` preserves the block size for hybrid models when the block size was not explicitly user-specified

Validated on remote with local weights and no explicit `block_size` override:

- `Qwen3.5-0.8B`: startup succeeds
- `Qwen3.5-35B-A3B`: startup succeeds

Also re-ran relevant platform unit tests:

- `tests/ut/test_platform.py -k preserves_hybrid_block_size`
- `tests/ut/test_platform.py -k cache_config_block_size`

### Scope

This PR is intentionally narrow.

It only fixes the Ascend-side hybrid block-size regression introduced by the interaction between:

- Ascend hybrid block-size alignment logic

It does not change any GDN runtime logic or speculative decoding behavior.

- vLLM version: v0.18.0

### DIFF
--- a/tests/ut/test_platform.py
+++ b/tests/ut/test_platform.py
@@ -317,6 +317,16 @@ class TestNPUPlatform(TestBase):
 
         self.assertEqual(vllm_config.cache_config.block_size, 1024)
 
+    def test_update_block_size_for_backend_preserves_user_block_size(self):
+        vllm_config = TestNPUPlatform.mock_vllm_config()
+        vllm_config.model_config.is_hybrid = False
+        vllm_config.cache_config.block_size = 512
+        vllm_config.cache_config.user_specified_block_size = True
+
+        self.platform.update_block_size_for_backend(vllm_config)
+
+        self.assertEqual(vllm_config.cache_config.block_size, 512)
+
     @patch("vllm_ascend.quantization.utils.maybe_auto_detect_quantization")
     @patch("vllm_ascend.utils.get_ascend_device_type", return_value=AscendDeviceType.A3)
     @patch("vllm_ascend.ascend_config.init_ascend_config")


vllm_ascend/platform.py
```diff
@@ -172,6 +172,16 @@ def get_device_uuid(cls, device_id: int = 0) -> str:
     def inference_mode(cls):
         return torch.inference_mode()
 
+    @classmethod
+    def update_block_size_for_backend(cls, vllm_config: VllmConfig) -> None:
+        model_config = vllm_config.model_config
```
MengqingCao, line 181: As `cache_config.user_specified_block_size` is introduced in v0.18.0, I prefer to keep the user specified block size at the first order.
```suggestion
        cache_config = vllm_config.cache_config
        if cache_config.user_specified_block_size:
            # User specified --block-size; keep it.
            return
        model_config = vllm_config.model_config
```

PR #7536, opened 2026-03-23
[Bugfix][CI] Fix aisbench installation to avoid Gitee authentication

zhangxinyuehfad:
- Pass GITEE_USERNAME (var) and GITEE_TOKEN (secret) as Docker build
  args in nightly image build so Dockerfile can authenticate to Gitee
- In Dockerfile.nightly.a2/a3, embed credentials into clone URL to
  avoid auth failure during `git clone`
- In single-node and multi-node PR test workflows, backup the
  pre-installed benchmark from the nightly image before wiping
  vllm-ascend, then restore it instead of re-cloning from Gitee,
  which is inaccessible from fork PR contexts

- vLLM version: v0.18.0

### DIFF
--- a/.github/workflows/dockerfiles/Dockerfile.nightly.a2
+++ b/.github/workflows/dockerfiles/Dockerfile.nightly.a2
@@ -20,6 +20,8 @@ FROM quay.io/ascend/vllm-ascend:main
 ARG PIP_INDEX_URL="https://mirrors.tuna.tsinghua.edu.cn/pypi/web/simple"
 ARG AIS_BENCH_TAG="v3.0-20250930-master"
 ARG AIS_BENCH_URL="https://gitee.com/aisbench/benchmark.git"
+ARG GITEE_USERNAME=""
+ARG GITEE_TOKEN=""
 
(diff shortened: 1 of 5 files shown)


.github/workflows/dockerfiles/Dockerfile.nightly.a2
```diff
@@ -35,7 +37,8 @@ RUN export PIP_EXTRA_INDEX_URL=https://mirrors.huaweicloud.com/ascend/repos/pypi
     python3 -m pip cache purge
 
 # Install benchmark tools
-RUN git clone -b ${AIS_BENCH_TAG} --depth 1 ${AIS_BENCH_URL} /vllm-workspace/vllm-ascend/benchmark && \
+RUN CLONE_URL=$(echo "${AIS_BENCH_URL}" | sed "s|https://|https://${GITEE_USERNAME}:${GITEE_TOKEN}@|") && \
```
Potabk, line 40: This may not be a robust way, maybe we should clone the benchmark dir in the workflow and `COPY`  it to the docker image

PR #7544, opened 2026-03-23
[Spec-Decode] Fix spec decode proposer in 0.18.0

MengqingCao:
As the vllm-ascend main doesn't maintain v0.17.0 now, we'd just apply the single branch in eagle proposer. Otherwise it will raise error in v0.18.0

CI passed with existing test.

- vLLM version: v0.18.0


wangxiyuan:
Please enable the test in the follow-up PR

PR #7546, opened 2026-03-23
[BugFix][310p]Handle null quantization config in ShardedStateLoader310

csoulnd:
---

Fixes 310P sharded-state saving / weight-compression metadata in `ShardedStateLoader310`. **VL models** often expose `quant_config` only on **`language_model`**, not on the multimodal root, so inferring from the root `model` was wrong; we now pass **`vllm_config.quant_config`** from `NPUWorker310`. **`generate_quant_description`** takes optional **`quant_config`**, handles **`None`** as FLOAT, and reads **`quant_description`** via **`getattr`** when present.

---

**No** — internal bugfix / robustness only.

---

- vLLM version: v0.18.0


vllm_ascend/_310p/worker_310p.py
```diff
@@ -48,7 +48,8 @@ def save_sharded_state(
             max_size=max_size,
         )
 
-        ShardedStateLoader310.generate_quant_description(self.model_runner.model, path)
+        quant_config = self.vllm_config.quant_config
+        ShardedStateLoader310.generate_quant_description(self.model_runner.model, path, quant_config)
```
pu-zhe: 建议直接写为：
```
ShardedStateLoader310.generate_quant_description(
    self.model_runner.model,
    path,
    self.vllm_config.quant_config
)
```
简洁直观

vllm_ascend/_310p/sharded_state_loader_310p.py
```diff
@@ -48,10 +49,18 @@ def save_model(
...
+        quant_config: QuantizationConfig | None = None,
+    ) -> None:
         """Generate a mapping of parameter names to their corresponding quantization types."""
         quant_description = {}
-        quantize_type = model.quant_config.quant_description.get("model_quant_type", "FLOAT")
+        if quant_config is None:
+            quantize_type = "FLOAT"
+        else:
+            qd = getattr(quant_config, "quant_description", None) or {}
+            quantize_type = qd.get("model_quant_type", "FLOAT")
```
pu-zhe: 直接写为：
```
try:
    quantize_type = quant_config.quant_description.get("model_quant_type", "FLOAT")
except AttributeError:
    quantize_type = "FLOAT"
```

PR #6979, opened 2026-03-04
[Triton][Sampler]add penalty-related triton kernel for better performance of penalties

linfeng-yuan:
### What this PR does / why we need it?
Implement `get_token_bin_count` and `apply_penalties` with triton kernels. It can significantly reduce latency of sampling process with penalties.

### Does this PR introduce _any_ user-facing change?
No.

### How was this patch tested?
E2E testing with Qwen3-480B (accuracy and perforamnce, update later)

- vLLM version: v0.16.0
- vLLM main: https://github.com/vllm-project/vllm/commit/15d76f74e2fdb12a95ea00f0ca283acf6219a2b7

PR #7569, opened 2026-03-23
[Triton][Sampler] Add penalty-related Triton kernel for better performance of penalties

realliujiaxu:
# [Triton][Sampler] Add penalty-related Triton kernel for better performance of penalties

Implement `get_token_bin_counts_and_mask` and `apply_penalties` with Triton-Ascend kernels. This significantly reduces latency of the sampling process when repetition/frequency/presence penalties are enabled.

Ref: #6979

---

### Benchmark config
- **Model**: Qwen3-0.6B config (vocab_size=151936)
- **Output**: 10 decode steps
- **repetition_penalty**: 1.05
- **warmup**: 10, **iterations**: 20

### Results

Numbers below are **avg over 10 decode steps**, where each step is averaged over **20 iterations**.

case (seqs, prompt_len, output_len) | This PR vllm_ascend.sample.penalties | main vllm.v1.sample.ops.penalties | pr #6979 vllm_ascend.sample.sampler
-- | -- | -- | --
(1, 65536, 8192) | 1.23 | 3.27 | 17.73
(4, 65536, 8192) | 3.29 | 11.72 | 18.99
(8, 16384, 4096) | 2.76 | 10.90 | 6.71
(32, 4096, 1024) | 3.05 | 13.17 | 3.37

</div></div><div class="ui-scroll-area__scrollbar" data-scrollable="true" role="presentation" aria-hidden="true" style="position: absolute; z-index: 10; inset: auto 0px 0px; width: auto; pointer-events: auto; height: 6px;"><div class="ui-scroll-area__thumb" style="position: absolute; inset: 0px auto 0px 0px; min-height: auto; width: 493px; border-radius: 9999px; background: none 0% 0% / auto repeat scroll padding-box border-box color(srgb 0.266667 0.301961 0.337255 / 0.08); pointer-events: auto; transform: translateX(207px); transition: background-color 0.15s, opacity 0.15s; height: 6px; min-width: 20px; opacity: 0;"></div></div></div><br class="Apple-interchange-newline">

No.

---

## How was this patch tested?

- E2E: `tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_apply_penalties_triton.py`
- Manual benchmark (Qwen3-0.6B)


vllm_ascend/ops/triton/bincount.py
```diff
@@ -0,0 +1,127 @@
...
+    tokens_ptr,
+    tokens_batch_stride,
+    tokens_seq_stride,
+    batch_size,
+    seq_len,
+    vocab_size,
+    bin_counts_ptr,
+    counts_batch_stride,
+    counts_vocab_stride,
+    BIG_CORE_NUM: tl.constexpr,
```
linfeng-yuan: These type hints treat `BIG_CORE_NUM` and `BIG_ROW_BLOCK_SIZE` as `tl.constexpr` would lead to recompilation problems when these values are updated and first met when calling. Should remove it.

vllm_ascend/ops/triton/bincount.py
```diff
@@ -0,0 +1,127 @@
...
+
+    big_row_block_size = triton.cdiv(n_rows, core_num)
+    big_core_num = max(
+        0,
+        min(
+            core_num - (big_row_block_size * core_num - n_rows),
+            core_num,
+        ),
+    )
+    grid = (min(n_rows, core_num),)
```
linfeng-yuan: Would be better if we can take the `seq_len` dim into account while doing tiling, bcause there are usually input sequences with small `bs` during prefill stage. Doing tiling with dim 0 only would lead to low-utilization among all vector cores.

vllm_ascend/ops/triton/bincount.py
realliujiaxu: done

vllm_ascend/ops/triton/bincount.py
realliujiaxu: done

PR #7572, opened 2026-03-23
[Graph][Bugfix] Set default cudagraph max capture size via platform defaults

linfeng-yuan:

This PR lets NPU platform provide its own default `max_cudagraph_capture_size` via `NPUPlatform.apply_config_platform_defaults()`.

Previously, when cudagraph sizing was left unset, Ascend inherited vLLM's upstream default heuristic in `_set_cudagraph_sizes()`, which uses `max_num_seqs * decode_query_len * 2`. This PR changes Ascend's default to `min(max_num_seqs * decode_query_len, 512)` while keeping the rest of vLLM's cudagraph sizing logic unchanged.

Yes, but only for Ascend when users do not explicitly configure cudagraph sizing.

If `max_cudagraph_capture_size` and `cudagraph_capture_sizes` are both unset, we now uses `max_num_seqs * decode_query_len` (capped at `512`) instead of the upstream `* 2` default. Explicit user settings are unchanged.

Add unit tests to cover:

- default max injection via `apply_config_platform_defaults()`
- explicit `max_cudagraph_capture_size` is preserved
- explicit `cudagraph_capture_sizes` are preserved
- Ascend default max no longer uses the upstream `* 2`
- late `_set_cudagraph_sizes()` recomputation reuses the current max input

- vLLM version: v0.18.0

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -193,10 +193,8 @@ class NPUPlatform(Platform):
     @classmethod
     def apply_config_platform_defaults(cls, vllm_config: VllmConfig) -> None:
         default_max_cg_capture_size = cls._get_default_max_cudagraph_capture_size(vllm_config)
-        if default_max_cg_capture_size is None:
-            return
-
-        vllm_config.compilation_config.max_cudagraph_capture_size = default_max_cg_capture_size
+        if default_max_cg_capture_size is not None:
+            vllm_config.compilation_config.max_cudagraph_capture_size = default_max_cg_capture_size
 
     @classmethod
     def get_device_capability(cls, device_id: int = 0):


linfeng-yuan:
This PR eliminates the `*2` behavior when setting `max_cudagraph_capture_size`, which may lead to performance degradation of  batched prefill requests with small tokens (e.g., 256~512) in  `FULL` mode. In these cases, users might set `max_cudagraph_capture_size` manually.

yiz-liu:
> This PR eliminates the `*2` behavior when setting `max_cudagraph_sizes`, which may lead to performance degradation of  batched prefill requests with small tokens (256~512) in  `FULL` mode.

This indeed is a problem, maybe we need to consider including both MC2 and all gather in the graphs.

PR #7596, opened 2026-03-24
[Refactor] Use forward mapping instead of reverse mapping in AscendModelSlimConfig

Feng-xiaosuo:

This PR refactors the `AscendModelSlimConfig` class to use **forward mapping** instead of reverse mapping for quantization config key transformation.

**Changes:**
1. Modified `apply_vllm_mapper()` to directly apply `hf_to_vllm_mapper.apply_dict()` to transform `quant_description` keys from HF format to vLLM format
2. Simplified `quant_prefix_mapper()` to return the prefix directly (no longer needs mapping since keys are already in vLLM format)
3. Removed `QUANT_MODEL_PREFIX_MAPPINGS` dictionary (~50 lines) - no longer needed
4. Removed `get_prefix_mapping()` function - no longer needed
5. Removed `vllm_to_hf_mapper` attribute - no longer needed

**Why this change is needed:**

The previous implementation used reverse mapping (vLLM → HF) which had several issues:
- Some keys might not be used in the forward direction but would be incorrectly used in reverse
- Empty values in the mapping would cause issues when reversed
- Required maintaining a separate `QUANT_MODEL_PREFIX_MAPPINGS` dict that duplicated information already available in vLLM's model-specific `WeightsMapper`

The new approach:
- Uses the forward mapping (HF → vLLM) directly from vLLM's `WeightsMapper`
- Eliminates the need for duplicate mapping definitions
- Avoids issues with reverse mapping (unused keys, empty values)
- Aligns with how `compressed_tensors_config.py` handles the same scenario

No. This is an internal refactoring that does not change any user-facing API or behavior.

- The changes were validated using an offline deployment script to launch and verify multiple multimodal models, including Qwen2.5-VL, Qwen3-VL, Kimi2.5, and Qwen-Omni.
- vLLM version: v0.18.0


MengqingCao:
plz make a backport on v0.18.0 branh

PR #7630, opened 2026-03-25
[bugfix]fixed block_size incorrect setting issue in dsv3.2

kunpengW-code:
https://github.com/vllm-project/vllm/pull/35122 This PR in the vllm community refactors the update mode of block_size. As a result, when the user does not specify `--block-size`, dsv3.2 obtains an incorrect block_size.

**The root cause of the problem is analyzed from the block_size update process as follows:**
1. In NPUPlatform, `check_and_update_config` calls `refresh_block_size` to set block_size to 128.
2. During Modelrunner initialization, the `self.block_size` parameter is generated. At this time, block_size is still 128. This parameter will be used for operations such as kvcache initialization.
3. `update_block_size_for_backend` updates block_size to the size set in attn_backend. The reason why the DSV3.2 is faulty is that it has an additional attn_backend `DeepseekV32IndexerBackend`, and this backend is not rewritten. The block_size obtained from attn_backend is 64. In this case, only `vllm_config.cache_config.block_size` is updated, and other parts are not modified. As a result, the block_size on the entire network is inconsistent.

**Modification solution:**
Skip `update_block_size_for_backend` and modify block_size only in the `check_and_update_config` method.

In the future, the block_size update logic can be migrated to the `update_block_size_for_backend` method. Ensure that all block_size values on the entire network are updated.
no

- vLLM version: v0.18.0


rjg-lyh (approved):

MengqingCao (approved):
LGTM, thx for this fix!

PR #7662, opened 2026-03-26
[CI][v0.18.0] Build nightly image for releases/v0.18.0 per pr

zhangxinyuehfad:

This patch add per pr image build for branch `releases/v0.18.0`, Due to the limitations of the quay naming convention, we should not name the image tag the same as branch name, we name the image tag`releases-v0.18.0` for daily build.

### DIFF
--- a/.github/workflows/_schedule_image_build.yaml
+++ b/.github/workflows/_schedule_image_build.yaml
@@ -47,6 +47,7 @@ jobs:
       with:
         fetch-depth: 0
         persist-credentials: false
+        ref: ${{ github.ref }}
 
     - uses: actions/checkout@v6
       if: ${{ github.event_name == 'workflow_dispatch' }}
@@ -113,6 +114,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
+        with:
+          ref: ${{ github.ref }}
 
       - name: Download arm64 digests
         uses: actions/download-artifact@v8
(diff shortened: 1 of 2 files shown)


.github/workflows/_schedule_image_build.yaml
```diff
@@ -154,9 +157,9 @@ jobs:
           #      which follow the rule from vLLM with prefix v
           # TODO(yikun): the post release might be considered as latest release
           tags: |
+            type=ref,event=branch,pattern=v0.18.0-dev,suffix=${{ env.SUFFIX }}
```
Potabk:
```suggestion
            type=ref,event=branch,suffix=${{ env.SUFFIX }}
            type=ref,event=pr,suffix=${{ env.SUFFIX }}
            type=pep440,pattern={{raw}},suffix=${{ env.SUFFIX }}
```

Potabk (changes requested):

.github/workflows/schedule_image_build_and_push.yaml
```diff
@@ -36,10 +56,14 @@ on:
...
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   image_build:
     name: Image Build and Push
-    if: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'image-build')  }}
+    if: ${{ github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'ready') }}
```
Potabk:
```suggestion
    if: ${{ github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'image-build') }}
```
the label `image-build` is more reasonable

.github/workflows/_schedule_image_build.yaml
Potabk: The final tag of the branch: `releases-v0.13.0-a3-openeuler`, why not `releases/v0.13.0-a3-openeuler`, this is for the quay' image tag name must not contain the `/`

PR #7717, opened 2026-03-27
[BugFix][Main][KV Pool]Fix KV transfer Put Logic

Pz1116:
Before when we do put for KV Pool, we find the first non-existing key and put all the blocks starting from that index; however, if the prefix cache blocks is from another request, and some of the blocks are evicted due to LRU, we will be putting blocks that still exist in the pool, and causing MooncakeStore printing unnecessary logs in master service.

co-authored-by: Skywalker-EP 173723846@qq.com

What this PR does:
1. Now we lookup all the keys and only put the ones that are missing.
2. Fix `lookup_scheduler` in `pool_worker` so it handles GQA correctly.
3. Fixes a few existing typos
4. Add UT, written by codex

- vLLM version: v0.18.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/kv_transfer.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/kv_transfer.py
@@ -366,7 +366,7 @@ class KVCacheStoreLayerSendingThread(KVTransferThread):
 
         logger.info(
             "Storing KV cache for %d out of %d blocks (missing_count=%d) for request %s",
-            len(keys),
+            len(key_list),
             total_block,
             len(missing_indices),
             req_meta.req_id,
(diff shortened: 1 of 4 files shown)


Pz1116:
please take a look @LCAIZJ @fems14 @baxingpiaochong

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/kv_transfer.py
```diff
@@ -88,22 +88,22 @@ def _handle_request(self, req_meta: Any):
...
+        :return: A bool list where True means the key exists in store.
         """
         try:
             res = self.m_store.exists(keys)  # type: ignore[assignment]
+            exists_list = [False] * len(keys)
             for index, value in enumerate(res):  # type: ignore[arg-type]
-                if value != 1:
-                    return index
-            # all tokens where found, return the maximal end
+                if index >= len(exists_list):
```
LCAIZJ: The condition for this if statement can never be satisfied?

Pz1116: right, this part is unnecessary, let me get rid of it

LCAIZJ:
https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py#L610 Let's fix the issue on this line as well.

foraxe:
Tests OK.

LCAIZJ (approved):
LGTM

PR #7729, opened 2026-03-27
[Bug][0.18.0] Added a validation check to prevent unsupported head dimensions of MHA

SunnyLee151064:

Validation Logic: Added a validation check to prevent unsupported head dimensions when using high tensor parallelism on Ascend platforms.
Error Handling: Introduced a ValueError to inform users about known limitations with torch_npu.npu_fused_infer_attention_score for specific head sizes.

No


MengqingCao:
we won't fix it in this way, let's wait for fia to support more head dim

PR #7756, opened 2026-03-28
[Feature] Optimize GDN non-spec prefill fallback metadata

maoxx241:

This PR optimizes the GDN non-spec prefill fallback path for the Ascend Qwen3.5 / Qwen3Next linear-attention backend.

Background:
- `causal_conv1d` in the current non-spec prefill path still needs host-side metadata, and the old path prepared that metadata lazily in forward.
- `chunk_gated_delta_rule` also relied on repeated hot-path `prepare_*` helpers (`prepare_chunk_indices`, `prepare_chunk_offsets`, etc.) instead of consuming a step-scoped prebuilt metadata bundle.
- The fallback contract was also too permissive: when required non-spec prefill metadata was missing, the code could silently fall back to legacy recomputation paths instead of failing explicitly.

Changes in this PR:
- Introduce a unified `non_spec_prefill_fallback_meta` for the GDN non-spec prefill path.
  - `causal_conv1d` host metadata is prepared during attention metadata build and consumed in forward.
  - `chunk_gated_delta_rule` prebuilt metadata is also prepared during metadata build and threaded through `gdn.py -> chunk.py`.
- Add `vllm_ascend/ops/triton/gdn_chunk_meta.py` to build:
  - `chunk_indices_chunk64`
  - `chunk_offsets_chunk64`
  - `update_chunk_offsets_chunk64`
  - `final_chunk_indices_chunk64`
  - `chunk_indices_large_block`
  - `block_indices_cumsum`
- The builder-side chunk-meta path now has two execution modes:
  - CPU path: generate metadata with the same semantics as the original runtime `prepare_*` helpers.
  - NPU path: write directly into pooled device tensors, reuse shared `seq_lens`, and avoid per-call helper fallback logic in the hot path.
- Thread prebuilt `chunk_offsets` through `chunk.py`, `chunk_o.py`, and `chunk_o_update.py` so wrapper code no longer re-prepares offsets when prebuilt metadata is available.
- Tighten the patched prefill contract:
  - remove legacy silent fallback branches that re-materialize metadata in unexpected paths
  - raise explicitly when required non-spec prefill fallback metadata is missing
- Keep metadata semantics aligned with the original runtime path for mixed spec/non-spec, padding, zero-length, and long-sequence cases.

Why this is needed:
- moves non-spec prefill fallback metadata generation to the metadata-builder stage instead of doing it lazily inside per-layer forward hot paths
- makes the fallback contract explicit and debuggable
- keeps `chunk_gated_delta_rule` operator inputs consistent with the original `prepare_*` path while reducing repeated hot-path preparation work
- preserves correctness for mixed/padded edge cases while improving TTFT / TPS on the target workload

There is no API or interface change.

User-visible impact is limited to backend behavior on Ascend:
- the GDN non-spec prefill path is stricter and will now fail loudly if the expected fallback metadata contract is violated
- performance for the targeted Qwen3.5 / Qwen3Next GDN prefill path is improved on the validated benchmark shape

Unit / correctness coverage:
- `tests/ut/patch/worker/patch_common/test_patch_gdn_attn.py`
- `tests/ut/ops/test_gdn_chunk_meta.py`
- `tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_gdn_chunk_meta.py`

The tests cover:
- `causal_conv1d` host args matching the original/runtime path
- chunk metadata parity with the original runtime `prepare_*` helpers
- CPU and NPU chunk-meta generation parity
- mixed spec/non-spec, padding, zero-length, and long-sequence edge cases
- strict failure when required non-spec prefill fallback metadata is missing
- prebuilt `chunk_offsets` wiring through the FLA wrappers

Additional validation:
- compared the current prebuilt `chunk_gated_delta_rule` metadata against the original non-prebuilt path on both CPU and NPU; for the tested long-sequence / mixed cases, the operator inputs were identical

Performance validation:
- methodology: for each code state, start the service once, run 5 benchmark rounds against the same warm service, and report the mean of the last 4 rounds
- model: `Qwen3.5-35B-A3B`
- target: `910B4 32G`
- benchmark config:
  - `tensor_parallel_size=4`
  - `--max-model-len 4096`
  - `--gpu-memory-utilization 0.9`
  - `--async-scheduling`
  - `--compilation-config '{"cudagraph_mode":"FULL_DECODE_ONLY","cudagraph_capture_sizes":[4,8,12,16]}'`
  - `--speculative-config '{"method":"qwen3_5_mtp","num_speculative_tokens":3}'`
  - `num_prompts=64`
  - `max_concurrency=16`
  - `output_len=1500`
- comparison states:
  - baseline: pre-PR state `dae3c99e`
  - original PR implementation: `cc3ca6f7`
  - current implementation in this PR: `8000a085`

Results (mean of last 4 rounds):

| State | TTFT | TPS |
| --- | ---: | ---: |
| baseline (`dae3c99e`) | `1048.64 ms` | `162.26` |
| original (`cc3ca6f7`) | `1008.81 ms` | `163.60` |
| current (`8000a085`) | `978.58 ms` | `164.67` |

Delta:
- current vs baseline:
  - TTFT: `-70.06 ms`
  - TPS: `+2.42`
- current vs original:
  - TTFT: `-30.22 ms`
  - TPS: `+1.07`

Summary:
- relative to the original PR implementation, the latest scheme keeps improving TTFT and TPS

### DIFF
--- a/tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_gdn_chunk_meta.py
+++ b/tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_gdn_chunk_meta.py
@@ -1,0 +1,94 @@
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+# This file is a part of the vllm-ascend project.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import gc
+
+import pytest
+import torch
+
+from vllm_ascend.ops.triton.gdn_chunk_meta import build_chunk_meta_device
+from vllm_ascend.ops.triton.fla.utils import (
+    prepare_chunk_indices,
+    prepare_chunk_offsets,
+    prepare_final_chunk_indices,
+    prepare_update_chunk_offsets,
+)
+from vllm_ascend.ops.triton.triton_utils import init_device_properties_triton
+
+
+@pytest.mark.parametrize(
+    ("cu_seqlens_data", "chunk_size", "input_dtype", "output_dtype"),
+    [
+        ([0, 4, 7], 4, torch.int32, torch.int32),
+        ([0, 4, 4, 12], 4, torch.int32, torch.int32),
+        ([0, 128, 1024, 1024, 4096], 64, torch.int64, torch.int32),
+    ],
+)
+@torch.inference_mode()
+def test_build_chunk_meta_device_correctness(
+    cu_seqlens_data: list[int],
+    chunk_size: int,
+    input_dtype: torch.dtype,
+    output_dtype: torch.dtype,
+):
+    init_device_properties_triton()
+    device = "npu"
+
+    cu_seqlens_cpu = torch.tensor(cu_seqlens_data, dtype=input_dtype)
+    cu_seqlens = cu_seqlens_cpu.to(device)
+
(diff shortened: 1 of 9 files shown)


weijinqian0:
The operator UT needs to be supplemented.

weijinqian0:
Supplementing the patch deletion plan.

weijinqian0 (approved):

PR #7762, opened 2026-03-28
[Feature] support qkv_rmsnorm_mrope for qwen3vl

jiangmengyu18:
Qwen3vl full attention supports enabling the split_qkv_rmsnorm_mrope fusion operator.

- [x] Run Qwen3-VL dense model with the fusion operator, verify correct output
- [x] Run Qwen3-VL MoE model with the fusion operator, verify correct output
- vLLM version: v0.18.0


weijinqian0:
The patch code needs to be added with a deletion plan.

PR #7764, opened 2026-03-28
[Triton] Fix triton-ascend version in Dockerfile

MengqingCao:
Triton-ascend occasionally encounters compilation errors, which is a known issue in triton-ascend 3.2.0. However, we want to use the official version rather than the development version, so we only changed the triton-ascend version in the Dockerfile and added a FAQ to explain this issue.

- vLLM version: v0.18.0


MengqingCao:
closing as wont fix in this way in branch main

PR #7789, opened 2026-03-28
[BugFix] fix bug that repetition_penalty is not effective in asynchronous scheduling

SumrainChan:

To fix the [issue:7769](https://github.com/vllm-project/vllm-ascend/issues/7769)

The GPUModelRunner updates the output_token_ids of sampling_metadata before each sample. However, the NPUModelRunner does not perform this step. As a result, the generated token cannot be obtained during apply_repetition_penalties, and the repetition_penalty is not effective when it is greater than 1.

No

- vLLM version: v0.18.0

### DIFF
--- a/tests/ut/worker/test_model_runner_v1.py
+++ b/tests/ut/worker/test_model_runner_v1.py
@@ -1,6 +1,6 @@
 import unittest
 from types import SimpleNamespace
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import torch
 from vllm.v1.kv_cache_interface import FullAttentionSpec, KVCacheConfig, KVCacheGroupSpec, KVCacheTensor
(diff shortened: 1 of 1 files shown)


realliujiaxu:
Thanks for your contribution! Let's merge it after CI pass.

realliujiaxu:
LGTM. Could you add a test case for this?

SumrainChan:
> Thanks for this bugfix! I have 2 suggestions.
> 
> 1. I checked the code and find this bug is introduced because we didn't adapt async scheduler + penalties before. And this is done in vllm by [vllm-project/vllm#26467](https://github.com/vllm-project/vllm/pull/26467).
> 
> And your fix seems lack of the logic of passing `logitsprocs_need_output_token_ids` by the `InputBatch`
> 
> 2. could you please add an test case of async scheduler + penalties scenario to prevent the problem from recurring?


Thanks for your suggestions. I will add logitsprocs_need_output_token_ids and test cases later.

Issue #6689, opened 2026-02-11
[Feature]: Verify / Support `Qwen/Qwen3-ASR-1.7B`


wjunLu:
### Description
Request to verify or support the Qwen3-ASR-1.7B automatic speech recognition model released in January 2026.

**Model Information:**
- **HuggingFace Path:** `Qwen/Qwen3-ASR-1.7B`
- **Release Date:** January 2026
- **Category:** Audio Processing Model (ASR)
- **Highlights:**
  - Specialized for automatic speech recognition
  - 1.7B parameter efficient architecture
  - High-accuracy transcription capabilities
  - Multi-language support

### Deliverables
1. **Test Cases:** Add comprehensive test cases for Qwen3-ASR-1.7B
   - Location: `tests/e2e/models/configs`
   - Include ASR-specific test scenarios
   
2. **Tutorial:** Create tutorial documentation
   - Location: `docs/source/tutorials/models`
   
3. **SKILL.md:** You must use AI assistance to complete this task and comment your agent SKILL.md or prompts used under this issue for reference
   - **What is SKILL:** A SKILL is a summary of lessons learned and best practices from the AI-assisted development process. It documents effective prompts, workflows, problem-solving approaches, and reusable patterns that can help others accomplish similar tasks more efficiently.
   - **SKILL.md format reference:** https://agentskills.io/specification

ascend-core:
@Marrry-png , You have not passed the open source internship qualification application, unable to claim this task. Please complete the [internship application](https://www.openeuler.org/en/internship/) first.

ascend-core:
@lihaokun-2026 , Thank you for claiming this task. Please communicate with the mentor in a timely manner. You can only undertake this task after the mentor approves, otherwise the task is invalid.

ascend-core:
@lihaokun-2026 , Congratulations, you have successfully claimed this task. Please handle the task in a timely manner. Claim task > Mentor reviews claim qualification > Process task > Submit task > Mentor reviews > PR merge > Earn points.

coolhuhu:
How is the model adaptation work for Qwen3-ASR-0.6B and 1.7B progressing? How can I join? Is there any relevant documentation available for reference?

ascend-core:
@lihaokun-2026 , 请关注您提交的pr审核进度, 跟进相关负责人审核, pr合入后可获得积分. 注: 提交pr时, 请务必在pr描述里添加此issue编号(#issue编号), 谢谢!

[...]

PR #7800, opened 2026-03-29
[Feature] Verify / Support Qwen/Qwen3-ASR-1.7B

lihaokun-2026:
### Overview

This document summarizes the AI-assisted debugging and evaluation workflow established while deploying and benchmarking the Qwen3-ASR-1.7B model through a local OpenAI-compatible API service. It serves as a reusable reference for handling model name mismatches, audio dataset loading issues, request formatting problems, context-length constraints, logging, and ASR benchmark design.

The workflow covers both functional deployment verification and baseline ASR evaluation, with the final validated setup successfully completing inference on 100 LibriSpeech samples and reporting WER/CER results.

#### Symptom:
The evaluation script failed during dataset iteration with:

ImportError: To support decoding audio data, please install 'torchcodec'.

#### Root Cause:
The datasets library attempted to automatically decode the audio column, but the required decoder backend was unavailable in the current environment.

#### Solution:
Disable automatic decoding and read the raw audio payload manually:
```python
dataset = dataset.cast_column("audio", Audio(decode=False))
```
Then load audio either from raw bytes or path using soundfile.

#### Reusable Pattern:
When evaluating ASR models in minimal containers or custom environments, prefer Audio(decode=False) and explicitly control the audio loading pipeline.

#### Symptom:
All samples failed with errors such as:

Error opening '6930-75918-0000.flac': System error.

#### Root Cause:
The audio["path"] returned by the dataset was only a relative filename, not a directly readable absolute path in the current runtime environment.

#### Solution:
Prefer reading from in-memory bytes when available:
```python
audio_info = sample["audio"]

if audio_info.get("bytes") is not None:
    audio_array, sampling_rate = sf.read(io.BytesIO(audio_info["bytes"]))
elif audio_info.get("path") is not None:
    audio_array, sampling_rate = sf.read(audio_info["path"])
else:
    raise ValueError("Invalid audio field")
```
Reusable Pattern:
For HF ASR datasets, always inspect the actual structure of the audio column before assuming path is directly readable.

### Challenge 3: Model Name Mismatch Between Client and Server

Symptom:
Every request failed with:

The model `qwen3-asr-1.7b` does not exist.

Root Cause:
The client request used the local directory name or model path, while the server exposed the model under a different alias through --served-model-name.

Solution:
Query the model list first:
```python
curl http://localhost:8000/v1/models
```

The server response showed:
```python
"id": "qwen3-asr"
```

So the request must use:
```python
MODEL_NAME = "qwen3-asr"
```
Reusable Pattern:
In OpenAI-compatible serving, always use the id returned by /v1/models in the client request, not the local model path.

### Challenge 4: Context-Length Error Caused by Output Token Setting

Symptom:
Requests failed with:

You passed 121 input characters and requested 448 output tokens. However, the model's context length is only 448 tokens

Root Cause:
The script set:

max_completion_tokens = 448

while the model’s maximum context length was also 448, leaving no room for the actual input.

Solution:
Reduce the generation length for ASR inference:

max_completion_tokens = 64
temperature = 0.0

Reusable Pattern:
For ASR transcription tasks, use small output budgets and deterministic decoding. Large generation limits are unnecessary and can cause immediate context overflow.
- vLLM version: v0.18.0

### DIFF
--- a/docs/source/tutorials/models/Qwen3-ASR-1.7B.md
+++ b/docs/source/tutorials/models/Qwen3-ASR-1.7B.md
@@ -63,7 +63,6 @@ In addition, if you don't want to use the docker image as above, you can also bu
 
 ```shell
 #!/bin/sh
-# Enable ModelScope download (optional, for domestic users)
 export VLLM_USE_MODELSCOPE=true
 
 # Start vLLM online service


lihaokun-2026:
[server.log](https://github.com/user-attachments/files/26328128/server.log)
[test_clean.log](https://github.com/user-attachments/files/26328129/test_clean.log)
[test_other.log](https://github.com/user-attachments/files/26328132/test_other.log)
image
image

lihaokun-2026:
The documentation build issues have been resolved and the basic checks are passing. Could you please help review this PR and approve the workflows? Thanks! @wjunLu
The issule is [6689](https://github.com/vllm-project/vllm-ascend/issues/6689)

wjunLu:
> The documentation build issues have been resolved and the basic checks are passing. Could you please help review this PR and approve the workflows? Thanks! @wjunLu The issule is [6689](https://github.com/vllm-project/vllm-ascend/issues/6689)

sovle your conflicts first

PR #7815, opened 2026-03-30
[BugFix] Fix norm_quant pass

### LINKED ISSUE #7811: [Bug]: The norm_quant pass does not take effect correctly when the bias in torch.ops._C.ascend.npu_add_rms_normbias() is None
### Your current environment

<details>

```text
Vllm ascend main branch
```

</details>

### 🐛 Describe the bug

When running the dots1 model in npuex_graph mode, we observed from profiling that the norm_quant pass is not taking effect.

After investigation, we found that the issue is caused by the bias argument being None in torch.ops._C.ascend.npu_add_rms_normbias().

This prevents the norm_quant pass from being properly applied.

We will submit a PR to address this issue.

Hibbert133:

The `norm_quant` pass does not take effect correctly when the `bias` input of  
`torch.ops._C.ascend.npu_add_rms_normbias()` is `None`.

Specifically, in this case, the pattern  
`AddRMSNormDynamicQuantPatternWithBias` in the `norm_quant` pass is matched at the operator level,  
but fails to be correctly activated when the `bias` argument is `None`. As a result, the optimization  
is not applied as expected.

Instead, this scenario should be handled by the  
`AddRMSNormDynamicQuantPattern`, which is designed for cases without bias.

### Proposed changes

- Update the condition in `AscendRMSNorm.forward_oot` in  
  `vllm-ascend/ops/layernorm.py`
- Ensure that the path calling  
  `torch.ops._C.ascend.npu_add_rms_normbias()` is only taken when `bias is not None`
- This guarantees correct pattern matching and enables the `norm_quant` pass properly

### Related Issue

Fixes #7811

---

N/A

---

The fix was validated through profiling.

We verified that after this change:
- The `norm_quant` pass is correctly applied
- The expected optimization behavior is observed in profiling results

- vLLM version: v0.27.1


Hibbert133:
Hi @zzzzwwjj, @realliujiaxu, @whx-sjtu, gentle ping on this PR when you have time.

This PR fixes the `norm_quant` path when `AscendRMSNorm.bias` is `None`. The patch is intentionally small: it prevents the bias-specific custom op from being selected without an actual bias, allowing the bias-free RMSNorm pattern to take effect.

The branch was refreshed on July 17, lint passes, and I verified the optimization through profiling. I noticed the automated review also raised a related case in `AscendGemmaRMSNorm`; I can include that case and add a regression test if maintainers prefer handling both paths in the same PR.

Could you please advise whether the current scope is acceptable, or whether the Gemma path and unit test should be included before review? Thanks!

PR #7841, opened 2026-03-31
[CI][Misc] modify ds3.2+dcp ci

weiguihua2:

Due to the current dcp solution of allgathering the KV cache, the performance deteriorates significantly, and the CI may get stuck. This PR temporarily removes the performance and accuracy benchmarks for DeepSeek-V3.2-W8A8-cp to prevent CI hangs until optimization is complete.

pcik-from:https://github.com/vllm-project/vllm-ascend/pull/7842

No

Verified that the configuration file remains valid and that the CI no longer attempts to run the problematic benchmarks.

pick-from: https://github.com/vllm-project/vllm-ascend/pull/7842


MengqingCao:
cc @rjg-lyh

PR #7858, opened 2026-03-31
[BugFix] Fix dimension mismatch error when SP padding causes num_tokens_padded != num_tokens_unpadded

huangfeifei1995:
Hardware: Huawei Ascend 910B
vLLM version: 0.17.0rc1
Model: Qwen3-Next-80B-A3B-Instruct
Parallel-config: TP4 DP1 
Enable Flashcomm: True
Max_num_seq: 64

This PR fixes a `RuntimeError` (dimension mismatch) that occurs when Sequence Parallelism (SP) is enabled and the padding added for SP causes `num_tokens_padded` to differ from `num_tokens_unpadded`. In such cases, `_pad_query_start_loc_for_fia` adds a dummy request, increasing `num_reqs_padded`. This mismatch between the actual number of requests and the padded number of requests leads to errors in downstream token count computations (e.g., `compute_num_computed_tokens`). 

The fix modifies the restrictive condition `num_tokens_padded == num_tokens_unpadded` when reverting the dummy request padding if SP is enabled, as SP padding is handled by stripping it after communication and should not be treated as an additional request in the attention metadata.

### Problem Description

When deploying Qwen-next 80B models with vllm-ascend, a RuntimeError occurs when request concurrency exceeds max_num_seqs with Flashcomm enabled:

`RuntimeError: The size of tensor a (64) must match the size of tensor b (65) at non-singleton dimension 0`

The error occurs in compute_num_computed_tokens(), where self.seq_lens has 64 elements while query_lens has 65 elements:
```
    def compute_num_computed_tokens(self) -> torch.Tensor:
        """Compute num_computed_tokens on device (seq_lens - query_lens)."""
        if self._num_computed_tokens_cache is None:
            query_lens = self.query_start_loc[1:] - self.query_start_loc[:-1]
            self._num_computed_tokens_cache = self.seq_lens - query_lens
        return self._num_computed_tokens_cache
```

### Root Cause Analysis
_pad_query_start_loc_for_fia function increases num_reqs_padded from 64 to 65 and writes data to position 65 in query_start_loc.
```
# Mixed-batch case: num_reqs must equal num_reqs_padded
            assert num_reqs == num_reqs_padded

            # Insert a dummy request instead of setting query_start_loc[num_reqs] = num_tokens_padded directly
            self.query_start_loc.np[num_reqs_padded + 1] = num_tokens_padded
            num_reqs_padded = num_reqs_padded + 1
```

this padding should be reverted when SP is enabled by following codes:
```
if enable_sp() and num_tokens_padded == num_tokens_unpadded:  # ← Problem here
    if num_reqs_padded > old_num_reqs_padded:
        num_reqs_padded = old_num_reqs_padded
        self.query_start_loc.np[num_reqs_padded + 1] = 0
```

However, when SP is enabled, _pad_for_sequence_parallelism pads tokens to multiples of TP size, which may cause num_tokens_padded != num_tokens_unpadded (in my case, num_tokens_padded =29292 which is multiples of 4, num_tokens_unpadded = 29290 which is not multiples of 4), so that num_reqs_padded remain 65 rather than 64.


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1339,10 +1339,22 @@ def execute_model(
...
+                    # != num_tokens_unpadded due to SP alignment (e.g., 29292 vs 29290).
+                    if enable_sp() and num_reqs_padded > old_num_reqs_padded:
+                        if num_tokens_padded == num_tokens_unpadded:
                             num_reqs_padded = old_num_reqs_padded
                             self.query_start_loc.np[num_reqs_padded + 1] = 0
+                        if num_tokens_padded != num_tokens_unpadded and not self.speculative_config:
+                            num_reqs_padded = old_num_reqs_padded
+                            self.query_start_loc.np[num_reqs_padded + 1] = 0
+                            self.query_start_loc.np[num_reqs_padded] = num_tokens_padded
+                            self.query_start_loc.gpu[num_reqs_padded] = num_tokens_padded
```
ZhaoJiangJiang, line 1423: why don't use `self.query_start_loc.copy_to_gpu()` method?

huangfeifei1995: Thanks for the question!
I initially used self.query_start_loc.copy_to_gpu() in the early version of this fix. However, during the code review, one of committers raised concerns that copy_to_gpu() may introduce potential synchronization issues or unnecessary overhead in this context.
Since we only need to update a single element at index num_reqs_padded, directly assigning to the GPU tensor via self.query_start_loc.gpu[num_reqs_padded] = num_tokens_padded is safer and more efficient. It avoids triggering a full host-to-device copy and eliminates any risk of implicit synchronization or race conditions that copy_to_gpu() might introduce.
Therefore, following the committer's suggestion, I switched to direct GPU tensor assignment to ensure correctness and performance.

ZhaoJiangJiang: > Thanks for the question! I initially used self.query_start_loc.copy_to_gpu() in the early version of this fix. However, during the code review, one of committers raised concerns that copy_to_gpu() may introduce potential synchronization issues or unnecessary overhead in this context. Since we only need to update a single element at index num_reqs_padded, directly assigning to the GPU tensor via self.query_start_loc.gpu[num_reqs_padded] = num_tokens_padded is safer and more efficient. It avoids triggering a full host-to-device copy and eliminates any risk of implicit synchronization or race conditions that copy_to_gpu() might introduce. Therefore, following the committer's suggestion, I switched to direct GPU tensor assignment to ensure correctness and performance.

Thanks for the detailed explanation! I see your point now.

PR #7873, opened 2026-03-31
[CI] runs-on/cache accelerates build speed

tfhddd:
Reduce the time spent building vllm-ascend when running test cases and prevent unnecessary use of computing power.

**Solution:** 
Utilize runs-on/cache to cache the compiled C++ code based on a hash of the relevant C++ code. If the C++ code remains unchanged, skip the compilation process and use the cache directly.

When merging a pull request, it will reduce the build time of vllm-ascend and speed up the execution time of the pull request CI.

Before cache takes effect：
image

After the cache takes effect
image

Full use cases：
image

- vLLM version: v0.18.0

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -2,6 +2,11 @@ name: 'e2e test'
 
 on:
   workflow_call:
+    secrets:
+      HW_OBS_AK:
+        required: false
+      HW_OBS_SK:
+        required: false
     inputs:
       vllm:
         required: true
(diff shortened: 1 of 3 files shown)


.github/workflows/_e2e_test.yaml
```diff
@@ -105,13 +127,37 @@ jobs:
           VLLM_TARGET_DEVICE=empty uv pip install -e .
           uv pip uninstall triton
 
+      - name: Get csrc hash
+        id: get_csrc_hash
+        run: |
+          CSRC_HASH=$(find ./csrc ./setup.py ./pyproject.toml ./CMakeLists.txt ./cmake \
+            ./vllm_ascend/envs.py ./requirements.txt ./packages.txt \
```
tfhddd: 这一行删除，这一行对于C编译来说不影响

tfhddd: fixed

PR #7875, opened 2026-03-31
[BugFix][main][KV Pool] Fix KV Pool not putting kv cache for vllm v0.18.0

Pz1116:
vLLM v0.18 defers KV connector finalization during target-modelforward when speculative decoding is enable, leading to KV Pool not doing Put Operation. This change is forgotten when we bumpped up the version for vllm-ascend. Fix by adding finalize_kv_connector for spec decode.

co-authored-by: Skywalker-EP 173723846@qq.com

- vLLM version: v0.18.0


LCAIZJ (approved):
LGTM

PR #7896, opened 2026-04-01
[Feature][PP] Support Dynamic Chunk for Chunked Pipeline Parallelism

gjc0824:

Chunked Pipeline Parallelism (CPP), with its relatively low communication overhead, has become a key feature for ultra-long sequences and variable-length sequence scenarios. However, ordinary CPP faces the problem of empty slots in long sequences as the KVCache grows, reducing computational efficiency.

We control the execution time of each scheduling batch to be equal by using a dynamic chunk strategy. Specifically, we profile requests of different lengths and fit the functional relationship between sequence length and execution time. During the scheduling process, we calculate the appropriate chunk size based on the given target execution time.

Specific modifications:

Patch modifications
1. Patch Scheduler to add dynamic chunking logic
2. Minimal patch to EngineCore to add logic for calling Profile

Changes on the vllm-ascend side
1. Add timing logic in worker & modelrunner, minor adaptation of dummy_run to CPP Profile
2. Create ProfileChunkManager to uniformly manage dynamic chunking resolution, function fitting, and other utility functions

- vLLM version: v0.19.0
- vLLM main: https://github.com/vllm-project/vllm/commit/620e8924d9c6b2a0b1d49ac0dcf2588fffcbe390

Performance Results:
- Fixed-Length Scene
256k Qwen3-235B chunk32k
TTFT
PP 61.4s -> CPP 53.5s, 15% benefit
- Variable-length scene
1k~32k
DeepSeek V3.1
CPP 
TTFT 9.9s, Throughput 23635 tps
PCP
TTFT 11.4s, Throughput 20734 tps

### DIFF
--- a/vllm_ascend/core/profiling_chunk_predictor.py
+++ b/vllm_ascend/core/profiling_chunk_predictor.py
@@ -23,7 +23,7 @@ latency and fitting a quadratic model.
 The approach:
 1. Profile: Run forward passes with different chunk sizes to measure latency
 2. Fit: Use quadratic model f(l) = a*l^2 + b*l + c to fit the latency data
-3. Predict: Given current history_len, solve for chunk size that achieves
+3. Predict: Given current num_computed_tokens, solve for chunk size that achieves
    target latency
 """
 
@@ -60,6 +60,19 @@ class ChunkSizePredictor:
         self.min_chunk = min_chunk
         self.history_fitted = False
 
+    def clamp_quadratic_and_linear_if_negative(self, fitted_a: float, fitted_b: float) -> tuple[float, float]:
+        """In theory, for the Transfomur structure of LLM, the fitted quadratic and linear
+        terms should not be negative. Can perform zero clamping for inaccurate fitting
+        """
+        if fitted_a < 0:
+            logger.warning("Fitted a=%.2e is not positive. Setting a=1e-9.", fitted_a)
+            fitted_a = 1e-9
+        if fitted_b < 0:
+            logger.warning("Fitted b=%.2e is not positive. Setting b=0.0.", fitted_b)
+            fitted_b = 1e-9
+
+        return fitted_a, fitted_b
+
     def fit(self, seq_lens: list[int], latencies: list[float]) -> bool:
         """Fit quadratic coefficients f(l) = al^2 + bl + c from data points.
 
@@ -68,8 +81,9 @@ class ChunkSizePredictor:
         """
         L = np.array(seq_lens, dtype=np.float64)
         T = np.array(latencies, dtype=np.float64)
+        MIN_FIT_POINTS_NO_CHUNK = 8
 
-        if len(L) < 8:
+        if len(L) < MIN_FIT_POINTS_NO_CHUNK:
             logger.warning(
                 "Not enough data points for quadratic fitting (%d < 8)",
                 len(L),
@@ -99,13 +113,7 @@ class ChunkSizePredictor:
                 logger.warning("Failed to fit quadratic model: %s", fallback_error)
                 return False
 
-        if fitted_a < 0:
-            logger.warning("Fitted a=%.2e is not positive. Setting a=1e-9.", fitted_a)
-            fitted_a = 1e-9
-
-        if fitted_b < 0:
-            logger.warning("Fitted b=%.2e is not positive. Setting b=0.0.", fitted_b)
-            fitted_b = 0.0
+        fitted_a, fitted_b = self.clamp_quadratic_and_linear_if_negative(fitted_a, fitted_b)
 
         self.quadratic_coeff_a = fitted_a
         self.linear_coeff_b = fitted_b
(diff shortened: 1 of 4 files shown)


vllm_ascend/core/profiling_chunk_predictor.py
```diff
@@ -0,0 +1,370 @@
...
+    def fit(self, seq_lens: list[int], latencies: list[float]) -> bool:
+        """Fit quadratic coefficients f(l) = al^2 + bl + c from data points.
+
+        Returns:
+            True if fitting succeeded, False otherwise
+        """
+        L = np.array(seq_lens, dtype=np.float64)
+        T = np.array(latencies, dtype=np.float64)
+
+        if len(L) < 8:
```
MengqingCao: why we constraint the length of `L` to 8 here?

vllm_ascend/core/profiling_chunk_predictor.py
```diff
@@ -0,0 +1,370 @@
...
+        return True
+
+    def fit_chunk(self, chunked_data: list) -> bool:
+        """Fit time with chunks: f(C,H) = a*C(C+H) + b*C + c*H.
+
+        Returns:
+            True if fitting succeeded, False otherwise
+        """
+        num_points = len(chunked_data)
+        if num_points < 5:
```
MengqingCao: Is `num_points < 5` an experience value?

vllm_ascend/core/profiling_chunk_predictor.py
```diff
@@ -0,0 +1,370 @@
...
+            logger.warning(
+                "Not enough data points for chunked data fitting (%d < 5)",
+                num_points,
+            )
+            return False
+        if num_points > 30:
```
MengqingCao: ditto, why when `num_points > 30` means fitted here?

vllm_ascend/core/profiling_chunk_predictor.py
```diff
@@ -0,0 +1,370 @@
...
+            logger.warning("Failed to fit chunked model: %s", e)
+            return False
+
+        if fitted_a < 0:
+            logger.warning("Fitted a=%.2e is not positive. Setting a=1e-9.", fitted_a)
+            fitted_a = 1e-9
+
+        if fitted_b < 0:
+            logger.warning("Fitted b=%.2e is not positive. Setting b=0.0.", fitted_b)
+            fitted_b = 0.0
```
MengqingCao: it seems the post process of a and b could be extract into a function.

vllm_ascend/patch/platform/patch_profiling_chunk.py
```diff
@@ -0,0 +1,179 @@
+#
```
MengqingCao, line 1: plz add the details of this patch like others in https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/patch/__init__.py

vllm_ascend/core/scheduler_profiling_chunk.py
```diff
@@ -0,0 +1,723 @@
...
+
+            # >>> PROFILING CHUNK: dynamic chunk sizing for RUNNING >>>
+            if (
+                self.profiling_chunk_manager is not None
+                and self.profiling_chunk_manager.is_ready
+                and num_new_tokens > 1
+                and request.num_computed_tokens > 0
+            ):
+                predicted_chunk = self.profiling_chunk_manager.predict_chunk_size(
+                    history_len=request.num_computed_tokens,
```
MengqingCao: I think it would be better to use consistent naming conventions for `num_computed_tokens` and `history_len`.

gjc0824:
> Thanks for your great work! Could you post more performance results on dynamic cpp and the design doc, usage doc, ~test cases~ should also be supplemented, I'm okay with adding them in a follow-up pr.

Sure. We plan to collect more test results and include them in the subsequent document PR. The subsequent PR will include interface specifications, user guides, and more performance data.

vllm_ascend/core/profiling_chunk_predictor.py
gjc0824: These values are all judgments of the amount of fitting data. For fitting the relationship between sequence length and time, a small amount of data can lead to inaccurate fitting. All are experience value. For clarity, I will create a variable name to represent.

vllm_ascend/core/profiling_chunk_predictor.py
gjc0824: ditto

vllm_ascend/core/profiling_chunk_predictor.py
gjc0824: Sure. We will extract it into a tool function.

vllm_ascend/core/scheduler_profiling_chunk.py
gjc0824: Sure.

vllm_ascend/patch/platform/patch_profiling_chunk.py
gjc0824: Sure.

vllm_ascend/core/profiling_chunk_predictor.py
gjc0824: ditto, addtionally, this is the online calibration stage, and users need to use several pieces of real data for preheating. It is not advisable to be too long here to affect the user experience, so the amount of fitting data is limited. The actual measurement of 30 data points has shown good fitting effect, and online calibration can be completed by consuming approximately 4 data points

vllm_ascend/core/scheduler_profiling_chunk.py
```diff
@@ -0,0 +1,723 @@
...
+            return float(result[0])
+        return None
+
+    # ------------------------------------------------------------------
+    # schedule() override
+    # ------------------------------------------------------------------
+    # The method below is based on the upstream Scheduler.schedule()
+    # with profiling-based chunk sizing applied to both RUNNING requests
+    # (chunked prefill continuation) and WAITING requests (new prefill).
+    # Modified sections are marked with ">>> PROFILING CHUNK" comments.
```
MengqingCao, line 231: This makes the code very clear! 👍

PR #7927, opened 2026-04-02
[CI] Clarify nightly test image tags

Potabk:
For nightly tests, there are too many messy variable passes running everywhere, which is confusing.
In my opinion, the only entry should be the tag `image=swr.cn-southwest-2.myhuaweicloud.com/base_image/ascend-ci/vllm-ascend:nightly-a3/a2`, the job itself should not be aware of branch information; its responsibility is simply to pull the image and execute tests. 
I've added a top-level variable management step where all the variables we use should be managed uniformly. This will allow us to more clearly observe what we're testing nightly and which branch's image we're using.

```shell
  setup-vars:
    name: Export global env vars as job outputs
    needs: parse-trigger
    runs-on: ubuntu-latest
    outputs:
      image: ${{ steps.vars.outputs.image }}
      cann_image: ${{ steps.vars.outputs.cann_image }}
      ascend_log_prefix: ${{ steps.vars.outputs.ascend_log_prefix }}
      vllm_version: ${{ steps.vars.outputs.vllm_version }}
      image_build_ref: ${{ steps.vars.outputs.image_build_ref }}
    steps:
      - name: Set variables
        id: vars
        run: |
          {
            echo "image=swr.cn-southwest-2.myhuaweicloud.com/base_image/ascend-ci/vllm-ascend:nightly-a2"
            echo "cann_image=swr.cn-southwest-2.myhuaweicloud.com/base_image/ascend-ci/cann:8.5.1-910b-ubuntu22.04-py3.11"
            echo "ascend_log_prefix=/root/.cache/ascend-logs"
            echo "vllm_version=v0.18.0"
            echo "image_build_ref=releases/v0.18.0"
          } >> "$GITHUB_OUTPUT"
```

- vLLM version: v0.18.0


.github/workflows/schedule_nightly_test_a2.yaml
zhangxinyuehfad, line 123: Delete it if you sure image.

PR #7943, opened 2026-04-02
[BugFix][xlite] Clamp block_size to 128 for xlite graph compatibility

SijieFu:

This PR fixes an `xlite` runtime issue when `cache_config.block_size` is not 128.

Changes:

- In `vllm_ascend/xlite/xlite.py`, clamp `xlite` `block_size` to 128 when the configured value is greater than 128, with warning logs.
- In `vllm_ascend/ascend_config.py`, replace the hard `RuntimeError` for non-128 block sizes with a warning to avoid failing early at init time.

Why:

- `xlite` currently behaves reliably with `block_size <= 128` (e.g., 16, 32, 63, 128), and `block_size = 256` can cause worker crash on first request (ACL stream synchronize failed: 507015).
- The previous strict check caused hard failure at startup for non-128 values and enforced user to explicitly pass `block_size=128` as an argument, while real behavior depends on value range.
- This change keeps startup resilient and applies a safe compatibility fallback for known-unsafe values.

Yes (behavioral):

- For `xlite` mode, when configured `block_size > 128`, `vllm_ascend.xlite` now logs warnings and automatically resets to 128 instead of proceeding with a value that may crash at runtime.
- Non-128 values no longer raise RuntimeError during config validation; a warning is emitted.

Environment notes:

- Test results were collected after skipping the `block_size` rectification in the function `vllm_ascend.utils.refresh_block_size` to isolate `xlite` behavior; this was to test `xlite`'s capabilities and the changes were not committed to this PR. 
-  `xlite` package was built from: <https://atomgit.com/openeuler/GVirt/commit/424ae8046a798eeca22956a002e015763f275de5?ref=master>.

Scripts:

```bash
# bring up the service
vllm serve /path/to/model \
--host 0.0.0.0 \
--port 8000 \
--api-server-count 1 \
--tensor-parallel-size 8 \
--served-model-name {model_name} \
--max-num-seqs 16 \
--max-model-len 40960 \
--max-num-batched-tokens 4096 \
--enable-expert-parallel \
--trust-remote-code \
--async-scheduling \
--additional-config='{"xlite_graph_config": {"enabled": true, "full_mode": true}}' \
--block_size {block_size} \
```

```bash
# make a request to the service
curl http://localhost:8000/v1/chat/completions \
    -d '{
        "model": "{model_name}",
        "messages": [
            {"role": "system", "content": "You are a helpful assistant."},
            {"role": "user", "content": "Tell me how to sleep well at night."}
        ],
        "max_tokens": 2048,
        "temperature": 0.7
    }'
```

Results:

| Model | `block_size` | Success? | Notes |
| ----- | ------------ | -------- | ----- |
| Qwen3-32B | 16 | ✅ |  |
| Qwen3-32B | 32 | ✅ |  |
| Qwen3-32B | 64 | ✅ |  |
| Qwen3-32B | 128 | ✅ |  |
| Qwen3-32B | 256 | ❌ | _RuntimeError: ACL stream synchronize failed, error code:507015_ |
| Qwen3-30B-A3B | 16 | ✅ |  |
| Qwen3-30B-A3B | 32 | ✅ |  |
| Qwen3-30B-A3B | 64 | ✅ |  |
| Qwen3-30B-A3B | 128 | ✅ |  |
| Qwen3-30B-A3B | 256 | ❌ | _RuntimeError: ACL stream synchronize failed, error code:507015_ |

Model mapping:

- `Qwen3-32B` -> ` vllm_ascend.xlite.xlite.LlamaXliteModel`
- `Qwen3-30B-A3B` -> ` vllm_ascend.xlite.xlite.QwenMoeXliteModel`

- vLLM version: v0.18.0

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -1119,11 +1119,7 @@ def refresh_block_size(vllm_config):
 
     ascend_config = get_ascend_config()
     if ascend_config.xlite_graph_config.enabled and cache_config.block_size > 128:
-        logger.warning(
-            f"Current block size {cache_config.block_size} is larger than the maximum supported block size of 128. "
-            f"Your vLLM server will likely crash when processing requests, even after a successful launch."
-        )
-        logger.warning("Resetting block size to 128 for xlite compatibility.")
+        logger.warning("Setting block size to 128 for xlite compatibility.")
         cache_config.block_size = 128
 
 


SijieFu:
The precisions from smaller block sizes (16, 32, 64) have been tested and are acceptable. (Source @lulina)

We have not benchmarked the performances from different block sizes, although we noted that [some user](https://github.com/vllm-project/vllm-ascend/issues/7563) reported better performance when block size was set to 16.

Under most scenarios, the block size would still be set to 128, if `cache_config.enable_prefix_caching` or `scheduler_config.enable_chunked_prefill` ([ref](https://github.com/vllm-project/vllm-ascend/pull/7943/changes#diff-452868d6534dcaf267ff38be22caa42c42fbecfa56515886740d977d0fdcb4b9L1115)).

PR #7962, opened 2026-04-02
[CI] Main2main upgrade vllm to 0330

22dimensions:
Main2main upgrade vllm to 0330
fix breaks:
1. https://github.com/vllm-project/vllm/pull/37728 add clear_row method for BlockTable
2. https://github.com/vllm-project/vllm/pull/37975 Adapt GatedDeltaNetAttention Refactor
3. https://github.com/vllm-project/vllm/pull/37698  update maybe_update_config in vllm_ascend/quantization/modelslim_config.py to adapt this pr change
4. https://github.com/vllm-project/vllm/pull/37880  This pr add the feat where we can set different moe backends between draft and target model, we should overwrite it in the draft proposer
5. https://github.com/vllm-project/vllm/pull/37853  for now just to skip test_cpu_offloading.py test case utils this feature has been adapted.

No

CI

- vLLM version: v0.18.0
- vLLM main: https://github.com/vllm-project/vllm/commit/29e48707e8144b78dd5d756f793c26a405043f3d


Potabk:
https://github.com/vllm-project/vllm/pull/37880 This pr add the feat where we can set different moe backends between draft and target model, we should overwrite it in the draft proposer

vllm_ascend/ops/gdn.py
```diff
@@ -0,0 +1,415 @@
...
+        hidden_states: torch.Tensor,
+        output: torch.Tensor,
+    ):
+        """
+        Forward pass with three parts:
+        1. Input projection
+        2. Core attention (custom op)
+        3. Output projection
+        """
+        if not self.gqa_interleaved_layout:
```
MengqingCao, line 57: add comment for both if and else branch

vllm_ascend/ops/gdn.py
```diff
@@ -0,0 +1,415 @@
...
+                validate_data=True,
+            )
+        else:
+            mixed_qkv_non_spec = None
+
+        query_spec, key_spec, value_spec = self.rearrange_mixed_qkv(mixed_qkv_spec)
+        query_non_spec, key_non_spec, value_non_spec = self.rearrange_mixed_qkv(mixed_qkv_non_spec)
+
+        # 2. Recurrent attention
+        if self.gqa_interleaved_layout:
```
MengqingCao, line 217: It seems better to split the forward functions of qwen3-next and qwen3.5.

PR #8017, opened 2026-04-07
[BugFix][310p] Fix post-sampling not working in graph mode on 310P

Tflowers-0129:

Enabling temperature in post-processing on 310P devices can cause the service to stall and eventually hang. We first traced the issue to a timeout where the temperature-related `div` operator was waiting for results from a sub-stream. After investigating the preceding operators, we finally identified the root cause as the `q.exponential_()` operator, which is not well supported on 310P and triggers an internal issue in the `add` kernel.

NA

This patch was thoroughly tested locally（accuracy-dataset test and stress test）. It is not easy to design a proper unit test for this case, and I appreciate your understanding.
- vLLM version: v0.18.0
- vLLM main: https://github.com/vllm-project/vllm/commit/14acf429ac08b6d538ca6feb3e06b6d13895804d

### DIFF
--- a/vllm_ascend/_310p/sample/sampler.py
+++ b/vllm_ascend/_310p/sample/sampler.py
@@ -16,9 +16,8 @@
 #
 
 import torch
-from vllm.model_executor.layers.batch_invariant import vllm_is_batch_invariant
+import vllm.envs as envs
 
-from vllm_ascend.ascend_config import get_ascend_config
 from vllm_ascend.sample.sampler import (
     DEFAULT_LOGPROBS_MODE,
     AscendSampler,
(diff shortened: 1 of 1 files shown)


vllm_ascend/_310p/model_runner_310p.py
```diff
@@ -35,7 +35,9 @@
     MambaSpec,
     UniformTypeKVCacheSpecs,
 )
+from vllm.v1.sample.rejection_sampler import RejectionSampler
 
+from vllm_ascend._310p.sample.sampler import AscendSampler310
```
YangShuai52, line 40: 可以直接从vllm_ascend._310p.sample import

vllm_ascend/_310p/sample/sampler.py
```diff
@@ -0,0 +1,81 @@
...
+    def __init__(self, logprobs_mode=DEFAULT_LOGPROBS_MODE):
+        super().__init__(logprobs_mode=logprobs_mode)
+        self.topk_topp_sampler = AscendTopKTopPSampler310(logprobs_mode=logprobs_mode)
+
+    def do_async_exponential(self, b_s, head_dim, generators):
+        with torch.npu.stream(global_stream()):
+            global_stream().wait_stream(torch.npu.current_stream())
+            q = torch.empty((b_s, head_dim), device="npu", dtype=torch.float32)
+            q = _generate_exponential_q(q, generators)
+            self.async_exponential_event.record()
```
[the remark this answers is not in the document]
Tflowers-0129: updated

vllm_ascend/_310p/model_runner_310p.py
Tflowers-0129: yeap, i see,  but i think current way to import ascendsampler310 hasnot any problem😀

PR #8032, opened 2026-04-08
[Misc] Remove useless Conv2d CustomOp

shen-shanshan:
Remove useless `AscendConv2dLayer` CustomOp.

Enforcing `Conv2dLayer` to use linear matmul hurts performance, since it only benefits performance for `Conv3dLayer`.

No.

No.

- vLLM version: v0.18.0
- vLLM main: https://github.com/vllm-project/vllm/commit/14acf429ac08b6d538ca6feb3e06b6d13895804d

### DIFF
--- a/vllm_ascend/ops/conv.py
+++ b/vllm_ascend/ops/conv.py
@@ -17,13 +17,7 @@
 
 
 import torch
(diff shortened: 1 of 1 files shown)


shen-shanshan:
CC @wangxiyuan @MengqingCao

PR #8043, opened 2026-04-08
[Performance][SFA] Reduce prefill KV all-gather communication for PCP/DCP

xiaocongtou6:
For SFA models with PCP/DCP enabled, prefill currently perform full-KV all_gather to support graph mode, which introduces substantial communication overhead. However, in the prefill stage, the SFA attention computation itself is not captured in graph mode, for either piecewise or full_decode_only. This PR changes the prefill path to gather only the required KV instead of the full KV, significantly reducing communication overhead and improving TTFT.

- vLLM version: v0.19.0
- vLLM main: https://github.com/vllm-project/vllm/commit/2a69949bdadf0e8942b7a1619b229cb475beef20

### DIFF
--- a/vllm_ascend/attention/context_parallel/sfa_cp.py
+++ b/vllm_ascend/attention/context_parallel/sfa_cp.py
@@ -301,7 +301,7 @@ class AscendSFACPImpl(AscendSFAImpl):
     def _align_to_graph_bucket_tokens(self, attn_output: torch.Tensor | None, attn_metadata: M) -> torch.Tensor | None:
         if attn_output is None or self.pcp_size == 1:
             return attn_output
-        # In graph/piecewise mode, output buffer uses graph bucket token size
+        # In graph mode, output buffer uses graph bucket token size
         # (forward_context.num_tokens), while PCP path may compute only valid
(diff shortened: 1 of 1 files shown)


MengqingCao:
@weiguihua2 @rjg-lyh please also take a look.

MengqingCao:
So could we re-enable the nightly test of deepseek-v3.2 with cp?

vllm_ascend/attention/context_parallel/sfa_cp.py
```diff
@@ -76,6 +76,28 @@ def build(
...
+            # Prefill is not graph-captured, so we keep a compact block view for
+            # it and let decode continue using the graph-friendly full KV view.
+            valid_block_ids, block_table_cp = self.build_prefill_compact_block_metadata(
+                metadata_cls.block_table, num_decodes
+            )
+            sfa_cp_metadata.valid_block_ids = valid_block_ids
+            sfa_cp_metadata.block_table_cp = block_table_cp
+
+        actual_seq_lengths_query = metadata_cls.cum_query_lens
+        if num_prefills > 0:
```
weiguihua2: The two parts of the if statement are combined together.

PR #8063, opened 2026-04-09
[0.18.0][BugFix][KV Pool] preserve drafted connector output on v0.18.0

Pz1116:

This fixes a KV Pool regression in PD + speculative decoding on `releases/v0.18.0`.

When speculative decoding is enabled, vLLM defers KV connector finalization
during target-model forward, and finalizes it after draft model execution.
`vllm-ascend` already calls `finalize_kv_connector()` in this path, but
`sample_tokens()` still returned the stale `kv_connector_output` captured at
the beginning of the function.

As a result, worker-side connector state updated during drafting/finalization
could be dropped before being sent back to the scheduler. For KV Pool, this
can prevent the scheduler from observing finished send events, so request
blocks remain in delay-free state and KV cache usage keeps increasing until
new waiting requests can no longer enter running.

This PR aligns `vllm-ascend` behavior with upstream:
- keep the PP non-final-rank passthrough path unchanged
- after speculative drafting and deferred connector finalization, re-read
  `self.kv_connector_output`
- return the post-drafting connector output instead of the stale pre-drafting
  snapshot

### Root cause

`sample_tokens()` did this too early:

- read `kv_connector_output = self.kv_connector_output`
- clear `self.kv_connector_output`

But in speculative decoding, connector output may be updated during:
- draft token generation
- deferred `finalize_kv_connector()`

So the returned connector output could miss the finished sending state needed
by the scheduler to free delayed KV blocks.

Yes.

It fixes a runtime regression where KV Pool blocks may not be released
correctly in PD + speculative decoding scenarios, causing KV cache usage to
grow continuously and eventually preventing waiting requests from being
scheduled.

pending

### DIFF
--- a/tests/ut/distributed/mooncake/test_ascend_store_connector.py
+++ b/tests/ut/distributed/mooncake/test_ascend_store_connector.py
@@ -1,0 +1,84 @@
+import unittest
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import torch
+
+if not hasattr(torch, "npu"):
+    torch.npu = SimpleNamespace(Event=object)  # type: ignore[attr-defined]
+
+from vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.ascend_store_connector import (  # noqa: E402
+    AscendStoreConnector,
+)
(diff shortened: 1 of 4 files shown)


WANG-WADE:
still not working under GLM-5 w4a8 with dual A2 machine

PR #8079, opened 2026-04-09
[Feature][Debug] Upgrade device-side debug print with launchHostFunc

yiz-liu:
This PR upgrades Ascend debug printing from the old `acl_graph_print` path to a
new `launchHostFunc`-based `device_print` implementation.

The previous design had two major issues:
- it could hit a deadlock path like `main thread -> device sync -> callback -> GIL -> main thread`
- it fails under Dynamo / `torch.compile`

To address that, this PR adds a custom-op-based device print path that can be
preserved through compile and graph execution, and removes the older graph-only helper.

- add `device_print(str)` and `device_print_tensor(Tensor)` custom ops in `csrc/torch_binding.cpp`
- add Meta registrations for the new print ops in `csrc/torch_binding_meta.cpp`
- mark the print ops as side-effectful in `vllm_ascend/utils.py` so FX/Inductor does not drop or reorder them
- remove `acl_graph_print` and its old stream subscription / cleanup path from `vllm_ascend/utils.py`
- keep `device_print` as the single Python debugging helper, with single-argument semantics
- add `examples/device_print_demo.py` to validate eager, `torch.compile(backend="aot_eager")`, and `torch.npu.graph` replay
- prefer CANN ACL headers over torch_npu's bundled ACL headers so `aclrtLaunchHostFunc`

```mermaid
sequenceDiagram
    autonumber
    participant Dev as 开发者代码
    participant Py as vllm_ascend.utils.device_print
    participant Reg as 注册与副作用标记
    participant FX as torch.compile / graph capture
    participant Meta as Meta kernel
    participant Op as torch.ops._C_ascend
    participant CPP as C++ custom op
    participant ACL as ACL Runtime
    participant Stream as NPU Stream
    participant CB as Host callback
    participant Out as stdout

    Dev->>Py: device_print(value)
    Py->>Reg: ensure custom op registered
    Reg->>Reg: mark device_print ops side-effectful
    Reg-->>Py: ready

    alt value 是 Tensor
        Py->>Op: device_print_tensor(value)
    else value 是普通值
        Py->>Op: device_print(str(value))
    end

    alt torch.compile 或 graph capture
        Op->>FX: custom op enters traced graph
        FX->>Meta: run no-op meta implementation
        Meta-->>FX: preserve op node
        FX-->>Op: runtime dispatch preserved op
    else eager
        Op-->>CPP: direct dispatch
    end

    alt 字符串打印
        Op->>CPP: device_print(msg)
        CPP->>CPP: build DevicePrintPayload(message)
    else Tensor 打印
        Op->>CPP: device_print_tensor(tensor)
        alt CPU Tensor
            CPP->>CPP: clone to host snapshot
        else NPU Tensor
            CPP->>CPP: contiguous + allocate pinned host snapshot
            CPP->>ACL: aclrtMemcpyAsync(D2H, same stream)
        end
    end

    CPP->>Stream: get current NPU stream
    CPP->>ACL: aclrtLaunchHostFunc(stream, callback, payload)
    ACL-->>Stream: enqueue host callback after prior stream work
    Stream->>CB: device_print_callback(payload)
    CB->>Out: print message / tensor snapshot
    CB->>Out: flush()

    Note over CB,Out: payload 保留以支持 torch.npu.graph replay，避免 replay 时悬空访问

    alt torch.npu.graph replay
        Dev->>Stream: graph.replay()
        Stream->>CB: invoke callback again
        CB->>Out: print again
    end
```

Yes, for developers debugging Ascend execution:
- `acl_graph_print` is removed
- `device_print(...)` becomes the supported debug helper
- the new helper is intended to work in eager mode, `torch.compile`, and NPU graph replay

- added manual validation through `examples/device_print_demo.py`
- vLLM version: 


CMakeLists.txt
```diff
@@ -109,12 +107,17 @@ set(
...
-  ${ASCEND_HOME_PATH}/x86_64-linux/include/experiment/platform
 )
 
 pybind11_add_module(vllm_ascend_C ${VLLM_ASCEND_SRC})
 
+# Prefer the CANN ACL headers over torch_npu's bundled third_party ACL copy.
+# torch_npu 2.9.0 ships an older acl_rt.h that does not declare
+# aclrtLaunchHostFunc, which breaks host-print compilation.
+target_include_directories(vllm_ascend_C BEFORE PRIVATE
+  ${ASCEND_HOME_PATH}/include
```
MengqingCao, line 118: plz double check this to make sure this is correct.

MengqingCao (approved):
LGTM

CMakeLists.txt
yiz-liu: Yeah I have checked it, I guess those dirs are long gone, no harm getting rid of them.

PR #8105, opened 2026-04-10
[BugFix] Fix swap_blocks wrong device and missing task_done in KV transfer

SuperMarioYL:
## Purpose

Fix two functional bugs: a wrong-device copy in `swap_blocks` and a missing `task_done()` call in the KV transfer error path that can cause queue deadlocks.

### Bug 1: `swap_blocks` copies value cache to wrong device

**File:** `vllm_ascend/attention/attention_v1.py`, line 114

```python
# Before (copies value cache to KEY cache's device)
dst_value_cache[dst_indices] = src_value_cache[src_indices].to(dst_key_cache.device)

# After (copies value cache to VALUE cache's device)
dst_value_cache[dst_indices] = src_value_cache[src_indices].to(dst_value_cache.device)
```

This is a copy-paste error from line 113 (the key cache line). While key and value caches typically share a device today, this would silently corrupt data in heterogeneous memory setups where they reside on different devices.

### Bug 2: Missing `task_done()` on exception in `KVTransferThread.run()`

**File:** `vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/kv_transfer.py`, line 82-83

When `_handle_request()` raises an exception, the `except` block logs the error but never calls `self.request_queue.task_done()`. This means any code that calls `queue.join()` will deadlock permanently, as the task counter is never decremented for failed requests.

## Test

- `ruff check` passes on both modified files.
- Both fixes are single-line corrections to existing logic.

- vLLM version: v0.24.0

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -121,7 +121,7 @@ class AscendAttentionBackend(AttentionBackend):
         dst_indices = src_to_dst[:, 1]
 
         dst_key_cache[dst_indices] = src_key_cache[src_indices].to(dst_key_cache.device)
-        dst_value_cache[dst_indices] = src_value_cache[src_indices].to(dst_key_cache.device)
+        dst_value_cache[dst_indices] = src_value_cache[src_indices].to(dst_value_cache.device)
 
     @staticmethod
     def copy_blocks(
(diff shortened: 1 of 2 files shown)


SuperMarioYL:
Rebased onto upstream/main to clear the base drift / merge conflict. The only conflict was in `kv_transfer.py`: upstream reworded the `logger.error` in `KVTransferThread.run()`'s except block, and this PR's `task_done()` addition is re-applied on top of the reworded log line. `attention_v1.py` merged clean. Both fixes are intact in 0c65d35cc:

- `attention_v1.py`: value cache copy now `.to(dst_value_cache.device)` (was the key cache's device — a copy-paste from the preceding key-cache line).
- `kv_transfer.py`: `self.request_queue.task_done()` added to `run()`'s error path, so `queue.join()` no longer deadlocks on a failed request.

Verified both bugs are still live on current main (`attention_v1.py:124` still targets the key cache's device; `run()`'s except block still lacks `task_done()`). Happy to address any further feedback.

Issue #5208, opened 2025-12-20
[RFC]: Model Runner V2


Ronald1995:
### 🚀 The feature, motivation and pitch

vllm has implemented model runner v2, see https://github.com/vllm-project/vllm/issues/23446
We will implement model runner v2 in vllm-ascend, here are related plans:

Features: 
- [x] implement basic framework #5051
- [x] e2e function for eager mode https://github.com/vllm-project/vllm-ascend/pull/5210 
- [x] refactor of ascend forward context. https://github.com/vllm-project/vllm-ascend/pull/5720
- [x] adapt to uvabuffer https://github.com/vllm-project/vllm-ascend/pull/6657
- [x] support aclgraph of main model https://github.com/vllm-project/vllm-ascend/pull/7110
- [ ] support aclgraph of draft model ongoing
- [x] support eagle spec decoding https://github.com/vllm-project/vllm-ascend/pull/5840 
- [x] async scheduling X spec decoding https://github.com/vllm-project/vllm-ascend/pull/5840
- [ ] support sequence parallel & flashcomm  https://github.com/vllm-project/vllm-ascend/pull/5627
- [ ] support data parallel @YuhanBai 
- [ ] support mm model
- [ ] support lora
- [ ] support pool model
- [ ] support hybrid attention model, such as qwen3-next.
- [ ] support context parallel
- [ ] support pipeline parallel 
- [x] support logprob @wangx700 
- [ ] support penalties
- [ ] support structure output
- [ ] async scheduling X  logprob
- [ ] async scheduling X  penalties
- [ ] PD disaggregation
- [ ] support eplb
- [ ] support lmhead tensor parallel
- [ ] support ngram spec decoding
- [ ] support suffix spec decoding

triton operators optimization:
- [ ] triton operator optimization in eagle spec decoding
- [ ] gumbel sampling triton operator optimization @shiyuan680
- [x] logprob triton operator optimization https://github.com/vllm-project/vllm-ascend/pull/7221
- [ ] min_p triton operator optimization
- [x] penalties triton operators optimization @shiyuan680  https://github.com/vllm-project/vllm-ascend/pull/5854 https://github.com/vllm-project/vllm-ascend/pull/7013
- [ ] rejection sample triton operator optimization
- [ ] triton operator optimization in `update_states`  @henryxuxu0716
- [ ] triton operator optimization in `prepare_inputs`  @cocacolafan

wxsIcey:
[...]

PR #8111, opened 2026-04-10
[BugFix] Fix the graph capturing OOM in model_runner_v2

slippersss:
This PR aims to fix the graph capturing OOM in model_runner_v2. The reason for the problem is that `workspace` for each graph is not released during capturing, and when the number of the graphs increases, OOM will happen somehow. Therefore, we wrap `torch.npu.graph` and add `weak_ref_workspace` as post-processing. Since this context manager will be called for exactly each graph, this is what we need.

RFC: #5208

N/A

by ci

- vLLM version: 


vllm_ascend/worker/v2/utils.py
```diff
@@ -38,3 +40,11 @@ def communicator_switch():
...
     finally:
         vllm.distributed.device_communicators.cuda_communicator.CudaCommunicator = CudaCommunicator
+
+
+@contextmanager
+def torch_npu_graph_wrapper(*args, **kwargs):
+    with torch.npu.graph(*args, **kwargs):
+        yield
+    weak_ref_workspaces(get_graph_params())
+    weak_ref_workspaces(get_draft_graph_params())
```
yiz-liu:
```suggestion
@contextmanager
def torch_npu_graph_wrapper(*args, **kwargs):
    # MRV2-specific cleanup hook: intentionally reuse the graph context
    # manager's exit to weak-ref graph workspaces after each capture,
    # without adding another upstream monkey patch.
    try:
        with torch.npu.graph(*args, **kwargs):
            yield
    finally:
        weak_ref_workspaces(get_graph_params())
        weak_ref_workspaces(get_draft_graph_params())
```
This is a reasonable MRV2-specific hotfix, but since it is effectively using the graph context manager as a capture-end cleanup hook rather than introducing a new graph abstraction, please add a clarifying comment and move the cleanup into finally so it still runs if capture fails.

yiz-liu (approved):

vllm_ascend/worker/v2/utils.py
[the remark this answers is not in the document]
yiz-liu: Indeed.

wangxiyuan (approved):

vllm_ascend/worker/v2/utils.py
slippersss: Done.

PR #8134, opened 2026-04-10
[BugFix] Fix GDN attention flash comm v1 bug under max concurrency

wangbj127:

- vLLM version: 


wangbj127:
This problem is already solved in https://github.com/vllm-project/vllm-ascend/pull/7858, so this PR should be closed.

PR #8172, opened 2026-04-13
[Misc] Upgrade vllm commit to 0414

Potabk:
Upgrade vllm commit to `6f786f2c506cb07f4566771fdc62e640e2c4a176`
1. fix https://github.com/vllm-project/vllm/pull/32936

This pr leads to the issue in Capture Phase (_dummy_run), https://github.com/vllm-project/vllm/issues/28207#issuecomment-3504004536
we should  re-read `compilation_config.cudagraph_capture_sizes` after the super() call in `_check_and_update_cudagraph_mode` to keep `self.cudagraph_batch_sizes` in sync with the (possibly rewritten) sizes in `model_runner_v1.NPUModelRunner_check_and_update_cudagraph_mode`.

For example, when speculative decoding (e.g. eagle3) is enabled and `cudagraph_capture_sizes` is explicitly specify as [5, 12], vLLM's `_check_and_update_cudagraph_mode` calls `adjust_cudagraph_sizes_for_spec_decode`, which rounds `cudagraph_capture_sizes` up to a multiple of
(num_speculative_tokens + 1). For example, with `num_speculative_tokens=2` , [5, 12] becomes [6, 12].

However, in vllm-ascend,  `self.cudagraph_batch_sizes` was cached during __init__ with the original [5, 12]. When `set_graph_params(self.cudagraph_batch_sizes)` runs later, it creates `graph_params.events` keyed by {5, 12}. Meanwhile, the `CudagraphDispatcher` uses the updated [6, 12] from `compilation_config`, so it tries to capture at num_tokens=6 — causing KeyError: 6 in `graph_params.events[num_tokens]` inside `full_graph_fia`.

you can also re-produce the issue with the script:
```python
import os

os.environ.setdefault("VLLM_WORKER_MULTIPROC_METHOD", "spawn")

from vllm.config import CompilationConfig

EXAMPLE_PROMPTS = [
    "Hello, my name is",
    "The president of the United States is",
    "The capital of France is",
    "The future of AI is",
]

SAMPLING_PARAMS = SamplingParams(
    max_tokens=300,
    temperature=0.0,
    ignore_eos=False,
)


def run_spec():
    """Run with eagle3 speculative decoding."""
    llm = LLM(
        model="Qwen/Qwen3-8B",
        tensor_parallel_size=1,
        pipeline_parallel_size=1,
        data_parallel_size=1,
        disable_log_stats=False,
        max_model_len=4096,
        seed=1024,
        async_scheduling=False,
        speculative_config={
            "disable_padded_drafter_batch": False,
            "method": "eagle3",
            "model": "RedHatAI/Qwen3-8B-speculator.eagle3",
            "num_speculative_tokens": 2,
            "draft_tensor_parallel_size": 1,
            "max_model_len": 128,
        },
        compilation_config=CompilationConfig(
            cudagraph_mode="FULL",
            cudagraph_capture_sizes=[5, 12],
        ),
    )
    spec_outputs = llm.generate(EXAMPLE_PROMPTS, SAMPLING_PARAMS)
    del llm
    return spec_outputs


def main():
    spec_outputs = run_spec()
    for o in spec_outputs:
        print(f"  PROMPT: {o.prompt!r}")
        print(f"  OUTPUT: {o.outputs[0].text[:80]}...")


if __name__ == "__main__":
    main()

```

3. fix https://github.com/vllm-project/vllm/pull/39604
1. For 310P, we are
- vLLM version: 


zhangxinyuehfad:
/e2e tests/e2e/multicard/2-cards/test_quantization.py tests/e2e/singlecard/model_runner_v2/test_basic.py tests/e2e/singlecard/spec_decode/test_mtp_eagle_correctness.py::test_llama_qwen3_eagle_correctness[None-True-False-eagle-1-LLM-Research/Meta-Llama-3.1-8B-Instruct-vllm-ascend/EAGLE-LLaMA3.1-Instruct-8B]

docs/source/conf.py
```diff
@@ -83,7 +83,7 @@
     "ci_vllm_version": "v0.18.0",
```
wangxiyuan: why this is still 0.18.0

GDzhu01: This is for the `get_version` in doc tests with the latest version we have released

Potabk: see https://github.com/vllm-project/vllm-ascend/blob/c977a9950b06a4691ba5725b6632297a4e4aa1b1/tests/e2e/common.sh#L69

yiz-liu (commented):
We should use `self.cudagraph_dispatcher.get_capture_descs()` instead of `self.cudagraph_batch_sizes` now.

PR #8208, opened 2026-04-13
[BugFix] fix dsv3.1 service failed to start

Yaphets24:
This PR fixes a service startup failure for DeepSeek-V3.1 models by removing a strict type assertion for MLAAttentionSpec in NPUModelRunner.get_kv_cache_spec. The assertion was failing due to class identity mismatches caused by the runtime patching of MLAAttentionSpec with AscendMLAAttentionSpec.

No.

Verified that the service starts correctly for DSV3.1 models.

- vLLM version: 

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -3412,7 +3412,7 @@ class NPUModelRunner(GPUModelRunner):
                     kv_cache_spec[layer_name] = spec
                     attn_layer_names.add(layer_name)
 
-            elif isinstance(attn_module, MLAAttention): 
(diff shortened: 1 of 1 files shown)


rjg-lyh (approved):

yiz-liu (approved):

rjg-lyh:
The assert check in the branch is intended to prevent potential issues in the future where the attn_module of the MLAAttention class in vllm might have other kv_cache_spec attributes. This could lead to errors when directly initializing AscendMLAAttentionSpec using its member attributes. 

However, based on the current code scenario, this potential issue does not exist and the assert check is redundant. It can be removed, which will also help avoid the conflict between the native and patched versions of MLAAttentionSpec mentioned in this PR.

PR #8206, opened 2026-04-13
[NOT MERGE][CI] TEST FULL CI ONLY

linfeng-yuan:
### How was this patch tested?

- vLLM version: 
- vLLM main: https://github.com/vllm-project/vllm/commit/v0.19.0

PR #8209, opened 2026-04-13
[MRV2][Spec Decoding][BugFix] fix mrv2 runtime error with speculative decoding

linfeng-yuan:

This PR fixes a runtime error in MRV2 when using speculative decoding (Eagle) by ensuring that the `gumbel_sample` function in the `speculator` module is correctly overridden with the Ascend-specific implementation. The root cause of this broken is that the patch of definition of `gumbel_sample` was removed in #8155.

No.

Verified that speculative decoding works correctly on Ascend hardware without runtime errors.

- vLLM version: 


linfeng-yuan:
Please refer to #8206 and https://github.com/vllm-project/vllm-ascend/actions/runs/24341538385/job/71071567870?pr=8206 to check the broken cases without this PR.

yiz-liu (approved):

PR #8210, opened 2026-04-13
[Feature] Support step3p5 main model

nomewang:
                                                                                                                                                                     
  This PR adds support for the step3p5 model on Ascend NPU. The step3p5 model is a hybrid architecture with mixed full-attention and sliding-window-attention (SWA)
  layers, and uses MoE with the SWIGLUSTEP activation function. The following changes are required:                                                                  
                                                                                                                                                                     
#### Attention:                                                                                                                                                         
  - Add sliding window attention support by configuring sparse_mode=4, pre_tokens, and next_tokens in npu_fused_infer_attention_score calls (both graph capture and
  eager paths).                                                                                                                                                      
  - Fix _forward_fia_slidingwindow to use proper TND layout, correct actual_seq_lengths / seq_lens_list parameters, and sparse_mode=4 instead of the previous BSH
  layout with hardcoded values.                                                                                                                                      
  - Introduce attn_layer_names tracking in GraphParams to record layer names during ACL graph capture in execution order. During replay, use this capture-order list 
  instead of attn_metadata.keys() to ensure correct parameter-to-layer mapping for hybrid models where metadata key order differs from capture order.               
  - Propagate layer_name parameter through forward_impl, forward_fused_infer_attention, forward_paged_attention, full_graph_fia, full_graph_pa, and their            
  context-parallel equivalents (attention_cp, mla_cp, mla_v1).                                                                                           
  - Remove unused swa_mask building in AscendAttentionMetadataBuilder.build.                                                                                         
                                                                            
  #### MoE / Activation:                                                                                                                                                  
  - Add Ascend wrapper for SwiGLU-step activation.                                                              
  - Extend the Triton swiglu_quant kernel with USE_STEP and LIMIT compile-time constants to support the step activation variant.                                     
  - Add apply_moe_activation and apply_moe_activation_triton dispatchers to route between SILU and SWIGLUSTEP in both quantized and unquantized MoE MLP paths.       
                                                                                                                                                                     
#### Expert Selector:                                                                                                                                                   
  - Add e_score_correction_bias handling in _native_select_experts with proper null check, using gate_prob_with_bias for top-k selection when bias is present.       
  - Add routed_scaling_factor parameter propagation through the expert selection pipeline, applying it after weight renormalization.                                 
  - Add epsilon (1e-20) to top-k weight renormalization denominator for numerical stability.                                                                         
  - Cast router_logits to float32 before scoring for numerical precision.                                                                                            
                                                                                                                                                                     
#### 310P:                                                                                                                                                              
  - Add layer_name parameter to 310P attention forward_paged_attention and forward_impl for interface consistency.                                                   
                                                                                                                                                                     
  Yes. This PR adds support for running the step3p5 model on Ascend NPU, including sliding window attention and the SWIGLUSTEP MoE activation function. No breaking  
  changes to existing model support.                                                                                                                                 
                                                                                                                                                                     
  - Unit test updated to expect MoEActivation enum instead of string in test_moe_mlp.py.                                                                             
  - Step3.5-Flash is loaded successfully and returns the correct answer.                                                                                                                           
                                                                                                                                                                     
### vLLM version:                                                                                                                                                      
                                                                                                                                                                     
  vLLM main: https://github.com/vllm-project/vllm/commit/v0.19.0

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -51,7 +51,9 @@ from vllm_ascend.attention.utils import (
 from vllm_ascend.compilation.acl_graph import (
     get_draft_graph_params,
     get_graph_params,
+    update_draft_graph_params_layer_name,
     update_draft_graph_params_workspaces,
+    update_graph_params_layer_name,
     update_graph_params_workspaces,
 )
 from vllm_ascend.device.device_op import DeviceOperator
@@ -302,12 +304,6 @@ class AscendAttentionMetadataBuilder(AttentionMetadataBuilder[AscendMetadata]):
         attn_mask = self.attn_mask_builder.get_attention_mask(self.model_config)
 
         swa_mask = None
-        is_swa = hasattr(self.model_config.hf_text_config, "sliding_window")
-        if self.model_config is not None and is_swa:
-            swa_mask = self.attn_mask_builder.get_swa_mask(
-                self.model_config.dtype, self.model_config.hf_text_config.sliding_window
-            )
-
         # TODO: Yet another unnecessary H2D while we already have a query_start_loc on device
         query_start_loc = query_start_loc_cpu.pin_memory().to(self.device, non_blocking=True)
 
@@ -409,9 +405,17 @@ class AscendAttentionBackendImpl(AttentionImpl):
                 graph_params = get_draft_graph_params()
             else:
                 graph_params = get_graph_params()
+            # Use capture order layer names for correct replay ordering
+            # For hybrid models, attn_metadata.keys() order differs from capture order
+            captured_layer_names = graph_params.attn_layer_names.get(num_tokens, [])
+            if captured_layer_names:
+                attn_keys = captured_layer_names
+            else:
+                # Fallback to metadata keys if no captured names (backward compatibility)
+                attn_keys = list(forward_context.attn_metadata.keys())
             with torch.npu.stream(update_stream):
                 for key, param, handle, event in zip(
-                    forward_context.attn_metadata,
+                    attn_keys,
                     graph_params.attn_params[num_tokens],
                     graph_params.handles[num_tokens],
                     graph_params.events[num_tokens],
@@ -460,11 +464,25 @@ class AscendAttentionBackendImpl(AttentionImpl):
             if _EXTRA_CTX.is_draft_model:
                 graph_params = get_draft_graph_params()
                 attn_metadata = draft_attn_metadatas
-                attn_keys = list(attn_metadata[0].keys())
+                # Use capture order layer names for correct replay ordering
+                # For hybrid models, attn_metadata.keys() order differs from capture order
+                captured_layer_names = graph_params.attn_layer_names.get(num_tokens, [])
+                if captured_layer_names:
+                    attn_keys = captured_layer_names
+                else:
+                    # Fallback to metadata keys if no captured names (backward compatibility)
+                    attn_keys = list(attn_metadata[0].keys())
             else:
                 graph_params = get_graph_params()
                 attn_metadata = forward_context.attn_metadata
-                attn_keys = list(attn_metadata.keys())
+                # Use capture order layer names for correct replay ordering
+                # For hybrid models, attn_metadata.keys() order differs from capture order
+                captured_layer_names = graph_params.attn_layer_names.get(num_tokens, [])
+                if captured_layer_names:
+                    attn_keys = captured_layer_names
+                else:
+                    # Fallback to metadata keys if no captured names (backward compatibility)
+                    attn_keys = list(attn_metadata.keys())
             # For Qwen3-next, since the kv_cache_config has already categorized
             # linear_attn and self_attn, the attn_metadata is first arranged with
             # self_attn followed by linear_attn. Therefore, using zip directly
@@ -497,6 +515,7 @@ class AscendAttentionBackendImpl(AttentionImpl):
                         num_kv_heads,
                         num_heads,
                         scale,
+                        sliding_window,
                         attn_output,
                         softmax_lse,
                     ) = param
@@ -526,7 +545,9 @@ class AscendAttentionBackendImpl(AttentionImpl):
                         num_key_value_heads=num_kv_heads,
                         num_heads=num_heads,
                         scale=scale,
-                        sparse_mode=3,
+                        sparse_mode=4 if sliding_window is not None else 3,
+                        pre_tokens=sliding_window - 1 if sliding_window is not None else SWA_INT_MAX,
+                        next_tokens=1 if sliding_window is not None else SWA_INT_MAX,
                         workspace=graph_params.workspaces.get(num_tokens),
                         out=[attn_output, softmax_lse],
                     )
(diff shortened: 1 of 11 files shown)


vllm_ascend/compilation/acl_graph.py
```diff
@@ -241,6 +241,9 @@ class GraphParams:
     workspaces: dict[int, torch.Tensor]
     handles: dict[int, list[torch_npu._C._NPUTaskGroupHandle]]
     attn_params: dict[int, list[tuple]]
+    # Record layer names in capture order to ensure correct replay order
+    # for hybrid models with mixed full/swa attention types
+    attn_layer_names: dict[int, list[str]]
```
realliujiaxu, line 246: Is it necessary to make `attn_layer_names` a dict rather than list? Since there should be only one kind of `attn_layer_names` no matter what key (i.e. `num_tokens`) is.

vllm_ascend/attention/attention_v1.py
```diff
@@ -302,12 +304,6 @@ def build(
         attn_mask = self.attn_mask_builder.get_attention_mask(self.model_config)
 
         swa_mask = None
-        is_swa = hasattr(self.model_config.hf_text_config, "sliding_window")
-        if self.model_config is not None and is_swa:
-            swa_mask = self.attn_mask_builder.get_swa_mask(
-                self.model_config.dtype, self.model_config.hf_text_config.sliding_window
-            )
-
```
weijinqian0, line 310: need check cann version.

wangxiyuan (commented):
SwigluStepAndMul custom ops need be added as well

vllm_ascend/attention/attention_v1.py
```diff
@@ -756,29 +795,29 @@ def _get_fia_params(self, key: torch.Tensor, value: torch.Tensor, attn_metadata:
...
+        key = self.key_cache.view(num_block, block_size, -1)
+        value = self.value_cache.view(num_block, block_size, -1)
         attn_output, _ = torch_npu.npu_fused_infer_attention_score(
             query,
             key,
             value,
             num_heads=self.num_heads,
             num_key_value_heads=self.num_kv_heads,
-            input_layout="BSH",
+            input_layout="TND",
```
NSDie, line 811: Does CANN support the TND layout?

vllm_ascend/compilation/acl_graph.py
weijinqian0, line 246: need transfer this to list.

vllm_ascend/attention/attention_v1.py
```diff
@@ -756,29 +795,29 @@ def _get_fia_params(self, key: torch.Tensor, value: torch.Tensor, attn_metadata:
...
+            pre_tokens=self.sliding_window - 1,
+            next_tokens=1,
```
NSDie, line 813: Is next_tokens=1 a design for step3p5? Can this be applied to other large models like SWA? For example, Gemma3 and GPT-oss.

realliujiaxu:
please add a model precision test case to nightly test

vllm_ascend/attention/attention_v1.py
```diff
@@ -853,6 +893,8 @@ def forward_fused_infer_attention(
                 query=query,
                 key=key,
                 value=value,
+                pre_tokens=self.sliding_window if self.sliding_window is not None else SWA_INT_MAX,
+                next_tokens=0 if self.sliding_window is not None else SWA_INT_MAX,
```
NSDie, line 897: Why is 0 set here, while 1 is set before?

vllm_ascend/ops/fused_moe/moe_mlp.py
```diff
@@ -320,14 +333,58 @@ def quant_apply_mlp(
...
+
+        hidden_states, swiglu_out_scale = swiglu_quant(
+            hidden_states,
+            group_list=group_list,
+            group_list_type=group_list_type,
+            use_step=True,
+            limit=7.0,
+        )
+    else:
+        raise ValueError(f"Unsupported FusedMoE activation: {activation}")
```
NSDie, line 352: Can more detailed error information be provided? Which activation functions are supported?

vllm_ascend/compilation/acl_graph.py
nomewang: Each graph needs to run `update_graph_params_layer_name`, and if using the list type, the management of layername would be more complex.

vllm_ascend/attention/attention_v1.py
```diff
@@ -574,8 +596,10 @@ def full_graph_fia(
                 actual_seq_lengths_kv=actual_seq_lengths_kv,
                 num_key_value_heads=self.num_kv_heads,
                 num_heads=self.num_heads,
-                sparse_mode=3,
+                sparse_mode=4 if self.sliding_window is not None else 3,
                 scale=self.scale,
+                pre_tokens=self.sliding_window - 1 if self.sliding_window is not None else SWA_INT_MAX,
+                next_tokens=1 if self.sliding_window is not None else SWA_INT_MAX,
```
gxy9808, line 602: Suggestion: Could we update this to pre_tokens=self.sliding_window and next_tokens=0?
In the MTP3 scenario (seq_len=4), setting next_tokens=1 would allow the current token to attend to the next token. Setting next_tokens=0 ensures that the current token only attends to itself and previous tokens, which seems to be the intended behavior here.

vllm_ascend/attention/attention_v1.py
```diff
@@ -620,8 +652,10 @@ def full_graph_fia(
             actual_seq_lengths_kv=actual_seq_lengths_kv,
             num_key_value_heads=self.num_kv_heads,
             num_heads=self.num_heads,
+            sparse_mode=4 if self.sliding_window is not None else 3,
             scale=self.scale,
-            sparse_mode=3,
+            pre_tokens=self.sliding_window - 1 if self.sliding_window is not None else SWA_INT_MAX,
+            next_tokens=1 if self.sliding_window is not None else SWA_INT_MAX,
```
gxy9808, line 658: Question: Should this be pre_tokens=self.sliding_window and next_tokens=0 instead?
I noticed that in MTP3 (seq_len=4), if next_tokens=1, the current token can attend to the next token. Setting it to 0 would ensure it only attends to the current and previous tokens.

vllm_ascend/attention/attention_v1.py
```diff
@@ -756,29 +795,29 @@ def _get_fia_params(self, key: torch.Tensor, value: torch.Tensor, attn_metadata:
...
+        value = self.value_cache.view(num_block, block_size, -1)
         attn_output, _ = torch_npu.npu_fused_infer_attention_score(
             query,
             key,
             value,
             num_heads=self.num_heads,
             num_key_value_heads=self.num_kv_heads,
-            input_layout="BSH",
+            input_layout="TND",
+            pre_tokens=self.sliding_window - 1,
```
gxy9808, line 812: In MTP3 (seq_len=4), using next_tokens=1 would unintentionally let the current token attend to the next one. Setting it to 0 guarantees strict causal attention (current and past tokens only), which aligns with our expected behavior.   Please update to pre_tokens=self.sliding_window and next_tokens=0

gxy9808 (commented):
Please update to pre_tokens=self.sliding_window and next_tokens=0.
For MTP3 (seq_len=4), we need next_tokens=0 to prevent the current token from attending to the next token. This ensures it only attends to current and past tokens

yiz-liu (commented):
Why do we need this `layer_names`?

PR #8220, opened 2026-04-14
[BugFix] Fix quant_bias missing in w8a8_static when flashcomm1 is enabled for GLM-5

triomino:
In a previous PR #7843 , the o_proj layer of GLM-5 was reverted to TP (Tensor Parallel) splitting when flashcomm1 was enabled. However, this was a temporary workaround and did not address the root cause of the precision issues observed in the o_proj layer under flashcomm1.

I am working on a definitive fix for this issue. Currently, a clear bug has been identified in https://github.com/vllm-project/vllm-ascend/blob/880e20fdde0b4e903685741b06e23b5fb8c287a1/vllm_ascend/quantization/methods/w8a8_static.py#L124: during quantized matrix multiplication, quant_bias is not added if tp_rank > 0. In the flashcomm1 scenario, all ranks actually require the addition of quant_bias, meaning tp_rank=0 should be passed to ensure the bias is applied correctly.

This PR aims to resolve this logic error and fix the underlying precision issue.

No.

glm5 e2e test


rjg-lyh (approved):
The quant_method implementation on the library is by default designed for the pd-mix scenario, where the quant_bias is only added to the computation result of tp_rank=0, as typically, in the pd-mix scenario, a reduce operation occurs after o_proj. However, in the PD separation scenario with dsa-cp, the token computations on each card are independent of each other. If the quant_bias is only added to the tokens on tp_rank=0, tokens on other cards will encounter precision issues due to the absence of the bias.

triomino:
/e2e tests/e2e/multicard/4-cards/long_sequence/test_prefix_caching_cp.py

PR #8290, opened 2026-04-15
[BugFix][DSv32] Fix DSA-CP PD role gating for deepseek v3.2

chenchuw886:
This PR fixes the DSA-CP PD role gating in `vllm_ascend.utils`.

The current helper logic does not distinguish the PD mixed-role case correctly when deciding whether layer sharding or TP `o_proj` handling should be enabled. In practice, layer sharding should stay limited to the P-side instance, while TP `o_proj` handling should remain enabled for normal non-PD deployments and for the PD mixed-role (`kv_both`) instance.

Such wrong condition lead to **vllm serve failures** in case: **FC1 + PD-colocated KV pooling + no layer_sharding**, specifically causing:
1. insufficient Available KV cache memory
2. o_proj shape error in sfa_v1 attention module

This patch makes those role checks explicit and adds unit coverage for the accepted and rejected combinations, including the DSA-CP-disabled path.

No.

E2E test with dsv32 + FC1 + FULL_DECODE_ONLY + kv_transfer_config(kv_both) + no layer_sharding

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -1298,7 +1298,7 @@ def enable_dsa_cp_with_o_proj_tp() -> bool:
 
     vllm_config = get_current_vllm_config()
     kv_transfer_config = vllm_config.kv_transfer_config
-    
+
     # In PD-mixed mode, keep the original TP o_proj weight when:
     # 1) KV pooling is disabled, or
     # 2) KV pooling is enabled with kv_role == "kv_both".
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -1283,11 +1283,10 @@ def enable_dsa_cp_with_layer_shard() -> bool:
     from vllm.config import get_current_vllm_config
 
     vllm_config = get_current_vllm_config()
-    # because the broadcast in layer sharding needs to be overlapped with a heavy compute stream to be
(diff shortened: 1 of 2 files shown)


vllm_ascend/utils.py
```diff
@@ -1298,9 +1297,12 @@ def enable_dsa_cp_with_o_proj_tp() -> bool:
...
 
     vllm_config = get_current_vllm_config()
-    # if is PD mix stage, using original TP o_proj weight, and also need to
-    # full gather for o_proj weight for prefill stage.
-    return vllm_config.kv_transfer_config is None
+    kv_transfer_config = vllm_config.kv_transfer_config
+    # In a normal non-PD deployment, or on the PD mixed-role instance, keep the
+    # original TP o_proj weight and full-gather it for the prefill stage.
+    is_non_pd_instance = kv_transfer_config is None
+    is_pd_mixed_instance = kv_transfer_config is not None and kv_transfer_config.kv_role == "kv_both"
```
rjg-lyh: I think the naming can better reflect the pooling and PD-mix concepts.

chenchuw886: Clarified, took KV pooling into consideration

wangxiyuan (approved):

PR #8314, opened 2026-04-15
[BugFix]: order acl graph updates before model forward for ENPU

keyi-zz:

For the ENPU scenario, it is required that device events follow the principle of "record first, wait later", otherwise the inference process may become stuck. However, in the current model_forward function, event.wait precedes event.record. Therefore, for the ENPU scenario, graph parameter updates should be performed before model execution.

N/A

- vLLM version: 


zzzzwwjj (approved):

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1972,7 +1963,41 @@ def _model_forward(
...
+        if self.enable_enpu:
+            self._update_full_graph_params_if_needed(
+                forward_context, num_tokens_padded, positions
+            )
+            hidden_states = run_model()
+        else:
+            hidden_states = run_model()
+            self._update_full_graph_params_if_needed(
+                forward_context, num_tokens_padded, positions
+            )
```
yiz-liu:
```suggestion
        if self.enable_enpu:
            self._update_full_graph_params_if_needed(
                forward_context, num_tokens_padded, positions
            )

        hidden_states = self.model(
            input_ids=input_ids,
            positions=positions,
            intermediate_tensors=intermediate_tensors,
            inputs_embeds=inputs_embeds,
            **model_kwargs,
        )

        if not self.enable_enpu:
            self._update_full_graph_params_if_needed(
                forward_context, num_tokens_padded, positions
            )
```

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1972,7 +1963,41 @@ def _model_forward(
                 self.speculative_config,
                 positions.shape[0],
```
yiz-liu, line 1965: This should cause issues with multi-modal models, I wonder why it's kept like this.

vllm_ascend/utils.py
```diff
@@ -1387,3 +1387,21 @@ def kv_cache_spec_uses_sparse_c8(kv_cache_spec) -> bool:
...
+    libc.getenv.restype = ctypes.c_char_p
+    return libc.getenv
+
+
+def get_c_env(name: str, encoding: str = "utf-8") -> str | None:
+    """Read env via C getenv; returns None if unset."""
+    raw = _libc_getenv()(name.encode(encoding))
+    if raw is None:
+        return None
+    return raw.decode(encoding)
```
yiz-liu, line 1407: Why do we need another helper to get envs? Why not use `env.py`?

vllm_ascend/worker/model_runner_v1.py
1kzk: I don't think so, we need to visualize the code logic, which affects readability.

wangxiyuan (approved):

PR #8333, opened 2026-04-16
[CI] Cache build artifacts with GitHub Actions-Caches

tfhddd:
Current Issue: Due to security concerns, pull requests (PRs) cannot use secrets for CI, making it impossible to leverage Huawei Cloud OBS as a cache.

Solution: Use GitHub Action Cache as the cache.

1. Add a scheduled task to build the vllm-ascend for the main branch daily;

2. When a user submits a pull request, if the C++ code has not changed, the pre-built vllm-ascend can be used directly, skipping the compilation process.

image

image

- vLLM version: v0.19.0
- vLLM main: https://github.com/vllm-project/vllm/commit/5af684c31912232e5c89484c2e8259e0fac6c55b


.github/workflows/schedule_build_vllm_ascend_cache.yaml
```diff
@@ -0,0 +1,256 @@
...
+        run: |
+          pip install uc-manager
+          uv pip install -r requirements-dev.txt
+          if find vllm_ascend -maxdepth 1 -name '*.so' -type f 2>/dev/null | grep -q .; then
+            COMPILE_CUSTOM_KERNELS=0 uv pip install -v -e .
+          else
+            uv pip install -v -e .
+          fi
+
+  image-a3:
```
Potabk: The same workflow should not be repeated; a matrix with different image should be added.

tfhddd: fixed

wangxiyuan (approved):

PR #8347, opened 2026-04-16
[Feature]In TP-asymmetric scenarios, the decoder stores KV-cache.

DreamerLeader:
This PR introduces support for Tensor Parallel (TP) mismatch between KV producers and consumers in the Ascend KV pool by implementing strided I/O operations for KV head mapping and refining scheduler logic for resumed requests. Review feedback identified critical issues: potential memory addressing errors in `_build_strided_addrs` when `use_sparse` is enabled, the absence of `BlockStored` events in the TP-mismatch storage path, and a lack of support for layerwise transfers.

Yes, it enables KV cache transfer between environments with asymmetric Tensor Parallel configurations.

CI passed with existing tests; however, specific tests for TP-asymmetric scenarios should be verified.

- vLLM version: v0.20.2

### DIFF
--- a/tests/ut/ops/test_token_dispatcher.py
+++ b/tests/ut/ops/test_token_dispatcher.py
@@ -275,7 +275,7 @@ class TestTokenDispatcherWithMC2(TestBase):
         self.assertEqual(kwargs["quant_mode"], 4)
         self.assertEqual(kwargs["y_dtype"], torch.float8_e4m3fn)
 
-    def test_mxfp4_dispatch_keeps_quantization_in_mlp_path(self):
+    def test_get_dispatch_mc2_kwargs_with_mxfp4_quant(self):
         hidden_states = torch.randn(10, 128)
         topk_weights = torch.randn(10, 1)
         topk_ids = torch.randint(0, 8, (10, 1))
@@ -292,9 +292,10 @@ class TestTokenDispatcherWithMC2(TestBase):
         )
         kwargs = self.dispatcher.get_dispatch_mc2_kwargs(token_dispatch_input)
 
-        self.assertFalse(token_dispatch_input.quant.dispatch_with_quant)
-        self.assertEqual(kwargs["quant_mode"], 0)
-        self.assertNotIn("y_dtype", kwargs)
+        self.assertTrue(token_dispatch_input.quant.dispatch_with_quant)
+        self.assertEqual(kwargs["quant_mode"], 4)
+        self.assertIn("y_dtype", kwargs)
+        self.assertNotEqual(kwargs["y_dtype"], torch.float8_e4m3fn)
 
         with patch(
             "torch_npu.npu_moe_distribute_dispatch_v2",
@@ -311,8 +312,8 @@ class TestTokenDispatcherWithMC2(TestBase):
             output = self.dispatcher.token_dispatch(token_dispatch_input=token_dispatch_input)
 
         mock_dispatch.assert_called_once()
-        self.assertIsNone(output.dynamic_scale)
-        self.assertFalse(output.combine_metadata.quant.dispatch_with_quant)
+        self.assertIsNotNone(output.dynamic_scale)
+        self.assertTrue(output.combine_metadata.quant.dispatch_with_quant)
 
 
 class TestTokenDispatcherWithAllGather(TestBase):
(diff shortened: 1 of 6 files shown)


Pz1116 (commented):
Here’s a consolidated review for PR #8347 (Ascend KV pool `ascend_store`), focusing on the new TP-mismatch + “decode-side store KV” path. The overall direction (use `effective_tp_size=max(local, peer)` to extend the key namespace and use strided I/O to map head-slices) makes sense, but there are a few high-risk correctness and maintainability issues to address.

## P0 (must-fix before merge)
1) Potential wrong memory addressing in `_build_strided_addrs` when `use_sparse` is enabled
- `per_token_bytes` is derived from `self.block_len[0]` and the same stride is reused for all entries in `self.kv_caches_base_addr`.
- With `use_sparse`, `self.block_len` can contain multiple different lengths (e.g., K/V or sparse layouts differ). Reusing `block_len[0]` for all entries can silently write/read the wrong addresses (data corruption).
- Suggested fix: compute strides per cache entry (e.g., `per_token_bytes[i] = block_len[i] // block_size`, and use entry-specific `per_token_bytes[i]` / `sub_size_bytes[i]` when building addresses). If that’s not feasible, explicitly disable/raise when `tp_mismatch && use_sparse` to avoid unsafe writes.

2) Missing `BlockStored` events on the TP-mismatch store path
- The normal path in `KVCacheStoreSendingThread._handle_request` generates and `update_kv_event(BlockStored)`.
- The TP-mismatch path calls `worker._store_kv_tp_mismatch()` directly and currently does not emit `BlockStored` events.
- This can break monitoring/metrics or any downstream logic relying on KV store events.
- Suggested fix: emit equivalent `BlockStored` events after TP-mismatch puts, and clarify the intended event granularity (per chunk vs per sub-key).

## P1 (strongly recommended)
3) TP-mismatch logic is bypassed when `use_layerwise=True`
- TP-mismatch handling is only wired into the non-layerwise threads (`KVCacheStoreSendingThread` / `KVCacheStoreRecvingThread`).
- If `use_layerwise` is enabled, the code goes through `KVCacheStoreLayer*Thread` and will not use `_load/_store_kv_tp_mismatch`.
- Suggested fix: either (a) add an explicit check at init time to warn/raise/fallback when `tp_mismatch && use_layerwise`, or (b) extend TP-mismatch support to the layerwise threads.

4) Thread-safety / TOCTOU risk around `stored_requests`
- `stored_requests` is protected by `done_task_lock` for add/dec/delete.
- `_store_kv_tp_mismatch` reads `kv_send_thread.stored_requests` without taking the same lock, then later calls `dec_stored_request()`.
- Under concurrency this can lead to TOCTOU issues (wrong early-return / missed decrement).
- Suggested fix: encapsulate the “should store + refcount” gating into a thread-safe method on the sending thread (or ensure the worker-side read is also guarded by the same lock).

5) Validate `lookup_scheduler()` key expansion matches the sub-key naming
- `lookup_scheduler()` now expands `@head_or_tp_rank:{i}` up to `effective_tp_size`, while TP-mismatch stores keys via `_make_sub_key_str(base_key, effective_rank)` where `effective_rank = tp_rank * num_sub_keys + sub_idx`.
- Please verify the sub-key strings are in the same namespace that `lookup_scheduler()` probes; otherwise scheduler hit/miss decisions may be incorrect.
- Suggested fix: add a minimal assertion/test that scheduler expansion covers all TP-mismatch sub-keys for a chunk.

## P2 (design/maintainability)
6) Consider a better integration point than thread -> worker callbacks
- Currently the transfer threads branch on `tp_mismatch` and then call back into `worker._store/_load_kv_tp_mismatch`.
- This increases coupling and tends to grow more conditionals as more layouts/modes appear (layerwise, sparse, etc.).
- Suggested refactor: introduce a small “transfer plan / address builder” (or move into `ChunkedTokenDatabase`) that produces `(keys, addrs, sizes)` for both normal and TP-mismatch paths; threads only execute `exists/get/put`. KV events (`BlockStored`) should be emitted by the sending thread after executing the plan.

7) Avoid string `replace()` in `lookup_scheduler()` if possible
- Expanding keys via `str.replace("@head_or_tp_rank:0", ...)` is fragile if key format evolves.
- Suggested improvement: provide a structured API on `KeyMetadata` to generate variants (e.g., `with_head_or_tp_rank(i)`).

Additional note: `pool_scheduler.py` changes the resumed-request handling around `new_block_ids` (moving the early-continue and removing the `num_computed_tokens` guard). Please regression-check edge cases to ensure it doesn’t generate extra `ReqMeta` or change behavior for requests that have already computed the full prompt.

Happy to follow up with concrete patch suggestions for entry-wise stride calculation and `BlockStored` emission semantics for sub-keys.

PR #8354, opened 2026-04-16
[BugFix] fix hang in async scheduling while open ENPU

keyi-zz:

1. there is no synchronization between steps. However, in async scheduling with aclgraph, it is possible that the CPU's record event for the current iteration completes before the previous iteration's graph execution has finished. If cpu is fast enough, device will hang on event_wait in interation i+1 (assume that event_record is executed immediately on update stream of device).
2. Under ENPU, eagle proposers also need to follow event.record first, and then event.Wait.

N/A


vllm_ascend/compilation/acl_graph.py
```diff
@@ -197,12 +205,10 @@ def __call__(self, *args, **kwargs):
...
-            self.vllm_config.speculative_config.method in ("eagle", "eagle3")
-            if self.vllm_config.speculative_config
-            else False
-        )
-        if self.runtime_mode != CUDAGraphMode.FULL or not _EXTRA_CTX.is_draft_model or not use_eagle:
+        # When enable_enpu is on, model_runner orders update vs replay; skip here.
+        # When FULL + EAGLE draft (merge path), replay does not need this barrier.
+        is_draft_eagle = _EXTRA_CTX.is_draft_model and self.use_eagle
+        if not self.enable_enpu and not is_draft_eagle:
             torch.npu.current_stream().synchronize()
```
yiz-liu, line 209: The only sync we need when `mode=FULL` is the one between two steps, i.e., after the draft model and before the next target model.

We don't need any sync in between the target model and draft model, or between different steps of spec, anymore.

yiz-liu (approved):

Issue #8336, opened 2026-04-16
[Bug]: qwen3.5 35B w8a8模型，910b main镜像，压测一压就崩（raise AttributeError(name) from None  AttributeError: WAITING_FOR_FSM）


xuyucheng220:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

python collect_env.py
Collecting environment information...
PyTorch version: 2.9.0+cpu
Is debug build: False

OS: Ubuntu 22.04.5 LTS (aarch64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
Clang version: 15.0.7
CMake version: version 4.3.1
Libc version: glibc-2.35

Python version: 3.11.14 (main, Feb 26 2026, 04:49:14) [GCC 11.4.0] (64-bit runtime)
Python platform: Linux-5.10.0-136.12.0.86.r1526_92.hce2.aarch64-aarch64-with-glibc2.35

CPU:
Architecture:                       aarch64
CPU op-mode(s):                     64-bit
Byte Order:                         Little Endian
CPU(s):                             192
On-line CPU(s) list:                0-191
Vendor ID:                          HiSilicon
BIOS Vendor ID:                     HiSilicon
Model name:                         Kunpeng-920
BIOS Model name:                    HUAWEI Kunpeng 920 5250
Model:                              0
Thread(s) per core:                 1
Core(s) per socket:                 48
Socket(s):                          4
Stepping:                           0x1
BogoMIPS:                           200.00
Flags:                              fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma dcpop asimddp asimdfhm ssbs
L1d cache:                          12 MiB (192 instances)
L1i cache:                          12 MiB (192 instances)
L2 cache:                           96 MiB (192 instances)
L3 cache:                           192 MiB (8 instances)
NUMA node(s):                       8
NUMA node0 CPU(s):                  0-23
NUMA node1 CPU(s):                  24-47
NUMA node2 CPU(s):                  48-71
NUMA node3 CPU(s):                  72-95
NUMA node4 CPU(s):                  96-119
NUMA node5 CPU(s):                  120-143
NUMA node6 CPU(s):                  144-167
NUMA node7 CPU(s):                  168-191
Vulnerability Gather data sampling: Not affected
Vulnerability Itlb multihit:        Not affected
[...]

PR #8358, opened 2026-04-17
[BugFix] Fix AttributeError for WAITING_FOR_FSM after upstream rename

ivyilike:
This PR fixes an AttributeError: WAITING_FOR_FSM error that occurs when running with the latest vLLM upstream.

No. This is an internal fix to align with upstream vLLM API changes.

- vLLM version: v0.19.0
- vLLM main: https://github.com/vllm-project/vllm/commit/5af684c31912232e5c89484c2e8259e0fac6c55b


xuyucheng220:
Hi, have you fixed this problem yet or not ? 
I got this problem too:  https://github.com/vllm-project/vllm-ascend/issues/8336
It seems to be the same problem.

PR #8389, opened 2026-04-17
[Feature][Attention] Add Quest sparse decode

MarcelFerrari:
Add an optional Quest sparse decode path for the v1 Ascend attention backend, enabled through `additional_config.quest_decode_config`. Quest builds compact per-page key metadata from the local KV cache, scores those metadata pages with the current decode query, selects the most relevant logical KV pages, maps them through the paged block table, and runs selected-page paged attention.

Quest request/layer bookkeeping is centralized in `vllm_ascend/attention/quest_decode.py`. The v1 model runner now only owns the metadata manager lifecycle and passes active request ids into attention metadata; dense decode remains the default path and runtime fallback.

This PR adds an optional sparse decode path for long-context v1 Ascend attention workloads.

Long-context decode can spend a large share of time attending over KV pages that contribute little to the next token. Quest keeps per-page max/min key metadata, uses the current query to select the top-k useful KV pages, and computes attention only over those selected pages. This keeps the existing dense attention path unchanged by default while providing a faster decode option for compatible long-context workloads. Reference paper: https://arxiv.org/abs/2406.10774

The feature is disabled by default and can be enabled through `additional_config.quest_decode_config`.

### How it works

At a high level, the PR adds four pieces:

- `QuestDecodeConfig` parsing and validation for `additional_config.quest_decode_config`.
- `QuestDecodeMetadataManager`, which validates runtime compatibility, allocates per-layer max/min metadata buffers, tracks which request owns each batch row, invalidates stale rows when requests change, and refreshes metadata when a request crosses a Quest page boundary.
- v1 attention integration that carries `QuestBatchMetadata` through `AscendCommonAttentionMetadata`/`AscendMetadata`, refreshes layer metadata after KV cache update, and dispatches to sparse selected-page attention for eligible decode-only batches.
- Ascend custom ops for Quest metadata generation, Quest page selection, and selected-page paged attention, including out/workspace variants used by graph capture/update paths.

The sparse path is selected only after the metadata manager confirms the model/runtime is compatible and the current decode batch is worth sparsifying. Short or already-small decode batches fall back to dense attention when the selected pages would cover at least half of the available logical KV pages on average.

### Main files added or changed

- `vllm_ascend/ascend_config.py`: adds `quest_decode_config` parsing and validation.
- `vllm_ascend/attention/quest_decode.py`: centralizes Quest config lookup, batch metadata handles, per-layer metadata freshness tracking, compatibility checks, metadata allocation, and sparse decode input assembly.
- `vllm_ascend/attention/attention_v1.py`: carries Quest metadata through the attention backend, refreshes metadata after KV cache writes, dispatches selected-page sparse decode, and updates graph tasks for Quest attention.
- `vllm_ascend/attention/utils.py`: propagates Quest manager/request ids through common attention metadata and unpadded metadata views.
- `vllm_ascend/worker/model_runner_v1.py`: creates and initializes the Quest metadata manager and passes active request ids into attention metadata.
- `vllm_ascend/ops/select_attention.py`: adds Python wrappers for the new custom ops.
- `csrc/attention/paged_select_attention/`: selected-page paged attention ACLNN op and kernel implementation.
- `csrc/attention/quest_prefill_metadata/`: Quest max/min metadata generation ACLNN op and kernel implementation.
- `csrc/attention/quest_block_select_paged/`: Quest metadata scoring and top-k page-selection ACLNN op and kernel implementation.
- `csrc/torch_binding.cpp`, `csrc/torch_binding_meta.cpp`, and `csrc/build_aclnn.sh`: register and build the new custom ops.

Yes. It adds a new optional `additional_config` entry:

```bash
--additional-config='{"quest_decode_config": {"enable": true, "topk_pages": 8}}'
```

`topk_pages` controls how many logical KV pages Quest selects per decode request. It must be a positive integer when Quest is enabled. Larger values keep more context in the sparse attention step; smaller values usually improve latency more aggressively.

The feature is currently gated to compatible v1 decode workloads:

- Ascend A2/A3 hardware supported by the new custom ops
- local KV cache only; `kv_transfer_config` is not supported
- v1 decode-only batches with `max_query_len == 1`
- standard 4D KV cache tuple layout
- block size 128 and head size 128
- at most 6 Quest metadata blocks per request, derived from `max_model_len`/`max_encoder_len`
- no context parallelism
- no MLA attention
- no xLite graph execution
- no full cudagraph mode
- no existing sparse attention mode
- no sliding window or sinks
- no encoder-decoder attention

If Quest is enabled but any model, runtime, or batch condition is incompatible, the manager disables Quest for that path and falls back to existing dense decode.

### Benchmark results

The benchmark below compares dense decode against Quest decode on long-context datasets from LongBench with `max_num_seqs=30`. Since the target is medium-long context, we filter the dataset by prompt length and report the number of samples.

- `multi_news` has mostly small queries, so we adjust the lower bound to 4800 tokens to include at least 20 samples.
- `gov_report` has many long queries, so we increase the minimum length to 10k tokens as well as the number of kept pages to 16.

Across the four measured workloads, Quest gives an average `speedup_pct` of about 27.5%, with individual speedups from 24.5% to 30.4%. ROUGE changes stay within -2.3% to +5.8%, with an average change of about +1.1%.

|    | dataset    |   max_num_seqs |   min_len |   max_len |   decode_tokens |   Quest_topk_pages |   num_samples |   dense_rouge |   rouge_score |   rouge_change_pct |   dense_time_s |   completion_time_s |   speedup_pct |
|---:|:-----------|---------------:|----------:|----------:|----------------:|-------------------:|--------------:|--------------:|--------------:|-------------------:|---------------:|--------------------:|--------------:|
|  1 | dureader   |             30 |      5000 |     32000 |             512 |                  8 |           200 |        0.1749 |        0.178  |           1.77244  |       568.337  |            446.374  |       27.3229 |
|  3 | gov_report |             30 |     10000 |     32000 |            1024 |                 16 |            86 |        0.3072 |        0.3002 |          -2.27865  |       609.998  |            467.68   |       30.4306 |
|  5 | multi_news |             30 |      4800 |     32000 |             512 |                  8 |            20 |        0.2137 |        0.2261 |           5.80253  |        42.2606 |             33.9486 |       24.484  |
|  7 | musique    |             30 |      5000 |     32000 |             512 |                  8 |           150 |        0.3765 |        0.3728 |          -0.982736 |       473.117  |            369.757  |       27.9533 |

benchmark_plots

- Ran the long-context benchmark above for dense decode and Quest decode on an Atlas A2 910B4 NPU machine.

- vLLM version: v0.23.0


csrc/paged_select_attention/op_kernel/paged_select_attention_kernel.h
```diff
@@ -0,0 +1,1024 @@
...
+            AscendC::SetFlag<AscendC::HardEvent::V_MTE2>(EVENT_ID0);
+            AscendC::SetFlag<AscendC::HardEvent::V_MTE2>(EVENT_ID1);
+            AscendC::SetFlag<AscendC::HardEvent::V_MTE2>(EVENT_ID2);
+            AscendC::SetFlag<AscendC::HardEvent::V_MTE2>(EVENT_ID3);
+
+            EpilogueOnlineSoftmax epilogueOnlineSoftmax(resource, scaleValue);
+            EpilogueRescaleO epilogueRescaleO(resource);
+            EpilogueInitOut epilogueInitOut(resource);
+
+            coreIdx = AscendC::GetBlockIdx() / AscendC::GetSubBlockNum();
```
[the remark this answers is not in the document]
MarcelFerrari: I do not think this should be removed. This logic is intentional for the `KERNEL_TYPE_MIX_AIC_1_2` execution model and was inherited from the ops-transformer `fused_infer_attention_score` donor kernel used to implement paged_select_attention.

In the AIV path, `GetBlockIdx()` identifies the vector subblock, while the task schedule and DB workspace are indexed by the AIC group. Dividing by `GetSubBlockNum()` maps the two AIV subblocks in a 1C2V group back to the shared AIC group index. The two AIV subblocks are then distinguished where needed through `GetSubBlockIdx()`.

This exact pattern exists in the original donor code:

`attention/fused_infer_attention_score/op_kernel/flash_attention_regular.h`

```cpp
coreIdx = AscendC::GetBlockIdx() / AscendC::GetSubBlockNum();
```

The donor op is launched with `KERNEL_TYPE_MIX_AIC_1_2`.

The same mapping also appears in other vllm-ascend mix kernels, including:

* `lightning_indexer`
* `lightning_indexer_quant`
* `sparse_flash_attention`
* `dispatch_gmm_combine_decode`

These kernels consistently map AIV block indices back to the shared AI-core group.

Removing this division would cause the vector path to index a different task/workspace slot than its paired cube core. Since workspace layout and synchronization flags are defined per AIC group, this would likely break producer–consumer synchronization rather than fix anything.

_Note: This analysis was generated by Codex._

csrc/paged_select_attention/op_kernel/paged_select_attention_kernel.h
```diff
@@ -0,0 +1,1024 @@
...
+                kvSIdx,
+                sparseHeadState,
+                windowPhysicalIds,
+                windowPhysicalCount);
+        }
+
+        __aicore__ inline
+        void operator()(FAIKernelParams const &params)
+        {
+            __gm__ PagedSelectAttentionTilingData *fATilingData =
```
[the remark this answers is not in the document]
MarcelFerrari: This looks like a valid hygiene/performance cleanup, but not a critical performance/correctness bug.

The current paged-select path copies tiling data locally at the kernel entry point:

- `csrc/paged_select_attention/op_kernel/paged_select_attention.cpp`
- `csrc/paged_select_attention/op_kernel/paged_select_attention_apt.cpp`

Both call:
```cpp
PAGED_SELECT_COPY_TILING_DATA(PagedSelectAttentionTilingData, tiling);
````

However, the templated bridge still receives the original GM `tiling` pointer, stores it in `FAIKernelParams`, and `FAInferKernel::operator()` later casts it back to:

```cpp
__gm__ PagedSelectAttentionTilingData *fATilingData =
    reinterpret_cast<__gm__ PagedSelectAttentionTilingData *>(params.tiling);
```

So the reviewer is correct that the local tiling copy is not reused in the split-fuse worker path.

That said, the comment appears to overstate the severity. The tiling fields are read from GM once at the beginning of `operator()` and copied into local scalar variables before the task loop. They are not re-read inside every task iteration. So this does not look like a correctness issue, and the performance impact should be limited to duplicated per-core setup reads.

The pattern also appears to have been inherited from the original ops-transformer donor kernel. In the clean `fused_infer_attention_score` donor path, the entry point calls:

```cpp
FIA_COPY_TILING_DATA(FAInferTilingData, tiling);
```

at [attention/fused_infer_attention_score/op_kernel/fused_infer_attention_score.cpp:50](https://gitcode.com/cann/ops-transformer/tree/8.5.0/attention/fused_infer_attention_score/op_kernel/fused_infer_attention_score.cpp#L50).

But the donor path then still passes the raw `tiling` pointer through `FAInfer(...)`; for example, `tiling` is part of the `FAInfer` signature at [attention/fused_infer_attention_score/op_kernel/flash_attention_interface.cpp:40](https://gitcode.com/cann/ops-transformer/tree/8.5.0/attention/fused_infer_attention_score/op_kernel/flash_attention_interface.cpp#L40), and it is stored in `FAIKernelParams` at [attention/fused_infer_attention_score/op_kernel/flash_attention_interface.cpp:104](https://gitcode.com/cann/ops-transformer/tree/8.5.0/attention/fused_infer_attention_score/op_kernel/flash_attention_interface.cpp#L104).

The donor `flash_attention_regular.h` then casts that pointer back to GM tiling data:

```cpp
__gm__ FAInferTilingData *fATilingData =
    reinterpret_cast<__gm__ FAInferTilingData *>(params.tiling);
```

at [attention/fused_infer_attention_score/op_kernel/flash_attention_regular.h:74](https://gitcode.com/cann/ops-transformer/tree/8.5.0/attention/fused_infer_attention_score/op_kernel/flash_attention_regular.h#L74).

So this is not a new suspicious access pattern introduced by this PR; it came from the ops-transformer split-fuse implementation.

At the same time, several vllm-ascend kernels use the cleaner local-copy style. For example, `sparse_flash_attention` and `lightning_indexer` copy tiling data at the entry point and pass `tiling_data` down into `op.Init(...)`. The QUEST helper kernels also use the local copy directly.

Conclusion: the feedback is reasonable as a cleanup suggestion. It is probably worth updating the paged-select path to pass `const PagedSelectAttentionTilingData *__restrict tiling_data` through `Run` / `operator()` instead of recasting `params.tiling` from GM. However, unless there is strong evidence that the current behavior is incorrect or causing measurable issues, I would be cautious about modifying code paths inherited from the donor kernel, especially where the original rationale for the design is not fully clear.

Note: this analysis was generated by Codex.

csrc/paged_select_attention/op_kernel/paged_select_attention_kernel_bridge.h
```diff
@@ -0,0 +1,117 @@
...
+ */
+
+#include "kernel_operator.h"
+#include "paged_select_attention_kernel.h"
+
+using namespace NpuArch;
+
+namespace PagedSelectAttentionKernel {
+template <typename InputDtype = half, typename IntermCalcPrec = float>
+__global__ __aicore__ void Run(
```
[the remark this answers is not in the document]
MarcelFerrari: This does not look like a real correctness issue. There is direct donor-code precedent for this exact pattern.

In the PR code, `Run` is declared as:

[csrc/paged_select_attention/op_kernel/paged_select_attention_kernel_bridge.h:23](https://github.com/vllm-project/vllm-ascend/tree/64c2a1b077f4bbdef7cfba44e2f4fb9a1d869884/csrc/paged_select_attention/op_kernel/paged_select_attention_kernel_bridge.h#L23)

```cpp
template <typename InputDtype = half, typename IntermCalcPrec = float>
__global__ __aicore__ void Run(...)
````

and it is called from the real op entry point here:

[csrc/paged_select_attention/op_kernel/paged_select_attention.cpp:51](https://github.com/vllm-project/vllm-ascend/tree/64c2a1b077f4bbdef7cfba44e2f4fb9a1d869884/csrc/paged_select_attention/op_kernel/paged_select_attention.cpp#L51)
[csrc/paged_select_attention/op_kernel/paged_select_attention_apt.cpp:51](https://github.com/vllm-project/vllm-ascend/tree/64c2a1b077f4bbdef7cfba44e2f4fb9a1d869884/csrc/paged_select_attention/op_kernel/paged_select_attention_apt.cpp#L51)

```cpp
PagedSelectAttentionKernel::Run<half, float>(...);
```

The reviewer’s claim would be correct under CUDA-style assumptions, where `__global__` kernels are not called like ordinary device functions. But the donor Ascend C code uses the same structure.

The clean ops-transformer donor declares the split-fuse helper as:

[attention/fused_infer_attention_score/op_kernel/flash_attention_interface.cpp:29](https://gitcode.com/cann/ops-transformer/tree/8.5.0/attention/fused_infer_attention_score/op_kernel/flash_attention_interface.cpp#L29)

```cpp
__global__ __aicore__ void FAInfer(...)
```

and then calls it from the top-level `fused_infer_attention_score` kernel as a regular function:

[attention/fused_infer_attention_score/op_kernel/fused_infer_attention_score.cpp:88](https://gitcode.com/cann/ops-transformer/tree/8.5.0/attention/fused_infer_attention_score/op_kernel/fused_infer_attention_score.cpp#L88)

```cpp
SplitFuse::FAInfer<half, half, float, false, ...>(...);
```

So this pattern was inherited from the original ops-transformer `fused_infer_attention_score` split-fuse implementation. It is not a new suspicious construct introduced by this PR.

Note: this analysis was generated by Codex.

csrc/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
```diff
@@ -0,0 +1,632 @@
...
+            static_cast<int32_t>(tiling_data->blockSize),
+            static_cast<int32_t>(tiling_data->headDim),
+            static_cast<int32_t>(tiling_data->maxMetadataBlocksPerRequest),
+            tiling_data->tokensSinceMetadataUpdate,
+            static_cast<int32_t>(tiling_data->k));
+        return;
+    }
+
+    if (TILING_KEY_IS(QUEST_BLOCK_SELECT_PAGED_TILING_BF16)) {
+        quest_block_select_paged_bfloat16_impl(
```
mitseng, line 613: quest_block_select_paged_bfloat16_impl & quest_block_select_paged_half_impl are highly duplicated, should use template.

csrc/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
```diff
@@ -0,0 +1,632 @@
...
+
+            AscendC::DataCopy(maxblock_lt, maxblocks_gm[meta_block_offset], gm_ub_cp);
+            AscendC::SetFlag<AscendC::HardEvent::MTE2_V>(EVENT_ID0);
+            AscendC::DataCopy(minblock_lt, minblocks_gm[meta_block_offset], gm_ub_cp);
+            AscendC::SetFlag<AscendC::HardEvent::MTE2_V>(EVENT_ID1);
+
+            uint8_t repeat_times = block_size;
+            AscendC::BinaryRepeatParams mul_repeat_params = {1, 1, 1, 8, 0, 8};
+            AscendC::WaitFlag<AscendC::HardEvent::MTE2_V>(EVENT_ID0);
+            AscendC::Mul(maxblock_lt, query_lt, maxblock_lt, mask, repeat_times, mul_repeat_params);
```
mitseng, line 518: in bf16, q and k are casted to float before Mul. should do same thing for fp16.

csrc/paged_select_attention/op_kernel/paged_select_attention_kernel.h
```diff
@@ -0,0 +1,1024 @@
...
+                    "selected_kv_indices sparse contract violation: logical page id rowBase=%llu idx=%u value=%d validBlocks=%u",
+                    static_cast<unsigned long long>(selectedRowBase), i, blockId, validLogicalBlockCount);
+                if (skipCount > 0U) {
+                    --skipCount;
+                    continue;
+                }
+                if (windowPhysicalCount >= KernelCommon::MAX_SELECTED_KV_STACK_BLOCKS) {
+                    break;
+                }
+                int32_t physicalBlockIdRaw = gBlockTable.GetValue(blockBOffset + static_cast<uint32_t>(blockId));
```
mitseng, line 209: should verify physicalBlockIdRaw

csrc/quest_prefill_metadata/op_kernel/quest_prefill_metadata.cpp
```diff
@@ -0,0 +1,376 @@
...
+            minblocks,
+            tiling_data);
+        return;
+    }
+    #else
+    if (TILING_KEY_IS(QUEST_PREFILL_METADATA_TILING_FP16)) {
+        RunQuestPrefillMetadata<half>(
+            k_cache,
+            block_tables,
+            seq_lens,
```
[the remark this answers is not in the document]
MarcelFerrari: Fixed in commit `baa4d0180c31f09b421d669d5f1972ef0d6cb59f` by removing the BF16-as-FP16 dispatch workaround and using chunked FP32 reduction for BF16 metadata.

csrc/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
MarcelFerrari: In bf16 we cast to fp32 because the required operations are not supported in bf16 datatype on A2 and A3. fp16 is supported natively so there is no need to cast.

csrc/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
MarcelFerrari: I rewrote/refactored the kernel to use templates + type traits. The branching logic for fp16 vs bf16 is handled with compile-time branches. The new code is not shorter, but hopefully it is easier to read and understand. Tested and confirmed correctness and no performance impact.

csrc/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
```diff
@@ -0,0 +1,683 @@
...
+    GM_ADDR metadata_block_tables,
+    GM_ADDR seq_lens,
+    GM_ADDR selected_indices,
+    GM_ADDR workspace,
+    GM_ADDR tiling)
+{
+    (void)workspace;
+    QUEST_BLOCK_SELECT_PAGED_COPY_TILING_DATA(QuestBlockSelectPagedTilingData, tiling);
+
+    if (TILING_KEY_IS(QUEST_BLOCK_SELECT_PAGED_TILING_FP16)) {
```
mitseng: tiling key doesn't need to depend on the dtype，it may lead to a larger binary size

csrc/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
mitseng: fp16 should also cast to fp32 for better numerical stability and accuracy, as the following operations is Mul

csrc/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
MarcelFerrari: Fixed, we pass datatype as a tiling data field and use the same key for both bf16 and fp16 paths

csrc/attention/paged_select_attention/op_kernel/paged_select_attention_kernel.h
```diff
@@ -0,0 +1,1024 @@
...
+                    "selected_kv_indices sparse contract violation: logical page id rowBase=%llu idx=%u value=%d validBlocks=%u",
+                    static_cast<unsigned long long>(selectedRowBase), i, blockId, validLogicalBlockCount);
+                if (skipCount > 0U) {
+                    --skipCount;
+                    continue;
+                }
+                if (windowPhysicalCount >= KernelCommon::MAX_SELECTED_KV_STACK_BLOCKS) {
+                    break;
+                }
+                int32_t physicalBlockIdRaw = gBlockTable.GetValue(blockBOffset + static_cast<uint32_t>(blockId));
```
MarcelFerrari: fixed, verified similarly to logical block id

csrc/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
MarcelFerrari: According to our tests and the original paper, i do not think there is a big stability/accuracy concern. In any case, it makes the kernel simpler and easier to template so I added cast for fp16 path too.

csrc/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
MarcelFerrari: Applied to metadata generation kernel too

MarcelFerrari:
Tested correctness and performance with vllm v0.20.1

MengqingCao:
@ZT-AIA could you also take a look at the kernel? thx!

vllm_ascend/attention/attention_v1.py
```diff
@@ -210,6 +223,9 @@ class AscendMetadata:
 
     kvcomp_metadata: KVCompMetaData | None = None
 
+    # QUEST active-batch handle.
+    quest_metadata: QuestBatchMetadata = field(default_factory=QuestBatchMetadata)
```
MengqingCao: maybe we should move this into the backend `quest_decode`, and we could inherit form the base classes in `attention_v1.py`

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,537 @@
...
+from vllm.model_executor.layers.attention_layer_base import AttentionLayerBase
+from vllm.utils.math_utils import cdiv
+
+from vllm_ascend.ascend_config import AscendConfig, QuestDecodeConfig
+from vllm_ascend.attention.utils import enable_cp
+from vllm_ascend.ops.select_attention import quest_prefill_metadata
+from vllm_ascend.utils import AscendDeviceType, get_ascend_device_type
+
+QUEST_PAGE_SIZE = 128
+QUEST_HEAD_SIZE = 128
```
MengqingCao, line 50: Could you explain more on why the `head_size` only supprt to be 128 here? Moreover, I think it is more reasonable to implement a `QuestBackend` to cover the `supports_head_size`, etc.

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,537 @@
...
+        self.refresh_start_seq_lens = torch.zeros((max_num_reqs,), dtype=torch.int32, device=device)
+        self.refresh_start_seq_lens_cpu_tensor = torch.zeros(
+            (max_num_reqs,),
+            device="cpu",
+            dtype=torch.int32,
+            pin_memory=pin_memory,
+        )
+        self.refresh_start_seq_lens_cpu = self.refresh_start_seq_lens_cpu_tensor.numpy()
+        self.refresh_seq_lens = torch.zeros((max_num_reqs,), dtype=torch.int32, device=device)
+        self.refresh_seq_lens_cpu_tensor = torch.zeros(
```
MengqingCao: Could you make some comments on what do `self.refresh_seq_lens` and `self.refresh_start_seq_lens` refer to?

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,537 @@
...
+            (max_num_reqs,),
+            device="cpu",
+            dtype=torch.int32,
+            pin_memory=pin_memory,
+        )
+        self.refresh_seq_lens_cpu = self.refresh_seq_lens_cpu_tensor.numpy()
+
+    def invalidate_rows(self, row_indices: Sequence[int]) -> None:
+        for row_idx in row_indices:
```
MengqingCao, line 180: Does the sparse selection of kv entries rely on this mask?

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,537 @@
...
+        if not self.quest_enabled_for_batch or self._manager is None:
+            return None
+        return self._manager._get_sparse_decode_inputs(
+            self,
+            layer_name=layer_name,
+        )
+
+
+class _QuestLayerMetadata:
+    """Per-layer QUEST metadata tensors and freshness bookkeeping."""
```
MengqingCao, line 146: Let's make some docstring for all the member functions.

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,537 @@
...
+            else:
+                crossed_page_boundary = seq_len // QUEST_PAGE_SIZE > valid_tokens // QUEST_PAGE_SIZE
+                if crossed_page_boundary:
+                    self.refresh_start_seq_lens_cpu[row_idx] = (valid_tokens // QUEST_PAGE_SIZE) * QUEST_PAGE_SIZE
+                    self.refresh_seq_lens_cpu[row_idx] = seq_len
+            if self.refresh_seq_lens_cpu[row_idx] > self.refresh_start_seq_lens_cpu[row_idx]:
+                refresh_required = True
+
+        if refresh_required:
+            self.refresh_start_seq_lens[:num_reqs].copy_(
```
MengqingCao: maybe we could use `CpuGpuBuffer` to avoid some redundant code.

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,537 @@
...
+    def prepare(
+        self,
+        batch_metadata: QuestBatchMetadata,
+    ) -> bool:
+        num_reqs = batch_metadata.batch_size
+        if num_reqs <= 0:
+            return False
+
+        self.refresh_start_seq_lens_cpu[:num_reqs].fill(0)
+        self.refresh_seq_lens_cpu[:num_reqs].fill(0)
```
MengqingCao: nits: Let's make the inplace operation more clearly.
```suggestion
        self.refresh_start_seq_lens_cpu[:num_reqs].fill_(0)
        self.refresh_seq_lens_cpu[:num_reqs].fill_(0)
```

MengqingCao (commented):
Thanks for your effort! I've not finish the review but will do it ASAP

vllm_ascend/attention/attention_v1.py
MarcelFerrari: Hi,
thank you for taking the time to review this big PR.

This is a good point. I initially ditched the idea because quest relied on layer names to fetch the correct metadata and this meant that implementing quest as a dedicated attention backend required a lot of code duplication. After some thoughts I think I found a way to get around this. We do not need to key runtime metadata specifically by layer_name; we can bind each _QuestLayerMetadata to the per-layer attention impl during manager initialization and use the impl instance as the stable layer identity. Then QuestAttentionBackendImpl can inherit dense fallback behavior from AscendAttentionBackendImpl, override only the sparse-decode hook/forward_impl, and call super().forward_impl(...) when QUEST is not applicable.

This is a big refactor so i am preparing a draft version of this on a different branch and will share it as soon as it is ready. One thing is that we will need to edit `vllm_ascend/platform.py` file to return the quest attention backend when it is enabled.

vllm_ascend/attention/quest_decode.py
MarcelFerrari: This is a limitation imposed by the QUEST selection and metadata kernels that i ported from [ops-transformer](https://gitcode.com/cann/ops-transformer/tree/9.0.0/experimental/select_attention_operators). 

For each KV page, QUEST stores only channelwise metadata: one max vector and one min vector, so `2 * head_size` values per KV page. The current kernels are specialized for the common case where `block_size = 128` and `head_size = 128`, so one metadata block contains summaries for 128 KV pages with shape `[128 pages, 128 channels]`.

During decode, `quest_block_select_paged` scores a whole metadata block against the query. It loads/processes the full `128 x 128` max/min metadata tile and keeps score/sort buffers for up to 6 such metadata blocks. The kernel does not currently tile the channel dimension, so increasing `head_size` would increase the width of every metadata row and the local buffer usage/tiling assumptions inside the kernel.

So supporting `head_size != 128` would require changing the metadata tiling logic in the QUEST kernels.

vllm_ascend/attention/quest_decode.py
MarcelFerrari: ~~`refresh_start_seq_lens[row]` is the inclusive token offset where QUEST metadata refresh should begin for that request row, and `refresh_seq_lens[row]` is the exclusive end/current sequence length for that refresh. Together they describe the token range `[start, end)` whose page metadata should be rebuilt. New, stale, or shrunk rows refresh from 0 to seq_len; normal decode refreshes only after crossing a QUEST page boundary, starting from the previous 128-token page boundary.~~

~~E.g. imagine you have a query with 350 input tokens. After prefill you have almost 3 pages of KV cache. The first metadata refresh needs to compress all _complete_ pages.~~

```
seq_len = 350

0          128         256     350 384         512
|-----------|-----------|--------   |           |
[== metadata refresh ===)

refresh_start_seq_lens[row] = 0
refresh_seq_lens[row]       = 256

refresh range = [0, 256)
```

~~Once you do enough decode, your KV history grows to 385 tokens. This triggers another refresh. Since the previously computed metadata is good, there is no need to refresh everything, rather just the newly completed block. This helps drastically reduce the overhead of metadata generation.~~

```
seq_len = 385

0          128         256         384         512
|-----------|-----------|-----------|-          |
                        [= refresh =)

refresh_start_seq_lens[row] = 256
refresh_seq_lens[row]       = 385

refresh range = [256, 384)
```

Note: this information is now partially outdated. The code has now been simplified to only store refresh_start_seq_lens and refresh_end_seq_lens. These track respectively the last token with valid metadata and the token up to which the refresh kernel  must run.

vllm_ascend/attention/quest_decode.py
MarcelFerrari: Not directly. `valid_tokens` is CPU-side freshness bookkeeping, not an input mask for top-k KV selection. It marks whether a batch row’s cached QUEST metadata is valid for the current request and sequence length. `prepare()` uses it to decide which metadata pages need refresh, and `is_ready_for_sparse_decode()` prevents sparse decode when metadata is stale. The actual sparse selection kernel uses the query, sequence lengths, block tables, and QUEST max/min metadata blocks and outputs a per-query-head sparsity mask.

`valid_tokens` is similar to `refresh_start_seq_lens` but they have different roles. `valid_tokens` is persistent CPU-side freshness bookkeeping: it records how far the cached QUEST metadata is valid for each continuous-batching row, and it is invalidated when a row is reused by another request.

`refresh_start_seq_lens` and `refresh_seq_lens` are temporary per-step kernel inputs. `prepare()` rewrites them each step. For rows that do not need refresh, we set `[start, end) = [0, 0)`, and the metadata kernel skips that row.

When a row does need refresh after crossing a QUEST page boundary, `refresh_start_seq_lens` is snapped down to the page boundary for the current partially-refreshed page, while `refresh_seq_lens` is set to the current sequence length.

Example:

```text
QUEST page size = 128

old valid_tokens = 350
new seq_len      = 385

0          128         256         384         512
|-----------|-----------|-----------|-          |
                        [= refresh =)

refresh_start_seq_lens = floor(valid_tokens/128) * 128 = floor(350 / 128) * 128 = 256
refresh_seq_lens       = 384

refresh range:  [256, 384)
```

So `refresh_start_seq_lens` is not `350`; it snaps back to `256` so the whole formerly-partial page is rebuilt.

vllm_ascend/attention/quest_decode.py
MarcelFerrari: Yes this is a good idea. I will do this as soon as soon as i am done with this:
> maybe we should move this into the backend `quest_decode`, and we could inherit form the base classes in `attention_v1.py`

vllm_ascend/attention/quest_decode.py
MarcelFerrari: Yes. I did not know there was a class for this but it seems like a good use case. I will update once i am done with the attention_v1 refactor.

vllm_ascend/attention/quest_decode.py
MarcelFerrari: In the current implementation these are CPU NumPy arrays so this does not work. Once we switch to `CpuGpuBuffers` then it makes sense.

vllm_ascend/attention/attention_v1.py
MarcelFerrari: ~~Here is a preview of a possible refactor strategy to move quest decode logic as its own backend https://github.com/vllm-project/vllm-ascend/commit/96082ee545c540e7f40bbe239d882e932cfcfafa~~

~~Please take a look at this and let me know if this is acceptable so that i can clean up the implementation.~~

~~Specifically take a look at how i changed `vllm_ascend/platform.py` and if a similar change to this file is acceptable.~~

~~Thanks!~~

See new commit

MarcelFerrari:
Hi @MengqingCao

I pushed a commit to refactor quest as its own attention backend + address the other issues you pointed out and simplify the code logic. Now no changes are made to `attention_v1.py`. I also added docstrings and renamed some variables/methods to be more explicit so this should make the review process simpler.

Thank you again for your help and time!

**Note about CI test:**
The Ruff failure is coming from the synthetic PR merge commit used by GitHub Actions, not from the branch head directly.

On my branch, `get_ascend_config` is already imported in `platform.py`, so local pre-commit passes. However, the target branch has changed in the same file since my branch point. In the merge result used by CI, the new QUEST helper still calls `get_ascend_config()`, but the merged import line only includes `init_ascend_config`, which triggers Ruff F821.

I can fix this, but it requires rebasing/merging the latest target branch and resolving the `platform.py` drift. I’ll do that once I get the OK from you, since it may touch more than the Ruff import line and i do not want to complicate the review process.

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+        )
+
+    def invalidate_rows(self, row_indices: Sequence[int]) -> None:
+        """Mark reused batch rows stale so their metadata is refreshed.
+        Row ownership changes when scheduler slots are reused for different request ids.
+        """
+        for row_idx in row_indices:
+            self.refresh_start_seq_lens.np[row_idx] = -1
+
+    def prepare_refresh_if_needed(
```
MengqingCao:
```suggestion
    def maybe_prepare_refresh(
```

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+            refreshed_seq_len = int(self.refresh_start_seq_lens.np[row_idx])
+            stale_or_shrunk = refreshed_seq_len < 0 or refreshed_seq_len > seq_len
+            if stale_or_shrunk:
+                self.refresh_start_seq_lens.np[row_idx] = 0
+                refreshed_seq_len = 0
+                self.refresh_end_seq_lens.np[row_idx] = seq_len
+            else:
+                crossed_page_boundary = seq_len // QUEST_PAGE_SIZE > refreshed_seq_len // QUEST_PAGE_SIZE
+                if crossed_page_boundary:
+                    self.refresh_end_seq_lens.np[row_idx] = seq_len
```
MengqingCao, line 209: QQ: don't we need fresh the `refresh_start_seq_lens` here?

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+    ) -> None:
+        """Validate QUEST support and allocate metadata once KV caches exist.
+        The runner creates this manager after KV cache initialization because layer shapes come from caches.
+        """
+        self.max_num_reqs = max_num_reqs
+        self.device = device
+        self.pin_memory = pin_memory
+        self.metadata_block_tables: torch.Tensor | None = None
+        self.max_num_metadata_blocks_per_req = 0
+        self.owner_req_ids: list[str | None] = [None] * max_num_reqs
```
MengqingCao, line 261: plz make some comment on the semantic meaning of `owner_req_ids`

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+        self.topk_pages = 0
+
+        if not ascend_config.quest_decode_config.enable:
+            return
+
+        max_num_metadata_blocks_per_req = _get_max_num_metadata_blocks_per_req(model_config, max_encoder_len)
+        is_disabled, disable_reason = self._quest_is_disabled(
+            vllm_config=vllm_config,
+            ascend_config=ascend_config,
+            model_config=model_config,
```
MengqingCao: actually `model_config` is included in `vllm_config`, let's just keep the `vllm_config`.

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+        num_metadata_blocks = max_num_reqs * max_num_metadata_blocks_per_req
+        self.metadata_block_tables = torch.arange(
+            num_metadata_blocks,
+            dtype=torch.int32,
+            device=device,
+        ).view(max_num_reqs, max_num_metadata_blocks_per_req)
+
+        base_layer_metadata: dict[str, _QuestLayerMetadata] = {}
+        for layer_name, k_cache in base_layers.items():
+            metadata_shape = (num_metadata_blocks, QUEST_PAGE_SIZE, k_cache.shape[2], QUEST_HEAD_SIZE)
```
MengqingCao: Let's explicitly declare `num_head = k_cache[2]` to make the semantics clearer and facilitate adaptation to more kvcache layouts.

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+    def _invalidate_rows(self, row_indices: Sequence[int]) -> None:
+        """Propagate row invalidation to every unique layer metadata object.
+        Shared-KV layers point at the same object, so iterating this list avoids duplicate work.
+        """
+        if not row_indices:
+            return
+        for layer_metadata in self.layer_metadata:
+            layer_metadata.invalidate_rows(row_indices)
+
+    def _quest_is_disabled(
```
MengqingCao:
```suggestion
    def _validate_model_support_quest(
```

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+            return None
+
+        kv_cache = kv_caches[layer_name]
+        if not isinstance(kv_cache, tuple) or len(kv_cache) < 2:
+            self._disable(f"attention layer {layer_name} does not expose a standard KV cache tuple.")
+            return None
+
+        k_cache = kv_cache[0]
+        if not isinstance(k_cache, torch.Tensor) or k_cache.ndim != 4:
+            self._disable(f"attention layer {layer_name} has an unsupported key-cache layout.")
```
MengqingCao, line 414: QQ: Is the supported key cache layout `[num_blocks, block_size, num_head, head_dim]`? maybe we should make some comments for it.

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+        if k_cache.shape[1] != QUEST_PAGE_SIZE or k_cache.shape[-1] != QUEST_HEAD_SIZE:
+            self._disable(
+                f"attention layer {layer_name} has block_size={k_cache.shape[1]} and "
+                f"head_size={k_cache.shape[-1]}, but QUEST requires block_size={QUEST_PAGE_SIZE} "
+                f"and head_size={QUEST_HEAD_SIZE}."
+            )
+            return None
+        return k_cache
+
+    def _collect_quest_k_caches(
```
MengqingCao:
```suggestion
    def _collect_quest_k_caches_layer_mappings(
```

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+            or max_query_len != 1
+            or block_table_width <= 0
+        ):
+            return QuestBatchMetadata()
+
+        num_reqs = min(num_reqs, len(req_ids), self.max_num_reqs)
+        if num_reqs <= 0:
+            return QuestBatchMetadata()
+
+        selected_blocks = min(self.topk_pages, block_table_width)
```
MengqingCao:
```suggestion
        num_selected_blocks = min(self.topk_pages, block_table_width)
```

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+        seq_lens_tensor = seq_lens[:batch_size]
+        if not isinstance(seq_lens_tensor, torch.Tensor):
+            seq_lens_tensor = torch.as_tensor(seq_lens_tensor)
+
+        total_blocks = torch.div(
+            seq_lens_tensor + QUEST_PAGE_SIZE - 1,
+            QUEST_PAGE_SIZE,
+            rounding_mode="floor",
+        )
+        total_blocks = total_blocks[total_blocks > 0]
```
MengqingCao: I recommend to use `torch.were` here to avoid the synchronization between host and device.

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+        if total_blocks.numel() == 0:
+            return False
+
+        selected_blocks_per_req = torch.clamp(total_blocks, max=selected_blocks)
+        avg_selected_ratio = (selected_blocks_per_req.float() / total_blocks.float()).mean()
```
MengqingCao: QQ: could you explain more about why we use the average calue of `selected_ratio` instead of the maximum?

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+            fast_build=fast_build,
+        )
+
+        num_reqs = common_attn_metadata.num_reqs
+        if common_attn_metadata._seq_lens_cpu is not None:
+            quest_seq_lens_cpu = common_attn_metadata._seq_lens_cpu[:num_reqs]
+        elif common_attn_metadata.seq_lens_cpu is not None:
+            quest_seq_lens_cpu = common_attn_metadata.seq_lens_cpu[:num_reqs]
+        else:
+            quest_seq_lens_cpu = common_attn_metadata.seq_lens[:num_reqs].to("cpu")
```
MengqingCao, line 674: plz add a todo here as a reminder to refactor this, I think we should keep the source of `seq_lens_cpu` a single one.

vllm_ascend/attention/quest_decode.py
```diff
@@ -0,0 +1,805 @@
...
+        )
+
+        if not metadata_ready:
+            return super().forward_impl(query, key, value, kv_cache, attn_metadata, output)
+
+        quest_inputs = quest_metadata.get_sparse_decode_inputs(layer_name)
+        if quest_inputs is not None:
+            return self.forward_quest_attention(query, key, value, attn_metadata, output, quest_inputs)
+
+        return super().forward_impl(query, key, value, kv_cache, attn_metadata, output)
```
MengqingCao:
```suggestion
        if metadata_ready:
            quest_inputs = quest_metadata.get_sparse_decode_inputs(layer_name)
            if quest_inputs is not None:
                return self.forward_quest_attention(query, key, value, attn_metadata, output, quest_inputs)

        return super().forward_impl(query, key, value, kv_cache, attn_metadata, output)
```

vllm_ascend/attention/quest_decode.py
MengqingCao: Thanks for the details, it is very clearly described. 👍

vllm_ascend/attention/quest_decode.py
MengqingCao: That makes sense, thanks!

MengqingCao (commented):
I'm very sorry for my delay, and great thanks for your refactor, the code now seems much more independent and clean. And I left some suggestions, most of which are about naming and tiny refactoring. LGTM if they're all fixed, thx for your effort!

vllm_ascend/attention/quest_decode.py
MarcelFerrari: `refresh_start_seq_lens` does not need to be overwritten before launch because it is the persistent token freshness marker. `maybe_prepare_refresh()` only sets `refresh_end_seq_lens` for rows that crossed a page boundary, then `quest_prefill_metadata()` maps the raw token range to the pages that need metadata. After the kernel finishes, `commit()` advances `refresh_start_seq_lens` to the raw end length.

For QUEST, metadata is only materialized for complete pages. The active tail page is intentionally skipped by the prefill metadata kernel because the decode selector always includes it as a fixed anchor.

```
block_size = 128
old marker = 350
new seq len = 385

tokens:   0        128       256       384   385
          |  page 0 |  page 1 |  page 2 |  p3 tail
range:                             [350 ----- 385)

metadata kernel:
  start_page = 350 // 128 = 2
  end_page   = 385 // 128 = 3

  refresh pages [2, 3) => page 2 only
  skip page 3, because it is the active tail page

decode selector:
  page 0 and page 3 are fixed anchors
  their scores are masked before sorting
  output = [0, 3, best scored interior pages...]
```

So the policy is: Python keeps raw token freshness markers; the metadata kernel snaps that range to complete metadata pages; the selection kernel handles first/tail pages as anchors, independent of their scores.

As a side note, while writing this answer I realized a few edge cases that could be problematic, so i patched and then re-tested the kernels.

vllm_ascend/attention/quest_decode.py
MarcelFerrari: Added this explanation
```
# Tracks which request id currently owns each active-batch row. If the
# scheduler reuses a row for another request, the per-layer QUEST
# metadata cached for that row no longer matches the new request.
```

The vLLM scheduler can decide to insert or move around queries continuously and Quest needs to track which query owns specific metadata. `owner_req_ids` maps `batch row id -> request id` so that we can check if metadata needs to be recomputed because a new query/request was inserted in the batch.

vllm_ascend/attention/quest_decode.py
MarcelFerrari: This is correct. I added a note about this in the docstring

vllm_ascend/attention/quest_decode.py
MarcelFerrari: We use the average selected ratio because Quest sparse decode is selected at batch granularity. Using the maximum would make the decision depend on the least sparse row, so one short request would disable sparse decode for a batch of long-context requests. The average better reflects expected batch-level savings while still falling back to dense when the selected-page ratio is high for the batch overall.

vllm_ascend/attention/quest_decode.py
MarcelFerrari: Implemented `torch.where` version. According to benchmark results it seems like the performance is the same, however it is ok for me to use `torch.where` if you prefer.

MarcelFerrari:
> I'm very sorry for my delay, and great thanks for your refactor, the code now seems much more independent and clean. And I left some suggestions, most of which are about naming and tiny refactoring. LGTM if they're all fixed, thx for your effort!

Hi @MengqingCao,

Thank you for your efforts in the review process. Your help is greatly appreciated!
I have updated the code and implemented the suggestions you shared.
Let me know if everything looks good to you. Once I have your ok I will squash and rebase the PR branch.

Thanks!

vllm_ascend/attention/quest_decode.py
MengqingCao: Thanks for the details!

MengqingCao (dismissed):
Thanks for your effort! this LGTM now. But let's wait for the review from @zzzzwwjj and @ZT-AIA , they will also take a look at especially on the kernels.

MengqingCao:
btw, could you add a test case for quest sparse decode?

csrc/attention/quest_prefill_metadata/op_kernel/quest_prefill_metadata.cpp
```diff
@@ -0,0 +1,409 @@
+/**
```
chenlongxiao, line 1: 这个kernel内似乎全是scalar运算，这在NPU内性能很差，
1.有无办法转成vector计算或者cube计算
2.算子性能整网占比高不高？

csrc/attention/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
```diff
@@ -0,0 +1,604 @@
...
+        if (num_selected_pages == 1) {
+            selected_indices_lt.SetValue(0, last_anchor);
+        } else {
+            selected_indices_lt.SetValue(0, 0U);
+            selected_indices_lt.SetValue(1, last_anchor);
+        }
+        return;
+    }
+
+    for (int32_t idx = 0; idx < k; ++idx) {
```
chenlongxiao: K有多大？比较大，SetValue性能极差，是否对算子性能有要求？
61行代码应该可以从copy指令
62行可以尝试duplicate指令

csrc/attention/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
```diff
@@ -0,0 +1,604 @@
...
+__aicore__ inline void quest_apply_sequential_selection(
+    LocalTensor<uint32_t> &selected_indices_lt,
+    int32_t valid_page_count,
+    int32_t k)
+{
+    int32_t num_selected_pages = valid_page_count > 0 ? MIN(k, valid_page_count) : 0;
+    for (int32_t idx = 0; idx < num_selected_pages; ++idx) {
+        selected_indices_lt.SetValue(idx, static_cast<uint32_t>(idx));
+    }
+    for (int32_t idx = num_selected_pages; idx < k; ++idx) {
```
chenlongxiao: duplicate指令

csrc/attention/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
```diff
@@ -0,0 +1,604 @@
...
+        AscendC::DataCopy(input_storage_lt, minblocks_gm_[meta_block_offset], gm_ub_cp);
+        AscendC::SetFlag<AscendC::HardEvent::MTE2_V>(EVENT_ID1);
+        AscendC::WaitFlag<AscendC::HardEvent::MTE2_V>(EVENT_ID1);
+        AscendC::Cast<ComputeT, StorageT>(
+            tensors.minblock,
+            input_storage_lt,
+            AscendC::RoundMode::CAST_NONE,
+            block_size_ * head_dim_);
+        AscendC::Mul(
+            tensors.minblock,
```
chenlongxiao: 这几个指令都在使用minblock  不需要PipeBarrier吗

csrc/attention/paged_select_attention/op_kernel/attn_infra/arch/resource.hpp
```diff
@@ -0,0 +1,43 @@
...
+    LocalTensorBuffer<ArchTag, AscendC::TPosition::C2> btBuf;
+    LocalTensorBuffer<ArchTag, AscendC::TPosition::CO1> l0CBuf;
+    LocalTensorBuffer<ArchTag, AscendC::TPosition::VECCALC> ubBuf;
+
+    __aicore__ inline
+    Resource()
+    {
+        // The initialization of AscendC::Tpipe will insert some synchronization interfaces,
+        // which may conflict with the usage by users. Therefore, the "destroy" interface is used for releasing.
+        pipe.Destroy();
```
chenlongxiao, line 37: ascendc里面需要自己去显示pipe.Destroy()吗

csrc/attention/paged_select_attention/op_kernel/attn_infra/epilogue/block/block_epilogue_init_outputs.hpp
```diff
@@ -0,0 +1,160 @@
...
+        AscendC::WaitFlag<AscendC::HardEvent::MTE3_V>(EVENT_ID6);
+        AscendC::Duplicate(attnOutUbTensor, static_cast<ElementAttnOut>(ATTN_OUT_INI), embedRoundV * qSThisSubBlock);
+        AscendC::SetFlag<AscendC::HardEvent::V_MTE3>(EVENT_ID6);
+        AscendC::WaitFlag<AscendC::HardEvent::V_MTE3>(EVENT_ID6);
+        for (uint32_t qNIdx = 0; qNIdx < qNThisSubBlock; qNIdx++) {
+            AscendC::DataCopyPad(
+                gOutput[qNIdx * embedV],
+                attnOutUbTensor,
+                AscendC::DataCopyExtParams(
+                    qSThisSubBlock, embedV * sizeof(ElementAttnOut),
```
chenlongxiao, line 85: sizeof(ElementAttnOut)多次调用 可以在前面算一次给一个scalar，减少sizeof运算

csrc/attention/paged_select_attention/op_kernel/attn_infra/epilogue/block/block_epilogue_init_outputs.hpp
```diff
@@ -0,0 +1,160 @@
...
+            AscendC::WaitFlag<AscendC::HardEvent::MTE3_V>(EVENT_ID7);
+            AscendC::Duplicate(lseOutUbTensor, LSE_OUT_INI, qSThisSubBlock * FLOAT_ELEM_NUM_PER_BLK);
+            AscendC::SetFlag<AscendC::HardEvent::V_MTE3>(EVENT_ID7);
+            AscendC::WaitFlag<AscendC::HardEvent::V_MTE3>(EVENT_ID7);
+            for (uint32_t qNIdx = 0; qNIdx < qNThisSubBlock; qNIdx++) {
+                AscendC::DataCopyPad(
+                    gLse[qNIdx],
+                    lseOutUbTensor,
+                    AscendC::DataCopyExtParams(
+                        qSThisSubBlock, sizeof(ElementLseOut),
```
chenlongxiao, line 100: 同上

csrc/attention/paged_select_attention/op_kernel/attn_infra/epilogue/block/block_epilogue_online_softmax.hpp
```diff
@@ -0,0 +1,1249 @@
...
+
+    __aicore__ inline
+    void RowmaxSPECTILE512(const AscendC::LocalTensor<float> &srcUb, const AscendC::LocalTensor<float> &rowmaxUb,
+        const AscendC::LocalTensor<float> &tvUbTensor, uint32_t numRowsRound, uint32_t numElems,
+        uint32_t numElemsAligned)
+    {
+        AscendC::BlockReduceMax<float, false>(
+            tvUbTensor,
+            srcUb,
+            numRowsRound * numElemsAligned / FLOAT_VECTOR_SIZE,
```
chenlongxiao, line 306: numRowsRound * numElemsAligned 多次运算，可以预先算好，多次使用结果，请节省NPU的scalar运算单元
请检查这里和其他地方类似问题

csrc/attention/paged_select_attention/op_host/op_api/aclnn_paged_select_attention.cpp
```diff
@@ -0,0 +1,70 @@
+/**
+ * Copyright (c) 2025 Huawei Technologies Co., Ltd.
```
chenlongxiao, line 2: 1.https://gitcode.com/cann/ops-transformer/tree/master/experimental/下面的代码是大家共享随便提交的  cann并不对他进行质量看护，是否有完整的质量看护策略？
2.没有UT，暂时不要合入

csrc/attention/quest_prefill_metadata/op_kernel/quest_prefill_metadata.cpp
MarcelFerrari: The kernel processes all the KV pages using vector. The only scalar operations are for indexing logic and page offset calculation logic. In our tests, the runtime of this kernel is < 0.5% end-to-end. The kernel is only executed roughly once every 128 decode steps.

I cleaned the code and added tests for this kernel.

csrc/attention/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
MarcelFerrari: K is usually small, for example 8 or 16.
I significantly improved the Quest selection kernel and removed any setValue loops.

csrc/attention/paged_select_attention/op_kernel/attn_infra/arch/resource.hpp
MarcelFerrari: This part of the code was inherited when we implemented sparse attention for decode. I do not know if this is necessary or not, but this is the way it is implemented in the production npu_fused_infer_attention_score kernel.

csrc/attention/paged_select_attention/op_kernel/attn_infra/epilogue/block/block_epilogue_init_outputs.hpp
MarcelFerrari: This was also inherited from npu_fused_infer_attention_score.
I would avoid changing this as the kernel already runs very efficiently.

csrc/attention/paged_select_attention/op_kernel/attn_infra/epilogue/block/block_epilogue_online_softmax.hpp
MarcelFerrari: Again, this was also inherited from npu_fused_infer_attention_score.

csrc/attention/paged_select_attention/op_host/op_api/aclnn_paged_select_attention.cpp
MarcelFerrari: I significantly improved the quest kernels and added unit tests.

MarcelFerrari:
Thank you for your valuable feedback.
I improved the quest kernels and addressed the problems that you pointed out, as well as implementing other improvements.
I also added unit tests for the quest kernels and the paged_select_attention kernel.

csrc/attention/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
MarcelFerrari: fixed

csrc/attention/quest_block_select_paged/op_kernel/quest_block_select_paged.cpp
MarcelFerrari: fixed

chenlongxiao (approved):

MarcelFerrari:
Squashed all commits and rebased. Tested and confirmed accuracy and performance.

PR #8448, opened 2026-04-20
[Misc]update vllm to v0.19.1

Meihan-chen:
1. update transformers to v5.5.3
1.1 The lm-eval package needs to be upgraded to v0.4.11; otherwise, there will be interface incompatibility
1.2 Transformers 5 drops add_bos_token/add_eos_token when a tokenizer_file is present, while TokenizersBackend defaults add_bos_token=False, so DeepSeek string prompts no longer get BOS injected automatically and TP/EP or golden outputs diverge. See [tokenization_utils_base.py#L1783-L1785](https://github.com/huggingface/transformers/blob/ded2b747bde5e9933c140c29ca3615d759f5744d/src/transformers/tokenization_utils_base.py#L1783-L1785) and [tokenization_utils_tokenizers.py#L417-L419](https://github.com/huggingface/transformers/blob/ded2b747bde5e9933c140c29ca3615d759f5744d/src/transformers/tokenization_utils_tokenizers.py#L417-L419).This PR updates the corresponding golden values for the affected DeepSeek test cases.
2. fix WAITING_FOR_FSM error by https://github.com/vllm-project/vllm/pull/38048

- vLLM version: v0.19.0


tests/e2e/model_utils.py
```diff
@@ -19,11 +19,31 @@
...
 
 
+def build_bos_tokenized_prompts(
+    model_name: str,
+    prompts: Sequence[str],
+) -> list[list[int]]:
+    tokenizer = AutoTokenizer.from_pretrained(model_name, trust_remote_code=True)
+    bos_token_id = tokenizer.bos_token_id
+    if bos_token_id is None:
+        raise ValueError(f"{model_name} does not define a BOS token")
```
wangxiyuan: just return prompts instead for raising error?

tests/e2e/model_utils.py
```diff
@@ -19,11 +19,31 @@
...
 from collections.abc import Sequence
 
+from transformers import AutoTokenizer
 from vllm.logprobs import PromptLogprobs, SampleLogprobs
 
 TokensText = tuple[list[int], str]
 
 
+def build_bos_tokenized_prompts(
+    model_name: str,
```
wangxiyuan: I think this is a bug for vllm. vllm should handle this case for ds v3/v3.1 in vllm/tokenizers

tests/e2e/multicard/4-cards/test_pipeline_parallel.py
```diff
@@ -28,7 +28,7 @@
 ]
 
 DATA_PARALLELS = [2]
-TENSOR_PARALLELS = [1, 2]
+TENSOR_PARALLELS = [1]
```
wangxiyuan, line 31: after this is changed, the test should be moved to 2-cards

tests/e2e/multicard/2-cards/test_expert_parallel.py
```diff
@@ -1,31 +0,0 @@
-import pytest
```
wangxiyuan, line 1: we should at least add one test to test ep works

wangxiyuan (approved):

tests/e2e/multicard/2-cards/test_expert_parallel.py
Meihan-chen: ok

tests/e2e/multicard/4-cards/test_pipeline_parallel.py
Meihan-chen: Currently, DP and PP test cases use the same golden value only when TP is 1. If TP uses 2 cards, a separate golden value needs to be added.

PR #8461, opened 2026-04-20
[Feature] adapter async spec decode to hybrid model and optimize performance

HF-001:
Delete configuration code that cannot enable speculative inference and asynchronous scheduling simultaneously
and  
refer to:  https://github.com/vllm-project/vllm/pull/38556  , Fix async spec decoding with hybrid models
and 
by asynchronously and non blocking synchronizing the correct seq_lens on the NPU to the CPU, ensure that the value of the attention layer is correct，the performance impact is almost negligible.

- vLLM version: v0.19.0


HF-001:
@wangxiyuan hi ,this pr is ready, Can you help take a look？

HF-001:
@wangxiyuan hi, this pr is ready, Can you help merge it?

wangxiyuan (approved):

HangjiHe:
Hi, from my understanding, this enabled the async mode config, but since a new sync point is introduced, it's still not technically fully async? When preparing next step's attention context, cpu still needs to wait for previous step's proposal from gpu?

HF-001:
> Hi, from my understanding, this enabled the async mode config, but since a new sync point is introduced, it's still not technically fully async? When preparing next step's attention context, cpu still needs to wait for previous step's proposal from gpu?

@HangjiHe  On Ascend, page attention requires seq_lens to be a CPU tensor, which is already the optimal PR for copying data asynchronously and non blocking, with minimal performance impact

PR #8462, opened 2026-04-20
[Draft][Feature][KV Pool]Support hybrid align prefix cache in Ascend Store

Pz1116:
Support hybrid align prefix cache in Ascend Store, did not test yet.

- vLLM version: v0.19.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
@@ -24,6 +24,8 @@ class KeyMetadata:
     dcp_rank: int
     """ Initialize the current pipeline parallel rank """
     pp_rank: int
+    """ Initialize the current kv cache group id """
+    kv_cache_group_id: int = 0
 
 
 @dataclass(order=True)
@@ -39,6 +41,7 @@ class PoolKey:
                 self.key_metadata.pcp_rank,
                 self.key_metadata.dcp_rank,
                 self.key_metadata.pp_rank,
+                self.key_metadata.kv_cache_group_id,
                 self.chunk_hash,
             )
         )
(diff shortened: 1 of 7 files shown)


vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
```diff
@@ -203,6 +259,20 @@ def decode_adaptor_prefill_pp(self, key, addr, size):
         return new_key, new_addr, new_size
 
 
+def normalize_block_ids_by_group(block_ids: tuple[list[int], ...] | list[int] | list[list[int]]) -> list[list[int]]:
+    if isinstance(block_ids, tuple):
+        return [group.copy() for group in block_ids]
+    if isinstance(block_ids, list):
+        if not block_ids:
+            return [[]]
```
Sparkheart, line 267: Check whether block_ids is empty.
If the input is an empty tuple (), the first if branch is executed and [] is returned. If the input is an empty list [], the second branch is executed and [[]] is returned. The returned results in the two cases have different semantics, which may cause difficulties in processing by the caller.

Pz1116: Thanks, the observed difference is real, but here empty tuple denotes zero KV cache groups, while empty list denotes one group with no blocks; for the HMA/hybrid path we do not expect zero groups, so I’d prefer to preserve the distinction. Thanks for calling this out. The difference is intentional here because the two empty inputs carry different upstream semantics.

- `[]` in this helper corresponds to the legacy/single-group shape (`list[int]`), so normalizing it to `[[]]` preserves the meaning of “one KV cache group with no blocks”.
- `()` corresponds to the grouped shape (`tuple[list[int], ...]`), and an empty tuple means “zero KV cache groups”, so normalizing it to `[]` preserves that meaning.

In other words, these are not two encodings of the same state. Collapsing them to the same normalized value would lose the distinction between:
1. one empty group, and
2. no groups at all.

So I’d prefer to keep the current behavior and make the semantics explicit with a comment/test if needed, rather than unifying the return values.

PR #8468, opened 2026-04-20
[Core][BugFix] Fix stuck chunked pipeline parallelism with async scheduling

rouye-he:

When chunked prefill is used with async scheduling + pipeline parallelism,
all requests in a non-final prefill chunk have their sampled tokens marked
as dummy (to be discarded). Broadcasting these dummy tokens across PP ranks
is unnecessary and causes the system to get stuck.

This PR adds a check via `_is_all_reqs_chunked_prefill()` to skip the
broadcast/receive of sampled token IDs when all scheduled requests are
in a non-final chunked prefill step.

Two locations fixed in `model_runner_v1.py`:
1. `_pp_broadcast_prev_sampled_token_ids` call (last PP rank sender)
2. `_pp_receive_prev_sampled_token_ids_to_input_batch` call (non-last PP rank receiver)

Mirrors vllm PR #38726.

Yes. Chunked prefill + async scheduling + pipeline parallelism no longer
gets stuck.

Verified fix logic mirrors the upstream vllm fix in PR #38726.
- vLLM version: v0.19.0


rouye-he:
@MengqingCao @wangxiyuan Could you please take a look when you have time? Thanks!

rouye-he:
Looks like this was addressed by #11076, which replaced the GPU broadcast path for PP sampled tokens with scheduler IPC entirely — the code paths this PR touched no longer exist. Closing as resolved via that refactor. Thanks for the earlier review!

PR #8469, opened 2026-04-20
[v0.18.0][BugFix] PIECEWISE mode also requires synchronization

keyi-zz:

This PR enables synchronization for the `PIECEWISE` runtime mode in ACL graph replay. Previously, synchronization was only performed in `FULL` mode. However, `PIECEWISE` mode also requires this barrier to ensure that parameter updates are completed before the graph is replayed, preventing accuracy loss.

The logic is also corrected to skip synchronization specifically for EAGLE draft models, as intended.

Fixes #

N/A

CI passed.

### DIFF
--- a/vllm_ascend/compilation/acl_graph.py
+++ b/vllm_ascend/compilation/acl_graph.py
@@ -202,9 +202,8 @@ class ACLGraphWrapper:
         # If we do not in main model and in full-graph mode when using merge-eagle-graph,
         # we do not need to synchronize.
         # When enable_enpu is on, model_runner orders update vs replay; skip here.
-        # When FULL + EAGLE draft (merge path), replay does not need this barrier.
-        is_draft_eagle = _EXTRA_CTX.is_draft_model and self.use_eagle
-        need_sync = self.runtime_mode == CUDAGraphMode.FULL and not is_draft_eagle
+        # When EAGLE draft (merge path), replay does not need this barrier.
+        need_sync = _EXTRA_CTX.is_draft_model and self.use_eagle
         if not self.enable_enpu and need_sync:
             torch.npu.current_stream().synchronize()
         entry.aclgraph.replay()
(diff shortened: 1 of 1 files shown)


jiangshaoping:
approve

yiz-liu (commented):
Theoretically, PIECEWISE does not require any synchronization. I suggest we leave an explicit FIXME here, otherwise this can be easily forgot.

vllm_ascend/compilation/acl_graph.py
```diff
@@ -202,9 +202,8 @@ def __call__(self, *args, **kwargs):
         # If we do not in main model and in full-graph mode when using merge-eagle-graph,
         # we do not need to synchronize.
         # When enable_enpu is on, model_runner orders update vs replay; skip here.
-        # When FULL + EAGLE draft (merge path), replay does not need this barrier.
-        is_draft_eagle = _EXTRA_CTX.is_draft_model and self.use_eagle
-        need_sync = self.runtime_mode == CUDAGraphMode.FULL and not is_draft_eagle
+        # When EAGLE draft (merge path), replay does not need this barrier.
+        need_sync = _EXTRA_CTX.is_draft_model and self.use_eagle
```
yiz-liu: Before: `not (full and draft and eagle)`, which means the only non-sync condition is the gap between target model and draft model.
After: `full and not (draft and eagle)`, this should be the final condition but we are encountering accuracy regression with PIECEWISE now
Current: `draft and eagle`, wrong

Please change to `not (draft and eagle)`, and we should recover the condition for FULL and address this accuracy issue in main branch.

yiz-liu (approved):

yiz-liu:
As @jiangshaoping required, we will merge this first, and fix for main branch later. Also, CI is not finished by now, we will have to revert this should anything fail.

PR #8479, opened 2026-04-21
[Test][CI] main2main 0918

zhangxinyuehfad:

- Add `multi_instance` service mode to the nightly single-node test framework
- Two `vllm serve` instances start on the same NPU card simultaneously (second starts while first holds memory), both must initialize and serve correctly


zhangxinyuehfad:
reopen

PR #8508, opened 2026-04-21
[BugFix] Fix sliding window attention computation and improve throughput

vvaen:

The original sliding window attention implementation has two critical issues:

1. **Output corruption in full-graph replay path**: Rebinding `block_tables` to the latest metadata tensor causes corrupted/garbled outputs on Ascend NPU for SWA models.

2. **Repetition issues**: Using `sparse_mode=3` and the `_forward_fia_slidingwindow` method both lead to repeated outputs, affecting model quality.

This PR fixes the above issues with the following changes:

- Fix block_tables handling in full-graph replay path: keep captured block_tables tensor for SWA models to avoid output corruption
- Fix SWA attention parameters: use `sparse_mode=4` with `pre_tokens=sliding_window` and `next_tokens=0`
- Add sliding window support in graph capture mode (full_graph_fia): pass `sparse_mode`, `pre_tokens`, `next_tokens` to graph params
- Remove deprecated `_forward_fia_slidingwindow` method, consolidate SWA handling into unified TND layout path
- Add missing `actual_seq_lengths_q` in test fixture

No. This PR only fixes internal implementation issues without changing any public API.

- vLLM version: v0.19.0

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -619,14 +619,16 @@ class AscendAttentionBackendImpl(AttentionImpl):
         input_layout = "TND"
         attn_mask = attn_metadata.attn_mask
 
+        # next_tokens only takes effect when sparse_mode=4 (sliding window attention);
+        # for other sparse modes it is ignored, so we set it to 0 by default.
+        next_tokens = 0
         if self.sliding_window is not None:
             sparse_mode = 4
             pre_tokens = self.sliding_window
-            next_tokens = 0
+
         else:
             sparse_mode = 3 if attn_metadata.causal else 0
             pre_tokens = SWA_INT_MAX
-            next_tokens = 0
 
         extra_args = {}
         if self.enable_c8_quant:
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -620,7 +620,7 @@ class AscendAttentionBackendImpl(AttentionImpl):
         attn_mask = attn_metadata.attn_mask
         sparse_mode = 4 if self.sliding_window else 3 if attn_metadata.causal else 0
         pre_tokens = self.sliding_window or SWA_INT_MAX
-        next_tokens= 0 if self.sliding_window else SWA_INT_MAX
+        next_tokens = 0 if self.sliding_window else SWA_INT_MAX
 
         extra_args = {}
         if self.enable_c8_quant:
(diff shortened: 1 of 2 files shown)


vllm_ascend/attention/attention_v1.py
```diff
@@ -513,7 +518,19 @@ def update_graph_params(
...
+                        # rebinding `block_tables` to the latest metadata tensor causes corrupted /
+                        # repeated outputs in our repro on Ascend NPU.
+                        #
+                        # Keep the captured block_tables tensor on this affected path.
+                        # Non-SWA models preserve the original behavior and continue to refresh
+                        # block_tables from attn_metadata.
+                        if hasattr(vllm_config.model_config.hf_text_config, "sliding_window"):
+                            block_tables = block_tables
+                        else:
+                            block_tables = attn_metadata[key].block_tables
```
Sparkheart, line 541: The code can be optimized through predicate statements, resulting in more concise and clear code.

vvaen: Thank you for the suggestion! I've simplified the conditional logic by removing the redundant self-assignment.

```python
if not hasattr(vllm_config.model_config.hf_text_config, "sliding_window"):
    block_tables = attn_metadata[key].block_tables
```

vllm_ascend/attention/attention_v1.py
```diff
@@ -605,7 +618,16 @@ def full_graph_fia(
         softmax_lse = torch.empty(1, dtype=query.dtype, device=query.device)
         input_layout = "TND"
         attn_mask = attn_metadata.attn_mask
-        sparse_mode = 3 if attn_metadata.causal else 0
+
+        if self.sliding_window is not None:
+            sparse_mode = 4
+            pre_tokens = self.sliding_window
+            next_tokens = 0
```
wangxiyuan: no need to init next_tokens

vllm_ascend/attention/attention_v1.py
```diff
@@ -605,7 +618,18 @@ def full_graph_fia(
         softmax_lse = torch.empty(1, dtype=query.dtype, device=query.device)
         input_layout = "TND"
         attn_mask = attn_metadata.attn_mask
-        sparse_mode = 3 if attn_metadata.causal else 0
+
+        # next_tokens only takes effect when sparse_mode=4 (sliding window attention);
+        # for other sparse modes it is ignored, so we set it to 0 by default.
+        next_tokens = 0
```
wangxiyuan: I think we can remove `next_tokens` totally, only set it to 0 at `npu_fused_infer_attention_score`?

vvaen: Thank you for the suggestion! We now conditionally set next_tokens to 0 only when configuring sliding window attention. For all other sparse attention modes, next_tokens retains its default value (SWA_INT_MAX) to ensure compatibility. Additionally, we have streamlined the assignment of sparse_mode, pre_tokens, and next_tokens to keep the configuration clear and consistent.

MengqingCao (approved):
LGTM now

PR #8557, opened 2026-04-22
[CI][1/N] Add hardware-awared test gating and CI routing

Potabk:
## What this PR do?

This PR introduces an npu_test decorator for hardware-based test gating and CI routing, alongside new unit tests for `NPUModelRunner`. Feedback highlights the need for more robust `torch_npu` mocking to prevent errors in non-NPU environments and suggests that the `@npu_test` decorator should verify the specific NPU chip type at runtime. 

We've long been plagued by excessively long end-to-end testing durations, which is extremely detrimental to community health and developer well-being. We aim to implement selective testing based on developer pull requests (PRs):

This will be addressed from two main perspectives:

1. Intelligently selecting test cases to trigger based on developer modifications. This is easily achievable; we simply need to ensure a one-to-one correspondence between the test directory (tests/ut) and the modules in the src directory. Furthermore, for some common, fundamental modules, we will allow them to be unconditionally tested in every PR (this is meaningful because certain modules...).

2. Stateful test cases that developers are aware of. Developers only need to add a `@npu_test` decorator to specify the required NPU device type and number of chips. The system will automatically route this test case to an appropriate node for testing.

## What's next?

1. Fill in more non-e2e cases that can run on the NPU to have some real verification.
2. Once the cases above are complete enough, we try to reduce the existing e2e cases until they can reach a healthy duration.
**This is a massive undertaking, and anyone who is interested can get involved.**


tests/ut/conftest.py
```diff
@@ -15,25 +15,90 @@
...
+    sys.modules["torch_npu._inductor"] = MagicMock()
 
 adapt_patch()
 adapt_patch(True)
 
 # register Ascend CustomOp here because uts will use this
 register_ascend_customop()
+
+
+class RunnerDeviceType(str, Enum):
```
leo-pony, line 54: RunnerDeviceType 在 2 处都定义是否有必要？是否可以单独起一个文件定义，这两处进行 import？

.github/workflows/scripts/determine_smart_e2e_scope.py
```diff
@@ -0,0 +1,567 @@
...
+    try:
+        tree = ast.parse(source)
+    except SyntaxError:
+        return {}
+
+    groups: dict[RunnerKey, list[str]] = defaultdict(list)
+    for node in ast.iter_child_nodes(tree):
+        if (isinstance(node, ast.FunctionDef) and node.name.startswith("test_")) or isinstance(node, ast.ClassDef):
+            key = _extract_runner_key(node)
+            groups[key].append(f"{filepath}::{node.name}")
```
leo-pony, line 255: For test class, should been format: pytest test_demo.py::TestMyClass::test_specific_func, this class could not recognize class name.

Potabk: In this system, the smallest output granularity is at the class level(test_demo.py::TestMyClass) if it is the node of `ClassDef`

.github/workflows/pr_test_light.yaml
```diff
@@ -43,18 +43,19 @@ jobs:
     with:
       vllm: 6f786f2c506cb07f4566771fdc62e640e2c4a176
   changes:
-    runs-on: linux-aarch64-a2b3-0
+    runs-on: ubuntu-latest
```
leo-pony: Change to vllm group GitHub runner, whether has large runner usage eclipse to vllm group?

Potabk: I think it's ok. This github runner `ubuntu-latest` is triggered every time a PR pushes a new commit, and the trigger duration is 9 seconds, which I think is acceptable.

Potabk: see https://github.com/vllm-project/vllm-ascend/actions/runs/24975070924/job/73125320037?pr=8557

tests/ut/conftest.py
```diff
@@ -15,25 +15,90 @@
...
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            if npu_type == RunnerDeviceType.CPU:
+                return func(*args, **kwargs)
+            if not _npu_available:
+                pytest.skip(f"NPU not available (need {npu_type.value} x{num_npus})")
+            import torch  # noqa
+
+            device_count = torch.npu.device_count()
```
leo-pony, line 102: After the system stabilizes, could we remove the device count check for each test case to speed up test case execution?  if device_count < num_npus that is in error state and this should been fixed?

Potabk: For now, I think we should give some more explicit and costly error messages if a case is routed to the wrong runtime environment.

PR #8645, opened 2026-04-24
[Ops][BugFix] Fix the issue of overlapping CPU binding ranges in specific scenarios

Rozwel-dx:

Adjust the CPU allocation logic for each NPU card to prevent overlapping CPU binding ranges when multiple processes share the same CPU affinity. By iterating over all logical NPUs instead of just the running ones, the code ensures a deterministic and non-overlapping global partitioning of CPU cores.

No

Tested by CI.


vllm_ascend/cpu_binding.py
```diff
@@ -328,7 +328,7 @@ def build_cpu_pools(self) -> None:
             self.build_global_slice_cpu_pool()
             return
 
-        for npu in self.device_info.running_npu_list:
+        for npu in self.device_info.all_logic_npus:
```
chenchuw886: When docker sets `--cpuset-cpus`, this change indeed causes RuntimeError in line 336.

chenchuw886:
```
        allowed_cpu_set = set(self.device_info.allowed_cpus)
        # Consider all logical NPUs to avoid CPU distribution overlap across worker processes.
        candidate_npus = [
            npu
            for npu in self.device_info.all_logic_npus
            # Always keep visible NPUs for strict conflict checks; include
            # non-visible NPUs only when they overlap this process's cpuset.
            if npu in self.device_info.running_npu_list
            or any(cpu in allowed_cpu_set for cpu in self.device_info.npu_affinity.get(npu, []))
        ]
        for npu in candidate_npus:
```

Consider this modification

vllm_ascend/cpu_binding.py
chenchuw886:
```
        # Keep only visible NPUs in the final binding pool. Non-visible NPUs are used only to avoid overlap.
        self.npu_cpu_pool = {npu: final[npu] for npu in self.device_info.running_npu_list}
```

and here, keep only visible NPUs in final pool

PR #8651, opened 2026-04-24
[BugFix] Routing replay support multi dp & tp and hybrid attention

cocacolafan:
This PR fixes a bug in the MoE routing replay feature for Ascend NPUs by introducing `AscendRoutedExpertsCapturer` and ensuring correct padding and buffer handling during expert routing capture in multi-DP/TP scenarios.This PR  also fixes buffer insufficient bug when use hybrid attention

vLLM version: v0.20.1
vLLM main: https://github.com/vllm-project/vllm/commit/c7aa186d67b6f051680831418e957c67f34ba7a2


weijinqian0 (dismissed):

vllm_ascend/ops/fused_moe/prepare_finalize.py
```diff
@@ -153,7 +153,8 @@ def prepare(
         padded_hidden_states_shape = hidden_states.shape
         if not (self.replace_allreduce or self.enable_shared_expert_dp):
             self.num_tokens, _ = hidden_states.shape
-            pad_size = self.tp_size - self.num_tokens  # Pad to TP size (cyclic)
+            target_pad_length = _EXTRA_CTX.padded_num_tokens
```
wangxiyuan: @zzzzwwjj  please double check this change

vllm_ascend/ops/fused_moe/prepare_finalize.py
```diff
@@ -153,7 +153,8 @@ def prepare(
         padded_hidden_states_shape = hidden_states.shape
         if not (self.replace_allreduce or self.enable_shared_expert_dp):
             self.num_tokens, _ = hidden_states.shape
-            pad_size = self.tp_size - self.num_tokens  # Pad to TP size (cyclic)
+            target_pad_length = _EXTRA_CTX.padded_num_tokens
+            pad_size = target_pad_length - self.num_tokens
```
realliujiaxu: Why do this change?
The original code pads `hidden_states.shape[0]` >= tp_size to make sure every rank has at least one token. The modified code here makes `hidden_states.shape[0]` multiple of tp_size.

cocacolafan: Implemented via an alternative approach, already rolled back this.

Li-Yongwen:
Hello, Im liyongwen of hw. I have noticed that you’re no longer with the company. Will you continue push for this pr to be merged recently? If not, can I resubmit this PR and adapt it to the main branch.

PR #8652, opened 2026-04-24
update v4 docs

GDzhu01:
Update v4 docs


Yikun (approved):
Better to do e2e test after https://github.com/vllm-project/vllm-ascend/actions/runs/24870728008/job/72816507065 image build ready.

PR #8656, opened 2026-04-24
[Doc] Update version policy

wangxiyuan:

Refresh version policy doc, add 0.13.0rc3 release mapping info

no
by ci

- vLLM version: v0.19.0


0001-Modify-the-docs-and-config-of-v4.patch
```diff
@@ -0,0 +1,851 @@
+From 3676573100a781009ee6be3816f420148d141921 Mon Sep 17 00:00:00 2001
```
Yikun: what's this

wangxiyuan: deleted

PR #8662, opened 2026-04-24
[Performance] Irregular mask build logic for DCP when num_spec_tokens > 1

SunnyLee151064:
This PR proposes a mask construction scheme for the DCP + speculative decoding scenario. Previously, one decode step for a single request in this scenario would be split into one decode step for (1 + num_spec_decodes) separate requests. Since the masks on each rank are not necessarily identical in this scenario, this approach omitted the mask construction process but resulted in more frequent KV cache transfers. This PR is specifically introduced to resolve this issue.

### Summary of changes
First of all，a func to build irregular mask is constructed in pcp_utils.py，it's result will be stored in long_seq_metadata as dcp_mtp_attn_mask:
`    def generate_mtp_attention_mask_for_decode(
        self,
        decode_num_computed_tokens: list[int],
        decode_num_scheduled_tokens: np.ndarray,
    ) -> dcp_mtp_attn_mask:`

Then, gqa and mla backend will fetch the mask and pass it to FIA operator. Notice that when the mask is irregular, we only can choose BSND layerout in FIA, so we have to do some changes in _forward_decode func both in graph and eager mode:

Graph update:
`if speculative_config is not None:
                    input_layout = "BSND"
                torch_npu.npu_fused_infer_attention_score.out(
                    q_nope,
                    k_nope,
                    value,
                    num_heads=num_heads,
                    num_key_value_heads=num_kv_heads,
                    input_layout=input_layout,
                    atten_mask=attn_mask,
                    scale=scale,
                    antiquant_mode=0,
                    antiquant_scale=None,
                    softmax_lse_flag=True,
                    block_table=block_table,
                    block_size=block_size,
                    actual_seq_lengths_kv=actual_seq_lengths_kv,
                    actual_seq_lengths=actual_seq_lengths_q,
                    workspace=graph_params.workspaces.get(num_tokens),
                    out=[attn_output, softmax_lse],
                )`

Eager mode:

`        if self.vllm_config.speculative_config is not None:
            input_layerout = "BSND"
            num_decodes = attn_metadata.num_decodes
            if attn_metadata.decode_meta.dcp_mtp_attn_mask is not None:
                attn_mask = attn_metadata.decode_meta.dcp_mtp_attn_mask
            else:
                attn_mask = None
            query = query.view(num_decodes, -1, query.shape[1], query.shape[-1])
            k_nope = self.key_cache.view(self.key_cache.shape[0], 1, self.key_cache.shape[1], -1)
            value = self.value_cache.view(self.value_cache.shape[0], 1, self.value_cache.shape[1], -1)`
No

- vLLM version: v0.20.2


vllm_ascend/attention/context_parallel/attention_cp.py
```diff
@@ -383,14 +391,17 @@ def update_graph_params(
 
                 torch.npu.graph_task_update_begin(update_stream, handle)
 
+                input_layout = "TND"
+                if speculative_config is not None:
+                    input_layout = "BSND"
```
MengqingCao, line 396: plz add note for why must use BSND in this scenario.

vllm_ascend/attention/utils.py
```diff
@@ -382,3 +384,17 @@ def enabling_mlapo(vllm_config: VllmConfig) -> bool:
         and not vllm_config.kv_transfer_config.is_kv_producer
     )
     return bool(envs.VLLM_ASCEND_ENABLE_MLAPO and is_decode_instance)
+
+
+def generate_dcp_mtp_mask(masks, query_lens, num_decodes) -> torch.Tensor:
+    if masks and masks[0] is not None:
+        query_len = masks[0].shape[0]
+        mtp_attn_mask = torch.ones(num_decodes, query_len, 16384, dtype=torch.bool)
```
weijinqian0, line 392: need check if the performance can be worse.

vllm_ascend/attention/context_parallel/attention_cp.py
SunnyLee151064: When irregular mask (spec decode scenario), FIA only support BSND layerout now.

PR #8718, opened 2026-04-25
[BugFix] MTP recurrent batch size after lmhead TP logits truncation

ichaoren:

-  With mtp and lmhead_tensor_parallel_size enabled together, real requests can fail with an error like:
```
RuntimeError: The expanded size of the tensor (1024) must match the 
existing size (32) at non-singleton dimension 0.
Target sizes: [1024]. Tensor sizes: [32]
```
- The failure happens in the recurrent draft loop at:
file

```
self.input_ids[:batch_size] = input_ids
```
- In this path:
  - the incoming batch_size still reflects the graph/padded drafter batch
  - but after compute_logits() under lmhead_tp , draft_token_ids reflects only the real valid sample count
- As a result, the recurrent writeback path can try to write a smaller tensor into a larger slice.

- In _run_merged_draft() , the first sampling step computes:
```
draft_token_ids = logits.argmax(dim=-1)
```
- Under lmhead_tp , logits can be padded for parallel computation and then truncated back to the real valid sample count.
- However, the recurrent draft loop still reuses the original batch_size from the function input, which may represent the padded graph batch rather than the effective sampled token count.
- This creates a semantic mismatch:
  - batch_size : padded/graph batch size
  - draft_token_ids.shape[0] : real effective recurrent batch size

No

### DIFF
--- a/vllm_ascend/spec_decode/eagle_proposer.py
+++ b/vllm_ascend/spec_decode/eagle_proposer.py
@@ -940,6 +940,7 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
             return torch.stack(draft_token_ids_list, dim=1)
 
         # Generate the remaining draft tokens.
+        batch_size = draft_token_ids.shape[0]
         draft_token_ids_tensor = torch.zeros(
             (self.num_speculative_tokens, *draft_token_ids.shape), dtype=draft_token_ids.dtype, device=self.device
         )
--- a/vllm_ascend/spec_decode/eagle_proposer.py
+++ b/vllm_ascend/spec_decode/eagle_proposer.py
@@ -944,7 +944,7 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
         # However, when lmhead_tp_enable() is disabled, the batch size uses the length after padding.
         # To decouple the scenarios, a judgment is required.
         # That is, the batch size needs to be modified only when lmhead_tp_enable() is enabled.
-        if lmhead_tp_enable():
+        if lmhead_tp_enable() and self.method == "mtp":
             batch_size = draft_token_ids.shape[0]
 
         # Generate the remaining draft tokens.
--- a/vllm_ascend/spec_decode/eagle_proposer.py
+++ b/vllm_ascend/spec_decode/eagle_proposer.py
@@ -499,13 +499,6 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
             if forward_context is not None:
                 forward_context.moe_layer_index = 0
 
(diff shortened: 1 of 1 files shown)


vllm_ascend/spec_decode/eagle_proposer.py
```diff
@@ -940,6 +940,7 @@ def _run_merged_draft(
             return torch.stack(draft_token_ids_list, dim=1)
 
         # Generate the remaining draft tokens.
+        batch_size = draft_token_ids.shape[0]
```
lilinsiman: 1. The logits are split and then merged only when lmhead_tp_enable() is enabled. As a result, the batch size length becomes the actual length 32. However, when lmhead_tp_enable() is disabled, the batch size uses the length after padding. To decouple the scenarios, a judgment is required. That is, the batch size needs to be modified only when lmhead_tp_enable() is enabled. 

2. In addition, detailed comments are added to explain why the batch size needs to be updated.

ichaoren: yes,  fixed

vllm_ascend/spec_decode/eagle_proposer.py
```diff
@@ -939,6 +939,14 @@ def _run_merged_draft(
                 draft_token_ids_list.append(draft_token_ids)
             return torch.stack(draft_token_ids_list, dim=1)
 
+        # The logits are split and then merged only when lmhead_tp_enable() is enabled.
+        # As a result, the batch size length becomes the actual length 32.
+        # However, when lmhead_tp_enable() is disabled, the batch size uses the length after padding.
+        # To decouple the scenarios, a judgment is required.
+        # That is, the batch size needs to be modified only when lmhead_tp_enable() is enabled.
+        if lmhead_tp_enable():
```
lilinsiman: MTP or eagle3? They are different method. Please check the case in this bug and for sure it' s needed in both eagle3 and MTP method

ichaoren: I encountered this problem in a GLM 5.1 model MTP scene.

lilinsiman: > I encountered this problem in a GLM 5.1 model MTP scene.
If only MTP scene, please add method == "mtp" in "if lmhead_tp_enable()"

ichaoren: OK，now is lmhead_tp_enable() and self.method == "mtp"

lilinsiman: > OK，now is lmhead_tp_enable() and self.method == "mtp"

OK， just wait for CI pass

lilinsiman (approved):

PR #8737, opened 2026-04-27
[BugFix] Fixed a bug when using FULL_DECODE_ONLY in model_runner_v2

drslark:

There is a pad bug in `FULL_DECODE_ONLY` mode for model_runner_v2.
So, We fixed the pad logic.
- Pad one more req for decodes graph, and the last element must be `num_tokens_padded`.
- Change the inputs of draft from `input_batch` (the information of contiguous batching) to `input_buffers` (the real drafter's input).

N/A

For inputs:

```python
prompts = [
    "Hello, my name is",
    "The president of the United States is",
    "The capital of France is",
    "The future of AI is",
]
```

with:

```python
max_num_seqs=2,
compilation_config={
    "cudagraph_mode": "FULL_DECODE_ONLY",
},
```

result is:

```text
--------------------------------------------------
total_num_output_tokens: 800
num_drafts: 461
num_draft_tokens: 1383
num_accepted_tokens: 340
mean acceptance length: 1.74
--------------------------------------------------
acceptance at token 0: 0.40
acceptance at token 1: 0.21
acceptance at token 2: 0.13
acceptance at token 3: 0.00
acceptance at token 4: 0.00
acceptance at token 5: 0.00
```

which is same with `eager` mode's result:

```text
--------------------------------------------------
total_num_output_tokens: 800
num_drafts: 461
num_draft_tokens: 1383
num_accepted_tokens: 340
mean acceptance length: 1.74
--------------------------------------------------
acceptance at token 0: 0.40
acceptance at token 1: 0.21
acceptance at token 2: 0.12
acceptance at token 3: 0.00
acceptance at token 4: 0.00
acceptance at token 5: 0.00
```

- vLLM version: v0.19.0

### DIFF
--- a/tests/e2e/singlecard/model_runner_v2/test_basic.py
+++ b/tests/e2e/singlecard/model_runner_v2/test_basic.py
@@ -70,9 +70,7 @@ def test_qwen3_dense_eager_mode(
 @pytest.mark.parametrize("eagle_model", EGALE_MODELS)
 @pytest.mark.parametrize("max_tokens", [32])
 @pytest.mark.parametrize("enforce_eager", [False])
-@pytest.mark.parametrize(
-    "compilation_config", [{"cudagraph_mode": "FULL_DECODE_ONLY", "cudagraph_capture_sizes": [4, 8]}, {}]
-)
+@pytest.mark.parametrize("compilation_config", [{"cudagraph_mode": "FULL_DECODE_ONLY"}, {}])
 @patch.dict(os.environ, {"VLLM_USE_V2_MODEL_RUNNER": "1"})
 def test_egale_spec_decoding(
     model: str,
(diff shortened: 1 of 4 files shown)


yiz-liu (commented):
Notice that draft model (step > 0) and target model have different `batch_desc` for graph keys, even if they have the same `batch_desc.num_tokens`, the `batch_desc.num_reqs` may still differ due to the change of `decode_query_len`.

drslark:
> Notice that draft model (step > 0) and target model have different `batch_desc` for graph keys, even if they have the same `batch_desc.num_tokens`, the `batch_desc.num_reqs` may still differ due to the change of `decode_query_len`.

You are right! We noticed that all failed `graph_capture_size`s are greater than `max_num_seqs`, which will be captured, but not replayed.
It is a bug in `vllm`, we will retry after next `main2main`.

PR #8756, opened 2026-04-27
[Refactor][Misc] Use lazy formatting for log

MrZ20:
This PR replaces eager log message formatting with lazy logging formatting across the repository.

- Converts `logger.*(f"...")` and `logging.*(f"...")` calls to lazy `%`-style logging arguments.
- Replaces deprecated `logger.warn(...)` usage with `logger.warning(...)`.
- Adds `logger.isEnabledFor(logging.DEBUG)` guards for debug logs whose arguments include function or method calls.
- Enables Ruff `G004` enforcement by removing it from `ignore` and registering `vllm.logger.logger` as a logger object.

This avoids unnecessary string formatting and expensive argument evaluation when the corresponding log level is disabled.

- vLLM version: v0.19.0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,12 +81,11 @@ ignore = [
     "B007",
     # f-string format
     "UP032",
-    # TODO: FIE ME
-    "G004",
     "B904",
     "SIM108",
     "SIM102"
 ]
+logger-objects = ["vllm.logger.logger"]
 
 [tool.ruff.format]
 docstring-code-format = true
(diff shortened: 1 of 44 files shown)


pyproject.toml
```diff
@@ -81,12 +81,11 @@ ignore = [
     "B007",
     # f-string format
     "UP032",
-    # TODO: FIE ME
-    "G004",
     "B904",
     "SIM108",
     "SIM102"
 ]
+logger-objects = ["vllm.logger.logger"]
```
leo-pony, line 88: What is the purpose of this line? Why is it being modified?

MrZ20: It tells Ruff to treat vllm.logger.logger as a logger.

We added it because many files use:
```python
from vllm.logger import logger
```
Without this config, Ruff may miss `G004` issues like `logger.info(f"...")`. With it, future f-string logging regressions are caught by lint.

leo-pony (approved):
LGTM

PR #8777, opened 2026-04-28
[CI][BugFix] Remove e2e tracker guard on smart e2e

Potabk:
This patch fixed an edge case where Smart e2e wasn't triggered as expected when only tests were added to the `tests/ut` directory. Ideally, Smart E2E shouldn't use E2E triggering conditions as its entry point. Furthermore, the specific routing logic should be left to Smart E2E to determine itself.

- vLLM version: v0.19.1

### DIFF
--- a/.github/workflows/pr_test_light.yaml
+++ b/.github/workflows/pr_test_light.yaml
@@ -87,7 +87,6 @@ jobs:
               - 'CMakeLists.txt'
 
       - name: Determine smart UT test scope
-        if: ${{ steps.filter.outputs.e2e_tracker == 'true' }}
         id: scope
         run: |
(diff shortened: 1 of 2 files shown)


leo-pony (approved):
LGTM

.github/workflows/pr_test_light.yaml
```diff
@@ -87,7 +87,6 @@ jobs:
               - 'CMakeLists.txt'
 
       - name: Determine smart UT test scope
-        if: ${{ steps.filter.outputs.e2e_tracker == 'true' }}
```
leo-pony, line 90: 这里把ut_tracker为true添加上是不是就可以了？避免跳过现在的e2e-filter 的功能。

PR #8749, opened 2026-04-27
[Performance] adapter dcp pcp  full graph and async schedule spec

HF-001:
### What this PR does / why we need it?
fix bug in dcp、spec decode and async schedule. refer to https://github.com/vllm-project/vllm-ascend/pull/7640

- vLLM version: v0.19.0
- vLLM main: https://github.com/vllm-project/vllm/commit/6f786f2c506cb07f4566771fdc62e640e2c4a176

PR #8835, opened 2026-04-30
[BugFix] fix _seq_lens_cpu in spec decode

HF-001:
refer to: https://github.com/vllm-project/vllm-ascend/pull/8461 , PR # 8461 has already fixed the value of _seq_lens_cpu through asynchronous non blocking methods, but currently there are still redundant operations in the proposer, which are redundant and result in performance loss.

- vLLM version: v0.19.1


KlyzhenkoVadim:
@HF-001 Hi! Thanks for performance testing. It's my fix and it needed for draft_model correctness. I'll fix perfomance in few days. Also I prepare some performance benchmarks to make sure that .to("cpu") degrades performance. If you have any please share. Thanks a lot!

HF-001:
The relevant modifications have been merged together with this PR：https://github.com/vllm-project/vllm-ascend/pull/8749

PR #8836, opened 2026-04-30
[BugFix] [P/D] release proxy resources on stream cancellation

liziyu179:
Solve the problem of unbalanced proxy load in the PD separation scenario. When a user cancels a request, an asyncio.CancelledError exception is thrown, which does not belong to Exception, causing the load score of the D node not to be released.

by nightly

- vLLM version: v0.19.1


MengqingCao (approved):
LGTM

PR #8838, opened 2026-04-30
[BugFix] adapt copy_and_expand_eagle_input op to 910c

HF-001:
adapt the copy_and_expand_eagle_input operator to 910c, which is used in unified parallel spec decode, like Pard、P-eagle etc.

change to this pr: https://github.com/vllm-project/vllm-ascend/pull/9127

- vLLM version: v0.19.1

Collaborate with @NickJudyHvv


HF-001:
change to this pr: https://github.com/vllm-project/vllm-ascend/pull/9127

PR #8855, opened 2026-05-01
[Misc] Drop vlllm 0.19.1 support

wangxiyuan:
This PR removes legacy version-specific logic and checks related to vLLM v0.19.1 across the repository, including Dockerfiles, core scheduling logic, attention implementations, and end-to-end tests. It also updates the Docker build process to fetch vLLM using a specific commit hash. Critical issues were identified in all modified Dockerfiles where the VLLM_REPO argument was commented out while still being used in the RUN command, which will lead to build failures.

No. This is an internal cleanup and build configuration update.

The changes involve removing version-specific skips in existing tests and updating build scripts. Testing should be verified via CI build pipelines.

- vLLM version: v0.19.1

### DIFF
--- a/.github/workflows/pr_test_full.yaml
+++ b/.github/workflows/pr_test_full.yaml
@@ -80,7 +80,7 @@ jobs:
     name: e2e-full
     strategy:
       matrix:
-        vllm_version: [d886c26d4d4fef7d079696beb4ece1cfb4b008a8, v0.19.1]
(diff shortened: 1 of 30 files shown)


gcanlin:
`tests/e2e/multicard/2-cards/test_qwen3_moe.py` seems to be breaking. https://github.com/vllm-project/vllm-ascend/pull/8864 is fixing it. And this bug also happened in other PRs. It could be a bug in current main branch.

Issue #8858, opened 2026-05-01
[Bug]: kvpool layerwise loading crashes when use_layerwise=true


YuzhengWang5:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
Collecting environment information...
PyTorch version: 2.8.0+cpu
Is debug build: False

OS: openEuler 24.03 (LTS-SP2) (aarch64)
GCC version: (GCC) 12.3.1 (openEuler 12.3.1-99.oe2403sp2)
Clang version: 17.0.6 ( 17.0.6-45.oe2403sp2)
CMake version: version 4.3.2
Libc version: glibc-2.38

Python version: 3.11.14 (main, Feb 26 2026, 04:51:44) [GCC 12.3.1 (openEuler 12.3.1-99.oe2403sp2)] (64-bit runtime)
Python platform: Linux-5.10.0-136.12.0.86.r1526_92.hce2.aarch64-aarch64-with-glibc2.38

CPU:
Architecture:                       aarch64
CPU op-mode(s):                     64-bit
Byte Order:                         Little Endian
CPU(s):                             192
On-line CPU(s) list:                0-191
Vendor ID:                          HiSilicon
BIOS Vendor ID:                     HiSilicon
Model name:                         Kunpeng-920
BIOS Model name:                    HUAWEI Kunpeng 920 5250 To be filled by O.E.M. CPU @ 2.6GHz
BIOS CPU family:                    280
Model:                              0
Thread(s) per core:                 1
Core(s) per socket:                 48
Socket(s):                          4
Stepping:                           0x1
BogoMIPS:                           200.00
Flags:                              fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma dcpop asimddp asimdfhm ssbs
L1d cache:                          12 MiB (192 instances)
L1i cache:                          12 MiB (192 instances)
L2 cache:                           96 MiB (192 instances)
L3 cache:                           192 MiB (8 instances)
NUMA node(s):                       8
NUMA node0 CPU(s):                  0-23
NUMA node1 CPU(s):                  24-47
NUMA node2 CPU(s):                  48-71
NUMA node3 CPU(s):                  72-95
NUMA node4 CPU(s):                  96-119
NUMA node5 CPU(s):                  120-143
NUMA node6 CPU(s):                  144-167
NUMA node7 CPU(s):                  168-191
[...]

PR #8859, opened 2026-05-01
[v0.13.0][Bugfix] Fix layerwise kvpool completion tracking

YuzhengWang5:
Fix issue #8858 

This PR fixes completion tracking for the layerwise kvpool store path.
The non-layerwise path uses `stored_requests` to track pending store jobs, but the layerwise path does not have the same bookkeeping. As a result, completion could be handled incorrectly in `get_and_clear_finished_requests()`. This PR adds an explicit layerwise branch so a request is only reported finished after the layerwise send thread reports completion and the scheduler also marks the request as finished.

It's also ok to add ```stored_requests``` member in ```KVCacheStoreLayerSendingThread```, however, considering the existing finished mechanism in ```_handle_request``` of ```KVCacheStoreLayerSendingThread```, I only modify the logic of ```get_and_clear_finished_requests```.

No.

I have tested different prefixes (1K,4k,16k) and suffixes (1k,4k,16k) input prompt length of Qwen3-32B, and as a result, vllm-ascend reports no errors in issue #8858 .

### DIFF
--- a/vllm_ascend/distributed/kvpool/pool_worker.py
+++ b/vllm_ascend/distributed/kvpool/pool_worker.py
@@ -484,30 +484,41 @@ class KVPoolWorker:
 
     def get_and_clear_finished_requests(
             self, finished_req_ids, meta: AscendConnectorMetadata) -> set[str]:
-        finished_sending = set()
-        for req_id in meta.preempted_req_ids:
-            self.kv_send_thread.delete_finished_stored_request(  # type: ignore[union-attr]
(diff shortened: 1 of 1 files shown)


LCAIZJ:
LGTM

LCAIZJ:
Can you fix the pipeline errors first?

YuzhengWang5:
I dont know why my commit can trigger CI environment build error in e2e_test??? It's weird

PR #8876, opened 2026-05-05
[Feature]add gpt-oss infer model

shikang-hangzhou:
This PR adds the inference method for gpt-oss-120b with “FULL_DECODE_ONLY” to improved performance.

- vLLM version: v0.19.1


yiz-liu (commented):
You should be able to get `hf_config` and all the attrs in `vllm_config` so that we don't have to set `additional_config` manually.

shikang-hangzhou:
> You should be able to get `hf_config` and all the attrs in `vllm_config` so that we don't have to set `additional_config` manually.

I find the sinks info not include in config.json but index.json, so 'vllm_config' can not support this case.

PR #8931, opened 2026-05-07
[BugFix] Fix `info_once` logging with `unhashable` config

MrZ20:
This PR fixes a `TypeError: unhashable type: 'list'` triggered when `layer_sharding` is enabled.

`logger.info_once` may use log arguments as part of its once-only cache key. When `layer_sharding` is configured as a list, passing it as a positional logging argument can fail because lists are unhashable.

This patch updates dynamic `info_once` logs to format the message before calling `info_once`, so the logger receives a plain string and no longer hashes mutable config objects.

It also fixes an invalid `logger.Error(...)` call to use the standard `logger.error(...)` method.

- vLLM version: v0.19.1


leo-pony (approved):
LGTM

PR #8943, opened 2026-05-07
[v0.18.0][Bugfix] Fix layerwise kvpool completion tracking and loading token length

YuzhengWang5:

Combine PR #8858 and #8859. Fix the same bug meet in v0.18.0.

No.

Same to PR #8858 and #8859


LCAIZJ (approved):
lgtm

PR #8958, opened 2026-05-07
[BugFix] Fix Qwen2 compiled-path outputs on Ascend

ShuhaoZhangTony:
## Summary

- Use a native rotary fallback for Qwen2 on Ascend while keeping the compiled path enabled.
- Avoid forcing the complete Qwen2 runtime into eager mode as the default safety behavior.

## Validation

- Single-request Qwen2.5-14B-Instruct functional check on Ascend with the default compiled path after the fix.
- Performance results will only be added after a fixed-commit, repeated real-hardware comparison against the PR base is complete.

- vLLM version: v0.24.0


ShuhaoZhangTony:
Continued as #12343 on the required `feature/...` branch. The replacement is rebased onto current `main`, reduces the active diff to the focused Qwen2 change, adds path-selection regression coverage, and remains Draft until fixed-commit real-NPU correctness and performance evidence is attached. This PR is closed to keep one active review target.

Issue #8427, opened 2026-04-20
[Bug]: Mooncake transfer failed后自愈，概率导致后续请求精度异常（答非所问，重复等）


1092626063:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
Your output of above commands here
```

</details>

### 🐛 Describe the bug

P/D节点间的hccl建链失败，造成D节点连接P节点时mooncake通信失败，报错如下：
Failed to transfer KV cache for request chatcmpl-xxxxxxxx: Mooncake transfer failed, ret: -1

后续单独curl一个请求，输出重复的'\n'

weiguihua2:
@liziyu179 plz take a look

WANG-WADE:
遇到了类似的问题，能问下启动脚本是什么吗

xhy15025776151-ops:
@1092626063 Did the problem occur after you had been running it for a while? Have you solved it?

1092626063:
> [@1092626063](https://github.com/1092626063) Did the problem occur after you had been running it for a while? Have you solved it?

Not yet, still working on it.
Best to first figure out why the HCCL connection is failing. We can look into Mooncake's robustness and version fixes afterward.

zzzzzmeng:
We have solved a similar issue. Please refer to PR 8959.

PR #8959, opened 2026-05-07
[P/D][BugFix] Fix for transmit kv cache failure

### LINKED ISSUE #7871: [Bug]: KV load failure path triggers metrics exception and internal server error with kv_load_failure_policy=fail
### Your current environment

<details>

```text
PyTorch: 2.9.0+cpu
torch_npu: 2.9.0
vLLM: 0.17.0
vLLM-ascend: 0.17.0rc2.dev0
CANN: 8.5.1 (V100R001C25SPC002B220)
OS: Huawei Cloud EulerOS 2.0 (aarch64)
Python: 3.11.4
glibc: 2.34
CPU: Kunpeng 920 7285Z, 320 CPUs, 8 NUMA nodes
NPU: 16 x Ascend910, all health OK

Warnings:
- CANN install path owner does not match current owner
- torch_npu plugin library owner does not match current owner

Key env:
- TORCH_DEVICE_BACKEND_AUTOLOAD=1
- PYTORCH_NVML_BASED_CUDA_CHECK=1
- TORCHINDUCTOR_COMPILE_THREADS=1

```

</details>

### 🐛 Describe the bug

When KV offloading is enabled and `kv_load_failure_policy=fail` is configured, a request may hit KV load inconsistency across TP ranks.
Failing the request in this case is understandable under the configured policy.

However, the actual issue is that the KV load failure path appears to trigger a **secondary metrics exception**:

```text
ValueError: Counters can only be incremented by non-negative amounts.
```

This causes the request to surface as an unexpected internal server error / HTTP 500, instead of remaining a clean policy-driven failure.

---

### To Reproduce

Run with KV offloading enabled and:

```yaml
kv_load_failure_policy: fail
```

In the observed case, the same request showed inconsistent KV load results across TP ranks, for example:

```text
[Rank1] success_blocks=173, total_blocks_num=173, status=SUCCESS
[Rank0] success_blocks=6, total_blocks_num=173, status=SUCCESS
```

The scheduler then failed the request due to KV load failure, and the failure path further triggered a metrics exception.

---

### Expected behavior

When `kv_load_failure_policy=fail` is configured and TP ranks observe inconsistent KV load results:

1. the request may fail according to policy;
2. the failure should be handled cleanly;
3. metrics/profiling code should not throw additional exceptions;
4. the system should not escalate the failure into an unrelated internal server error.

---

### Actual behavior

The observed behavior was:

1. most TP ranks reported full async load success;
2. a few TP ranks reported partial block load only;
3. scheduler failed the request due to KV load failure;
4. metrics recording then threw:

   ```text
   ValueError: Counters can only be incremented by non-negative amounts.
   ```
5. the request surfaced as internal server error / HTTP 500, with frontend first-token timeout.

---

### Environment

* `vllm-ascend`
* KV offloading: enabled
* external KV backend: enabled
* deployment mode: PD separation
* observed topology: multi-node, TP16

Configuration:

```yaml
kv_load_failure_policy: fail
```

---

### Relevant logs

KV load results for the same request were inconsistent across TP ranks:

```text
[adapter.py:228] [GetResult] Req <request_id> async load done, success_blocks=173, total_blocks_num=173, status=SUCCESS, cost_ms=4275.57
[adapter.py:228] [GetResult] Req <request_id> async load done, success_blocks=173, total_blocks_num=173, status=SUCCESS, cost_ms=4262.91
[adapter.py:228] [GetResult] Req <request_id> async load done, success_blocks=6, total_blocks_num=173, status=SUCCESS, cost_ms=4275.40
[adapter.py:228] [GetResult] Req <request_id> async load done, success_blocks=6, total_blocks_num=173, status=SUCCESS, cost_ms=4277.36
```

Scheduler failed the request:

```text
[scheduler.py:2192] Failing 1 request(s) due to KV load failure (failure_policy=fail, 21376 tokens affected). Request IDs: {'<request_id>'}
```

Frontend / serving side then saw timeout and internal error:

```text
request get chunk timeout, is_first_token=True, exceeds 20.0 s
```

```text
Request <request_id> failed with an internal error during generation
```

```text
Internal server error
```

Secondary exception from metrics path:

```text
  File ".../async_llm.py", line 345, in output_handler
    logger_ref[0].record(
  File ".../vllm/v1/metrics/loggers.py", line 1316, in record
    stat_logger.record(
  File ".../vllm/v1/metrics/loggers.py", line 1120, in record
    self.counter_prompt_tokens_by_source[source][engine_idx].inc(
  File ".../prometheus_client/metrics.py", line 313, in inc
    raise ValueError('Counters can only be incremented by non-negative amounts.')
ValueError: Counters can only be incremented by non-negative amounts.
```

---

### Impact

This issue makes diagnosis harder and worsens the user-visible failure mode:

* KV load failure under configured `fail` policy is expected;
* but the additional metrics exception turns it into a secondary internal error;
* frontend sees timeout / 500 rather than a bounded failure path.

In one workload we observed:

* 1066 total requests
* 183 affected requests

---

### Possible root cause

There may be two separate issues:

1. **KV load inconsistency across TP ranks**

   * same request reports different `success_blocks` on different TP ranks

2. **metrics accounting bug on failure path**

   * a negative delta appears to be passed into a Prometheus counter during failure handling / fallback accounting

The second one is the primary issue being reported here.

---

### Suggested fix

1. Ensure failure-path metrics never pass negative values into Prometheus counters.
2. Keep `kv_load_failure_policy=fail` behavior isolated to request failure handling, without triggering secondary metrics exceptions.
3. Investigate TP-level KV load inconsistency separately if needed.

---

### Additional context

Using:

```yaml
kv_load_failure_policy: recompute
```

may reduce user-visible failures, but that is not the main issue here.

The main issue is that under `fail` policy, the system should still avoid throwing:

```text
ValueError: Counters can only be incremented by non-negative amounts.
```

> The issue is that once KV load failure happens, the metrics/profiling path should not throw a secondary exception and turn it into an internal server error.

zzzzzmeng:
-- When kv cache transmite failed, blocks with load errors need to be marked invalid and end this request.
-- For mooncake_layerwise_connector, add a new 'FAILED_SENDING_MSG' signal to communicate transmission failures between the producer and consumer, ensuring both sides can clean up resources appropriately.
-- For recving thread, implement tracking for invalid block IDs that fail to load, providing a way for the scheduler to identify and potentially retry failed blocks.

fixes: https://github.com/vllm-project/vllm-ascend/issues/7871 & https://github.com/vllm-project/vllm-ascend/issues/8427
No
Test: Use the gsm8k dataset to construct a scenario where the KV cache transmission of the first part of.
- vLLM version: v0.19.1

### DIFF
--- a/tests/ut/kv_connector/test_mooncake_connector.py
+++ b/tests/ut/kv_connector/test_mooncake_connector.py
@@ -277,6 +277,23 @@ class TestKVCacheRecvingThreadBasic(unittest.TestCase):
         self.assertEqual(queued["request_id"], "req1")
         self.assertEqual(queued["remote_host"], "localhost")
 
+    def test_mark_and_is_failed(self):
+        self.thread._mark_failed_recv_request("req1", [10, 20])
+        self.assertTrue(self.thread._is_failed_recv_request("req1"))
+        self.assertIn(10, self.thread.invalid_block_ids)
+        self.assertIn(20, self.thread.invalid_block_ids)
+
+    def test_clear_failed_recv_request(self):
+        self.thread._mark_failed_recv_request("req2", [30])
+        self.thread._clear_failed_recv_request("req2")
+        self.assertFalse(self.thread._is_failed_recv_request("req2"))
+
+    def test_get_and_clear_invalid_block_ids(self):
+        self.thread.invalid_block_ids = {1, 2, 3}
+        result = self.thread.get_and_clear_invalid_block_ids()
+        self.assertSetEqual(result, {1, 2, 3})
+        self.assertEqual(self.thread.invalid_block_ids, set())
+
     @patch.object(KVCacheTaskTracker, "get_and_clear_finished_requests")
     def test_get_finished_requests(self, mock_tracker):
         mock_tracker.return_value = {"req1", "req2"}
--- a/tests/ut/kv_connector/test_mooncake_layerwise_connector.py
+++ b/tests/ut/kv_connector/test_mooncake_layerwise_connector.py
@@ -211,6 +211,25 @@ class TestKVCacheRecvingLayerThread(unittest.TestCase):
         self.meta = MooncakeAgentMetadata(te_rpc_port=6000, kv_caches_base_addr=[0x1, 0x2])
         self.ready_event = threading.Event()
 
+    def test_get_and_clear_done_requests(self):
+        th = KVCacheRecvingLayerThread(
+            tp_rank=0,
+            side_channel_port=5555,
+            tp_size=2,
+            pd_head_ratio=1,
+            local_engine_id="engineA",
+            metadata=self.meta,
+            ready_event=self.ready_event,
+        )
+
+        with th.lock:
+            th.done_requests.update({"r1", "r2"})
+        got = th.get_and_clear_done_requests()
+        self.assertEqual(got, {"r1", "r2"})
+
+        got2 = th.get_and_clear_done_requests()
+        self.assertEqual(got2, set())
+
     def test_get_and_clear_finished_requests(self):
         th = KVCacheRecvingLayerThread(
             tp_rank=0,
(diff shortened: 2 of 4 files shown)


weijinqian0 (approved):

MengqingCao (approved):
LGTM

PR #8971, opened 2026-05-08
Update DeepSeek-V4-Pro.md

TaDa-aaa:
Update DeepSeek-V4-Pro.md

for Update DeepSeek-V4-Pro.md，A2-Prefill-Decode Disaggregation


docs/source/tutorials/models/DeepSeek-V4-Pro.md
```diff
@@ -1045,7 +1050,7 @@ Before you start, please
                       }
```
MengqingCao, line 1050: the parallel machenism is not aligned with A3

MengqingCao (commented):
plz check again and make sure the scripts are aligned with that of A3

Issue #8772, opened 2026-04-28
[Doc]: qwen35-27B代码仓库测试流程和体验以及改进建议


aim-to-be:
### 📚 The doc issue

参考链接是https://docs.vllm.ai/projects/ascend/en/latest/tutorials/models/Qwen3.5-27B.html
1.installation
步骤quay.io/ascend/vllm-ascend:v0.17.0rc1(for Atlas 800 A2) 和https://docs.vllm.ai/projects/ascend/en/latest/installation.html#set-up-using-docker
链接给到的适配A2的镜像是vllm-ascend:v0.18.0rc1，前后明显不一致，经过测试v0.18.0镜像可以正常拉起来但是0.17.0不行，建议更正，以免用户拉取错误的镜像

MengqingCao:
@shaopeng-666 could you make a check and update the doc?

shaopeng-666:
Wait pr 9006 merge.

LoganJane:
@aim-to-be This issue has been resolved in the latest version. Please use the latest release or rc version (https://github.com/vllm-project/vllm-ascend/releases). If the issue persists, feel free to submit a new issue.

PR #9006, opened 2026-05-09
[Doc]Update Qwen3.5 tutorial image versions to v0.19.1rc1

wyu0-0:

Fix the inconsistent vllm-ascend image version in the Qwen3.5-27B and Qwen3.5-397B deployment tutorial. Update the outdated image tag v0.17.0rc1 to the correct and available v0.19.1rc1 for Atlas 800 A2 and A3, aligning with the official installation documentation. This avoids users pulling an invalid image and encountering deployment startup failures.

This change is made based on user feedback from Issue #8772, where users reported that the old image version cannot be pulled or started correctly. Updating the document ensures users can use the valid image and avoid deployment failures.

No

Verified that the v0.19.1rc1 image can be pulled and started normally.

- vLLM version: v0.19.1
- vLLM main: https://github.com/vllm-project/vllm/commit/4d51588e2381018348f1022dfa3a7698899805b7

### DIFF
--- a/docs/source/tutorials/models/Qwen3.5-27B.md
+++ b/docs/source/tutorials/models/Qwen3.5-27B.md
@@ -32,7 +32,7 @@ If you want to deploy multi-node environment, you need to verify multi-node comm
 :::::{tab-set}
 ::::{tab-item} Use docker image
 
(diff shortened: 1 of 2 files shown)


MengqingCao:
I think the doc on main should update to the latest version 0.19.1rc1?

wyu0-0:
> I think the doc on main should update to the latest version 0.19.1rc1?

@MengqingCao Fixed the comments. Please take a look.

PR #9017, opened 2026-05-09
[Feature]:Modify the UvaBufferWrapper class to enable UVA features

orangeBoxun:

GPUs have the UVA feature, which allows the GPU side to directly access CPU tensors based on the same virtual address, without manual copying. The NPU side also needs to follow up on this.

This feature needs to be enabled through os.environ['PYTORCH_NPU_ALLOC_CONF'] = 'pinned_mem_register:True'; otherwise, it is disabled by default.

Enable os.environ['PYTORCH_NPU_ALLOC_CONF'] = 'pinned_mem_register:True', and then use the regular vllm test cases.

- vLLM version: v0.23.0


vllm_ascend/patch/worker/patch_v2/patch_uva.py
```diff
@@ -89,14 +90,41 @@ def __getattr__(self, name):
...
+                error_info = (
+                    "UVA is not available, because cpu tensor is not pinned, "
+                    "try to use .pin_memory() to cpu_tensor wantted to use uva feature"
+                )
+                raise RuntimeError(error_info)
+            self._np = self._cpu.numpy()
+            self._uva = self._cpu
+        else:
+            self._cpu = torch.zeros(size, dtype=dtype, device="cpu", pin_memory=True)
+            self._np = self._cpu.numpy()
```
Ronald1995: it's better to define self._cpu and self._np in the begging of function.

vllm_ascend/patch/worker/patch_v2/patch_uva.py
```diff
@@ -89,14 +90,41 @@ def __getattr__(self, name):
...
+            value = os.environ[key]
+            if "pinned_mem_register:True" not in value:
+                error_info = (
+                    f"UVA is not available, because environment_param {key} "
+                    f"lack value 'pinned_mem_register:True', "
+                    f"try to add os.environ['PYTORCH_NPU_ALLOC_CONF'] = 'pinned_mem_register:True'"
+                )
+                raise RuntimeError(error_info)
+            self._cpu = torch.zeros(size, dtype=dtype, device="cpu", pin_memory=True)
+            if not self._cpu.is_pinned():
```
Ronald1995: it's not neccessary of this branch, because we have set `torch.zeros(size, dtype=dtype, device="cpu", pin_memory=True)`

vllm_ascend/patch/worker/patch_v2/patch_uva.py
```diff
@@ -16,13 +16,24 @@
...
 
+def is_uva_available() -> bool:
+    """check if uva feature is supported in this environment"""
+    result = False
+    key = "PYTORCH_NPU_ALLOC_CONF"
+    if key in os.environ:
+        value = os.environ[key]
+        if "pinned_mem_register:True" in value:
+            result = True
+    return result
```
Ronald1995: you could implement it like this

```python
def is_uva_available() -> bool:
    """check if uva feature is supported in this environment"""
    return "pinned_mem_register:True" in os.environ.get("PYTORCH_NPU_ALLOC_CONF", {})

```

vllm_ascend/patch/worker/patch_v2/patch_uva.py
```diff
@@ -16,13 +16,19 @@
...
 from collections.abc import Callable, Sequence
 
 import numpy as np
 import torch
 import vllm.v1.worker.gpu.buffer_utils
 
 
+def is_uva_available() -> bool:
+    """check if uva feature is supported in this environment"""
+    return "pinned_mem_register:True" in os.environ.get("PYTORCH_NPU_ALLOC_CONF", {})
```
Ronald1995: it's better to validate the version of triton_ascend and add fixme. such as
```python
def is_uva_available() -> bool:
    """check if uva feature is supported in this environment"""
     #FIXME(): why we don't support triton-ascend 3.2.1 and fix plan.
    return "pinned_mem_register:True" in os.environ.get("PYTORCH_NPU_ALLOC_CONF", {}) and triton_version <= 3.2.0
```

PR #9037, opened 2026-05-10
[CI] Add csrc cache for image build

Potabk:
This pull request introduces a csrc build artifact caching mechanism into image building, which is expected to reduce image build time from 50 minutes to 10 minutes.

image

- vLLM version: v0.20.1

### DIFF
--- a/.github/workflows/push_build_csrc_cache.yaml
+++ b/.github/workflows/push_build_csrc_cache.yaml
@@ -67,7 +67,7 @@ jobs:
         include:
           # arch → runner
           - arch: X64
-            runner: linux-amd64-cpu-8-hk
+            runner: linux-amd64-cpu-16-hk
           - arch: ARM64
             runner: linux-arm64-cpu-16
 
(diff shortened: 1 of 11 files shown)


MengqingCao (approved):
lgtm

.github/workflows/push_build_csrc_cache.yaml
```diff
@@ -48,55 +44,112 @@ env:
...
+        run: |
+          yum install -y gcc gcc-c++ cmake numactl-devel clang
+          pip install uv
+
+      - name: Install clang (ubuntu, non-310p only)
+        if: matrix.os == 'ubuntu' && matrix.chip_type != '310p'
+        run: |
+          apt-get -y install clang-15
+          update-alternatives --install /usr/bin/clang clang /usr/bin/clang-15 20
+          update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-15 20
```
MengqingCao: why not move this to L132?

Potabk: ok, let's make it more concise

MengqingCao (approved):
LGTM now, I think we could do some optimization on clearing the history kernel compilation cache, but I'm okay with this initial support, thx for your work!

PR #9038, opened 2026-05-10
[Test][Feature] Add unit tests for check_and_update_config

ZT-AIA:
Testing the functionality of `check_and_update_config`

No

ut

- vLLM version: v0.20.2


tests/ut/test_check_and_update_config.py
```diff
@@ -0,0 +1,77 @@
...
+
+        if parallel_config_enable_eplb:
+            args = parser.parse_args(["--enable-eplb"])
+            vllm_config = EngineArgs.from_cli_args(args=args).create_engine_config()
+      
+        if additional_config:
+            args = parser.parse_args([
+                "--additional-config", '{"enable_cuda_graph": true}'
+            ])
+            vllm_config = EngineArgs.from_cli_args(args=args).create_engine_config()
```
yiz-liu: What is this? Have not seen this arg before.

tests/ut/test_check_and_update_config.py
```diff
@@ -0,0 +1,77 @@
...
+        (False, False, False, CUDAGraphMode.FULL_DECODE_ONLY, "JackFram/llama-68m", "abhigoyal/vllm-medusa-llama-68m-random", False,  AttentionBackendEnum.XPU_MLA_SPARSE, CUDAGraphMode.FULL_DECODE_ONLY),
+        (True, True, False, CUDAGraphMode.FULL_AND_PIECEWISE, "JackFram/llama-68m", "abhigoyal/vllm-medusa-llama-68m-random", False,  AttentionBackendEnum.XPU_MLA_SPARSE, CUDAGraphMode.PIECEWISE),
+        (False, True, False, CUDAGraphMode.FULL_AND_PIECEWISE, "JackFram/llama-68m", "abhigoyal/vllm-medusa-llama-68m-random", False,  AttentionBackendEnum.XPU_MLA_SPARSE, CUDAGraphMode.PIECEWISE),
+        (True, False, False, CUDAGraphMode.FULL_AND_PIECEWISE, "JackFram/llama-68m", "abhigoyal/vllm-medusa-llama-68m-random", False,  AttentionBackendEnum.XPU_MLA_SPARSE, CUDAGraphMode.PIECEWISE),
+        (False, False, False, CUDAGraphMode.FULL_AND_PIECEWISE, "JackFram/llama-68m", "abhigoyal/vllm-medusa-llama-68m-random", False,  AttentionBackendEnum.XPU_MLA_SPARSE, CUDAGraphMode.PIECEWISE),
+   
+    ])
```
yiz-liu: Why is `enable_eplb=False` in every case?

yiz-liu: `speculative_config_target_model` is not used in any case?

yiz-liu: Ah the three speculative config args are all identical in every cases, not what we expected. You should have `suffix decoding`, `ngram`, and etc. as the options, not like this.

yiz-liu: The `attn_backend` should use ours, not from upstream vLLM.

yiz-liu: Also you need to have encoder-decoder model, `compilation_config.mode`, and those args you have in your notes as params.

yiz-liu: Overall, there is still a lot more to fix in this PR.

tests/ut/test_check_and_update_config.py
```diff
@@ -0,0 +1,196 @@
...
+                "cudagraph_mode": compilation_config_cudagraph_mode.name
+            }
+            compilation_config_str = json.dumps(compilation_config)
+            cli_args.extend(["--compilation-config", compilation_config_str])
+            
+            # Add tensor parallel size if specified
+            if parallel_config_tensor_parallel_size is not None:
+                cli_args.extend(["--tensor-parallel-size", str(parallel_config_tensor_parallel_size)])
+            
+            # Add speculative config if needed
```
yiz-liu:
```python
speculative_config = {
    "enforce_eager" = speculative_config_enforce_eager,
    "method" = speculative_config_method,
    ...
}
```

yiz-liu (commented):
Remove attn backend as CGSupport is validated in runner. And why are we having tp size as a param?

PR #9060, opened 2026-05-11
[Feature] support Flash Attention 3 (FA3) for training-inference consistency.

wangx700:
This PR introduces support for Flash Attention 3 (FA3) on Ascend NPUs, providing a FA3 attention backend for training-inference consistency. It includes the implementation of the FA3 backend, integration into the platform's attention selector, and comprehensive tests to ensure correctness with existing backends.([issue 10072](https://github.com/vllm-project/vllm-ascend/issues/10072))

Currently, external users cannot use it directly because the Flash Attention 3 package has not been open-sourced yet (**expected to be open-sourced at the 20th of May**). 
Once it is open-sourced, users will be able to use Flash Attention 3 by specifying the flash_attn backend (by setting the LLM parameter **attention_backend="FLASH_ATTN"**) and enabling batch invariance (set **VLLM_BATCH_INVARIANT=1**).

- End-to-end tests comparing FA3 output with Fused Infer Attention (FIA) for consistency across various prompt lengths and batch sizes.
- Unit tests for the custom FA3 operations.

- vLLM version: v0.20.1


vllm_ascend/platform.py
```diff
@@ -901,7 +939,10 @@ def _fix_incompatible_config(vllm_config: VllmConfig) -> None:
                 att_config.flash_attn_version = None
 
             # Notify user that the backend will be managed by Ascend plugins
-            if getattr(att_config, "backend", None) is not None:
+            if (
+                getattr(att_config, "backend", None) is not None
+                and att_config.backend != AttentionBackendEnum.FLASH_ATTN
```
Ronald1995, line 941: add some comments to explain why we need to use AttentionBackendEnum.FLASH_ATTN

vllm_ascend/platform.py
```diff
@@ -609,7 +618,36 @@ def get_attn_backend_cls(cls, selected_backend, attn_selector_config, num_heads:
         if is_310p():
             return backend_map_310.get(key, backend_map_310[(False, False)])
 
-        return backend_map[key]
+        backend_cls_path = backend_map[key]
+
+        if selected_backend == AttentionBackendEnum.FLASH_ATTN:
+            backend_cls_path = cls._resolve_fa3_backend(key, attn_selector_config)
```
Ronald1995: move the logic to the beginning of the method

vllm_ascend/platform.py
```diff
@@ -609,7 +618,36 @@ def get_attn_backend_cls(cls, selected_backend, attn_selector_config, num_heads:
...
+
+        return backend_cls_path
+
+    @classmethod
+    def _resolve_fa3_backend(cls, key, attn_selector_config):
+        if not attn_selector_config.use_batch_invariant:
```
Ronald1995, line 622: if not attn_selector_config.use_batch_invariant or not fa3:
    return 
logging.info("training-inference consistency scenario....., we will use fa3,which will make performance")
.....

vllm_ascend/platform.py
wangx700: ok, the logic was moved to the beginning of the method

vllm_ascend/platform.py
wangx700: ok，logging.info for training-inference consistency scenario has been added.

vllm_ascend/platform.py
wangx700: ok，added

MengqingCao (commented):
I think this pr still has somethings to confirm:
1. confirm if the semantics of `flash_attn_v3.flash_attn_with_kvcache` and the fa3 of gpu are same, make sure there is no ambiguity. we should not use the same enumeration value as the GPU's flash attn backend.
2. what's the plan of opening packages `flash_attn_v3`? If we could not use it immediately, plz declare that limitation clearly in the pr message.
3. plz add doc for the usage of this feature and note the limitation in the doc also.

wangx700:
> I think this pr still has somethings to confirm:
> 
> 1. confirm if the semantics of `flash_attn_v3.flash_attn_with_kvcache` and the fa3 of gpu are same, make sure there is no ambiguity. we should not use the same enumeration value as the GPU's flash attn backend.

The `flash_attn_with_kvcache` interface on NPU is semantically consistent with the GPU FA3 version in terms of API parameters. The key differences are:
(1) **Unsupported features on NPU FA3**: Sliding window attention, RoPE, ALiBi, Softcapping, and FP8 quantization are not yet supported.
(2) **Graph capture**: The tiling of `flash_attn_with_kvcache` is processed on the host side and is currently being optimized. It does not support ACL graph capture (i.e., cannot be captured into a computational graph for acceleration). Please use `enforce_eager=True` when enabling FA3.
Plz see the https://github.com/vllm-project/vllm-ascend/pull/9060/changes#diff-bcef3a49122cb59cb9323acf2f841294451d525a9ba56ce8d4f0fe905bb2f8ff
> 2. what's the plan of opening packages `flash_attn_v3`? If we could not use it immediately, plz declare that limitation clearly in the pr message.

The day of opening packages `flash_attn_v3` expected to be released after May 20th.
> 3. plz add doc for the usage of this feature and note the limitation in the doc also.

Plz see the https://github.com/vllm-project/vllm-ascend/pull/9060/changes#diff-bcef3a49122cb59cb9323acf2f841294451d525a9ba56ce8d4f0fe905bb2f8ff

weijinqian0 (approved):

vllm_ascend/ops/rotary_embedding.py
```diff
@@ -561,11 +561,12 @@ def __init__(
...
-            enforce_enable=enforce_enable,
-            is_neox_style=is_neox_style,
-            enable_fp32_compute=enable_fp32_compute,
-        )
+        from vllm.model_executor.custom_op import CustomOp
+
+        CustomOp.__init__(self, enforce_enable=enforce_enable)
+        self.is_neox_style = is_neox_style
+        self.enable_fp32_compute = enable_fp32_compute
+        self.apply_rotary_emb_flash_attn = None
```
weijinqian0, line 569: need remove this.

wangx700: don‘t need remove after reviewed.

tests/e2e/singlecard/test_attention_fa3.py
```diff
@@ -0,0 +1,154 @@
...
+        max_model_len=MAX_MODEL_LEN,
+        enforce_eager=True,
+        gpu_memory_utilization=0.7,
+        **runner_kwargs,
+    ) as runner:
+        return runner.generate_greedy(prompts, max_tokens)
+
+
+def _generate_logprobs_with_backend(prompts, max_tokens=5, num_logprobs=5, **runner_kwargs):
+    with VllmRunner(
```
MengqingCao, line 56: let's move this e2e test to https://github.com/vllm-project/vllm-ascend/tree/main/tests/e2e/singlecard instead of in ops.

vllm_ascend/ops/rotary_embedding.py
```diff
@@ -561,11 +561,12 @@ def __init__(
...
         enable_fp32_compute: bool = False,
     ) -> None:
-        super().__init__(
-            enforce_enable=enforce_enable,
-            is_neox_style=is_neox_style,
-            enable_fp32_compute=enable_fp32_compute,
-        )
+        from vllm.model_executor.custom_op import CustomOp
+
+        CustomOp.__init__(self, enforce_enable=enforce_enable)
```
MengqingCao, line 566: why changing this from super to this way here?

vllm_ascend/platform.py
```diff
@@ -611,6 +623,33 @@ def get_attn_backend_cls(cls, selected_backend, attn_selector_config, num_heads:
 
         return backend_map[key]
 
+    @classmethod
+    def _use_fa3_backend(cls, key, attn_selector_config):
```
MengqingCao:
```suggestion
    def _validate_fa3_backend(cls, key, attn_selector_config):
```

vllm_ascend/ops/rotary_embedding.py
wangx700: https://github.com/vllm-project/vllm/blob/releases/v0.20.1/vllm/model_executor/layers/rotary_embedding/common.py#L137-L138
Because the flash_attn package for NPU also contains the flash_attn module, but it does not have flash_attn.ops, importing from flash_attn.ops.triton.rotary import apply_rotary would raise a ModuleNotFoundError. To avoid this error, we used inheritance to remove this part of the logic.

tests/e2e/singlecard/test_attention_fa3.py
wangx700: ok，this e2e test was moved to '/tests/e2e/singlecard'

vllm_ascend/platform.py
wangx700: ok, the _use_fa3_backend had been changed to _validate_fa3_backend.

MengqingCao (approved):

PR #9064, opened 2026-05-11
[Refactor]: migrate env vars to AscendConfig

lizy124:
Migrate selected vllm-ascend environment variables from direct env access to the
  `AscendConfig` / `additional_config` configuration path, enabling users to
  configure these options through `--additional-config` while keeping the existing
  environment variables as backward-compatible fallbacks during the transition
  period.

  The migration follows this precedence:

  `additional_config` explicit value > deprecated environment variable fallback >
  default value

  Migrated environment variables:

  - `VLLM_ASCEND_BALANCE_SCHEDULING` → `enable_balance_scheduling` (default:
  `False`)
  - `VLLM_ASCEND_ENABLE_FLASHCOMM1` → `enable_flashcomm1` (default: `False`)
  - `VLLM_ASCEND_ENABLE_MATMUL_ALLREDUCE` → `enable_matmul_allreduce` (default:
  `False`)
  - `VLLM_ASCEND_FLASHCOMM2_PARALLEL_SIZE` → `enable_flashcomm2_parallel_size`
  (default: `0`)
  - `MSMONITOR_USE_DAEMON` → `msmonitor_use_daemon` (default: `False`)
  - `VLLM_ASCEND_ENABLE_MLAPO` → `enable_mlapo` (default: `True`)
  - `VLLM_ASCEND_ENABLE_NZ` → `weight_nz_mode` (default: `1`)
  - `VLLM_ASCEND_ENABLE_CONTEXT_PARALLEL` → `enable_context_parallel` (default:
  `False`)
  - `VLLM_ASCEND_ENABLE_FUSED_MC2` → `enable_fused_mc2` (default: `0`)
  - `VLLM_ASCEND_FUSION_OP_TRANSPOSE_KV_CACHE_BY_BLOCK` →
  `enable_transpose_kv_cache_by_block` (default: `True`)

  The original environment variables remain functional for now and are marked as
  deprecated where applicable, so existing deployments continue to work while users
   migrate to `--additional-config`.

Yes. Users can now use --additional-config '{"enable_mlapo": true, "enable_flashcomm2_parallel_size": 2}' instead of setting environment variables. Original environment variables are still supported for backward compatibility.

- vLLM version: v0.20.1

### DIFF
--- a/tests/ut/spec_decode/test_eagle_proposer.py
+++ b/tests/ut/spec_decode/test_eagle_proposer.py
@@ -17,7 +17,7 @@ from vllm.v1.spec_decode.draft_model import DraftModelProposer
 import vllm_ascend.spec_decode.eagle_proposer as eagle_proposer
 from tests.ut.base import TestBase
 from tests.ut.conftest import npu_test
-from vllm_ascend.ascend_config import init_ascend_config
+from vllm_ascend.ascend_config import clear_ascend_config, init_ascend_config
 from vllm_ascend.attention.attention_v1 import AscendAttentionState
 from vllm_ascend.attention.utils import AscendCommonAttentionMetadata
 from vllm_ascend.spec_decode.draft_proposer import AscendDraftModelProposer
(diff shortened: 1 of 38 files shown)


weijinqian0:
TMG review is required.

wangxiyuan:
we should deprecate the env and let it work at least one release. Then remove it in the next release

tests/ut/spec_decode/test_eagle_proposer.py
```diff
@@ -888,6 +690,16 @@ def setUp_and_tearDown(self):
         self.vllm_config.parallel_config.data_parallel_rank = 0
         self.vllm_config.parallel_config.data_parallel_size = 1
         self.vllm_config.parallel_config.prefill_context_parallel_size = 1
+        self.vllm_config.parallel_config.enable_expert_parallel = False
```
lilinsiman, line 965: Why is it necessary to specifically add these mocks and assignments？

lizy124: In the test, `init_ascend_config(self.vllm_config)` has already been called, which creates a real `AscendConfig` object and stores it in the global `_ASCEND_CONFIG`. After that, `get_ascend_config()` would normally return this real object.

However, the issue is that the test sets `self.vllm_config.additional_config = None`, while `AscendConfig.__init__` performs extensive initialization work (creating `AscendCompilationConfig`, `ProfilingChunkConfig`, etc.), which requires many attributes of `vllm_config` to be correctly set. If the `vllm_config` attributes are incomplete, `init_ascend_config` will raise an error.

Therefore, the current approach is:

1. Call `init_ascend_config(self.vllm_config)` → creates a real `AscendConfig` (though the initialization may be incomplete)
2. Then mock `get_ascend_config()` → overrides the real object and returns a `MagicMock` with 11 manually configured attributes.

linfeng-yuan (approved):
This PR was reviewed and approved by @wangxiyuan @MengqingCao in May 21. Let us merge it to build our test image.

vllm_ascend/envs.py
```diff
@@ -71,6 +71,7 @@
     "VLLM_ASCEND_ENABLE_MATMUL_ALLREDUCE": lambda: bool(int(os.getenv("VLLM_ASCEND_ENABLE_MATMUL_ALLREDUCE", "0"))),
     # Whether to enable FlashComm optimization when tensor parallel is enabled.
     # This feature will get better performance when concurrency is large.
+    # DEPRECATED: use additional_config.enable_flashcomm1 instead.
```
MengqingCao, line 74: we should add this warning in the log instead of comment, otherwise the users could not get this info.

PR #9084, opened 2026-05-12
[BugFix][310p] fix 310p ops compile failed

Tflowers-0129:
### What this PR does / why we need it?

Ascend 310P Support: Added 'ascend310p' to the supported compute units in both CMake configuration and the build script.
Build Configuration Update: Updated SOC_VERSION_LIST and ARCH_DIRECTORY_LIST in CMakeLists.txt to include the 310P architecture.
Bug Fix: Corrected a variable naming error in the CMake list search logic to ensure proper architecture detection.

### Does this PR introduce _any_ user-facing change?

NA

### How was this patch tested?

local test and ci
- vLLM version: v0.20.1
- vLLM main: https://github.com/vllm-project/vllm/commit/c7aa186d67b6f051680831418e957c67f34ba7a2

PR #9094, opened 2026-05-12
[Ops][BugFix] Fix arch32 detection in ARCH_DIRECTORY setup

MarcelFerrari:

This PR fixes a typo in the CMake logic for `arch32` detection. The result of `list(FIND)` was stored in `INDXE` but checked using `INDEX`, which could lead to incorrect build configurations if `INDEX` held a stale value from a previous lookup.

No. This is a build-system correctness fix.

- vLLM version: v0.20.1


MarcelFerrari:
Duplicate of #9084

PR #9106, opened 2026-05-12
[Performance][310p] use recurrent gdn custom op with aclgraph

Tflowers-0129:

[310P] Fix linear attention graph capture and dtype issues

- Fix graph capture issue for causal-conv1d operator.
- Fix dtype handling issue in recurrent-gated-delta-rule-310.
- Integrate fused operators for the 310P linear attention path.

NA

some UT and local test

- vLLM version: v0.20.2


Tflowers-0129:
我调试不出来了，causal-conv1d一直出现bad-alloc，先关闭吧，重头开始弄

PR #9110, opened 2026-05-12
[Feature] Add KV consumer partial-group caching for hybrid Mamba models

underfituu:

**Background**

In P/D (Prefill/Decode) disaggregated inference, the KV producer (prefill node) computes KV cache and transfers it to the KV consumer (decode node) via a KV connector. The consumer leverages prefix caching to reuse previously received KV blocks and only computes the remaining tokens.

For hybrid attention models (e.g. Jamba, which combines FullAttention layers with Mamba/SSM layers), this flow has a fundamental problem: **Mamba states are NOT transferred via the KV connector**. The producer computes and caches both FullAttention KV and Mamba states, but only FullAttention KV is sent to the consumer. On the consumer side, the Mamba layers have no cached state — they must recompute from scratch.

**Problem**

Without this patch, vLLM's `HybridKVCacheCoordinator` treats all KV cache groups (FullAttention + Mamba) uniformly. This causes three failures on the KV consumer:

1. **Cache hit length is clamped by Mamba**: `find_longest_cache_hit` iterates over all attention groups and converges on a length that satisfies ALL groups. Since Mamba has no cached blocks on the consumer, the hit length drops to 0 — effectively disabling prefix caching entirely.
2. **LCM block alignment is too coarse**: `lcm_block_size` is the LCM of all group block sizes (e.g. `LCM(128, max_model_len)`). This forces cache hit alignment to a huge granularity, wasting most prefix cache hits.
3. **Mamba block-aligned split blocks external tokens**: `_mamba_block_aligned_split` asserts `num_external_computed_tokens == 0`, which fails for KV consumers that receive external tokens from the producer.

**Solution**

Enable partial-group caching on KV consumers: only FullAttention groups participate in prefix caching; Mamba groups are treated as uncached (single-block with `block_size = max_model_len`).

The core flag `enable_kv_consumer_partial_group_caching` is set when:
- `kv_transfer_config is not None` (P/D mode)
- `kv_transfer_config.is_kv_consumer` (this is the decode node)
- `enable_prefix_caching` is True

When this flag is active:

| Component | Change | Why |
|-----------|--------|-----|
| `mamba_block_size` | Set to `max_model_len` | Makes Mamba a single block — no prefix caching hash computed |
| `hash_block_size` | Stays at GCD granularity | Skips the Mamba back-off in `resolve_kv_cache_block_sizes` |
| `lcm_block_size` | Overridden to `hash_block_size` | Reduces alignment so FullAttention cache hits are not wasted |
| `find_longest_cache_hit` | Uses `full_attn_hit_length` only | Ignores Mamba's zero-hit that would clamp the result |
| `need_mamba_block_aligned_split` | Disabled | Consumer doesn't need block-aligned chunking |
| `_mamba_block_aligned_split` assert | Removed | External computed tokens are expected on KV consumers |

All changes are implemented as platform-phase monkey-patches following established vllm-ascend patterns.

**Files Changed**

| File | Type | Description |
|------|------|-------------|
| `patch/platform/patch_kv_consumer_hybrid.py` | New | Core patch: 6 monkey-patches for coordinator, manager, scheduler |
| `patch/platform/patch_kv_cache_utils.py` | Modified | Skip Mamba block-size back-off for KV consumers |
| `patch/platform/patch_mamba_config.py` | Modified | Set `mamba_block_size=max_model_len` for KV consumers |
| `patch/platform/__init__.py` | Modified | Register the new patch module |
| `patch/__init__.py` | Modified | Add patch documentation entry |

No. This change is transparent to users. It automatically activates only when P/D disaggregated inference is configured with a hybrid Mamba model and prefix caching enabled.

- P/D serving with hybrid Mamba model (e.g. Jamba) on Ascend NPU
- Verified KV consumer correctly uses FullAttention-only prefix cache hits
- Verified Mamba layers recompute from scratch on consumer (no stale state)
- Verified non-P/D hybrid Mamba serving remains unaffected (flag stays False)
- Verified non-hybrid P/D serving (pure FullAttention models) remains unaffected
- vLLM version: v0.20.1

### DIFF
--- a/vllm_ascend/patch/platform/patch_kv_consumer_hybrid.py
+++ b/vllm_ascend/patch/platform/patch_kv_consumer_hybrid.py
@@ -1,0 +1,304 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Ascend project
+"""Patch: Enable KV Consumer partial-group caching for hybrid Mamba models.
+
+When running P/D disaggregated inference on hybrid attention models
+(FullAttention + Mamba, e.g. Jamba), the KV consumer side needs special
+handling because Mamba states are *not* transferred via the KV connector.
+Only the FullAttention KV cache should participate in prefix caching on
+the consumer.
+
+This patch monkey-patches four upstream vLLM modules so that, when the
+current instance is a KV consumer with prefix caching enabled:
+
+1. ``HybridKVCacheCoordinator.__init__``: sets the
+   ``enable_kv_consumer_partial_group_caching`` flag and overrides
+   ``lcm_block_size`` to ``hash_block_size``.
+2. ``HybridKVCacheCoordinator.find_longest_cache_hit``: uses only the
+   FullAttention hit length (ignoring Mamba hit results).
+3. ``KVCacheManager.__init__``: threads ``vllm_config`` through to the
+   coordinator factory.
+4. ``get_kv_cache_coordinator``: accepts and forwards ``vllm_config``.
+5. ``Scheduler.__init__``: passes ``vllm_config`` to ``KVCacheManager``
+   and disables ``need_mamba_block_aligned_split`` for KV consumers.
+6. ``Scheduler._mamba_block_aligned_split``: removes the assertion that
+   blocks external computed tokens.
+"""
+
+import vllm.v1.core.kv_cache_coordinator
+import vllm.v1.core.kv_cache_manager
+import vllm.v1.core.sched.scheduler
+from vllm.config import VllmConfig
+from vllm.logger import logger
+from vllm.v1.core.kv_cache_coordinator import (
+    FullAttentionSpec,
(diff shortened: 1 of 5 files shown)


vllm_ascend/patch/platform/patch_kv_consumer_hybrid.py
```diff
@@ -0,0 +1,304 @@
...
+        if is_simple_hybrid:
+            break
+
+    # For KV consumers, the final hit length is determined solely by
+    # FullAttention groups since Mamba states are not transferred.
+    if getattr(self, "enable_kv_consumer_partial_group_caching", False):
+        hit_length = full_attn_hit_length
+
+    # Truncate full attention blocks to final hit_length (if present)
+    spec, group_ids, _ = self.attention_groups[0]
```
nwpu-zxr, line 153: Maybe there is not only one attention group with FullAttentionSpec, it will be better to check the types of all the groups.

PR #9127, opened 2026-05-13
[Feature] adapt copy_and_expand_eagle_input op to 910c and fix shift bug

HF-001:
same to :  https://github.com/vllm-project/vllm-ascend/pull/8838 , adapt the copy_and_expand_eagle_input operator to 910c, which is used in unified parallel spec decode, like Pard、P-eagle etc. And fix shift bug.

Collaborate with @NickJudyHvv


HF-001:
@wangxiyuan hi, this pr is ready, could you help merge it?

HF-001:
@MengqingCao hi, this pr is ready, could you help merge it?

PR #9134, opened 2026-05-13
[BugFix][310P] Fix the 310P RC memory verification logic.

YangShuai52:
Modify the PR to fix the memory verification logic for the 310P RC. In the work_310p.py file, add a method to determine the RC device using the lspci command. Since the RC device shares memory and video memory, it is necessary to modify the method for obtaining free_memory and the calculation of the kv_cache allocation size. Additionally, modify Dockerfile.310p to install pciutils.
RFC #9730 
No
vLLM version: v0.20.2
vLLM main: https://github.com/vllm-project/vllm/commit/0d4d334eaa583b9c09aa4eb7538c22db99fd84b3


vllm_ascend/_310p/worker_310p.py
```diff
@@ -72,6 +93,8 @@ def determine_available_memory(self) -> int:
         ) as profile_result:
             self.model_runner.profile_run()
             free_memory, total_memory = torch.npu.mem_get_info()
+            if _is_rc_device():
+                free_memory = psutil.virtual_memory().available
```
Tflowers-0129, line 99: 这里也可以加一下注释来解释一下i think

YangShuai52: 已添加

PR #9145, opened 2026-05-14
[Feature]Add Support for hybrid model

UnifiedCacheManager:
- **`UCMConnectorV1`** now subclasses **`SupportsHMA`**, so vLLM correctly reports this connector as HMA-capable (`supports_hma()`).
- Forwards **`request_finished_all_groups`**, **`get_finished`**, **`build_connector_worker_meta`**, and **`update_connector_output`** to **`ucm.integration.vllm.ucm_connector.UCMConnector`**, aligning with UCM’s façade API for hybrid / multi-group KV.
- **`UCMConnector`** is constructed with **`kv_cache_config`** (the same instance vLLM passes into the adaptor), so UCM connector selection—including **FAWA / Ascend FAWA** routing via **`can_handle_*_kv_cache_config`**—no longer receives `None`.
**Yes (behavioral, for setups using UCM + `UCMConnector`).**
- **Intended improvement:** Hybrid KV / HMA configurations should now select the intended UCM implementation and receive the correct lifecycle callbacks instead of behaving like a plain non-HMA connector or omitting façade behavior.
- **Compatibility note:** Requires a UCM version whose **`UCMConnector`** exposes the forwarded methods above. Older UCM packages missing those APIs should upgrade together with this change.

- vLLM version: v0.20.1

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ucm_connector.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ucm_connector.py
@@ -4,8 +4,14 @@ from typing import TYPE_CHECKING, Any, Optional
 import torch
 from ucm.integration.vllm.ucm_connector import UCMConnector
 from vllm.config import VllmConfig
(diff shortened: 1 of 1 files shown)


MengqingCao (approved):
LGTM

PR #9158, opened 2026-05-14
[Ascend950][BugFix] Fix eagle3 speculative decode: use use_eagle() to cover eagle/eagle3/mtp methods

gaozihao-shy:
This PR fixes a dimension mismatch issue in eagle3 speculative decoding on Ascend NPUs. The sequence length alignment logic in `AscendMLAImpl` previously only checked for the "mtp" method. By replacing this check with `speculative_config.use_eagle()`, the logic now correctly applies to all eagle-family methods (eagle, eagle3, and mtp). This ensures that `seq_lens_list` is properly padded and `actual_seq_lengths` is correctly calculated for target models using these speculative decoding methods.

No. This is a bug fix for internal attention state handling during speculative decoding on Ascend NPUs.

The patch was tested with eagle3 speculative decoding on Ascend NPU to verify that the dimension mismatch errors are resolved and the model produces correct output.
- vLLM version: v0.20.1


linfeng-yuan (approved):
LGTM

PR #9176, opened 2026-05-15
[CI]Main2main 0515

Potabk:
Upstream PR [vllm-project/vllm#39568](https://github.com/vllm-project/vllm/pull/39568) is a complete rewrite of the routed-experts capture/transport pipeline. It supersedes both:

- The original 0.20.2 design — `RoutedExpertsCapturer.get_instance()` singleton, `save_captured_experts(indices=...)`, shared-memory + `fcntl.flock` cross-process transport.
- The intermediate PR #39917 design — module-level `get_global_experts_capturer()`, `init_routed_experts_capturer_with_shared_cache()`, `issue_routing_d2h_copy()`, `extract_routed_experts_for_current_batch()`. This API existed in main for only a few days and was never in a stable release; it has been **fully removed**.

After the upgrade to vLLM 0515, vllm-ascend faces two API surfaces that are incompatible at the source level:

| Aspect | 0.20.2 | main |
|---|---|---|
| Capturer access | `RoutedExpertsCapturer.get_instance()` (singleton) | `runner.routed_experts_capturer` (per-runner instance, no global) |
| Per-step `clear_buffer` | via singleton | via runner attribute |
| Per-step D2H + ship | `capturer.save_captured_experts(indices=cpu_slot_mapping)` (sync, shm write) | runner-managed pinned `routed_experts_cpu` D2H + `RoutedExpertsLists` on `ModelRunnerOutput.routed_experts` |
| Output channel | shm/flock to scheduler | `ModelRunnerOutput.routed_experts: RoutedExpertsLists` (NamedTuple, msgpack + zmq IPC) |
| `slot_mapping` source | `slot_mapping.cpu().numpy()` saved to `self.cpu_slot_mapping` | private device snapshot `routed_experts_slot_mapping_device`, then pinned `routed_experts_slot_mapping_cpu` |
| Layer hook injection | `select_experts` calls singleton from inside `apply()` | `module.router.set_capture_fn(...)` from `_bind_routed_experts_capturer` |

## Strategy Overview

1. **Keep the 0.20.2 path intact.** It already works end-to-end. All 0.20.2-specific call sites stay byte-identical.
2. **Add a parallel main path** gated by `vllm_version_is("0.20.2") == False`. Reuse upstream `GPUModelRunner.init_routed_experts_capturer()` (inherited) for buffer allocation; override only `_bind_routed_experts_capturer` because Ascend's `select_experts` does not go through upstream `BaseRouter`.
3. **Async scheduling: piggyback on upstream `AsyncGPUModelRunnerOutput`.** vllm-ascend already constructs that wrapper directly, so adding the `routed_experts=` kwarg is enough — the wrapper handles `to_cpu_nonblocking()` on its copy stream and `tolists()` finalization in `get_output()` for free.
4. **No new compat module, no monkey patches.** Branching is inline at each call site; total surface is one new method (`_bind_routed_experts_capturer`) plus three branched call sites in `model_runner_v1.py` and one in `fused_moe.py`.

- vLLM version: v0.20.2
- vLLM main: https://github.com/vllm-project/vllm/commit/ce29c26b31d432b1b4bc028c46bb2c3b07a667d8

### DIFF
--- a/vllm_ascend/ops/fused_moe/fused_moe.py
+++ b/vllm_ascend/ops/fused_moe/fused_moe.py
@@ -172,7 +172,11 @@ class AscendUnquantizedFusedMoEMethod(UnquantizedFusedMoEMethod):
             num_experts=num_logical_experts,
         )
         if layer.vllm_config.model_config is not None and layer.vllm_config.model_config.enable_return_routed_experts:
-            capturer = RoutedExpertsCapturer.get_instance()
+            if vllm_version_is("0.20.2"):
+                # 0.20.2: capturer is a process-wide singleton.
+                capturer = RoutedExpertsCapturer.get_instance()
+            else:
+                capturer = getattr(layer, "_ascend_routed_experts_capturer", None)
             if capturer is not None:
                 capturer.capture(layer_id=layer.layer_id, topk_ids=topk_ids)
 
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -75,6 +75,11 @@ from vllm.v1.outputs import (
     SamplerOutput,
     make_empty_encoder_model_runner_output,
 )
+
+from vllm_ascend.utils import vllm_version_is
+
+if not vllm_version_is("0.20.2"):
+    from vllm.v1.outputs import RoutedExpertsLists
 from vllm.v1.sample.logits_processor import build_logitsprocs
 from vllm.v1.sample.metadata import SamplingMetadata
 from vllm.v1.sample.rejection_sampler import RejectionSampler
@@ -1576,9 +1581,12 @@ class NPUModelRunner(GPUModelRunner):
         intermediate_tensors: IntermediateTensors | None = None,
     ) -> ModelRunnerOutput | IntermediateTensors | None:
         if self.vllm_config.model_config.enable_return_routed_experts:
-            capturer = RoutedExpertsCapturer.get_instance()
-            if capturer is not None:
-                capturer.clear_buffer()
+            if vllm_version_is("0.20.2"):
+                capturer = RoutedExpertsCapturer.get_instance()
+                if capturer is not None:
+                    capturer.clear_buffer()
+            elif self.routed_experts_initialized:
+                self.routed_experts_capturer.clear_buffer()
 
         if self.ascend_config.profiling_chunk_config.need_timing:
             # Check if the scheduler signaled that calibration is complete.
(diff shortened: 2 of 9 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2625,7 +2651,16 @@ def _get_block_table_and_slot_mapping(kv_cache_gid: int):
                     kv_cache_gid,
                 )
             if self.model_config.enable_return_routed_experts and kv_cache_gid == 0:
-                self.cpu_slot_mapping = slot_mapping.cpu().numpy()
+                if vllm_version_is("0.20.2"):
```
gcanlin, line 2654: I think we can add a compatibility for v0.21.0 here. Otherwise it's difficult to realize whether it's appropriate to only v0.20.2 or both them. We haven't upgraded the CI to v0.21.0 though. But it's necessary to do it in advance. Currently, all breaking from vLLM that this PR is fixing is definitely after vLLM v0.21.0, which means that these compatibilities should be applied for v0.21.0 as well. cc @wangxiyuan @MengqingCao @wjunLu

Potabk: I'm not sure what our versioning policy is, but it seems we only need to release a v0.20.2 for rc? Perhaps we could pull a new `releases/v0.20.2` branch and drop v0.20.2 on the main to adapt to v0.21? Because for developers, it's very difficult to achieve full compatibility with versions without CI oversight.

gcanlin: > I'm not sure what our versioning policy is, but it seems we only need to release a v0.20.2 for rc? Perhaps we could pull a new releases/v0.20.2 branch and drop v0.20.2 on the main to adapt to v0.21? Because for developers, it's very difficult to achieve full compatibility with versions without CI oversight.

Yes. Releasing v0.20.2rc1 would be better. (But to be honest, it's hard to push for me) And we shouldn't expect we can totally cover v0.21.0 without CI. Just add a simple compatibility as a workaround to avoid too heavy burden when we upgrade the CI to v0.21.0 in the future.
All compatibility in this PR also works for v0.20.2 and v0.21.0. It wouldn't add more burden for developing from my understanding. Because this PR is pinning the commit that's after v0.21.0.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -75,6 +75,11 @@
     SamplerOutput,
     make_empty_encoder_model_runner_output,
 )
+
+from vllm_ascend.utils import vllm_version_is
+
+if not vllm_version_is("0.20.2"):
```
gcanlin, line 81:
```suggestion
if not vllm_version_is("0.20.2") and not vllm_version_is("0.21.0"):
```

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1576,9 +1581,12 @@ def execute_model(
         intermediate_tensors: IntermediateTensors | None = None,
     ) -> ModelRunnerOutput | IntermediateTensors | None:
         if self.vllm_config.model_config.enable_return_routed_experts:
-            capturer = RoutedExpertsCapturer.get_instance()
-            if capturer is not None:
-                capturer.clear_buffer()
+            if vllm_version_is("0.20.2"):
```
gcanlin, line 1584:
```suggestion
            if vllm_version_is("0.20.2") or vllm_version_is("0.21.0"):
```

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2082,10 +2090,25 @@ def propose_draft_token_ids(sampled_token_ids):
             if self.speculative_config is not None:
                 self.finalize_kv_connector()
 
+        routed_experts_lists = None
         if self.model_config.enable_return_routed_experts:
-            capturer = RoutedExpertsCapturer.get_instance()
-            if capturer is not None:
-                capturer.save_captured_experts(indices=self.cpu_slot_mapping)
+            if vllm_version_is("0.20.2"):
```
gcanlin, line 2095:
```suggestion
            if vllm_version_is("0.20.2") or vllm_version_is("0.21.0"):
```

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2097,6 +2120,9 @@ def propose_draft_token_ids(sampled_token_ids):
             pooler_output=[],
             ec_connector_output=ec_connector_output if self.supports_mm_inputs else None,
             cudagraph_stats=cudagraph_stats,
+            **(
+                {} if vllm_version_is("0.20.2") else {"routed_experts": routed_experts_lists}
```
gcanlin, line 2124:
```suggestion
                {} if vllm_version_is("0.20.2") or vllm_version_is("0.21.0") else {"routed_experts": routed_experts_lists}
```

vllm_ascend/worker/model_runner_v1.py
gcanlin, line 2654:
```suggestion
                if vllm_version_is("0.20.2") or vllm_version_is("0.21.0"):
```

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -172,7 +172,11 @@ def apply(
             num_experts=num_logical_experts,
         )
         if layer.vllm_config.model_config is not None and layer.vllm_config.model_config.enable_return_routed_experts:
-            capturer = RoutedExpertsCapturer.get_instance()
+            if vllm_version_is("0.20.2"):
```
gcanlin, line 175:
```suggestion
            if vllm_version_is("0.20.2") or vllm_version_is("0.21.0"):
```

gcanlin (commented):
Just changing these 6 lines will be enough and it wouldn't break anything for main or v0.20.2.

wangxiyuan:
@gcanlin create a new PR if you need main for 0.21.0

PR #9179, opened 2026-05-15
[Feature] broadcast reset_prefix_cache to all PD nodes in layerwise proxy

underfituu:

Adds a `/reset_prefix_cache` endpoint to the layerwise disaggregated prefill proxy (`load_balance_proxy_layerwise_server_example.py`). When called, the proxy fans out the request concurrently to all prefiller and decoder nodes, collects failures, and returns 500 with the list of failed nodes if any backend call fails.

This mirrors the existing `/reset_prefix_cache` API on individual vLLM servers and makes it usable in a PD-disaggregated deployment where clients only talk to the proxy.

Yes — adds a new POST `/reset_prefix_cache` endpoint on the proxy. Query parameters (e.g. `reset_running_requests`, `reset_external`) are forwarded as-is to each backend.

Manually verified the endpoint structure and fan-out logic. CI will cover the rest.

> AI assistance was used to draft this change. This is not duplicating any existing open PR.
- vLLM version: v0.20.2
- vLLM main: https://github.com/vllm-project/vllm/commit/ce29c26b31d432b1b4bc028c46bb2c3b07a667d8

### DIFF
--- a/examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
+++ b/examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
@@ -566,6 +566,34 @@ async def handle_chat_completions(request: Request):
     return await _handle_completions("/chat/completions", request)
 
 
+@app.post("/reset_prefix_cache")
+async def reset_prefix_cache(request: Request):
+    params = dict(request.query_params)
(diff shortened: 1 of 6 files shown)


underfituu:
Superseded by a cleaner branch based on main with only the reset_prefix_cache change.

PR #9180, opened 2026-05-15
[Main][Feature] Broadcast reset_prefix_cache to all PD nodes in layerwise proxy

underfituu:

Adds a `/reset_prefix_cache` endpoint to the disaggregated prefill proxy (`load_balance_proxy_layerwise_server_example.py`  and `load_balance_proxy_server_example.py`). When called, the proxy fans out the request concurrently to all prefiller and decoder nodes, collects failures, and returns 500 with the list of failed nodes if any backend call fails.

This mirrors the existing `/reset_prefix_cache` API on individual vLLM servers and makes it usable in a PD-disaggregated deployment where clients only talk to the proxy.

Yes — adds a new POST `/reset_prefix_cache` endpoint on the proxy. Query parameters (e.g. `reset_running_requests`, `reset_external`) are forwarded as-is to each backend.

Manually verified the endpoint structure and fan-out logic.

- vLLM version: v0.20.2

### DIFF
--- a/examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
+++ b/examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
@@ -566,6 +566,34 @@ async def handle_chat_completions(request: Request):
     return await _handle_completions("/chat/completions", request)
 
 
+@app.post("/reset_prefix_cache")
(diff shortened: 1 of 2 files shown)


linfeng-yuan (approved):
LGTM

PR #9181, opened 2026-05-15
[Doc] Correcting the list information

herizhen:
deleting DeepSeek v4 data. we remove this in model support list as it is not supported in the branch 0.18.0.

No.

Verified the markdown rendering and ensured the status icons match the intended experimental support level.


herizhen:
Modified according to the review,please merge.
@MengqingCao

MengqingCao (approved):
we remove this in model support list as it is not supported in the branch 0.18.0.

PR #9229, opened 2026-05-18
[Test] Add a case and fix upstream workflow

jiangyunfan1:
This PR adds an upstream case and fix workflow, we need it for testing upstream cases
No
by running the case
- vLLM version: v0.20.2


jiangyunfan1:
https://github.com/vllm-project/vllm-ascend/actions/runs/26010814093/job/76450989102?pr=9229

PR #9241, opened 2026-05-18
[Misc] Remove useless patch

wangxiyuan:
Looks that torch_npu 2.10.0 has fixed the bug in `torch.argsort`. The patch can be dropped now.

test script:
```
import torch
import torch_npu

bool_tensor = torch.tensor([True, False, True, False, True], device='npu')

print('Input:', bool_tensor)

result_unstable = torch.argsort(bool_tensor, stable=False)
print('stable=False:', result_unstable)

result_stable = torch.argsort(bool_tensor, stable=True)
print('stable=True: ', result_stable)
```

result:
```
Input: tensor([ True, False,  True, False,  True], device='npu:0')
stable=False: tensor([1, 3, 0, 2, 4], device='npu:0')
stable=True:  tensor([1, 3, 0, 2, 4], device='npu:0')
```

- vLLM version: v0.20.2


drslark:
LGTM.

PR #9272, opened 2026-05-18
[BugFix][Feature] Fix ACLGraph attn metadata key resolution and add dflash speculative decoding support

Liuchenbing-2026:

This PR contains two bug fixes and one feature for speculative decoding on Ascend NPU:

#### 1. [BugFix] Fix attention metadata key resolution in ACLGraph FULL_DECODE_ONLY mode (`attention_v1.py`)

In `full_graph_fia`, the `layer_name` is now captured from the `AttentionLayer` object and stored as part of `attn_params`. During `update_graph_params`, a new helper `_resolve_attn_key` is used to look up the correct metadata key by matching against the captured layer name (exact match, suffix, or prefix fallback).

Without this fix, models whose attention layer name does not exactly match the positional key (e.g. Qwen3.5 variants) raise a `KeyError` when updating ACLGraph parameters in FULL_DECODE_ONLY mode.

**Root cause**: ACLGraph captures attention parameters by positional key at graph-capture time, but at update time the metadata dict is keyed by layer name. For architectures where these differ, the lookup fails silently or raises an error.

#### 2. [BugFix] Fix speculative tensor slicing in GDN forward pass (`ops/gdn.py`)

Pre-slice `spec_state_indices_tensor`, `num_accepted_tokens`, and `spec_query_start_loc` to `num_spec_decodes` length before passing into `causal_conv1d_update_npu` and `chunk_gated_delta_rule_npu`. Previously the full-length tensors were passed, causing shape mismatches when `num_spec_decodes < tensor.size(0)` in speculative decode steps.

#### 3. [Feature] Add `dflash` method support for speculative decoding (`worker/model_runner_v1.py`)

- Adds `elif method == "dflash"` branch in `_set_up_drafter`, asserting `AscendDflashProposer` and setting `use_aux_hidden_state_outputs = True`.
- Propagates `speculative_config` and `device` to `RejectionSampler` constructor.
- Handles `aux_hidden_states` being either a pre-concatenated `torch.Tensor` or a list of tensors, avoiding an unnecessary `torch.cat` when the drafter already returns a single tensor.

Yes: users can now enable DFlash speculative decoding via `--speculative-config '{"method": "dflash", ...}'` on Ascend NPU. Existing ACLGraph FULL_DECODE_ONLY workflows on Qwen3.5 are also fixed.

- Validated on Ascend 910B2, CANN 8.5.0
- Model: Qwen3-14B / Qwen3.5 series
- Speculative config: dflash method, FULL_DECODE_ONLY ACLGraph mode
- Confirmed no regression on standard (non-speculative) decode path
- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -487,6 +487,15 @@ class AscendAttentionBackendImpl(AttentionImpl):
                 return
             if _EXTRA_CTX.is_draft_model:
                 attn_keys = attn_keys * (len(graph_params.attn_params[num_tokens]) // num_layers)
+            def _resolve_attn_key(metadata: dict, captured_key, fallback_key):
+                if captured_key in metadata:
+                    return captured_key
+                if captured_key is not None:
+                    for candidate in metadata:
+                        if candidate.endswith(captured_key) or captured_key.endswith(candidate):
+                            return candidate
+                return fallback_key
+
             attn_count = 0
             with torch.npu.stream(update_stream):
                 for key, param, handle, event in zip(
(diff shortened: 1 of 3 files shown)


Liuchenbing-2026:
Closing — the ACLGraph attn_metadata key resolution and dflash spec-decode wiring have been superseded by later PRs. Branch has drifted with unresolved conflicts.

PR #9280, opened 2026-05-19
[Ascend950] [Feature]EPLB support W8A8_MXFP and W4A4_MXFP

Eric-dot:
This PR enables EPLB for `W8A8_MXFP` and `W4A4_MXFP` quantized MoE models on Ascend 950.

MXFP quantization requires both quantized expert weights and their corresponding scale tensors for correct dequantization and computation. Previously, the EPLB adaptor did not include the MXFP scale parameters when handling expert parameters, which could lead to incorrect results when EPLB was used together with MXFP quantized experts.

This PR adds the required scale inputs, such as `w13_weight_scale` and `w2_weight_scale`, to the EPLB expert parameter list. As a result, EPLB can correctly remap and manage MXFP expert weights together with their scales, ensuring accuracy correctness for `W8A8_MXFP` and `W4A4_MXFP` scenarios.

- vLLM version: v0.20.2


weijinqian0 (approved):

linfeng-yuan (approved):
LGTM.

PR #9304, opened 2026-05-19
[CI] Solve the problems of slow download speed and UV

tfhddd:
1. Replace triton-ascend source.
2. Add uv.

Speed ​​up PR execution.

Check the installation time of vllm-ascend in the pr.

- vLLM version: v0.20.2


.github/workflows/_e2e_test.yaml
```diff
@@ -53,8 +53,10 @@ on:
         default: ''
 env:
   UV_INDEX_URL: http://cache-service.nginx-pypi-cache.svc.cluster.local/pypi/simple
-  UV_EXTRA_INDEX_URL: "https://mirrors.huaweicloud.com/ascend/repos/pypi https://triton-ascend.osinfra.cn/pypi/simple https://download.pytorch.org/whl/cpu/"
```
MengqingCao, line 56: Don't we need the the resource of torch cpu?

requirements.txt
```diff
@@ -1,4 +1,3 @@
---extra-index-url https://triton-ascend.osinfra.cn/pypi/simple
```
MengqingCao, line 1: maybe we should add `https://docs.github.com/zh/github-cli/github-cli/quickstart` for users to install triton-ascend by this file?

wangxiyuan: ++, for users, how to install triton-ascend 3.2.1 by hand? we should update the install doc as well.

.github/workflows/_e2e_test.yaml
tfhddd: The download link for https://download.pytorch.org/whl/cpu/ has been restored.

requirements.txt
tfhddd: done

PR #9309, opened 2026-05-19
[Doc] Add Qwen3.6-27B and Qwen3.6-35B-A3B deployment guides and supported-m…

mazhixin000:
This PR adds documentation and tutorials for the Qwen3.6-27B and Qwen3.6-35B-A3B models, providing deployment scripts and updating the support matrix.

Documentation-only updates are not considered user-facing changes.

Documentation was reviewed for technical accuracy.


wangxiyuan:
does qwen3.6 is almost the same with qwen3.5? if yes, I don't think we need create a new guide. Just update qwen3.5 tutorials instead. similar with https://docs.vllm.ai/projects/recipes/en/latest/Qwen/Qwen3.5.html

mazhixin000:
> does qwen3.6 is almost the same with qwen3.5? if yes, I don't think we need create a new guide. Just update qwen3.5 tutorials instead. similar with https://docs.vllm.ai/projects/recipes/en/latest/Qwen/Qwen3.5.html

Changed to: Qwen3.5-27B/Qwen3.6-27B
Qwen3.6-35B
Qwen3.5-397B-A17B

PR #9322, opened 2026-05-19
[BugFix] Fix the accuracy issue in DFlash FULL_DESCODE_ONLY mode caused by attn group sorting

chenaoxuan:
In some speculative methods (such as DFlash), the order of attn_keys in the Target model will be disrupted instead of increasing by layer index, so need regular expressions to reorder the attn_keys and stor the results in _ATTN_KEYS_BUFFER.

No.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -67,6 +67,7 @@ from vllm_ascend.worker.kvcomp_utils import KVCompMetaData
 
 # default max value of sliding window size
 SWA_INT_MAX = 2147483647
+_ATTN_KEYS_BUFFER = None
 
 
 @register_backend(AttentionBackendEnum.CUSTOM, "ASCEND")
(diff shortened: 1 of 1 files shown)


yiz-liu (approved):

MengqingCao (approved):
Thanks for your effort! this is triggered by the HMA grouping mechanism, and sorting it before updating the graph params seems reasonable.

vllm_ascend/attention/attention_v1.py
```diff
@@ -475,6 +476,22 @@ def update_graph_params(
                 graph_params = get_graph_params()
                 attn_metadata = forward_context.attn_metadata
                 attn_keys = list(attn_metadata.keys())
+                # In some speculative methods (such as DFlash), the order of attn_keys in the Target model
+                # will be disrupted instead of increasing by layer index, so need regular expressions to
+                # reorder the attn_keys and stor the results in _ATTN_KEYS_BUFFER.
+                attn_keys_length = len(graph_params.attn_params[num_tokens])
+                global _ATTN_KEYS_BUFFER
+                if _ATTN_KEYS_BUFFER is None:
```
keyi-zz, line 484: after dummy_run, the _ATTN_KEYS_BUFFER  will be []. The condition should be: 
     if _ATTN_KEYS_BUFFER is None or not _ATTN_KEYS_BUFFER:

PR #9323, opened 2026-05-19
[KV-Cache][Feature] Enable prefix caching with PCP/DCP

cfq0:

Upstream prefix caching assumes logical KV block size equals kv_cache_spec.block_size. With PCP/DCP enabled, vllm-ascend hashes and allocates full-attention KV using virtual block size (block_size * pcp_world_size * dcp_world_size), which breaks prefix cache hit/miss unless scheduler, hasher, and worker paths agree on the same effective size.

This change registers the ascend_patch_general vLLM general plugin and adds CP-aware patches so prefix caching works under context parallel:

- Hybrid KV cache coordinator: AscendHybridKVCacheCoordinator scales effective block size by pcp * dcp for full-attention cache lookup and passes CP world sizes into find_longest_cache_hit.
- Request block hasher: normalize hash_block_size when EngineCore already passes the scheduler virtual block size (block_size * pcp * dcp).
- Mamba manager / worker utils: CP-aware block indexing for mamba state copy when prefix caching is enabled.
- PCP hybrid attention: treat qwen3_5_moe like qwen3_next.

DCP handling for MambaSpec only (linear / hybrid attention state):

Full-attention KV blocks follow PCP and DCP virtual sizing, but linear attention (Mamba) state is not DCP-sharded on the worker—each rank keeps a full mamba state block. Prefix-cache logic must therefore undo the DCP factor for MambaSpec only:

- AscendHybridKVCacheCoordinator._get_effective_block_size: for MambaSpec, use (block_size * pcp * dcp) // dcp_world_size instead of the full virtual size used for FullAttentionSpec.
- AscendMambaManager: when dcp_world_size > 1, divide internal block_size by dcp_world_size so longest-cache-hit and block walks match per-rank mamba state granularity.

Yes. Users can enable --enable-prefix-caching together with PCP and/or DCP; prefix cache hits and Mamba state copies use consistent virtual block sizing instead of silently missing or mis-aligning blocks.

- Unit tests: tests/ut/patch/general/test_prefix_cache_cp_patches.py (get_hash_size, effective block size for FullAttentionSpec vs MambaSpec, AscendMambaManager DCP scaling, mamba_utils PCP block size, PCPManager hybrid model types).
- Manual NPU testing with PCP/DCP and --enable-prefix-caching is still recommended for end-to-end validation.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/patch/general/patch_hybrid_kv_cache_coordinator.py
+++ b/vllm_ascend/patch/general/patch_hybrid_kv_cache_coordinator.py
@@ -1,0 +1,179 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+# This file is a part of the vllm-ascend project.
+#
+from math import lcm
+
+import vllm.v1.core.kv_cache_coordinator as kv_cache_coordinator
+from vllm.v1.core.kv_cache_coordinator import (
+    BlockHash,
+    BlockHashList,
+    BlockHashListWithBlockSize,
+    FullAttentionSpec,
+    HybridKVCacheCoordinator,
+    KVCacheBlock,
+    KVCacheConfig,
+    KVCacheCoordinator,
+    KVCacheMetricsCollector,
+    KVCacheSpec,
+    SingleTypeKVCacheManager,
+)
+from vllm.v1.kv_cache_interface import MambaSpec
+
+
+class AscendHybridKVCacheCoordinator(HybridKVCacheCoordinator):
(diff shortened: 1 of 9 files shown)


vllm_ascend/patch/general/patch_hybrid_kv_cache_coordinator.py
```diff
@@ -0,0 +1,179 @@
...
+    KVCacheConfig,
+    KVCacheCoordinator,
+    KVCacheMetricsCollector,
+    KVCacheSpec,
+    SingleTypeKVCacheManager,
+)
+from vllm.v1.kv_cache_interface import MambaSpec
+
+
+class AscendHybridKVCacheCoordinator(HybridKVCacheCoordinator):
```
weiguihua2, line 26: The latest code already includes this patch; rebase the code.

PR #9325, opened 2026-05-19
[CI] Support running one UT case on multiple devices

MrZ20:
This PR supports running one UT case on multiple device types in Smart UT.

Previously, one `@npu_test` case could only declare a single `(num_npus, npu_type)` requirement, so a test that should be validated on multiple NPU devices had to be duplicated or manually adjusted. 
With this change, the same test case can stack multiple `@npu_test` decorators and Smart UT will route it to each declared device group.

Changes included:
- Allow `npu_test` to accumulate multiple NPU requirements for the same test wrapper.
- Update Smart UT AST parsing so one test node can be added to multiple runner groups.
- Add documentation and a dummy test case for stacked `@npu_test` usage.

`@npu_test` decorators should be stacked consecutively without other decorators inserted between them.

- vLLM version: v0.20.2

### DIFF
--- a/tests/ut/conftest.py
+++ b/tests/ut/conftest.py
@@ -82,8 +82,6 @@ def npu_test(num_npus: int = 1, npu_type: str | RunnerDeviceType = RunnerDeviceT
         num_npus: Number of NPU devices required (default 1).
         npu_type: The NPU chip type required (default A2).
     """
-    if not isinstance(npu_type, RunnerDeviceType):
-        RunnerDeviceType(npu_type)
 
     def decorator(obj):
         return obj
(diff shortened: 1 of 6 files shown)


tests/ut/conftest.py
```diff
@@ -67,64 +65,27 @@ class RunnerDeviceType(str, Enum):
...
-
-            device_count = torch.npu.device_count()
-            if device_count < num_npus:
-                raise RuntimeError(
-                    f"Insufficient NPUs on this runner: need {num_npus}, have {device_count}. Check runner_label.json."
-                )
-            return func(*args, **kwargs)
-
-        return wrapper
+        RunnerDeviceType(npu_type)
```
Potabk: Why this change?

ZT-AIA (approved):

PR #9339, opened 2026-05-19
[Ops][BugFix] Register DeepSeek V4 custom ops for 910B

maoxx241:

PR #9228 added the DeepSeek V4 custom operators and registered them for 910C/950 builds, but the 910B `CUSTOM_OPS_ARRAY` was not updated. This PR registers the same newly added operators in the 910B ACLNN build list so 910B builds include the DeepSeek V4 custom ops.

No user-facing API change. This only updates the 910B custom operator build list.

- `git diff --check -- csrc/build_aclnn.sh`
- Verified every newly registered 910B operator has a matching directory under `csrc/`

No local NPU build was run from this macOS workspace.
- vLLM version: v0.20.2


csrc/build_aclnn.sh
```diff
@@ -83,19 +83,34 @@ elif [[ "$SOC_VERSION" =~ ^ascend910b ]]; then
     log "catlass include=${ABSOLUTE_CATLASS_PATH}"
 
     CUSTOM_OPS_ARRAY=(
+        "scatter_nd_update_v2"
```
MengqingCao, line 86: could we rename this operator like `scatter_nd_update_with_uncontiguous_cache`? I think it is not intuitive to name it as `xxx_v2`

csrc/build_aclnn.sh
```diff
@@ -83,19 +83,34 @@ elif [[ "$SOC_VERSION" =~ ^ascend910b ]]; then
...
+        "sparse_attn_sharedkv"
+        "sparse_attn_sharedkv_metadata"
+        "hc_pre_sinkhorn"
+        "hc_pre_inv_rms"
+        "hc_post"
+        "inplace_partial_rotary_mul"
+        "rms_norm_dynamic_quant"
+        "dequant_swiglu_quant"
+        "grouped_matmul_swiglu_quant"
+        "grouped_matmul_swiglu_quant_v2"
```
MengqingCao, line 113: ditto, what's the difference between v2 and v1?

maoxx241: `grouped_matmul_swiglu_quant_v2` is used for the W4A8 per-channel quantization fused path and the EPLB scenario. The corresponding `grouped_matmul_swiglu_quant` and `grouped_matmul_swiglu_quant_weight_nz_tensor_list` operators are used for the W8A8 fused path and the EPLB scenario.

csrc/build_aclnn.sh
maoxx241: Thanks for the suggestion. This operator is not limited to the non-contiguous cache case; it can be used in both contiguous and non-contiguous scenarios. Naming it `scatter_nd_update_with_uncontiguous_cache` would make the scope look narrower than the actual supported usage, so I would prefer to keep the existing `scatter_nd_update_v2` name here.

MengqingCao: okay, that make sense.

csrc/build_aclnn.sh
MengqingCao: It seems not a easy work to name it clearly then...

maoxx241: I also plan to align the 910B and 910C paths in follow-up work by gradually normalizing them onto `grouped_matmul_swiglu_quant_v2`.

MengqingCao: okay, that would be better.

MengqingCao (approved):

PR #9363, opened 2026-05-20
[BugFix][Ops] Support mixed precision rotary mul

maoxx241:

Ports the mixed precision `inplace_partial_rotary_mul` changes from cann-recipes-infer PR 456 into the vLLM Ascend operator layout.

- Allows FP16/BF16 `x` with FP32 `cos`/`sin` in the custom op schema and tiling path.
- Adds mixed precision tiling keys and regbase kernels for the custom operator.

This PR intentionally keeps the scope to the operator implementation only. A dedicated E2E test can be added separately.

No public API change. This extends the supported dtype combination for the custom operator.

- `git diff upstream/main --check`
- Confirmed the PR diff only contains `csrc/attention/inplace_partial_rotary_mul` operator changes and no test file additions.

NPU hardware execution was not run locally.

- vLLM version: v0.20.2

### DIFF
--- a/csrc/attention/inplace_partial_rotary_mul/op_host/inplace_partial_rotary_mul_def.cpp
+++ b/csrc/attention/inplace_partial_rotary_mul/op_host/inplace_partial_rotary_mul_def.cpp
@@ -9,7 +9,7 @@
  */
 
 /*!
- * \file rotary_position_embedding.cpp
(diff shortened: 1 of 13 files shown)


linfeng-yuan (approved):
LGTM

PR #9364, opened 2026-05-20
[Ops][BugFix] Fix stale custom ops build cache

### LINKED ISSUE #9360: [Build] Support true incremental custom ops builds
### Motivation

Currently, running `pip install -e .` repeatedly can rebuild the custom ops under `csrc` even when none of the custom-op inputs changed. This is expensive during local development because custom-op packaging invokes CMake, generated scripts, OPC kernel generation, and installer generation.

For editable installs and iterative custom-op development, the build should behave incrementally:

- no relevant input change: skip the custom-op build entirely
- changed custom-op source/config: rebuild the affected generated artifacts and kernels
- unchanged custom-op sources: keep their generated artifacts cached
- generated package/install artifacts should still be refreshed when required

### Current behavior

`csrc/build_aclnn.sh` historically removed `csrc/build`, `csrc/output`, and `csrc/build_out` before invoking `csrc/build.sh`. That guarantees correctness but prevents incremental builds. Even after preserving the build tree, some generated custom-op steps need explicit dependencies and content-stable generated files, otherwise unchanged outputs can still be rewritten or stale outputs can remain.

### Desired behavior

`pip install -e . --no-build-isolation` should not rebuild custom ops when the selected SOC/op inputs are unchanged. If one op source changes, only that op's source copy / dynamic Python copy / OPC kernel generation path should rerun where practical. File additions and deletions under an op should also be detected.

### Proposed direction

A practical implementation can be kept inside the existing shell/CMake flow:

1. Preserve `csrc/build` and `csrc/output` for an explicit incremental build mode.
2. Add an input fingerprint in `build_aclnn.sh` covering selected ops, shared custom-op sources/config/scripts, package inputs, SOC, CANN paths, and relevant compiler env.
3. Skip `build.sh` entirely when the fingerprint matches and the installed vendor tree exists.
4. Make custom-op CMake commands depend on copied-source stamps, dynamic Python outputs, compile scripts, and generated kernel stamps.
5. Write generated JSON/Python/shell files only when content changes.
6. Track generated compile artifacts with manifests so removed variants do not leave stale `.sh` / parameter files behind.

### Validation target

At minimum, this should be validated with a custom-op SOC path such as `ascend310p1`:

- fresh `build_aclnn.sh` custom-op build succeeds
- immediate rerun logs that custom-op inputs are up to date and skips build
- changing one op source triggers OPC generation for that op only
- adding/removing a source under one op triggers that op only
- `pip install -v -e . --no-build-isolation --no-deps` reaches `build_aclnn` and skips custom ops on no-op installs

### Notes

This is mainly a build-system correctness/performance issue, not a unit-testable runtime behavior. A small focused build-script/CMake change with manual build evidence should be sufficient for review.

QwertyJack:

This PR fixes stale-cache correctness issues in the custom-ops incremental build path after #9890 introduced Ninja and preserved `csrc/build` across editable installs.

#9890 is the main performance improvement: it enables Ninja and lets CI restore/reuse `csrc/build`. This PR is a follow-up hardening patch for the cases where reusing that build directory can become incorrect or incomplete.

What #9890 does not fully cover:

- **Existing build directories created by another CMake generator.** After #9890, Ninja is selected by default when available. If a developer already has a `csrc/build` configured with `Unix Makefiles`, reusing it with `-G Ninja` fails with CMake's generator-mismatch error. This PR detects that mismatch and refreshes `csrc/build` automatically.
- **Generated custom-op dependency tracking.** Ninja can only rebuild what CMake declares in the build graph. Several generated custom-op artifacts are driven by copied sources, generated dynamic Python files, compile scripts, and `.done` stamps. This PR wires those inputs into the CMake dependencies so changed custom-op inputs rebuild the affected generated artifacts instead of reusing stale outputs.
- **Stale installer packages in a preserved build directory.** Once `csrc/build` is kept, old `cann-ops-transformer*.run` files can remain beside the newly generated installer. This PR removes stale installers before rebuilding so the install step still sees exactly one candidate.

This PR also refreshes `csrc/build` when the cached CANN package path differs from the current one. In CI, a CANN upgrade normally means a new image and therefore a different cache key, so this is not the primary CI scenario. It is still a useful low-cost guard for local editable installs and non-standard environments where `ASCEND_HOME_PATH`, `ASCEND_OPP_PATH`, or a `latest` symlink can change while the same `csrc/build` directory is reused.

This patch keeps the change focused:

- Refresh `csrc/build` when the cached CMake generator no longer matches the selected generator.
- Refresh `csrc/build` when the cached CANN package path no longer matches the current CANN package path.
- Track source-copy, generated dynamic Python, compile-script, and kernel stamp dependencies so changed custom-op inputs rebuild the affected generated artifacts.
- Keep generated parameter and adapter files content-stable, avoiding unnecessary downstream rebuilds when generated content does not change.
- Remove stale `cann-ops-transformer*.run` installers before rebuilding so installer selection remains unambiguous.
- Canonicalize `ROOT_DIR` before passing the install path to the generated CANN installer.

Fixes #9360.

Yes. Re-running `pip install -e .` continues to use the incremental custom-ops build path from #9890, but stale CMake caches are now discarded automatically when the CANN package path or CMake generator changes. Changed generated custom-op inputs are also tracked more accurately by CMake/Ninja.

No unit tests were added because this changes custom-ops build scripts and generated build dependencies rather than Python runtime logic.

Manual checks:

```bash
git diff --check upstream/main...HEAD
bash -n csrc/build.sh csrc/build_aclnn.sh
python3 -m py_compile \
  csrc/cmake/scripts/util/ascendc_bin_param_build.py \
  csrc/cmake/scripts/util/ascendc_impl_build.py
ruff check \
  csrc/cmake/scripts/util/ascendc_bin_param_build.py \
  csrc/cmake/scripts/util/ascendc_impl_build.py
```

Manual build validation on 310P custom ops:

```bash
SOC_VERSION=ascend310p1 MAX_JOBS=8 bash csrc/build_aclnn.sh . ascend310p1
```

This completed successfully, generated exactly one `cann-ops-transformer*.run` installer, and installed it under `vllm_ascend/_cann_ops_custom`.

The validation started from an existing `csrc/build/CMakeCache.txt` using `CMAKE_GENERATOR=Unix Makefiles` while Ninja was available. The build printed `Info: CMake generator changed, refresh CMake cache.`, reconfigured with Ninja, and completed successfully. The local build environment emitted many clock-skew warnings, but the command completed successfully.

- vLLM version: v0.20.2

### DIFF
--- a/csrc/build.sh
+++ b/csrc/build.sh
@@ -386,6 +386,34 @@ function clean_third_party()
     fi
 }
 
+function refresh_cmake_cache()
+{
+    local current_cann_path
+    local cached_cann_path
+    local cached_generator
+    local current_generator
+    current_cann_path=$(realpath -m "${ASCEND_CANN_PACKAGE_PATH}")
+    current_generator="${CMAKE_GENERATOR:-Unix Makefiles}"
+    if [ "${#CMAKE_GENERATOR_ARGS[@]}" -ge 2 ]; then
+        current_generator="${CMAKE_GENERATOR_ARGS[1]}"
+    fi
+
+    if [[ -f "${BUILD_DIR}/CMakeCache.txt" ]]; then
+        cached_cann_path=$(sed -n 's/^CUSTOM_ASCEND_CANN_PACKAGE_PATH:.*=//p' "${BUILD_DIR}/CMakeCache.txt")
+        if [[ -n "${cached_cann_path}" && "$(realpath -m "${cached_cann_path}")" != "${current_cann_path}" ]]; then
+            log "Info: CANN package path changed, refresh CMake cache."
+            rm -rf "${BUILD_DIR}"
+            return
+        fi
+
+        cached_generator=$(sed -n 's/^CMAKE_GENERATOR:INTERNAL=//p' "${BUILD_DIR}/CMakeCache.txt")
+        if [[ -n "${cached_generator}" && "${cached_generator}" != "${current_generator}" ]]; then
+            log "Info: CMake generator changed, refresh CMake cache."
+            rm -rf "${BUILD_DIR}"
+        fi
+    fi
+}
+
 function cmake_config()
 {
     local extra_option="$1"
@@ -1416,6 +1444,7 @@ else
     log "Info: ninja is not found; use default CMake generator"
 fi
 
+refresh_cmake_cache
 clean_build_out
 clean_output
 
--- a/csrc/build_aclnn.sh
+++ b/csrc/build_aclnn.sh
@@ -3,6 +3,7 @@
 ROOT_DIR=$1
 SOC_VERSION=$2
 : "${ROOT_DIR:?ROOT_DIR is not set}"
+ROOT_DIR=$(cd "${ROOT_DIR}" && pwd -P) || exit 1
 
 log() {
     echo "[build_aclnn] $*"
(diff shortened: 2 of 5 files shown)


QwertyJack:
Updated the patch to keep the incremental build logic simpler:

- Removed the outer input-fingerprint/skip logic from `build_aclnn.sh`; it now always delegates incrementality to CMake via `build.sh --incremental`.
- Removed the generated-file manifest cleanup code and the extra CMake manifest dependency.
- Kept the CMake dependencies needed for source-copy, generated dynamic Python, compile scripts, and kernel stamps.
- Removed stale `cann-ops-transformer*.run` packages before rebuilding so the installer selection remains unambiguous.
- Canonicalized `ROOT_DIR` in `build_aclnn.sh` so relative invocation still passes an absolute install path to the CANN installer.

Re-ran static checks and the 310P `build_aclnn.sh . ascend310p1` validation successfully.

PR #9391, opened 2026-05-20
[Ascend950] [Feature] Enable MXFP4 flatquant and support for row parallelism

lenghuixing0330:

- This PR implements the `AscendW4A4MXFP4FlatQuantDynamicLinearMethod` to support W4A4 MXFP4 dynamic linear quantization on Ascend A5 devices. 

- It introduces Kronecker quantization. To support flat quantization under row parallelism, the input size of the activation layer is split, and the shape of the left rotation matrix cannot be determined. Therefore, the minimum left rotation matrix `block_unit` is obtained based on the `max_supported_tp` parameter in the weight file. In the row parallelism scenario, the left rotation matrix is loaded in diagonal blocks based on the `tp_size.`

  The specific method is to add the logic for determining the layer type to the basic `get_pertensor_param` method. For row parallelism, the corresponding blocks are loaded to different ranks in the `process_weights_after_loading` method.

- `Precision` and `Performance` Results

  For Qwen3-32B, apply FlatQuant MXFP4 quantization to MLP layers and keep MXFP8 quantization for attention layers.
  The precision achieves `96.3` on GSM8K and `80` on the AIME2024 dataset.
  In single-card inference with 3.5K input tokens, 1.5K output tokens and 50ms TPOT latency, FlatQuant MXFP4 yields `1.2` times higher output throughput than MXFP8 quantization.

It adds the `W4A4_MXFP4_FLATQUANT` quantization scheme

- vLLM version: v0.20.2


MengqingCao:
cloud you offer more precision and performance results?

linfeng-yuan (commented):
LGTM

linfeng-yuan (approved):
LGTM

PR #9394, opened 2026-05-21
[Feature][Model] Enable W4A8 per-channel GMM Swiglu quant

maoxx241:

This PR enables the custom `grouped_matmul_swiglu_quant_v2` path for the W4A8 per-channel MoE scenario.

- Carries `is_per_channel_weight` through the fused MoE runtime metadata.
- Uses `grouped_matmul_swiglu_quant_v2` only when the path is W4A8 + per-channel weight + custom ops enabled.
- Keeps grouped W4A8 and W8A8 on the existing grouped-matmul paths.
- Uses count-style MC2 `expert_token_nums` only for the W4A8 per-channel GMM path.
- Adjusts W4A8 per-channel `w13_weight_scale` to the `[E, N]` / `[N]` shape expected by the op.

This intentionally does not include the storage-shape helper torch bindings from the source commit.

No. This adds an internal W4A8 MoE execution path and does not expose a user-facing API or interface change.

- `python -m py_compile vllm_ascend/ops/fused_moe/moe_stage_params.py vllm_ascend/ops/fused_moe/moe_runtime_args.py vllm_ascend/ops/fused_moe/token_dispatcher.py vllm_ascend/ops/fused_moe/moe_mlp.py vllm_ascend/quantization/methods/w4a8.py tests/ut/ops/test_moe_mlp.py tests/ut/ops/test_token_dispatcher.py tests/ut/quantization/methods/test_w4a8.py`
- `uvx ruff check ...`
- `uvx ruff format --check ...`

`pytest` was not runnable in this local worktree because the Python environment is missing project runtime dependencies such as `numpy`, `torch`, `torch_npu`, and `vllm`.
- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/ops/fused_moe/moe_mlp.py
+++ b/vllm_ascend/ops/fused_moe/moe_mlp.py
@@ -102,6 +102,7 @@ def quant_apply_mlp(
     per_token_scale_type: torch.dtype | None = None,
     use_bf16: bool = True,
     swiglu_limit: int = 0,
+    use_w4a8_per_channel_gmm_swiglu: bool = False,
 ) -> torch.Tensor:
     input_hidden_dtype = hidden_states.dtype
     use_gmm_swiglu_quant_fusion = use_mxfp_quant or (fusion and not dynamic_eplb)
@@ -250,7 +251,19 @@ def quant_apply_mlp(
             # TODO w4a8 scene: dynamic acquisition of dtype in the future
             _output_dtype = torch.bfloat16
 
-        if _custom_gmm_swiglu_enabled(fusion, dynamic_eplb) and not use_mxfp_quant:
+        if use_w4a8_per_channel_gmm_swiglu and enable_custom_op():
+            hidden_states, swiglu_out_scale = torch.ops._C_ascend.grouped_matmul_swiglu_quant_v2(
+                x=hidden_states,
+                weight=w1,
+                weight_scale=w1_scale if isinstance(w1_scale, list) else [w1_scale],
+                x_scale=pertoken_scale,
+                group_list=group_list,
+                weight_assist_matrix=bias1,
+                dequant_mode=0,
+                group_list_type=group_list_type,
+                swiglu_limit=swiglu_limit,
+            )
+        elif _custom_gmm_swiglu_enabled(fusion, dynamic_eplb) and not use_mxfp_quant:
             # gmm1: gate_up_proj & act_fn: swiglu
             hidden_states, swiglu_out_scale, _ = torch.ops._C_ascend.grouped_matmul_swiglu_quant_weight_nz_tensor_list(
                 x=hidden_states,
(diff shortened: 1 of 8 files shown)


vllm_ascend/ops/fused_moe/moe_mlp.py
```diff
@@ -250,7 +251,19 @@ def quant_apply_mlp(
             # TODO w4a8 scene: dynamic acquisition of dtype in the future
             _output_dtype = torch.bfloat16
 
-        if _custom_gmm_swiglu_enabled(fusion, dynamic_eplb) and not use_mxfp_quant:
```
MengqingCao, line 253: Does `grouped_matmul_swiglu_quant_v2` also support enabling `use_mxfp_quant`?

vllm_ascend/quantization/methods/w4a8.py
```diff
@@ -481,6 +482,12 @@ def pack_to_int32(self, weight: torch.Tensor):
                 weight.to(torch.float32), torch.tensor([1.0]).npu(), None, torch.quint4x2, -1, False
             )
 
+    @staticmethod
+    def squeeze_per_channel_weight_scale(scale: torch.Tensor) -> torch.Tensor:
```
MengqingCao:
```suggestion
    def maybe_squeeze_per_channel_weight_scale(scale: torch.Tensor) -> torch.Tensor:
```

vllm_ascend/quantization/methods/w4a8.py
```diff
@@ -524,6 +531,8 @@ def update_bias_compressed_tensors(weight: torch.Tensor, scale: torch.Tensor, st
 
         layer.w13_weight_scale.data = process_scale_compressed_tensors(layer.w13_weight_scale.data)
```
MengqingCao, line 534: it seems we could move the `squeeze_per_channel_weight_scale` into `process_scale_compressed_tensors`

vllm_ascend/quantization/methods/w4a8.py
MengqingCao, line 556: Similar question

vllm_ascend/ops/fused_moe/moe_mlp.py
maoxx241: The current vLLM Ascend version of `grouped_matmul_swiglu_quant_v2` does not support enabling `use_mxfp_quant` yet. Compared with the CANN version, this PR only wires the W4A8 per-channel support path. W8A8 and MXFP8 support will be added incrementally after validation.

vllm_ascend/quantization/methods/w4a8.py
maoxx241: Updated in the latest push. I renamed it to `maybe_squeeze_per_channel_weight_scale` because the helper may return the original tensor when there is no singleton channel dimension to squeeze.

vllm_ascend/quantization/methods/w4a8.py
maoxx241: Done in the latest push. The compressed tensors path now applies the per-channel squeeze inside `process_scale_compressed_tensors`, so both scale processing and layout adjustment stay together in that path.

vllm_ascend/quantization/methods/w4a8.py
maoxx241: For the modelslim path, I kept the squeeze after `process_scale` because `process_scale` is shared by the non-compressed scale flow, while the per-channel squeeze is only needed after that processing step. The helper has also been renamed to `maybe_squeeze_per_channel_weight_scale`.

MengqingCao (approved):
LGTM now, thx!

zzzzwwjj (approved):

PR #9396, opened 2026-05-21
[Feature][Model] Switch DeepSeekV4 hc_pre to fused op

maoxx241:

This PR switches DeepSeekV4 `hc_pre` from the composite small-op path to the `aclnnHcPre`-backed `npu_hc_pre_v2` interface.

It also aligns the runtime `hc_pre` torch binding contract with the CANN recipe binding while leaving the Meta implementation as shape inference only, so torch compile does not trip over runtime-only checks:

- `x` must be 3D or 4D BF16
- `hc_mult` / `hc` must be 4
- `d` must be 4096 or 7168
- `hc_fn` must be `[24, hc * d]`
- `hc_scale` must be `[3]`
- `hc_base` must be `[24]`
- non-`x` tensors must be FP32

For Ascend950, `npu_hc_pre_v2` follows the CANN recipe's batch filter and falls back to the composite path when `bs > 512` and `bs` is not aligned to 8192. Other SoCs continue to use the fused path.

Reference: https://gitcode.com/cann/cann-recipes-infer/blob/master/ops/ascendc/torch_ops_extension/custom_ops/csrc/npu_hc_pre.cpp

No public API change. DeepSeekV4 internally uses the fused `hc_pre` interface when supported, with the existing composite implementation kept as fallback.

- `PATH="$HOME/Library/Python/3.9/bin:$PATH" bash format.sh ci`
- `python3 -m py_compile vllm_ascend/models/deepseek_v4.py tests/e2e/nightly/single_node/ops/singlecard_ops/test_npu_hc_pre.py`
- `git diff --check`

Added `tests/e2e/nightly/single_node/ops/singlecard_ops/test_npu_hc_pre.py` to compare `npu_hc_pre` and `npu_hc_pre_v2` on 3D and 4D BF16 inputs. Local NPU runtime execution was not run from this macOS workspace.

- vLLM version: v0.20.2

### DIFF
--- a/csrc/torch_binding.cpp
+++ b/csrc/torch_binding.cpp
@@ -1479,6 +1479,14 @@ at::Tensor npu_hc_post_npu(
     return out;
 }
 
+constexpr int64_t HC_PRE_HC_LIMIT = 4;
+constexpr int64_t HC_PRE_D_LIMIT = 4096;
+constexpr int64_t HC_PRE_D_LIMIT_EXTEND = 7168;
+constexpr int64_t HC_PRE_MIX_HC_LIMIT = 24;
+constexpr int64_t HC_PRE_FUSION_BASE_BS = 8192;
+constexpr int64_t HC_PRE_FUSION_SPLIT_K_MAX_BS = 512;
+constexpr const char* ASCEND_950_PREFIX = "Ascend950";
+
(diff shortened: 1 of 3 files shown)


MengqingCao (approved):
LGTM, thx!

PR #9437, opened 2026-05-22
[Feature] Support DSA PCP/DCP with FlashComm

xiaocongtou6:

This PR enables DSA models to use PCP/DCP together with SP/FlashComm in SFA CP.
Previously, when PCP/DCP was enabled, the sequence could not be split in the TP domain through FlashComm. This PR updates `AscendSFACPImpl.forward()` to use FlashComm for sequence partitioning in the TP domain when `pcp_size` / `dcp_size` is greater than 1, and performs an all-gather in the TP domain to restore the original sequence length before returning.
This PR also updates `AscendSFACPMetadataBuilder.build()` to pre-build the reordered slot mapping, so that KV reordering does not need to be performed during computation.
With these changes, DSA models can now enable PCP and SP/FlashComm at the same time, reducing redundant computation and improving the SFA CP execution path.

Yes. DSA models now support enabling PCP/DCP together with SP/FlashComm.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/ops/linear_op.py
+++ b/vllm_ascend/ops/linear_op.py
@@ -61,6 +61,7 @@ from vllm.model_executor.models.utils import extract_layer_index
 
 from vllm_ascend.ascend_config import get_ascend_config
 from vllm_ascend.ascend_forward_context import _EXTRA_CTX
+from vllm_ascend.attention.utils import enable_cp
 from vllm_ascend.distributed.parallel_state import (
     get_flashcomm2_odp_group,
     get_flashcomm2_otp_group,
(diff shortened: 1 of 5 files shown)


vllm_ascend/ops/linear_op.py
```diff
@@ -681,7 +682,6 @@ def _get_row_parallel_op(
             "out_proj",  # attn output linear of Qwen3 Next
             "down_proj",  # second MLP of most LLMs
             "attention.dense",  # attn output linear of Bailing
-            "wo_b",  # attn output linear of v4
```
hwhaokun, line 684: Why delete this?

xiaocongtou6: Accidentally deleted, already restored

PR #9438, opened 2026-05-22
[Ops][BugFix] Disable SwiGLU clamp by default

maoxx241:

This PR changes SwiGLU clamp/limit defaults from sentinel large values to 0 across the Python config path, torch binding schemas, op definitions, and tiling defaults. Kernel clamp vector ops now run only when the received clamp/limit value is positive, so the default value disables clamp instead of doing no-op-ish clamp work.

For grouped_matmul_swiglu_quant_weight_nz_tensor_list, the existing kernel skip guard is preserved and this PR only changes the default/validation side. Covered paths include dequant_swiglu_quant, grouped_matmul_swiglu_quant, grouped_matmul_swiglu_quant_v2, and grouped_matmul_swiglu_quant_weight_nz_tensor_list.

No intentional user-facing API change. Explicit positive swiglu_limit/clamp_limit values keep the previous clamp behavior. Omitting the value now means clamp is disabled with 0.

- bash -n csrc/build_aclnn.sh
- git diff HEAD~1..HEAD --check
- git diff upstream/main..HEAD -- csrc/build_aclnn.sh vllm_ascend/models/deepseek_v4.py
- git diff upstream/main..HEAD -- csrc/torch_binding.cpp csrc/gmm csrc/moe | rg -n '^\+.*isfinite|^\+.*finite|^\+\#include <cmath>|^\+.*weinachuan'
- rg -n 'weinachuan' csrc/gmm csrc/moe csrc/torch_binding.cpp
- python3 -m py_compile vllm_ascend/ops/fused_moe/fused_moe.py
- python3 -m py_compile vllm_ascend/models/deepseek_v4.py
- Not run locally: NPU build/runtime validation.

- vLLM version: v0.20.2

### DIFF
--- a/csrc/gmm/grouped_matmul_swiglu_quant/op_host/grouped_matmul_swiglu_quant_def.cpp
+++ b/csrc/gmm/grouped_matmul_swiglu_quant/op_host/grouped_matmul_swiglu_quant_def.cpp
@@ -54,7 +54,7 @@ public:
             .Format({ge::FORMAT_ND, ge::FORMAT_ND, ge::FORMAT_ND, ge::FORMAT_ND, ge::FORMAT_ND});
         this->Attr("is_enable_weight_assistance_matrix").AttrType(OPTIONAL).Bool(true);
         this->Attr("dequant_mode").AttrType(OPTIONAL).Int(0);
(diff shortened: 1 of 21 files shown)


linfeng-yuan (approved):
LGTM

PR #9455, opened 2026-05-22
[Test]Fix upstream workflow and add a case

jiangyunfan1:
This PR fix some problems on upstream test workflow and add a case, we need to test it regularly
No
by running the test
- vLLM version: v0.20.2


jiangyunfan1:
https://github.com/vllm-project/vllm-ascend/actions/runs/26386997764/job/77668104529?pr=9455

ZT-AIA (approved):

PR #9468, opened 2026-05-22
[Feature] Support layerwise KV cache events

lizy124:
This PR implements layerwise KV cache events support for distributed inference, which provides precise synchronization and event tracking at the final layer of model processing.

Key Features:

1. Layerwise Event Generation :
   
   - In layerwise mode ( use_layerwise=true ), KV cache events are generated only at the final layer
   - Previous layers only record information, final layer generates a cumulative event containing all layers' block information
   - This ensures precise synchronization in distributed inference scenarios
2. Thread Safety Improvements :
   
   - Added layerwise_event_lock to protect layerwise_event_starts data structure
   - Added done_task_lock to protect stored_requests during concurrent access
   - Fixed race conditions in multi-threaded environments
3. Metadata Expansion :
   
   - Extended LayerMultiBlockReqMeta to include:
     - token_ids : Token ID list for granular tracking
     - original_block_size : Original block size information
     - block_hashes : Block hash list
     - kv_cache_group_id : KV cache group identifier
4. Request Lifecycle Management :
   
   - Fixed dec_stored_request and set_finished_request to only execute on final layer
   - Proper cleanup of layerwise_event_starts in all code paths
   - Memory leak prevention
1. New Configuration Option :
   
   - use_layerwise : Enable layerwise KV cache event mode (default: false)
   - When enabled, events are generated only at the final layer
2. Behavior Change :
   
   - Before : KV events generated for each layer independently
   - After : With use_layerwise=true , one cumulative event generated at final layer
3. Backward Compatibility :
   
   - Default behavior unchanged ( use_layerwise=false )
   - Existing configurations continue to work as before
   - No breaking changes to API or configuration
Unit Tests :

- Modified tests/ut/distributed/ascend_store/test_kv_transfer.py
- Added test for update_and_get_kv_events with proper BlockStored instantiation
- Verified event generation and retrieval logic

- vLLM version: v0.20.2

### DIFF
--- a/tests/ut/distributed/ascend_store/test_kv_transfer.py
+++ b/tests/ut/distributed/ascend_store/test_kv_transfer.py
@@ -371,6 +371,7 @@ class TestKVCacheStoreLayerSendingThread(unittest.TestCase):
     def test_handle_request_puts_missing(self):
         t, store = self._make_thread([1, 0])
         req = self._make_layer_req(layer_id=0)
+        t.add_stored_request(req.req_id)
         t.request_queue.put(req)
         t._handle_request(req)
         self.assertEqual(len(store.put_calls), 1)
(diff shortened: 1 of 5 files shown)


ZT-AIA (approved):

MengqingCao:
plz add your pr message

MengqingCao (approved):
lgtm now, let's merge this as the CI failure is not related to this pr.

PR #9469, opened 2026-05-22
[CI] Resolves the issue of occasional 503 errors when downloading packages

tfhddd:
1. `pip install` scans all specified indexes to check if the package to be downloaded exists. When there are many CIs (Content Query Objects), the query frequency is high.

2. frequent access to the upstream PyTorch repository will result in a 503 error.

PyTorch's repository is not placed in public access repositories; the PyTorch repository must be specified when downloading PyTorch-cpu.

- vLLM version: v0.20.2

### DIFF
--- a/.github/workflows/_e2e_test.yaml
+++ b/.github/workflows/_e2e_test.yaml
@@ -53,7 +53,7 @@ on:
         default: ''
 env:
   UV_INDEX_URL: http://cache-service.nginx-pypi-cache.svc.cluster.local/pypi/simple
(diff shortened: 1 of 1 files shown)


.github/workflows/_e2e_test.yaml
```diff
@@ -140,7 +140,10 @@ jobs:
           MAX_JOBS: "23"
         run: |
           pip install uc-manager
+          # Use CPU version package
+          uv pip install torch==2.10.0 torchvision==0.25.0 torchaudio==2.10.0 --extra-index-url  http://cache-service.nginx-pypi-cache.svc.cluster.local/whl/cpu/
```
MengqingCao, line 146: plz add comment to describe the reason of setting this source separately

tfhddd: done

MengqingCao (approved):

PR #9473, opened 2026-05-22
[BugFix] Resolve multiple DeepSeek V4 PP issues

jianzs:

**Compared with DP2TP16, PP4TP8 delivers nearly 2x higher prefill throughput.**

Fix multiple issues that caused PP to fail or degrade output on DeepSeek V4:

1. make_empty_intermediate_tensors created upstream 2D tensors and carried an unused residual tensor across PP boundaries. Fix: use a custom factory that only allocates 3D hidden_states with shape (batch, hc_mult, hidden_size).

2. hidden_states.unsqueeze(1).repeat(...) ran on every PP rank, so non-first ranks expanded already-expanded activations to the wrong shape. Fix: only expand hidden_states on the first PP rank.

3. hc_head ran before every PP boundary, compressing stage outputs and then forcing later stages to reconstruct from degraded activations. Fix: pass full hidden_states between PP stages and only run hc_head on the last PP rank.

4. sink parameter loading could raise KeyError on ranks where the sink parameter belongs to a PPMissingLayer. Fix: guard sink loading with a params_dict membership check.

- vLLM version: v0.20.2


linfeng-yuan (approved):
LGTM

PR #9476, opened 2026-05-22
[BugFix][Attention] Fix DSA v1 W8A8 dynamic conflict

evan-ai-arg:

1. Share one `hidden_states` dynamic quantization between `wq_a` and `wkv` when both layers use W8A8 dynamic quantization, and reuse `(hs_int8, hs_pertoken_scale)` for the two `npu_quant_matmul` calls.
2. Add `_is_w8a8_dynamic(linear)` to centralize W8A8 dynamic detection and replace duplicated quantization-method checks.
3. Move decode `wq_a(hidden_states)` computation before the `wq_b` branch and pass `qr_pertoken_scale` to the indexer path so it can reuse the existing quantized `qr`.

No.

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/attention/dsa_v1.py
+++ b/vllm_ascend/attention/dsa_v1.py
@@ -108,6 +108,15 @@ def hadamard_scale(out: torch.Tensor, x_shape: tuple[int, ...], dim: int, scale:
     return out[..., :dim].reshape(*x_shape)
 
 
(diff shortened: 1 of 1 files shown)


linfeng-yuan (approved):
LGTM. As our discussion, please make full verification of this PR and refactor it to decouple quantization and attention backend implementation in one week.

PR #9489, opened 2026-05-23
[Performance][310p] Discard native RmsNormGated ops on 310P

Tflowers-0129:
Remove the native RMSNormGated implementation and replace it with a decomposed implementation using RMSNorm, Mul, and activation operators.
NA
UT and local test
- vLLM version: v0.20.2


Tflowers-0129:
trace RFC: #9711

PR #9509, opened 2026-05-25
[CI] fix v2 e2e case

shikang-hangzhou:
Fix the bug causing failures in V2 E2E test cases due to code changes, and restore the V2 test cases.

- vLLM version: v0.20.2


linfeng-yuan (approved):
LGTM

PR #8308, opened 2026-04-15
[Performance]Reduce sampling

hzx55906:
What this PR does / why we need it?
This PR introduces a significant optimization to the sampling process in a distributed (Tensor Parallel) environment, named "reduce sampling". The core goal is to minimize communication overhead during token selection, thereby improving the performance of vLLM Ascend in distributed deployment scenarios.

In traditional distributed sampling, full vocabulary gathering across tensor parallel ranks leads to high communication costs, especially in large-scale parallel settings. This optimization addresses the issue by designing specialized sampling logic that intelligently aggregates and processes vocabulary information, reducing unnecessary data transmission while ensuring sampling accuracy.

Does this PR introduce any user-facing change?

No breaking user-facing changes. A new configuration option enable_reduce_sample is added to AscendConfig, which users can optionally enable to activate the optimized distributed sampling scheme. Existing sampling behaviors remain unchanged by default.

How was this patch tested?

Tested with vLLM version based on v0.20.1. 
- Validation of distributed greedy sampling logic, ensuring correct global argmax selection across tensor parallel ranks.

- Verification of distributed top-k/top-p sampling, confirming accurate candidate aggregation and filtering.

- Functional testing of speculative decoding rejection sampling with the new compressed vocabulary mode.

- End-to-end testing of the integrated sampling pipeline in ModelRunnerV1, covering both regular and speculative decoding workflows.

Summary of Changes

- Optimized Distributed Sampling Scheme: Added enable_reduce_sample in AscendConfig to toggle the optimized sampling mode for Tensor Parallelism, reducing communication overhead.

- Distributed Greedy Sampling: Implemented a new greedy_sample function that gathers local maximum logits and their global indices across ranks to determine the global argmax, avoiding full vocabulary gathering.

- Distributed Top-K/Top-P Sampling: Enhanced top-k/top-p logic to perform local top-k selection first, then gather candidates across ranks and apply top-p filtering on the combined global candidates.

- Updated Rejection Sampling: Modified core rejection sampling functions (rejection_sample, rejection_random_sample_kernel, etc.) to support compressed vocabulary mode, working with reduced candidate tokens and global indices for efficiency.

- Model-Specific Logits Handling: Patched Eagle3LlamaForCausalLM.compute_logits to return only selected draft token IDs (and bias) when reduce sampling is enabled, optimizing draft generation.

[...]

PR #9545, opened 2026-05-25
[BugFix] fix reduce_sampling

hzx55906:

Fix the issue in reduce_sampling where enabling speculative sampling causes an error with a single curl request.
https://github.com/vllm-project/vllm-ascend/pull/8308

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2287,11 +2287,7 @@ class NPUModelRunner(GPUModelRunner):
         if spec_decode_metadata is None:
             if lmhead_tp_enable() and logits is not None:
                 logits = logits[: self.input_batch.num_reqs]
-            if (
-                self.input_batch.sampling_metadata.top_k is not None
-                and self.input_batch.top_k_cpu is not None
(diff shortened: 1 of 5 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2287,7 +2287,11 @@ def _sample(self, logits, spec_decode_metadata):
         if spec_decode_metadata is None:
             if lmhead_tp_enable() and logits is not None:
                 logits = logits[: self.input_batch.num_reqs]
-            if self.input_batch.top_k_cpu is not None and get_ascend_config().enable_reduce_sample:
+            if (
+                self.input_batch.sampling_metadata.top_k is not None
+                and self.input_batch.top_k_cpu is not None
+                and get_ascend_config().enable_reduce_sample
+            ):
```
MengqingCao: if the `self.input_batch.top_k_cpu` is always not None, let's just remove this.
```suggestion
            if (
                self.input_batch.sampling_metadata.top_k is not None
                and get_ascend_config().enable_reduce_sample
            ):
```

PR #9549, opened 2026-05-26
[Attention][BugFix] Fix scheduler dead lock for long requests by adding admission cap to KV cache manager

li-lizhe:
This PR fixes a scheduler deadlock where long requests (e.g., >32k tokens) would hang indefinitely in the WAITING queue. The root cause was the missing admission cap for  and  in the Ascend-specific KV cache manager, which caused block allocations to exceed available free blocks. This fix aligns the implementation with upstream vLLM by correctly computing and applying the  cap.

No.

- End-to-end NPU hardware testing with sequences up to 32k tokens (DeepSeek-V4 Flash).
- Verified resolution of the scheduler deadlock without regressions.

Signed-off-by: lizhe.li <147392333@qq.com>

- vLLM version: v0.20.2


yangqinghao-cmss:
这个问题还有吗

PR #9560, opened 2026-05-26
[Feature][Attention] Support Model GLM4.7-Flash

dsxsteven:

  This PR adds **support for model GLM-4.7-Flash**.

  Background: Ascend NPU fused attention ops (npu_fused_infer_attention_score / npu_fused_infer_attention_score_v2) require the number of query heads to be a power of 2. When num_heads is not (e.g. 20), the op will fail or produce incorrect results.

  Changes in vllm_ascend/attention/**mla_v1**.py:

  1. Head padding computation in **__init__**:
    - Compute num_heads_padded as the next power of 2: 1 << (num_heads - 1).bit_length()
    - Compute head_padding = num_heads_padded - num_heads
  2. Prefill path (**_forward_prefill** & **_compute_prefill_context**):
    - The fia operator has specific requirements for the shapes of qk_nope_head_dim and qj_rope_head_dim, v_head_dim,  which currently only affects glm4.7-flash now, so we we use head_padding > 0 to judge. Then, concatenate q_nope with q_pe (and k_nope with k_pe) along the last dimension, and pass them as the main query/key tensors to npu_fused_infer_attention_score
    - Omit the query_rope and key_rope kwargs, which are not compatible with the padded-head path
  4. Decode path (**_forward_decode**):
    - Pad q_nope and q_pe with F.pad along the head dimension to num_heads_padded
    - Pass num_query_heads = num_heads_padded to npu_fused_infer_attention_score_v2
    - Slice the attention output back to the original num_heads with attn_output[:self.num_heads] before feeding to _v_up_proj
    - Covers all decode branches: TND_NTD (SpecDecoding), BSND_NBSD (fa_quant / enable_kv_nz), and BNSD_NBSD (normal decode)

  Changes in tests/ut/attention/test_mla_v1.py:

  - Added test_init_head_padding_for_non_power_of_two to verify padding arithmetic (20 -> 32)
  - Added test_forward_prefill_non_power_of_two_heads to verify prefill uses concat without query_rope/key_rope
  - Added test_compute_prefill_context_non_power_of_two_heads to verify prefill context uses concat
  - Added test_forward_decode_non_power_of_two_heads to verify decode with SpecDecoding state pads num_query_heads and slices output
  - Added test_forward_decode_non_power_of_two_heads_normal to verify decode with DecodeOnly + BNSD_NBSD path

 ###  Does this PR introduce any user-facing change?

  No. This is a backend compatibility fix for Ascend NPU attention ops. Users can now run GLM-4.7-Flash on Ascend without changing APIs or configurations.

###  How was this patch tested?

Accuracy Test: 
The markdown format results is as below:

| gsm8k | 401e4c | accuracy | gen | 86.05 |
| aime25 |  | accuracy |  | 93.33 |

  - All new unit tests pass locally via the existing pytest suite in tests/ut/attention/test_mla_v1.py
  - The 5 new test cases collectively cover:
    - __init__ padding computation
    - Prefill forward (concat path)
    - Prefill context (concat path)
    - Decode forward with SpecDecoding / TND_NTD (pad + slice)
    - Decode forward with DecodeOnly / BNSD_NBSD (pad + slice)

- vLLM version: v0.20.2

### DIFF
--- a/tests/ut/attention/test_mla_v1.py
+++ b/tests/ut/attention/test_mla_v1.py
@@ -1071,6 +1071,47 @@ class TestAscendMLAImpl(TestBase):
         self.assertIsNotNone(self.impl.kv_a_proj_with_mqa)
         self.assertIsNotNone(self.impl.kv_a_layernorm)
         self.assertEqual(self.impl.num_queries_per_kv, 32)
+        # 256 is power of 2, so padding should be 0
+        self.assertEqual(self.impl.num_heads_padded, 256)
+        self.assertEqual(self.impl.head_padding, 0)
+
+    @patch("vllm_ascend.attention.mla_v1.get_current_vllm_config")
+    def test_init_head_padding_for_non_power_of_two(self, mock_get_current_vllm_config):
+        """Test head padding computation for num_heads that are not power of 2 (e.g. GLM-4.7-Flash with 20 heads)."""
+        mock_get_current_vllm_config.return_value = MagicMock()
+        kwargs = {
+            "kv_lora_rank": 32,
+            "qk_nope_head_dim": 64,
+            "qk_rope_head_dim": 32,
+            "qk_head_dim": 96,
+            "v_head_dim": 128,
+            "q_lora_rank": 64,
+            "q_proj": MagicMock(),
+            "q_b_proj": MagicMock(),
+            "kv_b_proj": MagicMock(),
+            "o_proj": MagicMock(),
+            "kv_a_proj_with_mqa": MagicMock(),
+            "fused_qkv_a_proj": MagicMock(),
+            "kv_a_layernorm": MagicMock(),
+            "rotary_emb": MagicMock(),
+        }
+        impl = AscendMLAImpl(
+            num_heads=20,
+            head_size=1024,
+            scale=0.1,
+            num_kv_heads=20,
+            alibi_slopes=None,
+            sliding_window=None,
+            kv_cache_dtype="auto",
+            blocksparse_params=None,
+            logits_soft_cap=None,
+            attn_type=None,
+            kv_sharing_target_layer_name=None,
+            **kwargs,
+        )
+        self.assertEqual(impl.num_heads, 20)
+        self.assertEqual(impl.num_heads_padded, 32)  # next power of 2
+        self.assertEqual(impl.head_padding, 12)  # 32 - 20
 
     @patch("vllm_ascend.attention.mla_v1.get_ascend_config")
     @patch("vllm_ascend.attention.mla_v1.register_all_layers_to_shard_weight_series")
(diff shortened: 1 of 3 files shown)


ZT-AIA:
@yiz-liu Plz see see this pr  MLA

vllm_ascend/attention/mla_v1.py
```diff
@@ -1244,25 +1273,44 @@ def _forward_prefill(
...
-            "block_table": None,
-            "block_size": 0,
-            "softmax_lse_flag": True,
-            "actual_seq_lengths": actual_seq_lengths_q,
-            "actual_seq_lengths_kv": actual_seq_lengths_kv,
-        }
-
-        attn_output, attn_lse = torch_npu.npu_fused_infer_attention_score(q_nope, k_nope, value, **common_kwargs)
+        if self.head_padding > 0:
+            query = torch.cat((q_nope, q_pe), dim=-1)
```
weijinqian0, line 1290: Plz merge these two branches.

dsxsteven: Done. I merged two branches both in  _forward_prefill and _compute_prefill_context functions

weijinqian0 (approved):

weijinqian0 (approved):

ader47:
Decode pads only query heads, while KV heads remain unpadded, which can break the operator or corrupt GQA/MHA mapping。[vllm_ascend/attention/mla_v1.py#L1502-L1506](https://github.com/vllm-project/vllm-ascend/blob/ac165c22f5bde730d1bda9f6b3af120c1c29a132/vllm_ascend/attention/mla_v1.py#L1502-L1506)
In `mla_v1.py`, `num_query_heads` is changed to `self.num_heads_padded`, for example 20 -> 32, but `num_key_value_heads` remains `self.num_kv_heads`, and `k_nope/k_pe` are not padded. If `num_kv_heads == 20, 32` is not divisible by 20 and the operator may fail. In GQA cases, the query-to-KV grouping ratio changes, which can silently produce incorrect attention results. Please either pad KV heads/cache consistently while preserving the original grouping semantics, or restrict this path to supported head configurations and use a fallback otherwise.

ader47:
In the `fa_quant` A5 path, head padding makes dequant scale and output slicing inconsistent [mla_v1.py#L1460-L1468](https://github.com/vllm-project/vllm-ascend/blob/ac165c22f5bde730d1bda9f6b3af120c1c29a132/vllm_ascend/attention/mla_v1.py#L1460-L1468). [mla_v1.py#L1590-L1591](https://github.com/vllm-project/vllm-ascend/blob/ac165c22f5bde730d1bda9f6b3af120c1c29a132/vllm_ascend/attention/mla_v1.py#L1590-L1591)
In the A5 `fa_quant branch`, `q_nope/q_pe` are padded from [B, 20, 1, D] to [B, 32, 1, D], but `dequant_scale_q_nope` is still reshaped to `[B, 20, 1]`. This does not match the padded query-head dimension. Also, the common slice `attn_output[: self.num_heads]` slices the batch dimension for the `[B, num_heads_padded, 1, D]` layout, not the head dimension, so padded heads may still be passed into `_v_up_proj`. Please slice according to the actual output layout and pad or validate the dequant-scale head dimension.

ader47:
The new decode tests mock away the real shape issue.
[tests/ut/attention/test_mla_v1.py#L2296-L2311](https://github.com/vllm-project/vllm-ascend/blob/ac165c22f5bde730d1bda9f6b3af120c1c29a132/tests/ut/attention/test_mla_v1.py#L2296-L2311)、[test_mla_v1.py#L2372-L2385](https://github.com/vllm-project/vllm-ascend/blob/ac165c22f5bde730d1bda9f6b3af120c1c29a132/tests/ut/attention/test_mla_v1.py#L2372-L2385)
The tests mock `_v_up_proj` to return a valid final shape, so they do not verify the actual `attn_output` shape passed into `_v_up_proj`. They also only assert `num_query_heads`, not whether the padded output was sliced on the correct dimension. Please assert `_v_up_proj.call_args` input shape and add coverage for `fa_quant_layer=True` + A5 + non-power-of-two heads.

tests/ut/attention/test_mla_v1.py
```diff
@@ -1071,6 +1071,47 @@ def test_init(self):
...
+            "q_b_proj": MagicMock(),
+            "kv_b_proj": MagicMock(),
+            "o_proj": MagicMock(),
+            "kv_a_proj_with_mqa": MagicMock(),
+            "fused_qkv_a_proj": MagicMock(),
+            "kv_a_layernorm": MagicMock(),
+            "rotary_emb": MagicMock(),
+        }
+        impl = AscendMLAImpl(
+            num_heads=20,
```
YzTongNiar, line 1099: if it is necessary to consider the eadge condition with 'nunm_heads==1'?

PR #9574, opened 2026-05-26
[BugFix] Correct prompt_tokens_details in layerwise proxy server

postwter:
This PR fixes an issue where the final usage returned from the last chunk of the Decoder is inaccurate (specifically cached_tokens in prompt_tokens_details) when using the disaggregated prefill proxy. It ensures that the precise prompt_tokens_details generated from the Prefill response is preserved and merged into the final decoder response usage.

- vLLM version: v0.20.2


PadawanZH:
To maintain the simplicity and better compatibility of the proxy, I suggest addressing this issue within vLLM. Specifically, Decode should obtain the prompt_tokens_details information from prefill internally (not through proxy hard coded logic) and incorporate this information into the final usage. Otherwise, the proxy would need to parse each line of output, increasing overhead and requiring all production-level proxies that replace this example proxy to also implement this functionality.

PR #9584, opened 2026-05-26
[BugFix] Fix the bug of repeatedly calling set_device.

zouyida2052:
First, in `model_runner_v1`, `torch.npu.config.allow_internal_format = True` was moved to set_device in the `worker.py`; otherwise, it would call `set_device` prematurely when importing model_runner, causing an error.

Second, in `adapt_patch_worker`, there is also a place where `set_device` is called prematurely. I have opened an [issue](https://github.com/vllm-project/vllm-ascend/issues/9615) to track this. The main actions taken are:

In model_runner, lazy loading was implemented for `patch_draft_quarot`.

`conftest.py` can be deleted directly, as verified by the full CI pipeline.

no.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/patch/worker/patch_kimi_k25.py
+++ b/vllm_ascend/patch/worker/patch_kimi_k25.py
@@ -39,7 +39,9 @@ class AscendLearnable2DInterpPosEmbDivided_fixed(nn.Module):
     def forward(self, x: torch.Tensor, grid_thws: torch.Tensor) -> torch.Tensor:
         pos_embs = []
         for t, h, w in grid_thws.tolist():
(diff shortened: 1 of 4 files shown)


vllm_ascend/worker/worker.py
```diff
@@ -257,6 +257,9 @@ def _init_device(self):
         device = torch.device(f"npu:{self.local_rank}")
         torch.npu.set_device(device)
 
+        # if true, allow tensor initialization and casting with internal format (e.g., NZ)
+        torch.npu.config.allow_internal_format = True
```
Tflowers-0129: 需同步310P设备更改

Issue #9589, opened 2026-05-26
[Bug]: DeepSeek V4 w4a8 finite SwiGLU clamp in fused MoE quant paths can produce abnormal tokens


QwertyJack:
### Your current environment

<details>
<summary>Reproduction environment used during local validation</summary>

```text
vllm-ascend upstream/main: b26970e597222029d624261a8738a40f2d33ad17
vllm: /vllm-workspace/vllm, reported version 0.20.2 at service startup
vllm_ascend: local upstream/main worktree with Ascend plugin
Model: /models/DeepSeek-V4-Pro-w4a8-0505
Model config: model_type=deepseek_v4, swiglu_limit=10.0, n_routed_experts=384, n_shared_experts=1
Quant metadata: model_quant_type=W8A8_DYNAMIC, group_size=0
torch: 2.10.0+cpu
torch_npu: 2.10.0
SOC_VERSION: ascend910_9391
Target topology: TP16 / DP1 / EP enabled on one node
```

</details>

### 🐛 Describe the bug

DeepSeek V4 Pro w4a8 can generate abnormal tokens when `swiglu_limit=10.0` is executed through the fused MoE SwiGLU+quant paths on Ascend.

The minimal bad case is a deterministic chat-completion request with `temperature=0`, `top_p=1`, prompt length 131 tokens, and the DeepSeek V4 Pro w4a8 checkpoint. On clean `upstream/main`, the output contains non-sensical tokens such as `Foracci`, `†source`, `dátummal`, and `böjnings`.

#### Launch command

```bash
export ASCEND_RT_VISIBLE_DEVICES=0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15
export HCCL_DETERMINISTIC=true
export HCCL_BUFFSIZE=512
export HCCL_OP_EXPANSION_MODE=AIV
unset VLLM_ASCEND_ENABLE_FUSED_MC2
unset VLLM_ASCEND_ENABLE_FLASHCOMM1
unset VLLM_ASCEND_ENABLE_FLASHCOMM2
unset VLLM_ASCEND_FLASHCOMM2_PARALLEL_SIZE
export VLLM_USE_V1=1
export ASCEND_A3_ENABLE=1
export VLLM_ASCEND_APPLY_DSV4_PATCH=1

vllm serve /models/DeepSeek-V4-Pro-w4a8-0505 \
  --host 0.0.0.0 \
  --port 8000 \
  --served-model-name dsv4 \
[...]

PR #9590, opened 2026-05-26
[BugFix][DeepSeek] Apply SwiGLU clamp before MoE quantization

QwertyJack:

This PR applies the DeepSeek V4 `swiglu_limit` before MoE dynamic quantization in the Ascend Python/PTA path.

DeepSeek V4 Pro w4a8 has `swiglu_limit=10.0`. On the current fused routed/shared expert paths, that finite limit is passed into fused custom ops. A deterministic `bad_12` request can then emit abnormal tokens such as `Foracci`, `†source`, `dátummal`, and `böjnings`.

Changes:

- Routed W4A8 per-channel MoE: when `0 < swiglu_limit < 1_000_000`, run GMM1 and explicitly apply `clamp -> silu(gate) * up -> dynamic_quant` before GMM2. Non-effective limits still use the existing fused op path.
- Quantized shared experts: explicitly dequantize, apply the same DeepSeek V4 clamp ordering, then dynamic-quantize before the down projection.
- Shared-expert MLP fallback: pass `swiglu_limit` so unquantized shared experts use upstream `SiluAndMulWithClamp`.
- Add focused UT coverage for both routed and shared expert paths.

This is the PTA-side mitigation for #9589. The issue remains open for the follow-up custom-op investigation and fix in `csrc`.

Yes. It fixes incorrect DeepSeek V4 Pro w4a8 generation when a finite `swiglu_limit` is used in Ascend MoE routed/shared expert paths.

Unit/static validation on the latest `upstream/main` (`b26970e597222029d624261a8738a40f2d33ad17`):

```bash
python -m py_compile \
  vllm_ascend/ops/fused_moe/moe_mlp.py \
  vllm_ascend/ops/fused_moe/fused_moe.py \
  vllm_ascend/models/deepseek_v4.py

pytest -q tests/ut/ops/test_moe_mlp.py tests/ut/ops/test_fused_moe.py
# 37 passed, 6 skipped

ruff check \
  vllm_ascend/ops/fused_moe/moe_mlp.py \
  vllm_ascend/ops/fused_moe/fused_moe.py \
  vllm_ascend/models/deepseek_v4.py \
  tests/ut/ops/test_moe_mlp.py \
  tests/ut/ops/test_fused_moe.py
# All checks passed

git -c core.whitespace=cr-at-eol diff --check
```

Manual real-weight validation was also run before rebasing the same patch to the latest main:

- Model: `/models/DeepSeek-V4-Pro-w4a8-0505`
- Topology: TP16 / DP1 / EP enabled, single node
- Runtime flags: `max-model-len=2048`, `gpu-memory-utilization=0.95`, `max-num-seqs=16`, `HCCL_DETERMINISTIC=true`, `HCCL_BUFFSIZE=512`, `HCCL_OP_EXPANSION_MODE=AIV`, `cudagraph_mode=FULL_DECODE_ONLY`
- Fused MC2 and FlashComm envs were unset.

Clean main reproduced abnormal `bad_12` markers:

```text
Foracci=True, †source=True, dátummal=True, böjnings=True
```

With this patch, the same request no longer contained those abnormal markers:

```text
Foracci=False, †source=False, dátummal=False, böjnings=False, 16:51=False
```

### DIFF
--- a/tests/ut/ops/test_fused_moe.py
+++ b/tests/ut/ops/test_fused_moe.py
@@ -733,6 +733,59 @@ class TestAscendFusedMoE:
 
 
 class TestAscendFusedMoESharedExperts:
+    def test_shared_swiglu_applies_deepseek_v4_clamp_before_dynamic_quant(self, monkeypatch):
+        hidden_states = torch.tensor([[20, -20, 40, -40]], dtype=torch.int32)
+        weight_scale = torch.ones(4, dtype=torch.float32)
+        activation_scale = torch.ones(1, dtype=torch.float32)
+        public_dequant = MagicMock()
+
+        def fake_dynamic_quant(tensor):
+            return tensor.round().to(torch.int8), tensor.abs().amax(dim=-1).to(torch.float32) / 127
+
+        monkeypatch.setattr(fused_moe_module.torch_npu, "npu_dynamic_quant", fake_dynamic_quant)
+        monkeypatch.setattr(
+            fused_moe_module.torch.ops._C_ascend,
+            "npu_dequant_swiglu_quant",
+            public_dequant,
+            raising=False,
+        )
+
+        output, output_scale = fused_moe_module._shared_dequant_swiglu_quant(
+            hidden_states,
+            weight_scale,
+            activation_scale,
+            10.0,
+            torch.bfloat16,
+        )
+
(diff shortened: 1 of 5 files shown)


QwertyJack:
Routed expert clamp is fixed by #9438 
Shared expert clamp is fixed by #9600 & #9741

PR #9597, opened 2026-05-26
[Feature] Adapt FlashComm1 SP for Step3.5

wangyichao1999:
Adapt FlashComm1 (SP) in vllm-ascend for the Step3.5 model with two fixes in `linear_op.py`:

  **1. Add `share_expert` prefix matching (3 places)**

  Step3.5 names its shared expert weights with `share_expert` (no trailing 's'), but the original code only matches `shared_expert` / `shared_experts`. This causes shared_expert  layers to incorrectly participate in SP column/row parallel ops. Added `share_expert` matching in `_get_column_parallel_op`, `_get_row_parallel_op`, and `get_parallel_op`.

**2.Add "g_proj" to SP column-parallel**
 Add "g_proj" to SP column-parallel dispatch prefixes for Step3p5 attention gate projection. Without this, the g_proj layer falls through to None and does not participate in sequence parallelism, causing a shape mismatch at the residual addition.

 No.

Tested on Ascend 910B with Step3.5 Flash model (MoE, W8A8, TP=8). Verified that shared_expert layers are correctly skipped and `g_proj` participates in SP column-parallel
  dispatch. Model runs without errors when `VLLM_ASCEND_ENABLE_FLASHCOMM1=1`.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/ops/test_linear.py
+++ b/tests/ut/ops/test_linear.py
@@ -169,12 +169,10 @@ class TestAscendReplicatedLinear(BaseLinearTest):
 
 
 class TestColumnParallelOpDispatch(unittest.TestCase):
-    """Tests for _get_column_parallel_op factory — DSV4 removal, share_expert, g_proj."""
+    """Tests for _get_column_parallel_op factory — share_expert, g_proj."""
 
     def setUp(self):
         self.mock_layer = MagicMock()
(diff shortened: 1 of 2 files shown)


zzzzwwjj (approved):

realliujiaxu (approved):
LGTM

PR #9600, opened 2026-05-26
[Feature] Fix the precision issue of dsV4 caused by dequant_swiglu_quant

ZT-AIA:
#### 1、Fix the precision issue of dsV4 caused by a bug in the operator:
    1. The default values of the last two parameters of the operator should be changed to 1, 0.
    2. If the groupindex parameter is not passed, it will cause the clamp function to be disabled.
    3. when clamp_limit==0，the clamp function is turned off.
no

ci

#9589 
#9590 

- vLLM version: v0.20.2


MengqingCao:
/nightly tests/e2e/nightly/single_node/ops/singlecard_ops/test_dequant_swiglu_quant.py

MengqingCao (approved):

PR #9613, opened 2026-05-27
[BugFix] Fix A5 custom opapi symbol shadowing

maoxx241:

Fix A5 custom opapi symbol shadowing in the legacy `BUILD_WITH_3_8_PACKAGE` custom-op build path.

The A5 custom `libcust_opapi.so` should not link against the full CANN `libopapi.so`, because that can expose or shadow opapi symbols from the CANN package instead of keeping the custom opapi symbols isolated in vLLM Ascend's build output.

This PR updates `csrc/cmake/custom_build.cmake` to use `opapi_math` for the legacy A5 path, matching the current upstream `cann/ops-transformer` custom opapi link model.

No. This is an internal custom-op build/linkage fix and does not add a public API, CLI flag, or default behavior change.

- `git diff --check`
- confirmed the diff is limited to `csrc/cmake/custom_build.cmake`

- vLLM version: v0.20.2

### DIFF
--- a/csrc/cmake/custom_build.cmake
+++ b/csrc/cmake/custom_build.cmake
@@ -80,13 +80,14 @@ if (BUILD_OPEN_PROJECT)
     target_compile_definitions(cust_opapi PRIVATE
             -DACLNN_LOG_FMT_CHECK
     )
-    if(BUILD_WITH_3_8_PACKAGE)  #3~8包 依赖opapi, 不依赖opbase
+    if(BUILD_WITH_3_8_PACKAGE)  # 3~8 package links opapi_math and does not depend on opsbase.
     target_link_libraries(cust_opapi PRIVATE
         $<BUILD_INTERFACE:intf_pub>
         -Wl,--whole-archive
         ops_aclnn
         -Wl,--no-whole-archive
-        -lopapi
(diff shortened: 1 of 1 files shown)


maoxx241:
Upstream `cann/ops-transformer` current master was checked against this change:

- repo: https://gitcode.com/cann/ops-transformer
- HEAD: `cc4905c06268b861b8bddfa8bc7bb6c698bebcbd`
- A5 is handled as `ASCEND_COMPUTE_UNIT=ascend950 -> arch35`; upstream no longer has the legacy `BUILD_WITH_3_8_PACKAGE` special path.
- In upstream `cmake/custom_build.cmake`, the `cust_opapi` link path has `-lopapi` commented out and links `opapi_math` separately. That matches the intent here: custom `libcust_opapi.so` should not gain `DT_NEEDED libopapi.so`, because `dlsym` on the custom opapi handle can otherwise resolve CANN built-in ACLNN symbols such as `aclnnMoeDistributeDispatch/CombineV2*` through the dependency.

This PR is the minimal backport of that behavior into vLLM Ascend's older inherited build framework, which still has `BUILD_WITH_3_8_PACKAGE` for A5.

MengqingCao (approved):
Thanks for this bug fix!

PR #9626, opened 2026-05-27
Add GMMSwigluQuantV2 DSV4 W4A8 e2e coverage

maoxx241:
## Summary
- keep GMMSwigluQuantV2 on the count-mode contract used by the W4A8 per-channel MoE path
- reject group_list_type=0 at the torch adapter boundary so the fused v2 op does not silently return mismatched results for cumsum input
- add a DSV4 W4A8 per-channel e2e operator test that loads real ModelSlim new-version w1/w3 expert weights, converts them through the same NZ/int32 + scale/bias layout, and compares fused v2 against the MoE small-op sequence

## Validation
- rebuilt vllm_ascend_C on 173.131.1.2 with SOC_VERSION=ascend910b4
- W4A8 ModelSlim per-channel direct compare, count mode, experts [0,1,2,3], counts [2,3,1,4], clamp 0.05: qmax=1, dqmax=0.00016533397138118744
- W4A8 ModelSlim per-channel direct cumsum input now fails fast with: grouped_matmul_swiglu_quant_v2 only supports count-mode group_list
- added e2e regression: tests/e2e/nightly/single_node/ops/singlecard_ops/test_grouped_matmul_swiglu_quant_v2_dsv4_w4a8.py
- current library result for the new e2e on 173.131.1.2: 1 passed, 1 failed; failing clamp-stress-one-token-per-expert has qmax=2 against the small-op reference, so fused v2 precision is still not fully aligned
- checked upstream/main build_aclnn.sh: grouped_matmul_swiglu_quant_v2 is already included for ascend910b and ascend910_93
- vLLM version: v0.20.2

### DIFF
--- a/csrc/gmm/grouped_matmul_swiglu_quant_v2/grouped_matmul_swiglu_quant_v2_torch_adpt.h
+++ b/csrc/gmm/grouped_matmul_swiglu_quant_v2/grouped_matmul_swiglu_quant_v2_torch_adpt.h
@@ -40,6 +40,11 @@ std::tuple<at::Tensor, at::Tensor> grouped_matmul_swiglu_quant_v2(
     int n = weight_scale[0].sizes().back();
     int m = x_size[0];
     int k = x_size[1];
+    TORCH_CHECK(
+        group_list_type == 1,
+        "grouped_matmul_swiglu_quant_v2 only supports count-mode group_list, "
+        "but received group_list_type=",
+        group_list_type);
 
     at::Tensor output =  at::empty({m, n/2}, x.options().dtype(at::kChar));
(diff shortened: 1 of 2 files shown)


maoxx241:
Closing this draft because the change targets the arch35/A5 kernel path, while the investigated A2/A3 path uses the generic arch32 GroupedMatmulSwigluQuantV2 kernel. Current upstream/main already registers grouped_matmul_swiglu_quant_v2 in csrc/build_aclnn.sh for both ascend910b and ascend910_93, so there is no build_aclnn delta to carry in a main-based PR. A build_aclnn-only fix should be opened against the older/backport branch where those entries are missing.

PR #9634, opened 2026-05-27
[Feature] [worker]add group_name param when create hccl_config

mfyCn-1204:
Profiler tool need save parallel info for user to analyze model communication problems.
So we add "group_name": group_name in hccl_config, then profiler can get parallel info by using torch_npu apis and save it to profiler data file([_add_group_info_to_metadata](https://github.com/Ascend/pytorch/blob/master/torch_npu/profiler/profiler_interface.py )).

No
Use profiler tool to collect vllm-ascend model data, then check if there is parallel info in profiler data file

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -645,9 +645,11 @@ class NPUModelRunner(GPUModelRunner):
             # Mixed-batch case: num_reqs must equal num_reqs_padded
             assert num_reqs == num_reqs_padded
 
-            # Insert a dummy request instead of setting query_start_loc[num_reqs] = num_tokens_padded directly
-            self.query_start_loc.np[num_reqs_padded + 1] = num_tokens_padded
-            num_reqs_padded = num_reqs_padded + 1
+            # Do not insert if the last value already equals the num_tokens
+            if self.query_start_loc.np[num_reqs_padded] < num_tokens_padded:
+                # Insert a dummy request instead of change the last value directly
+                self.query_start_loc.np[num_reqs_padded + 1] = num_tokens_padded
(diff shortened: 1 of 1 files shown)


vllm_ascend/utils.py
```diff
@@ -1075,6 +1075,9 @@ def create_hccl_pg_options(group_name: str):
     hccl_config = get_hccl_config_for_pg_options(group_name)
     if hccl_config is not None:
         options.hccl_config = hccl_config
+    hccl_config = options.hccl_config
+    hccl_config.update({"group_name": group_name})
+    options.hccl_config = hccl_config
```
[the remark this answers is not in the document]
mfyCn-1204: use new variable config to add group_name param for options.hccl_config

vllm_ascend/utils.py
```diff
@@ -1075,6 +1075,9 @@ def create_hccl_pg_options(group_name: str):
     hccl_config = get_hccl_config_for_pg_options(group_name)
     if hccl_config is not None:
         options.hccl_config = hccl_config
+    config = options.hccl_config
+    config.update({"group_name": group_name})
+    options.hccl_config = config
```
yiz-liu:
```suggestion
    if hccl_config:
        hccl_config["group_name"] = group_name
        options.hccl_config = hccl_config
```

yiz-liu (commented):
Please change to this.

vllm_ascend/utils.py
mfyCn-1204: If hccl_config returned by get_hccl_config_for_pg_options() is None, options.hccl_config will not add group_name param. So we use new variable 'config' from options.hccl_config, add group_name param, then set options.hccl_config to it.

PR #9636, opened 2026-05-27
Fix W4A8 GMMSwigluQuantV2 shape handling

maoxx241:
## Summary

Fix the W4A8 per-channel GMMSwigluQuantV2 path so it stays on the fused v2 op when `swiglu_limit` is set and uses the same effective clamp behavior as the MoE MLP small-op reference.

The root cause was in the fused op shape path: ModelSlim W4A8 weights arrive as int32-packed NZ tensors, and the aclnn entry was still letting the packed storage shape leak into the W4 tiling path. The fix derives the unpacked NZ storage shape from `output` and `x`, and the tiling side recovers the pre-SwiGLU `n` from the output shape.

This branch is based on upstream PR #9590. PR #9590 already adds `grouped_matmul_swiglu_quant_v2` to the A2/A3 build path and keeps A2/A3 on `arch32`; this follow-up does not add any new `arch35` mapping. Until #9590 lands, this PR's diff includes those base changes as well.

## Validation

- `git diff --check`
- 131 single-op rebuild: `bash build.sh --pkg --ops=grouped_matmul_swiglu_quant_v2 --soc=ascend910b -j16`
- 131 W4A8 compare script: `/tmp/pr9590_mlp_v2_compare.py`
  - custom op enabled: `enable_custom_op True has_v2 True`
  - W4A8 format path: `w_nz (4, 7168, 768) scale2d (4, 6144) scale3d (4, 1, 6144)`
  - fused vs small-op quantized max diff: `1`
  - exact match range: `0.9658203125` to `0.9818522334098816`
- 131 unit test: `python3 -m pytest tests/ut/ops/test_moe_mlp.py::TestQuantApplyMlpW4A8PerChannel -q`
  - `2 passed, 16 warnings`

- vLLM version: v0.20.2

### DIFF
--- a/csrc/gmm/grouped_matmul_swiglu_quant_v2/op_host/grouped_matmul_swiglu_quant_v2_base_tiling.cpp
+++ b/csrc/gmm/grouped_matmul_swiglu_quant_v2/op_host/grouped_matmul_swiglu_quant_v2_base_tiling.cpp
@@ -23,6 +23,8 @@ namespace GroupedMatmulSwigluQuantV2Tiling {
 
 constexpr int64_t ND_WEIGHT_MULTI_TENSOR_DIM = 2;
 constexpr int64_t NZ_WEIGHT_MULTI_TENSOR_DIM = 4;
+constexpr int64_t OUTPUT_INDEX = 0;
+constexpr int64_t SWIGLU_SPLIT_FACTOR = 2;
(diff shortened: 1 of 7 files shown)


maoxx241:
Closing duplicate. Reusing #9626 for this GMMSwigluQuantV2 follow-up.

PR #9654, opened 2026-05-28
[Feature] Allow causal DFlash

HF-001:
refer to: https://github.com/vllm-project/vllm/pull/43445  , This PR is a first step to supporting causal DFlash models. As a starting point, I allow dflash checkpoints to enable full causal or non-causality. This will help enable sliding-window DFlash models for which non-causal attention kernels do not exist.

The goal of this PR is not to fully support SWA or hybrid causal-non-causal models, but just to make configurable whether we require DFlash to be causal or not. The default is unchanged.

todo

- vLLM version: v0.23.0


HF-001:
@wangxiyuan hi, this pr is ready, cloud you help review it?

PR #9671, opened 2026-05-28
[Ascend950][Feature]Support MXFP8 FlashcommV3 on 950

Eric-dot:

This PR adds support for MXFP8 FlashCommV3 on Ascend 950.

Specifically, it updates the W8A8_MXFP8 MoE path to reuse `topk_weights` and `topk_ids` from `FlashCommon3Context` when `multistream_overlap_gate` is enabled, instead of calling `select_experts` again. The original `select_experts` path is kept for non-overlap scenarios.

This is needed to make the MXFP8 fused MoE path compatible with FlashCommV3 multistream overlap on Ascend 950, avoid duplicated gating computation, and ensure that MoE execution uses the routing results generated by the FlashCommV3 overlap path.

Performance on Ascend 950 with DeepSeek V3.1 prefill scenario using variable-length input data from 0 to 16K:

| Scenario | FlashCommV3 | Input Length | QPS |
| --- | --- | --- | --- |
| DeepSeek V3.1 Prefill | Disabled | 0-16K variable length | 0.62 |
| DeepSeek V3.1 Prefill | Enabled | 0-16K variable length | 0.64 |

This shows that enabling FlashCommV3 improves QPS from 0.62 to 0.64 in the DeepSeek V3.1 prefill scenario on Ascend 950.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/quantization/methods/w8a8_mxfp8.py
+++ b/vllm_ascend/quantization/methods/w8a8_mxfp8.py
@@ -288,6 +288,9 @@ class AscendW8A8MXFP8DynamicFusedMoEMethod(AscendMoEScheme):
                 tid2eid=tid2eid,
             )
 
+        if topk_weights is None or topk_ids is None:
+            raise RuntimeError("topk_weights and topk_ids must be set before fused MoE execution.")
+
         # this is a naive implementation for experts load balance so as
         # to avoid accumulating too much tokens on a single rank.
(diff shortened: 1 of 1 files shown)


realliujiaxu (approved):

realliujiaxu:
LGTM. Let's merge it after v0.20.2 is released.

PR #9676, opened 2026-05-28
[BugFix][Model] Recompute DeepSeek V4 hash routing weights

QwertyJack:

DeepSeek V4 W4A8 can hit NaN router weights in the `sqrtsoftplus + tid2eid/hash routing` path. The hash-routing custom op still returns finite `topk_ids`, but its returned `topk_weights` can become NaN and then propagate into MoE output for long generations.

This PR keeps using `moe_gating_top_k_hash` for selected expert ids, but recomputes the corresponding weights from `router_logits` in FP32:

```python
sqrt(softplus(router_logits)).gather(topk_ids)
```

The recomputed weights are then renormalized and scaled exactly like the existing DeepSeek V4 path. The W4A8 MoE path also keeps `topk_weights` in FP32 before expert scaling to avoid losing precision before MC2/ALLTOALL combine can promote them again.

The branch was created from `main@b26970e597222029d624261a8738a40f2d33ad17`.

No. This is an internal accuracy fix for DeepSeek V4 W4A8 MoE routing on Ascend.

- `python -m py_compile vllm_ascend/ops/fused_moe/experts_selector.py vllm_ascend/quantization/methods/w4a8.py tests/ut/ops/test_moe_hash_routing_input_ids.py`
- `ruff check vllm_ascend/ops/fused_moe/experts_selector.py vllm_ascend/quantization/methods/w4a8.py tests/ut/ops/test_moe_hash_routing_input_ids.py`
- `pytest -q tests/ut/ops/test_moe_hash_routing_input_ids.py tests/ut/quantization/methods/test_w4a8.py::TestAscendW4A8DynamicFusedMoEMethod::test_apply_comprehensive`

The new unit tests cover comm-aligned input ids for hash routing and the regression case where the custom op returns NaN weights while `topk_ids` and `router_logits` remain finite.

Manual validation was also run on DeepSeek-V4-Pro W4A8 with TP16DP1EP, eager mode, prefix cache disabled, FlashComm/FUSED_MC2 disabled, `HCCL_DETERMINISTIC=true`, and `HCCL_BUFFSIZE=512`. A long deterministic completion that previously produced abnormal tokens no longer propagated non-finite values through the effective MoE/hidden/logits path after this fix.

- vLLM version: v0.20.2

### DIFF
--- a/tests/ut/ops/test_moe_hash_routing_input_ids.py
+++ b/tests/ut/ops/test_moe_hash_routing_input_ids.py
@@ -1,0 +1,119 @@
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+import torch
+
+from vllm_ascend.ascend_forward_context import MoECommType
+from vllm_ascend.ops.fused_moe import experts_selector as experts_selector_module
+
+
+@pytest.mark.parametrize(
+    "moe_comm_type",
+    [
+        MoECommType.ALLGATHER,
+        MoECommType.MC2,
+        MoECommType.ALLTOALL,
+        MoECommType.FUSED_MC2,
+    ],
+)
+def test_hash_routing_uses_comm_aligned_input_ids(monkeypatch, moe_comm_type):
+    raw_input_ids = torch.tensor([7, 8, 9, 10], dtype=torch.int64)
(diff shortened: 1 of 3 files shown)


QwertyJack:
fixed by PR #9692

PR #9689, opened 2026-05-29
[BugFix]Fix multi-DP deadlock for extract_hidden_states proposer on MoE targets

learning-sketch:

`AscendExtractHiddenStatesProposer` inherits the upstream `ExtractHiddenStatesProposer._determine_batch_execution_and_padding` unchanged, which on Ascend causes failures when running `extract_hidden_states` on a MoE target model with `DP > 1` and sequence parallelism enabled (e.g. MiniMax-M2 with `VLLM_ASCEND_ENABLE_FLASHCOMM1=1`).

#### Bug 1 — gloo shape mismatch on the DP cpu_group

```
what(): [enforce fail at .../gloo/transport/tcp/pair.cc:456]
op.preamble.length <= op.nbytes. 8 vs 4.
Received data size doesn't match expected size.
Is there a distributed collective mismatch in your code?
```

Upstream `coordinate_batch_across_dp` posts a `[4, dp_size]` int32 tensor to the DP cpu_group, while Ascend's main runner uses `_sync_metadata_across_dp` with a `[2, dp_size]` tensor on the same cpu_group. The two shapes collide within one step.

#### Bug 2 — reduce_scatter shape-not-divisible assertion on the idle DP rank

```
File ".../vllm_ascend/ops/linear_op.py", line 574, in matmul_and_reduce
    output = tensor_model_parallel_reduce_scatter(output_parallel, 0)
File ".../base_device_communicator.py", line 234, in reduce_scatter
    assert input_tensor.shape[0] % world_size == 0
AssertionError
```

The proposer's own `cudagraph_dispatcher` is initialized as PIECEWISE/NONE only (never FULL), so `dispatch(num_tokens=6)` returns 6 as-is (no SP padding). That 6 enters DP sync, the synced max stays 6, and the idle DP rank's main MoE forward then crashes in SP `reduce_scatter` because `6 % TP=4 != 0`.

Eagle3/MTP do not reproduce this because Ascend's `AscendEagleProposer` uses `runner.cudagraph_dispatcher.dispatch(...)` which dispatches against the runner's FULL-mode capture sizes (always TP-aligned).

#### Bug 3 / Bug 4 — multi-DP deadlock on MoE targets

When the drafter sync performs a real all_reduce (e.g. hierarchical communication enabled), `MoE target + DP>1 + extract_hidden_states` serving hangs:

- `is_drafter_moe_model()` misclassified the cache-only extract_hidden_states drafter as MoE because its `hf_config` copies the (MoE) target `hf_config`, so the expert-key scan matched. This made `_sync_metadata_across_dp(is_draft_model=True)` perform a real gloo all_reduce on the DP cpu_group in the drafter path that idle DP ranks never match.
- `AscendExtractHiddenStatesProposer.dummy_run` never issued the drafter DP sync, unlike `llm_base_proposer.dummy_run`. Busy DP ranks (propose path) post one more cpu_group collective per step than idle ranks (dummy path), the collectives pair across different steps, token counts get garbled, and all ranks deadlock in the device-side MoE collectives.

### Fix

Override `AscendExtractHiddenStatesProposer._determine_batch_execution_and_padding`:

- SP-pad `num_tokens` via `runner._pad_for_sequence_parallelism` before dispatch, so the contribution to DP sync is always TP-aligned. Mirrors what the runner's main path does at `model_runner_v1.py:_determine_batch_execution_and_padding`.
- Use `runner._sync_metadata_across_dp` (packed_tensor shape `[2, dp_size]`) for DP coordination instead of upstream `coordinate_batch_across_dp` (shape `[4, dp_size]`), so all DP collectives in a single step that hit the cpu_group use a consistent tensor shape.
- Fail fast at the entry of the override with a clear `AssertionError` if the proposer was constructed without a runner reference, instead of letting the unguarded `runner._pad_for_sequence_parallelism` call raise a confusing `AttributeError`.
- Document the `is_draft_model=True` semantics: it intentionally makes `should_skip_allreduce_across_dp_group` short-circuit (the cache-only "draft" here is not MoE), so the call degenerates to a local broadcast. The actual cross-DP all_reduce has already been done by the main runner earlier in the step; the SP padding above is what keeps the value TP-aligned regardless.

Additionally fix the two DP-collective bugs that hang `MoE target + DP>1 + extract_hidden_states` serving:

- **`is_drafter_moe_model()` misclassification** (`vllm_ascend/utils.py`): the cache-only extract_hidden_states drafter is a single attention layer and never contains MoE layers, but its `hf_config` is built by copying the (possibly MoE) target `hf_config`, so the generic expert-key scan marked it MoE. Special-case `method == "extract_hidden_states"` to always return `False` (merged with the existing `Eagle3DeepseekV2ForCausalLM` special-case).
- **Missing drafter DP sync in `dummy_run`** (`extract_hidden_states_proposer.py`): mirror `llm_base_proposer.dummy_run` and issue the same `_sync_metadata_across_dp(num_tokens, is_draft_model=True)` in `dummy_run`, so busy and idle DP ranks always stay collective-balanced.

No user-facing API change. Fixes a runtime crash **and a multi-DP hang** for users running `extract_hidden_states` speculative decoding with `--data-parallel-size > 1` on MoE target models on Ascend NPU. Single-DP runs and dense target models (e.g. Qwen3-8B) are unaffected.

Reproduced the crash/hang and verified the fix on MiniMax-M2:

- 2x8 NPU, `--tensor-parallel-size 8 --data-parallel-size 2`
- `--enable-expert-parallel`
- `--compilation-config '{"cudagraph_mode": "FULL_DECODE_ONLY"}'`
- `--speculative-config '{"method": "extract_hidden_states", "num_speculative_tokens": 1, "draft_model_config": {"hf_config": {"eagle_aux_hidden_state_layer_ids": [2, 18, 34]}}}'`
- `--kv-transfer-config '{"kv_connector": "ExampleHiddenStatesConnector", "kv_role": "kv_producer", ...}'`

Sending one `/v1/completions` request with `max_tokens=1`:

- Before fix: idle DP rank crashes on first `execute_dummy_batch` with the `AssertionError` shown above (and hangs once the drafter sync does a real all_reduce).
- After fix: request returns `200 OK`, hidden_states `.safetensors` file is written with the expected `(prompt_len, len(layer_ids), hidden_size)` shape.

Also verified the existing dense Qwen3-8B + extract_hidden_states path still works unchanged.

MiniMax-M2:

Unit tests added in `tests/ut/spec_decode/test_extract_hidden_states_proposer.py`:

- `test_determine_batch_execution_and_padding_asserts_when_runner_is_none`: regression guard for the `AttributeError` that would otherwise be raised on the unguarded `self.runner._pad_for_sequence_parallelism` call at the entry of the override.
- `test_determine_batch_execution_and_padding_dp1_sp_pads_and_skips_sync`: with `DP=1` the runner's `_pad_for_sequence_parallelism` is still consulted (so `cache_only` forward gets an SP-aligned input) but `_sync_metadata_across_dp` is not called.
- `test_determine_batch_execution_and_padding_dp2_uses_runner_sync`: with `DP>1` the override calls `runner._sync_metadata_across_dp` with the SP-padded `num_tokens` and `is_draft_model=True`, and does NOT call the upstream `coordinate_batch_across_dp` (regression guard for the gloo `8 vs 4` shape mismatch).
- `test_determine_batch_execution_and_padding_dp2_keeps_tp_aligned_for_main_forward`: regression guard for the `reduce_scatter` `input.shape[0] % world_size == 0` assertion — the final `num_tokens_padded` returned to the caller is always TP-aligned.
- `test_dummy_run_syncs_metadata_across_dp_as_draft_model`: regression guard for the multi-DP deadlock — `dummy_run` must issue the same `_sync_metadata_across_dp(..., is_draft_model=True)` drafter sync as the busy `propose()` path, and forward the synced tensor to `set_forward_context`.

Unit tests added in `tests/ut/test_utils.py`:

- `test_is_drafter_moe_model_extract_hidden_states_is_never_moe`: the extract_hidden_states drafter is never classified as MoE even when its inherited `hf_config` carries MoE keys (`num_local_experts`, `num_experts_per_tok`).
- `test_is_drafter_moe_model_eagle_moe_drafter_detected`: non-extract_hidden_states drafters keep the expert-key MoE detection.

Related upstream PR: vllm-project/vllm#39949 (introduced extract_hidden_states speculative method).

- vLLM version: v0.22.1


Ronald1995 (approved):
LGTM

vllm_ascend/spec_decode/extract_hidden_states_proposer.py
```diff
@@ -39,6 +39,78 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device, runner=None):
...
+                )
+            # Reuse the runner's DP sync so the collective shape matches the
+            # main forward. ``is_draft_model=True`` short-circuits the
+            # all_reduce (cache-only drafter is not MoE); ``dummy_run`` issues
+            # the identical call to keep busy and idle DP ranks balanced.
+            (
+                _max_tokens_across_dp,
+                num_tokens_across_dp,
+                synced_cudagraph_mode,
+            ) = self.runner._sync_metadata_across_dp(
```
lilinsiman, line 96: Check whether the synchronization operation here affects the graph mode.

learning-sketch: the sync does interact with cudagraph_mode, but only in the
intended, safe way that mirrors the main runner:
  1. _sync_metadata_across_dp returns a synced_cudagraph_mode via
     _post_process_cudagraph_mode, i.e. the min across DP ranks (if any rank is
     NONE, all fall back to NONE). We re-dispatch with
     valid_modes={synced_cudagraph_mode} immediately after, exactly like the
     runner's own _determine_batch_execution_and_padding. This is required so all
     DP ranks agree on one graph mode; otherwise the device-side collectives
     desync (the deadlock this PR fixes).
  2. For the extract_hidden_states drafter itself, is_draft_model=True together
     with is_drafter_moe_model()==False makes should_skip_allreduce_across_dp_group
     return True (when hierarchy comm is off), so the call short-circuits and
     returns the input cudagraph_mode unchanged — no all_reduce, graph mode
     untouched. It only gets coordinated on the real all_reduce branch (hierarchy
     comm enabled).
  3. The proposer's own dispatcher only ever uses PIECEWISE/NONE (never FULL), and
     this runs during the dispatch/planning phase before set_forward_context /
     forward, i.e. outside any graph capture/replay region, so it cannot corrupt
     graph capture. The re-dispatch is also guarded by
     assert batch_desc.num_tokens == num_tokens_padded.

Net effect: the sync can only downgrade the proposer's graph mode to the common
DP mode (never upgrade), which is intended. Verified on MiniMax-M2 with
--compilation-config '{"cudagraph_mode": "FULL_DECODE_ONLY"}' and DP=2.

realliujiaxu (approved):
LGTM. please update PR description

learning-sketch:
> LGTM. please update PR description

PR #9697, opened 2026-05-29
[Misc](dsa_cp): introduce additional_config for dsa_cp

pisceskkk:

This PR introduces an additional_config option to control dsa_cp.

Previously, dsa_cp was tightly coupled with FC1. On A2 hardware, due to the relatively poor performance of the All2All operator, enabling dsa_cp can lead to significant performance degradation. At the same time, there was no way to disable dsa_cp while keeping FC1 enabled through the existing configuration switches. This PR decouples the two by providing a dedicated configuration option for dsa_cp.

Yes, introduces an additional_config option. Now, we should enable both FC1 and DSA_CP if want to use dsa_cp.

- vLLM version: v0.20.2


linfeng-yuan (commented):
Please update your e2e test configuration

pisceskkk:
> Please update your e2e test configuration

have updated.

PR #9700, opened 2026-05-29
[Refactor] Align swap blocks ops with vLLM

Pz1116:
## Summary
- Align Ascend swap_blocks custom op signature with vLLM by accepting src, dst, block_size_in_bytes, and block_mapping.
- Align swap_blocks_batch with the upstream three-argument src_ptrs, dst_ptrs, and sizes interface.
- Keep Ascend memcpy direction handling inside the C++ implementation and update CPU offloading to use the community-shaped batch API.

## Validation
vLLM version: v0.20.1

- vLLM version: v0.20.2

### DIFF
--- a/csrc/torch_binding.cpp
+++ b/csrc/torch_binding.cpp
@@ -148,7 +148,8 @@ aclrtMemcpyKind infer_memcpy_kind(const void* src, const void* dst)
     if ((src_is_host || src_ret != ACL_SUCCESS) && dst_is_device) {
         return ACL_MEMCPY_HOST_TO_DEVICE;
     }
-    if (src_is_host && dst_is_host) {
+    if ((src_is_host || src_ret != ACL_SUCCESS) &&
+        (dst_is_host || dst_ret != ACL_SUCCESS)) {
         return ACL_MEMCPY_HOST_TO_HOST;
     }
     return ACL_MEMCPY_DEFAULT;
@@ -290,6 +291,8 @@ void swap_blocks_impl(torch::Tensor& src, torch::Tensor& dst,
     }
 
     TORCH_CHECK(block_mapping.device().is_cpu(), "block_mapping must be on CPU");
+    TORCH_CHECK(block_mapping.dtype() == torch::kInt64,
+                "block_mapping must be int64");
     TORCH_CHECK(block_size_in_bytes > 0,
                 "block_size_in_bytes must be positive, got ",
                 block_size_in_bytes);
--- a/csrc/torch_binding.cpp
+++ b/csrc/torch_binding.cpp
@@ -119,6 +119,9 @@ void enqueue_device_print(std::unique_ptr<DevicePrintPayload> payload,
 
 aclrtMemcpyKind infer_memcpy_kind(const void* src, const void* dst)
 {
+    TORCH_CHECK(src != nullptr, "src pointer must not be null");
+    TORCH_CHECK(dst != nullptr, "dst pointer must not be null");
+
     aclrtPtrAttributes src_attrs = {};
     aclrtPtrAttributes dst_attrs = {};
     const aclError src_ret =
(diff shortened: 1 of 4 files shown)


Pz1116:
@HF-001 please take a look

csrc/torch_binding.cpp
```diff
@@ -143,28 +180,16 @@ void swap_blocks_batch(const torch::Tensor& src_ptrs,
...
-            memcpy_kind = ACL_MEMCPY_DEVICE_TO_DEVICE;
-            break;
-        default:
-            TORCH_CHECK(false,
-                        "swap_blocks_batch: invalid direction ", direction,
-                        " (expected 0=H2D, 1=D2H, 2=D2D)");
-    }
+    const aclrtMemcpyKind memcpy_kind = infer_memcpy_kind(
+        reinterpret_cast<const void*>(src_data[0]),
+        reinterpret_cast<const void*>(dst_data[0]));
```
HF-001, line 193: Only look at the direction inferred from item 0 and use this direction to run the entire batch. If the upper layer mixes pointers from different directions (D2H+H2D)
Mixed together, all subsequent copies will be submitted in the wrong direction, possibly written to the wrong memory or silently rejected by ACL. The previous explicit parameter of direction, although ugly, strongly constrained one direction at a time. Now that it has been changed to automatic inference, this constraint has become "dependent on the caller's self-awareness" without any code layer defense。

csrc/torch_binding.cpp
```diff
@@ -117,12 +117,49 @@ void enqueue_device_print(std::unique_ptr<DevicePrintPayload> payload,
...
+        src_attrs.location.type == ACL_MEM_LOCATION_TYPE_HOST;
+    const bool dst_is_host =
+        dst_ret == ACL_SUCCESS &&
+        dst_attrs.location.type == ACL_MEM_LOCATION_TYPE_HOST;
+
+    if (src_is_device && dst_is_device) {
+        return ACL_MEMCPY_DEVICE_TO_DEVICE;
+    }
+    if (src_is_device && (dst_is_host || dst_ret != ACL_SUCCESS)) {
+        return ACL_MEMCPY_DEVICE_TO_HOST;
```
HF-001, line 149: The failure of aclrtPointerVNet does not equal 'pointer pointing to host memory'. Possible scenarios: illegal pointer itself, cross process pointer, non ACL allocated by CUDAMalloc/libnuma
Register memory and even null pointers. These are all silently classified as hosts, so ACL will submit D2H → write to illegal addresses → segment errors or worse.

csrc/torch_binding.cpp
```diff
@@ -117,12 +117,49 @@ void enqueue_device_print(std::unique_ptr<DevicePrintPayload> payload,
...
+    return ACL_MEMCPY_DEFAULT;
+}
+
 }
 
 void swap_blocks_batch(const torch::Tensor& src_ptrs,
                        const torch::Tensor& dst_ptrs,
-                       const torch::Tensor& sizes,
-                       int64_t direction) {
+                       const torch::Tensor& sizes) {
```
HF-001, line 170: The Python wrapper is missing the is_strc_ccess_order/any parameter, and the ABI is not aligned with the upstream

HF-001:
> @HF-001 please take a look

@Pz1116 Previously, the Ascend community stated that they would no longer maintain native kvcache offloading. Have you decided to continue developing and maintaining it now

csrc/torch_binding.cpp
Pz1116: Thanks, agreed. I updated this in b1cf1bc1: the aclrtMemcpyBatchAsync path is now used only when every entry infers the same H2D or D2H direction. If the batch contains mixed directions, or any entry falls back to ACL_MEMCPY_DEFAULT, we skip the batch path and use the per-entry aclrtMemcpyAsync loop with direction inferred for each src/dst pair. This avoids reusing item 0 direction for later copies.

csrc/torch_binding.cpp
Pz1116: Thanks, agreed. I changed infer_memcpy_kind in b1cf1bc1 so aclrtPointerGetAttributes failure is no longer treated as host memory. We now return a concrete H2D/D2H/D2D/H2H kind only when the pointer attributes identify both sides as supported host/device locations, otherwise the per-copy path uses ACL_MEMCPY_DEFAULT. Null pointers are also rejected explicitly.

csrc/torch_binding.cpp
Pz1116: Thanks for checking this. I rechecked upstream vLLM and kept the ABI unchanged intentionally: the public C++ schema and Python wrapper for swap_blocks_batch are still the three arguments src_ptrs, dst_ptrs, and sizes. srcAccessOrder is an internal CUDA implementation detail for cuMemcpyBatchAsync, not a Python wrapper parameter. If Ascend needs a separate access-order control later, we can discuss it independently, but adding that parameter here would diverge from the upstream swap_blocks_batch ABI.

Pz1116:
> > @HF-001 please take a look
> 
> @Pz1116 Previously, the Ascend community stated that they would no longer maintain native kvcache offloading. Have you decided to continue developing and maintaining it now

Thanks for the comments! The reason we do this is that we have an independent demand of swap blocks ops, which requires us to allign the usage and input parameters to the ones upstream.

csrc/torch_binding.cpp
```diff
@@ -117,12 +117,57 @@ void enqueue_device_print(std::unique_ptr<DevicePrintPayload> payload,
     TORCH_CHECK(ret == ACL_SUCCESS, "aclrtLaunchHostFunc failed, error code: ", ret);
 }
 
+aclrtMemcpyKind infer_memcpy_kind(const void* src, const void* dst)
```
MengqingCao, line 120: @ZT-AIA could you plz take a look at this kernel?

ZT-AIA (approved):
No problem, if possible, remember to add test cases.

PR #7476, opened 2026-03-19
[Feature][KV Pool]KV Pool supports kv load failure block ids

Pz1116:
### What this PR does / why we need it?
Currently, KV Pool cannot handle the case where get operation fails, and the output will have accuracy issues. This PR tries to handle this issue by providing failure block ids needed for vllm scheduler.

### How was this patch tested?

- vLLM version: v0.20.2
- vLLM main: https://github.com/vllm-project/vllm/commit/1ac10f159a09897baada01b14b6a0dd6442aefd6

PR #9701, opened 2026-05-29
[Feature][KV Pool] Support load failure block ids without hybrid recompute

Pz1116:

Based on #7476, this PR keeps KV Pool load failure block-id reporting for non-hybrid models and removes the hybrid peer-block failure grouping logic.

Hybrid models now explicitly assert when kv_load_failure_policy is set to recompute, because KV load failure recompute recovery is not supported for hybrid KV cache layouts yet.

Hybrid models fail fast if kv_load_failure_policy="recompute" is configured.

### Validation

vLLM version: v0.20.2
vLLM main: vllm-project/vllm@1ac10f159a09897baada01b14b6a0dd6442aefd6


LCAIZJ (commented):
Why were modifications to the vllm_ascend/platform.py file added?

LCAIZJ (approved):

Pz1116:
> Why were modifications to the vllm_ascend/platform.py file added?

This is to assert that for hybrid models, recompute policy is not supported yet

PR #9706, opened 2026-05-29
[BugFix] fix EVS for vllm-ascend

garrygale:
cf. Issue #9709 
We have confirmed that the Efficient Video Sampling (EVS) feature in vllm can not be properly triggered in vllm-ascend.
This bug covers at least 0.16, 0.18 & 0.19 versions of vllm-ascend, and the main branch.

This PR fixes it.

### Fix
Modified model_runner_v1.py and vocab_parallel_embedding.py to fix the logic. 
~~Patched qwen3-vl to revert one of the changes in upstream vllm such that it can work with NPU.~~

No.

Launching the service with option --video_pruning_rate [a_non_zero_value], and send a request with a video input.
Fix works on vllm 0.18.0 and 0.19.1 with corresponding vllm-ascend releases.
Fix works on vllm 0.20.2 with vllm-ascend main branch.

~~### Why patching Qwen3-VL?~~
PR #34246 in vllm (Apr. 1, 2026) introduces a simplification of the code in the masking operation following EVS calculation in _create_final_video_embeddings method in Qwen3-VL. 
When testing on A2 it leads to an error in method wrapper_NPU__embedding as the indices cannot be in CPU while other tensors are on NPU. 
~~We therefore patch the method such that repl_token_ids are created on NPU rather than CPU.~~
Edit: No longer patching on qwen3-vl. Instead, in vocab_parallel_embedding.py, move the index to the same device as the embedding weight if they are on different devices (i.e. one on CPU and one on NPU).

Further edit: We managed to fix this qwen3-vl issue in the upstream vllm main branch (vllm-project/vllm#44205). The temporary fix in vocab_parallel_embedding.py in vllm-ascend might still be needed at this moment, but can removed in later versions.
* The fix in npu_model_runner.py is still necessary.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/ops/vocab_parallel_embedding.py
+++ b/vllm_ascend/ops/vocab_parallel_embedding.py
@@ -161,7 +161,8 @@ class AscendVocabParallelEmbedding(VocabParallelEmbedding):
         return input_, ~vocab_mask
 
     def forward(self, input_):
-        if input_.device != self.weight.device:
+        if input_.device != self.weight.device: 
+            # TODO: Remove this temporal fix after 0.23.0
             input_ = input_.to(self.weight.device)
         if self.forward_type == "embed_tp":
             return self._forward_embed_tp(input_)
--- a/vllm_ascend/ops/vocab_parallel_embedding.py
+++ b/vllm_ascend/ops/vocab_parallel_embedding.py
@@ -161,9 +161,6 @@ class AscendVocabParallelEmbedding(VocabParallelEmbedding):
         return input_, ~vocab_mask
 
     def forward(self, input_):
-        # TODO: Remove this temporal fix after 0.23.0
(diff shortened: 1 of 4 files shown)


weijinqian0 (approved):

vllm_ascend/ops/vocab_parallel_embedding.py
```diff
@@ -161,6 +161,8 @@ def _get_masked_input_and_mask(
         return input_, ~vocab_mask
 
     def forward(self, input_):
+        if input_.device != self.weight.device:
```
kunpengW-code: Add a TODO to describe the removal plan.

garrygale: Done, please check.

PR #9713, opened 2026-05-29
[Feature] TP/DP async parallel process startup

hwhaokun:

- vLLM version: v0.20.2


hwhaokun:
@zzzzwwjj

PR #9717, opened 2026-05-29
[BugFix][SpecDecode] Add validation of the DCP for the draft model

lilinsiman:
When DCP and speculative decoding are enabled, there is no validation for the parallel strategy configuration of the draft model's DCP. As a result, when the backend of the main model and the draft model is inconsistent, the parallel strategy becomes entirely incorrect due to the lack of validation.

no

tests and ut

- vLLM version: v0.20.2


linfeng-yuan (approved):
LGTM

PR #9720, opened 2026-05-29
[BugFix][310P] Fix the precision of the causal_conv1d_v310 operator on 310P

YangShuai52:
Fix precision issues in the 310P causal_conv1d_v310 custom op caused by incorrect state initialization, ring buffer handling, and missing Ascend pipeline synchronization.

Fix hasInit default: When hasInitialStateMode is not provided, only treat initial state as present in decode mode (runMode == 1). Prefill no longer incorrectly reads stale conv_states.
Initialize unused ring slots: Zero-fill ring buffer slots before use to avoid garbage values affecting convolution when width < MAX_WIDTH.
Restore V-pipe barrier in RunSeq: Add PipeBarrier<PIPE_V>() after Cast and before MulAddDst to prevent read-before-write races on the vector pipe.
Fix state writeback indexing: Use RetreatRingSlot() to walk back from the last ring slot when writing back history tokens.
Improve cross-engine sync: Replace generic PipeBarrier with HardEvent-based synchronization in WriteBackStateSpec, and add writeback sync before/after state updates in the main loop.
RFC #9730
RFC #9711 
No

- vLLM version: v0.20.2

### DIFF
--- a/csrc/moe/causal_conv1d_v310/op_kernel/causal_conv1d_v310.h
+++ b/csrc/moe/causal_conv1d_v310/op_kernel/causal_conv1d_v310.h
@@ -64,6 +64,9 @@ private:
     TEventID outVToMte3Event_[2];
     TEventID stateWritebackMte3ToVEvent_;
     TEventID stateWritebackMte3ToMte2Event_;
+    TEventID stateShiftMte2ToMte3Event_;
+    TEventID stateShiftMte3ToMte2Event_;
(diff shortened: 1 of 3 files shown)


linfeng-yuan:
Please update the e2e case of this op with bs=16,32 and more, etc.

ZT-AIA (approved):

linfeng-yuan (approved):

PR #9724, opened 2026-05-30
[BugFix]fix hybrid block table out of range bug when pcp or dcp

weiguihua2:
Fix the block table out-of-bounds issue in scenarios where PCP or DCP hybrid linear encoding is applied.

In the PCP or DCP mixed linear scenario, the kv cache of full attention is partitioned, while the Mamba layer remains intact. Therefore, when applying for blocks, the Mamba does not need to divide by pcp_size/dcp_size.

No

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/worker/block_table.py
+++ b/vllm_ascend/worker/block_table.py
@@ -3,7 +3,7 @@ import torch
 from vllm.distributed import get_dcp_group, get_pcp_group
 from vllm.utils.math_utils import cdiv
 from vllm.v1.attention.backends.utils import PAD_SLOT_ID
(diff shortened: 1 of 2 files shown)


vllm_ascend/worker/block_table.py
```diff
@@ -24,32 +24,38 @@ def __init__(
         kv_cache_group: KVCacheGroupSpec = None,
     ):
         self.max_num_reqs = max_num_reqs
+        try:
```
linfeng-yuan: remove this try block

linfeng-yuan (approved):
LGTM

PR #9735, opened 2026-05-30
[BugFix]Reduce sampling is reconstructed to eliminate all patch behaviors and support DFlash and MTP

hzx55906:

The Reduce_sampling optimization is reconstructed to eliminate all patch behaviors and support the DFlash and MTP.
When sampling optimization is enabled, if speculative decoding is Eagle3 or DFlash, sampling can be optimized for both the main model and the MTP layer. If the speculative decoding method is MTP, some models can optimize sampling for both the main model and the MTP layer, while others can only optimize sampling for the main model.

- vLLM version: v0.20.2


vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -347,15 +365,16 @@ def _maybe_share_lm_head(self, model: nn.Module) -> None:
                 if torch.equal(layer_module.shared_head.head.weight, model.lm_head.weight):
                     layer_module.shared_head.head = model.lm_head
 
-        if self.vllm_config.compilation_config.cudagraph_mode.has_full_cudagraphs() and self.use_cuda_graph:
+        if self.vllm_config.compilation_config.cudagraph_mode.has_full_cudagraphs():
```
lilinsiman: Check this place, self.update_stream = torch.npu.Stream()

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -879,6 +897,17 @@ def _propose(
                 self._update_full_graph_params_if_needed(forward_context, num_input_tokens, multi_steps_attn_metadata)
         return draft_token_ids
 
+    def compute_draft_token_ids(self, hidden_states: torch.Tensor):
```
MengqingCao, line 900: this rename makes the context more clear 👍

MengqingCao (approved):

PR #9763, opened 2026-05-30
[Feature] Support extract_hidden_states for hybrid models on Ascend NPU

learning-sketch:

Sync vllm-ascend with upstream vLLM PR #39949 so that the `extract_hidden_states`
speculative decoding method works on hybrid (Mamba/GatedDeltaNet + full-attention)
target models such as Qwen3.5. Dense models (e.g. Qwen3-8B) already worked and
stay unaffected.

`extract_hidden_states` adds a cache-only draft layer (`CacheOnlyAttentionLayer`)
whose KV spec is the upstream marker `HiddenStateCacheSpec` (a subclass of
`MLAAttentionSpec`). On hybrid models this layer was mishandled on NPU at several
stages of KV cache init:

- **Engine-side grouping (`get_kv_cache_groups`)**: vLLM isolates
  `HiddenStateCacheSpec` layers into their own KV cache group via
  `isinstance(spec, HiddenStateCacheSpec)`. NPU's `_get_layer_kv_cache_specs`
  rebuilt the cache-only spec as a plain `MLAAttentionSpec` (rebuild is required
  because `patch_kv_cache_interface.py` shadows `MLAAttentionSpec`, so the
  originally-imported class is not picklable). This dropped the marker type, so
  the hidden-state layer was not pulled out, its page size leaked into
  `unify_kv_cache_spec_page_size`, and startup failed with
  `NotImplementedError: The page size of the layer is not divisible by the maximum page size`.

- **Worker-side allocate/reshape (`_allocate_kv_cache_tensors` /
  `_reshape_kv_cache_tensors`)**: cache-only layers were identified only by the
  `"cache_only_layers"` substring, and the hybrid attn+mamba branch did not
  exclude them, so on hybrid models they were routed into the `(k, v)` path and
  produced shape/page-size mismatches. Additionally, once isolated, the layer's
  page is aligned to the hybrid common (padded) page, so the single-tensor
  reshape must account for the per-block padding.

- **Mamba cache mode (`patch_mamba_config.py`)**: hybrid models with a KV store
  force `mamba_cache_mode = "align"`, which routes them onto vLLM's fused GPU
  postprocess Triton kernel (vLLM #40172) that the Ascend Triton backend cannot
  compile. The `extract_hidden_states` connector only manages its dedicated
  cache-only layer and does not migrate mamba KV blocks, so it must not be forced
  into `"align"`.

### Changes

- Import `HiddenStateCacheSpec` directly and add an `is_hidden_state_cache_spec()`
  helper in `vllm_ascend/utils.py`.
- Preserve the `HiddenStateCacheSpec` marker when rebuilding the
  `CacheOnlyAttentionLayer` spec in `_get_layer_kv_cache_specs`, so upstream
  grouping isolates it into its own KV cache group. The marker class is not
  shadowed by the patch, so the rebuilt spec stays picklable.
- Identify cache-only layers by spec type (in addition to the layer name) in
  `_allocate_kv_cache_tensors` and `_reshape_kv_cache_tensors`, and exclude them
  from the hybrid attn+mamba `(k, v)` branch so they always take the dedicated
  single-tensor path.
- In the single-tensor branch, use a strided view (`torch.as_strided`) when
  `page_size_padded` is set, so the per-block padding introduced by aligning to
  the hybrid common page is skipped (mirrors upstream `GPUModelRunner`
  `page_size_padded` handling).
- Exclude `extract_hidden_states` from forcing the block-aligned mamba cache mode
  in `patch_mamba_config.py`, leaving the mode as vLLM derived it.

Yes. `extract_hidden_states` speculative decoding can now be used with hybrid
attention models (e.g. Qwen3.5) on Ascend NPU. There is no change for dense
models or for models that do not use `extract_hidden_states`.

Consolidated the dense and hybrid coverage into a single parametrized
`tests/e2e/pull_request/one_card/spec_decode/test_extract_hidden_states.py`:

- **Dense (Qwen3-8B)** in both eager and ACL graph modes, using real weights so
  the extracted hidden states can be asserted non-zero.
- **Hybrid (Qwen3.5-0.8B, GatedDeltaNet + full_attention)** loaded with
  `load_format="dummy"` (no checkpoint download) as a shape/round-trip smoke
  test, mirroring upstream `test_extract_hidden_states_qwen35_hybrid_smoke`. It
  validates that the hidden-state layer is allocated as a single tensor, HMA
  stays enabled, and the extracted hidden states have the expected shape
  `(num_prompt_tokens, len(layer_ids), hidden_size)`.

The estimated time for this test is bumped to 400s in
`.github/workflows/scripts/test_config.yaml`.

Verified end-to-end on Ascend NPU: Qwen3.5 + `extract_hidden_states` with
`ExampleHiddenStatesConnector` now starts and serves successfully (previously
failed at KV cache init). The dense Qwen3-8B path continues to pass.

`bash format.sh` (ruff check/format, codespell, typos, import/lint hooks) passes.

- vLLM version: v0.22.1


learning-sketch:
/e2e tests/e2e/pull_request/one_card/aclgraph/test_aclgraph_mem.py
[Bot]: e2e command triggered. See [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/27320735531) for details.
[Bot]: e2e command failed.

.github/workflows/scripts/test_config.yaml
```diff
@@ -493,6 +493,7 @@
     - vllm_ascend/spec_decode/extract_hidden_states_proposer.py
   tests:
     - tests/e2e/pull_request/one_card/spec_decode/test_extract_hidden_states.py
+    - tests/e2e/pull_request/one_card/spec_decode/test_extract_hidden_states_qwen35.py
```
Ronald1995: it's better to combine the `test_extract_hidden_states` and `test_extract_hidden_states_qwen35.py` into on file with pytest.parameterize.

learning-sketch: Done — combined both into test_extract_hidden_states.py with pytest.mark.parametrize (dense_eager, dense_aclgraph, hybrid_dummy_eager), removed the qwen35 file, and updated test_config.yaml.

Ronald1995 (approved):

PR #9776, opened 2026-05-30
[BugFix] Honor tool_choice="none" in Chat Completions streaming

### LINKED ISSUE #42747: [Bug]: Chat Completions streaming invokes tool parser despite `tool_choice="none"`
## Current environment

```text
vLLM v0.21.0
```

### 🐛 Describe the bug

# The bug

When using the OpenAI-compatible Chat Completions API in streaming mode, `tool_choice="none"` does not prevent tool parsing if the server was launched with a tool parser.

In non-streaming Chat Completions, `tool_choice="none"` is respected: generated tool-call-looking text remains in `message.content` and is not converted into `message.tool_calls`.

In streaming Chat Completions, after the migration to the unified `DelegatingParser.parse_delta()` path, the streaming parser can still enter the tool-call phase and call `extract_tool_calls_streaming(...)` even when the request has `tool_choice="none"`. If the model output matches the configured parser's tool-call format, vLLM can emit `delta.tool_calls` and finish with `finish_reason="tool_calls"` despite the explicit `tool_choice="none"` request.

**Expected behavior:**
- tool_choice="none" should prevent streamed delta.tool_calls.
- Model-generated tool-call-looking text should remain ordinary content, matching non-streaming behavior.
- The final streaming chunk should not use finish_reason="tool_calls".

**Actual behavior:**
- The streaming path can still invoke the configured tool parser.
- Matching model output can be converted into streamed delta.tool_calls.

## Example launch

Example with Kimi K2:

```bash
vllm serve moonshotai/Kimi-K2.6 \
  --enable-auto-tool-choice \
  --tool-call-parser kimi_k2 \
  --reasoning-parser kimi_k2
```

This is not specific to Kimi K2. The issue appears generic to Chat Completions streaming whenever a tool parser is configured. Similar behavior should be possible for other parser/model pairs such as Qwen XML/Coder, DeepSeek, Hermes, Gemma, etc.

## Example API request

```json
{
  "model": "moonshotai/Kimi-K2.6",
  "stream": true,
  "tool_choice": "none",
  "messages": [
    {
      "role": "system",
      "content": "You are a weather inquiry agent."
    },
    {
      "role": "user",
      "content": "Say hi first, and then call the tool to tell me the current temperatures in New York City and San Francisco?"
    }
  ],
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "get_weather",
        "description": "Get the current weather in a given location",
        "parameters": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "City and state, e.g. San Francisco, CA"
            },
            "unit": {
              "type": "string",
              "enum": ["celsius", "fahrenheit"]
            }
          },
          "required": ["location", "unit"]
        }
      }
    }
  ]
}
```

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

Liuchenbing-2026:
## Summary

Mirror of upstream [vllm-project/vllm#42752](https://github.com/vllm-project/vllm/pull/42752) (fixes [vllm-project/vllm#42747](https://github.com/vllm-project/vllm/issues/42747)) for vllm-ascend.

Streaming Chat Completions with `tool_choice="none"` (or omitted on a no-tools request) could still produce `delta.tool_calls` and finish with `finish_reason="tool_calls"` because `DelegatingParser.parse_delta` invokes `_extract_tool_calls_streaming` unconditionally once the stream enters the tool-call phase, ignoring `request.tool_choice`.

Non-streaming Chat Completions already short-circuits this in `chat_completion/serving.py:1250`:

```python
elif not request.tool_choice or request.tool_choice == "none":
    message = ChatMessage(role=role, reasoning=reasoning, content=content)
```

The streaming path was missing the equivalent guard.

## Fix

Monkey-patch `DelegatingParser.parse_delta` in vllm-ascend. When `tool_choice` is `None` or `"none"`, skip the tool parser inside the tool-call phase and surface the (post-reasoning) `delta_text` as plain `DeltaMessage.content`.

Effect:
- DSV4 DSML markup (and any other parser's tool-call-looking output) stays in `delta.content`, matching the non-streaming behavior.
- `finish_reason` falls back to `"stop"`.
- `delta.tool_calls` is never emitted when `tool_choice="none"`.

## Guard scope

Gate covers both cases per the discussion under #42747:
- explicit `tool_choice="none"`
- omitted `tool_choice` on a no-tools request (`tool_choice` defaults to `None`)

Upstream PR #42752 only checks `== "none"`; this PR additionally covers the `tool_choice is None` case to fully match the non-streaming guard `not request.tool_choice or request.tool_choice == "none"`.

## Behavior alignment (streaming vs non-streaming)

| Scenario | Non-streaming `content` | Streaming `delta.content` | `tool_calls` | `finish_reason` |
|---|---|---|---|---|
| `tool_choice="none"` / omitted | raw (incl. DSML) | raw (incl. DSML) | None | `"stop"` |
| `tool_choice="auto"` + tools | parsed (DSML stripped) | parsed + streamed tool_calls | extracted | `"tool_calls"` |

## Files

| File | Change |
|---|---|
| `vllm_ascend/patch/platform/patch_tool_choice_none_streaming.py` | +152 (new) |
| `vllm_ascend/patch/platform/__init__.py` | +1 (import) |

## History

This PR replaces an earlier attempt that monkey-patched `OpenAIServing._parse_tool_calls_from_content` to **strip** DSML markup from non-streaming `content`. The upstream consensus in #42747 is the opposite — both modes should **leave** the markup in `content`, neither should strip it. The earlier two commits (`f7ad97b2`, `3ad57170`) have been removed via force-push to the branch base.

- vLLM version: v0.20.2


QwertyJack:
I don’t think this should be fixed in the serving layer.

When no `tools` are provided, `tool_choice=none` is the expected behavior, and the model output should be treated as normal assistant content. If the model emits DSML markup in that mode, returning it verbatim is acceptable.

Silently stripping DSML on the no-tools path changes user-visible model output and may remove text the user intentionally asked the model to generate. The caller should either provide `tools` when tool calling is intended, or handle any desired cleanup at the application layer.

Liuchenbing-2026:
Closing — branch has drifted with unresolved conflicts; will reopen against current main if still needed.

PR #9780, opened 2026-05-30
[BugFix] Chunk wq_b matmul for NPU 65536 dimension limit

jianzs:
## Summary

- `torch_npu.npu_quant_matmul` does not support weight dimensions >= 65536. When DSA context parallel is enabled, the `wq_b` linear layer's output dimension exceeds this limit, causing `aclnnQuantMatmulV5` to fail with error code 161002.
- Split the `wq_b` weight into two chunks along the output dimension so each chunk stays below 65536. Two `npu_quant_matmul` calls are issued and their outputs concatenated along the last dimension.
- Guard the chunking with `enable_dsa_cp()` since only DSA-CP models hit the 65536 limit on `wq_b`.

## Changes

- **w8a8_dynamic.py**: Add chunked path in `process_weights_after_loading()` that splits weight/scale/offset into halves, flattens scales/offsets to 1D, and applies `maybe_trans_nz` for NZ format conversion. Add chunked forward in `apply()` with two matmul + `torch.cat`.
- **dsa_cp.py**: Add chunked forward in the DSA CP attention path for `wq_b`, matching the same two-matmul-and-cat pattern.

## Test Plan

- [x] Deploy DeepSeek V4 prefill service on Ascend A3 (2 nodes, TP=4, DP=4) with DSA-CP enabled
- [x] Verified `npu_quant_matmul` no longer fails with AclNN_Parameter_Error
- [x] Passed prefill CI perf test (50 requests, throughput 19460 tok/s; 10 requests, mean TTFT 5.7s)

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/attention/context_parallel/dsa_cp.py
+++ b/vllm_ascend/attention/context_parallel/dsa_cp.py
@@ -1009,14 +1009,37 @@ class AscendDSACPImpl(DSAAttentionImpl):
             qr_local, qr_pertoken_scale_local = torch.ops._C_ascend.npu_rms_norm_dynamic_quant(
                 q_a, self.q_norm.weight, epsilon=self.eps
             )
-            q = torch_npu.npu_quant_matmul(
-                qr_local,
-                self.wq_b.weight,
-                self.wq_b.weight_scale,
-                pertoken_scale=qr_pertoken_scale_local,
-                bias=self.wq_b.bias,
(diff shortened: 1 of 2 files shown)


linfeng-yuan (approved):
LGTM. @jianzs Make sure the concrete time of torch_npu.quant_mamul can support `65536` and we can remove this processs.

Issue #8327, opened 2026-04-16
[Bug]: (vLLM 0.18.0rc1&vllm-ascend:v0.18.0rc1)GLM5模型开启Function-Call后，“流式”输出tools-call消息中`finish_reason` 内容与最后一段数据封装在一个chunk中，不符合OpenAI标准定义


komawong168:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>
镜像：quay.io/ascend/vllm-ascend:v0.18.0rc1
模型：GLM-5-w8a8
HDK:  npu-smi 25.2.3    Version: 25.2.3  

启动脚本（开启function call功能）：
/usr/local/python3.11.14/bin/python3 /usr/local/python3.11.14/bin/vllm serve /workdir/GLM-5-w8a8 --host 0.0.0.0 --port 6600 --data-parallel-size 4 --data-parallel-rank 0 --data-parallel-address x.x.x.x --data-parallel-rpc-port 10521 --tensor-parallel-size 8 --enable-expert-parallel --enable-chunked-prefill --enable-prefix-caching --seed 1024 --served-model-name glm-5 --max-model-len 135168 --max-num-batched-tokens 4096 --trust-remote-code --max-num-seqs 48 --gpu-memory-utilization 0.95 --quantization ascend --async-scheduling --enforce-eager **--enable-auto-tool-choice --tool-call-parser glm47 --reasoning-parser glm45** --kv-transfer-config { "kv_connector": "MooncakeConnectorV1", "kv_role": "kv_producer", "kv_port": "30000", "engine_id": "0", "kv_connector_module_path": "vllm_ascend.distributed.mooncake_connector", "kv_connector_extra_config": { "use_ascend_direct": true, "prefill": { "dp_size": 4, "tp_size": 8 }, "decode": { "dp_size": 8, "tp_size": 4 } } } --additional-config { "recompute_scheduler_enable":true, "multistream_overlap_shared_expert":false, "fuse_qknorm_rope": false, "fuse_muls_add": true, "enable_npugraph_ex": true } --speculative-config {"num_speculative_tokens": 3, "method":"deepseek_mtp"}
```text
Your output of above commands here
```

</details>

### 🐛 Describe the bug

“流式”输出tools-call的消息中`finish_reason` 内容与最后一段数据封装在一个chunk中，不符合OpenAI标准定义，会导致下游系统丢弃包含最后一段数据的风险，实例如下：
.....
[...]

PR #9787, opened 2026-05-31
[BugFix][Platform] Patch GLM tool-call final chunks

QwertyJack:

Backports the serving-layer GLM tool-call final chunk fixes while keeping the upstream vLLM GLM parser implementation.

The patch fixes the two remaining GLM streaming protocol issues tracked by #8327:

- final remaining-argument chunks no longer re-emit `id`, `type`, or `function.name` by default
- terminal argument chunks with `finish_reason="tool_calls"` are split into an argument chunk with `finish_reason=null`, followed by an empty finish chunk

This intentionally does not restore the old GLM parser state-machine monkey patch; upstream vLLM already contains the parser-side terminal argument drainage / incremental streaming behavior. The new patch only covers the serving-layer final chunk behavior that is still missing upstream.

Related upstream tracking:

Yes. GLM OpenAI-compatible streaming tool-call responses no longer repeat function metadata in final argument chunks, and final argument bytes are sent before the empty `tool_calls` finish chunk.

- `ruff check vllm_ascend/patch/platform/patch_glm_tool_call_streaming.py vllm_ascend/patch/platform/__init__.py tests/ut/patch/platform/test_patch_glm_tool_call_streaming.py`
- `python -m py_compile vllm_ascend/patch/platform/patch_glm_tool_call_streaming.py tests/ut/patch/platform/test_patch_glm_tool_call_streaming.py`
- `python -m pytest -q tests/ut/patch/platform/test_patch_glm_tool_call_streaming.py` (`4 passed`)
- import probe confirmed both MiniMax usage wrapper and GLM final-chunk wrapper are installed on `OpenAIServingChat`

- vLLM version: v0.20.2

### DIFF
--- a/tests/ut/patch/platform/test_patch_glm_tool_call_streaming.py
+++ b/tests/ut/patch/platform/test_patch_glm_tool_call_streaming.py
@@ -1,0 +1,114 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import json
+
+from vllm.entrypoints.openai.chat_completion.serving import OpenAIServingChat
+from vllm.entrypoints.openai.engine.protocol import (
+    DeltaFunctionCall,
+    DeltaMessage,
+    DeltaToolCall,
+)
+
+from vllm_ascend.patch.platform import (
+    patch_glm_tool_call_streaming as glm_streaming_patch,
+)
+
+
+def test_remaining_args_delta_omits_metadata_by_default():
+    original_delta = DeltaMessage(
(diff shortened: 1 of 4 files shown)


linfeng-yuan (approved):
UT passed here: https://github.com/vllm-project/vllm-ascend/actions/runs/26757222872/job/78861039019?pr=9787 and tool call parser is irrelevant with E2E cases. We can merge it first.

PR #9789, opened 2026-05-31
[Misc] Rehash AscendStore grouped keys for DSV4/compress layouts

Pz1116:
Rehash grouped AscendStore block hashes with a framed SHA-256 digest when the store key granularity is larger than the original hash granularity. This keeps the key suffix fixed length while preserving all child block hashes in the digest input, reducing KV Pool metadata and backend key pressure for DSV4/compress layouts.

No. This only changes internal KV Pool key generation.

### Validation
vLLM version: v0.20.1

- vLLM version: v0.20.2


LCAIZJ:
Why is the hash so long? We use SHA-256 for hash calculation, which should theoretically be 32 bytes, so it shouldn't be too long.

LCAIZJ (approved):

Pz1116:
> Why is the hash so long? We use SHA-256 for hash calculation, which should theoretically be 32 bytes, so it shouldn't be too long.

This is because in the previous approach, we concat all the block hashes with `"".join(block_hashes[idx : idx + scale_factor])`, which results in super long keys. We decide to change our appoarch to rehashing grouped keys.

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
```diff
@@ -459,13 +463,28 @@ def get_block_hashes(
...
+    return [
+        _rehash_block_hash_group(block_hashes[idx : idx + scale_factor])
+        for idx in range(0, len(block_hashes) // scale_factor * scale_factor, scale_factor)
+    ]
+
+
+def _rehash_block_hash_group(block_hashes: Sequence[BlockHash | str]) -> BlockHash:
+    hasher = hashlib.sha256()
+    hasher.update(_GROUPED_BLOCK_HASH_DOMAIN)
+    hasher.update(len(block_hashes).to_bytes(4, "big"))
```
Ronald1995: This is a magic number, it's better to define a global variable.

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
```diff
@@ -459,13 +463,28 @@ def get_block_hashes(
...
+    for block_hash in block_hashes:
+        hash_bytes = _block_hash_to_bytes(block_hash)
+        hasher.update(len(hash_bytes).to_bytes(4, "big"))
```
Ronald1995: ditto

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
Pz1116: Thanks, updated in 3297272d. I replaced the raw `4` with `_GROUPED_BLOCK_HASH_LENGTH_PREFIX_BYTES`.

I kept it as a module-private constant rather than an externally configurable global value, because this framing is only used by the KV Pool internal grouped-key format and both sides should use the same fixed encoding.

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
Pz1116: Updated this one in the same commit as well. The per-block-hash length prefix now also uses `_GROUPED_BLOCK_HASH_LENGTH_PREFIX_BYTES`, so the grouped-key framing no longer has a raw magic number.

MengqingCao (approved):

PR #9801, opened 2026-06-01
[CI][Feature] Add vLLM verified ref anchors

### LINKED ISSUE #9800: [CI] Add vLLM main LKG anchor for incremental main-to-main updates
## Motivation

vLLM Ascend currently references the vLLM main-compatible commit in multiple places, such as workflow matrices and docs metadata. When the main-to-main adaptation process advances incrementally, these duplicated references make it easy to update one location but miss another.

The TPU inference repository uses a simple LKG-style commit file as a source of truth for its vLLM baseline. vLLM Ascend could use the same idea while keeping a repository-specific filename.

## Proposal

Add a small Last Known Good vLLM commit anchor for vLLM Ascend and wire CI/helper code to read from it. The goal is to support an incremental main-to-main workflow:

1. keep the current verified vLLM main commit in one tracked file;
2. let PR/full/light CI consume that commit instead of duplicating the SHA inline;
3. keep the release tag check alongside the LKG commit so CI still verifies the supported vLLM tag;
4. make helper scripts use the same source of truth when auto-detecting the known-good commit.

The filename should not be the same as TPU inference's `vllm_lkg.version`.

## Acceptance criteria

- A repository-specific LKG file records the current verified vLLM main commit.
- PR light/full workflows read the main commit from that file while keeping the release-tag matrix entry.
- The vLLM bisect helper can auto-detect the commit from the same source.
- The implementation is validated with local static checks where possible.

robellliu-dev:

Fixes #9800.

This PR adds two repository-local verified vLLM refs as the shared source of truth for PR selected tests and docs metadata:

- `.github/vllm-main-verified.commit`: verified upstream vLLM main commit.
- `.github/vllm-release-tag.commit`: matched vLLM release tag.

The refs stay in separate files because they have different semantics. The workflow wiring keeps the variable flow explicit and small:

- `lint-and-select-tests` reads both files in one step after checkout.
- `main_commit` is validated as a 7-40 character hex commit and used for the vLLM checkout plus selected-test matrix.
- `release_tag` is validated as a vLLM release tag and used as the second selected-test matrix entry.
- `/e2e` command selected tests in `.github/workflows/pr_e2e_command.yml` use the same `main_commit` / `release_tag` outputs, so they no longer keep a separate hardcoded `vllm_version` list.
- `docs/source/conf.py` directly reads the same anchor files for `main_vllm_commit` and `main_vllm_tag`.

This intentionally avoids a separate helper job. Dockerfile defaults and scheduled workflow hardcode cleanup are split into follow-up PRs.

No. This is CI/tooling and docs metadata plumbing only.

Local checks:

- `python docs/source/conf.py`
- `python -m ruff check docs/source/conf.py`
- `python -m ruff format --check docs/source/conf.py`
- `actionlint .github/workflows/pr_test.yaml .github/workflows/pr_e2e_command.yml`
- `git diff --check`

Conflict resolution:

- Rebased onto latest `vllm-project:main`.
- Preserved the latest `lint-and-select-tests` source-filter flow.
- Added the missing `trigger-selected-tests` `vllm_version` wiring in `.github/workflows/pr_e2e_command.yml`.

- vLLM version: v0.21.0


tools/bisect_helper.py
```diff
@@ -233,15 +234,42 @@ def get_ascend_install_for_test_cmd(test_cmd: str) -> str:
...
             content = subprocess.check_output(
                 ["git", "show", f"{ref}:{rel_path}"], text=True, stderr=subprocess.DEVNULL
             )
         except subprocess.CalledProcessError:
             return None
+    elif ref:
+        return None
     else:
         try:
             content = Path(yaml_path).read_text()
```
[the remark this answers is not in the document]
robellliu-dev: Addressed in bc364d3: ref-based lookup now stays fully git-based, returns None when repo_root cannot be resolved, and local fallback uses repo_root when available.

.github/vllm-main-verified.commit
```diff
@@ -0,0 +1 @@
+1ac10f159a09897baada01b14b6a0dd6442aefd6
```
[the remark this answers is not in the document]
robellliu-dev: Updated the PR title to `[CI][Feature] Add vLLM main verified commit anchor` and refreshed the PR summary/test section.

.github/workflows/pr_test_light.yaml
```diff
@@ -38,10 +38,25 @@ concurrency:
...
+      commit: ${{ steps.vllm.outputs.commit }}
+      matrix: ${{ steps.vllm.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v6
+      - name: Read verified vLLM main commit
+        id: vllm
+        run: |
+          commit="$(tr -d '[:space:]' < .github/vllm-main-verified.commit)"
+          echo "commit=${commit}" >> "$GITHUB_OUTPUT"
+          echo "matrix=[\"${commit}\",\"v0.20.2\"]" >> "$GITHUB_OUTPUT"
```
wjunLu: Please move `v0.20.2` into the same xx.commit, too

aflyingto (commented):
Thanks for centralizing the upstream vLLM main commit into `.github/vllm-main-verified.commit`. I found two remaining hard-coded paths that look like they can drift from the new anchor:

1. `.github/workflows/pr_test_full.yaml:118`

The comment-triggered E2E matrix still hard-codes the vLLM main commit:

```yaml
vllm_version: [39910f2b25aacc09f5e7f166cdf0030b19f8b9e8,v0.20.2]
```

Since `/e2e` comment runs are another path for testing the same upstream vLLM main baseline, this will keep using the stale commit after `.github/vllm-main-verified.commit` is advanced. Could this job also consume the `vllm-main-verified` output, the same way `e2e-test` now does?

2. `.github/workflows/dockerfiles/Dockerfile.lint:30` / `.github/workflows/schedule_lint_image_build.yaml`

There is still a hard-coded `ARG VLLM_COMMIT=39910...` in the lint image Dockerfile, and the lint image build workflow does not pass a build arg from `.github/vllm-main-verified.commit`. When the anchor advances, the scheduled/manual lint image rebuild can still bake dependencies from the old vLLM commit, while `_pre_commit.yml` checks out the new commit, which can create a mismatched lint/mypy environment. Could the image build read the anchor file and pass it as `VLLM_COMMIT`, and include `.github/vllm-main-verified.commit` in the workflow trigger paths?

aflyingto (commented):
One more hard-coded baseline remains for the release side of the same matrix: both PR workflows still embed `v0.20.2` directly when constructing the matrix:

- `.github/workflows/pr_test_light.yaml:53`
- `.github/workflows/pr_test_full.yaml:56`

```yaml
echo matrix=["${commit}","v0.20.2"] >> $GITHUB_OUTPUT
```

This is very similar to the commit hard-coding issue fixed by this PR: once the verified release tag needs to move, maintainers still have to update multiple workflow files. Could the release tag be centralized as well, for example in a small `.github/vllm-main-verified.tag` file or by reading the existing `main_vllm_tag` value from `docs/source/conf.py`, and then use that value when building the PR light/full matrices?

.github/workflows/pr_test_light.yaml
robellliu-dev: Addressed in a7efb7c: moved the matched release tag into .github/vllm-release-verified.commit and updated the PR/full/scheduled E2E matrices to read both verified refs from the .commit anchor files. I also removed the same hardcoded refs from Dockerfile.lint, the root Dockerfile variants, and docs/source/conf.py.

docs/source/conf.py
```diff
@@ -64,6 +64,18 @@
 myst_enable_extensions = ["colon_fence", "amsmath", "dollarmath", "substitution"]
 
 # Change this when cut down release
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+_VLLM_MAIN_VERIFIED_COMMIT_PATH = _REPO_ROOT / ".github" / "vllm-main-verified.commit"
+_VLLM_RELEASE_TAG_PATH = _REPO_ROOT / ".github" / "vllm-release-tag.commit"
+_VLLM_MAIN_VERIFIED_COMMIT = (
```
wangxiyuan: just `_VLLM_MAIN_VERIFIED_COMMIT =_ VLLM_MAIN_VERIFIED_COMMIT_PATH.read_text(encoding="utf-8").strip()` is enough IMO

.github/workflows/pr_test.yaml
```diff
@@ -41,7 +41,34 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  vllm-verified-refs:
```
wangxiyuan: this new job is totally useless IMO.  Just add a new step in `lint-and-select-tests` is enough IMO.

zhao-stack:
In the latest vllm-ascend version, the vllm_version configuration item is also added to trigger-selected-tests in pr_e2e_command.yml. This PR needs to be supplemented.

.github/workflows/pr_test.yaml
robellliu-dev: delete the new job ,and add the logic in the step of "lint-and-select-tests"

 if: ${{ needs.lint-and-select-tests.outputs.has_tests == 'true' && contains(github.event.pull_request.labels.*.name, 'ready') }}
    strategy:
      matrix:
        vllm_version:
          - ${{ needs.lint-and-select-tests.outputs.main_commit }}
          - ${{ needs.lint-and-select-tests.outputs.release_tag }}

robellliu-dev:
> In the latest vllm-ascend version, the vllm_version configuration item is also added to trigger-selected-tests in pr_e2e_command.yml. This PR needs to be supplemented.

upgraded the vllm commit version and add it into pr_e2e_command.yml as well

docs/source/conf.py
robellliu-dev: have modified it already

wangxiyuan (approved):

PR #9809, opened 2026-06-01
[Feature][P/D] Mooncake Connector Support Hybrid PCP/DCP for QWen3.5

zzzzzmeng:
Support Mooncake PD disaggregation KV transfer for hybrid attention models when PCP is enabled. Split attention KV by context-parallel shards while transferring Mamba state only from the final shard, and keep remote port completion accounting for hybrid transfers.

No user-facing API change.

Unit tests added/updated for Mooncake connector metadata split and group pull construction.

- vLLM version: v0.20.2


linfeng-yuan (approved):
LGTM

PR #9826, opened 2026-06-01
[BugFix][Ops] Fix ACL dtype mapping table

maoxx241:

This PR syncs the torch `ScalarType` to `aclDataType` table in `csrc/aclnn_torch_adapter/op_api_common.h` with the PyTorch 2.10 scalar enum layout used by the current main branch and the CANN 9.0 `aclDataType` enum set.

The main fix is keeping table indices aligned with `at::ScalarType` and mapping low-precision dtypes when CANN 9.0 provides corresponding ACL data types:
- `Float8_e5m2` -> `ACL_FLOAT8_E5M2`
- `Float8_e4m3fn` -> `ACL_FLOAT8_E4M3FN`
- `Float8_e8m0fnu` -> `ACL_FLOAT8_E8M0`
- `Float4_e2m1fn_x2` -> `ACL_FLOAT4_E2M1`
- `UInt1` -> `ACL_UINT1`
- `Int4` -> `ACL_INT4`

Torch-only scalar types without matching CANN 9.0 ACL enum values, such as FP8 `fnuz`, Bits types, and other sub-byte integer widths, remain mapped to `ACL_DT_UNDEFINED`.

No public API change. It fixes internal dtype conversion for custom torch binding paths using supported low-precision tensors.

- `git diff --check`
- `PATH=/tmp/vllm-ascend-lint-venv/bin:$PATH bash format.sh ci`

Build was not run locally because this macOS checkout does not have torch/CANN installed.

- vLLM version: v0.20.2

### DIFF
--- a/csrc/aclnn_torch_adapter/op_api_common.h
+++ b/csrc/aclnn_torch_adapter/op_api_common.h
@@ -86,7 +86,7 @@ extern thread_local int g_hashOffset;
   _(at::ScalarType::Half, ACL_FLOAT16)              \
   _(at::ScalarType::Float, ACL_FLOAT)               \
   _(at::ScalarType::Double, ACL_DOUBLE)             \
-  _(at::ScalarType::ComplexHalf, ACL_DT_UNDEFINED)  \
+  _(at::ScalarType::ComplexHalf, ACL_COMPLEX32)     \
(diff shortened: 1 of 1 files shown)


linfeng-yuan (approved):
LGTM

PR #9830, opened 2026-06-01
[Performance] Optimize split_qkv_tp_rmsnorm_rope with grid stride andh loading

He1pa:
Refactor Triton kernel to use Grid Stride Loop with batch tensor loading instead  
  of per-token processing. Pre-compute reciprocals to avoid division in kernel.     
  Reduce BLOCK_SIZE configs to 1-4 to prevent UB overflow on large hidden_size.     
                                                                                    

                                                                                    
  This PR refactors `_split_qkv_and_compute_local_qk_var_kernel` in
  `split_qkv_tp_rmsnorm_rope.py` with the following improvements:

  - **Grid Stride Loop + batch loading**: Replaces the per-token scalar loading loop
   with a Grid Stride Loop that loads `BLOCK_SIZE` tokens at a time using 2D tensor
  loading. This improves memory access efficiency on the NPU.
image

  - **Precomputed reciprocals**: Computes `q_inv_size = 1.0 / q_cols` and
  `k_inv_size = 1.0 / k_cols` on the host side and passes them as compile-time
  constants, avoiding division operations inside the Triton kernel.
  - **Reduced BLOCK_SIZE configs**: Limits autotune `BLOCK_SIZE` options to 1-4
  (from the previous BLOCK 64-256) to prevent undefined behavior overflow for models
   with large hidden sizes.

  No user-facing change. This is a pure kernel optimization with equivalent
  numerical results.

  CI passed with existing tests. Kernel produces numerically equivalent output to
  the previous implementation.

  ### Performance

  Tested on minimax-m2.5-w8a8-quarot with TP=8 prefill node (PD disaggregation):

  | Kernel | Before | After | Speedup |
  |--------|--------|-------|---------|
  | `_split_qkv_and_compute_local_qk_var_kernel` | ~0.67ms | ~0.134ms | **~5x** |

  Test setup: `max-num-seqs=320`, `max-num-batched-tokens=32768`, bfloat16,
  CANN 8.5.0, FlashComm1 enabled, NZ=2.

- vLLM version: v0.22.1


ZT-AIA (approved):

vllm_ascend/ops/triton/linearnorm/split_qkv_tp_rmsnorm_rope.py
```diff
@@ -28,9 +28,9 @@
```
linfeng-yuan, line 27: Please add a comment TODO here around UB difference between different chips and whether we can dynamically tune this parameter with formula

linfeng-yuan (approved):

PR #9833, opened 2026-06-01
[Feature] add ascendc ops store_kv_block

ZT-AIA:
add ascendc ops store_kv_block
Control whether to use the operator by using additional-config '{"enable_sparse_c8":true,"c8_enable_reshape_optim":true}'. It is only used on the P node.

1. The operator supports input graphs.

By enabling c8_enable_reshape_optim on the P node

nightly custom op

- vLLM version: v0.21.0


vllm_ascend/envs.py
```diff
@@ -112,6 +112,8 @@
     "VLLM_ASCEND_ENABLE_BATCH_MEMCPY": lambda: os.getenv("VLLM_ASCEND_ENABLE_BATCH_MEMCPY", None),
     # Whether to use MultiBlockPool for KV cache management
     "VLLM_ASCEND_APPLY_DSV4_PATCH": lambda: bool(int(os.getenv("VLLM_ASCEND_APPLY_DSV4_PATCH", "0"))),
+    # DeepSeek V3.2 only on P product on
```
zzzzwwjj: When adding feature level switches, add a new env is not allowed. Please add it in `additional_config`, and modify the documentation synchronously.

vllm_ascend/attention/sfa_v1.py
```diff
@@ -238,6 +243,10 @@ def build(
         slot_mapping = common_attn_metadata.slot_mapping[:num_input_tokens]
         input_positions = common_attn_metadata.positions[:num_input_tokens].long()
 
+        block_size = 128
+        if ascend_envs.VLLM_ASCEND_ENABLE_RESHAPE_OPTIM:
+            slot_mapping_cpu = slot_mapping.cpu()
```
zzzzwwjj: Will this code occur async-scheduling interrupt?

zzzzwwjj (changes requested):

vllm_ascend/envs.py
ZT-AIA: It has been modified to additional config.

vllm_ascend/attention/sfa_v1.py
ZT-AIA: It has been modified, and I have confirmed it with Liu Rong.

csrc/attention/store_kv_block/store_kv_block_torch_adpt.h
```diff
@@ -0,0 +1,134 @@
...
+            length.reserve(new_capacity);
+            key_idx.reserve(new_capacity);
+            key_cache_idx.reserve(new_capacity);
+
+            for (int32_t k = idx_groups; k < new_capacity; ++k){
+                length.emplace_back(0);
+                key_idx.emplace_back(0);
+                key_cache_idx.emplace_back(0);
+            } 
+        }
```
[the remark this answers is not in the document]
ZT-AIA: You only need to assign the value 0 to the newly allocated memory. The old address does not need to be modified.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3071,6 +3071,7 @@ def _get_block_table_and_slot_mapping(kv_cache_gid: int, total_num_scheduled_tok
             max_seq_len=max_seq_len,
             block_table_tensor=block_table_gid_0,
             slot_mapping=slot_mapping_gid_0,
+            slot_mapping_cpu=torch.from_numpy(self.cpu_slot_mapping) if self.cpu_slot_mapping is not None else None,
```
Ronald1995: why use torch.from_numpy here, it seems like `self.cpu_slot_mapping` is actually a tesnor.

zzzzwwjj (approved):

PR #9857, opened 2026-06-02
[BugFix][Worker] Fix embedding TP deadlock with multimodal models

coder-fny:

When `finegrained_tp_config.embedding_tensor_parallel_size` is enabled with multimodal models (e.g. kimi-k2.5), the embedding TP `all_gather`/`reduce_scatter` collectives deadlock. Root causes:

1. `_preprocess` mm branch calls `embed_input_ids` with `num_scheduled_tokens` (varies per DP rank) → embedding TP `all_gather` receives mismatched tensor sizes → HCCL deadlock
2. `_dummy_run` skips `embed_input_ids` entirely → embedding TP group has ranks doing different operations → HCCL deadlock

## Changes

All changes in `vllm_ascend/worker/model_runner_v1.py`:

| # | Location | Change | Purpose |
|---|----------|--------|---------|
| 1 | `_sync_metadata_across_dp` | Add `needs_dp_sync_for_finegrained_tp = embedding_tp_enable()`, prevent skipping DP all_reduce when true | All DP ranks agree on padding |
| 2 | `allow_dp_padding` | Add `or embedding_tp_enable()` | Force DP padding in eager mode for uniform `num_input_tokens` |
| 3 | `_preprocess` | Override `total_num_scheduled_tokens = num_input_tokens` before calling `embed_input_ids`; zero out positions for padding tokens after restore | All DP ranks in same embedding TP group produce identical-sized tensors for collectives |
| 4 | `_dummy_run` | Call `embed_input_ids(torch.zeros(...))` with a zero tensor when `embedding_tp_enable()` | Warmup embedding TP HCCL communication, avoid uninitialized buffer causing OOB embedding lookup |

## CI

Added unit tests in `tests/ut/worker/test_embedding_tp_mm.py`:

| # | Test | Purpose |
|---|------|---------|
| 1 | `test_sync_metadata_does_not_skip_dp_allreduce_with_embedding_tp` | Verify DP all_reduce is not skipped when embedding TP enabled |
| 2 | `test_sync_metadata_skips_dp_allreduce_without_embedding_tp` | Verify DP all_reduce is skipped when no embedding TP |
| 3 | `test_preprocess_overrides_num_scheduled_tokens_with_embedding_tp` | Verify `total_num_scheduled_tokens` is overridden to `num_input_tokens` |
| 4 | `test_preprocess_restores_num_scheduled_tokens_after_call` | Verify original value is restored after `_preprocess` |
| 5 | `test_preprocess_zeros_positions_for_padding_tokens` | Verify padding positions are zeroed out |
| 6 | `test_preprocess_does_not_override_without_embedding_tp` | Verify no override when embedding TP disabled |
| 7 | `test_preprocess_does_not_override_for_non_first_pp_rank` | Verify no override for non-first PP ranks |
| 8 | `test_preprocess_does_not_override_for_encoder_decoder` | Verify no override for encoder-decoder models |
| 9 | `test_preprocess_does_not_override_for_non_mm_model` | Verify no override for text-only models |

Run: `pytest -sv tests/ut/worker/test_embedding_tp_mm.py`

No
- Model: kimi-k2.5 (W4A8 quantization, Eagle3 speculative decoding)
- Config: DP=32, TP=1, EP=32, `embedding_tensor_parallel_size=8`
- Hardware: 2×A3 (16 NPU each)
- Graph mode: `FULL_DECODE_ONLY` (ACL graph)
- Both eager mode (`--enforce-eager`) and graph mode verified

- vLLM version: v0.23.0


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1441,11 +1443,32 @@ def _preprocess(
...
+        need_mm_embed_override = (
+            embedding_tp_enable()
+            and self.supports_mm_inputs
+            and get_pp_group().is_first_rank
+            and not self.model_config.is_encoder_decoder
+        )
+        saved_num_scheduled = None
+        if need_mm_embed_override:
+            saved_num_scheduled = scheduler_output.total_num_scheduled_tokens
+            scheduler_output.total_num_scheduled_tokens = num_input_tokens
```
maoxx241, line 1461: This override looks right for the normal MM path, but I think the PCP case still needs a guard or padding fix. When `pcp_size > 1`, our `_gather_mm_embeddings()` override builds `is_mm_embed` from `local_num_scheduled_tokens` instead of this overridden `num_input_tokens`. That can leave `input_ids[:num_input_tokens]` longer than `is_multimodal`, which will fail when `embed_input_ids()` merges multimodal embeddings. Could we either explicitly reject PCP + MM + embedding TP for now, or pad the PCP `is_mm_embed` mask to `num_input_tokens` and cover it with a unit test?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1441,11 +1443,32 @@ def _preprocess(
...
         try:
-            return super()._preprocess(
+            result = super()._preprocess(
                 scheduler_output, num_input_tokens, intermediate_tensors
             )
         finally:
+            if saved_num_scheduled is not None:
+                scheduler_output.total_num_scheduled_tokens = saved_num_scheduled
+                if num_input_tokens > saved_num_scheduled:
+                    self.positions[saved_num_scheduled:num_input_tokens].zero_()
```
maoxx241, line 1471: This clears the plain position buffer, but M-RoPE/XD-RoPE models return `self.mrope_positions.gpu[:, :num_input_tokens]` or `self.xdrope_positions.gpu[:, :num_input_tokens]` from the parent path. For Qwen-VL/Hunyuan-VL-style models, the padded tail can still contain stale positions. Could we clear the matching position buffer based on `uses_mrope` / `uses_xdrope_dim` here?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3567,6 +3592,11 @@ def _dummy_run(
             # Make sure padding doesn't exceed max_num_tokens
             assert num_tokens_padded <= self.max_num_tokens
             if self.supports_mm_inputs and not self.model_config.is_encoder_decoder or self.enable_prompt_embeds:
+                if embedding_tp_enable():
```
maoxx241, line 3595: This warmup call should probably be gated to the first PP rank, matching the real `_preprocess()` embedding path. Non-first PP ranks may not own the embedding module, and calling this there can also make the dummy communication path differ from real execution. Something like `embedding_tp_enable() and get_pp_group().is_first_rank` would make this narrower.

tests/ut/worker/test_embedding_tp_mm.py
```diff
@@ -0,0 +1,305 @@
...
+        runner.pcp_size = 1
+        return runner
+
+    @patch("vllm_ascend.worker.model_runner_v1.embedding_tp_enable", return_value=True)
+    @patch("vllm_ascend.worker.model_runner_v1.get_pp_group")
+    @patch("vllm_ascend.worker.model_runner_v1.should_skip_allreduce_across_dp_group", return_value=True)
+    def test_sync_metadata_does_not_skip_dp_allreduce_with_embedding_tp(self, mock_skip, mock_pp_group, mock_embed_tp):
+        """When embedding TP is enabled, DP all_reduce must not be skipped
+        even if should_skip_allreduce_across_dp_group returns True."""
+        runner = self._build_runner(embedding_tp=True)
```
maoxx241, line 42: This test currently enters the non-skip branch of `_sync_metadata_across_dp()`, but the synthetic runner does not set `runner.ascend_config`. The production code reads `self.ascend_config.dp_allreduce_on_npu` before calling `dist.all_reduce`, so the test can fail before it verifies the intended behavior. Could we set a minimal `ascend_config` mock in `_build_runner()` and assert the returned padding shape/value as well?

tests/ut/worker/test_embedding_tp_mm.py
```diff
@@ -0,0 +1,305 @@
...
+        # Capture the value at call time, since the finally block restores it
+        captured_scheduled = []
+
+        def capture_and_return(*args, **kwargs):
+            captured_scheduled.append(args[1].total_num_scheduled_tokens)
+            return (None, None, None, None, None, None)
+
+        with (
+            patch.object(GPUModelRunner, "_preprocess", side_effect=capture_and_return, autospec=True),
+            contextlib.suppress(Exception),
```
maoxx241, line 112: Could we avoid suppressing all exceptions here? These tests are meant to prove that the parent preprocess sees the padded token count and that state is restored, but `contextlib.suppress(Exception)` can hide failures before the parent path is exercised. A stricter mock parent `_preprocess()` plus explicit `assert_called_once()` / captured-shape assertions would give this regression test much better signal.

maoxx241 (changes requested):
Thanks for working on this. The overall direction makes sense to me, especially aligning the multimodal embedding path with the DP-padded token count for embedding TP. I left a few focused comments for edge cases and test signal that I think should be handled before merging.

PR #9870, opened 2026-06-02
[BugFix][v0.20.2rc]: NPU MoE quantization methods support TP only.

Fishermanykx:
This PR fixes the MoE (Mixture of Experts) quantization method initialization failure on Ascend NPU when **only Tensor Parallelism is enabled** without Expert Parallelism. https://github.com/vllm-project/vllm-ascend/pull/8532 only fix non-quantized cases.
- Remove unused `self.ep_group = get_ep_group()` initialization from MoE quantization methods that do not use the EP group.
- Use `get_tensor_model_parallel_world_size()` directly for W4A8 TP checks instead of deriving TP size from the EP group.
- This aligns with vLLM upstream's convention (`ep_size == 1` means no EP) and keeps the quantization code robust across TP-only MoE deployments.

No. This is a bug fix that restores expected behavior for TP-only MoE quantized inference on NPU. No API, CLI flag, or model interface is changed.

- [x] Verified that MoE quantization methods no longer unconditionally access `get_ep_group()` during initialization.
- [x] Verified that `enable-expert-parallel` paths remain intact and still avoid TP-size checks that only apply to TP-only mode.
- [x] `python -m compileall -q vllm_ascend/quantization/methods/w4a4_mxfp4.py vllm_ascend/quantization/methods/w4a8.py vllm_ascend/quantization/methods/w8a8_dynamic.py vllm_ascend/quantization/methods/w8a8_mxfp8.py tests/ut/quantization/methods/test_w4a4_mxfp4.py tests/ut/quantization/methods/test_w4a8.py tests/ut/quantization/methods/test_w8a8_dynamic.py tests/ut/quantization/methods/test_w8a8_mxfp8.py tests/ut/quantization/methods/test_w8a8_pdmix.py`
- [x] `git diff --check`
- [ ] `pytest` not run locally because the environment is missing the `regex` Python dependency.

- vLLM version: v0.20.2rc
- vLLM main:


Fishermanykx:
> This pull request has conflicts, please resolve those before we can evaluate the pull request.

resolved

weijinqian0 (approved):

weijinqian0 (approved):

PR #9883, opened 2026-06-02
[Performance][SFA][CP] Optimize SFA decode by avoiding KV all-gather

xiaocongtou6:
  This PR optimizes the SFA decode path under PCP/DCP by avoiding the cross-CP KV all-gather before `lightning_indexer`
  and `sparse_flash_attention`.

  Instead of gathering full KV on every rank, each rank now computes with its local KV cache:
  - `lightning_indexer` returns local topk indices and scores.
  - Topk scores are gathered across CP ranks to select the global topk.
  - `sparse_flash_attention` returns partial attention output with softmax statistics.
  - The final decode attention output is merged across CP ranks using softmax LSE correction.

  This keeps the decode result mathematically equivalent to the full-KV path while reducing KV communication and per-
  rank KV memory traffic in CP decode.

- vLLM version: v0.20.2


linfeng-yuan:
Please validate your PR with GLM5 on GPQA and swe-bench, and let @ZYang6263 sees the results.

PR #9890, opened 2026-06-02
[CI] Add ninja incremental build cache for csrc compilation

AuroraEmiya:
     
  Introduce a ninja incremental build cache (`csrc/build`) shared across CI workflows to
  accelerate C++ custom-op compilation.

  **Background:** vllm-ascend ships custom Ascend kernels under `csrc/` compiled via CMake.
  Previously the CI used a hash-based product cache (SHA256 of all csrc files → exact-match
  `.so` artifacts), which had a near-zero hit rate — any single-file change invalidated
  the entire cache, forcing full recompilation on almost every PR.

  **This PR** replaces that with a ninja intermediate build cache:

  - Ninja's own dependency tracking detects changed files and only recompiles the diff,
    so even PRs with csrc modifications get partial cache hits.
  - Single producer (`schedule_cache_csrc_build.yaml`, daily cron at 23:59 UTC+8) builds
    across 2 arch × 2 OS × 3 chip_type = 12 matrix jobs, saves `csrc/build` to GHA cache.
  - Two consumers (`_e2e_test.yaml` — 8 E2E jobs, `_optional_smart_e2e.yaml`) restore
    `csrc/build` before `pip install -e .`, ninja handles the rest automatically.
  - All three files share a unified cache key:
    `vllm-ascend-build-ninja-v1-{base_branch}-{os}-{arch}-{image}-{cache_hour}`

  **Commit history on this branch:**

  | Commit | Description |
  |--------|-------------|
  | `ac6db1d` | build: support ninja incremental build for custom ops — rewrote `csrc/build.sh` / `build_aclnn.sh` /
  cmake rules |
  | `81b1278` | [BugFix][Feature] add ninja incremental cache to `_e2e_test.yaml`, fix `build_aclnn.sh` |
  | `3d27a66` | [CI] unify ninja cache across all workflows, remove hash-based cache, align keys, fix actionlint |

  **Files changed in this PR (vs main):**

  | File | Status | Delta |
  |------|--------|-------|
  | `csrc/build.sh` | Modified | +42/−18 |
  | `csrc/build_aclnn.sh` | Modified | +20/−11 |
  | `csrc/cmake/custom_build.cmake` | Modified | +2/−0 |
  | `csrc/cmake/third_party/ascend_protobuf.cmake` | Modified | +76/−100 |
  | `.github/workflows/_e2e_test.yaml` | Modified | +300/−270 |
  | `.github/workflows/_optional_smart_e2e.yaml` | Modified | +47/−20 |
  | `.github/workflows/schedule_cache_csrc_build.yaml` | New | +188 |
  | `.github/workflows/push_build_csrc_cache.yaml` | Deleted | −184 |

  No.

  - `pre-commit` all hooks passed (actionlint, codespell, typos)
  - Cache key components verified byte-for-byte across all 3 workflow files
  - Runtime dependency audit: `ninja-build`, `clang-15` confirmed in all compilation paths
  - Build tested locally with `bash csrc/build.sh` using ninja generator

- vLLM version: v0.21.0


.github/workflows/schedule_cache_csrc_build.yaml
```diff
@@ -126,44 +121,44 @@ jobs:
...
 
-      - name: Get csrc hash
-        id: get_csrc_hash
+      - name: Get current cache hour
+        id: get_cache_time
         run: |
-          CSRC_HASH=$(find ./csrc ./setup.py ./CMakeLists.txt ./cmake \
-            -type f -not -path '*/.*' | sort | xargs sha256sum | sha256sum | awk '{print $1}')
-          echo "CSRC_HASH=$CSRC_HASH" >> $GITHUB_OUTPUT
+          cache_time=$(date -u +'%Y%m%d%H')
```
menogrey, line 128: Is this means that every hour we should refresh our cache?  PR in 12:01 reuse caches created in 11:01 seems also reasonable.

QwertyJack:
Thanks for working on this. Using Ninja plus a shared `csrc/build` cache is a promising direction for CI acceleration, but I think there are a few correctness issues to address before relying on this as the incremental custom-op build path.

1. Existing `csrc/build` may have a different CMake generator.

   This PR preserves `csrc/build` and defaults to Ninja whenever `ninja` is available. For users or CI caches that already have a `csrc/build` configured with `Unix Makefiles`, the next configure with `-G Ninja` will fail with CMake's generator mismatch error. The script should detect the cached `CMAKE_GENERATOR` and drop/reconfigure `csrc/build` when it does not match the selected generator.

2. The cached build directory still needs environment invalidation.

   `csrc/build/CMakeCache.txt` records values such as `CUSTOM_ASCEND_CANN_PACKAGE_PATH`. If the CANN package path/version changes, preserving the old build directory can keep stale include/tool paths and cause mixed-version builds. For CANN specifically, I think the simple behavior should be: if the cached CANN path differs from the current one, delete `csrc/build` and let CMake configure from a clean cache.

3. Ninja only works as well as the dependency graph CMake gives it.

   This PR enables Ninja, but it does not appear to update the custom-op CMake rules that produce copied sources, generated dynamic Python files, compile scripts, and `.done` stamps. If those custom commands do not declare their real inputs via `DEPENDS`, Ninja may correctly decide that the declared output is up to date while an actual source/script changed. The dependency fixes in #9364 are aimed at this part of the problem and should probably be incorporated or reconciled here.

4. There are still generator-specific clean calls.

   I still see `make clean` in `csrc/build.sh`. With a Ninja-generated build directory this should be replaced with a generator-neutral command, for example `cmake --build . --target clean`, or skipped in the incremental path where preserving the build tree is intentional.

So I see this PR as complementary to #9364 rather than a replacement: this PR can improve CI cache reuse and use Ninja for speed, while #9364 handles the correctness side of incremental custom-op rebuilds. It would be safer to combine the two ideas with explicit cache invalidation for generator/CANN changes and complete CMake dependencies for generated custom-op artifacts.

.github/workflows/schedule_cache_csrc_build.yaml
```diff
@@ -126,44 +121,44 @@ jobs:
...
+        uses: runs-on/cache/restore@v5
         with:
           path: |
-            vllm_ascend/_cann_ops_custom
-            vllm_ascend/*.so
-            vllm_ascend/lib
-            vllm_ascend/include
-          key: vllm-ascend-build-v1-${{ matrix.arch }}-${{ steps.get_image_tag.outputs.IMAGE_TAG }}-${{ steps.get_csrc_hash.outputs.CSRC_HASH }}
+            csrc/build
+          key: vllm-ascend-build-ninja-v1-${{ steps.get_csrc_build_cache_base_branch.outputs.base_branch }}-${{ runner.os }}-${{ matrix.arch }}-${{ matrix.image }}-${{ steps.get_cache_time.outputs.cache_time }}
```
menogrey: I think we only need to keep the latest cache of each branches. And when needing build new caches in installing vllm-ascend, we just refresh them. That's simple, and we don't need to take the key name into concern.

menogrey: 1. key: main-lastest
2. install vllm-ascend, when csrc cache changed, use a file as a flag.
3. upload cache if the file flag exist.

menogrey: But wait, consider this situation:

main with cache-1:
PR-A at 12:01: modify csrc code -> create new cache-2 -> upload at 12:10
PR-B at 12:30: don't modify any csrc code -> should reuse the main cache-1

Will the PR-A take effect on the PR-B cache even if the PR is not merged? This will diff from the original cache impl.

The key is the original use file as a cache key, although it is not precise as you said, but can ensure correctness.

menogrey: The comment from @QwertyJack provided many insightful suggestions. From a bigger perspective, the cache GHA is less reliable than ccache or bazel, which take every inputs of the compiler command as parts of the Key , including gcc, header files, source files, etc.

.github/workflows/schedule_cache_csrc_build.yaml
AuroraEmiya: In fact we schedule daily at 23:59 . The cache_time is only for record as a part of the key.  
```  schedule:
    # UTC+8: 23：59pm every day
    - cron: '59 15 * * *'
 ```
The time is scheduled at line 23 in schedule_cache_csrc_build.yaml.

File: .github/workflows/schedule_cache_csrc_build.yaml (hunk 2 above)
Comment (AuroraEmiya) [reply .github/workflows/schedule_cache_csrc_build.yaml]: > I think we only need to keep the latest cache of each branches. And when needing build new caches in installing vllm-ascend, we just refresh them. That's simple, and we don't need to take the key name into concern.

Yes ,in fact we should keep in pace with the latest cache of eache branches. However ,the **main branch is flooded** with tons of PR in a day and  the compilation env will change daily or hourly. If we refresh the cache per PR , the storage pace of cache will be eaten up,which is only 10GB as far as I am concerned. On the other hand , as a part of stability , we should use a reliable version that is not so far from the HEAD ,that 's why we set it daily.

Comment (AuroraEmiya) [reply .github/workflows/schedule_cache_csrc_build.yaml]: > 1. key: main-lastest
> 2. install vllm-ascend, when csrc cache changed, use a file as a flag.
> 3. upload cache if the file flag exist.

Ninja would skip the part that is labeled with "done". This PR is a tradeoff between The challenge of storage space and The challenge of frequency. And I think it's acceptable when the main branch is added or pushed with a few changes inside the csrc where this PR mainly focuses in a period like a day.

Comment (AuroraEmiya) [reply .github/workflows/schedule_cache_csrc_build.yaml]: > But wait, consider this situation:
> 
> main with cache-1: PR-A at 12:01: modify csrc code -> create new cache-2 -> upload at 12:10 PR-B at 12:30: don't modify any csrc code -> should reuse the main cache-1
> 
> Will the PR-A take effect on the PR-B cache even if the PR is not merged? This will diff from the original cache impl.
> 
> The key is the original use file as a cache key, although it is not precise as you said, but can ensure correctness.

Yes , it will take place whenever a PR hits the key exactly or with prefix that ignores the time stamp.
Let me take an example 
PR 1 : csrc/A B : Luckily scheduled and cached with key  env + time_stamp so we have A B in build/
PR 2 : csrc/A-new B : The same env and new A , at first not hit the exact key with different time_stamp but can restore the PR 1 build dir with the same prefix "env" since we have two rules to match the build/ cache
```
key: vllm-ascend-build-ninja-v1-${{ steps.get_csrc_build_cache_base_branch.outputs.base_branch }}-${{ runner.os }}-${{ steps.get_csrc_build_cache_arch.outputs.cache_arch }}-${{ env.CSRC_BUILD_CACHE_IMAGE }}-${{ steps.get_cache_time.outputs.cache_time }}
restore-keys: |
         vllm-ascend-build-ninja-v1-${{ steps.get_csrc_build_cache_base_branch.outputs.base_branch }}-${{ runner.os }}-${{ steps.get_csrc_build_cache_arch.outputs.cache_arch }}-${{ env.CSRC_BUILD_CACHE_IMAGE }}-
```
where the first hits with the same key and the other hits when prefix matches.
Result : skip unchanged B and rebuild A (not refresh until next schdule time)
PR 3 : csrc/A(merged PR 2) B-new , hits with prefix match but rebuild both because we only stored the PR 1 build dir.

AuroraEmiya:
> Thanks for working on this. Using Ninja plus a shared `csrc/build` cache is a promising direction for CI acceleration, but I think there are a few correctness issues to address before relying on this as the incremental custom-op build path.
> 
> 1. Existing `csrc/build` may have a different CMake generator.
>    This PR preserves `csrc/build` and defaults to Ninja whenever `ninja` is available. For users or CI caches that already have a `csrc/build` configured with `Unix Makefiles`, the next configure with `-G Ninja` will fail with CMake's generator mismatch error. The script should detect the cached `CMAKE_GENERATOR` and drop/reconfigure `csrc/build` when it does not match the selected generator.
> 2. The cached build directory still needs environment invalidation.
>    `csrc/build/CMakeCache.txt` records values such as `CUSTOM_ASCEND_CANN_PACKAGE_PATH`. If the CANN package path/version changes, preserving the old build directory can keep stale include/tool paths and cause mixed-version builds. For CANN specifically, I think the simple behavior should be: if the cached CANN path differs from the current one, delete `csrc/build` and let CMake configure from a clean cache.
> 3. Ninja only works as well as the dependency graph CMake gives it.
>    This PR enables Ninja, but it does not appear to update the custom-op CMake rules that produce copied sources, generated dynamic Python files, compile scripts, and `.done` stamps. If those custom commands do not declare their real inputs via `DEPENDS`, Ninja may correctly decide that the declared output is up to date while an actual source/script changed. The dependency fixes in [[Ops][BugFix] Make custom ops build incremental #9364](https://github.com/vllm-project/vllm-ascend/pull/9364) are aimed at this part of the problem and should probably be incorporated or reconciled here.
> 4. There are still generator-specific clean calls.
>    I still see `make clean` in `csrc/build.sh`. With a Ninja-generated build directory this should be replaced with a generator-neutral command, for example `cmake --build . --target clean`, or skipped in the incremental path where preserving the build tree is intentional.
> 
> So I see this PR as complementary to #9364 rather than a replacement: this PR can improve CI cache reuse and use Ninja for speed, while #9364 handles the correctness side of incremental custom-op rebuilds. It would be safer to combine the two ideas with explicit cache invalidation for generator/CANN changes and complete CMake dependencies for generated custom-op artifacts.

Thanks very much for remarkable insights. And I will also answer with sequence number below
1. Existing `csrc/build` may have a different CMake generator.
>This PR preserves `csrc/build` and defaults to Ninja whenever `ninja` is available. 

The dir of csrc/build will be generated whenever a Developer installs the project. The default action of this dir is being deleted as part of next installation procudure. What this pr does is preserve it before compilation.As for different CMake generators, it only happens when the project is built with CMake generator A and re-built with B afterwards.And I think in this situation the developer should remove the old build output and re-build. And I agree with your  `CMAKE_GENERATOR` detector idea for a near future update.
2. The cached build directory still needs environment invalidation.
This PR has a very critical assumption that the build cache is reused resently, like a local develop situation . To ensure this in different PR , We will refresh the build dir daily. And different branch and env will be split by key setting.
3.Ninja only works as well as the dependency graph CMake gives it.
It reminds me! And I will learn further about it. 
4. There are still generator-specific clean calls.
I have changed the procedure of build_aclnn.sh. I remains the clean function for others to use and actually it does  not take place in the real compilation.
Thanks again for your advices and I agree with you that we focus on different parts of installation of project , the correctness should be appreciated too! 
And I will learn further to improve the stability and correctness of this feature!

PR #9025, opened 2026-05-09
[Performance] Remove sync for PIECEWISE

yiz-liu:
### What this PR does / why we need it?
As mentioned in #4233, in `FULL` mode we need to update parameters for certain kernels. To avoid race conditions, we introduced a hard barrier before `replay`. However, this caused severe performance regressions in `PIECEWISE`, as reported in #8877 and #8854.

We did not have the bandwidth to address this for quite some time, since `PIECEWISE` generally performed worse than `FULL` and `FULL_DECODE_ONLY` in most cases, so there was little incentive to optimize it. However, as attention mechanisms become increasingly hybrid and complex, fewer new models are able to fully benefit from `FULL`, making `FULL_AND_PIECEWISE` more and more important.

So this PR revisits `PIECEWISE` in preparation for `FULL_AND_PIECEWISE`.

The change itself is relatively small, so why was this not done earlier?

We did try, more than once, in #5761 and #8354.

Each attempt appeared to introduce some accuracy issues.

And it seems I cannot reproduce those accuracy issues on the `main` branch. So I am proposing this PR to share the change and gather feedback.

This is also effectively a "help wanted" request: if you encounter any accuracy regressions with this PR, please report them here, with detailed repro guide (preferably a simple one). Thanks a lot!

### Does this PR introduce _any_ user-facing change?
None.

### How was this patch tested?
None.
- vLLM version: v0.19.1
- vLLM main: https://github.com/vllm-project/vllm/commit/4d51588e2381018348f1022dfa3a7698899805b7

PR #9896, opened 2026-06-02
[BugFix]fix DeepSeek-V4 PIECEWISE ACLGraph accuracy

panther-zhu:
## Summary

This PR fixes an accuracy issue for DeepSeek-V4 when running with PIECEWISE ACLGraph mode on Ascend.

DeepSeek-V4 DSA attention uses the Ascend custom op `vllm::dsa_forward`, but PIECEWISE splitting only added `vllm::mla_forward` as an Ascend custom attention boundary. As a result, the DSA attention path was not treated as a split boundary and could be captured inside a regular PIECEWISE graph segment, leading to incorrect outputs.

This change adds `vllm::dsa_forward` to `compilation_config.splitting_ops` together with `vllm::mla_forward`.

## Changes

- Add `vllm::dsa_forward` to PIECEWISE `splitting_ops`.

## Why

`vllm::dsa_forward` wraps the DeepSeek-V4 DSA attention path. It depends on runtime `forward_context` metadata, KV cache state, sparse attention metadata,and mutates the provided `output` tensor.

Because of these dynamic side effects, it must be handled as a PIECEWISE split boundary instead of being captured inside a normal graph segment.

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -421,7 +421,7 @@ class NPUPlatform(Platform):
             # This will cause in scenarios where both piecewise and splitting ops are configured simultaneously,
             # If splitting ops does not contain the vllm::mla forward value, this configuration issue will
             # not be detected in advance assert.
-            compilation_config.splitting_ops.extend(["vllm::mla_forward"])
+            compilation_config.splitting_ops.extend(["vllm::mla_forward", "vllm::dsa_forward"])
             update_aclgraph_sizes(vllm_config)
             ascend_config.ascend_compilation_config.enable_npugraph_ex = False
         elif (


vllm_ascend/compilation/acl_graph.py
```diff
@@ -204,10 +207,15 @@ def __call__(self, *args, **kwargs):
...
+        # Piecewise graph replay has the same async ordering risk between
+        # adjacent graph pieces and eager split ops.
         # When enable_enpu is on, model_runner orders update vs replay; skip here.
         # When FULL + EAGLE draft (merge path), replay does not need this barrier.
         is_draft_eagle = _EXTRA_CTX.is_draft_model and self.use_eagle
-        need_sync = self.runtime_mode == CUDAGraphMode.FULL and not is_draft_eagle
+        need_sync = (
+            self.runtime_mode in (CUDAGraphMode.FULL, CUDAGraphMode.PIECEWISE)
+            and not is_draft_eagle
+        )
```
yiz-liu: No, please don't add synchronization since it will cause serious performance regression. Find out why we need it and see if there is another way.

yiz-liu (changes requested):

vllm_ascend/compilation/acl_graph.py
yiz-liu: See full context in #9025 .

yiz-liu:
@panther-zhu What about the accuracy benchmark test now? Is everything OK?

PR #9897, opened 2026-06-02
[Feature] [Ops] Add swiglu_limit support in unquant_apply_mlp

Debonex:

In RL scenario, we need to run dsv4 with bf16 dtype, which has swiglu limit.

This PR adds swiglu_limit support to the unquantized MoE MLP path, keep swiglu_limit behavior consistent across MoE MLP execution paths.

No

- vLLM version: v0.23.0


realliujiaxu (approved):
LGTM

Issue #9845, opened 2026-06-01
[Bug]: 昇腾A2四机PD混部，适配DSv4优化，添加MTP后Prefix Cache无命中，去掉相关配置后，有命中


rogerlc-163-com:
### Your current environment

昇腾A2四机PD混部场景，MTP的相关配置为：

--speculative-config '{
     "num_speculative_tokens": 1,
     "method": "mtp",
     "enforce_eager": true
  }' \

叠加MTP的配置如下：
Node0配置：
local_ip="xxx"
node0_ip="xxxx"

export HCCL_IF_IP=$local_ip
export IFNAME="xxx"
export GLOO_SOCKET_IFNAME="$IFNAME"
export TP_SOCKET_IFNAME="$IFNAME"
export HCCL_SOCKET_IFNAME="$IFNAME"
export HCCL_BUFFSIZE=512
export ASCEND_RT_VISIBLE_DEVICES=0,1,2,3,4,5,6,7

export OMP_PROC_BIND=false
export OMP_NUM_THREADS=10
export PYTORCH_NPU_ALLOC_CONF=expandable_segments:True
export ACL_OP_INIT_MODE=1
export VLLM_ENGINE_READY_TIMEOUT_S=3600
export HCCL_OP_EXPANSION_MODE="AIV"

export TASK_QUEUE_ENABLE=1

export LD_PRELOAD=/usr/lib/aarch64-linux-gnu/libjemalloc.so.2:$LD_PRELOAD

export HCCL_CONNECT_TIMEOUT=7200
export ASCEND_CONNECT_TIMEOUT=10000
export ASCEND_TRANSFER_TIMEOUT=10000
export VLLM_RPC_TIMEOUT=1800000
export VLLM_ASCEND_APPLY_DSV4_PATCH=1

vllm serve /mnt/share/weight/DeepSeek-V4-Pro-w4a8-0505 \
  --host 0.0.0.0 \
  --port 10010 \
  --max_model_len 135000 \
  --max-num-batched-tokens 4096 \
  --served-model-name dsv4 \
  --gpu-memory-utilization 0.9 \
  --max-num-seqs 16 \
  --data-parallel-size 4 \
  --tensor-parallel-size 8 \
  --data-parallel-size-local 1 \
  --data-parallel-start-rank 0 \
  --data-parallel-address $node0_ip  \
  --enable-expert-parallel \
  --quantization ascend \
  --no-enable-prefix-caching \
[...]

PR #9903, opened 2026-06-02
[PrefixCache][BugFix] Fix Deepseek-V4 compress attn groups prefix caching hit

nwpu-zxr:
The kvcache caching and hash of the deepseek-v4 `compressor_kv` do not consider `compress_ratio`. This PR processes `compress_ratio` so that the prefix cache can be correctly hit.

No.

By CI.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/core/single_type_kv_cache_manager.py
+++ b/vllm_ascend/core/single_type_kv_cache_manager.py
@@ -20,8 +20,6 @@ from vllm.v1.kv_cache_interface import (
 )
 from vllm.v1.request import Request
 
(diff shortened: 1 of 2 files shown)


yiz-liu:
Fixes #9845

MengqingCao (approved):
LGTM

MengqingCao:
> Fixes #9845

FYI, this won't fix #9845, as the issue this pr fixed actually causes particial hit, instead of hit rate reduction on prefix cache.

MengqingCao (approved):

PR #9904, opened 2026-06-02
[0.20.2][PrefixCache][BugFix] Fix Deepseek-V4 compress attn groups prefix caching hit

nwpu-zxr:
The kvcache caching and hash of the deepseek-v4 `compressor_kv` do not consider `compress_ratio`. This PR processes `compress_ratio` so that the prefix cache can be correctly hit.

No.

By CI.

### DIFF
--- a/vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
+++ b/vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
@@ -21,8 +21,6 @@ from vllm.v1.kv_cache_interface import (
     FullAttentionSpec,
     KVCacheConfig,
     KVCacheSpec,
-    UniformTypeKVCacheSpecs,
-    MLAAttentionSpec,
 )
 
 from vllm_ascend.core.single_type_kv_cache_manager import get_manager_for_kv_cache_spec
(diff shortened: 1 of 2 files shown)


vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
```diff
@@ -170,6 +174,13 @@ def find_longest_cache_hit(
         """
 
         def _get_block_hashes(kv_cache_spec: KVCacheSpec) -> BlockHashList:
+            if hasattr(kv_cache_spec, "compress_ratio"):
+                compress_ratio = kv_cache_spec.compress_ratio if kv_cache_spec.compress_ratio >= 1 else 1
```
hwhaokun, line 178: Consider unifying the two branches with effective_block_size = kv_cache_spec.block_size * max(getattr(kv_cache_spec, "compress_ratio", 1), 1), then compare against self.hash_block_size. This avoids the separate hasattr path and keeps hash granularity consistent for all specs.
e.g.
effective_block_size = kv_cache_spec.block_size * max(getattr(kv_cache_spec, "compress_ratio", 1), 1)
if effective_block_size == self.hash_block_size:
    return block_hashes
return BlockHashListWithBlockSize(block_hashes, self.hash_block_size, effective_block_size)

PR #9941, opened 2026-06-03
[Feature] Add VLLM_ASCEND_AUTO_DETECT_QUANTIZATION environment variable

quancs:
In RL case, the quantization method shouldn't be auto-detected as it's not what users wanted. This PR proposes to add an env variable `VLLM_ASCEND_AUTO_DETECT_QUANTIZATION` (1 by default for compatibility)  to control the auto-detection behavior.

No

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/envs.py
+++ b/vllm_ascend/envs.py
@@ -113,7 +113,7 @@ env_variables: dict[str, Callable[[], Any]] = {
     # Whether to use MultiBlockPool for KV cache management
     "VLLM_ASCEND_APPLY_DSV4_PATCH": lambda: bool(int(os.getenv("VLLM_ASCEND_APPLY_DSV4_PATCH", "0"))),
     # Whether to enable auto-detect quantization from model files.
-    # This feature is enabled by default. When disabled, the model will be treated as unquantized.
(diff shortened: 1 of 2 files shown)


vllm_ascend/envs.py
```diff
@@ -112,6 +112,9 @@
     "VLLM_ASCEND_ENABLE_BATCH_MEMCPY": lambda: os.getenv("VLLM_ASCEND_ENABLE_BATCH_MEMCPY", None),
     # Whether to use MultiBlockPool for KV cache management
     "VLLM_ASCEND_APPLY_DSV4_PATCH": lambda: bool(int(os.getenv("VLLM_ASCEND_APPLY_DSV4_PATCH", "0"))),
+    # Whether to enable auto-detect quantization from model files.
+    # This feature is enabled by default.
+    "VLLM_ASCEND_AUTO_DETECT_QUANTIZATION": lambda: bool(int(os.getenv("VLLM_ASCEND_AUTO_DETECT_QUANTIZATION", "1"))),
```
Tflowers-0129, line 117: --additional-config

PR #9949, opened 2026-06-03
[Performance][SpecDecode] Enable vocab-parallel local argmax for Qwen3 MTP draft sampling

kichanll:
This pull request enables vocab-parallel local argmax for Qwen3 MTP draft sampling. It introduces a patch for `Qwen3_5MTP` and `Qwen3NextMTP` to add the `get_top_tokens` method, and refactors `AscendSpecDecodeBaseProposer` to utilize `_draft_argmax` and related helper methods instead of relying on the old `enable_reduce_sample` configuration. Additionally, comprehensive unit tests have been added to verify these changes.

Feedback:
- In `llm_base_proposer.py`, directly accessing `decode_metadata.sas_metadata` can raise an `AttributeError` if the attention metadata class does not define it. Using `getattr` is recommended for safety.
- In `patch_qwen3_mtp_local_argmax.py`, unconditional imports of `Qwen3_5MTP` and `Qwen3NextMTP` may crash engine startup if the installed vLLM version does not contain these models. These imports should be wrapped in `try-except` blocks.

No.

Tested via newly added unit tests in `test_patch_qwen3_mtp_local_argmax.py` and `test_base_proposer.py`.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/patch/worker/patch_qwen3_mtp_local_argmax.py
+++ b/vllm_ascend/patch/worker/patch_qwen3_mtp_local_argmax.py
@@ -1,0 +1,38 @@
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+# This file is a part of the vllm-ascend project.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


vllm_ascend/patch/worker/patch_qwen3_mtp_local_argmax.py
```diff
@@ -0,0 +1,46 @@
...
+except ImportError:
+    Qwen3_5MTP = None
+
+try:
+    from vllm.model_executor.models.qwen3_next_mtp import Qwen3NextMTP
+except ImportError:
+    Qwen3NextMTP = None
+
+
+def _qwen_mtp_get_top_tokens(
```
MengqingCao, line 33: let's use the registry mechanism of `PluggableLayer` instead of introducing patch.

PR #9950, opened 2026-06-03
[Feature]Add pcp+dp validation interception

2416602906:

- vLLM version: v0.20.2

### DIFF
--- a/.github/workflows/schedule_stale_manage.yaml
+++ b/.github/workflows/schedule_stale_manage.yaml
@@ -7,13 +7,13 @@ on:
 
 jobs:
   remove-wait-feedback-on-comment:
-    if: ${{ github.event_name == 'issue_comment' }}
+    if: ${{ github.event_name == 'issue_comment' && contains(github.event.issue.labels.*.name, 'wait-feedback') }}
     runs-on: ubuntu-latest
     permissions:
       issues: write
     steps:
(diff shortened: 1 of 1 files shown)


jianzs:
Why can't these two features be enabled at the same time?

2416602906:
> 为什么这两个功能不能同时启用？

Hello, PCP is used as a replacement for DP. Currently, there are no applicable scenarios for stacking both PCP and DP.

jianzs:
> > 为什么这两个功能不能同时启用？
> 
> Hello, PCP is used as a replacement for DP. Currently, there are no applicable scenarios for stacking both PCP and DP.

I want to know whether the two features theoretically conflict with each other. If there is no conflict, why restrict user behavior?

2416602906:
> > > 为什么这两个功能不能同时启用？
> > 
> > 
> > 您好，PCP 用作 DP 的替代品。目前，尚无同时使用 PCP 和 DP 的适用场景。
> 
> 我想知道这两个功能在理论上是否存在冲突。如果没有冲突，为什么要限制用户行为？

Theoretically there's no conflict, but we don't recommend this setup. PCP is meant to save VRAM, so enabling DP to clone weights defeats the purpose. We should block this user behavior.

Tflowers-0129 (approved):

yiz-liu:
I think there may be a rebase issue here. Could you clean it up a bit?

PR #9963, opened 2026-06-03
[Test][BugFix] Recover full custom ops test

### LINKED ISSUE #5336: [Help Wanted]: Recover full custom ops test
### Your current environment

None

### 🐛 Describe the bug

Some tests are skipped in the nightly https://github.com/vllm-project/vllm-ascend/blob/6d25372baaa0ef018a75b427b387fab8dd2e92b4/.github/workflows/_e2e_nightly_single_node.yaml#L151, anyone interested can help to recover.

arnavnagzirkar:

This PR recovers the nightly custom ops test for fused MoE (`test_fused_moe.py`).

Background: the nightly CI workflow was ignoring the whole file through a
file-level `--ignore` flag in `_e2e_nightly_single_node.yaml`, and all four
tests also carried `@pytest.mark.skip("Probabilistic failure, need zengiant
after fix")` that was added in a prior repair commit. The root reason the file
was broken originally was a stale import (`from
vllm.model_executor.layers.activation import SiluAndMul`), which was already
fixed earlier by adding a local `SiluAndMul`. The file-level ignore and the
blanket skips were left behind as conservative measures.

What this PR changes:

- Removes the redundant `--ignore` for `test_fused_moe.py` in
  `_e2e_nightly_single_node.yaml` so the file is collected during nightly CI.
- Removes `@pytest.mark.skip` from `test_select_experts_invalid_scoring_func`.
  This test is deterministic: it patches `get_weight_prefetch_method`, then
  calls `select_experts` with `scoring_func="invalid"`, which raises
  `ValueError` before any NPU computation runs. There is no probabilistic
  failure path.

The other three tests keep their skip marks because they exercise real NPU
precision or compute paths:

- `test_token_dispatcher_with_all_gather`: numerical comparison with
  `atol=4e-2`, real precision risk on NPU.
- `test_token_dispatcher_with_all_gather_quant`: runs actual NPU
  dispatch/combine ops.
- `test_select_experts`: can take the fused path and call
  `DeviceOperator.moe_gating_top_k` in some parameter combinations, so it is
  not purely mocking based.

Fixes #5336

No.

Ran the relevant tests and the linter for the changed files while developing.
The change is kept minimal and focused on this one issue.

- vLLM version: v0.23.0


arnavnagzirkar:
Thanks for the review. I addressed the feedback:

- The workflow change is now committed. `.github/workflows/_e2e_nightly_single_node.yaml` had the `--ignore` for `test_fused_moe.py` removed, so the file is collected in nightly CI. This was missing from the earlier push and is now included.
- Fixed DCO. The branch is now a single signed-off commit rebased on the latest `main`.
- Updated the PR title and summary to the `[Test][BugFix]` format with the three-section description.

The change stays minimal: remove the redundant file-level ignore and unskip only the deterministic `test_select_experts_invalid_scoring_func`. The other three tests keep their skip marks due to real NPU precision and compute paths.

Issue #9170, opened 2026-05-14
[Bug]: DeepSeek-V4 W8A8 Concurrency Failure: Output Corruption and Malformed begin of sentence with MTP Enabled


w1027-1:
### Your current environment

vLLM Version: 0.18.0+empty (Editable install)

Model: DeepSeek-V4 (DSV4)

Quantization: W8A8

Attention Backend: FlashAttention (Ascend)

Speculative Config: {"num_speculative_tokens": 1, "method": "deepseek_mtp"}

Hardware: Ascend 910B (or equivalent)

### 🐛 Describe the bug

A critical execution error has been observed when running concurrent requests on DeepSeek-V4 (W8A8 quantized) with DeepSeek MTP and FlashAttention on Ascend NPUs.

The Phenomenon:
During concurrent load, the model's output suddenly injects a <｜begin of sentence｜> (or malformed begin?of?sentence) string, and the response terminates immediately after this injection. This prevents the model from completing its response and results in truncated, nonsensical outputs.

eg:
{"id":"chatcmpl-279128cb9fcfda5c6d7ba6699f7c605b","object":"chat.completion.chunk","created":1778151413,"model":"deepseek_v4","choices":[{"index":0,"delta":{"role":"assistant","content":""},"logprobs":null,"finish_reason":null}],"prompt_token_ids":null}

data: {"id":"chatcmpl-279128cb9fcfda5c6d7ba6699f7c605b","object":"chat.completion.chunk","created":1778151413,"model":"deepseek_v4","choices":[{"index":0,"delta":{"content":"<｜begin▁of▁sentence｜>"},"logprobs":null,"finish_reason":null,"token_ids":null}]}

data: {"id":"chatcmpl-279128cb9fcfda5c6d7ba6699f7c605b","object":"chat.completion.chunk","created":1778151413,"model":"deepseek_v4","choices":[{"index":0,"delta":{},"logprobs":null,"finish_reason":"stop","stop_reason":null,"token_ids":null}]}

data: [DONE]

Single Request: Works normally without any unexpected tokens.

Concurrent Requests (MTP Enabled): Frequent early termination preceded by BOS token injection.

Concurrent Requests (MTP Disabled): 100% stable; no early termination or BOS injection observed.

weijinqian0:
@nomewang @GDzhu01 Plz see this issue.

onkevin:
[...]

PR #9983, opened 2026-06-04
[Ops][BugFix] Disable fused MC2 under speculative decoding to prevent logit corruption

ppppangu:
### What this does
On Ascend A3, `select_moe_comm_method` routes MoE comm to the fused `dispatch_ffn_combine` op (FUSED_MC2). Under speculative decoding (e.g. **DeepSeek-V4 + MTP**, `num_speculative_tokens=1`) this op is handed a tiny/ragged token shape and corrupts the model's **first-decode-step logits**, yielding intermittent **empty / malformed `<｜begin▁of▁sentence｜>` first responses** (gh #9170).

The current guard `dispatch_ffn_combine_enable = ... and (not is_draft_model)` only covers the **draft** forward. The corruption appears on the **target** model's verify-step forward, which runs with `is_draft_model=False`, so the fused op is still selected there.

This PR additionally gates on `vllm_config.speculative_config`, disabling fused MC2 for **both** the draft and the target-verify forwards whenever speculative decoding is configured (routing to MC2 / ALLTOALL). **Non-speculative runs are bit-identical** — zero behavior change when `speculative_config is None`. Under speculative decoding the MoE comm falls back to the decomposed MC2/ALLTOALL path instead of the fused op; `dispatch_ffn_combine` was already unusable for these tiny/ragged shapes (it is precisely what corrupts them), and MTP's own decode speedup dominates, so there is no net throughput regression for spec-decode either.

### Root cause (kernel detail in #9170)
`dispatch_ffn_combine` (AscendC) targets a dense, large decode batch and never zero-initializes its output / combine staging (sized by a fixed `max_output_size=65536`). On the tiny/ragged spec shape (~2 query positions/seq split unevenly across EP ranks, so a rank's local token count can be `< AIV core count`), the output-assembly tiling `unpermute/moe_token_unpermute_tiling.h:30-34` computes `tokens_core_length = outTokens/coreNum = 0` then divides by it (`0/0`), corrupting the loop bounds that write per-token output; under-filled rows are read from never-initialized HBM → NaN/garbage logits. A proper kernel fix (min-1 tiling guard + output zero-init) belongs in the op and touches the hot non-spec MoE path; this PR is the safe, side-effect-free control-plane mitigation.

### Validation
DeepSeek-V4-Flash W8A8 + MTP (`num_speculative_tokens=1`), 910C A3, TP4×DP4 + EP16: with fused MC2 enabled the first agent turn intermittently returns empty / `<｜begin▁of▁sentence｜>`-only output; setting `VLLM_ASCEND_ENABLE_FUSED_MC2=0` eliminates it (spec-decode sub-agents then complete normally). This PR makes that automatic under `speculative_config`.

Related to #9170 (mitigates the MTP/speculative trigger; the underlying kernel defect stays tracked there).

- vLLM version: v0.25.1

### DIFF
--- a/tests/ut/test_ascend_forward_context.py
+++ b/tests/ut/test_ascend_forward_context.py
@@ -201,6 +201,31 @@ def test_select_moe_comm_method_a3_enable_fused_mc2_mode_1(
         (129, MoECommType.ALLTOALL),
     ],
 )
+def test_select_moe_comm_method_a3_spec_decode_disables_dispatch_ffn_combine(
+    monkeypatch,
+    num_tokens,
+    expected,
+):
+    _patch_select_moe_comm_method_deps(
+        monkeypatch,
+        device_type=afc.AscendDeviceType.A3,
+        capacity=128,
+        ep_world_size=8,
+        enable_fused_mc2=1,
+    )
+    vllm_config = _make_vllm_config(quant_type="w4a16")
+    vllm_config.speculative_config = SimpleNamespace()
+
+    assert afc.select_moe_comm_method(num_tokens, vllm_config) == expected
+
+
+@pytest.mark.parametrize(
+    ("num_tokens", "expected"),
+    [
+        (128, MoECommType.MC2),
+        (129, MoECommType.ALLTOALL),
+    ],
+)
 def test_select_moe_comm_method_a3_without_fused_mc2(
     monkeypatch,
     num_tokens,
--- a/vllm_ascend/ascend_forward_context.py
+++ b/vllm_ascend/ascend_forward_context.py
@@ -309,7 +309,12 @@ def _select_a3_moe_comm_method(
     if get_ascend_config().enable_fused_mc2 == 1:
         # TODO: drop the EP-size guard when mega_moe supports larger EP sizes
         mega_moe_enable = get_ep_group().world_size <= 64 and _cann_megamoe_supported_by_config(vllm_config)
-        dispatch_ffn_combine_enable = get_ep_group().world_size <= 32
+        # dispatch_ffn_combine corrupts output for the small/ragged token
+        # shapes used by speculative decoding. The target verification forward
+        # shares the speculative config, so this guard covers both target and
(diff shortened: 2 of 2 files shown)


ppppangu:
@wangxiyuan friendly ping 🙏 Rebased onto latest `main`, fixed the title prefix, and addressed the review note (switched to `getattr(vllm_config, "speculative_config", None)`). CI is green now — ready for review. This guards `dispatch_ffn_combine` off under speculative decoding, including the *target* verify-forward (which runs with `is_draft_model=False`), fixing the #9170 logit corruption (empty/garbage first-decode responses). Complementary to #9992, which hardens the op itself via zero-init. Thanks!

fourierr:
在A3上关闭了fused_mc2，还是出现了这个问题，启动命令如下
export OMP_PROC_BIND=false
export OMP_NUM_THREADS=10
export PYTORCH_NPU_ALLOC_CONF=expandable_segments:True
export HCCL_BUFFSIZE=1024
export TASK_QUEUE_ENABLE=1
export ENABLE_TRACE_LOGGING=true

vllm serve /DeepSeek-V4-Flash-w4a8-mtp/ \
    --max-model-len 262144 \
    --served-model-name illm \
    --max-num-seqs 32 \
    --data-parallel-size 1 \
    --tensor-parallel-size 8 \
    --enable-expert-parallel \
    --enable-auto-tool-choice \
    --model-loader-extra-config='{"enable_multithread_load": true, "num_threads": 128}' \
    --quantization ascend \
    --port 8080 \
    --block-size 32 \
    --speculative-config '{"num_speculative_tokens": 3,"method": "mtp"}' \
    --compilation-config '{"cudagraph_mode": "FULL_DECODE_ONLY"}' \
    --additional-config '
    {"ascend_compilation_config":{
        "enable_npugraph_ex":true,
        "enable_static_kernel":false
        },
    "enable_cpu_binding": true,
    "enable_flashcomm1": true,
    "multistream_overlap_shared_expert": true}'

PR #10004, opened 2026-06-04
[BugFix] Fix reduce sampling where top_k and top_p could be None

zouzy5137:
This pull request updates the sampler logic in `vllm_ascend/sample/sampler.py` to correctly handle cases where no filtering is needed (`p is None and k is None`) by returning the full gathered logits paired with a global identity index. This prevents out-of-bounds indexing when sampling.

- Sampling Logic Fix: Updated the sampling logic to correctly handle cases where top-k is None, ensuring consistent behavior across different tensor parallel configurations.
- Contract Consistency: Ensured that the sampler maintains the expected (candidate logits, candidate indices) contract even when no filtering is applied, preventing out-of-bounds errors.
- Ascend Backend Updates: Refactored the Ascend-specific sampling path to align with the general logic, improving robustness for reduce-sampling scenarios.

No.

It was tested by end2end cases. Find more in pr test report.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/sample/sampler.py
+++ b/vllm_ascend/sample/sampler.py
@@ -143,7 +143,7 @@ class AscendTopKTopPSampler(TopKTopPSampler):
             elif self.logprobs_mode == "processed_logprobs":
                 logits_to_return = cand_logits.log_softmax(dim=-1, dtype=torch.float32)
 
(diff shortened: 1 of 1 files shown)


MengqingCao (approved):
LGTM, thx for the fix!

PR #10044, opened 2026-06-05
[Feature]graph sampling stage1

calyoung80:

  This PR advances the first stage of full-graph speculative sampling refactoring on Ascend by moving key speculative sampling work out of AscendRejectionSampler and into the
  model runner, while keeping the existing path compatible.

  Main changes:

  - Add explicit prepared-input handling for speculative sampling inputs and state propagation in model_runner_v1.py.
  - Move speculative bonus-token sampling out of AscendRejectionSampler.forward() into the model runner path.
  - Move speculative target-logits preprocessing out of AscendRejectionSampler.forward() into the model runner path.
  - Move speculative SamplerOutput assembly into the model runner path.
  - Add and use a prepared-input entry in AscendRejectionSampler for the migrated path.
  - Reduce reliance on mutable sampler/rejection-sampler object state by making the new path consume prepared execution state.
  - Refactor sample_tokens() into clearer internal stages for sampling, bookkeeping, post-processing, and output finalization.

  Why this is needed:

  - The previous implementation mixed sampling execution, state preparation, bookkeeping, and output assembly across a few large functions and object-local mutable state.
  - That structure made it difficult to continue toward full-graph sampling because key speculative inputs were still prepared and consumed too late in the execution chain.
  - This PR establishes a cleaner execution boundary for speculative sampling and prepares the codebase for deeper full-graph sampling migration in follow-up work.

  This PR is intentionally scoped to the first functional migration stage. It does not attempt to fully complete full-graph sampling in one step.

  No API or interface change.

  Behaviorally, this changes internal execution organization for speculative sampling on Ascend, but it is intended to preserve existing outputs while enabling further full-graph
  sampling work.

  Tested in the target environment with repeated unit-test validation after each migration step.

  Commands used:

  cd /vllm-workspace/vllm-ascend
  pytest -q tests/ut/sample/test_sampler.py tests/ut/sample/test_rejection_sampler.py

  Observed result:

  - 16 passed

  Additional validation:

  - Repeated smoke checks were run during development to ensure the new speculative path did not introduce obvious regressions.
  - Python syntax validation was also checked during development with py_compile on modified files.

  Tests were focused on the sample/rejection-sample path because this PR primarily restructures and migrates speculative sampling execution boundaries rather than changing
  unrelated model behavior.

- vLLM version: v0.20.2


ader47:
Unused Variable

**File:** `vllm_ascend/sample/rejection_sampler.py`, line 219

```
Ruff (F841): Local variable `bonus_token_ids` is assigned to but never used
```

When `bonus_sampler_output` is passed as a pre-computed argument (not `None`), the code executes:

```python
bonus_token_ids = bonus_sampler_output.sampled_token_ids  # line 219
```

However, this value is never subsequently used in that code path. The `forward_with_prepared_inputs()` method receives the full `bonus_sampler_output` object and extracts `bonus_token_ids` internally:

```python
def forward_with_prepared_inputs(self, ..., bonus_sampler_output, ...):
    bonus_token_ids = bonus_sampler_output.sampled_token_ids  # re-extracted here
```

**Fix:** Remove the dead assignment or restructure so `bonus_token_ids` is actually used in the pre-computed path.

ader47:
Mutable State Side Effect in `forward_with_prepared_inputs`

**File:** `vllm_ascend/sample/rejection_sampler.py`

```python
def forward_with_prepared_inputs(self, ..., random_tensors=None):
    if random_tensors is None:
        random_tensors = self.random_tensors
        self.random_tensors = None  # silent side effect: clears object state
```

This method silently reads and clears `self.random_tensors` when the parameter is `None`. This "consume-once" pattern has several risks:

1. **Non-obvious API contract**: A caller that passes `random_tensors=None` may not expect the method to fall back to (and clear) object state.
2. **Ordering sensitivity**: If `forward_with_prepared_inputs()` is called twice without re-setting `random_tensors`, the second call silently gets `None`.
3. **No logging/warning**: The state is consumed without any trace.

**Suggestion:** Either:
- Make `random_tensors` a required parameter (no fallback to `self`), or
- Add a clear docstring warning about the consume-once semantics, and consider logging when the fallback is used.

PR #10046, opened 2026-06-05
[BugFix] Add env var to control DP metadata all_reduce communication

lijiahang226:

On certain devices, CPU-side all_reduce communication may carry dirty data during DP metadata synchronization in `_sync_metadata_across_dp`, leading to incorrect `num_tokens` and `cudagraph_mode` values across DP ranks, which can cause hangs or crashes.

This PR adds an environment variable `VLLM_ASCEND_DP_ALLREDUCE_ON_NPU` to switch the DP metadata all_reduce from the CPU group to the NPU device group.

**Behavior** (in `model_runner_v1._sync_metadata_across_dp`):

| `VLLM_ASCEND_DP_ALLREDUCE_ON_NPU` | device | group |
|-----------------------------------|--------|-------|
| `0` (default) | CPU | `cpu_group` |
| `1` | NPU | `device_group` |

No. The default behavior is unchanged (uses CPU group). Users on affected devices can set `dp_allreduce_on_npu:True` in additional config to work around the dirty data issue.

- Verified on affected A5 devices with `dp_allreduce_on_npu:True`

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -253,6 +253,10 @@ class AscendConfig:
         # Enable dispatch/combine op inter-node communication by ROCE
         self.enable_mc2_hierarchy_comm = additional_config.get("enable_mc2_hierarchy_comm", False)
 
+        # Whether to use NPU device group for DP metadata all_reduce.
+        # "True": use NPU device group, "False" (default): use CPU group.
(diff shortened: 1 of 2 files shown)


MengqingCao (commented):
Let's use the if condition instead of env var to avoid gloo communication on A5

MengqingCao (approved):

PR #10051, opened 2026-06-05
[Feature] Add PP and MTP support

gjc0824:
This PR backports vLLM Pipeline Parallelism (PP) + Multi-Token Prediction (MTP) runtime support to `vllm-ascend`. It introduces scheduler-level PP+MTP alignment in `scheduler_profiling_chunk.py`, patches `ModelRunnerOutput` and `EngineCore` to support `spec_token_ids` and skip global post-step updates in PP batch_queue mode, and implements worker-state updates and token broadcasting in `model_runner_v1.py`.

No, this is an internal backend enablement and backport of PP+MTP support.

New unit tests were added in `test_profiling_chunk.py`, `test_patch_pp_mtp.py`, `test_patch_qwen3_5_mtp.py`, and `test_model_runner_v1.py`. Note that some of the new tests currently call unimplemented methods and will fail until those methods are implemented or the tests are corrected.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -674,6 +674,105 @@ class NPUModelRunner(GPUModelRunner):
         self.query_start_loc.copy_to_gpu()
 
         return num_reqs_padded
+    
+    def _is_non_last_pp_mtp(self) -> bool:
+        return (
+            self.need_accepted_tokens
+            and self.speculative_config is not None
+            and not get_pp_group().is_last_rank
+            and not self.use_async_scheduling
+        )
+
+    def _sync_num_accepted_tokens_to_gpu(
+        self, accepted_counts: dict[str, int]
+    ) -> None:
+        if not accepted_counts:
+            return
+        for req_id, count in accepted_counts.items():
+            req_index = self.input_batch.req_id_to_index.get(req_id)
+            if req_index is not None:
+                self.input_batch.num_accepted_tokens_cpu[req_index] = count
+        num_reqs = self.input_batch.num_reqs
+        self.num_accepted_tokens.np[:num_reqs] = (
+            self.input_batch.num_accepted_tokens_cpu[:num_reqs]
+        )
+        self.num_accepted_tokens.copy_to_gpu(num_reqs)
+
+    def _compute_non_last_pp_mtp_accepted_counts(
+        self,
+        scheduler_output: SchedulerOutput,
+        prev_scheduler_output: SchedulerOutput | None,
+    ) -> dict[str, int]:
+        req_data = scheduler_output.scheduled_cached_reqs
+        prev_spec_tokens = (
+            prev_scheduler_output.scheduled_spec_decode_tokens
+            if prev_scheduler_output is not None
+            else {}
+        )
+        accepted_counts: dict[str, int] = {}
+
+        for i, req_id in enumerate(req_data.req_ids):
+            count = 1
+            req_state = self.requests.get(req_id)
+            if req_state is not None and req_id in prev_spec_tokens:
+                prev_num_computed = req_state.num_computed_tokens
+                num_computed = int(req_data.num_computed_tokens[i])
+                delta = num_computed - prev_num_computed
+                max_count = len(prev_spec_tokens[req_id]) + 1
+                logger.info(f">>>>>>>>>> max_count: {max_count}, delta: {delta}, num_computed: {num_computed}, prev_num_computed: {prev_num_computed}")
+                if 0 < delta <= max_count:
+                    count = delta
+            accepted_counts[req_id] = max(1, count)
+
+        return accepted_counts
+    
+    def _update_states(self, scheduler_output: "SchedulerOutput"):
+        # For non-last PP + MTP, compute accepted counts via delta of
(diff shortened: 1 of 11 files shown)


MengqingCao (commented):
Thanks for this great work, but I think it is more reasonable to support this in vLLM instead of  vllm-ascend.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1679,6 +1780,87 @@ def _copy_draft_token_ids_to_cpu(
                 self.draft_token_ids_cpu[:num_reqs] = 0
             self.draft_token_ids_event.record()
 
+    def _collect_pp_mtp_readded_token(
```
LookAround0301, line 1783: pp相关的函数可以统一用一个类管理，类似pcp_manager

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1679,6 +1780,87 @@ def _copy_draft_token_ids_to_cpu(
...
+                continue
+            if new_tokens := new_token_ids[i]:
+                token_fixes[req_id] = (
+                    list(new_tokens),
+                    req_data.num_computed_tokens[i],
+                )
+        return token_fixes
+
+    @contextmanager
+    def _pp_mtp_update_req_spec_token_ids(
```
LookAround0301, line 1816: 同上，不直接写在model runner里

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -675,6 +683,92 @@ def _pad_query_start_loc_for_fia(
...
+                num_computed = int(req_data.num_computed_tokens[i])
+                delta = num_computed - prev_num_computed
+                max_count = len(prev_spec_tokens[req_id]) + 1
+                if 0 < delta <= max_count:
+                    count = delta
+            accepted_counts[req_id] = max(1, count)
+
+        return accepted_counts
+
+    def _prepare_non_last_pp_mtp_state_update(
```
LookAround0301, line 736: 函数命名需要规范，并添加相关注释，其他函数同理

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -675,6 +683,92 @@ def _pad_query_start_loc_for_fia(
...
+    def _compute_non_last_pp_mtp_accepted_counts(
+        self,
+        scheduler_output: SchedulerOutput,
+        prev_scheduler_output: SchedulerOutput | None,
+    ) -> dict[str, int]:
+        req_data = scheduler_output.scheduled_cached_reqs
+        prev_spec_tokens = (
+            prev_scheduler_output.scheduled_spec_decode_tokens
+            if prev_scheduler_output is not None
+            else {}
```
LookAround0301, line 718: 这里可能得改改，可能过不了CI

knight0528:
I have also implemented the same functionality in vLLM. If possible, I would like to discuss with you the adaptation approach for PP and MTP.

knight0528:
https://github.com/vllm-project/vllm/pull/39704

lidenghui1110:
> [vllm-project/vllm#39704](https://github.com/vllm-project/vllm/pull/39704)

@gjc0824 Compared to use scheduler process to sync draft_token between PP stages in this PR, I would rather to use broadcast following the strategy in async-scheduling in this PR: [#39074](https://github.com/vllm-project/vllm/pull/39704) , for the future maintaining of PP with async-scheduling, I recommend the second one which implemented in vllm, how do you think about it? cc @MengqingCao @knight0528 @jianzs

vllm_ascend/worker/model_runner_v1.py
gjc0824: ok

vllm_ascend/worker/model_runner_v1.py
gjc0824: ok

vllm_ascend/worker/model_runner_v1.py
gjc0824: ok

vllm_ascend/worker/model_runner_v1.py
gjc0824: ok

gjc0824:
> > [vllm-project/vllm#39704](https://github.com/vllm-project/vllm/pull/39704)
> 
> @gjc0824 Compared to use scheduler process to sync draft_token between PP stages in this PR, I would rather to use broadcast following the strategy in async-scheduling in this PR: [#39074](https://github.com/vllm-project/vllm/pull/39704) , for the future maintaining of PP with async-scheduling, I recommend the second one which implemented in vllm, how do you think about it? cc @MengqingCao @knight0528 @jianzs

The current implementation is the adaptation of your solution to the vllm-ascend solution, which was only for urgent needs. I think the core issue of PP+MTP is to update the metadata of non-last PP rank. The better method is following the broadcast of PP+async_scheduling. However, I may not continue with the follow-up work.

gjc0824:
> I have also implemented the same functionality in vLLM. If possible, I would like to discuss with you the adaptation approach for PP and MTP.

Thanks for your great work. Actually, current implementation is the adaptation of your solution to the vllm-ascend solution, which was only for urgent needs.

PR #10053, opened 2026-06-05
Pp mtp pd

LostFox11:

- vLLM version: v0.21.0


ader47:
VllmEplbAdaptor crashes when `_registered_moe_layers` is empty

**File:** `vllm_ascend/eplb/adaptor/vllm_adaptor.py`

`VllmEplbAdaptor.__init__` unconditionally indexes into `self.moe_layers[0]` at line ~85:

```python
self.moe_layers = VllmEplbAdaptor._registered_moe_layers
self.num_moe_layers = len(self.moe_layers)
# ...
first_layer = self.moe_layers[0]  # <-- IndexError if empty
self.num_local_experts = first_layer.local_num_experts
self.ep_rank = first_layer.ep_rank
```

If the model has no MoE layers on this PP rank (e.g., non-MoE models, or PP ranks that only hold dense layers), `_registered_moe_layers` will be an empty list and `self.moe_layers[0]` will raise an `IndexError`. The same issue applies to `init_expert_param_per_layer`, which also indexes `first_layer = self.moe_layers[0]`.

**Recommendation:** Guard with an early return or raise a clear `ValueError` when `_registered_moe_layers` is empty. For example:

```python
if not self.moe_layers:
    raise ValueError("No MoE layers registered for EPLB on this rank. "
                     "Ensure AscendFusedMoE layers are initialized before "
                     "creating VllmEplbAdaptor.")
```

ader47:
Weight name prefix mismatch in `_patched_mtp_load_weights`

**File:** `vllm_ascend/patch/worker/patch_deepseek_mtp_pp.py`

The patched `load_weights` iterates over `weights_list` using the raw checkpoint `name` as key to look up in `params_dict`:

```python
params_dict = dict(self.named_parameters())
for name, loaded_weight in weights_list:
    if name in loaded:
        continue
    if name in params_dict:
        param = params_dict[name]
        ...
```

However, `self.named_parameters()` returns names that include the `prefix` argument passed during model construction (e.g., the model is typically instantiated with `prefix="model"` or similar). Checkpoint weight names from the safetensors file may or may not include this prefix. If there is a mismatch (e.g., checkpoint has `model.embed_tokens.weight` but the model parameter is named `embed_tokens.weight` due to how `DeepSeekMTP` sets `prefix`), the lookup will silently skip loading these weights.

Additionally, `_original_mtp_load_weights` is called first and returns a `set[str]` of loaded names. The fallback loop then attempts to load anything NOT in that set. But the original method's weight name matching logic may differ from the raw `params_dict` key lookup, leading to either double-loading or silent skips.

**Recommendation:** Use `AutoWeightsLoader` (which is already used elsewhere in this PR for `GlmMoeDsaForCausalLM`) or apply the same prefix-remapping logic that the original `load_weights` uses. At minimum, log a warning when a weight name from the checkpoint is not found in `params_dict`, so that silent load failures are detectable.

PR #10069, opened 2026-06-05
[Misc][Ops]add trion ops log

Fager10086:

Standardize logging for Triton operators to improve issue diagnosis.

Add the [TritonOps] log tag for Triton operator errors and print operator input shapes to facilitate debugging and troubleshooting.

Does not.

Logs containing the [TritonOps] keyword are generated when starting the model inference service or running standalone operator tests.

- vLLM version: v0.21.0


vllm_ascend/ops/triton/layernorm_gated.py
```diff
@@ -138,7 +138,11 @@ def layer_norm_fwd_npu(
     MAX_FUSED_SIZE = 65536 // x.element_size()
     BLOCK_N = min(MAX_FUSED_SIZE, triton.next_power_of_2(group_size))
     if group_size > BLOCK_N:
-        raise RuntimeError("Feature dim too large.")
+        raise RuntimeError(
```
linfeng-yuan, line 148: Why do we only raise RuntimeError here with logger.error but print error log above (e.g., in `vllm_ascend/ops/triton/mamba/causal_conv1d.py`).

vllm_ascend/ops/triton/mamba/causal_conv1d.py
```diff
@@ -653,6 +663,11 @@ def causal_conv1d_update_npu(
     # vector core 40
     # TODO: use driver to get the vector core num
     CORE_HINT = 40
```
linfeng-yuan: It is such a bad implementation here. What we need to do is not raising warning here but using get_vectorcore_num to avoid hardcoding.

linfeng-yuan (approved):
LGTM

PR #10071, opened 2026-06-05
[CI] add git clone try support to pr e2e test

leo-pony:
Add git clone try support to pr e2e test.

Fix the bug network unstable lead code clone failed.
7b8ae023-4fe8-41a0-abe4-2e2d5c82b5e7

https://github.com/vllm-project/vllm-ascend/actions/runs/26873282806/job/79254495068?pr=9933

NA

NA

- vLLM version: v0.20.2

### DIFF
--- a/.github/workflows/pr_test.yaml
+++ b/.github/workflows/pr_test.yaml
@@ -73,6 +73,14 @@ jobs:
           echo "✓ PR title prefix is valid"
 
       - name: Checkout vllm-project/vllm-ascend repo
+        id: checkout_ascend_1
+        uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


Yikun (changes requested):
I don't think we need such 3ed party action

@pkking Infra team should ensure this.

leo-pony:
> I don't think we need such 3ed party action
> 
> @pkking Infra team should ensure this.

@Yikun Thanks for review！
I've implemented the retry logic using the official GitHub Action; could you take another look at it for me?

Yikun:
better add note why we need this in commit msg or note, but sounds like a infra bug

PR #10078, opened 2026-06-05
[Feature][Worker] Layerwise KV cache pool with prefill layer reuse support (Prefill offload)

ader47:

- vLLM version: v0.22.1


vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/layerwise_config.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def get_layerwise_kv_cache_reuse_layers(
+    num_layers: int,
+    extra_config: dict[str, Any] | None = None,
+) -> int | None:
+    layerwise_config = get_layerwise_config(num_layers, extra_config)
+    if not layerwise_config.has_layer_reuse:
+        return None
+    return layerwise_config.num_shared_buffers
```
dsxsteven, line 158: Can you add more uts for covering edge cases such as empty extra_config, extra_config=None, out-of-range indices, invalid types

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/layerwise_config.py
```diff
@@ -0,0 +1,158 @@
...
+    has_layer_reuse: bool
+
+
+def _parse_int_config(value: Any, name: str) -> int:
+    if isinstance(value, bool):
+        raise TypeError(f"{name} must be an integer, got bool")
+    try:
+        return int(value)
+    except (TypeError, ValueError) as err:
+        raise TypeError(f"{name} must be an integer, got {value!r}") from err
```
dsxsteven, line 27: If a user sets "num_prefetch_layers": true in extra_config, int(True) returns 1, which silently accepts a boolean as a valid integer. This is a common foot-gun because bool is a subclass of int in Python. You can add an explicit isinstance(value, bool) check before int(value),

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/layerwise_config.py
```diff
@@ -0,0 +1,158 @@
...
+        value = extra_config[_EXTRA_CONFIG_KEY_NUM_SHARED_BUFFERS]
+    if value is None:
+        value = envs_ascend.VLLM_ASCEND_KV_POOL_LAYERWISE_NUM_SHARED_BUFFERS
+    if value is None:
+        return num_layers
+    num_shared_buffers = _parse_int_config(
+        value,
+        _EXTRA_CONFIG_KEY_NUM_SHARED_BUFFERS,
+    )
+    if num_shared_buffers < 1:
```
dsxsteven, line 45: There is no check for num_shared_buffers > num_layers. If a user sets layerwise_num_shared_buffers = 100 for a 32-layer model, has_layer_reuse becomes False (line 134) and layer reuse is silently disabled with no warning.
You can add a warning or raise an error.

vllm_ascend/patch/platform/patch_layerwise_kv_cache_reuse.py
```diff
@@ -0,0 +1,291 @@
...
+    extra_config: dict | None = None,
+) -> int | None:
+    """Return configured KV cache reuse slots, or None when disabled."""
+    if max_layers is None:
+        return None
+    reuse_layers = get_layerwise_kv_cache_reuse_layers(
+        max_layers, extra_config)
+    if reuse_layers is None:
+        return None
+    if max_layers is not None and reuse_layers > max_layers:
```
dsxsteven: max_layers is already checked for None at line 38 (if max_layers is None: return None). The second check is dead code.

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/layerwise_config.py
ader47: Hi @dsxsteven, thanks for the suggestion! This is already handled in the current implementation — _parse_int_config explicitly checks for bool before calling int(value) :
```
def _parse_int_config(value: Any, name: str) -> int:
    if isinstance(value, bool):
        raise TypeError(f"{name} must be an integer, got bool")
    try:
        return int(value)
    except (TypeError, ValueError) as err:
        raise TypeError(f"{name} must be an integer, got {value!r}") from err
```

ader47: This guard is applied to all integer config parsing ( num_shared_buffers , prefetch_layers , and independent_layers indices). So passing true / false as a boolean value will raise a TypeError rather than being silently accepted.

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/layerwise_config.py
ader47: Good catch, thanks @dsxsteven! I'll add a warning when num_shared_buffers >= num_layers since it effectively disables layer reuse. I think a warning is more appropriate than an error here — the behavior is still correct, just likely unintended.

vllm_ascend/patch/platform/patch_layerwise_kv_cache_reuse.py
ader47: Good point, thanks @dsxsteven! The max_layers is not None check is indeed redundant since we already return early when it's None . I'll simplify it.

vllm_ascend/patch/platform/__init__.py
```diff
@@ -37,16 +37,14 @@
 import vllm_ascend.patch.platform.patch_deepseek_v4_thinking  # noqa
 import vllm_ascend.patch.platform.patch_torch_accelerator  # noqa
 import vllm_ascend.patch.platform.patch_tool_choice_none_content  # noqa
-import vllm_ascend.patch.platform.patch_mamba_manager  # noqa
 
 if os.getenv("DYNAMIC_EPLB", "false").lower() in ("true", "1") or os.getenv("EXPERT_MAP_RECORD", "false") == "true":
     import vllm_ascend.patch.platform.patch_multiproc_executor  # noqa
 
+if envs.VLLM_ASCEND_KV_POOL_LAYERWISE_NUM_SHARED_BUFFERS is not None:
+    import vllm_ascend.patch.platform.patch_kv_cache_utils  # noqa
```
Zhang-Bryan: This line appears to duplicate with line 22. Should it be "import patch_layerwise_kv_cache_reuse" instead?

PR #10087, opened 2026-06-05
[BugFix] Add AscendKVBlockZeroer to clean used mamba block for full attn

slippersss:
This PR aims to add block zeroer to clean used mamba block when assigning it to full attn. The problem is originated from the fact that spec_step>1 dose not obtain accurate seq_len and thus will use dirty block containing nan. Thus, though different from the reason why vllm use block zeroer, we try to reuse the existing logic to do the same thing.

N/A

by ci

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/worker/utils.py
+++ b/vllm_ascend/worker/utils.py
@@ -1,0 +1,176 @@
+from collections.abc import Iterable
+from itertools import product as iprod
+from typing import Any
+
+import torch
+from vllm.triton_utils import tl, triton
+from vllm.utils.math_utils import largest_power_of_2_divisor
+from vllm.v1.kv_cache_interface import FullAttentionSpec
+from vllm.v1.worker.utils import AttentionGroup, KVBlockZeroer
+
+from vllm_ascend.ops.triton.triton_utils import get_vectorcore_num
+
+
+@triton.jit
+def _zero_kv_blocks_kernel(
+    seg_addrs_ptr,
+    block_ids_ptr,
+    n_blocks,
+    N_SEGS: tl.constexpr,
+    PAGE_SIZE_EL: tl.constexpr,
+    BLOCK_SIZE: tl.constexpr,
+    GRID_SIZE: tl.constexpr,
+):
+    """Zero KV cache blocks across all segments in a single launch.
+
+    Each segment is a contiguous region of one block's data.  For backends
+    where blocks are outermost (block_dim=0) there is one segment per
+    buffer.  For backends where K/V is outermost (block_dim=1) there are
+    two segments per buffer (one for K, one for V).
+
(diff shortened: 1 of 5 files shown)


vllm_ascend/worker/utils.py
```diff
@@ -0,0 +1,175 @@
...
+                    outer_strides = [kv.stride(d) * el for d in outer_dims]
+                    for outer in iprod(*(range(kv.shape[d]) for d in outer_dims)):
+                        off_bytes = sum(i * s for i, s in zip(outer, outer_strides))
+                        seg_addrs.append(dp + off_bytes)
+
+        if not seg_addrs or page_size_el is None:
+            self._meta = None
+            return
+
+        blk_size = min(largest_power_of_2_divisor(page_size_el), 8192)
```
MengqingCao, line 130: what's this magic number 8192 refers to?

slippersss: Vllm sets it as 1024. We set it as 8192, due to the recommendation that 8192*8B=64B is more preferred to meet the UB size (8B is about int64 in the above triton kernel, line 49).

MengqingCao: okay, maybe a comment is needed here.

slippersss: Done.

vllm_ascend/worker/utils.py
```diff
@@ -0,0 +1,176 @@
...
+        chunk_index = remainder % chunks
+        block_id = tl.load(block_ids_ptr + block_index)
+        seg_addr = tl.load(seg_addrs_ptr + seg_index)
+        ptr = tl.cast(seg_addr, tl.pointer_type(tl.int32))
+        offset = block_id.to(tl.int64) * PAGE_SIZE_EL + chunk_index.to(tl.int64) * BLOCK_SIZE
+        cols = tl.arange(0, BLOCK_SIZE).to(tl.int64)
+        tl.store(ptr + offset + cols, tl.zeros([BLOCK_SIZE], dtype=tl.int32))
+
+
+class AscendKVBlockZeroer(KVBlockZeroer):
```
MengqingCao, line 53: it seems we did 2 different things compared with `KVBlockZeroer` in vLLM:
1. process the tuple of kvcache tensors instead of single kvcache tensor
2. change the `blk_size` to meet the UB size
if there is any misunderstanding, plz point it out, thx!

MengqingCao (approved):
LGTM now, actually I prefer to solve this by passing by a device tensor `seqs_len`, but I'm okay to merge this as a temporary solution.

slippersss:
> LGTM now, actually I prefer to solve this by passing by a device tensor `seqs_len`, but I'm okay to merge this as a temporary solution.

When ops are ready, we will remove this temporary solution.

PR #10096, opened 2026-06-05
[CI][v0.20.2]: support v0.20.2rc1 dispatch and 310p manylinux_2_34 baseimage

zhangxinyuehfad:

Fix [the error](https://github.com/vllm-project/vllm-ascend/actions/runs/26894814044/job/79477689281) verification baseimage not updated after https://github.com/vllm-project/vllm-ascend/pull/9710/changes#diff-e648e6ecc4ed55f550fc621019e61d85e00c2ab5c3f647733512ee48ba901f29L18 

- Add v0.20.2rc1 to workflow_dispatch tag choice list so wheel and variant can be re-published via UI on the release branch
- Insert a workflow_dispatch-only sed step in the 310p build job to bump Dockerfile.buildwheel.310p base image from 9.1.0-beta.1-310p-manylinux_2_28 to manylinux_2_34, addressing the 310p base image requirement for the v0.20.2rc1 release

### DIFF
--- a/.github/workflows/schedule_release_code_and_wheel.yml
+++ b/.github/workflows/schedule_release_code_and_wheel.yml
@@ -33,6 +33,7 @@ on:
         type: choice
         options:
           - main
+          - v0.20.2rc1
           - v0.19.1rc1
           - v0.18.0
           - v0.18.0rc1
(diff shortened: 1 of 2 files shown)


.github/workflows/schedule_release_code_and_wheel.yml
```diff
@@ -278,6 +279,17 @@ jobs:
           tool-cache: true
           docker-images: false
 
+      - name: Patch 310p Dockerfile base image to manylinux_2_34
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        run: |
+          FILE=.github/workflows/dockerfiles/Dockerfile.buildwheel.310p
+          if ! grep -q 'quay.io/ascend/manylinux:9\.1\.0-beta\.1-310p-manylinux_2_28' "$FILE"; then
```
MengqingCao, line 286: I don't get the meaning of checking this?

zhangxinyuehfad: This check is a safeguard for that. The whl package for tag v0.20.2rc1 failed to build because the base image wasn't updated from manylinux_2_28 to manylinux_2_34. We need to swap the old base image for the new one so the whl builds properly, and this grep ensures the old image is indeed replaced.

PR #10100, opened 2026-06-05
[BugFix] Solve the lmhead parallel feature assert lack and nightly bug

lilinsiman:
The verification that the lmhead parallel feature is not supported by the dense model is added, and the incorrect writing of the nightly test case is modified.

no

ut and tests

- vLLM version: v0.20.2


linfeng-yuan (approved):
LGTM

PR #10102, opened 2026-06-05
[BugFix][Refactor] Reduce Mooncake KV cache register regions for sparse C8

ZYang6263:

This PR refactors Mooncake KV cache buffer registration to reduce duplicated logic between the normal Mooncake connector and the layerwise connector.

It also reduces HCCL/Mooncake register region count for sparse-c8 KV caches by:
- allocating dsa_k and dsa_k_scale from one aligned raw allocation;
- keeping logical KV cache metadata unchanged;
- registering storage-merged memory ranges instead of registering every logical tensor separately.

For Mamba / attention-Mamba hybrid paths, the existing special registration ranges are preserved to avoid losing MTP padding handling.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
@@ -57,13 +57,13 @@ from vllm.v1.request import RequestStatus
 from vllm_ascend import envs as ascend_envs
 from vllm_ascend.ascend_config import get_ascend_config, init_ascend_config
 from vllm_ascend.distributed.kv_transfer.utils.mooncake_transfer_engine import global_te
-from vllm_ascend.distributed.kv_transfer.utils.utils import get_transfer_timeout_value
-from vllm_ascend.utils import enable_custom_op
 from vllm_ascend.distributed.kv_transfer.utils.utils import (
     RegisterRegions,
     collect_storage_merged_register_regions,
-    warn_if_register_regions_exceed_limit,
+    get_transfer_timeout_value,
+    validate_register_region_count,
 )
+from vllm_ascend.utils import enable_custom_op
 
 # isort: off
 if TYPE_CHECKING:
@@ -1768,7 +1768,7 @@ class MooncakeConnectorWorker:
             # storage to avoid exceeding the HCCL per-process region limit.
             register_regions = collect_storage_merged_register_regions(kv_caches)
 
-        warn_if_register_regions_exceed_limit(register_regions)
+        validate_register_region_count(register_regions)
         global_te.register_buffer(register_regions.ptrs, register_regions.lengths)
 
         logger.debug(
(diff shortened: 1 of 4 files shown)


ader47:
`tensor_nbytes` is redundant with `tensor.nbytes`

The helper function:

```python
def tensor_nbytes(tensor: torch.Tensor) -> int:
    return tensor.numel() * tensor.element_size()
```

is semantically identical to PyTorch's built-in `tensor.nbytes` property, which also returns `numel() * element_size()`. The custom wrapper adds no value, adds an extra function call in a loop, and could confuse future readers into thinking there is a semantic difference where none exists. All call sites should use `tensor.nbytes` directly.

ader47:
`warn_if_register_regions_exceed_limit` only logs a `logger.warning` when the merged region count exceeds `MAX_HCCL_REGISTER_REGIONS (256)`:

```python
def warn_if_register_regions_exceed_limit(regions: RegisterRegions) -> None:
    if len(regions.ptrs) <= MAX_HCCL_REGISTER_REGIONS:
        return
    logger.warning(
        "Mooncake register_buffer merged region count %d exceeds "
        "HCCL per-process limit %d. This may still fail. ...",
        len(regions.ptrs),
        MAX_HCCL_REGISTER_REGIONS,
    )
```

The function name says "warn" but the consequence is that `register_buffer` is still called with an excessive number of regions, which will **deterministically fail** at the HCCL level. This should either:
- Raise an exception to prevent silent failure at registration time.
- Attempt further merging (e.g., increase the merge gap dynamically).
- Return a boolean/status so the caller can decide how to handle it.

A warning that says "this may still fail" shifts responsibility to the operator reading logs rather than failing fast and clearly.

ZYang6263:
> `tensor_nbytes` is redundant with `tensor.nbytes`
> 
> The helper function:
> 
> ```python
> def tensor_nbytes(tensor: torch.Tensor) -> int:
>     return tensor.numel() * tensor.element_size()
> ```
> 
> is semantically identical to PyTorch's built-in `tensor.nbytes` property, which also returns `numel() * element_size()`. The custom wrapper adds no value, adds an extra function call in a loop, and could confuse future readers into thinking there is a semantic difference where none exists. All call sites should use `tensor.nbytes` directly.

Thanks for pointing this out. Updated to use tensor.nbytes directly and removed the redundant tensor_nbytes helper.

ZYang6263:
> `tensor_nbytes` is redundant with `tensor.nbytes`
> 
> The helper function:
> 
> ```python
> def tensor_nbytes(tensor: torch.Tensor) -> int:
>     return tensor.numel() * tensor.element_size()
> ```
> 
> is semantically identical to PyTorch's built-in `tensor.nbytes` property, which also returns `numel() * element_size()`. The custom wrapper adds no value, adds an extra function call in a loop, and could confuse future readers into thinking there is a semantic difference where none exists. All call sites should use `tensor.nbytes` directly.

I agree that continuing to call register_buffer after exceeding the HCCL region limit is not useful. I changed the limit check to fail fast with a RuntimeError before calling register_buffer, and split it from the normal logging path.

linfeng-yuan (approved):
LGTM

PR #10117, opened 2026-06-05
[main][BugFix] Avoid the issue of moe hanging in multi-DP scenarios without introducing new problems

drslark:

It is a problem descibbed in https://github.com/vllm-project/vllm/issues/44185.

We solved it in vllm-ascend, by dropping problematic input_fits_in_drafter.

Also, we changed the mask to fill `1` to solve the original problem solved in https://github.com/vllm-project/vllm-ascend/pull/9221.

N/A

By ci.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/spec_decode/llm_base_proposer.py
+++ b/vllm_ascend/spec_decode/llm_base_proposer.py
@@ -1509,15 +1509,16 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
         common_attn_metadata.seq_lens[:batch_size] += 1
         # For the requests that exceed the max model length, we set the
         # sequence length to 1 to minimize their overheads in attention.
-        common_attn_metadata.seq_lens[:batch_size].masked_fill_(exceeds_max_model_len, 1)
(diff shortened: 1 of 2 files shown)


vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -1509,15 +1509,16 @@ def attn_update_stack_num_spec_norm(
         common_attn_metadata.seq_lens[:batch_size] += 1
         # For the requests that exceed the max model length, we set the
         # sequence length to 1 to minimize their overheads in attention.
-        common_attn_metadata.seq_lens[:batch_size].masked_fill_(exceeds_max_model_len, 1)
+        exceeds_mask = common_attn_metadata.seq_lens[:batch_size] > self.max_model_len
```
linfeng-yuan, line 1512: I think this would lead to D2H synchronization here. Could we use seq_lens_cpu instead?

linfeng-yuan: Not involved.

linfeng-yuan (approved):
LGTM

PR #10120, opened 2026-06-06
[Feature] Support configuring RFork settings via environment variables

GoMarck:
This PR introduces support for configuring RFork settings via environment variables as fallbacks when they are absent from `--model-loader-extra-config`. It also updates the documentation and error messages to reflect these changes.

Feedback: In `rfork_loader.py`, `isinstance(value, (int, float, str))` will evaluate to `True` for boolean values because `bool` is a subclass of `int` in Python. This can cause boolean values to be incorrectly parsed as `1.0` or `0.0` instead of falling back to the environment variables. An explicit check for `isinstance(value, bool)` should be added.

Yes, users can now configure RFork settings using environment variables.

No tests were added in this PR.

- vLLM version: v0.20.2

### DIFF
--- a/docs/source/user_guide/feature_guide/rfork.md
+++ b/docs/source/user_guide/feature_guide/rfork.md
@@ -31,7 +31,9 @@ The RFork loading flow in the current implementation is:
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


linfeng-yuan (commented):
please move this env to additional_config

PR #10121, opened 2026-06-06
[feature]Fused MTP Graph Merge

calyoung80:
Base on the closest community commit fcb30beb and apply source-compatible file-level changes while preserving community-only modules that are not wired into the current source runtime path.

Sign-off: calyoung80
Sign-off: BobQiu1981

  This PR ports the fused MTP full-graph support from the current DeepSeek V4 source snapshot onto the closest community baseline, fcb30beb, using a minimal file-level adaptation
  strategy.

  The main goal is to land the fused MTP runtime path needed by the source code while preserving community consistency wherever possible instead of replacing the whole tree with
  the source snapshot.

  Core changes:

  - vllm_ascend/worker/model_runner_v1.py
    Main fused MTP integration:
      - full-graph enable checks
      - fused wrapper creation
      - capture / replay parameter updates
      - runtime buffers
      - graph-internal sampling output handling
      - draft token writeback
      - dtype fixes
      - get_model() unwrap fix

    Adds _FusedModelWithMTP wrapper:
      - target forward + in-graph verifier
      - in-graph sampling
      - MTP draft propose
      - previous hidden-state shape handling

    Exposes _FusedModelWithMTP for reuse from model_runner_v1.py.

  - vllm_ascend/sample/sampler.py
    Adds / wires sample_with_runtime_state() for fused-graph runtime-state sampling, reusing gumbel_sample.

  - vllm_ascend/compilation/acl_graph.py
    Adds ACLGraphWrapper.is_captured() so replay can check whether capture metadata already exists.

  Config and env support:

  - vllm_ascend/ascend_config.py
    Adds additional_config.enable_fused_mtp_full_graph.

  - vllm_ascend/envs.py
    Adds VLLM_ASCEND_ENABLE_FUSED_MTP_FULL_GRAPH.

  Model compatibility support:

  - vllm_ascend/models/deepseek_v4.py
    Stores / exposes MTP target hidden states through _mtp_hidden_buffer and get_mtp_target_hidden_states().

  - vllm_ascend/models/deepseek_v4_mtp.py
    Handles fused-path previous_hidden_states in both 2D and expanded forms, fixing the previous [-1, 4, 4096] reshape issue.

  Implementation approach:

  - Files that already exist in community history with content exactly matching the source snapshot were restored from community history.
  - Only files that do not have an exact upstream match were taken directly from source.
  - Generated / packaged artifacts from the source snapshot were intentionally excluded:
      - vllm_ascend/_build_info.py
      - vllm_ascend/_cann_ops_custom/vendors/

  This is needed because the source snapshot is not a clean export of a single upstream commit; it is a mixed snapshot with a small number of non-upstream file variants. A full
  overwrite would unnecessarily discard valid community code.

  Yes.

  This PR introduces user-facing runtime behavior changes for Ascend-specific fused MTP / speculative decoding paths, especially for DeepSeek V4 related execution.

  User-visible impact may include:

  - different fused graph behavior for MTP execution
  - changed speculative decoding behavior in affected paths
  - new config / env control for fused MTP full-graph mode

  No new public API surface is intentionally introduced beyond the new config/env controls.

  - Identified fcb30beb as the closest community baseline to the source snapshot.
  - Compared file blobs against local community Git history and restored exact upstream-matching file versions where possible.
  - Applied only the required source-only files directly.
  - Ran syntax validation with python -m py_compile on the source-applied files.
  - Reviewed format_ci.log / format_ci_01.log and fixed the issues in the source-applied scope:
      - long-line cleanup in vllm_ascend/worker/model_runner_v1.py
      - removal of unused local variables in the same file

  - Confirmed that remaining shellcheck findings are from existing community shell scripts under csrc/ and are outside the scope of this patch.

  No new unit tests were added in this PR. The work is primarily a source-to-community adaptation for fused MTP full-graph support, and the immediate goal was to preserve required
  runtime behavior while minimizing unrelated divergence from community code.

- vLLM version: v0.20.2


yiz-liu:
This is too large for a single PR and would place a significant burden on reviewers, so I'm inclined not to merge it.

I'm also curious how host-side operations between the target model and draft model (sampling, for example) are handled when the they are fused into a single static graph.

PR #10140, opened 2026-06-07
[Misc]M2m

nofushanquan:

- vLLM version: v0.20.2


ader47:
PP tests unconditionally skipped with an empty reason (`tests/e2e/pull_request/four_card/test_pipeline_parallel.py`, lines 261, 269)

   Two pipeline-parallel test functions (`test_models_pp2_tp2` and the DP+PP variant) are marked `@pytest.mark.skip(reason=(""))`. The reason string is empty, giving no indication of why these tests are disabled, whether they are temporarily or permanently skipped, or what it would take to re-enable them. For a version-upgrade PR this is especially concerning -- pipeline parallelism is a core distributed feature and silently disabling its entire test suite without any documented justification makes regressions invisible. At minimum the reason should explain the failure (e.g., "PP tests hang on v0.21.0 due to ...") and link to a tracking issue.

PR #10153, opened 2026-06-08
[BugFix]fix w4a8 mxfp quantization in shared experts

lcfenglinwan:
This pull request refactors the shared experts gate and down projection calls in fused_moe.py to use the unified .apply() method of the quantization schemes, passing the quantized inputs and scales as a tuple. It also updates the apply methods in w4a8_mxfp4.py and w8a8_mxfp8.py to handle tuple inputs and set the appropriate output data type. Feedback highlights a critical NameError in w8a8_mxfp8.py where dynamic_scale was renamed to pertoken_scale but still referenced by its old name, as well as minor indentation issues. Additionally, the reviewer requested updating the PR title and description to comply with the repository's style guide.

- vLLM version: v0.20.2

### DIFF
--- a/vllm_ascend/ops/fused_moe/fused_moe.py
+++ b/vllm_ascend/ops/fused_moe/fused_moe.py
@@ -786,17 +786,7 @@ class AscendFusedMoE(FusedMoE):
                 # Execute the gate projection and activation concurrently with the
                 # dispatch communication.
                 maybe_wait_event(fused_moe_evts.before_dispatch)
-                hidden_states = torch_npu.npu_quant_matmul(
-                    quantized_x,
-                    self._shared_experts.gate_up_proj.weight,
-                    self._shared_experts.gate_up_proj.weight_scale,
-                    scale_dtype=torch_npu.float8_e8m0fnu,
-                    pertoken_scale=pertoken_scale,
-                    pertoken_scale_dtype=torch_npu.float8_e8m0fnu,
-                    bias=None,
-                    output_dtype=original_dtype,
-                    group_sizes=[1, 1, 32],
-                )
+                hidden_states = self._shared_experts.gate_up_proj((quantized_x, pertoken_scale))[0]
                 # Execute activation concurrently with gmm2.
                 maybe_wait_event(fused_moe_evts.before_gmm2)
                 quantized_x, swiglu_out_scale, _ = torch.ops._C_ascend.npu_swiglu_group_quant(
@@ -810,17 +800,7 @@ class AscendFusedMoE(FusedMoE):
                 # Execute the down projection concurrently with the combine
                 # communication.
                 maybe_wait_event(fused_moe_evts.before_combine)
-                shared_out = torch_npu.npu_quant_matmul(
-                    quantized_x,
-                    self._shared_experts.down_proj.weight,
-                    self._shared_experts.down_proj.weight_scale,
-                    scale_dtype=torch_npu.float8_e8m0fnu,
-                    pertoken_scale=swiglu_out_scale,
-                    pertoken_scale_dtype=torch_npu.float8_e8m0fnu,
-                    bias=None,
-                    output_dtype=original_dtype,
-                    group_sizes=[1, 1, 32],
-                )
+                shared_out = self._shared_experts.down_proj((quantized_x, swiglu_out_scale))[0]
             else:
                 # Ensure the shared experts wait for hidden_states to be ready.
                 torch.npu.current_stream().wait_event(fused_moe_evts.before_routed_experts)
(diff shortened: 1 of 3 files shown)


linsheng1:
I found a possible dtype regression in the MXFP tuple-input path.

The new tuple path hardcodes:

```python
output_dtype = torch.bfloat16
```

This is now used by the shared-expert path in `fused_moe.py`, where the previous direct `torch_npu.npu_quant_matmul(...)` call used:

```python
output_dtype=original_dtype
```

So for non-bf16 runs, shared expert outputs can change dtype from the model's original dtype to bf16.

This PR also removes the previous `w8a8_mxfp8.py` bias normalization:

```python
if bias is not None and bias.dtype != torch.float32:
    bias = bias.to(torch.float32)
```

If `npu_quant_matmul` still expects fp32 bias for this path, this can reintroduce dtype errors for biased layers.

Suggested fix: preserve the caller/model dtype for tuple inputs, either by passing the desired output dtype alongside the pre-quantized tuple or deriving it from the surrounding layer/input contract. I would also keep the bias cast unless the operator contract is verified to accept all bias dtypes used here.

vllm_ascend/quantization/methods/w8a8_mxfp8.py
```diff
@@ -71,18 +71,21 @@ def get_pergroup_param(
...
+           quantized_x, pertoken_scale = x
+           original_shape = quantized_x.shape
+           output_dtype = torch.bfloat16
+        else:
+            # reshape x for Qwen VL models
+            original_shape = x.shape
+            if x.dim() > 2:
+                x = x.view(-1, x.shape[-1])
+            quantized_x, pertoken_scale = torch_npu.npu_dynamic_mx_quant(x, dst_type=torch.float8_e4m3fn)
+            output_dtype = x.dtype
```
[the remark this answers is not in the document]
lcfenglinwan: indentation fixed, and now all names used here is pertoken_scale

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -810,17 +800,7 @@ def maybe_wait_event(evt: torch.npu.Event | None):
...
-                    self._shared_experts.down_proj.weight,
-                    self._shared_experts.down_proj.weight_scale,
-                    scale_dtype=torch_npu.float8_e8m0fnu,
-                    pertoken_scale=swiglu_out_scale,
-                    pertoken_scale_dtype=torch_npu.float8_e8m0fnu,
-                    bias=None,
-                    output_dtype=original_dtype,
-                    group_sizes=[1, 1, 32],
-                )
+                shared_out = self._shared_experts.down_proj((quantized_x, swiglu_out_scale))[0]
```
kunpengW-code, line 803: Is `[0]` necessary here? The `apply` method returns the output of the `torch_npu.npu_quant_matmul` interface, which should just be a Tensor. The same question about R789.

vllm_ascend/quantization/methods/w4a8_mxfp4.py
```diff
@@ -65,9 +65,12 @@ def apply(
         bias: torch.Tensor | None = None,
         tp_rank: int | None = 0,
     ) -> torch.Tensor:
-        quantized_x, dynamic_scale = torch_npu.npu_dynamic_mx_quant(x, dst_type=torch.float8_e4m3fn)
-
-        output_dtype = x.dtype if not isinstance(x, tuple) else x[0].dtype
+        if isinstance(x, tuple):
+            quantized_x, dynamic_scale = x
+            output_dtype = torch.bfloat16
```
TmacAaron, line 70: Here `output_dtype` is forced to `torch.bfloat16`. The previous passed `output_dtype=original_dtype`, where `original_dtype` is derived from the input `hidden states`. So this may silently convert fp16 outputs to bf16. 

I recommend pass the original_dtype together with the per-quantized tuple, e.g. `(quantized_x, dynamic_scale, original_dtype)`. The same concern applies to `w8a8_mxfp8.py`.

vllm_ascend/ops/fused_moe/fused_moe.py
lcfenglinwan: it is necessary here because self._shared_experts.down_proj is initialized without return_bias input, so the output of the linear module is (output, output_bias)

> https://github.com/vllm-project/vllm/blob/469f3dcf1d70349d0776ef82bdc5c4aff20a93f2/vllm/model_executor/layers/linear.py#L1563

vllm_ascend/quantization/methods/w4a8_mxfp4.py
lcfenglinwan: I think passing in a new parameter might also cause some issues, such as the original_dtype potentially needing to be passed through multiple levels. Also, since the current standard is bfloat16, bfloat16 was specified here.

linfeng-yuan (approved):
LGTM. Please 1) use `dynamic_scales` instead of `pertoken_scales` and 2) make sure how to avoid dtype regression here with hard-coding `bf16` in the next PR @lcfenglinwan

PR #10154, opened 2026-06-08
[Ops][BugFix] Support disabling reduce_sample when unverified sampling parameters are requested

hzx55906:
This PR introduces a mechanism to temporarily override and disable the reduce_sample optimization for unverified sampling parameters. To this end, a new global variable "_reduce_sample_force_disabled" is introduced to dynamically manage this variable in each sampling.

Yes. When unverified sampling parameters are requested with reduce_sample enabled, a warning is now triggered to inform the user that the optimization will be temporarily disabled for that request, and the content of the unsupported sampling parameters.

- vLLM version: v0.23.0


linsheng1:
I think this PR leaks the reduce-sample override context.

In `model_runner_v1.py`, both `execute_model()` and `_sample()` manually enter the context manager:

```python
if needs_reduce_sample_override:
    reduce_sample_ctx = override_reduce_sample(False)
    reduce_sample_ctx.__enter__()
```

but there is no matching `__exit__()` / reset. Since `_REDUCE_SAMPLE_OVERRIDE` is a `ContextVar`, a logprobs request can leave `is_reduce_sample_enabled()` forced to `False` for later work on the same worker/task context.

Impact: later requests that do not ask for logprobs may silently lose the reduce-sample optimized path and potentially change logits/sampler behavior.

Suggested fix: use the context manager in the existing `with (...)` stack, or wrap the affected block in `try/finally`. For example, set `reduce_sample_ctx = override_reduce_sample(False) if needs_reduce_sample_override else nullcontext()` and include it in the `with`, matching the safer pattern already used in `llm_base_proposer.py`.

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -971,6 +991,7 @@ def _propose(
             else:
                 draft_token_ids = run_draft()
                 self._update_full_graph_params_if_needed(forward_context, num_input_tokens, multi_steps_attn_metadata)
+        set_reduce_sample_force_disabled(_prev_reduce_sample_disabled)
```
lilinsiman: Check if this call back is needed or not

linfeng-yuan (commented):
1. @wangxiyuan @MengqingCao Please check the dynamic `reduce_sample_enabled` usage here, and the influence of maintainence of model_runner and llm_proposer.
2. Please add E2E case in your PR. (e.g., enable the switch in a performance test with spec decode with sampling data.)
3. Based on 2, verify the performance influence of your PR.  

After 1-3, we must verify it in a full test setting with both performance and accuracy. Because this with will be turned on later, it is necessary to verify this function in this PR.

linfeng-yuan (dismissed):
Let's trigger CI test first, then nightly with specific cases. Then we should set it to `False` by default and merge it.

PR #10158, opened 2026-06-08
[CI] Fix comment CI auth error

wangxiyuan:

- vLLM version: v0.20.2

### DIFF
--- a/.github/workflows/pr_nightly_command.yml
+++ b/.github/workflows/pr_nightly_command.yml
@@ -35,19 +35,21 @@ jobs:
       test_cases: ${{ steps.resolve.outputs.test_cases }}
       branch: ${{ steps.resolve.outputs.branch }}
       is_authorized: ${{ steps.auth.outputs.is_authorized }}
+      dispatch_a2: ${{ steps.resolve.outputs.dispatch_a2 }}
+      dispatch_a3: ${{ steps.resolve.outputs.dispatch_a3 }}
     steps:
       - name: Check user authorization
         id: auth
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.PAT_TOKEN }}
         run: |
           ACTOR='${{ github.event.client_payload.github.payload.comment.user.login }}'
           REPO='${{ github.event.client_payload.github.payload.repository.full_name }}'
 
-          PERMISSION=$(gh api "repos/$REPO/collaborators/$ACTOR/permission" --jq '.permission')
-          case "$PERMISSION" in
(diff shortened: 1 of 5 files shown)


linsheng1:
There is a GitHub Actions context typo in the new `post-links` job.

The workflow uses:

```yaml
repository: ${{ github.event_client_payload.github.payload.repository.full_name }}
```

but GitHub Actions exposes this under `github.event`, and the same workflow uses the correct form elsewhere:

```yaml
${{ github.event.client_payload.github.payload.repository.full_name }}
```

Impact: the `Post workflow URLs` step can fail expression evaluation or pass an invalid repository, so the `/nightly` command may dispatch jobs but fail to report the A2/A3 workflow links back to the PR.

Suggested fix: change it to:

```yaml
repository: ${{ github.event.client_payload.github.payload.repository.full_name }}
```

It is also worth testing the full `/nightly` path, since the run-id lookup fetches the most recent workflow run after a fixed sleep and may be racy under concurrent dispatches.

.github/workflows/pr_nightly_command.yml
```diff
@@ -118,42 +113,140 @@ jobs:
...
+            echo "branch=$BRANCH"
+          } >> "$GITHUB_OUTPUT"
+
+          # Determine which platform(s) to dispatch based on test name conventions:
+          # - Names ending with -A2 or in A2-only list -> A2
+          # - Names ending with -A3 or in A3-only list -> A3
+          # - Names without platform suffix -> dispatch both (conservative)
+          # - "accuracy-group" -> dispatch both
+          A2_ONLY="test_custom_op test_custom_op_multi_card"
+          A3_ONLY="custom-multi-ops"
```
zhangxinyuehfad, line 136: need to add more cases, it will run all if case not in here, like: https://github.com/nv-action/vllm-benchmarks/pull/254#issuecomment-4647004011

.github/workflows/pr_nightly_command.yml
```diff
@@ -118,42 +113,140 @@ jobs:
...
+      a2_run_id: ${{ steps.dispatch_a2.outputs.a2_run_id }}
     steps:
       - name: Dispatch nightly-a2 workflow
+        id: dispatch_a2
         env:
           GH_TOKEN: ${{ secrets.PAT_TOKEN }}
         run: |
           REPO='${{ github.event.client_payload.github.payload.repository.full_name }}'
+          REQUEST_ID='${{ github.run_id }}'
           gh workflow run schedule_nightly_test_a2.yaml \
```
zhangxinyuehfad, line 182: Need to skip build image in schedule_nightly_test_a2/a3 when pr command `/nightly`

PR #10159, opened 2026-06-08
[E2E] Add E2E tests for VWN Eagle3

YzTongNiar:
增添E2E用例看护VWN Eagle3性能，具体增加的E2E用例，以及其看护范围如下：

| Test Method| Scheduling Mode | Batch Configuration| max\_tokens | Assertion Method| Coverage Scope|
| -----------| --------- | ----------------- | ----------- | ------------------ | ------------------------------------|
| `test_vwn_eagle3_acceptance_tp4`                   | async=True      | padded batch              | 256         | acceptance-per-pos vs golden \[0.50, 0.25, 0.10], tolerance=0.15 | Acceptance rate regression under default async scheduling  |
| `test_vwn_eagle3_correctness_tp4`                  | default         | padded batch              | 64          | Greedy decoding output non-empty                                 | Basic functional correctness                               |
| `test_vwn_eagle3_acceptance_tp4_async_false`       | async=False     | padded batch              | 256         | acceptance-per-pos vs golden, tolerance=0.20                     | Acceptance rate under synchronous scheduling mode          |
| `test_vwn_eagle3_disable_padded_drafter_batch_tp4` | default         | padded batch **disabled** | 64          | Greedy decoding output non-empty                                 | Correctness with padded drafter batch disabled             |
| `test_vwn_eagle3_longer_generation_tp4`            | default         | padded batch              | 512         | Output non-empty + token count > 0                               | Spec decode loop stability under long generation scenarios |

Nope

- vLLM version: v0.20.2

### DIFF
--- a/tests/e2e/pull_request/four_card/spec_decode/test_vwn_eagle3.py
+++ b/tests/e2e/pull_request/four_card/spec_decode/test_vwn_eagle3.py
@@ -1,0 +1,411 @@
+"""E2E acceptance tests for VWN-Eagle3 speculative decoding (TP=4, 4 NPU).
+
+Test Matrix
+-----------
++----------------------------------------------+-------------+-------------+------------+---------------------+
+| Test                                         | Scheduling   | Padded Batch | max_tokens | Assertion           |
++----------------------------------------------+-------------+-------------+------------+---------------------+
+| test_vwn_eagle3_acceptance_tp4               | async=True  | enabled     | 256        | acceptance vs golden |
+| test_vwn_eagle3_correctness_tp4              | default     | enabled     | 64         | non-empty output    |
+| test_vwn_eagle3_acceptance_tp4_async_false   | async=False | enabled     | 256        | acceptance vs golden |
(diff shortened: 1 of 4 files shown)


linsheng1:
There seems to be a weight-prefix bug in `VwnLlamaDecoderLayer`.

`downward_and_forgot_after_attn` is constructed with the same prefix as `downward_and_forgot`:

```python
self.downward_and_forgot = _linear(..., mp(prefix, "downward_and_forgot"))
...
self.downward_and_forgot_after_attn = _linear(
    wd // m, (hs + wd) // m, vllm_config, qc, mp(prefix, "downward_and_forgot"))
```

but the checkpoint/test key is distinct: `layers.0.downward_and_forgot_after_attn.weight`.

Impact: the after-attention projection can load the wrong weight, fail to load its intended weight, or collide with the pre-attention projection metadata depending on how `ReplicatedLinear` uses the prefix.

Suggested fix: use:

```python
mp(prefix, "downward_and_forgot_after_attn")
```

It would also help to add a test that verifies the `downward_and_forgot_after_attn` checkpoint key actually populates that module, not just that the module has the expected shape.

PR #10161, opened 2026-06-08
[Test]update upstream cases to v0.21.0

jiangyunfan1:
This PR remove 3 unavailable cases and add passed cases at v0.21.0, we need to test them regularly
No
by running the test

- vLLM version: v0.21.0

### DIFF
--- a/.github/workflows/scripts/upstream_config.yaml
+++ b/.github/workflows/scripts/upstream_config.yaml
@@ -371,10 +371,6 @@ e2e-upstream_singlecard:
     estimated_time: 20
   - name: tests/entrypoints/openai/responses/test_sampling_params.py
     estimated_time: 20
-  - name: tests/kernels/moe/test_unquantized_backend_selection.py
-    estimated_time: 20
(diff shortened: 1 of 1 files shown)


linsheng1:
I found a CI coverage risk in this PR.

`run_suite.py` changes the shared upstream config path from `upstream_config.yaml` to `upstream_config1.yaml`. The new file looks like a small debug subset, while the original `upstream_config.yaml` still contains the maintained broad upstream suite.

Because `run_suite.py` is the shared suite entry point, this would make upstream suite invocations use the temporary/debug config and silently reduce E2E coverage.

Suggested fix: keep `_CONFIG_UPSTREAM` pointing at `upstream_config.yaml`. If this subset is needed, pass it explicitly or add a separate suite/config selector instead of replacing the global default.

jiangyunfan1:
https://github.com/vllm-project/vllm-ascend/actions/runs/27190423766/job/80269051806?pr=10161

realliujiaxu (approved):

PR #10174, opened 2026-06-08
[Feature][Attention] Refactor DSA attention, clean up device overrides, and introduce fused MTP full graph execution

calyoung80:

  The main goal is to land the fused MTP runtime path needed by the source code while preserving community consistency wherever possible instead of replacing the whole tree with the source snapshot.

  Core changes:

  - vllm_ascend/worker/model_runner_v1.py
    Main fused MTP integration:
      - full-graph enable checks
      - fused wrapper creation
      - capture / replay parameter updates
      - runtime buffers
      - graph-internal sampling output handling
      - draft token writeback
      - dtype fixes
      - get_model() unwrap fix

    Adds _FusedModelWithMTP wrapper:
      - target forward + in-graph verifier
      - in-graph sampling
      - MTP draft propose
      - previous hidden-state shape handling

    Exposes _FusedModelWithMTP for reuse from model_runner_v1.py.

  - vllm_ascend/sample/sampler.py
    Adds / wires sample_with_runtime_state() for fused-graph runtime-state sampling, reusing gumbel_sample.

  - vllm_ascend/compilation/acl_graph.py
    Adds ACLGraphWrapper.is_captured() so replay can check whether capture metadata already exists.

  Config and env support:

  - vllm_ascend/ascend_config.py
    Adds additional_config.enable_fused_mtp_full_graph.

  - vllm_ascend/envs.py
    Adds VLLM_ASCEND_ENABLE_FUSED_MTP_FULL_GRAPH.

  Model compatibility support:

  - vllm_ascend/models/deepseek_v4.py
    Stores / exposes MTP target hidden states through _mtp_hidden_buffer and get_mtp_target_hidden_states().

  - vllm_ascend/models/deepseek_v4_mtp.py
    Handles fused-path previous_hidden_states in both 2D and expanded forms, fixing the previous [-1, 4, 4096] reshape issue.

  Implementation approach:

  - Files that already exist in community history with content exactly matching the source snapshot were restored from community history.
  - Only files that do not have an exact upstream match were taken directly from source.
  - Generated / packaged artifacts from the source snapshot were intentionally excluded:
      - vllm_ascend/_build_info.py
      - vllm_ascend/_cann_ops_custom/vendors/

  This is needed because the source snapshot is not a clean export of a single upstream commit; it is a mixed snapshot with a small number of non-upstream file variants. A full
  overwrite would unnecessarily discard valid community code.

  Yes.

  This PR introduces user-facing runtime behavior changes for Ascend-specific fused MTP / speculative decoding paths, especially for DeepSeek V4 related execution.

  User-visible impact may include:

  - different fused graph behavior for MTP execution
  - changed speculative decoding behavior in affected paths
  - new config / env control for fused MTP full-graph mode

  No new public API surface is intentionally introduced beyond the new config/env controls.

  - Identified fcb30beb as the closest community baseline to the source snapshot.
  - Compared file blobs against local community Git history and restored exact upstream-matching file versions where possible.
  - Applied only the required source-only files directly.
  - Ran syntax validation with python -m py_compile on the source-applied files.
  - Reviewed format_ci.log / format_ci_01.log and fixed the issues in the source-applied scope:
      - long-line cleanup in vllm_ascend/worker/model_runner_v1.py
      - removal of unused local variables in the same file

  - Confirmed that remaining shellcheck findings are from existing community shell scripts under csrc/ and are outside the scope of this patch.

  No new unit tests were added in this PR. The work is primarily a source-to-community adaptation for fused MTP full-graph support, and the immediate goal was to preserve required
  runtime behavior while minimizing unrelated divergence from community code.

Sign-off: [calyoung80](zsucal@hotmail.com)
Sign-off: [BobQiu1981](qiumingbo81@163.com)

- vLLM version: v0.21.0


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -430,6 +434,24 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
...
+        self._fused_mtp_runtime_common_attn_metadata: AscendCommonAttentionMetadata | None = None
+        self._fused_mtp_debug_enabled = os.getenv(
+            "VLLM_ASCEND_FUSED_MTP_DEBUG", "0"
+        ).lower() in ("1", "true", "yes", "on")
+        try:
+            fused_mtp_debug_interval = int(
+                os.getenv("VLLM_ASCEND_FUSED_MTP_DEBUG_INTERVAL", "200")
+            )
+        except ValueError:
+            fused_mtp_debug_interval = 200
```
TmacAaron: It's better to load all ENV VARIABLES in envs.py ?

calyoung80: Yes, absolutely. It is a highly recommended best practice. I have fixed it.

vllm_ascend/models/deepseek_v4_mtp.py
```diff
@@ -61,12 +60,8 @@ def __init__(self, vllm_config: VllmConfig, prefix: str) -> None:
...
         quant_config = vllm_config.quant_config
 
-        self.e_proj = ReplicatedLinear(
-            config.hidden_size, config.hidden_size, bias=False, quant_config=quant_config, return_bias=False
-        )
-        self.h_proj = ReplicatedLinear(
-            config.hidden_size, config.hidden_size, bias=False, quant_config=quant_config, return_bias=False
-        )
+        self.e_proj = nn.Linear(config.hidden_size, config.hidden_size, bias=False)
+        self.h_proj = nn.Linear(config.hidden_size, config.hidden_size, bias=False)
```
Hailey-Zh: ReplicatedLinear replicates the full weight on every TP rank, while nn.Linear does not. If the MTP module runs under TP > 1, nn.Linear will participate in tensor-parallel sharding, causing weight-loading shape mismatches or incorrect computation results.

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -93,21 +95,266 @@ def split_inputs_tp_to_sp(hidden_states, out):
...
+            input_ids=input_ids,
+            positions=positions,
+            intermediate_tensors=intermediate_tensors,
+            inputs_embeds=inputs_embeds,
+            **kwargs,
+        )
+        forward_context = get_forward_context()
+        is_capturing = getattr(forward_context, "capturing", False)
+        if is_capturing and self.capture_mtp_enabled and not isinstance(hidden_states, IntermediateTensors):
+            raw_hidden = hidden_states[0] if isinstance(hidden_states, tuple) else hidden_states
```
Hailey-Zh: suggest add a length check as a safeguard against downstream logic break:
raw_hidden = hidden_states[0] if isinstance(hidden_states, tuple) and len(hidden_states) > 0 else hidden_states

vllm_ascend/models/deepseek_v4_mtp.py
calyoung80: Agree. For TP > 1, these two projections must stay replicated across ranks. Using nn.Linear here is unsafe because it can introduce sharded-loading / compute mismatch. Updated
  to ReplicatedLinear and fixed the forward path accordingly.

vllm_ascend/spec_decode/llm_base_proposer.py
calyoung80: Agree. This is a low-risk defensive fix. Added the tuple length check to avoid unexpected IndexError if downstream behavior changes.

PR #10193, opened 2026-06-08
[BugFix][v0.20.2rc] Port DSv4 SWA prefix-cache retention to vllm-ascend

Liuchenbing-2026:

Lifts DSv4 SWA prefix-cache hit rate on vllm-ascend `releases/v0.20.2rc`
from near-zero to ~50% by allowing the DSv4 compressor's KV-cache
`block_size` to be configured per-deployment instead of being hard-pinned
to 128 tokens.

### Symptom

```
--max-model-len 16384+, --enable-prefix-caching
8K-input × concurrency 32 × repeat 0.9 sequential replay:
  prefix_cache_hit:        ~0%
  TTFT avg:                ~16400 ms (= cold prefill)
```

### Root cause

The DSv4 compressor's KV-cache block size is hard-coded to **128 tokens**
in two places:

- `vllm_ascend/models/layer/attention/layer.py::DSAAttention.get_kv_cache_spec()`
  returns `MLAAttentionSpec(block_size=128, ...)`.
- `vllm_ascend/patch/worker/patch_deepseek_compressor.py::AscendDeepseekV4IndexerCache.get_kv_cache_spec()`
  returns `AscendMLAAttentionSpec(block_size=128, ...)`.

Prompts whose token count is not an exact multiple of 128 leave their
tail block uncached. Under the standard prefix-matching loop, the moment
one tail block is uncached the matcher stops returning more blocks past
that point, so the entire trailing region of the prompt also misses
even when the body would otherwise have hit. Concurrent long-context
workloads aggravate this because every request has its own un-aligned
tail.

### Fix

Make the DSv4 compressor's KV-cache `block_size` configurable through a
new env switch (default 128, so legacy behaviour is preserved
bit-for-bit). At `block_size = 32` the alignment requirement shrinks by
4×, so 4× more prompt prefixes can match exactly.

Single env switch:

```bash
VLLM_ASCEND_DSV4_COMPRESSED_KV_BLOCK_SIZE=32 vllm serve ...
# valid: 32 / 64 / 128
# unset -> legacy 128 behaviour
# only effective when compress_ratio == 128
```

### What changed

- **`patch_kv_cache_interface.py`** — new `get_dsv4_compressed_kv_block_size(compress_ratio)`
  resolver. Reads the env, validates the value is in `{32, 64, 128}`,
  short-circuits to default `128` when `compress_ratio != 128`.
  Relaxes the DSV4 C128 `block_size % compress_ratio == 0` check (which
  previously rejected `block_size < compress_ratio`).
- **`models/layer/attention/layer.py`** —
  `DSAAttention.get_kv_cache_spec()` calls the new resolver instead of
  hard-coding 128.
- **`patch/worker/patch_deepseek_compressor.py`** —
  `AscendDeepseekV4IndexerCache.get_kv_cache_spec()` calls the same
  resolver so the indexer cache stays consistent with the DSA attention
  cache spec.
- **`patch_kv_cache_utils.py`** — KV-cache page size is now aggregated
  across all KV groups (not only the first full-MLA group), and the
  `KVCacheTensor` allocator skips tensors whose `shared_by` is empty.
  Required because at `block_size = 32` different KV groups can have
  different page sizes.
- **`core/single_type_kv_cache_manager.py`** —
  `CompressAttentionManager` tracks `_prefix_block_size` separately from
  the allocation block_size. Prefix hashing keeps using the original
  token range so cache entries stay comparable across deployments with
  different compressed block sizes.
- **`worker/model_runner_v1.py`** — `kv_cache_raw_tensors` layer-name
  mismatch now raises a structured error with missing / extra / tensor
  summary instead of bare `assert`. Diagnostic-only; no inference math
  change.
- **`patch_kv_cache_coordinator.py`** + **`worker/block_table.py`** —
  small `_prefix_block_size` plumbing and debug-print updates to keep
  the multi-KV-group page-size logic legible.

## NPU production measurements

Validated on Ascend A3 (TP8), DeepSeek-V4-Flash-w8a8-mtp, served on
vllm v0.20.2. Workload `input 8192 x output 1024 x concurrency 32 x
repeat 0.9`, baseline `367b8e62` vs this PR.

### Prefix cache hit rate + serving performance

| Metric | baseline | this PR | Gain |
|---|---:|---:|---|
| prefix cache hit rate | `0.00%` | `81.54%` | hits |
| TTFT avg | `14593.2 ms` | `6098.9 ms` | **2.39x** faster |
| TTFT P90 | `24798.3 ms` | `7018.8 ms` | **3.53x** faster |
| TPOT avg | `36.9 ms` | `27.9 ms` | **1.32x** faster (not degraded) |
| TPOT SLO_P90 | `46.0 ms` | `30.0 ms` | **1.53x** faster |
| E2E time | `53.13 s` | `36.96 s` | **1.44x** faster |
| QPS | `0.6023` | `0.8658` | +43.75% |
| output throughput | `616.8 tok/s` | `886.6 tok/s` | +43.74% |
| E2E throughput | `5553.6 tok/s` | `7982.9 tok/s` | +43.74% |
| prefill token throughput | `561.6 tok/s` | `1343.8 tok/s` | **2.39x** |

Key point: at `output_len=1024` the TPOT **improves** rather than
regresses, so the old "more hits -> worse TPOT" shape no longer applies.

### Second workload (baseline already partially hitting)

Where the baseline already hit 44.53% on the same workload, this PR
pushes it to 77.70% (+33.17 pp), still with no TPOT degradation:

| Metric | baseline | this PR | Gain |
|---|---:|---:|---|
| prefix cache hit rate | `44.53%` | `77.70%` | +33.17 pp |
| TTFT avg | `14298.8 ms` | `8623.7 ms` | **1.66x** faster |
| TPOT avg | `35.4 ms` | `33.7 ms` | not degraded |
| E2E throughput | `20945.7 tok/s` | `24432.4 tok/s` | +16.65% |
| prefill token throughput | `2291.9 tok/s` | `3800.2 tok/s` | +65.81% |

### Accuracy (regression check)

GPQA Diamond (198 questions, `temperature=0`, 0-shot CoT):

| item | value |
|---|---:|
| `GPQA_diamond.accuracy` | **`89.8989898989899`** |
| summary | `89.90` |
| completed | `198 / 198` |
| failed | `0` |

GPQA is a non-repeating workload (does not consume prefix-cache entries);
it is the precision regression check. Accuracy matches the production
baseline -- no precision regression from the smaller KV-cache block size.

### Service environment used during validation

| item | value |
|---|---|
| platform | `Ascend A3 (8 cards, TP8)` |
| image | `quay.io/ascend/vllm-ascend:nightly-releases-v0.20.2rc-a3` |
| model weights | `DeepSeek-V4-Flash-w8a8-mtp` |
| `--max-model-len` | `135000` |
| `--max-num-batched-tokens` | `8192` |
| `--tensor-parallel-size` | `8` |
| `--enable-expert-parallel` | yes |

## Expected impact

| Scenario | hit-rate change | Reason |
|---|---|---|
| Prompts already 128-aligned | unchanged | No new tail to recover |
| Prompts with 1-127 token tail past the last 128-boundary | improves | The tail rounds to a 32-block boundary instead of being thrown away |
| Repeat-prompt traffic (`repeat_rate=0.9`, c32, 8K) | `0% -> 81.54%` | Validated above |
| Short context (< 32 tokens) | unchanged | Below `block_size = 32` floor; no caching possible either way |

No when `VLLM_ASCEND_DSV4_COMPRESSED_KV_BLOCK_SIZE` is unset or set to
`128`. The resolver returns the legacy `128`, the cache spec is
bit-for-bit identical, and the rest of the modified files take their
no-op default branches.

When set to `32` or `64`, the only effective change is the page-layout
of the DSv4 compressor KV cache. Block contents, the cache lookup
math, and the attention compute itself are untouched. A previously
cacheable replay either still hits (its tail block survives) or misses
(its tail block landed in a smaller bucket that was already evicted);
on miss the request recomputes the same tokens it would have
recomputed under any cache miss.

## References

- vLLM version: v0.20.2 (commit `bc150f50299199599673614f80d12a196f377655`)

### DIFF
--- a/tests/ut/core/test_dsv4_partial_prefix_cache.py
+++ b/tests/ut/core/test_dsv4_partial_prefix_cache.py
@@ -1,0 +1,108 @@
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+# This file is a part of the vllm-ascend project.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""Unit tests for the DSv4 partial compressed prefix-cache helpers.
+
+These cover the pure-Python pieces that back ``CompressAttentionManager``'s
+partial-hit support: range hashing and the per-``BlockPool`` partial-cache maps.
+The maps must live on the ``BlockPool`` instance (not module globals), so the
+isolation test below is a regression guard for that requirement.
+"""
+
+from types import SimpleNamespace
+
+from tests.ut.base import TestBase
+from vllm_ascend.core.single_type_kv_cache_manager import (
+    _hash_range,
+    _insert_partial_cache,
+    get_partial_cached_block,
+    remove_partial_cache_entries_for_block,
+)
+
+
+def _block(block_id: int) -> SimpleNamespace:
+    # _insert/get/remove only touch ``block_id``.
+    return SimpleNamespace(block_id=block_id)
+
+
+def _pool() -> SimpleNamespace:
+    # A bare object is enough: the partial-cache maps are attached lazily.
+    return SimpleNamespace()
+
+
+class TestHashRange(TestBase):
+    def setUp(self):
+        # 4 hash-blocks of 8 tokens each -> hashes for tokens [0,8,16,24).
+        self.block_hashes = [b"h0", b"h1", b"h2", b"h3"]
+        self.hash_block_size = 8
+
+    def test_unaligned_returns_none(self):
+        self.assertIsNone(_hash_range(self.block_hashes, self.hash_block_size, 0, 12))
+        self.assertIsNone(_hash_range(self.block_hashes, self.hash_block_size, 4, 16))
+
+    def test_empty_or_inverted_range_returns_none(self):
+        self.assertIsNone(_hash_range(self.block_hashes, self.hash_block_size, 8, 8))
+        self.assertIsNone(_hash_range(self.block_hashes, self.hash_block_size, 16, 8))
+
+    def test_out_of_range_returns_none(self):
+        # end token 40 -> needs 5 hashes, only 4 available.
+        self.assertIsNone(_hash_range(self.block_hashes, self.hash_block_size, 0, 40))
+
+    def test_valid_range_concatenates_hashes(self):
+        self.assertEqual(_hash_range(self.block_hashes, self.hash_block_size, 0, 16), b"h0h1")
+        self.assertEqual(_hash_range(self.block_hashes, self.hash_block_size, 8, 24), b"h1h2")
+
+
+class TestPartialPrefixCache(TestBase):
+    def setUp(self):
+        self.pool = _pool()
+        self.group_id = 0
+
+    def test_insert_and_get(self):
+        block = _block(7)
+        _insert_partial_cache(self.pool, b"abc", self.group_id, block)
+        self.assertIs(get_partial_cached_block(self.pool, b"abc", self.group_id), block)
+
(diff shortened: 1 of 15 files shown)


jhonzy0928:
In scenarios involving separate PDs (8 machines and a large EP), a phenomenon occurs where prefill is completed, but decoding loops infinitely, preventing inference from completing.
pd

Liuchenbing-2026:

PR #10197, opened 2026-06-08
[Feature] Support AscendStore retention masks

Pz1116:
## What

Adapt AscendStore KV Pool to use prefix-cache retention masks for external store save/load, following the MooncakeStore direction in vllm-project/vllm#44774.

This adds retention-aware store/load masking for AscendStore while preserving DeepSeek V4 compressed-family store granularity.

## Why

DeepSeek V4 sparse/sliding-window attention should not require saving/loading every external KV chunk for long contexts. AscendStore needs the same reachable-block retention semantics as the vLLM-side MooncakeStore integration, but mapped onto AscendStore's group-aware keys and cache-family granularity.

## Changes

- Add store/load mask support around AscendStore send/recv paths.
- Pass prompt token counts through scheduler metadata so retention decisions can match vLLM prefix-cache behavior.
- Reuse vLLM cache-manager mask semantics via lightweight test doubles and external-hit adapters.
- Preserve DSV4 C4/C128 cache-family granularity for external store keys.
- Add unit coverage for mask generation and transfer filtering.

## Validation

vLLM version: v0.20.1

- vLLM version: v0.21.0

### DIFF
--- a/tests/ut/distributed/ascend_store/_mock_deps.py
+++ b/tests/ut/distributed/ascend_store/_mock_deps.py
@@ -83,11 +83,13 @@ _vllm_mock_modules = [
     "vllm.v1.attention",
     "vllm.v1.attention.backend",
     "vllm.v1.core",
+    "vllm.v1.core.block_pool",
     "vllm.v1.core.kv_cache_manager",
     "vllm.v1.core.kv_cache_utils",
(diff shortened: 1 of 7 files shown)


Pz1116:
closed as #10393 covers the same functions

PR #10198, opened 2026-06-08
[Feature] Support prefix cache retention patch

Pz1116:
## What

Backport the local prefix-cache retention support from vllm-project/vllm#43447 into vLLM-Ascend's platform patch layer for current vLLM 0.20.x based deployments.

This keeps the change separate from the AscendStore external-store retention work and focuses on local sliding-window prefix-cache retention semantics.

## Why

Current vLLM-Ascend main still carries Ascend-specific KV cache coordinator and cache manager patches. For DeepSeek V4 long-context runs, those patches need to understand the same selective sliding-window retention behavior as vLLM core, otherwise local prefix-cache checkpointing can remain dense and over-retain blocks.

## Changes

- Register `VLLM_PREFIX_CACHE_RETENTION_INTERVAL` for current vLLM versions that do not expose it yet.
- Patch free-queue prepend and `BlockPool` masked caching/freeing support used by sparse retention.
- Add sliding-window reachable-block retention masks following vllm-project/vllm#43447.
- Propagate retention interval through Ascend's hybrid KV cache coordinator.
- Keep compressed MLA groups dense while allowing sliding-window groups to retain sparse checkpoints.
- Add focused unit coverage for env registration, validation, SWA masks, and AscendHybrid propagation.

## Validation

vLLM version: v0.20.1

- vLLM version: v0.23.0


Pz1116:
closed as main2main to vLLM 23.0 is done

PR #10199, opened 2026-06-08
[BugFix] Fix the verification when pipeline parallel + mtp in PD disaggregation scenario

gjc0824:
This pull request backports vLLM Pipeline Parallel (PP) + Multi-Token Prediction (MTP) speculative decoding support for NPU. It introduces patches to handle the alignment of model runner outputs, post-step scheduling, and model config validation. Specifically, it ensures that local Eagle/MTP drafters are validated with a pipeline parallel size of 1 on the last PP stage, and that speculative token IDs are correctly tracked and propagated through the scheduler and model runner.

Feedback on the changes:
- In `scheduler_profiling_chunk.py`, slicing `request.spec_token_ids` in-place without resetting it can lead to duplicate scheduling. It is recommended to slice without in-place mutation and clear the list.
- In `scheduler_profiling_chunk.py`, accessing `model_runner_output.req_id_to_index[req_id]` directly may raise a `KeyError` in distributed environments; using `.get()` is safer.
- In `patch_pp_mtp.py`, importing `MTPModelTypes` directly can cause an `ImportError` on older/newer vLLM versions; wrapping it in a `try...except` block is recommended.
- In `model_runner_v1.py`, fallback to `self.input_batch.req_ids` when `self._draft_token_req_ids` is `None` to prevent silent dropping of speculative tokens.

No, this PR only introduces internal runtime patches and backports for PP+MTP speculative decoding support on NPU.

The changes are covered by new unit tests added in `tests/ut/core/test_profiling_chunk.py`, `tests/ut/patch/platform/test_patch_pp_mtp.py`, `tests/ut/patch/worker/test_patch_qwen3_5_mtp.py`, and `tests/ut/test_platform.py`.

- vLLM version: v0.22.1


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -4581,13 +4584,50 @@ def _check_and_update_cudagraph_mode(
...
+            self.speculative_config
+            and get_pp_group().is_last_rank
+            and (
+                self.speculative_config.use_eagle()
+                or self.speculative_config.uses_extract_hidden_states()
+            )
+        ):
+            assert isinstance(
+                self.drafter,
+                AscendEagleProposer
```
lilinsiman: If only the method is mtp or eagle3, only the first one is required.

vllm_ascend/patch/platform/patch_pp_mtp.py
```diff
@@ -0,0 +1,88 @@
...
+            and getattr(parallel_config, "pipeline_parallel_size", 1) > 1
+        ):
+            # Local Eagle/MTP drafters are loaded on the last PP stage rather
+            # than partitioned across all PP stages. Keep normal target-model
+            # validation intact, but validate these draft models as PP=1.
+            logger.warning(
+                "Validating local Eagle/MTP drafter with pipeline_parallel_size=1 "
+                "because it is loaded locally on the last pipeline stage."
+            )
+            patched_config = copy.copy(parallel_config)
```
Ronald1995, line 67: should we use deepcopy here?

gjc0824: We only modify the int attribute pipeline_parallel_size. I think shallow copy is enough?

vllm_ascend/worker/model_runner_v1.py
gjc0824: We found that in the existing modelrunner, self.drafter is created only on pp_last_rank. The specific code is [_set_up_drafter](https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/worker/model_runner_v1.py#L562). Therefore, we can add a judgment on self.drafter here, which can include all the draft methods that are created only on pp_last_rank.

realliujiaxu (approved):
LGTM

PR #10202, opened 2026-06-08
[BugFix] Fix layer index mapping and block ID flattening in Mooncake connector

gjc0824:
This PR fixes a KV transfer failure in the Mooncake connector under Pipeline Parallel (PP) configurations.
**Root cause**: `self.total_layers` was initialized via `get_num_layers()`, which returns the **per-stage** layer count
(divided by PP rank). However, `extract_layer_index()` returns **global** layer numbers. For example, with a 48-layer model and PP=2, each stage sees `total_layers=24`, but PP rank 1 holds layers 24–47. `next_mtp_layer_idx` was initialized to 24, causing MTP virtual-layer indices to collide with actual base-layer indices. This corrupted per-layer KV cache metadata and produced incorrect KV transfers in P/D scenarios.
**Fix**:
- Use `get_total_num_hidden_layers()` instead of `get_num_layers()` to obtain the global layer count when initializing `self.total_layers`, so MTP virtual-layer indices always start above all base-layer indices.
- Fix `_mark_failed_recv_request` to only invalidate blocks in the full-attention group (`local_block_ids[0]`), as invalidation is only meaningful for the full-attention KV cache group.
No.
Verified manually in a PP=2, TP=2 (Prefill) + DP=2, TP=2 (Decode)
deployment with a 48-layer model and 1 MTP layer.

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
@@ -1710,12 +1710,25 @@ class MooncakeConnectorWorker:
         next_mtp_layer_idx = self.total_layers
         for group_id, group_spec in enumerate(self.kv_cache_config.kv_cache_groups):
             layer_indices = []
+            # For eagle3 method there is no "mtp" in layer names, and upstream model initiation assigns the layer id 
+            # that is sliced by Pipeline Parallel. So the eagle layer id will confilt with target model layers. 
+            # Here we determine whether the current layer is an eagle layer based on whether the layer id has been 
+            # assigned to previous layers. If the layer id has been assigned, we treat the current layer as 
+            # an eagle layer and assign a new layer id starting from total_layers.
+            assigned_indices = set()
             for layer_name in group_spec.layer_names:
                 if "mtp" in layer_name:
                     layer_idx = next_mtp_layer_idx
                     next_mtp_layer_idx += 1
                 else:
                     layer_idx = extract_layer_index(layer_name, num_attn_module)
+                    if (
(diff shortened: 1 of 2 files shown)


LCAIZJ:
cc @lidenghui1110

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
```diff
@@ -1591,7 +1591,7 @@ def __init__(self, vllm_config: VllmConfig, engine_id: str, kv_cache_config: KVC
         self.kv_caches: dict[str, torch.Tensor] = {}
         self.side_channel_host = get_ip()
         self.pcp_size = get_pcp_group().world_size
-        self.total_layers = vllm_config.model_config.get_num_layers(vllm_config.parallel_config)
+        self.total_layers = vllm_config.model_config.get_total_num_hidden_layers()
```
LCAIZJ, line 1594: Was this issue introduced by the previous PR? @liziyu179

liziyu179: > Was this issue introduced by the previous PR? [@liziyu179](https://github.com/liziyu179)

The function _build_kv_group2layeridx constructs the layer-to-group information for all PP groups. During subsequent transmission, this global information is used to obtain the layer index corresponding to the layer name.

@LCAIZJ

Kurumi5210:
the change makes sense and resolves the bug introduced when hybrid attention support was added

ZT-AIA (approved):

liziyu179:
approve

wangxiyuan (approved):

PR #10205, opened 2026-06-08
[BugFix][Performance] Fix MTP copy_valid_sampled_token_count sync

nwpu-zxr:
According to #9656, the `copy_` in 、`copy_valid_sampled_token_count` is forcibly block the host. The root cause is that the dtype of `valid_sampled_token_count_cpu` is `torch.int32`, while that of `valid_sampled_token_count_gpu` is `torch.int64`. As a result, the operator generates implicit dtype cast, causing the blocking synchronization. After the dtype is explicitly converted, the synchronization is eliminated. Therefore, the content of #9656 is rolled back.

No.

By CI.

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -1463,6 +1463,7 @@ class NPUModelRunner(GPUModelRunner):
         default_stream = torch.npu.current_stream()
         with torch.npu.stream(self.valid_sampled_token_count_copy_stream): 
             self.valid_sampled_token_count_copy_stream.wait_stream(default_stream)
+            counts = valid_sampled_tokens_count
             counts_cpu = self.valid_sampled_token_count_cpu
             assert counts_cpu is not None
             counts_cpu[: counts.shape[0]].copy_(counts, non_blocking=True)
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1451,9 +1451,9 @@ def _copy_valid_sampled_token_count(
         # Initialize a new stream to overlap the copy operation with
         # prepare_input of draft model.
         default_stream = torch.npu.current_stream()
-        with torch.npu.stream(self.valid_sampled_token_count_copy_stream):  
-            self.valid_sampled_token_count_copy_stream.wait_stream(default_stream)  
-            counts = valid_sampled_tokens_count
+        with torch.npu.stream(self.valid_sampled_token_count_copy_stream): 
+            self.valid_sampled_token_count_copy_stream.wait_stream(default_stream)
+            counts = valid_sampled_tokens_count.to(self.valid_sampled_token_count_cpu.dtype)
```
Ronald1995: we can just redefine dtype of `self.valid_sampled_token_count_cpu` in init phase, not convert the dtype of valid_sampled_tokens_count in every decode phase.

nwpu-zxr: `gpu_model_runner` use Triton to compute `valid_sampled_token_count`, and its output format is `torch.int32`. However, currently, NPU use the torch operators with `sample_token_ids`  and `torch.sum()` to calculate it, whose output format is `torch.int64`. I will re-init `valid_sampled_token_count_cpu` in place of cast dtype.

ZT-AIA (approved):

linfeng-yuan (approved):
LGTM

PR #10217, opened 2026-06-09
[BugFix] Align AscendStore grouped hash lookup encoding

Pz1116:

Follow-up to #9789. The scheduler lookup RPC serializes block hashes as hex strings, while the worker store path rehashes raw `BlockHash` bytes. After grouped rehashing was introduced, hashing hex string bytes and raw bytes produced different grouped digests, so lookup and store keys could diverge.

This PR minimally updates `_block_hash_to_bytes()` to decode hex strings back to raw bytes before rehashing, while preserving the existing UTF-8 fallback for non-hex strings.

No. This only fixes internal AscendStore KV pool key generation consistency.

### Validation

vLLM version: v0.20.1

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
@@ -476,8 +476,11 @@ def _rehash_block_hash_group(block_hashes: Sequence[BlockHash | str]) -> BlockHa
 
 def _block_hash_to_bytes(block_hash: BlockHash | str) -> bytes:
     if isinstance(block_hash, str):
-        if len(block_hash) == 64 and all(char in "0123456789abcdefABCDEF" for char in block_hash):
-            return bytes.fromhex(block_hash)
+        if len(block_hash) == 64:
+            try:
+                return bytes.fromhex(block_hash)
+            except ValueError:
+                return block_hash.encode("utf-8")
         return block_hash.encode("utf-8")
     return bytes(block_hash)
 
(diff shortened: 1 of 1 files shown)


vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
```diff
@@ -476,6 +476,10 @@ def _rehash_block_hash_group(block_hashes: Sequence[BlockHash | str]) -> BlockHa
 
 def _block_hash_to_bytes(block_hash: BlockHash | str) -> bytes:
     if isinstance(block_hash, str):
+        try:
+            return bytes.fromhex(block_hash)
+        except ValueError:
+            pass
```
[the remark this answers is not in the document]
Pz1116: Updated in 50d83aff: hex decoding is now limited to 64-character strings, which matches the SHA-256 block hash hex representation from the lookup RPC path. Short mock/non-hash strings keep the existing UTF-8 fallback behavior.

ZT-AIA (approved):

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
```diff
@@ -476,6 +476,11 @@ def _rehash_block_hash_group(block_hashes: Sequence[BlockHash | str]) -> BlockHa
 
 def _block_hash_to_bytes(block_hash: BlockHash | str) -> bytes:
     if isinstance(block_hash, str):
+        if len(block_hash) == 64:
+            try:
+                return bytes.fromhex(block_hash)
+            except ValueError:
+                pass
```
Ronald1995: why just do noting when there is an exception? don't we need to raise it?

Pz1116: Fixed. We now explicitly check for a 64-character hex string before decoding; non-hex strings keep the UTF-8 fallback.

LCAIZJ (approved):

PR #10222, opened 2026-06-09
[Misc] improve error messages for quantization mismatches and netloader config validation  

xuchi-0808:

This PR improves error messages for three user-facing scenarios identified during code audit, replacing bare Python exceptions with clear guidance:

**1. Unsupported ModelSlim quantization scheme (`modelslim_config.py`)**

When a model is quantized with a scheme that vLLM Ascend recognizes but doesn't fully implement (e.g., `quant_type=W4A16` registered for `moe` but not for `linear`), the error previously only stated the fact without suggesting a resolution:

```text
Before: NotImplementedError: Currently, vLLM Ascend doesn't support quant_type=W4A16 for layer_type=linear.
After:  NotImplementedError: ... Please use a supported quantization format or load the model with its original float weights.
```

**2. NetLoader crashes with bare AttributeError on malformed config JSON (`netloader.py`)**

When a user passes a non-dict JSON value (e.g., a string or array) to `--model-loader-extra-config`, `config.get()` on a `str`/`list` raises a bare `AttributeError`:

```text
Before: AttributeError: 'str' object has no attribute 'get'
After:  RuntimeError: NetLoader requires --model-loader-extra-config to be a JSON object.
```

The fix adds an `isinstance(extra, dict)` check, following the same pattern already used by `rfork_loader.py:43-46`.

- Fixes # (internal code audit, no GitHub issue)

Yes — error messages for three scenarios are changed:

1. When loading a ModelSlim model with an unsupported quantization scheme (e.g., W4A16 + linear layer), the `NotImplementedError` now includes usage guidance.
2. When `--model-loader-extra-config` contains a non-dict JSON value for NetLoader, a `RuntimeError` with format guidance replaces the bare `AttributeError`.

- Pre-commit CI passed (ruff-check, ruff-format, codespell, typos, signoff-commit, etc.)
- All three changes are purely error message/logging improvements — no functional logic change
- Verified on Ascend 910B4 (NPU 4) with `vllm serve` end-to-end for compressed-tensors and netloader; modelslim verified via Python unit test

#### End-to-end test results (actual server output)

**1. `vllm serve /home/model_weights/Qwen3-0.6B --quantization compressed-tensors`**

```text
(APIServer pid=41671) ERROR 06-09 15:25:57 [compressed_tensors_config.py:70]
    The model you are trying to load does not contain a valid compressed-tensors
    quantization configuration. Please ensure the model was quantized with
    compressed-tensors, or remove '--quantization' if loading an unquantized float model.
...
ValueError: The model you are trying to load does not contain a valid compressed-tensors
    quantization configuration. Please ensure the model was quantized with compressed-tensors,
    or remove '--quantization' if loading an unquantized float model.
```

**2. Python unit test for modelslim unsupported quant_type (via unittest.mock)**

```text
ERROR 06-09 15:25:32 [modelslim_config.py:401]
    Currently, vLLM Ascend doesn't support quant_type={...}
    for layer_type=linear. Please use a supported quantization format
    or load the model with its original float weights.
NotImplementedError: ... Please use a supported quantization format
    or load the model with its original float weights.
```

- vLLM version: v0.22.1


vllm_ascend/quantization/compressed_tensors_config.py
```diff
@@ -59,15 +59,29 @@ class AscendCompressedTensorsConfig(QuantizationConfig):
...
+                "compressed-tensors quantization configuration, "
+                "but you specified '--quantization compressed-tensors'. "
+                "Please remove --quantization for float models, "
+                "or ensure the model was quantized with compressed-tensors."
+            )
+            raise ValueError(
+                "The model is not quantized with compressed-tensors. "
+                "Please remove '--quantization compressed-tensors' from your command "
+                "when loading a float model."
+            )
```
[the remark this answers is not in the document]
xuchi-0808: Thank you for the catch. You're right — this code path can also be reached via auto-detection when the model's config.json contains a malformed quantization_config, not only when the user explicitly passes --quantization compressed-tensors.      
                                                                                                                             
I've updated both the logger.error and ValueError messages to be more general: they now describe the actual problem (invalid quantization config) and offer both possible resolutions (ensure the model is quantized, or remove --quantization for float models), without assuming the user specified the flag.

ZT-AIA (approved):

vllm_ascend/quantization/modelslim_config.py
```diff
@@ -393,7 +393,11 @@ def create_scheme_for_layer(
     if scheme_cls is not None:
         return scheme_cls()
 
-    err_msg = f"Currently, vLLM Ascend doesn't support quant_type={quant_type} for layer_type={layer_type}."
+    err_msg = (
+        f"Currently, vLLM Ascend doesn't support quant_type={quant_type} "
+        f"for layer_type={layer_type}. Please use a supported quantization format "
+        "or load the model with its original float weights."
+    )
```
Ronald1995, line 400: we should % string instead of using f-string in logger.

vllm_ascend/quantization/compressed_tensors_config.py
```diff
@@ -59,12 +59,23 @@ class AscendCompressedTensorsConfig(QuantizationConfig):
...
-        ignore: list[str],
-        quant_format: str,
+        target_scheme_map: dict[str, Any] | None = None,
+        ignore: list[str] | None = None,
+        quant_format: str | None = None,
         config: dict[str, Any] | None = None,
     ):
         super().__init__()
+
+        if target_scheme_map is None or ignore is None or quant_format is None:
```
MengqingCao: I think it is enough to keep the input args as is, instead of giving it a default value but raise error when it is the default value.

MengqingCao (approved):

PR #10237, opened 2026-06-09
[BugFix] Guard against all-placeholder speculative output

QwertyJack:
### Problem
Speculative rejection sampling can silently return a row that is entirely `PLACEHOLDER_TOKEN_ID` (`-1`). When `RejectionSampler.parse_output` filters that row, the scheduler receives an empty token list, so the request can remain `RUNNING` without making progress in PD serving.

This has been observed under parallel PD workloads; using TP=2 only lowers the reproduction probability and does not eliminate the failure, so this guard is still needed as a defensive fallback while the hidden-state NaN root cause is investigated.

### Fix
Add a final guard before `rejection_sample` returns `output_token_ids`. The guard only repairs rows that are all placeholders:

- rows with draft tokens first use the corresponding first `recovered_token_ids` entry
- if that recovered token is still `-1`, fall back to the target argmax token, including `target_indices` mapping for reduce sampling
- rows without draft tokens use `bonus_token_ids`
- rows that already contain any real token are left unchanged

### Verification
- `python -m py_compile vllm_ascend/sample/rejection_sampler.py tests/ut/sample/test_rejection_sampler.py`
- `python -m pytest tests/ut/sample/test_rejection_sampler.py::TestAscendRejectionSampler::test_repair_empty_output_rows -q`
- `python -m pytest tests/ut/sample/test_rejection_sampler.py -q`
- `python -m ruff check vllm_ascend/sample/rejection_sampler.py tests/ut/sample/test_rejection_sampler.py`
- `python -m ruff format --check vllm_ascend/sample/rejection_sampler.py tests/ut/sample/test_rejection_sampler.py`

### DIFF
--- a/tests/ut/sample/test_rejection_sampler.py
+++ b/tests/ut/sample/test_rejection_sampler.py
@@ -4,6 +4,7 @@ import torch
 
 from tests.ut.base import TestBase
 from vllm_ascend.sample.rejection_sampler import (
+    _repair_empty_output_rows,
     expand_batch_to_tokens,
     expand_pytorch,
     rejection_greedy_sample_pytorch,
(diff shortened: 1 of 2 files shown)


QwertyJack:
Fixed by PR #10254 / #10255

Issue #8938, opened 2026-05-07
[Bug]: 在 1P2 + KV Pool 场景下，P0 服务拉起成功后，偶现P1 服务拉起失败，报错 ZMQError: Address already in use，P0、P1 服务在极短时间内同时拉起时 ZMQ 端口绑定无重试机制，存在端口抢占现象


zmc1997:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
[root@cloudideworkspaceapp-0-010233118009 work]# python collect_evn.py
Collecting environment information...
PyTorch version: 2.9.0+cpu
Is debug build: False

OS: openEuler 24.03 (LTS) (aarch64)
GCC version: (GCC) 12.3.1 (openEuler 12.3.1-38.oe2403)
Clang version: Could not collect
CMake version: version 4.3.1
Libc version: glibc-2.38

Python version: 3.11.14 (main, Feb  1 2026, 10:20:28) [GCC 12.3.1 (openEuler 12.3.1-38.oe2403)] (64-bit runtime)
Python platform: Linux-5.10.0-182.0.0.95.r1941_123.hce2.aarch64-aarch64-with-glibc2.38

CPU:
Architecture:                         aarch64
CPU op-mode(s):                       64-bit
Byte Order:                           Little Endian
CPU(s):                               320
On-line CPU(s) list:                  0-319
Vendor ID:                            HiSilicon
Model name:                           unknown
Model:                                0
Thread(s) per core:                   1
Core(s) per cluster:                  80
Socket(s):                            -
Cluster(s):                           4
Stepping:                             0x0
Frequency boost:                      disabled
CPU(s) scaling MHz:                   100%
CPU max MHz:                          3000.0000
CPU min MHz:                          400.0000
BogoMIPS:                             200.00
Flags:                                fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 sm3 sm4 asimddp sha512 sve asimdfhm dit uscat ilrcpc flagm ssbs sb paca pacg dcpodp flagm2 frint svei8mm svef32mm svef64mm svebf16 i8mm bf16 dgh rng ecv
L1d cache:                            20 MiB (320 instances)
L1i cache:                            20 MiB (320 instances)
L2 cache:                             400 MiB (320 instances)
L3 cache:                             560 MiB (8 instances)
NUMA node(s):                         8
[...]

PR #10260, opened 2026-06-10
[EPLB][Test] Change mooncake port

shenchuxiaofugui:
If mooncake uses port 36000, it is prone to being occupied. Therefore, the port is changed to 30000 to remain consistent with other use cases.

The port occupation issue is tracked by issue #8938 .

- vLLM version: v0.21.0


realliujiaxu (approved):
LGTM

PR #10266, opened 2026-06-10
[BugFix] [Kimi] Fix dense draft model (Kimi-MLA) misidentified as MoE

baolongsun:

Background & Issue:
Currently, vLLM's configuration loading mechanism automatically propagates the target (main) model's configuration attributes to the draft model during speculative decoding setup.

This causes a critical issue when using Kimi-K2.6 based speculative draft models (specifically lightseekorg/kimi-k2.6-eagle3-mla) alongside an MoE main model in an Ascend (NPU) environment where Sequence Parallelism is enabled via export VLLM_ASCEND_ENABLE_FLASHCOMM1=1.

Although Kimi-K2.6 is a dense model utilizing the Eagle3 architecture, the configuration inheritance erroneously passes the main model's MoE attributes to it. As a result, vLLM misidentifies the dense draft model as an MoE model.

Under the VLLM_ASCEND_ENABLE_FLASHCOMM1=1 environment, vLLM executes specialized optimization and communication pathways for MoE models. Because the draft model is misclassified as MoE, it illegally enters these pathways during initialization, throwing fatal errors and failing to boot.

No. This is an internal bugfix for the configuration propagation logic in speculative decoding. It does not alter any public APIs, but it enables proper out-of-the-box support for Kimi-K2.6 as a draft model.

1. Manual Verification:Verified by launching the vLLM server on an Ascend environment with an MoE main model and `lightseekorg/kimi-k2.6-eagle3-mla` as the speculative draft model.
   
   Environment Setup:
```bash
     export VLLM_ASCEND_ENABLE_FLASHCOMM1=1
  ```
   Before this patch:The initialization failed because the draft model was misidentified as MoE, causing the `VLLM_ASCEND_ENABLE_FLASHCOMM1` workflow to throw a configuration/runtime error on the dense draft model.
   After this patch: The server launched successfully; the draft model was correctly treated as dense and successfully bypassed the MoE-FlashComm execution path.

2. Regression Testing:Checked against other standard dense/MoE draft model combinations to ensure no side effects on standard speculative decoding pipelines.

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -894,7 +894,7 @@ def is_drafter_moe_model(vllm_config: VllmConfig):
         model_configs = vllm_config.speculative_config.draft_model_config.hf_text_config.to_dict()
         _IS_DRAFTER_MOE_MODEL = _is_contain_expert(model_configs)
         if not model_configs or not model_configs.get("architectures"):
-            return _IS_DRAFTER_MOE_MODEL       
+            return _IS_DRAFTER_MOE_MODEL
         if "Eagle3DeepseekV2ForCausalLM" in model_configs["architectures"]:
             _IS_DRAFTER_MOE_MODEL = False
     return _IS_DRAFTER_MOE_MODEL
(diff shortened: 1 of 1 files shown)


realliujiaxu (approved):
LGTM

PR #10285, opened 2026-06-10
[Scheduler][Feature] Implement LAPS-Inspired Prefill Scheduling for vllm-ascend

immengzi:

This PR implements **LAPS (Length-Aware Prefill Scheduling)** for the Ascend recompute scheduler in `vllm-ascend`.

Under FCFS prefill scheduling, a long prompt at the head of the waiting queue blocks every shorter prompt behind it (head-of-line blocking), inflating the latency of short requests under load. LAPS splits the waiting queue into three length-aware lanes (immediate, short, and long) and schedules short prefills ahead of long ones, while bounding how badly long requests can be starved using a token-bucket anti-starvation aging mechanism.

Yes, but it is **off by default** and additive:
- A new optional `additional_config["laps_config"]` block is introduced.
- When enabled without `recompute_scheduler_enable=true`, or with a non-FCFS policy, LAPS is not activated and a warning is logged.

- **Unit tests** (`tests/ut/core/test_laps_scheduler.py`, `test_laps_scheduler_mixin.py`, `test_recompute_scheduler_laps.py`, and `test_ascend_config.py`) covering config validation, queue routing, token-bucket aging, and recompute-scheduler integration.
- **E2E tests** (`tests/e2e/singlecard/test_laps_scheduling.py`) asserting LAPS produces output consistent with the default scheduler on a single card.

### DIFF
--- a/.github/workflows/scripts/config.yaml
+++ b/.github/workflows/scripts/config.yaml
@@ -15,6 +15,8 @@ e2e-singlecard:
   estimated_time: 187
 - name: tests/e2e/singlecard/test_async_scheduling.py
   estimated_time: 252
+- name: tests/e2e/singlecard/test_laps_scheduling.py
+  estimated_time: 150
 - name: tests/e2e/singlecard/test_batch_invariant.py
   estimated_time: 506
 - name: tests/e2e/singlecard/test_camem.py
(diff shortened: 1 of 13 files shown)


immengzi:
#11576  merged

PR #10289, opened 2026-06-10
[CI][BugFix] Checkout PR ref instead of default branch in /e2e workflow

zhangxinyuehfad:

The Checkout repo step previously checked out the repository's default branch, causing new test files added in a PR to fail validation since they did not exist on the default branch. This prevented PR authors from running /e2e on newly added test cases.

- vLLM version: v0.25.1


wxsIcey (approved):
Thanks for the fix!

menogrey:
LGTM, add a /e2e command test will be better.

wangxiyuan (approved):

PR #10295, opened 2026-06-10
[Feat][Attention] Support DSA-CP o-proj TP allgather

TmacAaron:
Depends on #10026. This PR is intentionally stacked on top of the DSV4 DSA-CP A5 support from #10026; the diff will become clean after #10026 is merged into `main`.

This PR adds the missing DSV4 DSA-CP o-proj TP full-weight allgather path for prefill.

When DSA-CP runs with o-proj TP enabled, prefill now asynchronously allgathers the TP-local `wo_a` / `wo_b` projection weights, including MXFP8 weight scales when present, switches the projection layers to the gathered full buffers for the o-proj computation, and restores the TP-local tensors afterward. Decode and spec-decode keep the existing activation all-to-all behavior.

The PR also patches post-load weight processing so `DSAAttention.process_weights_after_loading()` is invoked from the same loader hook as other attention modules, allowing DSA-CP full-weight buffers to be initialized after weights are loaded instead of lazily in forward.

No user-facing API, configuration, environment variable, or schema changes. The change only affects the internal DSV4 DSA-CP o-proj TP execution path when the existing o-proj TP enablement is active.

- `python -m py_compile vllm_ascend/attention/context_parallel/dsa_cp.py vllm_ascend/patch/worker/patch_process_weights_after_loading.py`
- `git diff --cached --check`

`ruff` was not run locally because it is not installed in the current environment.

- vLLM version: v0.21.0

### DIFF
--- a/csrc/attention/quant_lightning_indexer/op_host/quant_lightning_indexer_tiling.cpp
+++ b/csrc/attention/quant_lightning_indexer/op_host/quant_lightning_indexer_tiling.cpp
@@ -873,13 +873,20 @@ ge::graphStatus QuantLightningIndexerTiling::DoTiling(QLITilingInfo *tilingInfo)
     constexpr uint32_t TOPK_MAX_SIZE = 2048;           // TopK选取个数
     uint32_t workspaceSize = ascendcPlatform.GetLibApiWorkSpaceSize();
     // 主流程需Workspace大小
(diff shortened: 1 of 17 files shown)


TmacAaron:
Closing this stacked draft PR because #10026 has merged and the clean main-based PR is now available at #10694.

PR #10306, opened 2026-06-11
[Feature] [Logger] Add file logging with daily rotation and configurable path for vllm-ascend

lizy124:
This PR introduces structured file logging for vllm-ascend. Previously, vllm-ascend log messages were indistinguishable from upstream vLLM logs in the console output, and there was no persistent log file for post-mortem debugging or production monitoring.

Yes. Users will see [vllm-ascend] prefixed log messages in the console output, and a new log file will be created at ~/ascend/log/vllm_ascend/vllm_ascend.log (or the custom path specified via additional_config.ascend_log_path ). Old log files rotate daily and are retained for 7 days.

Unit tests added in tests/ut/test_logger.py covering:
- _is_ascend_module — positive/negative matching, empty string edge case
- _infer_module_name — root-level files, nested modules, edge cases (empty path, non-ascend path)
- AscendFormatter — prefix insertion for root files, nested files, pass-through for vLLM logs
- AscendColoredFormatter — prefix insertion works correctly

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/logger.py
+++ b/vllm_ascend/logger.py
@@ -9,8 +9,10 @@ Module identification is inferred from record.pathname.
 
 import logging
 import os
+import sys
 from datetime import datetime, timedelta
 
+from vllm import envs
 from vllm.logging_utils import ColoredFormatter, NewLineFormatter
 
 _FORMAT = "%(levelname)s %(asctime)s [%(fileinfo)s:%(lineno)d] %(message)s"
@@ -19,6 +21,19 @@ _DATE_FORMAT = "%m-%d %H:%M:%S"
 _LOG_DIR = os.path.join(os.path.expanduser("~"), "ascend", "log", "vllm_ascend")
 _LOG_RETENTION_DAYS = 7
 _LOG_MAX_BYTES = 20 * 1024 * 1024
+
+
+def _use_color() -> bool:
+    """Determine if colored output should be used."""
+    if envs.NO_COLOR or envs.VLLM_LOGGING_COLOR == "0":
+        return False
+    if envs.VLLM_LOGGING_COLOR == "1":
+        return True
+    if envs.VLLM_LOGGING_STREAM == "ext://sys.stdout":
+        return hasattr(sys.stdout, "isatty") and sys.stdout.isatty()
+    elif envs.VLLM_LOGGING_STREAM == "ext://sys.stderr":
+        return hasattr(sys.stderr, "isatty") and sys.stderr.isatty()
+    return False
 
 
 def _is_ascend_module(pathname: str) -> bool:
@@ -75,13 +90,6 @@ class AscendColoredFormatter(ColoredFormatter):
 
     def format(self, record):
         return _format_with_ascend_prefix(self, record, super().format)
-
-
-def _patch_handler(handler: logging.Handler) -> None:
-    if isinstance(handler.formatter, ColoredFormatter):
-        handler.formatter = AscendColoredFormatter(fmt=_FORMAT, datefmt=_DATE_FORMAT)
-    elif isinstance(handler.formatter, NewLineFormatter):
-        handler.formatter = AscendFormatter(fmt=_FORMAT, datefmt=_DATE_FORMAT)
 
 
 class RotatingAscendFileHandler(logging.FileHandler):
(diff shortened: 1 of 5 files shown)


vllm_ascend/logger.py
```diff
@@ -78,6 +83,55 @@ def _patch_handler(handler: logging.Handler) -> None:
...
+_file_handler: logging.Handler | None = None
+
+
+def _setup_file_logging(log_dir: str | None = None) -> None:
+    global _file_logging_configured, _file_handler
+    if _file_logging_configured:
+        return
+    target_dir = log_dir or _LOG_DIR
+    os.makedirs(target_dir, exist_ok=True)
+    log_file = os.path.join(target_dir, "vllm_ascend.log")
```
momo609: It is necessary to ensure that the directory has write permission. Should this part include a validation check?

vllm_ascend/logger.py
```diff
@@ -78,6 +83,66 @@ def _patch_handler(handler: logging.Handler) -> None:
...
+            if not entry.name.startswith("vllm_ascend_") or not entry.name.endswith(".log"):
+                continue
+            mtime = datetime.fromtimestamp(entry.stat().st_mtime)
+            if mtime < cutoff:
+                os.remove(entry.path)
+    except OSError:
+        pass
+
+
+def _setup_file_logging(log_dir: str | None = None) -> None:
```
momo609, line 137: 1.Any code that imports vllm_ascend.logger (or any module that indirectly imports it) will trigger file system operations (os.makedirs) and log handler registration.

2.At this point, ascend_config has not been initialized, so the default path ~/ascend/log/vllm_ascend is always used.

3. Later, on line 439 of platform.py, configure_ascend_file_logging() is called again. If the user has configured a different ascend_log_path, this results in a handler being created for the default directory first, then torn down and recreated, causing unnecessary I/O.

vllm_ascend/logger.py
lizy124: Thanks for the review! Regarding the write permission check — the current code follows Python's EAFP ("Easier to Ask for Forgiveness than Permission") pattern. If the directory is not writable, os.makedirs() or the subsequent RotatingAscendFileHandler construction will raise a PermissionError / OSError , which propagates up through configure_ascend_file_logging() and ultimately causes the engine initialization to fail with a clear error message. This is actually the desired behavior — if we can't write logs, the user should know immediately rather than silently skipping file logging.

Adding a separate os.access() pre-check would be redundant and also introduce a TOCTOU race condition (permissions could change between the check and the actual operation). So I think the current approach is sufficient here.

vllm_ascend/logger.py
lizy124: Good catch on the concern about import-time side effects! However, _setup_file_logging() is not called at module level in the current code — the module body of logger.py only defines constants, functions, and classes, with zero function calls or filesystem operations. The os.makedirs call is only reachable through configure_ascend_file_logging() , which is explicitly invoked at two call sites:

- platform.py:L442 — called after init_ascend_config() at L437
- worker.py:L115 — called after init_ascend_config() at L109
Since ascend_config is already initialized at both call sites, get_ascend_config() returns the user-configured ascend_log_path on the very first call, so there is no scenario where a handler is created for the default directory first and then torn down/recreated. The _file_logging_configured guard also prevents any re-initialization.

The import vllm_ascend.logger in __init__.py only loads the module definitions — it does not trigger any filesystem I/O. This is different from the old _patch_vllm_formatter() approach, which was explicitly removed in this PR.

vllm_ascend/platform.py
```diff
@@ -444,6 +444,12 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
 
         ascend_config = init_ascend_config(vllm_config)
 
+        from vllm_ascend.logger import configure_ascend_file_logging
+        from vllm_ascend.logger import configure_ascend_logging
```
MengqingCao, line 448: plz avoid to use lazy import.

lizy124: Thank you for the review!

The lazy import here is intentional to avoid circular import issues. Importing these modules at the top level could create circular dependencies with vLLM's internal modules. Using lazy imports inside the function ensures they are only loaded when needed, avoiding circular import errors at module initialization time.

MengqingCao (approved):
LGTM now.

PR #10307, opened 2026-06-11
[BugFix][SpecDecode] Sanitize MTP placeholders before model forward

### LINKED ISSUE #10305: [BugFix] PD MTP placeholder token can reach model forward input_ids
## Summary

In DeepSeek V4 Flash 1P1D PD-disaggregated serving, the D-side can crash when MTP/spec decode placeholder tokens are scheduled under a small token budget, for example:

```bash
--max-num-batched-tokens 8 \
--max-num-seqs 4
```

A single `hi` request is enough to trigger the D-side failure in the affected setup.

## Reproduction Context

- Model: DeepSeek V4 Flash w8a8 MTP
- Scenario: 1P1D PD disaggregation
- P side: first 8 cards, `dp=2,tp=4,ep`
- D side: last 8 cards, `dp=8,tp=1,ep`
- D side: MTP/spec decode enabled, async scheduling enabled, recompute scheduler enabled
- Triggering D-side scheduler output includes:

```text
scheduled_spec_decode_tokens={...: [-1]}
num_scheduled_tokens={...: 2}
```

The observed NPU fault is a GatherV2 out-of-range/aivec error, later surfaced through async runtime synchronization as `507011` / `rtMemcpyAsyncEx` failure.

## Root Cause

`PLACEHOLDER_TOKEN_ID = -1` is valid as a speculative decoding placeholder for scheduler/sampler semantics. It must not be used as a real model token id.

The sampler-side hardening in #10043 / #10259 prevents placeholder draft tokens from being used as probability table indices, but the D-side model runner can still use the same `input_ids.gpu` buffer for both:

1. building `SpecDecodeMetadata` with draft placeholder semantics, and
2. feeding target/drafter model forward.

As a result, the placeholder `-1` can reach model forward embedding/GatherV2 and crash the NPU runtime.

## Expected Behavior

The scheduler/sampler path should preserve placeholder semantics, but model forward should only receive valid token ids.

## Proposed Fix

After `SpecDecodeMetadata` has been built, sanitize placeholder token ids only in the current forward input range before `_preprocess()`/model forward. This preserves `draft_token_ids=-1` for the rejection sampler while preventing invalid token ids from entering model forward.

Additional tests should cover:

- sampler metadata still sees placeholder draft ids;
- forward input buffer is sanitized only for the current forward range;
- `num_speculative_tokens > 1` / MTP3 all-placeholder case rejects at the first placeholder and does not submit dummy-token results.

## Validation

A local real-machine smoke run with the fix used:

- P: devices `0-7`, `dp2tp4`
- D: devices `8-15`, `dp8tp1`
- D: `--max-num-batched-tokens 8 --max-num-seqs 4`
- `D_ENABLE_SPEC=1`, `D_ENABLE_ASYNC=1`, `D_ENABLE_RECOMPUTE=1`

Results:

- P `/v1/models`: 200
- D `/v1/models`: 200
- single `hi` streaming request: completed successfully
- 4 concurrent `hi` requests: all completed successfully
- D metrics after requests: running=0, waiting=0
- server logs did not contain `507011`, `GatherV2`, `out of range`, or scheduler dump errors

QwertyJack:
### What this PR does
Fixes #10305.

MTP placeholder token IDs are valid scheduler/sampler metadata, especially on the PD kv-consumer first step, but they must not reach model forward `input_ids` because embedding kernels reject `-1` token IDs.

This PR sanitizes only the current forward GPU input buffer after `SpecDecodeMetadata` has been built, replacing `PLACEHOLDER_TOKEN_ID` with token id `0`. It complements the sampler-side placeholder handling in #10043 and #10259.

### Why it is safe
- `SpecDecodeMetadata.draft_token_ids` is computed before sanitization, so sampler/rejection metadata keeps the original placeholders.
- Only `self.input_ids.gpu[:num_forward_tokens]` is mutated; CPU token history is unchanged.
- Tokens outside the current forward range are unchanged.

### Validation
Real-machine smoke: DeepSeek V4 Flash PD 1P1D, P `dp2tp4` on devices `0-7`, D `dp8tp1` on devices `8-15`, D launched with `--max-num-batched-tokens 8 --max-num-seqs 4`; single `hi` and 4 concurrent `hi` requests completed, D metrics returned `running=0 waiting=0`, and logs had no `507011`, `GatherV2`, `out of range`, or scheduler dump.

- vLLM version: v0.21.0

### DIFF
--- a/tests/ut/sample/test_rejection_sampler.py
+++ b/tests/ut/sample/test_rejection_sampler.py
@@ -152,6 +152,42 @@ class TestAscendRejectionSampler(TestBase):
     @patch("torch.ones", new=mock_pin_memory(torch.ones))
     @patch("torch.full", new=mock_pin_memory(torch.full))
     @patch("torch.tensor", new=mock_pin_memory(torch.tensor))
+    def test_rejection_random_sample_pytorch_rejects_all_placeholder_mtp3(self):
+        batch_size = 1
+        max_spec_len = 3
+        output_token_ids = torch.full((batch_size, max_spec_len + 1), PLACEHOLDER_TOKEN_ID)
+
+        cu_num_draft_tokens = torch.tensor([3])
+        draft_token_ids = torch.tensor([PLACEHOLDER_TOKEN_ID, PLACEHOLDER_TOKEN_ID, PLACEHOLDER_TOKEN_ID])
+        # Placeholder draft tokens must reject regardless of target probability.
+        # The recovered token is passed in after recovery sampling.
+        target_probs = torch.zeros((max_spec_len, 3))
+        bonus_token_ids = torch.tensor([[100]])
+        recovered_token_ids = torch.tensor([2, 1, 0])
+        uniform_probs = torch.tensor([0.0, 0.0, 0.0])
+        is_greedy = torch.tensor([False])
+
+        rejection_random_sample_pytorch(
+            output_token_ids,
+            cu_num_draft_tokens,
+            draft_token_ids,
+            None,
+            target_probs,
+            bonus_token_ids,
+            recovered_token_ids,
+            uniform_probs,
+            is_greedy,
+            max_spec_len,
+            vocab_size=3,
+            IS_NGRAM=True,
+        )
+
+        assert output_token_ids.tolist() == [[2, PLACEHOLDER_TOKEN_ID, PLACEHOLDER_TOKEN_ID, PLACEHOLDER_TOKEN_ID]]
+
+    @patch("torch.arange", new=mock_pin_memory(torch.arange))
+    @patch("torch.ones", new=mock_pin_memory(torch.ones))
+    @patch("torch.full", new=mock_pin_memory(torch.full))
+    @patch("torch.tensor", new=mock_pin_memory(torch.tensor))
     def test_sample_recovered_tokens_pytorch_keeps_placeholder_distribution(self):
         output_token_ids = torch.empty(1, dtype=torch.int32)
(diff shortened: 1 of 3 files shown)


tests/ut/sample/test_rejection_sampler.py
```diff
@@ -148,6 +148,46 @@ def test_rejection_random_sample_pytorch_rejects_placeholder(self):
 
         assert output_token_ids.tolist() == [[2, PLACEHOLDER_TOKEN_ID]]
 
+    @patch("torch.arange", new=mock_pin_memory(torch.arange))
```
momo609, line 151: all draft_token_ids are PLACEHOLDER_TOKEN_ID: in this scenario, target_probs is still set to specific probability values. In a real scenario, if all draft tokens are placeholders, would target_probs still be meaningful?

weijinqian0 (approved):

weijinqian0 (approved):

tests/ut/sample/test_rejection_sampler.py
QwertyJack: Good point. I updated this test to make the intent explicit: placeholder draft tokens are forced to reject regardless of target probability, and `recovered_token_ids` is already passed in after recovery sampling. The `target_probs` value in this direct rejection-sampler test is now just a zero-shaped placeholder instead of one-hot probabilities. The separate `sample_recovered_tokens_pytorch` test still covers the case where target probabilities are meaningful for producing the recovered token.

PR #10311, opened 2026-06-11
[Feature] [BugFix] DSV3.1-style fine-grained TP for DeepSeek-V4 DSA o_proj and embedding (backup) 

xmpp777:
Enable fine-grained tensor parallelism on the o_proj (wo_a/wo_b) weights of DeepSeek-V4's DSA (DeepSeek Attention) module via all-to-all collective communication, and make the embedding TP path graph-stable so it runs correctly under ACL graph capture/replay.

Key changes:
- Replace the enforce_eager restriction for oproj_tensor_parallel_size with a tensor_parallel_size == 1 assertion, so oproj TP works in both eager and graph (CUDAGraph/ACL graph) modes but is currently valid only when standard TP size is 1 (wo_a/wo_b are sharded by the OTP group, orthogonal to standard TP).
- Add _forward_o_proj method that performs all-to-all exchange on the group dimension, then applies wo_a batch matmul + wo_b linear with reduce_scatter, enabling independent TP for the o_proj weights.
- Rewrite _forward_embed_tp in AscendVocabParallelEmbedding to use address-stable static all_gather/reduce_scatter buffers (dist.all_gather_into_tensor/reduce_scatter_tensor) instead of fresh per-call tensors, fixing HCCL operator desync during ACL graph replay that corrupted embedding output.
- Force allow_dp_padding when OTP is enabled, so all DP ranks exchange the same num_tokens and the all_to_all/reduce_scatter shapes stay identical across capture and replay.
- Use graph-stable static HCCL buffers in both paths (allocated once to the max capture capacity on the profiling run) and address them via a contiguous prefix slice of num_tokens rather than padding to the full max bucket, so collective traffic is only num_tokens wide.
- Update the dsa_forward profiling path to delegate to impl.forward() so that the o_proj TP collective ops are captured in the ACL graph during warmup, keeping graph capture consistent.
- Add DSV4OProjColumnParallelOp (for wo_a) and DSV4OProjRowParallelOp (for wo_b) custom parallel ops that use the OTP (oproj TP) comm group.
- Route wo_a through DSV4OProjColumnParallelOp and wo_b through DSV4OProjRowParallelOp when oproj_tp_enable() is active.

This pull request enables oproj_tensor_parallel_size support in the DSA attention backend. It introduces _forward_o_proj in AscendDSABackend to handle o_proj tensor parallel execution using static exchange buffers, all_to_all_single, and reduce_scatter collectives. It registers new column and row parallel operations (DSV4OProjColumnParallelOp and DSV4OProjRowParallelOp) for wo_a and wo_b projections when oproj_tp_enable() is active, and updates the profiling/warmup path to ensure consistent ACL graph capture. It also rewrites _forward_embed_tp to use graph-stable static buffers so HCCL collectives replay against the same device address recorded at capture, fixing the embedding TP path under ACL graph mode.

Yes, it enables tensor parallel support for the o_proj operator in the DSA backend (with the tensor_parallel_size == 1 constraint above), and makes the embedding TP path correct under ACL graph mode.

- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/attention/dsa_v1.py
+++ b/vllm_ascend/attention/dsa_v1.py
@@ -3,6 +3,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, ClassVar, TypeAlias
 
 import torch
+import torch.distributed as dist
 import torch.nn.functional as F
 import torch_npu
 import vllm.envs as envs_vllm
@@ -19,6 +20,7 @@ from vllm_ascend.attention.attention_mask import AttentionMaskBuilder
 from vllm_ascend.attention.attention_v1 import AscendAttentionState
 from vllm_ascend.attention.utils import AscendCommonAttentionMetadata, split_decodes_and_prefills
 from vllm_ascend.device.device_op import DeviceOperator
+from vllm_ascend.distributed.parallel_state import get_otp_group
 from vllm_ascend.ops.cv_linear import CVLinearWrapper
 from vllm_ascend.ops.linear import AscendUnquantizedLinearMethod
 from vllm_ascend.ops.rope_dsv4 import get_cos_and_sin_dsa
@@ -29,6 +31,7 @@ from vllm_ascend.utils import (
     get_ascend_device_type,
     npu_stream_switch,
     olora_tp_enable,
+    oproj_tp_enable,
 )
 from vllm_ascend.worker.npu_input_batch import NPUInputBatch
 
@@ -1550,6 +1553,10 @@ class AscendDSAImpl(DSAAttentionImpl):
         topk_indices_buffer.copy_(topk_indices_to_cache)
 
     def process_weights_after_loading(self, act_dtype: torch.dtype):
+        # Attention impls are not walked by vllm's process_weights_after_loading
+        # dispatcher (only LinearMethodBase subclasses are). OTP buffers are
+        # allocated lazily on the first _forward_o_proj call, which always runs
+        # before ACL graph capture (profiling run triggers it).
         pass
 
     # TODO: cast to bfloat16 to speed up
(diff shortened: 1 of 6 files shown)


jianzs:
To my knowledge, this feature cannot be enabled when both graph mode and recompute_scheduler are active simultaneously. When different data parallelism ranks replay distinct graphs, the OTP communication may freeze.

vllm_ascend/attention/dsa_v1.py
```diff
@@ -1563,6 +1575,81 @@ def rope_single(self, x, cos, sin, inverse=False):
...
+            recv = torch.empty_like(send)
+            dist.all_to_all_single(recv, send, group=oproj_group.device_group)
+            o_proj_input = recv.view(oproj_tp_size * exchange_num_tokens,
+                                     groups_per_oproj_rank, group_hidden_dim)
+            o_proj_input = torch_npu.npu_transpose_batchmatmul(
+                o_proj_input,
+                self.wo_a.weight,
+                bias=None,
+                scale=None,
+                perm_x1=(1, 0, 2),
```
momo609, line 1642: The permute parameter is hardcoded and may need adjustment if the dimensions change in the future.

vllm_ascend/attention/dsa_v1.py
```diff
@@ -1573,11 +1660,19 @@ def forward(  # type: ignore[override]
         output: torch.Tensor | None = None,
     ) -> torch.Tensor:
         assert output is not None, "Output tensor must be provided."
+        output_padded = output
         if attn_metadata is None:
-            return output.fill_(0)
+            # Profiling run.
+            if not oproj_tp_enable():
```
momo609: Why is it necessary to check that oproj_tp is not enabled before performing zero‑padding here?

vllm_ascend/ops/linear_op.py
```diff
@@ -220,6 +220,39 @@ def apply_impl(self, input_: torch.Tensor) -> torch.Tensor | tuple[torch.Tensor,
         return output, output_bias
 
 
+class DSV4OProjColumnParallelOp(CustomColumnParallelOp):
```
momo609, line 223: What is the reason for making DSV4OProjColumnParallelOp a separate class? Looking at the logic, there don't seem to be any special operations.

vllm_ascend/attention/dsa_v1.py
```diff
@@ -168,6 +171,15 @@ def pad_to_blocks(x: torch.Tensor, length_list: torch.Tensor, block_size: int =
     return out
 
 
+def _get_oproj_static_exchange_tokens(vllm_config: VllmConfig) -> int:
```
maoxx241: We don't actually need to do the padding inside this forward method. Instead, we can add a check where the ModelRunner sync_metadata_cross_dp. If oproj_tp is enabled, we can just handle the padding outside by default. This way, we avoid having to pad to the maximum capture size internally

vllm_ascend/attention/dsa_v1.py
xmpp777: This is the transposed arrangement of x1.

vllm_ascend/attention/dsa_v1.py
xmpp777: This is not the previous one; it is another branch.

vllm_ascend/ops/linear_op.py
xmpp777: Because the weights need to be divided according to the OTP group, not the TP division.

vllm_ascend/attention/dsa_v1.py
xmpp777: Adopted — thanks. We no longer pad to the max capture size inside the forward. Since allow_dp_padding is already forced in the model runner when oproj_tp is enabled, every OTP rank sees the same num_tokens per graph bucket, so we now communicate only num_tokens via a contiguous prefix slice of the static buffer (keeping graph-replay address stability while avoiding max-size traffic).

vllm_ascend/attention/dsa_v1.py
```diff
@@ -1563,6 +1575,81 @@ def rope_single(self, x, cos, sin, inverse=False):
...
+                self.wo_a.weight,
+                bias=None,
+                scale=None,
+                perm_x1=(1, 0, 2),
+                perm_x2=(0, 1, 2),
+                perm_y=(1, 0, 2),
+                batch_split_factor=1)
+            o_proj_input = o_proj_input.reshape(num_tokens, -1)
+            output[...] = self.wo_b(o_proj_input)
+        return output
```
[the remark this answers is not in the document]
xmpp777: This code only adds the OTP part, and the original process remains unchanged.

xmpp777:
> To my knowledge, this feature cannot be enabled when both graph mode and recompute_scheduler are active simultaneously. When different data parallelism ranks replay distinct graphs, the OTP communication may freeze.

1. This is the PR that precisely resolves the issue of freezing during graph replay.
2. With the current PR modifications included, verify that there is no freezing issue when it is enabled simultaneously.

linfeng-yuan:
1. Need to check the modifications around static buffer and HCCL collective calls caused by ACLGraph capturing; @yiz-liu 
2. Check implementation of DSV4OprojLinearOp
3. Split Embedding and add commit message.

xmpp777:
> 1. Need to check the modifications around static buffer and HCCL collective calls caused by ACLGraph capturing; @yiz-liu

The collectives use lazily-allocated static buffers (sized to the max capture capacity and addressed via contiguous prefix slices) so each all_to_all / all_gather / reduce_scatter hits the same device address across capture and replay, and allow_dp_padding keeps the collective shapes identical across all OTP ranks/buckets.

> 2. Check implementation of DSV4OprojLinearOp

The two classes exist only to rebind the weight-sharding group to OTP — comm_group determines which group shards the weight, and wo_a/wo_b must shard by the OTP group rather than the standard TP group (size 1 here). vLLM has no reference here: ParallelLinear hardcodes the standard TP group with no override, and vLLM-Ascend's only mechanism to change the sharding group is a custom op; OProjRowParallelOp can't be reused for wo_b because its self-contained all_to_all + reduce_scatter conflicts with DSV4's wo_a-in-the-middle layout.

> 3. Split Embedding and add commit message.

The description of modifying the embedding has been added.

PR #10327, opened 2026-06-11
[Refactor][Model] Move DeepSeek V4 cache hooks into model

QwertyJack:
Moves the Ascend-specific DeepSeek V4 cache overrides from the worker patch into `vllm_ascend.models.deepseek_v4`, where the DeepSeek V4 model implementation is maintained.

This removes `vllm_ascend/patch/worker/patch_deepseek_compressor.py` and its worker patch registration. `DeepseekV4Attention` now creates the Ascend SWA cache directly and passes it through `DSAModules`, so `AscendDeepseekSparseAttention` no longer depends on monkey-patching upstream DeepSeek V4 cache classes.

Co-author: Mengqing Cao <cmq0113@163.com>

No. This is an internal refactor of the DeepSeek V4 implementation path; the cache spec behavior is kept unchanged.

- `python -m py_compile vllm_ascend/models/deepseek_v4.py vllm_ascend/ops/dsa.py vllm_ascend/patch/worker/__init__.py`
- `ruff check vllm_ascend/models/deepseek_v4.py vllm_ascend/ops/dsa.py vllm_ascend/patch/worker/__init__.py`
- `git diff --check`
- Source-level check confirmed the removed worker patch is no longer referenced.

NPU serve validation was not run for this refactor.

- vLLM version: v0.21.0


vllm_ascend/models/deepseek_v4.py
```diff
@@ -82,11 +84,148 @@
...
+    ):
+        torch.nn.Module.__init__(self)
+        self.state_dim = state_dim
+        self.dtype = dtype
+        self.prefix = prefix
+        self.kv_cache = torch.tensor([])
+        compilation_config = get_current_vllm_config().compilation_config
+        if prefix in compilation_config.static_forward_context:
+            raise ValueError(f"Duplicate layer name: {prefix}")
+        compilation_config.static_forward_context[prefix] = self
```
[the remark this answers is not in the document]
QwertyJack: Fixed in the latest push. The exact suggested call included `block_size`, but the upstream `VllmCompressorStateCache` constructor signature is `(state_dim, dtype, compress_ratio, prefix)`. The updated code now calls `super().__init__(state_dim, dtype, compress_ratio, prefix)` and then preserves the Ascend-specific `compress_ratio` attribute and `block_size` override used by this cache path.

vllm_ascend/models/deepseek_v4.py
```diff
@@ -82,11 +84,132 @@
...
+        CompressorStateCache as VllmCompressorStateCache,
+    )
+    from vllm.model_executor.layers.deepseek_v4_attention import (  # type:ignore
+        DeepseekV4IndexerCache as VllmDeepseekV4IndexerCache,
+    )
 else:
-    from vllm.models.deepseek_v4.attention import DeepseekV4IndexerCache
-    from vllm.models.deepseek_v4.compressor import CompressorStateCache
+    from vllm.models.deepseek_v4.attention import DeepseekV4IndexerCache as VllmDeepseekV4IndexerCache
+    from vllm.models.deepseek_v4.compressor import CompressorStateCache as VllmCompressorStateCache
```
MengqingCao: no need to make version compatibility here.

vllm_ascend/models/deepseek_v4.py
```diff
@@ -82,11 +84,132 @@
...
+            dtype=self.dtype,
+            sliding_window=self.sliding_window,
+            alignment=None,
+            page_size_padded=page_size_padded,
+        )
+
+    def forward(self): ...
+
+    def get_attn_backend(self):
+        from vllm_ascend.attention.dsa_v1 import AscendDSABackend
```
MengqingCao: we could move this lazy import on the top of this script now, as it is not a patch anymore.

vllm_ascend/models/deepseek_v4.py
```diff
@@ -82,11 +84,132 @@
...
+
+        return AscendDSABackend
+
+
+CompressorStateCache = AscendCompressorStateCache
+DeepseekV4IndexerCache = AscendDeepseekV4IndexerCache
```
MengqingCao: plz remove this patch.

vllm_ascend/ops/dsa.py
```diff
@@ -112,15 +112,8 @@ def __init__(
...
-        k_dtype = torch.float8_e4m3fn if ascend_device_type == AscendDeviceType.A5 else torch.bfloat16
-        self.swa_cache_layer = DeepseekV4SWACache(
-            head_dim=self.head_dim,
-            window_size=self.window_size,
-            dtype=k_dtype,
-            prefix=f"{prefix}.swa_cache",
-            cache_config=cache_config,
-        )
+        self.swa_cache_layer = dsa_modules.swa_cache_layer
+        assert self.swa_cache_layer is not None
```
MengqingCao: plz add message for the assertion.

vllm_ascend/models/deepseek_v4.py
QwertyJack: Done. Removed the vLLM 0.21 compatibility branch and now import the DeepSeek V4 cache base classes directly from `vllm.models.deepseek_v4` for main.

vllm_ascend/models/deepseek_v4.py
QwertyJack: Done. Moved `AscendDSABackend` to the top-level imports in `deepseek_v4.py`.

vllm_ascend/models/deepseek_v4.py
QwertyJack: Done. Removed the alias-based patch lines and changed the call sites to instantiate `AscendCompressorStateCache` and `AscendDeepseekV4IndexerCache` explicitly.

vllm_ascend/ops/dsa.py
QwertyJack: Done. Added an assertion message for the required DeepSeek V4 SWA cache layer.

vllm_ascend/models/deepseek_v4.py
QwertyJack: Restored this compatibility branch after re-checking the repo support matrix. `main` still documents CI commitment for both vLLM main and the vLLM v0.21.0 tag (`README.md` and `.github/vllm-release-tag.commit`), and the v0.21.0 path still uses `vllm.model_executor.layers.deepseek_*`. The alias-style patch has still been removed; only the import compatibility remains.

vllm_ascend/models/deepseek_v4.py
```diff
@@ -82,11 +85,122 @@
 )
 
 if vllm_version_is("0.21.0"):
-    from vllm.model_executor.layers.deepseek_compressor import CompressorStateCache  # type:ignore
-    from vllm.model_executor.layers.deepseek_v4_attention import DeepseekV4IndexerCache  # type:ignore
+    from vllm.model_executor.layers.deepseek_compressor import (  # type: ignore[no-redef]
```
MengqingCao: nit: no need to rename `CompressorStateCache` to `VllmCompressorStateCache`

QwertyJack: Done. Removed the `Vllm*` base-class aliases and now inherit from the imported `CompressorStateCache` / `DeepseekV4IndexerCache` names directly. The version compatibility branch is still kept because the repo currently supports both vLLM main and the v0.21.0 tag; the extra `type: ignore[import-not-found,no-redef]` only covers the static-checker view of the mutually exclusive import paths.

QwertyJack: fixed

weijinqian0 (approved):

PR #10342, opened 2026-06-11
[Feature] Mooncake connectot support dsv4

liziyu179:

This PR adds Mooncake connector support for DeepSeek V4 / hybrid KV cache scenarios.

DeepSeek V4 introduces mixed KV cache layouts, including attention KV, compressed KV, sliding-window KV, and Mamba/state cache groups. The previous Mooncake transfer logic assumed that block ids could be translated with a simple `block_id * block_len` offset and that prompt blocks were aligned in the same way across all KV groups. That assumption is not always valid for DSV4/hybrid cache, and can lead to transferring extra blocks, using incorrect byte offsets, or computing one extra state token on the P side.

Main changes:

- Add per-block byte stride metadata to Mooncake handshake metadata.
  - `MooncakeAgentMetadata` now carries `block_strides`.
  - Sender/receiver use `stride(0) * element_size` for block offsets instead of assuming `block_len`.
  - Transfer grouping now checks byte contiguity with both block id continuity and actual memory stride.

- Support compressed and hybrid KV transfer block calculation.
  - Add per-group transfer metadata for `tokens_per_block`, sliding-window size, and state-group detection.
  - Use `compress_ratio` when calculating how many prompt blocks should be transferred.
  - Keep Mamba/state groups intact, while trimming attention-like groups to the blocks that actually contain transferable prompt KV.
  - Handle sliding-window groups by keeping only the active window blocks and dropping placeholder zero blocks where needed.

- Fix P/D-side token alignment for hybrid attention + Mamba prefill.
  - For remote prefill, the D side only counts `N - 1` prompt tokens when state cache requires the decoder to recompute the last token.
  - For remote decode on the P side, truncate the last prompt token once and mark the request with `_p_side_truncated` to avoid repeated truncation after preemption/reschedule.
  - Move this logic out of `RecomputeScheduler` and into the Mooncake connector scheduler where the KV-transfer semantics are known.

- Improve failure handling for multi-group block ids.
  - Failed receive requests now mark invalid block ids from the first block-id group instead of treating the full multi-group `BlockIds` tuple as a flat list.

- Update Mooncake worker metadata registration for hybrid KV cache.
  - Use total hidden layer count from model config.
  - Register shared hybrid KV tensor buffers when hybrid cache is enabled.

Yes. Mooncake connector can now correctly support DeepSeek V4 / hybrid KV cache disaggregated prefill scenarios.

There is no new public API or configuration option. Existing Mooncake metadata is extended internally with block stride information, so both sides of the connector should run with this updated code.

Added and updated unit tests for Mooncake connector and layerwise connector:

- `tests/ut/kv_offload/test_mooncake_connector.py`
  - compressed KV prefix-cache offset calculation
  - block-stride-based transfer offsets
  - Mamba transfer metadata offsets with block stride
  - byte-contiguous grouping and split behavior
  - attention MTP block trimming
  - state-group block preservation
  - compressed prompt length block calculation
  - sliding-window block clipping and zero-block filtering
  - remote metadata propagation of `block_strides`

- `tests/ut/kv_offload/test_mooncake_layerwise_connector.py`
  - hybrid prefill token count excludes the last token
  - P-side hybrid prefill request truncation
  - remote block trimming when the last token occupies a new block

Suggested commands:

```bash
pytest -sv tests/ut/kv_offload/test_mooncake_connector.py
pytest -sv tests/ut/kv_offload/test_mooncake_layerwise_connector.py


- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
@@ -2003,6 +2003,8 @@ class MooncakeConnectorWorker:
         # [layer_idx][cache_idx] -> stride(0) * element_size.
         self.block_stride_per_addr: list[list[int]] = [[] for _ in range(metadata_layers)]
 
+        # TODO: For DSV4 use_compress, metadata/transfer can be optimized by
+        # aggregating layer views that share the same raw KVCacheTensor.
         for layer_name, kv_cache_tuple in kv_caches.items():
             layer_idx = layer_name_to_idx[layer_name]
             for single_kv_cache in self._as_kv_cache_tuple(kv_cache_tuple):
--- a/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
@@ -449,6 +449,7 @@ class KVCacheRecvingThread(threading.Thread):
             self.group_compress_ratios[group_id] = compress_ratio
         self.remote_te_port: dict[str, dict[int, int]] = SizedDict()
         self.remote_block_size_scale: dict[str, dict[int, list[list[int]]]] = SizedDict()
+        self.remote_block_stride_per_addr: dict[str, dict[int, list[list[int]]]] = SizedDict()
         self.remote_kv_group2layeridx: dict[str, dict[int, dict[int, tuple[dict[str, Any], list[int]]]]] = SizedDict()
 
         self.request_queue: queue.Queue[Any] = queue.Queue()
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by LCAIZJ** [comment]
Could you describe the main logic behind the changes?

**Comment by liziyu179** [comment]
> Could you describe the main logic behind the changes?

ok, we have updated the PR description.

**Comment by nwpu-zxr** [comment]
I approve this PR.

**Review by wangxiaoteng888 — APPROVED** [review]

**Review by nwpu-zxr — APPROVED** [review]

**Review by zzzzwwjj — APPROVED** [review]

File: vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py [hunk 1]
```diff
@@ -745,10 +772,19 @@ def pp_layer_indices(layer_indices: list[int], prefill_pp_rank: int) -> list[int
...
+                        grouped_remote_block_ids,
+                        grouped_local_block_ids,
+                        src_block_stride=remote_block_stride,
+                        dst_block_stride=block_stride,
+                        block_len=inner_block_len,
+                    )
+                    for remote_block_id, local_block_id in zip(transfer_remote_block_ids, transfer_local_block_ids):
+                        src = src_layer_base_addr + local_block_id[0] * block_stride + inner_offset * inner_block_len
+                        dst = dst_layer_base_addr + remote_block_id[0] * remote_block_stride
                         length = inner_block_len * len(local_block_id)
```
nwpu-zxr, line 788: Using `block_stride` to send the blocks can simplify the transfer and make it easier to merge transfer tasks and get full use of bandwidth.

PR #10349, opened 2026-06-11
[Feature][DSv4] Support compressor block size [32,64,128] to improve APC

Liuchenbing-2026:
Makes the DeepSeek V4 compressor KV-cache block_size configurable (32 / 64 /
128) instead of hard-pinned to 128, via a lookup table that scales the MLA /
SWA / C4-state / C128-state block sizes and page-size padding together. A
smaller block size shrinks the alignment requirement of the compressed KV
cache, so more prompt prefixes can hit the prefix cache (better APC hit rate).

Block size defaults to 32 for deepseek_v4; pass --block-size 128 to keep the
legacy layout (bit-for-bit identical to the previous behavior).

---
Co-authored with @wangzhao-11a and @MengqingCao.

### DIFF
--- a/vllm_ascend/attention/dsa_v1.py
+++ b/vllm_ascend/attention/dsa_v1.py
@@ -207,7 +207,7 @@ class AscendDSABackend(AttentionBackend):
 
     @staticmethod
     def get_supported_kernel_block_sizes() -> list[int]:
-        return [8, 32, 128]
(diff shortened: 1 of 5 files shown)


MengqingCao:
closing as done by https://github.com/vllm-project/vllm-ascend/pull/10354

PR #10350, opened 2026-06-11
[Feature][DSv4] Support partial compressed prefix cache hits

Liuchenbing-2026:
Adds partial compressed prefix-cache hits for DeepSeek V4. When a prompt is
shorter than one compressed cache block and gets zero full-block hits, the
first block's prefix boundaries are registered in a private partial table; a
later request then reuses the longest matching prefix by copying the source KV
into a private block (in-cache device-to-device copy) before the forward pass.
Boundary registration is latched once per request to avoid per-decode-step
re-hashing. Builds on the prefix-cache core block-reuse patch and aligns hits
to the scheduler block size.

No.

---
Co-authored with @wangzhao-11a.


Liuchenbing-2026:

PR #10353, opened 2026-06-11
[Feature][DSv4] Backport vLLM #43447 prefix-cache (block-reuse + sliding-window retention) for vLLM 0.21.0

Liuchenbing-2026:
Backports vLLM PR #43447 (DeepSeek-V4 selective prefix-cache retention for sliding-window KV cache) to vLLM versions that lack it (e.g. v0.21.0), via capability/source guarded monkey-patches. It carries two parts of #43447:

- **block-reuse**: free-list front-insertion (`prepend_n` / `free_blocks(prepend=)`), SWA/EAGLE cache-block mask, and the `remove_skipped_blocks` cached/uncached split.
- **selective prefix-cache retention** for sliding-window KV cache: retention interval + reachable block mask, controlled by the `VLLM_ASCEND_PREFIX_CACHE_RETENTION_INTERVAL` env.

Each patch is guarded by a capability/source check, so it is a no-op and is automatically skipped once the underlying vLLM already provides the behavior.

## Relationship to upstream PRs
- Backports [Prefix Caching] DeepSeekv4 - Support selective prefix-cache retention for sliding-window KV cache vllm-project/vllm#43447 in full (both block-reuse and retention).
- Orthogonal to the official compressed-prefix lookup fix already on main: this PR leaves the lookup / read path untouched; retention operates on the write path (`__init__` + `cache_blocks`), sharing only the read-only alignment.

Adds an opt-in env `VLLM_ASCEND_PREFIX_CACHE_RETENTION_INTERVAL` to control sliding-window retention (unset = legacy dense behavior, bit-for-bit identical to before). All patches are guarded and skipped when the underlying vLLM already provides them.

## How was this patch tested?
- `lint-and-select-tests` passed (ruff / format / codespell / mypy).
- New unit tests `tests/ut/core/test_prefix_cache_retention.py` assert the retention reachable-block masks (EAGLE and interval cases) and run on `cpu-0card`.
- `run-selected-tests` on `cpu-0card` / `a2-1card`.

Co-authored with @wangzhao-11a and @HF-001.

- vLLM version: v0.21.0

### DIFF
--- a/.github/workflows/scripts/test_config.yaml
+++ b/.github/workflows/scripts/test_config.yaml
@@ -372,6 +372,7 @@
   tests:
     - tests/ut/patch
     - tests/ut/test_compressed_prefix_cache.py
+    - tests/ut/core/test_prefix_cache_retention.py
     - tests/e2e/pull_request/one_card/test_qwen3_0_6b.py
     - tests/e2e/pull_request/one_card/test_qwen3_5_0_8b.py
 
--- a/tests/ut/core/test_prefix_cache_retention.py
+++ b/tests/ut/core/test_prefix_cache_retention.py
@@ -1,0 +1,304 @@
+#
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+# This file is a part of the vllm-ascend project.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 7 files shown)


realliujiaxu (approved):

realliujiaxu:
please update **How was this patch tested?** and add UT

PR #10354, opened 2026-06-11
[Feature][DSv4] Support compressor block size [32,64,128] to improve automatic prefix cache hit rate

Liuchenbing-2026:
Makes the DeepSeek V4 compressor KV-cache block_size configurable (32 / 64 / 128) instead of hard-pinned to 128, via a lookup table that scales the MLA / SWA / C4-state / C128-state block sizes and page-size padding together. A smaller block size shrinks the alignment requirement of the compressed KV cache, so more prompt prefixes can hit the prefix cache (better automatic prefix cache hit rate).

For DeepSeek V4, `--block-size` now accepts 32 / 64 / 128.

- **Unset**: vLLM's `CacheConfig.DEFAULT_BLOCK_SIZE` is 16, which is not in the allowed set; `refresh_block_size` resets it to **32** with a warning log. This means **the effective default for DeepSeek V4 has changed from 128 (pre-PR, hard override) to 32 (this PR)**. `--block-size 32` is also the recommended setting for prefix-cache workloads (best automatic prefix cache hit rate; see benchmarks below).
- `--block-size 32 / 64 / 128`: respected as-is.
- Any other explicit value: falls back to **32** with a warning log.

If you want to preserve the previous behavior (block_size = 128), pass `--block-size 128` explicitly.

> ⚠️ **PD-disaggregated deployments**: The Prefiller (P) and Decoder (D) instances **must use the same `--block-size`**. Mismatched block sizes produce incompatible KV-cache page layouts and the D side will fail to consume the KV cache transferred from P. When upgrading to this PR in a PD setup, either set `--block-size 128` on both sides (preserve old behavior) or set `--block-size 32` on both sides (recommended for prefix-cache workloads) — do not leave one side unset while the other is pinned.

Validated on DeepSeek-V4 (w8a8 + MTP) with vLLM 0.21.0, prefix caching enabled, tensor-parallel 8, `num_speculative_tokens=1`, `--max-model-len 150000`, `--max-num-batched-tokens 8192`, `--max-num-seqs 32`. Service `/health` returned 200.

**Block-size comparison** — 32k input / 1k output, concurrency 8, repeat_rate 0.9, prefix-cache enabled:

| block-size | HBM hit rate | Total Token Throughput | Avg TTFT | Avg TPOT |
|---:|---:|---:|---:|---:|
| **32** | **87.49%** | **8906 tok/s** | **1471 ms** | **27.9 ms** |
| 64 | 74.99% | 7922 tok/s | 2661 ms | 30.6 ms |
| 128 | 45.31% | 6612 tok/s | 5430 ms | 34.3 ms |

Smaller block size gives finer cache-key granularity and higher automatic prefix cache hit rate. Ordering: hit rate & throughput **32 > 64 > 128**; TTFT/TPOT **32 < 64 < 128**. `--block-size 32` is recommended for prefix-cache workloads.

**Latest PR HEAD validation** (block-size 32):

| Scenario | input/output | concurrency | HBM hit rate | Total Token Throughput | Avg TTFT | Avg TPOT |
|---|---|---:|---:|---:|---:|---:|
| 8k/1k | 8192/1024 | 32 | 48.80% | 5996 tok/s | 3297 ms | 44.3 ms |
| 32k/1k | 32768/1024 | 8 | 87.49% | 8711 tok/s | 1832 ms | 28.3 ms |

- All requests succeeded (128/128 and 32/32); all prefix-cache hits were local HBM (no external hits).
- CI: `lint-and-select-tests` and `a3-4card` DSv4 e2e passed.

---
Co-authored with @MengqingCao.

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/models/deepseek_v4.py
+++ b/vllm_ascend/models/deepseek_v4.py
@@ -102,6 +102,14 @@ def _get_ascend_dsa_backend():
     return AscendDSABackend
 
 
+def _dsv4_block_sizes():
+    # Lazy import to avoid the circular import chain (layer -> dsa_v1 ->
+    # attention_v1 -> device_op) hit during vLLM subprocess model inspection.
+    from vllm_ascend.models.layer.attention.layer import DSV4_BLOCK_SIZES
+
+    return DSV4_BLOCK_SIZES
+
+
 class AscendCompressorStateCache(CompressorStateCache):
     def __init__(
         self,
@@ -116,10 +124,8 @@ class AscendCompressorStateCache(CompressorStateCache):
         self.block_size = block_size
 
     def get_kv_cache_spec(self, vllm_config: VllmConfig) -> KVCacheSpec:
-        if get_ascend_device_type() in {AscendDeviceType.A5}:
-            page_size_padded = 16896 if self.state_dim == 2 * 256 and self.compress_ratio == 4 else 81920
-        else:
-            page_size_padded = 16640 if self.state_dim == 2 * 256 and self.compress_ratio == 4 else 131072
+        pads = _dsv4_block_sizes()[vllm_config.cache_config.block_size][1]
+        page_size_padded = pads[0] if self.state_dim == 2 * 256 and self.compress_ratio == 4 else pads[1]
 
         return SlidingWindowMLASpec(
             block_size=self.block_size,
@@ -156,7 +162,7 @@ class AscendDeepseekV4IndexerCache(DeepseekV4IndexerCache):
         from vllm.v1.kv_cache_interface import MLAAttentionSpec
 
         return MLAAttentionSpec(
-            block_size=128,
+            block_size=_dsv4_block_sizes()[vllm_config.cache_config.block_size][0][0],
             num_kv_heads=1,
             head_size=self.head_dim,
             dtype=self.dtype,
@@ -185,7 +191,7 @@ class AscendDeepseekV4SWACache(VllmDeepseekV4SWACache):
         super().__init__(head_dim, window_size, torch.uint8, prefix, cache_config)
         self.dtype = dtype
 
-        self.block_size = 128
+        self.block_size = _dsv4_block_sizes()[cache_config.block_size][0][1]
 
     def get_kv_cache_spec(self, vllm_config: VllmConfig) -> KVCacheSpec:
         if get_ascend_device_type() in {AscendDeviceType.A5}:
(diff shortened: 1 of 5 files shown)


vllm_ascend/models/deepseek_v4.py
```diff
@@ -524,15 +525,17 @@ def __init__(
...
         elif compress_ratio == 128:
             self.state_cache = CompressorStateCache(
                 state_dim=2 * self.head_dim,  # kv_state + score_state
                 dtype=state_dtype,
                 compress_ratio=compress_ratio,
                 prefix=f"{prefix}.state_cache",
-                block_size=16 if get_ascend_device_type() in {AscendDeviceType.A5} else 32,
+                block_size=16
+                if get_ascend_device_type() in {AscendDeviceType.A5}
+                else DSV4_BLOCK_SIZES[cache_config.block_size][0][3],  # type: ignore[union-attr]
```
MengqingCao:
```suggestion
                block_size=DSV4_BLOCK_SIZES[cache_config.block_size][0][3],  # type: ignore[union-attr]
```

realliujiaxu:
suggest PR title：[Feature][DSv4] Support compressor block size [32,64,128] to improve APC hit rate

MengqingCao (changes requested):
there is a bug that the block size won't set correctly.

MengqingCao (approved):
I've checked that the block size won't be changed until the model is initialized, that is, the block size we used in the `KVCacheSpec` and `SingleTypeKVCacheManager` is correct, thus LGTM now.

vllm_ascend/utils.py
```diff
@@ -1261,8 +1264,27 @@ def refresh_block_size(vllm_config):
         return
 
     if model_config.hf_config.model_type == "deepseek_v4":
-        # TODO(qcs): generalize the block_size
-        cache_config.block_size = 128
+        # NOTE(cmq): use _ENIGINE_CORE_BLOCK_SIZE as the cache_config.block_size
+        # is updated in the core process, but it will be updated to the minimized
+        # block_size among all the kvcache groups, which will leading to a wrong
+        # block_size in DeepSeek V4 scenario.
+        global _ENIGINE_CORE_BLOCK_SIZE
```
MengqingCao: btw, this global block size is not needed now cause the `block_size` won't be refreshed before the model is initialized. Let's remove this.

nayihz:
Even if the input length is less than 4096, it still cannot hit the cache, right?

PR #9310, opened 2026-05-19
[Performance] Reuse prebuilt chunk host metadata for Ascend chunk ops and earse synconize for qwen3.5 model

shaopeng-666:
### What this PR does / why we need it?
1.Erease the synconize before chunk o and chunk fwd h op.
2.decrease the time between 2 decode steps
3.refactor the gdn attention backend（delete the gdn attn builder patch）.
### Does this PR introduce _any_ user-facing change?
No
### How was this patch tested?
1.performence:
1)chunk o and chunk fwd h:
2)decode time between 2 decode steps decrease from 6ms to 3ms
3)3.5K/1.5K@TPOT50ms on Qwen3.5-397B-A13B model on A3:
330 performence: 2400 tokens/s @TPOT43ms
main code with this pr:2700 tokens/s @TPOT32ms
2.Accuracy:
Test GPQA on Qwen3.6-35B-A3B:
0 day accuracy:83.23.
main code with this pr:83.33

- vLLM version: v0.21.0
- vLLM main: https://github.com/vllm-project/vllm/commit/967c5c3bc38891f4465d3f4e99917ed837bb3833

PR #10359, opened 2026-06-11
[Performance] Optimize chunk_gated_delta_relu_fwd operator

socrahow:
In chunk_gated_delta_relu_fwd, vllm-ascend use triton and ascendc operators. triton need tentor, ascendc need list. in every gdn layer do tolist(). This PR moves the tolist to the front of GDN, so that it only needs to be calculated once per round.
No
It test in Atlas 800T A2, use qwen3.5-0.8b model
- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/patch/worker/patch_gdn_attn.py
+++ b/vllm_ascend/patch/worker/patch_gdn_attn.py
@@ -16,6 +16,7 @@
 from dataclasses import dataclass
 
 import torch
+from typing import List, Optional
 import vllm.v1.attention.backends.gdn_attn as gdn_attn
 from vllm.v1.attention.backends.utils import NULL_BLOCK_ID
 
@@ -77,6 +78,17 @@ class GDNDecodeFallbackMeta:
 @dataclass
 class GDNSpecDecodeFallbackMeta:
     spec_causal_conv1d: GDNSpecCausalConv1dHostMetadata
+
+
+@dataclass
+class GDNChunkedPrefillReluListMeta:
+    chunk_list: _GDNChunkedReluListMeta
+
+
+@dataclass
+class _GDNChunkedReluListMeta:
+    non_spec_query_start_loc_list: Optional[List[int]] = None,
+    chunk_indices_list: Optional[List[int]] = None,
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


vllm_ascend/patch/worker/patch_gdn_attn.py
```diff
@@ -706,6 +718,19 @@ def _build_non_spec_chunked_prefill_meta(
     return _build_chunked_prefill_metadata(builder, tensors, slot=slot)
 
 
+def _build_chunked_list_meta(
+    builder,
+    cu_seqlens,
+    chunk_indices,
+) -> _GDNChunkedReluListMeta:
+    cu_seqlens_list: Optional[List[int]] = cu_seqlens.to(torch.int64).tolist()
+    chunk_indices_list: Optional[List[int]] = chunk_indices.flatten().tolist()
```
maoxx241, line 727: Can we avoid deriving this host list from `chunk_indices` here? In the NPU chunk-metadata path, `chunk_indices` can be a device tensor, so `chunk_indices.flatten().tolist()` still creates a device-to-host synchronization. Since `cu_seqlens` / `non_spec_query_start_loc_cpu` is already CPU-side, this list can be built directly from CPU chunk counts, matching the direction in upstream vLLM's GDN metadata builder and the approach in #9310.

maoxx241 (commented):
Thanks for the optimization. I agree with the goal of avoiding repeated per-layer `tolist()` conversions before the Ascend chunk ops, but I don't think this shape is the right one to merge as-is.

The current patch still derives the host-side chunk list from `chunk_indices_chunk64` via `chunk_indices.flatten().tolist()`. In the NPU chunk-metadata path that tensor can be on device, so this moves the synchronization from every GDN layer to metadata build time, but it does not remove the NPU-to-CPU synchronization.

Upstream vLLM `main` has already moved this area in a cleaner direction: `GDNAttentionMetadataBuilder` builds chunk metadata from the CPU-side `prefill_query_start_loc_cpu`, then copies the tensor metadata to device, explicitly to avoid the `.tolist()` sync in `prepare_chunk_indices`. Current upstream main also precomputes `prefill_query_start_loc`, `prefill_state_indices`, and `prefill_has_initial_state` for the mixed decode+prefill path.

For vLLM Ascend, I think we should follow that pattern, or align this with #9310: build the AscendC host list/tuple directly from `non_spec_query_start_loc_cpu` / CPU chunk counts, and pass that into the custom ops, instead of converting a device `chunk_indices` tensor back to a Python list.

vllm_ascend/patch/worker/patch_gdn_attn.py
socrahow: This PR [#9310](https://github.com/vllm-project/vllm-ascend/pull/9310) incorporated my optimizations and has already been merged into the release. Should I close my pr ?

Issue #10376, opened 2026-06-12
[Bug]: v0.20.2rc1 run minimax-m2.7 meet an error when using streaming tool parser


frankie-ys:
### Your current environment

v0.20.2rc1 + 910B

### 🐛 Describe the bug

the error message is:
(APIServer pid=58284) INFO:     172.30.8.42:41346 - "GET /metrics HTTP/1.1" 200 OK
(APIServer pid=58284) ERROR 06-09 14:11:06 [serving.py:1272] Error in chat completion stream generator.
(APIServer pid=58284) ERROR 06-09 14:11:06 [serving.py:1272] Traceback (most recent call last):
(APIServer pid=58284) ERROR 06-09 14:11:06 [serving.py:1272]   File "/vllm-workspace/vllm/vllm/entrypoints/openai/chat_completion/serving.py", line 1005, in chat_completion_stream_generator
(APIServer pid=58284) ERROR 06-09 14:11:06 [serving.py:1272]     delta_message = parser.parse_delta(
(APIServer pid=58284) ERROR 06-09 14:11:06 [serving.py:1272]                     ^^^^^^^^^^^^^^^^^^^
(APIServer pid=58284) ERROR 06-09 14:11:06 [serving.py:1272]   File "/vllm-workspace/vllm/vllm/parser/abstract_parser.py", line 631, in parse_delta
(APIServer pid=58284) ERROR 06-09 14:11:06 [serving.py:1272]     delta_message = self.extract_tool_calls_streaming(
(APIServer pid=58284) ERROR 06-09 14:11:06 [serving.py:1272]                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(APIServer pid=58284) ERROR 06-09 14:11:06 [serving.py:1272]   File "/vllm-workspace/vllm/vllm/parser/abstract_parser.py", line 551, in extract_tool_calls_streaming
(APIServer pid=58284) ERROR 06-09 14:11:06 [serving.py:1272]     return self._tool_parser.extract_tool_calls_streaming(
(APIServer pid=58284) ERROR 06-09 14:11:06 [serving.py:1272]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #10377, opened 2026-06-12
[BugFix] fix the streaming tool parser's DeltaMessage miss tool_calls list in v0.20.2rc1

frankie-ys:
fix issue #10376 
add tool_calls list in DeltaMessage.

- vLLM version: v0.22.1


QwertyJack:
Thanks for fixing #10376. I think the current patch fixes the Pydantic validation error, but it does so by adding `tool_calls=[]` to content-only streaming deltas. That creates a separate OpenAI-compatibility problem.

`tool_calls=[]` should not be emitted when there is no actual tool call. vLLM issue vllm-project/vllm#44104 is exactly about empty `tool_calls` arrays in normal assistant messages / streaming text deltas causing SDK clients to enter the tool-call path. The upstream fix in vllm-project/vllm#44105 is to omit empty `tool_calls` from the serialized payload.

The #10376 crash looks narrower: this branch can pass `tool_calls=None` into `DeltaMessage`, whose `tool_calls` field is a list:

```python
if delta_tool_call or content_before:
    return DeltaMessage(
        content=content_before,
        tool_calls=[delta_tool_call] if delta_tool_call else None,
    )
```

A safer fix would be to split the tool-call and content-only cases, and leave the existing content-only returns without an explicit `tool_calls` argument:

```python
if delta_tool_call:
    return DeltaMessage(
        content=content_before,
        tool_calls=[delta_tool_call],
    )

if content_before:
    return DeltaMessage(content=content_before)
```

Then keep these as-is:

```python
return DeltaMessage(content=delta_text) if delta_text else None
return DeltaMessage(content="")
```

That avoids the validation error while preserving the desired wire contract: no `tool_calls` field when no tool call exists, and a non-empty `tool_calls` list only when there is an actual tool-call delta.

Could you also add a small regression test for this parser path? The important assertions are:

1. the `content_before`-only path no longer raises a `DeltaMessage` validation error;
2. serialized streaming text deltas without an actual tool call do not contain `tool_calls: []`;
3. non-empty tool-call deltas still preserve `tool_calls`.

QwertyJack:
Thanks for your PR! It's no longer needed as vLLM has rewritten the minimax parser.

PR #10380, opened 2026-06-12
[BugFix] Fix garbled output for Qwen3.6-27B with dflash/suffix num_speculative_tokens > 7

nevin001:
Fix garbled output for Qwen3.6-27B with dflash/suffix when num_speculative_tokens > 7

This PR fixes garbled output for Qwen3.6-27B when using dflash/suffix speculative decoding with num_speculative_tokens > 7. The root cause is that the UB (Unified Buffer) memory allocation in the recurrent gated delta rule tiling logic did not correctly account for cases where the actual speculative token count exceeds certain internal buffer limits (8), leading to memory corruption during attention computation. This fix ensures correct buffer sizing and output correctness for all valid num_speculative_tokens values (1–15).

No. This is a bugfix with no API or interface changes. Users can now correctly set num_speculative_tokens to values greater than 7 without encountering garbled output.

Tested manually on Ascend 910B with Qwen3.6-27B using dflash/suffix speculative decoding:

- Run inference with num_speculative_tokens=8 (previously garbled) → output is now correct.
- Run inference with num_speculative_tokens=7 (previously working) → output unchanged, no regression.

- vLLM version: v0.21.0

### DIFF
--- a/csrc/attention/recurrent_gated_delta_rule/op_host/recurrent_gated_delta_rule_tiling.cpp
+++ b/csrc/attention/recurrent_gated_delta_rule/op_host/recurrent_gated_delta_rule_tiling.cpp
@@ -49,7 +49,7 @@ const size_t DIM_1 = 1;
 const size_t DIM_2 = 2;
 const size_t DIM_3 = 3;
 
-const size_t MAX_MTP = 8;
+const size_t MAX_MTP = 16;
 
 template <typename T1, typename T2>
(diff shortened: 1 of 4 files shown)


leixinW:
您好，我们在qwen3.5-27B的dflash模型上也遇到了相同的问题，num_speculative_tokens设置超过7就会乱码，请问这个issue预计什么时间合入解决

PR #10384, opened 2026-06-12
[BugFix][v0.20.2rc] Preserve GLM streaming tool call name

QwertyJack:
## Summary
- backport #10361 to releases/v0.20.2rc
- preserve GLM streaming tool-call id/type/function.name when appending remaining arguments
- keep the parser-produced delta unchanged when remaining_call is empty, so zero-arg tool calls keep function.name and arguments

## Trigger condition
This is triggered when GLM streaming tool calls run with the thinking parser disabled. The GLM47 parser can emit the correct tool-call name and empty arguments, but the vllm-ascend streaming monkeypatch may rebuild the final delta via _create_remaining_args_delta. When remaining_call is empty, the old code can replace the parser delta with a new tool call whose function.name is missing/empty.

## Validation
- ruff format --check vllm_ascend/patch/platform/patch_glm_tool_call_streaming.py tests/ut/patch/platform/test_patch_glm_tool_call_streaming.py tests/ut/patch/platform/test_patch_glm47_tool_call_parser.py
- ruff check vllm_ascend/patch/platform/patch_glm_tool_call_streaming.py tests/ut/patch/platform/test_patch_glm_tool_call_streaming.py tests/ut/patch/platform/test_patch_glm47_tool_call_parser.py
- git diff origin/releases/v0.20.2rc --check
- python -m py_compile vllm_ascend/patch/platform/patch_glm_tool_call_streaming.py tests/ut/patch/platform/test_patch_glm_tool_call_streaming.py tests/ut/patch/platform/test_patch_glm47_tool_call_parser.py
- local GLM47 inline zero-arg parser/helper smoke: PASS
- patch applyability checked on a clean origin/releases/v0.20.2rc worktree

Focused pytest was attempted but could not run locally because this container's /vllm-workspace/vllm does not match the release CI vLLM baseline and fails during tests/ut/conftest.py import before collecting these tests.

### DIFF
--- a/tests/ut/patch/platform/test_patch_glm47_tool_call_parser.py
+++ b/tests/ut/patch/platform/test_patch_glm47_tool_call_parser.py
@@ -4,6 +4,7 @@ import json
 from unittest.mock import MagicMock
 
 from vllm.entrypoints.openai.chat_completion.protocol import ChatCompletionRequest
+from vllm.entrypoints.openai.chat_completion.serving import OpenAIServingChat
 from vllm.parser.abstract_parser import _WrappedParser
 from vllm.reasoning.deepseek_v3_reasoning_parser import (
     DeepSeekV3ReasoningWithThinkingParser,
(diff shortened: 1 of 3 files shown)


QwertyJack:
fixed on main

PR #10393, opened 2026-06-12
[Feature][KV Pool] Introduce AscendStoreCoordinator for hybrid cache-hit and mask coordination

Pz1116:
## Reference Design

This PR ports the external-store cache-hit coordination model from vLLM MooncakeStore HybridKVCacheCoordinator and MooncakeStoreCoordinator into AscendStore KV Pool. The implementation keeps AscendStore key and address generation, while using coordinator-style convergence and per-group reachable masks for hybrid cache groups.
Ref:https://github.com/vllm-project/vllm/pull/44774

## Main Changes

- Add AscendStoreCoordinator to mirror MooncakeStore external-cache hit coordination over AscendStore keys.
- Wire the coordinator into KVPoolWorker and ChunkedTokenDatabase so lookup, save, and load share the same per-group mask semantics.
- Preserve DeepSeek V4 compressed cache-family granularity: c4 and c128 keys are generated over raw-token spans, while transfer addresses remain in cache-domain blocks.
- Carry num_prompt_tokens through request metadata so retention-aware reachable masks can match the scheduler/cache-manager behavior.
- Keep the path disabled for non-hybrid or single-group KV Pool users, preserving the legacy behavior.
- Add unit coverage for coordinator hit convergence, compressed effective granularity, and transfer mask application.

## Expected Behavior

- Hybrid AscendStore lookup returns the common external prefix only when required KV cache groups satisfy the coordinator policy.
- Sliding-window, Mamba align, and compressed DSV4 groups use manager-compatible reachable masks instead of storing or loading every physical chunk.
- DSV4 c128 participates at the 16K effective boundary when block size is 128, rather than as 128-token chunks.
- Store skips chunks that cannot serve a future hit, reducing external-store footprint for long-context hybrid models.
- Load fetches only chunks required by the per-group mask and does not regress single-group KV Pool behavior.

## Validation

vLLM version: v0.20.1
vLLM main: vllm-project/vllm@7021be66e

- vLLM version: v0.23.0


vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/coordinator.py
```diff
@@ -0,0 +1,368 @@
...
+        )
+
+        self.kv_cache_groups = kv_cache_groups
+        self.hash_block_size = hash_block_size
+        self.lcm_block_size = scheduler_block_size
+        self.use_eagle = use_eagle
+        self.retention_interval = retention_interval
+        self.group_block_sizes = group_block_sizes
+        self.group_cache_families = group_cache_families
+        self.group_effective_block_sizes = [
```
nwpu-zxr, line 82: Has the impact of `compress_ratio` been considered in the block size? I want to align with the VLLM community, that is, the scheduler keeps the original block size unchanged, and the worker side detects `compress_kv`. Is the current code logic also adaptable to this situation?

Pz1116: Let's mark it as a TODO when we actually align with upstream

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
```diff
@@ -516,8 +576,24 @@ class RequestTracker:
     # NOTE: This field will only be used when you enable kv-event
     token_ids: list[int] | None = None
 
+    # Full prompt length before chunk truncation, used by sparse retention masks.
+    num_prompt_tokens: int | None = None
+
+    block_gvas: list[int] = field(default_factory=list)
```
realliujiaxu: please check `block_gvas` to `last_block_key` is acutally used

Pz1116: good catch, already removed

LCAIZJ (approved):

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
```diff
@@ -519,100 +545,97 @@ def start_load_kv(self, metadata: AscendConnectorMetadata):
...
-                            request.block_hashes,
+                addr_list = []
+                size_list = []
+                key_list = []
+                block_id_list: list[int] = []
+                load_masks = self.token_database.load_mask(request.block_hashes, token_len)
+                for group_id in load_group_ids:
+                    block_ids = request.block_ids_by_group[group_id]
+                    group_block_size = self.grouped_block_size[group_id]
+                    mask_num = load_spec.vllm_cached_tokens // group_block_size * group_block_size
```
LCAIZJ, line 561: In the existing process_tokens() implementation, mask_num filtering is based on the raw start_idx. After #10393 changes group_block_size for c4/c128 to the effective granularity, this filtering can skip a compressed chunk that only partially overlaps the locally cached prefix but still contains tokens that need to be loaded.

Pz1116: process_tokens() filters chunks with `start_idx < mask_num` before mapping
compressed-family chunks back to cache-domain offsets. This is safe for the
current DSV4 default path only because cache_transfer_granularity includes the
C128 effective granularity, so mask_num is expected to be aligned to
block_size * 128.

If a future path allows mask_num to be smaller or not aligned to a compressed
family's effective granularity, this start-only check would skip a chunk that
partially overlaps the local prefix. In that case the condition should be based
on `end_idx <= mask_num`, or we should explicitly assert/alignment-normalize
mask_num to the effective group granularity.

Pz1116: let me mark this as a TODO

PR #10394, opened 2026-06-12
[Feature][DSv4] Backport vLLM #43447 prefix-cache (block-reuse + sliding-window retention) for vLLM 0.21.0

HF-001:
Backports vLLM PR #43447 (DeepSeek-V4 selective prefix-cache retention for sliding-window KV cache) to vLLM versions that lack it (e.g. v0.21.0), via capability/source guarded monkey-patches. It carries two parts of #43447:

- **block-reuse**: free-list front-insertion (`prepend_n` / `free_blocks(prepend=)`), SWA/EAGLE cache-block mask, and the `remove_skipped_blocks` cached/uncached split.
- **selective prefix-cache retention** for sliding-window KV cache: retention interval + reachable block mask, controlled by the `VLLM_PREFIX_CACHE_RETENTION_INTERVAL` env.

Each patch is guarded by a capability/source check, so it is a no-op and is automatically skipped once the underlying vLLM already provides the behavior.

## Relationship to upstream PRs
- Backports [Prefix Caching] DeepSeekv4 - Support selective prefix-cache retention for sliding-window KV cache vllm-project/vllm#43447 in full (both block-reuse and retention).
- Orthogonal to the official compressed-prefix lookup fix already on main: this PR leaves the lookup / read path untouched; retention operates on the write path (`__init__` + `cache_blocks`), sharing only the read-only alignment.

Adds an opt-in env `VLLM_ASCEND_PREFIX_CACHE_RETENTION_INTERVAL` to control sliding-window retention (unset = legacy dense behavior, bit-for-bit identical to before). All patches are guarded and skipped when the underlying vLLM already provides them.

## How was this patch tested?
- `lint-and-select-tests` passed (ruff / format / codespell / mypy).
- New unit tests `tests/ut/core/test_prefix_cache_retention.py` assert the retention reachable-block masks (EAGLE and interval cases) and run on `cpu-0card`.
- `run-selected-tests` on `cpu-0card` / `a2-1card`.

Co-authored with @wangzhao-11a and @FutureSkyFly .

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/envs.py
+++ b/vllm_ascend/envs.py
@@ -110,6 +110,16 @@ env_variables: dict[str, Callable[[], Any]] = {
     # Control the aclrtMemcpyBatchAsync compile path for KV cache offloading.
     # "1": force enable, "0": force disable, None: auto-detect from CANN headers.
     "VLLM_ASCEND_ENABLE_BATCH_MEMCPY": lambda: os.getenv("VLLM_ASCEND_ENABLE_BATCH_MEMCPY", None),
+    # Retain sparse sliding-window KV checkpoints for prefix caching.
+    # Unset (None, default): dense local checkpointing (current behavior, byte-for-byte).
+    # 0: keep only the latest completed prompt (replay) boundary.
+    # >0: keep one tail per interval-sized segment (must be a multiple of
+    #     scheduler_block_size). Applies to SWA only, not Mamba/linear attn.
+    "VLLM_ASCEND_PREFIX_CACHE_RETENTION_INTERVAL": lambda: (
+        int(os.environ["VLLM_ASCEND_PREFIX_CACHE_RETENTION_INTERVAL"])
+        if "VLLM_ASCEND_PREFIX_CACHE_RETENTION_INTERVAL" in os.environ
+        else None
+    ),
 }
 
 # end-env-vars-definition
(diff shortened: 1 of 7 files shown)


vllm_ascend/envs.py
```diff
@@ -110,6 +110,16 @@
     # Control the aclrtMemcpyBatchAsync compile path for KV cache offloading.
     # "1": force enable, "0": force disable, None: auto-detect from CANN headers.
     "VLLM_ASCEND_ENABLE_BATCH_MEMCPY": lambda: os.getenv("VLLM_ASCEND_ENABLE_BATCH_MEMCPY", None),
+    # Retain sparse sliding-window KV checkpoints for prefix caching.
+    # Unset (None, default): dense local checkpointing (current behavior, byte-for-byte).
+    # 0: keep only the latest completed prompt (replay) boundary.
+    # >0: keep one tail per interval-sized segment (must be a multiple of
+    #     scheduler_block_size). Applies to SWA only, not Mamba/linear attn.
+    "VLLM_ASCEND_PREFIX_CACHE_RETENTION_INTERVAL": lambda: (
```
MengqingCao: to keep the same usage, let's just use `VLLM_PREFIX_CACHE_RETENTION_INTERVAL`, thus won't introduce userface changes when the version is upgraded.

HF-001: > to keep the same usage, let's just use `VLLM_PREFIX_CACHE_RETENTION_INTERVAL`, thus won't introduce userface changes when the version is upgraded.

@MengqingCao thank you, i have fixed it

wangxiyuan (approved):

underfituu:
Thanks a lot! We’ve actually been brainstorming how to align #43447 with our main branch lately. Since many of the modifications in `block_pool.py` and `kv_cache_utils.py` touch shared utility methods, it might be cleaner to handle the fine-grained adaptation after vLLM upgrades to **v0.23.0**. This will save us from dealing with a massive amount of patches.

PR #10412, opened 2026-06-12
[Misc][310P] 310P device print compatibility

YangShuai52:
310P uses a separate TORCH_LIBRARY registration path and previously did not expose device_print / device_print_tensor. This PR registers those ops on 310P (C++ binding + Meta kernels) and adds Python custom-op wrappers, so vllm_ascend.utils.device_print() works under graph capture/replay (e.g. ACL graph).

None

- vLLM version: v0.25.1

### DIFF
--- a/csrc/torch_binding.cpp
+++ b/csrc/torch_binding.cpp
@@ -2125,6 +2125,14 @@ TORCH_LIBRARY_EXPAND(CONCAT(_C, _ascend), ops)
         "                               Tensor? gk=None) -> Tensor");
     ops.impl("npu_recurrent_gated_delta_rule", torch::kPrivateUse1, &vllm_ascend::npu_recurrent_gated_delta_rule);
 
+    ops.def("device_print(str msg) -> ()");
+    ops.impl("device_print", c10::DispatchKey::CompositeExplicitAutograd,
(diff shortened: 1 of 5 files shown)


csrc/torch_binding.cpp
ZT-AIA, line 2181: Isn't there someone with the same name? What are the differences between the two?

yiz-liu (commented):
1. Consider remove `_mark_op_side_effect` if you already wrap it with custom ops;
2. Update demo to prove it's now compilation-compatible.

vllm_ascend/ops/register_custom_ops.py
```diff
@@ -296,3 +319,19 @@ def _muls_add_impl_fake(
...
     dispatch_key="PrivateUse1",
 )
+
+direct_register_custom_op(
+    op_name="device_print_op",
+    op_func=_device_print_op_impl,
+    fake_impl=_device_print_op_fake,
+    mutates_args=[],
+    dispatch_key="PrivateUse1",
+)
```
ZT-AIA, line 286: Confirm whether it is necessary to register again as a torch custom op. Previously, it has been registered as torch AscendC.

PR #10423, opened 2026-06-12
[CI][Build] Fix dockerfile env error

wangxiyuan:

This PR fixes an issue where the `VLLM_BATCH_INVARIANT` environment variable was not correctly set during the Docker image build process. Previously, it was defined as an `ENV` variable, which made it persist in the final image by default. To prevent this, the PR removes it from the global `ENV` block. However, the current implementation attempts to set it inline before sourcing `set_env.sh`, which only applies the variable to that specific command and does not persist it for the subsequent `pip install` command. This PR fixes this by correctly exporting the variable during the build process.

No.

Docker build verified.

- vLLM version: v0.21.0


wangxiyuan:
@gemini-code-assist review

PR #10435, opened 2026-06-13
[Feature] Support AscendStore load masks

Pz1116:
## What

Add load-only masking for AscendStore KV Pool get paths.

## Why

Hybrid/sliding-window groups such as DSV4 C1 should not fetch chunks that the local KV cache manager would not populate. This keeps the change focused on reducing unnecessary get-side reads without introducing store-side retention semantics.

## Changes

- Resolve the per-group KV cache manager from kv_cache_config when available.
- Generate per-group load masks using an all-present external block pool.
- Apply the mask in both synchronous worker get and asynchronous recv-thread get paths.
- Preserve the existing store path and fall back to full load if no compatible manager is available.

## Validation

vLLM version: v0.20.1

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
@@ -16,6 +16,7 @@ from vllm.v1.core.sched.output import NewRequestData
 _GROUPED_BLOCK_HASH_DOMAIN = b"vllm-ascend-grouped-block-hash-v1\0"
 _GROUPED_BLOCK_HASH_LENGTH_PREFIX_BYTES = 4
 _CACHE_MISSING = object()
+_MANAGER_CLASS_CACHE_ATTR = "_manager_class_cache"
 
 
 # Parameters related to the key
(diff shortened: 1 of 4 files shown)


vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
```diff
@@ -256,6 +341,67 @@ def get_block_size(self, kv_cache_group_id: int) -> int:
...
         return self.block_size[kv_cache_group_id]
 
+    def load_mask(
+        self,
+        block_hashes: BlockHashList | list[str],
+        token_len: int,
+        kv_cache_group_id: int,
+        cache_role: str = "kv",
+    ) -> list[bool] | None:
+        if cache_role != "kv" or not block_hashes or kv_cache_group_id >= len(self.kv_cache_groups):
```
LCAIZJ, line 351: `cache_role != "kv"`, What is the proportion of other types, such as the state type?

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
```diff
@@ -256,6 +341,67 @@ def get_block_size(self, kv_cache_group_id: int) -> int:
...
+            return None
+
+        cache_family = self.group_cache_families.get(cache_role, {}).get(kv_cache_group_id, "default")
+        if infer_cache_family_ratio(cache_family) > 1:
```
LCAIZJ, line 355: Is this only for SWA?

Pz1116:
closed as #10393 is merged

PR #10439, opened 2026-06-13
[Feature] Support AscendStore retention masks for v0.20.2

Pz1116:

This PR backports the AscendStore retention and mask flow to the v0.20.2rc branch.

It integrates the relevant pieces from #10198 and #10393 so the release branch uses coordinator-derived hybrid group semantics instead of a per-group local mask path.

It also backports the AscendStore side of #11370: when an EAGLE/MTP group is merged with same-spec sibling groups, store-mask generation uses the full eagle-containing attention group instead of only the originally marked MTP group.

The expected behavior is:

- AscendStore store only writes chunks allowed by the coordinator-derived retention/reachable mask.
- AscendStore async and sync get only load chunks allowed by the same coordinator-derived load mask.
- Hybrid group lookup is coordinated across required groups.
- Non-hybrid or partial-group fallback keeps the existing per-group exists path.
- If the runtime does not expose the newer KV cache spec registry API, AscendStore falls back to the older manager map for vLLM 0.20.x/0.21.x compatibility.

Yes. AscendStore on the v0.20.2rc branch supports coordinator-derived store/load masks for hybrid KV Pool models.

CI plus AscendStore unit tests.

- vLLM version: 

### DIFF
--- a/tests/ut/distributed/ascend_store/_mock_deps.py
+++ b/tests/ut/distributed/ascend_store/_mock_deps.py
@@ -23,6 +23,7 @@ Usage at the top of each test file:
     import tests.ut.distributed.ascend_store._mock_deps  # noqa: F401, E402
 """
 
+import logging
 import os
 import sys
 import types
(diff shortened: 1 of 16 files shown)


LCAIZJ:
Could you provide a detailed explanation of the concept and logic behind the "retention mask"? Alternatively, could you share any relevant documentation or descriptions from vLLM?

Pz1116:
> Could you provide a detailed explanation of the concept and logic behind the "retention mask"? Alternatively, could you share any relevant documentation or descriptions from vLLM?

Updated in main thread, this is a backport of https://github.com/vllm-project/vllm/pull/43447 and https://github.com/vllm-project/vllm/pull/44774 from upstream for so we can achieve less kv pool usage for dsv4 due to less kv block needed to be stored, for details you can refer to the original prs.

PR #10440, opened 2026-06-13
[Prefix Cache][0.21.0rc] DeepSeek-V4 partial compressed prefix-cache hit

Liuchenbing-2026:

DeepSeek-V4 compresses its KV cache: at `--block-size 128` one cache block already represents `128 × compress_ratio = 16384` tokens. With vLLM prefix caching, a block is only reused on a **full-block** hit, so any prompt **shorter than 16K** (e.g. 2K–8K multi-turn requests) gets **0 hit** and pays full prefill every turn.

This PR adds a **partial (sub-block) prefix-cache hit** path for the DSv4 compressed MLA cache:

- when there is no full-block hit, look up the **longest cached sub-block prefix** and copy those tokens’ KV into the request’s first block, so the prefill only recomputes the uncached tail;
- a small **per-device copy** path (one D2H/H2D batched copy per device) materializes the partial KV without a custom kernel;
- a hit-rate-driven **gate** (`VLLM_ASCEND_DSV4_PARTIAL_MIN_HIT_TOKENS`, default `0`) lets deployments skip very short partial hits when the copy cost outweighs the prefill saved.

It is **off by default** (zero behavior change unless the env/threshold is set) and only affects the DSv4 compressed-cache path.

No user-facing API change. One **optional** env knob is added:

- `VLLM_ASCEND_DSV4_PARTIAL_MIN_HIT_TOKENS` (default `0`): minimum matched-token length for a partial hit to be used; `0` keeps every hit.

Default behavior is unchanged.

- **Unit**: `tests/ut/.../test_dsv4_partial_prefix_cache.py` covering boundary registration, longest-prefix lookup, and the gate.
- **E2E**: DeepSeek-V4 on Atlas A3, `--block-size 128`, multi-turn prefix-cache workload (2K–8K prompts); verified prefix-cache hit rate rising from 0 and TTFT dropping on cache hits, output identical to the no-cache baseline.

## Prefix cache performance data

Test configuration:

```text
input_len: 2048 / 4096
output_len: 1024
concurrency: 32
data_num: 128
repeat_rate: 0.9
request_rate: 0
seed: 123

Steady-state average over run2-run3. partial means prefix cache enabled, and baseline means prefix cache disabled.

 Input    Mode        Prefix Hit     TTFT avg    TPOT avg     Req/s    Out tok/s    Total tok/s    Compared with baseline
━━━━━━━  ━━━━━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━━  ━━━━━━━━━━  ━━━━━━━━  ━━━━━━━━━━━  ━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    2k    partial        78.587%    1043.3 ms     34.8 ms    0.8539     874.4778      2626.9060    Total tok/s +6.94%, TTFT -21.99%
───────  ──────────  ────────────  ───────────  ──────────  ────────  ───────────  ─────────────  ───────────────────────────────────
    2k    baseline        0.000%    1337.5 ms     37.3 ms    0.7985     817.7283      2456.4383    -
───────  ──────────  ────────────  ───────────  ──────────  ────────  ───────────  ─────────────  ───────────────────────────────────
    4k    partial        81.741%    1168.2 ms     36.0 ms    0.8326     852.5774      4266.2760    Total tok/s +24.76%, TTFT -62.01%
───────  ──────────  ────────────  ───────────  ──────────  ────────  ───────────  ─────────────  ───────────────────────────────────
    4k    baseline        0.000%    3074.9 ms     43.5 ms    0.6674     683.3978      3419.6729    -
```

### DIFF
--- a/vllm_ascend/core/single_type_kv_cache_manager.py
+++ b/vllm_ascend/core/single_type_kv_cache_manager.py
@@ -1,14 +1,18 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
 import itertools
+from collections import defaultdict
 from collections.abc import Sequence
 
 from vllm.utils.math_utils import cdiv
 from vllm.v1.core.block_pool import BlockPool
 from vllm.v1.core.kv_cache_utils import (
+    BlockHash,
     BlockHashList,
     BlockHashListWithBlockSize,
+    BlockHashWithGroupId,
     KVCacheBlock,
+    make_block_hash_with_group_id,
 )
 from vllm.v1.core.single_type_kv_cache_manager import (
     FullAttentionManager,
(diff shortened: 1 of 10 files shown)


Angazenn:
With this partial cache hit, how do you handle the different num_computed_tokens in different cache group?

Angazenn:
Generally, I'm fine with it on 0.21.0. But we should track how vllm plans to implement it on main branch. See [#49539](https://github.com/vllm-project/vllm/pull/45939). cc @underfituu

frankie-ys:
> Generally, I'm fine with it on 0.21.0. But we should track how vllm plans to implement it on main branch. See [#49539](https://github.com/vllm-project/vllm/pull/45939). cc @underfituu

I have implemented this pr on main branch of vllm-ascend, but it meet a problem which reuse a partial private block across requests.This problem happened in 0.21.0rc1 and main branch. And I have fix it, i will add some comment on this pr.

frankie-ys (changes requested):

vllm_ascend/core/single_type_kv_cache_manager.py
```diff
@@ -169,21 +284,89 @@ def cache_blocks(
...
+        # block is request.block_hashes[i] (the raw 128-token chained hash).
+        # When block_size == hash_block_size the base class uses the raw
+        # chained hashes directly; otherwise both the base class and the
+        # lookup side in find_longest_cache_hit compose them with the same
+        # BlockHashListWithBlockSize(target=self.block_size), so the two
+        # sides always match. A key only attests the first
+        # (i+1)*hash_block_size tokens; the tail block claimed beyond
+        # max_length is backed by the private copy made in
+        # allocate_new_computed_blocks.
+        super().cache_blocks(request, compressed_tokens)
```
frankie-ys, line 298: Full-block cache using logical_block_size (block_size * compress_ratio) so that one hash covers all 16384 logical tokens that a compressed block represents.  Using block_size=128 (as the PR did via super().cache_blocks) would only verify 128 tokens per hash entry, letting unmatched tail KV from the source request leak into the consuming request — the root cause of cross-request KV contamination.Partial (sub-block) boundary hashing — independent of full-block cache.

Liuchenbing-2026:

PR #10441, opened 2026-06-13
[Feature] Adapt GLM-5.2

ZYang6263:
This PR adds support for GLM-5.2 on Ascend.

GLM-5.2 uses a layer-wise DSA indexer pattern. Some backbone layers skip Lightning Indexer computation and reuse top-k indices generated by shared indexer layers through the index cache. However, the existing SFA implementation assumes that every DSA attention layer owns an Indexer, which causes initialization and runtime failures for layers configured with skip_topk.

The main changes include: 
- Support DSA layers without a local `Indexer` when `skip_topk` is enabled. 
- Reuse cached top-k indices for shared indexer layers. 
- Adapt SFA and DSA-CP paths to handle layers without indexer KV cache. 
- Add a temporary vLLM patch and a multi-node GLM-5.2 W8A8 nightly test.

- vLLM version: v0.23.0

### DIFF
--- a/.github/workflows/schedule_nightly_test_a3.yaml
+++ b/.github/workflows/schedule_nightly_test_a3.yaml
@@ -223,6 +223,9 @@ jobs:
           - name: multi-node-GLM-5.1-w8a8-A3
             config_file_path: GLM5_1-W8A8-A3-dual-nodes.yaml
             size: 2
+          - name: multi-node-GLM-5.2-w8a8-A3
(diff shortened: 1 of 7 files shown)


ZYang6263:
/nightly multi-node-GLM-5.1-w8a8-A3
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/28216322672)

weijinqian0 (approved):

PR #10442, opened 2026-06-13
[BugFix] fix qwen3.5 accuracy bug when sett cp-kv-cache-interleave-si…

weiguihua2:
Fix the issue of incorrect precision when the qwen3.5 cp-kv-cache-interleave-size setting is set to 1024.

When block_size > 1024, the block_size in the block_table is set to the kernel block size, causing issues in the calculation of slot mapping. As a result, the allocation is not evenly distributed across the cards, leading to precision problems.

No

- vLLM version: v0.21.0


linfeng-yuan (approved):
LGTM

pisceskkk (approved):
LGTM, thanks for contribution!

PR #10454, opened 2026-06-14
[CI] main2main 0703 commit: 1f486d9

zhangxinyuehfad:
#### vllm_ascend/distributed/weight_transfer/hccl_engine.py
- Conditionally implement start_weight_update() and finish_weight_update() as no-op methods for non-0.23.0 releases.
- Keep the NPU IPC weight transfer engine compatible with the updated WeightTransferEngine interface.
- Upstream source: vllm#44353 (https://github.com/vllm-project/vllm/pull/44353).

---

#### vllm_ascend/patch/platform/patch_torch_accelerator.py
- Redirect torch.accelerator.get_memory_info() to torch.npu.mem_get_info() on non-0.23.0.
- Avoid crashes caused by the unsupported C10 DeviceAllocator path when constructing MemorySnapshot.
- Align with the existing NPU-specific memory API patches.
- Upstream source: commit 747b068 (v0.24.0+ MemorySnapshot(device=device) path).

--- 

#### vllm_ascend/patch/worker/patch_qwen3_dflash.py
- Wrap DFlashQwen3ForCausalLM._read_mask_embedding() to ignore optional mask embedding download failures.
- Preserve the expected "mask embedding not present" behavior when the file is unavailable.
- Upstream source: vllm#46104 (https://github.com/vllm-project/vllm/pull/46104).

--- 

#### vllm_ascend/patch/worker/patch_v2/patch_input_batch.py
- Forward is_padding and prompt_lens when constructing AscendInputBatch.
- Match the updated upstream InputBatch interface and avoid initialization failures on newer releases.
- Upstream source: vllm#40654 (https://github.com/vllm-project/vllm/pull/40654).

--- 

#### vllm_ascend/patch/worker/patch_deepseek_v2.py
- Add the reduce_results argument to DeepseekV2MLAAttention.
- Forward the parameter to RowParallelLinear to stay compatible with the updated upstream attention initialization.
- Keep the implementation compatible across all supported vLLM versions.
- Upstream source: vllm#45895 (https://github.com/vllm-project/vllm/pull/45895).

--- 

#### vllm_ascend/distributed/device_communicators/npu_communicator.py
- Register a no-op all2all_manager for NPUCommunicator.
- Bypass the upstream MoE fault-tolerance check (which queries all2all_manager when data_parallel_size > 1 and is_moe) while preserving the existing MC2 communication path.
- Keep compatibility with the updated distributed initialization.
- Related upstream changes:
- vllm#46892 (https://github.com/vllm-project/vllm/pull/46892)

--- 

#### vllm_ascend/ops/fused_moe/fused_moe.py
- Share routed expert parameters through direct nn.Parameter aliasing instead of creating wrapper parameters.
- Ensure both legacy and routed_experts parameter paths reference the same underlying weights.
- Apply the aliasing strategy to all routed-expert MoE models on newer vLLM releases.
- Related upstream changes:
- vllm#40996 (https://github.com/vllm-project/vllm/pull/40996)
- vllm#46892 (https://github.com/vllm-project/vllm/pull/46892)

--- 

#### vllm_ascend/patch/platform/patch_weight_transfer_engine.py
- Adapt WeightTransferEngineFactory.create_engine() and WeightTransferEngine.__init__() to support both legacy and current upstream signatures.
- Keep weight transfer compatible across v0.23.0, v0.24.0, and newer upstream releases.
- Upstream source: vllm#44353 (https://github.com/vllm-project/vllm/pull/44353).

--- 

#### vllm_ascend/patch/worker/patch_deepseek_v2.py
- Remove the upstream model-level all-gather path for DeepSeek-V2 on non-0.23.0.
- Keep the implementation compatible with the Ascend MC2 dispatch flow.
- Avoid tensor shape mismatches and residual concatenation failures introduced by the upstream refactor.
- Related upstream changes:
- vllm#41184 (https://github.com/vllm-project/vllm/pull/41184)

--- 

#### vllm_ascend/ops/fused_moe/fused_moe.py
- Remove the unnecessary .contiguous() call after weight transposition on non-0.23.0.
- Reduce transient NPU peak memory during MoE weight loading.
- Prevent OOM caused by duplicate temporary tensor allocations.
- Related upstream changes:
- vllm#44589 (https://github.com/vllm-project/vllm/pull/44589)

- vLLM version: v0.23.0


realliujiaxu (changes requested):
Lets's merge this after v0.21.0 is released

MrZ20:
The upgrade has been completed, this PR can be closed. Thank you for your participation.
- https://github.com/vllm-project/vllm-ascend/pull/10476

vllm_ascend/_310p/fused_moe/fused_moe.py
```diff
@@ -66,14 +66,22 @@ def maybe_make_prepare_finalize(self, routing_tables=None):
...
-        # Fused gate_up_proj (column parallel)
-        w13_data = self._maybe_pad_weight(layer.w13_weight.data).transpose(1, 2).contiguous()
-        w13_data = maybe_trans_nz(w13_data)
-        layer.w13_weight = torch.nn.Parameter(w13_data, requires_grad=False)
-        # down_proj (row parallel)
-        w2_data = self._maybe_pad_weight(layer.w2_weight.data).transpose(1, 2).contiguous()
-        w2_data = maybe_trans_nz(w2_data)
-        layer.w2_weight = torch.nn.Parameter(w2_data, requires_grad=False)
+        if not vllm_version_is("0.23.0"):
+            w13_data = self._maybe_pad_weight(layer.w13_weight.data).transpose(1, 2)
```
MengqingCao, line 70: why we need remove contiguous here?

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -112,11 +112,18 @@ def maybe_make_prepare_finalize(self, routing_tables=None):
     def process_weights_after_loading(self, layer):
         super(UnquantizedFusedMoEMethod, self).process_weights_after_loading(layer)
 
-        w13_data = self._maybe_pad_weight(layer.w13_weight.data).transpose(1, 2).contiguous()
-        layer.w13_weight = torch.nn.Parameter(w13_data, requires_grad=False)
+        if not vllm_version_is("0.23.0"):
+            w13_data = self._maybe_pad_weight(layer.w13_weight.data).transpose(1, 2)
```
MengqingCao: ditto, why remove contiguous here?

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -496,27 +503,11 @@ def _get_quant_type(self) -> QuantType:
...
-                return original_process_weights(layer, *args, **kwargs)
-
-            self._quant_method.process_weights_after_loading = wrapped_process_weights  # type: ignore[method-assign]
+                object.__setattr__(self, name, param)
 
         def _needs_routed_expert_parameter_aliases(self) -> bool:
-            vllm_config = get_current_vllm_config()
-            hf_config = getattr(vllm_config.model_config, "hf_config", None)
-            return getattr(hf_config, "model_type", None) == "gpt_oss"
+            return True
```
MengqingCao: let's remove these 2 functions if we don't need to alia parameters

vllm_ascend/_310p/fused_moe/fused_moe.py
zhangxinyuehfad: 1. Peak memory reduction. After upstream PR #44589 changed the weight loading timing, all MoE weights reside on NPU simultaneously during process_weights_after_loading. .transpose(1,2) followed by .contiguous() materializes a full copy, doubling the memory footprint — causing OOM on large MoE models like Qwen3-Next-80B-A3B (TP=4).
2. Redundant operation. .transpose(1,2) is already a zero-copy view with correct logical layout. The subsequent maybe_trans_nz() handles non-contiguous inputs and outputs contiguous NZ storage, so the intermediate .contiguous() is unnecessary.

vllm_ascend/ops/fused_moe/fused_moe.py
zhangxinyuehfad: yep

vllm_ascend/patch/worker/patch_deepseek_v2.py
```diff
@@ -277,3 +281,78 @@ def _deepseek_v2_mla_attention_init(
...
+            )
+        else:
+            llama_4_scaling = None
+
+        aux_hidden_states = []
+        for idx, layer in enumerate(
+            islice(self.layers, self.start_layer, self.end_layer),
+            start=self.start_layer,
+        ):
+            if idx in self.aux_hidden_state_layers:
```
MengqingCao, line 332: I think we could keep the same implementation with vLLM, cause the all-gather won't be triggered if we don't enable sp with the same way of vLLM. @linfeng-yuan please also take a look.

zhangxinyuehfad: The patch is necessary, confirmed by experiment: removing it causes test_deepseek_v2_lite_fc1_tp2 to crash immediately (torch.cat([hidden_states, None]) → TypeError at deepseek_v2.py:1460).
vLLM SP and Ascend MC2 are mutually incompatible communication paths. vLLM SP requires model-level all-gather to expand hidden_states to full size; MC2 requires hidden_states to stay partition-shaped, with mc2_mask handling slicing internally. Enabling both causes xActiveMask dim0 != x dim0. This is also why platform.py:648-650 explicitly disables use_sequence_parallel_moe.
The per-layer all-gather in upstream is still needed for GPU non-SP paths, so it's not suitable for upstream removal. However, the missing residual is None guard in torch.cat is a cross-platform defect worth upstreaming. The current patch remains the cleanest fix for Ascend.

MengqingCao (approved):

realliujiaxu (approved):

PR #10459, opened 2026-06-14
[Misc] main2main 0612

MrZ20:

This PR updates vLLM-Ascend for the vLLM main upgrade:

- Previous vLLM main: [`dc68bd8c4199b00631fe71eb37313f406cc66ac1`](https://github.com/vllm-project/vllm/commit/dc68bd8c4199b00631fe71eb37313f406cc66ac1)
- New vLLM main: [`a30addc7548a9a8b9b3323a7bc3eb7d7c4895d1c`](https://github.com/vllm-project/vllm/commit/a30addc7548a9a8b9b3323a7bc3eb7d7c4895d1c)
- Upstream compare: [`dc68bd8c...a30addc7`](https://github.com/vllm-project/vllm/compare/dc68bd8c4199b00631fe71eb37313f406cc66ac1...a30addc7548a9a8b9b3323a7bc3eb7d7c4895d1c)

The final patch changes 6 files. The mapping below explains why each changed file is needed and which upstream vLLM PR/commit in this upgrade range caused it.

#### `tests/e2e/pull_request/one_card/spec_decode/test_extract_hidden_states.py`

- Upstream source: [vLLM #43805](https://github.com/vllm-project/vllm/pull/43805/changes#diff-72eec9c1ef7542e20f0c2610fd9e255959d9574e45321b0c2c8996301599a90fR287).
- What changed upstream: `ExampleHiddenStatesConnector` was refactored to save hidden states asynchronously, pre-create `.lock` files, and expose connector helpers such as `load_hidden_states()` / `cleanup_hidden_states()` as the synchronized read path.
- Local change: replace direct `safe_open(hidden_states_path, "pt")` reads with `example_hidden_states_connector.load_hidden_states(hidden_states_path)`, then read `hidden_states` and `token_ids` from the returned object. For vLLM main, cleanup is performed through `cleanup_hidden_states()` after the read; the v0.23.0 path keeps the existing existence check for release compatibility.
- Why: after upstream async writes, the output path can be visible before the safetensors write has completed. Reading the file directly can race the writer. The connector helper respects the upstream lock protocol and avoids flaky hidden-state extraction tests.

#### `tests/ut/patch/platform/test_patch_tool_choice_none_content.py`

- Upstream source: [vLLM #45190](https://github.com/vllm-project/vllm/pull/45190/changes#diff-42bfc2c85e7e8c3e45317685ea1cd396c6f617a5eee3540d9b8758b8b06d60d6).
- What changed upstream: Responses API parsing was unified to use `parser.parse()`, and the old `DelegatingParser._parse_tool_calls()` helper was removed. The current hook is `DelegatingParser._extract_tool_calls(content=..., request=..., enable_auto_tools=...)`.
- Local change: update the unit test to call `_extract_tool_calls()` with the current argument order.
- Why: the Ascend patch for forced tool-choice requests with `content=None` is applied to the current upstream hook. The test must exercise the method that is actually used by vLLM main; otherwise it either fails with `AttributeError` or no longer verifies the live parser path.

#### `vllm_ascend/ops/fused_moe/fused_moe.py`

- Upstream source: the routed-expert weight-loading fixes in [vLLM #45054](https://github.com/vllm-project/vllm/pull/45054), commit [`f4966f8b3ddf757c607d57d4bb35624e2ee4f6b4`](https://github.com/vllm-project/vllm/commit/f4966f8b3ddf757c607d57d4bb35624e2ee4f6b4), and [vLLM #45067](https://github.com/vllm-project/vllm/pull/45067), commit [`7920ccb97c2d27d0a1a822e42d11ba369c2f255c`](https://github.com/vllm-project/vllm/commit/7920ccb97c2d27d0a1a822e42d11ba369c2f255c). These are follow-up fixes to the earlier fused-MoE routed-expert refactor.
- What changed upstream: vLLM moved model-specific loading paths to the new `experts.routed_experts.*` / `mlp.experts.routed_experts.*` parameter ownership. #45054 updates Aria, Qwen3-VL-MoE, Step3, Step3.5, and related quant loading paths. #45067 updates GPT-OSS legacy checkpoint name remapping after the fused-MoE refactor.
- Local change: import `FusedMoE` and `MoERunner` from `vllm.model_executor.layers.fused_moe.layer` at module scope, while leaving only `UnquantizedFusedMoEMethod` under the v0.23.0/main version branch. Also narrow `_needs_routed_expert_parameter_aliases()` from the broad local list of model types to `model_type == "gpt_oss"`.
- Why: the broad alias workaround was needed while upstream model loaders still looked for expert weights on the old MoE module path. In this vLLM range, upstream fixed the affected model loaders to use `routed_experts` directly, so keeping aliases for Aria/Qwen3/Step3/etc. would preserve stale duplicate parameter names. The remaining local alias path is kept only for the Ascend GPT-OSS compatibility case, while the import cleanup follows the current main package shape and keeps the release/main `UnquantizedFusedMoEMethod` path difference explicit.

#### `vllm_ascend/worker/v2/attn_utils.py`

- Upstream source: [vLLM #44586](https://github.com/vllm-project/vllm/pull/44586/changes#diff-9b864c13232e1f03b906ccc83311fa78d1c37988616ae88b4afdbb5c0d186a75R397).
- What changed upstream: GPU `build_attn_metadata()` gained a `causal: bool = True` argument and passes it into `CommonAttentionMetadata(causal=causal)`.
- Local change: add the same `causal` keyword to Ascend `build_attn_metadata()` and pass it through to `AscendCommonAttentionMetadata`.
- Why: vLLM main can now call `build_attn_metadata(..., causal=...)`, especially from newer speculative decoding paths. Ascend must accept the same keyword to avoid `unexpected keyword argument 'causal'`, and must preserve `causal=False` instead of silently forcing all metadata to causal attention.

#### `vllm_ascend/worker/v2/model_runner.py`

- Upstream source: [vLLM #45163](https://github.com/vllm-project/vllm/pull/45163/changes#diff-5823f988fc0264681a80db24ccaba4d364f14394815d83ec1d90944c09f571f0L266).
- What changed upstream: `GPUModelRunner.__init__` removed the early instance-attribute assignment `self.decode_query_len = self.num_speculative_steps + 1`. `self.decode_query_len` is now created later in `load_model()` after `model_state` exists, using `self.num_speculative_steps + self.model_state.num_new_sampled_tokens_per_step`.
- Local change: in `NPUModelRunner.__init__`, replace the early read of `self.decode_query_len` with a local variable `decode_query_len = self.num_speculative_steps + 1` only for the immediate Ascend setup calls `set_cos_and_sin()` and `set_mc2_tokens_capacity()`. This does not reintroduce `self.decode_query_len` assignment in `__init__`.
- Why: Ascend runs these setup calls immediately after `GPUModelRunner.__init__`, at a point where upstream main no longer has `self.decode_query_len` yet. The local value is a compatibility fallback matching the pre-#45163 formula for the currently supported Ascend path, while leaving upstream's later `self.decode_query_len` initialization model intact.

#### Does this PR introduce _any_ user-facing change?

No intended user-facing API change. This is a compatibility update for the new vLLM main baseline. It fixes test/runtime compatibility with upstream parser, connector, attention metadata, model runner, and fused-MoE loading changes.

- CI: https://github.com/vllm-project/vllm-ascend/actions/runs/27770885645

Compatibility targets:

- vLLM version: v0.23.0


vllm_ascend/patch/worker/patch_gpt_oss.py
```diff
@@ -0,0 +1,67 @@
...
+# limitations under the License.
+#
+
+from collections.abc import Iterable
+
+import torch
+from vllm.model_executor.models.gpt_oss import GptOssModel
+
+
+def _remap_gpt_oss_routed_expert_weight_names(
```
menogrey: why adding this patch? Can this PR resolve this issue? https://github.com/vllm-project/vllm/pull/45067

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -145,6 +152,8 @@ def apply(
         global_redundant_expert_num: int = 0,
         pertoken_scale: torch.Tensor | None = None,
         mc2_mask: torch.Tensor | None = None,
+        topk_weights: torch.Tensor | None = None,
```
menogrey: this may be a meaningful change? where is the `select_experts` in vllm? Is it possible that we are using vllm default `select_experts` and using a native impl (not efficient)?

vllm_ascend/eplb/utils.py
```diff
@@ -16,23 +16,40 @@
 #
 # Todo: Once https://github.com/vllm-project/vllm/pull/23553 is merged in vllm. Remove this model register.
 import types
+from typing import Any
 
 import torch
 
 
+def get_moe_weight_owner(experts: Any) -> Any:
```
menogrey: No need to wrap it, just use `.routed_experts`

vllm_ascend/quantization/compressed_tensors_config.py
```diff
@@ -49,6 +49,12 @@ def _remove_quantization_method():
 QUANTIZATION_SCHEME_MAP_TYPE = dict[str, dict[str, "QuantizationArgs"] | None]
 
 
+def _is_fused_moe_layer(layer: torch.nn.Module) -> bool:
```
menogrey: No need to wrap it. use `isinstance(layer, RoutedExperts)` directly.

vllm_ascend/quantization/compressed_tensors_config.py
```diff
@@ -167,23 +175,27 @@ def get_quant_method(
...
-        if isinstance(layer, FusedMoE):
+        if _is_fused_moe_layer(layer):
             # Delayed import to avoid circular import
             from vllm_ascend.ops.fused_moe.fused_moe import AscendUnquantizedFusedMoEMethod
 
-            layer.ascend_quant_method = COMPRESSED_TENSORS_METHOD
+            # vLLM PR #41184 makes this branch accept either legacy FusedMoE
+            # or target-main RoutedExperts. After the runtime predicate, use a
+            # narrow local cast for MoE-only dynamic attributes.
+            moe_layer = cast(Any, layer)
```
menogrey: No need to cast, we just add a attribute. Any lint error here?

vllm_ascend/quantization/fp8_config.py
```diff
@@ -15,6 +15,12 @@
 QUANTIZATION_SCHEME_MAP_TYPE = dict[str, dict[str, QuantizationArgs] | None]
 
 
+def _is_fused_moe_layer(layer: torch.nn.Module) -> bool:
```
menogrey, line 20: No need to wrap it.

vllm_ascend/quantization/fp8_config.py
```diff
@@ -113,10 +119,14 @@ def get_quant_method(
             scheme = create_scheme_for_layer(self.quant_description, prefix, "ds_linear", self.packed_modules_mapping)
             quant_method = AscendLinearMethod(scheme)
             return quant_method
-        if isinstance(layer, FusedMoE):
-            layer.ascend_quant_method = FP8_METHOD
+        if _is_fused_moe_layer(layer):
+            # vLLM PR #41184 makes this branch accept either legacy FusedMoE
+            # or target-main RoutedExperts. After the runtime predicate, use a
+            # narrow local cast for MoE-only dynamic attributes.
+            moe_layer = cast(Any, layer)
```
menogrey: No need to cast.

vllm_ascend/quantization/method_adapters.py
```diff
@@ -270,7 +282,14 @@ def apply(
...
         apply_router_weight_on_input: bool = False,
         mc2_mask: torch.Tensor | None = None,
+        topk_weights: torch.Tensor | None = None,
+        topk_ids: torch.Tensor | None = None,
     ) -> torch.Tensor:
+        # vLLM PR #41184 passes precomputed routing tensors from
+        # RoutedExperts.forward_modular into quant_method.apply. Ascend's
+        # quantized schemes still select experts inside their own apply path, so
+        # the adapter accepts the keywords but does not forward them.
+        del topk_weights, topk_ids
```
menogrey: Why `UnquantizedFusedMoeMethod` use the `topk_weights` and `topk_ids` args, but `QuantizedFusedMoeMethod` del the args?

vllm_ascend/patch/worker/patch_gpt_oss.py
MrZ20: https://github.com/vllm-project/vllm/pull/45067 only fixed the MXFP4/Quark paths for GPT-OSS; the generic/FP8 path under `_load_weights_other` still uses legacy `mlp.experts.*` to look up `params_dict[name]`.

The value of the current patch lies in filling in the generic/FP8 path, which means PR #45067 cannot provide full coverage.

vllm_ascend/ops/fused_moe/fused_moe.py
MrZ20: We use `vllm_ascend.ops.fused_moe.experts_selector.select_experts` here instead of the default vLLM implementation.
It prioritizes the Ascend NPU gating top-k fused operator and falls back to the native selector only when the fused operator is unsupported; this is not an accidental adoption of the inefficient default upstream path.

vllm_ascend/quantization/method_adapters.py
MrZ20: `AscendUnquantizedFusedMoEMethod` can reuse the `top-k` passed from upstream, as this layer directly governs the execution of Ascend fused MoE.

`AscendFusedMoEMethod` serves as a quantization scheme adapter. Multiple underlying quantization implementations currently handle expert selection, placement, and quantization-specific logic within their respective `apply()` functions. Directly passing `top-k` through would alter the semantics of all quantization `kernels`.

The current `del topk_weights, topk_ids` maintains signature compatibility. In the long term, a separate optimization PR can be created to enable quantized MoE to reuse preselected top-k values.

vllm_ascend/eplb/utils.py
MrZ20: done

vllm_ascend/quantization/compressed_tensors_config.py
MrZ20: done.

vllm_ascend/quantization/fp8_config.py
MrZ20: done.

vllm_ascend/quantization/fp8_config.py
MrZ20: done.

vllm_ascend/quantization/compressed_tensors_config.py
MrZ20: done.

vllm_ascend/patch/worker/patch_gpt_oss.py
menogrey: Okay, I see it, we are using GPT-OSS bf16 on our testcase, and there is an open upstream PR https://github.com/vllm-project/vllm/pull/45818

vllm_ascend/ops/fused_moe/fused_moe.py
menogrey: As you mentioned, I means we are using the forward_modular path, and using the topk_weights / topk_ids args, how the vllm upstream produce the topk_weights / topk_ids, can we use it?

> Make AscendUnquantizedFusedMoEMethod.apply() accept upstream precomputed topk_weights / topk_ids and reuse them when available. 

> Upstream #41184 makes MoERunner choose monolithic vs modular dispatch based on quant_method.is_monolithic: [moe_runner.py#L549](https://github.com/vllm-project/vllm/blob/dc68bd8c4199b00631fe71eb37313f406cc66ac1/vllm/model_executor/layers/fused_moe/runner/moe_runner.py#L549). In the modular path, the runner computes topk_weights/topk_ids and passes them to RoutedExperts.forward_modular(): [moe_runner.py#L558](https://github.com/vllm-project/vllm/blob/dc68bd8c4199b00631fe71eb37313f406cc66ac1/vllm/model_executor/layers/fused_moe/runner/moe_runner.py#L558), [moe_runner.py#L565](https://github.com/vllm-project/vllm/blob/dc68bd8c4199b00631fe71eb37313f406cc66ac1/vllm/model_executor/layers/fused_moe/runner/moe_runner.py#L565). RoutedExperts.forward_modular() then forwards these tensors to the quant method: [routed_experts.py#L1066](https://github.com/vllm-project/vllm/blob/dc68bd8c4199b00631fe71eb37313f406cc66ac1/vllm/model_executor/layers/fused_moe/routed_experts.py#L1066), [routed_experts.py#L1097](https://github.com/vllm-project/vllm/blob/dc68bd8c4199b00631fe71eb37313f406cc66ac1/vllm/model_executor/layers/fused_moe/routed_experts.py#L1097).

vllm_ascend/quantization/method_adapters.py
menogrey: if pass the topk_weights/topk_ids args, we are doing unnecessary calculation before, which is a bad signal. All I want to say is that we must cal topk_weights/topk_ids ourself, not doing by upstream, should check this.

tests/e2e/pull_request/one_card/spec_decode/test_extract_hidden_states.py
```diff
@@ -123,20 +127,23 @@ def _verify_output(output, expected_shape, *, verify_nonzero, verify_token_ids):
...
-    assert os.path.exists(hidden_states_path)
 
-    with safe_open(hidden_states_path, "pt") as f:
-        tensor_names = f.keys()
-        assert "hidden_states" in tensor_names
-        hidden_states = f.get_tensor("hidden_states")
-        assert hidden_states.shape == expected_shape
+    if vllm_version_is("0.23.0"):
+        assert os.path.exists(hidden_states_path)
+        obj = example_hidden_states_connector.load_hidden_states(hidden_states_path)
```
MengqingCao: plz revert this change

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -123,8 +123,9 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
         # is necessary for weight_prfetching function, and MoE communication optimization.
         set_weight_prefetch_method(self.ascend_config.weight_prefetch_config)
         # TODO: remove set_cos_and_sin (together with update_cos_sin) when mla can properly handle cos/sin internally
-        set_cos_and_sin(vllm_config, self.max_num_reqs, self.decode_query_len, self.dtype, self.device)
-        set_mc2_tokens_capacity(vllm_config, self.max_num_reqs, self.decode_query_len)
+        decode_query_len = self.num_speculative_steps + 1
```
MengqingCao: we need to adapt dllm with the following code in the `load_model` period.

```suggestion
        self.decode_query_len = (
            self.num_speculative_steps
            + self.model_state.num_new_sampled_tokens_per_step
        )
```

MengqingCao:
```suggestion
        self.decode_query_len = self.num_speculative_steps + 1
```

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -123,6 +123,7 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
         # is necessary for weight_prfetching function, and MoE communication optimization.
         set_weight_prefetch_method(self.ascend_config.weight_prefetch_config)
         # TODO: remove set_cos_and_sin (together with update_cos_sin) when mla can properly handle cos/sin internally
+        self.decode_query_len = self.num_speculative_steps + 1
```
wxsIcey, line 126: why add it?

wangxiyuan (approved):

PR #10486, opened 2026-06-15
[BugFix]Combine the features of PP and EPLB.

LostFox11:
This PR makes the EPLB feature compatible with PP (Pipeline Parallelism).

  In PP mode, each rank only holds a subset of model layers (`PPMissingLayer` replaces layers not on the current rank). The existing EPLB implementation accessed MoE layers via hardcoded paths (`model.model.layers[layer_id].mlp.experts`), which
   fails in PP since those layers may not exist on the current rank.

  Key changes:
  - Introduce `VllmEplbAdaptor.register_layer()` — a registration mechanism where MoE layers register themselves during initialization (`AscendFusedMoE.__init__`). `PPMissingLayer` (nn.Identity) never calls this, so only real layers on the
  current PP rank are registered.
  - Refactor `VllmEplbAdaptor` to read from the registry (`_registered_moe_layers`) instead of walking `model.model.layers`.
  - Remove `vllm_ascend/eplb/utils.py` (model monkey-patching) — no longer needed since layer info is now provided via registration.
  - Fix D2D transfer `dist.P2POp` calls to use global rank IDs via `comm_group.ranks[]`.
  - Fix the EPLB iteration count issue for non-last PP ranks in the DP scenario.
  
No

  - `test_pp_eplb_adaptor_init_with_registered_layer` — verifies the adaptor correctly picks up MoE layers registered via `register_layer()` under the PP scenario.
  - Existing EPLB tests continue to pass.
 

- vLLM version: v0.22.1


linfeng-yuan (approved):
LGTM. Please add a PP e2e case in test_config.yaml to maintain this function,

freyfwt:
lgtm

Angazenn (approved):

PR #10490, opened 2026-06-15
[Test]update upstream workflow

jiangyunfan1:
This PR updates the upstream workflow by adding hf online workflow and cache the compile of vllm-ascend, which improves the quality of the workflow.
This PR also removes some cases that no longer available on vllm 0.21.0
No
by running the test
- vLLM version: v0.22.1

### DIFF
--- a/.github/workflows/schedule_vllm_e2e_test.yaml
+++ b/.github/workflows/schedule_vllm_e2e_test.yaml
@@ -35,7 +35,7 @@ concurrency:
 
 env:
   UV_INDEX_URL: http://cache-service.nginx-pypi-cache.svc.cluster.local/pypi/simple
-  UV_EXTRA_INDEX_URL: https://repo.huaweicloud.com/ascend/repos/pypi
+  UV_EXTRA_INDEX_URL: "https://repo.huaweicloud.com/ascend/repos/pypi http://cache-service.nginx-pypi-cache.svc.cluster.local/whl/cpu/"
   UV_INDEX_STRATEGY: unsafe-best-match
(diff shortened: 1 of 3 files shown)


.github/workflows/schedule_vllm_e2e_test.yaml
```diff
@@ -99,19 +109,54 @@ jobs:
...
+           
+          - name: Install vllm-project/vllm-ascend with device
+            run: |
+              export MAX_JOBS=$(( ${{ matrix.num_npus }} * 23 ))
               pip install uc-manager
-              uv pip install torch==2.10.0 torchvision==0.25.0 torchaudio==2.10.0 --extra-index-url  http://cache-service.nginx-pypi-cache.svc.cluster.local/whl/cpu/
               uv pip install -r requirements-dev.txt
               uv pip install --force-reinstall --no-deps triton-ascend==3.2.1
-              uv pip install -e .
+              if find vllm_ascend -maxdepth 1 -name '*.so' -type f 2>/dev/null | grep -q .; then
```
Tflowers-0129, line 148: 是不是粒度太大，有so就认为cache hit吗

Tflowers-0129 (approved):

.github/workflows/schedule_vllm_e2e_test.yaml
tfhddd: Incremental compilation acceleration is currently not feasible. Therefore, the hash of the calculated C++-related code is used as the cache key. As long as the C++ code remains unchanged, the artifacts from the previous build are used, thus improving CI speed.

wangxiyuan (approved):

PR #10505, opened 2026-06-15
[CI] Extend PR-triggered nightly timeout to 240 minutes

MINGJING-WU:

PR-triggered nightly tests may need additional time to check out and install the PR code before running the selected test cases. The reusable nightly workflows currently either hard-code a 120-minute timeout or do not define a consistent job timeout, causing long-running PR nightly jobs to be cancelled before completion.

This PR applies the same request-aware timeout to all reusable nightly test workflows:

- use 240 minutes when `request_id` is non-empty, which identifies a PR `/nightly` dispatch;
- keep 120 minutes for scheduled nightly runs and regular manual dispatches without a `request_id`.

The change covers single-node, multi-node, and single-node model accuracy nightly tests. Image build and orchestration jobs are unchanged because they do not execute nightly test cases, and PR `/nightly` dispatches already skip image builds.

No. This only changes GitHub Actions timeout behavior for nightly CI jobs.

- Ran `actionlint` v1.7.12 against all three changed workflow files.
- Parsed all three workflow files with Ruby's YAML parser.
- Ran `git diff --check`.

- vLLM version: v0.22.1

### DIFF
--- a/.github/workflows/_e2e_nightly_multi_node.yaml
+++ b/.github/workflows/_e2e_nightly_multi_node.yaml
@@ -92,7 +92,9 @@ concurrency:
 jobs:
   e2e:
     name: ${{ inputs.config_file_path }}
-    timeout-minutes: 120
+    # PR-triggered nightly runs carry a request_id and use a 4-hour timeout;
+    # workflow-triggered nightly runs keep the default 2-hour timeout.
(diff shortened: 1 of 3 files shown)


weixinAc:
weekly下存在单用例配置多性能场景的情况，也会出现执行超时中断，需要适配更长的超时时间，能否兼容处理

PR #10510, opened 2026-06-15
[Feature]Support IndexCache top-k index reuse for GLM5.2

jiajinzhu2:
This PR aligns the Ascend backend with the upstream implementation of IndexCache top-k index reuse for DeepSeek Sparse MLA's MTP Layer, inspired by IndexCache: Accelerating Sparse Attention via Cross-Layer Index Reuse.{GLM5.2}
vllm PR: https://github.com/vllm-project/vllm/pull/44420/changes

No
vllm serve /home/jjz/mnt/GLM-5.2-0610-Provider
--host 0.0.0.0
--port 8077
--data-parallel-size 1
--tensor-parallel-size 16
--enable-expert-parallel
--seed 1024
--served-model-name glm-5
--max-num-seqs 16
--max-model-len 40960
--max-num-batched-tokens 4096
--trust-remote-code
--gpu-memory-utilization 0.95
--enable-chunked-prefill
--enable-prefix-caching
--additional-config '{"fuse_muls_add": true, "multistream_overlap_shared_expert": true, "ascend_compilation_config": {"enable_npugraph_ex": true}}'
--compilation-config '{"cudagraph_mode": "FULL_DECODE_ONLY"}'
--speculative-config '{"num_speculative_tokens": 3, "method": "deepseek_mtp"}'

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/spec_decode/a2/test_eagle_proposer.py
+++ b/tests/ut/spec_decode/a2/test_eagle_proposer.py
@@ -4382,3 +4382,120 @@ def test_split_pcp_input_hybrid_preserves_hidden_size():
     )
 
     assert out_hidden_states.shape[1] == hidden_size
+
+
+class TestDeepSeekMTPIndicesSharing(unittest.TestCase):
+    """
+    Unit tests for DeepSeek Sparse MLA MTP Layer's Top-K Index reuse feature (PR #10510)
+    """
+
+    def setUp(self):
+        # Prepare the base Config Mock
+        self.vllm_config = MagicMock(spec=VllmConfig)
+        self.vllm_config.speculative_config = MagicMock()
+        self.vllm_config.speculative_config.draft_model_config = MagicMock()
+        self.vllm_config.speculative_config.draft_model_config.hf_config = MagicMock()
+        self.device = torch.device("cpu")
+        self.runner = MagicMock()
+
+    def test_init_mtp_indices_flag(self):
+        """Test whether index_share_for_mtp_iteration is correctly read in __init__."""
+        # Scenario 1: Set to True in config
+        self.vllm_config.speculative_config.draft_model_config.hf_config.index_share_for_mtp_iteration = True
+
+        with patch.object(AscendEagleProposer, "__init__", lambda self, vllm_config, device, runner: None):
+            proposer = AscendEagleProposer(self.vllm_config, self.device, self.runner)
+            # Manually trigger the newly added initialization logic for validation
+            proposer.vllm_config = self.vllm_config
+            proposer._share_mtp_indices = getattr(
+                self.vllm_config.speculative_config.draft_model_config.hf_config, "index_share_for_mtp_iteration", False
+            )
+            self.assertTrue(proposer._share_mtp_indices, "MTP share flag should be True when configured.")
+
+        # Scenario 2: Not set in config (defaults to False)
+        del self.vllm_config.speculative_config.draft_model_config.hf_config.index_share_for_mtp_iteration
+        with patch.object(AscendEagleProposer, "__init__", lambda self, vllm_config, device, runner: None):
+            proposer2 = AscendEagleProposer(self.vllm_config, self.device, self.runner)
+            proposer2.vllm_config = self.vllm_config
+            proposer2._share_mtp_indices = getattr(
+                self.vllm_config.speculative_config.draft_model_config.hf_config, "index_share_for_mtp_iteration", False
+            )
+            self.assertFalse(proposer2._share_mtp_indices, "MTP share flag should default to False.")
+
+    def test_maybe_share_topk_indices_submodules(self):
+        """Test if _maybe_share_topk_indices correctly updates topk_indices_buffer for all submodules."""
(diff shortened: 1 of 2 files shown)


vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -1017,13 +1031,23 @@ def _run_merged_draft(
...
+            draft_model.set_skip_topk(False)
+
         ret_hidden_states = self.model(**model_kwargs)
         if not self.model_returns_tuple():
             last_hidden_states = ret_hidden_states
             hidden_states = last_hidden_states
         else:
             last_hidden_states, hidden_states = ret_hidden_states
 
+        # step 1+ skip indexer
```
YzTongNiar, line 1055: just wonder what is the usage of index cache for MTP since MTP usually has only one layer? But certainly, I understand that skiping the mtp-layer's indexer also has performance gain.

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -469,6 +476,12 @@ def _maybe_share_topk_indices(self, target_language_model: nn.Module) -> None:
                 "[spec_decode/base] Detected MTP model with topk_indices_buffer."
                 " Sharing target model topk_indices_buffer with the draft model."
             )
+            target_buffer = target_language_model.model.topk_indices_buffer
```
lilinsiman, line 478: Determine whether the logic of vllm can be reused.

YzTongNiar:
I jsut wonder, whether this feature affects MTP's acceptance rate? especially under some difficult datasets like GPQA? If it does have affects, please add a switch in 'speculative config' to allow disabling it.

lilinsiman:
1. Add the corresponding UT and modify the nightly test cases to achieve the maintenance effect.
2. Paste the upstream VLLM PR to the message.
3. Resolve the comment.

wu-yushan:
I have a couple of questions regarding the current PR:

1. Unlike the inter-layer reuse of the index in the base model, the context changes across different steps in MTP. Given this, is it reasonable to reuse the indexer's results under these conditions?
2. Assuming the first point is valid, do we only reuse the first step regardless of the number of speculative steps? This might be acceptable when MTP=3, but would reusing just the first layer still have a significant impact when MTP=7?

jiajinzhu2:
Thanks for the thoughtful review! 
This PR adapts the optimization from the upstream vLLM community PR (https://github.com/vllm-project/vllm/pull/44420/changes) and applies it to the vLLM Ascend backend with GLM model adaptations.
I will try to answer your question.

This paper (https://arxiv.org/abs/2603.12201) demonstrates through experiments that the top-k indices of adjacent layers have a high degree of overlap (70% to 100%), indicating that most of the indexer computations are redundant. Context changes are incremental—each MTP step adds only 1 token, so reusing the indexer's results provides efficiency gains with minimal approximation.

vllm_ascend/spec_decode/llm_base_proposer.py
jiajinzhu2: When top-k indices selection can be reused, the sparse attention indexer is avoided from running in each layer or **each MTP inference step**.

vllm_ascend/spec_decode/llm_base_proposer.py
jiajinzhu2: I don't think we can just reuse vllm's logic.
This modification is situated at the conclusion of the `def load_eagle_model()` function within the file `vllm/v1/worker/gpu/spec_decode/eagle/utils.py`. The `load_eagle_model()` function is meant to fully load an EAGLE draft model and hand it back to you. If you just call load_eagle_model directly, the model ends up loading over and over again.

jiajinzhu2:
This function is enabled only when the 'index_share_for_mtp_iteration configuration' exists in the config.json file. It is used for adaptation to specific models.

vllm_ascend/spec_decode/llm_base_proposer.py
lilinsiman: OK, fine

tests/ut/spec_decode/a2/test_eagle_proposer.py
```diff
@@ -4214,3 +4170,127 @@ def test_set_inputs_first_pass_draft_model(self):
 
         for attrition in attrs_from_cad:
             assert_attr_equal(attrition, expected_cad, out_cad)
+
+class TestDeepSeekMTPIndicesSharing(unittest.TestCase):
```
lilinsiman, line 4387: Try to add nightly for this feature

jiajinzhu2: The quantized version of the new model with this feature hasn't been released on ModelScope yet, so we can't add nightly for this feature.

lilinsiman: OK, fine

lilinsiman (approved):

zzzzwwjj (approved):

PR #10511, opened 2026-06-15
"[Ascend950] [BugFix]Fix the issue where MOE model output is garbled when communicating with allgether while SP is disabled.“

Eric-dot:
This PR fixes a garbled output issue for MoE models when using the ALLGATHER communication path without sequence parallelism enabled.

In the non-SP MoE scenario, the final MoE output still needs to be reduced across TP/EP ranks. However, when running in graph mode, the communication decision may be specialized during graph compilation, especially when switching between different MoE communication paths such as MC2 and ALLGATHER. As a result, the required final all-reduce for the ALLGATHER path can be skipped, which leads to incorrect MoE output and eventually garbled model responses.

To fix this, this PR moves the final MoE output reduction decision into the graph/runtime path, so the correct reduce behavior is preserved for non-sequence-parallel configurations. This ensures that MoE models produce correct results when ALLGATHER is selected while SP is disabled.

MXFP8 Deepseekv3.1 qpga AISBenchtest:
| Dataset      | Version | Metric   | Mode | vLLM API General Chat |
|--------------|---------|----------|------|------------------------|
| GPQA_diamond | 53064e  | accuracy | gen  | 76.77                 |

- vLLM version: v0.22.1


vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -293,6 +293,25 @@ def _maybe_reduce_shared_expert_output(
...
+
+        The upstream implementation branches in Python on
+        ``_fused_output_is_reduced``. On Ascend, that value depends on
+        ``_EXTRA_CTX.moe_comm_type`` and may switch between MC2 and ALLGATHER
+        for different token counts. In compiled graph mode the Python branch can
+        be specialized from an earlier MC2 trace, dropping the late all-reduce
+        needed by ALLGATHER. Keep a custom op in the graph and let its runtime
+        implementation decide whether to all-reduce for the current context.
+        """
+        if not enable_sp() and (self.moe_config.tp_size > 1 or self.moe_config.ep_size > 1):
```
zzzzwwjj: why we don't need to judge `self._fused_output_is_reduced` here?

PR #10518, opened 2026-06-16
[Attention] [BugFix]Enable multistream_dsv4_dsa_overlap and remove redundant code

Shelleyaaa:

1. This PR fixes cv_parallel bug.
2. This PR removes the dedicated `prefill_comm_compute_overlap` special path in DSA v1 and keeps prefill on the unified FlashComm gather flow.
3. This PR removes `multistream_dsa_preprocess` which is useless.
4. This PR removes `dsa_warmup_with_multistream` in dsa.py and its implementation in dsa_v1.py.

No. This is an internal correctness and robustness fix for DSA v1 quantization/indexer execution paths.

- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -140,10 +140,7 @@ class AscendConfig:
         # PD-disaggregated only (kv_producer/kv_consumer); invalid in PD-mixed (kv_both / no kv_transfer_config).
         self.recompute_scheduler_enable = additional_config.get("recompute_scheduler_enable", False)
         self.enable_cpu_binding = additional_config.get("enable_cpu_binding", True)
-        self.multistream_dsa_preprocess = additional_config.get("multistream_dsa_preprocess", False)
-        self.multistream_dsv4_dsa_overlap = additional_config.get("multistream_dsv4_dsa_overlap", False)
-        self.prefill_comm_compute_overlap = additional_config.get("prefill_comm_compute_overlap", False)
-
+        self.multistream_dsv4_dsa_overlap = additional_config.get("multistream_dsv4_dsa_overlap", True)
         self.enable_matmul_allreduce = self._get_config_value(
             additional_config,
             "enable_matmul_allreduce",
--- a/vllm_ascend/attention/dsa_v1.py
+++ b/vllm_ascend/attention/dsa_v1.py
@@ -1841,7 +1841,14 @@ class AscendDSAImpl(DSAAttentionImpl):
                     bias=self.wq_b.bias,
                     output_dtype=hidden_states.dtype,
                 ).unflatten(-1, (self.n_local_heads, self.head_dim))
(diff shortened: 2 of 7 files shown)


pisceskkk (approved):
LGTM

vllm_ascend/ascend_config.py
```diff
@@ -140,10 +140,7 @@ def __init__(self, vllm_config: "VllmConfig"):
         # PD-disaggregated only (kv_producer/kv_consumer); invalid in PD-mixed (kv_both / no kv_transfer_config).
         self.recompute_scheduler_enable = additional_config.get("recompute_scheduler_enable", False)
         self.enable_cpu_binding = additional_config.get("enable_cpu_binding", True)
-        self.multistream_dsa_preprocess = additional_config.get("multistream_dsa_preprocess", False)
-        self.multistream_dsv4_dsa_overlap = additional_config.get("multistream_dsv4_dsa_overlap", False)
-        self.prefill_comm_compute_overlap = additional_config.get("prefill_comm_compute_overlap", False)
-
+        self.multistream_dsv4_dsa_overlap = additional_config.get("multistream_dsv4_dsa_overlap", True)
```
MengqingCao, line 143: we should add instruction for these configs in https://docs.vllm.ai/projects/ascend/en/latest/user_guide/configuration/additional_config.html

vllm_ascend/attention/dsa_v1.py
```diff
@@ -1484,51 +1481,6 @@ def __init__(
             False,
         )
 
-    def dsa_warmup_with_multistream(self, hidden_states: torch.Tensor) -> None:
```
MengqingCao, line 1487: could you explain more on why we need the warm up before and don't need it now?

vllm_ascend/attention/dsa_v1.py
MengqingCao, line 1758: Now we just use this multistream, and no need to handle a new stream, right?

vllm_ascend/attention/dsa_v1.py
Shelleyaaa: When using PTA 2.9.0 for inference, graph capture fails in pure DP scenarios. The operators on the auxiliary stream must be warmed up again for inference to run correctly. However, in the current main version, PTA has already been upgraded, so this warmup is no longer required and inference can run normally.

vllm_ascend/ascend_config.py
Shelleyaaa: Sure, I will add the instructions for this config in a follow-up update.

vllm_ascend/attention/dsa_v1.py
Shelleyaaa: Yes. We use multistream by default now, but the serial path is still available. Setting `multistream_dsv4_dsa_overlap` to `false` will fall back to the original serial logic.

MengqingCao (approved):
LGTM now, thx for this clean up!

PR #10523, opened 2026-06-16
[BugFix] fix w4a8 weight loading of DSACP in sfa_v1.py

pisceskkk:
This PR refactors the parameter switching mechanism for `o_proj` in the Ascend SFA attention backend. Instead of hardcoding specific parameters like `aclnn_input_scale`, it introduces configurable parameter groups (`O_PROJ_ACLNN_INPUT_PARAMS` and `O_PROJ_INPUT_SHARDED_QUANT_PARAMS`) and a helper method `_switch_o_proj_params` to dynamically switch between TP-mode and Full-mode.

Feedback:
- There is a potential correctness issue when gathering multi-dimensional sharded quantization parameters (e.g., 2-D tensors), as the current shape allocation and gathering assume dimension 0 is the sharded dimension, which might not be correct if `input_dim == 1`.

No.

GSM8K accuracy test for glm5-w4a8 weight:

| gsm8k | 50861a | accuracy | gen | 93.25 |

Wait for MXFP8's accuracy test.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/attention/sfa_v1.py
+++ b/vllm_ascend/attention/sfa_v1.py
@@ -911,7 +911,7 @@ class AscendSFAImpl(MLAAttentionImpl):
         attn_output: torch.Tensor,
         output: torch.Tensor,
         o_proj_full_handle: torch.distributed.Work | None,
-        o_proj_full_weight_scale_handle: torch.distributed.Work | None,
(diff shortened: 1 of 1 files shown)


MengqingCao (approved):
LGTM

PR #10548, opened 2026-06-16
[Performance][PD] Parallelize KV cache receive via ThreadPoolExecutor

jianzs:
## Summary

This PR improves KV cache receive throughput in PD disaggregation by running independent receive work concurrently while preserving FIFO ordering for the same remote peer.

The original `KVCacheRecvingThread` processed every transfer task sequentially:

```python
request_data = self.request_queue.get()
self._handle_request(request_data)
```

That makes unrelated transfers block behind each other. Under concurrent PD traffic, tasks from different prefill ranks can queue behind a slow transfer even though they target independent remote peers and can safely progress in parallel.

The final implementation dispatches through a per-peer FIFO scheduler:

- requests for the same `(remote_host, remote_handshake_port)` are serialized;
- requests for different peers can run concurrently in the existing `ThreadPoolExecutor(max_workers=32)`;
- a busy peer yields after a small batch so it cannot monopolize executor workers;
- request completion accounting waits for all submitted transfer tasks before marking the request done.

## Main changes

1. **Parallel KV receive with per-peer FIFO**

   `KVCacheRecvingThread.run()` now records a request as submitted and queues it by remote peer. Each active peer has at most one handler running, so REQ/REP ordering for that peer is preserved while independent peers can transfer concurrently.

2. **Bounded peer handler batch**

   `MAX_REQUESTS_PER_PEER_HANDLER = 5` prevents one busy peer from holding an executor worker forever when other peers are already waiting.

3. **Thread-safe shared state**

   - removes the shared `zmq.Poller` receive path and uses per-socket `RCVTIMEO`;
   - protects `proc_not_transfer_request` with a lock;
   - closes and discards failed REQ sockets instead of returning invalid sockets to the pool;
   - protects remote metadata cache population/readout with `remote_metadata_lock`;
   - tracks pending transfer tasks per request before calling `task_tracker.update_done_task_count()`.

## Latest E2E validation

Test date: 2026-06-29.

Setup:

- 4-node DeepSeek V4 PD deployment, 32 NPUs total
- Prefill: DP4TP8
- Decode: TP4DP8
- Requests sent from `dev0` to `localhost:9000`; proxy forwards to prefill/decode
- Workload: 500 requests, max concurrency 48, input length 256
- Baseline: PR changes fully reverted to merge-base for comparison
- Current: this PR head
- Timing data was collected with temporary measurement scaffolding during validation; that scaffolding is not included in the final patch.

### Request-level result

| Metric | Baseline | Current | Change |
|---|---:|---:|---:|
| OK | 500/500 | 500/500 | no errors |
| Total time | 12.6s | 12.5s | -0.8% |
| Request p50 | 0.96s | 0.97s | roughly flat |
| Request p99 | 3.44s | 3.10s | -9.9% |

### KV receive timing

Aggregate across decode nodes, 2000 KV receive tasks per run:

| Metric | Baseline | Current | Change |
|---|---:|---:|---:|
| queue p50 | 18.14ms | 0.70ms | improved |
| queue p90 | 237.79ms | 1.87ms | improved |
| queue p99 | 603.63ms | 2.64ms | -99.56% |
| lifecycle p99 | 735.16ms | 512.80ms | -30.25% |
| active handler max | 1 | 7 | parallel receive active |

Per-node queue p99:

| Decode node | Baseline | Current | Change |
|---|---:|---:|---:|
| dev2 | 602.23ms | 4.11ms | -99.32% |
| dev3 | 595.88ms | 2.41ms | -99.60% |

The queue tail result is consistent with the earlier validation data: queue p99 drops from hundreds of milliseconds to single-digit milliseconds, showing that the tail was caused by serialized receive handling.

## How was this patch tested?

- `pre-commit run ruff-check --all-files`
- `pre-commit run ruff-format --all-files`
- `git diff --check`
- `python3 -m py_compile vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_hybrid_connector.py`
- `python3 -m pytest tests/ut/kv_offload/test_mooncake_connector.py tests/ut/kv_offload/test_mooncake_hybrid_connector.py -q`
  - Result: 86 passed, 16 warnings
- Full 4-node E2E restart and proxy benchmark:
  - baseline: 500/500 OK
  - current: 500/500 OK
- Log scan after E2E found no `ZMQError`, `EFSM`, `Traceback`, duplicate done signal, request loss, or address conflict errors.

No. This changes internal KV receive scheduling and synchronization only.

- vLLM version: v0.23.0


LCAIZJ (approved):
LGTM

wangxiaoteng888:
This PR might cause precision issues when Tensor Parallelism (TP) is unbalanced/asymmetrical, because multi-threaded transmission can lead to incorrect reordering timing. For the fix, please see the PR at https://github.com/vllm-project/vllm-ascend/pull/12359

PR #10564, opened 2026-06-16
[CI][Feature] Add nightly auto-bisect tool for E2E test failures

DreamerLeader:

This PR introduces an automated bisect tool (`auto_bisect`) for nightly E2E test failures in `vllm-ascend`. It binary-searches the commit history between the last known-good commit (tracked in a CSV table) and the failing commit to identify the first bad commit and its associated PR. It supports both single-node and multi-node environments, reuses existing nightly launch entries, and optimizes the build process by only rebuilding when native or build-definition files are modified.

No. This is a developer/CI tool and does not introduce any user-facing changes.

Tested locally by running the auto-bisect tool on single-node and multi-node scenarios.

- vLLM version: v0.23.0


tools/bisect/__init__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
```
wangxiyuan, line 1: let's move this tool to `tools/bisect` folder.

linfeng-yuan (approved):

tools/bisect/__init__.py
DreamerLeader: already fixed

PR #10566, opened 2026-06-16
[SpecDecode][BugFix] Support Eagle3 with Chunked Pipeline Parallelism

845473182:
This pull request introduces support for propagating Eagle3 auxiliary hidden states through the Pipeline Parallelism (PP) pipeline, specifically targeting `DeepseekV2Model`. It wraps the model's forward pass and intermediate tensor allocation to pass auxiliary states across PP stages transparently. Additionally, it addresses scheduling and speculative decoding issues, such as handling empty scheduled tokens and ensuring draft token IDs are correctly retrieved on NPU.

No, this is an internal enablement and bug fix for Eagle3 speculative decoding with Pipeline Parallelism.

The patch includes unit tests in `tests/ut/patch/worker/test_patch_eagle3_pp_aux.py` to verify auxiliary state propagation and intermediate tensor allocation.

Co-authored-by: Jingchun Gao [gaojingchun1@huawei.com](mailto:gaojingchun1@huawei.com)

- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/spec_decode/llm_base_proposer.py
+++ b/vllm_ascend/spec_decode/llm_base_proposer.py
@@ -502,7 +502,6 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
         batch_descriptor=None,
         dummy_compute_logits=lambda hidden_states: None,
         is_profile=False,
-        profile_cpp=False,
     ):
         (
             num_tokens,
@@ -590,8 +589,6 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
         # TODO: temporarily hack here, we should find out batch_size for profile_run
         if is_profile:
             batch_size = min(batch_size, self.runner.max_num_reqs)
-        elif profile_cpp:
-            batch_size = 1
 
         if self.supports_mm_inputs:
             mm_embeds, is_mm_embed = (None, None)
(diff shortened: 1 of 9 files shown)


vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -587,6 +588,8 @@ def dummy_run(
         # TODO: temporarily hack here, we should find out batch_size for profile_run
         if is_profile:
             batch_size = min(batch_size, self.runner.max_num_reqs)
+        elif profile_cpp:
+            batch_size = 1
```
lilinsiman: CPP in dummyrun causes OOM here, why can setting the batch size to 1 avoid it, and are there better solutions? It is best to avoid hardcoding.

845473182: In normal dummy runs, num_tokens roughly means `batch_size * (1 + num_speculative_tokens)`. So drafter can infer batch_size from it.
In profile_cpp, num_tokens means a request with a long prefill chunk. If drafter uses the same formula, it treats one long request as many request, allocates huge draft-side buffers, and OOMs.
Here we either set the `batch_size = 1` or we can just simply skip drafter's dummy run in profile_cpp.

realliujiaxu (approved):
LGTM

linfeng-yuan (approved):

PR #10581, opened 2026-06-17
[DSv4][MTP] Enable MTP drafter cudagraph capture under DSA (graph-mode MTP without enforce_eager)

wangzhao-11a:
## Summary

DeepSeek-V4 (Flash/Pro) MTP speculative decoding crashes under cudagraph (`--compilation-config '{"cudagraph_mode":"FULL_DECODE_ONLY"}'`) because the **drafter graph-capture path** in `llm_base_proposer.dummy_run` does not populate the DSA metadata that `dsa_v1` requires:

- `dsa_v1.build_decode_metadata` indexes `common_attn_metadata.positions_cpu`, but the drafter's `AscendCommonAttentionMetadata` in the graph-capture branch never sets `positions_cpu` -> `None` indexing crash.
- `dsa_v1.build` asserts `prefill/decode/common_ratio_to_sas_metadata is not None`, but the drafter graph-capture path passes none of them -> empty `AssertionError`.

The main-model path already handles both (`model_runner_v1` sets `positions_cpu=_dsa_positions_cpu_buf` when `use_compress`, and the non-graph `_propose()` builds `extra_attn_metadata_args`); only the drafter graph-capture branch was missing them. As a consequence DeepSeek-V4 MTP could previously only run with speculative `enforce_eager`.

## Fix

Mirror the non-graph `_propose()` path inside the graph-capture branch of `dummy_run`:

- pass `positions_cpu = runner._dsa_positions_cpu_buf` when `use_compress`;
- build empty `prefill/decode/common_ratio_to_sas_metadata` + `block_size` and forward them into `build_for_graph_capture` (passed through to `build` via `**kwargs`).

This lets the MTP drafter complete cudagraph capture, so DeepSeek-V4 MTP runs in graph mode without `enforce_eager`. The change is confined to `spec_decode/llm_base_proposer.py` (14 lines); model layers are unchanged, so the fix is model-agnostic and covers both DeepSeek-V4 Flash and Pro.

## Testing

- `python -m py_compile` and `import vllm_ascend` pass.
- Serve config: `--speculative-config '{"method":"mtp","num_speculative_tokens":1}'` (no `enforce_eager`) + `--compilation-config '{"cudagraph_mode":"FULL_DECODE_ONLY"}'`.
- End-to-end serve validation (graph capture 8/8, acceptance rate, accuracy) in progress.

### DIFF
--- a/vllm_ascend/attention/context_parallel/dsa_cp.py
+++ b/vllm_ascend/attention/context_parallel/dsa_cp.py
@@ -873,6 +873,21 @@ class AscendDSACPImpl(DSAAttentionImpl):
     understand this class
     """
 
+    @staticmethod
+    def update_graph_params(
+        update_stream,
+        forward_context,
+        num_tokens,
+        vllm_config=None,
+        speculative_config=None,
+        num_dcp_pcp_tokens=None,
+        draft_attn_metadatas=None,
+    ):
+        # MTP graph-mode fix: DSA-CP does not need per-step graph param updates
+        # (same as SFA). Required so the generic full-graph drafter path does
+        # not raise AttributeError under FULL_DECODE_ONLY + MTP.
+        pass
+
     def __init__(
         self,
         n_heads: int,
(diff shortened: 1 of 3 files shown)


nwpu-zxr:
I do not believe this can properly enable MTP FullGraph. It is necessary to fix the addresses of some variables in `dsa_metadata`, such as `sas_metadata`, etc. Please conduct a thorough check.

PR #10586, opened 2026-06-17
[Feature] [BugFix] Support fine-grained Tensor Parallelism (OTP and Embedding TP) in DSA

xmpp777:
Enable fine-grained tensor parallelism on the o_proj (wo_a/wo_b) weights of DeepSeek-V4's DSA (DeepSeek Attention) module via static-buffer all_to_all/reduce_scatter across the OTP group, and rewrite the embedding TP path to use graph-stable static raw-dist buffers so both run correctly under ACL graph capture/replay.

#### Key changes

- Fine-grained Tensor Parallelism for O-Proj: Implemented fine-grained TP for DSA o_proj (wo_a/wo_b) via the OTP group using graph-stable static send/recv buffers padded to the max capture size, with dist.all_to_all_single + raw dist.reduce_scatter_tensor into an address-stable output buffer for efficient, ACL-graph-safe distribution.
- Fine-grained Tensor Parallelism for Embedding: Rewrote input-embedding TP to use static all_gather_into_tensor / reduce_scatter_tensor buffers (allocated once to the max capture capacity), improving performance and fixing the accuracy regression caused by the previous list-based wrappers.
- Recompute-Scheduler Binding for DP Tokens: Bound oproj/embedding TP to recompute_scheduler_enable and forced allow_dp_padding under oproj TP, so cross-DP num_tokens stays uniform and the collectives don't deadlock — enforced as a config-time assertion rather than a runtime hang.

Enables oproj_tensor_parallel_size support in the DSA attention backend and makes the embedding_tensor_parallel_size path correct and faster under ACL graph mode — using graph-stable static exchange buffers rather than per-call dynamic allocation. o_proj (wo_a/wo_b) is sharded across the OTP group via raw dist.all_to_all_single + dist.reduce_scatter_tensor into static buffers, with new DSV4OProjColumnParallelOp / DSV4OProjRowParallelOp routing wo through the OTP group; embedding TP is rewritten with static all_gather_into_tensor / reduce_scatter_tensor buffers. Because both paths' cross-DP collectives require uniform num_tokens across DP ranks, the features are bound to recompute_scheduler_enable and enforced at config load instead of deadlocking at runtime.

Yes. It enables tensor-parallel support for the o_proj operator in the DSA backend (currently constrained to tensor_parallel_size == 1) and makes embedding TP correct and faster under ACL graph mode. Both oproj_tensor_parallel_size and embedding_tensor_parallel_size now require recompute_scheduler_enable = true; enabling either without recompute raises a clear config-time error instead of a runtime collective deadlock.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -139,6 +139,25 @@ class AscendConfig:
         self.multistream_overlap_gate = additional_config.get("multistream_overlap_gate", False)
         # PD-disaggregated only (kv_producer/kv_consumer); invalid in PD-mixed (kv_both / no kv_transfer_config).
         self.recompute_scheduler_enable = additional_config.get("recompute_scheduler_enable", False)
+        # DSV4 oproj / embedding fine-grained TP (oproj_tensor_parallel_size /
+        # embedding_tensor_parallel_size) use static, graph-stable exchange
+        # buffers and run cross-DP HCCL collectives (all_to_all / all_gather /
+        # reduce_scatter) that require uniform num_tokens across all DP ranks.
+        # Only the recompute scheduler balances num_tokens across DP ranks; the
+        # MC2 uneven-token skip path leaves each rank at its own num_tokens and
+        # would deadlock these collectives on a shape mismatch. So bind both
+        # features to recompute_scheduler_enable: they are only supported when
+        # recompute is on.
+        if (
+            self.finegrained_tp_config.oproj_tensor_parallel_size > 0
+            or self.finegrained_tp_config.embedding_tensor_parallel_size > 0
+        ) and not self.recompute_scheduler_enable:
+            raise AssertionError(
+                "oproj_tensor_parallel_size / embedding_tensor_parallel_size "
+                "require recompute_scheduler_enable=true: their cross-DP HCCL "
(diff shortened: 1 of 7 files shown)


maoxx241 (approved):

xmpp777:
### Verification Results

| Item | Result |
| --- | --- |
| Branch status | Rebased onto latest `main`; resolved `utils.py` conflict by keeping both `is_pd_decode_recompute_scheduler_enabled` (upstream) and the `_compute_potential_max_tokens` family (this PR); fixed import order for the pre-commit/isort hook; added `Signed-off-by` for DCO |
| GPQA — oproj + embedding TP (graph mode) | Non-thinking mode, **> 71** across 10 runs |
| GPQA — embedding TP + eager mode | Non-thinking mode, **> 71** across 10 runs |
| GPQA — oproj + embedding TP + eager mode | Errors out as designed (eager mode does not support oproj TP) |
| Memory (enabled vs. disabled) | **~76.8 GB** saved in total; Loading model weights: 23.2812 GB (after) vs. 28.0937 GB (before) |

GDzhu01 (approved):

linfeng-yuan (approved):

Issue #10403, opened 2026-06-12
[RFC]: PD Disaggregation Cached Tokens Semantics


lio1226:
### Motivation.

Public cloud and hosted model APIs commonly price cached input tokens differently from non-cached input tokens. Therefore, returning the wrong `cached_tokens` value can make cost attribution, customer chargeback, and cache hit-rate observability materially wrong.

Price snapshot from official pages on 2026-06-12:

| Provider/API | Example model | Non-cached input | Cached input / cache read | Notes |
| --- | --- | ---: | ---: | --- |
| OpenAI API | GPT-5.5 | $5.00 / 1M tokens | $0.50 / 1M tokens | Cached input is 10% of normal input price. |
| OpenAI API | GPT-5.4 | $2.50 / 1M tokens | $0.25 / 1M tokens | Same 10% ratio. |
| DeepSeek API | deepseek-v4-flash | $0.14 / 1M cache-miss input tokens | $0.0028 / 1M cache-hit input tokens | Cache-hit input is 2% of cache-miss input price. |

In vLLM, we can add the --enable-prompt-tokens-details flag to return the cache hit count for one request. And the number of hit tokens is a multiple of the block size.

```bash
{"prompt_tokens":3678,"total_tokens":4614,"completion_tokens":936,"prompt_tokens_details":{"cached_tokens":3072}}} 
```

But in PD disaggregation, the external OpenAI-compatible response currently can expose `usage.prompt_tokens_details.cached_tokens` from the D node. This is misleading because the request input is prefilling on the P node, and the P node is the authoritative place where prefix-cache hit or miss is decided.

```bash
Prefill usage: 
{"prompt_tokens":3678,"total_tokens":3679,"completion_tokens":1,"prompt_tokens_details":{"cached_tokens":3072}}}
Decode usage:
{'prompt_tokens': 3678, 'total_tokens': 4614, 'completion_tokens': 936, 'prompt_tokens_details': {'cached_tokens': 3677}}
proxy "usage":
{'prompt_tokens': 3678, 'total_tokens': 4614, 'completion_tokens': 936, 'prompt_tokens_details': {'cached_tokens': 3677}}
```

This RFC proposes that the PD proxy rewrites the user-facing `cached_tokens` value to the P-node prefill cache-hit token count. D-node cache tokens must not be used as the user-facing input cache-hit metric.

### Proposed Change.

[...]

Issue #8910, opened 2026-05-06
[Bug]: vllm-ascend 018.0 pd 分离部署 cached token 数量永远比 prompted 小1


kdshuo:
### Your current environment

<details>
<summary>
使用 vllm-ascend 0.18.0 A2单机跑 glm5-w4a8

拉起：
--enable-chunked-prefill \
--enable-prefix-caching \
--enable-prompt-tokens-details \
请求：
"stream": true,
"stream_options": {"include_usage": true}
</summary>

</details>

### 🐛 Describe the bug

```text
在单机混布场景下，没复现小1（部分block命中）
在pd分离场景下(eg:A2 vllm-ascend 0.18.rc1 glm5.1 w8a8 1*4p+1*4d)，cached token 数量永远比 prompted 小1



```

yiz-liu:
@liziyu179 Is this a known issue? I vaguely remember that we’ve discussed it before.

liziyu179:
in vllm\vllm\v1\core\kv_cache_manager.py function get_computed_blocks set max_cache_hit_length = request.num_tokens - 1

postwter:
```shell
Prefill usage: 
{'prompt_tokens': 4609, 'total_tokens': 4610, 'completion_tokens': 1, 'prompt_tokens_details': {'cached_tokens': 3072}, 'completion_tokens_details': None}
Decode usage:
{'prompt_tokens': 4609, 'total_tokens': 4614, 'completion_tokens': 5, 'prompt_tokens_details': {'cached_tokens': 4608}}
the correct "usage":
{"prompt_tokens":4609,"total_tokens":4614,"completion_tokens":5,"prompt_tokens_details":{"cached_tokens":3072}}} 
``` 

the precise `prompt_tokens_details` is generated from the ‌Prefill‌ response, but when using the proxy, the final `usage` is returned‌ directly from the ‌last chunk of the Decoder‌ without recomputation. This leads to ‌the an inaccurate `cached_token` result (which should be an integer multiple of the block_size)

the layerwise proxy use #9574

yiz-liu:
@liziyu179 Looks like a metrics issue, can you confirm it again?

liziyu179:
The vllm proxy only returns response information from node D, not from node P. Manual querying is required if needed.

PR #10598, opened 2026-06-17
[BugFix][Misc] Report prefiller cached tokens in PD proxy

lio1226:

This PR fixes the cached token usage reported by the disaggregated prefill load-balance proxy, as mentioned in #8910 and #10403.

When the proxy sends a request to the prefiller, it now extracts `usage.prompt_tokens_details.cached_tokens` from the prefiller response and keeps it in the selected `InstanceInfo`. When forwarding decoder responses back to the client, the proxy writes that prefiller-side cached token count into response chunks that contain `usage`, while preserving both SSE streaming chunks and non-streaming JSON responses.

This is needed because in PD disaggregated serving the decoder response does not have enough context to report the P node's prefix-cache hits. As a result, clients of the proxy could observe missing or incorrect `cached_tokens` in `usage.prompt_tokens_details`, especially in streaming responses. The reported value should reflect the prefiller's cache-hit tokens.

For example, before this PR the proxy could pass through the decoder-side `cached_tokens` even though the prefiller had a different cache-hit count:

```text
Prefill usage:
{'prompt_tokens': 8193, 'total_tokens': 8194, 'completion_tokens': 1, 'prompt_tokens_details': {'cached_tokens': 6144}, 'completion_tokens_details': None}

decode usage chunk:
{'prompt_tokens': 8193, 'total_tokens': 8201, 'completion_tokens': 8, 'prompt_tokens_details': {'cached_tokens': 8192}}

Final usage chunk before this PR:
{'prompt_tokens': 8193, 'total_tokens': 8201, 'completion_tokens': 8, 'prompt_tokens_details': {'cached_tokens': 8192}}

Correct final usage chunk after this PR:
{"prompt_tokens":8193,"total_tokens":8201,"completion_tokens":8,"prompt_tokens_details":{"cached_tokens":6144}}
```

The change also keeps the cached token value up to date when the proxy retries and reassigns prefiller/decoder instances.

Yes. For users of `examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py`, responses that include `usage` now report `usage.prompt_tokens_details.cached_tokens` from the prefiller side of the PD request.

This makes the OpenAI-compatible usage metadata more accurate for disaggregated prefill proxy requests. No new API fields are introduced.

Syntax check:

```bash
python3 -c "from pathlib import Path; p=Path('examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py'); compile(p.read_text(), str(p), 'exec')"
```

`ruff` was not run locally because the `ruff` command is not available in the current environment:

```bash
ruff check examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
```

No new unit test was added because this change is limited to the example PD proxy path and depends on prefiller/decoder backend response integration. The expected behavior can be verified manually by running the disaggregated prefill example with prefix caching enabled and checking that the proxy response includes the prefiller-reported `usage.prompt_tokens_details.cached_tokens` for both streaming and non-streaming requests.

- vLLM version: v0.22.1

### DIFF
--- a/examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
+++ b/examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
@@ -161,9 +161,37 @@ class InstanceInfo:
     decoder_score: float
     decoder_host: str
     decoder_port: int
+    prefiller_cached_tokens: int | None = None
 
 
 TAINT_PRIORITY = 1e15
+
+
+def extract_cached_tokens(response_json: dict) -> int | None:
+    usage = response_json.get("usage") or {}
+    prompt_tokens_details = usage.get("prompt_tokens_details") or {}
+    cached_tokens = prompt_tokens_details.get("cached_tokens")
+    return cached_tokens if isinstance(cached_tokens, int) else None
+
+
+def update_cached_tokens_in_chunk(chunk_json: dict, cached_tokens: int | None) -> bool:
+    if cached_tokens is None:
+        return False
+    usage = chunk_json.get("usage")
+    if not isinstance(usage, dict):
+        return False
+    prompt_tokens_details = usage.get("prompt_tokens_details")
+    if not isinstance(prompt_tokens_details, dict):
+        prompt_tokens_details = {}
+        usage["prompt_tokens_details"] = prompt_tokens_details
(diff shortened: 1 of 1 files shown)


weijinqian0:
Ref: https://github.com/vllm-project/vllm-ascend/pull/11630/changes

PR #10602, opened 2026-06-17
[BugFix] Add compatibility guard for unsupported CLI params on Ascend NPU

underfituu:
## What this PR does

Add silent reset handling in `_fix_incompatible_config` for four upstream CLI parameters that are incompatible with Ascend NPU:

| Parameter | Config Section | Reset Value | Reason |
|-----------|---------------|-------------|--------|
| `--calculate-kv-scales` | Cache Config | `False` | Deprecated upstream; not supported on Ascend |
| `--compilation-config.use_inductor_graph_partition` | Compilation Config | `False` | Ascend forces `use_inductor=False`, making this irrelevant |

## User-facing change

Users who pass `--calculate-kv-scales`, `--compilation-config '{"use_inductor_graph_partition": true}'` on Ascend NPU will now see a warning log and the parameter will be silently reset to its safe default, rather than potentially causing errors or undefined behavior.

## How was this patch tested

- Code review of the `_fix_incompatible_config` method to verify the pattern matches existing silent-reset entries.

- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -1205,7 +1205,7 @@ class NPUPlatform(Platform):
             if getattr(vllm_config.compilation_config, "use_inductor_graph_partition", False) is not None:
                 logger.warning(
                     "Parameter is not supported on Ascend NPU (use_inductor is False). "
-                    "parameter=use_inductor_graph_partition, action: resetting to None."
+                    "parameter=use_inductor_graph_partition, action: resetting to False."
                 )
                 vllm_config.compilation_config.use_inductor_graph_partition = False
 
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -1018,6 +1018,13 @@ class NPUPlatform(Platform):
                 )
                 vllm_config.cache_config.cpu_kvcache_space_bytes = None
 
+            if getattr(vllm_config.cache_config, "calculate_kv_scales", False):
+                logger.warning(
+                    "Parameter is not supported on Ascend NPU. "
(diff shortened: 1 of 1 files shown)


vllm_ascend/platform.py
```diff
@@ -1205,6 +1200,34 @@ def _fix_incompatible_config(vllm_config: VllmConfig) -> None:
...
+                    "action: resetting to False."
+                )
+                vllm_config.parallel_config.enable_elastic_ep = False
+
+        # ==================== 10. Compilation Config ====================
+        if vllm_config.compilation_config:
+            if getattr(vllm_config.compilation_config, "use_inductor_graph_partition", False) is not None:
+                logger.warning(
+                    "Parameter is not supported on Ascend NPU (use_inductor is False). "
+                    "parameter=use_inductor_graph_partition, action: resetting to None."
```
Tflowers-0129: 这里是不是写错了，应该是False吧

underfituu: 已改正

vllm_ascend/platform.py
```diff
@@ -1205,6 +1200,15 @@ def _fix_incompatible_config(vllm_config: VllmConfig) -> None:
                 )
                 vllm_config.parallel_config.ubatch_size = 0
 
+        # ==================== 10. Compilation Config ====================
+        if vllm_config.compilation_config:
+            if getattr(vllm_config.compilation_config, "use_inductor_graph_partition", False) is not None:
```
realliujiaxu: why `is not None` here?

zzzzwwjj (approved):

PR #10694, opened 2026-06-18
[Feature][Performance] Support o-proj TP weights async allgather in DSA-CP

TmacAaron:
This PR adds the missing DSV4 DSA-CP o-proj TP full-weight allgather path for prefill on top of the DSV4 DSA-CP A5 support merged by #10026.

When DSA-CP runs with o-proj TP enabled, prefill asynchronously allgathers the TP-local `wo_a` / `wo_b` projection weights, including MXFP8 weight scales when present, switches the projection layers to the gathered full buffers for the o-proj computation, and restores the TP-local tensors afterward. Decode and spec-decode keep the existing activation all-to-all behavior.

The weight allgather is started after hidden-state gathering and before the remaining KV projection / KV cache preparation work, so the communication can be overlapped and hidden by the following compute path instead of being paid as a fully serial step immediately before o-proj.

The PR also patches post-load weight processing so `DSAAttention.process_weights_after_loading()` is invoked from the same loader hook as other attention modules, allowing DSA-CP full-weight buffers to be initialized after weights are loaded instead of lazily in forward.

No user-facing API, configuration, environment variable, or schema changes. The change only affects the internal DSV4 DSA-CP o-proj TP execution path when the existing o-proj TP enablement is active.

- `python -m py_compile vllm_ascend/attention/context_parallel/dsa_cp.py vllm_ascend/patch/worker/patch_process_weights_after_loading.py`
- `git diff --check`

`ruff` was not run locally because it is not installed in the current environment.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/patch/worker/patch_process_weights_after_loading.py
+++ b/vllm_ascend/patch/worker/patch_process_weights_after_loading.py
@@ -1,0 +1,63 @@
+import sys
+
+import torch
+from torch import nn
+
+from vllm.config import ModelConfig
+from vllm.model_executor.layers.attention import (
+    Attention,
+    MLAAttention,
+    MMEncoderAttention,
+)
+from vllm.model_executor.layers.quantization.base_config import (
+    QuantizeMethodBase,
+)
+from vllm.model_executor.model_loader import base_loader, utils
+from vllm.model_executor.model_loader.utils import device_loading_context
+from vllm.model_executor.model_loader.reload import set_torchao_reload_attrs
+
+from vllm_ascend.models.layer.attention.layer import DSAAttention
+
+
+def ascend_process_weights_after_loading(
+    model: nn.Module, model_config: ModelConfig, target_device: torch.device
+) -> None:
+    for _, module in model.named_modules():
+        quant_method = getattr(module, "quant_method", None)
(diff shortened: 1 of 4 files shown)


pisceskkk (approved):
LGTM! Thanks for contribution!

vllm_ascend/patch/worker/patch_process_weights_after_loading.py
```diff
@@ -0,0 +1,62 @@
+import sys
+
+import torch
+from torch import nn
```
weiguihua2, line 4: This patch needs to specify the mitigation plan in vllm-ascend/vllm_ascend/patch/__init__.py.

TmacAaron: already add it

yiz-liu (approved):

pisceskkk:
/cherry-pick releases/v0.23.0
[Bot]: cherry-pick command failed: you do not have permission. Only the PR author or users with triage+ permission can trigger /cherry-pick.
[Bot]: cherry-pick command failed. Please check the [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/29316787647) for details.

PR #10701, opened 2026-06-18
[CI] Adjust test case timeout to fit different scenarios

weixinAc:

Background: 
1. Before running the selected test cases, the nightly test triggered by PR may require additional time to check and install the PR code. Currently, the reusable nightly workflow either hardcodes a timeout of 120 minutes or does not define a consistent job timeout. As a result, the long-running PR nightly job is canceled before it is completed. 
2. In the current weekly workflow, a test case is configured with more than 10 performance scenarios, and the execution times out and is interrupted.
 
Modification: 
- When the request_id is not empty (PR or nightly scheduling), the timeout is set to 240 minutes.
- For scheduled nightly running and regular manual scheduling without request_id, if a value is passed, the timeout of the passed value is used. If no value is passed, the default timeout of 120 minutes is used.
- The timeout is set to 600 minutes during the scheduling of the weekly scheduled task.

NO

by the running the test

- vLLM version: v0.22.1

### DIFF
--- a/.github/workflows/_e2e_nightly_multi_node.yaml
+++ b/.github/workflows/_e2e_nightly_multi_node.yaml
@@ -98,7 +98,7 @@ jobs:
   e2e:
     name: ${{ inputs.config_file_path }}
     # timeout-minutes: 120
-    timeout-minutes: ${{ inputs.request_id != '' && 240 || (inputs.testcase_timeout != '' && inputs.testcase_timeout || 120) }}
+    timeout-minutes: ${{ inputs.request_id != '' && 240 || fromJSON(inputs.testcase_timeout) }}
     # This is the runner with no NPU for k8s controller
     runs-on: ${{ inputs.runner }}
     if: ${{ inputs.should_run }}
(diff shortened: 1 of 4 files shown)


weijinqian0 (approved):

.github/workflows/_e2e_nightly_multi_node.yaml
```diff
@@ -92,7 +97,8 @@ concurrency:
 jobs:
   e2e:
     name: ${{ inputs.config_file_path }}
-    timeout-minutes: 120
+    # timeout-minutes: 120
```
MengqingCao: plz remove this comment.

.github/workflows/_e2e_nightly_single_node.yaml
```diff
@@ -83,7 +88,8 @@ jobs:
     name: ${{ inputs.name || inputs.config_file_path || inputs.tests }}
     runs-on: ${{ inputs.runner }}
     if: ${{ inputs.should_run }}
-    timeout-minutes: 120
+    # timeout-minutes: 120
```
MengqingCao: ditto

MengqingCao (approved):

PR #10711, opened 2026-06-18
[BugFix] Disable get_device_total_memory on Ascend

chenchuw886:

This PR intentionally disables `NPUPlatform.get_device_total_memory()` by making the override raise `NotImplementedError`. It also removes the `npu-smi` memory-query helpers and their obsolete unit tests.

When this API returns successfully, upstream `EngineArgs.get_batch_defaults()` immediately calls `current_platform.get_device_name()`. On Ascend, that query can initialize the NPU runtime before DP ranks finish binding their devices, causing multiple ranks to initialize the same NPUs.

The explicit override and comment preserve this platform contract and warn future maintainers not to reimplement the query. Raising `NotImplementedError` keeps upstream vLLM on its safe fallback path. A runtime-safe device metadata solution can be handled separately.

No.

- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -19,7 +19,6 @@ from __future__ import annotations
 
 import math
 import os
-import subprocess
 from importlib import import_module, util
 from typing import TYPE_CHECKING, Any
 from uuid import uuid4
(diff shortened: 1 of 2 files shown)


vllm_ascend/platform.py
```diff
@@ -298,21 +269,6 @@ def get_device_uuid(cls, device_id: int = 0) -> str:
             raise RuntimeError(f"Device {device_id} does not have a valid UUID.")
         return device_props.uuid
 
-    @classmethod
-    def get_device_total_memory(cls, device_id: int = 0) -> int:
```
wangxiyuan, line 302: ```do not override ```
 raise NotImplementedError

MengqingCao:
Thanks for this fix, could you also raise a pr for branch `releases/v0.22.1rc`?

vllm_ascend/platform.py
chenchuw886: Done

chenchuw886:
> Thanks for this fix, could you also raise a pr for branch `releases/v0.22.1rc`?

done
#10794

PR #10077, opened 2026-06-05
[KV Cache][Feature] Support Layerwise KV Pooling

ader47:
### 1. Per-Layer Key Management

- Each KV cache block in every layer was assigned a unique key.
- The number of keys increased proportionally with the number of layers and blocks, resulting in significant metadata management overhead.

### 2. High Metadata Lookup Latency

- Both `put` and `get` operations required querying the Meta Server to resolve the memory address associated with a key.
- Frequent metadata lookups introduced additional latency on the critical access path.

### 3. Inefficient HBM Address Computation

- HBM addresses were computed using Python lists.
- This approach incurred noticeable CPU overhead when processing large numbers of blocks.

---

### 1. Unified Key Management

- Keys are now managed at the request ID level.
- KV cache blocks belonging to different layers but the same request share a single key.
- This significantly reduces the number of keys and metadata management overhead.

### 2. One-Time Address Resolution

- Key-to-address mapping is performed only once during initialization.
- Subsequent accesses locate layer-specific blocks using a base address plus a layer offset.
- This eliminates repeated Meta Server lookups during runtime.

### 3. New Address-Based Memcache Interface

- Introduced a new `batch_copy` API in Memcache.
- Memory can be accessed directly through physical addresses without key resolution.
- This reduces access latency and improves batch transfer efficiency.

### 4. Accelerated HBM Address Computation

- Replaced Python list-based address computation with NumPy vectorized operations.
- This reduces CPU overhead and improves address generation efficiency.

### 5. CPU Affinity Optimization

- Applied CPU affinity binding to Memcache worker threads.
- This minimizes thread migration, reduces cache misses, and improves scheduling efficiency.

### 6. Controlled Overlap Between Data Transfer and Attention Computation

- Limited the degree of concurrency between KV cache transfers and Attention computation.
- This reduces resource contention and minimizes interference with the inference execution path.

## Results

The optimized design delivers the following benefits:

- Reduced metadata lookup overhead.
- Lower access latency through address-based memory operations.
- Improved HBM address computation efficiency.
- Reduced CPU overhead through vectorized computation and CPU affinity optimization.
- Lower runtime contention between KV cache transfers and model execution.
- Improved overall Layerwise KV Cache Pool performance and scalability.

### DeepSeek V3.2 TP16 hit rate 90%
| kv cache pool | Input tokens | Output tokens | Concurrency | Average TTFT |
|----|----|----|----|----|
| no-layerwise | 64k | 1 | 36.9 | 312519.2 ms |
[...]

PR #10733, opened 2026-06-18
[Feature][Worker] Layerwise KV cache pool with prefill layer reuse support (Prefill offload)

ader47:
# Background

  This feature targets PD-disaggregated (prefill/decode separated) deployments. On the prefill node, long-context prefill is the bottleneck: it is slow and memory-hungry, and each layer's KV cache is computed and used only once per iteration, yet it holds a dedicated HBM buffer for the entire run. At the same time, the KV produced by every layer must
  be shipped to the decode node so decode can pick up seamlessly.

  Today, the prefill node allocates a full buffer for every layer.

 # What this PR does

  This PR adds layer reuse to the layerwise KV pool (from #10077), on the prefill (P) node only. Instead of each layer owning its KV buffer, layers time-share a small set of shared HBM buffers. A 78-layer model can run on just 4 buffers. The lifecycle:

  - Before a layer runs, its KV is loaded from the pool into a shared buffer (overlapped with the previous layer's compute).
  - After the layer computes, its freshly produced KV is incrementally saved back to the pool, freeing the buffer for reuse.
  - When chunked prefill returns to a layer whose turn already passed, its prior KV is no longer in a buffer (the shared slot was overwritten by other layers), so it is reloaded from the pool. Results stay bit-for-bit identical — only where the KV lives changes, not its contents.

  Works alongside the existing layerwise mooncake connector

  This feature is designed to be used alongside the existing layerwise mooncake connector (also from #10077), which streams each layer's KV to the decode (D) node as soon as it is produced. In a PD deployment the two can run side by side on P:

  - the ascend_store pool reuses buffers to minimize P-node HBM,
  - the mooncake connector ships each layer's KV out while it is still live.

  Since reuse overwrites the shared buffers, it could race an in-flight PD transfer without coordination. So this PR adds a per-layer handshake: the pool will not overwrite a shared buffer until mooncake confirms it has delivered that layer's KV to D — so reuse never breaks an ongoing PD transfer.

#  Why it helps

  - Lower P-node HBM footprint — buffer count drops from num_layers to a configurable few, freeing memory for larger KV caches and bigger prefill batches.
  - Overlapped data movement — pool load/save is hidden behind prefill compute, so reuse costs no throughput.
  - No precision loss — KV is preserved across chunks; only its location changes.

# Usage

  Layer reuse is off by default and enabled through kv_connector_extra_config (no environment variables):

  {
    "layerwise_num_shared_buffers": 4,
    "layerwise_prefetch_layers": "...",
    "layerwise_independent_layers": "..."
  }

  Decode cache saving is also off by default (save_decode_cache=False); chunked-prefill save is always on.

Note: layer reuse applies only to the prefill (P) node. For PD deployments, pair it with the existing layerwise mooncake connector to carry the reused KV to the decode node.

# Implementation
  
  Layer reuse is built on top of the layerwise KV pool (#10077): multiple layers time-multiplex a single shared transfer buffer, lowering HBM footprint. To keep vLLM's block accounting whole, the reported available memory is inflated by total_layers / reuse_layers, so vLLM still sizes num_blocks for the full layer count; the model runner then merges
  the per-layer KVCacheTensor entries into those shared buffers (shared_by).

  - layerwise_config.py — parse num_shared_buffers / prefetch_layers / independent_layers from kv_connector_extra_config and compute the shared storage slot layout (get_layerwise_storage_indices).
  - worker.py — inflate the reported available memory by total_layers / reuse_layers so vLLM sizes num_blocks for the full layer count.
  - model_runner_v1.py — merge per-layer KVCacheTensor entries into the shared buffers, and reject hybrid attention+mamba reuse.
  - pool_worker / pool_scheduler — drive prefetch and offload from the parsed LayerwiseConfig.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/distributed/ascend_store/test_layerwise_config.py
+++ b/tests/ut/distributed/ascend_store/test_layerwise_config.py
@@ -1,0 +1,165 @@
+"""Unit tests for layerwise KV cache reuse config parsing."""
+
+import pytest
+
+from vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.layerwise_config import (
+    get_layer_load_start_block,
+    get_layerwise_config,
+    get_layerwise_independent_layers,
+    get_layerwise_kv_cache_num_tensors,
+    get_layerwise_kv_cache_reuse_layers,
+    get_layerwise_num_prefetch_layers,
+    get_layerwise_num_shared_buffers,
+    get_layerwise_storage_indices,
+)
+
+
+class TestNumSharedBuffers:
+    def test_default_equals_num_layers(self):
+        assert get_layerwise_num_shared_buffers(27, None) == 27
+        assert get_layerwise_num_shared_buffers(27, {}) == 27
+
+    def test_from_extra_config(self):
+        cfg = {"layerwise_num_shared_buffers": 6}
+        assert get_layerwise_num_shared_buffers(27, cfg) == 6
+
+    def test_string_value_parsed(self):
+        cfg = {"layerwise_num_shared_buffers": "6"}
+        assert get_layerwise_num_shared_buffers(27, cfg) == 6
+
+    def test_bool_rejected(self):
+        with pytest.raises(TypeError):
+            get_layerwise_num_shared_buffers(27, {"layerwise_num_shared_buffers": True})
+
+    def test_less_than_one_rejected(self):
+        with pytest.raises(ValueError):
+            get_layerwise_num_shared_buffers(27, {"layerwise_num_shared_buffers": 0})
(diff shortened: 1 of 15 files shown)


LCAIZJ:
Does this change only allow decode to fetch KV cache from the KV pool?

ader47:
> Does this change only allow decode to fetch KV cache from the KV pool?

This change supports both the prefill and decode stages; however, it results in extremely poor performance during the decode stage. Therefore, it is strongly recommended to use it only in the prefill stage.

PR #10735, opened 2026-06-19
[BugFix] Persist enable_npugraph_ex override for worker processes

ShuhaoZhangTony:
## Problem

When `platform.py` overrides `enable_npugraph_ex = False` for PIECEWISE or NONE cudagraph modes, the change only affects the main-process `AscendConfig`. Spawned workers reconstruct their configuration from `vllm_config.additional_config`, which still contains the default value. Workers can therefore select a different compilation path from the main process.

## Fix

Add `_sync_npugraph_ex_to_additional_config()` and call it after cudagraph-mode selection. The final `enable_npugraph_ex` value is persisted into `additional_config` before workers are spawned.

- Full-graph mode remains unchanged when the value is already true.
- PIECEWISE and NONE overrides now propagate to workers.

## Reproduction and verification

Start a TP=4 server with PIECEWISE compilation and verify that every worker receives the same final `enable_npugraph_ex` value selected by the main process. Performance results will only be added after a fixed-commit, repeated real-hardware comparison is complete.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -74,6 +74,28 @@ _CUSTOM_OP_REGISTERED = False
 MAX_CAPTURE_SIZES_FOR_950 = 4
 
 
+def _sync_npugraph_ex_to_additional_config(
+    vllm_config: VllmConfig,
+    ascend_config,
+) -> None:
+    """Persist *enable_npugraph_ex* into ``vllm_config.additional_config``.
+
+    ``platform.py`` may override ``enable_npugraph_ex`` (e.g. to ``False`` for
+    PIECEWISE / NONE cudagraph modes).  Worker processes spawned via
+    ``multiprocessing`` re-initialize their own ``AscendConfig`` from
+    ``vllm_config.additional_config``, so the override must be persisted there
+    as well.  Without this sync, workers silently fall back to the default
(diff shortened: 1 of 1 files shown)


winson-00178005:
## npugraph_ex 配置同步修复

修复 `enable_npugraph_ex` 在 worker 进程中丢失的问题。当 platform.py 将 `enable_npugraph_ex` 设为 False（例如 PIECEWISE/NONE cudagraph 模式），worker 进程通过 `vllm_config.additional_config` 重新初始化 AscendConfig 时会恢复为默认 True，导致使用错误的编译路径。

修复思路正确：在 `check_and_update_config` 中将 override 后的值同步回 `vllm_config.additional_config`。

一处建议：`_sync_npugraph_ex_to_additional_config` 函数只同步了一个字段。如果未来还有其他 `AscendConfig` 字段也需要类似的同步（例如 `enable_fused_mc2` 等），建议改为通用的 sync 函数，一次性同步所有 override 字段，避免逐个字段添加 sync 逻辑。

ShuhaoZhangTony:
Continued as #12342 on the required `feature/...` branch. The replacement is rebased onto current `main`, adds regression tests, and fixes the review finding for a `None` or non-dictionary `ascend_compilation_config`. This PR is closed to keep one active review target.

PR #10476, opened 2026-06-15
[Misc] Main2main upgrade vllm to v0.22.1

MrZ20:
### What this PR does / why we need it?

This PR updates the vLLM release tag used by vLLM Ascend from `v0.21.0` to `v0.22.1`.

Main changes:

- Bump `.github/vllm-release-tag.commit` from `v0.21.0` to `v0.22.1`.
- Update Dockerfiles to build against vLLM `v0.22.1`.
- Update scheduled image build workflow options from `v0.21.0rc1` to `v0.22.1rc1`.
- Update README and slash-command docs to reference the new maintained vLLM release tag.
- Adapt Ascend code paths to vLLM `v0.22.1` API changes, including:
  - routed experts output handling
  - v2 model runner input batch fields
  - block table constructor changes
  - EAGLE/autoregressive speculative decoding module layout
  - GDN import path and operator naming changes
  - compressed prefix cache manager signature changes
- Remove obsolete `v0.21.0` compatibility branches where the code now targets `v0.22.1`.
- Update related unit/e2e test expectations and skips for the `v0.22.1` tag.

### How was this patch tested?
**CI action link for test**: https://github.com/vllm-project/vllm-ascend/actions/runs/27563316722

- vLLM version: v0.21.0
- vLLM main: https://github.com/vllm-project/vllm/commit/967c5c3bc38891f4465d3f4e99917ed837bb3833

PR #10704, opened 2026-06-18
[Misc] Drop vLLM v0.22.1 compatibility from main

MrZ20:
### What this PR does / why we need it?

Upstream has cut the `releases/v0.22.1rc` release branch off `main`. From now on
`main` only needs to track vLLM `main`, so the dual-version compatibility shims
added during the main2main upgrade are dead weight on the `main` line. This PR:

1. Removes **every** `vllm_version_is(...)` branch, keeping only the vLLM-`main`
   path, and deletes the now-unused helper and its supporting machinery.
2. Repoints the nightly CI at the new `releases/v0.22.1rc` branch (following the
   pattern of #9778, which did this for `releases/v0.20.2rc`).

#### 1. Drop v0.22.1 (and stale 0.21.0 / 0.19.1) version branching

- Collapse all `vllm_version_is(...)` conditionals to the vLLM-`main` side across
  core / KV-cache / ops / spec-decode / patch / worker modules. This also clears
  leftover `0.21.0` and `0.19.1` checks.
- Align override signatures with upstream main (e.g. drop the legacy `use_eagle`
  kwarg from `find_longest_cache_hit` overrides in favor of `drop_eagle_block`).
- Remove the helper and scaffolding now that all uses are gone:
  - `vllm_version_is` in `vllm_ascend/utils.py` (and the now-unused
    `packaging.version` import).
  - the `VLLM_VERSION` env entry in `vllm_ascend/envs.py`.
  - the helper self-test in `tests/ut/test_utils.py`.
  - dead flag machinery in `patch_tool_choice_none_content.py` whose only setter
    lived in the removed 0.22.1 patch.
- Update the README / README.zh support matrix to advertise "vLLM main branch"
  only.

#### 2. CI: keep the release-tag legs restorable, switch nightly to the rc

- **Comment out (not delete)** the v0.22.1 release-tag CI legs so the next
  version-compat round can restore them easily (restore notes point at
  `.github/vllm-release-tag.commit`):
  - `release_tag` matrix legs in `pr_test.yaml` and `pr_e2e_command.yml`.
  - the `E2E-upstream` schedule cron in `schedule_vllm_e2e_test.yaml` (kept
    `workflow_dispatch`).
  - the `run-tests-release` job in `schedule_update_estimated_times.yaml`.
- **Switch nightly to `releases/v0.22.1rc`:**
  - `schedule_nightly_test_a2.yaml` / `schedule_nightly_test_a3.yaml`: switch the
[...]

PR #10740, opened 2026-06-21
[BugFix] Fix logic bug in patch_kv_coordinator to properly support vLLM 0.22.1

underfituu:
https://github.com/vllm-project/vllm-ascend/pull/10476

**What:** This PR fixes a syntax/logic error in `patch_kv_coordinator` where multiple `vllm_version_is` conditions were incorrectly chained or nested. It corrects the version check logic so that `eagle_kwarg` is accurately set based on the vLLM version.

**Why:**
In the original code, `if vllm_version_is("0.22.1"):` was placed directly under `if vllm_version_is("0.21.0"):` without proper control flow (like `elif`), which leads to dead code or unexpected behavior when determining whether to use `use_eagle` or `drop_eagle_block`.

**Corrected Logic Applied:**

```python
# Eagle needs to match one more block and then pop the last.
_max_length = min(curr_hit_length + spec.block_size, max_cache_hit_length)
# vLLM B renamed the ``use_eagle`` kwarg to ``drop_eagle_block``.
if vllm_version_is("0.22.1"):
    eagle_kwarg = {"use_eagle": use_eagle}
else:
    eagle_kwarg = {"drop_eagle_block": use_eagle}

```

No. This is a purely internal bug fix to adapt to different vLLM versions and does not change any public APIs or user-facing behavior.

* **Unit Tests:** Verified that the integration with vLLM `0.21.0`, `0.22.1`, and newer versions passes the local coordination tests. [CI]
* **Manual Verification:** Verified that `eagle_kwarg` correctly resolves to `{"use_eagle": ...}` when mock version is set to `0.22.1`. [CI]
- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
+++ b/vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
@@ -381,7 +381,7 @@ class AscendHybridKVCacheCoordinator(HybridKVCacheCoordinator):
                     # Eagle needs to match one more block and then pop the last.
                     _max_length = min(curr_hit_length + spec.block_size, max_cache_hit_length)
                 # vLLM B renamed the ``use_eagle`` kwarg to ``drop_eagle_block``.
-                if vllm_version_is("0.21.0"):
+                if vllm_version_is("0.22.1"):
                     eagle_kwarg = {"use_eagle": use_eagle}
                 else:
                     eagle_kwarg = {"drop_eagle_block": use_eagle}


liziyu179 (approved):

MengqingCao (approved):

MengqingCao:
Thanks for your fix, but this has already done in #10704

zzzzwwjj (approved):

underfituu:
> Thanks for your fix, but this has already done in #10704

Thanks! I will keep an eye on the community updates once this test version is out. Since the partial hit feature is heading upstream in vLLM v0.23.0 (https://github.com/vllm-project/vllm/pull/44243), we’ll be sunsetting this local patch as soon as we make the switch.

PR #10742, opened 2026-06-21
[Feature][P/D][Core][KVOffload] Support recompute KVCache offloading on P/D disaggregation decoder nodes

nwpu-zxr:
This PR adds a new `RecomputeCPUOffloadConnector` for vLLM Ascend.

The connector is used with `RecomputeScheduler` to reduce the recomputation cost caused by request preemption. The `max_num_batch_tokens` of the decode node is usually set to `max_num_seqs * (1 + num_spec_tokens)`. Therefore, if the Decode node performs recompute for prefill, there is a high probability that out of memory (OOM) occurs and the performance is greatly affected. In the original `RecomputeScheduler`, the request is recombined and then sent to the prefill node for prefill again. During prompt recombination, tool calling and function call cannot be used, which also causes accuracy issues. This PR introduces recompute offload to solve the preceding problems. When a running request is preempted because HBM KV cache blocks are insufficient, the scheduler can offload the preempted request's existing KV cache blocks from HBM to CPU DRAM. When the request is scheduled again, the connector reloads these blocks back to HBM before model forward.

Main changes:

- Add `vllm_ascend.distributed.kv_transfer.kv_pool.recompute_cpu_offload`.
  - Scheduler-side manager tracks preempted request state, CPU block allocation, D2H store metadata, and H2D reload metadata.
  - Worker-side handler performs basic CPU/NPU KV cache copies with torch `copy_`.
  - Metadata classes carry preemption store/load events between scheduler and worker.

- Register `RecomputeCPUOffloadConnector` in the Ascend KV connector factory.

- Extend `RecomputeScheduler` recompute-preemption logic.
  - If offload succeeds, the preempted request is preserved in CPU memory and restored when it is scheduled again.
  - If no offload connector is configured, or CPU offload capacity is insufficient, it falls back to the original recompute behavior.

- Add `AscendMultiConnector` support for recompute offload.
  - Recompute-offloaded requests are prioritized during `get_num_new_matched_tokens`, so private/unhashed blocks can be restored correctly even when other KV connectors are also configured.

- Add model runner handling so preemption D2H stores are executed before `_update_states()` can zero or reuse the same physical KV blocks.

- Add connector capacity configuration.
  - `cpu_bytes_to_use`: total CPU memory budget for recompute offload. Default is 8 GB.
  - `cpu_bytes_to_use_per_rank`: optional per-rank override.
  - `enable_offload_prefix_caching`: experimental prefix-cache sharing for offloaded blocks. It is disabled by default and is not recommended for normal use yet because it has not been fully tested.

This PR is intended to improve recompute-preemption behavior for normal full-attention models. More complex hybrid/sliding-window/async edge cases can be refined in follow-up PRs. Currently, there are some known issues, including that the Qwen3.5 series models can only function properly when `async-scheduling` is disabled, and deepseek-v4 series models have precision issues with the sliding window attention kvcache.

Update: The accuracy issues of the deepseek-v4 series models have been fixed through self-testing and verification.

Yes.

A new KV transfer connector is available:

```python
KVTransferConfig(
    kv_connector="RecomputeCPUOffloadConnector",
    kv_role="kv_consumer",
    kv_connector_extra_config={
        # Optional. Default is 8 GB.
        "cpu_bytes_to_use": 17179869184,

        # Experimental, disabled by default.
        # Not recommended unless specifically testing prefix sharing.
        "enable_offload_prefix_caching": False,
    },
)
```
For online P/D disaggregation, you can use `MultiConnector` to enable recompute kv-offload on decoder nodes, `RecomputeScheduler` must be enable:
```bash
--kv-transfer-config \
    '{
    "kv_connector": "MultiConnector",
    "kv_role": "kv_consumer",
        "use_layerwise": false,
        "connectors": [
            {
                "kv_connector": "MooncakeConnectorV1",
                "kv_role": "kv_consumer",
                "kv_port": "26010",
                    "prefill": {
                        "dp_size": 2,
                        "tp_size": 2
                     },
                    "decode": {
                        "dp_size": 2,
                        "tp_size": 2
                     }
                }
            },
            {
                "kv_connector": "RecomputeCPUOffloadConnector",
                "kv_role": "kv_consumer",
                    "cpu_bytes_to_use": 17179869184
                }
            }  
        ]
    }
```

The feature is opt-in and only takes effect when the recompute scheduler is enabled. Existing behavior is preserved when this connector is not configured, or when the connector cannot offload because CPU cache capacity is insufficient.

Added unit tests for `RecomputeCPUOffloadConnector`:

```bash
pytest tests/ut/distributed/kv_transfer/test_recompute_cpu_offload.py -q
```

The tests cover:

- scheduler-side preempted request state transitions;
- CPU block allocation and cleanup;
- store/load metadata construction;
- worker-side metadata aggregation;
- D2H/H2D entrypoints;
- fallback behavior when offload is unavailable;
- recompute scheduler remote-KV restore token-position handling.

Also, I conducted an E2E accuracy test. To increase the probability of triggering recomputation, it is necessary to adjust the `gpu-memory-utilization` to a relatively low value. The test results for the `qwen3-30b-a3b-instruct-2507` model on the PD-separated GPQA dataset showed the same accuracy as the score obtained without offloading and directly performing recomputation.

Update: The accuracy issues of the deepseek-v4 series models have been verified and fixed through self-testing, with the results as follows:

| GPQA_diamond | b1ed2c | accuracy | gen | 85.86 |

According to the test, when the number of concurrent requests reaches the maximum and offload is frequently triggered, the precision is the same as that of the test result when no offload is triggered. The sampling parameters are as follows: temp=1, top_p=0.95, enable_thinking, max_len=35000.

The UTs were run on an Ascend server and passed.

- vLLM version: v0.22.1


LCAIZJ:
Could this design be submitted to the upstream vLLM community?

nwpu-zxr:
> Could this design be submitted to the upstream vLLM community?

I will try, but for now, I will first fully verify on vllm ascend.

liziyu179 (approved):

vllm_ascend/distributed/kv_transfer/kv_pool/recompute_cpu_offload/recompute_cpu_offload_connector.py
```diff
@@ -0,0 +1,246 @@
...
+    from vllm.v1.core.kv_cache_manager import KVCacheBlocks
+    from vllm.v1.kv_cache_interface import KVCacheConfig
+    from vllm.v1.request import Request
+
+# Default CPU capacity: 8 GB
+DEFAULT_CPU_CAPACITY_BYTES = 8 * (1024**3)
+
+
+class RecomputeCPUOffloadConnectorV1(KVConnectorBase_V1, SupportsHMA):
+    """CPU KV cache preservation for recompute-preempted requests."""
```
MengqingCao, line 44: plz add docstring to describe which scenario does this connector work for. I'm okay with doing this in a follo-up pr

vllm_ascend/distributed/kv_transfer/kv_pool/recompute_cpu_offload/recompute_cpu_offload_connector.py
```diff
@@ -0,0 +1,246 @@
...
+if TYPE_CHECKING:
+    from vllm.forward_context import ForwardContext
+    from vllm.v1.attention.backend import AttentionMetadata
+    from vllm.v1.core.block_pool import BlockPool
+    from vllm.v1.core.kv_cache_manager import KVCacheBlocks
+    from vllm.v1.kv_cache_interface import KVCacheConfig
+    from vllm.v1.request import Request
+
+# Default CPU capacity: 8 GB
+DEFAULT_CPU_CAPACITY_BYTES = 8 * (1024**3)
```
MengqingCao, line 40: why setting this to 8GB?

vllm_ascend/distributed/kv_transfer/kv_pool/recompute_cpu_offload/recompute_cpu_offload_connector.py
MengqingCao: and also for the member classes.

nwpu-zxr: I will add more information about this connector on next PR.

vllm_ascend/distributed/kv_transfer/kv_pool/recompute_cpu_offload/recompute_cpu_offload_connector.py
nwpu-zxr: This inherits the default setting of the vllm `SimpleCpuOffloadConnector`, that is, each `dp_rank` uses `8 GiB/dp_world_size` DRAM buffer space.

vllm_ascend/distributed/kv_transfer/kv_pool/recompute_cpu_offload/worker.py
```diff
@@ -0,0 +1,319 @@
...
+
+        unique_gpu_caches: dict[str, torch.Tensor] = {}
+        register_cache_ptrs = []
+        for layer_name, layer_tensor in kv_caches.items():
+            if isinstance(layer_tensor, (tuple, list)):
+                for idx, single_tensor in enumerate(layer_tensor):
+                    if single_tensor.data_ptr() not in register_cache_ptrs:
+                        unique_gpu_caches[f"{layer_name}.{idx}"] = single_tensor.view(single_tensor.shape[0], -1)
+                        register_cache_ptrs.append(single_tensor.data_ptr())
+                        self.block_size_scale[f"{layer_name}.{idx}"] = single_tensor.shape[0] // self.num_gpu_blocks
```
MengqingCao, line 83: it is better to rename it to `block_num_scale`

vllm_ascend/distributed/kv_transfer/kv_pool/recompute_cpu_offload/worker.py
```diff
@@ -0,0 +1,319 @@
...
+                "worker-side num_block is set to %d to align with scheduler.",
+                scheduler_num_cpu_blocks,
+                scheduler_num_cpu_blocks,
+            )
+
+        self.gpu_kv_caches = unique_gpu_caches
+        self.cpu_kv_caches = {}
+        for name, gpu_tensor in unique_gpu_caches.items():
+            tensor_block_size_scale = self.block_size_scale[name]
+            cpu_shape = (self.num_cpu_blocks * tensor_block_size_scale,) + gpu_tensor.shape[1:]
```
MengqingCao, line 106:
```suggestion
            cpu_kv_cache_shape = (self.num_cpu_blocks * tensor_block_size_scale,) + gpu_tensor.shape[1:]
```

vllm_ascend/distributed/kv_transfer/kv_pool/recompute_cpu_offload/worker.py
```diff
@@ -0,0 +1,319 @@
...
+        with torch.npu.stream(stream):
+            for src_block_id, dst_block_id in zip(src_block_ids, dst_block_ids):
+                for name, gpu_tensor in self.gpu_kv_caches.items():
+                    cpu_tensor = self.cpu_kv_caches[name]
+                    tensor_block_size_scale = self.block_size_scale[name]
+                    if is_store:
+                        # TODO: Replace this D2H torch copy with the NPU copy
+                        # backend dedicated kernel.
+                        if tensor_block_size_scale > 1:
+                            cpu_tensor[
```
MengqingCao, line 240: maybe we could replace this by `torch.ops._C_ascend.swap_blocks_batch`, but actually I've not tested its performance.

vllm_ascend/distributed/kv_transfer/kv_pool/recompute_cpu_offload/worker.py
```diff
@@ -0,0 +1,319 @@
...
+                self._load_hwm = max(self._load_hwm, event_idx)
+            return
+
+        assert not is_store
+        self._load_events.append((event_idx, event))
+
+    def get_finished(
+        self,
+        finished_req_ids: set[str],
+    ) -> tuple[set[str] | None, set[str] | None]:
```
MengqingCao, line 289: what does the return values refers to? seems the first one always be None?

vllm_ascend/distributed/kv_transfer/kv_pool/recompute_cpu_offload/worker.py
nwpu-zxr: I will try to use this later.

vllm_ascend/distributed/kv_transfer/kv_pool/recompute_cpu_offload/worker.py
nwpu-zxr: This function is also inherited from vllm simpleoffload. Its first return value represents done_sending, which is the set of request_ids that have completed transmission to node D. This is meaningless for offloading, so it returns None.

vllm_ascend/distributed/kv_transfer/kv_pool/recompute_cpu_offload/worker.py
nwpu-zxr: I will rename this later.

MengqingCao:
It seems only the tests with 1 card is passed https://github.com/vllm-project/vllm-ascend/actions/runs/27956060159/job/82725900120, but let's merge this quickly to test it tonight, and if any ci issue introduced by this pr, we'll revert it ASAP.

MengqingCao (commented):
overall I agree to merge this quickly, but I think there are still many works to do in the follow-up pr:
- [ ] using batch copy operators
- [ ] add more docstring for new introduced classes and functions
- [ ] rename some variables
- [ ] unified with existing simple connectors

MengqingCao (approved):

nwpu-zxr:
> overall I agree to merge this quickly, but I think there are still many works to do in the follow-up pr:
> 
> * [ ]  using batch copy operators
> * [ ]  add more docstring for new introduced classes and functions
> * [ ]  rename some variables
> * [ ]  unified with existing simple connectors

All right, I will complete these tasks later. Thanks for reviewing and merging..

bupt-ycq:
返回时是否一次性全部加载kv cache，如果hbm还是不够的话 可能会存在kv cache较大没办法一次加载会导致请求反而会比重算还慢的情况 这面请问是否测过高并发情况下tpot的对比

PR #10746, opened 2026-06-22
[BugFix] Avoid the problem that the CPP is suspended due to profiling failure on qwen3.5

gjc0824:

This PR addresses an issue where the Chunk Prefill Predictor (CPP) hangs/suspends the scheduling of requests when profiling fails (e.g., when the predicted chunk size is `None` or `<= 0`). 

To resolve this, the PR:
1. Updates the scheduler to degenerate to a fixed chunk size instead of breaking the scheduling loop when profiling fails.
2. Introduces a configurable `max_fit_chunk` parameter to allow more adequate fitting.
3. Removes the negative clamping for the fitted quadratic term to perform calculations based on the actual fitting relationship.

Yes, adds a new configuration option `max_fit_chunk` under `profiling_chunk_config` to control the maximum number of data points used for chunked fitting.

Tested with Qwen 3.5 model profiling and verified that the scheduler no longer hangs when profiling fails.

- vLLM version: v0.22.1


weiguihua2 (approved):

MengqingCao (approved):

MengqingCao:
plz refresh the doc on ascend_config when you're adding a new one.

gjc0824:
> plz refresh the doc on ascend_config when you're adding a new one.

Sure. We have refreshed the relative Docs.

MengqingCao:
all ci passed in https://github.com/vllm-project/vllm-ascend/actions/runs/27944691179/job/82687166452?pr=10746

PR #10759, opened 2026-06-22
[CI] Remove the validation of the nightly precision test case limit

chen-commits:

This PR removes the upper-bound validation for the nightly precision test cases in `tools/aisbench.py`. Previously, the accuracy verification asserted that the accuracy value was within a threshold relative to the baseline (both lower and upper bounds). This PR changes the assertion to only verify that the accuracy is greater than or equal to the minimum acceptable value (lower bound), allowing higher accuracy values without failing the test.

No.

Tested via existing CI workflows.

- vLLM version: v0.22.1

### DIFF
--- a/tools/aisbench.py
+++ b/tools/aisbench.py
@@ -243,7 +243,7 @@ class AisbenchRunner:
         acc_value = self.result
         lower_bound = self.baseline - self.threshold
         assert acc_value >= lower_bound, (
-            f"Accuracy verification failed. "
+            "Accuracy verification failed. "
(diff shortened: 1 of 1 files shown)


zhangxinyuehfad:
/nightly multi-node-GLM-5.1-w8a8-A3
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/27933956924)

jiangyunfan1 (approved):

PR #10763, opened 2026-06-22
[BugFix] Fix MTP DSA graph capture metadata

naliazheli:
## Summary

Fix MTP speculative decoding with DSA compressed attention when FULL decode graph capture is enabled.

## Root cause

When MTP speculative decoding captures FULL decode graphs with DSA compressed attention enabled, the drafter dummy-run path builds an `AscendCommonAttentionMetadata` with `positions` but without `positions_cpu`. The DSA-CP metadata builder later indexes `common_attn_metadata.positions_cpu`, which causes graph capture to fail with `TypeError: 'NoneType' object is not subscriptable`.

After startup succeeds, the first request replays the draft ACL graph and updates graph params through `update_full_graph_params()`. DSA/DSA-CP attention does not register mutable attention graph params, but the common graph-update path still calls `impl_cls.update_graph_params()`. Since `AscendDSACPImpl` inherited no such method, request-time graph replay failed with `AttributeError: type object 'AscendDSACPImpl' has no attribute 'update_graph_params'`.

These paths are only exercised when the speculative drafter is allowed to use graph capture. Deployments that set `"enforce_eager": true` in `--speculative-config` bypass this path because `AscendSpecDecodeBaseProposer.use_cuda_graph` becomes false.

## Fix

- Reuse the runner's `_dsa_positions_cpu_buf` for compressed-attention graph capture metadata, matching the main runner metadata path. Non-compressed paths keep `positions_cpu=None`.
- Add a no-op `update_graph_params()` to `DSAAttentionImpl`, matching DSA's lack of registered mutable graph attention params and covering both regular DSA and DSA-CP implementations.

## Validation

- `python -m py_compile vllm_ascend/spec_decode/llm_base_proposer.py vllm_ascend/attention/abstract.py`
- Reproduced the startup failure with DeepSeek-V4-Flash w8a8 MTP on Ascend using MTP graph capture enabled; after the metadata fix, `Capturing CUDA graphs (decode, FULL)` completed `8/8`, the server logged `Application startup complete`, and `GET /v1/models` returned the served model.
- Reproduced the request-time failure on `/v1/chat/completions` after startup; after adding the DSA graph-update no-op, two chat completion requests returned `200 OK` while the worker logged `Replaying aclgraph` without `AttributeError` or worker failure.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/attention/abstract.py
+++ b/vllm_ascend/attention/abstract.py
@@ -16,6 +16,18 @@ T = TypeVar("T", bound=AttentionMetadata)
 
 
 class DSAAttentionImpl(AttentionImpl[T], Generic[T]):
+    @staticmethod
+    def update_graph_params(
+        update_stream,
+        forward_context,
+        num_tokens,
+        vllm_config=None,
+        speculative_config=None,
+        num_dcp_pcp_tokens=None,
+        draft_attn_metadatas=None,
+    ):
+        pass
+
     @abstractmethod
     def __init__(
         self,
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -1258,6 +1258,14 @@ class NPUModelRunner(GPUModelRunner):
         if self._needs_seq_lens_cpu_sync and async_spec_decode_active:
             self._correct_optimistic_seq_lens_cpu(num_reqs)
 
+        if self.use_compress:
+            dsa_base_tokens = (
+                base_num_computed_tokens_np
+                if base_num_computed_tokens_np is not None
+                else self.input_batch.num_computed_tokens_cpu[:num_reqs]
+            )
+            self._dsa_num_computed_tokens_for_positions_np = dsa_base_tokens.copy()
+
         # For non-PCP, compute slot_mapping on GPU. PCP slot_mapping was
         # already computed on GPU before PCP split the positions.
         if self.pcp_size <= 1:
@@ -2042,6 +2050,13 @@ class NPUModelRunner(GPUModelRunner):
         num_scheduled_tokens = scheduler_output.total_num_scheduled_tokens
         with record_function_or_nullcontext("prepare input"):
             with self.synchronize_input_prep():
+                if self.use_async_scheduling and self.use_compress:
+                    # DSA CP graph capture reuses input/attention metadata
+                    # buffers by address. Some DSA graph work may outlive the
+                    # Python current stream, so wait on the device before
+                    # _prepare_inputs overwrites them for the next async step.
+                    torch.npu.synchronize()
+
                 # Fix up prev_req_id_to_index for requests that were discarded
                 # in the previous sample_tokens step. If a request has
                 # prev_num_draft_len > 0 but is missing from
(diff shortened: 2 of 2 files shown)


vllm_ascend/attention/abstract.py
```diff
@@ -16,6 +16,18 @@ class AttentionMetadata:
```
winson-00178005, line 17: ## MTP DSA graph capture metadata 修复

修复在 MTP speculative decoding 场景下 DSA 的 ACL graph capture metadata 问题。新增了 `DSAAttentionImpl.update_graph_params` 静态方法和 `positions_cpu` 参数传给 dummy_run。

一处建议：`update_graph_params` 在 `DSAAttentionImpl` 基类中定义为空实现（pass），而具体实现在子类中。但 `abstract.py` 中这个方法没有用 `@abstractmethod` 标注，这意味着子类可能忘记实现它。建议要么标注为 `@abstractmethod` 强制子类实现，要么在基类中抛出 `NotImplementedError` 以提供明确的错误提示。

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -554,6 +554,11 @@ def dummy_run(
                 slot_mapping=self.runner.input_batch.block_table[0].slot_mapping.gpu,
                 slot_mapping_cpu=self.runner.input_batch.block_table[0].slot_mapping.cpu,
                 positions=self.runner.positions,
+                positions_cpu=(
+                    self.runner._dsa_positions_cpu_buf
+                    if getattr(self.runner, "use_compress", False)
+                    else None
+                ),
```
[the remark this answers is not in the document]
naliazheli: Thanks, good point. I updated this to use `getattr(self.runner, "_dsa_positions_cpu_buf", None)` defensively, so tests or alternate runner implementations without this field will not hit an `AttributeError`.

vllm_ascend/attention/abstract.py
naliazheli: 这里保留 no-op 是有意的。

`update_graph_params` 是 `update_full_graph_params` 统一调用的 graph hook，但不是所有 attention backend 都需要在 ACL graph replay 前更新 graph params。普通 attention / MLA 路径会在 capture 阶段保存 `graph_params.attn_params / handles / events`，replay 前通过 `graph_task_update_begin/end` 动态 patch attention op 参数，所以这些 backend 需要实现实际更新逻辑。

当前 DSA / DSA-CP 路径没有这套 capture-time graph params 注册逻辑，也没有 replay-time graph task patch 需求；动态信息是通过 metadata / tensor 输入传给 DSA sparse attention op 的。因此这里如果改成 `@abstractmethod` 或 `raise NotImplementedError`，反而会让当前 DSA / DSA-CP 必须额外实现一个空方法，否则会重新破坏这次 MTP DSA graph capture 路径。

我倾向于保留 no-op，并把它作为该统一 hook 的默认行为：DSA backend 当前不需要更新 graph params；后续如果某个 DSA 子类引入 replay 前参数更新需求，再 override 这个 hook。

naliazheli: 这里也和现有 SFA backend 保持一致
image

naliazheli:
Additional update for the runtime crash case:

- Reproduced the issue with MTP + DSA CP + FULL_DECODE_ONLY when async scheduling is enabled: sustained 3 concurrent chat requests eventually hit rank6/rank7 `fftsplus aivector` / MTE out-of-range errors, later surfacing at `_seq_lens_cpu_event.synchronize()`.
- Verified that keeping `enforce_eager=false` and FULL_DECODE_ONLY but disabling async scheduling avoids the crash: 90 requests with max 3 in-flight all returned 200, with no ERROR/Traceback/MTE logs.
- Added a platform guard to reset `async_scheduling=False` for the unsupported MTP + DSA CP combo on Ascend, while leaving ACL graph enabled.

naliazheli:
Updated the PR to keep `enforce_eager=false` supported instead of disabling async scheduling for MTP + DSA CP.

What changed:
- Rebased onto latest `main` and resolved conflicts. Latest `main` already carries the `positions_cpu` draft metadata path, so the final diff now only touches the DSA attention hook and model runner.
- Removed the async-scheduling guard from the final diff.
- Added a synchronization before async input/attention metadata buffers are reused in DSA CP graph mode.
- Use the async-corrected `base_num_computed_tokens_np` when building DSA CPU positions, so graph metadata matches rebuilt CP/spec inputs.

Validation:
- `python -m py_compile vllm_ascend/attention/abstract.py vllm_ascend/spec_decode/llm_base_proposer.py vllm_ascend/worker/model_runner_v1.py vllm_ascend/platform.py`
- `git diff --check`
- DeepSeek V4 Flash Ascend env with `enforce_eager=false` and async scheduling: service started successfully, 90 chat requests at 3 concurrency all returned 200 with no MTE/ERROR logs.
- AIPerf 8k/1k CON20, 60 requests: errors=0, output throughput 316.66 tokens/s, MTP accepted/drafted 24593/35364 (69.5%).

PR #10778, opened 2026-06-22
[Profiler][Misc] Update AI Core metrics to PipeUtilization in torch_npu profiler

zhangtao0408:
This PR fixes the Ascend torch profiler configuration in `TorchNPUProfilerWrapper`.

Previously, vLLM Ascend created the `torch_npu.profiler._ExperimentalConfig` with `aic_metrics=torch_npu.profiler.AiCMetrics.AiCoreNone`. That disables AiCore metric collection, so traces collected through `--profiler-config '{"profiler": "torch", ...}'` could miss the compute and pipeline utilization units needed for profiler analysis.

This patch changes the profiler AIC metric mode to `torch_npu.profiler.AiCMetrics.PipeUtilization`, allowing the generated profiling output to include pipeline utilization data while keeping the existing export type, profiler level, stack, memory, and trace handler behavior unchanged.

Yes. Users who enable Ascend torch profiling can now see compute and pipeline utilization information in the generated profiler output.

There is no CLI/API change.

- before this pr
image

- after this pr
image

- vLLM version: v0.22.1

### DIFF
--- a/tests/ut/profiler/test_torch_npu_profiler.py
+++ b/tests/ut/profiler/test_torch_npu_profiler.py
@@ -74,7 +74,7 @@ class TestTorchNPUProfilerWrapper(TestBase):
 
         mock_export_type.Text = "Text"
         mock_profiler_level.Level1 = "Level1"
-        mock_aic_metrics.AiCoreNone = "AiCoreNone"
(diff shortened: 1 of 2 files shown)


realliujiaxu (approved):
LGTM

PR #10781, opened 2026-06-22
[Refactor][BugFix] P-eagle and lmhead bugfix and Reduce sample reactor

lilinsiman:
When both the P-eagle and lmhead features are enabled, dummyrun is used for accompaniment, which causes token_indices_to_sample to be too long and exceed the value of max_num_reqs_across_dp. As a result, in lmhead, no padding is performed. Therefore, when P-eagle is enabled, an error is reported at position 1 exit. The solution is to pad the corresponding length to the initial length and restore the length when num_indices is too long.

no

UT and Tests

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/spec_decode/llm_base_proposer.py
+++ b/vllm_ascend/spec_decode/llm_base_proposer.py
@@ -1032,6 +1032,15 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
             )
 
         num_indices = token_indices_to_sample.shape[0]
+        if lmhead_tp_enable():
+            max_num_reqs_across_dp = (
+                self.vllm_config.scheduler_config.max_num_seqs * self.runner.uniform_decode_query_len
+            )
+            # It is necessary to evaluate the case where num_indices becomes large
+            # in the context of the dummy‑run accompaniment of p‑eagle.
+            if num_indices > max_num_reqs_across_dp:
+                ori_token_indices_to_sample = token_indices_to_sample
+
         if self.pcp_size > 1:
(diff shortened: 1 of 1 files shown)


slippersss (approved):

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -976,15 +976,19 @@ def _propose(
         return draft_token_ids
 
     def compute_draft_token_ids(self, hidden_states: torch.Tensor):
-        logits = self.model.logits_processor(self.model.lm_head, hidden_states)
-        if not hasattr(self.model, "draft_id_to_target_id") or self.model.draft_id_to_target_id is None:
+        if self.method in ("eagle3", "dflash"):
```
winson-00178005, line 979: ## `compute_draft_token_ids` 分支逻辑建议

```python
if self.method in ("eagle3", "dflash"):
    logits = self.model.logits_processor(self.model.lm_head, hidden_states)
    ...
else:
    logits = self.model.compute_logits(hidden_states)
    return greedy_sample(logits)
```

`compute_draft_token_ids` 根据 `self.method` 选择不同的 logits 计算路径，但 `eagle3/dflash` 走 `logits_processor(lm_head)`，其他走 `compute_logits`。这两条路径的语义差异（是否经过 quant 方法等）没有在代码或注释中说明。建议补充说明为什么 eagle3/dflash 需要走 `logits_processor` 而不是 `compute_logits`，以及两者在量化场景下的行为差异。

zzzzwwjj (approved):

PR #10804, opened 2026-06-22
适配Minimax m3

CXY-Katrina:

- vLLM version: v0.21.0


vllm_ascend/quantization/methods/w8a8_dynamic.py
```diff
@@ -53,6 +53,8 @@ class AscendW8A8DynamicLinearMethod(AscendLinearScheme):
     and per-channel quantization for weights.
     """
 
+    act_quant_type: torch.dtype = torch.int8
```
menogrey, line 56: Why adding this? Seems meaningless, this is a w8a8 quant method.

vllm_ascend/quantization/methods/w8a8_dynamic.py
```diff
@@ -77,7 +79,7 @@ def apply(
         bias: torch.Tensor | None = None,
         tp_rank: int | None = 0,
     ) -> torch.Tensor:
-        quantized_x, pertoken_scale = torch_npu.npu_dynamic_quant(x)
+        quantized_x, pertoken_scale = torch_npu.npu_dynamic_quant(x, dst_type=self.act_quant_type)
```
menogrey, line 82: When the dst_type is None, the default type is int8.

vllm_ascend/quantization/methods/w8a8_dynamic.py
```diff
@@ -115,7 +117,7 @@ def apply(
                 layer.weight,
                 layer.weight_scale,
                 pertoken_scale=pertoken_scale,
-                bias=bias,
+                bias=bias if self.act_quant_type == torch.int8 else None,
```
menogrey, line 120: self.act_quant_type will not be changed, so why adding this

vllm_ascend/quantization/methods/w8a8_dynamic.py
```diff
@@ -143,7 +145,8 @@ def process_weights_after_loading(self, layer):
             del layer.weight_offset
         else:
             # cast quantized weight tensors in NZ format for higher inference speed
-            layer.weight.data = maybe_trans_nz(layer.weight.data)
+            if self.act_quant_type == torch.int8:
```
menogrey, line 148: same, will we change self.act_quant_type?

vllm_ascend/quantization/modelslim_config.py
```diff
@@ -50,7 +50,15 @@
 
 # key: model_type
 # value: dict of fused module name -> list of original module names
+_MINIMAX_M3_PACKED_MODULES = {
```
menogrey, line 53: No need to use a var, just write as below.

vllm_ascend/quantization/modelslim_config.py
```diff
@@ -571,7 +579,10 @@ def get_quant_method(self, layer: torch.nn.Module, prefix: str, tid2eid=None) ->
             logger.debug("Select AscendFusedMoEMethod for %s (layer=%s)", prefix, "FusedMoE")
             return AscendFusedMoEMethod(scheme, layer.moe_config, tid2eid)
         elif isinstance(layer, VocabParallelEmbedding):
-            if self.is_layer_skipped_ascend(prefix, self.packed_modules_mapping):
+            if (
+                self.is_layer_skipped_ascend(prefix, self.packed_modules_mapping)
+                or f"{prefix}.weight" not in self.quant_description
```
menogrey: We real has this scenario? Can you give a model weight example? And if has this scenario, should we integrate it to the is_layer_skipped_ascend func?

PR #10816, opened 2026-06-23
[BugFix][WeightTransfer] Lazy-load weight transfer engines to avoid eager ray import

learning-sketch:

PR vllm-project#10592 (`[Feature] WeightTransfer: Add NPUIPCWeightTransferEngine backend for Ascend NPU`) passed all gates but broke the smoke test with:

```
ModuleNotFoundError: No module named 'ray'
```

**Root cause**

`vllm_ascend/patch/platform/patch_weight_transfer_engine.py` runs during *global plugin patching* — extremely early in startup, triggered even by `AsyncEngineArgs.add_cli_args(parser)` → `load_general_plugins()`. This happens long before any weight-transfer backend is selected.

The patch eagerly imported the engine classes at module top level:

```python
from vllm_ascend.distributed.weight_transfer.npu_ipc_engine import (
    NPUIPCWeightTransferEngine,
)
```

`npu_ipc_engine` transitively imports `vllm.distributed.weight_transfer.ipc_engine`, which does `import ray` at module top level. `ray` is an optional dependency, so when it is not installed the entire `vllm_ascend` plugin load aborts and **every** `vllm serve` invocation crashes — even workloads that never use weight transfer.

Import chain that crashed:

```
vllm_ascend/__init__.py:_ensure_global_patch()
  → utils.py:adapt_patch(is_global_patch=True)
    → patch/platform/__init__.py
      → patch_weight_transfer_engine.py
        → distributed/weight_transfer/npu_ipc_engine.py
          → vllm/distributed/weight_transfer/ipc_engine.py
            → import ray   # ModuleNotFoundError
```

**Fix**

Replace the eager top-level imports with lazy loader functions. `WeightTransferEngineFactory._registry[backend]` is a zero-arg callable that returns the engine class and is only invoked from `create_engine()` when the backend is actually requested. Deferring the import into those callables matches the factory's existing lazy-loading contract, so the `ray`-importing module is only loaded when the `ipc` backend is selected (at which point ray is genuinely needed):

```python
def _load_npu_ipc_engine() -> "type[WeightTransferEngine]":
    from vllm_ascend.distributed.weight_transfer.npu_ipc_engine import (
        NPUIPCWeightTransferEngine,
    )
    return NPUIPCWeightTransferEngine

WeightTransferEngineFactory._registry["ipc"] = _load_npu_ipc_engine
```

Only the `ipc` registration is made lazy, since the IPC engine is the one that transitively imports `ray`. The `nccl` → HCCL registration keeps its eager import, as `hccl_engine` has no optional-dependency import problem.

No. The `nccl`/`ipc` backend resolution behaves identically; only the timing of the underlying module import changes (deferred until the backend is actually used).

- `python -m py_compile` and `ruff check` / `ruff format --check` pass on the changed file.
- The fix removes the only eager import of the IPC engine on the global-patch path, so plugin loading no longer requires `ray`. The other registration site (`distributed/weight_transfer/__init__.py:register_engine`) already uses lazy string-based registration and was unaffected.

- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/patch/platform/patch_weight_transfer_engine.py
+++ b/vllm_ascend/patch/platform/patch_weight_transfer_engine.py
@@ -53,16 +53,12 @@ from typing import TYPE_CHECKING
 
 from vllm.distributed.weight_transfer.factory import WeightTransferEngineFactory
 
+from vllm_ascend.distributed.weight_transfer.hccl_engine import (
+    HCCLWeightTransferEngine,
+)
+
 if TYPE_CHECKING:
     from vllm.distributed.weight_transfer.base import WeightTransferEngine
-
-
-def _load_hccl_engine() -> "type[WeightTransferEngine]":
-    from vllm_ascend.distributed.weight_transfer.hccl_engine import (
-        HCCLWeightTransferEngine,
-    )
-
-    return HCCLWeightTransferEngine
 
 
 def _load_npu_ipc_engine() -> "type[WeightTransferEngine]":
@@ -73,5 +69,5 @@ def _load_npu_ipc_engine() -> "type[WeightTransferEngine]":
     return NPUIPCWeightTransferEngine
 
 
-WeightTransferEngineFactory._registry["nccl"] = _load_hccl_engine
+WeightTransferEngineFactory._registry["nccl"] = lambda: HCCLWeightTransferEngine
 WeightTransferEngineFactory._registry["ipc"] = _load_npu_ipc_engine
--- a/vllm_ascend/patch/platform/patch_weight_transfer_engine.py
+++ b/vllm_ascend/patch/platform/patch_weight_transfer_engine.py
@@ -49,14 +49,29 @@
 #   Remove this patch when upstream vllm relaxes the Literal type to str
 #   or provides an extension point for out-of-tree backends.
 
+from typing import TYPE_CHECKING
+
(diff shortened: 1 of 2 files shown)


vllm_ascend/patch/platform/patch_weight_transfer_engine.py
```diff
@@ -49,14 +49,29 @@
...
+if TYPE_CHECKING:
+    from vllm.distributed.weight_transfer.base import WeightTransferEngine
+
+
+def _load_hccl_engine() -> "type[WeightTransferEngine]":
+    from vllm_ascend.distributed.weight_transfer.hccl_engine import (
+        HCCLWeightTransferEngine,
+    )
+
+    return HCCLWeightTransferEngine
```
Ronald1995: according to your pr description, it seems like there is no problem with HCCLWeightTransferEngine

learning-sketch: > according to your pr description, it seems like there is no problem with HCCLWeightTransferEngine

Correct — The latest commit reverts the HCCL side to eager import.

Ronald1995 (approved):

zzzzwwjj (approved):

zzzzwwjj (approved):

PR #10825, opened 2026-06-23
[BugFix][SpecDecode] Fix low MTP acceptance rate for SFA+DSA_CP (MTP>1)

dsxsteven:
Fixes a degraded multi-token-prediction (MTP) / speculative-decoding acceptance rate on Ascend when num_speculative_tokens > 1 when enable dsa_cp for sfa models.

Root cause: Root cause: In the SFA v1 attention-metadata builder (vllm_ascend/attention/sfa_v1.py), the actual_seq_lengths_query / actual_seq_lengths_key buffers handed to lightning indexer shared the same memory buffer across all draft steps. With MTP > 1, the attention-metadata is overwritten and only the last step is kept, which leads to wrong inputs. Meanwhile, rotary cos/sin were also fetched from an out-dated cache while drafting.

Changes:

- vllm_ascend/attention/sfa_v1.py: pre-allocate per-draft-step buffers spec_actual_seq_lengths_query/key (one tensor per step, sized max_num_reqs * (num_speculative_tokens + 1) + 1) so each step owns independent, with no step overwrites another's metadata. Split the builder into build() (non-draft/verify path), build_for_drafting(draft_step, ...), and a shared _build(..., draft_step) that selects the correct per-step buffer; pass use_cache=False to get_cos_and_sin_mla while drafting so the rotary tables are recomputed for each step.
- vllm_ascend/spec_decode/llm_base_proposer.py: when a builder exposes build_for_drafting, route each draft step's attention-metadata build through it instead of calling build(0, ...) for every step, so the per-step buffers are actually used and cross-step aliasing is avoided.

Effect: correct per-step attention metadata and rotary embeddings → restored MTP/spec-decode acceptance rate for num_speculative_tokens > 1.

No

- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/attention/sfa_v1.py
+++ b/vllm_ascend/attention/sfa_v1.py
@@ -200,6 +200,11 @@ class AscendSFAMetadataBuilder(MLACommonMetadataBuilder[AscendSFAMetadata]):
 
         self.speculative_config = vllm_config.speculative_config
         self.decode_threshold = 1
+        max_num_reqs = vllm_config.scheduler_config.max_num_seqs
+        self.actual_seq_lengths_query = torch.zeros(max_num_reqs + 1, dtype=torch.int32, device=device)
+        self.actual_seq_lengths_key = torch.empty_like(self.actual_seq_lengths_query)
+        self.spec_actual_seq_lengths_query: list[torch.Tensor] | None = None
+        self.spec_actual_seq_lengths_key: list[torch.Tensor] | None = None
         if self.speculative_config:
             spec_token_num = self.speculative_config.num_speculative_tokens
             self.decode_threshold += spec_token_num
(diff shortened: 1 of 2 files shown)


vllm_ascend/attention/sfa_v1.py
```diff
@@ -240,6 +250,22 @@ def build(
...
+    ) -> AscendSFAMetadata:
+        # common_prefix_len / fast_build are unused; kept for API compatibility.
+        return self._build(common_attn_metadata, draft_step=None)
+
+    def build_for_drafting(
+        self,
+        draft_step: int,
+        common_attn_metadata: AscendCommonAttentionMetadata,
+        **kwargs,
+    ) -> AscendSFAMetadata:
```
slippersss, line 262: Vllm has the same interface in the base class `AttentionMetadataBuilder`, but its signature is (`self`, `common_attn_metadata`, `draft_index`). I feel ok about the current implementation just like what dsa_v1 does, but if we can align to vllm it would be better, just a suggestion.

slippersss:
LGTM.

vllm_ascend/attention/sfa_v1.py
dsxsteven: I've discussed this with the maintainer of MTP, and the interface names will be uniformly refactored in the future.

slippersss (approved):

zzzzwwjj (approved):

PR #10842, opened 2026-06-23
[BugFix] Fix Error when GLM5 MTP layer has BF16 weights

dsxsteven:

- Root Cause:

  o_proj_full_pool is a class-level attribute of AscendSFAImpl, so a single buffer is shared across every model instance (main model + MTP drafters) and is allocated only once — during the main model's process_weights_after_loading, using self.o_proj.weight.dtype. In the full_gather_o_proj_enabled path this pool is passed explicitly as the output of all_gather_async. The helper only auto-allocates when output is None; when an output is supplied it is used as-is with no dtype check. So the all-gather output is forever locked to the main model's dtype.

  The MTP drafter is a separate model instance carrying a different quantization dtype (w8a8 → int8, bf16 → bfloat16). When the drafter runs the same all-gather, dist.all_gather_into_tensor sees output.dtype (int32) != input.dtype (int8 / bfloat16) and aborts:

  RuntimeError: output tensor must have the same type as input tensor

- Fix:

  Stop passing the shared, fixed-dtype pool as the all-gather output. Let all_gather_async allocate the output internally from input.dtype, then assign the new tensor back to the shared pool for the later set_():

  With output=None, the helper runs output = torch.empty(..., dtype=input.dtype, ...), so output.dtype == input.dtype is guaranteed regardless of which instance is running — main model (int32), w8a8 drafter (int8), or bf16 drafter (bfloat16). The dtype mismatch is eliminated at the source rather than worked around, which is why the same change fixes all non-matching drafter dtypes at once.

NO

- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/attention/sfa_v1.py
+++ b/vllm_ascend/attention/sfa_v1.py
@@ -1255,9 +1255,8 @@ class AscendSFAImpl(MLAAttentionImpl):
                         if is_hidden_layer(layer):
                             reach_layer_for_shard_weight_series(layer)
                 elif full_gather_o_proj_enabled:
-                    _, o_proj_full_handle = all_gather_async(
-                        self.o_proj_tp_weight, get_tp_group(), output=AscendSFAImpl.o_proj_full_pool
-                    )
+                    pool, o_proj_full_handle = all_gather_async(self.o_proj_tp_weight, get_tp_group())
+                    AscendSFAImpl.o_proj_full_pool = pool
 
                 if kv_cache is not None:
                     assert fused_kv_no_split is not None


vllm_ascend/attention/sfa_v1.py
```diff
@@ -1255,9 +1255,8 @@ def forward(
                         if is_hidden_layer(layer):
                             reach_layer_for_shard_weight_series(layer)
                 elif full_gather_o_proj_enabled:
-                    _, o_proj_full_handle = all_gather_async(
-                        self.o_proj_tp_weight, get_tp_group(), output=AscendSFAImpl.o_proj_full_pool
-                    )
+                    pool, o_proj_full_handle = all_gather_async(self.o_proj_tp_weight, get_tp_group())
```
winson-00178005, line 1258: ## 代码检视

修复逻辑正确：将 `o_proj_full_pool` 从 `all_gather_async` 的固定 dtype output 参数中移除，改为让 helper 自行分配匹配 input dtype 的 buffer，然后赋值回共享 pool。

### 🟡 潜在风险：多实例并发写入 `o_proj_full_pool`

`AscendSFAImpl.o_proj_full_pool` 是类级别属性，所有 model 实例共享。修改后每次 `all_gather_async` 调用都会将新分配的 tensor 赋值回 `AscendSFAImpl.o_proj_full_pool = pool`。如果不同 model 实例（main model + MTP drafters）在多 stream 并发场景下同时执行这段代码，可能出现竞态写入。

请确认在 PP+MTP 场景下，main model 和 drafter 的 SFA forward 是否在同一 stream 上串行执行，或者是否存在并发执行的可能性。如果存在并发风险，建议使用实例级别属性而非类级别属性。

winson-00178005, line 1258: ## 🟢 修复正确，建议补充详细说明

修复思路清晰：停止传递共享的固定dtype pool，让 `all_gather_async` 内部按 input.dtype 分配 output，避免 dtype mismatch。

但 `pool, o_proj_full_handle = all_gather_async(...)` 的修改意味着 `AscendSFAImpl.o_proj_full_pool` 从静态类属性变成了动态实例属性（每次 all_gather 后重新赋值）。这可能导致：
1. **并发安全问题**：如果多个实例同时运行 all_gather，o_proj_full_pool 可能被交替覆盖。请确认在当前架构中是否存在并发执行 all_gather 的场景。
2. **内存生命周期**：原来 pool 在类级别只分配一次，现在每次 all_gather 可能创建新 tensor。请确认旧的 pool tensor 是否会被正确释放，避免内存泄漏。

建议在代码或 PR description 中补充这些影响的分析。

PR #10854, opened 2026-06-23
[BugFix][v0.22.1] Fix Error when GLM5 MTP layer has BF16 weights

dsxsteven:
🔄 Backport of https://github.com/vllm-project/vllm-ascend/pull/10842 to releases/v0.22.1rc.

- Root Cause:

o_proj_full_pool is a class-level attribute of AscendSFAImpl, so a single buffer is shared across every model instance (main model + MTP drafters) and is allocated only once — during the main model's process_weights_after_loading, using self.o_proj.weight.dtype. 
In the full_gather_o_proj_enabled path this pool is passed explicitly as the output of all_gather_async. The helper only auto-allocates when output is None; when an output is supplied it is used as-is with no dtype check. So the all-gather output is forever locked to the main model's dtype.

The MTP drafter is a separate model instance carrying a different quantization dtype (w8a8 → int8, bf16 → bfloat16). When the drafter runs the same all-gather, dist.all_gather_into_tensor sees output.dtype (int32) != input.dtype (int8 / bfloat16) and aborts:

RuntimeError: output tensor must have the same type as input tensor

- Fix:

Stop passing the shared, fixed-dtype pool as the all-gather output. Let all_gather_async allocate the output internally from input.dtype, then assign the new tensor back to the shared pool for the later set_():

With output=None, the helper runs output = torch.empty(..., dtype=input.dtype, ...), so output.dtype == input.dtype is guaranteed regardless of which instance is running — main model (int32), w8a8 drafter (int8), or bf16 drafter (bfloat16). The dtype mismatch is eliminated at the source rather than worked around, which is why the same change fixes all non-matching drafter dtypes at once.

No

- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/attention/sfa_v1.py
+++ b/vllm_ascend/attention/sfa_v1.py
@@ -1250,9 +1250,8 @@ class AscendSFAImpl(MLAAttentionImpl):
                         if is_hidden_layer(layer):
                             reach_layer_for_shard_weight_series(layer)
                 elif full_gather_o_proj_enabled:
-                    _, o_proj_full_handle = all_gather_async(
-                        self.o_proj_tp_weight, get_tp_group(), output=AscendSFAImpl.o_proj_full_pool
-                    )
+                    pool, o_proj_full_handle = all_gather_async(self.o_proj_tp_weight, get_tp_group())
+                    AscendSFAImpl.o_proj_full_pool = pool
 
                 if kv_cache is not None:
                     assert fused_kv_no_split is not None


winson-00178005:
## 代码检视意见

**检视维度**: 可读性/可维护性/可测试性/安全/可靠/高效/可移植

| 维度 | 评价 | 说明 |
|------|------|------|
| **可读性** | ✅ 好 | 修复简洁，将`all_gather_async`返回值正确解构为`(pool, handle)` |
| **可维护性** | ✅ 好 | 变更仅2行，符合patch模式 |
| **可测试性** | ⚠️ 中 | 无新增UT，但修复逻辑清晰 |
| **安全** | ✅ 好 | 无安全风险 |
| **可靠** | ✅ 好 | 修复原代码忽略pool返回值导致o_proj_full_pool可能为None的问题 |
| **高效** | ✅ 好 | 正确获取pool避免重复分配 |
| **可移植** | ✅ 好 | 纯逻辑修复 |

**总评**: ✅ **建议合并** - 关键BugFix，修复all_gather_async返回值处理错误

PR #9704, opened 2026-05-29
[CI]fix dflash test

zhao-stack:
### What this PR does / why we need it?
The upstream PR of vllm https://github.com/vllm-project/vllm/pull/40727 fixed the dflash acceptance rate issue, which caused inconsistency between the acceptance rate and the use case. This PR modified the use case baseline.

### Does this PR introduce _any_ user-facing change?
None
### How was this patch tested?

- vLLM version: v0.20.2
- vLLM main: https://github.com/vllm-project/vllm/commit/39910f2b25aacc09f5e7f166cdf0030b19f8b9e8

PR #10858, opened 2026-06-24
[Feature][Worker] Apply structured-output bitmask on device in the v1 path

Aeolion-mu:

In the vLLM v1 path, `NPUModelRunner.sample_tokens` applied the structured-output (guided decode) grammar bitmask via a per-step **device→CPU→device round-trip**: it copied the full-vocab logits to host, ran the xgrammar CPU scan (`apply_grammar_bitmask`), then copied the masked logits back. On Ascend 910B this costs ~6ms/step at batch 16 (D2H ~1.9ms + CPU scan ~2.9ms + H2D ~1.3ms) and scales with batch × vocab.

This PR wires the **existing NPU Triton kernel** `_apply_grammar_bitmask_kernel` (already used by the v2 path) into v1 so the packed `int32` bitmask is applied **in-place on the device logits, in their native dtype, with no host transfer**. A small reusable helper `apply_grammar_bitmask(...)` builds the bitmask-row→logits-row index map — mirroring the upstream device branch: it honors the reordered structured-output subset (the guided requests are a subset of the batch, in a different order) and the `1 + num_spec_tokens` row expansion under speculative decoding — and launches the kernel on a compacted bitmask (row `i` is the mask for `logits_indices[i]`).

The kernel's one-time ~5s Triton JIT is **warmed up at startup** from the worker's `compile_or_warm_up_model` (called in both eager and graph modes), so it does not stall the first guided-decode request; a once-guard makes the additional `capture_model` warmup a no-op. The warmup allocates its dummy logits in the model's **real logits dtype** (e.g. bf16) — Triton caches the kernel per logits-pointer dtype, so a float32 warmup would leave the bf16 entry cold and the JIT would still land on the first real request.

**Performance** (Ascend 910B, Qwen3-8B, xgrammar backend, eager): guided-decode throughput **+18.8% tok/s (134.5 → 159.8)**, per-step mask-apply **~6ms → ~1.2ms**, with **byte-identical output** vs the prior CPU path. The guided-vs-plain throughput penalty is roughly halved (-29% → -16%); plain (non-guided) throughput is unchanged.

**Correctness**: verified on NPU against the prior CPU path on identical real tensors — the device path produces an identical `-inf` masked-position set and finite-allclose kept logits across multi-request reordered batches, speculative-decode row expansion (including simultaneous multi-request different spec counts), and mixed batches where non-grammar rows must stay untouched. Mask densities up to ~911k positions, never a mismatch.

**Notes for reviewers** (transparency, no behavior change):
- *Helper layering*: the shared helper currently lives in `vllm_ascend/worker/v2/structured_outputs.py` (next to the kernel) and is imported by the v1 runner, to avoid duplicating the kernel and keep the diff minimal. Happy to relocate it to a neutral module if the v1→v2 import is undesirable.
- *Padded-vocab tail*: if a model's sampler logits are padded **wider** than the bitmask vocab, this kernel masks the padded tail columns to `-inf`, whereas the old CPU path left them untouched. This is benign for valid sampleable tokens and is not triggered on the tested models (Qwen3 vocab 151936 == bitmask vocab). Worth revisiting only if a model surfaces sampleable padded columns.

This extends the device kernel (landed via #9704, used by v2) to the v1 path that PR left on the CPU round-trip.

No. Behavior is unchanged for users — guided-decode output is byte-identical to the previous CPU path. The change is a performance optimization (and a startup warmup) internal to the NPU v1 model runner. No new config, env var, or API.

- **Unit tests added** (`tests/ut/worker/v2/test_structured_outputs.py`, CPU-comparable, no NPU/triton required for the logic under test): the bitmask-row→logits-row index/reorder math (`build_grammar_bitmask_indices`) across in-order subset, reordered subset, spec-decode row expansion, request-absent-from-batch alignment, and the empty case; plus a numpy CPU reference of the kernel's masking semantics including the int32 bit-31 sign-bit boundary. 6/6 pass.
- **NPU end-to-end** (Ascend 910B3, Qwen3-8B, xgrammar, eager): device path vs the upstream CPU path on the SAME pre-mask logits — identical masked-position sets + finite-allclose kept logits across mixed reordered batches, ngram spec-decode (1+num_spec rows, incl. simultaneous multi-request different spec counts), and non-grammar rows verified untouched. Before/after throughput on an identical harness (batch 16, byte-identical output): guided +18.8% tok/s; per-step mask-apply ~6ms → ~1.2ms. Cold-Triton-cache run confirms the bf16-dtype warmup removes the first-step JIT stall (first guided step ~1.15ms instead of ~5.3s).

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -91,7 +91,6 @@ from vllm.v1.sample.metadata import SamplingMetadata
 from vllm.v1.sample.rejection_sampler import PLACEHOLDER_TOKEN_ID, RejectionSampler
 from vllm.v1.spec_decode.metadata import SpecDecodeMetadata
 from vllm.v1.spec_decode.ngram_proposer_gpu import copy_num_valid_draft_tokens
-from vllm.v1.structured_output.utils import apply_grammar_bitmask
 from vllm.v1.utils import record_function_or_nullcontext
 from vllm.v1.worker import mamba_utils
 from vllm.v1.worker.cp_utils import (
@@ -160,6 +159,10 @@ from vllm_ascend.utils import (
 from vllm_ascend.worker.npu_input_batch import NPUInputBatch
 from vllm_ascend.worker.pcp_utils import PCPManager
 from vllm_ascend.worker.utils import AscendKVBlockZeroer
+from vllm_ascend.worker.v2.structured_outputs import (
+    apply_grammar_bitmask as apply_grammar_bitmask_npu,
+)
+from vllm_ascend.worker.v2.structured_outputs import warmup_grammar_bitmask_kernel
 
 from vllm_ascend.ascend_forward_context import (  # isort: skip
     MoECommType,
@@ -486,6 +489,13 @@ class NPUModelRunner(GPUModelRunner):
         self.decode_threshold = 1 + (self.speculative_config.num_speculative_tokens if self.speculative_config else 0)
 
         self.use_aclgraph = self._use_aclgraph()
+
+        # One-time guard so the structured-output bitmask Triton kernel only pays
+        # its ~5s JIT compile once (warmed up during graph capture, see
+        # capture_model). Structured outputs are enabled per request, so we
+        # cannot reliably gate this on a global config flag; the warmup is a
+        # cheap one-shot no-op kernel launch.
+        self._grammar_bitmask_kernel_warmed = False
 
         eplb_config = self.ascend_config.eplb_config
         self.dynamic_eplb = eplb_config.dynamic_eplb
@@ -2418,12 +2428,14 @@ class NPUModelRunner(GPUModelRunner):
 
         # Apply structured output bitmasks if present.
         if grammar_output is not None:
-            # here we are different from gpu_model_runner,
-            # the apply_grammar_bitmask uses torch.compile to optimize this,ascend does not support it now
-            logits_dtype = logits.dtype
-            logits = logits.to("cpu").float()
-            apply_grammar_bitmask(scheduler_output, grammar_output, self.input_batch, logits)
-            logits = logits.to(self.device).to(logits_dtype)
+            # Apply the bitmask in-place on the device logits via the NPU Triton
+            # kernel. This avoids the device->CPU->device round-trip the upstream
+            # util incurs on Ascend (its device branch routes through a
+            # torch.compile path that NPU does not support); logits stay on the
+            # NPU in their native dtype.
+            apply_grammar_bitmask_npu(
+                logits, scheduler_output, grammar_output, self.input_batch
+            )
 
         with record_function_or_nullcontext("sample_token"):
             sampler_output = self._sample(logits, spec_decode_metadata)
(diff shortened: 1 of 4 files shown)


vllm_ascend/worker/v2/structured_outputs.py
```diff
@@ -66,3 +80,133 @@ def _apply_grammar_bitmask_kernel(
...
+        scheduler_output.scheduled_spec_decode_tokens,
+    )
+    if not out_indices:
+        return
+
+    # Gather only the applied rows so bitmask row i aligns with out_indices[i].
+    # The common case (all structured-output requests present and contiguous)
+    # gathers every row, which numpy returns as a cheap copy.
+    selected_bitmask = grammar_output.grammar_bitmask[bitmask_rows]
+    bitmask = torch.from_numpy(selected_bitmask).to(logits.device, non_blocking=True)
```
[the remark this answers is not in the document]
Aeolion-mu: Good catch - applied `torch.as_tensor` in 50e97b8. Thanks for the review!

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -160,6 +159,10 @@
 from vllm_ascend.worker.npu_input_batch import NPUInputBatch
 from vllm_ascend.worker.pcp_utils import PCPManager
 from vllm_ascend.worker.utils import AscendKVBlockZeroer
+from vllm_ascend.worker.v2.structured_outputs import (
```
winson-00178005, line 170: ## 🟡 跨模块导入建议

`model_runner_v1.py` 从 `vllm_ascend/worker/v2/structured_outputs` 导入 helper，这在 v1 和 v2 runner 之间创建了依赖关系。PR description 中也提到了这一点。建议考虑将这些共享 helper（`build_grammar_bitmask_indices`, `apply_grammar_bitmask`, `warmup_grammar_bitmask_kernel`）提取到一个更中性的位置，例如 `vllm_ascend/ops/structured_outputs.py` 或 `vllm_ascend/utils/structured_outputs.py`，避免 v1→v2 的交叉依赖。

如果当前设计是合理的临时方案，请至少在导入处添加注释说明为什么从 v2 模块导入，以及后续是否有迁移计划。

vllm_ascend/worker/worker.py
```diff
@@ -789,6 +789,13 @@ def compile_or_warm_up_model(self) -> CompilationTimes:
         # may cause performance degradation at runtime.
         if get_ascend_device_type() != AscendDeviceType.A5:
             self._warm_up_atb()
+        # Warm up the structured-output bitmask kernel here (always called, in
+        # both eager and graph modes) so its one-time Triton JIT does not stall
+        # the first guided-decode request. capture_model() also warms it but only
+        # runs when graphs are captured (not enforce_eager); the once-guard makes
```
winson-00178005, line 795: ## 🟡 `worker.py` 中直接调用 model_runner 的内部方法

```python
if hasattr(self.model_runner, "_maybe_warmup_grammar_bitmask_kernel"):
    self.model_runner._maybe_warmup_grammar_bitmask_kernel()
```

使用 `hasattr` + 直接调用 `_maybe_warmup_grammar_bitmask_kernel`（前缀 `_` 表示内部方法）的方式不够优雅。建议在 model_runner 上提供一个公开的 warmup 接口（如 `warmup_kernels()`），避免 worker 直接访问内部方法。这也有利于后续其他 kernel warmup 的统一管理。

vllm_ascend/worker/model_runner_v1.py
Aeolion-mu: Thanks for flagging the v1->v2 coupling. The bitmask Triton kernel is actually v2-native: it predates this PR and is also imported by vllm_ascend/patch/worker/patch_v2/patch_triton.py, so fully hoisting it into a neutral ops/ module would also move pre-existing v2 code and widen this PR's scope. To keep this PR focused I kept the helpers in place and added a NOTE at the v1 import explaining the dependency, plus a TODO to hoist them into vllm_ascend/ops/structured_outputs.py in a dedicated follow-up. Applied in 4998c7e.

vllm_ascend/worker/worker.py
Aeolion-mu: Agreed, thanks. I exposed it as a public method maybe_warmup_grammar_bitmask_kernel (dropping the underscore) and updated the worker to call the public name instead of the name-mangled internal one. I considered the generic warmup_kernels() you suggested, but vllm already defines a warmup_kernels in v1/worker/gpu/warmup.py, so I kept a specific public name to avoid shadowing/confusion. Applied in 4998c7e.

Aeolion-mu:
Hi @MengqingCao @wangxiyuan — friendly ping on this one. All review feedback so far (Gemini bot's `torch.as_tensor` suggestion, winson-00178005's public-API rename for the warmup hook) has been addressed in 4998c7e. Let me know if there's anything else blocking review — happy to make further changes. Thanks for your time!

PR #10865, opened 2026-06-24
[BugFix] add logic when KVCacheSpec group has compress ratio > 1

KlyzhenkoVadim:

DeepSeek-V4-Flash or self_attn and indexer_k_cache group (when compress_ratio=4) max_num_block_per_req changing incorrect because
```py
if (kv_cache_group is not None
    and hasattr(kv_cache_group, "kv_cache_spec")
    and hasattr(kv_cache_group.kv_cache_spec, "compress_ratio")
     ):
         compress_ratio = kv_cache_group.kv_cache_spec.compress_ratio
```
is unreachable. I changed the logic.
No

- vLLM version: v0.22.1


MengqingCao (approved):
LGTM, thx for this fix!

PR #10873, opened 2026-06-24
[Test] Enable default graph mode for test case test_guided_json_completion

zhiyu-wa:
For the test case `test_guided_json_completion`, use the default graph mode `FULL_AND_PIECEWISE`

No

Ran `test_guided_json_completion` locally with the change applied and verified it passes.

- vLLM version: v0.22.1


realliujiaxu (approved):
LGTM

PR #10878, opened 2026-06-24
[BugFix][SpecDecode] Fix low MTP acceptance rate for SFA+DSA_CP (MTP>1)

dsxsteven:
Fixes a degraded multi-token-prediction (MTP) / speculative-decoding acceptance rate on Ascend when num_speculative_tokens > 1 when enable dsa_cp for sfa models.

Root cause: Root cause: In the SFA v1 attention-metadata builder (vllm_ascend/attention/sfa_v1.py), the actual_seq_lengths_query / actual_seq_lengths_key buffers handed to lightning indexer shared the same memory buffer across all draft steps. With MTP > 1, the attention-metadata is overwritten and only the last step is kept, which leads to wrong inputs. Meanwhile, rotary cos/sin were also fetched from an out-dated cache while drafting.

Align Ascend attention builders' build_for_drafting with vllm's base convention (common_attn_metadata, draft_index) so all spec-decode draft paths go through one consistent entry point.

- sfa_v1.py: split build() into build()/build_for_drafting()/_build(); pre-allocate per-draft-step spec_actual_seq_lengths_query/key buffers so each draft step owns independent metadata (no cross-step aliasing); recompute rotary (use_cache=False) while drafting.
- llm_base_proposer.py: drop the else->build(0,...,model) branch; route all draft steps through build_for_drafting(common_attn_metadata, draft_index, **extra). Standard-attention drafters (eagle/eagle3/mtp on Qwen/Llama) now hit the base build_for_drafting -> build(0, m, fast_build=True), fixing the 'int' object has no attribute 'num_reqs' crash from the swapped-positional elif. Rename draft_step -> draft_index.
- dsa_v1.py / dsa_cp.py: swap build_for_drafting arg order to (common_attn_metadata, draft_index, ...) and rename draft_step -> draft_index for consistency.

No

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/attention/dsa_v1.py
+++ b/vllm_ascend/attention/dsa_v1.py
@@ -1122,8 +1122,8 @@ class AscendDSAMetadataBuilder(AttentionMetadataBuilder[AscendDSAMetadata]):
 
     def build_for_drafting(
         self,
-        draft_step: int,
         common_attn_metadata: AscendCommonAttentionMetadata,
+        draft_index: int,
         fast_build: bool = False,
         **kwargs,
     ) -> AscendDSADecodeMetadata:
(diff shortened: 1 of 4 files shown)


MengqingCao (approved):

vllm_ascend/attention/dsa_v1.py
```diff
@@ -1136,19 +1136,19 @@ def build_for_drafting(
         if num_prefills:
```
YzTongNiar, line 1136: like sfa_v1's implementation, a single line can cover differenr scenario  'cos, sin = get_cos_and_sin_mla(input_positions, use_cache=(draft_index is None))'

PR #10886, opened 2026-06-24
[Test] Limit DeepSeek R1 longseq accuracy prompts

wzx0726:

This PR adds num_prompts: 360 to the DeepSeek-R1-W8A8 longseq nightly accuracy benchmark config.

The original benchmark may timeout because it runs too many samples from the dataset. This change limits the number of evaluated prompts to 360, reducing the nightly benchmark runtime while still keeping enough samples for accuracy validation.

Changes:

Add num_prompts: 360 to the acc benchmark case.
increase `batch_size` from 16 to 32 to lower benchmark pressure.
Keep dataset_path, request_conf, dataset_conf, max_out_len, baseline, and threshold unchanged.

No. This PR only updates an internal nightly benchmark configuration and does not introduce any user-facing API, interface, or behavior change.

## How was this patch tested?

Not run locally. This change only updates the nightly benchmark configuration.

The updated benchmark is expected to be validated by the nightly CI.

- vLLM version: v0.23.0


realliujiaxu (approved):
LGTM

PR #10937, opened 2026-06-25
[Test] add AOP failure interception and bisect pipeline for single/multi-node testsFeature bridge bisect

czydyy:
Adds an AOP (Aspect-Oriented Programming) failure interception and bisect pipeline for single-node and multi-node nightly E2E tests. When a test fails, the pipeline:

1. Classify: Scans pod logs for environment error patterns (rules-env.txt) to distinguish environment issues from real code bugs
2. Check commit age: Looks up the last successful run in good_table.csv; skips bisect for failures older than 3 days
3. Trigger bisect: Calls auto_bisect for recent, non-environment failures to identify the first bad commit
4. Update good_table: On success, updates the shared good_table.csv with the current commit info
Both single-node (GHA workflow steps) and multi-node (in-pod aop_pipeline()) share the same decision logic.

No. All AOP features are disabled by default (aop_single_enabled: false, aop_multi_enabled: 'false'). Enable by passing aop_single_enabled: true or aop_multi_enabled: 'true' from the calling workflow.

- Multi-node AOP pipeline tested on A2 cluster with Kimi-K2_5-W4A8-A2-dual-nodes: verified classify (env/not_env), commit age lookup, worker-leader synchronization
- Single-node AOP pipeline tested with Qwen3.5-27B-w8a8-A2: verified capture → classify → check age → trigger bisect flow
- CSV table lookup tested with good_table.csv: name matching, success row filtering, date parsing
- Worker pod log collection verified via PVC shared storage

- vLLM version: v0.23.0


kunpengW-code:
Please check the feedback provided by Gemini above.

zhangxinyuehfad (approved):

realliujiaxu (approved):
LGTM

Issue #10500, opened 2026-06-15
[Bug]: 含有VIT的多模态大模型开启FlashComm1和FlashComm2报错


wangyichao1999:
### Your current environment

PyTorch version: 2.9.0+cpu
Torch-npu version: 2.9.0
Is debug build: False
OS: Ubuntu 22.04.5 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
Clang version: 15.0.7
CMake version: version 4.3.2
Libc version: glibc-2.35

Python version: 3.11.14 (main, Feb 26 2026, 04:48:50) [GCC 11.4.0] (64-bit runtime)
Python platform: Linux-5.15.0-89-generic-x86_64-with-glibc2.35
Is XNNPACK available: True

CPU:
Architecture:                       x86_64
CPU op-mode(s):                     32-bit, 64-bit
Address sizes:                      46 bits physical, 57 bits virtual
Byte Order:                         Little Endian
CPU(s):                             192
On-line CPU(s) list:                0-175
Off-line CPU(s) list:               176-191
Vendor ID:                          GenuineIntel
Model name:                         Intel(R) Xeon(R) Platinum 8468
CPU family:                         6
Model:                              143
Thread(s) per core:                 2
Core(s) per socket:                 48
Socket(s):                          2
Stepping:                           8
BogoMIPS:                           4200.00
[...]

PR #10941, opened 2026-06-25
[BugFix] Skip FlashComm1/2 For ViT

wangyichao1999:
Fix VIT crash when enabling FC1/FC2 on Ascend NPU for VL models (e.g. Step3p7):
  #10500

  vLLM precomputes inputs_embeds outside forward_context for multimodal models, causing VIT layers to be incorrectly routed by FC1/FC2 dispatch to context-dependent ops. Since the context is not yet set, _EXTRA_CTX is empty and FC1/FC2 flags are unavailable → crash.

  Add "vision_model" not in prefix guards on three dispatch paths:
  - SP column-parallel dispatch (SequenceColumnParallelOp)
  - SP row-parallel dispatch (SequenceRowParallelOp)
  - FlashComm2 row-parallel dispatch (Flashcomm2OProjRowParallelOp)

No user-facing changes. VL models crashed on startup with FC1/FC2 enabled before the fix; after the fix they run correctly.Text-only models are unaffected (their prefixes do not contain vision_model).

Step3p7 VL model, 8× Ascend NPU, TP=8, FC1+FC2+MTP all enabled. Before the fix, the VIT forward phase crashed. After the fix, both PD disaggregated and PD colocated deployments start correctly and pass stress testing.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/ops/linear_op.py
+++ b/vllm_ascend/ops/linear_op.py
@@ -695,7 +695,7 @@ def _get_column_parallel_op(
             "g_proj",  # attention gate projection of Step3p5
         ]
         for a_prefix in sp_column_prefix:
-            if a_prefix in prefix:
+            if a_prefix in prefix and "vision_model" not in prefix:
                 return SequenceColumnParallelOp(layer)
 
     return None
@@ -725,7 +725,8 @@ def _get_row_parallel_op(
         return MatmulAllreduceRowParallelOp(layer)
     if flashcomm2_enable():
         if "o_proj" in prefix or "out_proj" in prefix:
-            return Flashcomm2OProjRowParallelOp(layer)
+            if "vision_model" not in prefix:
+                return Flashcomm2OProjRowParallelOp(layer)
     if enable_sp():
         # "share_expert" added for Step3p5
         if "shared_expert" in prefix or "share_expert" in prefix:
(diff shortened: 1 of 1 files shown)


tests/ut/ops/test_linear.py
```diff
@@ -168,5 +169,94 @@ def test_init_without_disable_tp(self):
         self.assertTrue(isinstance(linear.quant_method, AscendUnquantizedLinearMethod))
 
 
+class TestVisionModelParallelOpExclusion(unittest.TestCase):
+    """FC1/FC2 (SequenceParallel / FlashComm2) must be skipped for ViT
```
zhenwenqi2024: What functionality does this UT cover? Don't add this kind of code.  I think it is unnecessary

vllm_ascend/ops/linear_op.py
```diff
@@ -651,7 +651,7 @@ def _get_column_parallel_op(
             "query_key_value",  # qkv linear of Bailing
         ]
         for a_prefix in sp_column_prefix:
-            if a_prefix in prefix:
+            if a_prefix in prefix and "vision_model" not in prefix:
```
zhenwenqi2024, line 698: Is vision_model universal across all vit models?

realliujiaxu:
change title to "Skip FlashComm1/2 For ViT"

realliujiaxu (approved):
LGTM

vllm_ascend/ops/linear_op.py
wangyichao1999: vision_model is not universal, about 20+ models

PR #10945, opened 2026-06-25
[Feature]:Modify the UvaBufferWrapper class to enable UVA features

orangeBoxun:

GPUs have the UVA feature, which allows the GPU side to directly access CPU tensors based on the same virtual address, without manual copying. The NPU side also needs to follow up on this.

This feature needs to be enabled through os.environ['PYTORCH_NPU_ALLOC_CONF'] = 'pinned_mem_register:True'; otherwise, it is disabled by default.

Enable os.environ['PYTORCH_NPU_ALLOC_CONF'] = 'pinned_mem_register:True', and then use the regular vllm test cases.

- vLLM version: v0.23.0

### DIFF
--- a/.github/workflows/scripts/test_config.yaml
+++ b/.github/workflows/scripts/test_config.yaml
@@ -392,6 +392,7 @@
     - tests/ut/test_compressed_prefix_cache.py
     - tests/e2e/pull_request/one_card/test_qwen3_0_6b.py
     - tests/e2e/pull_request/one_card/test_qwen3_5_0_8b.py
+    - tests/e2e/pull_request/one_card/model_runner_v2/test_uva.py
 
 - name: patch_mamba_utils_310p
   optional: false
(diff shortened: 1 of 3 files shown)


vllm_ascend/patch/worker/patch_v2/patch_uva.py
```diff
@@ -16,13 +16,46 @@
...
+    # Target version that disables UVA feature
+    DISABLE_UVA_VERSION = "3.2.1"
+    installed_version = version("triton-ascend")
+    # Check if current version is the one with UVA disabled
+    if installed_version == DISABLE_UVA_VERSION:
+        warning_msg = (
+            f"WARNING: triton-ascend {DISABLE_UVA_VERSION} disables the UVA feature.\n"
+            f"Related bug issue: https://github.com/triton-lang/triton-ascend/issues/783"
+        )
+        print(warning_msg, file=sys.stderr)
```
weijinqian0: Plz use log.

weijinqian0 (approved):

.github/workflows/scripts/test_config.yaml
```diff
@@ -675,6 +675,7 @@ estimated_times:
   tests/e2e/pull_request/one_card/lora/test_qwen3_multi_loras.py: 120
   tests/e2e/pull_request/one_card/lora/test_qwen3_reranker_lora.py: 170
   tests/e2e/pull_request/one_card/model_runner_v2/test_basic.py: 770
+  tests/e2e/pull_request/one_card/model_runner_v2/test_uva.py: 770
```
kunpengW-code: UT trigger rule synchronization modification

PR #10947, opened 2026-06-25
[Bugfix] reject mooncakeV1 c8 consumer KVcache for GQA backend

zzzzzmeng:

Reject MooncakeConnector startup when C8 KV cache quantization is enabled on a KV consumer node with GQA attention backend, because the producer keeps bf16 KV cache while the consumer allocates int8 cache. Add a regression test for the unsupported configuration.

No

UT test

- vLLM version: v0.23.0


linfeng-yuan:
Please make sure whether kvcahce c8 is not supported in all model scenarios.

zzzzzmeng:
> Please make sure whether kvcahce c8 is not supported in all model scenarios.



> 请确认 kvcahce c8 是否在所有模型场景中都受支持。

It has been confirmed that this restriction is only needed for GQA architectures.

linfeng-yuan (approved):

PR #10956, opened 2026-06-25
[BugFix] Pad actual_seq_lengths with 0 for EOD scenario in npu_fusion_attention

bowgneo:
On CANN 9.0, a strict validation was added for EOD in npu_fusion_attention. The actual_seq_qlen and actual_seq_kvlen parameters must end with 0 to pass the validation.
In the current fuse_attention path, when handling EOD scenarios, these parameters may not end with 0, causing npu_fusion_attention call to fail on CANN 9.0.

#### EOD:
When the last batch item's actual sequence length does not account for all remaining tokens — specifically when:
T_q > actualSeqQLen[-1] (sum of actual_seq_qlen is less than total query tokens)
T_kv > actualSeqKvLen[-1] (sum of actual_seq_kvlen is less than total kv tokens)
This means there are trailing tokens that belong to an "end-of-document" padding region beyond the described sequences. For the npu_fusion_attention operator, this is the EOD scenario — the actual_seq_qlen and actual_seq_kvlen arrays must explicitly end with a 0 to signal the end of document to the CANN runtime.

NO

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -1195,6 +1195,10 @@ class AscendAttentionBackendImpl(AttentionImpl):
         _: torch.Tensor,
     ) -> torch.Tensor:
         # use default sparse_mode 0 in normal scenario, which means no mask works on it
+        # Pad actual_seq_len with 0 when num_tokens > actual_seq_len in TND layout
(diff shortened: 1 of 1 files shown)


realliujiaxu (approved):
LGTM

PR #10976, opened 2026-06-25
[feature] ai接入冒烟问题定位

xiaoshudian555:

- vLLM version: v0.22.1


.agents/skills/ci-ai-diagnosis-agent/references/output-format.md
```diff
@@ -0,0 +1,304 @@
...
+    "missing_sources": [],
+    "source_conflicts": [],
+    "matched_playbooks": ["log-diagnosis-vllm-inference-timeout"]
+  },
+  "failure_family": "vllm_engine_runtime_timeout",
+  "root_cause": "Inference batch timed out due to prefill queue bloat",
+  "classification": "product_bug",
+  "confidence": "medium",
+  "root_layers": [
+    {"layer": "表象", "cause": "EngineDeadError at CI summary"},
```
zzzzwwjj: Please change to English.

.agents/skills/ci-ai-diagnosis-agent/scripts/ci_diagnosis_agent.py
zzzzwwjj, line 1: Where should the tools called by skills be placed? @maoxx241

vllm_ascend/envs.py
```diff
@@ -110,6 +110,43 @@
     # Control the aclrtMemcpyBatchAsync compile path for KV cache offloading.
     # "1": force enable, "0": force disable, None: auto-detect from CANN headers.
     "VLLM_ASCEND_ENABLE_BATCH_MEMCPY": lambda: os.getenv("VLLM_ASCEND_ENABLE_BATCH_MEMCPY", None),
+    # --- CI AI diagnosis agent (optional; see
```
zzzzwwjj: Please check these newly added env. @wangxiyuan

wangxiyuan: these envs are totally useless for vllm-ascend. If you want to use/set these env. Just use `os.env` in your script. No need to add them here.

.agents/skills/ci-ai-diagnosis-agent/SKILL.md
```diff
@@ -0,0 +1,313 @@
...
+  "question": "Which benchmark task failed and by how much?",
+  "tool": "get_benchmark_summary",
+  "expected_fact": "failed task, metric, value, baseline, threshold",
+  "fallback": "If benchmark results are missing, record missing_source and inspect main log around aisbench output"
+}
+```
+
+Use source-specific tools before generic search:
+
+1. Benchmark outcome: `get_benchmark_summary()` first, then `search()` for
```
maoxx241, line 120: Thanks for spelling out the source routing model. One thing that may need tightening: this section asks the agent to call source-specific tools such as `get_benchmark_summary()`, `get_k8s_summary()`, `search_artifacts()`, and `get_wrapper_upstream_context()`, but the current function-calling loop only exposes `list_dir`, `read_file`, `search`, and `diagnose`. That mismatch can make the documented workflow hard for the model to follow. Could we either add these source-specific tools or adjust the skill instructions to match the actual tool surface?

maoxx241 (commented):
Thanks for the work here. I left a few inline comments around the skill/runtime contracts that would be good to align before we rely on this in CI.

.agents/skills/ci-ai-diagnosis-agent/scripts/ci_diagnosis_agent.py
```diff
@@ -0,0 +1,1639 @@
...
+
+# ---------------------------------------------------------------------------
+# Rendering
+# ---------------------------------------------------------------------------
+
+
+def render_markdown(diag: dict[str, Any]) -> str:
+    routing = diag.get("routing") or {}
+    lines: list[str] = []
+    lines.append("## CI AI Diagnosis")
```
maoxx241, line 1382: The renderer currently emits a compact summary, but `references/output-format.md` defines the Markdown report as an 11-section contract and notes that downstream tools may match on those headings. It would be good to make the implementation and the documented report contract consistent before merging. The no-LLM path also seems to return `type: evidence_only`, while the reference doc describes an `evidence_bundle` shape, so that contract may need the same alignment.

.agents/skills/ci-ai-diagnosis-agent/scripts/ci_diagnosis_agent.py
```diff
@@ -0,0 +1,1639 @@
...
+import os
+import subprocess
+import sys
+import urllib.error
+import urllib.request
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import regex as re
```
maoxx241, line 41: This script now depends on the third-party `regex` package. Some jobs may already have it through test dependencies, but the multi-node workflow invokes this helper from the `nightly-cpu` container without an explicit dependency install step. To keep the CI helper robust, could we either switch to Python's built-in `re` module here, or make the runtime dependency explicit before the script is called?

.github/workflows/_e2e_nightly_single_node.yaml
```diff
@@ -271,3 +279,69 @@ jobs:
...
+          SHA=$(git rev-parse HEAD)
+          python3 "$GITHUB_WORKSPACE/.agents/skills/ci-ai-diagnosis-agent/scripts/ci_diagnosis_agent.py" \
+            --log-file "$LOGFILE" \
+            --step-name "${{ inputs.name || inputs.config_file_path || inputs.tests }}" \
+            --artifact-dir /tmp/ci-diagnosis-artifacts \
+            --benchmark-dir /tmp/benchmark_results \
+            --ref "${{ inputs.vllm_ascend_ref || github.ref_name }}" \
+            --sha "$SHA" \
+            --repo-dir "$GITHUB_WORKSPACE" \
+            --output-json /tmp/ci-diagnosis-evidence.json \
```
maoxx241, line 346: This writes a diagnosis JSON file, but the single-node workflow does not upload it afterward. The multi-node workflow uploads its diagnosis evidence bundle, so the artifact behavior is currently inconsistent. If this JSON is intended for post-mortem debugging, could we upload it here as well (and similarly in the single-node-models workflow)? If the GitHub step summary is the only intended output for single-node jobs, we can avoid writing the JSON to reduce confusion.

zzzzwwjj (approved):

PR #10991, opened 2026-06-26
[Feature] Fix Mooncake HMA KV spec group mapping

liziyu179:

This PR fixes Mooncake KV transfer group mapping for HMA scenarios where the target model and Eagle draft model share one KV cache manager group but use different KV specs.

In the MLA + QGA case, the main model has 1 KV head while the Eagle model has 8 KV heads. With P/D TP set to tp8/tp4, treating both specs as one transfer group causes incorrect TP pull mapping. This PR makes Mooncake transfer metadata distinguish KV specs by KV head count while preserving the mapping back to the original KV cache manager group.

It also fixes P-side delayed-free rank calculation so the prefill worker computes the union of prefill ranks that may be pulled by any decode rank, instead of reusing D-side group pull logic.

No user-facing API change.

This fixes internal KV transfer behavior for HMA + MLA main model + QGA Eagle model deployments.

## How was this patch tested?

Unit tests updated:

```bash
$env:PYTHONPATH='C:\Users\34754\Desktop\liziyu\vllm;' + (Get-Location).Path + ';' + $env:PYTHONPATH
$env:VLLM_VERSION='0.22.1'
pytest -q tests\ut\kv_offload\test_mooncake_connector.py

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/kv_offload/test_mooncake_connector.py
+++ b/tests/ut/kv_offload/test_mooncake_connector.py
@@ -14,6 +14,7 @@ import msgspec
 import torch
 import zmq
 from vllm.utils.network_utils import make_zmq_path
+from vllm.v1.kv_cache_interface import FullAttentionSpec, UniformTypeKVCacheSpecs
 
 fake_engine = types.ModuleType("mooncake.engine")
 fake_engine.TransferEngine = MagicMock()  # type: ignore[attr-defined]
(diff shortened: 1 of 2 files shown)


wangxiaoteng888 (approved):

linfeng-yuan:
Please check whether head_dim influences.

Yuli-yx:
Thanks for pointing this out. For the current supported models, grouping by spec type and num_kv_heads is sufficient, as we don't have layers with the same num_kv_heads but different incompatible KV cache dimensions/layouts. We've added a todo to document this assumption and flag the future extension point. Plz check if this is appropiate.

MengqingCao (approved):
I'm okay with taking different head_dim into account in a follow-up pr, let's merge this quickly.

PR #10992, opened 2026-06-26
[Doc] Increased the restrictions on the transformer version

wangzhishenghw:
In the Single-Node Online Deployment section, users encounter transformer version issues when starting the service with vllm serve. Although it was mentioned in the Introduction section, some users may not have noticed it, so it is reminded again in the Single-Node Online Deployment section.
Added transformers version limit prompt and version upgrade command

- vLLM version: v0.23.0


wangxiyuan:
https://github.com/vllm-project/vllm-ascend/blob/main/requirements.txt#L38 it'll be installed by vllm-ascend by default. Why add the note here again?

wangzhishenghw:
> https://github.com/vllm-project/vllm-ascend/blob/main/requirements.txt#L38 it'll be installed by vllm-ascend by default. Why add the note here again?

Because some users have encountered problems, a reminder is given before the code where problems may occur.

MengqingCao:
> > https://github.com/vllm-project/vllm-ascend/blob/main/requirements.txt#L38 it'll be installed by vllm-ascend by default. Why add the note here again?
> 
> Because some users have encountered problems, a reminder is given before the code where problems may occur.

Then I think we could add a note for this, instead of adding this step as a required one.

wangzhishenghw:
> > > https://github.com/vllm-project/vllm-ascend/blob/main/requirements.txt#L38 it'll be installed by vllm-ascend by default. Why add the note here again?
> > 
> > 
> > Because some users have encountered problems, a reminder is given before the code where problems may occur.
> 
> Then I think we could add a note for this, instead of adding this step as a required one.

The two added lines have comments

MengqingCao:
> > > > https://github.com/vllm-project/vllm-ascend/blob/main/requirements.txt#L38 it'll be installed by vllm-ascend by default. Why add the note here again?
> > > 
> > > 
> > > Because some users have encountered problems, a reminder is given before the code where problems may occur.
> > 
> > 
> > Then I think we could add a note for this, instead of adding this step as a required one.
> 
> The two added lines have comments

okay, I'll merge it then, thx!

MengqingCao (approved):

PR #11021, opened 2026-06-26
Revert "[KV Cache][Feature] Support Layerwise KV Pooling (#10077)"

ader47:
This PR reverts #10077.
PR #10077 introduced new functionality for MooncakeLayerwiseConnector. However, this change causes compatibility issues in some model scenarios.
For example, in models such as Qwen3.5, attn_metadata can be a dict whose keys are layer_name. In attention_v1.py, this variable may default to ascend_metadata, which can cause the new layerwise connector path to handle the attention metadata incorrectly and lead to runtime errors.
Since #10077 contains a relatively large set of changes around Layerwise KV Pooling and MooncakeLayerwiseConnector, this PR reverts the entire change set instead of applying a partial fix.

- vLLM version: v0.23.0


wangxiyuan:
wait CI before merging

wangxiyuan (approved):

PR #11044, opened 2026-06-27
[Misc] Add PCP reorder debug logs for long-sequence scenarios

wzx0726:

This PR adds lightweight debug logs for PCP long-sequence DFX in `pcp_utils.py`.

The logs focus on reorder-related variables in the PCP path, including:

* PCP token split and padding information
* PCP-local positions and allgather positions
* `pcp_allgather_restore_idx`
* hybrid attention FA reorder metadata, such as `pcp_enter_fa_restore_idx`, `pcp_exit_fa_scatter_idx`, and `pcp_fa_query_idx`
* hidden states restore shape information after PCP allgather

These debug logs are intended to provide better observability for long-sequence issues, especially when investigating whether token positions, padding masks, allgather restore indices, or hidden states reorder logic are incorrect.

To avoid excessive logs in long-sequence scenarios, this PR only prints compact summaries such as shape, dtype, device, and small samples of numpy arrays. It does not dump full token sequences or full tensors.

No.

This PR only adds debug-level logs for DFX purposes and does not change APIs, interfaces, model behavior, scheduling behavior, or default runtime behavior.

This patch was tested by running the PCP long-sequence path and verifying that the newly added debug logs can be emitted correctly when debug logging is enabled.

The added logs are debug-only and do not affect normal execution when debug logging is disabled. No new unit tests were added because this change only improves internal debug observability and does not modify functional behavior.

- vLLM version: v0.23.0


vllm_ascend/worker/pcp_utils.py
```diff
@@ -1307,6 +1330,23 @@ def generate_pcp_metadata(
                     long_seq_metadata.pcp_enter_fa_restore_idx = self.pcp_enter_fa_restore_idx[
                         : pcp_unpad_mask.sum() + self.num_decode_tokens * (self.pcp_world_size - 1)
                     ]
+                    
+                    if logger.isEnabledFor(logging.DEBUG):
```
weiguihua2, line 1321: Can these three prints be printed together?

PR #11050, opened 2026-06-27
[BugFix][Platform] Scope MiniMax usage accounting patch

QwertyJack:

This PR narrows the MiniMax reasoning-token usage-accounting backport so it fixes vllm-project/vllm#37988 without routing other models through the MiniMax request path.

Changes:
- move the chat usage accounting backport into `patch_minimax_usage_accounting.py` and remove the generic `patch_chat_usage_accounting.py` import/file
- keep the global OpenAI usage schema backfill needed to serialize `completion_tokens_details.reasoning_tokens`
- bind full/stream chat usage wrappers only for MiniMax reasoning parser instances
- avoid wrapping `OpenAIServingChat.__init__`, so unrelated parser initialization errors do not point at the MiniMax patch
- preserve class/default `reasoning_parser_cls` access when using the descriptor-based MiniMax instance binding
- keep MiniMax parser wrapper handling and skip non-MiniMax parsers instead of returning fake reasoning-token values

This preserves the MiniMax fix while avoiding the broad per-request overhead that caused performance regression on other models.

Yes. MiniMax chat completions continue to report non-negative `completion_tokens_details.reasoning_tokens`. Non-MiniMax models should no longer be routed through the MiniMax usage-accounting wrapper.

On this `main` branch checkout:

```bash
ruff check vllm_ascend/patch/platform/patch_minimax_usage_accounting.py tests/ut/patch/platform/test_patch_minimax_usage_accounting.py vllm_ascend/patch/__init__.py vllm_ascend/patch/platform/__init__.py
ruff format --check vllm_ascend/patch/platform/patch_minimax_usage_accounting.py tests/ut/patch/platform/test_patch_minimax_usage_accounting.py vllm_ascend/patch/__init__.py vllm_ascend/patch/platform/__init__.py
python3 -m py_compile vllm_ascend/patch/platform/patch_minimax_usage_accounting.py tests/ut/patch/platform/test_patch_minimax_usage_accounting.py vllm_ascend/patch/__init__.py vllm_ascend/patch/platform/__init__.py
git diff --check
```

Result: passed.

Direct descriptor smoke test also passed:
- class-level `OpenAIServingChat.reasoning_parser_cls` returns the preserved default, not the descriptor object
- untouched instances return the preserved default
- non-MiniMax instances do not bind stream/full wrappers
- MiniMax instances bind the MiniMax usage wrapper

I also attempted:

```bash
python3 -m pytest -q tests/ut/patch/platform/test_patch_minimax_usage_accounting.py
python3 -m pytest -q tests/ut/patch/platform/test_patch_glm_tool_call_streaming.py tests/ut/patch/platform/test_patch_glm47_tool_call_parser.py tests/ut/patch/platform/test_patch_minimax_usage_accounting.py
```

Both failed during test collection before reaching these tests because the local active vLLM runtime is `/vllm-workspace/vllm`, which does not contain `vllm.model_executor.layers.minimax_rms_norm`, while current `upstream/main` imports that module from `vllm_ascend/patch/worker/patch_minimax_m2_linear_attn.py` through `tests/ut/conftest.py`.

- vLLM version: v0.23.0


vllm_ascend/patch/platform/patch_minimax_usage_accounting.py
```diff
@@ -43,3 +90,372 @@ def _patched_count_reasoning_tokens(self, token_ids: Sequence[int]) -> int:
...
+        instance.__dict__["_ascend_reasoning_parser_cls"] = value
+        if _is_minimax_reasoning_parser_cls(value):
+            _patch_chat_usage_instance(instance)
+
+
+if not isinstance(
+    getattr(OpenAIServingChat, "reasoning_parser_cls", None),
+    _ReasoningParserClsDescriptor,
+):
+    OpenAIServingChat.reasoning_parser_cls = _ReasoningParserClsDescriptor()
```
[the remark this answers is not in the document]
QwertyJack: Fixed in `8056127e9`: `_ReasoningParserClsDescriptor` now stores the original default value, class-level and untouched-instance access return that default instead of the descriptor object, and the idempotence check uses `OpenAIServingChat.__dict__` so re-imports do not reinstall the descriptor accidentally. Added `test_reasoning_parser_cls_descriptor_preserves_default_access` to cover this.

MINGJING-WU:
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/28344935652)

vllm_ascend/patch/platform/patch_minimax_usage_accounting.py
```diff
@@ -43,3 +90,373 @@ def _patched_count_reasoning_tokens(self, token_ids: Sequence[int]) -> int:
...
+    reasoning_parser,
+) -> int | None:
+    if reasoning_parser is None:
+        return None
+    reasoning_token_counts = [
+        _count_minimax_reasoning_tokens_for_usage(token_ids, reasoning_parser) for token_ids in raw_output_token_ids
+    ]
+    if all(reasoning_tokens is None for reasoning_tokens in reasoning_token_counts):
+        return None
+    return sum(reasoning_tokens or 0 for reasoning_tokens in reasoning_token_counts)
```
Lyr-GW, line 225: The list comprehension result is iterated twice — once for all() and once for sum():

```python
reasoning_token_counts = [
    _count_minimax_reasoning_tokens_for_usage(token_ids, reasoning_parser)
    for token_ids in raw_output_token_ids
]
if all(reasoning_tokens is None for reasoning_tokens in reasoning_token_counts):
    return None
return sum(reasoning_tokens or 0 for reasoning_tokens in reasoning_token_counts)
```

In practice n is small for chat completions so this has zero real-world impact. If you wanted to clean it up anyway, a single-pass loop avoids the double scan:

```python
total = 0
any_valid = False
for token_ids in raw_output_token_ids:
    rt = _count_minimax_reasoning_tokens_for_usage(token_ids, reasoning_parser)
    if rt is not None:
        any_valid = True
        total += rt
return total if any_valid else None
```

Feel free to ignore this — purely stylistic.

QwertyJack: Thanks for checking. I’ll keep this as-is for now since `n` is small in chat completions, this only runs on the MiniMax-scoped usage path, and the current form is straightforward. I’d prefer to avoid extra churn in this backport.

linfeng-yuan (approved):
LGTM

PR #11059, opened 2026-06-27
[Refactor][Attn] Reuse on-device query_start_loc, drop redundant per-forward H2D

Aeolion-mu:

`AscendAttentionMetadataBuilder.build` re-copies `query_start_loc` to the device on every forward via `query_start_loc_cpu.pin_memory().to(self.device, non_blocking=True)`, even though the synced device tensor is already available as `common_attn_metadata.query_start_loc` (the on-device counterpart of the same `CpuGpuBuffer` as `query_start_loc_cpu`, set together in `AscendCommonAttentionMetadata`). This resolves the existing in-code TODO ("Yet another unnecessary H2D while we already have a query_start_loc on device") by reusing the on-device tensor and dropping the redundant per-forward H2D copy. `query_start_loc_cpu` is kept (still used for `actual_seq_lengths_q`).

No. Internal cleanup; behavior is unchanged.

## How was this patch tested?

Verified on Qwen3-8B (single card, greedy): generated token ids are bit-identical (sha256 match) before and after, and the per-forward `pin_memory().to(device)` H2D call count on this path drops from 48 to 0 over the same run.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -309,8 +309,9 @@ class AscendAttentionMetadataBuilder(AttentionMetadataBuilder[AscendMetadata]):
         # Get attn_mask from singleton AttentionMaskBuilder
         attn_mask = self.attn_mask_builder.get_attention_mask(common_attn_metadata.causal, self.model_config)
 
-        # TODO: Yet another unnecessary H2D while we already have a query_start_loc on device
-        query_start_loc = query_start_loc_cpu.pin_memory().to(self.device, non_blocking=True)
+        # Reuse the already-on-device query_start_loc (a synced view of the same
+        # CpuGpuBuffer as query_start_loc_cpu) instead of re-doing a per-forward H2D copy.
+        query_start_loc = common_attn_metadata.query_start_loc[: num_reqs + 1]
 
         attn_metadata = AscendMetadata(
             num_actual_tokens=num_actual_tokens,


Aeolion-mu:
Hi @weijinqian0 @whx-sjtu — friendly ping on this one when you have a chance. It's a small, bit-identical drive-by fix for the redundant per-forward H2D noted in a maintainer TODO (reuses the existing device `query_start_loc`). Let me know if anything needs adjusting. Thanks!

Ruiqiu-Zheng:
Independent current-main validation from 2026-09-12: the same `AscendAttentionMetadataBuilder.build` TODO/redundant `query_start_loc_cpu.pin_memory().to(self.device, non_blocking=True)` seam is still present on upstream main `e11eaf0c78fa79813ea74d5c64be11abe95bf485`.

I also ran a bounded component check of the exact replacement used by this PR (reuse the already-populated device `query_start_loc` slice):
- exact integer/dtype/device/shape/contiguity equivalence: 7/7 cardinalities `N={1,2,63,64,65,128,129}`; `N=65` includes the B64 + dummy/padded-request case;
- B64, exactly 50 paired AB/BA observations: current duplicate-H2D median `154.7585 us`, existing-device-slice median `49.641 us`, paired median reduction `105.127 us` (`67.93%`), candidate faster `50/50`.

Claim ceiling is intentionally component-level only: this does not establish TPOT/E2E or production gain, and it does not replace focused current-main integration coverage. But it independently confirms that the source seam is still live and that the redundant copy remains materially expensive at the component level.

PR #11065, opened 2026-06-27
[BugFix] Avoid allocating KV cache for skipped indexers

ZYang6263:
## What this PR does

GLM-5.2 reuses top-k indices on most backbone layers, so those layers no longer initialize a local Indexer. The model runner still described and allocated an Indexer KV cache for every sparse MLA layer, wasting device memory and underestimating the number of KV blocks that fit in the available memory.

During KV cache grouping, the Ascend MLA merge path also accepted the 704- and 576-dimension layouts as identical and expanded every layer back to the larger layout. This change keeps those layouts heterogeneous through block planning. This PR builds a smaller per-layer KV cache spec for sparse MLA layers that do not own an Indexer.

## How was this patch tested?

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -160,6 +160,7 @@ from vllm_ascend.utils import (
     set_potential_max_tokens,
     set_weight_prefetch_method,
     should_skip_allreduce_across_dp_group,
+    sparse_kv_cache_has_indexer,
     vllm_version_is,
 )
 from vllm_ascend.worker.npu_input_batch import NPUInputBatch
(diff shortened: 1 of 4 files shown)


linfeng-yuan (approved):

linfeng-yuan:
@MengqingCao  Please have a look.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -4005,6 +3978,11 @@ def _get_attention_kv_cache_dims(self, layer_name: str, kv_cache_spec: Attention
         head_size_v = kv_cache_spec.head_size_v if hasattr(kv_cache_spec, "head_size_v") else kv_cache_spec.head_size
         return kv_cache_spec.head_size, head_size_v
 
+    @staticmethod
+    def _sparse_kv_cache_has_indexer(kv_cache_spec: AttentionSpec) -> bool:
```
MengqingCao: it is more reasonable to move this function to utils.py

ZYang6263: done!

PR #11068, opened 2026-06-27
[BugFix][Model] Remove redundant SFA o_proj TP weights

maoxx241:

This PR fixes the SFA v1 DSA-CP mixed `o_proj` path so the original TP shard remains the only persistent source of truth.

SFA DSA-CP mixed execution already couples DSA-CP with the `o_proj` TP path. This PR keeps that behavior and documents the intended storage invariant instead of adding a new user-facing switch:

- Decode-only batches exchange SFA outputs with TP all-to-all, then run the original TP-sharded `o_proj`.
- Prefill/mixed batches temporarily all-gather the TP-sharded `o_proj` weight and input-sharded quantization parameters into full-weight buffers before the `o_proj` forward.
- `o_proj_tp_*` tensors alias the original parameter storage instead of cloning it.
- `o_proj_full_*` tensors remain temporary gather destinations for the prefill/mixed path only.
- Input-sharded `o_proj` quant params are discovered by `input_dim == 1` instead of hard-coded parameter names.
- The DSA-CP mixed `o_proj` data path is documented in `docs/source/developer_guide/Design_Documents/context_parallel.md`.
- Load-time `.contiguous()` is kept for W4A4 MXFP4 and W8A8 MXFP8 linear transpose paths that feed NPU quant matmul. W8A8 restore now uses `reshape` instead of `view` for the restored scale shape so the contiguous transformed scale can still be restored for weight loading.

The previous clone-based setup made the mixed path look like two persistent `o_proj` weights. This patch keeps the existing runtime behavior but removes the redundant persistent TP copy and records the intended storage invariant in the design doc and code comments.

No. There is no new config switch or public API change. The existing SFA DSA-CP mixed behavior remains coupled with the `o_proj` TP path.

Static/unit checks run in an A5 environment with this PR checkout:

```bash
python -m pytest tests/ut/attention/test_sfa_o_proj_tp.py -q
# 2 passed

python -m pytest \
  tests/ut/quantization/methods/test_w4a4_mxfp4.py \
  tests/ut/quantization/methods/test_w8a8_mxfp8.py -q
# 18 passed

python -m ruff check \
  vllm_ascend/attention/sfa_v1.py \
  vllm_ascend/quantization/methods/w4a4_mxfp4.py \
  vllm_ascend/quantization/methods/w8a8_mxfp8.py \
  tests/ut/attention/test_sfa_o_proj_tp.py \
  tests/ut/quantization/methods/test_w4a4_mxfp4.py \
  tests/ut/quantization/methods/test_w8a8_mxfp8.py

python -m ruff format --check \
  vllm_ascend/attention/sfa_v1.py \
  vllm_ascend/quantization/methods/w4a4_mxfp4.py \
  vllm_ascend/quantization/methods/w8a8_mxfp8.py \
  tests/ut/attention/test_sfa_o_proj_tp.py \
  tests/ut/quantization/methods/test_w4a4_mxfp4.py \
  tests/ut/quantization/methods/test_w8a8_mxfp8.py

python -m py_compile vllm_ascend/quantization/methods/w8a8_mxfp8.py
```

NPU validation:

| Case | Checkpoint | TP | DSA-CP | Loading model weights | Available / current KV cache |
| --- | --- | ---: | --- | ---: | ---: |
| W4A4 baseline | GLM-5.1-W4A4C8-mxfp4-rot | 8 | off | 52.3663 GB | 27.59 GiB / 27.59 GiB |
| W4A4 DSA-CP | GLM-5.1-W4A4C8-mxfp4-rot | 8 | on | 57.6788 GB | 22.26 GiB / 22.26 GiB |

The W4A4 before/after DSA-CP comparison keeps the same TP size, model length, batch-token limit, quantization mode, and memory utilization. The DSA-CP run has higher persistent weight accounting than the non-DSA baseline because DSA-CP changes the attention-side layout beyond `o_proj`; the PR removes the redundant persistent `o_proj` TP clone within that mixed path.

- vLLM version: v0.23.0

### DIFF
--- a/docs/source/developer_guide/Design_Documents/context_parallel.md
+++ b/docs/source/developer_guide/Design_Documents/context_parallel.md
@@ -121,6 +121,25 @@ By predefining the maximum amount of KV cache processed per round, we sequential
 
 ![PCP-ChunkedPrefill](../../assets/cp/chunkedprefill.png)
 
+### SFA DSA-CP Mixed `o_proj` Path
+
+SFA DSA-CP mixed execution intentionally reuses the normal TP-sharded `o_proj`.
+This is part of the DSA-CP mixed data path, not a standalone user-facing `o_proj` TP switch.
+The mixed path is used when one instance may handle both decode-only and prefill/mixed batches, so `o_proj` must support two layouts at runtime:
+
+- **Decode-only batches** keep the decode TP path.
+  SFA outputs are exchanged with an all-to-all in the TP group, then the original TP-sharded `o_proj` runs normally.
+- **Prefill or mixed batches** produce SFA outputs that are not directly compatible with the TP-sharded `o_proj` input layout.
+  Before `o_proj` forward, each rank all-gathers the TP-sharded `o_proj` weight and all input-sharded quantization parameters into temporary full-weight buffers.
+  The full-weight `o_proj` forward runs once for that batch, and the module is then restored to the TP parameter aliases.
+
+The storage invariant is that the original TP-sharded `o_proj` parameter remains the only persistent source of truth.
+`o_proj_tp_*` tensors are aliases of the original parameter storage.
+`o_proj_full_*` tensors are reusable communication buffers for prefill/mixed full-gather execution only.
+They must not become a second persistent copy of the TP weight.
+
+This coupling preserves the existing decode TP behavior, supports prefill/mixed DSA-CP batches, and avoids adding an extra configuration path whose state can drift from DSA-CP mixed execution.
+
 ### Related Files
 
 - slot_mapping computation: `vllm_ascend/worker/block_table.py`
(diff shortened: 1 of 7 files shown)


vllm_ascend/quantization/methods/w4a8_mxfp4.py
```diff
@@ -91,9 +91,9 @@ def process_weights_after_loading(self, layer):
         layer.weight.data = torch_npu.npu_format_cast(
             layer.weight.data, 29, customize_dtype=torch.float8_e4m3fn, input_dtype=torch_npu.float4_e2m1fn_x2
         )
-        layer.weight.data = layer.weight.data.transpose(-1, -2)
+        layer.weight.data = layer.weight.data.transpose(-1, -2).contiguous()
```
lijiahang226: The weight has already been converted to NZ format and can no longer undergo contiguous operation.

maoxx241: Thanks for pointing this out. Fixed by removing `.contiguous()` after the NZ format cast on the weight.

zzzzwwjj (approved):

PR #11086, opened 2026-06-29
[Refactor][Distributed] Remove all_to_all patch and switch lm-head TP to all_to_all_single

xuchi-0808:

Removes the `GroupCoordinator.all_to_all` monkey-patch from vllm-ascend and switches all remaining consumers to a new `lmhead_all_to_all` helper that uses `torch.distributed.all_to_all_single`, returning to the upstream PyTorch standard interface.

**Background** — the old path required two layers of patch:
```
consumer → get_lmhead_tp_group().all_to_all()   ← GroupCoordinatorPatch.all_to_all
        → device_communicator.all_to_all()      ← NPUCommunicator.all_to_all
        → dist.all_to_all(list version)          ← real HCCL call
```

**Changes:**
- Delete `GroupCoordinatorPatch.all_to_all` and `NPUCommunicator.all_to_all`.
- Add `lmhead_all_to_all(logits, comm_group)` helper in `vocab_parallel_embedding.py`, used by all 3 consumers:
  - `vocab_parallel_embedding.py:_get_logits_lmheadtp`
  - `spec_decode/llm_base_proposer.py:_run_merged_draft` (mrope + normal branches)

**The redistribution `[N, V/P] → [N/P, V]`** (all tokens × partial vocab → partial tokens × full vocab) is done with a `view + all_to_all_single + permute + view` sequence:
```
[N, V/P] -> view(P, N/P, V/P) -> all_to_all_single -> permute(1,0,2) -> view(N/P, V)
```
This is mathematically equivalent to the removed patch's `tensor_split(dim=0) + all_to_all(list) + cat(dim=-1)`, but keeps a single contiguous buffer for better HCCL fusion. Microbenchmarks on Ascend 910 (P=2/4, realistic lm-head shapes) show **1.2x–1.4x speedup** over the list version.

**Why not a bare `all_to_all_single`?** A naive `all_to_all_single` only redistributes on dim 0 (token) and never gathers the vocab dimension — it silently produces `[N, V/P]` instead of `[N/P, V]`. The `view` reshapes the input so dim 0 carries the per-rank token shard, letting the single collective route correctly; `permute` then interleaves the vocab shards of each token. (Thanks to the reviewer who caught this — the initial draft had this bug.)

**Safety:** equal split is guaranteed because `pad_vocab_size` + `divide` make the vocab shard `V/P` identical on every rank. `world_size == 1` returns the input unchanged.

Also removes the patch registration comment block in `patch/__init__.py` and cleans up 4 affected test files.

No. The behavior of lm-head tensor parallel is unchanged; the refactor only replaces the internal communication call with an equivalent (and faster) one. No API, config, or model interface changes.

- **Unit tests (Ascend NPU):** All lm-head-TP-related UTs pass, including `test_run_merged_draft_mtp_mrope_graph_and_lmhead_tp_preparation` which exercises the mtp + lmhead_tp path. (`test_vocab_parallel_embedding.py`, `test_fused_moe.py`, `test_patch_distributed.py`, `test_eagle_proposer.py`.)
- **Correctness microbench (2-rank gloo + 2/4-rank HCCL):** Verified the `view + all_to_all_single + permute` output is element-wise identical to the old list-based `tensor_split + all_to_all + cat` on representative shapes.
- **Performance microbench (Ascend 910, HCCL):** view version is 1.2x–1.4x faster than the list version across P=2/4 × decode/prefill shapes.
- **Lint:** `ruff check` clean.

- vLLM version: v0.25.1


xuchi-0808:
Closing in favor of a fresh PR rebased on latest main. The original commit was based on a 591-commit-old base; the new PR keeps the same branch name (`A00283`) and same design, with conflicts resolved against recent main changes (notably preserving the new `_NpuAll2AllManager` for vllm 0.25.x MoE fault-tolerance compatibility, and adapting to the `_apply_head` version-dispatch abstraction).

PR #11088, opened 2026-06-29
[BugFix] fix dsv32 nightly

Toneymiller:
fix dsv32 nightly：The original batch size is too small, which causes the performance test to time out. Therefore, the batch size is increased to 256.
no
nightly test
- vLLM version: v0.23.0


zhangxinyuehfad:
/nightly multi-node-dpsk3.2-2node
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/28350121932)

linfeng-yuan (approved):

PR #11109, opened 2026-06-29
[Doc] Add ascend-vllm serving auto-benchmark and serving-tune-loop skills

Jonxjdong:

Adds two new Claude Code skills under `.agents/skills/` and registers them in `.agents/README.md`:

- **`ascend-vllm-serving-auto-benchmark`**: automated 5-phase benchmark workflow for vLLM-Ascend serving on Ascend 910C (env/config validation, server launch, evalscope perf stress test, metrics collection, report generation). Ships pre-built configs for DeepSeek-V3, DeepSeek-V4-Flash, Qwen3-32B, Qwen3.5-27B, Qwen3.5-35B-A3B.
- **`ascend-vllm-serving-tune-loop`**: autonomous 7-phase performance tuning loop (RLCR principle, one lever per iteration, fixed benchmark contract, ledger discipline for crash-recovery, auto-resume wrapper for cross-session continuation, final synthesis report).
- **`.agents/README.md`**: add "Serving Auto Benchmark" and "Serving Tune Loop" sections with TOC entries, matching the existing skill documentation style.

These two skills form a pair: the tune-loop skill establishes its baseline using the same benchmarking contract as the auto-benchmark skill, and references its prerequisites. Submitting them together keeps the cross-skill reference self-consistent.

No. Adds agent skill definitions and docs only; no runtime code changes.

- Skill files reviewed for internal consistency (cross-skill prerequisite reference, file layout, invocation names match directory names).
- No executable code paths added to the vllm-ascend package; skills are invoked by Claude Code, not imported at runtime.
- The tune-loop skill was exercised end-to-end on a real run (Qwen3-8B, TP=2, Ascend 910C ×2) — see the example output below. The full 7-phase campaign completed, producing a structured ledger and final report.

---

### Example output from a real run

The tune-loop skill was run end-to-end against Qwen3-8B (dense, bf16, TP=2) on Ascend 910C ×2. After completion, the working directory contains one subdirectory per iteration plus the ledger and final report:

```
# tree -L 1 ./
./
├── baseline
├── extra-info
├── fa3_v1.py.orig.bak
├── final_report.md
├── iter_01
├── iter_02
├── iter_03
├── iter_04
├── iter_05
├── iter_06
├── iter_07
├── iter_08
├── iter_09
├── iter_10
├── iter_12
├── iter_13
├── iter_14
├── iter_15
├── iter_16
├── iter_17
├── iter_24
├── iter_25
├── iter_26
├── iter_27
├── kernel_meta
├── ledger.md
├── profiling
├── rms_norm.py.orig.bak
└── server.log
```

Each `iter_NN/` holds that iteration's evalscope outputs; `ledger.md` is the single source of truth (one entry per verdict, written immediately); `final_report.md` is synthesized when the loop exits. The `.orig.bak` files are backups of source edits that were reverted after measurement.

The generated `final_report.md` (26 post-baseline iterations, all 7 phases attempted):

<details>
<summary><b>final_report.md</b> (click to expand)</summary>

# vLLM-Ascend Tuning Report — qwen3-8b-chat (Qwen3-8B)

> Generated: 2026-06-25 14:30  |  Run: `./tuning_run/qwen_run_03`  |  evalscope 1.8.1  |  NPU: Ascend910 ×2 (devices 0,1)

## Summary

| Item | Value |
|------|-------|
| Model | `qwen3-8b-chat` (Qwen3-8B, dense, bf16) |
| Tensor parallel | 2 |
| Target metric (frozen) | `tpot_avg` @ c=1 (user-overridden; guard = `ttft_avg` @ c=1, >5% regression downgrades WIN→NEUTRAL) |
| Fixed workload | c=1/4, 20/80 requests, in=1024 / out=512, warmup 20 @ c=1 |
| Total iterations | 26 (post-baseline) |
| Winning levers | **0** |
| NEUTRAL | 12 |
| LOSS | 1 |
| STARTUP_FAIL | 4 |
| SKIPPED_CONFLICT | 9 |
| Overall improvement on target metric (c=1) | **0.0%** (baseline unchanged; best config = baseline) |
| Completion status | `completed` — all Phases 1–7 attempted |

**Bottom line:** No tuning lever improved `tpot_avg` at c=1. The baseline configuration is already at the performance ceiling for this workload. Every config knob that could move the decode critical path was either (a) a no-op at c=1/short-context decode, (b) an environmental incompatibility (missing module / ACL runtime error / tensor-layout assertion), or (c) inapplicable to a dense, unquantized, non-DeepSeek model. The decode loop runs inside a captured ACL graph (`FULL_DECODE_ONLY`), which already fuses and optimally schedules the kernel sequence — leaving no scheduler/fusion/comm lever with headroom at c=1.

## Baseline vs Best Configuration

Best = baseline (0 winning levers). Metrics identical.

| Concurrency | Metric | Baseline | Best | Improvement |
|-------------|--------|----------|------|-------------|
| 1 | TTFT avg (ms) | 63.0 | 63.0 | 0.0% |
| 1 | **TPOT avg (ms)** ★ target | **10.2** | **10.2** | **0.0%** |
| 1 | Output tok/s | 96.57 | 96.57 | 0.0% |
| 1 | Latency avg (ms) | 5302 | 5302 | 0.0% |
| 4 | TTFT avg (ms) | 142.5 | 142.5 | 0.0% |
| 4 | TPOT avg (ms) | 11.2 | 11.2 | 0.0% |
| 4 | Output tok/s | 348.27 | 348.27 | 0.0% |

## Winning Configuration (copy-pasteable)

No lever beat the baseline. The baseline user-provided command remains the recommended serving config:

```bash
ASCEND_RT_VISIBLE_DEVICES=0,1 vllm serve /data/models/qwen3-8b \
  --served-model-name qwen3-8b-chat \
  --tensor-parallel-size 2 \
  --host 0.0.0.0 --port 5001 \
  --trust-remote-code \
  --compilation-config '{"cudagraph_mode": "FULL_DECODE_ONLY"}'
```

Effective engine state at this config (from EngineCore log): `dtype=torch.bfloat16`, `enable_prefix_caching=True`, `enable_chunked_prefill=True`, custom fusions `norm_quant` + `act_quant` enabled, ACL Graph mode `FULL_DECODE_ONLY`, `cudagraph_capture_sizes=[1,2,4,...,256]`, `ir_op_priority rms_norm=['native']`.

## Optimization History (condensed)

| # | Phase / Lever | Verdict | c=1 tpot Δ | One-line reason |
|---|---------------|---------|-----------|-----------------|
| 0 | Baseline | — | — | tpot=10.2ms ttft=63.0ms @ c=1 |
| 1 | 2.1 balance_scheduling | NEUTRAL | 0.0% | no mixed-batch contention at c=1/4 |
| 2 | 2.2 cpu_binding=false | NEUTRAL | +1.0% | default(true) is better; disabling hurt |
| 3 | 2.3 block_size=128 | NEUTRAL | 0.0% | default already 128; KV-ptr chase not bottleneck at c=1 |
| 4 | 2.4 chunked_prefill mnbt=2048 | NEUTRAL | 0.0% | 1024 prefill fits one chunk; chunking is a no-op |
| 5 | 2.5 max_num_seqs=32 | NEUTRAL | 0.0% | cap not binding at c≤4 |
| 6 | 3.1 VLLM_ASCEND_ENABLE_NZ=2 | **LOSS** | **+10.8%** | forcing NZ on all ops adds conversion overhead on decode |
| 7 | 3.2 npugraph_ex+static_kernel | NEUTRAL | 0.0% | graph already captured; static precomp no extra gain |
| 8 | 3.3 fuse_norm_quant=false | SKIPPED | — | flag lives in `pass_config`, not `ascend_compilation_config`; not applied |
| 9 | 3.4 fuse_allreduce_rms=true | STARTUP_FAIL | — | `fused_add_rms_norm` op not registered in this build |
| 10 | 3.5 MATMUL_ALLREDUCE=1 | NEUTRAL | 0.0% | no compute to overlap at c=1 |
| 11 | 3.6 GMM-SwiGLU-Quant | SKIPPED | — | MoE-only; model is dense |
| 12 | 3.7 XliteGraph | STARTUP_FAIL | — | `ModuleNotFoundError: No module named 'xlite'` |
| 13 | 4.1 flashcomm1 | STARTUP_FAIL | — | `ACL stream synchronize failed, error code:507015` |
| 14 | 4.2 flashcomm2 PS=1 | STARTUP_FAIL | — | tensor stride assertion (4096 vs 2048) |
| 15 | 4.3 prefill_comm_compute_overlap | NEUTRAL | 0.0% | overlap overhead > gain at c=1; ttft +6.8% |
| 16 | 4.4 HCCL_BUFFSIZE=200 | NEUTRAL | 0.0% | decode allreduce payload tiny; buffer irrelevant |
| 17 | 5.1 weight_prefetch | NEUTRAL | 0.0% | graph already schedules weights optimally |
| 18 | 5.2 MLAPO | SKIPPED | — | DeepSeek W8A8 only; model is bf16 |
| 19 | 5.3 fused MC2 | SKIPPED | — | MoE W8A8+EP only; dense/unquantized |
| 20 | 6.1 shared_expert overlap | SKIPPED | — | MoE only |
| 21 | 6.2 gate overlap | SKIPPED | — | MoE only |
| 22 | 6.3 dsa_preprocess | SKIPPED | — | DeepSeek Sparse Attention only |
| 23 | 6.4 dsv4_dsa_overlap | SKIPPED | — | DeepSeek V4 only |
| 24 | 7.1/7.2 profiling trace | SKIPPED | — | graph mode hides per-op kernels from torch profiler |
| 25 | 7.3a rms_norm ROW_BLOCK_SIZE 16→32 | NEUTRAL | 0.0% | rms_norm routes to native ACLNN, not this Triton kernel |
| 26 | 7.3b fa3 num_splits 1→2 | NEUTRAL | 0.0% | single-token/short-KV decode has no work to split |

## Levers That Did Not Help (no-ops at c=1, reverted)

- **2.1 balance_scheduling** — only helps mixed prefill+decode batching contention; absent at c=1/4.
- **2.2 cpu_binding=false** — default `true` (NUMA binding) is already optimal here; disabling regressed c=1 by ~1%.
- **2.3 block_size=128** — engine default already resolves to 128; KV-cache pointer chasing is not the c=1 bottleneck.
- **2.4 chunked_prefill (mnbt=2048)** — at 1024 input tokens the prefill fits one chunk, so chunking never engages.
- **2.5 max_num_seqs=32** — cap is not binding when ≤4 sequences are concurrent.
- **3.2 npugraph_ex + static_kernel** — `FULL_DECODE_ONLY` ACL Graph already captures the decode graph; static pre-compilation adds startup time with no per-step gain.
- **3.5 MATMUL_ALLREDUCE=1** — at c=1 there is too little matmul compute to overlap with the allreduce.
- **4.3 prefill_comm_compute_overlap** — overlap machinery costs more than it saves for a single 1024-token prefill (TTFT +6.8%); TPOT unaffected.
- **4.4 HCCL_BUFFSIZE=200** — the per-decode-step allreduce payload is a few KB; buffer fragmentation is not the bottleneck.
- **5.1 weight_prefetch** — captured graph already schedules weight loads optimally; explicit prefetch adds no HBM-latency hiding at c=1.
- **7.3a rms_norm ROW_BLOCK_SIZE=32** — `ir_op_priority rms_norm=['native']`: rms_norm uses the native ACLNN path, not this Triton kernel.
- **7.3b fa3 num_splits=2** — single decode token against short KV has insufficient work to amortize split-across-cores overhead.

## Failed / Incompatible Levers

| Lever | Verdict | Stage | Reason | Evidence |
|-------|---------|-------|--------|----------|
| `3.4 fuse_allreduce_rms=true` | `STARTUP_FAIL` | `server_startup` | fused op missing in this vllm-ascend/CANN build | `AttributeError: '_OpNamespace' '_C' object has no attribute 'fused_add_rms_norm'` → `Engine core initialization failed` (flag WAS applied via corrected `pass_config` path) |
| `3.7 XliteGraph enabled=true` | `STARTUP_FAIL` | `server_startup` | required `xlite` Python package not installed | `ModuleNotFoundError: No module named 'xlite'` at `vllm_ascend/xlite/xlite.py:34` |
| `4.1 enable_flashcomm1=true` | `STARTUP_FAIL` | `server_startup` | ACL stream sync incompatible with this CANN runtime / graph-mode interaction | `RuntimeError: Worker failed with error 'ACL stream synchronize failed, error code:507015'` |
| `4.2 FLASHCOMM2_PARALLEL_SIZE=1` | `STARTUP_FAIL` | `server_startup` | FlashComm2 overlap kernel tensor-layout mismatch with Qwen3-8B | `Worker failed with error 'expected size 4096==4096, stride 4096==2048 at dim=0; ...'` (independent of PS value; tp/2, tp not retried) |
| `3.3 fuse_norm_quant=false` | `SKIPPED_CONFLICT` | `precheck` | documented `--additional-config` path does not reach the flag; it lives in `compilation_config.pass_config` | log still showed `pass_config: {'fuse_norm_quant': True}` + `Enabled custom fusions: norm_quant` after attempt. Corrected mechanism: `--compilation-config '{"pass_config":{"fuse_norm_quant":false}}'` |
| `3.6 GMM-SwiGLu-Quant` | `SKIPPED_CONFLICT` | `precheck` | MoE-only; Qwen3-8B is dense | no MoE expert/grouped-matmul layers |
| `5.2 MLAPO` | `SKIPPED_CONFLICT` | `precheck` | DeepSeek W8A8 only; model is bf16 unquantized | `quantization=None` |
| `5.3 fused MC2` | `SKIPPED_CONFLICT` | `precheck` | MoE W8A8 + EP only; dense/unquantized/TP-only | no EP config |
| `6.1 multistream_overlap_shared_expert` | `SKIPPED_CONFLICT` | `precheck` | MoE only | no shared expert |
| `6.2 multistream_overlap_gate` | `SKIPPED_CONFLICT` | `precheck` | MoE only | no router/gate |
| `6.3 multistream_dsa_preprocess` | `SKIPPED_CONFLICT` | `precheck` | DeepSeek Sparse Attention only | standard attention |
| `6.4 multistream_dsv4_dsa_overlap` | `SKIPPED_CONFLICT` | `precheck` | DeepSeek V4 only | not a DeepSeek model |
| `7.1/7.2 profiling op-classification` | `SKIPPED_CONFLICT` | `profiling analysis` | torch trace captured only host-side Python events; no NPU kernel visibility | 5.67M events all `python_function`; zero `aclnn`/`matmul`/`rms_norm`/`attention` kernel events — `FULL_DECODE_ONLY` graph mode hides per-op kernels |

## Manual Intervention Required

**1. Per-op kernel profiling is blocked by graph mode.**
The torch profiler trace (`profiling/*.trace.json.gz`, 121MB, 5.67M events) contains **only host-side Python function events** — zero NPU kernel events. Root cause: `cudagraph_mode=FULL_DECODE_ONLY` executes decode as a single captured ACL graph replay, so individual kernels (`aclnn*`, matmul, rms_norm, attention, hccl) are fused behind the graph-replay boundary and never appear as separate profiler events. The Ascend per-rank dirs (`rank0/1_*_ascend_pt`) emitted only raw binary `slice_*` host/device slices; no CSV op-summary was produced via the torch-profiler path.

To identify the actual top-N slowest operators and tune them, run **`msprof` directly** (outside the torch profiler), which captures device-side op-level stats even under graph mode:
```bash
msprof --application="vllm serve <baseline_cmd>" --output=./msprof_out --aic-metrics=PipeUtilization
# then inspect msprof_out/PROF_*/summary/csv/op_summary_*.csv for top ops by aicore time
```
This was not done in this run (msprof availability was optional and the torch-profiler path was attempted first per the phase doc).

**2. Non-tunable (ACLNN) ops likely dominate the decode path.**
Even with msprof, several decode ops are not source-tunable:
- `rms_norm` → routed to native ACLNN (`ir_op_priority rms_norm=['native']`); the Triton `ROW_BLOCK_SIZE` in `ops/triton/rms_norm.py` is NOT on the hot path (confirmed: editing 16→32 had zero effect).
- `matmul`/`gemm` → ACLNN-backed on Ascend; block sizes (`BLOCK_M/N/K`) in `ops/triton/batch_invariant/matmul.py` are not on the path.
- `flash_attention` (fa3) → calls `torch_npu.flash_attn_with_kvcache` (ACLNN); `num_splits` is the only user-knob and had no effect at c=1/short-KV.
- Any `aclnn*` op → tiling decided at CANN op-compile time.

If msprof shows an `aclnn*` op dominating (e.g. `aclnnFlashAttention`, `aclnnMatmul`, `aclnnGroupedMatmul`), the tuning levers are: (a) CANN AOE offline tiling for the specific input shapes, or (b) a newer CANN/vllm-ascend build with an optimized op binary.

**3. Environmental fixes that would unlock crashed levers.**
- `3.7 XliteGraph`: install the `xlite` wheel matching the installed CANN/vllm-ascend version.
- `3.4 fuse_allreduce_rms` / `4.1 flashcomm1` / `4.2 flashcomm2`: require a vllm-ascend/CANN build where `fused_add_rms_norm` is registered and the FlashComm ACL-stream / tensor-layout paths are compatible with this model. Verify against the vllm-ascend release notes for the installed CANN version.

## Recommended Next Steps

1. **Run `msprof` directly** (command above) to get device-side per-op timing under graph mode — this is the only way to see whether decode is matmul-bound, attention-bound, or comm-bound at c=1, and is the prerequisite for any further kernel-level work.
2. **Re-benchmark at higher concurrency / longer context** if the real deployment target is c≥8 or input >>2048 tokens: several no-op levers here (chunked_prefill, balance_scheduling, weight_prefetch, matmul_allreduce, HCCL_BUFFSIZE, fa3 num_splits) are designed for exactly those regimes and would show real effects there. The c=1/short-context workload simply has no headroom for them.
3. **Upgrade vllm-ascend/CANN** to re-test the 4 STARTUP_FAIL levers (fuse_allreduce_rms, XliteGraph, flashcomm1/2) — these are the levers most likely to help TP=2 decode if the build incompatibilities are resolved.
4. **If TPOT 10.2ms/c=1 must drop further**, the ceiling is the ACL graph's fused kernel sequence; the only remaining levers are CANN-version upgrades or AOE offline tiling of the dominant `aclnn*` op identified by msprof — both are off-repo, vendor-side actions.

## Artifacts

- Ledger (single source of truth): `tuning_run/qwen_run_03/ledger.md`
- Per-iteration evalscope outputs: `tuning_run/qwen_run_03/iter_NN/outputs/*/qwen3-8b-chat/`
- Baseline: `tuning_run/qwen_run_03/baseline/outputs/20260625_105733/qwen3-8b-chat/`
- Profiling trace: `tuning_run/qwen_run_03/profiling/` (chrome trace + ascend_pt ranks)
- Source-edit backups (reverted): `rms_norm.py.orig.bak`, `fa3_v1.py.orig.bak` (both restored to original in the vllm-ascend tree)

</details>

- vLLM version: v0.25.1

### DIFF
--- a/.agents/README.md
+++ b/.agents/README.md
@@ -19,6 +19,16 @@ Note: Please copy the skills directory `.agents/skills` to `.claude/skills` if y
     - [File layout](#file-layout-1)
     - [Quick start](#quick-start-2)
     - [Key guidelines](#key-guidelines)
+  - [vLLM Ascend Serving Auto Benchmark Skill](#vllm-ascend-serving-auto-benchmark-skill)
+    - [What it does](#what-it-does-3)
+    - [File layout](#file-layout-2)
+    - [Quick start](#quick-start-3)
+    - [Key guidelines](#key-guidelines-2)
+  - [vLLM Ascend Serving Tune Loop Skill](#vllm-ascend-serving-tune-loop-skill)
+    - [What it does](#what-it-does-4)
+    - [File layout](#file-layout-3)
+    - [Quick start](#quick-start-4)
+    - [Key guidelines](#key-guidelines-3)
 
 
 ## vLLM Ascend Model Adapter Skill
@@ -105,3 +115,85 @@ This skill guides you through a structured workflow to:
 - Focus on user-facing impact and include context for practical usage.
 - Verify details by checking linked PRs (use GitHub API for descriptions if needed).
 - Keep notes concise and avoid unnecessary technical details.
+
+## vLLM Ascend Serving Auto Benchmark Skill
+
+Automated benchmark skill for vLLM-Ascend serving on Huawei Ascend 910C NPU. Launches vLLM serving, runs evalscope perf stress tests, and produces a complete Markdown/CSV performance report.
+
+### What it does
+
+This skill runs an automated five-phase benchmark workflow:
+
+1. Validate environment and config (NPU devices, model path, evalscope/vllm-ascend install).
+2. Launch the vLLM-Ascend OpenAI-compatible server with the chosen model and parallelism.
+3. Run evalscope perf stress tests across configurable concurrency levels.
+4. Collect metrics (`ttft_avg`, `tpot_avg`, `output_token_throughput`, `latency_avg`) at each level.
+5. Generate a Markdown/CSV performance report via `scripts/generate_report.py`.
+
+### File layout
+
+| File | Purpose |
+| ---- | ------- |
+| `SKILL.md` | Skill definition, phases, input parameters, and report format |
+| `configs/*.yaml` | Pre-built model configs (DeepSeek-V3, DeepSeek-V4-Flash, Qwen3-32B, Qwen3.5-27B, Qwen3.5-35B-A3B) |
+| `scripts/run_benchmark.sh` | Orchestrates server launch + evalscope run + report generation |
+| `scripts/validate_configs.py` | Validates a YAML config before launch |
+| `scripts/generate_report.py` | Produces the Markdown/CSV performance report |
+
+### Quick start
+
+1. Open a conversation with the AI agent inside the vllm-ascend dev container.
+2. Invoke the skill (e.g. `/ascend-vllm-serving-auto-benchmark`).
+3. Provide `model_path` and `model_name` (or pass a `config_file` from `configs/`).
+4. The agent runs the five-phase workflow and writes the report to the output directory.
+
+### Key guidelines
+
+- Concurrency levels and requests-per-level are fixed for a given run — do not change them mid-run.
+- Always launch the server on a free port and clean it up after the run.
+- A failed server startup must still produce a report entry marked as failure, not a silent skip.
+- Reuse a ready-to-use `configs/*.yaml` when benchmarking a known model instead of hand-tuning flags.
+
+## vLLM Ascend Serving Tune Loop Skill
+
+Autonomous, evidence-driven performance optimization loop for vLLM-Ascend serving on Huawei Ascend 910C NPU. Runs a fixed baseline, then iterates through a layered tuning plan (scheduler → ACLGraph → kernel) and produces an optimization ledger plus a final comparison report.
+
+### What it does
+
+This skill drives a long-running optimization campaign across 7 phases:
+
+1. Establish a fixed baseline benchmark (the contract that never changes between iterations).
+2. Iterate over ranked tuning levers — one lever per iteration, change one thing at a time (RLCR principle).
+3. Revalidate at c=1/4/8 with the fixed workload after every change; gate WIN/LOSS on `ttft_avg` at c=1.
+4. Append every verdict (WIN/LOSS/NEUTRAL/STARTUP_FAIL/BENCHMARK_FAIL/SKIPPED_CONFLICT) to the ledger immediately.
+5. Resume cleanly across session interruptions via the ledger; drive the campaign to completion with the auto-resume wrapper.
+6. Synthesize a final report (`final_report.md`) with baseline-vs-best, winning copy-pasteable config, and failed levers.
+
+### File layout
+
+| File | Purpose |
+| ---- | ------- |
+| `SKILL.md` | Skill definition, phases, ledger discipline, resume logic, report format |
+| `references/tuning-knobs-reference.md` | Catalog of tunable Ascend knobs and their expected impact |
+| `scripts/auto_resume_wrapper.sh` | Relaunches the `claude` session until `final_report.md` exists (cross-session resume) |
+| `scripts/generate_tuning_report.py` | Synthesizes `final_report.md` from the ledger |
+
+### Quick start
+
+1. Open a conversation with the AI agent inside the vllm-ascend dev container.
+2. Invoke the skill (e.g. `/ascend-vllm-serving-tune-loop`).
+3. Provide `model_path`, `model_name`, and `tensor_parallel_size`; optionally `baseline_launch_command` to preserve a known-good config across the campaign.
+4. To run the full campaign unattended, drive it with the wrapper:
+   ```bash
+   bash scripts/auto_resume_wrapper.sh --work-dir <work_dir> -- claude -p "..." --max-turns 0
+   ```
+5. The agent writes the ledger incrementally and `final_report.md` when the loop exits.
+
+### Key guidelines
+
+- One lever per iteration — never bundle changes; it destroys attribution.
+- The benchmark contract (concurrency levels, token sizes, requests-per-level, target metric) is frozen after Phase 1.
+- Append ledger entries immediately after each verdict; never defer writes — they are the crash-recovery hand-off.
+- c=1 `ttft_avg` is the authoritative WIN/LOSS metric; c=4/8 alone never justifies a WIN.
+- Drive the campaign through the auto-resume wrapper; do not manually relaunch `claude` mid-campaign (it races with the wrapper over the ledger).
+- Run the mandatory post-iteration cleanup between every iteration to avoid zombie processes and NPU memory leaks.
--- a/.agents/skills/ascend-vllm-serving-auto-benchmark/SKILL.md
+++ b/.agents/skills/ascend-vllm-serving-auto-benchmark/SKILL.md
@@ -1,0 +1,173 @@
+# ascend-vllm-serving-auto-benchmark
+
+Automated benchmark skill for vLLM-Ascend serving on Huawei Ascend 910C NPU. Launches vLLM serving, runs evalscope perf stress tests, and produces a complete Markdown/CSV performance report.
+
+## When to invoke
+
+Use this skill when the user wants to:
+- Benchmark LLM serving performance on Ascend 910C NPU
+- Compare performance across different model configurations or parallelism settings
+- Generate a reproducible performance report for vllm-ascend
+
+Trigger phrases: "benchmark", "pressure test", "性能测试", "压测", "ascend benchmark", "vllm-ascend serving 性能"
+
+## Prerequisites
+
+Before running, the system must have:
+
+1. **Hardware**: Huawei Ascend 910C NPU (Atlas 800T A2 or Atlas 800I A2/A3 series)
+2. **CANN**: >= 8.1.0 (recommend 9.0.0)
+3. **torch-npu**: matched to your PyTorch version
+4. **vllm-ascend**: installed (`pip install vllm-ascend` or from source)
+5. **evalscope[perf]**: installed (`pip install evalscope[perf] -U`)
+6. **Model weights**: downloaded locally or accessible via HuggingFace/ModelScope path
+
+## Inputs the user must supply
+
+| Parameter | Required | Description |
+|-----------|----------|-------------|
+| `model_path` | Yes | Local path or HuggingFace/ModelScope model ID |
+| `model_name` | Yes | Short name for the model (used in reports and API) |
+| `tensor_parallel_size` | No | Number of NPU cards (default: 1 for 910C single card) |
+| `pipeline_parallel_size` | No | Pipeline parallel size (default: 1) |
+| `max_model_len` | No | Max context length (default: 8192) |
+| `dtype` | No | Weight dtype: `float16`, `bfloat16`, `auto` (default: `float16`) |
+| `quantization` | No | Quantization method: `w8a8`, `w4a16`, `None` (default: None) |
+| `port` | No | Server port (default: 5000) |
+| `parallel_levels` | No | Concurrency levels to test (default: `1 4 8 16 32 64`) |
+| `requests_per_level` | No | Requests per concurrency level (default: `10 40 80 160 320 640`) |
+| `input_tokens` | No | Input token length (default: 1024) |
+| `output_tokens` | No | Output token length (default: 512) |
+| `config_file` | No | Path to a YAML config file from `configs/` directory |
+
+## Workflow
+
+This skill runs in **five phases**:
+
+### Phase 1 — Environment & Config Validation
+- Detect available NPU devices via `npu-smi info`
+- Validate model path exists
+- Validate config file if provided (via `scripts/validate_configs.py`)
+- Check evalscope and vllm-ascend installations
+
+### Phase 2 — Launch vLLM-Ascend Server
+Start the serving endpoint:
+```bash
+ASCEND_RT_VISIBLE_DEVICES=0,1,...  \
+python -m vllm.entrypoints.openai.api_server \
+  --model <model_path> \
+  --served-model-name <model_name> \
+  --tensor-parallel-size <tp> \
+  --pipeline-parallel-size <pp> \
+  --max-model-len <max_len> \
+  --dtype <dtype> \
+  --port <port> \
+  --trust-remote-code
+```
+
+Wait for server readiness by polling `/health` endpoint (max 300s).
+
+### Phase 3 — Warm-up
+Run 20 warm-up requests at concurrency=1 before measuring:
+```bash
+evalscope perf \
+  --parallel 1 \
+  --number 20 \
+  --model <model_name> \
+  --url http://127.0.0.1:<port>/v1/chat/completions \
+  --api openai \
+  --dataset random \
+  --max-tokens <output_tokens> \
+  --min-tokens <output_tokens> \
+  --min-prompt-length <input_tokens> \
+  --max-prompt-length <input_tokens> \
+  --tokenizer-path <model_path>
+```
+
+### Phase 4 — Stress Test (evalscope perf)
+Run the main benchmark across all concurrency levels:
+```bash
+evalscope perf \
+  --parallel <parallel_levels> \
+  --number <requests_per_level> \
+  --model <model_name> \
+  --url http://127.0.0.1:<port>/v1/chat/completions \
+  --api openai \
+  --stream \
+  --dataset random \
+  --max-tokens <output_tokens> \
+  --min-tokens <output_tokens> \
+  --min-prompt-length <input_tokens> \
+  --max-prompt-length <input_tokens> \
+  --tokenizer-path <model_path>
+```
+
+Results are saved to `outputs/<timestamp>/<model_name>/`.
+
+### Phase 5 — Report Generation
+Call `scripts/generate_report.py` to produce:
+- `benchmark_report.md` — full Markdown report with tables and charts description
+- `benchmark_results.csv` — raw metrics for spreadsheet import
+
+## Output report structure
+
+The generated `benchmark_report.md` includes:
+
+1. **Environment Summary** — Hardware (NPU model, count), CANN version, vllm-ascend version, evalscope version
+2. **Model Configuration** — model path, TP/PP size, dtype, quantization, max_model_len
+3. **Workload Specification** — input/output token lengths, dataset type, request distribution
+4. **Performance Overview Table** — per-concurrency: RPS, success rate, throughput (tokens/s)
+5. **Latency Distribution Table** — per-concurrency: avg/P50/P90/P99 latency, TTFT, TPOT
+6. **Best Configuration Summary** — peak throughput point, best TTFT point, recommended concurrency
+7. **Fairness & Reproducibility** — exact server command, evalscope command, git commit hashes
+
+## Constraints & fairness rules
+
+- Always restart the server between different TP/PP configurations; never reuse a warm server across configs
+- Warm up before every measurement run
+- Record exact vllm-ascend git commit: `git -C $(pip show vllm-ascend | grep Location | awk '{print $2}')/vllm_ascend rev-parse HEAD`
+- Report NPU utilization if `npu-smi` is available
+- All concurrency levels must use identical model weights and tokenizer
+- Never compare results from different NPU types in the same report
+
+## Config files
+
+YAML configs in `configs/` encode model-specific defaults. The user can pass `--config_file configs/qwen2.5-7b-instruct.yaml` to skip specifying individual flags. See `configs/qwen2.5-7b-instruct.yaml` for the format.
+
+## Example invocations
+
+**Quick single-card benchmark:**
+```
+/ascend-vllm-serving-auto-benchmark
+model_path=/data/models/Qwen2.5-7B-Instruct
+model_name=Qwen2.5-7B-Instruct
+```
+
+**Multi-card benchmark with custom config:**
+```
+/ascend-vllm-serving-auto-benchmark
+config_file=configs/deepseek-r1-7b.yaml
+tensor_parallel_size=2
+```
+
+**Full custom benchmark:**
+```
+/ascend-vllm-serving-auto-benchmark
+model_path=/data/models/Qwen2.5-72B-Instruct
+model_name=Qwen2.5-72B-Instruct
+tensor_parallel_size=8
+max_model_len=16384
+dtype=bfloat16
+parallel_levels="1 8 16 32 64 128"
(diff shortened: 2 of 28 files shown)


maoxx241 (changes requested):
Thank you for the contribution! The PR direction is highly valuable and addresses a key need. I've left some feedback below to help get this ready for merging.

###  Blocking Issues
1. **Missing Frontmatter**: Both new `SKILL.md` files are missing the required Agent Skills YAML frontmatter (`name`, `description`). Existing repo skills can be used as a reference for this format.
2. **File Length**: `ascend-vllm-serving-tune-loop/SKILL.md` is currently 867 lines. Please consider splitting this via progressive disclosure into separate references/scripts to maintain readability.
3. **Unsafe Cleanup Contract**: The tune-loop “Operation-first contract” is unsafe. Instructing the agent to immediately run a broad `pkill -9 -f vllm/ray` cleanup can accidentally kill unrelated workloads on shared Ascend machines. 
   - *Fix:* Cleanup must be strictly scoped by PID, process group, work_dir, port, or devices. Destructive actions should require user confirmation or explicit invocation.
4. **Data Loss**: `generate_report.py` currently discards parsed `.json` evalscope results and returns empty results.
5. **Metrics Logic**: `generate_tuning_report.py` does not populate per-iteration metrics, misclassifies non-WIN/LOSS/NEUTRAL verdicts, and computes improvement with the wrong sign for lower-is-better metrics.
6. **Process Handling**: `auto_resume_wrapper.sh` should trap and exit on `SIGINT` (exit code 130) instead of automatically relaunching.
7. **Config & Flags**: `run_benchmark.sh` does not apply most `server.env` values from YAML configs and appears to use the wrong evalscope output directory flag (`--work-dir` vs the current `outputs_dir`/`--outputs-dir`).

###  Suggested Next Steps
- Add proper skill frontmatter and set `disable-model-invocation: true`.
- Split long tune-loop content into dedicated reference files.
- Replace broad cleanup with a scoped, safe cleanup script.
- Add unit tests/fixtures for JSONL/JSON evalscope parsing, ledger parsing, and report generation.
- Add shellcheck for bash scripts and a minimal dry-run CI pipeline.

Let me know if you have any questions or need clarification on any of these points. Looking forward to the next revision!

Jonxjdong:
> Thank you for the contribution! The PR direction is highly valuable and addresses a key need. I've left some feedback below to help get this ready for merging.
> 
> ### Blocking Issues
> 1. **Missing Frontmatter**: Both new `SKILL.md` files are missing the required Agent Skills YAML frontmatter (`name`, `description`). Existing repo skills can be used as a reference for this format.
> 2. **File Length**: `ascend-vllm-serving-tune-loop/SKILL.md` is currently 867 lines. Please consider splitting this via progressive disclosure into separate references/scripts to maintain readability.
> 3. **Unsafe Cleanup Contract**: The tune-loop “Operation-first contract” is unsafe. Instructing the agent to immediately run a broad `pkill -9 -f vllm/ray` cleanup can accidentally kill unrelated workloads on shared Ascend machines.
>    
>    * _Fix:_ Cleanup must be strictly scoped by PID, process group, work_dir, port, or devices. Destructive actions should require user confirmation or explicit invocation.
> 4. **Data Loss**: `generate_report.py` currently discards parsed `.json` evalscope results and returns empty results.
> 5. **Metrics Logic**: `generate_tuning_report.py` does not populate per-iteration metrics, misclassifies non-WIN/LOSS/NEUTRAL verdicts, and computes improvement with the wrong sign for lower-is-better metrics.
> 6. **Process Handling**: `auto_resume_wrapper.sh` should trap and exit on `SIGINT` (exit code 130) instead of automatically relaunching.
> 7. **Config & Flags**: `run_benchmark.sh` does not apply most `server.env` values from YAML configs and appears to use the wrong evalscope output directory flag (`--work-dir` vs the current `outputs_dir`/`--outputs-dir`).
> 
> ### Suggested Next Steps
> * Add proper skill frontmatter and set `disable-model-invocation: true`.
> * Split long tune-loop content into dedicated reference files.
> * Replace broad cleanup with a scoped, safe cleanup script.
> * Add unit tests/fixtures for JSONL/JSON evalscope parsing, ledger parsing, and report generation.
> * Add shellcheck for bash scripts and a minimal dry-run CI pipeline.
> 
> Let me know if you have any questions or need clarification on any of these points. Looking forward to the next revision!

Thanks for the detailed review — I’ve pushed a follow-up update addressing the blocking issues.

What changed:
- Added required YAML frontmatter to both SKILL.md files and set `disable-model-invocation: true`
- Split the long tune-loop skill into smaller reference docs for progressive disclosure
- Replaced the unsafe broad cleanup guidance with a scoped managed-process cleanup script
- Fixed `generate_report.py` so JSON evalscope results are parsed instead of discarded
- Fixed `generate_tuning_report.py` to populate per-iteration metrics, preserve non-WIN/LOSS/NEUTRAL verdicts, and compute improvement with the correct direction for lower-is-better metrics
- Updated `auto_resume_wrapper.sh` to exit on SIGINT / code 130 instead of relaunching
- Updated `run_benchmark.sh` to apply YAML `server.env` values and use the current evalscope output-directory handling
- Added focused unit tests plus a minimal dry-run CI workflow and shellcheck coverage for the skill scripts

Validation:
- `bash -n` on updated shell scripts
- `python3 -m py_compile` on updated Python scripts
- `pytest --noconftest -q tests/ut/agents/test_serving_skill_scripts.py`
- dry-run validation for `run_benchmark.sh`

maoxx241 (changes requested):
Thanks for the updates — the skill structure and documentation are much clearer now. I took another pass against the current vLLM-Ascend runtime and the official Codex/Claude Code skill conventions. There are still a few correctness and safety issues that I think should be addressed before merging.

1. **Codex and Claude Code need separate invocation controls**

   Please keep `disable-model-invocation: true`, since that is the correct Claude Code control. For Codex, please add the following file under **both** skill directories:

   ```text
   .agents/skills/<skill-name>/agents/openai.yaml
   ```

   ```yaml
   policy:
     allow_implicit_invocation: false
   ```

   Codex defaults to allowing implicit invocation unless this policy is set. Explicit examples should also be separated:

   - Codex: `$ascend-vllm-serving-auto-benchmark`
   - Claude Code: `/ascend-vllm-serving-auto-benchmark`

   This is particularly important because these skills start serving processes, occupy NPU resources, and may run long tuning campaigns. See the [Codex skill invocation policy](https://learn.chatgpt.com/docs/build-skills) and [Claude Code skill documentation](https://code.claude.com/docs/en/skills).

2. **Please test against real EvalScope output**

   EvalScope is explicitly part of this skill's contract: the skill description says it runs `evalscope perf`, Phase 4 is the EvalScope benchmark, and the report generator is required to consume its JSON/JSONL output.

   The current UT fixture uses a normalized custom shape, but real EvalScope v1.8.1 output is nested under timestamp/name/parallel directories and uses files such as `benchmark_summary.json` and `benchmark_percentile.json`, with fields including `Concurrency`, `TTFT (ms)`, and `TPOT (ms)`. With a real-shaped fixture, the auto-benchmark parser returns no records and the tuning parser may return zero-valued metrics.

   Could you please add a small sanitized fixture captured from real EvalScope output and make both report generators parse the paired summary/percentile files? The test does not need to install or run EvalScope; checking in a minimal representative fixture should be sufficient.

3. **`VLLM_ASCEND_ENABLE_ACLGRAPH` is currently an end-to-end no-op**

   This variable is present in all five benchmark configs, is generated by `--no-aclgraph`, and is included in the final report. However, it is not defined in `vllm_ascend/envs.py` and is not read anywhere in the runtime.

   As a result, `--no-aclgraph` does not actually disable ACLGraph, while the generated report can claim that it did. This can invalidate benchmark comparisons.

   I suggest removing this environment variable from the configs, runner, and report, then wiring `--no-aclgraph` to the existing supported mechanism such as `--enforce-eager`. Graph-enabled configurations should use the existing compilation/cudagraph configuration path. Please also add a dry-run test that verifies the generated server command actually changes graph mode.

4. **`block_size=16` should be treated as a supported-configuration issue, not only a documentation typo**

   The current Ascend platform default is 128. In addition:

   - the Mamba patch explicitly states that upstream block size 16 is unsupported on Ascend and changes it to 128;
   - prefix caching and chunked prefill force block size 128;
   - DeepSeek V4 accepts only 32, 64, or 128;
   - Xlite graph recommends 128.

   Therefore, listing 16 as the generic 910C default can lead the tuning agent to treat an unsupported or automatically rewritten configuration as a valid baseline. It can also waste an iteration by presenting 128 as an optimization even though it is already the platform default.

   Could you please update the table to use 128 as the general default and document model/feature-specific supported values? It would also be good to audit every knob in this reference against the current runtime, since `prefill_comm_compute_overlap` and `multistream_dsa_preprocess` currently appear only in this new document and have no implementation in the repository.

These points are not cosmetic documentation issues: they can cause the skill to run with ineffective settings or generate misleading benchmark conclusions. Once they are addressed with representative tests, I would be happy to take another look.

Jonxjdong (commented):
Thank you for the detailed review — very helpful. Here's how I addressed each point in commit `e8ed0e72`.

### Point 1 — Explicit invocation per platform

Done. Both `SKILL.md` files now show the Codex (`$<skill>`) and Claude Code (`/<skill>`) forms side-by-side in the Quick-start section, with a one-line note that these skills start serving processes and occupy NPU resources, so they must be invoked explicitly (the frontmatter already sets `disable-model-invocation: true`).

### Point 2 — Parse real evalscope perf v1.8.1 output

Done. Both report generators now parse the real layout instead of the assumed snake_case schema:

- Walk `<results_dir>/<timestamp>/<name>/parallel_<P>_number_<N>/` and read the paired `benchmark_summary.json` (flat dict) + `benchmark_percentile.json` (list) per concurrency level.
- Read the human-readable keys directly: `Concurrency`, `TTFT (ms)`, `TPOT (ms)`, `Output Throughput (tok/s)`, `Avg Latency (s)`, plus `TTFT (ms)`/`TPOT (ms)`/`Latency (s)` from the percentile rows for p50/p90/p99.
- `Avg Latency (s)` / `Latency (s)` are converted from seconds to ms; TTFT/TPOT stay in ms.
- A sanitized fixture (`tests/ut/agents/fixtures/evalscope_v181/`, c=1/4/8 paired files) is checked in, and UT now asserts parsing + the unit conversion against it. JSONL-of-flat-summaries is kept as a fallback.

### Point 3 — Remove the no-op `VLLM_ASCEND_ENABLE_ACLGRAPH`; wire `--no-aclgraph` to `--enforce-eager`

Done. Confirmed `VLLM_ASCEND_ENABLE_ACLGRAPH` is not defined in `vllm_ascend/envs.py` and has zero reads in the runtime. Removed it from the 5 configs, `run_benchmark.sh`, and the report's env display list. `--no-aclgraph` now sets `NO_ACLGRAPH` and appends `--enforce-eager` to the vLLM server args (vLLM-Ascend honors it: `_use_aclgraph()` returns False, `capture_model()` is skipped). A dry-run UT asserts the generated `server_cmd.txt` contains `--enforce-eager` only when `--no-aclgraph` is passed.

### Point 4 — `block_size=128` default + audit unimplemented knobs

Done. `tuning-knobs-reference.md` `block_size` default corrected `16` → `128` (910C platform default; required for xlite_graph). Audited every knob against `vllm_ascend/`; removed the 5 with zero runtime implementation: `fuse_allreduce_rms`, `prefill_comm_compute_overlap`, `weight_prefetch_config`, `multistream_overlap_gate`, `multistream_dsa_preprocess`. Kept `multistream_overlap_shared_expert` and `multistream_dsv4_dsa_overlap` (implemented in `ascend_config.py`). Tuning-order lists updated accordingly.

All in commit `e8ed0e72`. `pre-commit` / `lint-and-dry-run` pass on CI; `pytest --noconftest tests/ut/agents/test_serving_skill_scripts.py` passes locally (5 tests).

PR #11120, opened 2026-06-29
[Refactor][EPLB] Support mixed expert weight layouts

freyfwt:

This PR refactors EPLB expert weight handling so layers can use different expert weight layouts in the same model.

- Adds per-layer expert weight keys based on quantization type and fused MC2 state.
- Builds EPLB buffer tensors per expert weight key, with shape validation before reusing buffers across layers.
- Uses the layer's expert weight key when generating D2D expert transfer tasks.
- Propagates EPLB heat collection status into draft model forward contexts so registered draft MoE layers do not stay at zero load.
- Fixes the non-last PP rank `forward_end()` path to use the current heat collection status.

This also avoids `Expert hotness` mean becoming `nan` when registered draft or PP-stage MoE layers did not collect heat.

No.

- Local formatting/preflight:
  - `bash format.sh ci`
- Local UT:
  - `tests/ut/eplb/adaptor/test_vllm_adaptor.py`
  - Result: `6 tests OK`
- NPU focused UT:
  - `python -m pytest -q --confcutdir=tests/ut/eplb tests/ut/eplb --tb=short`
  - Result: `19 passed, 17 warnings in 24.77s`

- vLLM version: v0.23.0

### DIFF
--- a/.github/workflows/scripts/run_selected_tests.sh
+++ b/.github/workflows/scripts/run_selected_tests.sh
@@ -32,6 +32,15 @@ pytest_log_dir="${RUNNER_TEMP:-/tmp}/selected-tests-${npu_type}-${num_npus}card"
 
 mkdir -p "${pytest_log_dir}"
 
+setup_vllm_cache_root() {
+  if [ "${CI:-}" != "true" ]; then
+    return
+  fi
+  export VLLM_CACHE_ROOT
+  VLLM_CACHE_ROOT="$(mktemp -d "${RUNNER_TEMP:-/tmp}/vllm-cache-${npu_type}-${num_npus}card.XXXXXX")"
+  echo "Using vLLM cache root: ${VLLM_CACHE_ROOT}"
+}
+
 print_test_info() {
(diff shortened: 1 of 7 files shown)


linfeng-yuan (approved):
1. Please update a RuntimeError when unsupported quantype are used with EPLB function.
2. Please add a CI with quantized verifier and float drafter.

Please update these problems in your next PR

QwertyJack:
The current CPU selected-test failure is from the MiniMax usage-accounting test on base, not from the EPLB diff:

`tests/ut/patch/platform/test_patch_minimax_usage_accounting.py` imports vLLM's private `_WrappedParser`, but the CI vLLM refs used here no longer export it, so pytest fails during collection.

I opened #11149 with a one-file fix that keeps the wrapped-parser coverage using `SimpleNamespace(reasoning_parser=...)` instead of the private helper. Focused validation against the CI-pinned vLLM commit `967c5c3b` passed:

`python -m pytest -q tests/ut/patch/platform/test_patch_minimax_usage_accounting.py`

Result: `24 passed, 16 warnings in 0.19s`.

PR #11123, opened 2026-06-29
[Misc] Add debug logging probes for speculative decoding diagnostics

chengduxiaowu:

  Adds 4 DEBUG-level log probes to the speculative decoding pipeline in
  `rejection_sampler.py` and `model_runner_v1.py` to make production-time
  triage feasible without code changes or reproduction.

  The probes are placed at customer-facing diagnostic points:
  - `propose_draft_token_ids` entry — which spec algorithm dispatched, `k`, batch size
  - `propose_draft_token_ids` after `_propose` — draft token shape & placeholder ratio
  - `rejection_sampler` entry — target-logits / metadata shape contract
  - `rejection_sampler` exit — per-batch acceptance signal

  All probes are gated by `logger.isEnabledFor(logging.DEBUG)`. At the default
  INFO level the cost is a single Python `int` compare per forward step — no
  NPU sync, no tensor allocation, no behavioral change.

  No. DEBUG-level logs only; default INFO level emits no new output.

  - Manually verified on NPU with `VLLM_LOGGING_LEVEL=DEBUG`: all 4 probes
    fire at the expected lifecycle points with field values matching
    runtime state.
  - Verified with default `VLLM_LOGGING_LEVEL=INFO`: gated blocks emit no
    output, throughput unchanged within measurement noise.
  - No new unit tests added: pure observability with no new code path or

  All probes are gated by `logger.isEnabledFor(logging.DEBUG)`. At the default
  INFO level the cost is a single Python `int` compare per forward step — no
  NPU sync, no tensor allocation, no behavioral change.

  No. DEBUG-level logs only; default INFO level emits no new output.

  - Manually verified on NPU with `VLLM_LOGGING_LEVEL=DEBUG`: all 4 probes
    fire at the expected lifecycle points with field values matching
    runtime state.
  - Verified with default `VLLM_LOGGING_LEVEL=INFO`: gated blocks emit no
    output, throughput unchanged within measurement noise.
  - No new unit tests added: pure observability with no new code path or
    branch; existing spec-decode tests continue to pass unchanged.

- vLLM version: v0.24.0

### DIFF
--- a/vllm_ascend/sample/rejection_sampler.py
+++ b/vllm_ascend/sample/rejection_sampler.py
@@ -175,7 +175,11 @@ class AscendRejectionSampler(RejectionSampler):
                 Contains the final output token IDs and their logprobs if
                 requested.
         """
-        assert metadata.max_spec_len <= MAX_SPEC_LEN
+        assert metadata.max_spec_len <= MAX_SPEC_LEN, (
+            f"rejection_sampler.forward: max_spec_len={metadata.max_spec_len} "
+            f"exceeds MAX_SPEC_LEN={MAX_SPEC_LEN}; rejection sampling cannot "
+            "proceed."
+        )
 
         self._log_rejection_sampler_entry(logits, metadata)
         bonus_logits_indices = metadata.bonus_logits_indices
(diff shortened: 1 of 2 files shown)


zzzzwwjj (approved):

vllm_ascend/sample/rejection_sampler.py
```diff
@@ -174,7 +175,39 @@ def forward(
...
                 requested.
         """
+        if metadata.max_spec_len > MAX_SPEC_LEN:
+            logger.error(
+                "[spec/dfx] rejection_sampler.forward: max_spec_len=%d exceeds "
+                "MAX_SPEC_LEN=%d; rejection sampling cannot proceed.",
+                metadata.max_spec_len,
+                MAX_SPEC_LEN,
+            )
         assert metadata.max_spec_len <= MAX_SPEC_LEN
```
zzzzwwjj: move `assert` into if, and use raise error.

vllm_ascend/sample/rejection_sampler.py
```diff
@@ -394,6 +452,17 @@ def rejection_sample(
...
+            "[spec/dfx] rejection_sample: target/draft row mismatch - "
+            "target_logits.shape[0]=%d != num_tokens(draft_token_ids)=%d "
+            "(batch_size=%d, max_spec_len=%d). Target output does not align "
+            "with draft tokens; spec decode results will be incorrect.",
+            target_logits.shape[0],
+            num_tokens,
+            batch_size,
+            max_spec_len,
+        )
     assert target_logits.shape[0] == num_tokens
```
zzzzwwjj: ditto

vllm_ascend/sample/rejection_sampler.py
```diff
@@ -493,6 +562,10 @@ def rejection_sample(
     # For random sampling with selected logits
     # target_logits is [num_tokens, top_k*tp_size] with indices [num_tokens, top_k*tp_size]
     if target_indices is not None:
+        if draft_probs is None:
+            logger.error(
```
zzzzwwjj: Does it not need to raise an error if the error log is printed here?

zzzzwwjj (changes requested):

vllm_ascend/sample/rejection_sampler.py
chengduxiaowu: you're right that logging at error level without raising is inconsistent. The reason it doesn't raise here is that
  draft_probs is None is a legitimate code path, not a bug: it's the ngram spec-decode case (the docstring notes draft_probs "Can be None ... for ngram spec decode")  So I changed this logger.error to logger.warning.

vllm_ascend/sample/rejection_sampler.py
chengduxiaowu: OK

vllm_ascend/sample/rejection_sampler.py
chengduxiaowu: OK

zzzzwwjj (approved):

vllm_ascend/sample/rejection_sampler.py
```diff
@@ -394,7 +450,15 @@ def rejection_sample(
     assert draft_probs is None or draft_probs.is_contiguous()
     assert target_logits.is_contiguous()
     assert bonus_token_ids.is_contiguous()
-    assert target_logits.shape[0] == num_tokens
+    if target_logits.shape[0] != num_tokens:
```
drslark: You can just add a assert message here.

vllm_ascend/sample/rejection_sampler.py
```diff
@@ -174,7 +175,37 @@ def forward(
                 Contains the final output token IDs and their logprobs if
                 requested.
         """
-        assert metadata.max_spec_len <= MAX_SPEC_LEN
+        if metadata.max_spec_len > MAX_SPEC_LEN:
```
drslark: I think an assert message is better.

vllm_ascend/sample/rejection_sampler.py
```diff
@@ -174,7 +175,37 @@ def forward(
...
+            )
+
+        # rejection sampler entry probe.
+        # Captures the shape/dtype baseline at the rejection-sampling boundary
+        # so that mismatches between the target-model output and the spec
+        # decode metadata can be diagnosed without re-running the workload.
+        # All payload fields are host scalars / tuple shapes -- no device sync.
+        # The whole block is gated by isEnabledFor(DEBUG) so production builds
+        # pay only the cost of one integer compare.
+        if logger.isEnabledFor(logging.DEBUG):
```
drslark: Why need this if?

vllm_ascend/sample/rejection_sampler.py
```diff
@@ -229,6 +260,31 @@ def forward(
             ori_target_logits=raw_target_logits,
         )
 
+        # rejection sampler exit probe (acceptance signal).
+        # Reports placeholder fill rate and an approximate acceptance ratio so
+        # operators can tell at a glance whether the draft/target pairing is
+        # healthy. The .ne()/.sum() calls force a host sync and only run when
+        # DEBUG is on; production paths skip the entire block.
+        if logger.isEnabledFor(logging.DEBUG):
```
drslark: Is it possible to separate the logging logic from the main logic?

drslark (changes requested):

drslark (approved):
LGTM.

zzzzwwjj (approved):

PR #11129, opened 2026-06-29
[BugFix]Block count calculation error in MambaManager when hitting the external cache

zqs164:
- When external cache hit, MambaManager will allocates 1 block for external cache + 1 block for new scheduled tokens, But upstream MambaManager#get_num_blocks_to_allocate says it only needs one extra block besides the speculative blocks, which isn't right, this pr overrides the get_num_blocks_to_allocate method to return one more block for this condition

- vLLM version: v0.23.0


MengqingCao (commented):
LGTM, thx! will remove this patch after https://github.com/vllm-project/vllm/pull/46892

MengqingCao (approved):

PR #11191, opened 2026-06-30
[BugFix][310P] Fix MTP spec decode crashes from index_fill and GatherV2 OOB

csoulnd:
**1. Replace `index_fill_` in `prepare_next_token_ids_padded`**
On 310P, `InplaceIndexFill` fails on `DT_INT32` tensors (`error code 161002`). This is hit in `prepare_next_token_ids_padded` when marking discarded requests in the padded `sampled_token_ids` batch (e.g. mixed prefill/decode batches or concurrent requests with non-empty `discard_request_indices`).
The fix replaces `index_fill_` with a `torch.where`-based mask and skips the discard path when the index tensor is empty. Semantics match the base `AscendSpecDecodeBaseProposer` implementation.
**2. Clamp token ids in `AscendVocabParallelEmbedding310.forward`**
When `tp_size == 1`, out-of-range token ids (notably `-1` placeholders from rejection sampling or scheduler spec tokens) reach NPU `GatherV2` without vocab masking, causing AICore hard faults (`retCode=0x26`, invalid GM access). The draft model path does not run target-side placeholder sanitization, so embedding lookup must tolerate these ids safely.
The fix clamps ids to `[0, num_embeddings - 1]` before the gather. Rejected/padding positions are discarded later by the rejection sampler, so this does not affect valid outputs.
#10640 
No

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/_310p/ops/vocab_parallel_embedding.py
+++ b/vllm_ascend/_310p/ops/vocab_parallel_embedding.py
@@ -57,6 +57,17 @@ class AscendVocabParallelEmbedding310(AscendVocabParallelEmbedding):
         if quant_config is None:
             self.quant_method = AscendUnquantizedEmbeddingMethod310()
 
+    def forward(self, input_):
+        # The base path only masks out-of-range token ids when tp_size > 1.
+        # When tp_size == 1 the raw ids go straight into the embedding gather,
+        # and the NPU GatherV2 kernel hard-faults on out-of-range ids (e.g. the
+        # -1 placeholders produced by speculative decoding, which the rejection
+        # sampler drops afterwards). Clamp into the valid range to keep results
+        # correct while avoiding invalid GM access on device.
+        if self.tp_size == 1:
+            input_ = input_.clamp(0, self.num_embeddings - 1)
+        return super().forward(input_)
+
 
 class AscendParallelLMHead310(AscendParallelLMHead):
     """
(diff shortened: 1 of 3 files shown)


vllm_ascend/_310p/spec_decode/llm_base_proposer_310.py
```diff
@@ -42,9 +42,22 @@ def prepare_next_token_ids_padded(
         self.backup_next_token_ids.copy_to_gpu(num_reqs)
 
         discard_sampled_tokens_req_indices = discard_request_indices[:num_discarded_requests]
-        valid_sampled_token_ids_gpu = sampled_token_ids.clone()
+        valid_sampled_token_ids_gpu = sampled_token_ids
         if discard_sampled_tokens_req_indices.numel() != 0:
-            valid_sampled_token_ids_gpu.index_fill_(0, discard_sampled_tokens_req_indices, -1)
+            request_indices = torch.arange(
```
zzzzwwjj, line 47: It is best to mask the operator differences introduced by the chip through DeviceAdaptor. Please implement this difference in DeviceAdaptor instead of directly modifying it here.

vllm_ascend/_310p/ops/vocab_parallel_embedding.py
```diff
@@ -57,6 +57,17 @@ def __init__(
...
 
+    def forward(self, input_):
+        # The base path only masks out-of-range token ids when tp_size > 1.
+        # When tp_size == 1 the raw ids go straight into the embedding gather,
+        # and the NPU GatherV2 kernel hard-faults on out-of-range ids (e.g. the
+        # -1 placeholders produced by speculative decoding, which the rejection
+        # sampler drops afterwards). Clamp into the valid range to keep results
+        # correct while avoiding invalid GM access on device.
+        if self.tp_size == 1:
+            input_ = input_.clamp(0, self.num_embeddings - 1)
```
zzzzwwjj, line 68: Why A2/A3 don't need to do this? Please identify the root cause of the problem.

zzzzwwjj (changes requested):

PR #11211, opened 2026-06-30
[Performance] Delete the tl.constexpr of total_batch param in triton_rms_kernel

LebudiPrince:
In the triton_rms_kernel, the parameter TOTAL_BATCH should not be marked as tl.constexpr, because it will cause frequent recompilation and degrade performance.

And according to profiling before and after the change, the kernel's performance is almost identical. Performance is also comparable for similarly shaped inputs.

Here are some cases:

|Shape|Averave Time Before(us)|Averave Time After(us)|
|---|---|---|
|[81984, 512]|141.420|148.260|
|[896, 512]|11.640|12.750|
|[704, 512]|11.0|12.560|
|[512, 512]|9.450|10.680|
|[320, 512]|10.450|10.190|
|[224, 512]|10.40|10.220|
|[128, 512]|7.0|7.130|

Related to [issue #11209](https://github.com/vllm-project/vllm-ascend/issues/11209)

No.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/ops/triton/rms_norm.py
+++ b/vllm_ascend/ops/triton/rms_norm.py
@@ -8,20 +8,20 @@ def triton_rms_kernel(
     hidden_state_stride_bs,
     norm_output_ptr,
     variance_epsilon,
-    TOTAL_BATCH: tl.constexpr,
+    total_batch,
     DIM: tl.constexpr,
     BLOCK_M: tl.constexpr,
 ):
     core_id = tl.program_id(0)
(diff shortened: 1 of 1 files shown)


weijinqian0 (approved):

realliujiaxu:
Is there any performance change for this kernel?

LebudiPrince:
> Is there any performance change for this kernel?

The PR description now includes performance comparison tables for a few cases. The performance differences are minor, and they are acceptable compared to the regression already introduced by compilation.

PR #11227, opened 2026-07-01
[CI] main2mian v0.24.0

zhangxinyuehfad:

This PR upgrades the verified vLLM main commit to ee0da84ab9e04ac7610e28580af62c365e898389 (v0.24.0 tag) and adapts vLLM Ascend to upstream API changes introduced after the previous verified commit.
Changes

# Changes

### `examples/offline_data_parallel.py`

- Slice `ASCEND_RT_VISIBLE_DEVICES` according to the local DP rank before creating the `LLM` instance.
- Fall back to `torch.npu.device_count()` when `ASCEND_RT_VISIBLE_DEVICES` is not set.
- Keep application-level DP compatible with the upstream removal of automatic device isolation.
- Upstream source: [vllm#45026](https://github.com/vllm-project/vllm/pull/45026).

---

### `tests/e2e/conftest.py`

- Slice `ASCEND_RT_VISIBLE_DEVICES` for each DP rank before creating `LLM()` instances or worker processes.
- Fall back to `torch.npu.device_count()` when the environment variable is unavailable.
- Ensure each test process uses an independent NPU subset after upstream removed automatic device isolation.
- Upstream source: [vllm#45026](https://github.com/vllm-project/vllm/pull/45026).

---

### `tests/ut/quantization/methods/test_w4a16_mxfp4.py`

- Skip the test to match the upstream behavior.

---

### `tests/ut/spec_decode/test_speculators_vwn_eagle3.py`

- Mock `vllm.v1.attention.selector._cached_get_attn_backend()` for non-NPU unit tests.
- Avoid attention backend initialization failures introduced by the updated v0.24.0 initialization path.
- Keep spec decode unit tests runnable without physical NPU devices.

---

### `vllm_ascend/_310p/worker_310p.py`

- Use `MemorySnapshot(device=device)` for non-0.23.0 releases.
- Align the 310P worker with the updated worker implementation.
- Related upstream changes:
  - [vllm#30868](https://github.com/vllm-project/vllm/pull/30868)

---

### `vllm_ascend/patch/platform/patch_dp_device_ids.py`

- Add a patch for `get_physical_gpu_ids_for_local_dp_rank()`.
- Support pre-sharded `ASCEND_RT_VISIBLE_DEVICES` by avoiding the upstream DP-rank offset when device isolation is already handled externally.
- Prevent `IndexError` for application-level DP after the upstream device isolation changes.
- Wire the patch into the platform initialization for non-0.23.0.
- Upstream source: [vllm#45026](https://github.com/vllm-project/vllm/pull/45026).

---

### `vllm_ascend/patch/platform/patch_kv_cache_utils.py`

- Patch `_get_kv_cache_config_deepseek_v4()` on v0.23.0.
- Patch `_get_kv_cache_config_packed()` for non-0.23.0 by reusing the existing non-packed implementation.
- Avoid excessive KV-cache allocation introduced by the packed KV-cache layout and prevent NPU OOM.
- Upstream source: [vllm#46205](https://github.com/vllm-project/vllm/pull/46205).

---

### `vllm_ascend/patch/worker/__init__.py`

- Monkey-patch `vllm.v1.utils.CpuGpuBuffer.__init__()` to remap `torch.uint64` to `torch.int64` on non-0.23.0.
- Preserve the original behavior for all other data types.
- Avoid the Ascend runtime failure caused by unsupported `DT_UINT64` in `aclnnInplaceZero`.
- Upstream source: [vllm#44665](https://github.com/vllm-project/vllm/pull/44665).

---

### `vllm_ascend/spec_decode/ngram_proposer.py`

- Preserve the original initialization path on v0.23.0.
- Delay the Ascend-specific `propose()` implementation during base-class initialization on newer releases.
- Avoid eager GPU-only initialization introduced upstream while preserving the original behavior afterward.
- Upstream source: [vllm#29184](https://github.com/vllm-project/vllm/pull/29184).

---

### `vllm_ascend/worker/utils.py`

- Correct the type annotation of `kernel_block_sizes` to match its actual nested-list usage.
- Keep the implementation consistent with the upstream interface.

---

### `vllm_ascend/worker/worker.py`

- Adjust the local device rank for single-node application-level DP on non-Ray backends.
- Match the upstream device assignment behavior after automatic device isolation was removed.
- Ensure each DP worker is mapped to the correct NPU subset.
- Upstream source: [vllm#45026](https://github.com/vllm-project/vllm/pull/45026).

---

### `vllm_ascend/__init__.py`

- Reorganize imports to eliminate circular dependencies during package initialization.

- vLLM version: v0.23.0


examples/offline_data_parallel.py
```diff
@@ -123,8 +123,20 @@ def main(
     os.environ["VLLM_DP_MASTER_IP"] = dp_master_ip
     os.environ["VLLM_DP_MASTER_PORT"] = str(dp_master_port)
 
-    # CUDA_VISIBLE_DEVICES for each DP rank is set automatically inside the
-    # engine processes.
+    # vLLM v0.24.0 (PR #45026) stopped isolating devices per worker process
```
MengqingCao: I think we should add `_build_device_ids` in `vllm/entrypoints/openai/dp_supervisor.py` in the example.

tests/e2e/conftest.py
```diff
@@ -872,6 +872,30 @@ def _run_vllm_runner_dp_worker(conn, llm_kwargs: dict[str, Any], dp_rank: int, d
...
 
+        # vLLM v0.24.0 (PR #45026) stopped isolating devices per worker
+        # process automatically. For application-level DP (one LLM() per
+        # rank), each child process must see its own slice of
+        # ASCEND_RT_VISIBLE_DEVICES, otherwise HCCL complains that two
+        # independent processes on the same node claim the same physical
+        # device.
+        from vllm_ascend.utils import vllm_version_is
+
+        if not vllm_version_is("0.23.0"):
```
MengqingCao, line 877: I think we should add `--device_ids` like the above comment, instead of setting the `ASCEND_RT_VISIBLE_DEVICES` here, this is conflict with vllm#45026

tests/ut/core/test_profiling_chunk.py
```diff
@@ -90,6 +90,7 @@ def test_invalid_min_chunk_raises(self):
             ProfilingChunkConfig({"min_chunk": 0})
 
     @patch("vllm.config.VllmConfig.__post_init__", MagicMock())
+    @patch("vllm.config.device.DeviceConfig.__post_init__", MagicMock())
```
MengqingCao, line 93: why introducing the mock of post init of device config here?

tests/ut/quantization/methods/test_w4a16_mxfp4.py
```diff
@@ -55,7 +60,9 @@ def test_process_weights_transposes_weights(self):
...
         layer.w2_weight_scale = nn.Parameter(torch.randint(0, 255, (8, 128, 8), dtype=torch.uint8), requires_grad=False)
         self.scheme.process_weights_after_loading(layer)
-        self.assertEqual(layer.w13_weight.shape, (8, 128, 32))
+        self.assertEqual(layer.w13_weight.shape, (8, 128, 256))
         self.assertEqual(layer.w13_weight_scale.shape, (8, 4, 256))
-        self.assertEqual(layer.w2_weight.shape, (8, 256, 16))
+        self.assertEqual(layer.w2_weight.shape, (8, 256, 128))
         self.assertEqual(layer.w2_weight_scale.shape, (8, 8, 128))
+        self.assertEqual(layer.w13_weight.dtype, torch.float32)
+        self.assertEqual(layer.w2_weight.dtype, torch.float32)
```
MengqingCao: why do we change the assertion check in the test case?

vllm_ascend/patch/worker/__init__.py
```diff
@@ -76,6 +76,38 @@
 if not vllm_version_is("0.23.0"):
     import vllm_ascend.patch.worker.patch_fused_moe  # noqa
 
+# Ascend NPU does not support DT_UINT64 in aclnnInplaceZero.
+# CpuGpuBuffer with dtype=torch.uint64 (used by MambaCopyBuffers)
```
MengqingCao: It seems more reasonable to patch `MambaCopyBuffers` instead of `CpuGpuBuffer`?

vllm_ascend/worker/v2/spec_decode/rejection_sampler_utils.py
```diff
@@ -337,6 +337,7 @@ def rejection_sample(
     # [num_speculative_steps]
     synthetic_conditional_rates: torch.Tensor | None = None,
     use_fp64: bool = False,
+    use_block_verification: bool = False,
```
MengqingCao, line 343: plz add a todo here as this is a new feature for speculative decoding in vllm, but we already implement the block verify in vllm-ascend before, maybe we should refactor it later

vllm_ascend/__init__.py
```diff
@@ -75,3 +73,6 @@ def register_model():
     from .models import register_model
 
     register_model()
+
+
+import vllm_ascend.logger  # noqa: E402, F401
```
MengqingCao, line 78: why moving this here?

vllm_ascend/worker/worker.py
```diff
@@ -395,6 +395,25 @@ def initialize_cache(self, num_gpu_blocks: int, num_cpu_blocks: int) -> None:
...
+                and parallel_config.nnodes_within_dp == 1
+            ):
+                tp_pp_world_size = parallel_config.pipeline_parallel_size * parallel_config.tensor_parallel_size
+                visible_device_count = torch.npu.device_count() if torch.npu.is_available() else 0
+                if visible_device_count > tp_pp_world_size:
+                    dp_local_rank = parallel_config.data_parallel_rank_local
+                    if dp_local_rank is None:
+                        dp_local_rank = parallel_config.data_parallel_index
+                    self.local_rank += dp_local_rank * tp_pp_world_size
+
```
MengqingCao, line 444: I think you can adapt the `init_device` just as that of `gpu_worker.py` https://github.com/vllm-project/vllm/pull/45026/changes?diff=unified#diff-62b35645ec2e0a565f1367fda869f582ca181cdc04d7b8ca36db71218216d6d7

examples/offline_data_parallel.py
zhangxinyuehfad: done

tests/e2e/conftest.py
zhangxinyuehfad: done

tests/ut/core/test_profiling_chunk.py
zhangxinyuehfad: In v0.24.0, DeviceConfig.__post_init__ calls torch.device("npu") and VllmConfig.__post_init__ calls try_verify_and_update_config(), both of which require actual NPU hardware. These are pure unit tests without NPU devices, so we mock these __post_init__ methods to bypass device initialization. v0.23.0 lacked these validations, hence no mocks were needed.

tests/ut/quantization/methods/test_w4a16_mxfp4.py
zhangxinyuehfad: [This pr](https://github.com/vllm-project/vllm-ascend/pull/11437) skips this use case and notes subsequent fixes.No need to modify it now

vllm_ascend/patch/worker/__init__.py
zhangxinyuehfad: Sure, this is more elegant

vllm_ascend/worker/v2/spec_decode/rejection_sampler_utils.py
zhangxinyuehfad: done

vllm_ascend/__init__.py
zhangxinyuehfad: To avoid circular imports when vLLM's plugin chain triggers during logger loading.

vllm_ascend/worker/worker.py
zhangxinyuehfad: I will align it

vllm_ascend/worker/worker.py
```diff
@@ -395,7 +396,57 @@ def initialize_cache(self, num_gpu_blocks: int, num_cpu_blocks: int) -> None:
...
+                parallel_config.distributed_executor_backend not in ("ray", "external_launcher")
+                and parallel_config.data_parallel_backend != "ray"
+                and parallel_config.nnodes_within_dp == 1
+                # vllm-ascend: when the user pre-shards devices via
+                # --device-ids (which becomes assigned_physical_gpu_ids),
+                # each child process already binds to its own NPU(s); the
+                # DP local_rank shift below would push local_rank past the
+                # length of the per-rank device list and trip the assert
+                # in this same method. Skip the shift in that case.
+                and parallel_config.assigned_physical_gpu_ids is None
```
MengqingCao, line 415: when using ray as the distributed backend, `device-ids` doesn't have effect, thus let's remove this check.

zhangxinyuehfad: ok

MengqingCao (approved):

MengqingCao:
all ci passed in https://github.com/vllm-project/vllm-ascend/actions/runs/28728597438/job/85190525764

PR #11236, opened 2026-07-01
[Feature][Ops] Increase MAX_MTP from 8 to 16 for recurrent_gated_delta_rule

Liuchenbing-2026:
## What this PR does

Bumps the `MAX_MTP` tiling constant in the `recurrent_gated_delta_rule` op from **8 → 16** in three files:

- `csrc/attention/recurrent_gated_delta_rule/op_host/recurrent_gated_delta_rule_tiling.cpp`
- `csrc/attention/recurrent_gated_delta_rule/op_kernel/recurrent_gated_delta_rule.h`
- `csrc/attention/recurrent_gated_delta_rule/op_kernel/arch35/recurrent_gated_delta_rule.h`

Plus a follow-up commit addressing review feedback (arch35 gama UB accounting, see below).

## Why

The old cap of 8 blocked MTP-3/5/7 speculative-decoding block sizes on Qwen3.5 GDN-style targets — the kernel would silently reject or misbehave once the per-request MTP-consumed token count exceeded 8.

## Scope

- **A2 / A3 (910b / 910c) and arch35 (950)**: covered by this PR.
- **310P (`recurrent_gated_delta_rule_v310`)**: intentionally **out of scope**. Qwen3.5-GDN MTP speculative decoding targets A2/A3, so `v310` keeps `MAX_MTP = 8` in both its host tiling and kernel. If 310P MTP > 8 is needed later, it should be a separate change (and a separate UB re-check for the v310 kernel).

## Review fix — arch35 gama UB accounting

Raised in review: the arch35 kernel carved an extra `gamaInUb` slice from `tmpBuff` (`InitLocalBuffers`) that the host-side UB sizing never reserved — `CalcWorkingUbBytes` accounts for the beta working buffer but not a gama one. With `MAX_MTP` doubled to 16, this unaccounted `betaNumAlign`-float requirement also doubles, so tiling could pick a `vStep` that leaves insufficient `tmpBuff` when `hasGama_` is true → risk of UB overflow / silent corruption.

Fix (aligns arch35 with the generic kernel rather than growing host accounting):
- Reuse `gamaInQueue_` directly: `gamaInUb = gamaInQueue_.DeQue<float>()` → in-place `Exp` → `PipeBarrier<PIPE_V>` → free at the batch boundary.
- Drop the `tmpBuff` `gamaInUb` carve-out (and the now-unused `betaUbSize`).
- `gamaInQueue_` is already counted in `CalcFixedUbBytes`, so the shared host UB accounting is now correct for both kernels.

## User-facing change

None (compile-time constant).

## Test plan

- [x] Compiles with the new constant on A2/A3.
- [ ] Regression: existing MTP-1/2 workloads unchanged.
- [ ] MTP-3/5 on Qwen3.5-GDN targets no longer capped.
- [ ] **`seqLen > 8` on the arch35 (950) path with `hasGama_ = true`** — validates the UB-accounting fix (no overflow / correct output vs. reference).

Co-authored with @mazhixin000 and @wangzhao-11a

- vLLM version: v0.23.0

### DIFF
--- a/csrc/attention/recurrent_gated_delta_rule/op_host/recurrent_gated_delta_rule_tiling.cpp
+++ b/csrc/attention/recurrent_gated_delta_rule/op_host/recurrent_gated_delta_rule_tiling.cpp
@@ -49,7 +49,7 @@ const size_t DIM_1 = 1;
 const size_t DIM_2 = 2;
 const size_t DIM_3 = 3;
 
-const size_t MAX_MTP = 8;
+const size_t MAX_MTP = 16;
 
 template <typename T1, typename T2>
 static T1 CeilDiv(T1 a, T2 b)
--- a/csrc/attention/recurrent_gated_delta_rule/op_kernel/arch35/recurrent_gated_delta_rule.h
+++ b/csrc/attention/recurrent_gated_delta_rule/op_kernel/arch35/recurrent_gated_delta_rule.h
@@ -25,7 +25,7 @@ using namespace AscendC;
 using namespace AscendC::MicroAPI;
 constexpr uint64_t BUFFER_NUM = 1;
 constexpr uint32_t MAX_OUT_BUFFER_NUM = 2;
-constexpr uint64_t MAX_MTP = 8;
+constexpr uint64_t MAX_MTP = 16;
 constexpr uint64_t BF16_NUM_PER_BLOCK = 16;
 constexpr uint64_t FP32_NUM_PER_BLOCK = 8;
 constexpr uint32_t REPEAT_LENTH = 64; // 256Byte for float
--- a/csrc/attention/recurrent_gated_delta_rule/op_kernel/arch35/recurrent_gated_delta_rule.h
+++ b/csrc/attention/recurrent_gated_delta_rule/op_kernel/arch35/recurrent_gated_delta_rule.h
@@ -116,7 +116,6 @@ public:
         uint32_t vSize = MAX_MTP * alignV_ * sizeof(float);
         uint32_t kSize = MAX_MTP * alignK_ * sizeof(float);
         uint32_t betaNumAlign = Ceil(MAX_MTP * NV_, BF16_NUM_PER_BLOCK) * BF16_NUM_PER_BLOCK;
-        uint32_t betaUbSize = betaNumAlign * sizeof(float); //  8: 8 * 4 = 32B;
         pipe_->InitBuffer(qInQueue_, BUFFER_NUM, MAX_MTP * alignK_ * sizeof(inType));
         pipe_->InitBuffer(kInQueue_, BUFFER_NUM, MAX_MTP * alignK_ * sizeof(inType));
         pipe_->InitBuffer(vInQueue_, BUFFER_NUM, MAX_MTP * alignV_ * sizeof(inType));
@@ -147,8 +146,11 @@ public:
         broadTmpInUb = tmpBuff.GetWithOffset<float>(static_cast<uint32_t>(alignK_ * vStep_), buffOffset);
         buffOffset += cubeSize;
         betaInUb = tmpBuff.GetWithOffset<float>(static_cast<uint32_t>(betaNumAlign), buffOffset);
-        buffOffset += betaUbSize;
-        gamaInUb = tmpBuff.GetWithOffset<float>(static_cast<uint32_t>(betaNumAlign), buffOffset);
+        // gamaInUb is NOT carved from tmpBuff. It reuses the gamaInQueue_ tensor
+        // directly (see CopyInGamaBeta), matching the generic kernel. Otherwise
+        // the host-side UB accounting (CalcWorkingUbBytes, which reserves beta
+        // but not gama in tmpBuff) under-counts by betaNumAlign floats, and the
+        // shortfall doubles with MAX_MTP -> risk of tmpBuff overflow.
     }
 
     __aicore__ inline void ComputeAvgload()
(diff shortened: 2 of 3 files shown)


ZT-AIA (approved):

drslark (approved):
It is very important for Qwen3.5 with draft.
LGTM.

Liuchenbing-2026:
Hi @wangshuai09 @Yikun,

CI has been consistently failing on **`310p-4 card` with `IndexPut aicpu 0x2a`**, which is **unrelated to this PR**. The same failure has reproduced 100% deterministically (not a flake) — requesting maintainer help to unblock.

## Full error

```
Aicpu kernel execute failed, device_id=2, stream_id=26, task_id=5306,
soName=libpt_kernels.so, funcName=RunCpuKernel, kernelName=IndexPut,
errorCode=0x2a
rtStreamSynchronize execution failed, reason=aicpu exception
RuntimeError: ACL stream synchronize failed, error code:507018
```

## Why this is unrelated to the PR

- This PR only modifies 3 files under `csrc/attention/recurrent_gated_delta_rule/` (bumping `MAX_MTP` from 8 to 16), affecting **arch35 (910B)** and the generic kernel.
- The crash happens on **Ascend310P** (not 910B).
- The crashing op is a **CANN system op `IndexPut`** (from `libpt_kernels.so`), completely unrelated to the GDN kernel touched by this PR.
- The failing tests (`test_dense_model_310p / test_moe_model_310p / test_vl_model_310p`) don't exercise `recurrent_gated_delta_rule` at all.
- **The same code fails identically on both vllm main and v0.23.0** → not a version compat issue with this PR.
- **All other ~40 jobs pass**, including every 910B / a3-2 / a3-4 job → confirms the PR code itself is correct.
- Notably, this PR is currently the only recent PR that touches `op_kernel/op_host` files with substantive changes, so it's the only PR that actually triggers the `310p-4 card` job in CI — other recent PRs skip this job because their file scope doesn't match, which is why they appear to pass. This makes this PR the first to hit the CANN 9.1.0-beta.1 `310p-4` regression.

## Hypothesis

The `310p-4 card` runner uses image `9.1.0-beta.1-310p-ubuntu22.04-py3.12`. This looks like a deterministic bug in the **CANN 9.1.0-beta.1 `IndexPut` aicpu implementation** on Ascend310P + 4-card setup.

Request:
1. Please approve / bypass the 310p-4 job for this PR, **or**
2. Forward to the infra team to investigate the CANN 9.1.0-beta.1 environment issue.

Thanks!

maoxx241 (changes requested):
Thanks for the contribution. Increasing `MAX_MTP` from 8 to 16 for `recurrent_gated_delta_rule` looks useful for unblocking larger MTP speculative-decoding cases, but I think there are a couple of scope/correctness points that should be clarified before this is merged.

One concern is the `arch35` path. It still appears to allocate an extra `gamaInUb` slice from `tmpBuff`, while the host-side UB sizing only accounts for the q/k/v/beta working buffers. With `MAX_MTP` doubled to 16, this unaccounted gamma tmp-buffer requirement also doubles, so tiling may select a `vStep` that does not actually leave enough `tmpBuff` space when `hasGama_` is true. This could lead to UB memory overflow or silent corruption on the arch35 path.

Could we either align the `arch35` implementation with the generic kernel by reusing `gamaInQueue_.DeQue<float>()` directly and removing the extra `gamaInUb` allocation from `tmpBuff`, or explicitly include this extra arch35 gamma tmp-buffer allocation in the host-side UB calculation?

Another point is scope: `recurrent_gated_delta_rule_v310` still appears to keep `MAX_MTP = 8` in both host tiling and kernel code. If 310P is intentionally out of scope for the Qwen3.5-GDN / A2/A3 use case, please mention that in the PR description. Otherwise, this PR may only partially remove the MTP cap.

It would also be helpful to add or report a runtime test covering `seqLen > 8`, especially for the `arch35` path with `hasGama_ = true`.

Liuchenbing-2026:
Thanks for the careful review @maoxx241 — all three points addressed in `b49d6ae7`.

**1. arch35 gama UB accounting (the correctness concern).** You're right: `InitLocalBuffers` carved an extra `gamaInUb` slice from `tmpBuff` that `CalcWorkingUbBytes` never reserved (it only counts the beta working buffer), so the shortfall doubled with `MAX_MTP`. I took your first suggested option — **align arch35 with the generic kernel** rather than growing the host accounting:
- `gamaInUb = gamaInQueue_.DeQue<float>()` → in-place `Exp` → `PipeBarrier<PIPE_V>` → `FreeTensor` at the batch boundary (mirrors the generic kernel's `Process()`),
- removed the `tmpBuff` `gamaInUb` carve-out and the now-unused `betaUbSize`.

Since `gamaInQueue_` is already counted in `CalcFixedUbBytes`, the shared host UB accounting is now correct for both kernels, and there's no unaccounted `tmpBuff` requirement to scale with `MAX_MTP`.

**2. 310P scope.** Intentional — `recurrent_gated_delta_rule_v310` keeps `MAX_MTP = 8`. Qwen3.5-GDN MTP targets A2/A3, so 310P is out of scope for this PR; I've documented that explicitly in the PR description (and noted a v310 bump would need its own UB re-check).

**3. Test.** Added a `seqLen > 8` / arch35 / `hasGama_ = true` item to the test plan. Running the NPU validation for that path now and will report the numbers here.

Could you take another look when you have a moment? Thanks!

maoxx241 (approved):
Thanks for the update. I re-checked the latest head b49d6ae7 against the current main. The arch35 fix now matches the generic kernel: gamaInUb is reused from gamaInQueue_, kept live across the per-batch head processing, and freed at the batch boundary. Since CalcFixedUbBytes already accounts for the gamma queue and CalcWorkingUbBytes only needs the beta tmpBuff slice, the previous unaccounted tmpBuff gamma requirement is gone.

The 310P scope is now documented as intentionally out of scope, and the current selected CI is green across the v0.23.0 and vLLM-main jobs. LGTM.

linfeng-yuan (approved):

PR #11254, opened 2026-07-01
[Test] Add and update Triton rejection sampling kernel tests

chengduxiaowu:
Here's the filled-out PR description:

  This PR adds kernel-level tests for the speculative-decoding rejection-sampling
  Triton kernels in `vllm_ascend/ops/triton/reject_sample.py`, and fixes the
  existing rejection-sample test that was broken by kernel signature changes.

  Motivation:
  - The Triton kernels `rejection_random_sample_kernel` and
    `rejection_random_sample_block_verify_kernel` had gained new parameters
    (`target_indices`, `global_vocab_size`, `ori_target_probs`,
    `NO_ORI_TARGET_PROBS`, `ENABLE_REDUCE_SAMPLING`, `ENTROPY_VERIFY`, ...), but
    the test still called them with the old signature, so the whole file failed
    at kernel launch with `TypeError: dynamic_func() missing ... arguments`.
  - Several rejection-sampling kernels had no dedicated kernel-level coverage, so
    regressions in these kernels could only surface via end-to-end spec-decode
    runs.

  Changes:
  - Add deterministic, manually-constructed PyTorch-vs-Triton parity tests for
    four kernels (each runs its input for many iterations):
    - `rejection_greedy_sample_spec_len_1_triton`
    - `rejection_greedy_sample_triton`
    - `expand_kernel`
    - `sample_recovered_tokens_kernel`
  - Update the existing kernel calls (`rejection_random_sample_kernel`,
    `rejection_random_sample_block_verify_kernel`) to match the current kernel
    signatures.
  - Switch `test_rejection_random_sample` to compare the Ascend Triton kernel
    against the Ascend PyTorch reference (`rejection_random_sample_pytorch`)
    instead of the upstream vLLM Triton kernel, so it no longer breaks when the
    upstream kernel signature changes, and is consistent with the other
    PyTorch-vs-Triton tests in this file.
  - Re-enable two tests that were previously skipped
    (`test_rejection_random_sample`, `test_rejection_sampler_block_verify_triton_kernel`).

  No. This PR only touches tests
  (`tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_rejection_sample.py`);
  there is no change to runtime behavior, APIs, or interfaces.

  Ran the full test file on Ascend NPU hardware:

  ```bash
  pytest -sv ./tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_rejection_sample.py --tb=short

  Result: 17 passed. All newly added kernel tests and the re-enabled tests pass.
```

- vLLM version: v0.23.0

### CODE REVIEW

File: tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_rejection_sample.py [hunk 1]
```diff
@@ -19,7 +31,208 @@ def setup_device_properties():
     yield
 
 
-@pytest.mark.skip("Probabilistic failure, need zengtian after fix")
+@torch.inference_mode()
```
zhenwenqi2024, line 34: Currently, there are three scheduling modes—pure prefill, pure decode, and prefill+decode—and we need to ensure all use cases are covered.

zhenwenqi2024 (commented):
Currently, there are three scheduling modes—pure prefill, pure decode, and prefill+decode—and we need to ensure all use cases are covered.

tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_rejection_sample.py
```diff
@@ -19,7 +31,208 @@ def setup_device_properties():
...
+        )
+        torch.npu.synchronize()
+        assert torch.equal(output_token_ids_ref, output_token_ids_triton), f"iteration {i}"
+
+    gc.collect()
+    torch.npu.empty_cache()
+    torch.npu.reset_peak_memory_stats()
+
+
+# @pytest.mark.skip("Probabilistic failure, need zengtian after fix")
```
realliujiaxu: delete this line

chengduxiaowu: ok

tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_rejection_sample.py
chengduxiaowu: pure prefill does not invoke the post‑sampling (rejection sampling) kernel, so there is no need to treat prefill separately. For pure decode and mixed prefill+decode scenarios, the kernel inputs are simply tensors like draft_token_ids, cu_num_draft_tokens, target_argmax, etc. The different draft_tokens_per_req values across requests (including the case where the number of draft tokens is 0) already reflect the differences introduced by mixed scheduling, while the kernel logic itself remains completely identical.

The current tests, with draft_tokens_per_req = [3,2,1,0,3,2] and parameterized batch_size, max_spec_len etc., already cover the input distributions of both pure decode and mixed scenarios.
Therefore, I believe the current tests already cover all the use cases you mentioned, and no additional scheduling‑mode‑specific handling is required for the rejection sampling kernel.

tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_rejection_sample.py
chengduxiaowu: please review again

PR #11257, opened 2026-07-01
[Performance] Reduce DSA CP KV projection redundancy

jianzs:

This updates the DSA context-parallel KV projection path to run `wkv` on `hidden_states_local`, then gather the normalized KV tensor before rotary and KV cache updates.

For TP execution this avoids redundant KV projection work on every rank while preserving the full-token KV layout required by the cache update path. In DeepSeek V4 Pro profiling with `max_num_batched_tokens=8192` and `tp_size=8`, the per-rank KV projection workload is reduced from 8192 tokens to 1024 tokens. The added all-gather is cheaper than the removed redundant projection work; the measured section dropped from about 400 us to about 200 us.

No. This is an internal performance optimization for DSA context-parallel attention.

- `python -m py_compile vllm_ascend/attention/context_parallel/dsa_cp.py`

- vLLM version: v0.23.0


jianzs:
@realliujiaxu  PTAL

PR #11258, opened 2026-07-01
[Refactor][ops]Replace hardcoded max_output_size with configurable mega_moe_max_tokens

justice-dance:
The mega moe (dispatch_ffn_combine) fused operator previously used a hardcoded max_output_size of 131072, which is both wasteful in typical scenarios and unsafe under long-context or small-EP configurations.
Add mega_moe_max_tokens to additional_config, defaulting to 65536. Users can override it via: --additional-config '{"mega_moe_max_tokens": 262144}'
This PR replaces the hardcoded max_output_size=131072 in the mega moe (dispatch_ffn_combine) fused operator with a configurable mega_moe_max_tokens option exposed through additional_config.
Why:
The original value 131072 was chosen for max_num_batched_tokens=16384 scenarios but is wasteful in typical workloads (M ≤ 2048 per rank after TP split → actual need ~16K, wasting ~2 GB workspace memory). Under long-context or small-EP configurations (e.g., max_num_batched_tokens=131072, EP=4), the actual per-rank receive count can exceed 131072, causing buffer overflow and silently dropping tokens — degrading accuracy. The workspace memory allocated by the operator scales linearly with this value (~21 KB per token slot for int8 path), so a reasonable default balances memory efficiency and correctness across scenarios.
Changes:
ascend_config.py: added mega_moe_max_tokens config, default 65536
moe_comm_method.py: replaced the hardcoded 131072 with get_ascend_config().mega_moe_max_tokens

Yes. A new additional_config key mega_moe_max_tokens is introduced for users to adjust the max token buffer limit. Default value is 65536 without extra config; users can specify custom values for long-context tasks via the launch argument: vllm serve model --additional-config '{"mega_moe_max_tokens": 262144}'

The code modification is lightweight and does not change original operator logic, only replacing hardcoded number with config reading. Existing E2E test suite tests/e2e/nightly/single_node/ops/multicard_ops_a3/test_dispatch_ffn_combine*.py covers dispatch_ffn_combine operator logic and will verify the new config after setting mega_moe_max_tokens in test env. Additional service validation will be performed: launch vLLM service with multiple mega_moe_max_tokens values and send curl requests to confirm normal inference service functionality.

- vLLM version: v0.23.0


realliujiaxu (changes requested):
please add discription to `additional_config.md`

realliujiaxu (approved):
LGTM

PR #11287, opened 2026-07-01
[Ascend950][Feature]Add allgatherEP MXFP4 quantization

Eric-dot:

This PR adds MXFP4 communication quantization support for the AllGather + EP MoE path on Ascend 950.

Previously, the AllGather EP path did not fully support MXFP4 activations in the MoE communication and dispatch flow. This PR updates the fused MoE path to correctly handle MXFP4 quantized activations during token dispatch and expert computation.

Main changes:

Add MXFP4 quantization handling for the AllGather + EP MoE path.
Pass activation quantization dtype into npu_moe_init_routing_v2 through act_quant_type, so the routing operator can correctly identify quantized activation dtype.
Add torch.uint8 as a supported hidden state dtype in fused MoE, which is required for MXFP4 packed activations.
Update token dispatch logic to use the correct quant_mode for MXFP4.
Keep MXFP4 dispatch unquantized in the AllGather path and perform quantization in the MoE MLP path.
Update W4A4 MXFP4 MoE apply logic to correctly handle torch.uint8 activations.
Add unit test coverage for quant mode selection in token dispatch.

This enables the AllGather EP path to work with MXFP4 communication quantization and improves compatibility for W4A4/MXFP4 MoE execution on Ascend 950.

No.

How was this patch tested?
Added unit test coverage for token dispatch quant mode selection.
Verified MXFP4 quant mode is correctly passed to npu_moe_init_routing_v2 in the AllGather EP path.
And already check the Deepseek v3.1 profiling to ensure it works.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/ops/fused_moe/token_dispatcher.py
+++ b/vllm_ascend/ops/fused_moe/token_dispatcher.py
@@ -374,9 +374,7 @@ class TokenDispatcherWithAllGather(MoETokenDispatcher[MoEAllGatherCombineMetadat
         dynamic_scale = token_dispatch_input.routing.pertoken_scale
         quant_type = token_dispatch_input.quant.quant_type
         act_quant_type = (
-            token_dispatch_input.quant.mxfp.act_quant_type
-            if token_dispatch_input.quant.mxfp is not None
-            else None
+            token_dispatch_input.quant.mxfp.act_quant_type if token_dispatch_input.quant.mxfp is not None else None
         )
         global_redundant_expert_num = token_dispatch_input.routing.global_redundant_expert_num
         restore_shape = hidden_states.shape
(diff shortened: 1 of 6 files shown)


vllm_ascend/ops/fused_moe/token_dispatcher.py
```diff
@@ -373,14 +372,20 @@ def token_dispatch(
...
         # dispatch_with_quant is on but got a None dynamic_scale.
-        if with_quant and dynamic_scale is None:
+        missing_dynamic_scale = dynamic_scale is None
+        quant_mode = -1
+        if missing_dynamic_scale and quant_type == QuantType.MXFP8:
             quant_mode = 3 if is_mxfp else 1
-        else:
-            quant_mode = -1
+        elif missing_dynamic_scale and quant_type == QuantType.MXFP4:
+            quant_mode = 9 if is_mxfp else 1
```
kunpengW-code: The modification here will cause quant_mode to fail to obtain the value 1, affecting INT quantization.

Eric-dot: Thanks for correction，Fixed

MengqingCao (approved):

PR #11294, opened 2026-07-02
[BugFix][KVCache] Update num_layers to include MTP/spec-decode draft layers

tyy0829:

`KVPoolWorker.num_layers` was not accounting for the MTP/spec-decode draft layer, causing a mismatch between the worker's layer count and the actual number of KV cache layers registered. For example, a 47-layer model with 1 MTP layer would report `num_layers=47` while 48 KV cache layers were actually registered.

This PR recalculates `num_layers` as the sum of `group_num_layers` (which is computed from the actual `kv_caches` dict including ALL attention layers: main + MTP), making it the authoritative layer count for the worker.

No. This only affects the internal `num_layers` accounting in `KVPoolWorker` for PD-disaggregated KV transfer with MTP/speculative decoding.

## How was this patch tested?

Tested on Atlas 800T A2 (Ascend 910B3) with glm-4.7-flash (MTP enabled, `num_nextn_predict_layers=1`) in PD-disaggregated mode with MooncakeLayerwiseConnector. Verified the log output:
```
KVPoolWorker: updated num_layers 47 -> 48 (includes MTP/spec-decode draft layers).
```

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/distributed/ascend_store/test_config_data.py
+++ b/tests/ut/distributed/ascend_store/test_config_data.py
@@ -118,7 +118,7 @@ class TestChunkedTokenDatabase(unittest.TestCase):
     def setUp(self):
         self.meta = KeyMetadata("llama", 0, 0, 0, 0)
         self.db = ChunkedTokenDatabase([self.meta], block_size=[16], partitions=None)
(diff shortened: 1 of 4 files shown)


Pz1116 (approved):
LGTM

zzzzwwjj (approved):

PR #11310, opened 2026-07-02
[Refactor][Platform] Refactor balance schedule: drop copied run_engine_core/run_busy_loop

xmpp777:
Refactors `patch_balance_schedule.py` to drop stale verbatim upstream copies (`DPEngineCoreProc.run_busy_loop`, `EngineCoreProc.run_engine_core`) that were drifting from the pinned vLLM, while strictly preserving `balance_flag` semantics (`any-rank-at-cap => global freeze` — each rank reaches the freeze decision independently from the same gathered snapshot; there is no leader).

- Move `balance_gather` into `BalanceScheduler` (signature uses `self.dp_group`, injected by the engine core). It runs once per busy-loop iteration from `BalanceDPEngineCoreProc._process_engine_step` — **NOT** from inside `schedule()`. A rank that has drained its local requests never enters `schedule()` (it runs a dummy batch), so an `all_gather` placed in `schedule()` would be skipped by that rank while busy ranks call it: a collective mismatch that deadlocks. `_process_engine_step` is called every iteration by upstream `run_busy_loop` (incl. dummy-batch iterations), so every rank participates every step; timing matches the pre-refactor engine-core gather (after schedule + execute + `update_from_output`).
- Slim `BalanceDPEngineCoreProc` to a minimal `_process_engine_step` override that injects `dp_group` and runs `balance_gather`; delete the copied `run_busy_loop` body.
- Delete the copied `run_engine_core`. Activate `BalanceDPEngineCoreProc` ONLY when balance is enabled, via a thin `run_engine_core` wrapper that swaps the module-level `DPEngineCoreProc` at call time (where `vllm_config` is available) and restores the upstream class otherwise. This relies on a load-bearing assumption: upstream resolves `DPEngineCoreProc` at call time, not at import — if upstream ever switches to an import-time binding the swap silently stops taking effect.
- Track the pinned vLLM `schedule` signature: `schedule(self, throttle_prefills=False)` with a version branch on the disabled fast-path, since v0.23.0 lacks the `throttle_prefills` argument added in v0.23.1+.
- Route the config probe's env-var fallback through `vllm_ascend.envs` (no bare `os.getenv`); keep the three-way fallback until AscendConfig init timing is moved earlier.

The `schedule()` body is still reproduced verbatim (the `balance_flag` gate is inlined mid-method and upstream has no finer hook yet); an intent-lock test asserts the signature tracks the installed vLLM and that the balance delta lines remain present.

No. `balance_flag` semantics and behavior remain strictly unchanged.

Unit tests in `tests/ut/patch/platform/test_patch_balance_schedule.py` covering signature/intent lock, deferred conditional engine-core swap, `_process_engine_step` seam guard, config precedence, gather cadence, disabled-path delegation, and admission-gate semantics. Full `SchedulerOutput` equivalence and NPU sync profiling remain in e2e/nightly (skipped stub).

---

- vLLM version: v0.23.0

### DIFF
--- a/docs/source/developer_guide/Design_Documents/balance_schedule_refactor.md
+++ b/docs/source/developer_guide/Design_Documents/balance_schedule_refactor.md
@@ -1,0 +1,608 @@
+# Balance Schedule Refactor
+
+**TL;DR** The existing `patch_balance_schedule.py` copied three large upstream
+units verbatim to inject roughly 5 lines of real logic: `Scheduler.schedule()`
+(~520 lines), `DPEngineCoreProc.run_busy_loop()` (~40 lines), and
+`EngineCoreProc.run_engine_core()` (~55 lines). This refactor, while **strictly
+preserving the `balance_flag` semantics**, first deletes the two already-stale
+copies `run_busy_loop()` / `run_engine_core()` (replacing them with an engine
+core hook on `_has_global_unfinished_reqs` plus a module-level name swap for
+conditional activation). The `schedule()` copy is **kept for now** — upstream
+exposes no finer-grained hook to borrow, and deleting it depends on contributing
+an override seam upstream, tracked as later Phase 2B. The file therefore does
+not shrink to a few dozen lines: the `schedule()` body is still a verbatim
+upstream copy (**aligned verbatim to release tag `v0.23.0`**, with only 3
+balance deltas), and the file is ~830 lines. Aligning to a stable release tag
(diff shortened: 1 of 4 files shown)


zzzzwwjj (approved):

ningjingbengxiaohai (approved):

DavidJiang9 (commented):
lgtm

PR #11315, opened 2026-07-02
[BugFix] Add put timeout circuit breaker for AscendStore KV pool

Alexander-Qiu:

This PR adds a request-level timeout circuit breaker for AscendStore KV cache `put` operations.

Before this change, if a backend `put` call became slow or stuck, each remaining chunk for the same request could wait on the same slow path again. That could turn one stalled transfer into accumulated back-pressure across the request.

This patch:
- adds `put_timeout_s` to `kv_connector_extra_config`, defaulting to `10.0` seconds;
- wraps AscendStore KV `put` with a timeout;
- marks the request as failed after a put timeout, so later puts for the same request are skipped;
- keeps request accounting in `finally`, so timed-out or failed puts still drain the queue and release request bookkeeping;
- clears the failed-request marker when the request is cleaned up.

Yes.

AscendStore KV pool users can now configure `put_timeout_s` through `kv_connector_extra_config`. The default is `10.0` seconds.

Behavior also changes when a KV `put` stalls: after the timeout, the remaining puts for that request are skipped instead of waiting chunk by chunk.

Added unit coverage in `tests/ut/distributed/ascend_store/test_kv_transfer.py` for:
- put timeout marks the request as failed and still runs accounting;
- already-failed requests skip remaining puts;
- normal fast puts still work;
- put exceptions still run accounting;
- request cleanup clears the failed-request marker.

I attempted to run the test locally with:

```bash
python3 -m pytest tests/ut/distributed/ascend_store/test_kv_transfer.py -q

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
@@ -100,6 +100,14 @@ class KVPoolWorker:
 
         self.kv_role = vllm_config.kv_transfer_config.kv_role
         self.load_async = vllm_config.kv_transfer_config.kv_connector_extra_config.get("load_async", False)
+        # Per-chunk put timeout (seconds) for the request-level circuit breaker in
+        # KVCacheStoreSendingThread: a chunk whose put exceeds this is dropped and
+        # the rest of that request's puts are short-circuited (bounds accumulated
+        # congestion from one slow/stuck transfer). Default 10s; set <= 0 to
+        # disable the timeout.
+        self.put_timeout_s = vllm_config.kv_transfer_config.kv_connector_extra_config.get("put_timeout_s", 10.0)
+        if self.put_timeout_s is not None and self.put_timeout_s <= 0:
+            self.put_timeout_s = None
         self._invalid_block_ids: set[int] = set()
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py [hunk 1]
```diff
@@ -100,6 +100,13 @@ def __init__(
 
         self.kv_role = vllm_config.kv_transfer_config.kv_role
         self.load_async = vllm_config.kv_transfer_config.kv_connector_extra_config.get("load_async", False)
+        # Per-chunk put timeout (seconds) for the request-level circuit breaker in
+        # KVCacheStoreSendingThread: a chunk whose put exceeds this is dropped and
+        # the rest of that request's puts are short-circuited (bounds accumulated
+        # congestion from one slow/stuck transfer). Default 10s. To effectively
+        # disable, set a very large value (e.g. 1e9). Do NOT set 0 -- with
+        # result(timeout=0) every put times out instantly and all stores break.
+        self.put_timeout_s = vllm_config.kv_transfer_config.kv_connector_extra_config.get("put_timeout_s", 10.0)
```
Pz1116, line 108: we should add this parameter in our document, kv_pool.md

Issue #11296, opened 2026-07-02
[Bug]: 高精度推理性能评测后报错 request.num_output_placeholders变负报错AssertionError AsyncLLM output_handler failed


wangx700:
### Your current environment

A5 环境
cann-9.1.T560 
(/mnt/share/vllm/conda_envs/wx_test_dts) [root@localhost w00899129]# pip list | grep -i torch
torch                                    2.10.0
torch_npu                                2.10.0
torchaudio                               2.10.0
torchvision                              0.25.0
(/mnt/share/vllm/conda_envs/wx_test_dts) [root@localhost w00899129]# pip list | grep -i triton
triton                                   3.5.0
triton_ascend                            3.2.2
(/mnt/share/vllm/conda_envs/wx_test_dts) [root@localhost w00899129]# pip list | grep -i vllm
vllm                                     0.23.1.dev0+g0fc695fc6.d20260625.empty /mnt/share/w00899129/vllm_code/vllm
vllm_ascend                              0.19.1rc2.dev688+gac4d54bfe.d20260625  /mnt/share/w00899129/vllm_code/vllm-ascend

### 🐛 Describe the bug

# 复现脚本
# 服务端
vllm serve  /mnt/share/Qwen3-32B\
    --port 26186 \
    --data-parallel-size 1 \
    --tensor-parallel-size 1 \
    --served-model-name qwen3 \
    --max-model-len 8192 \
    --max-num-seqs 200 \
    --trust-remote-code

# 客户端
ais_bench --models vllm_api_stream_chat --datasets gsm8k_gen_0_shot_cot_str_perf --debug --summarizer default_perf --mode perf --num-prompts 24

# vllm_api_stream_chat
from ais_bench.benchmark.models import VLLMCustomAPIChatStream
from ais_bench.benchmark.utils.model_postprocessors import extract_non_reasoning_content

models = [
    dict(
        attr="service",
        type=VLLMCustomAPIChatStream,
        abbr='vllm-api-stream-chat',
        path="", # 置为空，否咋会报错404
        model="qwen3",
        request_rate = 0,
        retry = 2,
        host_ip = "localhost",
        host_port = 26186,
        max_out_len = 4096,
        batch_size=24,
        trust_remote_code=False,
        generation_kwargs = dict(
            temperature = 0,
            ignore_eos = True,
        ),
    )
]

# 报错日志
[...]

PR #11334, opened 2026-07-02
[Bugfix] Fix preempted-then-resumed requests for using right num_tokens

wangx700:

Ensure that when a preempted request is resumed, the vLLM-ascend model_runner_v1 correctly uses the updated num_tokens (reflecting the already-computed token length) rather than the original num_prompt_tokens, so that the prompt phase boundary is consistent with the vLLM scheduler's view and chunk_prefill is correctly identified. [issues 11296](https://github.com/vllm-project/vllm-ascend/issues/11296)

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -732,6 +732,12 @@ class NPUModelRunner(GPUModelRunner):
 
         result = super()._update_states(scheduler_output)
 
+        # For preempted-then-resumed requests, attach the prefill token
+        # count to the CachedRequestState so that its (patched) num_tokens
+        # property returns the full token history instead of just the
+        # original prompt length. The request only appears in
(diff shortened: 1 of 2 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -732,7 +730,23 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> Callable | None
                 if num_computed_tokens < req_state.num_computed_tokens:
                     req_state.prev_num_draft_len = 0
 
-        return super()._update_states(scheduler_output)
+        result = super()._update_states(scheduler_output)
```
Ronald1995, line 733: the root cause is that vllm's scheduler use model runner v2 by default, we can set model runner v2 env to false in vllm-ascend

PR #11338, opened 2026-07-02
[BugFix] Add validation for avoiding disable_padded with full_graph

slippersss:
This PR aims to fix potential errors when enabling `disable_padded_drafter_batch=True` and `cudagraph_mode` with full graph inside. Since `disable_padded_drafter_batch` can result in irregular input for draft model, e.g., different requests with different decode tokens lengths (rejected tokens are removed), full graph thus will have troubles in solving token padding. We now disable this scenario.

- vLLM version: v0.23.0


drslark (approved):
`disable_padded_drafter_batch` is a virtually useless configuration.
So, we only support `disable_padded_drafter_batch=False` with graph.
LGTM.

MengqingCao (approved):

PR #11342, opened 2026-07-02
[Bugfix] Fix ZMQ port mapping error in multi-node PP deployment

MarinaMiao:
背景issue：https://github.com/vllm-project/vllm-ascend/issues/11343

#### Problem
In multi-node PP disaggregated serving, the D-side uses P0's base port (meta.remote_port) to calculate handshake ports for ALL PP ranks, including those on other PP nodes. For ranks on P1, the computed port (remote_port + rank) does not match P1's actual listening port (side_channel_port + device_index), causing all P1 connections to time out.

#### Root cause
The code assumes all PP ranks share the same side_channel_port (base port). However, when different PP nodes have different kv_port or data_parallel_rank, each node gets a different side_channel_port from:
  side_channel_port = kv_port + dp_rank * tp_size * pp_size * pcp_size

#### Fix
Propagate per-node side_channel_port through MooncakeAgentMetadata and multi_nodes_meta_mapping, then replace hardcoded (meta.remote_port + rank) with _compute_remote_handshake_port() that uses the correct per-rank base port. Also replace _get_remote_host_info_by_port() with _get_remote_host_info_by_rank() to correctly look up host info by rank instead of deriving rank from incorrect port arithmetic.

No. This fixes a bug in multi-node PP PD disaggregated serving. Single-node deployments are unaffected (the port arithmetic happens to be correct when all nodes share the same `side_channel_port`).

Tested on internal multi-node PP disaggregated serving cluster (2 P nodes + 2 D nodes, PP=4, TP=8). Without this fix, D→P1 connections time out; with this fix, all KV cache transfers complete successfully.

- vLLM version: v0.23.0


linfeng-yuan:
@nwpu-zxr @wangxiaoteng888 Please review this PR which fixes port conflicts. And it is better to extract redundant code in a utils funtion.

wangxiaoteng888:
Has the case of multiple machines per DP been considered？

PR #11375, opened 2026-07-03
[Performance][310P] Access _npu_flash_attention_unpad_v2

YangShuai52:
Feature Integration: Added support for approximate flash attention calculation using the _npu_flash_attention_unpad_v2 operator.
Dynamic Operator Selection: Implemented a check to detect if the environment supports the new approximate calculation operator, allowing for a fallback to the original implementation.
No

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/_310p/ops/mm_encoder_attention.py
+++ b/vllm_ascend/_310p/ops/mm_encoder_attention.py
@@ -124,6 +124,9 @@ class AscendMMEncoderAttention310(MMEncoderAttention):
             v = F.pad(v, (0, pad_len), mode="constant", value=0)
 
         context_layer = torch.empty_like(q)
+        # TODO: The current torch_npu version does not support
+        # _npu_flash_attention_unpad_v2. Once it is supported, drop the
+        # else branch below and always use the v2 op.
         if self.support_approximate_calculation:
             torch_npu._npu_flash_attention_unpad_v2(
                 query=q,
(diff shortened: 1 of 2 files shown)


vllm_ascend/_310p/ops/mm_encoder_attention.py
```diff
@@ -35,6 +35,10 @@
 seq_lens_cpu_cache: torch.Tensor = None
 
 
+def is_approximate_calculation_supported() -> bool:
+    return hasattr(torch_npu, "_npu_flash_attention_unpad_v2")
```
MengqingCao, line 39: plz make sure if this is included in torch-npu 2.10.0, if so, let's use this by default for a better performance.

YangShuai52: I tried it, and version 2.10.0 does not have this interface, so it needs to be added to the branch.

vllm_ascend/_310p/ops/mm_encoder_attention.py
```diff
@@ -119,18 +124,32 @@ def forward_oot(
...
-            query=q,
-            key=k,
-            value=v,
-            seq_len=seq_lens_cpu,
-            scale_value=self.scale_value,
-            num_heads=self.num_heads,
-            num_kv_heads=self.num_kv_heads,
-            out=context_layer,
-        )
+        # TODO: The current torch_npu version does not support
```
MengqingCao:
```suggestion
        # TODO: The current torch_npu version 2.10.0 does not support
```

MengqingCao (approved):

PR #11377, opened 2026-07-03
[BugFix]fix vllm use v2 but vllm-ascend use v1

wangx700:

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/patch/worker/__init__.py
+++ b/vllm_ascend/patch/worker/__init__.py
@@ -66,6 +66,7 @@ import vllm_ascend.patch.worker.patch_cudagraph  # noqa
 import vllm_ascend.patch.worker.patch_deepseek_mtp  # noqa
 import vllm_ascend.patch.worker.patch_deepseek_v2  # noqa
 import vllm_ascend.patch.worker.patch_gqa_c8  # noqa
+import vllm_ascend.patch.worker.patch_v2.patch_use_v2_model_runner  # noqa
 
 if not vllm_version_is("0.23.0"):
     import vllm_ascend.patch.worker.patch_fused_moe  # noqa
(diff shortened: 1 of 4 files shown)


vllm_ascend/patch/worker/__init__.py
```diff
@@ -67,6 +67,13 @@
 import vllm_ascend.patch.worker.patch_deepseek_v2  # noqa
 import vllm_ascend.patch.worker.patch_gqa_c8  # noqa
 
+# vLLM's use_v2_model_runner may enable the v2 runner without the
+# VLLM_USE_V2_MODEL_RUNNER env var (e.g. based on model architecture).
+# We always patch it so that on Ascend the v2 runner is enabled only
+# when the env var is explicitly set.
+if not _V2_MODEL_RUNNER_SUPPORTED:
```
Ronald1995: why add this branch, we need to patch this in main branch

wangx700: ok, 'if not _V2_MODEL_RUNNER_SUPPORTED' was removed

vllm_ascend/patch/worker/__init__.py
```diff
@@ -1,82 +1,88 @@
...
+    import vllm_ascend.patch.worker.patch_fused_moe  # noqa
+
+if _V2_MODEL_RUNNER_SUPPORTED:
+    import vllm_ascend.patch.worker.patch_v2.patch_uva  # noqa
+    import vllm_ascend.patch.worker.patch_v2.patch_input_batch  # noqa
+    import vllm_ascend.patch.worker.patch_v2.patch_model_state  # noqa
+    import vllm_ascend.patch.worker.patch_v2.patch_block_table  # noqa
+    import vllm_ascend.patch.worker.patch_v2.patch_attn_utils  # noqa
+
+# only patch routed experts capture in main2main.
```
Ronald1995: add docstring in vllm_ascend/patch/__init__.py to explain this patch and future plan,

Ronald1995:
LGTM, the root cause is that vllm enable model_runner_v2 by default, we need to change the default value temporarily

Issue #10517, opened 2026-06-16
[RFC]: Improve Prefix-Caching Hit Rate for Hybrid Models


underfituu:
# Motivation.

This RFC tracks our efforts to optimize prefix-caching for hybrid models, specifically under PD disaggregation scenarios. It covers tracking for both upstream community PRs and implementations within the vllm-ascend repository.

# Proposed Change.

### Partial hit (FA)
- [x]  vllm
        vllm-project/vllm#42524 
        vllm-project/vllm/pull/44243

- [x] vllm-ascend
        https://github.com/vllm-project/vllm-ascend/pull/10009

  We propose introducing a targeted platform patch in vllm-ascend to exclude the Mamba group from D-side KV cache hit rate accounting under specific conditions.  (Using recomputed_scheduler) (--additional-config '{"recompute_scheduler_enable":true}')
  > This patch will be removed once the vllm-ascend repository tracks the latest upstream main branch that incorporates the aforementioned changes, or after version 0.23.1rc0
   - [ ] Remove patch
### cache retention interval:
- [x]  vllm
        https://github.com/vllm-project/vllm/pull/45845
- [x] vllm-ascend
        No adaptation required
### Related Issues

- [x] https://github.com/vllm-project/vllm-ascend/issues/7944

### Partial hit :
- [ ]  vllm
        vllm-project/vllm#45804

- [ ]  vllm-ascend
        TBD

### Ops:

- [x] https://github.com/vllm-project/vllm-ascend/pull/10354

    Makes the DeepSeek V4 compressor KV-cache block_size configurable (32 / 64 / 128) instead of hard-pinned to 128, via a lookup table that scales the MLA / SWA / C4-state / C128-state block sizes and page-size padding together. A smaller block size shrinks the alignment requirement of the compressed KV cache, so more prompt prefixes can hit the prefix cache (better automatic prefix cache hit rate).
  
    This allows the cache matching granularity to be reduced by up to 4×. Optimization of the specific operators is currently underway.

### cache retention interval:
- [x]  vllm
        vllm-project/vllm#43447
        vllm-project/vllm#44082  （fix MTP + prefix-caching bug）

- [ ]  vllm-ascend 
        https://github.com/vllm-project/vllm-ascend/pull/11107  （refer to #44082）
        https://github.com/vllm-project/vllm-ascend/pull/11383   （fix alignment_tokens != lcm_block_size bug）

### Related Issues

[...]

PR #11383, opened 2026-07-03
[BugFix][Prefix Caching][DSv4] Align SlidingWindowManager scheduler block size with LCM block size

Csrayz:

When validated against an agent-scenario dataset, the prefix-cache hit rate improved rising from 33% to 86%.

The root cause is a unit ambiguity in `AscendHybridKVCacheCoordinator`'s SWA alignment logic. Since Ascend's `block_size` semantics refer to the number of **physical slots** rather than the number of tokens (tracked in vllm-ascend RFC #10517), the `alignment_tokens` value flowing into `SlidingWindowManager.reachable_block_mask` and `SlidingWindowManager.find_longest_cache_hit` was previously expressed in mixed units of `token / max(compression_ratio)`. This produced wrong block counts when dividing `alignment_tokens` by per-block token capacity.

This PR unifies the `alignment_tokens` unit across all SWA computations: the value is now the actual `lcm_block_size` computed by the coordinator (`self.lcm_block_size` in `verify_and_split_kv_cache_groups()`), and every SWA manager receives this value as its `scheduler_block_size`. As a result, `window_size`, `block_size`, and `alignment_tokens` are uniformly expressed in actual token counts inside both `reachable_block_mask` and `find_longest_cache_hit`, and the SWA module maintains consistent unit semantics end-to-end.

No. Behavior is gated by the existing upstream env `VLLM_PREFIX_CACHE_RETENTION_INTERVAL` (vLLM PR #43447). When unset, `reachable_block_mask` returns the dense default and no SWA block selection changes. When set, the WRITE/READ paths now use the same alignment granularity instead of the previous mixed-unit one.

Validated in DeepSeek-V4 Flash (mtp=1):

- **Prefix-cache hit rate** on an agent-scenario dataset: 33%(main) → 86%(this pr)
- **SWA block cache occupation**:
  - groups 2 and 3: full retention → 9/256
  - group 4: 1/4 → 1/512
  - group 5: full retention → 1/32

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
+++ b/vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
@@ -21,6 +21,7 @@ from vllm.v1.core.kv_cache_utils import (
 )
 from vllm.v1.core.single_type_kv_cache_manager import (
     SingleTypeKVCacheManager,
+    SlidingWindowManager,
 )
 from vllm.v1.kv_cache_interface import (
     FullAttentionSpec,
@@ -145,6 +146,15 @@ class AscendHybridKVCacheCoordinator(HybridKVCacheCoordinator):
                 for g in kv_cache_config.kv_cache_groups
             ), "block_size must be divisible by hash_block_size"
         self.verify_and_split_kv_cache_groups()
+
+        # Align the WRITE-path mask granularity (reachable_block_mask) with the
+        # READ-path hit granularity (find_longest_cache_hit) so SlidingWindowManager
+        # only caches blocks that land on a boundary where future cache hits can
+        # actually be matched.
+        # TODO (Csrayz): Consider unified all single_type_managers to simplify logic.
+        for mgr in self.single_type_managers:
+            if isinstance(mgr, SlidingWindowManager):
+                mgr.scheduler_block_size = self.lcm_block_size
 
         self.use_eagle = use_eagle
 
--- a/tests/ut/patch/platform/test_prefix_cache_cp_patches.py
+++ b/tests/ut/patch/platform/test_prefix_cache_cp_patches.py
@@ -7,6 +7,9 @@ from unittest.mock import MagicMock
 import pytest
 import torch
 from vllm.v1.core.block_pool import BlockPool
+from vllm.v1.core.single_type_kv_cache_manager import (
+    SlidingWindowManager,
+)
 from vllm.v1.kv_cache_interface import (
     FullAttentionSpec,
     KVCacheConfig,
@@ -14,6 +17,7 @@ from vllm.v1.kv_cache_interface import (
     KVCacheTensor,
     MambaSpec,
     MLAAttentionSpec,
+    SlidingWindowMLASpec,
     UniformTypeKVCacheSpecs,
 )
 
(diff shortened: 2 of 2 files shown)


Csrayz:
~I am currently analyzing the feasibility of unifying the units for window_size, block_size, and alignment_tokens across all SingleTypeKVCacheManager variants to actual token counts. If this proves feasible, I believe this change is worthwhile, as it would align the internal semantics of SingleTypeKVCacheManager more closely with upstream logic. Especially given that, in the current implementation, spec.block_size for both C4A and C128A does not represent actual token counts. This would also make it easier to adapt to future upstream changes.~

After further analysis, I've found that refactoring this area would have a relatively large impact. I also noticed that the community is currently pushing forward related refactoring efforts. If there is a need to discuss the design or implementation plan, I would be more than happy to participate and contribute.

Csrayz:
Hi @wangxiyuan @MengqingCao @zzzzwwjj 

Following up on the CACHE_RETENTION_INTERVAL track for DeepSeek-V4 outlined in this RFC:

[vllm-ascend#11383](https://github.com/vllm-project/vllm-ascend/pull/11383) is a small alignment fix that complements vLLM upstream #43447. It propagates the coordinator-computed lcm_block_size to every SlidingWindowManager, ensuring that reachable_block_mask and find_longest_cache_hit operate on the same alignment token count. Previously, the mixed-unit alignment_tokens led to incorrect block counts on compressed groups.

Validated on DeepSeek-V4 Flash (mtp=1): prefix-cache hit rate improved from 33% to 86%, and SWA block retention dropped by 1–2 orders of magnitude across all swa attention groups.

linfeng-yuan:
@underfituu Please review this PR.

underfituu:
Thanks for fixing this bug.
 My understanding is that, this bugfix could avoid unnecessary block retention/waste. It also makes the replay-boundary tail handling consistent, which is important for maintaining prefix-cache hit rate when the exact same prompt is replayed. 

In particular, when `VLLM_PREFIX_CACHE_RETENTION_INTERVAL` is configured ( >0 ). Does this mean the bug no longer exists?

underfituu (approved):

vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
```diff
@@ -146,6 +147,15 @@ def __init__(
...
         self.verify_and_split_kv_cache_groups()
 
+        # Align the WRITE-path mask granularity (reachable_block_mask) with the
+        # READ-path hit granularity (find_longest_cache_hit) so SlidingWindowManager
+        # only caches blocks that land on a boundary where future cache hits can
+        # actually be matched.
+        # TODO (Csrayz): Consider unified all single_type_managers to simplify logic.
+        for mgr in self.single_type_managers:
+            if isinstance(mgr, SlidingWindowManager):
+                mgr.scheduler_block_size = self.lcm_block_size
```
underfituu, line 157: Considering that only `alignment_tokens` utilizes `scheduler_block_size`, applying a forced modification is an acceptable workaround for now. However, we will ultimately need to remove the current patch to align with the upstream community. thanks!

Csrayz:
> In particular, when `VLLM_PREFIX_CACHE_RETENTION_INTERVAL` is configured ( >0 ). Does this mean the bug no longer exists?

It still exists. The behavior is close, but there is a subtle difference in the tail alignment at the prompt boundary. With VLLM_PREFIX_CACHE_RETENTION_INTERVAL=4096 alone (while block_size=32, no fix), it retains an extra 4 blocks of SWA cache beyond the num_prompt_tokens alignment boundary. And alignment_tokens=32 rather than the correct 4096, so those 4 blocks are unlikely to be reused by a subsequent request, since the replay boundary almost never lands on the 4 blocks.

Separately, manually configuring VLLM_PREFIX_CACHE_RETENTION_INTERVAL requires a good understanding of how SWA operates under hybrid attention,  which carries a non-trivial learning curve to configure correctly. This PR removes the need for that tuning.

vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
Csrayz: Agreed, and that's why this PR keeps the change minimal. Here I just a assignment on scheduler_block_size.

I think the right long-term approach is to unify the block_size, window, and alignment_tokens semantics inside all SingleTypeKVCacheManager variants to match upstream (actual token counts). Currently, block_size = token / compression_ratio and alignment_tokens = lcm(token / compression_ratio) across groups, which causes the mixed-unit issue described in this PR. Resolving this inconsistency would make future maintenance simpler and ultimately allow removing this patch entirely.

linfeng-yuan (approved):

Csrayz:
Looks like there's something wrong with the CI environment @linfeng-yuan .

underfituu:
> Looks like there's something wrong with the CI environment @linfeng-yuan .

Could you please do a manual rebase to trigger the CI? Thanks.

Csrayz:
> > Looks like there's something wrong with the CI environment @linfeng-yuan .
> 
> Could you please do a manual rebase to trigger the CI? Thanks.

Done. I've rebased onto main and pushed the changes. I'll keep an eye on the CI status and follow up if anything unexpected comes up.

Csrayz:
> > Looks like there's something wrong with the CI environment @linfeng-yuan .
> 
> Could you please do a manual rebase to trigger the CI? Thanks.

@underfituu CI passed. Thanks!

PR #11386, opened 2026-07-03
[Feature] Support IndexCache in context-parallel DSA path

GDzhu01:

The `index_cache` feature (enabled via `--hf-overrides '{"use_index_cache": true, "index_topk_pattern": "..."}'`) lets DeepSeek-V4 layers reuse the indexer's top-k indices across layers instead of recomputing them every layer. The `index_topk_pattern` string marks each layer as `"F"` (compute top-k and write it to a shared buffer) or `"S"` (skip the indexer and read the cached top-k). This was already supported on the serial DSA path (`dsa_v1.py`) but was missing from the context-parallel (CP) DSA path (`dsa_cp.py`), so running DSV4 with `enable_dsa_cp()` + IndexCache silently computed top-k on every layer instead of reusing it.

This PR ports the IndexCache support to the CP path in `vllm_ascend/attention/context_parallel/dsa_cp.py`:

- **`AscendDSACPImpl.__init__`**: read `skip_topk`, `topk_indices_buffer` from kwargs and derive `use_index_cache` from `hf_config`, mirroring `AscendDSAImpl`.
- **Two new helpers** `_get_indexcache_topk_indices` / `_update_indexcache_topk_indices`: read/write the shared top-k buffer (copied from the serial impl).
- **`_forward`** (`compress_ratio == 4`):
  - On **skip layers** (`skip_topk=True`), bypass `_update_indexer_cache` + `_indexer_select_topk` entirely and reuse the cached top-k.
  - On **indexer layers**, run the indexer as before and refresh the shared buffer after the compressor scatter so downstream skip layers can read it.

### CP-specific design notes

- **Offset is always 0.** Unlike the serial path, which splits the batch into `[decode | prefill]` and uses `offset=num_decode_tokens` for the prefill segment, CP processes a single contiguous local token shard per step (no decode/prefill row split), so the buffer is always accessed at offset 0.
- **Buffer sizing is safe.** `topk_indices_buffer` is `[max_num_batched_tokens, index_topk]`. CP pads the input to `tokens_per_rank = ceil(num_input_tokens / tp_size) <= max_num_batched_tokens`, so each rank's local shard fits. Each TP rank holds its own model replica (and its own buffer), consistent with CP's per-rank-local semantics — no cross-rank sharing is needed.
- **Token count uses `compress_topk_idxs.shape[0]`** (== `hidden_states_local.shape[0]`, the padded local count produced by `npu_vllm_quant_lightning_indexer` from the query shape), matching the serial path's approach.

## How was this patch tested?

- Syntax-validated the modified file.
- Verified the commit applies cleanly on top of `origin/main` and only touches `vllm_ascend/attention/context_parallel/dsa_cp.py`.

End-to-end validation requires running DSV4 with `enable_dsa_cp()` and `--hf-overrides '{"use_index_cache": true, "index_topk_pattern": "..."}'` on NPU hardware; there is no existing CP+IndexCache e2e test.

- vLLM version: v0.23.0

## Performance Validation (CP + IndexCache ablation)

Ablation measured on DeepSeek-V4-Flash (w4a8-int4), 8 cards (DP=2 × TP=4), `enable_dsa_cp=true` + flashcomm1, 48 requests × 16388 input tokens × 1024 output tokens, concurrency 12.

| Metric | IndexCache OFF | IndexCache ON | Delta |
|---|---|---|---|
| TTFT (avg) | 5773.8 ms | 5435.2 ms | **-5.9%** |
| TTFT (median) | 5518.9 ms | 5267.7 ms | -4.5% |
| TPOT (avg) | 44.6 ms | 42.8 ms | **-4.0%** |
| TPOT (median) | 45.0 ms | 42.9 ms | -4.7% |
| Output Token Throughput | 238.42 tok/s | 249.19 tok/s | +4.5% |
| Total Token Throughput | 4054.06 tok/s | 4237.23 tok/s | **+4.5%** |
| Prefill Token Throughput | 2838.32 tok/s | 3015.19 tok/s | +6.2% |
| Request Throughput | 0.2328 req/s | 0.2434 req/s | +4.6% |
| E2EL (avg) | 51378.6 ms | 49249.4 ms | -4.1% |
| Benchmark Duration | 206157.9 ms | 197245.8 ms | -4.3% |

Enabling IndexCache on the CP path yields a uniform ~4–6% throughput improvement. The prefill stage benefits most (+6.2% prefill token throughput → -5.9% TTFT): with many query tokens per prefill step, skipping the indexer top-k computation on `"S"` layers (per `index_topk_pattern`) removes substantial compute. Decode steps have far fewer tokens per step, so the TPOT gain is smaller but still positive (-4.0%). The result confirms the CP IndexCache path is functionally correct and delivers the expected compute savings end-to-end.

### DIFF
--- a/vllm_ascend/attention/context_parallel/dsa_cp.py
+++ b/vllm_ascend/attention/context_parallel/dsa_cp.py
@@ -999,6 +999,21 @@ class AscendDSACPImpl(DSAAttentionImpl):
             self.compressor_norm = self.compressor.norm
             self.compressor_norm_eps = self.compressor.norm_eps
 
+        # IndexCache: skip_topk indicates this layer reuses topk from a previous
+        # indexer-bearing layer; use_index_cache marks whether the buffer must
+        # be kept fresh on non-skip layers so downstream skip layers can read.
+        # In CP, every rank holds its own model replica (and thus its own
+        # topk_indices_buffer), and each rank processes a single contiguous
+        # local token shard per step, so the buffer is always accessed at
+        # offset 0 for compress_topk_idxs.shape[0] rows.
+        self.skip_topk = kwargs.get("skip_topk", False)
+        self.topk_indices_buffer = kwargs.get("topk_indices_buffer")
+        self.use_index_cache = self.skip_topk or getattr(
+            self.vllm_config.model_config.hf_config,
+            "use_index_cache",
+            False,
+        )
+
     def _compute_compressor_metadata(
         self,
         metadata: AscendDSAReqMetadata,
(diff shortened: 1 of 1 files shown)


pisceskkk (approved):
LGTM! Thanks for contribution!

vllm_ascend/attention/context_parallel/dsa_cp.py
```diff
@@ -1029,6 +1044,25 @@ def _compute_compressor_metadata(
...
+        return topk_indices
+
+    def _update_indexcache_topk_indices(self, topk_indices: torch.Tensor, offset: int = 0) -> None:
+        if self.topk_indices_buffer is None:
+            return
+        num_tokens = topk_indices.shape[0]
+        topk_tokens = topk_indices.shape[-1]
+        topk_indices_buffer = self.topk_indices_buffer[offset : offset + num_tokens, :topk_tokens]
+        if topk_indices.dim() == 3 and topk_indices_buffer.dim() == 2:
+            assert topk_indices.shape[1] == 1
```
MengqingCao, line 1062: plz add assertion message here.

linfeng-yuan (approved):

linfeng-yuan:
Could you please add an e2e case with indexer_cache and DSA_CP？

pisceskkk:
> Could you please add an e2e case with indexer_cache and DSA_CP？

@GDzhu01 please add indexer_cache config here, and maybe modify the GOLDEN for this case.
https://github.com/GDzhu01/vllm-ascend/blob/0b6603ee89931cec5b30a967d64bbd43eeabca79/tests/e2e/pull_request/four_card/context_parallel/test_accuracy.py#L419-L421

PR #10935, opened 2026-06-25
[CI] upgrade vllm to 0619

MrZ20:
### What this PR does / why we need it?

This PR upgrades the verified vLLM main commit to `b9a7cd464c9ae9b1b450f8982b76d7be4de73724` and adapts vLLM Ascend to upstream API changes introduced after the previous verified commit.

`vllm_ascend/__init__.py`

- Set `VLLM_USE_V2_MODEL_RUNNER=0` for non-`0.23.0`.
- Reason: this branch keeps the main compatibility path on V1 model runner. V2 still has
  partial compatibility updates, but it is not enabled as the default path here.
- Related upstream changes: vLLM Model Runner V2 migration work, including
  [vllm#42667](https://github.com/vllm-project/vllm/pull/42667) and [vllm#44446](https://github.com/vllm-project/vllm/pull/44446).

`tests/e2e/pull_request/one_card/test_guided_decoding.py`

- For non-`0.23.0` and `xgrammar` / `outlines`, set
  `VLLM_REGEX_COMPILATION_TIMEOUT_S=30` during the test.
- Keep `0.23.0` behavior unchanged.
- Reason: upstream added a structured-output regex compilation timeout. The existing
  complex JSON schema can exceed the default 5s timeout during grammar compilation.
- Upstream source: [vllm#45118](https://github.com/vllm-project/vllm/pull/45118).

`vllm_ascend/core/scheduler_profiling_chunk.py`

- Add `throttle_prefills: bool = False` to `ProfilingChunkScheduler.schedule`.
- Reason: upstream scheduler `schedule()` now accepts `throttle_prefills`.
- Upstream source: [vllm#44558](https://github.com/vllm-project/vllm/pull/44558).

`vllm_ascend/patch/platform/patch_balance_schedule.py`

- Add `throttle_prefills` to `BalanceScheduler.schedule`.
- For `0.23.0`, continue calling `super().schedule()` with no argument.
- For non-`0.23.0`, forward `throttle_prefills` to upstream scheduler.
- Reason: keep the balance scheduler wrapper compatible with both old and new upstream
  scheduler signatures.
- Upstream source: [vllm#44558](https://github.com/vllm-project/vllm/pull/44558).

`vllm_ascend/patch/platform/patch_profiling_chunk.py`

- Change the schedule wrapper to accept `*args, **kwargs`.
- Reason: profiling chunk patch wraps scheduler implementations whose signatures now
  differ between `0.23.0` and upstream main.
[...]

PR #11389, opened 2026-07-03
[BugFix] fix vllm use v2 but vllm-ascend use v1

wangx700:
Add patch_use_v2_model_runner.py — gate v2 runner by env var only. Upstream vLLM enables the v2 model runner based on model architecture whitelists, Triton availability, and feature compatibility — even without VLLM_USE_V2_MODEL_RUNNER set. On Ascend the NPU v2 path is not yet compatible with all upstream-defaulted models, which can cause crashes([issues 11296](https://github.com/vllm-project/vllm-ascend/issues/11296)). This patch overrides VllmConfig.use_v2_model_runner to read only the environment variable, deferring model compatibility checks to the NPU runner itself.

- Updated worker.py and platform.py to call vllm_config.use_v2_model_runner instead of reading envs_vllm.VLLM_USE_V2_MODEL_RUNNER directly, so they benefit from the patched logic.

No.

[issues 11296](https://github.com/vllm-project/vllm-ascend/issues/11296) was resolved with this PR.

- vLLM version: v0.23.0


Ronald1995 (approved):
LGTM

MengqingCao (approved):
LGTM

MengqingCao:
let's merge this quickly as it block the main2main, and the test will be done with #10935

PR #11399, opened 2026-07-03
[Feature] Gemma4 MTP speculative decoding support on Ascend NPU on A5

swimming2007-doge:
### Gemma4 MTP k=3 acceptance rate

Single card, greedy, coding prompts:

| pos0 (1st draft) | pos1 (2nd) | pos2 (3rd) | avg accepted / turn |
|---|---|---|---|
| 74.0% | 49.2% | 26.4% | 1.50 / 3 |

- Sample: +504 drafts, +754 accepted, 15 requests, 24.9s (from `/metrics` delta).
- Strictly decreasing `pos0 > pos1 > pos2` healthy speculative-decoding signature (later drafts are progressively harder to guess correctly in sequence).
- Throughput implication: each forward produces 1 (target) + 1.5 (accepted draft) 2.5 tokens  ~2.5 theoretical speedup upper bound (actual discount applies from draft/verify overhead).

### Health checklist

| Dimension | Result |
|---|---|
| Gather path OOM bug | healthy 25K, not reproduced |
| HBM peak | flat over 120s, no peak |
| Acceptance rate | pos0 74% / 1.5 token per turn, healthy |

TODO ：GPU and Precision Test

- vLLM version: v0.25.1


vllm_ascend/ops/rotary_embedding.py
```diff
@@ -160,6 +160,54 @@ def rope_forward_oot(
...
+                cos = cos.index_select(0, positions).view(num_tokens, 1, -1)
+                sin = sin.index_select(0, positions).view(num_tokens, 1, -1)
+                if is_neox_style:
+                    q1, q2 = query.chunk(2, dim=-1)
+                    query = torch.cat(
+                        [q1 * cos - q2 * sin, q2 * cos + q1 * sin], dim=-1
+                    )
+                else:
+                    query = query * cos + query * sin
+                query = query.flatten(-2, -1)
```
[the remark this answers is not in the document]
swimming2007-doge: Q-only RoPE 的非 triton fallback 原手写 cos/sin 有 interleaved 公式错和 neox 切片 shape 错两个 bug，改成复用 torch_npu._npu_rotary_embedding + dummy key buffer，与同文件正常路径对称

vllm_ascend/platform.py
```diff
@@ -564,6 +565,11 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
                 compilation_config.cudagraph_capture_sizes = sp_aclgraph_sizes
                 update_cudagraph_capture_sizes(vllm_config, sp_aclgraph_sizes)
 
+        # MTP: FULL_AND_PIECEWISE is not yet fully supported; downgrade to
+        # PIECEWISE. FULL_DECODE_ONLY is allowed through for FDO graph mode.
+        if compilation_config.cudagraph_mode == CUDAGraphMode.FULL_AND_PIECEWISE:
+            compilation_config.cudagraph_mode = CUDAGraphMode.PIECEWISE
```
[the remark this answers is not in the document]
swimming2007-doge: 已经更新t Pull Request title and description

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2311,6 +2314,16 @@ def execute_model(
...
+        # MTP: Record NPU event after target forward completes, so the draft
+        # model can wait on it before reading the KV cache. Required when
+        # the target model runs in FDO graph mode because reshape_and_cache
+        # KV writes are async on the NPU stream.
+        _target_is_fdo = self.compilation_config.cudagraph_mode.has_full_cudagraphs()
+        if ((_target_is_fdo or os.path.exists("/tmp/vllm_sync_target_kv"))
+                and hasattr(self, 'drafter') and self.drafter is not None):
+            _target_done = torch.npu.Event()
+            _target_done.record()
+            self.drafter._target_done_event = _target_done
```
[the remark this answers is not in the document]
swimming2007-doge: 复用单个 NPU event 跨步 record()，替代每步新建 torch.npu.Event()，避免热路径上的开销和 NPU 事件资源耗尽，行为不变。

vllm_ascend/utils.py
```diff
@@ -1693,3 +1742,109 @@ def get_c_env(name: str, encoding: str = "utf-8") -> str | None:
...
+    # If original sizes exceed maximum, sample a representative subset
+    if max_num_batch_sizes < len(original_sizes):
+        step = (len(original_sizes) - 1) / (max_num_batch_sizes - 1)
+        indices = [round(i * step) for i in range(max_num_batch_sizes)]
+
+        # Ensure first and last elements are preserved
+        indices[0], indices[-1] = 0, len(original_sizes) - 1
+
+        sampled_sizes = [original_sizes[i] for i in indices]
+        update_cudagraph_capture_sizes(vllm_config, sampled_sizes)
```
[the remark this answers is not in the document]
swimming2007-doge: 已经按照建议修改

vllm_ascend/spec_decode/gemma4_proposer.py
swimming2007-doge, line 1: - 新增 Gemma4 MTP 的 Ascend draft proposer，多重继承复用上游 Gemma4 逻辑 + vllm-ascend NPU 运行时；
- 核心：跨模型 KV 共享下的 KV head 数量对齐和lm_head 不共享，
- 范围：全部逻辑仅对 model_type=="gemma4_mtp" 生效，不影响其他 spec-decode 路径。

linfeng-yuan:
@slippersss Please review the modification around model_runner_v1 ,llm_base_proposer and gemma4_proposer. CC @lilinsiman

linfeng-yuan:
@swimming2007 Please verify on A2/A3.

vllm_ascend/ops/rotary_embedding.py
```diff
@@ -160,6 +160,59 @@ def rope_forward_oot(
...
+                    positions,
+                    query,
+                    k_dummy,
+                    head_size,
+                    cos_sin_cache,
+                    is_neox_style,
+                )
+                query = query.view(num_tokens, -1, head_size).flatten(-2, -1)
+        return query, None
+
```
swimming2007-doge: 7.6 上会意见：需要单独提出来，尽量不放在公共的文件，用gemma4的门控封装

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -729,6 +743,18 @@ def _update_full_graph_params_if_needed(
         if forward_context.cudagraph_runtime_mode == CUDAGraphMode.FULL:
             self._update_full_graph_params(forward_context, num_input_tokens, multi_steps_attn_metadata)
 
+    def _sync_wait_target_events(self) -> None:
+        """Wait for NPU events recorded after target forward completes.
```
swimming2007-doge: 7.6上会意见：这里要用gemma4的门控控制，另外主要代码应当放到gemma4_proposer.py这里面

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2338,6 +2341,19 @@ def execute_model(
             hidden_states = self._model_forward(
                 num_tokens_padded, input_ids, positions, intermediate_tensors, inputs_embeds, **model_kwargs
             )
+        # MTP: Record NPU event after target forward completes, so the draft
+        # model can wait on it before reading the KV cache. Required when
+        # the target model runs in FDO graph mode because reshape_and_cache
+        # KV writes are async on the NPU stream.
+        _target_is_fdo = self.compilation_config.cudagraph_mode.has_full_cudagraphs()
+        if ((_target_is_fdo or os.path.exists("/tmp/vllm_sync_target_kv"))
```
swimming2007-doge: 7.6上会意见：测试代码应当删除

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -497,7 +502,15 @@ def _maybe_share_lm_head(self, model: nn.Module) -> None:
...
                     layer_module.shared_head.head = model.lm_head
 
-        if self.vllm_config.compilation_config.cudagraph_mode.has_full_cudagraphs() and self.use_cuda_graph:
+        _has_full = self.vllm_config.compilation_config.cudagraph_mode.has_full_cudagraphs()
+        _is_gemma4_mtp = self.speculative_config is not None and self.speculative_config.use_gemma4_mtp()
+        # On Ascend, Gemma4 MTP draft runs in eager mode (FDO is not supported
+        # for the draft). Disable cuda graph only for Gemma4 MTP so other MTP
+        # models keep the upstream FULL-cudagraph draft wrapping below.
+        if _has_full and _is_gemma4_mtp and self.use_cuda_graph:
+            self.use_cuda_graph = False
```
slippersss: If you wanna disable cudagraph, then do it in `__init__`. DO NOT put this kind of logic everywhere carelessly.

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -707,6 +720,10 @@ def dummy_run(
             if forward_context is not None:
                 forward_context.moe_layer_index = 0
 
+            # Gemma4 MTP: draft 读 target 的 KV cache，需等 target 异步 KV 写完成。
+            # 非 Gemma4 模型 _sync_wait_target_events 是 no-op，这里门控跳过调用。
+            if self._is_gemma4_mtp:
+                self._sync_wait_target_events()
```
slippersss: Does vllm community achieve gemma4 mtp in this way? I don't think this piece of code should be placed here.

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -730,6 +747,25 @@ def _update_full_graph_params_if_needed(
...
+
+    def _sync_wait_target_events(self) -> None:
+        """Wait for NPU events recorded after target forward completes.
+
+        No-op in the shared base: only Gemma4 MTP reads the target model's
+        KV cache and thus needs to wait on the target's async KV writes.
+        AscendGemma4Proposer overrides this with the real event-wait; all
+        other draft proposers own their KV cache and fall through here.
+        """
+        return
```
slippersss: Same question about the way how vllm community achieve it. If you have no-op here, why do you judge `if self._is_gemma4_mtp:` in `dummy_run` and `_propose`?

vllm_ascend/platform.py
```diff
@@ -590,7 +624,17 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
             ascend_config.ascend_compilation_config.enable_npugraph_ex = False
+        elif (
+            compilation_config.cudagraph_mode == CUDAGraphMode.FULL_DECODE_ONLY
+            or compilation_config.cudagraph_mode == CUDAGraphMode.FULL
+        ):
+            logger.info(
+                "FULL_DECODE_ONLY compilation enabled on NPU. use_inductor not supported - using only ACL Graph mode"
+            )
+            compilation_config.use_inductor = False
+            compilation_config.splitting_ops = []
```
slippersss: Is it also a part of this PR? Until now, I feel the entire pull request is ai-style and irresponsible, without any prior checks.

slippersss:
Inconsistent comment styles (Chinese and English) and messy code content (plenty of code content about cudagraph, etc.) indicate that the authors did not check their PR carefully. I think this PR is not ready for reviewing right now. I suggest referring to the implementation of vllm community and carefully revising it.

vllm_ascend/attention/attention_v1.py
```diff
@@ -1473,6 +1492,10 @@ def reshape_and_cache(
                 # see: https://github.com/vllm-project/vllm/blob/ce88756b967c2c5006746a424c15dd59a284ed8c/vllm/model_executor/layers/attention/cross_attention.py#L117
                 slot_mapping=slots[: attn_metadata.num_actual_tokens] if not encoder_decoder else slots.to(torch.int32),
             )
+            # 通知 KV-transfer connector KV 已写入。无 connector 时 no-op
+            # （Gemma4 MTP 用 in-process KV-sharing，无 connector）。不用
+            # _is_gemma4_mtp() 门控（见上，draft forward 误判）。
+            notify_kv_cache_written()
```
drslark, line 1579: Why does `attention_v1` include the logic for KV transfer?
This reflects an unreasonable layering of logic.

vllm_ascend/spec_decode/gemma4_proposer.py
```diff
@@ -0,0 +1,432 @@
...
+    Gemma4-MTP-specific (no other model passes key=None), hence the helper
+    lives in the Gemma4 proposer module rather than the shared rotary op.
+
+    Returns the rotated query; the caller discards the (None) key.
+    """
+    import torch_npu
+    from vllm.triton_utils import HAS_TRITON
+
+    num_tokens = query.shape[0]
+    if HAS_TRITON:
```
drslark: Why is there an operator in proposer?

drslark (changes requested):

drslark:
> Inconsistent comment styles (Chinese and English) and messy code content (plenty of code content about cudagraph, etc.) indicate that the authors did not check their PR carefully. I think this PR is not ready for reviewing right now. I suggest referring to the implementation of vllm community and carefully revising it.

I couldn't agree more. 
`vllm-ascend` is merely a plugin for `vllm`; it shouldn't involve so much logic that completely rewrites the original implementation.

vllm_ascend/utils.py
```diff
@@ -560,7 +560,55 @@ def attention_calculation_stream() -> torch.npu.Stream:
     return _ATNN_CALCULATION_STREAM
 
 
+def _patch_get_config_hf_overrides():
```
Alex-stack-hub: This global monkey patch of `ModelConfig.__post_init__` feels too broad for this PR. It is not scoped to Gemma4 MTP, has no idempotency guard, and can wrap `__post_init__` multiple times if `adapt_patch()` is called more than once. Since recent vLLM already contains the Gemma4 assistant -> `gemma4_mtp` config override path, could we remove this patch and rely on the normal override mechanism, or move any remaining compatibility fix into a much narrower/version-gated path?

vllm_ascend/utils.py
```diff
@@ -1747,3 +1795,105 @@ def get_c_env(name: str, encoding: str = "utf-8") -> str | None:
...
+    # Maximum number of graphs that can be captured by ACL Graph
+    MAX_CAPTURE_SIZE = 1800
+
+    # enable pcp or dcp will add new communication and consume additional
+    # approximately less than 100 streams
+    CP_ADDITIONAL_STREAM_NUM = 100
+
+    # Store original configuration and temporarily clear it
+    compilation_config = vllm_config.compilation_config
+    original_sizes, compilation_config.cudagraph_capture_sizes = (compilation_config.cudagraph_capture_sizes, None)
```
Alex-stack-hub: This clears `compilation_config.cudagraph_capture_sizes` before the function knows whether it will write back a sampled list. If `max_num_batch_sizes >= len(original_sizes)` or the function returns early above, the original capture sizes are lost and left as `None`. This also seems unrelated to Gemma4 MTP support, so I suggest removing it from this PR or making it a separate, self-contained fix that preserves the original sizes on all paths.

vllm_ascend/attention/utils.py
```diff
@@ -542,3 +549,325 @@ def enabling_mlapo(vllm_config: VllmConfig) -> bool:
...
+    q = query[:num_tokens]  # [T, H, D]
+    k = shared_key  # [S, Hkv, D]
+    v = shared_value  # [S, Hkv, D]
+
+    # Create causal cross-attention mask.  Query position i can attend
+    # to KV positions [0, kv_len - q_len + i].  For sliding-window
+    # layers, additionally restrict to the last `sliding_window` tokens.
+    S = k.shape[0]
+    offset = S - num_tokens
+    mask = torch.ones(num_tokens, S, dtype=q.dtype, device=q.device) * float("-inf")
```
Alex-stack-hub: This constructs a single `[num_tokens, S]` causal mask over the flattened TND batch. For batch size > 1, tokens from one request can attend to KV tokens from another request because request boundaries from `actual_seq_lengths_q` / `seq_lens_list` are not represented. Could we make this mask block-diagonal per request, or reuse a varlen attention path that respects the per-request sequence lengths?

Alex-stack-hub (commented):
Thanks for working on Gemma4 MTP support. I reviewed this against the current main branch and the recent Gemma4/A2/A5 refactors. My overall suggestion is to narrow this PR to the minimum Ascend-specific MTP delta: reuse upstream vLLM Gemma4Proposer logic where possible, reuse the existing vllm-ascend large-head fallback and paged-KV gather utilities, and remove unrelated global platform/config changes from this PR. The current patch mixes the core Gemma4 MTP implementation with broader graph-size/config monkey patches and duplicated proposer/attention logic, which makes the review surface much larger and introduces a few correctness risks.

vllm_ascend/attention/utils.py
```diff
@@ -542,3 +549,325 @@ def enabling_mlapo(vllm_config: VllmConfig) -> bool:
...
+            read_vc,
+            block_table,
+            seq_lens,
+            impl.num_kv_heads,
+            impl.head_size,
+        )
+        # Zero-value fallback: if the routed block_table gave all-zero KV
+        # (wrong pool), try the other groups' block_tables.  Reverse
+        # iteration prefers the global-attention group (highest gid).
+        _k_mean = dense_key.float().mean().item()
```
Alex-stack-hub: Using `dense_key.float().mean().item()` as a wrong-pool detector is risky in the attention hot path: `.item()` forces an NPU->CPU sync, and the zero/non-zero heuristic is data dependent. The broad `except Exception` fallback below also hides real routing or shape bugs. It would be safer to rely on explicit `gid -> block_table` routing and fail/log clearly if the expected group metadata is missing, rather than silently trying other groups based on tensor values.

vllm_ascend/spec_decode/gemma4_proposer.py
```diff
@@ -0,0 +1,428 @@
...
+        draft_index: int = 0,
+    ):
+        from copy import copy
+
+        per_group_attn_metadata: list[object] = []
+        per_layer_attn_metadata: dict[str, object] = {}
+        for attn_group in self.draft_attn_groups:
+            gid = attn_group.kv_cache_group_id
+            if gid in self._per_group_block_tables:
+                cm = copy(common_attn_metadata)
```
Alex-stack-hub: This method duplicates upstream vLLM `Gemma4Proposer.build_per_group_and_layer_attn_metadata()`, but the upstream implementation slices the stored block table to the actual batch size before building drafting metadata. Without that slice, padded block-table rows can leak into the draft metadata. Can we reuse the upstream implementation here instead of overriding it, or at least keep the same `[:batch_size]` behavior?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2426,6 +2428,18 @@ def execute_model(
             hidden_states = self._model_forward(
                 num_tokens_padded, input_ids, positions, intermediate_tensors, inputs_embeds, **model_kwargs
             )
+        # MTP: Record NPU event after target forward completes, so the draft
+        # model can wait on it before reading the KV cache. Required when
+        # the target model runs in FDO graph mode because reshape_and_cache
+        # KV writes are async on the NPU stream.
+        _target_is_fdo = self.compilation_config.cudagraph_mode.has_full_cudagraphs()
+        if (_target_is_fdo and hasattr(self, 'drafter') and self.drafter is not None):
```
Alex-stack-hub: The event is recorded for any drafter whenever the target uses full cudagraphs, but only Gemma4 MTP reads the target KV cache and waits on this event. To avoid extra hot-path work and NPU event pressure for Eagle/DFlash/normal draft-model speculation, could we gate the record side on `AscendGemma4Proposer` or `speculative_config.use_gemma4_mtp()` as well?

Alex-stack-hub:
From an upstreamability perspective, I suggest narrowing this PR to the minimum Ascend-specific Gemma4 MTP delta.

Please remove unrelated global changes such as `ModelConfig.__post_init__` monkey patching and ACL graph size recalculation from this PR. These are not Gemma4 MTP-specific and should be handled separately if still needed.

For Gemma4 proposer logic, please reuse upstream vLLM `Gemma4Proposer` as much as possible. Avoid copying methods like `build_per_group_and_layer_attn_metadata` unless Ascend needs a real behavior difference. If overridden, please preserve upstream semantics such as slicing per-group block tables to the actual batch size.

For attention, please keep `attention_v1.py` as thin as possible. The PR should only add small hooks for Gemma4 MTP KV-sharing and large-head graph capture gaps, while reusing existing vllm-ascend utilities for paged KV gather and A2/A3 512-head fallback.

Please also fix the correctness issues before merge:

1. shared-KV prefill attention must respect per-request boundaries instead of using one global flattened causal mask;
2. avoid broad exception swallowing and tensor-value heuristics when selecting KV cache groups;
3. gate target KV event recording to Gemma4 MTP only;
4. make Q-only RoPE return shape consistent across Triton and non-Triton paths.

swimming2007-doge:
> @swimming2007 Please verify on A2/A3.

A2/A3 need another patch which relay on PA and lots of Customized Implement. I will support A2/A3‘s MTP with a patch until 930 PA offline

swimming2007-doge:
## A5 (vLLM v0.25.1) e2e verification 

### Gemma4 MTP k=3 acceptance rate

Single card, greedy, coding prompts:

| pos0 (1st draft) | pos1 (2nd) | pos2 (3rd) | avg accepted / turn |
|---|---|---|---|
| 74.0% | 49.2% | 26.4% | 1.50 / 3 |

- Sample: +504 drafts, +754 accepted, 15 requests, 24.9s (from `/metrics` delta).
- Strictly decreasing `pos0 > pos1 > pos2` healthy speculative-decoding signature (later drafts are progressively harder to guess correctly in sequence).
- Throughput implication: each forward produces 1 (target) + 1.5 (accepted draft) 2.5 tokens  ~2.5 theoretical speedup upper bound (actual discount applies from draft/verify overhead).

### Health checklist

| Dimension | Result |
|---|---|
| Gather path OOM bug | healthy 25K, not reproduced |
| HBM leak | flat over 120s, no leak |
| Acceptance rate | pos0 74% / 1.5 token per turn, healthy |

vllm_ascend/spec_decode/gemma4_proposer.py
```diff
@@ -0,0 +1,285 @@
...
+                    "(heads=%d, kv_heads=%d)",
+                    draft_idx,
+                    draft_nh,
+                    draft_nkv,
+                    tgt_name,
+                    tgt_nh,
+                    tgt_nkv,
+                )
+                object.__setattr__(attn, "num_kv_heads", tgt_nkv)
+                object.__setattr__(attn, "num_heads", tgt_nh)
```
Alex-stack-hub, line 250: suggestion: Could we double-check whether the draft model's `num_heads` needs to be overwritten here? KV sharing mainly requires the K/V layout (`num_kv_heads`, head size, cache spec) to match, while the query-head count is also tied to the draft Q projection that was created during initialization. If supported checkpoints always have `tgt_nh == draft_nh`, a small assertion or comment would make that assumption clear; otherwise, it may be safer to keep the draft query-head count and adapt only the shared KV layout.

.github/workflows/scripts/test_config.yaml
```diff
@@ -454,6 +454,7 @@
   source_file_dependencies:
     - vllm_ascend/spec_decode/llm_base_proposer.py
     - vllm_ascend/spec_decode/eagle_proposer.py
+    - vllm_ascend/spec_decode/gemma4_proposer.py
```
Alex-stack-hub, line 468: suggestion: It would be helpful to add a focused Gemma4 MTP regression test for this new proposer path. The tests currently listed in this bucket mainly cover Eagle/generic spec-decode, while Gemma4 also exercises multi-group block-table routing, Q-only RoPE, and target-to-draft synchronization. Even a small smoke test—ideally including batch > 1—would make future refactors safer.

Alex-stack-hub (changes requested):
Thanks for narrowing the patch and reusing the upstream Gemma4 proposer. I still see two correctness/coverage blockers in the current revision: the draft query-head layout is mutated after module construction, and the new path has no Gemma4 MTP regression coverage in CI. Please address these before merge.

vllm_ascend/ops/rotary_embedding.py
```diff
@@ -242,6 +243,14 @@ def forward_oot(
         is_neox_style = self.is_neox_style
         if is_neox_style_override is not None:
             is_neox_style = is_neox_style_override
+        # Gemma4 MTP Q-only: key=None (K/V from target cache). Gate on key,
+        # not config — draft forward's config has no speculative_config.
+        # See ops.rope_q_only.gemma4_q_only_rope.
+        if key is None:
```
Alex-stack-hub, line 249: nit: Since this new branch intentionally accepts `key=None` and returns `(q, None)`, could we also update the method annotation to `key: torch.Tensor | None` and add an explicit return type? That would make the Q-only contract visible to type checkers and future callers.

vllm_ascend/patch/platform/patch_speculative_config.py
```diff
@@ -16,6 +16,12 @@
 
 def hf_config_override(hf_config: PretrainedConfig) -> PretrainedConfig:
     initial_architecture = hf_config.architectures[0]
+    if hf_config.model_type in ("gemma4_assistant", "gemma4_unified_assistant"):
+        hf_config.model_type = "gemma4_mtp"
+        text_config = getattr(hf_config, "text_config", hf_config)
+        if hasattr(text_config, "num_kv_shared_layers"):
+            text_config.num_kv_shared_layers = 0
```
Alex-stack-hub, line 23: suggestion: Setting `num_kv_shared_layers` to zero is a little counter-intuitive in a feature built around KV sharing. Could we add a short comment explaining whether this disables checkpoint-level/shared-layer construction because runtime KV sharing is wired by the proposer? That context would help prevent someone from “fixing” this later.

vllm_ascend/spec_decode/gemma4_proposer.py
```diff
@@ -0,0 +1,285 @@
...
+    # ---- _fix_draft_kv_head_counts -------------------------------------------
+    # When the draft model's attention layer is configured with a different
+    # number of KV heads than the target layer it shares with (e.g. draft
+    # reports 16 KV heads from HuggingFace config but the target's
+    # full_attention uses num_global_kv_heads=4), reading that cache under the
+    # draft's head counts would misinterpret the layout.  Align the draft's
+    # num_kv_heads (and num_heads) to the target layer's values so the shared
+    # KV cache is decoded correctly on both the FIA (A5) and PA (A2/A3) paths.
+
+    def _fix_draft_kv_head_counts(self, target_model) -> None:
```
Alex-stack-hub, line 191: nit: `target_model` does not appear to be used in this method; the lookup comes from `self.vllm_config`. Would you mind removing the parameter (and updating the call site), or using it explicitly? Keeping it currently suggests that the head counts are derived from the passed module when they are not.

vllm_ascend/spec_decode/gemma4_proposer.py
```diff
@@ -0,0 +1,285 @@
...
+        from vllm.logger import logger
```
Alex-stack-hub, line 192: nit: Could we define the logger once at module scope with the repository's usual `init_logger(__name__)` pattern? There are several method-local `from vllm.logger import logger` imports in this class, and a module logger would be a bit easier to follow and mock in tests.

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -932,6 +975,22 @@ def _propose(
...
 
+        # Gemma4 MTP has multiple KV cache groups (sliding + full attention).
+        # During chunked prefill, build_draft_attn_metadata inherits the
+        # target's ChunkedPrefill attn_state.  Patch every group's output
+        # metadata to SpecDecoding so the FIA backend picks the decode
+        # kernel path.  We patch outputs rather than mutating the shared
+        # common_attn_metadata to avoid side effects on the target object.
+        # Deferred import avoids circular dependency.
+        if self.method == "mtp":
+            from vllm_ascend.spec_decode.gemma4_proposer import AscendGemma4Proposer
```
Alex-stack-hub: suggestion: The deferred import works, but it makes the shared base depend on a concrete proposer subtype. A small overridable hook such as `_prepare_draft_attn_metadata(...)` (no-op in the base, Gemma4 implementation in its proposer) might keep the layering cleaner and remove the circular-import concern. Non-blocking, but it would make future proposer-specific metadata adjustments easier.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2354,6 +2356,10 @@ def execute_model(
             hidden_states = self._model_forward(
                 num_tokens_padded, input_ids, positions, intermediate_tensors, inputs_embeds, **model_kwargs
             )
+        # Gemma4 MTP reads the target's KV cache; record an NPU event so the
+        # draft can wait on the target's async KV writes before reading.
+        if getattr(self, "drafter", None) is not None and isinstance(self.drafter, AscendGemma4Proposer):
```
Alex-stack-hub, line 2073: nit: `isinstance(None, AscendGemma4Proposer)` is already false, so this could be simplified to `if isinstance(getattr(self, "drafter", None), AscendGemma4Proposer):`. It is a small readability improvement and avoids accessing `self.drafter` through two different forms in the same condition.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3925,11 +3942,28 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
-            self.drafter.initialize_attn_backend(kv_cache_config, block_size)
+            if isinstance(self.drafter, AscendGemma4Proposer):
+                # Gemma4 MTP needs per-group kernel_block_sizes (list of ints),
+                # not a single block_size.  vllm-ascend stores
+                # self.kernel_block_sizes as list[list[int]] (one inner list
+                # per kv_cache_group, e.g. [[128], [128]]), but upstream
+                # vLLM's initialize_attn_backend expects a flat list[int]
+                # indexed by kv_cache_group_id.  Flatten the inner lists.
+                kernel_block_sizes = [
+                    sz if isinstance(sz, int) else sz[0]
```
Alex-stack-hub: suggestion: Could we validate the nested entry before indexing `sz[0]` (or centralize this flattening in a helper)? An empty group or an unexpected multi-size group would otherwise fail here with a fairly opaque `IndexError`, while a short assertion mentioning `kv_cache_group_id` would make configuration issues much easier to diagnose.

Alex-stack-hub (commented):
A few additional non-blocking suggestions for readability and maintainability. None of these are intended as merge blockers; please take whichever ones are useful.

vllm_ascend/spec_decode/gemma4_proposer.py
```diff
@@ -0,0 +1,285 @@
...
+
+        # Build a lookup from target layer name → (num_heads, num_kv_heads)
+        # using the already-computed target_attn_layer_names.
+        target_attn_layers = get_layers_from_vllm_config(
+            self.vllm_config,
+            AttentionLayerBase,
+        )
+
+        draft_model = self.get_model()
+        if not (hasattr(draft_model, "model") and hasattr(draft_model.model, "layers")):
```
drslark: There are so many fallback codes like `getattr` and `hasattr`, which are so ai-style and unreadable for humankind.

vllm_ascend/spec_decode/gemma4_proposer.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def notify_target_forward_done(self) -> None:
+        """Record an NPU event now that the target forward has finished.
+
+        The draft reads the target's KV cache, whose FDO-graph replay writes
+        land asynchronously on the NPU stream; ``_sync_wait_target_events``
+        waits on this event before the draft reads the cache.  A single event
+        is reused across steps to avoid NPU event resource pressure.
+        """
+        if not hasattr(self, "_target_done_event"):
```
drslark: There are many sync events which vllm does not have.

drslark (changes requested):

drslark:
> { }
> 
> * vLLM version: v0.25.1
> * vLLM main: [vllm-project/vllm@54503ec](https://github.com/vllm-project/vllm/commit/54503ecec0f3ac31e5ecfc5f28652e4cc42307b5)

What kind of half-hearted information is this?
Can this code actually run?
Where are the test results?
Are you kidding me?

PR #11411, opened 2026-07-03
[Test][Misc] Merge two test cases for batch invariance into one

wangx700:
This PR merges the two separate test files for batch invariance (`test_aclgraph_batch_invariant.py` and `test_batch_invariant.py`) into a single unified test file `test_batch_invariant.py`. It also updates the compilation configuration to use \"PIECEWISE\" mode instead of \"FULL_DECODE_ONLY\" or `enforce_eager=True` and reduces the default needle batch size to 64.

No.

- vLLM version: v0.23.0


yiz-liu (commented):
Please update the doc to stress that `FULL/FULL_DECODE_ONLY` is not compatible with batch invariancy.

yiz-liu (approved):

PR #11417, opened 2026-07-03
[BugFix] Fix packaged custom opapi lookup

ShuhaoZhangTony:
## Summary

Fix custom op API lookup for packaged vLLM-Ascend custom ops.

This updates the extension RPATH to point at the packaged `vendors/vllm-ascend/op_api/lib` directory and teaches the ACLNN adapter to derive that packaged custom op library path from the loaded extension location at runtime.

## Root Cause

`aclnnAddRmsNormBias` is not exported by the CANN system `libopapi.so`; it is exported by the packaged `libcust_opapi.so` under `vllm_ascend/_cann_ops_custom/vendors/vllm-ascend/op_api/lib`.

The previous lookup path could miss that packaged library, especially when `ASCEND_CUSTOM_OPP_PATH` was not set before the Python extension was imported. In that case the resolver fell through to system `libopapi.so` and failed at runtime with missing `aclnnAddRmsNormBias` / `aclnnAddRmsNormBiasGetWorkspaceSize`.

## Changes

- Add CANN 9 AscendC CMake path candidates under `aarch64-linux/tikcpp` and `arm64-linux/tikcpp`.
- Update `vllm_ascend_C` RPATH from the stale `custom_transformer` vendor path to `vendors/vllm-ascend/op_api/lib`.
- Recompute custom op paths at lookup time instead of relying only on early global initialization.
- Add a packaged custom op library lookup path derived via `dladdr` from the loaded extension path.

## Validation

- Cherry-picked onto current `vllm-project/vllm-ascend/main`.
- `git diff --check upstream/main...HEAD` passed.
- Rebuilt and installed `vllm_ascend_C` in a prepared dev-hub Ascend development container on the downstream branch containing the same patch.
- Verified the rebuilt extension RPATH contains `_cann_ops_custom/vendors/vllm-ascend/op_api/lib`.
- Verified packaged `libcust_opapi.so` exports `aclnnAddRmsNormBias` and `aclnnAddRmsNormBiasGetWorkspaceSize`.
- Verified Python registration exposes `torch.ops._C_ascend.npu_add_rms_norm_bias`.

- vLLM version: v0.23.0


ShuhaoZhangTony:
Branch renamed to remove the codex prefix. Follow-up PR: #12316. The replacement also addresses both outstanding review comments.

PR #11421, opened 2026-07-03
Fix NPU runtime compatibility for Python 3.12

ShuhaoZhangTony:
## Summary

- Map `torch.accelerator.get_memory_info` and `memory_allocated` to the NPU implementations so generic vLLM memory profiling works on Ascend.
- Avoid eager `vllm_ascend.ops` imports while `vllm_ascend.device.device_op` is still initializing.
- Make the spec-decode rejection sampler helper imports compatible with both old and renamed vLLM helper symbols.

## Why

With the Python 3.12 / torch-npu runtime, `torch.accelerator.get_memory_info` may exist but still route through the generic allocator path instead of `torch.npu.mem_get_info`. vLLM's `MemorySnapshot` can then fail during worker device initialization with an allocator assertion.

The eager `vllm_ascend.ops` imports can also create an import-order cycle through `device_op`, and recent vLLM changes renamed the spec-decode helper kernels used by the Ascend wrapper.

## Validation

- Syntax-compiled the touched Python files.
- Verified the downstream HUST stack with Python 3.12 on Ascend NPU 6,7, TP=2, Qwen2.5-7B; `/health` returned 200.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/ops/__init__.py
+++ b/vllm_ascend/ops/__init__.py
@@ -15,12 +15,18 @@
 # This file is a part of the vllm-ascend project.
 #
 
+import sys
+
 import torch
 from vllm.triton_utils import HAS_TRITON
 
-import vllm_ascend.ops.fused_moe.fused_moe  # noqa
-import vllm_ascend.ops.layernorm  # noqa
-import vllm_ascend.ops.register_custom_ops  # noqa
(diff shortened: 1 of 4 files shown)


ShuhaoZhangTony:
Closing in favor of a replacement PR with an upstream-compliant title, non-codex branch name, and DCO sign-off.

PR #11422, opened 2026-07-03
[Ops][BugFix] Fix NPU memory profiling on Python 3.12

ShuhaoZhangTony:
## Summary

Map torch.accelerator memory queries to torch.npu implementations in the centralized accelerator patch module. This keeps generic vLLM memory profiling on the NPU allocator path under Python 3.12 / torch-npu.

The earlier circular-import workaround, duplicate platform patch, and compatibility import shim were removed while rebasing because current main either already provides the required symbol names or has a centralized patch path.

## Validation

- git diff --check
- Python syntax compilation for the changed patch module
- downstream HUST stack previously verified on Ascend with Python 3.12, TP=2, Qwen2.5-7B

### DIFF
--- a/vllm_ascend/patch/platform/patch_torch_accelerator.py
+++ b/vllm_ascend/patch/platform/patch_torch_accelerator.py
@@ -1,6 +1,29 @@
 import torch
 
 
+def _get_npu_memory_info(
+    device: int | str | torch.device | None = None,
+) -> tuple[int, int]:
+    if device is None:
+        device = torch.npu.current_device()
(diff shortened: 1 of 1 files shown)


ShuhaoZhangTony:
Continued as #12344 on the required `feature/...` branch. The replacement is rebased onto current `main`, keeps the patch in the centralized accelerator module, removes the import-cycle workaround and duplicate patch sites flagged in review, and adds focused device-handling tests. This PR is closed to keep one active review target.

PR #11433, opened 2026-07-04
[BugFix] Exclude MTP draft from FlashComm2 o_proj and align all-to-all buffer shape

wangyichao1999:
 This PR fixes two issues in the FlashComm2 (FC2) o_proj communication path (Flashcomm2OProjRowParallelOp in linear_op.py) that cause failures when running Step3.5 with MTP speculative decoding and FC2 enabled.

  1. Exclude MTP draft layers from the FC2 o_proj path _get_row_parallel_op assigns Flashcomm2OProjRowParallelOp to any layer whose prefix contains o_proj/out_proj and does not contain vision_model. The Step3.5 MTP draft decoder is nested under .mtp_block. (e.g.
  model.layers.45.mtp_block.self_attn.o_proj), so its prefix matches this condition and the draft's o_proj is erroneously assigned the FC2 op.

  This op is bound at construction time to the main model's communication state (reorgnized_batch_ids, group_indices, odp_group). When the draft forward runs through otp_maybe_quant_comm, its batch/token shapes
  don't match these main-model-bound indices, causing failures (assert batch_size % chunk_num == 0 or all-to-all shape/semantic mismatch).

  Fix: Add and "mtp_block" not in prefix to the FC2 o_proj condition so draft layers are excluded at the assignment stage.

  2. Align send_buf / recv_buf dimensions in the all-to-all

  In otp_maybe_quant_comm, recv_buf is always 1D (torch.empty(total_intermediate_size * chunk_size, ...)), but send_buf = chunked.flatten(1, 2) is multi-dimensional (2D when otp_size == 1, 3D when otp_size > 1 due
  to the group_indices advanced indexing). dist.all_to_all_single expects matching shapes; this dimension inconsistency is a latent bug that surfaces when the abnormal draft path exercises it.

  Fix: send_buf = chunked.flatten(1, 2).contiguous().view(-1) flattens to 1D to match recv_buf. .contiguous() is required before .view(-1) because the advanced-indexed tensor may not be reinterpretable as a flat view.

No.

Step3.5 Flash model, 8× Ascend NPU, TP=8, FC1+FC2+MTP all enabled. After the fix, both PD disaggregated and PD colocated deployments start correctly and pass random dataset benchmark stress testing (batch 1–128).

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/ops/test_linear.py
+++ b/tests/ut/ops/test_linear.py
@@ -208,7 +208,7 @@ class TestColumnParallelOpDispatch(unittest.TestCase):
 
 
 class TestRowParallelOpDispatch(unittest.TestCase):
-    """Tests for _get_row_parallel_op factory — share_expert."""
+    """Tests for _get_row_parallel_op — mtp_block, share_expert."""
 
     def setUp(self):
         self.mock_layer = MagicMock()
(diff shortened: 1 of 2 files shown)


tests/ut/ops/test_linear.py
```diff
@@ -168,5 +168,149 @@ def test_init_without_disable_tp(self):
         self.assertTrue(isinstance(linear.quant_method, AscendUnquantizedLinearMethod))
 
 
+class TestColumnParallelOpDispatch(unittest.TestCase):
+    """Tests for _get_column_parallel_op: DSV4 removal, share_expert, g_proj."""
+
+    def setUp(self):
+        self.mock_layer = MagicMock()
+        self.mock_layer.tp_size = 2
+        self.mock_layer.tp_rank = 0
```
zhenwenqi2024: What functionality does this UT cover? Don't add this kind of code.

vllm_ascend/ops/linear_op.py
```diff
@@ -722,7 +722,7 @@ def _get_row_parallel_op(
     if matmul_allreduce_enable():
         return MatmulAllreduceRowParallelOp(layer)
     if flashcomm2_enable():
-        if "o_proj" in prefix or "out_proj" in prefix:
+        if ("o_proj" in prefix or "out_proj" in prefix) and "mtp_block" not in prefix:
```
zhenwenqi2024, line 727: Is mtp_block universal across all models?

vllm_ascend/ops/linear_op.py
```diff
@@ -357,7 +357,7 @@ def otp_maybe_quant_comm(x):
             chunked = x.view(chunk_num, batch_size_per_chunk, x.shape[1])
             if self.otp_size != 1:
                 chunked = chunked[self.group_indices]
-            send_buf = chunked.flatten(1, 2)
+            send_buf = chunked.flatten(1, 2).contiguous().view(-1)
```
zhenwenqi2024, line 360: Is continuous required?

realliujiaxu (approved):
LGTM

vllm_ascend/ops/linear_op.py
wangyichao1999: .contiguous() guards the otp_size > 1 path.

tests/ut/ops/test_linear.py
wangyichao1999: Deleted.

vllm_ascend/ops/linear_op.py
wangyichao1999: Not 100% universal, but mtp_block is the dominant convention for MTP draft decoders across vLLM — 14+ MTP implementations (Step3p5, DeepSeek, LongCatFlash, Ernie, Glm4Moe, MiMo,HYV3, GlmOcr, OpenPangu, DeepSeekV4, …) all name the draft submodule self.mtp_block = SomeDecoderLayer(...), which puts .mtp_block. in the parameter prefix.

PR #11438, opened 2026-07-04
[BugFix] Fix zero chunk problem of Qwen3.5 + PCP

SunnyLee151064:
  
  Fixes a runtime 507015 (MTE write out-of-bounds) crash in the AscendC chunk_gated_delta_rule_fwd_h / chunk_fwd_o kernels when running Qwen3.5 with PCP
  (prefill-context-parallel) world_size > 1 under a mixed-length concurrent batch.

 #### Root cause 
  PCP splits each request's prefill across ranks with an interleave granularity (cp_kv_cache_interleave_size, default 128). A request shorter than the interleave size
  produces zero tokens on rank>0, leaving a zero-length segment in that rank's cu_seqlens.

  chunk_indices carries compact (non-empty) segment ranks (empty segments are skipped when built), but the chunk_fwd_o kernel indexes gmSeqlen (= the original cu_seqlens,
  with empty segments) by that same compact rank. When a zero-length segment is sandwiched between non-empty ones, the rank and the original segment index go out of sync.
  
####  Example

  Suppose rank>0 has 3 requests with lengths [128, 0, 128] tokens (the middle request is shorter than the interleave size, so it gets 0 tokens on this rank):

  cu_seqlens      = [0, 128, 128, 256]   # seg0=128, seg1=EMPTY, seg2=128
  chunk_indices   = [(0,0),(0,1),(1,0),(1,1)]   # compact rank: seg0→0, seg2→1 (seg1 skipped)

  chunk_fwd_o processing chunk3 (the last chunk of seg2, within-chunk idx=1):
  
  tokenOffset   = gmSeqlen[1]              = 128        # cu_seqlens[1]  (boundary of EMPTY seg1!)
  batchTokens   = gmSeqlen[2] - tokenOffset = 128 - 128 = 0
  isFinalState  = (chunkIdx == numChunks-1) = true
  blockTokens   = batchTokens - batchChunkIdx*chunkSize
                = 0 - 1*64                   = -64       # uint32 underflow → huge value
                → MTE write out-of-bounds → runtime 507015
  #### The fix
  
  Add _compact_empty_segments(cu_seqlens_host, initial_state) and call it before the AscendC fwd_h / fwd_o kernels. It drops zero-length segments from cu_seqlens (and the
  corresponding rows of initial_state), so the compact rank carried by chunk_indices lines up with the compacted gmSeqlen. The compacted final_state is scattered back to
  the original [N, H, K, V] layout via the keep-mask (empty segments keep their initial_state). It is a no-op when there are no empty segments (rank0 / non-PCP / uniform
  batch).

  chunk_indices is already compact-ranked (built by skipping empty segments in gdn_attn_builder._fill_chunk_indices_cpu), so it is reused as-is; only cu_seqlens /
  initial_state need compacting. Zero-length segments contribute exactly zero to the GDN math, so compaction preserves the token layout (empty segment = 0 tokens) and is
  numerically equivalent.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/ops/triton/fla/chunk.py
+++ b/vllm_ascend/ops/triton/fla/chunk.py
@@ -111,12 +111,16 @@ def chunk_gated_delta_rule_fwd(
         cu_seqlens_kern = cu_seqlens_host if prebuilt_meta.cu_seqlens_kern is None else prebuilt_meta.cu_seqlens_kern
         keep_meta = prebuilt_meta.keep_meta
         initial_state_kern = (
-            initial_state[keep_meta.to(initial_state.device)]
+            initial_state[keep_meta]
             if initial_state is not None and keep_meta is not None
             else initial_state
         )
     else:
-        cu_seqlens_kern, initial_state_kern, keep_meta = _compact_empty_segments(cu_seqlens_host, initial_state)
+        cu_seqlens_kern, initial_state_kern, keep_meta = _compact_empty_segments(
+            cu_seqlens_host,
+            initial_state,
+            device=initial_state.device if initial_state is not None else None,
+        )
     h, v_new, final_state = torch.ops._C_ascend.chunk_gated_delta_rule_fwd_h(
         k_ascendc,
         w_ascendc,
--- a/vllm_ascend/ops/triton/fla/chunk.py
+++ b/vllm_ascend/ops/triton/fla/chunk.py
@@ -133,7 +133,19 @@ def chunk_gated_delta_rule_fwd(
     # Compact zero-length segments for the AscendC kernels (see
     # _compact_empty_segments). chunk_indices_chunk64 is already compact-ranked and
     # is reused as-is; only cu_seqlens / initial_state need compacting.
-    cu_seqlens_kern, initial_state_kern, keep_meta = _compact_empty_segments(cu_seqlens_host, initial_state)
+    # Pre-computed compact info from metadata avoids recomputation across layers:
+    # cu_seqlens_kern and keep_meta only depend on cu_seqlens_host (batch-wide),
+    # so they are built once and cached in GDNChunkedPrefillMetadata.
+    if prebuilt_meta is not None and hasattr(prebuilt_meta, 'keep_meta'):
+        cu_seqlens_kern = cu_seqlens_host if prebuilt_meta.cu_seqlens_kern is None else prebuilt_meta.cu_seqlens_kern
+        keep_meta = prebuilt_meta.keep_meta
+        initial_state_kern = (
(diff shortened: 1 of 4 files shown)


SunnyLee151064:
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/28699766404)

vllm_ascend/ops/triton/fla/chunk.py
```diff
@@ -99,21 +130,31 @@ def chunk_gated_delta_rule_fwd(
         cu_seqlens_host = tuple(cu_seqlens.tolist())
     if chunk_indices_chunk64_host is None and chunk_indices is not None:
         chunk_indices_chunk64_host = tuple(chunk_indices.flatten().tolist())
+    # Compact zero-length segments for the AscendC kernels (see
+    # _compact_empty_segments). chunk_indices_chunk64 is already compact-ranked and
+    # is reused as-is; only cu_seqlens / initial_state need compacting.
+    cu_seqlens_kern, initial_state_kern, keep_meta = _compact_empty_segments(cu_seqlens_host, initial_state)
```
weiguihua2: This block is placed in the build metadata for calculation and is directly obtained as metadata.

vllm_ascend/ops/triton/fla/chunk.py
```diff
@@ -99,21 +130,31 @@ def chunk_gated_delta_rule_fwd(
...
+        cu_seqlens=cu_seqlens_kern,
         chunk_indices=chunk_indices_chunk64_host,
         use_exp2=False,
         transpose_state_layout=False,
     )
+    if keep_meta is not None:
+        # Scatter the compacted final_state back to the original [N, H, K, V] layout
+        # the PCP state recursion expects; empty segments keep their initial state.
+        _fs_full = initial_state.clone()
+        _fs_full[keep_meta] = final_state
```
weiguihua2, line 141: The number of true values in keep_meta is variable, and counting the number of true values may introduce synchronization operations. It is recommended to perform profiling for a look.

linfeng-yuan:
@ZT-AIA Please review this PR.

linfeng-yuan (dismissed):

vllm_ascend/ops/triton/fla/chunk.py
SunnyLee151064: solved/

vllm_ascend/ops/triton/fla/chunk.py
SunnyLee151064: solved

PR #11444, opened 2026-07-04
[Cherry-pick][main][KV Cache][Feature] Support Layerwise KV Pooling with Memcache Backendche

tyy0829:
### 1. Per-Layer Key Management
- Each KV cache block in every layer was assigned a unique key.
- The number of keys increased proportionally with the number of layers and blocks, resulting in significant metadata management overhead.

### 2. High Metadata Lookup Latency
- Both `put` and `get` operations required querying the Meta Server to resolve the memory address associated with a key.
- Frequent metadata lookups introduced additional latency on the critical access path.

### 3. Inefficient HBM Address Computation
- HBM addresses were computed using Python lists.
- This approach incurred noticeable CPU overhead when processing large numbers of blocks.

---

### 1. Unified Key Management
- Keys are now managed at the request ID level.
- KV cache blocks belonging to different layers but the same request share a single key.
- This significantly reduces the number of keys and metadata management overhead.

### 2. One-Time Address Resolution
- Key-to-address mapping is performed only once during initialization.
- Subsequent accesses locate layer-specific blocks using a base address plus a layer offset.
- This eliminates repeated Meta Server lookups during runtime.

### 3. New Address-Based Memcache Interface
- Introduced a new `batch_copy` API in Memcache.
- Memory can be accessed directly through physical addresses without key resolution.
- This reduces access latency and improves batch transfer efficiency.

### 4. Accelerated HBM Address Computation
- Replaced Python list-based address computation with NumPy vectorized operations.
- This reduces CPU overhead and improves address generation efficiency.

### 5. Controlled Overlap Between Data Transfer and Attention Computation
- Limited the degree of concurrency between KV cache transfers and Attention computation.
- This reduces resource contention and minimizes interference with the inference execution path.

---

### Per-Rank Key Design
- Store key format: `{model}@{block_hash}@{head_or_tp_rank}` where `head_or_tp_rank = tp_rank // put_step`
- Ranks in the same `put_step` group share one key (same KV cache for MLA latent)
- MLA (`put_step=8`): all 8 ranks share key `@0`, only 1 key per block
- GQA (`put_step=1`): each rank has its own key `@{tp_rank}`

### GVA Alloc Moved from Scheduler to Worker
- Memcache `gvaBlobTracker` is per-process, `batch_alloc` and `batch_copy` must run in the same process
- Each worker allocates its own GVA via `batch_alloc` right before `batch_copy`
- `batch_alloc` is non-idempotent, so already-allocated keys are reused from `_allocated_gvas` cache

### Dynamic Save by put_step
- Only `tp_rank % put_step == 0` saves to the pool (avoids redundant writes for MLA)
- MLA: only rank 0 saves (1 copy, no redundancy)
- GQA: every rank saves independently (8 different copies)
- All ranks load from the pool via `batch_get_key_info` + `batch_add_lease` + `batch_copy G2L`

### Full k+v Write (No Partial Write)
- Removed `put_step` slicing on save/load — each saving rank writes the full k+v for each block
- 27 layers accumulate to fill the entire blob (`alloc_size = page_size × num_layers`)
- Fixes memcache `ConsumePendingHole` error caused by partial blob writes

### Hit Block Skip in Save
- Hit blocks are loaded via `batch_add_lease` which changes blob state to READABLE
- Re-saving READABLE blobs fails with `MMC_UNMATCHED_KEY`
- Save range excludes hit blocks: `save_start_block = max(save_start_block, hit_full_blocks)`

### Separate Save/Load GVA Fields
- `block_gvas_np` + `gva_block_offset` for save path
- `load_block_gvas_np` + `load_gva_block_offset` for load path
- Prevents overwrite between `_alloc_gvas_for_save` and `_prepare_load_gvas`
- `build_shared` / `_require_request_arrays` accept `is_save` parameter

### hit_check Uses `batch_get_key_info`
- `batch_is_exist` only checks if a key was alloc'd, not if save is complete → false hits
- `batch_get_key_info` returns valid GVA only when save is complete → no false hits
- Prevents cascading failures (false hit → load fail → save thread list not cleared → crash)

### Lease Lifecycle
- `batch_add_lease` before load (registers blob in per-process `gvaBlobTracker` + acquires lease)
- `batch_copy G2L` reads data (27 layers)
- `batch_remove_lease` after final layer's read completes (releases lease immediately)
- No more waiting for 5min TTL expiration

### Error Handling
- All memcache API failures (`batch_alloc`, `batch_get_key_info`, `batch_add_lease`, `batch_copy`) use `logger.error` instead of `raise`
- No crash on transient memcache failures — load failure uses stale KV cache, save failure skips pool storage

---

### DeepSeek V3.2 TP16 hit rate 90%
| kv cache pool | Input tokens | Output tokens | Concurrency | Average TTFT |
|----|----|----|----|----|
| no-layerwise | 64k | 1 | 36.9 | 312519.2 ms |
| layerwise | 64k | 1 | 36.9 | 191323.1 ms |
| no-layerwise | 32k | 1 | 37.1 | 149498.4 ms |
| layerwise | 32k | 1 | 36.8 | 64203.3 ms |
| HBM hit | 32k | 1 | 37.1 | 64276.3 ms |

The optimized design delivers the following benefits:
- Reduced metadata lookup overhead.
- Lower access latency through address-based memory operations.
- Improved HBM address computation efficiency.
- Reduced CPU overhead through vectorized computation.
- Lower runtime contention between KV cache transfers and model execution.
- Improved overall Layerwise KV Cache Pool performance and scalability.

---

Supports memcache backend for layerwise KV cache transfer, enabling per-layer save/load overlap with attention computation for improved TTFT.

New `use_layerwise` config option for `AscendStoreConnector` with memcache backend.

- DeepSeek-V2-Lite-Chat (MLA, TP=8): GSM8K accuracy + TTFT performance
- Multi-block requests (2+ blocks) with save → hit → load → save verification
- Concurrent requests (batch_size=32) with lease release verification
- No `ConsumePendingHole`, `MMC_UNMATCHED_KEY`, or `LEASE_EXPIRED` errors

- vLLM version: v0.24.0


vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/ascend_store_connector.py
```diff
@@ -84,6 +84,9 @@ def __init__(self, vllm_config: VllmConfig, role: KVConnectorRole, kv_cache_conf
         self.kv_role = vllm_config.kv_transfer_config.kv_role
 
         self.use_layerwise = vllm_config.kv_transfer_config.kv_connector_extra_config.get("use_layerwise", False)
+        backend_name = vllm_config.kv_transfer_config.kv_connector_extra_config.get("backend", "mooncake")
+        self.backend_name = backend_name.lower()
+        self.use_gva_layerwise = self.use_layerwise and self.backend_name == "memcache"
```
tyy0829, line 89: The current function of layerwise is to distinguish backends, with different backends following different code branches. `self.use_gva_layerwise` uses the memcache backend.

LCAIZJ:
Could you provide a detailed explanation of how KV cache transfer and Attention computation are overlapped? A profiling trace would be especially helpful.

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_scheduler.py
```diff
@@ -350,158 +597,324 @@ def update_state_after_alloc(self, request: "Request", blocks: "KVCacheBlocks",
...
+            block_sizes=self.grouped_block_size,
+        )
+        self._request_trackers[request.req_id] = request_tracker
+        num_blocks = num_tokens_to_compute // self._block_size
+        has_last_block = num_tokens_to_compute % self._block_size != 0
+        self._allocate_gva_if_needed(
+            request_tracker,
+            request_real.block_hashes,
+            num_blocks,
+            has_last_block,
```
LCAIZJ, line 705: What is the reason for saving last_block_key?

tyy0829: The last_block_key is only recorded, but the KV cache of last_block_key is not stored.

Pz1116 (approved):
please make CI happy

Pz1116 (approved):
Make sure to cherry-pick all layerwise related PRs from 23.0 to main branch if needed

wangxiyuan (approved):

PR #11445, opened 2026-07-04
[BugFix] fix  for new verision of cann

ZT-AIA:
The changes to log.h in the new version of cann caused compilation errors, so compatibility with the new version of cann was added.
No
pip install 

- vLLM version: v0.23.0


zzzzwwjj (approved):

zzzzwwjj:
This PR looks like an AI's work. This is a temporary bug fix, please confirm the correct way to call the CANN interface.

PR #11449, opened 2026-07-05
[Performance] Reuse DP metadata sync buffers

ShuhaoZhangTony:
This extracts the DP metadata sync buffer reuse optimization from vllm-ascend-hust for upstream review.

The current DP metadata sync path allocates fresh CPU tensors on every `_sync_metadata_across_dp` call. This PR keeps reusable per-runner buffers for:

- the packed DP metadata tensor
- the `num_tokens_after_padding` result tensor

It also preserves the existing `dp_allreduce_on_npu` correctness path by using a reusable NPU metadata tensor when that mode is enabled, then copying the reduced result back to the CPU buffer for the existing post-processing path.

Validation:

- `git diff --check`
- `python -m py_compile vllm_ascend/worker/model_runner_v1.py tests/ut/worker/a2/test_model_runner_v1.py`

Local pytest could not run in this checkout because the environment is missing the `vllm` Python package:

```text
ModuleNotFoundError: No module named 'vllm'
```

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/worker/a2/test_model_runner_v1.py
+++ b/tests/ut/worker/a2/test_model_runner_v1.py
@@ -4,6 +4,7 @@ from unittest.mock import MagicMock, patch
 
 import numpy as np
 import torch
+from vllm.config import CUDAGraphMode
 from vllm.v1.kv_cache_interface import FullAttentionSpec, KVCacheConfig, KVCacheGroupSpec, KVCacheTensor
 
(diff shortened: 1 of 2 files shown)


ShuhaoZhangTony:
Branch renamed to remove the codex prefix. Follow-up PR: #12317. The replacement also addresses the outstanding buffer-allocation review comment and rebases onto current main.

PR #9731, opened 2026-05-30
[Feature]Add Mooncake SSD offload with embedded client

huangdong2022:
# What this PR does / why we need it?
This PR adds Mooncake SSD offload `Mode A` (Embedded Real Client) support for the KV Pool MooncakeBackend used by AscendStoreConnector, aligned with Mooncake ssd-offload.md `Step 3A`.
see : [https://github.com/kvcache-ai/Mooncake/blob/main/docs/source/deployment/ssd-offload.md#step-3a-mode-a-start-the-application-with-embedded-real-client](url)

# Does this PR introduce _any_ user-facing change?
Yes.
1）enable SSD depends on `Mooncake ≥ v0.3.11`.
2）enable on `mooncake master` side : 
```shell
mooncake_master --eviction_high_watermark_ratio=0.7 --eviction_ratio=0.2 --rpc_port=50088 --enable_offload=true --offload_on_evict=true
```
3）enable on client side：add in `mooncake.json`
```json
{
    "enable_ssd_offload": true,
    "ssd_offload_path": "/XXX/mooncake_offload",
}
```

# How was this patch tested?
## Function Test
1）enable SSD as above mentioned.
2）Run some test cases to check whether the SSD has storage when DRAM eviction occurs. see log on `mooncake_master` node.
```log
I20260601 21:23:38.529359   511 master_service.cpp:3181] [EVICT-DONE] BatchEvict execution completed.
I20260601 21:23:38.539436   511 master_service.cpp:3170] [EVICT-TRIGGER] memory_ratio=0.870361 high_watermark=0.7 need_mem_eviction=0 eviction_ratio=0.2
I20260601 21:23:38.541512   511 master_service.cpp:3181] [EVICT-DONE] BatchEvict execution completed.
[...]

PR #11459, opened 2026-07-06
[BugFix][KV Pool] Fix MooncakeBackend lazy init failure on cross-thread store setup

DHX98:
Upstream PR：https://github.com/vllm-project/vllm-ascend/pull/9731
When `MooncakeBackend` is created with `lazy_init=True` (DSV4 compress model + `ASCEND_ENABLE_USE_FABRIC_MEM=1`), `_setup_store()` is deferred from `__init__` to the first `put()`, which runs on the KV transfer sending thread (`KVCacheStoreLayerSendingThread`), not the main worker thread.

`_setup_store()` calls `get_global_rank()` with no arguments. The no-arg form falls back to `get_current_vllm_config().parallel_config`, which reads the process-global `_current_vllm_config` set by `set_current_vllm_config()` on the main thread. On the KV transfer thread that global is `None`, so `get_current_vllm_config()` raises `AssertionError: Current vLLM config is not set`, the store never initializes, and every subsequent `put()` fails.

This was introduced by the per-DP-rank SSD sub-directory change (`get_global_rank()` call inside `_setup_store()`).

**Fix**: stash `parallel_config` on `self` in `__init__` and pass it explicitly to `get_global_rank(self.parallel_config)` in `_setup_store()`.

No.

Dual-node PD separation with DeepSeek-V4-Flash-w8a8-mtp, `--data-parallel-size 4 --tensor-parallel-size 4`, `kv_role=kv_producer`, `ASCEND_ENABLE_USE_FABRIC_MEM=1`, `enable_multithread_load=true`, Mooncake store with `enable_ssd_offload=true`.

Before fix: `AssertionError: Current vLLM config is not set` x43, store never initialized.
截屏2026-07-06 10 00 57

After fix: `AssertionError` x0.
image

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/backend/mooncake_backend.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/backend/mooncake_backend.py
@@ -61,6 +61,7 @@ def _ssd_setup_kwargs(config: "MooncakeStoreConfig") -> dict[str, object]:
 
 class MooncakeBackend(Backend):
     def __init__(self, parallel_config: ParallelConfig, lazy_init: bool = False):
+        self.parallel_config = parallel_config
         self.config = MooncakeStoreConfig.load_from_env()
         if self.config.protocol != "ascend":
             raise NotImplementedError(f"MooncakeBackend does not support protocol {self.config.protocol!r}.")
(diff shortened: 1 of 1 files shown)


Pz1116 (approved):
LGTM

linfeng-yuan (approved):

PR #11489, opened 2026-07-06
[Test] Added test cases for the precision and performance of qwen3

taozi01319:
Add qwen3-8/14/32 performance and accuracy tests

no

- vLLM version: v0.23.0


.github/workflows/configs/weekly_config.yaml
```diff
@@ -1,4 +1,4 @@
-# Single source of truth for the Weekly-A2 / Weekly-A3 test matrix.
+ # Single source of truth for the Weekly-A2 / Weekly-A3 test matrix.
```
jiangyunfan1: not necessary to change this.

wenjun91 (approved):

.github/workflows/configs/weekly_config.yaml
taozi01319: Modified

wangxiyuan (approved):

PR #11496, opened 2026-07-06
[Performance][SpecDecode] Avoid H2D sync in CP proposer metadata

pisceskkk:

This PR reduces host/device synchronization in the CP speculative proposer metadata path.

The first commit keeps CP proposer metadata calculations on device by deriving `cu_num_tokens_pcp_full`, decode query lengths, reject/accept counts, and slot index bases from `query_start_loc_pcp_full.gpu`. It also precomputes CP slot offsets and builds slot indices with a vectorized device operation instead of per-request Python `torch.arange` calls. `_get_cp_local_seq_lens` now creates rank offsets on the same device and dtype as `seq_lens`, avoiding a CPU tensor in the device path.

The follow-up pass removes additional CP-only sync candidates in `llm_base_proposer.py`: decode hidden-state unpadding indices now stay on device instead of using a host Python index list, CP prefill query-start offsets use the existing CPU mirror instead of `cad.query_start_loc[...] .item()`, and non-MLA CP sequence metadata uses the CPU seq-lens mirror when it must be converted to numpy.

No. This is an internal performance-path optimization for CP speculative decoding metadata construction.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/spec_decode/llm_base_proposer.py
+++ b/vllm_ascend/spec_decode/llm_base_proposer.py
@@ -980,11 +980,11 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
                 if ori_seq_len_cpu is not None:
                     ori_seq_len_cpu = ori_seq_len_cpu[:batch_size].clone()
                 slot_idx_base = (
-                    query_start_loc_pcp_full[:num_decode_reqs].to(torch.int64) * self.pcp_size
+                    query_start_loc_pcp_full[:num_decode_reqs] * self.pcp_size
                     + torch.arange(num_decode_reqs, device=self.device, dtype=torch.int64)
                     * (self.num_speculative_tokens - 1)
                     * self.pcp_size
-                    + (num_accept_tokens.to(torch.int64) - 1) * self.pcp_size
+                    + (num_accept_tokens - 1) * self.pcp_size
                 )
                 slot_indices = (
                     slot_idx_base[:, None]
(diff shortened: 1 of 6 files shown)


vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -1440,19 +1447,23 @@ def set_inputs_first_pass(
...
                     else:
-                        # remove padding (from pcp all-gather) in decode part
-                        mask_start_loc = torch.cat(
-                            [
-                                torch.tensor([0], dtype=torch.int32, device=query_lens_d.device),
-                                torch.cumsum(query_lens_d * self.pcp_size, dim=0)[:-1],
-                            ]
+                        # Keep decode unpadding indices on device to avoid a host index list.
+                        decode_req_starts = query_start_loc_pcp_full[:num_decode_reqs].to(torch.int64)
+                        decode_query_lens = query_lens_d_device.to(torch.int64)
```
zhenwenqi2024: int32 够吗，decode阶段少点这种to的操作？

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -1440,19 +1447,23 @@ def set_inputs_first_pass(
...
-                        for req_id in range(num_decode_reqs):
-                            assert None not in (mask_start_loc, mask_len)
-                            mask += list(range(mask_start_loc[req_id], mask_start_loc[req_id] + mask_len[req_id]))
-                        target_hidden_states_d = target_hidden_states_d_padded[mask]
+                        decode_padded_starts_per_token = torch.repeat_interleave(
+                            decode_padded_starts,
+                            decode_query_lens,
+                            output_size=num_tokens_d,
+                        )
+                        decode_offsets = self.arange[:num_tokens_d].to(torch.int64) - decode_req_starts_per_token
```
zhenwenqi2024: int64

weiguihua2 (approved):

slippersss (approved):

linfeng-yuan (approved):

PR #11497, opened 2026-07-06
[BugFix][RFork] Fix RFork GLM fallback crashes

### LINKED ISSUE #11409: [Bug]: Rfork第二次实例启动：FusedMoE shared experts split computation does not match the integrated computation.max absolute difference:nan integrated output-sum:nan,norm:nan split output-sum:nan,norm:nan
### Your current environment

<details>

```text
Your output of above commands here
```

</details>

### 🐛 Describe the bug

[](url)`

`

yangsonglin13:

Fixes #11409.

This PR fixes two RFork failure paths seen when processed-layout transfer is used with quantized DeepSeekV2-family models, for example GLM-5 W4A8 + MTP with `multistream_overlap_shared_expert`.

1. During the RFork pre-transfer post-process, `process_weights_after_loading` can run the `AscendFusedMoE` shared-expert consistency forward while receiver weights are still empty. That can produce NaNs and abort the transfer with `FusedMoE shared experts split computation does not match the integrated computation`. This PR keeps the workaround scoped to the RFork loader: only for the pre-transfer layout-processing call, it temporarily uses each wrapped FusedMoE quant method's original `process_weights_after_loading` implementation saved by `functools.wraps`, then restores the wrapper immediately after the call.

2. If RFork transfer fails after a worker has initialized a temporary RFork model, fallback re-runs the default loader in the same process. The failed model's process-global registrations survive `del model` and can make re-initialization fail with `Duplicate layer name`. This PR removes only the registry entries owned by that discarded RFork model before fallback re-initialization, while leaving unrelated or newly rebuilt fallback registrations intact for later profile/compile. If fallback happens before RFork model initialization, such as a seed miss, the registry cleanup is skipped.

3. CI pre-commit was failing before reaching the changed Python files because the self-hosted runner's `markdownlint-cli` hook setup used nodeenv's moving default Node version and hit missing nodeenv/node_modules files. This PR pins the markdownlint hook to Node 20.19.0 so the hook environment is deterministic.

Together these changes avoid the common pre-transfer shared-expert validation failure and keep the fallback path from taking the worker down on future RFork failures, without changing the non-RFork FusedMoE wrapper path.

No public API or configuration change.

Users running RFork with affected quantized DeepSeekV2-family models no longer hit the worker startup crash in this path. Users not using RFork, processed-layout transfer, or the affected shared-expert validation path are unchanged.

Unit tests were added/updated in `tests/ut/model_loader/rfork/test_rfork_loader.py` for:

- RFork fallback removing only failed-model process-global registrations before the fallback `get_model` call, while preserving unrelated state and fallback registrations needed by later profile/compile.
- Seed-miss fallback preserving existing process-global state because no RFork model was initialized yet.
- The registry reset helper remaining safe when optional attributes are absent.
- RFork pre-transfer weight processing temporarily unwrapping FusedMoE quant-method post-load wrappers and restoring them afterward, including exception paths.

Local checks run for this PR:

- `python -m ruff check vllm_ascend/model_loader/rfork/rfork_loader.py tests/ut/model_loader/rfork/test_rfork_loader.py`
- `python -m ruff format --check vllm_ascend/model_loader/rfork/rfork_loader.py tests/ut/model_loader/rfork/test_rfork_loader.py`
- `python -m py_compile vllm_ascend/model_loader/rfork/rfork_loader.py tests/ut/model_loader/rfork/test_rfork_loader.py`
- `python -m pre_commit clean; python -m pre_commit run markdownlint --hook-stage manual --files README.md`
- `git diff --check main/main...HEAD`
- `git diff --check`
- `git merge-tree --write-tree main/main HEAD`

Targeted pytest was attempted but this local Windows environment is missing the upstream `vllm` package, so collection stopped at `ModuleNotFoundError: No module named 'vllm'` before the tests could run locally.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/model_loader/rfork/test_rfork_loader.py
+++ b/tests/ut/model_loader/rfork/test_rfork_loader.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 
+from functools import wraps
 from types import SimpleNamespace
 
 import pytest
@@ -28,6 +29,8 @@ from vllm_ascend.model_loader.rfork.rfork_loader import (
     _is_dynamic_eplb_enabled,
     _is_layer_sharding_enabled,
     _make_fallback_load_config,
+    _reset_process_global_model_state,
+    _rfork_pre_transfer_weight_processing,
 )
 from vllm_ascend.model_loader.rfork.seed_protocol import get_local_seed_key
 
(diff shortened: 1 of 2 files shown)


yangsonglin13:
Superseded by #12995, which consolidates this RFork fallback fix with the related NPU synchronization and runtime-layout fixes into one PR.

PR #11527, opened 2026-07-07
[0-day] Hy3 0-day adaptation

rjg-lyh:
  Fix two Hunyuan3 (Hy3) compatibility issues on vllm-ascend 0.22.1rc:
  
  1. **MTP model type conversion**: `patch_speculative_config` was missing `hy_v3` → `hy_v3_mtp` conversion, causing
  `Unsupported speculative method: 'mtp'` when MTP is enabled.
  2. **MoE quant type lookup**: `get_linear_quant_type` failed with KeyError for MoE layers because it looked up
  `experts.weight` as a single key, while the quant description only has per-expert entries (`experts.0.gate_proj.weight`). 

No.

- vLLM version: v0.22.1

### DIFF
--- a/vllm_ascend/patch/platform/patch_speculative_config.py
+++ b/vllm_ascend/patch/platform/patch_speculative_config.py
@@ -124,6 +124,13 @@ def hf_config_override(hf_config: PretrainedConfig) -> PretrainedConfig:
     if initial_architecture == "MistralLarge3ForCausalLM":
         hf_config.update({"architectures": ["EagleMistralLarge3ForCausalLM"]})
 
+    if hf_config.model_type == "hy_v3":
+        hf_config.model_type = "hy_v3_mtp"
+        n_predict = getattr(hf_config, "num_nextn_predict_layers", None)
+        hf_config.update(
+            {"n_predict": n_predict, "architectures": ["HYV3MTPModel"]}
+        )
+
     return hf_config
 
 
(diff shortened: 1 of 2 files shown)


vllm_ascend/patch/platform/patch_speculative_config.py
```diff
@@ -124,6 +124,13 @@ def hf_config_override(hf_config: PretrainedConfig) -> PretrainedConfig:
     if initial_architecture == "MistralLarge3ForCausalLM":
         hf_config.update({"architectures": ["EagleMistralLarge3ForCausalLM"]})
 
+    if hf_config.model_type == "hy_v3":
```
zhenwenqi2024, line 127: Hy3.0's MTP support is already available in vLLM, correct? If so, what's the reason for applying a patch here?"

vllm_ascend/quantization/modelslim_config.py
```diff
@@ -327,7 +327,22 @@ def get_linear_quant_type(
...
+            # instead of a single experts.weight key. Find any sub-entry.
+            prefix_dot = prefix + "."
+            for k, v in quant_description.items():
+                if k.startswith(prefix_dot) and k.endswith(".weight"):
+                    quant_type = v
+                    break
+            else:
+                raise KeyError(
+                    f"Neither '{key}' nor any key starting with "
+                    f"'{prefix_dot}' found in quant_description."
```
zhenwenqi2024, line 342: maybe the after descrition is better?"quant_description does not contain the key '{key}' nor any key with the prefix '{prefix_dot}'."

ponyioy:
请问这个修改什么时候可以合并？

PR #11555, opened 2026-07-07
[LoRA] Register Ascend dense-linear LoRA wrappers for MLP/o_proj under TP>1

Liuchenbing-2026:
## Summary

On the `v0.21.0rc` line, `refresh_all_lora_classes()` registers only the four QKV Ascend LoRA wrappers. The dense linear layers (`gate_up_proj` / `down_proj` / `o_proj`) are `AscendMergedColumnParallelLinear` / `AscendRowParallelLinear` / `AscendColumnParallelLinear` subclasses, but the upstream `MergedColumn* / Row* / Column*WithLoRA.can_replace_layer` gate on a strict `type(source_layer) is <upstream base>` check (the merged-column class additionally has a `tp_size == 1` fallback for subclasses).

Under **TP > 1**, none of these match the Ascend concrete types, so the MLP + `o_proj` LoRA deltas are silently dropped and the adapter appears to have no effect (LoRA output == base output).

## Fix

Add six Ascend dense-linear LoRA wrappers mirroring the existing QKV ones (plain + sharded variants for MergedColumn / Row / Column) whose `can_replace_layer` matches the Ascend concrete types, and register them in `refresh_all_lora_classes()`. The LoRA math is fully reused from the upstream wrappers — this change only fixes layer replacement/registration, not the compute path.

## Effect

Fixes dense (MLP / `o_proj`) LoRA being ineffective at TP > 1 for Qwen3.5-27B (Qwen3-Next hybrid GDN, dense). In our testing, an r=8 adapter under TP=2 in eager mode now yields a non-zero LoRA delta with this change, whereas before the outputs were bit-identical to the base model.

## Scope

- Single file: `vllm_ascend/lora/utils.py` (+110), registration only.
- Verified in eager (single-operator) mode. Graph modes (piecewise / full-decode) have separate, unrelated issues and are out of scope for this PR.

- vLLM version: v0.21.0

### DIFF
--- a/vllm_ascend/lora/utils.py
+++ b/vllm_ascend/lora/utils.py
@@ -3,15 +3,24 @@ from torch import nn
 from transformers import PretrainedConfig
 from vllm.config import LoRAConfig
 from vllm.lora.layers import (
+    ColumnParallelLinearWithLoRA,
+    ColumnParallelLinearWithShardedLoRA,
+    MergedColumnParallelLinearWithLoRA,
+    MergedColumnParallelLinearWithShardedLoRA,
     MergedQKVParallelLinearWithLoRA,
     MergedQKVParallelLinearWithShardedLoRA,
     QKVParallelLinearWithLoRA,
     QKVParallelLinearWithShardedLoRA,
+    RowParallelLinearWithLoRA,
+    RowParallelLinearWithShardedLoRA,
 )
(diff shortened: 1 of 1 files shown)


Liuchenbing-2026:
Closing — v0.21.0rc is no longer the active target for this dense-linear LoRA wrapper fix.

PR #11575, opened 2026-07-07
[feature] Support Gemma4 ModelSlim quantization

Alex-stack-hub:

This PR adds ModelSlim W8A8_DYNAMIC quantization support for Gemma4 / Gemma4 text models.

For Gemma4 full-attention `k_eq_v` layers, the ModelSlim quant description contains `q_proj` and `k_proj`, but does not contain a dedicated `v_proj` entry. The model loader duplicates `k_proj` into `v_proj` at load time, while the Ascend ModelSlim packed `qkv_proj` lookup previously required all `q_proj` / `k_proj` / `v_proj` quant entries to exist. As a result, loading a Gemma4 quantized checkpoint could fail with `KeyError` on the missing `v_proj.weight` quant description key.

This PR handles only that known missing-`v_proj` case when both `q_proj` and `k_proj` quant entries are present, so other malformed packed quant descriptions still keep the original failure behavior.

This PR also adds Gemma4-scoped ModelSlim mapping for MoE expert layers. vLLM routes the FusedMoE module through a `.moe.experts` prefix, while the ModelSlim quant description keeps the checkpoint-style `.experts` naming. The added prefix mapping is scoped to `gemma4` / `gemma4_text`, avoiding a global `.experts` rewrite that could affect other model types.

Related issue/RFC: N/A.

No. This only extends ModelSlim quantization config handling for Gemma4 quantized checkpoints. Existing non-Gemma4 logic is preserved, and missing packed shards still raise as before except for the Gemma4 `k_eq_v` missing-`v_proj` layout described above.

- `python3 -m ruff check vllm_ascend/quantization/modelslim_config.py tests/ut/quantization/test_modelslim_config.py`
- `python3 -m ruff format --check vllm_ascend/quantization/modelslim_config.py tests/ut/quantization/test_modelslim_config.py`
- `python3 -m py_compile vllm_ascend/quantization/modelslim_config.py tests/ut/quantization/test_modelslim_config.py`
- `git diff --check`

Unit coverage was added for:

- Gemma4 `k_eq_v` missing-`v_proj` handling.
- Preserving the existing error behavior for other missing packed shards.
- Gemma4 MoE expert prefix mapping.
- Keeping the MoE expert prefix adaptation scoped to Gemma4 / Gemma4 text.

Not run locally:

- `python3 -m pytest tests/ut/quantization/test_modelslim_config.py -q`, because the current local Python environment is missing `torch`.

Base references used during development:

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/quantization/modelslim_config.py
+++ b/vllm_ascend/quantization/modelslim_config.py
@@ -345,6 +345,19 @@ def get_packed_modules_mapping(model_type: str) -> dict[str, list[str]]:
     return packed_modules_model_mapping.get(model_type, {})
 
 
+def _is_missing_k_eq_v_shard(shard_key: str, quant_description: Mapping[str, Any]) -> bool:
+    """Return whether the missing shard matches Gemma4's k_eq_v layout.
+
+    k_eq_v layers do not have a dedicated v_proj entry because v_proj is
+    replicated from k_proj at load time. Keep q_proj/k_proj mandatory so other
+    malformed packed quant descriptions still fail as before.
+    """
+    if not shard_key.endswith(".v_proj.weight"):
+        return False
+    shard_prefix = shard_key[: -len("v_proj.weight")]
+    return f"{shard_prefix}q_proj.weight" in quant_description and f"{shard_prefix}k_proj.weight" in quant_description
+
+
 def get_linear_quant_type(
     quant_description: dict[str, Any], prefix: str, packed_modules_mapping: dict[str, Any]
 ) -> str | None:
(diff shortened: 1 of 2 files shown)


realliujiaxu:
please use PR template

vllm_ascend/quantization/modelslim_config.py
```diff
@@ -345,6 +379,19 @@ def get_packed_modules_mapping(model_type: str) -> dict[str, list[str]]:
     return packed_modules_model_mapping.get(model_type, {})
 
 
+def _is_missing_k_eq_v_shard(shard_key: str, quant_description: Mapping[str, Any]) -> bool:
```
realliujiaxu: How about `_is_missing_v_shard`

Tflowers-0129: 看一下上游这个逻辑的取名

Alex-stack-hub: Upstream Gemma4 handles this by detecting `attention_k_eq_v` full-attention layers and reusing `k_proj` as `v_proj` during weight loading. It does not relax all missing shards. This PR follows the same semantics for ModelSlim quant config: only the missing `v_proj` entry in Gemma4 k_eq_v layout is skipped when `q_proj` and `k_proj` exist; other missing shards still fail as before. I renamed the helper to `_is_missing_v_shard` per the review suggestion and kept the k_eq_v explanation in the comment.

Tflowers-0129 (approved):

realliujiaxu (approved):
LGTM

Issue #6628, opened 2026-02-09
[RFC]: Design of the decode-based KV cache storage solution


DreamerLeader:
### Motivation.

11

### Proposed Change.

11

DreamerLeader:
Building upon the original logic of storing heads, and by integrating the capabilities of the pooling interface, each transmission is implemented based on the TP structure of the prefill node's heads. The decode node is required to store the corresponding TP heads from the prefill node in order to construct the keys; the content stored consists of the starting address of each block shown in the diagram, plus an offset address.

<img width="1342" height="1176" alt="Image" src="https://github.com/user-attachments/assets/c78b8676-5b9d-4d17-a389-9cc4f9cc61a6" />

Nagisa125:
Hello, how is the current progress? Has the issue been resolved in the new version?

PR #11582, opened 2026-07-07
[Feature][KV Pool] Support TP-mismatch PD disaggregated KV pooling for GQA

DreamerLeader:
…r GQA

Add AscendStore pooling adaptation for TP-unequal prefill/decode (e.g. prefill tp4 / decode tp2) on GQA dense-attention models. Decode stores KV via strided put; prefill loads via strided get, sharing a unified effective_tp namespace so cached blocks are mutually reachable.

All new paths are gated behind tp_mismatch (default False): when prefill_tp_size/decode_tp_size are not configured, behavior is line-by-line identical to before.

- pool_worker: parse prefill/decode_tp_size into effective_tp and num_sub_keys; strided address build + tp_mismatch load/store
- kv_transfer: thread dispatch hooks for tp_mismatch store/load
- pool_scheduler: gate empty new_block_ids handling under tp_mismatch
- tests: add TP-mismatch unit tests for worker and transfer dispatch

This PR introduces support for TP-asymmetric (tensor parallel mismatch) prefill/decode strided KV transfer. This allows disaggregated prefill and decode nodes to transfer KV caches even when they run with different tensor parallel sizes (e.g., decode TP=2 and prefill TP=4). It implements strided I/O address calculations, sub-key generation, and specialized load/store logic to handle the mismatch.

Feedback on the implementation:
- **Race Condition**: `self._invalid_block_ids` is updated in `_load_kv_tp_mismatch` without acquiring `self._invalid_block_ids_lock`, which can lead to concurrent modification issues.
- **Potential IndexError**: In `_build_tp_mismatch_keys_and_addrs`, `block_ids` is accessed using `start // self.block_size` without a boundary check.
- **Exception Handling in `_store_kv_tp_mismatch`**: If an exception occurs during the store operation, `dec_stored_request` is skipped, which could cause hangs.
- **Exception Handling in `KVCacheStoreRecvingThread._handle_request`**: If an exception occurs, `task_done()` is skipped, and a missing `None` check on `load_spec` could lead to an `AttributeError`.

No, this is an internal optimization and feature enablement for disaggregated KV transfer under TP-mismatched configurations.

New unit tests have been added in `test_kv_transfer.py` and `test_pool_worker.py` to verify the TP-mismatch dispatching, key/address building, and worker loading/storing logic.

report：
https://github.com/DreamerLeader/vllm-ascend/blob/main/PHASE1_TP_MISMATCH_ANALYSIS.md

- vLLM version: v0.23.0

rfc：https://github.com/vllm-project/vllm-ascend/issues/6628


LCAIZJ (approved):

LCAIZJ:
Could you add a startup script for this feature?

PR #11589, opened 2026-07-07
[BugFix] Guard mixed structured output backends

### LINKED ISSUE #11588: V1 structured outputs can crash on mixed auto backends
### Problem

When `structured_outputs_config.backend` is `auto`, two structured-output requests can resolve to different backends in the same V1 engine lifetime.

One reproduced sequence is:

1. Send an object JSON schema request. It resolves to `xgrammar` and returns HTTP 200.
2. Send a later JSON schema request that `auto` resolves away from `xgrammar` to `guidance`, for example a tuple-array style schema:

```json
{"type":"array","items":[{"type":"string"}]}
```

V1 structured outputs initialize one engine-level backend manager. After the first request latches `xgrammar`, the later request can still reach the initialized `xgrammar` backend even though request validation resolved it to `guidance`. That backend mismatch can fail during grammar compilation.

In the failing live run, the API returned HTTP 500 and the engine became unavailable. The server log contained:

```text
json_schema_converter.cc:2999: items must be a boolean or an object
```

This is not caused by an invalid `null` schema. A request using `{"type":"null"}` returned HTTP 200 with content `null` in the same investigation.

### Related upstream context

This overlaps with https://github.com/vllm-project/vllm/issues/43920. The shared root cause is the V1 single-backend-per-engine invariant combined with per-request `backend=auto` fallback and uncaught grammar compilation exceptions.

The reproducers are not identical:

- upstream vLLM #43920 uses malformed EBNF/regex after xgrammar is latched;
- this vllm-ascend report uses JSON schema that resolves to `guidance` after xgrammar is latched.

The most relevant upstream fix candidate is https://github.com/vllm-project/vllm/pull/44401. That PR is broader: it changes EngineCore/scheduler/serving error propagation so grammar compilation failures fail a single request instead of killing the engine. The vllm-ascend fix in #11589 is intentionally narrower: reject mixed backend transitions before grammar compilation, matching the existing one-backend-per-engine constraint.

### Expected behavior

The service should not crash the engine or return a server-side 500 for this backend mismatch. It should either keep using one backend consistently or reject the mixed-backend request before grammar compilation.

### Proposed fix

Add a vllm-ascend guard for V1 structured outputs:

- record the first resolved backend on the engine structured-output config;
- reject later requests that resolve to a different backend with `VLLMValidationError`;
- keep a fallback guard in `StructuredOutputManager.grammar_init` so requests that bypass API-side validation still fail before backend grammar compilation.

With the patch, the mixed `auto` case returns a client-side validation error and the service remains healthy.

QwertyJack:

Fixes #11588.

This PR adds a vllm-ascend guard for V1 structured outputs when `backend=auto` resolves different requests to different backends in the same engine lifetime. V1 structured outputs use one engine-level backend manager. In the reported failure sequence, the first schema initialized `xgrammar`, and a later schema resolved to `guidance`; the later request could still reach the initialized `xgrammar` backend and crash during grammar compilation.

This PR:

- Records the first resolved structured-output backend on `StructuredOutputsConfig`.
- Rejects later requests that resolve to a different backend with `VLLMValidationError` before grammar compilation.
- Adds a fallback guard in `StructuredOutputManager.grammar_init` for requests that bypass API-side validation.
- Supports both active vLLM `_validate_structured_outputs` signatures used by CI validation refs.
- Adds focused unit coverage for mixed-backend rejection, same-backend success paths, and subclassed backend instances.

Related upstream context: this overlaps with https://github.com/vllm-project/vllm/issues/43920. The shared root cause is the V1 single-backend-per-engine invariant combined with per-request `backend=auto` fallback and uncaught grammar compilation exceptions. The most relevant upstream fix candidate is https://github.com/vllm-project/vllm/pull/44401, but this PR intentionally keeps the vllm-ascend fix narrower by rejecting mixed backend transitions before grammar compilation.

Yes. Requests that resolve to a structured-output backend different from the backend already initialized by the engine now fail with a clear validation error instead of surfacing as a server-side grammar compilation crash / HTTP 500. Requests that continue using the same backend are unchanged.

- Added unit tests in `tests/ut/patch/platform/test_patch_structured_output.py` for mixed-backend rejection, same-backend success paths, failed-first-validation behavior, and subclassed backend detection.
- Ran local checks:
  - `python -m py_compile vllm_ascend/patch/platform/patch_structured_output.py tests/ut/patch/platform/test_patch_structured_output.py`
  - `ruff check vllm_ascend/patch/platform/patch_structured_output.py tests/ut/patch/platform/test_patch_structured_output.py`
  - `ruff format --check vllm_ascend/patch/platform/patch_structured_output.py tests/ut/patch/platform/test_patch_structured_output.py`
  - `PYTHONPATH=<vllm-src>:<vllm-ascend-src> python -m pytest -q tests/ut/patch/platform/test_patch_structured_output.py` -> `7 passed`
- Checked the updated GitHub Actions run for this PR: `lint-and-select-tests` passed, and the CPU selected tests that previously failed with `_validate_structured_outputs` signature mismatches passed for both `1f486d96a17303ce8db8e02be39545b2be338446` and `v0.23.0`.
- Validated with a real-weight live service using `/models/DeepSeek-V4-Flash-w8a8-mtp-0507`, TP4/DP4/EP, ACLGraph, async scheduling, MTP1 eager via `--speculative-config`, `max_model_len=133120`, and `max_num_seqs=16`:
  - unpatched: the user repro case `items=[]` returned HTTP 500 and logged xgrammar `items must be a boolean or an object` from EngineCore;
  - patched: the same case returned HTTP 400 with the mixed-backend validation message;
  - patched: `/v1/models` and `/health` stayed HTTP 200 after the rejected request;
  - patched: the full user repro loop did not reproduce a server-side 500.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/patch/platform/patch_structured_output.py
+++ b/vllm_ascend/patch/platform/patch_structured_output.py
@@ -1,0 +1,122 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (c) 2026 Huawei Technologies Co., Ltd. All Rights Reserved.
+
+from __future__ import annotations
+
+from typing import Any
+
+from vllm.exceptions import VLLMValidationError
+from vllm.sampling_params import SamplingParams
+from vllm.v1.structured_output import StructuredOutputManager
+
+_BACKEND_ATTR = "_vllm_ascend_structured_output_backend"
+_ORIGINAL_GRAMMAR_INIT_ATTR = "_vllm_ascend_original_grammar_init"
+_ORIGINAL_VALIDATE_ATTR = "_vllm_ascend_original_validate_structured_outputs"
+_PATCHED_MANAGER_ATTR = "_vllm_ascend_backend_guard_patched"
+_PATCHED_SAMPLING_PARAMS_ATTR = "_vllm_ascend_backend_validation_guard_patched"
+
+
+def _request_backend(request: Any) -> str | None:
+    if getattr(request, "structured_output_request", None) is None:
+        return None
+
+    sampling_params = getattr(request, "sampling_params", None)
+    structured_outputs = getattr(sampling_params, "structured_outputs", None)
+    backend = getattr(structured_outputs, "_backend", None)
+    return backend if isinstance(backend, str) else None
+
+
+def _backend_name_from_instance(backend: Any) -> str | None:
+    if backend is None:
+        return None
+
+    backend_names = {
+        "XgrammarBackend": "xgrammar",
+        "GuidanceBackend": "guidance",
+        "OutlinesBackend": "outlines",
+        "LMFormatEnforcerBackend": "lm-format-enforcer",
+    }
+    for backend_cls in type(backend).__mro__:
+        for class_name, backend_name in backend_names.items():
+            if class_name in backend_cls.__name__:
+                return backend_name
+    return None
+
+
+def _raise_mixed_backend(initialized_backend: str, request_backend: str) -> None:
+    raise VLLMValidationError(
+        "V1 structured outputs only supports one backend per engine. "
+        f"The engine is already using '{initialized_backend}', but "
+        f"this request resolved to '{request_backend}'. Configure "
+        "`structured_outputs_config.backend` explicitly or use schemas "
+        "supported by the initialized backend."
+    )
+
+
+def _sampling_params_backend(sampling_params: SamplingParams) -> str | None:
+    structured_outputs = getattr(sampling_params, "structured_outputs", None)
+    backend = getattr(structured_outputs, "_backend", None)
+    return backend if isinstance(backend, str) else None
+
+
+def _patch_sampling_params_validation() -> None:
+    if getattr(SamplingParams, _PATCHED_SAMPLING_PARAMS_ATTR, False):
+        return
+
+    original_validate = SamplingParams._validate_structured_outputs
+    setattr(SamplingParams, _ORIGINAL_VALIDATE_ATTR, original_validate)
+
+    def _validate_structured_outputs(
+        self: SamplingParams,
+        structured_outputs_config: Any,
+        tokenizer: Any,
+    ) -> None:
+        result = original_validate(self, structured_outputs_config, tokenizer)
+        request_backend = _sampling_params_backend(self)
+        if structured_outputs_config is None or request_backend is None:
+            return result
+
+        initialized_backend = getattr(structured_outputs_config, _BACKEND_ATTR, None)
+        if initialized_backend is not None and request_backend != initialized_backend:
+            _raise_mixed_backend(initialized_backend, request_backend)
+
+        setattr(structured_outputs_config, _BACKEND_ATTR, request_backend)
+        return result
+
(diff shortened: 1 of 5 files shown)


vllm_ascend/patch/platform/patch_structured_output.py
```diff
@@ -0,0 +1,117 @@
...
+def _backend_name_from_instance(backend: Any) -> str | None:
+    if backend is None:
+        return None
+
+    return {
+        "XgrammarBackend": "xgrammar",
+        "GuidanceBackend": "guidance",
+        "OutlinesBackend": "outlines",
+        "LMFormatEnforcerBackend": "lm-format-enforcer",
+    }.get(type(backend).__name__)
```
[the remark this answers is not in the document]
QwertyJack: Addressed in f735791c5. `_backend_name_from_instance` now walks the backend class MRO and recognizes backend-family class names, and `test_manager_rejects_mixed_backend_after_subclassed_backend_is_initialized` covers the subclass/pre-initialized fallback path.

Tflowers-0129 (approved):

realliujiaxu (approved):
LGTM. @zzzzwwjj please review this patch

vllm_ascend/patch/platform/patch_structured_output.py
```diff
@@ -0,0 +1,144 @@
...
+
+        setattr(structured_outputs_config, _BACKEND_ATTR, request_backend)
+        return result
+
+    SamplingParams._validate_structured_outputs = _validate_structured_outputs
+    setattr(SamplingParams, _PATCHED_SAMPLING_PARAMS_ATTR, True)
+
+
+def _patch_structured_output_manager() -> None:
+    if getattr(StructuredOutputManager, _PATCHED_MANAGER_ATTR, False):
```
MengqingCao: I think the `adapt_patch` will just do once in platform, why we need this const value to early exit the patch? It makes the code more complex.

QwertyJack: Good point. These guards were defensive against repeated module reloads/direct imports, but the normal platform `adapt_patch` path imports the platform patch module only once via Python module cache. They are not required for the production path and did make the patch noisier.

Removed the `_PATCHED_*` early-exit attrs in 336e6ba72. The `_ORIGINAL_*` attrs remain because they keep the original vLLM methods available for delegation/test inspection.

wangxiyuan (approved):

MengqingCao:
@QwertyJack plz also add the test case https://github.com/vllm-project/vllm-ascend/pull/11635#issuecomment-4926214310 into this pr, I forgot to update the comment in this pr before.

QwertyJack:
Thanks for the reminder. Added the same e2e guided-decoding case in 941ef6ed8. It starts with `backend=auto`, initializes xgrammar with a supported schema, then uses a valid `multipleOf` schema that resolves to guidance and verifies the mixed-backend validation error. I also compared the structured-output patch, unit test, and this e2e file against #11635; the three files are now identical across both PRs.

PR #11609, opened 2026-07-08
[BugFix][FusedMoE] Use model activation instead of SwiGLU in quantized MoE MLP

0moyi0-2024:

`quant_apply_mlp` hardcoded SwiGLU: every quantized MoE layer ran through the fused SwiGLU+quant NPU ops (`npu_grouped_matmul_swiglu_quant` / `npu_dequant_swiglu_quant` / `npu_swiglu`) regardless of the model's configured activation. Models that use GELU (e.g. Gemma4, `activation='gelu_tanh'`) therefore ran the wrong activation in their quantized experts, while the float path (`unquant_apply_mlp`) already dispatched correctly — causing a large accuracy drop.

This adds a GELU path to `quant_apply_mlp` that runs dequant GMM -> GELU -> (re)quant -> GMM2, mirroring the float activation. It returns early **before** the SwiGLU paths, so the existing SwiGLU logic is untouched. The activation is threaded from `unified_apply_mlp` (already available on `MoEMlpComputeInput`) via a new `activation` kwarg. Covers both the W4A16 antiquant and the W8A8/W4A8 int-quant layouts; MXFP is excluded.

The change is purely additive (single file, +87 lines, 0 deletions); no existing SwiGLU code path is modified.

No API change. Quantized MoE models using GELU (e.g. Gemma4) now use the correct activation instead of SwiGLU — an accuracy fix.

## How was this patch tested?

Manual testing on A5 (Ascend 950PR) with `gemma-4-26b-a4b-it-w8a8` (W8A8_DYNAMIC, `gelu_tanh`), TP=4, `cudagraph_mode=FULL_DECODE_ONLY`. GPQA Diamond (198q):

| Activation path | mean_acc |
|---|---|
| Before (SwiGLU, buggy) | 0.5404 |
| After (GELU, fixed) | 0.7121 |

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/ops/test_moe_mlp_gelu.py
+++ b/tests/ut/ops/test_moe_mlp_gelu.py
@@ -1,0 +1,351 @@
+"""Unit tests for GELU activation support in quantized MoE MLP.
+
+These tests cover the GELU early-return path added to ``quant_apply_mlp``
+(PR #11609). ``quant_apply_mlp`` previously hardcoded SwiGLU, so quantized MoE
+experts of GELU models (e.g. Gemma4, ``activation='gelu_tanh'``) ran the wrong
+activation. The new path runs dequant GMM -> GELU -> (re)quant -> GMM2 and
+returns before any SwiGLU branch.
+
+All NPU operators are mocked, so the tests run on CPU without Ascend hardware.
+"""
+
+import unittest
+from unittest.mock import MagicMock, patch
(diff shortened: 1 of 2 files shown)


realliujiaxu:
@weijinqian0 please help review this, the `moe_mlp` is getting more and more complicated

PR #11620, opened 2026-07-08
[bugFix] refresh block size in mooncake connector schedule

liziyu179:

- vLLM version: v0.23.0


LCAIZJ:
Could you describe the purpose of this change or what problem it addresses?

PR #11641, opened 2026-07-08
[BugFix] Filter linear-attention keys out of FIA graph replay for hybrid models

Liuchenbing-2026:

Hybrid models such as Qwen3.5 / Qwen3-Next mix self-attention layers (updated through `npu_fused_infer_attention_score`, "FIA") with linear-attention (GatedDeltaNet, "GDN") layers.

In `update_graph_params`, the non-draft branch builds the key list via `list(attn_metadata.keys())`, which contains **both** the FIA and the GDN layers. Only the FIA layers are captured into `graph_params.attn_params`; the GDN layers use a separate conv1d update path and their metadata (`GDNAttentionMetadata`) has neither `seq_lens_list` nor `actual_seq_lengths_q`.

Under **full-graph decode replay** this raises, on the first request:

```
AttributeError: 'GDNAttentionMetadata' object has no attribute 'seq_lens_list'
```

which kills the worker. The existing "self_attn first + zip truncation" ordering assumption does not hold on the layer-aware replay path, where the key list is expanded to the captured op count and cycles back into the GDN keys.

### Fix

Add `_filter_fia_attn_keys`, which keeps only keys whose metadata carries both `seq_lens_list` and `actual_seq_lengths_q` (i.e. real FIA metadata), and apply it to the non-draft `attn_keys` before the FIA graph-replay loop. GDN layers are correctly skipped; their conv1d state update path is unaffected.

Single file, `vllm_ascend/attention/attention_v1.py`, filtering only.

No. It fixes a crash; there is no API or behavioural change for models that do not hit the GDN-in-FIA-replay path.

Verified on Ascend NPU with a hybrid Qwen3.5 model under `cudagraph_mode=FULL_DECODE_ONLY`, TP=2: without the change the first decode request crashes with the `seq_lens_list` AttributeError above; with the change graph capture and replay proceed and requests return normally.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -72,6 +72,26 @@ SWA_INT_MAX = 2147483647
 _ATTN_KEYS_BUFFER = None
 
 
+def _filter_fia_attn_keys(attn_keys, attn_metadata):
+    """Keep only the self-attention (FIA) layers in the graph-replay key list.
+
+    Hybrid models such as Qwen3.5 / Qwen3-Next mix self-attention layers
+    (updated through npu_fused_infer_attention_score) with linear-attention
+    (GatedDeltaNet) layers. Only the self-attention layers are captured into
+    `graph_params.attn_params`; the linear-attention layers use a separate
+    conv1d update path and their metadata (`GDNAttentionMetadata`) has neither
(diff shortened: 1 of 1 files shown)


Liuchenbing-2026:
Closing as duplicate of #10256, which fixes the same crash (`GDNAttentionMetadata` has no `seq_lens_list` in `update_graph_params`) at the same location and was opened earlier. Thanks!

PR #11691, opened 2026-07-09
[Performance][310p] fix performance degradation

ShySummer:
This issue occurs in the GDN conv1d parameter update path in the Qwen3.5 310P + MTP + ACL Graph replay scenario. In the old implementation, when the metadata parameters are updated before each replay, the `spec_query_start_loc / spec_state_indices / num_accepted_tokens` is converted into a host tuple and then written back to the graph buffer. As a result, each layer in the decoding process has an additional NPU→CPU→NPU round-trip, causing performance deterioration. This modification directly updates the real metadata parameters on the NPU and then copies them back to the fixed buffer for replay, reducing the H2D copy latency and improving the inference performance.

NA

How was this patch tested?
Local test

vLLM version: v0.23.0
vLLM main: https://github.com/vllm-project/vllm/commit/0fc695fc6

- vLLM version: v0.23.0


Alex-stack-hub:
LGTM overall. Please add the 310P MTP ACL graph replay validation/perf data, and double-check the new padding path matches the old host-padding behavior.

PR #11697, opened 2026-07-09
[LoRA] Fix zero LoRA delta for VLM-wrapped hybrid dense models

Liuchenbing-2026:

For VLM-wrapped hybrid **dense** models — where the text tower is nested under a
submodule such as `language_model` — LoRA silently has **no effect**: the LoRA
output is bit-identical to the base model.

**Root cause.** The adapter tensors load correctly but are keyed by the bare
module name, e.g. `model.layers.N.mlp.down_proj`, while the model registers its
LoRA modules under the wrapped name,
e.g. `language_model.model.layers.N.mlp.down_proj`. The activation-time weight
lookup uses the wrapped model name, misses the bare-keyed adapter weights, and
`set_lora` receives a zero buffer — so the LoRA delta is zero.

**Fix.** Right after `WorkerLoRAManager._load_adapter` loads the `LoRAModel`
(but before registration / packed-merge), auto-detect the wrapper prefix from
`LoRAModelManager.modules` (find a model module whose name ends with a
non-packed adapter key) and remap the `loras` dict keys to line up with the
model's real module names. Once aligned, plain modules match directly and
`_create_merged_loras_inplace` packs `qkv_proj` / `gate_up_proj` automatically.

Implemented as a worker patch, consistent with the existing
`vllm_ascend/patch/worker/` mechanism.

No behavior change for models that already align (keys are left untouched).
For wrapped hybrid dense models, LoRA now applies a non-zero delta as expected.
The behavior can be disabled with `VLLM_ASCEND_LORA_VLM_PREFIX=0`.

### How was it tested?

Verified on a wrapped hybrid dense checkpoint: without the fix the LoRA output
matches the base model exactly (delta = 0); with the fix the adapter keys are
remapped, `set_lora` receives non-zero weights, and the LoRA output diverges
from base as expected.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/patch/worker/__init__.py
+++ b/vllm_ascend/patch/worker/__init__.py
@@ -67,6 +67,7 @@ import vllm_ascend.patch.worker.patch_cudagraph  # noqa
 import vllm_ascend.patch.worker.patch_deepseek_mtp  # noqa
 import vllm_ascend.patch.worker.patch_deepseek_v2  # noqa
 import vllm_ascend.patch.worker.patch_gqa_c8  # noqa
+import vllm_ascend.patch.worker.patch_lora_vlm_prefix  # noqa
 
 # vLLM's use_v2_model_runner may enable the v2 runner without the
 # VLLM_USE_V2_MODEL_RUNNER env var (e.g. based on model architecture).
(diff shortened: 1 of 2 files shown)


Liuchenbing-2026:
Superseded by #11708, which bundles this name-mapping fix together with the GDN graph capture/replay fixes needed to run Qwen3.5 hybrid dense LoRA end-to-end in both eager and full-graph modes. Closing in favor of #11708.

PR #11699, opened 2026-07-09

singzhou:

No.

Local machine tested.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/ops/test_moe_mlp.py
+++ b/tests/ut/ops/test_moe_mlp.py
@@ -309,7 +309,7 @@ def _mock_w8a8_gelu_compute(gate_up, *, gmm2_out=None, capture_quant=False):
     stream_patch, evt = _patch_npu_stream()
     captured = {}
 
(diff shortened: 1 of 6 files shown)


vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -1319,6 +1320,9 @@ def _run_merged_draft(
 
             model_hidden_states, model_positions = self.maybe_pad_and_reduce(model_hidden_states, model_positions)
 
+            # Align position cache for multi-step drafting to ensure cross-device compatibility:
```
singzhou: The reason why adding update_cos_sin is that:

Non-310P device: Tolerates skipping `update_cos_sin` by passing 'position' to the position embedding operator (torch_npu._npu_rotary_embedding or triton impl), the cache slice is actually wrong here.

310P device: Strictly requires the correct position embedding cache slice (required by torch_npu.npu_apply_rotary_pos_emb).

PR #10977, opened 2026-06-25
[Feature][LoRA][MoE] Support LoRA with unquantified MoE models

paulyu12:
### What this PR does / why we need it?
By injecting lora w13/w2 delta logics into `unquant_apply_mlp`, now this pr makes vLLM Ascend support lora with unquantified MoE models, such as Qwen3-30B-A3B. `TP>1` and `ACLGraph mode` are supported simultaneously. Performance improvement will be considered in the near future.

### Does this PR introduce _any_ user-facing change?
No.

### How was this patch tested?
pytest -sv tests/e2e/pull_request/one_card/lora/test_olmoe_lora.py
pytest -sv tests/e2e/pull_request/two_card/lora/test_qwen3moe_lora_tp.py

Co-authored with @liuchenbing and @wangzhao-11a.

- vLLM version: v0.23.0
- vLLM main: https://github.com/vllm-project/vllm/commit/1f486d96a17303ce8db8e02be39545b2be338446

PR #11702, opened 2026-07-09
[Feature][LoRA][MoE] Support MoE LoRA on models with shared experts (opt-in)

Liuchenbing-2026:

vllm-ascend MoE LoRA (added in #10977) currently **rejects** models with
**shared experts** up-front — it raises `AssertionError` when
`base_layer._shared_experts is not None`. This excludes models such as
**Qwen3.5-MoE** and **DeepSeek-V3**.

The reason is that the shared-experts LoRA runs through vLLM's standard dense
LoRA wrappers, whose expand-slice path does not match the
`_C_ascend.sgmv_expand` stacked `lora_b` layout, and vLLM's `torch_ops` einsum
fallback fails on the same padded `lora_b_stacked` layout.

This PR adds an **opt-in** path so those models can run:

- New env `VLLM_ASCEND_MOE_LORA_ALLOW_SHARED_EXPERTS` (default **off**).
- When set to `1`, `_assert_ascend_moe_lora_supported` no longer rejects
  shared-experts layers, and
  `PunicaWrapperNPU._expand_slice_prefill` / `_expand_slice_decode` route the
  expand step through a **`torch.bmm` fallback** (per-token gather + bmm),
  dispatched via a mutating `torch.library` custom op so it stays opaque to
  Dynamo fullgraph. This bypasses both `sgmv_expand_slice` and
  `bgmv_expand_slice`.
- Default off means models **without** shared experts (e.g. `Qwen3-30B-A3B`)
  keep the fused `_C_ascend` path and are completely unaffected.

Only behind an opt-in env var. With `VLLM_ASCEND_MOE_LORA_ALLOW_SHARED_EXPERTS`
unset/`0` (the default), behavior is unchanged. When `1`, MoE LoRA becomes
usable on shared-experts models at the cost of throughput (the `torch.bmm`
fallback is slower than the fused NPU op).

### How was it tested?

Validated that a shared-experts MoE model serves and activates LoRA
end-to-end through the fallback path without errors (the `bmm` expand-slice
op is exercised on every decode step). This path is **experimental** and its
numerical parity against a production adapter is still being validated, hence
the opt-in gating and the throughput caveat — feedback welcome.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/envs.py
+++ b/vllm_ascend/envs.py
@@ -100,6 +100,19 @@ env_variables: dict[str, Callable[[], Any]] = {
     # `dispatch_gmm_combine_decode` can be used only for **decode node** moe layer
     # with W8A8. And MTP layer must be W8A8.
     "VLLM_ASCEND_ENABLE_FUSED_MC2": lambda: int(os.getenv("VLLM_ASCEND_ENABLE_FUSED_MC2", "0")),
+    # Allow MoE LoRA on models with shared experts (e.g. Qwen3.5-MoE,
+    # DeepSeek-V3). The shared-experts LoRA runs through vLLM's dense
+    # wrappers, whose expand-slice path does not match the
+    # _C_ascend.sgmv_expand stacked lora_b layout (and vLLM's torch_ops
+    # einsum fallback fails on the same layout). When set to 1,
+    # PunicaWrapperNPU._expand_slice_prefill / _expand_slice_decode bypass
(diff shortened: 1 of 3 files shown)


Liuchenbing-2026:
Closing in favor of the consolidated #12295 (Qwen3.5 hybrid-dense + shared-experts MoE LoRA with isolated ACL compile graphs), which supersedes this opt-in base.

PR #11708, opened 2026-07-09
[Feature][LoRA] Enable Qwen3.5 hybrid dense LoRA (eager + full-graph)

Liuchenbing-2026:

Enables LoRA on **Qwen3.5 hybrid dense** models in both **eager** and
**full-graph (full-decode)** modes. These models wrap the text tower under a
submodule (`language_model`) and mix self-attention with GatedDeltaNet (GDN)
linear-attention layers. Three issues stop LoRA from working; this PR fixes
all three together so the model runs end-to-end.

**1. Zero LoRA delta (name mapping).**
Adapter tensors are keyed by the bare module name (`model.layers.N...`) while
the model registers LoRA modules under the wrapped name
(`language_model.model.layers.N...`). The activation-time lookup misses,
`set_lora` receives a zero buffer, and the LoRA delta is zero (LoRA output ==
base). A worker patch auto-detects the wrapper prefix from
`LoRAModelManager.modules` and remaps the `loras` dict keys right after
`WorkerLoRAManager._load_adapter`, before registration/packed-merge. Keys are
left untouched when already aligned, so plain models are unaffected. Guarded
by `VLLM_ASCEND_LORA_VLM_PREFIX` (default on).

**2. Graph-capture crash.**
Graph replay walks the full `attn_metadata` key list and reaches a GDN layer
whose `GDNAttentionMetadata` has no `seq_lens_list` / `actual_seq_lengths_q`,
raising `AttributeError`. Filter the replay key list down to the captured
self-attention (FIA) layers.

**3. Graph-replay crash under MTP.**
With spec-decode sequences present, the builder folds plain decodes into
prefills and sets `num_decodes` to 0, so `non_spec_decode_fallback_meta` is
None; the `non_spec_decode` conv1d graph update then fetches host args off the
missing meta and kills the engine. Treat the task as a state no-op
(`cache_indices = -1`) when the fallback meta is absent, mirroring the spec
branch.

No change for existing models: the name remap is a no-op when adapter keys
already align, and the two GDN graph fixes only skip/short-circuit metadata
that is otherwise invalid. Qwen3.5 hybrid dense LoRA now runs in eager and
full-graph modes.

### How was it tested?

Verified end-to-end on a Qwen3.5 hybrid dense checkpoint: LoRA output diverges
from base as expected (non-zero delta), and the model serves in both eager and
full-decode (graph) modes without the capture/replay crashes above.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -74,6 +74,26 @@ SWA_INT_MAX = 2147483647
 _ATTN_KEYS_BUFFER = None
 
 
+def _filter_fia_attn_keys(attn_keys, attn_metadata):
+    """Keep only the self-attention (FIA) layers in the graph-replay key list.
+
+    Hybrid models such as Qwen3.5 / Qwen3-Next mix self-attention layers
+    (updated through npu_fused_infer_attention_score) with linear-attention
+    (GatedDeltaNet) layers. Only the self-attention layers are captured into
+    `graph_params.attn_params`; the linear-attention layers use a separate
+    conv1d update path and their metadata (`GDNAttentionMetadata`) has neither
+    `seq_lens_list` nor `actual_seq_lengths_q`. Iterating the raw key list in
+    the FIA graph-replay loop therefore reaches a linear-attention key and
+    raises ``AttributeError: 'GDNAttentionMetadata' object has no attribute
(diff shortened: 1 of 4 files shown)


Liuchenbing-2026:
Closing — superseded by #12295 which bundles this LoRA hybrid-dense fix together with the MoE shared-experts opt-in and the ACL base/adapter graph isolation. Thanks!

PR #11725, opened 2026-07-09
[BugFix][KVTransfer] Make kv_connector_extra_config optional for Mooncake PD separation

zhaotyer:

The official [DeepSeek-V4-Flash PD separation guide](https://docs.vllm.ai/projects/ascend/en/latest/tutorials/models/DeepSeek-V4-Flash.html#52-multi-node-pd-separation-deployment) uses `MooncakeHybridConnector` with heterogeneous TP (P `tp=4,dp=4`, D `tp=1,dp=16`). Currently users must manually specify the full `kv_connector_extra_config` with both prefill/decode `dp_size`/`tp_size` on every node, otherwise the connectors crash or misroute KV. This PR makes the config optional so topology is auto-discovered.

The prior auto-discovery attempt had several blocking bugs:

1. **Startup crash**: `MooncakeConnectorWorker.__init__` called `_resolve_parallel_sizes` as its first line, before `self.kv_role` was assigned → `AttributeError`. Both connectors affected.
2. **D-side never discovered remote topology**: `_update_remote_sizes_from_metadata` read from the ZMQ metadata cache with empty default `engine_id`/`handshake_port`, so `remote_tp_size.get("", {}).get(0)` always returned `None` and `_prefill_tp_size` stayed at the default 1 → KV shard routing wrong under heterogeneous TP.
3. **Hybrid `DONE_RECVING_MSG` dropped `dp_size`**: sent a 4-tuple vs the standard connector's 5-tuple, so P-side `_remote_decode_dp_size` was never populated.
4. **`check_kv_extra_config` warning contradicted behavior**: logged "Using local value" on mismatch without actually overriding, so the wrong config value was still consumed.

### Fixes
- Move `_resolve_parallel_sizes` after `self.kv_role` assignment in both connectors (and before the Mamba assertion in the hybrid connector).
- D-side now discovers P's `tp_size` from the per-request `meta.remote_ptp_size` (already populated by the prefiller in `build_connector_meta` and forwarded by the proxy), avoiding the empty ZMQ cache timing race.
- Hybrid `DONE_RECVING_MSG` now sends `(msg, req_id, port_num, tp_size, dp_size)`, matching the standard connector.
- `check_kv_extra_config` restores strict `ValueError` on local-role config mismatch (P checks `prefill`, D checks `decode`), and logs a debug message for the auto-discovery path when config is absent.
- Restore the `_prefill_tp_size < _decode_tp_size` validation, guarded by `_remote_sizes_resolved` so it does not fire on the default placeholder.

### Scope
MLA models (e.g. DeepSeek-V4) skip the `ascend_config` `pd_tp_ratio`/`pd_head_ratio`/`num_head_replica` derivation, so heterogeneous TP auto-discovery works end-to-end without `extra_config`. Non-MLA heterogeneous TP still requires `extra_config` — the `_P_TP` alltoall group is built irreversibly at `init_ascend_model_parallel` and cannot be reconstructed from runtime discovery. This is out of scope for this PR.

Yes — `kv_connector_extra_config` is now optional for Mooncake PD separation on MLA models. Users who continue to provide it see unchanged behavior (extra_config values take priority). Users who provide a mismatched *local-role* config now get an early `ValueError` instead of a silent warning.

## How was this patch tested?

- Unit tests added: `tests/ut/distributed/kv_transfer/test_mooncake_topology_resolution.py` (8 cases covering call order, D-side meta discovery, P-side extra_config fallback, hybrid connector discovery, and `check_kv_extra_config` scenarios).
- Manual e2e verification (planned on NPU): launch DeepSeek-V4 P(tp4,dp4)/D(tp1,dp16) per the official guide **without** `kv_connector_extra_config`, confirm no `AttributeError`, D-side `_prefill_tp_size=4` discovered from `meta.remote_ptp_size`, and KV transfer output matches the baseline with full `extra_config`.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/distributed/kv_transfer/test_mooncake_topology_resolution.py
+++ b/tests/ut/distributed/kv_transfer/test_mooncake_topology_resolution.py
@@ -1,0 +1,269 @@
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+# Copyright 2023 The vLLM team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# This file is a part of the vllm-ascend project.
+#
+"""Unit tests for Mooncake connector remote-topology auto-discovery.
+
+Covers the fixes that make ``kv_connector_extra_config`` optional for
+PD separation with heterogeneous TP (e.g. DeepSeek-V4 P(tp4)/D(tp1)):
+
+1. ``_resolve_parallel_sizes`` must run after ``self.kv_role`` is assigned
+   (regression: previously called as the first line of ``__init__``).
+2. D-side discovers P's tp_size from per-request ``meta.remote_ptp_size``,
+   not from the ZMQ metadata cache (which is empty before the first recv).
+3. ``check_kv_extra_config`` raises on local-role config mismatch but
+   tolerates a missing config (auto-discovery path).
+"""
+
+import types
+import unittest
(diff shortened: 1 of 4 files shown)


zhaotyer:
@wangxiyuan @MengqingCao
I don’t have a node test environment—could either of you two ask someone to help verify this?

PR #11732, opened 2026-07-09
[BugFix][FusedMoE] Use model activation instead of SwiGLU in quantized MoE MLP (in-branch variant)

0moyi0-2024:

`quant_apply_mlp` hardcoded SwiGLU: every quantized MoE layer ran through the fused SwiGLU+quant NPU ops (`npu_grouped_matmul_swiglu_quant` / `npu_dequant_swiglu_quant` / `npu_swiglu`) regardless of the model's configured activation. Models that use GELU (e.g. Gemma4, `activation='gelu_tanh'`) therefore ran the wrong activation in their quantized experts, while the float path (`unquant_apply_mlp`) already dispatched correctly — causing a large accuracy drop.

This PR adds GELU handling at the existing non-fused activation dispatch sites (W4A16 antiquant and W8A8/W4A8 int-quant) and guards the fused SwiGLU+quant sub-branches (`use_w4a8_per_channel_gmm_swiglu`, `_custom_gmm_swiglu_enabled`, `use_gmm_swiglu_quant_fusion`, and the MC2 entry) with `and not is_gelu_activation`, so GELU activations fall through to the non-fused `GMM -> GELU -> (re)quant -> GMM2` path. GELU reuses the existing GMM1/GMM2 calls (no duplication); SwiGLU models keep their original fused paths unchanged.

This is the **in-branch/guard variant** — minimal diff (+16/-4), reuses existing GMM calls. It is an alternative to #11609 (early-return variant, +84/-0, fully isolated). Correctness is equivalent; the two differ in tradeoff: this one touches 4 existing branch conditions (smaller diff) while #11609 is pure addition (zero impact on existing SwiGLU logic, larger diff).

No API change. Quantized MoE models using GELU (e.g. Gemma4) now use the correct activation instead of SwiGLU — an accuracy fix.

## How was this patch tested?

Unit tests: `tests/ut/ops/test_moe_mlp_gelu.py` (12 tests, all NPU ops mocked, run on CPU):
- W8A8/W4A8 int-quant: gelu_tanh vs gelu (approximate tanh/none) math, dequant GMM1 form, requant + GMM2 wiring, scale_bias handling (bias propagation, group_list_type 0->1), w1_scale dtype cast.
- W4A16 antiquant layout: antiquant GMM1 -> gelu*up -> antiquant GMM2, no requant.
- **Guard coverage**: fusion-on + GELU must skip `npu_grouped_matmul_swiglu_quant`; MC2 + GELU must skip the MC2 fused branch (`npu_dequant_swiglu_quant`). Both tests fail if the `and not is_gelu_activation` guards are removed.
- No-impact: silu / swiglustep / swigluoai do not enter the GELU path; `unified_apply_mlp` forwards `gelu_tanh` to `quant_apply_mlp`.
- Existing `tests/ut/ops/test_moe_mlp.py` still passes (8 tests).

Manual testing on A5 (Ascend 950PR) with `gemma-4-26b-a4b-it-w8a8` (W8A8_DYNAMIC, gelu_tanh), TP=4: GPQA Diamond (198q) improved from 0.5404 (SwiGLU) to 0.7121 (GELU).

- vLLM version: v0.23.0


realliujiaxu (approved):
LGTM

PR #11762, opened 2026-07-09
[Performance] Vectorize local seq len computation in SFAMetadataBuilder

zhenwenqi2024:

Replace the per-request Python loop that computed local query/key
lengths with on-device vectorized ops (clamp, cumsum, torch.where).
The loop issued 2 `.item()` NPU->CPU syncs per request (2 * num_reqs
syncs/step); the new path is fully on-device with zero syncs.

Behavior is preserved: each request's [global_start, global_end) range
is clipped to the local [local_start, local_end_with_pad) slice and
accumulated into actual_seq_lengths_query / actual_seq_lengths_key.
the profiling before is:

image

the profiling after is :
image

NO

- vLLM version: v0.23.0


yiz-liu (approved):

MengqingCao (approved):
LGTM, thx for this improvement!

PR #11767, opened 2026-07-09
[Feature] support ubatch overlap to improve MoE model prefill throughput

eillon:

This PR introduces **UBatch Overlap** (dual-batch compute/comm overlap) to vllm-ascend, improving MoE model prefill throughput on Ascend NPUs.

#### Motivation

For MoE models on Ascend, a single prefill forward pass spends a significant fraction of time inside MoE collective communication (`tensor_model_parallel_*`, MC2 dispatch/combine, EP all-gather/reduce-scatter). When the whole batch runs on one stream, compute and communication are serialized — AI cores sit idle during comm, and HCCS comm engines sit idle during compute, leaving hardware utilization low.

#### Approach

UBatch overlap splits a single scheduler step's token batch into N micro-batches (N=2) and executes them **concurrently** on 2 dedicated NPU streams. While one ubatch performs dense compute (matmul/attention), the other performs MoE collective communication, hiding MoE comm latency behind compute and shortening the critical path.

Key design:

- **Process-wide singleton `UBatchRuntimeManager`** (`vllm_ascend/worker/ubatch_utils.py`): orchestrates NPU streams, events, worker thread pool, and forward contexts. The two streams perform a lock-step ping-pong handoff via `compute_done_event` / `comm_done_event` + `cpu_event`.
- **`comm_section()` context manager**: inlined around `prepare()` / `finalize()` calls in the shared MoE base class `AscendFusedMoE.forward_impl` (`vllm_ascend/ops/fused_moe/fused_moe.py`). This is the **single** compute/comm switching point; it is a no-op when ubatch is disabled, so MoE call sites need no `is_ubatch_running` guard. Since every Ascend MoE model routes experts through `AscendFusedMoE` / `AscendSharedFusedMoE`, the overlap is **universal** — no per-model monkey-patch is required.
- **Attention metadata splitting** (`split_ascend_common_metadata`): handles the case where a single request spans the ubatch boundary, chunked-prefill attention-state remapping, and M-RoPE `(3, seq_len)` dimension-aware slicing.
- **Error recovery**: `_model_forward_ubatches` on worker exception first calls `release_parked_workers()` to wake parked threads, then `reset_thread_pool()` to discard the old pool, and `forward_finished(skip_event_wait=True)` skips waiting on NPU events that may never be recorded — preventing main-thread blocking.
- **Thread safety**: rotary cos/sin tables are sliced via a thread-local `token_slice` (not the shared `curr_batch` cursor, which would race under overlap); DeepStack cross-layer buffers are supported via `patch_qwen3_vl_deepstack`.

#### Configuration

Configured via vLLM `additional_config` (consistent with all other Ascend feature toggles), declared on `AscendConfig`:

| Key | Default | Meaning |
|----------|---------|---------|
| `num_ubatches` | `1` | Number of concurrent ubatches. **Only `1` (disabled) and `2` (enabled) are supported.** Values > 2 are rejected at config init (the lock-step handoff is a two-stream ping-pong by design; > 2 yields no additional overlap). |
| `ubatch_trigger_threshold` | `2048` | Minimum padded token count required to enable ubatch overlap for a given step. |

```bash
vllm serve <model> --additional-config='{"num_ubatches": 2}'
```

Activation conditions (all must hold per step): `num_ubatches>1` AND padded token count exceeds threshold AND last ubatch is non-empty AND all DP ranks vote consistently AND step contains prefill tokens (`with_prefill`) AND model is not blocked by `_ubatch_blocked_reason`.

#### Performance Gains (Highlight)

UBatch overlap delivers significant prefill-stage performance gains for MoE models. Test data across different dimensions:

**(1) Prefill performance across different input lengths**

| Input Length | Baseline (tokens/s) | UBatch (tokens/s) | Speedup |
|--------------|--------------------|-------------------|---------|
| 2K           | ____               | ____              | ____    |
| 4K           | ____               | ____              | ____    |
| 8K           | ____               | ____              | ____    |
| 16K          | ____               | ____              | ____    |
| 32K          | ____               | ____              | ____    |

> Note: Larger input lengths have higher MoE comm ratio, yielding greater ubatch overlap benefit. Batches below `ubatch_trigger_threshold` (default 2048) do not trigger ubatch.

**(2) Performance across different prefix hit rates (long-sequence scenario)**

| Prefix Hit Rate | Baseline (tokens/s) | UBatch (tokens/s) | Speedup |
|-----------------|--------------------|-------------------|---------|
| 0%              | ____               | ____              | ____    |
| 50%             | ____               | ____              | ____    |
| 90%             | ____               | ____              | ____    |
| 99%             | ____               | ____              | ____    |

> Note: Prefix hit rate affects prefill token count (higher hit rate = fewer actual prefill tokens). UBatch benefits most when there are sufficient effective prefill tokens.

**(3) Performance across different MoE models**

| Model | TP/EP | Baseline (tokens/s) | UBatch (tokens/s) | Speedup |
|-------|-------|---------------------|-------------------|---------|
| DeepSeek-V4 | ____  | ____                | ____              | ____    |
| Qwen3-MoE   | ____  | ____                | ____              | ____    |
| ____        | ____  | ____                | ____              | ____    |

> Note: Any MoE model using the MC2 dispatch/combine communication path benefits. AllReduce/AllGather paths are not supported due to cross-rank ordering dependencies.

#### Support Matrix

- **MoE communication paths**: MC2 `dispatch_v2`/`combine_v2` (CommSectionGuard + multi-stream safe, no cross-rank ordering dependency). Unsupported: AllReduce/AllGather/ReduceScatter/Send/Recv/Broadcast (cross-rank ordering dependency) and fused op `DispatchFFNCombine` (comm/compute not separable).
- **Multimodal models**:

| Model family | M-RoPE | DeepStack | Supported? |
|--------------|--------|-----------|------------|
| qwen2-vl / qwen2.5-vl | Yes | No | Yes |
| internvl / glm4v | No | No | Yes |
| qwen3-vl / qwen3-vl-moe | Yes | Yes | Yes (requires `patch_qwen3_vl_deepstack`) |
| qwen3-omni-moe-thinker | Yes | Yes | Yes (requires `patch_qwen3_vl_deepstack`) |

#### Limitations

- **Prefill only** — decode / spec decoding steps always run on the FULL ACL graph and are never ubatched.
- **First PP rank only** — `AscendConfig` rejects `num_ubatches>1` when `pipeline_parallel_size>1`.
- **DSA models not supported** (Distributed Sparse Attention, e.g. GLM-5, DeepSeek-V3.2) — SFA path relies on thread-unsafe global cos/sin buffers.
- **PCP not supported** — per-ubatch attention metadata splitter drops PCP slot-mapping state.
- Prefill runs eager (cannot reuse captured FULL ACL graph), hence the 2048-token threshold to avoid small-batch overhead.
- Known limitation: `reset_thread_pool` join may hang if HCCL is blocked, requiring process-level restart (documented in the error path comments).

---

Yes. Introduces two `additional_config` keys:

- `num_ubatches` (int, default `1`): set to `2` to enable ubatch overlap.
- `ubatch_trigger_threshold` (int, default `2048`): ubatch trigger threshold.

When enabled, MoE model prefill steps automatically use the dual-stream compute/comm overlap path when conditions are met; decode steps are unaffected. Users can monitor the actual enablement ratio via `ubatch_rate` and `tokens_ubatch_rate` in the logs. No API/interface breaking changes — when `num_ubatches=1` (default), behavior is identical to before.

---

**Unit tests** (`tests/ut/worker/test_ubatch_utils.py`, pure Python, no NPU required):

- `slice_query_start_locs` arithmetic correctness.
- `split_ascend_common_metadata`: whole-request split, request-spanning-ubatch-boundary (first/last request continuation), chunked-prefill attention-state remapping.
- Decision logic: `should_enable_ubatch` / `maybe_use_ubatch` / `get_ubatch_trigger_threshold`, including `is_last_ubatch_empty` short-circuit and `ubatch_trigger_threshold` override.
- Thread-local token slice binding / `get_cos_and_sin_slice` ubatch fallback.
- Concurrency (mocked `torch.npu.Stream`/`Event`): `comm_section` symmetry, `yield_thread` timeout, `reset_thread_pool` error recovery.

```bash
pytest -sv tests/ut/worker/test_ubatch_utils.py
```

**Config tests** (`tests/ut/test_ascend_config.py`): `num_ubatches` valid-value validation, PP>1 rejection, `ubatch_trigger_threshold` override.

**ST / Nightly** (NPU hardware, `tests/e2e/nightly/`):

- End-to-end accuracy: ubatch enabled vs disabled, logits numerical consistency.
- Performance benchmarks: prefill throughput across different input lengths, prefix hit rates, and MoE models.
- Error path: worker exception correctly recovers runtime state; subsequent steps run normally.
- Multimodal: qwen3-vl / qwen3-omni-moe-thinker (with M-RoPE + DeepStack patch) accuracy and performance validation.

**Manual verification commands**:

```bash
# Enable ubatch
vllm serve <model> --additional-config='{"num_ubatches": 2}'

# Verify ubatch enablement ratio (in logs)
# Expected: running ubatch, ubatch_rate: XX%, tokens_ubatch_rate: XX%
```

- vLLM version: v0.25.1


eillon:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

PR #11772, opened 2026-07-09
[feature] reduce cases that required DP padding

zzzzwwjj:

Reducing the cases that required DP padding.

Before this pr, if one of the three cases is met, padding is required:
1. cudagraphmode != None;
2. enable_sp == True;
3. is_draft_model == True;

This is not reasonable, some cases do not require padding across dp.

And after this pr, there are two cases where padding between DPs is required:
1. comm_method == ALLGATHER;
2. comm_method == MC2 and is hierarchical communication, in this case, the mc2 operator does not support dynamic batch size.
3. when finegrained_tp is open, we need to ensure num_tokens remains consistent within finegrained_tp_group.

None

- vLLM version: v0.25.1

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -151,7 +151,6 @@ from vllm_ascend.utils import (
     AscendDeviceType,
     calc_split_factor,
     check_gdn_layer,
-    embedding_tp_enable,
     enable_sfa_dcp_replicated_indexer,
     enable_sp,
     enable_sp_by_pass,
@@ -159,9 +158,9 @@ from vllm_ascend.utils import (
     get_c_env,
     global_stream,
     is_hidden_state_cache_spec,
+    is_hierarchical_communication_enabled,
     kv_cache_spec_uses_sparse_c8,
     lmhead_tp_enable,
-    oproj_tp_enable,
     set_potential_max_tokens,
     should_skip_allreduce_across_dp_group,
 )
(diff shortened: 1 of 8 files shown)


jianzs:
@zzzzwwjj  If the cudagraph mode is NONE, there’s no need to pad inside the model runner. We should add padding tokens before communication and drop them afterwards, right? Even ALLGATHER is selected.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -704,7 +704,13 @@ def _sync_metadata_across_dp(
         synced_cudagraph_mode = CUDAGraphMode(_post_process_cudagraph_mode(packed_tensor))
 
         # Create a tensor for num_tokens_after_padding
-        if allow_dp_padding or is_draft_model:
+        comm_method = select_moe_comm_method(max_tokens_across_dp, self.vllm_config)
+        is_mc2_with_hierarchical = (comm_method == MoECommType.MC2 and is_hierarchical_communication_enabled())
```
jianzs, line 663: Is hierarchical communication supported in A3?

zzzzwwjj:
> @zzzzwwjj If the cudagraph mode is NONE, there’s no need to pad inside the model runner. We should add padding tokens before communication and drop them afterwards, right? Even ALLGATHER is selected.

Uniforming the padding logic in ModelRunner is better for architecture, and we don't have to make the architecture very complex just for a little of performance optimization.

weijinqian0 (approved):

vllm_ascend/worker/model_runner_v1.py
zzzzwwjj: dispatch/combine/all2allv op have already supported A3 hierarchical communication, we need to set some param because we cannot automatically identify the server's network configuration. Fused_mc2 op don't now.

pisceskkk:
/revert
[Bot]: revert command failed: you do not have permission. Only the PR author or users with triage+ permission can trigger /revert.
[Bot]: revert command failed. Please check the [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/30345752592) for details.

zhangxinyuehfad:
/revert
[Bot]: revert completed successfully. New PR created: https://github.com/vllm-project/vllm-ascend/pull/13004

PR #11774, opened 2026-07-09
[BugFix][Worker] Reset slot_mapping to pad id for dummy graph capture

zhaochuang001:
Clear stale slot_mapping for all kv cache groups during dummy/graph-capture runs, where slot_mapping is not populated via _prepare_inputs(), to avoid graph capture reading uninitialized values.

No.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -3655,6 +3655,9 @@ class NPUModelRunner(GPUModelRunner):
                 for_cudagraph_capture=is_graph_capturing,
                 num_scheduled_tokens_np=num_scheduled_tokens,
             )
+            for kv_cache_gid in range(len(self.kv_cache_config.kv_cache_groups)):
+                blk_table = self.input_batch.block_table[kv_cache_gid]
+                blk_table.slot_mapping.gpu.fill_(-1)
 
         with self.maybe_dummy_run_with_lora(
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3655,6 +3655,9 @@ def _dummy_run(
                 for_cudagraph_capture=is_graph_capturing,
                 num_scheduled_tokens_np=num_scheduled_tokens,
             )
+            for kv_cache_gid in range(len(self.kv_cache_config.kv_cache_groups)):
+                blk_table = self.input_batch.block_table[kv_cache_gid]
+                blk_table.slot_mapping.gpu.fill_(-1)
```
yiz-liu:
```suggestion
            if not is_capturing:
                for kv_cache_gid in range(len(self.kv_cache_config.kv_cache_groups)):
                    blk_table = self.input_batch.block_table[kv_cache_gid]
                    blk_table.slot_mapping.gpu.fill_(-1)
```

zhaochuang001: There is no ‘’is_capturing‘’; it needs to be changed to ‘’is_graph_capturing‘’.

yiz-liu (approved):

linfeng-yuan (approved):

PR #11791, opened 2026-07-10
[Feature]  Gemma4  E2B and E4B

0moyi0-2024:

Gemma4 (E2B/E4B) uses cross-layer KV sharing: certain "target" layers do not maintain their own KV cache but read from a "producer" layer's cache. On Ascend, `reshape_and_cache` was unconditionally writing K/V for every layer, so a KV-sharing target layer would overwrite the producer's shared slots with its own (dummy/current) K/V, corrupting the shared cache and breaking correctness.

This PR skips the cache write for KV-sharing target layers in the Ascend attention backend:

- In `vllm_ascend/attention/attention_v1.py`, both `AscendAttentionBackendImpl.reshape_and_cache` and `AscendC8AttentionBackendImpl._reshape_and_cache` now return early without writing K/V when the layer is a KV-sharing target (`kv_sharing_target_layer_name is not None`). The producer layer still records `reshape_cache_event` before returning so write ordering is preserved; non-producer target layers neither write nor record.

This is required to support Gemma4 E2B/E4B correctly on Ascend.

No. This is an internal correctness fix with no new public API, flag, or config. The only behavior change (internal) is that KV-sharing target layers no longer write into the shared KV cache. Users running Gemma4 E2B/E4B on Ascend get correct outputs without any config change.

- Added unit tests in `tests/ut/attention/a2/test_attention_v1.py`:
  - `test_kv_sharing_target_skips_cache_write`: asserts `DeviceOperator.reshape_and_cache` is **not** called for a KV-sharing target (non-producer) layer, and that `key_cache`/`value_cache` are bound without any write.
  - `test_c8_kv_sharing_target_skips_nz_cache_write`: asserts `torch_npu.npu_scatter_pa_kv_cache` is **not** called on the C8/NZ cache path.
- Validated locally on Ascend950PR (A5 / 957b) with Gemma4: KV-sharing target layers no longer overwrite the shared KV cache and inference output is correct.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/attention/a2/test_attention_v1.py
+++ b/tests/ut/attention/a2/test_attention_v1.py
@@ -10,6 +10,7 @@ from vllm_ascend.attention.attention_v1 import (
     AscendAttentionBackendImpl,
     AscendAttentionMetadataBuilder,
     AscendAttentionState,
+    AscendC8AttentionBackendImpl,
 )
 from vllm_ascend.attention.kvcomp_attn.attention_utils import get_kvcomp_decode_params, reshape_and_cache_kvcomp
 from vllm_ascend.attention.utils import (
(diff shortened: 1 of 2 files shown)


yiz-liu (approved):
Please add nightly test case for Gemma4.

PR #11807, opened 2026-07-10
[BugFix] Fix small chunk problem for pcp+qwen3.5

SunnyLee151064:
What this PR does / why we need it
Fixes a runtime 507015 (MTE write out-of-bounds) crash in the AscendC chunk_gated_delta_rule_fwd_h / chunk_fwd_o kernels when running Qwen3.5 with PCP
(prefill-context-parallel) world_size > 1 under a mixed-length concurrent batch.

Root cause
PCP splits each request's prefill across ranks with an interleave granularity (cp_kv_cache_interleave_size, default 128). A request shorter than the interleave size
produces zero tokens on rank>0, leaving a zero-length segment in that rank's cu_seqlens.

chunk_indices carries compact (non-empty) segment ranks (empty segments are skipped when built), but the chunk_fwd_o kernel indexes gmSeqlen (= the original cu_seqlens,
with empty segments) by that same compact rank. When a zero-length segment is sandwiched between non-empty ones, the rank and the original segment index go out of sync.

Example
Suppose rank>0 has 3 requests with lengths [128, 0, 128] tokens (the middle request is shorter than the interleave size, so it gets 0 tokens on this rank):

cu_seqlens = [0, 128, 128, 256] # seg0=128, seg1=EMPTY, seg2=128
chunk_indices = [(0,0),(0,1),(1,0),(1,1)] # compact rank: seg0→0, seg2→1 (seg1 skipped)

chunk_fwd_o processing chunk3 (the last chunk of seg2, within-chunk idx=1):

tokenOffset = gmSeqlen[1] = 128 # cu_seqlens[1] (boundary of EMPTY seg1!)
batchTokens = gmSeqlen[2] - tokenOffset = 128 - 128 = 0
isFinalState = (chunkIdx == numChunks-1) = true
blockTokens = batchTokens - batchChunkIdxchunkSize
= 0 - 164 = -64 # uint32 underflow → huge value
→ MTE write out-of-bounds → runtime 507015

The fix
Add _compact_empty_segments(cu_seqlens_host, initial_state) and call it before the AscendC fwd_h / fwd_o kernels. It drops zero-length segments from cu_seqlens (and the
corresponding rows of initial_state), so the compact rank carried by chunk_indices lines up with the compacted gmSeqlen. The compacted final_state is scattered back to
the original [N, H, K, V] layout via the keep-mask (empty segments keep their initial_state). It is a no-op when there are no empty segments (rank0 / non-PCP / uniform
batch).

chunk_indices is already compact-ranked (built by skipping empty segments in gdn_attn_builder._fill_chunk_indices_cpu), so it is reused as-is; only cu_seqlens /
initial_state need compacting. Zero-length segments contribute exactly zero to the GDN math, so compaction preserves the token layout (empty segment = 0 tokens) and is
numerically equivalent.

vLLM version: v0.23.0
vLLM main: https://github.com/vllm-project/vllm/commit/1f486d96a17303ce8db8e02be39545b2be338446

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/ops/triton/fla/chunk.py
+++ b/vllm_ascend/ops/triton/fla/chunk.py
@@ -16,6 +16,8 @@ from vllm.distributed import get_pcp_group
 from vllm.forward_context import get_forward_context
 from vllm.model_executor.layers.fla.ops.utils import SUPPRESS_LEVEL
 
+from vllm_ascend.ops.gdn_attn_builder import _compact_empty_segments
+
 from .chunk_delta_h import chunk_gated_delta_rule_fwd_h  # noqa: F401
 from .chunk_delta_hupdate import chunk_gated_delta_rule_fwd_hupdate
 from .chunk_o import chunk_fwd_o  # noqa: F401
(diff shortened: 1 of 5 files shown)


vllm_ascend/ops/triton/fla/chunk.py
```diff
@@ -16,6 +16,8 @@
 from vllm.forward_context import get_forward_context
 from vllm.model_executor.layers.fla.ops.utils import SUPPRESS_LEVEL
 
+from vllm_ascend.ops.gdn_attn_builder import _compact_empty_segments
```
weiguihua2: Useless, remove it.

SunnyLee151064:
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/29078984216)

weiguihua2 (approved):

SunnyLee151064:
/cherry-pick releases/v0.23.0
[Bot]: cherry-pick command failed. Please check the [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/29217551486) for details.

linfeng-yuan (approved):

PR #11829, opened 2026-07-10
[v0.23.0][BugFix] Fix layerwise KV pool IndexError when MTP is enabled

tyy0829:
### Problem
When using layerwise KV pool (MemCache backend) with MTP speculative decoding enabled, vLLM crashes with `IndexError: list index out of range` on the first inference request.

### Root Cause
1. `_infer_cache_group_metadata` counts ALL layer names (including MTP draft layers) in `group_num_layers`, causing `num_layers` to be inflated (e.g., 47 → 48 for GLM-4.7-Flash with 1 MTP layer).
2. `layer_save_tasks` / `layer_load_tasks` are initialized with the original `num_layers` (47), but iterated with the inflated `num_layers` (48) → **IndexError**.
3. `wait_for_layer_load` lacks the MTP guard that `save_kv_layer` already has, causing a second IndexError when MTP draft model executes.

### Fix
1. **Add `_extract_physical_layer_index` method**: identifies MTP layers by name pattern (`mtp.N.xxx` → `num_hidden_layers + N`), which maps to an index >= `num_layers`.
2. **Skip MTP layers in `_infer_cache_group_metadata`**: MTP layers (`phys >= num_layers`) are excluded from `group_num_layers`, so `num_layers` stays at the correct value (47).
3. **Add `wait_for_layer_load` MTP guard**: `if self.current_layer >= self.num_layers: return`, matching the existing guard in `save_kv_layer`.

### Verification
Tested with GLM-4.7-Flash (TP=4) + MTP + layerwise MemCache:
- Service starts successfully (no IndexError)
- Inference works correctly
- KV pool cache hits on repeated requests (`hit_blocks=8/8`, External prefix cache hit rate: 45.4%)

### Note
This fix is already validated on the `feature/multi-group-v0.23` branch (commit 170094407). This PR cherry-picks the relevant changes to `releases/v0.23.0`.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
@@ -522,11 +522,32 @@ class KVPoolWorker:
         except AttributeError:
             return cache.storage().data_ptr()
 
+    def _extract_physical_layer_index(self, layer_name: str) -> int:
+        import regex as re
+
+        m = re.search(r"layers\.(\d+)", layer_name)
+        if m:
+            return int(m.group(1))
+        # MTP layers have names like "mtp.0.self_attn.xxx" without "layers."
+        # prefix. Map them after the main model layers.
+        if ".mtp." in f".{layer_name}.":
+            m = re.search(r"mtp\.(\d+)", layer_name)
+            if m:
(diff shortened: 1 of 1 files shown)


Pz1116 (approved):
TODO：Is there a more robust way to calculate all these layer indexes？

yiz-liu (approved):

PR #11831, opened 2026-07-10
[Bugfix] Add error log, eplb not support w4a8mxfp

Spicy-Stick:
Error Handling: Added an explicit check to raise a ValueError when the quantization type is set to W4A8MXFP, as it is currently unsupported by EPLB.

- vLLM version: v0.23.0


vllm_ascend/eplb/adaptor/vllm_adaptor.py
```diff
@@ -129,6 +129,8 @@ def init_expert_param_per_layer(self):
         for local_idx, layer in enumerate(self.moe_layers):
             quant_type = QuantType.NONE if self.model.quant_config is None else layer.quant_type
             expert_weight_key = (quant_type, get_ascend_config().enable_fused_mc2 == 1)
+            if expert_weight_key[0] == QuantType.W4A8MXFP:
```
zhenwenqi2024, line 132: RuntimeError？

Spicy-Stick: okay,  that sounds better

PR #11832, opened 2026-07-10
[BugFix][Worker] Restore DSA compress dummy metadata zeroing

maoxx241:

This PR restores explicit cache-metadata initialization for DSA/compress dummy runs.

Compress dummy and ACL graph-capture runs do not go through the normal `_prepare_inputs()` path. Their `slot_mapping` and block-table buffers can therefore retain values from a previous real request. Feeding those stale block and `[block, offset]` scatter indices into DSA kernels can cause intermittent accuracy errors, hangs, or out-of-bounds accesses.

The fix:

- marks `_build_attention_metadata()` calls from `_dummy_run()` explicitly;
- zeroes the complete dummy-run `slot_mapping` and block-table ranges only when compression is enabled;
- preserves the existing `-1` padding behavior for normal requests;
- includes a prominent `Do not remove` comment explaining the safety requirement.

This restores the safety behavior originally introduced by #9818 and lost during the synchronization in #11513.

No API or configuration change. It prevents DSA/compress dummy runs from consuming stale cache metadata.

- `python -m py_compile vllm_ascend/worker/model_runner_v1.py tests/ut/worker/a2/test_model_runner_v1.py`
- `PATH="$PWD/.venv/bin:$PATH" bash format.sh ci`
- `git diff --check`

No dedicated unit test was added because this branch mutates NPU-backed cache metadata inside the complete attention-metadata builder. The relevant DSA/compress dummy ACL graph path requires validation in the supported Ascend runtime; this macOS checkout does not have `pytest`, `torch`, `vllm`, or NPU hardware.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -3166,6 +3166,7 @@ class NPUModelRunner(GPUModelRunner):
         logits_indices: torch.Tensor | None = None,
         use_spec_decode: bool = False,
         for_cudagraph_capture: bool = False,
+        is_dummy_run: bool = False,
         num_scheduled_tokens: dict[str, int] | None = None,
         num_scheduled_tokens_np: np.ndarray | None = None,
         cascade_attn_prefix_lens: list[list[int]] | None = None,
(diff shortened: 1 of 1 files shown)


maoxx241:
Closing this draft because the issue is in the DSA/compress dummy-run slot_mapping initialization path, not the GDN core output buffer.

maoxx241:
Reopened after replacing the original GDN output-buffer patch with the DSA/compress dummy-run cache-metadata fix. Current head c66df419b only changes model_runner_v1.py and its worker regression test.

PR #11841, opened 2026-07-11
[Feature] Support AlltoAll + EP + LoRA in non‑quantized mode

wangx700:
This pull request enables LoRA support for Mixture of Experts (MoE) models running with Expert Parallelism (EP) on Ascend hardware. Previously, this configuration was broken and caused crashes. The changes introduce infrastructure to track and exchange per-token LoRA indices across Tensor Parallel (TP) ranks and during AlltoAll communication, ensuring that LoRA adapters are correctly applied to the distributed expert layers. It also includes updates to validation logic and adds a new end-to-end test case to ensure coverage for this feature.

- AlltoAll EP Support: Enabled LoRA support for MoE models using Expert Parallelism (AlltoAll) on Ascend hardware by allowing the AlltoAll communication backend.
- LoRA Index Exchange: Implemented a mechanism to synchronize and exchange per-token LoRA indices alongside hidden states during AlltoAll communication, ensuring alignment across TP ranks.
- Routing Recovery: Added utility functions to recover LoRA routing metadata specifically for the AlltoAll dispatch path, enabling correct LoRA delta application.
- E2E Test Coverage: Added a new end-to-end test case, test_qwen3moe_lora_ep, to verify LoRA + EP functionality.

Currently, LoRA for RL scenarios has not yet considered quantization.

No. EP + LoRA was previously broken (crashed); this PR makes it work. The existing TP-only LoRA path is unaffected. 

- vLLM version: v0.25.1


vllm_ascend/ops/fused_moe/fused_moe_0_23_0.py
```diff
@@ -423,6 +425,23 @@ def forward_impl(  # type: ignore[override]
...
+        # only the lora_ids for the tokens it owns.
+        split_lora_indices = None
+        if self.moe_config.ep_size > 1:
+            lora_context = getattr(self, "_ascend_moe_lora_context", None)
+            if lora_context is not None:
+                token_lora_indices = lora_context.punica_wrapper.token_lora_indices
+                tp_size = get_tensor_model_parallel_world_size()
+                tp_rank = get_tensor_model_parallel_rank()
+                num_tokens = hidden_states.shape[0]
+                pad_size = tp_size - num_tokens
```
Tflowers-0129: Could you explain a bit more?

vllm_ascend/ops/fused_moe/token_dispatcher.py
```diff
@@ -523,6 +541,15 @@ def token_dispatch(
...
         )
 
+        # Apply the second LoRA permute (sort by local expert within the rank).
+        # if there are multiple local experts.
+        if (
+            self.num_local_experts > 1
+            and exchanged_lora_pending is not None
+            and reversed_global_input_permutation_mapping is not None
+        ):
+            exchanged_lora_indices = exchanged_lora_pending[reversed_global_input_permutation_mapping]
```
Tflowers-0129: else中没有赋值，不会有问题吗

Tflowers-0129:
The PR title says this only supports non-quantized mode, but I do not see an explicit guard here that rejects quantized MoE + EP + LoRA before inference.

vllm_ascend/ops/fused_moe/token_dispatcher.py
wangx700: if self.num_local_experts <= 1 no need permute for only one expert.

vllm_ascend/ops/fused_moe/fused_moe_0_23_0.py
wangx700: lora split process was added to all2all prepare and split align with hiddenstates for tp.

wangx700:
change to https://github.com/vllm-project/vllm-ascend/pull/12451.

PR #11854, opened 2026-07-11
[BugFix][KV Pool][KV offload] Skip SSD offload registration for scheduler-only Mooncake clients

bowgneo:
Scheduler-only clients (contribute_memory=False) do not contribute KV cache memory to the pool. However, the current code unconditionally passes enable_ssd_offload to MooncakeDistributedStore.setup(), which causes Mooncake to register an extra active client on the master for each scheduler rank. This inflates both the perceived client count and the reported SSD storage usage.

This PR ensures that when _contribute_memory is False, SSD offload parameters are stripped before calling setup(), so schedulers are not counted as SSD-enabled clients.

No. The change only affects internal Mooncake store initialization.

CI with new/existing tests.

- vLLM version: v0.23.0


yiz-liu (approved):

bowgneo:
/cherry-pick main
[Bot]: cherry-pick completed successfully. New PR created: https://github.com/vllm-project/vllm-ascend/pull/12221

PR #11883, opened 2026-07-12
[BugFix] [MLAPO] Keep worker alive instead of crashing when kv_consumer must local-prefill (fixes #11882)

### LINKED ISSUE #11882: [P/D] Mis-parametrized request crashes decode worker (UndefinedTensorImpl) when MLAPO freed prefill weights on kv_consumer
## Summary

In PD disaggregation, a single request with incorrect `kv_transfer_params` crashes the **entire decode (D) worker process** with `NotImplementedError: Cannot access storage of UndefinedTensorImpl`, taking down all in-flight requests on that D instance. The engine should fail the single bad request gracefully instead of crashing the worker.

## Environment

- vllm-ascend `main` (verified at `2b41938d`)
- PD disaggregated prefill, mooncake KV transfer (`kv_p2p/mooncake_connector.py`)
- D nodes configured as `kv_consumer`
- `VLLM_ASCEND_ENABLE_MLAPO=1` (default), `max_num_batched_tokens <= 1024`
- Model: MLA-based (Kimi K2.6 / kimi_k25), W8A8

## Trigger

The D worker receives a request whose `kv_transfer_params` are the **P-side (pusher) parameters** instead of the D-side (puller) parameters:

```json
{ "do_remote_decode": true, "do_remote_prefill": false }   // ← P's params, forwarded to D
```

D should receive:
```json
{ "do_remote_decode": false, "do_remote_prefill": true, "remote_block_ids": [...], "remote_host": "...", ... }
```

This mis-parameterization happens whenever the prefill step on P does **not** finish as `FINISHED_LENGTH_CAPPED` (e.g. the proxy failed to cap prefill to 1 token — such as the known `max_completion_tokens`-overrides-`max_tokens=1` proxy bug — and P ran full generation, stopping on EOS/tool-call). In that case P's `request_finished` returns `None`, the params are **not flipped** to the D-side form, and the proxy forwards the un-flipped P-side params to D.

## Root cause chain (with code refs)

1. P only pushes KV and returns D-side params when prefill finishes `LENGTH_CAPPED` — [`mooncake_connector.py:1660-1693`](https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py#L1660-L1693):
   ```python
   if (params is None
       or not params.get("do_remote_decode")
       or request.status != RequestStatus.FINISHED_LENGTH_CAPPED):
       return False, None          # ← no KV push, params not flipped for D
   return delay_free_blocks, dict(do_remote_prefill=True, do_remote_decode=False, ...)
   ```
2. Proxy forwards the un-flipped `{do_remote_decode:True, do_remote_prefill:False}` to D.
3. D's `get_num_new_matched_tokens` sees `do_remote_prefill=False` → returns `(0, False)` → D does **not** pull KV — [`mooncake_connector.py:1594`](https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py#L1594).
4. D falls back to **local prefill (recompute)** for the request.
5. But on a `kv_consumer` with MLAPO enabled and `max_num_batched_tokens <= MLAPO_MAX_SUPPORTED_TOKENS` (1024), the original prefill weights were **freed at init** — [`mla_v1.py:1074-1086`](https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/attention/mla_v1.py#L1074-L1086):
   ```python
   if (kv_transfer_config is not None
       and kv_transfer_config.is_kv_consumer
       and scheduler_config.max_num_batched_tokens <= MLAPO_MAX_SUPPORTED_TOKENS):
       self.fused_qkv_a_proj.weight = None
       self.fused_qkv_a_proj.deq_scale = None
       self.fused_qkv_a_proj.quant_bias = None
       self.q_proj.weight = None
       self.q_proj.deq_scale = None
       self.q_proj.quant_bias = None
   ```
6. D's local recompute forward needs the original `fused_qkv_a_proj` weights → they are `None` → `UndefinedTensor` → worker crash:
   ```
   File "vllm_ascend/attention/mla_v1.py", line 1657, in _mla_preprocess
     qkv_lora = self.fused_qkv_a_proj(hidden_states)[0]
   NotImplementedError: Cannot access storage of UndefinedTensorImpl
   ```

## Why `kv_load_failure_policy` does not catch it

`recompute_kv_load_failures` / `_handle_invalid_blocks` only fires when the worker reports `invalid_block_ids` from `get_block_ids_with_load_errors`, which is populated **only after a pull was attempted and failed** ([`mooncake_connector.py:2130`](https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py#L2130)). Here D never attempts a pull (`do_remote_prefill=False`), so no load error is recorded and the graceful-failure path never triggers.

## Crash log

D-side log shows the crashing request has **no** `KV cache transfer for request ...` line (every healthy request has 8, one per TP rank), immediately followed by:
```
(Worker_DP2_TP3_EP19 pid=283) ERROR ... NotImplementedError: Cannot access storage of UndefinedTensorImpl
  File "vllm_ascend/attention/mla_v1.py", line 1657, in _mla_preprocess
    qkv_lora = self.fused_qkv_a_proj(hidden_states)[0]
```
i.e. D never pulled KV for that request, then crashed in local recompute.

## Impact

A single malformed/mis-routed request crashes the whole D worker, affecting every in-flight and subsequent request on that instance until restart. The blast radius is far larger than the offending request.

### Option A (structural, vllm-ascend-only, simplest) — recommended as immediate fix
Do **not** free the original prefill weights on `kv_consumer` (guard [`mla_v1.py:1074-1086`](https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/attention/mla_v1.py#L1074-L1086)). Then D's local recompute has the weights it needs → no `UndefinedTensor` → no worker crash. Trade-off: D keeps an extra copy of `fused_qkv_a_proj`/`q_proj` weights (memory cost), but recompute/fallback always works.

### Option B (graceful per-request failure — ideal, needs vllm-core cooperation)
Detect the role contract violation on D (`is_kv_consumer` receiving `do_remote_decode=True`) and route the single request through `kv_load_failure_policy` (fail/recompute) instead of crashing. This requires a vllm-core connector API extension: the scheduler-side connector currently has no way to fail a single request (it cannot call `finish_requests`, and raising in `update_state_after_alloc` fatally crashes the engine core via `_send_engine_dead`). A small vllm-core API (e.g. a connector-reported "unservable request" set that the scheduler finishes with `FINISHED_ERROR`) would enable this.

## Workaround

Set `VLLM_ASCEND_ENABLE_MLAPO=0` on D nodes. This skips `_process_weights_for_fused_mlapo`, so the original prefill weights are not freed → D's local recompute works → no crash. Cost: MLA decode performance regression on D.

## Note

The proximate trigger is a proxy bug (prefill not capped to 1 token → P does not push KV). The upstream `load_balance_proxy_server_example.py` already mitigates the `max_completion_tokens`-overrides-`max_tokens=1` case. But regardless of proxy correctness, the engine should not let a single bad request crash the entire D worker — hence this issue focuses on engine-side robustness.

xlshaoscu:
## What & Why

Fixes #11882.

On `kv_consumer` (decode-only D) with MLAPO enabled and `max_num_batched_tokens <= MLAPO_MAX_SUPPORTED_TOKENS` (1024), `_process_weights_for_fused_mlapo` freed the original `fused_qkv_a_proj`/`q_proj` weights (set to `None`) to save NPU memory.

This breaks **local prefill on D**: vLLM has several normal paths where D must local-prefill (recompute on KV load failure, P2P session fallback, preempt resume), and a mis-parametrized request (e.g. proxy forwards un-flipped `kv_transfer_params` without valid remote KV) also forces D to local-prefill. In all these cases `_mla_preprocess` dereferences the freed weights and the **entire D worker crashes** with `Cannot access storage of UndefinedTensorImpl` — taking down all in-flight requests on that instance.

## Fix

Keep the NPU memory saving during normal decode, but make the weights **recoverable**:

1. In `_process_weights_for_fused_mlapo`, **offload** the 6 tensors (`fused_qkv_a_proj.{weight,deq_scale,quant_bias}` + `q_proj.{weight,deq_scale,quant_bias}`) **to CPU** (instead of freeing to `None`), and set `self._prefill_weights_offloaded = True`. NPU memory is still saved during normal decode.
2. **Lazily restore** them to NPU (synchronous `.to(device)`) at the top of `forward()`'s else branch — the only place they are referenced — on the first step where D must local-prefill. Restored **once per layer**; afterwards the flag is cleared and the weights stay on NPU. Subsequent steps pay no copy cost.

This converts a worker-fatal crash into a successful (if slow) local recompute.

## Properties

- **Normal decode**: unchanged. The MLAPO fast path never touches these weights; they sit on CPU. Zero perf impact, NPU memory saved.
- **First local-prefill step** (recompute / fallback / preempt / mis-parametrized): one-time synchronous CPU→NPU copy of the 6 tensors per layer, then they stay on NPU. The prefill completes normally.
- **Memory**: trades the previous "free to None" (0 memory, but crashes on any prefill) for "offload to CPU" (1 CPU copy of `fused_qkv_a_proj`+`q_proj` per layer, reclaimable NPU memory). After the first recompute, NPU memory converges to the "keep weights" state.
- **No vllm-core changes** — fully contained in vllm-ascend.

## Why this approach over alternatives

- **Free to None (original code)**: simplest, saves the most NPU memory, but crashes the worker on any prefill path. This is the bug being fixed.
- **Keep weights on NPU (don't free)**: simplest and safest, but pays the NPU memory cost unconditionally. This PR preserves the NPU saving during normal operation and only pays it after the first recompute (rare).
- **Zero-fill output on guard**: keeps the worker alive but produces empty/garbage output for the affected request (and the rest of the batch in that step). The offload-and-restore approach produces correct results instead.

## Testing

- `python -c "import ast; ast.parse(open('vllm_ascend/attention/mla_v1.py').read())"` passes.
- **Validated on real Ascend NPU hardware (Atlas 800T A2 / 910B3).** In a PD-disaggregated deployment (kv_consumer + MLAPO), the original crash (`UndefinedTensorImpl` in `_mla_preprocess` dereferencing freed `fused_qkv_a_proj.weight`) was reproduced and confirmed fixed: with this patch the D worker survives, the prefill weights are restored from CPU on the first local-prefill step, and the request completes normally instead of crashing the worker.

## Related

- Issue: #11882
- The proximate trigger is often a proxy bug (prefill not capped to 1 token → P does not push KV). The upstream `examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py` already mitigates the `max_completion_tokens`-overrides-`max_tokens=1` case. This PR is the engine-side robustness fix so that a single bad request does not crash the whole D worker regardless of proxy correctness.

- vLLM version: v0.26.0

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -1051,19 +1051,24 @@ class AscendMLAImpl(MLAAttentionImpl):
         self.q_nope_scale = torch.tensor([1], dtype=act_dtype, device=device)
 
         # On KV consumers (decode-only) MLAPO uses the transformed weights built above;
-        # the original fused_qkv_a_proj/q_proj weights and quant params are no longer
-        # referenced, so drop them to save memory.
+        # the original fused_qkv_a_proj/q_proj weights and quant params are not
+        # referenced by the MLAPO decode fast path, so offload them to CPU to save
+        # NPU memory. They are lazily restored to NPU in forward() if D must
+        # local-prefill (recompute / fallback / preempt, see issue #11882).
         if (
             self.vllm_config.kv_transfer_config is not None
             and self.vllm_config.kv_transfer_config.is_kv_consumer
             and self.vllm_config.scheduler_config.max_num_batched_tokens <= MLAPO_MAX_SUPPORTED_TOKENS
         ):
-            self.fused_qkv_a_proj.weight = None  # type: ignore[union-attr]
-            self.fused_qkv_a_proj.deq_scale = None  # type: ignore[union-attr]
-            self.fused_qkv_a_proj.quant_bias = None  # type: ignore[union-attr]
-            self.q_proj.weight = None
-            self.q_proj.deq_scale = None
-            self.q_proj.quant_bias = None
+            cpu_device = torch.device("cpu")
+            for module in (self.fused_qkv_a_proj, self.q_proj):
+                if module is None:
+                    continue
+                for attr in ("weight", "deq_scale", "quant_bias"):
+                    val = getattr(module, attr, None)
+                    if val is not None:
+                        val.data = val.data.to(cpu_device, non_blocking=True)
+            self._prefill_weights_offloaded = True
             torch.npu.empty_cache()
 
     def _process_weights_for_fused_mlapo_a5(self, act_dtype: torch.dtype):
@@ -1734,6 +1739,30 @@ class AscendMLAImpl(MLAAttentionImpl):
                 self, hidden_states, kv_cache, attn_metadata
             )
         else:
+            # On a kv_consumer with MLAPO, the prefill weights were offloaded to CPU
+            # to save NPU memory. This branch is reached when D must local-prefill
+            # (recompute / fallback / preempt, or miss-parametrized kv_transfer_params
+            # without valid remote KV, see #11882). Lazily restore them to NPU once
+            # per layer on the first such step, then clear the flag so subsequent
+            # steps pay no copy cost.
+            if getattr(self, "_prefill_weights_offloaded", False):
+                logger.warning(
+                    "Layer %s: restoring MLAPO prefill weights from CPU to NPU for "
+                    "local prefill on kv_consumer (see issue #11882).",
+                    layer_name,
+                )
+                device = hidden_states.device
+                for module in (self.fused_qkv_a_proj, self.q_proj):
+                    if module is None:
+                        continue
+                    for attr in ("weight", "deq_scale", "quant_bias"):
+                        val = getattr(module, attr, None)
+                        if val is not None and val.device.type == "cpu":
+                            # Synchronous copy: _mla_preprocess dereferences these
+                            # weights immediately after, so the transfer must be
+                            # complete before we proceed.
+                            val.data = val.data.to(device)
+                self._prefill_weights_offloaded = False
             decode_preprocess_res, prefill_preprocess_res = self._mla_preprocess(
                 layer_name, hidden_states, kv_cache, attn_metadata, need_gather_q_kv
             )


xlshaoscu:
## Real-world reproduction confirmation

We hit exactly this crash in production. Adding the on-site details to help reviewers prioritize and verify.

### Environment
- PD disaggregated prefill, mooncake `kv_p2p` connector
- D nodes = `kv_consumer`, `VLLM_ASCEND_ENABLE_MLAPO=1` (default), `max_num_batched_tokens <= 1024`
- Model: MLA-based (Kimi K2.6 / `kimi_k25`), W8A8
- A custom PD load-balance proxy in front (adapted from `examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py`)

### Trigger (content-dependent — only specific messages)
The offending client request carried `max_completion_tokens` (not `max_tokens`). In vllm `max_completion_tokens` takes precedence over `max_tokens`, so the proxy's `max_tokens=1` prefill cap was silently overridden → P ran full generation and finished `FINISHED_STOPPED` (not `FINISHED_LENGTH_CAPPED`) → P did not push KV and did not flip `kv_transfer_params` → the proxy forwarded the un-flipped P-side params `{do_remote_decode:True, do_remote_prefill:False}` to D.

Only messages that cause P to not length-cap trigger it; normal messages (where P length-caps at 1 token) work fine. This matches "only specific messages crash".

### Crash signature (D-side log)
The crashing request has **no** `KV cache transfer for request ...` line (every healthy request has 8, one per TP rank), immediately followed by:
```
(Worker_DP2_TP3_EP19 pid=283) ERROR ... NotImplementedError: Cannot access storage of UndefinedTensorImpl
  File "vllm_ascend/attention/mla_v1.py", line 1657, in _mla_preprocess
    qkv_lora = self.fused_qkv_a_proj(hidden_states)[0]
```
i.e. D never pulled KV for that request, fell back to local prefill, and dereferenced the MLAPO-freed weights → worker process crash, taking down all in-flight requests on that D instance.

### Workaround / fix validation
- **`VLLM_ASCEND_ENABLE_MLAPO=0` on D** stops the crash (weights are not freed → D's local recompute works). Confirms the crash is exactly the MLAPO weight-freeing + recompute path described in #11882.
- **Proxy-side fix** (pop `max_completion_tokens` + `ignore_eos` + validate `do_remote_prefill` before forwarding to D) prevents the bad request from reaching D at source.

This PR is the engine-side robustness fix so that even if a bad request slips through any proxy, the D worker no longer crashes — it lazily restores the offloaded weights and recomputes.

### Note on testing
I do not have Ascend NPU access in my current environment, so I could only verify syntax (`ast.parse` passes). The restore path (`_mla_preprocess` → `.to(device)`) is straightforward, but I'd appreciate a maintainer running the PD + kv_consumer + MLAPO + forced-recompute path on real NPU to confirm end-to-end. If there's a preferred test script, point me at it and I can help extend it.

xlshaoscu:
Thanks @gemini-code-assist for the review — both concerns were valid. I've updated the PR (force-pushed `f265c4b2`):

1. **`TypeError` on assigning a plain `Tensor` to an `nn.Parameter` slot** — confirmed `weight`/`deq_scale` are `nn.Parameter` (`w8a8_static.py:161`). Fixed by rebinding `.data` in place (`_t.data = _t.data.cpu()` / `.to(device)`) for `Parameter` attributes, which keeps the `Parameter` object and avoids the slot-assignment `TypeError`.

2. **`AttributeError` when `quant_bias` is `None`** — confirmed `quant_bias` is `None` on non-zero tp ranks (`w8a8_static.py:123`). Fixed with an explicit `if _t is None: continue` guard.

Both the offload and restore paths now use a unified loop over `(fused_qkv_a_proj, q_proj) × (weight, deq_scale, quant_bias)` that:
- skips `None` (e.g. `quant_bias` on non-zero tp rank),
- uses `.data =` for `nn.Parameter` (rebind underlying tensor),
- uses `setattr(layer, name, t.cpu())` for plain `Tensor` attributes.

Also verified no code path checks `<attr> is None` on these (only the module-level `fused_qkv_a_proj is not None`), so keeping the `Parameter` non-None (with CPU data) does not change branching.

Still not validated on real Ascend NPU (no NPU in my environment) — syntax check passes. A maintainer run of the PD + kv_consumer + MLAPO + forced-recompute path would confirm end-to-end.

xlshaoscu:
Hi @weijiqian0 @whx-sjtu, friendly ping for review. 🙏

This PR fixes #11882 (kv_consumer D-worker crash with MLAPO when local prefill/recompute is needed). The earlier CI blockers are all resolved on the latest push:

- PR title prefix → [BugFix] ✅
- typos lint (comment wording) ✅
- DCO sign-off ✅

All required CI checks are now green (lint-and-select-tests, DCO, main). Branch rebased onto latest main (mergeable: clean).

Note: the fix is not yet validated on real Ascend NPU on my side. It'd help if you can run the PD-disagg path (kv_consumer + MLAPO + a request that triggers local recompute) to confirm the restore path end-to-end. Happy to address any feedback. Thanks!

vllm_ascend/attention/mla_v1.py
```diff
@@ -1734,6 +1739,30 @@ def forward(
...
+                for module in (self.fused_qkv_a_proj, self.q_proj):
+                    if module is None:
+                        continue
+                    for attr in ("weight", "deq_scale", "quant_bias"):
+                        val = getattr(module, attr, None)
+                        if val is not None and val.device.type == "cpu":
+                            # Synchronous copy: _mla_preprocess dereferences these
+                            # weights immediately after, so the transfer must be
+                            # complete before we proceed.
+                            val.data = val.data.to(device)
```
linfeng-yuan, line 1764: This implementation is not acceptable. Since it would reduce free memory (~3GB in pure DP scenarios) and lead to OOM when executing P requests.

linfeng-yuan (changes requested):

vllm_ascend/attention/mla_v1.py
linfeng-yuan: How about add a configuration to keep the q_proj weights? It can keep the accuracy and robustness with less kvcache which should be determined by our users?

PR #11911, opened 2026-07-13
[Doc][Misc] Remove invalid fuse_muls_add from additional-config

kunpengW-code:
The fuse_muls_add parameter only takes effect inside ascend_compilation_config, not at the top level of additional-config. All examples and test configs that placed it at the top level were ineffective. Since the default value is already True, these invalid entries are removed rather than moved, keeping configs clean without changing behavior.
no

- vLLM version: v0.23.0


weiguihua2 (approved):

linfeng-yuan (approved):

kunpengW-code:
/cherry-pick releases/v0.23.0
[Bot]: cherry-pick command failed. Please check the [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/29246212813) for details.

PR #11916, opened 2026-07-13
[Feature] Add SwigluOAI activation variant to dispatch_ffn_combine

iceflysnow:
# feat(mc2): add SwigluOAI activation variant to `dispatch_ffn_combine`

## Summary

This contributes the **SwigluOAI activation variant** of the fused
`dispatch_ffn_combine` (grouped-matmul + SwiGLU + per-token clamp)
operator on Ascend, so that models using an OAI-style SwiGLU — such as
**MiniMax-M3** — can run on a single fused kernel instead of an unfused
grouped-matmul + separate-activation path, recovering the quant-fusion
benefit.

The change is **purely additive at the operator level** and does not
register or depend on any specific model.

## Background

MiniMax-M3's MLP uses an OAI-style SwiGLU:

| path | formula | `up` clamp |
|---|---|---|
| standard SwiGLU (`alpha == 0`) | `out = silu(gate) * up` | single-sided `ClampMin(-limit)` |
| **SwigluOAI** (`alpha > 0`) | `out = (up + 1) * gate * sigmoid(gate * alpha)` | double-sided `ClampMin(-limit) + ClampMax(limit)` |

Because the currently fused `dispatch_ffn_combine` op only implements
the standard path, models using SwigluOAI had to fall back to an unfused
grouped-matmul + separate swiglu kernel, which is slower and loses the
quant-fusion benefit on Ascend.

## What this PR does

Adds an optional `swigluAlpha` attribute (default `0.0`) to the
`dispatch_ffn_combine` op:

- `alpha > 0` → kernel takes the SwigluOAI branch (new math + double-sided
  `up` clamp).
- `alpha == 0` → **bit-identical** to the existing standard SwiGLU path,
  so all current callers are unaffected.

The Python side:

- `vllm_ascend/ops/activation.py` — adds an inline native reference
  implementation of SwigluOAI (`(up+1)*gate*sigmoid(gate*alpha)`).
- `vllm_ascend/ops/fused_moe/*` + `quantization/methods/w8a8_dynamic.py`
  — threads `swiglu_alpha` (default `0.0`) from the layer config to the
  operator, and routes layers whose activation is `"swigluoai"` through
  this fused op instead of the generic gmm+swiglu fusion. Layers using
  other activations are unchanged.
- `vllm_ascend/ops/fused_moe/fused_moe.py` — for `"swigluoai"` layers the
  flashcommon3 gate-stream overlap (`multistream_overlap_gate`) is
  disabled, because the fused single-kernel `dispatch_ffn_combine` path
  is incompatible with the separate gate-stream overlap mechanism.
  Non-OAI layers keep the original `ascend_config.multistream_overlap_gate`
  behavior, so they are unaffected.

## Files (18, no new files, no binaries)

**C++ op side** (`csrc/mc2/dispatch_ffn_combine/`)
- `op_host/dispatch_ffn_combine_def.cpp` — register `swigluAlpha` attr.
- `op_host/dispatch_ffn_combine_tiling.cpp`, `op_kernel/dispatch_ffn_combine_tiling.h` — read attr into tiling info.
- `op_host/op_api/aclnn_dispatch_ffn_combine.{cpp,h}` — expose alpha in op_api.
- `dispatch_ffn_combine_torch_adpt.h` — alpha param on the torch adapter.
- `op_kernel/dispatch_ffn_combine.h`, `op_kernel/dispatch_ffn_combine_kernel.hpp` — propagate alpha to the epilogue.
- `op_kernel/utils/block_epilogue_pertoken_swiglu.hpp` — **the kernel math** (OAI branch + double-sided clamp; standard branch preserved).
- `csrc/torch_binding.cpp`, `csrc/torch_binding_meta.cpp` — python binding + schema.

**Python integration** (`vllm_ascend/`)
- `ops/activation.py` — native reference impl.
- `ops/fused_moe/{fused_moe,moe_mlp,moe_comm_method,moe_runtime_args,moe_stage_contracts}.py` — thread `swiglu_alpha`.
- `quantization/methods/w8a8_dynamic.py` — pass `swiglu_alpha` to the op.

## Out of scope (intentionally not in this PR)

This PR is scoped to the **operator contribution** only. The following
are model-specific and are intentionally excluded (they can land as a
separate PR that depends on this one):

- MiniMax-M3 model registration (`models/__init__.py`,
  `ascend_forward_context.py`) and the `minimax_m3_vl.py` model module.
- `patch/platform/patch_minimax_m3.py` and its import.
- `utils.is_minimax_m3_model()` and the M3-specific slot-mapping /
  cudagraph-skip logic in `worker/model_runner_v1.py`.
- M3 quant packed-modules mapping in `modelslim_config.py`.
- M3 triton kernels (`ops/triton/*`) and reasoning / tool parsers.

Excluding these keeps this PR self-contained: it applies cleanly to
`vllm-ascend@5f6faa0c` and `import vllm_ascend` still works without any
out-of-tree files.

## Compatibility

- `alpha == 0` is the default and is bit-compatible with the pre-PR
  kernel (the standard SwiGLU branch keeps the same ops, same order, same
  clamp sides).
- Only layers that opt in (`swiglu_alpha > 0`) take the new path;
  existing models see no change.

## Tested

- Base commit: `5f6faa0c` (vllm-ascend v0.22.1rc1).
- Built and ran on Ascend A3 910C + CANN 9.0.0, Python 3.12.
- Verified the fused SwigluOAI op produces identical outputs to the
  reference native implementation `(up+1)*gate*sigmoid(gate*alpha)`
  within fp tolerance.
- Serving benchmark (MiniMax-M3 w8a8, A3 910C ×8 (16 NPU), TP=4 DP=4, random
  16K-in / 1024-out, CON32, 32 prompts), fused `dispatch_ffn_combine`
  path (with `swiglu_alpha`) vs unfused gmm + separate SwigluOAI:

  | metric | fused | unfused | delta |
  |---|---|---|---|
  | output throughput (tok/s) | 82.10 | 60.42 | **+36%** |
  | total token throughput (tok/s) | 1409.74 | 1037.53 | +36% |
  | mean TPOT (ms) | 174.55 | 277.84 | **−37%** |
  | mean TTFT (ms) | 143571 | 255402 | −44% |
  | mean E2EL (ms) | 322135 | 539629 | −40% |

  The fused op collapses dispatch + GMM1 + activation + quant + GMM2 +
  combine into one kernel, eliminating the unfused path's repeated
  comms and intermediate buffer writes.

## Notes for reviewers

- This touches the custom MC2 `dispatch_ffn_combine` operator in
  `csrc/mc2/`. The operator is vendored into vllm-ascend's csrc, so the
  host / tiling / op_api / kernel changes all live in this repo (no
  CANN-side prerequisite).
- The change is additive: a new optional attribute + a new branch guarded
  by `alpha > 0`; the existing path is untouched and bit-compatible.

### DIFF
--- a/csrc/mc2/dispatch_ffn_combine/dispatch_ffn_combine_torch_adpt.h
+++ b/csrc/mc2/dispatch_ffn_combine/dispatch_ffn_combine_torch_adpt.h
@@ -32,11 +32,14 @@ std::tuple<at::Tensor&, at::Tensor&> dispatch_ffn_combine(
     at::Tensor& out,
     at::Tensor& expert_token_nums,
     const c10::optional<at::Tensor>& x_active_mask,
-    double swiglu_limit
+    double swiglu_limit,
+    double swiglu_alpha = 0.0
 ) {
     char *group_ep_ptr = const_cast<char *>(group.data());
     bool is_int8 = weight1[0].dtype() == at::kChar;
     bool is_int4 = weight1[0].dtype() == at::kInt;
+    TORCH_CHECK(is_int8 || swiglu_alpha == 0.0,
+        "swiglu_alpha is only supported for int8 (W8A8) quantization path.");
     if (is_int8) {
         EXEC_NPU_CMD(aclnnDispatchFFNCombine,
                  x,
@@ -50,6 +53,7 @@ std::tuple<at::Tensor&, at::Tensor&> dispatch_ffn_combine(
                  group_ep_ptr,
                  max_output_size,
                  swiglu_limit,
+                 swiglu_alpha,
                  out,
                  expert_token_nums);
     } else if (is_int4){
--- a/csrc/mc2/dispatch_ffn_combine/op_host/dispatch_ffn_combine_def.cpp
+++ b/csrc/mc2/dispatch_ffn_combine/op_host/dispatch_ffn_combine_def.cpp
@@ -78,6 +78,8 @@ class DispatchFFNCombine : public OpDef {
     this->Attr("transB").AttrType(OPTIONAL).Bool(false);
     this->Attr("weightNz").AttrType(OPTIONAL).Bool(false);
     this->Attr("swigluLimit").AttrType(OPTIONAL).Float(0.0f);
+    // SwigluOAI alpha: >0 uses (up+1)*gate*sigmoid(gate*alpha); ==0 falls back to standard swiglu (M2.7 compatible)
+    this->Attr("swigluAlpha").AttrType(OPTIONAL).Float(0.0f);
 
     OpAICoreConfig aicore_config;
     aicore_config.DynamicCompileStaticFlag(true)
--- a/csrc/mc2/dispatch_ffn_combine/op_host/dispatch_ffn_combine_tiling.cpp
+++ b/csrc/mc2/dispatch_ffn_combine/op_host/dispatch_ffn_combine_tiling.cpp
@@ -36,6 +36,7 @@ namespace {
     constexpr uint32_t ATTR_IS_TRANS_B = 2;
     constexpr uint32_t ATTR_WEIGHT_NZ = 3;
     constexpr uint32_t ATTR_SWIGLU_LIMIT = 4;
+    constexpr uint32_t ATTR_SWIGLU_ALPHA = 5;   // SwigluOAI alpha
     constexpr uint64_t INIT_TILINGKEY = 1000000;
     constexpr uint64_t TILINGKEY_TRANS_B = 1U;
     constexpr uint64_t TILINGKEY_WEIGHT_NZ = 10;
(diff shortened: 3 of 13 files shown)


iceflysnow:
@zzzzwwjj @CXY-Katrina friendly ping for review. This PR has been rebased onto the latest `main` (zero conflicts, MERGEABLE, CI green).

## Updated value proposition

With MiniMax-M3 support landed via #12448 on `releases/v0.24.0rc`, M3's OAI SwiGLU currently runs through `torch_npu.npu_clipped_swiglu` as a **split** path (grouped-matmul -> npu_clipped_swiglu -> ...).

This PR complements that by enabling the **fully-fused** `dispatch_ffn_combine` path for OAI SwiGLU:
- Adds a `swigluAlpha` attribute to the `dispatch_ffn_combine` C++ kernel; `alpha > 0` selects the OAI branch (`out = (up + 1) * gate * sigmoid(gate * alpha)` with double-sided clamp on `up`), `alpha == 0` is bit-identical to the existing standard SwiGLU path (no impact on M2/M2.5 callers).
- Lets M3 use a single fused kernel (grouped-matmul + SwiGLU + per-token clamp + ALLTOALL) in EP<=32 + W8A8 + `enable_fused_mc2` scenarios, reducing kernel launch overhead and intermediate GM traffic vs the split path.

Currently `dispatch_ffn_combine` only supports standard SwiGLU (passes `swiglu_limit` only), so M3 cannot benefit from the fused_mc2 path for its OAI activation -- this PR fills that gap. Purely additive at the operator level; does not register or depend on any specific model.

Would appreciate your review.

PR #11918, opened 2026-07-13

singzhou:

No.

Local machine tested.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/_310p/spec_decode/llm_base_proposer_310.py
+++ b/vllm_ascend/_310p/spec_decode/llm_base_proposer_310.py
@@ -20,7 +20,7 @@ from typing import Any
 import torch
 from vllm.v1.attention.backends.utils import CommonAttentionMetadata
 
-from vllm_ascend._310p.ops.rotary_embedding import set_rope_position_flag_310p
+from vllm_ascend._310p.ops.rotary_embedding import AscendRotaryEmbedding310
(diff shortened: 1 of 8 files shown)


vllm_ascend/_310p/spec_decode/llm_base_proposer_310.py
```diff
@@ -20,12 +20,43 @@
...
+        num_input_tokens,
+        batch_size,
+        token_indices_to_sample,
+        target_positions,
+        inputs_embeds,
+        multi_steps_attn_metadata,
+        num_tokens,
+        is_prefill=None,
+    ) -> torch.Tensor:
+        set_rope_position_flag_310p(True)
```
singzhou: The reason why adding update_cos_sin is that:

Non-310P device: Tolerates skipping update_cos_sin by passing 'position' to the position embedding operator (torch_npu._npu_rotary_embedding or triton impl), the cache slice is actually wrong here.

310P device: Strictly requires the correct position embedding cache slice (required by torch_npu.npu_apply_rotary_pos_emb).

Tflowers-0129:
\rerun

Tflowers-0129 (approved):

singzhou:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

weiguihua2 (approved):

linfeng-yuan (approved):

PR #11920, opened 2026-07-13

singzhou:
Backports #11918 to releases/v0.23.0

No.

Local machine tested.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/_310p/ops/rotary_embedding.py
+++ b/vllm_ascend/_310p/ops/rotary_embedding.py
@@ -25,11 +25,18 @@ from vllm.model_executor.layers.rotary_embedding import MRotaryEmbedding
 from vllm.model_executor.layers.rotary_embedding.common import ApplyRotaryEmb
 from vllm.model_executor.layers.rotary_embedding.mrope import apply_interleaved_rope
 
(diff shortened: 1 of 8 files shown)


Tflowers-0129 (approved):

vllm_ascend/_310p/ops/rotary_embedding.py
```diff
@@ -25,12 +25,19 @@
...
 from vllm.model_executor.layers.rotary_embedding.mrope import apply_interleaved_rope
 
-from vllm_ascend.ops.rotary_embedding import AscendRotaryEmbedding, get_cos_and_sin_slice
+from vllm_ascend.ops.rotary_embedding import AscendRotaryEmbedding, get_cos_and_sin_slice, update_cos_sin
 
 # Filled once per model forward in NPUModelRunner310._model_forward; read by every MRoPE layer.
 _mrope_cos_slice: torch.Tensor | None = None
 _mrope_sin_slice: torch.Tensor | None = None
 
+_ROPE_310P_UPDATE_FLAG = False
```
weiguihua2: Do not add all variables; implement it in a different way.

weiguihua2 (approved):

linfeng-yuan (approved):

PR #11937, opened 2026-07-13
[Feat][Worker] Support NZ static buffers for prefetch offload

KINGFIOX:

This PR supports Ascend NZ-format static buffers for prefetch CPU offload.

- Marks parameters whose quantized weights are transformed to `ACL_FORMAT_FRACTAL_NZ` during `process_weights_after_loading`, after `maybe_trans_nz` has run.
- Casts the corresponding prefetch `StaticBufferPool` buffers to `ACL_FORMAT_FRACTAL_NZ` during `post_init`, so offloaded weights are copied back into buffers with the expected Ascend storage format.
- Refactors the Ascend prefetch offloader to reuse upstream vLLM `PrefetchOffloader` behavior while installing Ascend-specific module offloaders and keeping CUDA stream/event aliases mapped to NPU APIs after model runner initialization.

In my setup, this makes it possible to run prefill for GLM-5.1-W8A8 with around 200K context on a single-node 16-card machine, although it is still very slow. I use prefetch CPU offload for selected MLP weights with:

```bash
--offload-backend prefetch
--offload-group-size 30
--offload-num-in-group 1
--offload-prefetch-step 1
--offload-params mlp
```

### Previous works

[vllm-project/vllm-ascend#10251](https://github.com/vllm-project/vllm-ascend/pull/10251) added the initial Ascend prefetch offloader by porting the upstream vLLM implementation and replacing CUDA stream/event APIs with NPU equivalents. However, that implementation still reused the normal-layout `StaticBufferPool` flow and did not preserve Ascend NZ weight format for static buffers.

This matters because vLLM Ascend has enabled NZ conversion for quantized weights by default since [#4878](https://github.com/vllm-project/vllm-ascend/pull/4878), included from `v0.13.0rc1`: `VLLM_ASCEND_ENABLE_NZ=1` means quantized weights are transformed to NZ by `maybe_trans_nz` during `process_weights_after_loading`. When prefetch offload copies those NZ weights through normal-layout static buffers, the offloaded weights may be restored with the wrong layout. I tested the #10251 behavior on Ascend A3 and reproduced accuracy issues from this mismatch.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/model_executor/test_offloader.py
+++ b/tests/ut/model_executor/test_offloader.py
@@ -1,0 +1,37 @@
+from types import SimpleNamespace
+
+from vllm.model_executor.offloader.base import NoopOffloader
+
+from vllm_ascend.model_executor.offloader.base import create_offloader
+from vllm_ascend.model_executor.offloader.prefetch import _is_using_nz_weight
+
+
+def test_create_offloader_without_config_returns_noop():
+    offloader = create_offloader(None)
+
+    assert isinstance(offloader, NoopOffloader)
+
+
+def test_create_offloader_for_non_prefetch_backend_returns_noop():
+    offload_config = SimpleNamespace(
+        offload_backend=None,
+        prefetch=None,
+    )
+
+    offloader = create_offloader(offload_config)
+
+    assert isinstance(offloader, NoopOffloader)
+
+
+def test_is_using_nz_weight_handles_invalid_npu_format(monkeypatch):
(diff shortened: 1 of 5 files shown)


KINGFIOX:
I am sorry that I am a freshhand in colaborating with Github, so I chose to use codex to help me out; however, the job above is done by myself. 让各位看笑话了

PR #11940, opened 2026-07-13
[Bugfix][LoRA][Qwen3.5]Fix Qwen3.5 densemodel LoRA scenarios with tp, fully_sharded_loras and aclgraph.

### LINKED ISSUE #10897: [Bug] LoRA cannot be used with graph mode under tensor parallelism
## Summary

LoRA currently cannot be used together with graph mode under tensor parallelism. The issue can be reproduced by disabling `enforce_eager` and enabling multi-TP in the Qwen3.5 LoRA e2e test.

## Reproducer

UT file:

`tests/e2e/pull_request/one_card/lora/test_qwen35_densemodel_lora.py`

The problematic case is equivalent to adding/running a graph-mode LoRA test with:

```python
llm = vllm.LLM(
    model=MODEL_PATH,
    max_model_len=4096,
    enable_lora=True,
    max_loras=2,
    max_num_seqs=4,
    max_lora_rank=8,
    enforce_eager=False,
    tensor_parallel_size=1,
    trust_remote_code=True,
)
```
The problematic case is equivalent to adding/running a tp LoRA test with:

```python
llm = vllm.LLM(
    model=MODEL_PATH,
    max_model_len=4096,
    enable_lora=True,
    max_loras=2,
    max_num_seqs=4,
    max_lora_rank=8,
    enforce_eager=True,
    tensor_parallel_size=4,
    trust_remote_code=True,
)
```

Run:

```bash
pytest tests/e2e/pull_request/one_card/lora/test_qwen35_densemodel_lora.py
```

## Expected behavior

LoRA inference should pass in graph mode with tensor parallelism, and the generated outputs should match the existing eager-mode LoRA expectations.

## Actual behavior

When `enforce_eager=False` and `tensor_parallel_size=4`, the LoRA path cannot be stacked with graph mode and the e2e test cannot pass reliably. The eager-mode LoRA path and eager-mode TP LoRA path are used as the baseline.
tp error log：

Image

graph mode log：

Image

## Notes

This indicates the LoRA graph-mode path still needs further design and validation, especially around tensor-parallel LoRA data flow and graph-mode execution. There is currently no explicit external customer performance requirement for LoRA graph mode, so this can be tracked and deferred to a later PoC-focused effort.

paulyu12:
- LoRA shrink kernels (bgmv_shrink.cpp, sgmv_shrink.cpp): CopyOut used
DataCopy, which requires block-aligned lengths. For LoRA ranks/dtypes
whose maxLoRARank_ * sizeof(Y_T) isn't block-aligned, this produced
incorrect/corrupted output. Switched to DataCopyPad with an explicit
byte length so arbitrary rank sizes copy out correctly.

- Cudagraph replay for attention (attention_v1.py): when LoRA
specialization (cudagraph_specialize_lora) is combined with TP,
graph_param_count becomes a multiple of the attention layer count, and
the previous zip-based key/param pairing either dropped
GDN/linear-attn metadata into the FIA loop or left the extra captured
ops unmatched to their layer. Now filters attn_keys to only FIA layers
(via seq_lens_list) and block-wise replicates keys when
graph_param_count % num_layers == 0, so every captured op is replayed
against the correct layer's metadata.

- LoRA dispatch cleanup (vllm_ascend/lora/utils.py): removed the
Ascend-specific *WithLoRA/*WithShardedLoRA QKV/merged-column
subclasses and the patch_mcp_apply_non_fully_sharded() monkeypatch.
The pinned upstream vLLM already resolves out-of-tree linear classes
via isinstance in can_replace_layer and only routes to _mcp_apply's
all-gather when fully_sharded_loras=True; for the non-fully-sharded +
TP>1 case it now goes through the local (no-communication)
shrink/expand path by construction. The Ascend-only classes and the
monkeypatch were redundant workarounds for behavior upstream already
fixed, so they're dead code against the current vLLM pin.

- Test coverage: moved test_qwen35_densemodel_lora.py (one-card only)
to test_qwen35_densemodel_lora_tp.py (two-card), parametrized over
tensor_parallel_size (1/2), enforce_eager (True/False), and
fully_sharded_loras (True/False), so TP + cudagraph +
sharded/non-sharded LoRA combinations are all covered.

No.

pytest -sv tests/e2e/pull_request/two_card/lora/test_qwen35_densemodel_lora_tp.py

- vLLM version: v0.25.1


csrc/kernels/sgmv_shrink.cpp
```diff
@@ -200,7 +200,8 @@ class SGMVShrink {
     __aicore__ inline void CopyOut(const int64_t idx)
     {
         AscendC::LocalTensor<Y_T> yOutLocal = outQueueY_.DeQue<Y_T>();
-        DataCopy(yOutGm_[maxLoRARank_ * idx], yOutLocal, maxLoRARank_);
+        uint16_t blockLen = static_cast<uint16_t>(maxLoRARank_ * sizeof(Y_T));
+        DataCopyPad(yOutGm_[maxLoRARank_ * idx], yOutLocal, {1, blockLen, 0, 0});
```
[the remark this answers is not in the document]
paulyu12: `DataCopyPad` will not introduce the risk, but the old method `DataCopy` will.

csrc/kernels/bgmv_shrink.cpp
```diff
@@ -188,7 +188,8 @@ class BGMVShrink {
     __aicore__ inline void CopyOut(const int64_t idx)
     {
         AscendC::LocalTensor<Y_T> yOutLocal = outQueueY_.DeQue<Y_T>();
-        DataCopy(yOutGm_[maxLoRARank_ * idx], yOutLocal, maxLoRARank_);
+        uint16_t blockLen = static_cast<uint16_t>(maxLoRARank_ * sizeof(Y_T));
+        DataCopyPad(yOutGm_[maxLoRARank_ * idx], yOutLocal, {1, blockLen, 0, 0});
```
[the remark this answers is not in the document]
paulyu12: `DataCopyPad` will not introduce the risk, but the old method `DataCopy` will.

Tflowers-0129, line 192: 这样改动需要确保算子测试用例覆盖，测试用例的lora rank取值需确保存在不对齐情况

vllm_ascend/attention/attention_v1.py
```diff
@@ -695,6 +698,13 @@ def is_direct_target_attn_key(key: str) -> bool:
                 # layer-specific metadata lookup below prevents global/sliding
                 # window layers from accidentally sharing the same metadata.
                 attn_keys = [attn_keys[index % num_layers] for index in range(graph_param_count)]
+            elif graph_param_count > num_layers and num_layers and graph_param_count % num_layers == 0:
```
Tflowers-0129: 如果不能整除的时候，可能存在隐含问题

csrc/kernels/bgmv_shrink.cpp
paulyu12: 当前 e2e 里 `fully_sharded_loras=True` + `tp=2` 组合已经覆盖了不对齐路径，此时“每 shard 4、float32格式为4B → 4×4=16B 不对齐32B“，因此会导致问题。

vllm_ascend/attention/attention_v1.py
paulyu12: 已将整除判断从进入条件改为分支内 assert。每个 LoRA specialization 会 capture 一整套 self-attn FIA op, 所以 graph_param_count 应是 self-attn 层数的整数倍。

Tflowers-0129 (approved):

vllm_ascend/attention/attention_v1.py
```diff
@@ -656,7 +689,10 @@ def update_graph_params(
             else:
                 graph_params = get_graph_params()
                 attn_metadata = forward_context.attn_metadata
-                attn_keys = list(attn_metadata.keys())
+                # Only standard (FIA) attention layers have captured graph
+                # params here; linear/GDN layers (GDNAttentionMetadata) are
+                # updated separately by update_conv1d_graph_params. So we filter by `seq_lens_list`
+                attn_keys = [k for k in attn_metadata if hasattr(attn_metadata[k], "seq_lens_list")]
```
voidvelocity, line 662: > Using `seq_lens_list` as the discriminator is not very robust, since this implementation detail may change or even be removed in the future. It would be more reliable to identify `self_attn` entries in `attn_metadata` using the same regex-based matching already used in `attention_v1.py` (e.g. around lines 679–684), rather than relying on the existence of a specific attribute.
>
> That said, considering the graph parameter logic is currently being refactored, I'm okay with accepting this bug fix as-is. **TODO:** once the graph parameter refactoring is complete, it would be good to revisit this logic and replace the `seq_lens_list` check with a more explicit and stable mechanism.

vllm_ascend/attention/attention_v1.py
```diff
@@ -725,6 +761,22 @@ def is_direct_target_attn_key(key: str) -> bool:
                 # layer-specific metadata lookup below prevents global/sliding
                 # window layers from accidentally sharing the same metadata.
                 attn_keys = [attn_keys[index % num_layers] for index in range(graph_param_count)]
+            elif graph_param_count > num_layers:
```
voidvelocity: > **I have two concerns about this approach:**
>
> 1. **The condition is based on an observed symptom rather than the actual semantic condition.** `graph_param_count > num_layers` does not necessarily imply a LoRA-specialized graph. If future graph capture strategies also produce more captured attention operators than runtime attention metadata, they may unintentionally enter this branch.
>
> 2. **This introduces graph-capture-specific knowledge into `attention_v1`.** The comments and logic now rely on concepts such as LoRA specialization and EAGLE3 draft keys. These seem to belong to the graph capture layer rather than the attention metadata update layer. Embedding these assumptions here increases coupling and could make future maintenance more difficult.
>
> **Overall, this patch looks like a reasonable fix for the current LoRA use case. If there are no existing or planned graph capture strategies that share the same `graph_param_count > num_layers` characteristic, I don't think this needs to block the PR. However, it would be worth adding a `TODO` to revisit this logic if additional graph capture modes are introduced in the future, so that the dispatch can be based on explicit graph semantics rather than this inferred condition.**

vllm_ascend/lora/utils.py
```diff
@@ -1,82 +1,13 @@
 import vllm
```
voidvelocity, line 1: This line can be removed as well.

paulyu12: `vllm` is still used by line 16 and line 18.

vllm_ascend/attention/attention_v1.py
paulyu12: Fixed. I tested locally and there is nothing wrong happened, so I removed relevant code. Thanks.

vllm_ascend/attention/attention_v1.py
paulyu12: Thanks. I'll find out a reliable way to locate `self_attn` layers when it's necessary.

voidvelocity (approved):
This PR fix bugs about lora. I think it can be merged now.
It fix:
1. graph param about qwen
2. ascend_c op precision problem

zzzzwwjj (approved):

PR #12000, opened 2026-07-14
[BugFix]Fix block size resolution in AscendSpecDecodeBaseProposer

lcfenglinwan:

This PR introduces a check for GDN layers using `check_gdn_layer` and conditionally sets the block size. If GDN is present, it uses `kernel_block_size`; otherwise, it attempts to use `block_size`.

No

- vLLM version: v0.23.0


MengqingCao (approved):
LGTM

slippersss (approved):

Issue #12002, opened 2026-07-14
[Bug]: [KV Pool][AscendStore] Incorrect parent_block_hash for partially missing blocks causes Conductor longest_matched to stay at one block


tobking-rain:
### Your current environment

vLLM-Ascend version/commit:  0.23
vLLM version:
Mooncake version/commit:  kv index
Deployment mode: PD disaggregation + AscendStore KV Pool
KV Pool backend: Mooncake / YuanRong / other
KV block size: 128
Model:
NPU: Ascend 910B / 910C / A3
OS:
Python:
PyTorch:
CANN:

Relevant environment variable:

export PYTHONHASHSEED=0

Relevant KV transfer configuration:

{
  "kv_connector": "AscendStoreConnector",
  "kv_role": "kv_producer",
  "kv_connector_extra_config": {
    "backend": "mooncake"
  }
}

KV cache events and Mooncake Conductor are enabled.

### 🐛 Describe the bug

When AscendStore KV Pool and Mooncake Conductor are used together, the longest_matched value returned by Conductor can remain at exactly one KV block, for example 128, even when the KV Pool contains a much longer cached prefix.

The issue becomes consistently reproducible after all vLLM-Ascend instances use the same fixed PYTHONHASHSEED, for example:

export PYTHONHASHSEED=0

PYTHONHASHSEED=0 is required to make vLLM block hashes reproducible across processes and nodes. It is not the root cause of this issue.

The actual problem is that AscendStore generates an incorrect parent_block_hash when some blocks already exist in the KV Pool and only the missing blocks are stored.

As a result, Mooncake Conductor cannot connect newly stored blocks to the existing prefix chain. The Conductor query matches the first existing block and stops at the next block, so:

longest_matched = block_size = 128

This makes KV-cache-aware routing inaccurate.

Tflowers-0129:
@bowgneo Pls take a look.

PR #12012, opened 2026-07-14
[Ops][BugFix] Fix parent block hash chain exception in AscendStore KV event

zh98530:
Fixes issue：https://github.com/vllm-project/vllm-ascend/issues/12002
This PR fixes an exception in the parent block hash chain for `AscendStore` KV events. Previously, `prev_key` was tracked across loop iterations of only the missing blocks being stored. This caused incorrect parent-child relationships or exceptions when some blocks in the sequence were already cached and thus skipped.
This change resolves the issue by pre-computing all block hashes for the group (`all_hashes`) and looking up the parent block hash directly using the block index (`block_idx - 1`), ensuring correct parent-child chain tracking.

No. This is an internal bugfix for KV transfer event tracking on Ascend devices.

Tested with existing unit tests and integration tests for KV transfer.

- vLLM version: v0.24.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/kv_transfer.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/kv_transfer.py
@@ -423,6 +423,8 @@ class KVCacheStoreSendingThread(KVTransferThread):
                         )
                         if block_size is not None:
                             block_idx = start // group_block_size
+                            if block_idx >= len(all_hashes):
+                                continue
                             current_hash = all_hashes[block_idx]
                             parent_hash = all_hashes[block_idx - 1] if block_idx > 0 else None
                             stored_event = BlockStored(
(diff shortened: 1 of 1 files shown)


tobking-rain:
Please add a regression test for the partial-cache-hit scenario that this PR fixes.

The current KV-event test covers only a single block, so it cannot verify the parent-chain behavior after filtering out already cached blocks. Please add a case with four blocks where `h0` and `h2` already exist and only `h1` and `h3` are stored. The emitted events should be `h1 -> h0` and `h3 -> h2`.

This test fails with the previous `prev_key` implementation (which incorrectly makes `h3` depend on `h1`) and passes with the proposed index-based lookup.

MengqingCao (approved):

MengqingCao:
plz add a ut in the follow-up pr

zh98530:
re-run

zh98530:
rerun

zh98530:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

PR #12019, opened 2026-07-14
[CI] vllm drop v0.23.0 and upgrade v0.24.0

MrZ20:

This PR is a follow-up to https://github.com/vllm-project/vllm-ascend/pull/11875.

PR #11875 introduced the vLLM v0.24.0 support. Those adaptation changes are treated as the baseline and are intentionally not repeated in this description.

This PR removes the remaining vLLM v0.23.0 compatibility paths from `main`. 

Most changes only remove `v0.23.0` branches and keep the existing v0.24/main implementation. Those mechanical removals are not listed individually below.

#### Release defaults and documentation

Update the default `VLLM_TAG` in all maintained Dockerfiles from `v0.23.0` to `v0.24.0`.

The main-branch support matrix, slash-command examples, and balance-scheduler design documents are updated accordingly. This prevents source-built images and contributor documentation from continuing to select the unsupported v0.23 release.

#### Fused MoE weight layout boundary

Keep an explicit `vllm_version_is("0.24.0")` branch in both the standard and 310P unquantized Fused MoE implementations.

Upstream vLLM PR https://github.com/vllm-project/vllm/pull/44589 was merged as `5051698e`, 26 commits after the v0.24.0 cut point, and is present in the verified-main revision. Therefore v0.24.0 and verified main do not share the same post-load weight-layout behavior:

- v0.24.0 explicitly materializes the transposed weights as contiguous tensors before the NPU layout conversion;
- verified main follows the post-PR #44589 path without forcing the same intermediate contiguous layout.

The standard and 310P unit tests cover both version-specific layouts, the current MoE runner contract, shared-expert handling, and the 310P-specific communication method.

#### Qwen3.5/Qwen3Next output contract

Change the version boundary in the GDN and Qwen3.5/Qwen3Next patches from `v0.23.0` to `v0.24.0`.

Upstream vLLM PR https://github.com/vllm-project/vllm/pull/46998 was merged as `300e3379`, after the v0.24.0 cut point and before the current verified-main revision. It changed the attention contract from writing into a caller-provided output buffer to returning the output tensor.

#### Balance scheduler alignment

`BalanceScheduler.schedule()` is a downstream copy of the upstream scheduler body because the balance admission logic cannot be implemented through a small wrapper.

The copied body is therefore updated from the v0.23.0 implementation to the v0.24.0 implementation while preserving only the existing balance-scheduling deltas.

Both supported upstream references now expose:

```python
schedule(self, throttle_prefills: bool = False)
```

The old signature-introspection compatibility code is removed and the disabled path delegates directly to `super().schedule(throttle_prefills)`.

The v0.24 scheduler alignment also preserves the corresponding upstream behavior for:

- DP prefill throttling;
- speculative-token and maximum-length accounting;
- hybrid Mamba KV-cache hit handling;
- resumed-request bookkeeping;
- dynamic speculative decoding;
- deferred KV-block freeing;
- MRV1-only previous-step request tracking.

The balance scheduler unit tests and English/Chinese design documents are updated to use v0.24.0 as the release reference. The drift test continues to verify that the copied scheduler body differs from the pinned upstream release only by the intended balance deltas.

#### Deferred removal of owner-maintained patches

The following compatibility patches and their unit tests are intentionally retained in this PR:

- GLM47 zero-argument tool-call streaming parser;
- MiniMax-M2 incremental tool-call parser;
- MiniMax usage accounting;
- `tool_choice=none` empty-`tool_calls` response cleanup.

The first three remain behind the existing `vllm_version_is("0.23.0")` condition, with a TODO explaining that their owners will remove them in a follow-up. The `patch_tool_choice_none_content` registration is also left unchanged for the same ownership reason.

These files are not required by the newly supported v0.24/main lanes, but deleting owner-maintained patches is intentionally outside the scope of this compatibility cleanup.

#### Test boundaries

The two Qwen3 MoE/EPLB E2E files previously ran only on v0.23.0 and were skipped on main. Since v0.23.0 is removed and the cases remain broken on both v0.24.0 and verified main, they are now explicitly skipped on both supported lanes instead of being unintentionally re-enabled.

The newly rebased MRV2 data-parallel test is skipped on v0.24.0 rather than v0.23.0. MRV2 remains supported only by the verified-main lane, as established by PR #11875.

The HunyuanVL release helper names are updated from `_v023_*` to `_v024_*` because the bundled-processor compatibility path now targets v0.24.0. This is a naming correction only; the HunyuanVL adaptation itself belongs to PR #11875.

Yes.

The vLLM Ascend main branch no longer supports vLLM v0.23.0. Docker builds that do not override `VLLM_TAG` now use vLLM v0.24.0 by default.

There is no additional API change for the supported v0.24.0 and verified-main lanes.

- vLLM version: v0.23.0


Tflowers-0129 (approved):

realliujiaxu (approved):
LGTM

PR #11195, opened 2026-06-30
[BugFix] fix qwen3.5+pcp accuray error

weiguihua2:
### What this PR does / why we need it?
Fixing the precision issue of qwen3.5 with pcp integration  
1. Uniformly use the `torch.ops._C_ascend.npu_causal_conv1d_custom` operator to prevent precision drift.  
2. When integrating mtp, since `mix_qkv` has been split, `num_decode` cannot be directly obtained from the metadata and needs to be recalculated.

[2026-06-29 11:59:55,012] [ais_bench] [INFO] Summarizing evaluation results...
dataset       version    metric    mode      vllm-api-general-chat

GPQA_diamond  b1ed2c     accuracy  gen                       86.00

### Does this PR introduce _any_ user-facing change?
No

### How was this patch tested?

- vLLM version: v0.23.0
- vLLM main: https://github.com/vllm-project/vllm/commit/a30addc7548a9a8b9b3323a7bc3eb7d7c4895d1c

PR #11161, opened 2026-06-29
[BugFix][Ops] Move GDN conv1d metadata to device tensors

maoxx241:
### What this PR does / why we need it?

Related RFC issue: Refs #11723

This PR moves the metadata inputs used by Ascend GDN `npu_causal_conv1d_custom` from host materialized arguments to optional device tensors. The goal is to remove hidden host synchronization in GDN + async scheduling + MTP paths, especially for cache indices, accepted-token counts, query start locations, and initial-state flags.

Key changes:

- Changes `query_start_loc_opt`, `cache_indices_opt`, `initial_state_mode_opt`, and `num_accepted_tokens_opt` to optional Tensor inputs in the Python schema and C++ binding.
- Removes `ValueDepend` and per-element host reads for these metadata inputs in the AscendC op host path.
- Lets the AscendC kernels read dtype-native metadata directly on device: `query_start_loc` supports int32/int64, `cache_indices` and `num_accepted_tokens` support int32/int64, and `initial_state_mode` supports bool/int32/int64.
- Keeps GDN metadata dtype-native at the Python boundary, avoiding runtime casts from upstream int32/bool metadata.
- Keeps GDN cache indices on the upstream device block-table path, so async MTP no longer depends on CPU block-table or CPU seq-len correction for conv1d cache index construction.
- Aligns chunked-prefill metadata with the vLLM community fields from the current vllm-ascend verified main commit (`b9a7cd464c9ae9b1b450f8982b76d7be4de73724`), including `prefill_query_start_loc`, `prefill_state_indices`, and `prefill_has_initial_state`.
- Uses one common GDN conv1d device-metadata path for A2/A3/A5/310P. The old conv1d graph-param update hook, host-argument compatibility path, and 310P buffer-replay monkeypatch are removed.
- Fixes FULL_DECODE_ONLY + MTP GDN spec conv1d metadata to keep request granularity for `query_start_loc`, `state_indices`, and `num_accepted_tokens` instead of padding those tensors to token granularity.
- Simplifies `gdn_attn_builder.py` by replacing the custom chunk-meta pool/shape/fill path with existing FLA helper construction, and removes the now-unused `gdn_chunk_meta` Triton helper and standalone nightly coverage entry.

### Does this PR introduce _any_ user-facing change?

No. This is an internal Ascend GDN/custom-op metadata path change.

### How was this patch tested?

Current PR head `6b7f65bb3` local checks:

- `git diff --check`
[...]

PR #12027, opened 2026-07-14
[v0.23.0][BugFix][GDN] Fix PD, PCP and DCP accuracy regressions

maoxx241:

This is a minimal `releases/v0.23.0` forward fix for four Qwen3.5 GDN correctness regressions exposed after #11735, the release backport of #11161:

1. PD disaggregation without MTP can enter a decode graph before the prompt is complete and return early.
2. PCP with `FULL_DECODE_ONLY` can replay stale padded non-spec GDN metadata and corrupt output.
3. PCP + MTP mixed batches can lose the chunk-local single-token count used by PCP state propagation.
4. With MTP + DCP, a rank with no runtime speculative requests can replay capture-time spec metadata and advance stale GDN state.

The fix keeps construction ownership in the model runner/GDN metadata builders and leaves GDN forward consumers and kernels unchanged.

#### 1. Decode graph selection

`num_computed_tokens > 0` is not sufficient to identify decode. A chunked prefill or PD handoff can have a computed prefix and a final one-token prompt chunk. The release code also bypassed `is_all_decode` when speculative decoding was disabled.

Uniform decode is now selected only after every request satisfies:

```text
num_computed_tokens >= num_prompt_tokens
```

This applies with and without MTP. An unfinished one-token prompt chunk remains on the prefill path; a completed prompt still enters the decode graph.

#### 2. Non-spec graph metadata construction

In pure non-spec decode, every real request contributes one token, but graph replay may append dummy requests. The counters therefore have distinct meanings:

| Value | Meaning during full-graph replay |
| --- | --- |
| `num_decodes` | Decode classification count, including graph dummy requests |
| `num_decode_tokens` | Real non-spec decode tokens and the real request boundary |
| `num_reqs` | Request-shaped graph input size, including dummy requests |
| `num_actual_tokens` | Real token count, not the captured request shape |

The builder refreshes all fixed-shape rows at `num_reqs`, using `num_decode_tokens` as the real boundary:

```text
state_indices       = [id0, id1, NULL, NULL]
query_start_loc     = [0, 1, 2, 2, 2]
actual_seq_lengths  = [0, 1, 1, 0, 0]
```

The same helper is used by the common and 310P builders. Dummy state rows are `NULL_BLOCK_ID`; repeated terminal offsets produce zero-length dummy recurrent rows.

#### 3. PCP chunk-local single-token count

In an MTP mixed batch, single-token non-spec requests are moved into the non-spec prefill tensor so spec and non-spec decode are not consumed together. The top-level `attn_metadata.num_decodes` is therefore intentionally reset to zero, but PCP still needs the number of leading single-token sequences when it builds `h_update` and replaces the prefill state slice.

#11195 stored that value in `GDNChunkedPrefillMetadata.num_decodes`. #11735 removed the producer while retaining a default value of zero, so the consumer silently received the wrong boundary.

The chunk metadata builder now recomputes the value from its authoritative `cu_seqlens`, passes it explicitly, and makes the dataclass field required so a future producer omission fails immediately.

#### 4. Empty runtime spec rank during graph replay

Full-graph capture can retain references to the builder's speculative input buffers. On a later MTP + DCP replay, a rank may have no runtime spec requests, so the newly returned metadata has no spec branch, but the captured spec task still sees the same stable buffer addresses. Before this patch those buffers retained capture-time values.

The no-spec graph path now refreshes those captured inputs as a state no-op:

```text
spec_state_indices      = PAD_SLOT_ID
spec_query_start_loc    = 0
num_accepted_tokens     = 0
spec_actual_seq_lengths = 0
```

`PAD_SLOT_ID` prevents causal-conv state writes, while zero actual lengths prevent recurrent state updates. This restores the device-tensor equivalent of the empty-spec protection introduced by #10901 and removed with the old host graph-task update path.

### `actual_seq_lengths` ownership

The previously proposed graph-internal derivation remains removed. `actual_seq_lengths` is derived by the GDN builder from `query_start_loc` and written into stable builder-owned buffers before replay. The only additional write in this revision is the zero-length reset for a captured spec branch when the current rank has no spec work.

### Scope and related PRs

- #12009 contains the completed-prompt graph predicate on `main` together with broader PCP empty-rank and chunk-kernel work. This release PR carries only the focused predicate and regression test.
- #11670 covers a separate PD + MTP scheduler/input issue and is not included here.
- #12180 and #12181 revert #11735. This PR is the minimal forward fix if #11735 remains; the revert alternatives should not be merged together with it.
- No GDN forward consumer, causal-conv/recurrent kernel, tiling, host graph-task hook, DCP attention kernel, or public configuration is changed.

No API or configuration change. It prevents invalid graph dispatch and restores Qwen3.5 GDN metadata invariants for PD, PCP, MTP, and DCP combinations.

Local checks on commit `9cfbe3c8b`:

- focused pre-commit hooks for the changed GDN builder/tests;
- `uvx ruff check` and `uvx ruff format --check`;
- `python -m py_compile`;
- `git diff --check origin/releases/v0.23.0...HEAD`;
- exactly one signed commit on the current `releases/v0.23.0` head.

All available local checks passed.

Regression coverage verifies:

- without MTP, an unfinished one-token prompt chunk is not uniform decode, while a completed prompt remains uniform decode;
- common GDN graph padding with stale block-table/builder rows, both without MTP and with MTP configured but no speculative requests;
- the same non-spec padding contract on 310P;
- an MTP mixed batch preserves its chunk-local single-token count for PCP;
- a no-spec replay clears every stable spec input retained by a preceding full-graph spec capture.

Target-environment testing of the preceding revision confirmed the PCP and PD-without-MTP scenarios were fixed.

Earlier hardware validation of the non-spec builder portion reported:

- targeted common/310P builder tests: 22 passed in a nightly container;
- PCP=2 plus `FULL_DECODE_ONLY`, 198 prompts: baseline produced 178 corrupted responses; the fixed metadata path produced 198/198 valid JSON responses and 0 garbled responses;
- focused 16-prompt run with up to 4096 output tokens: 0 corrupted responses.

The local Mac does not have the target pytest/vLLM/NPU runtime, so the new mixed MTP + PCP and empty-spec MTP + DCP regressions still require CI and target-hardware accuracy validation.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/_310p/ops/test_gdn_310.py
+++ b/tests/ut/_310p/ops/test_gdn_310.py
@@ -99,3 +99,43 @@ def test_builder310_pads_spec_decode_metadata_with_dummy_requests():
     assert spec_meta.cache_indices.data_ptr() == attn_metadata.spec_state_indices_tensor.data_ptr()
     assert spec_meta.num_accepted_tokens.data_ptr() == attn_metadata.num_accepted_tokens.data_ptr()
     assert attn_metadata.spec_decode_metadata.actual_seq_lengths.tolist() == [0, 4, 4, 0, 0]
+
+
+def test_builder310_refreshes_non_spec_decode_graph_metadata():
+    builder = object.__new__(AscendGDNAttentionMetadataBuilder310)
+    builder.non_spec_state_indices_tensor = torch.full((4,), 77, dtype=torch.int32)
+    builder.non_spec_query_start_loc = torch.full((5,), 77, dtype=torch.int32)
+    builder.non_spec_actual_seq_lengths = torch.full((5,), 77, dtype=torch.int32)
+    builder.use_full_cuda_graph = True
+    attn_metadata = SimpleNamespace(
+        num_prefills=0,
+        num_decodes=4,
+        num_decode_tokens=2,
+        num_spec_decodes=0,
+        non_spec_state_indices_tensor=torch.tensor(
+            [10, 11, 98, 99],
+            dtype=torch.int32,
+        ),
+        non_spec_query_start_loc=torch.tensor(
+            [0, 1, 2, 2, 2],
+            dtype=torch.int32,
+        ),
+    )
+
+    builder._pad_decode_metadata(attn_metadata, graph_batch_size=4)
+
+    assert attn_metadata.non_spec_state_indices_tensor.tolist() == [
+        10,
+        11,
+        NULL_BLOCK_ID,
+        NULL_BLOCK_ID,
+    ]
+    assert attn_metadata.non_spec_query_start_loc.tolist() == [0, 1, 2, 2, 2]
+    assert attn_metadata.non_spec_state_indices_tensor.data_ptr() == builder.non_spec_state_indices_tensor.data_ptr()
+    assert attn_metadata.non_spec_query_start_loc.data_ptr() == builder.non_spec_query_start_loc.data_ptr()
+    decode_meta = attn_metadata.non_spec_decode_metadata
+    conv_meta = decode_meta.causal_conv1d
+    assert conv_meta.query_start_loc.data_ptr() == attn_metadata.non_spec_query_start_loc.data_ptr()
+    assert conv_meta.cache_indices.data_ptr() == attn_metadata.non_spec_state_indices_tensor.data_ptr()
+    assert decode_meta.actual_seq_lengths.data_ptr() == builder.non_spec_actual_seq_lengths.data_ptr()
+    assert decode_meta.actual_seq_lengths.tolist() == [0, 1, 1, 0, 0]
--- a/tests/ut/ops/test_gdn_attn_builder.py
+++ b/tests/ut/ops/test_gdn_attn_builder.py
@@ -9,6 +9,7 @@ import torch
 from vllm.config.compilation import CUDAGraphMode
 from vllm.model_executor.layers.fla.ops import index as _fla_index
 from vllm.v1.attention.backend import CommonAttentionMetadata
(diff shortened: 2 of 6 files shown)


maoxx241:
Update after the latest target-environment A/B test:

- Base: `releases/v0.23.0`
- Head: `0c5db818cf199eeeb384f8c40d42a08c0bf847e7` (one signed commit)
- Final diff: 6 files: the completed-prompt graph predicate and focused runner UT, plus the common/310P GDN metadata builders and their regression tests.
- The builder-only forward fix still reproduced premature return in PD without MTP. The runner guard is therefore restored while the combined target rerun proceeds.
- #11161 did not edit `model_runner_v1.py`; the runner change closes a pre-existing one-token prefill misdispatch that becomes correctness-critical when retaining #11161's device-metadata path.
- The implementation does not add synchronization, pending-state tracking, host graph-task rebinding, or GDN consumer/kernel changes.

The earlier Gemini summary, conflict notice, and the previous version of this comment describe superseded history. Please review the current commit and PR description.

LoganJane (approved):

czydyy:
/nightly Qwen3.5-397B-A17B-w8a8-mtp-longseq --branch releases/v0.23.0
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/29504257195)

ningjingbengxiaohai (approved):

zzzzwwjj (approved):

weijinqian0 (approved):

PR #12040, opened 2026-07-14
[BugFix] Fix GDN graph padding state indices with PCP

maoxx241:
## What this PR does

Fixes corrupted Qwen3.5/Qwen3.6 GDN decode output when PCP is combined with `FULL_DECODE_ONLY` graph execution and the active request count is smaller than the captured graph batch size.

The GDN metadata builder now:

- keeps non-spec decode metadata at the padded graph request batch size;
- copies state indices and query offsets only for real one-token decode requests;
- fills padded state-index rows with `NULL_BLOCK_ID` and repeats the final query offset.

No host metadata updater, event, or device-wide synchronization is added.

## Root cause

In full graph decode, `num_decodes`/`num_reqs` include graph padding requests, while `num_actual_tokens` is the number of real one-token decode requests.

The previous builder cleared the tail of its stable state-index buffer using `num_actual_tokens`, but then copied `num_decodes` rows from the block table. Under PCP, padded block-table rows are intentionally not initialized by the common model runner and can retain a valid state slot from an earlier batch. The copy therefore overwrote the cleared tail. The Python metadata view was then shortened to the real-token count, but the captured GDN conv1d kernel still consumed the fixed graph-sized buffer and could update a real request's conv state through the padded lane. That state corruption propagated through subsequent GDN decode steps and produced garbled output.

SFA/DSA are unaffected because their PCP paths consume explicitly padded slot-mapping/request metadata rather than constructing GDN conv/SSM state indices from padded block-table rows.

## Validation

- `pytest -q tests/ut/ops/test_gdn_attn_builder.py`: 16 passed.
- Added a regression test with two real decode requests and two padded rows containing stale valid state slots; the resulting fixed graph input is `[10, 11, 0, 0]` with query offsets `[0, 1, 2, 2, 2]`.
- Qwen3.6-35B-A3B W8A8 (Qwen3.5 GDN architecture), vLLM 0.23.0, latest vLLM Ascend main, TP2 + PCP2 + expert parallel + `FULL_DECODE_ONLY`:
  - focused 16-request GPQA run at 4096 max output tokens: 16/16 complete, 0 garbled responses;
  - full 198-request GPQA run at concurrency 16: 198/198 valid JSON responses, 0 responses with the baseline corruption signature (random control/script characters or hundreds-to-thousands of repeated spaces).

The exact Qwen3.5 checkpoint was not present on the validation host, so end-to-end validation used the available Qwen3.6-35B-A3B W8A8 checkpoint, which resolves to the same `Qwen3_5MoeForConditionalGeneration` GDN implementation.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/ops/test_gdn_attn_builder.py
+++ b/tests/ut/ops/test_gdn_attn_builder.py
@@ -573,10 +573,10 @@ def test_full_graph_spec_actual_seq_lengths_use_padded_builder_buffer():
     )
 
 
-def test_full_graph_non_spec_actual_seq_lengths_use_padded_builder_buffer():
+def test_full_graph_non_spec_metadata_nulls_padded_state_indices():
     batch_spec = BatchSpec(
-        seq_lens=[1, 1],
-        query_lens=[1, 1],
+        seq_lens=[1, 1, 0, 0],
+        query_lens=[1, 1, 0, 0],
         name="full_graph_padded_non_spec_actual_seq_lengths",
     )
     common_attn_metadata = create_common_attn_metadata(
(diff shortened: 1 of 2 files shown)


maoxx241:
Superseded by #12027, which now contains the root-cause GDN metadata-builder fix for both the common and 310P paths. The model-runner synchronization workaround is not used.

PR #12071, opened 2026-07-15
[BugFix][Worker] Snapshot query start locations before async H2D copies

freyfwt:

`query_start_loc` buffers reuse their CPU storage across scheduling iterations while `copy_to_gpu()` issues a non-blocking H2D transfer. A later CPU update can therefore overwrite the transfer source before the queued copy has consumed it, producing incorrect device-side query start locations and derived indices.

This PR:

- adds the public `copy_snapshot_to_gpu()` helper, which clones and pins the CPU buffer before copying it asynchronously into the existing GPU buffer;
- routes the v1, 310P, GDN, speculative proposer, Step3.5, and PCP query-start-location copies through that helper;
- preserves existing GPU buffer objects and addresses, so their consumers and ACL Graph paths do not need to change; and
- adds focused coverage for snapshot ownership, pinning, and non-blocking copy behavior.

No. This only prevents an internal asynchronous H2D source-buffer race.

- CPU-only stubbed unit test:
  - `run_stubbed_ut.py tests/ut/worker/test_utils.py`
  - Result: `2 passed`
- Targeted mypy for `vllm_ascend/worker/utils.py` with Python 3.10, 3.11, and 3.12: passed.
- NPU unit test:
  - `python -m pytest -q tests/ut/worker/test_utils.py`
  - Result: `2 passed, 16 warnings`
- Real NPU pinned H2D stress test:
  - reused one pinned `CpuGpuBuffer` for 1,000 snapshot copies;
  - overwrote the original CPU buffer immediately after every non-blocking copy;
  - synchronized and verified that the device buffer retained the final snapshot;
  - Result: passed, with `buffer.cpu.is_pinned() == True`.

No model weights were required.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/worker/test_pcp_manager.py
+++ b/tests/ut/worker/test_pcp_manager.py
@@ -11,7 +11,7 @@
 # limitations under the License.
 # This file is a part of the vllm-ascend project.
 
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import numpy as np
 import pytest
@@ -476,8 +476,9 @@ def test_generate_pcp_mtp_input(
         input_batch.num_computed_tokens_cpu,
         input_batch.num_prompt_tokens,
     )
-    pcp_manager.generate_pcp_mtp_input(total_num_scheduled_tokens, num_scheduled_tokens, False,
-                                       input_batch, arange_np)
+    with patch.object(torch.Tensor, "pin_memory", lambda tensor: tensor):
+        pcp_manager.generate_pcp_mtp_input(total_num_scheduled_tokens, num_scheduled_tokens, False,
+                                           input_batch, arange_np)
     assert torch.equal(
         pcp_manager.input_ids_pcp_full.cpu[:total_num_scheduled_tokens],
         target_input_ids_pcp_full)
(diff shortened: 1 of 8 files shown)


Ronald1995 (approved):
After reviewing the issue, it theoretically looks like a read-write race condition. While vLLM uses synchronization to guarantee the execution order, vllm-ascend might have other special operations that caused this issue.I have no objection to using this PR as a temporary workaround to fix the problem. However, repeatedly copying CPU memory and pinning it each time carries a certain performance risk. MRv2 completely resolved this issue by using a round-robin approach. I suggest addressing this performance risk in the future, and there are two possible ways to do so:

1.Eliminate the special write operations on query_start_loc by vllm-ascend.
2.Reference the round-robin approach to avoid copying query_start_loc from the CPU during every decode step.

zzzzwwjj (approved):

yiz-liu:
You sure this will not lead to potential host memory leak since this is creating host tensors with pinned memory continuously? We may need to investigate it further later. Personally I do not accept this, but `query_start_loc` does not take much space so this can be a workaround, for a very short time, because this will make a very bad example for the other host buffers related issues, and we don't want our codes filled up with these `copy_snapshot_to_gpu` right?

PR #11663, opened 2026-07-09
[Ascend950][Bugfix]Fix allgatherEP MXFPW4A8 quantization

Eric-dot:
### What this PR does / why we need it?
This PR fixes the allgatherEP W4A8_MXFP4 quantization path on Ascend 950.

For `torch.float8_e4m3fn` inputs, the previous logic cast `topk_weights` to the same dtype as the hidden states. However, routing weights should not be forced to float8 for the downstream fused experts kernel. This PR skips the float8 cast for `topk_weights` and enables the bf16-compatible MXFP execution mode when the input dtype is `torch.float8_e4m3fn`.

This is needed to make W4A8_MXFP4 MoE inference work correctly with allgatherEP quantization on Ascend 950.

### How was this patch tested?

- vLLM version: v0.23.0
- vLLM main: https://github.com/vllm-project/vllm/commit/1f486d96a17303ce8db8e02be39545b2be338446

PR #11113, opened 2026-06-29
[Bugfix] [Ascend950] Fix Triton kernel performance fluctuation issue when enabling prefix cache

Bybbbb11:
`Change-Id:` Icfffcd2c998866903dea35861b3bb0038e8e51f8

[Ascend950] [Bugfix] Fix Triton kernel performance fluctuation issue when enabling prefix cache

### What this PR does / why we need it?

This PR fixes the Triton kernel performance fluctuation issue observed on Qwen3-32B and Qwen3-235B for Ascend950.

**Background & Motivation**
The Triton kernels precompute_rope_cos_sin_kernel and split_qkv_rmsnorm_rope_simt_kernel were always recompiled when the batch size changed, because the parameter batch_size_per_vec was marked as tl.constexpr and its value varies with batch size. This recompilation causes intolerable compilation overhead, especially with prefix caching enabled where batch sizes can fluctuate.

**Changes Made**
1. **Removed the tl.constexpr compile‑time constant qualifier from batch_size_per_vec in both kernels**
   - In precompute_rope_cos_sin_kernel: replaced the parameter batch_size_per_vec: tl.constexpr with a fixed BLOCK_SIZE loop, computing each program’s range dynamically from pid and num_vectorcore.

   - In split_qkv_rmsnorm_rope_simt_kernel: removed the batch_size_per_vec: tl.constexpr parameter entirely; the kernel now computes the per‑program batch range internally using pid and num_vectorcore, using the already‑present fixed tile sizes for iteration.
2. **Cleaned up dead host‑side code that calculated batch_size_per_vec (no longer passed to the kernels)**

### Does this PR introduce _any_ user-facing change?

No. All changes are backend compilation and operator registration fixes. No Python API or CLI behavior is changed.

### How was this PR tested?

**Environment**

**Ascend 950 PR**
- Model: Qwen3-32B
- Before the change:
  The number of compilations **fluctuates randomly** each time the dataset test is run when enabling prefix cache.
- After the change:

  **Input length: 3500, Output length: 1500**

  |                         | dummy_run | first_bench | second_bench | third_bench |
  |-------------------------|-----------|-------------|--------------|-------------|
  | Compilation time        | 6         | 0           | 0            | 0           |
  | Output token throughput (token/s) | -   | 483.054     | 487.316      | 487.15      |

- aime2024 accuracy:
    - `mxfp8`: **76.67**, **86.67**
- GPQA accuracy:
    - `mxfp8`: **65.15**, **62.63**，**66.67**

[...]

PR #10643, opened 2026-06-18
[Feature] A5 Support Gemma4

Alex-stack-hub:
### What this PR does / why we need it?

This PR adds A5 support updates for Gemma4 graph execution.

Main changes include:
- Bind FIA graph replay metadata by layer name so captured attention ops can update with the matching per-layer metadata during graph replay.
- Reuse graph workspaces through the existing graph params update path, while using a Gemma4-specific max-workspace cache for FIA graph capture to cover layers with different workspace requirements.
- Add Gemma4 MoE compatibility for config/routing/activation differences, including `top_k_experts` fallback, custom routing functions without `num_experts`, and `gelu`/`gelu_tanh` activation handling.
- Move shared attention graph helper logic into `vllm_ascend/attention/utils.py` to keep `attention_v1.py` focused on the execution path.
- Add unit coverage for attention graph helper behavior.

These changes keep the existing non-Gemma4 workspace behavior unchanged while enabling Gemma4 graph execution on A5 to use the correct per-layer metadata and workspace state.

### Does this PR introduce _any_ user-facing change?

No. This PR does not introduce API, CLI, or configuration changes.

### How was this patch tested?

Tested Gemma4 graph execution on an A5 server.

Added unit tests for attention graph helper behavior.

Local checks:
- `python3 -m py_compile vllm_ascend/attention/attention_v1.py vllm_ascend/attention/utils.py tests/ut/attention/a2/test_attention_v1.py`
- `python3 -m ruff check vllm_ascend/attention/attention_v1.py vllm_ascend/attention/utils.py tests/ut/attention/a2/test_attention_v1.py`
- `python3 -m ruff format --check vllm_ascend/attention/attention_v1.py vllm_ascend/attention/utils.py tests/ut/attention/a2/test_attention_v1.py`

Co-authored-by: Tongpan
Co-authored-by: [swimming2007-doge](https://github.com/swimming2007-doge)

- vLLM version: v0.23.0
- vLLM main: https://github.com/vllm-project/vllm/commit/967c5c3bc38891f4465d3f4e99917ed837bb3833

PR #12087, opened 2026-07-15
[Test][Platform] Add A5 chip-detection regression guards

winson-00178005:

A5 (Ascend 950) is the only device type mapped from a single discrete SoC version value (`260`), unlike A2/A3/310P which use integer ranges — see `vllm_ascend/utils.py::check_ascend_device_type`. A5 has extensive runtime support but **no dedicated CI runner yet**, so chip-detection regressions only surface on real A5 hardware, not in the PR pipeline.

This PR adds two CPU-mock regression guards (run on every PR via `default_cpu_ut`):

1. `test_a5_soc_version_260_maps_uniquely` — pins that SoC `260` resolves to A5, and that adjacent values `259`/`261` are **rejected** (not silently treated as A5). This documents the fragile single-value mapping so a new A5 SoC variant is an intentional, reviewed change.
2. `test_is_950_only_true_for_a5` — pins that `is_950()` returns `True` only for `AscendDeviceType.A5` (several A5 code paths branch on this helper; a broken helper would silently route A5 through non-A5 paths).

This is the first of a small series of A5 pipeline-watch PRs inspired by recent A5 landings/bugfixes (#10643, #11451, #11287, #11663, #11113).

No. Test-only change.

## How was this patch tested?

- New CPU UT: `pytest -sv tests/ut/test_ascend_device_type.py` (runs on CPU via the existing `torch_npu`/`acl` mock in `tests/ut/conftest.py`, no NPU required).
- Auto-collected by the always-on `default_cpu_ut` module (`tests/ut`, `cpu_only: true`) — no `test_config.yaml` change needed.

- vLLM version: v0.24.0


winson-00178005:

PR #12089, opened 2026-07-15
[Test][Platform] Add A5 aclgraph capture-size and SFA-C8 platform guards

winson-00178005:

Two A5-only platform safety guards live in `vllm_ascend/platform.py`, but because there is **no A5 CI runner yet**, removing them during refactoring would let A5 regressions slip through silently:

- `prune_capture_sizes_for_950` — A5 HDK has an ACL graph capture-size incompatibility; capture sizes are pruned to `MAX_CAPTURE_SIZES_FOR_950 = 4` (see the TODO in `platform.py`).
- The SFA-DCP sparse-C8 path raises `NotImplementedError` on A5 (A5 uses a fused CKV quant sparse attention path that needs a separate DCP LSE merge).

Recent A5 regressions (e.g. prefix-cache performance fluctuation #11113) are exactly the class of issues these guards exist to prevent.

This PR adds two CPU tests (run on every PR via `default_cpu_ut`):

1. `test_prune_capture_sizes_for_950_behavior` — verifies pruning samples down to `MAX_CAPTURE_SIZES_FOR_950`, **preserves the first/last capture sizes**, and is a **no-op** when already within the limit.
2. `test_a5_platform_safety_guards_present` — source-level drift guard asserting the A5 pruning call-site and the SFA-DCP `not supported on A5` raise remain in place.

No. Test-only change.

## How was this patch tested?

- New CPU UT: `pytest -sv tests/ut/test_a5_platform_guards.py`.
- Auto-collected by the always-on `default_cpu_ut` module — no `test_config.yaml` change needed.

- vLLM version: v0.24.0

### DIFF
--- a/.github/workflows/scripts/test_config.yaml
+++ b/.github/workflows/scripts/test_config.yaml
@@ -616,6 +616,7 @@
     - vllm_ascend/utils.py
     - setup.py
   tests:
+    - tests/ut/test_a5_platform_guards.py
     - tests/ut/test_ascend_config.py
     - tests/ut/test_ascend_forward_context.py
     - tests/ut/test_cpu_binding.py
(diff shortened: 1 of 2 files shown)


winson-00178005:

PR #12090, opened 2026-07-15
[Test][Ops] Add A5 MC2 token-dispatch MXFP quant path guards

winson-00178005:

On A5 (Ascend 950), MC2 token dispatch selects MXFP communication quantization (`quant_mode=4`) and passes an explicit `y_dtype` through MC2 — see `vllm_ascend/ops/fused_moe/token_dispatcher.py::get_dispatch_mc2_kwargs`. This A5 path **recently regressed**:

- #11663 `[Ascend950][Bugfix] Fix allgatherEP MXFPW4A8 quantization`
- #11287 `[Ascend950][Feature] Add allgatherEP MXFP4 quantization`

Because there is **no A5 CI runner yet**, these regressions only surfaced on real A5 hardware. This PR adds two CPU-mock guards (run on every PR via `default_cpu_ut`) that pin the A5 branch so it cannot silently drift back to the non-A5 path:

1. `test_a5_mxfp_dispatch_uses_quant_mode_4_and_y_dtype` — A5 + MXFP dispatch must select `quant_mode=4` and inject `y_dtype=float8_e4m3fn`.
2. `test_non_a5_dispatch_uses_quant_mode_2_without_y_dtype` — non-A5 (A2) dispatch must stay at `quant_mode=2` and must **not** inject `y_dtype` (i.e. the A5 branch must not leak into the non-A5 path).

The dispatcher is constructed via `object.__new__` to bypass the heavy `__init__` (HCCL/distributed setup), with `_get_expert_token_nums_type` mocked — keeping the test fast (<1s) and NPU-free.

No. Test-only change.

## How was this patch tested?

- New CPU UT: `pytest -sv tests/ut/ops/test_a5_dispatch_quant.py`.
- Auto-collected by the always-on `default_cpu_ut` module (`tests/ut`, `cpu_only: true`) — no `test_config.yaml` change needed.

- vLLM version: v0.24.0


winson-00178005:

PR #12092, opened 2026-07-15
[Refactor][Ops] Remove unused indexer_compress_epilog operator

maoxx241:

Removes the unused legacy `indexer_compress_epilog` operator for Ascend 950.

- Deletes the operator host and kernel implementation.
- Removes it from the A5 custom-operator build list.
- Removes its Torch runtime and Meta registrations.
- Keeps `indexer_compress_epilog_v2`, which is the implementation used by the DeepSeek V4 A5 path.

The legacy operator had no Python runtime callers and was retained only in the build and registration layers.

No. The active DeepSeek V4 A5 path continues to use `indexer_compress_epilog_v2`.

- `bash format.sh ci`
- `git diff --check`
- Verified there are no remaining non-v2 `indexer_compress_epilog` references.

NPU execution was not run because this PR only removes an operator with no runtime callers.

- vLLM version: v0.23.0


maoxx241:
Superseded by #12896, which consolidates the unused DSv4 A5 KV-cache operators with the related legacy HcPre cleanup.

PR #12096, opened 2026-07-15
[Performance][Ops][Feature]optimize slot mapping kernel:

ckeyever:
This PR optimizes the Ascend slot mapping Triton kernel in the v1 block table path.

The previous path reused the generic upstream vLLM `_compute_slot_mapping_kernel`, which can generate excessive scalar computation on Ascend NPUs due to int64 arithmetic, modulo operations, and non-contiguous block table loads. It also used one final program to serially fill the padded slot mapping range, which is inefficient when `num_tokens` is much smaller than `max_num_tokens`.

This PR adds a local Ascend-optimized `_compute_slot_mapping_kernel` with the following improvements:

- use int32 arithmetic in the hot path;
- rewrite modulo operations with div/mul/sub expressions;
- replace non-contiguous block table loads with bounded contiguous window loads plus `tl.gather`;
- bound the block table window by the current token tile instead of the full block table row;
- preserve physical KV block to kernel block mapping with `KV_CACHE_BLOCK_SIZE` and `BLOCKS_PER_KV_BLOCK`;
- parallelize padding by launching one padding program per `BLOCK_SIZE` tile instead of using a single trailing program.

This reduces scalar computation and avoids a single-core padding bottleneck for cases where `num_tokens << max_num_tokens`.

Performacne
|cases | shape | time before opt  | time after opt| gains|
| ------ | ------ | ------ |--------| -------- |
| case1 | num_tokens=4,num_reqs=2,CP=1,max_num_batched_tokens=8192 | 173.19us | 3.071us | 5639.5% |
| case2 | num_tokens=32\*256,num_reqs=256,CP=1,max_num_batched_tokens=32\*256 | 1056.497us | 21.155us | 4994.1% |

No. This is an internal performance optimization for Ascend slot mapping.

Tested with new unit tests in `tests/ut/worker/a2/test_block_table.py` verifying the kernel launch grid with different numbers of tokens and requests.

- vLLM version: v0.25.1

### DIFF
--- a/tests/ut/worker/a2/test_block_table.py
+++ b/tests/ut/worker/a2/test_block_table.py
@@ -172,6 +172,69 @@ class TestBlockTableComputeSlotMapping(TestBase):
                     f"dcp_rank={dcp_rank}, pcp_rank={pcp_rank}",
                 )
 
+    def _run_slot_mapping_kernel_launch_case(
+        self,
+        num_reqs: int,
+        num_tokens: int,
+        max_num_batched_tokens: int,
+        expected_pad_blocks: int,
+    ):
+        old_max_num_batched_tokens = self.max_num_batched_tokens
+        self.max_num_batched_tokens = max_num_batched_tokens
+        try:
+            block_table = self.create_block_table(
+                dcp_world_size=1,
+                dcp_rank=0,
+                pcp_world_size=1,
+                pcp_rank=0,
+                cp_kv_cache_interleave_size=1,
+            )
+        finally:
+            self.max_num_batched_tokens = old_max_num_batched_tokens
+
+        self.setup_block_table_data(block_table, num_reqs=num_reqs)
+        counts = np.full(num_reqs, num_tokens // num_reqs, dtype=np.int32)
+        counts[: num_tokens % num_reqs] += 1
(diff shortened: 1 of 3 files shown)


vllm_ascend/worker/block_table.py
```diff
@@ -326,6 +322,88 @@ def _make_buffer(self, *size: int | torch.SymInt, dtype: torch.dtype) -> CpuGpuB
         return CpuGpuBuffer(*size, dtype=dtype, device=self.device, pin_memory=self.pin_memory)
 
 
+def _next_power_of_2(value: int) -> int:
+    return 1 << (value - 1).bit_length()
+
+
+@triton.jit(do_not_specialize=["num_tokens", "max_num_tokens"])
```
chenlongxiao: 归档到统一ops文件夹下

ckeyever: OK

tests/ut/worker/a2/test_block_table.py
```diff
@@ -172,6 +172,69 @@ def _test_slot_mapping_for_ranks(self, dcp_world_size, pcp_world_size, cp_kv_cac
                     f"dcp_rank={dcp_rank}, pcp_rank={pcp_rank}",
                 )
 
+    def _run_slot_mapping_kernel_launch_case(
```
chenlongxiao, line 175: 补充MD

PR #12097, opened 2026-07-15
 [BugFix] Forward upstream HTTP error status code and body to client

paopaozhang:
When the Prefill (P) node returns an HTTP error (e.g. 400 for context‑length exceeded), the proxy swallows the original status code and response body, and returns a bare 500 "Internal Server Error" (text/plain) to the client.

Root cause: The except Exception: block inside the request handler unconditionally re‑raises all exceptions. FastAPI’s default exception handler converts them into plain 500 responses, discarding the original 400 status code and JSON error payload from the Prefill node.

#### Impact
Clients receive HTTP 500 "Internal Server Error" (plain text), instead of the real error (e.g. HTTP 400 with JSON indicating context exceeds max_model_len).

Original error type, status code and root cause from the Prefill node are lost, making troubleshooting difficult.

None.

None.

- vLLM version: v0.26.0


wxsIcey (approved):
LGTM!

yiz-liu (approved):

PR #12127, opened 2026-07-15
[v0.23.0][Bugfix] Bind Mooncake receiver threads to KV cache device

maoxx241:
## What this PR does

Backport of #12126 to `releases/v0.23.0`.

- Bind every Mooncake receiver `ThreadPoolExecutor` worker to the NPU device of the registered KV cache.
- Apply the fix to both the standard and hybrid Mooncake connectors.
- Add regression tests that force two executor workers and verify each worker calls `torch.npu.set_device` before handling a request.

## Why this is needed

NPU current-device state is thread-local. Mooncake receiver executor workers do not inherit the device selected by the decode model worker thread and start with logical device 0.

During TP-mismatch KV cache reformatting, current-device-dependent NPU calls can therefore create an unintended context on device 0. This makes the same non-zero-rank worker PID appear on both its assigned NPU and device 0.

The registered KV cache tensor is used as the authoritative device source instead of `tp_rank`, which is not reliable with DP/PP/PCP layouts or `ASCEND_RT_VISIBLE_DEVICES` remapping.

## User impact

This prevents an extra NPU context from leaking onto device 0 in Mooncake PD-disaggregated deployments. There is no API change.

## Validation

- Cherry-picked cleanly onto `origin/releases/v0.23.0@c83bf9bd`.
- Both related release unit-test files: `110 passed, 18 warnings`.
- `git diff --check` and `py_compile` pass.
- The equivalent main change was validated on Ascend A3 with Qwen3-4B and a P TP4 / D TP2 topology; after the fix, four KV transfers and generation completed successfully while every decode PID remained on one chip.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/kv_offload/test_mooncake_connector.py
+++ b/tests/ut/kv_offload/test_mooncake_connector.py
@@ -618,6 +618,69 @@ class TestKVCacheRecvingThreadBasic(unittest.TestCase):
         result = self.thread.get_and_clear_finished_requests()
         self.assertEqual(result, {"req1", "req2"})
 
+    def test_executor_workers_bind_kv_cache_device_before_handling_requests(self):
+        expected_device = torch.device("npu:5")
+        kv_cache = MagicMock(device=expected_device)
+        worker_events: defaultdict[int, list[tuple[str, int | str]]] = defaultdict(list)
+        events_lock = threading.Lock()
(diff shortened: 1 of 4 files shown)


Pz1116 (approved):
LGTM

nwpu-zxr (approved):

PR #12130, opened 2026-07-16
[BugFix] Fix block_table overwrite during CPU-to-NPU transfer

bubaishenhua112-netizen:

In the PD disaggregated inference scenario, `block_table` may be overwritten during the asynchronous CPU-to-device transfer, resulting in incorrect data being copied to the device and causing inference accuracy issues.

This PR fixes the issue by creating a cloned copy of `block_table` before the device transfer. The cloned tensor is then used as the source of the asynchronous `copy_()` operation, ensuring that the transferred data remains unchanged even if the original `block_table` is updated concurrently.

No.

Verified in the PD disaggregated inference scenario.

- Reproduced the inference accuracy issue before the fix.
- Applied this patch and reran the same workload.
- Confirmed that `block_table` is no longer overwritten during CPU-to-device transfer and the inference accuracy issue is resolved.

- vLLM version: v0.24.0

### DIFF
--- a/vllm_ascend/worker/block_table.py
+++ b/vllm_ascend/worker/block_table.py
@@ -285,7 +285,10 @@ class BlockTable:
             self.slot_mapping.cpu[: req_indices.shape[0]] = torch.where(mask, slot_mapping, -1)
 
     def commit_block_table(self, num_reqs: int) -> None:
-        self.block_table.copy_to_gpu(num_reqs)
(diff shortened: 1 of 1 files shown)


weijinqian0 (approved):

bubaishenhua112-netizen:
/cherry-pick releases/v0.23.0
[Bot]: cherry-pick completed successfully. New PR created: https://github.com/vllm-project/vllm-ascend/pull/12189

PR #12136, opened 2026-07-16
[Performance][KVCache] Optimize AscendStore key construction and miss path

DreamerLeader:

This adapts the AscendStore key-construction and KVPool miss-path optimizations to `releases/v0.23.0` while preserving the hybrid/layerwise architecture introduced on the release branch.

Key construction and lookup:

- Keeps `process_tokens()` as a direct lazy-hash-to-`PoolKey` path for layerwise `split_layers()` operations.
- Generates serialized non-layerwise keys directly, caches immutable prefixes, and computes grouped hashes lazily.
- Propagates HBM hit tokens through the lookup RPC, skips external lookup on a full HBM hit, and avoids querying the HBM-resident prefix again.
- Optionally applies hybrid-cache reachable masks before key construction.

Put and save path:

- Applies store/load masks before key construction.
- Adds validated sparse store-mask key construction and optional pre-shard key construction.
- Preserves the raw KVPool hit length so an already pooled prefix is not queried or stored again.
- Adds global-queue, per-request, and asynchronous save modes with exception-safe queue/request lifecycle handling.
- Skips external lookup for retention-enabled prompts that are too short to produce a reusable retained prefix.

Scope boundaries:

- This PR intentionally does **not** add C128-first lookup, a C128 miss gate, or `LOOKUP_FULL_GUARD` behavior.
- Layerwise mode continues to use `PoolKey` and per-layer keys. The non-layerwise string fast path and asynchronous save controls do not replace the layerwise transfer path.

The default key format and cache behavior remain unchanged. The direct string path is enabled by default. More aggressive sparse/pre-shard lookup and save behavior remains opt-in through documented environment variables, including `VLLM_ASCEND_ASYNC_SAVE`; asynchronous save can allow a following identical lookup to race with save visibility.

- 137 targeted AscendStore unit tests passed with repository mock dependencies; 10 existing deprecated layerwise tests were skipped.
- `ruff check` and `ruff format --check` passed for all changed Python files.
- Targeted mypy checks passed with missing external imports ignored locally; CI provides the full vLLM/PyTorch environment.
- Python compile checks, repository custom Python checks, and `git diff --check` passed.
- Local key-build microbenchmarks show about 4x speedup for direct 512-key construction, 6.9x for a 50%-masked 512-block path, and effectively unchanged layerwise key generation.
- NPU hardware testing was not available locally; CI/NPU validation is required.

- vLLM version: v0.23.0

### DIFF
--- a/docs/source/user_guide/feature_guide/kv_pool.md
+++ b/docs/source/user_guide/feature_guide/kv_pool.md
@@ -49,6 +49,19 @@ To guarantee uniform hash generation, it is required to synchronize the PYTHONHA
 export PYTHONHASHSEED=0
 ```
 
+AscendStore also provides the following performance switches. Except for the
+direct key-string path, they are disabled by default so deployments can opt in
+after validating their cache-visibility and retention requirements.
+
+| Environment variable | Default | Description |
+| :--- | :--- | :--- |
+| `VLLM_ASCEND_PUT_KEY_STRING_FAST_PATH` | `1` | Builds backend key strings without allocating intermediate `PoolKey` objects. Layerwise mode keeps using `PoolKey` because it needs per-layer keys. |
+| `VLLM_ASCEND_PUT_SPARSE_STORE_MASK` | `0` | Builds keys only for chunks selected by the hybrid-cache store mask. The implementation falls back automatically when mask and key granularities do not match. |
+| `VLLM_ASCEND_PUT_PRE_SHARD_KEY_BUILD` | `0` | With sparse store-mask key building enabled, builds only the keys owned by the local put shard. |
+| `VLLM_ASCEND_LOOKUP_REACHABLE_MASK` | `0` | Skips hybrid-cache chunks that cannot contribute to an aligned external-cache hit. |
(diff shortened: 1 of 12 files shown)


docs/source/user_guide/feature_guide/kv_pool.md
```diff
@@ -49,6 +49,19 @@ To guarantee uniform hash generation, it is required to synchronize the PYTHONHA
 export PYTHONHASHSEED=0
 ```
 
+AscendStore also provides the following performance switches. Except for the
+direct key-string path, they are disabled by default so deployments can opt in
+after validating their cache-visibility and retention requirements.
+
+| Environment variable | Default | Description |
+| :--- | :--- | :--- |
+| `VLLM_ASCEND_PUT_KEY_STRING_FAST_PATH` | `1` | Builds backend key strings without allocating intermediate `PoolKey` objects. Layerwise mode keeps using `PoolKey` because it needs per-layer keys. |
```
zzzzwwjj, line 58: 1. Why do we need these parameters?
2. We need to control the number of environment variables. If these parameters are necessary, please configure them through `additional_config`, and add params description in `additional_config.md`.

zzzzwwjj (changes requested):

PR #12168, opened 2026-07-16
[BugFix] [load_balance_proxy] Forward decode backend errors to client instead of empty 200

### LINKED ISSUE #12166: [load_balance_proxy] Decode backend errors are swallowed into an empty HTTP 200 response
## Summary

## Reproduction

```bash

```

## Root cause

```python
except Exception as exc:
finally:
```

Two aggravating factors:

## Impact

## Suggested fix

xlshaoscu:
## Description

Fixes #12166.

When the decode backend returns an error (e.g. 4xx because `prompt_tokens + max_tokens > max_model_len`, or 5xx/OOM/connection), the load-balance proxy currently swallows the exception inside `generate_stream()` and ends the `StreamingResponse` generator without yielding any body. Because `StreamingResponse` had already sent the HTTP 200 head, the client received a `200` with an **empty body** and no way to learn the cause.

This PR makes the decode-phase error visible to the client.

## Changes (`examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py`)

1. **`open_stream_service_response_with_retry`** (renamed from the streaming wrapper): open the decode response with `client.send(..., stream=True)` and return `(response, chunk_iterator, first_chunk)` instead of consuming it as a generator. On a non-2xx response it buffers the upstream error body (`await response.aread()`), retries 5xx, and **returns the 4xx/last-5xx immediately without raising** so the caller can forward the real upstream status. 4xx is not retried (deterministic). The first chunk is pre-read on success so connection-level failures surface within the retry budget.

2. **`_open_decoder_stream_or_early_response`**: open the decode response *before* creating `StreamingResponse`. If the decode backend failed up front, return a verbatim `Response(status_code=upstream.status_code, content=upstream.content)` (or a `502` for connection failures) — i.e. the real upstream HTTP status reaches the client before the ASGI response head is committed. Only on success is a `StreamingResponse` created (carrying the upstream 200).

3. **`_DecodeStreamSession`** (replaces the nested `generate_stream` closure): per-request state for chunk forwarding, recompute retry, and error forwarding. Its `except Exception` forwards a proxy-side/late fault to the client as an SSE `error` event + `data: [DONE]` (streaming) or a JSON `{"error": {...}}` body (non-streaming), instead of dropping it. A mid-stream late decode error (only reachable on a recompute retry after the head is committed) is forwarded the same way.

4. Helpers: `DecodeUpstreamError`, `build_error_payload` (returns the standard `{"error":{...}}` shape verbatim, with a flat-`{"message":}` wrap and a raw-body-truncation fallback), `_encode_error_body`, `_parse_stream_chunk`, `_upstream_error_response`, `_decode_unavailable_response`, `_late_decode_error_body`.

## Behavior

| Scenario | Before | After |
| --- | --- | --- |
| Streaming + initial decode 4xx/5xx | `200` + 0 bytes | **real `4xx`/`5xx`** + upstream body verbatim |
| Non-streaming + initial decode 4xx/5xx | `200` + 0 bytes | **real `4xx`/`5xx`** + upstream body verbatim |
| Decode backend unreachable (retries exhausted) | `200` + 0 bytes | **`502`** + `{"error":{...,"code":"decode_backend_unavailable"}}` |
| Mid-stream late decode error (recompute retry) | `200` + 0 bytes (or truncated) | `200` head already sent + SSE/JSON `{"error":{...}}` body |
| Normal request | unchanged | unchanged |

Note on the streaming status code: the original draft only did body-level forwarding because `StreamingResponse` commits the 200 head before the generator runs. The merged implementation goes further — by pre-opening the decode response before `StreamingResponse` is constructed, **initial** decode 4xx/5xx are returned with the real upstream status code. Only errors that occur *after* the head is committed (a recompute retry failing mid-stream) fall back to body-level forwarding, since the status can no longer be changed.

## Follow-up fixes in this PR

- `max_retries=0` boundary: `open_stream_service_response_with_retry` now uses `max(1, max_retries)` so `max_retries=0` still makes one attempt (previously `range(1, 0+1)` was empty and raised `RuntimeError` without sending any request). Matches the prefill path's existing convention.
- Non-streaming late-error corruption: once a non-streaming body chunk has been forwarded, a subsequent late decode error body is dropped instead of appended (appending would concatenate with the already-sent single JSON into invalid JSON). Streaming is unaffected (SSE errors are independent events).

## Verification

Unit-checked locally with a mocked decode backend (`max_retries=3`):
- 4xx → returned immediately, not retried; real status + body forwarded.
- 5xx → retried, then last 5xx returned on exhaustion; 5xx-then-200 succeeds.
- `RequestError` → retried, then `502 decode_backend_unavailable`.
- `build_error_payload` extracts vLLM's `{"error":{"message":...,"type":"BadRequestError"}}` verbatim; flat `{"message":}` wrapped; non-JSON body truncated to 1000B; non-upstream exceptions fall back to `502`.
- `max_retries=0` now makes one attempt (no `RuntimeError`).
- 54 mock unit cases pass (0 failures): covers all branches of `open_stream_service_response_with_retry`, `_open_decoder_stream_or_early_response`, `_DecodeStreamSession` (incl. recompute loop, mid-stream late error, raw/skip chunk parsing, cancellation), and `handle_completions_impl` (incl. `/v1/completions` prompt path and outer `HTTPStatusError` handling).

Reproduction from the issue against a deployed 1P1D instance (Qwen2.5-7B-Instruct on Atlas 800T A2):
- Non-streaming over-limit (`max_tokens` > `max_model_len`): previously `200` + 0 bytes → now returns the real upstream `400` + `{"error":{"message":"max_tokens=... cannot be greater than max_model_len..."}}`.
- Streaming over-limit: previously `200` + 0 bytes → now returns the real upstream `400` + the same error body.
- Decode backend down: previously `200` + 0 bytes → now `502` + `decode_backend_unavailable`.
- Normal requests: unchanged.

### DIFF
--- a/examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
+++ b/examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
@@ -875,7 +875,13 @@ async def stream_service_response(
         first_chunk_sent = False
         try:
             async with client.stream("POST", endpoint, json=req_data, headers=headers) as response:
-                response.raise_for_status()
+                if response.status_code >= 400:
+                    # Buffer the upstream error body so callers can forward the real
+                    # cause (e.g. vLLM's "maximum context length" message) instead of
+                    # dropping it into an empty 200. The body is unreadable once this
+                    # streaming context exits, so read it before raising.
+                    await response.aread()
+                    response.raise_for_status()
                 async for chunk in response.aiter_bytes():
                     first_chunk_sent = True
                     yield chunk
@@ -993,6 +999,20 @@ async def reassign_instances(
     return await assign_instances(api, req_data, request_length, is_initial_request=False)
 
 
+async def _replay_first_chunk(first_chunk: bytes, gen: Any):
+    """Replay a pre-read first chunk, then stream the rest of ``gen``.
+
+    ``handle_completions_impl`` pre-reads the first chunk of the decode stream
+    before committing the ASGI response head so an initial decode error can be
+    returned with the real HTTP status; this replays that chunk so the client
+    still receives the full stream.
+    """
+    if first_chunk:
+        yield first_chunk
+    async for chunk in gen:
+        yield chunk
+
+
 async def handle_completions_impl(api: str, request: Request):
     runtime = get_runtime()
     args = get_global_args()
@@ -1014,9 +1034,53 @@ async def handle_completions_impl(api: str, request: Request):
             origin_prompt = ""
         origin_max_tokens = req_data.get("max_tokens", 16)
 
+        # Pre-open the decode response BEFORE committing the ASGI 200 head so an
+        # initial decode 4xx/5xx/connection failure can be returned with the real
+        # HTTP status code instead of an empty 200. On success the opened stream
+        # (first chunk already read) is handed to generate_stream via these vars.
+        preopened_gen = None
+        preopened_first = b""
+        decoder_client = await runtime.get_client(ServerRole.DECODE, instance_info.decoder_key)
+        preopened_gen = stream_service_response(
+            decoder_client,
+            api,
+            req_data,
+            request_id=instance_info.request_id,
+            max_retries=args.max_retries,
+            base_delay=args.retry_delay,
+        )
+        try:
+            preopened_first = await preopened_gen.__anext__()
+        except asyncio.CancelledError:
+            await _finish_instance(runtime, instance_info, release_prefill_kv=True)
+            request_released = True
+            raise
+        except StopAsyncIteration:
+            preopened_first = b""
+        except httpx.HTTPStatusError as exc:
+            await _finish_instance(runtime, instance_info, release_prefill_kv=True)
+            request_released = True
+            return Response(
+                content=exc.response.content, status_code=exc.response.status_code, media_type="application/json"
+            )
+        except httpx.RequestError as exc:
+            await _finish_instance(runtime, instance_info, release_prefill_kv=True)
+            request_released = True
+            return JSONResponse(
+                status_code=502,
+                content={
+                    "error": {
+                        "message": f"Decode backend request to {api} failed: {exc}",
+                        "type": "upstream_error",
+                        "code": "decode_backend_unavailable",
+                    }
+                },
+            )
+
         async def generate_stream():
             nonlocal instance_info
             nonlocal request_released
+            nonlocal preopened_gen, preopened_first
             generated_token = ""
             released_kv = False
             retry_count = 0
@@ -1035,15 +1099,22 @@ async def handle_completions_impl(api: str, request: Request):
             try:
                 while retry:
                     retry = False
-                    decoder_client = await runtime.get_client(ServerRole.DECODE, instance_info.decoder_key)
-                    async for chunk in stream_service_response(
-                        decoder_client,
-                        api,
-                        req_data,
-                        request_id=instance_info.request_id,
-                        max_retries=args.max_retries,
-                        base_delay=args.retry_delay,
-                    ):
+                    if preopened_gen is not None:
+                        # First iteration: consume the decode stream pre-opened by the
+                        # caller so initial 4xx/5xx could be returned with the real status.
+                        gen = _replay_first_chunk(preopened_first, preopened_gen)
+                        preopened_gen = None
+                    else:
+                        decoder_client = await runtime.get_client(ServerRole.DECODE, instance_info.decoder_key)
+                        gen = stream_service_response(
+                            decoder_client,
+                            api,
+                            req_data,
+                            request_id=instance_info.request_id,
+                            max_retries=args.max_retries,
+                            base_delay=args.retry_delay,
+                        )
+                    async for chunk in gen:
                         if not released_kv and chunk:
                             await release_prefill_kv_once()
                         try:


xlshaoscu:
Added unit tests in `examples/disaggregated_prefill_v1/test_load_balance_proxy_server_example.py` (commit c93f8ac). They stub out the runtime/scheduler, so no real backend is required.

## Test cases (9)

**`build_error_payload`**
- `test_build_error_payload_upstream_4xx` — decodes vLLM's `{"error":{"message":...,"type":"BadRequestError"}}` body, keeps the original message/type/code.
- `test_build_error_payload_non_upstream_falls_back_to_502` — non-upstream exceptions map to `502 / proxy_error`.

**`stream_service_response_with_retry`**
- `test_stream_retry_200_yields_chunks` — 200 streams chunks normally.
- `test_stream_retry_4xx_raises_immediately_without_retry` — 400 raises `DecodeUpstreamError(400)` immediately, no retry even with `max_retries=3`, no body streamed.
- `test_stream_retry_5xx_retries_then_raises` — 500 retries up to `max_retries`, then raises `DecodeUpstreamError(500)`.

**End-to-end via `handle_completions_impl`** (decode stubbed to fail / succeed)
- `test_streaming_decode_error_forwards_sse_error_event` — streaming + 400 → body contains `data: {"error":{...}}` with the vLLM message + `data: [DONE]`, body is non-empty (no longer an empty 200).
- `test_non_streaming_decode_error_forwards_json_error_body` — non-streaming + 400 → body contains `{"error":{...}}`, non-empty.
- `test_streaming_success_unchanged` — streaming success → normal `content` + `[DONE]`.
- `test_non_streaming_success_unchanged` — non-streaming success → normal `content`.

## Result

```
$ pytest examples/disaggregated_prefill_v1/test_load_balance_proxy_server_example.py -v
============================= test session starts =============================
platform win32 -- Python 3.14.3, pytest-9.1.1, pluggy-1.6.0
configfile: pyproject.toml
plugins: anyio-4.12.1
collected 9 items

test_load_balance_proxy_server_example.py
  ::test_build_error_payload_upstream_4xx                              PASSED [ 11%]
  ::test_build_error_payload_non_upstream_falls_back_to_502            PASSED [ 22%]
  ::test_stream_retry_200_yields_chunks                                PASSED [ 33%]
  ::test_stream_retry_4xx_raises_immediately_without_retry             PASSED [ 44%]
  ::test_stream_retry_5xx_retries_then_raises                          PASSED [ 55%]
  ::test_streaming_decode_error_forwards_sse_error_event               PASSED [ 66%]
  ::test_non_streaming_decode_error_forwards_json_error_body           PASSED [ 77%]
  ::test_streaming_success_unchanged                                   PASSED [ 88%]
  ::test_non_streaming_success_unchanged                               PASSED [100%]

============================== 9 passed in 0.52s ==============================
```

## Deployment verification (against a running proxy with this fix)

Also verified end-to-end against a deployed instance (`kimi_k26`, `max_model_len=131072`, `prompt_tokens=9`, boundary `max_tokens=131063`). Before the fix, every over-limit request returned `200` + 0 bytes. After:

| # | Scenario | stream | max_tokens | HTTP | size | Before (same proxy) |
| --- | --- | --- | --- | --- | --- | --- |
| 1 | normal | yes | 64 | 200 | 4631 B | 200 normal (unchanged) |
| 2 | normal | no | 64 | 200 | 677 B | 200 normal (unchanged) |
| 3 | over-limit (boundary+1) | yes | 131064 | 200 | 395 B | **200 + 0 bytes** |
| 4 | over-limit (boundary+1) | no | 131064 | 400 | 374 B | **200 + 0 bytes** |
| 5 | boundary OK | yes | 131063 | 200 | 16435 B | 200 normal (unchanged) |
| 6 | boundary OK | no | 131063 | 200 | 861 B | 200 normal (unchanged) |
| 7 | huge (1M) | yes | 1000000 | 200 | 242 B | **200 + 0 bytes** |
| 8 | huge (1M) | no | 1000000 | 400 | 221 B | **200 + 0 bytes** |

Streaming over-limit body (scenario 3) is now:
```
data: {"error": {"message": "This model's maximum context length is 131072 tokens. However, you requested 131064 output tokens and your prompt contains at least 9 input tokens, for a total of at least 131073 tokens. ...", "type": "BadRequestError", "code": 400}}

data: [DONE]
```

Non-streaming over-limit (scenario 4) now returns `HTTP 400` with the same `{"error":{...}}` body. Normal and boundary requests are unchanged.

c5566b:
Hi @xlshaoscu, thanks for working on #12166.

While preparing a fix based on the latest `main`, we reviewed this PR and noticed one possible remaining gap: the decode request still appears to be opened inside the `StreamingResponse` body iterator. That means for non-streaming requests the proxy may have already committed HTTP 200 before it observes the decoder 4xx/5xx response, so the backend error body may become visible but the real HTTP status may still not be forwarded.

We prepared an alternative PR here:

https://github.com/vllm-project/vllm-ascend/pull/13578

The main difference is that #13578 opens the decoder response before returning the proxy response. This lets the proxy forward initial decoder 4xx/5xx status and body before the ASGI response starts. It also adds focused UT coverage under `tests/ut/`, including assertions that initial decode 400 returns real HTTP 400 for both streaming and non-streaming requests.

Could you take a look and see whether this matches the issue you intended to fix? If maintainers prefer continuing with this PR, we are also happy to fold the relevant implementation idea or tests back here.

xlshaoscu:
Hi @c5566b, thanks for the detailed review and for #13578 — you're right that opening the decode response before the `StreamingResponse` is the correct way to forward the real HTTP status code. I've rewritten this PR to adopt that architecture.

**Adopted from #13578:**
- Split `stream_service_response_with_retry` into `open_stream_service_response_with_retry` (opens the response, returns `(response, chunk_iterator, first_chunk)`) + a thin backward-compatible generator wrapper.
- The decode response is now opened **before** `StreamingResponse(...)` is created, so initial 4xx/5xx returns the real upstream status code and body instead of an empty 200.
- First-chunk pre-read so connection errors surface within retry budget.
- 4xx not retried; 5xx keeps retry. Late errors (after the head is committed) are still forwarded as SSE/JSON error bodies.

**Kept / optimized from this PR:**
- **One unified error renderer instead of five.** #13578 adds `backend_response_headers` / `decoder_error_payload` / `decoder_error_chunk` / `decoder_stream_error_chunk` / `decoder_unavailable_response`; this PR collapses them into a single `build_error_payload(exc)` used only for late in-stream errors (the pre-head path forwards the backend body verbatim, see below).
- **Pre-head 4xx/5xx forwards the backend body verbatim** — `Response(content=decoder_response.content, status_code=...)` without re-parsing, preserving the original content-type and all JSON fields. `build_error_payload` runs only when the head is already committed and we must render an error body.
- **Two-layer fallback for non-standard error bodies** — `build_error_payload` returns the standard `{"error": {...}}` shape verbatim when present (preserving `param`/`code`/`type`), then falls back to wrapping a flat `{"message": ...}` body, then to the raw body. #13578's `decoder_error_payload` only handles the standard shape and otherwise drops to raw text.

**Tests** moved to `tests/ut/test_load_balance_proxy_errors.py` and registered in `test_config.yaml` (`cpu_only: true`). 8 cases, all passing: `build_error_payload` (verbatim / flat / 502 fallback), decode 4xx returns real 400 (stream + non-stream), 5xx retried, connection error → 502, streaming success replays first chunk once. Rebased onto latest `main` (resolves cleanly with the prefiller-cached-tokens changes from #13040).

Head is `5e3af3c`. Happy to adjust if you see gaps — and if maintainers still prefer landing #13578 instead, I'm glad to close this in favor of it.

examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
```diff
@@ -993,6 +1219,234 @@ async def reassign_instances(
...
+                if self._decoder_stream is None:
+                    self._decoder_stream = await self._open_stream()
+                response, chunk_iterator, first_chunk = self._decoder_stream
+                self._decoder_stream = None
+                if not response.is_success:
+                    # The ASGI response head is already committed here, so a
+                    # late decode error is forwarded inside the stream body
+                    # instead of leaving an empty 200.
+                    #
+                    # 非流式保护：若本轮 recompute 前已向客户端转发过 body（上一轮 200
```
lijiahang226: Translate the content into English.

examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
```diff
@@ -861,44 +1043,88 @@ async def send_request_to_service(
...
+    pre-read so that connection-level failures surface here (within retry
+    budget) instead of after ``StreamingResponse`` has started.
+    """
     headers = auth_headers(request_id)
-    max_attempts = max(1, max_retries)
-    for attempt in range(1, max_attempts + 1):
-        first_chunk_sent = False
+    # max(1, max_retries) 保证至少一次尝试：max_retries=0 时 range(1,1) 为空会跳过整个循环，
+    # 直接抛 RuntimeError 而不发任何请求（原 main 用 max(1, max_retries) 保证至少一次）。
+    for attempt in range(1, max(1, max_retries) + 1):
```
Yuli-yx: Plz use english in comments

examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
```diff
@@ -993,6 +999,54 @@ async def reassign_instances(
...
+    """Build a client-visible error body when decode fails mid-stream.
+
+    Minimal fix for the "empty 200" bug (issue #12166): ``StreamingResponse``
+    has already committed the 200 head before the decode request is sent, so a
+    decode-phase error can no longer change the HTTP status. Instead of dropping
+    the error into an empty body, forward it to the client as an SSE ``error``
+    event + ``data: [DONE]`` (streaming) or a JSON ``{"error": {...}}`` body
+    (non-streaming), preserving the upstream backend's original error message
+    when it is an ``HTTPStatusError``; otherwise synthesize a 502 Bad Gateway.
+    """
```
wangxiaoteng888: Please update the comment to only describe what this function does.

examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
```diff
@@ -1014,9 +1068,49 @@ async def handle_completions_impl(api: str, request: Request):
...
+                status_code=502,
+                content={
+                    "error": {
+                        "message": f"Decode backend request to {api} failed: {exc}",
+                        "type": "upstream_error",
+                        "code": "decode_backend_unavailable",
+                    }
+                },
+            )
+
```
wangxiaoteng888, line 1079: please add cacnel error

examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
```diff
@@ -1118,12 +1214,14 @@ async def release_prefill_kv_once() -> None:
...
                 logger.error(
-                    "Error during streaming from decoder %s:%s: %s while handling request %s; releasing prefiller KV",
+                    "Error during streaming from decoder %s:%s: %s while handling request %s; "
+                    "forwarding error to client",
                     instance_info.decoder_host,
                     instance_info.decoder_port,
                     exc,
                     instance_info.request_id,
                 )
+                yield _build_error_payload(exc, stream_flag)
```
Yuli-yx: `_build_error_payload()` appears out of scope for the minimal initial-error fix. Initial decode failures are already returned before StreamingResponse with the real HTTP status and body, while post-first-chunk transport failures are still swallowed by stream_service_response(). For non-streaming responses, appending another JSON error after partial output can also produce an invalid body. Could we remove this late-error rendering from this PR and handle mid-stream error semantics separately with dedicated tests?

examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
```diff
@@ -1035,15 +1129,22 @@ async def release_prefill_kv_once() -> None:
...
-                    decoder_client = await runtime.get_client(ServerRole.DECODE, instance_info.decoder_key)
-                    async for chunk in stream_service_response(
-                        decoder_client,
-                        api,
-                        req_data,
-                        request_id=instance_info.request_id,
-                        max_retries=args.max_retries,
-                        base_delay=args.retry_delay,
-                    ):
+                    if preopened_gen is not None:
```
Yuli-yx, line 1102: This first-chunk pre-read and replay approach is acceptable as a short-term fix for forwarding the initial decode error before the ASGI response head is committed. However, it couples response opening with body consumption and requires state to be passed between `handle_completions_impl()` and `generate_stream()`. As a follow-up, may we consider a follow-up refactor that explicitly separates opening the decode response, consuming its body iterator, and closing the response? We can work on that later once unit test is ready for example proxy.

Yuli-yx (approved):

wangxiaoteng888 (approved):

Tflowers-0129 (approved):

xlshaoscu:
Credit to @c5566b for this fix — he proposed an instructive approach that guided the bug fix. Thanks!

PR #12173, opened 2026-07-16
[BugFix][RFork] Synchronize NPU before quantized weight transfer

yangsonglin13:

This PR synchronizes the NPU after post-load tensor layout conversion and before RFork starts transferring quantized model weights.

Qwen3-32B-W8A8 can reproduce this issue: `process_weights_after_loading()` may enqueue asynchronous NPU layout-conversion operations, while RFork immediately proceeds to `pre_transfer()` and exposes the converted buffers for transfer. Without an explicit synchronization boundary, RFork may transfer weights before the layout conversion has completed, resulting in corrupted model output that appears as garbled responses.

Calling `torch.npu.synchronize()` at this boundary ensures that RFork transfers fully converted weights.

Yes. RFork loading for quantized models such as Qwen3-32B-W8A8 no longer risks producing garbled responses due to incomplete asynchronous weight layout conversion.

- `python -m py_compile vllm_ascend/model_loader/rfork/rfork_loader.py`
- `git diff --check`

- vLLM version: v0.24.0

### DIFF
--- a/vllm_ascend/model_loader/rfork/rfork_loader.py
+++ b/vllm_ascend/model_loader/rfork/rfork_loader.py
@@ -268,6 +268,9 @@ class RForkModelLoader(BaseModelLoader):
                 if processed_layout_transfer:
                     logger.info("RFork uses post-load tensor layout transfer for quantized model.")
                     process_weights_after_loading(model, model_config, target_device)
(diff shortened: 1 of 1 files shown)


yangsonglin13:
Superseded by #12995, which consolidates this NPU synchronization fix with the related RFork fallback and runtime-layout fixes into one PR.

PR #12193, opened 2026-07-16
[Performance][DSA-CP] Optimize local token metadata computation with fused Triton kernel and caching

Csrayz:

In DSA context-parallel mode, `AscendDSACPMetadataBuilder.build()` runs once per attention group within a step. Under DeepSeek-V4's dsv4 attention, there can be 8 build() per step, and each build independently recomputes metadata from scratch.

Across all groups, the following data are identical:

- `query_start_loc`, `seq_lens`, `num_reqs`, `num_input_tokens`
- Clamp + cumsum + offset + mask results
- RoPE local slices
- `start_pos`

This PR caches these intermediate results so that only the first group computes them; subsequent groups reuse the cached values. Additionally, the fused `_build_local_metadata_npu_kernel` is introduced to reduce kernel launch overhead.

The metadata building phase dropped from approximately 23ms to 10ms per step (measured on DeepSeek-V4 Flash, mtp=1).

before (main)

c5fa490ea22b40dbaafa4c2dab1f760d_IMAGE_1784202330638

after (this pr)

7281c93267ea4205b34526f9fb6ebd3a_IMAGE_1784202321384

**Changes summary**

- New `_ensure_gpu_local_metadata` method that computes GPU local metadata once and caches the result in `common_ratio_to_sas_metadata["_gpu_local"]`.
- Cross-group caching for CPU local metadata (`"_cpu_local"`) and RoPE local slices (`"_rope_local"`).
- New Triton kernel `_build_local_metadata_npu_kernel` that fuses clamp, cumsum, offset, mask, and start_pos into a single kernel launch.
- Refactored `build_req_metadata` to use the cached results and inlined RoPE slicing.
- Cleaned up `_build_local_token_metadata` by removing unused parameters (`input_positions`, `use_cache`, `seq_lens_q`).

No. Behavior is identical.

- Added a unit test for `RopeDataProxy.pad_to`.
- GMS8K evaluation results passed.

```
before (main)
+---------+-----------+----------+----------+-------+---------+---------+
| Model   | Dataset   | Metric   | Subset   |   Num |   Score | Cat.0   |
+=========+===========+==========+==========+=======+=========+=========+
| dsv4-f    | gsm8k     | mean_acc | main     |  1319 |  0.9682 | default |
+---------+-----------+----------+----------+-------+---------+---------+ 
```

aftet (this pr)

```
+---------+-----------+----------+----------+-------+---------+---------+
| Model   | Dataset   | Metric   | Subset   |   Num |   Score | Cat.0   |
+=========+===========+==========+==========+=======+=========+=========+
| dsv4-f    | gsm8k     | mean_acc | main     |  1319 |  0.9712 | default |
+---------+-----------+----------+----------+-------+---------+---------+ 
```

- vLLM version: v0.25.1

### DIFF
--- a/vllm_ascend/attention/context_parallel/dsa_cp.py
+++ b/vllm_ascend/attention/context_parallel/dsa_cp.py
@@ -8,6 +8,7 @@ import torch.nn.functional as F
 import torch_npu
 from vllm.config import VllmConfig, get_current_vllm_config
 from vllm.distributed import get_tp_group
+from vllm.triton_utils import HAS_TRITON, triton
 from vllm.v1.attention.backend import AttentionCGSupport, AttentionMetadataBuilder
 from vllm.v1.kv_cache_interface import AttentionSpec
 
@@ -25,7 +26,8 @@ from vllm_ascend.device.device_op import DeviceOperator
 from vllm_ascend.distributed.utils import all_gather_async
 from vllm_ascend.memcache_comm_fence import record_attention_compute_start
 from vllm_ascend.ops.linear import AscendUnquantizedLinearMethod
-from vllm_ascend.ops.rope_dsv4 import get_cos_and_sin_dsa, get_full_cos_and_sin_dsa
+from vllm_ascend.ops.rope_dsv4 import RopeDataProxy, get_cos_and_sin_dsa, get_full_cos_and_sin_dsa
+from vllm_ascend.ops.triton.dsa_cp import build_local_metadata_triton
 from vllm_ascend.quantization.methods.w8a8_dynamic import AscendW8A8DynamicLinearMethod
 from vllm_ascend.utils import (
     AscendDeviceType,
@@ -151,7 +153,7 @@ class AscendDSACPMetadataBuilder(AttentionMetadataBuilder[AscendDSAMetadata]):
     # Does this backend/builder support ACL Graphs for attention (default: no).
     aclgraph_support: ClassVar[AttentionCGSupport] = AttentionCGSupport.UNIFORM_BATCH
     hadamard = None
-    start_pos_prefill: torch.Tensor | None = None
+    start_pos_prefill: torch.Tensor
     req_sas_metadata: torch.Tensor
     req_qli_metadata: torch.Tensor
     block_size: int = 128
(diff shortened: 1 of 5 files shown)


vllm_ascend/attention/context_parallel/dsa_cp.py
```diff
@@ -733,51 +837,58 @@ def _build_local_token_metadata(
...
-        if local_query_start_loc is not None:
-            local_query_start_loc[1 : num_reqs + 1] = torch.cumsum(local_query_lens, dim=0)
-        else:
-            local_query_start_loc = torch.cat(
-                [
-                    torch.tensor([0], dtype=local_query_lens.dtype, device=local_query_lens.device),
-                    torch.cumsum(local_query_lens, dim=0),
-                ],
-                0,
+        if query_start_loc.device.type != "cpu" and HAS_TRITON and num_reqs <= 1024:
```
linfeng-yuan: 1. Why we should set a magic number 1024 here?
2. PTAL @ZT-AIA 
3.  Move triton implementation to triton dir.
4.  add e2e cases

linfeng-yuan:
@pisceskkk PTAL Threre's a performance optimization for dsa_cp with dsv4.

vllm_ascend/attention/context_parallel/dsa_cp.py
Csrayz: 1. 1024 was just a conservative upper bound; our actual batch size never reaches it, so the check is always true and removing it is safe.
2. ~
3. done
4. done

Csrayz:
> @pisceskkk PTAL Threre's a performance optimization for dsa_cp with dsv4.

@pisceskkk, sorry to ping again. I have another fix for dsa_cp in the MTP scenario, which modifies the same function and will cause merge conflicts. Could you please take a look at the current performance optimization PR? Thanks!

vllm_ascend/attention/context_parallel/dsa_cp.py
```diff
@@ -554,66 +551,135 @@ def _num_compressor_metadata_rows(
         num_tokens = self.num_actual_tokens
         return min(num_tokens, num_tokens // self.compressor_ratio + common_attn_metadata.num_reqs)
 
+    def _ensure_gpu_local_metadata(
```
pisceskkk: I think this should be renamed as `_ensure_device_local_metadata`

pisceskkk (approved):
Sorry for later reply.  LGTM, leave a little suggestion.

vllm_ascend/attention/context_parallel/dsa_cp.py
Csrayz: Sure, I'll push a new commit shortly to address it.

Also, just a heads-up: I'll be submitting a few more optimizations and bug fixes for dsa_cp in the coming days. I'd really appreciate your review on those as well if you have the bandwidth. Thanks in advance!s

Csrayz:
@kunpengW-code, linfeng-yuan is currently unreachable. This PR's CI is all green and pisceskkk has already approved. Could you take a look and approve/merge when you get a chance?

linfeng-yuan (approved):

PR #12208, opened 2026-07-16
[v0.23.0][Performance] Use CPU slot mapping for regular decode

Liuchenbing-2026:
## Summary

Reuse the existing CPU slot mapping implementation for regular non-PCP decode to avoid the Triton slot mapping kernel overhead on the `releases/v0.23.0` baseline.

## Motivation

The Triton slot mapping kernel path introduced upstream adds meaningful overhead for regular (non-PCP) decode steps. Since the CPU path is already implemented and well-validated, routing regular decode through it removes the extra kernel launch and index-math cost without any correctness change.

## Changes

- `vllm_ascend/worker/model_runner_v1.py` (+3 / -6): route regular decode slot mapping to the existing CPU path.

## Validation

- Non-PCP decode functional tests pass.
- Slot mapping output matches the Triton kernel path bit-for-bit on the covered shapes.

## Test plan

- [ ] Run e2e single-card Qwen tests
- [ ] Confirm no regression in decode throughput on the target SoC
- [ ] Verify PCP path is untouched

---

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -1229,13 +1229,10 @@ class NPUModelRunner(GPUModelRunner):
         if self._needs_seq_lens_cpu_sync and async_spec_decode_active:
             self._correct_optimistic_seq_lens_cpu(num_reqs)
 
-        # For non-PCP, compute slot_mapping on GPU. PCP slot_mapping was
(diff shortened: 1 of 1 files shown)


Liuchenbing-2026:
Closing — v0.23.0 CPU slot mapping fix is no longer needed on the current release line.

PR #12230, opened 2026-07-17
[Misc][Ops][Sampler] use torch_npu top-k top-p op

linfeng-yuan:
Replace the AscendC custom operator call with the torch_npu API while preserving the existing A2 and A3 dispatch behavior.

No.

Tests with original op test of AscendC Implementation with torch_npu API.

- vLLM version: v0.24.0


linfeng-yuan:
#12232  is merged. So we can close this PR.

PR #12237, opened 2026-07-17
[BugFix] Disable multistream_overlap_shared_expert when enable_fused_mc2 is enabled

kunpengW-code:
### What this PR does / why we need it?
This PR disables `multistream_overlap_shared_expert` when `enable_fused_mc2` is enabled. These two features cannot be enabled at the same time, so we automatically set `multistream_overlap_shared_expert` to `False` and log a warning if both are configured. 

Cause:
1. When fused mc2 is enabled, the shared expert multi-stream feature cannot be used with CV parallelism, and there is no performance gain.
2. When fused mc2 is enabled, multistream_overlap_shared_expert (shared expert multi-stream) is enabled, but enable_shared_expert_dp (shared expert DP) is not enabled.
As a result, the MOE large fusion operator and the TP communication (AIV mode) of the shared expert exist on two streams respectively. If one card enters the MOE large fusion operator and the other card enters the TP communication, the two operators will be deadlocked. As a result, the system is suspended.

Co-authored-by: guanguan0308 <1546542263@qq.com>
### Does this PR introduce _any_ user-facing change?
no
### How was this patch tested?

- vLLM version: v0.24.0
- vLLM main: https://github.com/vllm-project/vllm/commit/85c09e9885e346ea1612da30ebff5a75f67d2350

PR #12245, opened 2026-07-17
[Cherry-pick][releases/v0.23.0] [BugFix] Disable multistream_overlap_shared_expert when enable_fused_mc2 is enabled

kunpengW-code:
Cherry-pick of PR #12237 onto releases/v0.23.0.

This PR disables `multistream_overlap_shared_expert` when `enable_fused_mc2` is enabled. These two features cannot be enabled at the same time, so we automatically set `multistream_overlap_shared_expert` to `False` and log a warning if both are configured. 

Cause:
1. When fused mc2 is enabled, the shared expert multi-stream feature cannot be used with CV parallelism, and there is no performance gain.
2. When fused mc2 is enabled, multistream_overlap_shared_expert (shared expert multi-stream) is enabled, but enable_shared_expert_dp (shared expert DP) is not enabled.
As a result, the MOE large fusion operator and the TP communication (AIV mode) of the shared expert exist on two streams respectively. If one card enters the MOE large fusion operator and the other card enters the TP communication, the two operators will be deadlocked. As a result, the system is suspended.

Co-authored-by: guanguan0308 <1546542263@qq.com>
no

- vLLM version: v0.23.0

### DIFF
--- a/docs/source/tutorials/models/GLM5.2.md
+++ b/docs/source/tutorials/models/GLM5.2.md
@@ -674,7 +674,7 @@ Before you start, please
             --served-model-name glm-52 \
             --max-model-len 135000 \
             --speculative-config '{"num_speculative_tokens": 1, "method":"deepseek_mtp", "enforce_eager": true}' \
-            --additional-config '{"enable_sparse_c8":false, "multistream_overlap_shared_expert": true, "enable_dsa_cp": true}' \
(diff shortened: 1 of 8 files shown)


kunpengW-code:
/rerun
[Bot]: rerun completed.

**Failed:**

linfeng-yuan (approved):

PR #12252, opened 2026-07-17
[Ops][BugFix] Fix parent block hash chain exception in AscendStore KV event

zh98530:
Fixes issue：https://github.com/vllm-project/vllm-ascend/issues/12002
This PR fixes an exception in the parent block hash chain for `AscendStore` KV events. Previously, `prev_key` was tracked across loop iterations of only the missing blocks being stored. This caused incorrect parent-child relationships or exceptions when some blocks in the sequence were already cached and thus skipped.
This change resolves the issue by pre-computing all block hashes for the group (`all_hashes`) and looking up the parent block hash directly using the block index (`block_idx - 1`), ensuring correct parent-child chain tracking.

No. This is an internal bugfix for KV transfer event tracking on Ascend devices.

Tested with existing unit tests and integration tests for KV transfer.

- vLLM version: v0.23.0


Pz1116 (approved):
LGTM

MengqingCao (approved):

PR #12295, opened 2026-07-17
[Feature][LoRA] Qwen3.5 hybrid-dense and shared-experts MoE LoRA with isolated ACL compile graphs

Liuchenbing-2026:

Enable and stabilize LoRA on the Qwen3.5 family in vllm-ascend across both eager and full-graph modes, and isolate base-model / adapter ACL compile graphs to avoid replay explosions when the adapter set changes.

The PR is three stacked commits:

1. **`[LoRA][MoE] Support MoE LoRA on models with shared experts (opt-in)`** — vllm-ascend MoE LoRA currently rejects models with shared experts (Qwen3.5-MoE, DeepSeek-V3): the shared-experts LoRA runs through vLLM's dense expand-slice path, whose stacked `lora_b` layout does not match `_C_ascend.sgmv_expand`, and the `torch_ops` einsum fallback fails on the same layout. Add an opt-in path (`VLLM_ASCEND_MOE_LORA_ALLOW_SHARED_EXPERTS=1`) that routes `PunicaWrapperNPU._expand_slice_{prefill,decode}` through a `torch.bmm` fallback (per-token gather + bmm via a mutating custom op that stays opaque to Dynamo fullgraph). Default off — plain MoE (e.g. Qwen3-30B-A3B) keeps the fused `_C_ascend` path.

2. **`[LoRA] Enable Qwen3.5 hybrid dense LoRA (eager + full-graph)`** — Qwen3.5 dense variants wrap the text tower under a `language_model` submodule and mix self-attention with GatedDeltaNet (GDN) linear-attention layers. Three issues stop LoRA from working; this fixes all three:
   - **Zero LoRA delta (name mapping).** Adapter tensors are keyed by `model.layers.N...` while the model registers LoRA modules under `language_model.model.layers.N...`. The lookup misses, `set_lora` gets a zero buffer, delta is zero. A worker patch auto-detects the wrapper prefix from `LoRAModelManager.modules` and remaps the `loras` dict keys right after `WorkerLoRAManager._load_adapter`, before registration/packed-merge. Keys are left untouched when already aligned, so plain models are unaffected. Guarded by `VLLM_ASCEND_LORA_VLM_PREFIX` (default on).
   - **Graph-capture crash.** Graph replay walks the full `attn_metadata` key list and reaches a GDN layer whose `GDNAttentionMetadata` has no `seq_lens_list` / `actual_seq_lengths_q`, raising `AttributeError`. Filter the replay key list down to the captured self-attention (FIA) layers.
   - **Graph-replay crash under MTP.** With spec-decode sequences present, the builder folds plain decodes into prefills and sets `num_decodes = 0`, so `non_spec_decode_fallback_meta` is `None`; the non-spec-decode conv1d graph update then fetches host args off the missing meta and kills the engine. Treat the task as a state no-op (`cache_indices = -1`) when the fallback meta is absent, mirroring the spec branch.

3. **`[LoRA] Isolate base and adapter ACL compile graphs`** — under full-graph mode the base-model and LoRA-adapter compile artifacts share one ACL graph, so any adapter set change contaminates the base graph and forces spurious recompiles / replay failures. Patch the ACL compile wrapper so base-model and adapter graphs are captured and replayed independently, keeping the base graph stable across adapter swaps.

Two new env switches, both guarding new behavior:

- `VLLM_ASCEND_LORA_VLM_PREFIX` (default **on**) — auto-remap adapter keys when the model is wrapped under `language_model.`. Off restores upstream behavior for plain models; keys already aligned are left untouched, so plain models are unaffected in either state.
- `VLLM_ASCEND_MOE_LORA_ALLOW_SHARED_EXPERTS` (default **off**) — opt-in `torch.bmm` fallback for MoE LoRA on shared-experts models. **Experimental** — throughput cost vs the fused NPU op.

No change for existing non-LoRA paths, plain LoRA models without a `language_model.` wrapper, or plain MoE LoRA (without shared experts).

### How was it tested?

- New UT `tests/ut/patch/worker/test_patch_lora_compile_wrapper.py` covers the base/adapter compile-graph isolation.
- E2E manual runs:
  - Qwen3.5-27B dense + LoRA in eager mode: adapter delta > 0, expected output.
  - Qwen3.5-27B dense + LoRA in full-graph mode: no capture/replay crash on GDN layers.
  - Qwen3.5-27B dense + LoRA + MTP: no null-fallback-meta crash.
  - Qwen3.5-MoE MoE LoRA with `VLLM_ASCEND_MOE_LORA_ALLOW_SHARED_EXPERTS=1`: expected output (throughput drop expected).
  - Qwen3-30B-A3B MoE LoRA with the flag off: no regression — still uses fused `_C_ascend`.

### Scope

- **A2 / A3 (910b / 910c)**: covered.
- **310P**: out of scope for the compile-graph isolation and the GDN patch (the 310P worker takes a different LoRA path).

Co-authored-by: zhaoyue <wzy85@local>

- vLLM version: v0.26.0


SkychenLee:
https://github.com/vllm-project/vllm-ascend/pull/11702 seems you have same feature confilct

PR #12296, opened 2026-07-17
Dspark FULL Graph support

StanislavII:
# Add DSpark FULL Decode Graph Support on vLLM Ascend

## Summary

This PR adds correct ACL/CUDA graph support for the DSpark speculative drafter, including `FULL_DECODE_ONLY` execution.

DSpark differs from existing EAGLE/DFlash proposers because the target and draft forwards use different per-request widths:

```text
Target verification width: K + 1
DSpark draft width:        K
```

For example, with `num_speculative_tokens = 7`:

```text
B = 1:
  target graph = 8 tokens
  draft graph  = 7 tokens

B = 16:
  target graph = 128 tokens
  draft graph  = 112 tokens
```

Previously, the DSpark drafter reused graph descriptors and capture sizes generated for the target model. This caused incorrect request-count calculation, missing graph captures, FIA padding failures, and incorrect DSpark context-KV capture.

This PR introduces a separate DSpark graph-dispatch path and ensures that target and drafter graphs are captured and dispatched using their respective token widths.

### 1. Target graph descriptors were incorrectly reused by DSpark

The target dispatcher is initialized with:

```text
uniform_decode_query_len = K + 1
```

A target FULL descriptor therefore interprets:

```text
num_reqs = num_tokens / (K + 1)
```

DSpark processes only `K` query tokens per request. Reusing the target descriptor for the drafter therefore produces an incorrect request count.

For example:

```text
DSpark tokens = 112
Correct interpretation: 112 / 7 = 16 requests
Target interpretation:  112 / 8 = 14 requests
```

This could result in:

- incorrect graph selection;
- `graph_num_reqs < actual_num_reqs`;
- incorrect FIA `query_start_loc` padding;
- missing graph replay;
- runtime graph capture attempts.

### 2. DSpark FULL graph shapes were not independently registered

The upstream proposer initializes drafter graph keys primarily for PIECEWISE execution and shares the target compilation configuration.

DSpark requires its own dispatcher configuration:

```text
Target dispatcher:
  query width = K + 1
  FULL shapes = B × (K + 1)

DSpark dispatcher:
  query width = K
  FULL shapes = B × K
```

Without this separation, the runtime dispatcher may select a descriptor that does not correspond to a physically captured DSpark graph.

### 3. DSpark graph capture used the wrong context length

DSpark uses:

```text
Draft query length: B × K
Context hidden states: B × (K + 1)
```

During dummy graph capture, `_dflash_num_context` was incorrectly overwritten with `num_input_tokens`, which is the draft query length `B × K`.

As a result, the captured graph precomputed context KV using only `B × K` hidden states instead of the full target context `B × (K + 1)`.

This did not necessarily crash, but reduced FULL-mode draft quality and acceptance.

The fix preserves:

```python
self._dflash_num_context = dspark_num_context_tokens
```

where:

```text
dspark_num_context_tokens = B × (K + 1)
```

The inherited DFlash input builder can then precompute context KV from the complete target hidden-state block.

### 4. DSpark draft-specific attention metadata was overwritten

`AscendDsparkProposer.set_inputs_first_pass()` prepares draft attention metadata using:

```text
query length = K
seq_lens = effective target seq_lens + K
causal = False
attention state = ChunkedPrefill
```

The generic FULL proposer path replaced `common_attn_metadata.seq_lens` with `runner.seq_lens`, discarding the DSpark-specific `+K` update.

The FULL path now preserves the metadata created by DSpark, similarly to DFlash:

```python
if self.method == "dflash" or is_dspark:
    common_attn_metadata.seq_lens = self._adjust_tensor(
        common_attn_metadata.seq_lens,
        num_reqs_padded,
    )
```

### 5. Dynamic graph inputs must use stable buffers

DSpark graph execution now uses preallocated buffers for graph-sensitive inputs, including:

- seed token IDs;
- draft token IDs;
- confidence logits;
- verification lengths;
- token indices to sample.

In particular, `token_indices_to_sample` now uses the proposer-owned static buffer rather than allocating a new tensor every iteration.

This keeps runtime input pointers aligned with graph-capture input pointers.

### `AscendDsparkProposer`

A DSpark-specific `initialize_cudagraph_keys()` override was added.

The override:

- creates an independent copy of the compilation configuration;
- sets the DSpark uniform decode query width to `K`;
- generates DSpark graph token sizes as `B × K`;
- avoids mutating the target model dispatcher;
- preserves the configured graph mode, including `FULL_DECODE_ONLY`;
- initializes independent FULL graph descriptors for the drafter.

Conceptually:

```python
target_capture_sizes = [
    batch_size * (K + 1)
    for batch_size in capture_batch_sizes
]

draft_capture_sizes = [
    batch_size * K
    for batch_size in capture_batch_sizes
]
```

### `ModelRunnerV1._check_and_update_cudagraph_mode`

The graph initialization path now initializes both dispatchers:

```text
Target dispatcher → target graph descriptors
DSpark dispatcher → draft graph descriptors
```

Target and draft graph parameters are configured separately:

```python
set_graph_params(target_capture_sizes)
set_draft_graph_params(draft_capture_sizes)
```

This prevents target and draft graph sizes from being treated as the same shape domain.

### `LLMBaseProposer._propose`

DSpark runtime dispatch now uses the drafter dispatcher directly instead of dispatching through the target descriptor and repairing the result afterward.

Dispatch is performed both:

1. before data-parallel graph-size synchronization;
2. after data-parallel synchronization.

The second dispatch selects the final graph descriptor after any required padding.

For FULL mode, the returned descriptor is expected to satisfy:

```python
batch_descriptor.num_tokens == (
    batch_descriptor.num_reqs * K
)
```

The descriptor request count is then used when padding FIA metadata.

### `ModelRunnerV1._dummy_run`

During target graph capture, the corresponding DSpark draft graph uses:

```text
drafter_num_reqs   = target padded request count
drafter_num_tokens = drafter_num_reqs × K
context tokens     = drafter_num_reqs × (K + 1)
```

For example:

```text
Target descriptor:
  num_tokens = 128
  num_reqs   = 16

DSpark descriptor:
  num_tokens = 112
  num_reqs   = 16

DSpark context:
  num_context_tokens = 128
```

The target context length is passed separately to `AscendDsparkProposer.dummy_run()`.

### `AscendDsparkProposer.dummy_run`

The dummy run now distinguishes between:

```text
num_input_tokens          = B × K
dspark_num_context_tokens = B × (K + 1)
```

The graph captures:

- context-KV precomputation over `B × (K + 1)` target hidden states;
- draft-model forward over `B × K` query tokens;
- DSpark non-causal draft attention metadata.

## Runtime Validation

The target and drafter now independently select FULL graphs with the expected shapes.

For `K = 7`, `B = 16`:

```text
Target:
  mode              = FULL
  num_tokens        = 128
  num_reqs          = 16
  query_len_per_req = 8

DSpark:
  mode              = FULL
  num_tokens        = 112
  num_reqs          = 16
  query_len_per_req = 7
```

For `B = 1`:

```text
Target:
  mode              = FULL
  num_tokens        = 8
  num_reqs          = 1

DSpark:
  mode              = FULL
  num_tokens        = 7
  num_reqs          = 1
```

The DSpark graph capture also uses the expected context/query split:

```text
B = 16:
  context tokens = 128
  query tokens   = 112

B = 1:
  context tokens = 8
  query tokens   = 7
```

## GSM8K End-to-End Validation

The implementation was evaluated on the full GSM8K test set with 1,300 prompts using Qwen3-8B as the target model, the DSpark block-7 drafter, and:

```python
num_speculative_tokens = 7
```

For this validation, all paired target and DSpark token shapes for batch sizes `1..16` were included:

```python
compilation_config = {
    "cudagraph_capture_sizes": sorted(
        set(
            shape
            for batch_size in range(1, 17)
            for shape in (
                7 * batch_size,  # DSpark draft: B * K
                8 * batch_size,  # target verify: B * (K + 1)
            )
        )
    ),
}
```

This exhaustive paired-shape configuration was used to validate correctness and avoid the remaining sparse-shape dispatch issue during testing. It is not intended as the final memory-efficient capture policy.

### Results

| Metric | PIECEWISE | FULL_DECODE_ONLY |
|---|---:|---:|
| Prompts | 1,300 | 1,300 |
| Accuracy | 0.8438 | 0.8477 |
| Invalid rate | 0.0000 | 0.0000 |
| Wall time | 193.97 s | 185.56 s |
| Queries/s | 6.70 | 7.01 |
| Input throughput | 595.29 tokens/s | 622.44 tokens/s |
| Output throughput | 2,042.87 tokens/s | 2,135.38 tokens/s |
| Number of draft blocks | 62,416 | 62,336 |
| Number of draft tokens | 436,912 | 436,352 |
| Accepted tokens | 332,483 | 332,424 |
| Mean acceptance length | 6.33 | 6.33 |

### Per-Position Acceptance

| Draft position | PIECEWISE | FULL_DECODE_ONLY |
|---:|---:|---:|
| 0 | 0.94 | 0.94 |
| 1 | 0.88 | 0.88 |
| 2 | 0.81 | 0.81 |
| 3 | 0.75 | 0.76 |
| 4 | 0.70 | 0.70 |
| 5 | 0.65 | 0.65 |
| 6 | 0.60 | 0.60 |

The acceptance distributions are effectively identical between PIECEWISE and FULL execution. Both modes obtain a mean acceptance length of `6.33`, showing that FULL graph replay preserves DSpark draft quality and attention semantics.

The small accuracy difference corresponds to approximately five GSM8K examples out of 1,300 and does not indicate a systematic FULL-graph quality regression. Both modes produced a zero invalid-output rate.

Compared with PIECEWISE, `FULL_DECODE_ONLY` achieved:

- approximately **4.3% lower wall-clock time**;
- approximately **4.6% higher query throughput**;
- approximately **4.6% higher input-token throughput**;
- approximately **4.5% higher output-token throughput**.

### Validation Conclusion

These results verify that:

- target verification runs with graph width `K + 1`;
- DSpark drafting runs with graph width `K`;
- FULL graph context-KV precomputation uses all `B × (K + 1)` target hidden states;
- PIECEWISE and FULL produce equivalent draft acceptance behavior;
- `FULL_DECODE_ONLY` provides a measurable end-to-end performance improvement without reducing acceptance quality.

The exhaustive capture configuration used for this test includes all `B × K` and `B × (K + 1)` shapes for `B = 1..16`.

## Known Issue: Sparse Graph Capture Sizes

This PR does not yet fully solve sparse target/drafter graph-shape coordination.

Currently, DSpark graph sizes may be generated for every request count:

```python
draft_capture_sizes = [
    batch_size * K
    for batch_size in range(1, max_num_reqs + 1)
]
```

However, due to graph memory consumption, users may configure only a sparse subset of physical graph captures.

For example, the physically captured request buckets may be:

```python
capture_batch_sizes = [
    1, 2, 4, 8, 16, 32, 60, 61, 62, 63, 64
]
```

For `K = 7`, these correspond to:

```text
Target shapes:
8, 16, 32, 64, 128, 256, 480, 488, 496, 504, 512

DSpark shapes:
7, 14, 28, 56, 112, 224, 420, 427, 434, 441, 448
```

If the DSpark dispatcher advertises every `B × K` key but only the sparse subset was physically captured, it may select an uncaptured exact shape.

For example, with 58 requests:

```text
Actual target tokens = 58 × 8 = 464
Actual draft tokens  = 58 × 7 = 406
```

The desired behavior is to pad both models to the same request bucket, for example `B = 60`:

```text
Target: 464 → 480
Draft:  406 → 420
```

Instead, if the DSpark dispatcher contains a key for `406`, it may select that uncaptured graph. The graph wrapper then attempts lazy graph capture during inference and fails with:

```text
RuntimeError: CUDA graph capturing detected at an inappropriate time.
```

This failure has been observed with sparse capture configurations.

### Current Workaround

Including all target and draft token shapes in `cudagraph_capture_sizes` avoids the missing-shape error:

```python
sorted({
    value
    for batch_size in range(1, max_batch_size + 1)
    for value in (
        batch_size * K,
        batch_size * (K + 1),
    )
})
```

However, this captures unnecessary graphs and can consume too much graph memory.

It is also not the desired final design for `FULL_DECODE_ONLY`, because target and DSpark FULL graphs belong to different dispatchers.

### Planned Follow-Up

The proper solution is to use a shared list of captured **request-count buckets**:

```python
capture_batch_sizes = [...]
```

and derive both shape sets from it:

```python
target_capture_sizes = [
    batch_size * (K + 1)
    for batch_size in capture_batch_sizes
]

draft_capture_sizes = [
    batch_size * K
    for batch_size in capture_batch_sizes
]
```

The following three sets must remain synchronized for each model:

```text
Dispatcher keys
Graph parameters passed to the capture manager
Physically captured graph sizes
```

For PIECEWISE mode, graph sizes represent generic total-token buckets and do not encode a fixed request count. PIECEWISE target/drafter shape coordination therefore requires separate handling and should not assume the `B × width` relationship used by FULL decode graphs.

## Scope

This PR focuses on:

- enabling correct DSpark FULL graph capture and replay;
- separating target and drafter graph descriptors;
- preserving DSpark context and attention semantics;
- restoring correct draft acceptance under graph execution;
- validating numerical quality and end-to-end performance on GSM8K.

### DIFF
--- a/vllm_ascend/spec_decode/dspark_proposer.py
+++ b/vllm_ascend/spec_decode/dspark_proposer.py
@@ -1,12 +1,18 @@
 from typing import Any
 
 import torch
+from copy import deepcopy
+
 from vllm.config import CUDAGraphMode, VllmConfig
+from vllm.forward_context import get_forward_context
 from vllm.v1.attention.backends.utils import CommonAttentionMetadata
 
-from vllm_ascend.ascend_forward_context import set_ascend_forward_context
+from vllm_ascend.ascend_forward_context import _EXTRA_CTX, set_ascend_forward_context
 from vllm_ascend.attention.attention_v1 import AscendAttentionState
-from vllm_ascend.ops.triton.spec_decode.utils import copy_and_expand_dflash_and_dspark_inputs_kernel_single_grid
+from vllm_ascend.attention.utils import AscendCommonAttentionMetadata
+from vllm_ascend.ops.triton.spec_decode.utils import (
+    copy_and_expand_dflash_and_dspark_inputs_kernel_single_grid,
+)
 from vllm_ascend.spec_decode.dflash_proposer import AscendDflashProposer
 
 
@@ -23,10 +29,112 @@ class AscendDsparkProposer(AscendDflashProposer):
             runner=runner,
         )
 
-        # Initialize and establish static address for graph mode
+        #   Static buffers for graph mode.
+        #   DSpark returns K draft tokens, but internally keeps seed + K tokens.
         blk = 1 + self.num_speculative_tokens
-        self._dspark_seed_buffer = torch.zeros(self.max_batch_size, dtype=torch.int64, device=device)
-        self._dspark_draft_buffer = torch.zeros((self.max_batch_size, blk), dtype=torch.int64, device=device)
+        self._dspark_seed_buffer = torch.zeros(
+            self.max_batch_size,
+            dtype=torch.int64,
+            device=device,
+        )
+        self._dspark_draft_buffer = torch.zeros(
+            (self.max_batch_size, blk),
+            dtype=torch.int64,
+            device=device,
+        )
+        self._dspark_confidence_logits_buffer = torch.zeros(
+            (self.max_batch_size, self.num_speculative_tokens),
+            dtype=torch.float32,
+            device=device,
+        )
+        self._dspark_num_verify_tokens_buffer = torch.zeros(
+            self.max_batch_size,
+            dtype=torch.int32,
+            device=device,
+        )
+        self._keep_lens = torch.zeros(
+            (self.max_batch_size,),
+            dtype=torch.int32,
+            device=self.device,
+        )
+        self._dspark_updates = torch.ones(
+            (self.max_batch_size * self.num_speculative_tokens,),
+            dtype=torch.int32,
+            device=self.device,
+        )
+
+    def initialize_cudagraph_keys(
+        self,
+        cudagraph_mode: CUDAGraphMode,
+    ) -> None:
+        """Initialize independent FULL graph keys for the DSpark drafter.
+        Target width: K + 1
+        DSpark width: K
+        """
+        dispatcher = self.cudagraph_dispatcher
+        draft_query_len = int(self.num_speculative_tokens)
+
+        #   Do not reuse the target model's CompilationConfig object directly.
+        #   The target dispatcher needs [K+1, 2(K+1), ...], while DSpark needs [K, 2K, ...].
+        draft_compilation_config = deepcopy(dispatcher.compilation_config)
+
+        max_num_reqs = int(
+            dispatcher.vllm_config.scheduler_config.max_num_seqs
+        )
+
+        target_max_capture_size = int(
+            dispatcher.compilation_config.max_cudagraph_capture_size
+        )
+
+        draft_capture_sizes = [
+            batch_size * draft_query_len
+            for batch_size in range(1, max_num_reqs + 1)
+            if batch_size * draft_query_len <= target_max_capture_size
+        ]
+
+        if not draft_capture_sizes:
+            raise RuntimeError(
+                "No valid DSpark cudagraph capture sizes: "
+                f"draft_query_len={draft_query_len}, "
+                f"max_num_reqs={max_num_reqs}, "
+                f"max_capture_size={target_max_capture_size}"
+            )
+
+        draft_compilation_config.cudagraph_capture_sizes = draft_capture_sizes
+        draft_compilation_config.max_cudagraph_capture_size = (
+            draft_capture_sizes[-1]
+        )
+
+        #   avoid validation against target compile sizes such as 8, 16, ...
+        if draft_compilation_config.compile_sizes:
+            draft_compilation_config.compile_sizes = list(
+                draft_capture_sizes
+            )
+
+        #   this dispatcher now owns a separate config from the target dispatcher.
+        dispatcher.compilation_config = draft_compilation_config
+
+        #    critical: _create_padded_batch_descriptor() reads this attribute,
+        #    not the initialize_cudagraph_keys() argument.
+        dispatcher.uniform_decode_query_len = draft_query_len
+
+        #    robust if initialization is invoked more than once.
+        for keys in dispatcher.cudagraph_keys.values():
+            keys.clear()
+        dispatcher.keys_initialized = False
+
+        if self.speculative_config.enforce_eager:
+            dspark_cudagraph_mode = CUDAGraphMode.NONE
+        else:
+            #   preserve FULL_DECODE_ONLY / PIECEWISE_AND_FULL semantics.
+            #   do not replace this with cudagraph_mode.mixed_mode(), because
+            #   doing so would discard the separate FULL decode routine.
+            dspark_cudagraph_mode = cudagraph_mode
+
+        dispatcher.initialize_cudagraph_keys(
+            dspark_cudagraph_mode,
+            uniform_decode_query_len=draft_query_len,
+        )
 
     def set_inputs_first_pass(
         self,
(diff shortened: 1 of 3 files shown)


StanislavII:
moved to https://github.com/vllm-project/vllm-ascend/pull/13124

PR #12283, opened 2026-07-17
[BugFix][v0.23.0][Worker] Restore pre-#12027 model runner dispatch

maoxx241:
### What this PR does / why we need it?

Revert the two model-runner dispatch changes introduced by #12027 on `releases/v0.23.0`.

- Restore `is_all_decode` to the pre-#12027 `num_computed_tokens > 0` check.
- Restore the non-speculative uniform-decode path to bypass the `is_all_decode` gate, as it did before #12027.
- Remove the model-runner test setup and completed-prompt regression case introduced with those dispatch changes.

This is a focused revert of the #12027 model-runner behavior and its associated test coverage. It does not include the broader #12255 changes.

### Does this PR introduce any user-facing change?

No API change. This restores the previous model-runner graph-dispatch behavior on the v0.23.0 release branch.

### How was this patch tested?

- `uvx ruff check tests/ut/worker/a2/test_model_runner_v1_with_device.py`
- `uvx ruff format --check tests/ut/worker/a2/test_model_runner_v1_with_device.py`
- `python3 -m py_compile vllm_ascend/worker/model_runner_v1.py tests/ut/worker/a2/test_model_runner_v1_with_device.py`
- `git diff --check`

The local macOS Python environment does not provide `pytest`; NPU runtime validation is delegated to PR CI.

- vLLM version: v0.23.0
- vLLM main: https://github.com/vllm-project/vllm/commit/ee0da84ab9e04ac7610e28580af62c365e898389

PR #12297, opened 2026-07-17
[v0.23.0][BugFix][Graph] Bypass decode-only replay for GDN prefill

maoxx241:

This is a minimal `releases/v0.23.0` fix for a Qwen3.5 GDN correctness boundary under `FULL_DECODE_ONLY`.

Graph dispatch is shape based. A final one-token chunk can therefore look like uniform decode even when `CommonAttentionMetadata.is_prefilling` still identifies it as prompt work. Replaying a decode-only graph in that state bypasses the GDN prefill/PCP path; in PCP, rank-local query lengths can be asymmetric (`1` on one rank and `0` on another), so normalizing the row to decode inside the GDN builder would change state-propagation semantics.

This PR leaves `model_runner_v1.py` unchanged. Instead, `ACLGraphWrapper` checks the already-built GDN metadata immediately before a `FULL_DECODE_ONLY` replay. If any GDN metadata still has `num_prefills > 0`, that invocation temporarily bypasses the full graph and compiled model, runs with eager prefill semantics, and restores the forward context afterward.

The guard is limited to:

- runtime `FULL` under configured `FULL_DECODE_ONLY`;
- GDN attention metadata; and
- batches that still contain semantic prefill work.

Ordinary GDN decode replay, non-GDN models, and configured `FULL` mode are unchanged. The performance cost is limited to the tail-prefill invocation that would otherwise use the wrong decode-only graph.

This preserves the non-MTP dispatch restored by #12283 and does not reclassify stateful one-token prefills as decodes.

No API or configuration change. It prevents a semantic GDN prefill from replaying a decode-only graph.

Added a focused ACL graph wrapper regression test that verifies:

- GDN metadata with `num_prefills > 0` bypasses `FULL_DECODE_ONLY` graph capture/replay;
- the eager runnable is used for that invocation;
- no graph entry is created; and
- `cudagraph_runtime_mode` and `skip_compiled` are restored afterward.

Local checks:

- `python3 -m py_compile vllm_ascend/compilation/acl_graph.py tests/ut/compilation/a2/test_acl_graph.py`
- `uvx ruff check vllm_ascend/compilation/acl_graph.py tests/ut/compilation/a2/test_acl_graph.py`
- `uvx ruff format --check vllm_ascend/compilation/acl_graph.py tests/ut/compilation/a2/test_acl_graph.py`
- `pre-commit run --files vllm_ascend/compilation/acl_graph.py tests/ut/compilation/a2/test_acl_graph.py`
- `git diff --check origin/releases/v0.23.0...HEAD`

All checks above passed. The local macOS environment does not provide `pytest`, `torch`, `vllm`, or an NPU runtime, so target CI and the exact Qwen3.5 nightly remain required before merge:

- `Qwen3.5-397B-A17B-w8a8-mtp-longseq`
- TP=8, PCP=2, DCP=4, MTP=3
- GPQA Diamond, 50 prompts
- `FULL_DECODE_ONLY`

Nightly status: the exact case was requested in the PR, but the
[`Handle /nightly Command` workflow](https://github.com/vllm-project/vllm-ascend/actions/runs/29613730469)
rejected the request because the initiator does not have `triage+` permission.
A maintainer-authorized trigger is still required.

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/compilation/a2/test_acl_graph.py
+++ b/tests/ut/compilation/a2/test_acl_graph.py
@@ -20,6 +20,7 @@ import torch
 from vllm.compilation.cuda_graph import CUDAGraphOptions
 from vllm.config import CUDAGraphMode, VllmConfig
 from vllm.forward_context import BatchDescriptor, ForwardContext
+from vllm.v1.attention.backends.gdn_attn import GDNAttentionMetadata
 
 from tests.ut.base import TestBase
 from vllm_ascend.attention.attention_v1 import AscendMetadata, AscendMetadataForDecode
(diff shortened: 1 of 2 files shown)


maoxx241:
/nightly Qwen3.5-397B-A17B-w8a8-mtp-longseq --branch releases/v0.23.0

PR #12336, opened 2026-07-18
[BugFix] Fix qsfa

lijiahang226:
  Fix QSFA (Quant Sparse Flash Attention) operator dispatch by routing through the device adaptor class hierarchy instead of hardcoding BaseDeviceAdaptor.

  Problem: The torch_npu C8 SFA binding only returns attention_out and rejects return_softmax_lse. However, DCP (Decoupled Context Parallel) needs the
  softmax max/sum tensors to build LSE for attention output merging. The _C_ascend custom op supports return_softmax_lse, but was previously only called on
  the LSE path — the non-LSE path incorrectly relied on torch_npu regardless of device type.

  Solution:
  - BaseDeviceAdaptor: Uses torch.ops._C_ascend.npu_kv_quant_sparse_flash_attention (custom op, LSE-capable). Always passes return_softmax_lse=True and
  strips the LSE output via _format_sparse_flash_attention_output when not needed.
  - A5DeviceAdaptor: Overrides execute_kv_quant_sparse_flash_attention with torch_npu.npu_kv_quant_sparse_flash_attention (no LSE support, matches A5's
  standard torch_npu binding). Raises RuntimeError if LSE is requested.
  - execute_sparse_flash_attention_process: Changed from @staticmethod to @classmethod, dispatches via cls.execute_kv_quant_sparse_flash_attention(...) so
  the correct device-specific implementation is invoked.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/device/device_op.py
+++ b/vllm_ascend/device/device_op.py
@@ -612,8 +612,9 @@ class BaseDeviceAdaptor:
             )
         return topk_indices
 
-    @staticmethod
+    @classmethod
     def execute_sparse_flash_attention_process(
+        cls,
         sfa_impl,
         ql_nope: torch.Tensor,
         q_pe: torch.Tensor,
@@ -641,7 +642,7 @@ class BaseDeviceAdaptor:
             torch.float8_e5m2,
         )
         if use_kv_quant_sparse_attention:
-            return BaseDeviceAdaptor.execute_kv_quant_sparse_flash_attention(
+            return cls.execute_kv_quant_sparse_flash_attention(
                 sfa_impl,
                 ql_nope,
                 q_pe,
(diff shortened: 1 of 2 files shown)


weiguihua2 (approved):

zhenwenqi2024 (approved):

vllm_ascend/device/device_op.py
```diff
@@ -1124,6 +1125,48 @@ def npu_fused_infer_attention_score(
             **kwargs,
         )
 
+    @staticmethod
+    def execute_kv_quant_sparse_flash_attention(
```
zzzzwwjj, line 1129: This interface is only called as an internal interface. Please add a "_" before the interface.

zzzzwwjj (approved):

PR #12342, opened 2026-07-18
[BugFix] Persist enable_npugraph_ex override for worker processes

ShuhaoZhangTony:
## Problem

When `platform.py` overrides `enable_npugraph_ex = False` for PIECEWISE or NONE cudagraph modes, the change only affects the main-process `AscendConfig`. Spawned workers reconstruct their configuration from `vllm_config.additional_config`, which can still contain the default value. Workers can therefore select a different compilation path from the main process.

## Fix

Persist the final `enable_npugraph_ex` value into `additional_config` after cudagraph-mode selection and before workers are spawned.

- Preserve unrelated `ascend_compilation_config` keys.
- Safely replace a missing, `None`, or non-dictionary nested config.
- Keep full-graph behavior unchanged when the final value is already true.

## Validation

- Added regression coverage for a missing top-level config, an empty config, a `None` nested config, and preservation of existing nested keys.
- `ruff check vllm_ascend/platform.py tests/ut/test_platform.py`
- `ruff format --check vllm_ascend/platform.py tests/ut/test_platform.py`
- `python3 -m compileall -q vllm_ascend/platform.py tests/ut/test_platform.py`

This is a configuration-correctness fix; it makes no throughput or latency claim.
- vLLM version: v0.24.0

### DIFF
--- a/tests/ut/test_platform.py
+++ b/tests/ut/test_platform.py
@@ -9,7 +9,7 @@ from vllm.v1.attention.selector import AttentionSelectorConfig  # type: ignore
 
 from tests.ut.base import TestBase
 from vllm_ascend.ascend_forward_context import MoECommType, override_mrv2_in_profile_run
-from vllm_ascend.platform import NPUPlatform
+from vllm_ascend.platform import NPUPlatform, _sync_npugraph_ex_to_additional_config
 from vllm_ascend.utils import (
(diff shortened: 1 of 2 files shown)


ShuhaoZhangTony:
Gentle ping for maintainer review. The current CI checks are green and the automated review has no unresolved thread. Feedback from a maintainer would be appreciated.

PR #12451, opened 2026-07-21
[Feature] Support AlltoAll + EP + LoRA in non‑quantized mode

wangx700:
This pull request enables LoRA support for Mixture of Experts (MoE) models running with Expert Parallelism (EP) on Ascend hardware. Previously, this configuration was broken and caused crashes. The changes introduce infrastructure to track and exchange per-token LoRA indices across Tensor Parallel (TP) ranks and during AlltoAll communication, ensuring that LoRA adapters are correctly applied to the distributed expert layers. It also includes updates to validation logic and adds a new end-to-end test case to ensure coverage for this feature.

- AlltoAll EP Support: Enabled LoRA support for MoE models using Expert Parallelism (AlltoAll) on Ascend hardware by allowing the AlltoAll communication backend.
- LoRA Index Exchange: Implemented a mechanism to synchronize and exchange per-token LoRA indices alongside hidden states during AlltoAll communication, ensuring alignment across TP ranks.
- Routing Recovery: Added utility functions to recover LoRA routing metadata specifically for the AlltoAll dispatch path, enabling correct LoRA delta application.
- E2E Test Coverage: Added a new end-to-end test case, test_qwen3moe_lora_ep, to verify LoRA + EP functionality.

Currently, LoRA for RL scenarios has not yet considered quantization.

No. EP + LoRA was previously broken (crashed); this PR makes it work. The existing TP-only LoRA path is unaffected. 

- vLLM version: v0.25.1


vllm_ascend/ops/fused_moe/token_dispatcher.py
```diff
@@ -643,6 +669,10 @@ def _dispatch_postprocess(
         dst_type,
         scale_type,
     ):
+        lora_context = self.lora_context
+        exchanged_lora_indices = (
+            getattr(lora_context, "exchanged_lora_indices", None) if lora_context is not None else None
+        )
```
weijinqian0: move this to it is used.

wangx700: all lora ops was moved  to lora files

weijinqian0 (approved):

Tflowers-0129 (approved):

PR #12494, opened 2026-07-21
[CI] Extract cleanup-fork-branches as composite action

zhangxinyuehfad:

The cherry-pick, doc-translate, and update-estimated-times workflows each create new branches on the `vllm-ascend-ci` fork per run. Over time this clutters the fork with hundreds of stale branches. This PR adds a cleanup step to each workflow via a shared composite action.

- vLLM version: v0.25.1

### DIFF
--- a/.github/actions/cleanup-fork-branches/action.yml
+++ b/.github/actions/cleanup-fork-branches/action.yml
@@ -1,0 +1,83 @@
+name: 'Cleanup stale fork branches'
+description: >-
+  Delete branches on a fork repo whose associated PRs in the upstream repo
+  are 'merged' or 'closed'. Branches with open PRs or no associated PRs are
+  left untouched. Results are written to $GITHUB_STEP_SUMMARY.
+
+inputs:
+  prefix:
+    description: 'Branch name prefix to match (e.g. auto-pr/doc-translate-)'
+    required: true
+  fork:
+    description: 'Fork repository (owner/name) holding the branches'
+    required: false
+    default: 'vllm-ascend-ci/vllm-ascend'
+  upstream:
+    description: 'Upstream repository to query PR state'
+    required: false
+    default: 'vllm-project/vllm-ascend'
+  token:
+    description: 'GitHub token with contents:write on the fork'
+    required: true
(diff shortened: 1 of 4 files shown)


.github/actions/cleanup-fork-branches/action.yml
```diff
@@ -0,0 +1,83 @@
...
+    - name: Run cleanup
+      env:
+        GH_TOKEN: ${{ inputs.token }}
+        PREFIX: ${{ inputs.prefix }}
+        FORK: ${{ inputs.fork }}
+        UPSTREAM: ${{ inputs.upstream }}
+      shell: bash
+      run: |
+        set +e
+        branches_json=$(gh api "repos/${FORK}/branches?per_page=100" \
```
wxsIcey: I think there is no need to impose a limit on this quantity. However, since operations such as document translation involve deleting branches each time, the current approach is also fine.

wxsIcey (approved):

menogrey:
I think we can get some action implement from github, and copy it. Another question is why we search with `per_page=100`, would we miss some branches?

.github/actions/cleanup-fork-branches/action.yml
zhangxinyuehfad: Modified, added --paginate flag, now fetches all branches instead of just the first 100.

.github/workflows/schedule_update_estimated_times.yaml
wangxiyuan: just notice that this job still use ubuntu-latest, please migrate it to self-host-runner

zhangxinyuehfad:
> I think we can get some action implement from github, and copy it. Another question is why we search with `per_page=100`, would we miss some branches?

Yes, we will investigate whether there is a better way to reuse it in a future GitHub action.

added --paginate flag, now fetches all branches instead of just the first 100.

.github/workflows/schedule_update_estimated_times.yaml
zhangxinyuehfad: Yes，I will do it.

PR #12496, opened 2026-07-21
[Feature] Remove additional config option 'dp_allreduce_on_npu'

lijiahang226:
  Removes the dp_allreduce_on_npu configuration option and its associated runtime logic. This option was introduced as a workaround for certain devices
  where CPU-side all_reduce could return dirty data, routing DP metadata synchronization through the NPU device group instead. Since this workaround is no
  longer needed, the conditional branch is removed and DP metadata all_reduce now unconditionally uses the CPU group, which simplifies the code path.

  Changes:
  - vllm_ascend/ascend_config.py: Removed dp_allreduce_on_npu configuration attribute
  - vllm_ascend/worker/model_runner_v1.py: Removed the NPU-vs-CPU device selection branch in the DP metadata all_reduce call — always uses cpu device and
  get_dp_group().cpu_group
Yes — the dp_allreduce_on_npu key in VLLM_ASCEND_CONFIG is no longer recognized. Users who previously set dp_allreduce_on_npu=True should remove it from
  their configuration; the value is silently ignored (no error is raised) but the NPU-path behavior is no longer available.

- vLLM version: v0.25.1

### DIFF
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -269,10 +269,6 @@ class AscendConfig:
         if self.mega_moe_max_tokens <= 0:
             raise ValueError(f"mega_moe_max_tokens must be a positive integer, got {self.mega_moe_max_tokens}")
 
-        # Whether to use NPU device group for DP metadata all_reduce.
-        # "True": use NPU device group, "False" (default): use CPU group.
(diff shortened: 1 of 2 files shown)


zzzzwwjj (approved):

lijiahang226:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

ningjingbengxiaohai (approved):

PR #12509, opened 2026-07-21
[CI] Make pre-commit explicit in PR workflow

### LINKED ISSUE #12508: [CI]: Make pre-commit explicit in PR workflow
﻿## Problem

The PR workflow currently combines PR title validation, pre-commit, mypy, coverage config validation, and selected-test discovery under one job named `lint-and-select-tests`.

This makes the check list less explicit than upstream vLLM's PR pipeline, where pre-commit is visible as its own job. It also makes it harder to tell whether a failure is from pre-commit/mypy or from selected-test discovery.

## Proposed minimal scope

- Rename/split the PR workflow so the pre-commit phase appears as a dedicated `pre-commit` job.
- Move selected-test discovery into a separate `select-tests` job that depends on `pre-commit` and exports the same outputs used by downstream jobs.
- Keep mypy in the `pre-commit` job for now because vLLM Ascend currently runs mypy through `tools/mypy.sh` across multiple Python versions rather than through `.pre-commit-config.yaml`.

## Out of scope

- No change to test selection logic.
- No change to E2E trigger behavior.
- No broad workflow refactor.
- No immediate mypy pre-commit hook migration unless validated separately.

robellliu-dev:

Fixes #12508.

This PR makes the PR workflow show the pre-commit phase explicitly, following the upstream vLLM pipeline shape where `pre-commit` is a visible job instead of being hidden under a combined lint/select job.

Minimal workflow changes:

- Rename the old combined lint phase into a dedicated `pre-commit` job.
- Keep PR title validation, pre-commit, and mypy in that job.
- Move selected-test discovery into a new `select-tests` job that depends on `pre-commit`.
- Preserve the existing selected-test outputs (`has_tests`, `test_groups`, `matched_modules`, `main_commit`, `release_tag`) for downstream E2E jobs.
- Update downstream `needs` references and `ci-gate` reporting from `lint-and-select-tests` to `pre-commit` / `select-tests`.

Mypy review:

- Upstream vLLM integrates mypy through pre-commit hooks, but vLLM Ascend currently runs mypy via `tools/mypy.sh` for Python 3.10, 3.11, and 3.12.
- This PR integrates mypy into the visible `pre-commit` job while keeping the existing `tools/mypy.sh` implementation.
- Moving mypy into `.pre-commit-config.yaml` would change hook dependencies and local developer behavior, so that is intentionally left out of this minimal CI display cleanup.

Scope Guard:

- Issue contract: make pre-commit visible as its own PR workflow job while preserving existing PR test selection behavior.
- Archetype: CI And Workflow Correctness.
- Changed file: `.github/workflows/pr_test.yaml` only.
- No test-selection logic, reusable workflow, Docker image, or pre-commit hook changes are included.

No. This is CI workflow structure only. It changes how PR checks are displayed and grouped, not runtime behavior.

- `pre-commit run actionlint --hook-stage manual --files .github/workflows/pr_test.yaml`
- PyYAML structure assertion verifying:
  - `pre-commit` and `select-tests` jobs exist
  - `lint-and-select-tests` is no longer referenced
  - downstream jobs consume `select-tests` outputs
  - `ci-gate` depends on `pre-commit`, `select-tests`, and `run-selected-tests`
- `rg -n "lint-and-select-tests|needs\.lint-and-select-tests" .github\workflows\pr_test.yaml` returned no stale references
- `git diff --check`
- Commit-time pre-commit hooks passed, including actionlint
- vLLM version: v0.25.1

### DIFF
--- a/.github/workflows/pr_test.yaml
+++ b/.github/workflows/pr_test.yaml
@@ -41,17 +41,11 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  lint-and-select-tests:
+  pre-commit:
     if: ${{ github.event.action != 'labeled' || github.event.label.name == 'ready' }}
     runs-on: linux-amd64-cpu-8-hk
     container:
       image: quay.io/ascend-ci/vllm-ascend:lint
-    outputs:
-      has_tests: ${{ steps.scope.outputs.has_tests || steps.noop.outputs.has_tests }}
-      test_groups: ${{ steps.scope.outputs.test_groups || steps.noop.outputs.test_groups }}
-      matched_modules: ${{ steps.scope.outputs.matched_modules || steps.noop.outputs.matched_modules }}
-      main_commit: ${{ steps.vllm.outputs.main_commit }}
-      release_tag: ${{ steps.vllm.outputs.release_tag }}
     steps:
       - name: Validate PR title prefix
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_test.yaml
```diff
@@ -184,6 +162,62 @@ jobs:
...
+      matched_modules: ${{ steps.scope.outputs.matched_modules || steps.noop.outputs.matched_modules }}
+      main_commit: ${{ steps.vllm.outputs.main_commit }}
+      release_tag: ${{ steps.vllm.outputs.release_tag }}
+    steps:
+      - name: Checkout vllm-project/vllm-ascend repo
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - name: Read verified vLLM refs
```
wangxiyuan: Looks that this step is redundant, since it's called at L88 already. Just reuse the result  from L88 in this yaml is enough.

for example:
change `needs.select-tests.outputs.main_commit` to `needs.pre-commit.outputs.main_commit`

wangxiyuan (approved):

PR #12262, opened 2026-07-17
[BugFix][GLM-5.2][SpecDecode] Fix DSpark W8A8 eager acceptance

Liuchenbing-2026:
### What this PR does / why we need it?

Fix the GLM-5.2 DSpark acceptance rate on the refactored `main` (post #11431). Before this PR, the GLM-5.2 W8A8 DSpark eager path had a token acceptance rate of only ~1.18%, effectively unusable. This PR restores acceptance to the expected level via three focused fixes:

- **Register `Qwen3DSparkModel` under the Ascend model registry** as `AscendQwen3DSparkForCausalLM`. After the #11431 refactor consolidated most DSpark logic into vLLM upstream, vllm-ascend still needs an Ascend-specific override to apply QuaRot rotation to `fc.*` weights at load time. The new class inherits the upstream `Qwen3DSparkForCausalLM` and only overrides `load_weights`, keeping non-QuaRot targets on the original path.
- **Introduce a `sample_from_anchor` toggle in `AscendDsparkProposer`**, driven by the `dspark_bonus_anchor` hf-config flag. The proposer now derives `num_query_per_req` from this flag (either `N` for anchor-first or `1+N` for bonus-anchor) instead of hard-coding one layout. All downstream places — kernel arguments, `token_indices_to_sample`, `query_start_loc`, `seq_lens` — use the new abstraction. Missing `dspark_bonus_anchor` defaults to `False`, preserving the current behavior for existing checkpoints.
- **Fix the MarkovHead sampling input**: replace `compute_logits(last_hidden_states)` with `compute_logits(sample_hidden_states)` in `AscendSpecDecodeBaseProposer._run_merged_draft`. The previous code fed the last hidden state to `compute_logits`, so all `N` speculative tokens received identical logits — the primary cause of the collapsed acceptance rate. The fix is scoped to `if self.method == "dspark"` and does not touch DFlash / EAGLE paths.

### Does this PR introduce _any_ user-facing change?

- New model class `AscendQwen3DSparkForCausalLM` is registered for the `Qwen3DSparkModel` architecture. Users deploying GLM-5.2 W8A8 DSpark with a QuaRot-quantized target will automatically go through the new load path; other models are unaffected.
- New optional hf-config flag `dspark_bonus_anchor` on draft models. Absent flag defaults to `False`, which matches the pre-PR behavior. Setting `True` enables the bonus-anchor draft layout.
- No new env variables, CLI flags, or config keys required in `vllm serve` invocations.

### How was this patch tested?

[...]

PR #12539, opened 2026-07-22
[BugFix] Fix GLM-5.2 DSpark acceptance rate problem

wangbj127:
GLM-5.2 DSpark acceptance rates are extremely low, due to the following reasons:
- GLM-5.2 should not sample from anchor token, since `dspark_bonus_anchor=True`, which is different from DeepSeek-V4 DSpark
- When target model uses rotary quantization, the first linear weight (`fc.weight` in this case) need to be processed with global rotation matrix.

No.

GLM-5.2 W4A8 + DSpark BF16 acceptance rates are as follows
```
num_drafts=16397.0
num_accepted_tokens_per_pos=[14302.0, 11935.0, 9840.0, 7981.0, 6458.0, 5168.0, 3929.0]
acceptance_per_pos=[0.8722327254985668, 0.7278770506800024, 0.6001097761785692, 0.4867353784228822, 0.393852534000122, 0.31517960602549244, 0.23961700311032505]
acceptance_len=4.63560407391596
```
- vLLM version: v0.25.1

### DIFF
--- a/tests/ut/model_executor/test_qwen3_dspark.py
+++ b/tests/ut/model_executor/test_qwen3_dspark.py
@@ -1,0 +1,173 @@
+from pathlib import Path
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+import torch
+from vllm.model_executor.models.qwen3_dspark import Qwen3DSparkForCausalLM
+
+import vllm_ascend.models as ascend_models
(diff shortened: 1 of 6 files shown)


wangbj127:
duplicated https://github.com/vllm-project/vllm-ascend/pull/12262

PR #12612, opened 2026-07-22
[Bugfix] Fix multi dp in dspark

slippersss:
This PR aims to fix dp padding in dsv4 dspark. The improper slicing for `positions` causes errors. We fix it and add ut for solving this problem.

N/A

by ci

- vLLM version: v0.25.1

### DIFF
--- a/vllm_ascend/spec_decode/dspark_proposer.py
+++ b/vllm_ascend/spec_decode/dspark_proposer.py
@@ -288,7 +288,7 @@ class AscendDSparkProposer(AscendDflashProposer):
         cad.max_query_len = block_size
         cad.max_seq_len = cad.max_seq_len + block_size
         cad.slot_mapping = self._per_group_query_slot_mapping_buffers[primary_gid][:num_query_total]
-        cad.positions = self.positions[:num_query_total]
+        cad.positions = self.positions  # this would be sliced in attention backend
         cad.causal = False
         cad.attn_mask = None
         cad.attn_state = AscendAttentionState.ChunkedPrefill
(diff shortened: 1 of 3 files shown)


vllm_ascend/spec_decode/dspark_proposer.py
```diff
@@ -288,7 +288,7 @@ def set_inputs_first_pass(
         cad.max_query_len = block_size
         cad.max_seq_len = cad.max_seq_len + block_size
         cad.slot_mapping = self._per_group_query_slot_mapping_buffers[primary_gid][:num_query_total]
-        cad.positions = self.positions[:num_query_total]
+        cad.positions = self.positions  # this would be sliced in attention backend
```
drslark, line 291: Maybe add an ut?
LGTM.

drslark (approved):

vllm_ascend/spec_decode/dspark_proposer.py
slippersss: Adding now...

QwertyJack:
I reran this on latest `main` (`f5b5514af`) with the real DeepSeek-V4-Flash DSpark W4A8 checkpoint, V1, TP4/DP4/EP, and `--enforce-eager`.

The 16x256 reproducer completed 4 requests and failed 12 with `aclnnInplacePartialRotaryMul: dim0 must be equal`, so this is not graph-only. Removing the `positions` slice covers one mismatch, but the same DP-padding contract also requires:

- padding positions/slots before draft metadata is built;
- using the padded decode-token count in non-causal DSA metadata;
- explicit `lens=0` / `slots=-1` SWA rows for the padded tail.

I opened #12640 with the concise reproducer/root cause and #12641 with the complete metadata fix plus unit and real-weight validation (`4/12` -> `16/0`).

slippersss:
/rerun
[Bot]: rerun completed.

**Failed:**

zzzzwwjj (approved):

slippersss:
/rerun

PR #12630, opened 2026-07-22
[BugFix][Worker] Replace copy_snapshot_to_gpu with synchronize_input_prep for query_start_loc buffers

xiayingqing:

PR #12079 fixed a race where back-to-back steps could overwrite pinned CPU
buffers while a non-blocking H2D was still reading from them, using per-call
clone-based snapshotting (`copy_snapshot_to_gpu`).

This PR replaces the clone approach with `synchronize_input_prep()`—the
event-based guard already used by upstream vLLM's `GPUModelRunner` for the
same purpose. Each step waits for the previous H2D to complete before
writing, without the per-call allocation overhead of cloning.

Changes:
- Wrap `_dummy_run` and proposer `query_start_loc` copy + copy_to_gpu
  blocks with `synchronize_input_prep()`
- Revert `copy_snapshot_to_gpu()` calls introduced by #12079 back to
  `copy_to_gpu()`

No. This is an internal refinement of the fix in #12079, replacing the
clone-based approach with a lighter-weight event protocol.

- vLLM version: v0.25.1


freyfwt (approved):

vllm_ascend/_310p/model_runner_310p.py
```diff
@@ -452,13 +451,13 @@ def _prepare_inputs(  # type: ignore[override]
...
             self.query_start_loc.np[num_reqs + 1 :].fill(-1)
-        copy_snapshot_to_gpu(self.query_start_loc)
+        self.query_start_loc.copy_to_gpu()
 
         if self._has_gdn:
             self.gdn_query_start_loc.np[0] = 0
             self.gdn_query_start_loc.np[1 : num_reqs + 1] = cu_num_tokens
             self.gdn_query_start_loc.np[num_reqs + 1 :].fill(cu_num_tokens[-1])
-            copy_snapshot_to_gpu(self.gdn_query_start_loc)
+            self.gdn_query_start_loc.copy_to_gpu()
```
weijinqian0, line 460: Plz check if there also need sync.

weijinqian0 (approved):

yiz-liu (commented):
I think this context manager should only be used in `runner` and `proposer`. Other places should not be allowed to modify CPU tensors.

Please contact the owners of 310P and `step3p5.py` to check whether any additional fixes are needed.

Also, @wangxiaoteng888 PTAL and see if #12189 needs to be reverted.

yiz-liu (commented):
Also, `_prepare_inputs` should also be wrapped, @xiayingqing

vllm_ascend/_310p/model_runner_310p.py
xiayingqing: This call is inside _prepare_inputs, which is invoked within execute_model's with self.synchronize_input_prep(): block .

xiayingqing:
> Also, `_prepare_inputs` should also be wrapped, @xiayingqing

_prepare_inputs is already inside execute_model's synchronize_input_prep() , no additional wrapper needed.

xiayingqing:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

xiayingqing:
/rerun

Issue #12078, opened 2026-07-15
[RFC]: vLLM-Ascend Nightly-A3 Test Pipeline Scheduling and Resource Optimization


zhangxinyuehfad:
# Motivation.

### 1.1 Original Pipeline Performance Bottlenecks

Currently, the Nightly-A3 test suite in the `vllm-ascend` repository takes **11 hours 25 minutes** to complete. A quantitative analysis of execution logs and test case durations revealed three core bottlenecks:

* **Sequential Blockage & Lack of Concurrency**: The multi-node, double-node, single-node, and multi-card test phases are executed entirely in serial. Even though the gaps between stages are less than 1 minute, this strictly sequential topology prevents the overall test cycle from converging efficiently.
* **Significant Long-tail Test Cases**:
  * **Single-node (16-card) Phase** (Totaling 4h 43m): This phase contains several massive models or long-context test cases. For instance, `mtpx-deepseek-r1-0528-w8a8` (1h 24m 37s), `deepseek-r1-0528-w8a8` (1h 15m 26s), and `MiniMax-M2.5-w8a8-QuaRot-A3` (1h 7m 52s) severely drag down the single-node execution path.
* **Low Utilization of High-Value Cluster Resources**: Because the multi-node and double-node test cases are not arranged concurrently based on physical machine topology, high-value A3 computing clusters sit idle or under-utilized during specific windows.

#### 1.2.1 Gantt Chart Analysis of Job Durations

<img width="865" height="361" alt="Image" src="https://github.com/user-attachments/assets/02aa286c-5d3a-4fd4-82e4-e90db6d4b5c6" />

#### 1.2.2 Gantt Chart Analysis of Test Case Durations

<img width="865" height="1168" alt="Image" src="https://github.com/user-attachments/assets/354e69c4-b5e5-4f04-a196-fbaa74ce3e1d" />

#### 1.2.3 Detailed Execution Times for Individual Test Cases

<img width="865" height="758" alt="Image" src="https://github.com/user-attachments/assets/3f36c576-6273-4052-8192-dcea72f7e9cb" />

<img width="865" height="1359" alt="Image" src="https://github.com/user-attachments/assets/203c9e91-1fa6-4f5c-8a95-7aab1af722f6" />

---

# Proposed Change.

## 2. Proposed Architecture & Design

To compress the feedback loop and maximize resource efficiency, the optimization is implemented in three progressive phases:

```text
[ Start Trigger ]
        |
[...]

PR #12633, opened 2026-07-22
[CI] Add pod readiness signaling for A3 560T nightly test

zhangxinyuehfad:

RFC: https://github.com/vllm-project/vllm-ascend/issues/12078

Adapt the double-node pod readiness coordination mechanism from the A3 nightly workflow to the A3 560T cluster, allowing single-node and multi-card tests to start as soon as double-node pods are ready rather than waiting for full test completion.

Changes:
- Add send_ready_signal to _e2e_nightly_multi_node_560t.yaml
- Create _nightly_wait_for_pods_ready.yaml reusable polling workflow
- Add double-node-pods-started job with isolated context prefix (double-node-pods-ready-560t) to avoid conflict with A3
- Restructure job dependencies: single-node/multi-card now depend on double-node-pods-started instead of double-node-tests
- Add remove-taints job for cleanup
- Make PAT_TOKEN optional to avoid requiring it for multi-node callers

- vLLM version: v0.25.1

### DIFF
--- a/.github/workflows/_e2e_nightly_multi_node_560t.yaml
+++ b/.github/workflows/_e2e_nightly_multi_node_560t.yaml
@@ -95,7 +95,19 @@ on:
         type: boolean
         default: false
         description: enable Pod AOP pipeline on failure
+      send_ready_signal:
+        required: false
+        type: boolean
+        default: false
+        description: send a commit status signal after LWS is created (used by double-node to unblock downstream jobs)
+      run_id:
+        required: false
+        type: string
+        default: ''
+        description: unique run identifier to isolate signals across concurrent runs (e.g. github.run_id)
     secrets:
+      PAT_TOKEN:
+        required: false
       KUBECONFIG_B64_560T:
         required: true
       OBS_ACCESS_KEY_ID:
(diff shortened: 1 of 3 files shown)


wxsIcey:
I think there might be an issue: single-node/multi-card is currently running in parallel with double-node-tests, but clear-pre-logs only cleans up the PVC after double-node-tests finishes. At this point, single-node/multi-card might still be writing logs to the same PVC. **The clear-pre-logs requirement should include single-node-tests and multi-card-tests.**

zhangxinyuehfad:
> I think there might be an issue: single-node/multi-card is currently running in parallel with double-node-tests, but clear-pre-logs only cleans up the PVC after double-node-tests finishes. At this point, single-node/multi-card might still be writing logs to the same PVC. **The clear-pre-logs requirement should include single-node-tests and multi-card-tests.**

It runs after multi-node/double-node because it cleans the shared PVC logs (/root/.cache/ascend-logs) on runner linux-aarch64-a3-0, which is used exclusively by multi-node tests. Its purpose is to clear stale logs from the PVC to avoid confusion for the next run — so it must execute after the multi-node jobs that use it.
It excludes single-node (single-node-tests, multi-card-tests) because those jobs run on separate standalone runners (determined by matrix.test_config.os). They don't share the multi-node PVC — each has its own local storage — so there are no shared logs to clean from them.

menogrey:
LGTM

wangxiyuan (approved):

PR #12641, opened 2026-07-22
[Bugfix] Fix DSpark DP-padded metadata

### LINKED ISSUE #12640: [Bug]: DSpark V1 crashes when DP pads drafter tokens
## Environment

- vLLM-Ascend: `f5b5514afdb9ed53e7382ce8abdefec0bd0e5c62` (latest `main`)
- vLLM: `54503ecec0f3ac31e5ecfc5f28652e4cc42307b5`
- CANN: `9.0.1`
- torch-npu: `2.10.0.post2`
- Model: `/models/DeepSeek-V4-Flash-DSpark-w4a8`
- Topology: TP4 / DP4 / EP on 16 NPUs
- Model runner: V1

## Reproduction

Start DSpark with graph capture disabled:

```bash
VLLM_USE_V2_MODEL_RUNNER=0 \
ASCEND_RT_VISIBLE_DEVICES=0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15 \
vllm serve /models/DeepSeek-V4-Flash-DSpark-w4a8 \
  --tensor-parallel-size 4 \
  --data-parallel-size 4 \
  --enable-expert-parallel \
  --max-model-len 4096 \
  --max-num-seqs 4 \
  --max-num-batched-tokens 2048 \
  --quantization ascend \
  --additional-config '{"enable_fused_mc2":false}' \
  --speculative-config '{"method":"dspark","num_speculative_tokens":5}' \
  --enforce-eager
```

Send 16 heterogeneous prompts with `temperature=0`, `max_tokens=256`, `ignore_eos=true`, and concurrency 4.

Observed on unmodified `main`:

- completed: 4
- failed: 12
- the engine exits when requests finish at different times across DP ranks

```text
RuntimeError: call aclnnInplacePartialRotaryMul failed
EZ9999: dim0 must be equal
CheckInputShapes is failed
```

This also reproduces with all graph capture disabled, so it is not an ACLGraph issue. The rotary operator is only the first place that rejects the inconsistent token dimensions.

## Root cause

After DP synchronization, `num_input_tokens` can be larger than the current rank's actual drafter token count. The DSpark path does not apply one consistent contract to that padded tail:

1. draft attention metadata is built before `positions` and slot buffers receive padding sentinels;
2. non-causal DSA metadata keeps the actual decode-token count while positions and hidden states use the DP-padded count;
3. SWA indices use `repeat_interleave(..., output_size=num_decode_tokens)` even when the repeat counts only cover actual tokens, so padded token rows have no defined metadata.

Related #12612 changes the `positions` view, but it does not cover the decode-token count or SWA padded-tail sentinels.

## Proposed fix

- pad DSpark token buffers before building draft attention metadata;
- pass padded positions and slot mappings with `positions=0` and `slot_mapping=-1` for inactive rows;
- align non-causal DSA decode-token count with `num_input_tokens`;
- generate SWA padded rows with `lens=0` and `slots=-1`.

The change should remain limited to DSpark/DSA metadata construction; rotary and SAS operator code do not need changes.

## Validation

Using the same real-weight eager workload:

| Revision | Completed | Failed | Output tokens |
| --- | ---: | ---: | ---: |
| latest `main` | 4 | 12 | 813 before engine failure |
| proposed metadata fix | 16 | 0 | 4096 |

Unit coverage should include actual token count smaller than DP-padded token count for positions, slot mappings, DSA decode count, and SWA sentinels.

QwertyJack:

Fixes #12640.

DSpark V1 can receive a DP-padded `num_input_tokens` larger than the current rank's actual drafter token count. Before this change, the metadata path mixed those two sizes:

- draft attention metadata was built before token buffers received padding sentinels;
- non-causal DSA metadata kept the actual decode-token count;
- SWA index expansion had no defined rows for the padded tail.

The mismatch eventually surfaced as `aclnnInplacePartialRotaryMul: dim0 must be equal`, including with `--enforce-eager`.

This PR keeps the fix in DSpark/DSA metadata construction:

- pad DSpark buffers before building draft attention metadata;
- pass padded positions and slots (`positions=0`, `slot_mapping=-1` for inactive rows);
- align the non-causal DSA decode-token count with `num_input_tokens`;
- generate SWA padded rows with `lens=0` and `slots=-1`.

Rotary, SAS operator, scheduler, graph-capture, and non-DSpark paths are unchanged.

Related #12612 fixes the positions slice only. This PR also covers the decode-token count and SWA padded-tail contract required by the real DP4 reproducer.

Yes. DSpark V1 no longer crashes when DP ranks require different amounts of drafter token padding.

Unit tests:

```text
tests/ut/spec_decode/a2/test_eagle_proposer.py: 73 passed, 13 skipped
tests/ut/spec_decode/test_llm_base_proposer.py: 9 passed
```

Static checks:

```text
ruff check: passed
ruff format --check: passed
py_compile: passed
git diff --check: passed
```

Real-weight test with `/models/DeepSeek-V4-Flash-DSpark-w4a8`, V1, TP4/DP4/EP, 16 NPUs, `--enforce-eager`, 16 heterogeneous prompts, output length 256, concurrency 4:

| Revision | Completed | Failed | Output tokens |
| --- | ---: | ---: | ---: |
| latest `main` (`f5b5514af`) | 4 | 12 | 813 before engine failure |
| this PR | 16 | 0 | 4096 |

- vLLM version: v0.25.1

### DIFF
--- a/vllm_ascend/attention/dsa_v1.py
+++ b/vllm_ascend/attention/dsa_v1.py
@@ -420,8 +420,16 @@ def build_dspark_swa_indices(
     slot_ids = (block_ids * block_size + block_offsets).to(torch.int32)
     slot_ids = slot_ids.where(col_mask, torch.full_like(slot_ids, -1))
 
-    per_token_slots = torch.repeat_interleave(slot_ids, query_lens, dim=0, output_size=num_decode_tokens).unsqueeze(1)
-    per_token_lens = torch.repeat_interleave(visible_lens, query_lens, dim=0, output_size=num_decode_tokens)
+    if num_decode_tokens is None:
+        num_decode_tokens = query_start_loc[-1]
+    token_rows = torch.arange(num_decode_tokens, dtype=torch.long, device=query_start_loc.device)
+    token_to_req = torch.bucketize(token_rows, query_start_loc[1:].long(), right=True)
+    inactive_tokens = token_to_req >= slot_ids.shape[0]
+    safe_token_to_req = token_to_req.clamp(max=slot_ids.shape[0] - 1)
+    per_token_slots = slot_ids.index_select(0, safe_token_to_req).unsqueeze(1)
+    per_token_slots.masked_fill_(inactive_tokens[:, None, None], -1)
+    per_token_lens = visible_lens.index_select(0, safe_token_to_req)
+    per_token_lens.masked_fill_(inactive_tokens, 0)
 
     return per_token_slots, per_token_lens
 
@@ -1188,6 +1196,8 @@ class AscendDSAMetadataBuilder(AttentionMetadataBuilder[AscendDSAMetadata]):
         num_decodes, num_prefills, num_decode_tokens, num_prefill_tokens = split_decodes_and_prefills(
             common_attn_metadata, decode_threshold=self.decode_threshold
         )
+        if not common_attn_metadata.causal and num_prefills == 0:
+            num_decode_tokens = common_attn_metadata.num_input_tokens
         num_input_tokens = common_attn_metadata.num_input_tokens
         input_positions = common_attn_metadata.positions[:num_input_tokens].long()
         if num_prefills:
--- a/vllm_ascend/spec_decode/llm_base_proposer.py
+++ b/vllm_ascend/spec_decode/llm_base_proposer.py
@@ -930,7 +930,8 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
         multi_steps_attn_metadata, attn_metadata_i = self.build_draft_attn_metadata(
             common_attn_metadata, num_input_tokens, num_tokens
         )
-        self._pad_draft_buffers(num_tokens, num_input_tokens)
+        if self.method != "dspark":
+            self._pad_draft_buffers(num_tokens, num_input_tokens)
 
         if self.uses_mrope:
             used_update_positions = self.mrope_positions[:, token_indices_to_sample]
@@ -2136,6 +2137,8 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
     ):
         # FIXME(woosuk): The below two ops cause synchronization. Optimize.
         assert len(self.draft_attn_groups) > 0
+        if self.method == "dspark":
+            self._pad_draft_buffers(num_actual_tokens, num_input_tokens)
         per_layer_attn_metadata: dict[str, Any] = {}
         for attn_group in self.draft_attn_groups:
             builder = attn_group.get_metadata_builder()
@@ -2150,6 +2153,7 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
             if self.method == "dspark":
                 gid = attn_group.kv_cache_group_id
                 common_attn_metadata = copy.copy(common_attn_metadata)
+                common_attn_metadata.positions = self.positions[:num_input_tokens]
                 block_table = getattr(self, "_per_group_block_table_buffers", {}).get(gid)
                 if block_table is not None:
                     common_attn_metadata.block_table_tensor = block_table[: common_attn_metadata.num_reqs]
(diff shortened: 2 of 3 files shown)


vllm_ascend/attention/dsa_v1.py
```diff
@@ -420,8 +420,16 @@ def build_dspark_swa_indices(
...
+    if num_decode_tokens is None:
+        num_decode_tokens = query_start_loc[-1]
+    token_rows = torch.arange(num_decode_tokens, dtype=torch.long, device=query_start_loc.device)
+    token_to_req = torch.bucketize(token_rows, query_start_loc[1:].long(), right=True)
+    inactive_tokens = token_to_req >= slot_ids.shape[0]
+    safe_token_to_req = token_to_req.clamp(max=slot_ids.shape[0] - 1)
+    per_token_slots = slot_ids.index_select(0, safe_token_to_req).unsqueeze(1)
+    per_token_slots.masked_fill_(inactive_tokens[:, None, None], -1)
+    per_token_lens = visible_lens.index_select(0, safe_token_to_req)
+    per_token_lens.masked_fill_(inactive_tokens, 0)
```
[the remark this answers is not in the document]
QwertyJack: Thanks for checking this. These tensors are already required to be co-located: `safe_nums` is derived from `query_start_loc`/`seq_lens` and is used by `torch.gather(block_table, 1, safe_nums)` immediately above. If `block_table` and `query_start_loc` were on different devices, that gather would fail before `slot_ids` is created. `visible_lens` and `token_to_req` also share the same source device. Moving only the index and mask to `slot_ids.device` would still leave `visible_lens.index_select` and its mask inconsistent in the hypothetical mixed-device case, while adding copies to the per-step metadata path. The CPU unit test and the real TP4/DP4 NPU run both cover this path, so I am keeping the tensors on their existing common device.

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -930,7 +930,8 @@ def _propose(
         multi_steps_attn_metadata, attn_metadata_i = self.build_draft_attn_metadata(
             common_attn_metadata, num_input_tokens, num_tokens
         )
-        self._pad_draft_buffers(num_tokens, num_input_tokens)
+        if self.method != "dspark":
+            self._pad_draft_buffers(num_tokens, num_input_tokens)
```
drslark, line 934: `_pad_draft_buffers` is necessary for all draft methods, otherwise, `slot_mapping` will be dirty.

vllm_ascend/attention/dsa_v1.py
```diff
@@ -420,8 +420,16 @@ def build_dspark_swa_indices(
...
-    per_token_slots = torch.repeat_interleave(slot_ids, query_lens, dim=0, output_size=num_decode_tokens).unsqueeze(1)
-    per_token_lens = torch.repeat_interleave(visible_lens, query_lens, dim=0, output_size=num_decode_tokens)
+    if num_decode_tokens is None:
+        num_decode_tokens = query_start_loc[-1]
+    token_rows = torch.arange(num_decode_tokens, dtype=torch.long, device=query_start_loc.device)
+    token_to_req = torch.bucketize(token_rows, query_start_loc[1:].long(), right=True)
+    inactive_tokens = token_to_req >= slot_ids.shape[0]
+    safe_token_to_req = token_to_req.clamp(max=slot_ids.shape[0] - 1)
+    per_token_slots = slot_ids.index_select(0, safe_token_to_req).unsqueeze(1)
+    per_token_slots.masked_fill_(inactive_tokens[:, None, None], -1)
```
drslark, line 430: The pad logic should only be in proposer.

vllm_ascend/spec_decode/llm_base_proposer.py
drslark: @lilinsiman Check this logic.

QwertyJack:
Closing this PR as superseded by #12612. Local verification successful.

PR #12650, opened 2026-07-22
[Bugfix] fix main2main eplb

Spicy-Stick:
This PR restores the compatibility of the legacy Ascend EPLB implementation with the latest upstream vLLM FusedMoE refactoring.
It:

- Propagates the configured redundant expert count to the upstream FusedMoE factory, ensuring that physical expert slots are allocated before checkpoint loading.

- Separates logical expert placement from physical expert capacity and validates that vLLM and Ascend use consistent expert counts.

- Aligns Ascend's initial physical-to-logical placement with the upstream checkpoint layout: logical experts first, followed by redundant replicas.

- Updates VllmEplbAdaptor to support the refactored MoERunner, where expert weights may be owned by RoutedExperts and accessed through get_eplb_parameter().

- Uses static attribute inspection to avoid dynamic proxy and unspecced MagicMock side effects.

- Fixes expert-load tensor dimensions: only the FlashLB policy uses a per-step collection window, while SwiftBalance consumes one load vector per layer and rank.

- Exposes the local expert count, expert map update, logical-to-physical map, and load-reset interfaces required by EPLB.

- Re-enables the Qwen3 MoE and dynamic EPLB E2E tests and adds corresponding unit-test coverage.

- Documents that container deployments must mount shared memory.

This is needed because upstream vLLM now allocates RoutedExperts parameters before constructing the Ascend MoE runner. Previously, Ascend added redundant-expert information too late, which caused the allocated weight capacity, checkpoint physical layout, and Ascend placement map to become inconsistent. Depending on the configuration, this resulted in missing attributes, expert-capacity mismatches, EPLB worker crashes, or incorrect inference results when redundant experts were enabled.
The adaptor also assumed that expert parameters were stored directly on the MoE runner. After the upstream refactoring, some parameters moved under RoutedExperts, so the adaptor must support both the legacy direct-attribute layout and the new accessor-based layout.

- vLLM version: v0.25.1

### DIFF
--- a/vllm_ascend/patch/platform/patch_fused_moe.py
+++ b/vllm_ascend/patch/platform/patch_fused_moe.py
@@ -30,6 +30,7 @@
 import vllm.model_executor.layers.fused_moe as _fused_moe_pkg
 import vllm.model_executor.layers.fused_moe.layer as _fused_moe_layer
 
+from vllm_ascend.ascend_config import get_ascend_config
 from vllm_ascend.utils import is_310p
 
 # Capture the real original before fused_moe.py's module-level code runs.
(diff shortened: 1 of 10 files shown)


freyfwt (approved):

ningjingbengxiaohai (approved):

vllm_ascend/patch/platform/patch_fused_moe.py
```diff
@@ -44,6 +45,19 @@
 def _ascend_FusedMoE(*args, runner_cls=None, runner_args=None, **kwargs):
     if runner_cls is None:
         runner_cls = _DefaultAscendMoERunner
+    # RoutedExperts allocates its parameters before AscendMoERunner is
+    # constructed. Propagate Ascend EPLB capacity into the upstream factory so
+    # redundant expert slots are present when weights are created and loaded.
+    eplb_config = get_ascend_config().eplb_config
```
wangxiyuan, line 51: why not override enable_eplb and num_redundant_experts in check_and_update_config function？

PR #12668, opened 2026-07-23
[Feature] Add NPU frequency regulator operator for A5

Viktorrrrrrrrrrr:
This PR introduces a new custom operator on A5, aclnnFrequencyRegulator, designed to dynamically regulate the frequency of Ascend NPU devices during model execution. It includes the necessary C++ header definitions, the integration into the Torch binding layer, and the required meta kernel implementation to support proper operator registration and execution within the vLLM Ascend backend.

**Usage Example**
The operator is designed to be injected into the computation graph without altering the tensor values. It uses an existing tensor (e.g., `hidden_states`) to anchor the device context and prevent graph optimization passes from removing the operator.

hidden_states = torch.ops._C_ascend.run_frequency_regulator(1400, hidden_states)

No.

Tested via local compilation on A5 and execution to ensure the custom operator is properly registered and functions correctly within the vLLM Ascend backend.

- vLLM version: v0.25.1

### DIFF
--- a/csrc/common/aicpu/aclnn_frequency_regulator.h
+++ b/csrc/common/aicpu/aclnn_frequency_regulator.h
@@ -1,0 +1,56 @@
+/**
+ * Copyright (c) 2025 Huawei Technologies Co., Ltd.
+ * This file is a part of the CANN Open Software.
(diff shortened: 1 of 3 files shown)


yiz-liu (commented):
Please add docs, demo and test case for this.

PR #12368, opened 2026-07-19
[Feature] SD spec decode op

HF-001:
### What this PR does / why we need it?
refer to: https://github.com/vllm-project/vllm-ascend/pull/10894 ,
Improve the SD of the current main branch in the following two aspects： 
Dynamically extending K to Medusa/extract_hidden_states/eagle/mtp.

### How was this patch tested?
```
qwen3-8b+eagle3, test results(bs 128): Output throughput：534 -> 652 , ttft : 63637-> 44785 , tpot: 83.9 -> 80
```
qwen3-8b+eagle3：
```
vllm serve /nas/disk1/Qwen3-8B --port 8898 --dtype bfloat16  \
--tensor-parallel-size 1 --gpu-memory-utilization 0.9 \
--max-model-len 32768 --trust-remote-code \
--no-enable-prefix-caching \
--async-scheduling \
--speculative_config '{"method":"eagle3","model":"/model/EAGLE3-Qwen3-8B","num_speculative_tokens":3,"draft_tensor_parallel_size":1, "num_speculative_tokens_per_batch_size":[[1,32,3],[33,64,2],[65,128,1]]}'
```

Test results of support for existing speculative decode algorithms：
- eagle3 : after testing, it supports SD
- dflash:  after testing, it supports SD
- mtp:  after testing, it supports SD
- ngram:  after testing, it supports SD
- suffix:   after testing, it not supports SD of this pr. Suffix itself is' per request dynamic ', and the outer dynamic K is redundant/conflicting for it.This is an upstream constraint, and vLLM itself will also collapse, not a unique problem of Ascend.
- dspark：Need to develop and adapt separately

vLLM version: v0.26.0
vLLM main: https://github.com/vllm-project/vllm/commit/d02df748bf9efd99022f1a062597dc3cb3808485

- vLLM version: v0.26.0
- vLLM main: https://github.com/vllm-project/vllm/commit/0351e9aa1fdf1a51329d1906881528dfe61fc88e

PR #12686, opened 2026-07-23
[Feature] Add Dynamic Speculative Decoding (DSD) for EAGLE/EAGLE3/MTP proposers on speculative decode

chengduxiaowu:
This PR enables Dynamic Speculative Decoding (DSD) for the EAGLE / EAGLE3 / MTP proposers on Ascend, so the per-step speculative-token count K varies with the running batch size instead of staying fixed.

  Upstream vLLM (v0.25.1) added DSD via the num_speculative_tokens_per_batch_size schedule; the scheduler-side lookup was already synced to Ascend's recompute / balance schedulers, and the ngram / suffix proposers already
  honor the per-step K. The EAGLE / EAGLE3 / MTP path did not. This PR closes that gap.

  Yes (opt-in, no breakage). num_speculative_tokens_per_batch_size now takes effect with eagle / eagle3 / mtp on Ascend — previously it was honored only for ngram / suffix. Behavior is unchanged when the new config key is not
  set. num_speculative_tokens becomes the maximum K; each schedule entry is clamped to it, and K=0 means a plain decode for that batch-size range. Documentation is added.
 Validated end-to-end with a GSM8K throughput/accuracy benchmark (max output tokens = 1500, concurrency = 128, `--max-num-seqs 128`). Example commands:

  ```shell
  # Static K=3
  vllm serve <model> --max-num-seqs 128 \
    --speculative-config '{"method":"eagle3","model":"<draft>","num_speculative_tokens":3}'

  # Static K=1
  vllm serve <model> --max-num-seqs 128 \
    --speculative-config '{"method":"eagle3","model":"<draft>","num_speculative_tokens":1}'

  # Dynamic SD (this PR): K=3 for BS 1-64, K=1 for BS 65-128
  vllm serve <model> --max-num-seqs 128 \
    --speculative-config '{"method":"eagle3","model":"<draft>","num_speculative_tokens":3,"num_speculative_tokens_per_batch_size":[[1,64,3],[65,128,1]]}'
  ```

  **Throughput (GSM8K, max_tokens=1500, concurrency=128)**

  | Mode | K schedule | Input Throughput (tok/s) | Output Throughput (tok/s) |
  | --- | --- | --- | --- |
  | Static SD | K = 3 | 40.78 | 641.33 |
  | Static SD | K = 1 | 163.46 | 2570.59 |
  | **Dynamic SD (this PR)** | 1–64 -> K=3, 65–128 -> K=1 | **193.16** | **3037.66** |

  At 128 concurrency, DSD improves output throughput by **~18% over the best static config (K=1)** and **~4.7× over static K=3** - the latter showing exactly why a fixed large `K` degrades performance once `BS*K` goes 
  compute-bound.

  **Accuracy (GSM8K)**

  | Mode | Accuracy |
  | --- | --- |
  | Without DSD (static SD) | 89% |
  | With DSD (this PR) | 90% |

  Accuracy is preserved - the 89% vs 90% difference is within run-to-run noise on GSM8K; DSD introduces no accuracy regression.

- vLLM version: v0.26.0


zhenwenqi2024 (approved):

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -772,6 +772,26 @@ def _propose(
...
+        # .num_spec_tokens_to_schedule``. This mirrors upstream
+        # ``SpecDecodeBaseProposer.propose``, which receives the per-step K as
+        # an explicit argument. The model runner already forwards the
+        # ``scheduler_output`` to us, so we recover the per-step K here and
+        # refresh the derived ``decode_threshold`` accordingly. All downstream
+        # loops / reshapes read ``self.num_speculative_tokens``, so updating it
+        # once at the entry propagates through the rest of ``_propose``.
+        if scheduler_output is not None:
+            self.num_speculative_tokens = scheduler_output.num_spec_tokens_to_schedule
+            self.decode_threshold = 1 + self.num_speculative_tokens
```
HF-001: Decode_threshold=1+ num_speculative_tokens is not an isolated variable, it participates in buffer slicing and graph allocation, and these buffer/ACLGraph are pre allocated according to the maximum K. For example, I saw in the source code that the calculation of slice length is similar to num_reqs_padded * self.decode_threshold. Changing the decode_threshold step by step without restoring it may not match the 'buffer/graph allocated according to the maximum K', resulting in under counting of slices or incorrect allocation of graph keys

HF-001:
@zhenwenqi2024 @chengduxiaowu hi, Hello everyone, I have already done this part of the work and submitted the PR  https://github.com/vllm-project/vllm-ascend/pull/12368. I think there are still issues with the decode_threshold and pcp aspects of this PR.

vllm_ascend/spec_decode/llm_base_proposer.py
chengduxiaowu: Hi HF-001,

Thanks for your careful review. You are absolutely right that `decode_threshold = 1 + num_speculative_tokens` is not an isolated variable – it directly affects buffer slicing and ACLGraph allocation.

chengduxiaowu: Thanks for the detailed review! You're absolutely right — decode_threshold is not an isolated variable. It is deeply coupled with pre-allocated buffers and ACLGraph capture sizes, all of which are sized according to the maximum K (1 + num_speculative_tokens).
After carefully tracing through the source code, I've identified the clear boundary between where decode_threshold must remain static and where a runtime dynamic K can be safely used:
Must remain static (tied to max K):
Buffer pre-allocation in __init__: e.g., dcp_mtp_attn_mask shape [max_num_reqs, decode_threshold, max_model_len]
ACLGraph capture sizes and graph key computation: e.g., slice lengths like num_reqs_padded * self.decode_threshold
mask_spec_decode_restore_idx_for_graph: the mask period pcp_world_size * decode_threshold must match the graph-captured layout
classify_decode_request_mask: used as a safe upper-bound classifier (backed by done_prefilling)
Can safely use runtime dynamic K (no buffer/graph dependency):
generate_pcp_mtp_input: the if decode_threshold > 2 guard can be replaced with if current_k > 1 to skip the entire MTP expansion block when K ≤ 1, avoiding unnecessary numpy ops and GPU tensor allocation (mtp_slot_pad)
generate_pcp_metadata: the MTP mask generation + copy_to_gpu can be skipped entirely when current_k == 0, saving a large H2D transfer
generate_mtp_attention_mask_for_decode: already uses dynamic slicing internally (max_q = max(decode_num_scheduled_tokens)), so the filled rows naturally shrink with dynamic K — no change needed
So the approach would be: keep self.decode_threshold as the static max-K upper bound for all buffer/graph allocations, and only pass a separate current_num_spec_tokens (runtime dynamic K) to the specific functions above for early-exit / skip optimizations. This way, buffer slicing and graph keys always match the pre-allocated max-K layout, while we still avoid unnecessary computation when dynamic K is small.
Does this align with your concern? Happy to adjust the implementation accordingly.

PR #12704, opened 2026-07-23
[BugFix][Spec Decode] Isolate eager draft compilation config

yiz-liu:

The original issue was an assertion failure when using an eager speculative draft with PIECEWISE CUDAGraph and FlashComm: `cudagraph_capture_sizes [] does not contain values that are multiples of tp_size 16`. The eager draft context changed the target model's shared `CompilationConfig` in place, and draft config initialization could then clear CUDAGraph capture sizes on the same object. As a result, the target model's PIECEWISE graph configuration was corrupted and the SP capture-size assertion failed.

`FULL_DECODE_ONLY` is not affected because it does not depend on compilation, so the upstream `check_and_update_config` path does not rewrite the related compilation fields when constructing the draft config. PIECEWISE depends on compilation and therefore exposes the shared `CompilationConfig` mutation.

The ideal fix is to resolve the draft-model compilation incompatibility and remove this eager workaround. Given the current time constraint, this PR takes a minimal approach: use a temporary copy of `CompilationConfig` while creating the eager draft, and skip SP capture-size validation when CUDAGraph mode is `NONE`.

This is correct because the target model keeps its original capture settings, while the runtime registries required by model layers remain shared with the draft. The original config object is restored after draft creation, including on exceptions.

No. This PR introduces no configuration or parameter changes; it only fixes the incorrect handling of the existing configuration.

A regression test was added, and the fix passed manual acceptance testing.

### DIFF
--- a/tests/ut/test_platform.py
+++ b/tests/ut/test_platform.py
@@ -224,6 +224,40 @@ class TestNPUPlatform(TestBase):
 
         self.assertEqual(observed_inputs, [77])
 
+    @patch("vllm_ascend.platform.refresh_block_size")
+    @patch("vllm_ascend.platform.get_ascend_device_type", return_value=AscendDeviceType.A3)
+    @patch("vllm_ascend.platform.enable_sp", return_value=True)
+    @patch("vllm_ascend.ascend_config.init_ascend_config")
+    @patch("vllm_ascend.quantization.utils.maybe_auto_detect_quantization")
+    def test_check_and_update_config_skips_sp_capture_sizes_without_cudagraph(
+        self,
+        mock_auto_detect,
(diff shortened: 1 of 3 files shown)


weijinqian0 (approved):

yiz-liu:
/cherry-pick releases/v0.24.0rc
[Bot]: cherry-pick completed successfully. New PR created: https://github.com/vllm-project/vllm-ascend/pull/12721

yiz-liu:
/cherry-pick releases/v0.23.0
[Bot]: cherry-pick completed successfully. New PR created: https://github.com/vllm-project/vllm-ascend/pull/12722

PR #12717, opened 2026-07-23
[CI] Restructure A2 nightly test order and add taint cleanup for A2/A3

zhangxinyuehfad:

A2 nightly resources were being preempted by other workloads during the single-node test phase, causing multi-node tests to fail due to insufficient dedicated nodes.

- Default taint applied at 22:30, A2 manual fallback at 2:45, A3 at 6:00

- A2: Move multi-node-tests to run before single-node-tests. Add remove-taints between multi-node and single-node phases to clean Kubernetes node taints proactively, ensuring single-node tests are not blocked waiting for the 2:45 fallback cleanup window.

- A3 / A3 : Add remove-taints at end of all tests for post-test cleanup.

- vLLM version: v0.25.1

### DIFF
--- a/.github/workflows/schedule_nightly_test_a2.yaml
+++ b/.github/workflows/schedule_nightly_test_a2.yaml
@@ -176,8 +176,8 @@ jobs:
       HW_TOKEN: ${{ secrets.HW_TOKEN }}
       GITEE_TOKEN: ${{ secrets.GITEE_TOKEN }}
 
-  single-node-tests:
-    name: single-node
+  multi-node-tests:
+    name: multi-node
     needs: [setup-vars, parse-trigger, build-image]
     if: >-
       always() &&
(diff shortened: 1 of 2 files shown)


.github/workflows/schedule_nightly_test_a2.yaml
```diff
@@ -176,29 +176,33 @@ jobs:
...
+  multi-node-tests:
+    name: multi-node
     needs: [setup-vars, parse-trigger, build-image]
     if: >-
       always() &&
       needs.parse-trigger.outputs.run == 'true' &&
       (needs.build-image.result == 'success' || needs.build-image.result == 'skipped')
     strategy:
       fail-fast: false
+      max-parallel: 3
```
menogrey, line 188: Why changing the max-parallel of multi-node-tests.

zhangxinyuehfad: Currently, there are only 3 multi-machine testing use cases, and only 6 machines are used for taint testing. The subsequent increase in use cases will increase the degree of parallelism.

.github/workflows/schedule_nightly_test_a2.yaml
```diff
@@ -289,7 +307,7 @@ jobs:
           python3 .github/workflows/scripts/resolve_nightly_tests.py --mode=matrix
 
   single-node-accuracy-tests:
-    needs: [setup-vars, parse-trigger, build-image, generate-accuracy-matrix]
+    needs: [setup-vars, parse-trigger, build-image, generate-accuracy-matrix, multi-node-tests]
```
menogrey: Why single-node-tests depends on remove-taints, but the single-node-accuracy-tests depends on multi-node-tests

zhangxinyuehfad: yep, single-node-accuracy-tests should also depends on remove-taints .

menogrey:
LGTM

wangxiyuan (approved):

Issue #12723, opened 2026-07-23
[Bug]: Ascend-dispatched rotary path (Triton rope) corrupts Q/K norms for GLM-5.2 DSpark draft (bf16, NeoX, head_dim=192); forward_native restores accept_len 3.06 -> 5.55


Frozen7771:
### Your current environment

<details>
<summary>Environment (collected inside the serving container; the container is offline so collect_env.py could not be downloaded)</summary>

```text
Ascend A3 (16 dies), driver 25.5.1
Image: quay.io/ascend/vllm-ascend:nightly-releases-v0.23.0-a3

Python 3.12.13
torch 2.10.0+cpu   (standard Ascend arrangement: CPU torch build, NPU backend via torch_npu)
torch_npu 2.10.0.post2
vllm 0.23.0
vllm_ascend: source tree bundled in the image at /vllm-workspace/vllm-ascend (no __version__ attribute)
triton 3.2.0
CANN: as bundled in the image

vllm_ascend/ops/rotary_embedding.py: HAS_TRITON = True
-> rope_forward_oot routes into rope_forward_triton (vllm_ascend/ops/triton/rope.py),
   not torch_npu._npu_rotary_embedding, for this stack.
```

</details>

## Summary

While bringing up GLM-5.2 DSpark speculative decoding on Ascend A3 (16 dies, DP2/TP8/EP, eager), we found that the **platform-dispatched rotary embedding path returns non-norm-preserving Q/K** for the DSpark draft attention shape (bf16, NeoX-style, `head_dim = rotary_dim = 192`). Calling `forward_native()` on the **same rotary object, same worker, same inputs** produces correct results and restores online acceptance length from ~3.0 to ~5.4–5.5.

This is not a checkpoint / config / acceptance-logic problem (cf. #12262, which fixes different root causes — Markov logits input, anchor sampling, QuaRot weight rotation). It is a numerical error in the dispatched rotary execution path itself.

## Environment

- Ascend A3, 16 dies, driver 25.5.1
- Image: `quay.io/ascend/vllm-ascend:nightly-releases-v0.23.0-a3` (vllm-ascend bundled as a source tree in the image)
- Python 3.12.13, `vllm 0.23.0`, `torch 2.10.0+cpu` + `torch_npu 2.10.0.post2` (standard Ascend arrangement: CPU torch build with the NPU backend provided by torch_npu), `triton 3.2.0`, CANN as bundled in the image
[...]

PR #11431, opened 2026-07-04
[Feature][Refactor] DSv4 DSpark

drslark:
### What this PR does / why we need it?

It is planned in https://github.com/vllm-project/vllm-ascend/issues/11163.
Refactor https://github.com/vllm-project/vllm-ascend/pull/11196 by @QwertyJack.
Following sample features from https://github.com/vllm-project/vllm-ascend/pull/11765 by @chenaoxuan.
Aligning with vllm's https://github.com/vllm-project/vllm/pull/46995.

Thanks for @QwertyJack for the rapid rough adaption of DSv4 DSpark.
Thanks for @chenaoxuan for the overall sampling architecture.

What does it follow?
- DSpark draft-modeling for DSv4.
- A trial-and-error version `sas` AscendC operator.
- Most logic of `DSparkProposer`.
- Anchor-first block drafting, sequential Markov sampling.

What does it refactor?
- Remove a lot of private branches and meaningless fallback ai codes.
- Remove a lot of non-compliant pass-through parameters in dsa_v1.
- Remove all redundant helper methods.
- Reuse the vllm layer and dsa_v1 interface as much as possible.
- Merge most of common logic between `Qwen3-DSpark` and `DSv4-DSpark`.
- Remove all v2 features like probabilistic sampler and double buffer input_buffers.

What does it add?
- Adapt for the `quarot-w4a8` weight of `DSv4-DSpark`.

Future works.
- Add ci for `DSv4-DSpark` when this model is ready in modelscope.

### Does this PR introduce _any_ user-facing change?

N/A.

### How was this patch tested?

We test 30 samples from gsm8k:

`Qwen3-DSpark` (without template):

```text
% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 53545  100 53545    0     0  3923k      0 --:--:-- --:--:-- --:--:-- 4022k
**********
6478.0 [5317.0, 4230.0, 3321.0, 2629.0, 2103.0, 1616.0, 1241.0]
**********
accepted: [5317.0, 4230.0, 3321.0, 2629.0, 2103.0, 1616.0, 1241.0]
num_drafts: 6478.0
acceptance rate: [0.8207780179067613, 0.6529793146032726, 0.5126582278481012, 0.4058351343007101, 0.3246372337141093, 0.2494597097869713, 0.1915714726767521]
acceptance_len: 4.157919110836678
```

`DSv4-DSpark`:

```text
[...]

PR #11066, opened 2026-06-27
[Feature][DSpark] Add GLM-5.2 DSpark speculative decoding

Liuchenbing-2026:
### What this PR does / why we need it?

Add GLM-5.2 DSpark speculative decoding to vLLM-Ascend (0.23), aligned with the upstream vLLM DSpark design.

Adds a self-contained DSpark draft model for the Qwen3-based GLM-5.2:

- `vllm_ascend/models/qwen3_dspark.py`: DFlash Qwen3 backbone + DSpark Markov / confidence heads + `compute_dspark_draft_tokens`. Derived from upstream vLLM `qwen3_dflash.py` / `qwen3_dspark.py` (same Apache-2.0 header), with Ascend-native RMSNorm / rotary fallbacks and an in-model sampling path.
- Model registration in `vllm_ascend/__init__.py` and `vllm_ascend/models/__init__.py`.
- DSpark wiring in the spec-decode proposers (`dflash_proposer.py`, `eagle_proposer.py`, `llm_base_proposer.py`) and `worker/model_runner_v1.py`.
- DSpark QuaRot: FC load-time weight transform (`vllm_ascend/models/dspark_quarot.py`) with unit tests.

Scope: changes are under `vllm_ascend/*` only; vLLM site-packages are untouched.

### Does this PR introduce _any_ user-facing change?

Adds DSpark draft-model support for GLM-5.2; non-DSpark workloads are unchanged.

### How was this patch tested?

Functionally runs on Ascend A3. DSpark vs MTP3 aisbench comparison (aggregated metrics):

#### Test Scope

- Model: GLM-5.2 W8A8
- Topology: DP2 TP8, 16 NPUs
- Input length: about 1k tokens
- Requests / concurrency: 32 / 8
- Request rate: 0
- DSpark: `num_speculative_tokens=7`, `method=dflash`
- MTP3: `num_speculative_tokens=3`, `method=mtp`

#### Performance

| Output Len | Mode | Output TPS | E2E TPS | QPS | E2E Time (s) | TTFT Avg (ms) | TPOT Avg (ms) | TPOT P90 (ms) |
|---:|---|---:|---:|---:|---:|---:|---:|---:|
| 300 | DSpark | 113.0854 | 494.5600 | 0.3770 | 84.8916 | 1177.2 | 63.8 | 71.3 |
| 300 | MTP3 | 84.6665 | 370.2747 | 0.2822 | 113.3861 | 1287.6 | 87.3 | 95.5 |
[...]

PR #12726, opened 2026-07-23
[BugFix][GLM-5.2][SpecDecode] Use forward_native for DSpark draft query RoPE on Ascend (stacked on #11066)

Frozen7771:
> **Stacked on #11066** (`Liuchenbing-2026:dspark` @ `fabc77a9`): this branch contains #11066's commits plus **one new commit on top — only the last commit is under review here**. Happy to rebase once #11066 lands, or to retarget the fix onto whichever GLM-5.2 DSpark line settles post-#11431.

Fixes a global acceptance collapse of GLM-5.2 DSpark on Ascend A3 caused by the
platform-dispatched rotary path returning non-norm-preserving Q/K for the draft
attention shape (bf16, NeoX-style, head_dim=rotary_dim=192). Root-cause evidence
chain (same-worker dispatched-vs-native A/B, two fresh-process reproductions,
token-level causality, cache exoneration) is in #12723

The change is a single line in the draft attention forward: call
`self.rotary_emb.forward_native(positions, q, k)` instead of the dispatched
`self.rotary_emb(positions, q, k)`. `forward_native` still executes on the NPU
(as unfused tensor ops); this is a correctness fix, not a performance
optimization. A shape-gated fix in the Ascend rotary backend itself would be the
long-term solution.

No. Same API, same weights, same config. Only the draft query RoPE execution
path changes.

E2E on Ascend A3 (16 dies, DP2/TP8/EP, eager, k=7, GLM-5.2 w8a8 target,
bf16 DSpark draft), identical requests before/after:

| metric | before | after |
|---|---:|---:|
| accept_len, 8 serial requests | 3.055556 | **5.547692** |
| accept_len, 32 requests / concurrency 8 | 2.910313 | **5.429936** |

40/40 requests HTTP 200, hard counters consistent (`draft_tokens = 7 × drafts`),
post-run health check green. Offline accept_len for the same checkpoint on the
same question domain is ~5.6, i.e. online now matches offline capability.

End-to-end throughput comparison (no-spec vs MTP-5 vs DSpark) is **in
progress**; numbers will be posted in this thread. A NeoX-192 bf16
norm-preservation unit test (production path vs `forward_native` within
tolerance) is planned — note a plain-CPU CI cannot reproduce the corrupted
dispatch, so the numerical regression test belongs in NPU e2e/hardware CI.
- vLLM version: v0.25.1

### DIFF
--- a/tests/ut/compilation/a2/test_acl_graph.py
+++ b/tests/ut/compilation/a2/test_acl_graph.py
@@ -33,6 +33,7 @@ from vllm_ascend.compilation.acl_graph import (
     GraphParams,
     get_draft_graph_params,
     get_graph_params,
+    reset_graph_params,
     set_draft_graph_params,
     set_graph_params,
     update_draft_graph_params_workspaces,
@@ -808,6 +809,26 @@ class TestSleepGraphParams(TestBase):
 
 
 class TestDraftGraphParams(TestBase):
+    def test_reset_graph_params(self):
+        graph_params = GraphParams(
+            events={},
+            workspaces={},
+            handles={},
+            attn_params={},
+        )
+        with (
+            patch("vllm_ascend.compilation.acl_graph._graph_params", graph_params),
+            patch("vllm_ascend.compilation.acl_graph._draft_graph_params", graph_params),
+            patch(
+                "vllm_ascend.compilation.acl_graph._draft_graph_prefill_params",
+                graph_params,
+            ),
+        ):
+            reset_graph_params()
+            self.assertIsNone(acl_graph._graph_params)
+            self.assertIsNone(acl_graph._draft_graph_params)
+            self.assertIsNone(acl_graph._draft_graph_prefill_params)
+
     def test_set_draft_graph_params(self):
         with patch("vllm_ascend.compilation.acl_graph._draft_graph_params", new=None):
             set_draft_graph_params([4])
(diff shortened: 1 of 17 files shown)


Frozen7771:
Status note for reviewers:

1. **Scope**: this PR is stacked on #11066 (`Liuchenbing-2026:dspark` @ `fabc77a9`); only the last commit (c64e5ae) is under review here — a one-line switch of the DSpark draft query RoPE to `forward_native`. Root-cause evidence is in #12723; the mainline Triton kernel tiling fix is tracked in #12733 / #12740.
2. **Merge conflicts / DCO**: both are inherited from the stacked base — the conflicts are between #11066's commits and current main, and the two commits flagged by DCO (57f24f2, a11e2e4) are from the stacked branch; this PR's own commit is signed off. I'll rebase once #11066 (or its successor after #11431) lands, which clears both.
3. **Gemini's inline comments** all target code introduced by #11066 (RMSNorm fallback precision, QuaRot `fc.weight` guards, env-var caching in hot paths), not this PR's change.

Frozen7771:
Heads-up: the root cause behind this workaround has been identified and fixed properly in #12963 (sin_offsets computed from the padded rather than the real rope_dim). Once that lands I'll close this PR — the forward_native fallback here is no longer needed and would only cost performance. Will re-verify accept_len on A3 after #12963 merges.

PR #12757, opened 2026-07-24
[Feat][KV Offload] Implement layerwise and sparse KV cache offloading

ader47:

> [!NOTE]
> This PR is a implementation of
> [vLLM RFC #48203: Layerwise and Sparse KV cache offloading to support longer sequence length](https://github.com/vllm-project/vllm/issues/48203).
>

## Motivation

For long-sequence inference, KV cache capacity is often one of the main limits on maximum sequence length and batch size.

Sparse-attention models provide an opportunity to reduce this device-memory usage. Although the full KV cache must be preserved, decode attention only consumes the KV rows selected by the sparse indexer. Therefore, the full main attention KV cache can remain in host memory while only the selected top-k rows are loaded into a small device-resident buffer.

This PR combines two complementary approaches proposed in the RFC:

1. **Layerwise KV offloading for prefill**
   - A bounded number of device KV buffers are shared by multiple transformer layers.
   - KV transfers are overlapped with layer execution.
   - Device memory consumption no longer grows linearly with the number of model layers.

2. **Sparse KV offloading for decode**
   - The full main SFA KV cache is stored in host memory.
   - The LightningIndexer cache remains device-resident.
   - Only the selected top-k KV rows are loaded into bounded device hot buffers.
   - KV rows already present in the hot buffer can be reused across decode steps.

Together, these approaches reduce KV cache HBM usage in both prefill and decode, allowing the saved device memory to be used for longer sequences or larger batches.

### 1. Layerwise KV offloading for prefill

The existing AscendStore layerwise transfer path is extended with bounded KV buffer reuse.

Instead of allocating one device KV tensor for every model layer:

- Multiple transformer layers share a bounded set of physical KV buffers.
- Selected layers may retain independent buffers.
- MTP and speculative-decoding cache layers are included in the layer mapping.
- A shared layer reloads its cached prefix before attention because a previous layer may have overwritten the physical buffer.
- Save-completion gates prevent a buffer from being reused before its previous contents have been transferred.
- Layer transfers are prefetched and overlapped with model execution.
- Transfer planning supports per-group block mappings and layer-specific memory offsets.

The existing allocation behavior is preserved when layer reuse is not enabled.

### 2. Sparse KV offloading for decode

On a PD decode node:

- The full main SFA K/V cache is allocated from a pinned host-memory pool.
- The full LightningIndexer cache remains on the NPU.
- Each offloaded SFA layer owns a bounded device hot buffer.
- Hot-buffer memory is accounted for during model profiling.
- Newly generated K/V rows are written to the host-resident cache.
- The indexer produces logical top-k token positions.
- The offload runtime checks which selected KV rows are already resident.
- Cache misses are loaded from host memory into the hot buffer.
- Logical token positions are remapped to physical hot-buffer slots before sparse attention executes.

This keeps the full sequence KV state available without requiring the complete main SFA cache to remain in HBM.

### 3. PD KV transfer

This PR introduces a dedicated `SFAPDCpuOffloadConnector` instead of changing the behavior of existing offloading connectors.

The connector uses an asymmetric prefill/decode design:

- The prefill node produces the normal layer KV cache and indexer cache.
- The decode node advertises the destination memory layout.
- The decode node pulls the main SFA KV directly into its host-memory pool.
- The LightningIndexer cache is transferred into device memory.
- Main and indexer caches may use independent block mappings.
- Layer-ready and transfer-completion events prevent source buffers from being reused too early.
- Main-cache transfers are divided across decode TP ranks.
- Request completion waits for all participating TP ranks.
- Transfer failures are propagated across TP ranks and reported to the scheduler.

The connector also supports requests containing multiple prompts by preserving per-child transfer metadata while dispatching the original prompt batch to the prefill node only once.

### 4. Attention backend integration

The normal SFA execution path remains unchanged when decode offloading is disabled.

When offloading is enabled, the existing Ascend SFA backend selects an offload-aware metadata builder and implementation. This isolates the offload-specific behavior while reusing the existing SFA preprocessing, top-k selection, and attention execution where possible.

The offload-aware path is responsible for:

- Correctly classifying the first PD decode boundary token.
- Writing newly generated KV rows into the host-memory pool.
- Loading top-k cache misses into the resident device buffer.
- Remapping logical top-k positions to resident-buffer slots.
- Running SFA against the resident top-k buffer.

No model architecture or weight-loading changes are introduced.

### 5. Sparse C8 cache separation

The sparse C8 capability is separated between the main SFA cache and the LightningIndexer cache.

This distinction is required because decode CPU offloading can retain a C8 LightningIndexer cache on the device, while the host-resident main SFA cache currently requires BF16.

## Scope and design decisions

This PR intentionally implements PD-disaggregated serving first.

Supporting PD-colocated execution would require the same worker to manage the normal device KV cache, indexer cache, host-resident cache, and top-k hot buffers. That would require broader hybrid KV cache management changes and make the initial implementation substantially more intrusive.

For the current PD-disaggregated design:

- The prefill node owns the layerwise main and indexer caches.
- The decode node owns the device indexer cache, host-resident main cache, and fixed top-k buffers.
- Top-k buffers are allocated outside the normal paged KV cache and accounted for during memory profiling.
- A dedicated connector encapsulates the asymmetric transfer protocol.
- Existing execution paths remain unchanged when the feature is disabled.

This PR currently targets the V1 model runner. V2 support and generalized PD-colocated support are outside the scope of this draft.

Detailed deployment instructions and configuration documentation will be provided separately.

## Additional changes

- Extend the layerwise PD proxy to coordinate batched requests and per-child KV transfer metadata.
- Add a standalone MemFabric pull-read diagnostic for NPU-to-host and NPU-to-NPU validation.
- Improve connector shutdown, retry, failure propagation, and request cleanup.
- Remove obsolete layerwise KV lookup and initialization paths.
- Extend unit-test coverage for layerwise and sparse KV offloading.

Yes.

This PR introduces:

- Layerwise KV buffer reuse for prefill.
- Decode-side CPU KV offloading for sparse-attention models.
- A dedicated PD KV transfer connector.
- Independent handling of the main SFA and LightningIndexer cache layouts.

Detailed configuration and usage instructions will be provided in follow-up documentation.

Current limitations:

- Only the V1 model runner is supported.
- Decode offloading only supports sparse-attention/SFA models.
- Production use currently requires PD-disaggregated serving.
- MemFabric is currently required for the new PD transfer path.
- Context-parallel SFA is not supported by decode offloading.
- The host-resident main SFA cache currently requires BF16.
- Layerwise transfer currently requires the AscendStore memcache backend.
- Layerwise buffer reuse does not support multiple KV cache group families.

- vLLM version: v0.26.0


vllm_ascend/worker/worker.py
```diff
@@ -580,10 +587,59 @@ def determine_available_memory(self) -> int:
...
+            if memory_info is not None:
+                num_layers, num_slots, factor = memory_info
+            else:
+                num_layers = getattr(
+                    self,
+                    "_gva_layerwise_num_layers",
+                    self.model_config.get_num_layers(self.parallel_config),
+                )
+                num_slots = get_layerwise_kv_cache_num_tensors(num_layers, extra_config)
+                factor = num_layers / num_slots if num_slots is not None else 1.0
```
zhangsicheng5, line 576: consider num_layers in pp case

vllm_ascend/distributed/kv_transfer/sfa_pd_cpu_offload/__init__.py
```diff
@@ -0,0 +1,16 @@
+#
```
zhangsicheng5, line 1: move to kv_p2p direction

vllm_ascend/attention/sfa_v1.py
```diff
@@ -1543,8 +1557,20 @@ def _update_indexcache_topk_indices(self, topk_indices: torch.Tensor) -> None:
...
-        self, ql_nope, q_pe, kv_cache, topk_indices, attn_metadata, actual_seq_lengths_query, actual_seq_lengths_key
+        self,
+        ql_nope,
+        q_pe,
+        kv_cache,
+        topk_indices,
+        attn_metadata,
+        actual_seq_lengths_query,
+        actual_seq_lengths_key,
+        block_table=None,
```
weiguihua2, line 1557: Does cp support it? sfa_cp also inherits the method _execute_sparse_flash_attention_process. Does it need to be modified?

ader47:
This PR has been split into several smaller PRs to make the changes easier to review and merge:

- #13026 — Sparse KV cache offloading for decode
- #12852 — Layerwise prefill KV buffer reuse
- #12853 — MTP and sparse C8 support for layerwise prefill offload
- #12831 — SFA PD RD2H connector for KV cache offloading

All of the above PRs have now been merged, so this original PR has been superseded and is no longer needed.

Closing this PR in favor of the smaller merged PRs. Thanks everyone for the reviews and discussions!

PR #12761, opened 2026-07-24
[1/N][Refactor][Core] remove function find_longest_cache_hit_per_group in patch and support DSV4 decode partial group hits

underfituu:
#### Problem

Hybrid models can have different decode-node prefix-cache hit lengths across KV cache groups. The existing all-group reconciliation uses a boundary that every group can satisfy. For remote-prefill P/D serving, a sliding-window, Mamba/state, or MTP/EAGLE group may report a shorter hit, often zero, even when a reusable full-attention prefix is already present on the decode node. That collapses the decode-node local hit and causes the connector to request KV that the full-attention group already has.

Align to https://github.com/vllm-project/vllm/pull/48425 and https://github.com/vllm-project/vllm/pull/45804

This affects the two target layouts in this PR:

- Qwen3.5 + MTP: the full-attention group can hit while the Mamba/state and MTP groups remain shorter.
- DeepSeek-V4 + MTP: the C4 and C128 full-attention groups can hit while the remaining hybrid groups report zero. The C128 group must also be truncated with its own effective boundary.

#### Solution requirements

The implementation follows these requirements:

1. Apply partial-group selection only to remote-prefill requests handled by a hybrid KV cache coordinator on the decode node.
2. Keep the normal cache-off and non-connector paths unchanged. Local lookup must still honor `enable_prefix_caching` and `skip_reading_prefix_cache`.
3. Use the first full-attention group as the dense reference group. Its hit is downward-closed and is the safe local boundary used by the decode scheduler.
4. Query the upstream per-group lookup API and select the dense reference hit only when no other group reports a hit longer than the reference. A longer non-reference hit means the per-group union is not consistent at one boundary and must not be selected as a partial-group hit.
5. Treat the result as a partial-group candidate only when at least one other group is shorter than the dense reference group.
6. Ask the KV connector only for the remaining external tokens. If the connector reports zero external tokens for a divergent candidate, fall back to the normal all-group-consistent lookup rather than accepting an incomplete local-only boundary.
7. Preserve MTP/EAGLE semantics when grouping identical cache specs and when dropping the EAGLE block during lookup.
8. Truncate every DeepSeek-V4 full-attention group, including C4 and C128, to the reconciled hit length using that group's own effective block size. No full-attention group may retain blocks beyond the final safe boundary.
9. Record prefix-cache statistics through the current vLLM interface and keep structured candidate/decision logs for mechanism validation.
10. Do not add a product feature flag. Cache-off/cache-on remains controlled by the existing decode-node `--enable-prefix-caching` setting.

Yes, it changes decode-node cache behavior for hybrid remote-prefill serving when decode-node prefix caching is enabled:

- The decode node can reuse a safe full-attention local prefix even when other hybrid groups have shorter hits.
- The connector transfers only the remaining KV tokens.
- Cache-off behavior, request/response APIs, and model configuration interfaces do not change.

#### Qwen3.5 P/D configuration

Model: `/mnt/weight/Qwen3.5-35B-A3B-w8a8-mtp`

| Item | Prefill node | Decode node |
|---|---|---|
| Devices | `0,1,2,3` | `4,5,6,7` |
| Parallelism | TP2, DP2, EP enabled | TP2, DP2, EP enabled |
| Service port | `30060` | `30050` |
| Prefix caching | Enabled | Cache-off: disabled; cache-on: enabled |
| KV connector | `MooncakeConnectorV1`, `kv_producer`, port `23010` | `MooncakeConnectorV1`, `kv_consumer`, port `36010` |
| Hybrid KV manager | Enabled | Enabled |
| MTP | Enabled, one speculative token | Enabled, one speculative token |
| Maximum model length | `32768` | `32768` |
| Maximum sequences | `64` | `64` |
| Maximum batched tokens | `4096` | `2048` |
| Execution | Eager | Eager + recompute scheduler |
| Other key settings | CPU binding, fused MC2 disabled, seed 1024 | CPU binding, fused MC2 disabled, seed 1024 |

The load-balancing proxy listened on port `8010` and routed prefill to `30060` and decode to `30050`.

#### DeepSeek-V4 P/D configuration

Model: `/mnt/weight/DeepSeek-V4-Flash-w8a8-mtp`

| Item | Prefill node | Decode node |
|---|---|---|
| Devices | `0-7` | `8-15` |
| Parallelism | TP8 | TP8 |
| Service port | `30060` | `30050` |
| Prefix caching | Enabled | Cache-off: disabled; cache-on: enabled |
| KV connector | `MooncakeHybridConnector`, `kv_producer`, port `23010` | `MooncakeHybridConnector`, `kv_consumer`, port `36010` |
| Hybrid KV manager | Enabled, block size 32 | Enabled, block size 32 |
| KV cache | FP8, `18360985190` bytes | FP8, `18360985190` bytes |
| MTP | Enabled, one speculative token | Enabled, one speculative token |
| Maximum model length | `132096` | `132096` |
| Maximum sequences | `16` | `60` |
| Maximum batched tokens | `8192` | `120` |
| Execution | Eager | Async scheduling + `FULL_DECODE_ONLY` graph |
| Other key settings | DeepSeek-V4 tokenizer/tool/reasoning parsers, 128-thread loader, CPU binding, fused MC2 disabled | Same parsers/loader, recompute scheduler, NPU graph extension, CPU binding, fused MC2 disabled |

The load-balancing proxy listened on port `8010` and routed prefill to `30060` and decode to `30050`.

#### Partial-group mechanism validation

| Model | Cache-off decode node | Cache-on decode node |
|---|---|---|
| Qwen3.5 + MTP | Local prefix-cache hit rate `0.0%`; no partial-group decision | 508 candidates and 508 decisions; `per_group_hits=(14336, 0, 0, 0)`; `selected_local_tokens=14336`; all `fallback=False`; observed local hit rate about `80.8%-83.5%` |
| DeepSeek-V4 + MTP | Local prefix-cache hit rate `0.0%`; no partial-group decision | 452 candidates and 452 decisions; all `fallback=False`; observed local hit rate increased from `49.8%` to `97.8%` |

DeepSeek-V4 additionally demonstrated both 131K and canonical-prefix boundaries:

```text
131K workload:  per_group_hits=(131584, 131072, 0, 0, 0, 0)
                 selected_local_tokens=131584
                 external_tokens=3

Canonical GPQA: per_group_hits=(10880, 8192, 0, 0, 0, 0)
```

The second full-attention group is truncated to `131072` or `8192`, both exact multiples of 4096, while the zero-hit hybrid groups no longer collapse the decode-node local hit to zero.

#### Performance comparison

https://github.com/vllm-project/vllm/pull/45804#issuecomment-5076112373

#### Accuracy comparison: GPQA canonical only

| Model | Cache-off | Cache-on | Accuracy delta | Paired answer changes |
|---|---:|---:|---:|---|
| Qwen3.5 + MTP, 64 workers | `162/198 = 81.82%` | `162/198 = 81.82%` | `0.00 pp` | 7 wrong-to-correct, 7 correct-to-wrong, 19 predictions changed |
| DeepSeek-V4 + MTP, 16 workers | `144/198 = 72.73%` | `146/198 = 73.74%` | `+1.01 pp` | 9 wrong-to-correct, 7 correct-to-wrong, 22 predictions changed |

Both cache states completed `198/198` requests with zero request errors. Prompt hashes and request hashes matched for all paired samples. Because generated reasoning is not token-for-token deterministic across repeated high-concurrency runs, GPQA is treated as a regression check; the score difference is reported but is not attributed as a cache-quality improvement.

#### Main/v0.26.0 sunset plan

This PR targets the vLLM main/v0.26.0 interface. It does not retain a v0.25.1 compatibility path.

The duplicated Ascend implementation of `find_longest_cache_hit_per_group()` is already removed in favor of the upstream API. The remaining Ascend-specific patch can be retired incrementally:

1. When upstream vLLM truncates every full-attention group with its own effective block size, remove the corresponding coordinator override from `patch_kv_cache_coordinator.py`.
2. When upstream remote-prefill scheduling natively supports safe dense-reference partial-group selection and zero-external fallback, remove `_get_computed_blocks_for_connector()` and the partial-group branch from `RecomputeScheduler`.

### DIFF
--- a/tests/ut/core/test_recompute_scheduler.py
+++ b/tests/ut/core/test_recompute_scheduler.py
@@ -5,8 +5,23 @@ from collections import defaultdict
 from types import MethodType, SimpleNamespace
 from unittest.mock import MagicMock
 
+import torch
 from vllm.sampling_params import SamplingParams
+from vllm.utils.hashing import sha256
+from vllm.v1.core.kv_cache_manager import KVCacheManager
+from vllm.v1.core.kv_cache_utils import (
+    get_group_id,
+    get_request_block_hasher,
+    init_none_hash,
+)
+from vllm.v1.core.single_type_kv_cache_manager import register_all_kvcache_specs
 from vllm.v1.engine import EngineCoreOutput, FinishReason
+from vllm.v1.kv_cache_interface import (
+    KVCacheConfig,
+    KVCacheGroupSpec,
+    MLAAttentionSpec,
+    SlidingWindowMLASpec,
+)
 from vllm.v1.request import Request, RequestStatus
 
 from vllm_ascend.core.recompute_scheduler import (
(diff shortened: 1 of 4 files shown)


linfeng-yuan (approved):
@underfituu @nwpu-zxr Please focus on the maintainence of RecomputeScheduler.

PR #12783, opened 2026-07-24
[v0.23.0][Performance] Optimize AscendStore key construction and miss path

DreamerLeader:

This supersedes #12223 and reapplies the AscendStore key-construction and
KVPool miss-path optimizations on top of the latest `releases/v0.23.0`.

- Generates serialized non-layerwise keys directly, caches immutable prefixes,
  and computes grouped hashes lazily.
- Centralizes token/chunk iteration and removes duplicated key-building paths.
- Propagates HBM hit tokens through lookup and avoids redundant external
  lookups and stores.
- Applies hybrid-cache lookup/store/load masks before key construction.
- Pre-shards filtered keys automatically when applicable.
- Uses asynchronous save by default and removes the old wait-mode lifecycle.
- Keeps the layerwise `PoolKey` transfer path unchanged.

Yes. AscendStore now selects the optimized key construction, mask handling,
pre-sharding, and asynchronous save paths automatically. The cache key format
is unchanged. Because save is asynchronous, a following identical lookup may
race with save visibility.

- `ruff check` passed for all 12 changed Python files.
- `ruff format --check` passed for all 12 changed Python files.
- Python syntax compilation passed for the three core conflict-sensitive files.
- `git diff --check` passed.
- Targeted pytest execution could not start in the local environment because
  `torch` is unavailable; CI/NPU validation is required.

### Co-authors

- Pz1116 <zpbzpb123123@gmail.com>
- zmc1997 <742133686@qq.com>

- vLLM version: v0.23.0

### DIFF
--- a/tests/ut/distributed/ascend_store/test_config_data.py
+++ b/tests/ut/distributed/ascend_store/test_config_data.py
@@ -155,7 +155,7 @@ class TestChunkedTokenDatabase(unittest.TestCase):
         hashes = [bytes([idx % 251]) * 32 for idx in range(128)]
 
         result = list(
-            db.process_tokens_with_block_ids(
+            db.process_token_key_strings_with_block_ids(
                 128 * 128,
                 hashes,
                 [1000, 1001, 1002, 1003],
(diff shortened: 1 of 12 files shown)


yiz-liu (approved):

Pz1116 (approved):
LGTM

PR #12852, opened 2026-07-25
[Feature][KV Offload] Add layerwise prefill KV buffer reuse

ader47:

Layerwise prefill offloading currently allocates a dedicated device KV cache buffer for every transformer layer. However, these buffers are not used concurrently during layerwise execution, resulting in unnecessary HBM reservation and limiting the available KV cache capacity.

This PR implements the prefill-side layerwise KV buffer reuse proposed in [RFC #48203](https://github.com/vllm-project/vllm/issues/48203). It allows multiple transformer layers to time-share a bounded set of physical device KV cache buffers while keeping the logical KV cache layout unchanged.

The main changes include:

- Reorganizing KV cache tensor descriptors before allocation so compatible transformer layers can share physical buffers.
- Keeping layers that cannot participate in buffer reuse on independent buffers.
- Updating KV cache memory accounting to expose the recovered HBM capacity.
- Reloading the required KV data before a shared buffer is reused.
- Ensuring a shared buffer is reused only after the previous layer has completed its save operation.
- Fixing the SFA execution path so the transfer gate and layer save hook are executed consistently, including for layers without an indexer.

This is the first PR in a stacked series and only covers prefill-side buffer reuse for base transformer layers. MTP and sparse C8 support, multi-group support, the layerwise KV pool refactor, and decode-side sparse offloading are outside the scope of this PR.

Yes. The supported layerwise prefill offload path can use fewer physical HBM KV cache buffers, making more device memory available for KV cache capacity while preserving the existing connector interface.

Configuration and usage instructions will be documented separately.

- Added unit tests for physical buffer reuse planning and configuration validation.
- Added coverage for independent and reused layer loading behavior.
- Added coverage for connector scoping.
- Added coverage for SFA layers without an indexer.
- Passed Ruff lint and formatting checks.
- Passed Python compilation and `git diff --check`.
- Full pytest execution is pending CI because the local environment does not provide `torch`.

- vLLM version: v0.26.0


vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/layerwise_cache_layout.py
```diff
@@ -0,0 +1,193 @@
...
+            continue
+        extra_config = connector_config.get("kv_connector_extra_config") or {}
+        if str(extra_config.get("backend", "mooncake")).lower() == "memcache" and extra_config.get(
+            "use_layerwise", False
+        ):
+            return extra_config
+    return None
+
+
+def _parse_int_config(value: Any, name: str) -> int:
```
Angazenn, line 66: Is this helper function necessary?

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/layerwise_cache_layout.py
```diff
@@ -0,0 +1,193 @@
...
+        num_prefetch_layers = min(num_shared_buffers, _DEFAULT_MAX_PREFETCH_LAYERS)
+    else:
+        num_prefetch_layers = _parse_int_config(prefetch_value, _PREFETCH_LAYERS)
+        if num_prefetch_layers < 1:
+            raise ValueError(f"{_PREFETCH_LAYERS} must be at least 1")
+
+    independent_value = extra_config.get(_INDEPENDENT_LAYERS) if extra_config else None
+    if independent_value is None:
+        layer_indices = [0, num_layers - 1]
+    elif isinstance(independent_value, str) and independent_value.strip().lower() == "all":
```
Angazenn, line 100: How should we configure independent_layers ?

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_scheduler.py
```diff
@@ -191,6 +194,12 @@ def __init__(
         else:
             self.put_step = 1
         self.num_layers = vllm_config.model_config.get_num_layers(vllm_config.parallel_config)
+        self.layerwise_offload = False
+        if self.use_gva_layerwise:
```
Angazenn, line 193: What if we use normal AscendStore kv pool but not use offload?

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/layerwise_cache_layout.py
ader47: already remove this function

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/layerwise_cache_layout.py
ader47: We can configure the independent layers via the `layerwise_indepedent_layers` param as mentioned in the documentation, https://github.com/vllm-project/vllm-ascend/pull/12852/changes#diff-230cd9743c733a3f4703cdb9cf4575f90650c22a1b7fbf9dacb7f41cf2885567R72

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_scheduler.py
ader47: This logic can handle this situation correctly.

realliujiaxu (approved):

PR #12868, opened 2026-07-25
[BugFix] Pack SFA DSA-CP KV tensors into one all-gather

pisceskkk:
In rare cases, issuing consecutive KV all-gathers can lead to first-forward correctness issues. For example, GLM5.2 can produce severely garbled output during the first forward pass when DSA_CP is enabled.

This PR packs the SFA KV payload, indexer KV payload, and indexer scale payload into one byte stream, submits a single KV all-gather, and restores each tensor to its original dtype and shape before the existing cache writes. This avoids multiple KV-related all-gather launches while preserving the separate o_proj weight gather path.

This is the main-branch counterpart of the v0.23.0 PR: #12867.

No

- python compile check for vllm_ascend/attention/sfa_v1.py and tests/ut/attention/a2/test_sfa_v1.py
- git diff --check
- Added unit coverage for mixed-dtype byte pack/restore and mismatched token-count validation

Local torch/torch_npu execution was not run in this workspace.

- vLLM version: v0.25.1


pisceskkk:
/cherry-pick releases/v0.24.0rc
[Bot]: cherry-pick command failed. Please check the [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/30334409982) for details.

PR #9690, opened 2026-05-29
[Worker][Doc] Add A5 server disaggregated PD endpoint configuration

Yuli-yx:
### What this PR does / why we need it?

Add A5 server disaggregated prefill-decode (PD) endpoint configuration support.

On A5 hardware, disaggregated PD requires per-device NPU endpoint JSON files to be loaded during worker initialization so that HCCL communication layers can correctly match the hardware topology.

Changes:

- `vllm_ascend/worker/worker.py`: In `init_device()`, when running on A5, read the per-device endpoint config directory from `kv_transfer_config.kv_connector_extra_config["ascend_local_comm_res_path"]` and set `ASCEND_LOCAL_COMM_RES` for HCCL in each worker process.

### Does this PR introduce _any_ user-facing change?

Yes. Users running disaggregated PD on A5 servers need to configure the endpoint config directory through `kv_connector_extra_config` before launching vLLM:

```json
{
  "kv_connector_extra_config": {
    "ascend_local_comm_res_path": "/etc/hixlep"
  }
}
```
- vLLM version: v0.20.2
- vLLM main: https://github.com/vllm-project/vllm/commit/9090368b650896bf5fc990c921df7eb4c20355a5

PR #12873, opened 2026-07-25
[BugFix][Worker] Map A5 endpoint config to physical device

Yuli-yx:

Fixes the device ID mapping issue introduced by #9690.

`setup_ascend_local_comm_res` previously used vLLM's logical `local_rank` to select a per-device endpoint file. Logical, user-visible, and physical NPU IDs can differ in containers, especially when only later host devices are mounted.

This PR performs the complete mapping inside `setup_ascend_local_comm_res`:

1. Resolve `local_rank` to the user-visible device ID.
2. Use `aclrtGetLogicDevIdByUserDevId` to obtain the Ascend logical device ID.
3. Use `aclrtGetPhyDevIdByLogicDevId` to obtain the physical NPU ID.
4. Load the endpoint file matching that physical ID.

This makes endpoint selection independent of `ASCEND_RT_VISIBLE_DEVICES`, container mount ordering, and `npu-smi` enumeration.

> [!IMPORTANT]
> This approach has an architectural compromise: it introduces direct `ctypes` calls to `libascendcl.so` and depends on low-level CANN Runtime APIs from framework code. This increases implementation and runtime dependency complexity, couples vLLM Ascend to the AscendCL ABI and supported CANN versions, and places user/logical/physical device conversion in a layer that should not ideally own it. The mapping and endpoint selection should ultimately be provided by HiXL or another lower-level runtime component.

This remains a temporary workaround for the HiXL transport dependency on 950DT server/pod deployments. A higher-level alternative is available in #12959; it avoids the direct AscendCL dependency but requires `ASCEND_RT_VISIBLE_DEVICES` to remain unset when a container mounts only a subset of host NPUs.

Yes. A5/950DT workers now load the endpoint configuration matching the physical NPU they actually use after container device remapping.

- Ruff format and lint passed.
- Python compilation passed.
- Added unit coverage for the Runtime user-to-logical-to-physical mapping and endpoint selection with a misleading visible-device environment variable.
- Local test execution requires the NPU test environment and is left to CI.

- vLLM version:
- vLLM main:

### DIFF
--- a/tests/ut/test_utils.py
+++ b/tests/ut/test_utils.py
@@ -133,6 +133,45 @@ class TestUtils(TestBase):
         with mock.patch("torch.npu.current_stream") as mock_current_stream:
             self.assertEqual(utils.current_stream(), mock_current_stream())
 
+    def test_get_physical_device_id(self):
+        acl = mock.MagicMock()
+
+        def get_logic_device_id(user_device_id, logic_device_id):
+            self.assertEqual(user_device_id, 0)
+            logic_device_id._obj.value = 1
+            return 0
+
+        def get_physical_device_id(logic_device_id, physical_device_id):
+            self.assertEqual(logic_device_id.value, 1)
+            physical_device_id._obj.value = 6
+            return 0
+
+        acl.aclrtGetLogicDevIdByUserDevId.side_effect = get_logic_device_id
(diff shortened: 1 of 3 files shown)


Yuli-yx:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

PR #12877, opened 2026-07-25
[BugFix] add warning log while sq/cq fail to bind

zouyida2052:
[BugFix] add warning log while sq/cq fail to bind

no

- vLLM version: v0.26.0

### DIFF
--- a/vllm_ascend/cpu_binding.py
+++ b/vllm_ascend/cpu_binding.py
@@ -705,7 +705,7 @@ class CpuAlloc:
             return
 
         if not os.access("/proc/irq", os.W_OK):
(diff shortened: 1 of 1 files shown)


vllm_ascend/cpu_binding.py
```diff
@@ -705,6 +705,9 @@ def bind_npu_irq(self) -> None:
             return
 
         if not os.access("/proc/irq", os.W_OK):
+            logger.warning(
```
weiguihua2: Do I need to be reminded every time?

linfeng-yuan (approved):

PR #12881, opened 2026-07-26
[Feature][KVOffload][P/D] Recompute offload supports linear hybrid model

nwpu-zxr:
Recompute offload supports linear hybrid model, such as Qwen3.5. When the MTP accept length changes, the offload Mamba cache block corresponding to the accept length needs to be placed at the start position of mamba cache groups, and this logic is implemented in this PR.

Specifically, when preemption occurs, the last `1+num_spec_tokens` mamba blocks are offloaded. When the blocks are reloaded, the index of the mamba block that needs to be reloaded is calculated based on the actual length of the last scheduling, that is, `accept_token_idx = self.num_spec_tokens - (state.num_computed_tokens - request.num_tokens + 1)`. In this case, we only need to reload the block to the 0th mamba block index of the resumed request.

No.

In local self-verification, I use Qwen3.6 35B-A3B for test. The GPU memory usage of D node is set to 0.75, and the DP1TP2 parallel configuration is used to trigger recompute offload at a high frequency. The verification result shows that the score of the GPQA dataset meets the expectation.
```
| GPQA_diamond | b1ed2c | accuracy | gen | 85.86 |
```

- vLLM version: v0.25.1

### DIFF
--- a/tests/ut/kv_offload/test_recompute_cpu_offload.py
+++ b/tests/ut/kv_offload/test_recompute_cpu_offload.py
@@ -197,6 +197,7 @@ def test_recompute_cpu_offload_scheduler_aligns_sliding_window_blocks():
 def test_recompute_cpu_offload_scheduler_d2h_keeps_sliding_window_offsets():
     scheduler = RecomputeCPUOffloadScheduler.__new__(RecomputeCPUOffloadScheduler)
     scheduler._group_is_sliding_window = [True]
+    scheduler._group_is_mamba = [False]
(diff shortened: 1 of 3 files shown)


nwpu-zxr:
@jianzs This PR supports recompute offload for linear hybrid model.

ZT-AIA (approved):

Angazenn (approved):

zzzzwwjj (approved):

nwpu-zxr:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

PR #12883, opened 2026-07-26
[Feature] Optimize MLA prolog by sharing quantization results + remove redundant stream syncsMla prolog multistream overlap

frankie-ys:
## Summary

Optimize `AscendDSAImpl._mla_prolog_multistream()` with two changes that
reduce duplicate work and increase dual-stream parallelism:

1. **Share quantized hidden_states** between wq_a and wkv when their
   quantization schemes are identical and communication status matches.
   Saves one `npu_dynamic_quant` call (full-tensor read + per-row absmax
   reduction) on W8A8, and one kernel launch on W4A8 or TP paths.

2. **Remove redundant `wait_stream(aux)`** at Part1→Part2 and Part2→Part3
   boundaries. The Q projection chain does not depend on any intermediate
   KV results; these syncs only narrow the parallel window between the
   main-stream Q path and aux-stream KV path.

## Motivation

In the 3-block multi-stream scheduling of MLA prolog (Part1/Part2/Part3),
the Q chain (wq_a → q_norm → wq_b → q_rms → rope) and KV chain (wkv →
kv_norm → rope → scatter) are dispatched to main/aux NPU streams
respectively to overlap Vector and Cube unit utilization.

Two inefficiencies were identified:

- **Duplicate quantization**: Both `cv_wq_a.quantize(hidden_states)` and
  `cv_wkv.quantize(hidden_states)` read the same input and produce the
  same output when quantization types match. For W8A8 this means two
  full-tensor reads + absmax reductions; for W4A8/TP this means two
  redundant kernel launches + event synchronization overhead.

- **Unnecessary stream syncs**: The `wait_stream(aux)` at the end of
  Part1 and Part2 block the main stream until aux completes its KV work,
  but the main stream's subsequent Q operations never read KV
  intermediates. Only the tail `wait_stream(aux)` before q_rms+rope is
  necessary (to ensure KV cache is fully written before attention reads
  it).

### `vllm-ascend/vllm_ascend/attention/dsa_v1.py`

- `_mla_prolog_multistream()` Part1: Add `share_quant` detection using
  `type(wq_a._quant_method) is type(wkv._quant_method)` (exact type
  match) plus `_has_communication` status equality. When true, call
  `quantize()` once and alias the result for both Q and KV paths.
  Falls back to original dual-quant path when quant types differ.

- Remove `main_stream.wait_stream(aux_stream)` after Part1 and Part2.
  Keep the tail `wait_stream(aux)` which is the sole hard dependency
  (KV scatter must complete before attention reads swa_kv_cache).

- vLLM version: v0.27.1


weijinqian0 (dismissed):

vllm_ascend/attention/dsa_v1.py
```diff
@@ -1387,16 +1387,28 @@ def _mla_prolog_multistream(self, hidden_states, cos, sin, swa_kv_cache, slot_ma
...
+            kv_quant, kv_pertoken_scale = q_quant, q_pertoken_scale
+        else:
+            q_quant, q_pertoken_scale = self.cv_wq_a.quantize(hidden_states)
+            e_q_quant_done = main_stream.record_event()
+            with npu_stream_switch(aux_stream, enabled=True):
+                torch.npu.current_stream().wait_event(e_q_quant_done)
+                kv_quant, kv_pertoken_scale = self.cv_wkv.quantize(hidden_states)
 
         wq_a_result = self.cv_wq_a.matmul(q_quant, q_pertoken_scale)
-        main_stream.wait_stream(aux_stream)
```
lcfenglinwan, line 1399: when this is remove, `q_norm` on line L1409  may execute with `cv_wkv.quantize` at the same time?

vllm_ascend/attention/dsa_v1.py
```diff
@@ -1419,8 +1431,6 @@ def _mla_prolog_multistream(self, hidden_states, cos, sin, swa_kv_cache, slot_ma
             q_b_quant, q_b_scale = qr, None
             qr_pertoken_scale = None
 
-        main_stream.wait_stream(aux_stream)
```
lcfenglinwan, line 1422: same as above

vllm_ascend/attention/dsa_v1.py
frankie-ys: > when this is remove, `q_norm` on line L1409 may execute with `cv_wkv.quantize` at the same time?

Yes, the modifications take into account the FIFO nature of task distribution for vectors and cubes, which avoids resource contention. Adding `wait_event` here allows for parallel computation without incurring additional overhead.

lijiahang226:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

weijinqian0 (approved):

PR #12930, opened 2026-07-27
[CI] support daily image and A5 build

du-ping1:
support daily image and A5 build
How was this patch tested?

- vLLM version: v0.25.1


.github/workflows/dockerfiles/Dockerfile.buildwheel.310p
```diff
@@ -31,7 +31,7 @@ WORKDIR /workspace
 COPY . /workspace/vllm-ascend/
 
 # Install req
-RUN python3 -m pip install -r vllm-ascend/requirements.txt --extra-index https://download.pytorch.org/whl/cpu/ --extra-index-url https://repo.huaweicloud.com/ascend/repos/pypi && \
+RUN python3 -m pip install -r vllm-ascend/requirements.txt --extra-index-url https://download.pytorch.org/whl/cpu/ --extra-index-url https://repo.huaweicloud.com/ascend/repos/pypi && \
```
Yikun, line 34: Please change this in a separate PR.

.github/workflows/dockerfiles/Dockerfile.nightly.310p
```diff
@@ -16,7 +16,9 @@
 #
 
 ARG VLLM_ASCEND_BRANCH=""
-FROM quay.io/ascend/vllm-ascend:nightly-${VLLM_ASCEND_BRANCH}-310p
+ARG BASE_IMAGE="quay.io/ascend/vllm-ascend"
+ARG TAG_SUFFIX=""
```
Yikun, line 20: Why we need this? What't the reall case? Otherwise we shouldn't change this.

.github/workflows/dockerfiles/Dockerfile.openeuler.cann.test
```diff
@@ -0,0 +1,200 @@
+# Stage 1: Install Python
```
Yikun, line 1: IMO, this should be here: https://github.com/Ascend/cann-container-image/tree/main

.github/workflows/_nightly_image_build.yaml
```diff
@@ -29,56 +29,198 @@ on:
         type: string
         default: 'main'
         description: "Branch/ref to checkout"
+      build_type:
+        required: false
+        type: string
+        default: 'release'
+        description: "Build type: 'daily' or 'release'. Daily uses swr.cn-north-4 and quay.io base image."
+      quay_daily_username:
```
Yikun, line 37: this is too specific.....

Dockerfile.310p.openEuler
```diff
@@ -101,6 +104,24 @@ RUN echo '#!/bin/bash' > /usr/local/bin/entrypoint.sh && \
...
+# All ARG definitions are in the same stage for better maintainability
+ARG BUILD_TYPE="release"
+ARG TORCH_NPU_VERSION="26.1.0"
+ARG TORCH_NPU_DATE="20260715.4"
+
+# Install daily packages only when BUILD_TYPE is set to "daily"
+RUN if [ "$BUILD_TYPE" = "daily" ]; then \
+        echo "Download, extract and install torch-npu..." && \
+        mkdir -p /tmp/torch_npu && \
+        wget -O /tmp/torch_npu/torch_npu.tar.gz "https://pytorch-package.obs.cn-north-4.myhuaweicloud.com/pta/Daily/v2.10.0-${TORCH_NPU_VERSION}/${TORCH_NPU_DATE}/pytorch_v2.10.0-${TORCH_NPU_VERSION}_py312.tar.gz" && \
```
Yikun, line 117: This is not something we mentioned in https://docs.vllm.ai/projects/ascend/en/latest/community/versioning_policy.html#software-dependency-management

Dockerfile.a3
```diff
@@ -77,4 +80,47 @@ RUN export PIP_EXTRA_INDEX_URL="https://mirrors.huaweicloud.com/ascend/repos/pyp
...
+# All ARG definitions are in the same stage for better maintainability
+ARG BUILD_TYPE="release"
+ARG MEMCACHE_VERSION="1.2.0"
+ARG MEMCACHE_DATE="20260725.3"
+ARG MEMFABRIC_VERSION="1.2.0"
+ARG MEMFABRIC_DATE="20260725.3"
+ARG TORCH_NPU_VERSION="26.1.0"
+ARG TORCH_NPU_DATE="20260715.4"
+ARG TRITON_ASCEND_VERSION="26.1.0.B110"
+ARG TRITON_ASCEND_PACKAGE_VERSION="3.2.2"
```
Yikun, line 93: Looks like internal version....we should NOT use them

Yikun (changes requested):

.github/workflows/dockerfiles/Dockerfile.openeuler.cann.test
wjunLu, line 1: Why do we need to build CANN on this repo? It should exist on the `cann-container-image` repo

du-ping1: We need a custom/intermediate CANN image built from this repo because the version we require has been validated and is not yet available in the official cann-container-image repo. That repo only contains the final release images

Dockerfile.a3
du-ping1: We need this intermediate version because it contains the exact dependency set we've already validated. The official repo doesn't provide it yet, so we build it here for now

.github/workflows/dockerfiles/Dockerfile.nightly.310p
du-ping1: We need to change this parameter to support validating intermediate versions during our iteration cycle. This allows us to keep validation builds separate from the official release artifacts. The parameter is necessary to distinguish between these two environments; otherwise, we cannot test pre-release candidates without affecting the stable release pipeline

.github/workflows/dockerfiles/Dockerfile.openeuler.cann.test
du-ping1: This intermediate image is only used by vLLM-Ascend for internal validation. It's not a formal release, so it does not belong in the cann-container-image repo, which is reserved for official CANN images. We'll keep it here to avoid polluting the official repo

Dockerfile.310p.openEuler
du-ping1: This internal iteration version is only used in our daily build/test pipeline for internal validation. It will never be included in any official release branch (e.g., main or releases/vX.Y.Z). The change here is scoped to the daily testing environment and does not affect the existing release process. We ensure that all official releases strictly follow the versioning policy.

.github/workflows/_nightly_image_build.yaml
du-ping1: The intention of this parameter is to isolate two distinct build pipelines within the same workflow:

release mode: follows the official repo's original design, using only stable dependencies as defined in the versioning policy.

daily mode: uses internal iteration versions for early validation of upcoming dependencies.

This allows the version team to proactively validate new dependencies in our daily builds, without any risk of affecting the official release process or the release pipeline. The parameter acts as a clear switch between these two isolated environments

Dockerfile.310p.openEuler
Yikun: We should push for the release of a usable torch-npu instead of using an internal version.

PR #12944, opened 2026-07-27
[BugFix][MRV2] Avoid global synchronization for ACL graph replay

yiz-liu:

This PR refines the synchronization introduced for the async-scheduling ACL graph hang described in #4233.

The hang occurs when the parameter-update stream records the current iteration event before the previous graph replay has completed. A device-wide synchronization prevents the race, but also introduces an unnecessary global barrier.

This change:

- moves ownership of the full-graph parameter update stream from `NPUModelRunner` to `ModelAclGraphManager`;
- makes the update stream wait for work already submitted to the current graph replay stream;
- preserves the required replay/update ordering without synchronizing the entire NPU device.

No.

None.

- vLLM version: v0.25.1


realliujiaxu (approved):
LGTM

PR #12424, opened 2026-07-20
[BugFix]Fix the precision issue of the npu_dequant_swiglu_quant operator when x.shape=[2,192]

ZT-AIA:
### What this PR does / why we need it?
Fix the precision issue caused by the bug in small shape tiling. Additionally, add corresponding single-operator test cases.

### Does this PR introduce _any_ user-facing change?
no
### How was this patch tested?
test

- vLLM version: v0.22.1
- vLLM main: https://github.com/vllm-project/vllm/commit/967c5c3bc38891f4465d3f4e99917ed837bb3833

PR #12965, opened 2026-07-28
[BugFix]Fix the precision issue of the npu_dequant_swiglu_quant operator when x.shape=[2,192]- #12424

ZT-AIA:

This PR backports the `dequant_swiglu_quant` UB capacity fix and test improvements from upstream PR #12424 (commit 368111f43) to the `releases/v0.24.0rc` branch for Atlas inference products.

The `dequant_swiglu_quant` AscendC operator is used in the MoE fused dequant → SwiGLU → dynamic quantization path. The original UB (Unified Buffer) capacity calculation under-reserved space for the scale buffer introduced in the `DynamicQuant()` kernel, leading to buffer overruns that caused silent data corruption on certain shapes (particularly small or wide non-aligned shapes such as `[4, 192]`, `[8, 384]`, and single-row `[1, 256]`).

Technical background and motivation

The `DynamicQuant()` kernel performs WholeReduceMax → Muls → Brcb (broadcast) → Copy → Div on the quantized output. During the broadcast step, `Brcb` was writing directly to `outLocal`, which aliased the same underlying UB memory region used by other active buffers. When the UB capacity wasn't large enough to allow these buffers to coexist at safe offsets, the broadcast write corrupted adjacent data read by the subsequent `Copy` step, producing wrong scale values and quantized output.

The size of the broadcast temp buffer (`(ceil(UbFactorDimx / BLOCK_ELEM)) * BLOCK_ELEM * BLOCK_ELEM * sizeof(float)`) was missing from both the `CountMaxDim()` tiling formulas, so UB was always under-reserved for any invocation path that reached the `DynamicQuant()` branch.

Changes in this PR

Operator fix (2 files in `csrc/moe/dequant_swiglu_quant/`):

| Area | File | Change |
|---|---|---|
| Tiling | `op_host/dequant_swiglu_quant_tiling.cpp` | Reserve `BLOCK_ELEM * BLOCK_ELEM * sizeof(float)` in `ubFactorDimy` path and `BLOCK_ELEM * sizeof(float)` in `ubFactorDimx` path |
| Kernel | `op_kernel/dequant_swiglu_quant.h` | Add dedicated `scaleBuf_` TBuf; use temporary `brcbTmp` LocalTensor in `DynamicQuant()` to isolate Brcb output from `outLocal` |

Test improvements (1 file in `tests/e2e/nightly/single_node/ops/singlecard_ops/`):

| Change | Detail |
|---|---|
| Parameterization | Switched to `@pytest.mark.parametrize` with 5 `_REPRO_CASES` covering large aligned (`4608x2048`), small misaligned (`2x192`, `4x192`, `8x384`), and single-row (`1x256`) shapes |
| Clamp guard | Skip clamp when `limit == 0.0` (treated as "no clamp") instead of hitting the small-ops fallback |
| Tolerance | int8 comparison at `atol=1` (1-ulp quantization rounding); scale comparison at `atol=1e-4, rtol=5e-3` |
| Cleanup | Removed unused `_has_effective_swiglu_limit()` helper and small-ops code path |

Files changed

- vLLM version: v0.25.1

### DIFF
--- a/csrc/moe/dequant_swiglu_quant/op_host/dequant_swiglu_quant_tiling.cpp
+++ b/csrc/moe/dequant_swiglu_quant/op_host/dequant_swiglu_quant_tiling.cpp
@@ -625,7 +625,8 @@ ge::graphStatus DequantSwigluQuantDskTiling::CountMaxDim(int64_t& ubFactorDimx) 
   int64_t quantOffsetSpace = quantMode_ == QUANT_MODE_DYNAMIC ? 0 : static_cast<int64_t>(sizeof(float));
 
   // UbFactorDimx is 1,compute maxOutDimy
-  int64_t numerator = static_cast<int64_t>(ubSize_) - UB_RESERVE - BLOCK_SIZE - db * BLOCK_SIZE - static_cast<int64_t>(sizeof(float));
+  int64_t numerator = static_cast<int64_t>(ubSize_) - UB_RESERVE - BLOCK_SIZE - db * BLOCK_SIZE - static_cast<int64_t>(sizeof(float)) -
+                       BLOCK_ELEM * BLOCK_ELEM * static_cast<int64_t>(sizeof(float));
   int64_t denominator =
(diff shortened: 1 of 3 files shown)


ZT-AIA:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

Tflowers-0129 (approved):

PR #12966, opened 2026-07-28
[Ops][Misc] Optimize 310P random sampling with inverse CDF

ShySummer:
This PR optimizes temperature-based random sampling on 310P without generating random values on the NPU.

For regular sampling, it replaces the `[batch, vocab]` CPU exponential-noise path and `Div + ArgMax` with inverse-CDF sampling: one CPU uniform random value per request, asynchronous H2D transfer, followed by NPU `Cumsum + SearchSorted`. For MTP recovered-token sampling, exponential values are still generated on CPU via `-log(U)`, with improved request-specific generator reuse, inactive-request handling, and removal of redundant copies and host-side stream synchronization.

This avoids the problematic 310P NPU random/in-place Add path while significantly reducing CPU RNG and H2D overhead.

NA

Local test

RC  without temperature (benchmark):
image

RC with temperature before this PR:
image1

RC with temperature after this PR:
image2

- vLLM version: v0.25.1

### DIFF
--- a/vllm_ascend/_310p/sample/sampler.py
+++ b/vllm_ascend/_310p/sample/sampler.py
@@ -10,7 +10,7 @@
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
+# See the License for the specific language govserning permissions and
 # limitations under the License.
 # This file is a part of the vllm-ascend project.
 #
@@ -132,7 +132,14 @@ def _random_sample_310p(
     probs: torch.Tensor,
     generators: dict[int, torch.Generator],
 ) -> torch.Tensor:
-    """310P random sampling without NPU random-number generation."""
+    """
+    310P does not support the required NPU random-generation path.
+    The previous implementation generated [batch, vocab] random values
+    on the CPU and copied them to the NPU, causing performance degradation on
+    small models and RC devices. This implementation generates only one CPU
+    random value per request and performs inverse-CDF sampling on the NPU,
+    reducing CPU computation and H2D transfer overhead.
+    """
     with npu_stream_switch(global_stream()):
         uniforms = _generate_request_uniforms_310p(
             probs.shape[0],
--- a/vllm_ascend/_310p/sample/sampler.py
+++ b/vllm_ascend/_310p/sample/sampler.py
@@ -26,71 +26,123 @@ from vllm_ascend.sample.sampler import (
 )
 from vllm_ascend.utils import global_stream, npu_stream_switch
 
-_CPU_GENERATOR_CACHE_310P: dict[int, tuple[torch.Generator, int]] = {}
+_CPU_GENERATOR_CACHE_310P: dict[int, tuple[torch.Generator, torch.Generator]] = {}
 
 
-def _get_cpu_generator_310p(i: int, generator: torch.Generator) -> torch.Generator:
-    cache_entry = _CPU_GENERATOR_CACHE_310P.get(i)
-    if cache_entry is None or cache_entry[1] != id(generator):
-        cpu_generator = torch.Generator(device="cpu")
-        try:
-            # Keep RNG stream consistent with the original generator.
-            cpu_generator.set_state(generator.get_state())
-        except Exception:
-            cpu_generator.manual_seed(generator.initial_seed())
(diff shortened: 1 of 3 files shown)


vllm_ascend/_310p/sample/sampler.py
```diff
@@ -26,73 +26,66 @@
...
-    _fill_cpu_exponential_310p(q_cpu, generators, has_draft_mask)
-    q.copy_(q_cpu.to(q.device))
-    # Ensure H2D of q is visible before rejection/recover consumes it.
-    torch.npu.current_stream().synchronize()
+        row = _generate_exponential_310p(q[i], generator)
+        if has_draft_mask is not None:
+            row = torch.where(has_draft_mask[i], row, q[i])
+        rows.append(row)
+
+    return torch.stack(rows)
```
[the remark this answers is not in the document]
ShySummer: SamplingMetadata.generators are not CPU generators in the current vLLM V1 GPU/NPU runner. They are created with torch.Generator(device=self.device), and the 310P runner uses an NPU device. Direct use of the generator advances its state automatically. CPU and NPU generator states are not interchangeable on this platform—their state sizes are 5056 and 16 respectively—so copying NPU state back to a CPU generator would raise a RuntimeError.  
Both sampling call sites already use float32 tensors. Microbenchmarks at vocabulary size 248320 showed no material improvement from replacing torch.stack with slice assignment, while slice assignment introduces aten::copy_. Therefore, the proposed generator cache, state synchronization, inplace assignment, and dtype conversion are not adopted.

vllm_ascend/_310p/sample/sampler.py
```diff
@@ -26,69 +26,59 @@
...
+    generator: torch.Generator | None = None,
+) -> torch.Tensor:
+    """Generate exponential values on NPU without using inplace add."""
+    uniform = torch.rand(
+        reference.shape,
+        dtype=reference.dtype,
+        device=reference.device,
+        generator=generator,
+    )
+    return -torch.log(uniform.clamp_min(torch.finfo(reference.dtype).tiny))
```
Tflowers-0129: Does previous ut need to change?

Alex-stack-hub (commented):
I found one blocker in the test coverage for this PR. The 310P sampler implementation removes `_CPU_GENERATOR_CACHE_310P` and `_fill_cpu_exponential_310p`, but `tests/ut/_310p/sample/test_sampler_310.py` still references those symbols in several tests (`tearDown` at line 92, cache assertions around lines 95/132/142/179/186/231/237, and `_fill_cpu_exponential_310p` at line 263 in the PR head). Once this test module is collected, those cases will fail with `AttributeError` instead of validating the new NPU RNG path. Please update or replace these tests so they assert the new `fill_exponential_310p` behavior, including the `has_draft_mask` path and generator-specific rows.

vllm_ascend/_310p/sample/sampler.py
ShySummer: UT changed

Tflowers-0129 (approved):

Tflowers-0129 (approved):

ShySummer:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

realliujiaxu (approved):

vllm_ascend/_310p/sample/sampler.py
```diff
@@ -26,71 +26,130 @@
...
+
+
+def _sample_from_cdf_310p(
+    weights: torch.Tensor,
+    uniforms: torch.Tensor,
+) -> torch.Tensor:
+    """Sample rows of non-negative weights with 310P-supported NPU ops."""
+    cdf = weights.cumsum(dim=-1, dtype=torch.float32)
+    thresholds = uniforms.unsqueeze(-1) * cdf[..., -1:]
+    return torch.searchsorted(cdf, thresholds, right=True).squeeze(-1)
```
Alex-stack-hub, line 85: `right=True` can return `weights.shape[-1]` when `thresholds` is exactly equal to the last CDF value. That can happen when `uniforms * cdf[..., -1:]` rounds up to the total in float32, even though `torch.rand` nominally returns values below 1. The returned index is then out of range for the subsequent `cand_idx.gather(...)` (or otherwise selects an invalid token). Please use `right=False` for the usual inverse-CDF convention, or clamp the result to `cdf.shape[-1] - 1` before returning.

Alex-stack-hub (commented):
Found one correctness issue in the inverse-CDF boundary handling.

PR #12982, opened 2026-07-28
[Refactor][DSA] Move TP weight switching into quantization methods

pisceskkk:

This PR moves the tensor-parallel weight split, gather, and TP/full-weight switching logic out of the SFA v1 and DSA-CP attention backends and into the quantization methods.

The attention implementations previously carried duplicated, quantization-specific knowledge about:

- which weight and auxiliary tensors must be gathered;
- which dimension each tensor is sharded on;
- which input-side parameters must be repeated for full-weight execution;
- how asynchronous gather handles are tracked and synchronized; and
- how a layer switches between its persistent TP tensors and temporary full tensors.

This made the DSA-CP o_proj path depend on specific quantization layouts and required attention code changes whenever another quantization type was added.

The new `TPWeightSwitchMixin` and declarative gather/repeat specifications centralize this behavior in the quantization layer. Each quantization method now describes its own TP tensor layout, while SFA v1 and DSA-CP use the common interface to allocate reusable full-weight buffers, launch asynchronous all-gathers, wait for completion, and switch layer parameters.

The refactor:

- removes duplicated TP/full-weight management from `sfa_v1.py` and `dsa_cp.py`;
- supports different shard dimensions and auxiliary quantization parameters;
- keeps the original TP tensors as the persistent source of truth;
- reuses pooled full-weight gather buffers;
- scopes DSA-CP o_proj switching to methods that explicitly declare support; and
- adds the unified TP weight-switch capability to all existing Ascend linear quantization types.

Yes. DSA-CP o_proj TP/full-weight switching is no longer limited to hard-coded quantization layouts. Existing Ascend quantization types can use the same execution path through their quantization-method declarations.

There is no new public API or configuration option.

Functional validation was completed on Ascend with:

- `DeepSeek-V4-Flash-w4a8`
- `DeepSeek-V4-Flash-w8a8-mtp`

Both cases used:

- tensor-parallel o_proj weight splitting and full-weight gather;
- `decode_context_parallel_size=1` with `enable_dsa_cp=true`;
- `enable_flashcomm1=true`;
- `FULL_DECODE_ONLY` graph mode; and
- MTP speculative decoding.

For each quantization type, the first 100 GSM8K samples were evaluated with deterministic decoding. Both W4A8 and W8A8 completed with 97% accuracy, 0% invalid outputs, and 0% request errors. The generated responses were also checked for empty output, obvious mojibake, and repeated text, with no issues found.

The remaining existing quantization types are covered by the unified declarative implementation but have not yet received the same end-to-end model validation.

Additional check:

- `git diff --check upstream/main...HEAD`

- vLLM version: v0.26.0


vllm_ascend/quantization/methods/w4a8.py
```diff
@@ -31,7 +31,12 @@
 from vllm_ascend.ops.fused_moe.moe_runtime_args import build_fused_experts_input
 from vllm_ascend.utils import COMPRESSED_TENSORS_METHOD, maybe_trans_nz
 
-from .base import AscendLinearScheme, AscendMoEScheme, QuantType, get_moe_num_logical_experts
+from .base import (
```
kunpengW-code: Some files have not been fully modified and need to delete useless code.

pisceskkk: have deleted.

kunpengW-code (approved):

zzzzwwjj (approved):

JC-ut0:
[Bot]: cherry-pick command failed: you do not have permission. Only the PR author or users with triage+ permission can trigger /cherry-pick.
[Bot]: cherry-pick command failed. Please check the [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/32237040800) for details.

PR #12279, opened 2026-07-17
[CI] Prebuild missing CSRC caches on CPU runners for PR tests

MrZ20:
### What this PR does / why we need it?
This PR moves csrc compilation out of NPU test jobs and builds missing csrc caches on CPU runners instead.

Previously, NPU jobs could spend significant time compiling the same custom kernels. Cache creation and cache consumption could also use different source snapshots, causing identical-looking PR runs to miss an existing cache and rebuild it.

The new workflow:

- resolves the csrc cache targets required by selected NPU runners;
- calculates the cache hash from a fixed source snapshot;
- looks up the required caches before starting NPU jobs;
- builds only missing caches on CPU runners;
- requires NPU jobs to restore the cache and installs the package with `COMPILE_CUSTOM_KERNELS=0`;
- prints each target's cache hit or miss in its corresponding GitHub Actions step.

#### Comparison chart before and after modification
image

#### NPU runner to csrc cache target mapping

```text
A2 ARM64 runners
├─ linux-aarch64-a2b3-1
├─ linux-aarch64-a2b3-2
└─ linux-aarch64-a2b3-4
        │
        └─ a2-arm64-ubuntu

A3 ARM64 runners
├─ linux-aarch64-a3-2
├─ linux-aarch64-a3-4
└─ linux-aarch64-a3-8
        │
        └─ a3-arm64-ubuntu

310P ARM64 runners
├─ linux-aarch64-310p-1
├─ linux-aarch64-310p-2
└─ linux-aarch64-310p-4
        │
        └─ 310p-arm64-ubuntu

...
```

#### PR and `/e2e` workflow

```text
PR ready label / /e2e command
        │
        ▼
Resolve the fixed source snapshot
├─ source repository and ref
├─ fixed base SHA
└─ rebase the source onto that base
        │
        ▼
Select tests
├─ selected test groups
├─ required NPU runners
└─ deduplicated csrc cache target IDs
        │
        ▼
Ensure csrc caches
├─ calculate the tracked-source manifest hash
├─ lookup a2-arm64-ubuntu
├─ lookup a3-arm64-ubuntu
├─ lookup 310p-arm64-ubuntu
├─ print the hit/miss result for each lookup step
└─ collect missing targets
        │
        ├─ no missing targets
        │        └─ start NPU tests
        │
        └─ missing targets
                 └─ build csrc caches on CPU runners
                          └─ save caches
                                   └─ start NPU tests
```

#### Cache production and consumption

```text
Fixed source snapshot
[...]

PR #12988, opened 2026-07-28
[CI] Fix Main2Main fresh-mode upstream reference

MrZ20:
Fix a Main2Main regression introduced by:
- https://github.com/vllm-project/vllm-ascend/pull/12279

PR moved `upstream` setup into the incremental-only `FIXED_BASE_SHA` condition. When a baseline rebase conflicts, the workflow falls back to fresh mode with an empty `FIXED_BASE_SHA`, so `upstream/main` is not created: https://github.com/vllm-project/vllm-ascend/actions/runs/30284552529/job/90040429009.

`main2main_flow` then fails during initialization:

```text
fatal: Not a valid object name upstream/main
```

This change:

- derives the fixed upstream base from `base_sha` or `source_ref`;
- configures `upstream` in both fresh and incremental modes;
- pins `upstream/main` to the resolved SHA instead of fetching a moving main;
- preserves the existing incremental rebase behavior.

- vLLM version: v0.25.1

### DIFF
--- a/.github/workflows/schedule_main2main.yaml
+++ b/.github/workflows/schedule_main2main.yaml
@@ -331,17 +331,22 @@ jobs:
         env:
           FIXED_BASE_REF: ${{ needs.resolve-source.outputs.base_ref }}
           FIXED_BASE_SHA: ${{ needs.resolve-source.outputs.base_sha }}
+          SOURCE_REF: ${{ needs.resolve-source.outputs.source_ref }}
         run: |
           set -eu
           BRANCH="main2main_auto_$(date +%Y-%m-%d_%H-%M)"
+          UPSTREAM_BASE_SHA="${FIXED_BASE_SHA:-$SOURCE_REF}"
+          git remote add upstream "https://github.com/${UPSTREAM_REPO}.git" || \
+            git remote set-url upstream "https://github.com/${UPSTREAM_REPO}.git"
           git checkout -B "$BRANCH"
           if [ -n "$FIXED_BASE_SHA" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/schedule_main2main.yaml
menogrey, line 333: https://github.com/vllm-project/vllm-ascend/pull/12279/changes delete some comments, please add it back, now I don't know about the `fresh-mode` in the PR title.

.github/workflows/schedule_main2main.yaml
```diff
@@ -331,17 +331,22 @@ jobs:
         env:
           FIXED_BASE_REF: ${{ needs.resolve-source.outputs.base_ref }}
           FIXED_BASE_SHA: ${{ needs.resolve-source.outputs.base_sha }}
```
menogrey: What is the differences between FIXED_BASE_REF and FIXED_BASE_SHA, the naming don't show the real meaning.

.github/workflows/schedule_main2main.yaml
MrZ20: done.

.github/workflows/schedule_main2main.yaml
MrZ20: Good point. `FIXED_BASE_REF` was the ref used to fetch the upstream commit, while `FIXED_BASE_SHA` was the immutable commit actually used as the rebase base.

I renamed them to `UPSTREAM_FETCH_REF` and `UPSTREAM_REBASE_SHA` and added a comment to make the distinction explicit. In fresh mode, both are empty because the workflow checks out the resolved upstream SHA directly.

wangxiyuan (approved):

PR #13008, opened 2026-07-28
[Test][E2E] Fail fast on DP runner worker errors

pisceskkk:
## What this PR does

- wait concurrently on every DP worker result Pipe and process sentinel
- fail immediately when any rank reports an error or exits unexpectedly
- stop all DP worker process groups together through graceful shutdown, SIGTERM, then SIGKILL
- explicitly shut down each DP worker's vLLM EngineCore
- add focused unit coverage for error propagation, sentinel handling, rank ordering, process-group cleanup, and EngineCore shutdown

## Why

In [the failing CI job](https://github.com/vllm-project/vllm-ascend/actions/runs/30324533946/job/90168231149), pytest printed a DP worker failure, but a peer worker or one of its descendants stayed alive and retained the pytest/tee pipe. As a result, the nonzero test result did not reach the workflow promptly and the job eventually ended at the runner timeout. Sequential per-rank Pipe polling also delayed observing failures from later ranks.

This change treats all Pipes and process sentinels as one wait set. Any worker failure enters one cleanup path that stops every worker process group, while each worker also explicitly shuts down its EngineCore in `finally`.

## Validation

- ruff check and format
- codespell and typos
- relevant repository pre-commit checks
- Python compileall
- targeted mypy with CI-compatible options

Ascend-dependent UT was not executed locally because this workspace has no configured Ascend remote endpoint; PR CI will provide the runtime validation.

- vLLM version: v0.26.0

### DIFF
--- a/.github/workflows/scripts/test_config.yaml
+++ b/.github/workflows/scripts/test_config.yaml
@@ -53,6 +53,14 @@
   tests:
     - tests/ut
 
+# E2E test harness
+- name: e2e_conftest
+  optional: true
+  source_file_dependencies:
+    - tests/e2e/conftest.py
+  tests:
+    - tests/ut/test_dp_vllm_runner.py
+
 # 310P
 - name: 310p
   optional: false
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -33,8 +33,9 @@ import sys
 import threading
 import time
 import traceback
+from multiprocessing.connection import Connection, wait
 from pathlib import Path
-from typing import Any, TypeVar
+from typing import Any, TypeVar, cast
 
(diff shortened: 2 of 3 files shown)


zhangxinyuehfad:
/e2e tests/e2e/pull_request/four_card/context_parallel/test_accuracy.py::test_models_dcp_full_feature_accuracy[dsv3_2_sfa_dcp_replicated_indexer]
[Bot]: e2e command triggered. See [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/30349832536) for details.
[Bot]: e2e command failed.

pisceskkk:
Closed because CI have timeout now.

PR #13081, opened 2026-07-29
[Misc] Suggest max_cudagraph_capture_size via computational modeling

zhiyu-wa:
Adds a heuristic function `_suggest_max_capture_size()` that computes and logs a recommended `max_cudagraph_capture_size` based on simple runtime modeling.

We conducted experiments to model the recommended value for `max_cudagraph_capture_size`. Using `num_tokens` as the independent variable, we measured inference latency under both graph mode and eager mode as the dependent variable. The results show that graph mode latency scales roughly linearly with `num_tokens`.

Based on this observation, we adopted the following approach:

1.Measure latency at a few representative `num_tokens` points for both cudagraph and eager modes.

2.Fit a linear model to the aclgraph latency data.

3.Treat eager mode latency as approximately constant(which holds before the intersection point), and compute the intersection point between the constant eager latency and the fitted line.

4.Use the `num_tokens` value at the intersection as the recommended `max_cudagraph_capture_size`.
**Note:** that eager latency variance may cause the computed value to fluctuate across runs; use it as a reference baseline.

Experimental data supporting this approach is shown below (red: eager, blue: aclgraph):

**Qwen3-0.6B**
Qwen3-0 6B

**Qwen3-32B**
Qwen3-32B

**Llama-3.2-3B-Instruct**
Llama-3 2-3B-Instruct

**Qwen3-30B-A3B**
Qwen3-30B-A3B

Yes. Users will see a new log message in the format. This is informative only and does not change any default behavior.

- vLLM version: v0.26.0

### DIFF
--- a/tests/e2e/pull_request/two_card/test_flashcomm_distributed.py
+++ b/tests/e2e/pull_request/two_card/test_flashcomm_distributed.py
@@ -22,6 +22,7 @@ Run `pytest tests/e2e/pull_request/two_card/test_flashcomm_distributed.py`.
 """
 
 import os
+import time
 from unittest.mock import patch
 
 import pytest
@@ -70,6 +71,7 @@ def test_qwen3_dense_fc1_tp2(model):
     ) as vllm_model:
         vllm_model.generate_greedy(example_prompts, max_tokens)
 
+    time.sleep()
 
 @pytest.mark.parametrize("model", QWEN_DENSE_MODELS)
 @patch.dict(os.environ, {"VLLM_ASCEND_ENABLE_FLASHCOMM1": "1"})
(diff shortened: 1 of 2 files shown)


zhiyu-wa:
/e2e tests/e2e/pull_request/two_card/test_flashcomm_distributed.py::test_qwen3_dense_fc1_tp2
[Bot]: e2e command triggered. See [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/30983313972) for details.
[Bot]: e2e command failed.

zhiyu-wa:
/e2e tests/e2e/pull_request/two_card/test_flashcomm_distributed.py::test_qwen3_dense_fc1_tp2
[Bot]: e2e command triggered. See [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/30986666622) for details.
[Bot]: e2e command failed.

PR #13153, opened 2026-07-30
perf(worker): overlap attn metadata build by removing CPU syncs in as…

hbamboo:
In async scheduling + MTP + hybrid (GDN) decode, two CPU blocking points in _prepare_inputs drained the CPU/GPU pipeline and exposed ~2ms of attention metadata build per decode step with the GPU idle (measured on Qwen3.5-122B-A10B-MTP, TP4, FULL_DECODE_ONLY):

num_accepted_tokens_event.synchronize() blocked ~10–14ms per step until the previous step's sampling finished. After the sync, only ~3ms of drafter tail remained to cover ~5ms of CPU work, so the last two (GDN) _build_attn_group_metadata calls ran with the GPU idle (~1.9ms/step exposed, 56.5ms over 31 profiled steps).

_correct_optimistic_seq_lens_cpu (with its valid-sampled-count event wait) ran in the middle of _prepare_inputs, further extending the CPU blocking window.

Fix:

Move num_accepted_tokens fixup entirely onto the GPU: gather by prev_positions, set new requests to 1, and use the event only as a stream-level dependency (wait_event) instead of a CPU sync. Falls back to the original synced path for non-async scheduling or mamba_cache_mode != "none" (where CPU values are still consumed by preprocess_mamba in "align" mode).

Defer _correct_optimistic_seq_lens_cpu (and its valid-sampled-count event wait) from _prepare_inputs into _build_attention_metadata. KV cache groups whose builders don't consume CPU seq_lens (GDN) are built first, so their construction overlaps with the previous step's GPU execution; the correction runs only just before the first group (FIA/MLA/compressed) that needs exact CPU seq_lens.

 no

Benchmark on Atlas A3 with Qwen3.5-122B-A10B-MTP (TP4, FULL_DECODE_ONLY) under async spec decode. Profiling confirms the two CPU syncs no longer block the pipeline, and decode throughput improves accordingly.

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -439,6 +439,11 @@ class NPUModelRunner(GPUModelRunner):
         self._needs_seq_lens_cpu_sync = self.use_compress or issubclass(
             self.attn_backend, (AscendAttentionBackend, AscendMLABackend)
         )
+        # Set by _prepare_inputs when optimistic_seq_lens_cpu still needs the
+        # async spec-decode correction; consumed (and cleared) by
+        # _build_attention_metadata just before the first builder that
+        # consumes CPU seq_lens.
+        self._seq_lens_cpu_correction_pending = False
 
         # kv role
         self.is_kv_producer = False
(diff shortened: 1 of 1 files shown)


Tflowers-0129:
此问题还存在吗？是否已经解决了呢

hbamboo:
> 此问题还存在吗？是否已经解决了呢

Qwen3.5-122B-A10B 在0.23 release版本仍然存在

PR #13155, opened 2026-07-30
[Feature][MRV2] Support DSA in ModelRunner V2

li1how:

This PR enables DeepSeek-V4 DSA execution with ModelRunner V2 while preserving the existing ModelRunner V1 behavior.

- Integrate DeepSeek-V4 with the unified Ascend forward context so DSA and MoE paths can access the current input tokens, sequence-parallel padding state, and communication metadata under both model runners.
- Implement the complete ModelRunner V2 KV-cache lifecycle for DeepSeek-V4 cache-only modules, including cache-spec discovery, physical allocation, page-strided data/scale views, A5 full-cache views, shared backing storage, and runner/layer binding. Existing GQA and MLA cache paths remain supported.
- Preserve DSA physical block sizing in the Ascend MLA cache spec, avoiding a second compression of pages whose scheduler-to-cache mapping is already handled by the cache manager and DSA metadata.
- Build DSA attention metadata with separate actual and padded input-token counts, derive block sizing from each cache group, and share common, prefill, and decode SAS metadata across ratio-specific groups within one model execution.
- Reuse the upstream slot-mapping implementation while retaining Ascend's int32 output buffers, removing the local fixed-size block-table lookup window that limited long-context requests.

Unit coverage validates DeepSeek-V4 forward-context propagation, A2/A5 cache discovery and initialization, shared physical storage, cache binding, DSA metadata sharing, and eager/full-graph token-count handling.

This PR focuses on the standalone DSA ModelRunner V2 adaptation. DSA prefill context-parallel execution support is handled separately.

No new CLI option or configuration is introduced. DeepSeek-V4 DSA models can use ModelRunner V2 through the existing runner switch.

https://wiki.huawei.com/domains/87904/wiki/153722/WIKI2026072311977224

- vLLM version: v0.26.0


vllm_ascend/worker/v2/block_table.py
```diff
@@ -61,106 +59,3 @@ def __init__(
...
-        else:
-            # Context parallelism is used.
-            is_local = block_offsets // CP_INTERLEAVE % CP_SIZE == cp_rank
-            rounds = block_offsets // (CP_INTERLEAVE * CP_SIZE)
-            remainder = block_offsets % CP_INTERLEAVE
-            local_offsets = rounds * CP_INTERLEAVE + remainder
-            slot_ids = block_numbers * block_size + local_offsets
-            slot_ids = tl.where(is_local, slot_ids, PAD_ID)
-
-        tl.store(slot_mapping_ptr + offset, slot_ids, mask=offset < end_idx)
```
weijinqian0, line 166: plz check this.

weijinqian0 (approved):

vllm_ascend/worker/v2/attn_utils.py
```diff
@@ -251,116 +381,122 @@ def _allocate_kv_cache(
...
-        assert k_dim > 0 and v_dim > 0
-        kv_head_dim_list = [k_dim, v_dim]
-        if enable_fa_quant(vllm_config):
-            k_tensor_split_factor, v_tensor_split_factor = vllm_config.quant_config.get_kv_quant_split_factor(
-                example_layer_name, kv_head_dim_list
-            )
+        example_spec = layer_kv_cache_spec[example_layer_name]
+        assert isinstance(example_spec, AttentionSpec)
+        k_dim, v_dim = _get_attention_kv_cache_dims(example_layer_name, example_spec)
+        if enable_fa_quant(vllm_config, example_layer_name):
```
pichangping: There is no need to pass `example_layer_name` to the `enable_fa_quant` function here.

li1how: modified

vllm_ascend/worker/v2/block_table.py
li1how: confirmed with the person in charge, prioritizing functionality.

PR #13182, opened 2026-07-30
[BugFix][MRV2]Fix synchronous H2D in UvaBufferWrapper fallback path

xiayingqing:
When running MRV2 on NPU, the H2D copy in `UvaBufferWrapper.uva` fallback path is not truly asynchronous, causing CPU bubbles.

**Root cause**: The original code uses advanced indexing `self._cpu[dirty_rows]`, which creates an unpinned copy. This causes `.to(device="npu", non_blocking=True)` to silently degrade to a synchronous copy in PyTorch.

**Fix**: Since `copy_to_uva` always writes a contiguous prefix via `dst[:n] = x`, `dirty_rows` is always `[0..n-1]`. Replace the indexed assignment with a contiguous slice `copy_`:

```python
# Before
self._uva[dirty_rows] = self._cpu[dirty_rows].to(device="npu", non_blocking=True)

# After
n = len(self._modified_indices)
self._uva[:n].copy_(self._cpu[:n], non_blocking=True)
```
This resolves three issues:

Slice indexing preserves the pinned property of the CPU source, so non_blocking takes effect.
Avoids the intermediate NPU tensor allocation from .to().
Eliminates the implicit stream sync between .to() and __setitem__ on NPU.

No.

NPU profiling: synchronous H2D copy and NPU bubble observed before the fix; async DMA and NPU bubble eliminated after the fix.

- vLLM version: v0.26.0

### DIFF
--- a/vllm_ascend/patch/worker/patch_v2/patch_uva.py
+++ b/vllm_ascend/patch/worker/patch_v2/patch_uva.py
@@ -150,11 +150,15 @@ class UvaBufferWrapper:
     def uva(self):
         """Get the device data of the buffer."""
         if not is_uva_available() and self._modified_indices:
-            # Sort for better memory access locality
-            dirty_rows = sorted(self._modified_indices)
-            # can't use copy_ method, because copy_ for index tensor
-            #  will malloc new memory.
-            self._uva[dirty_rows] = self._cpu[dirty_rows].to(device="npu", non_blocking=True)
+            # copy_to_uva always writes a contiguous prefix dst[:n] = x,
+            # so dirty rows are always [0..n-1].  Use a contiguous slice
+            # copy_ (instead of indexed assignment) to:
+            #  1. keep the CPU source pinned (slice is a view, not a copy),
+            #     so non_blocking is not silently degraded.
+            #  2. avoid the intermediate tensor from .to() and the implicit
+            #     stream sync between .to() and __setitem__ on NPU.
+            n = len(self._modified_indices)
+            self._uva[:n].copy_(self._cpu[:n], non_blocking=True)
             self._modified_indices.clear()
         return self._uva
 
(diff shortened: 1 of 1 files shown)


vllm_ascend/patch/worker/patch_v2/patch_uva.py
```diff
@@ -150,11 +150,15 @@ def np(self):
...
-            self._uva[dirty_rows] = self._cpu[dirty_rows].to(device="npu", non_blocking=True)
+            # copy_to_uva always writes a contiguous prefix dst[:n] = x,
+            # so dirty rows are always [0..n-1].  Use a contiguous slice
+            # copy_ (instead of indexed assignment) to:
+            #  1. keep the CPU source pinned (slice is a view, not a copy),
+            #     so non_blocking is not silently degraded.
+            #  2. avoid the intermediate tensor from .to() and the implicit
+            #     stream sync between .to() and __setitem__ on NPU.
+            n = len(self._modified_indices)
+            self._uva[:n].copy_(self._cpu[:n], non_blocking=True)
```
freyfwt: 这里最好加个保护，如果_modified_indices不是连续值，fallback回原来的写法。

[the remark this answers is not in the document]
xiayingqing: The guard is already in place in the current revision — the code checks dirty_rows[0] == 0 and dirty_rows[-1] == n_dirty - 1 before taking the fast path. The else branch also explicitly re-pins the CPU source with pin_memory() to prevent the same non_blocking degradation that the original indexed assignment had.

xiayingqing: 已加，当前版本的 guard 就是检查 dirty_rows[0] == 0 and dirty_rows[-1] == n_dirty - 1，不连续时走 else 分支做 indexed copy。

linfeng-yuan (approved):

freyfwt (approved):

PR #13232, opened 2026-07-31
[v0.23.0][Bugfix]Modify the VLLM_EXECUTE_MODEL_TIMEOUT_SECONDS timeout duration

luoxiaolin712:
The timeout interval of the HCCL operator is 1836s. Timeout in seconds for execute_model RPC calls in multiprocessing must be greater than 1836s. The timeout interval of the DP engine  is shorter than that of the collective communication (1836s). As a result, the communication times out and is suspended without logs.

NA

- vLLM version: v0.23.0


yiz-liu (commented):
Please change it to `envs_vllm`.

luoxiaolin712:
Already change it to envs_vllm.

yiz-liu (approved):

PR #12950, opened 2026-07-27
[Feature][v0.26.0rc] Support Kimi K3 on Ascend

maoxx241:
## Overview

Add Kimi K3 model support to vllm-ascend, including the model architecture, KDA attention, quantized MoE execution, multimodal processing, tool-call parsing, KV-cache integration, and Ascend custom operators.

## Key Components

1. Model Architecture and Hybrid Attention (`vllm_ascend/models/`, `vllm_ascend/core/`, `vllm_ascend/attention/`)
   - Adds the Kimi K3 text network, model registration, hybrid KV-cache specifications, MLA no-RoPE handling, and output gating
   - Integrates MLA/KDA layers, MoE routing, quantized expert execution, and model-specific rotary embeddings

2. KDA/GDN Attention and Recurrent State (`vllm_ascend/ops/`, `csrc/attention/`, `csrc/moe/chunk_gated_delta_rule_fwd_h/`)
   - Adds KDA orchestration and recurrent-state management for prefill, decode, variable-length sequences, and graph capture
   - Adds chunk KDA, recurrent KDA, gate cumulative-sum, layout-swap, and chunk gated-delta-rule operators across AscendC and Triton paths
   - Pre-packs causal-convolution weights during the standard post-load lifecycle and preserves the packed parameter across full and kernel-format checkpoint reloads

3. Quantized MoE, SiTU, and Parallel Execution (`vllm_ascend/ops/fused_moe/`, `vllm_ascend/quantization/`, `csrc/moe/`)
   - Adds SiTU MX quantization and dequantization kernels and integrates the Kimi K3 quantized expert execution paths
   - Extends ModelSlim and compressed-tensors mappings and supports the greater-than-512-expert AllGather fallback
   - Decouples shared-expert DP from FlashComm sequence parallelism, routes FC1 full-rank gates correctly, and validates multistream shared-expert input widths

4. Multimodal and Transformers Utilities (`vllm_ascend/transformers_utils/`, `vllm_ascend/models/`, `vllm_ascend/patch/platform/`)
   - Registers the upstream-aligned Kimi K3 configuration and processor contracts
   - Handles image placeholders, NaViT sizing, multimodal token expansion, preprocessing, media loading, and vision parallelism

5. MLA FIA Split and Graph Execution (`vllm_ascend/attention/`)
   - Splits long decode block tables into fixed FIA shards and merges local outputs with device-side LSE
   - Masks graph-padding rows and extends MLA, context-parallel, and ACLGraph coverage for the split path

[...]

PR #13277, opened 2026-07-31
[Feature][v0.26.0rc] Support Kimi K3 MLA DSpark speculative decoding on Ascend

maoxx241:

Adds the Kimi-K3 dense **MLA** draft model for **DSpark** speculative decoding
on Ascend, together with the correctness and asynchronous-scheduling fixes
required by MLA DSpark, Prefix Cache, ACLGraph, P/D, and QuaRot serving. The
target branch is `releases/v0.26.0rc`, on top of the Kimi-K3 support from
#12950.

This is deliberately an **MLA-only** PR:

- Kimi-K3 **GQA DSpark is out of scope**. The patch does not change the
  ordinary attention backend for GQA drafting.
- The Kimi-K3 **four-plane cache layout is out of scope**. Its implementation
  and history are absent from the rewritten branch.
- `vllm_ascend/attention/attention_v1.py` and its focused A2 unit test are
  unchanged from the release baseline.

The low-level MLA preprocess RoPE switch is already supplied by release PR
#13467, so this PR does not duplicate its C++ operator changes.

### Commit layout

| Commit | Scope |
|---|---|
| `069f3ee52` | Backport the net stateful GDN graph-replay fixes from #12027 and #12255, including the 310P path, without restoring their reverted ModelRunner approach. |
| `756a328fe` | Add the Ascend Triton fused norm-gate path used by Kimi-K3 sigmoid gating and focused A2 coverage. |
| `64a223a10` | Add and register `K3DSparkModel` / `K3DSparkForCausalLM`, MLA draft KV mapping, YaRN metadata, non-causal draft attention, and release-baseline proposer compatibility. |
| `9e91bd2bc` | Keep shared DSpark boundary weights in the correct coordinate space for QuaRot targets and fail closed on incompatible shapes. |
| `e9181131d` | Release the native KDA workspace event on the empty-Vnew return path. |
| `1f84ddedb` | Reset stale speculative state inputs before full-graph replay. |
| `20f7bb3c3` | Reconstruct exact MLA draft FIA KV lengths on the host and remove the early device `Tensor.tolist()` synchronization. |
| `46d2c0fd3` | Align the Step3.5 proposer override with the release interface. |
| `90697a039` | Trim Prefix Cache hits consistently across all full-attention groups. |
| `ceccb99ce` | Preserve producer-side Mamba Prefix Cache hits for P/D scheduling. |
| `3cd7136db` | Give idle dummy graph runs zero sequence length so GDN conv/KDA state updates are skipped. |
| `82ca7b4c5` | Defer rejected-token synchronization to the first MLA draft attention instead of blocking earlier in the proposer. |
| `be4dfdb69` | Align the five K3 MLA `context_proj` auxiliary hidden-state blocks with the target QuaRot basis. |
| `4fd628914` | Fix pre-commit mypy by narrowing optional rejected-token copy resources and the padded request count without changing runtime flow. |
| `214a56771` | Preserve the narrowed `ast.BoolOp` type in the balance-schedule AST test while keeping `generic_visit` behavior unchanged. |
| `840fe0dda` | Update the A2 Eagle proposer signature contract for the new draft-token host metadata parameter. |

Yes. Kimi-K3 can use an MLA DSpark draft model on Ascend in colocated and P/D
serving, including Prefix Cache, ACLGraph `FULL_DECODE_ONLY`, asynchronous
scheduling, and QuaRot targets. This PR intentionally does not add Kimi-K3 GQA
DSpark or the former four-plane cache layout.

### Focused unit and source checks

- `tests/ut/models/test_kimi_k3_dspark.py` and
  `tests/ut/spec_decode/test_dspark_proposer.py`: **33 passed** on Ascend A3.
- Ruff format/check, codespell, typos, clang-format, Markdown/workflow/shell
  lint, Python compile checks, and `git diff --check`: passed.
- The local validation environment lacks the gitleaks bootstrap dependency, so
  the gitleaks hook is left to GitHub CI.
- The ordinary GQA attention implementation and focused test match
  `releases/v0.26.0rc`; no four-plane layout symbols remain in the patch.

### Fast MLA functional validation at rewritten head

Validated commit: `be4dfdb69a9292b3aea38aa0dfcd51644cee7711`.

The smoke target used a reduced 16-expert Kimi-K3 W4A8 checkpoint and an MLA
DSpark draft checkpoint, with TP16, Prefix Cache enabled, asynchronous
scheduling enabled by default, and ACLGraph `FULL_DECODE_ONLY` at capture size
16.

#### Colocated

- 6/6 smoke requests completed: identity, GSM8K, two shared-prefix requests,
  and two concurrent requests.
- 38 draft steps, 266 draft tokens, 106 accepted tokens: **39.85%** per-token
  acceptance.
- Prefix Cache: 5,893 queried tokens, 1,920 hits.

#### Disaggregated prefill/decode

- 6/6 smoke requests completed through one producer, one consumer, and the
  standard load-balancing proxy.
- Consumer logs recorded real Mooncake KV transfers across TP ranks.
- Consumer: 37 draft steps, 259 draft tokens, 108 accepted tokens: **41.70%**
  per-token acceptance.
- Consumer external Prefix Cache: 3,973 queried tokens, 3,967 hits.

Both modes completed without a request traceback, transfer error, literal
`None`, replacement-character output, or binary/mojibake corruption. The
reduced 16-expert checkpoint is a feature-path smoke target, not a semantic
accuracy gate; full-checkpoint answer quality is therefore not inferred from
this run. All validation services and the proxy were stopped afterward.

Made with [Cursor](https://cursor.com)

- vLLM version: v0.26.0

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -354,6 +354,7 @@ class AscendMLAMetadata:
     attn_mask: torch.Tensor = None
     # chunked prefill by default if no attn_states passed
     attn_state: AscendAttentionState = AscendAttentionState.ChunkedPrefill
+    causal: bool = True
 
     decode: AscendMLADecodeMetadata | None = None
     prefill: AscendMLAPrefillMetadata | None = None
@@ -628,6 +629,7 @@ class AscendMLAMetadataBuilder(MLACommonMetadataBuilder[AscendMLAMetadata]):
             num_prefills=self.num_prefills,
             attn_mask=self.attn_mask_builder.get_splitfuse_attn_mask(),
             attn_state=common_attn_metadata.attn_state,
+            causal=common_attn_metadata.causal,
             prefill=prefill_metadata,
             decode=decode_metadata,
             query_start_loc=query_start_loc,
@@ -1692,8 +1694,18 @@ class AscendMLAImpl(MLAAttentionImpl):
                 q_nope = F.pad(q_nope, (0, 0, 0, self.head_padding), "constant", 0)
             # Output shape: [num_heads, num_tokens, dim]
             attn_output_shape = (self.num_heads_padded, num_tokens, self.kv_lora_rank)
-            sparse_mode = 3
-            attn_mask = attn_metadata.decode.attn_mask  # type:ignore
+            if not attn_metadata.causal:
+                # The DSpark MLA draft block is non-causal (bidirectional):
+                # every query token attends to the trailing context window plus
+                # all other query tokens in the draft block. On Ascend FIA this
+                # is sparse_mode=0 with NO atten_mask (a mask under sparse_mode=0
+                # is applied as defaultMask and would wrongly hide the upper
+                # triangle -- see vllm_ascend/attention/attention_mask.py).
+                sparse_mode = 0
+                attn_mask = None
+            else:
+                sparse_mode = 3
+                attn_mask = decode_meta.attn_mask
             actual_seq_lengths = decode_meta.actual_seq_lengths_q
             if self.fa_quant_layer:
                 dequant_scale_q_nope = dequant_scale_q_nope.view(num_tokens, self.num_heads)
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -598,9 +598,15 @@ class AscendMLAMetadataBuilder(MLACommonMetadataBuilder[AscendMLAMetadata]):
 
         query_seq_lens_cpu = query_start_loc_cpu[1:] - query_start_loc_cpu[:-1]
         self.query_lens = query_seq_lens_cpu[:num_reqs]
-        # Prefer _seq_lens_cpu (always available, updated during draft
-        # iterations) over seq_lens_cpu (None in async spec decode mode).
-        if common_attn_metadata._seq_lens_cpu is not None:
+        # Parallel drafting needs the exact post-rejection length. In async
+        # mode the regular CPU mirrors are optimistic; the proposer supplies
+        # an exact host copy reconstructed from accepted-token counts. Other
+        # paths retain the existing CPU-mirror contract.
+        if common_attn_metadata.parallel_draft_seq_lens_cpu is not None:
+            self.seq_lens = common_attn_metadata.parallel_draft_seq_lens_cpu[:num_reqs]
+        # Prefer _seq_lens_cpu (always available, updated during non-parallel
+        # draft iterations) over seq_lens_cpu (None in async spec mode).
+        elif common_attn_metadata._seq_lens_cpu is not None:
             self.seq_lens = common_attn_metadata._seq_lens_cpu[:num_reqs]
         elif common_attn_metadata.seq_lens_cpu is not None:
             self.seq_lens = common_attn_metadata.seq_lens_cpu[:num_reqs]
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -360,6 +360,17 @@ class AscendMLAMetadata:
     prefill: AscendMLAPrefillMetadata | None = None
     reshape_cache_event: torch.npu.Event = None
 
+    # Deferred parallel-draft reject finalize (dspark async spec decode). The
+    # builder copies these from AscendCommonAttentionMetadata when the proposer
+    # published an optimistic parallel_draft_seq_lens_cpu; _forward_decode
+    # synchronizes the event and subtracts the per-request reject count from
+    # ``decode.seq_lens_list`` exactly once (``reject_finalized`` guards re-entry
+    # across layers sharing this metadata object). None when no finalize is pending.
+    pending_reject_cpu: torch.Tensor = None
+    pending_reject_event: torch.npu.Event = None
+    pending_reject_num_reqs: int = 0
+    reject_finalized: bool = False
+
     def __post_init__(self):
         pass
         # supported_head_sizes = AscendMLABackend.get_supported_head_sizes()
@@ -598,10 +609,10 @@ class AscendMLAMetadataBuilder(MLACommonMetadataBuilder[AscendMLAMetadata]):
 
         query_seq_lens_cpu = query_start_loc_cpu[1:] - query_start_loc_cpu[:-1]
         self.query_lens = query_seq_lens_cpu[:num_reqs]
-        # Parallel drafting needs the exact post-rejection length. In async
-        # mode the regular CPU mirrors are optimistic; the proposer supplies
-        # an exact host copy reconstructed from accepted-token counts. Other
-        # paths retain the existing CPU-mirror contract.
+        # Async parallel drafting publishes an optimistic host length here.
+        # The first FIA forward waits for the overlapped reject-count copy and
+        # applies the exact post-rejection correction before launching FIA.
+        # Other paths retain the existing CPU-mirror contract.
         if common_attn_metadata.parallel_draft_seq_lens_cpu is not None:
             self.seq_lens = common_attn_metadata.parallel_draft_seq_lens_cpu[:num_reqs]
         # Prefer _seq_lens_cpu (always available, updated during non-parallel
@@ -642,6 +653,10 @@ class AscendMLAMetadataBuilder(MLACommonMetadataBuilder[AscendMLAMetadata]):
             block_tables=self.block_table,
             seq_lens=self.seq_lens,
             seq_lens_cpu=self.seq_lens,
+            pending_reject_cpu=common_attn_metadata.parallel_draft_num_reject_cpu,
+            pending_reject_event=common_attn_metadata.parallel_draft_num_reject_event,
+            pending_reject_num_reqs=common_attn_metadata.parallel_draft_num_reject_num_reqs,
+            reject_finalized=False,
         )
 
     def build_chunked_metadata(
@@ -1649,6 +1664,20 @@ class AscendMLAImpl(MLAAttentionImpl):
     ) -> torch.Tensor:
         decode_meta = attn_metadata.decode
         assert decode_meta is not None
+        # Deferred parallel-draft reject finalize (dspark async spec decode):
+        # synchronize the side-stream D2H of num_rejected_tokens and subtract
+        # the per-request reject count from decode_meta.seq_lens_list -- the
+        # host list FIA consumes as actual_seq_kvlen. Runs once per metadata
+        # object; layers sharing this metadata skip via reject_finalized. The
+        # D2H was launched right after prepare_inputs_padded, so by the first
+        # attention layer the copy is typically already complete (sync no-op).
+        if attn_metadata.pending_reject_event is not None and not attn_metadata.reject_finalized:
+            attn_metadata.pending_reject_event.synchronize()
+            reject = attn_metadata.pending_reject_cpu[: attn_metadata.pending_reject_num_reqs].tolist()
+            seq_lens_list = decode_meta.seq_lens_list
+            for i in range(attn_metadata.pending_reject_num_reqs):
+                seq_lens_list[i] -= reject[i]
+            attn_metadata.reject_finalized = True
         # TODO: The CANN package is expected to support num_heads that are not
         # powers of 2 in 2026 Q2. Once supported, all padding operations under
         # `if self.head_padding > 0` in this function can be removed.
--- a/.github/workflows/scripts/test_config.yaml
+++ b/.github/workflows/scripts/test_config.yaml
@@ -833,6 +833,7 @@ estimated_times:
   tests/ut/eplb/core/a2/test_eplb_utils.py: 30
   tests/ut/kv_offload/a2/test_remote_decode_lifecycle.py: 60
   tests/ut/kv_offload/a2/test_remote_prefill_lifecycle.py: 60
+  tests/ut/ops/a2/test_fused_norm_gate.py: 30
   tests/ut/ops/a2/test_gdn_chunk_meta.py: 30
   tests/ut/ops/a2/test_gdn_layerwise_kv.py: 30
   tests/ut/ops/a2/test_token_dispatcher.py: 30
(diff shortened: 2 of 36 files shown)


zhenwenqi2024 (approved):

vllm_ascend/attention/mla_v1.py
```diff
@@ -1692,8 +1696,21 @@ def _forward_decode(
                 q_nope = F.pad(q_nope, (0, 0, 0, self.head_padding), "constant", 0)
             # Output shape: [num_heads, num_tokens, dim]
             attn_output_shape = (self.num_heads_padded, num_tokens, self.kv_lora_rank)
-            sparse_mode = 3
-            attn_mask = attn_metadata.decode.attn_mask  # type:ignore
+            if _EXTRA_CTX.is_draft_model and self.speculative_config.use_dspark():
```
drslark: Please use `causal` like `attention_v1` to distinguish.

drslark (changes requested):

H100-H200-B200:
Great PR! I was wondering if it would be possible to access the draft model `Kimi-K3-DSpark-GQA-20260801` somewhere. As far as I can tell, there doesn't seem to be an open-weights release of the Ascend-compatible kimi-k3-dspark version yet. Even a pre-release version would be immensely helpful if available.

maoxx241:
@H100-H200-B200 Thanks for asking! The Kimi-K3 DSpark draft-model weights are available here: https://huggingface.co/RadixArk/Kimi-K3-DSpark

ZT-AIA (approved):

drslark (approved):

maoxx241:
## Historical four-node A3 performance evidence (pre-rewrite)

These measurements were collected at pre-rewrite head `2c16a64d5` with a
QuaRot target and MLA DSpark draft. They are retained as historical performance
evidence and are **not** current-head correctness or acceptance validation.

Test contract: DP4 x TP16 x EP64, global concurrency 4, one request per DP,
draft K=7, greedy sampling, asynchronous scheduling enabled, and ACLGraph
`FULL_DECODE_ONLY` with capture size 16. Every measured request returned the
requested 1,024 output tokens with no request failure.

| Client/workload | Prefix Cache | Cache result | Mean TTFT | Mean TPOT | Mean ITL | Acceptance length | Acceptance rate |
|:---|:---:|:---|---:|---:|---:|---:|---:|
| AISBench exact 8K/1K c4 | off | disabled | 3485.8 ms | 18.0 ms | **108.3 ms** | 6.06 | 72.27% |
| AISBench exact 8K/1K c4 | on | cached 8K prompt | 847.5 ms | 18.0 ms | **109.8 ms** | 6.13 | 73.22% |
| vLLM Bench exact 8K/1K c4 | on | forced no-hit: 0/8192 per DP | 4935.5 ms | 18.69 ms | **113.98 ms** | 6.12 | 73.09% |
| vLLM Bench exact 128K/1K c4 | on | 129408/131072 hit per DP | 2610.4 ms | 80.63 ms | **117.62 ms** | 1.46 | 6.55% |

For the 128K case, 129,792 of 131,072 input tokens were shared by construction
(99.023%). The measured reusable prefix was 129,408 tokens (98.73%) because the
384-token cache block had to be completely reusable. The 80.63 ms TPOT tracks
the low 1.46-token acceptance length; mean ITL is the more comparable runtime
metric for this case.

Prefix Cache changed cached 8K TTFT from 3.49 s to 0.85 s while AISBench decode
ITL remained essentially flat (108.3 ms to 109.8 ms).

Deployment-specific filesystem paths, network addresses, interface names, and
service endpoints are intentionally omitted.

maoxx241:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

maoxx241:
## Final commit-by-commit problem/fix map

This summary follows the **final merged history** at `840fe0dda`. It supersedes
the pre-rewrite SHA descriptions: the merged PR is MLA-only and does not contain
the former Kimi K3 GQA DSpark/four-plane cache implementation.

1. [`069f3ee52`](https://github.com/vllm-project/vllm-ascend/commit/069f3ee52884cf7e6020de9631e246c1798ba035) — **stateful GDN graph replay**
   - **Problem:** FULL graph replay could reuse capture-time sequence lengths,
     query starts, recurrent/conv state indices, and accepted-token metadata.
     Speculative buffers could leak into a later non-spec replay; one-token or
     speculative-width prefills were classified by shape without considering
     recurrent-state ownership; padded rows could retain valid-looking state IDs.
   - **Fix:** Refresh stable buffers from the current replay, clear stale
     speculative inputs, preserve per-chunk token counts, include state ownership
     in path selection, and write null/PAD sentinels for padded rows. The same
     contract is applied to the 310P builder.

2. [`756a328fe`](https://github.com/vllm-project/vllm-ascend/commit/756a328fe662729e6107d256c7d9dbba148b80c0) — **Kimi K3 fused norm gate**
   - **Problem:** The Ascend gated-norm path was effectively tied to
     SiLU/swish and did not carry the activation choice through
     `AscendRMSNormGated`, so K3's RMSNorm + sigmoid-gate equation had no correct
     fused implementation.
   - **Fix:** Add an Ascend Triton fused LayerNorm/RMSNorm gate supporting
     sigmoid and swish, gate-before/gate-after normalization, bounded UB tiling,
     and fallback to the legacy path for unsupported forms; forward the
     activation from the layer and add A2 operator coverage.

3. [`64a223a10`](https://github.com/vllm-project/vllm-ascend/commit/64a223a10f00637144e49e516103fca0fb917755) — **Kimi K3 dense MLA DSpark support**
   - **Problem:** vLLM Ascend had no registered model/config/execution contract
     for the released dense K3 MLA draft checkpoint.
   - **Fix:** Add `K3DSparkConfig`, `K3DSparkModel`, and
     `K3DSparkForCausalLM`; register the architecture; project five target aux
     hidden-state blocks through `context_proj`; precompute latent context KV per
     draft layer; use the draft's YaRN RoPE metadata; and execute the seven-token
     draft block with non-causal MLA attention. It also adds per-layer context
     slot mapping and focused model/proposer/MLA tests.

4. [`9e91bd2bc`](https://github.com/vllm-project/vllm-ascend/commit/9e91bd2bcef9b059ca7dbba73251aae5c483721d) — **QuaRot target/draft boundary alignment**
   - **Problem:** A QuaRot target exposes rotated hidden states and boundary
     weights, while the DSpark draft is trained in the original basis. Directly
     aliasing the target embedding/`lm_head`, or consuming rotated aux hidden
     states, is mathematically wrong even when shapes match and can collapse
     draft acceptance.
   - **Fix:** Load the target's global rotation, align each draft projection
     input block, keep/materialize draft-owned embedding and `lm_head` modules,
     copy target boundary weights back to the draft basis, leave target tensors
     unchanged, and fail closed on incompatible shapes.

5. [`e9181131d`](https://github.com/vllm-project/vllm-ascend/commit/e9181131dad37b935a4d189bc51196140dc08055) — **native KDA multi-wave hang**
   - **Problem:** An empty Vnew AIV subblock returned without consuming and
     restoring its E0 workspace `HardEvent`. When the ping-pong stream was reused
     on a later scheduling wave, the next task waited forever. A direct A3 probe
     gave a stable boundary: 40 tasks completed, while 41 tasks timed out.
   - **Fix:** Determine the ping-pong slot before the empty-row branch, wait for
     the current workspace state, restore `V_MTE2`, and only then publish
     `vec1Done` and return. The fixed operator completed the 40/41/42/80/96-task
     cases without adding a global synchronization.

6. [`1f84ddedb`](https://github.com/vllm-project/vllm-ascend/commit/1f84ddedb92631a6649f7fd706a2b51e6faa68ba) — **stale state in no-spec FULL replay**
   - **Problem:** The stale-spec reset ran only in the pure non-spec decode
     branch. A FULL replay containing no speculative decode but containing
     dummy/prefill work could therefore reuse the preceding request's captured
     recurrent inputs and mutate a completed Mamba checkpoint.
   - **Fix:** Reset the captured spec task for every FULL no-spec replay before
     the prefill/decode branch is selected. The same commit also makes the K3
     draft config an explicit no-default-init contract and removes reliance on
     a serialized `block_size` that older checkpoints may omit.

7. [`20f7bb3c3`](https://github.com/vllm-project/vllm-ascend/commit/20f7bb3c30447b6966f640ea8ee65b31ffb94a37) — **MLA draft sequence-length D2H sync**
   - **Problem:** Async scheduling leaves host sequence-length mirrors
     optimistic, but MLA FIA requires exact post-rejection KV lengths as a host
     list. Falling back to device `Tensor.tolist()` synchronizes the NPU default
     stream on the draft critical path.
   - **Fix:** Reconstruct the exact parallel-draft lengths from the
     accepted-token counts already copied to pinned host memory, add the current
     draft query width, and publish a dedicated host length field for MLA. This
     removes the device `tolist()` sync while preserving rejected-tail
     correctness; commit `82ca7b4c5` later moves the remaining side-stream wait
     farther down the pipeline.

8. [`46d2c0fd3`](https://github.com/vllm-project/vllm-ascend/commit/46d2c0fd33e74c1c520086c9186a6b96b604465e) — **Step3.5 proposer interface mismatch**
   - **Problem:** The common proposer call gained `num_draft_tokens_cpu`, but
     `AscendStep3p5MTPProposer._propose` did not accept it. This produced an
     incompatible-override mypy error and could raise an unexpected-keyword
     error when the common runner invoked that drafter.
   - **Fix:** Add the same optional parameter to the Step3.5 override. Step3.5
     does not consume it, so runtime behavior is otherwise unchanged.

9. [`90697a039`](https://github.com/vllm-project/vllm-ascend/commit/90697a03956d72a5e0a95e3c26c3edee3e9c13ab) — **hybrid Prefix Cache group reconciliation**
   - **Problem:** After reconciling one reusable prefix across full-attention
     and Mamba groups, finalization truncated only `attention_groups[0]`. Later
     full-attention groups could retain blocks beyond the recurrent checkpoint.
   - **Fix:** Trim every `FullAttentionSpec` group to the reconciled hit length
     using its effective block size and record the same final hit on every
     group.

10. [`ceccb99ce`](https://github.com/vllm-project/vllm-ascend/commit/ceccb99cec0d44b9915f2ba51677139d6d82dbfa) — **producer-side Mamba Prefix Cache correctness**
    - **Problem:** The connector's consumer-only partial-group lookup was also
      used by KV producers. A producer could accept an attention-only local hit,
      skip prefill tokens, and continue without the corresponding Mamba state;
      unlike a consumer, it does not receive that missing state from a peer.
    - **Fix:** Restrict partial-group lookup to consumers. Producers use the
      normal all-group/Mamba-aware lookup, while the consumer optimization is
      preserved.

11. [`3cd7136db`](https://github.com/vllm-project/vllm-ascend/commit/3cd7136db63cdef17a2b0f5bff4459b29be8c3d3) — **idle DP dummy state corruption**
    - **Problem:** An idle DP rank still runs a dummy model step for collective
      symmetry. Describing `_dummy_run(1)` to GDN as a real one-token decode
      could write conv/KDA state through persistent padded graph buffers even
      though the rank owned no request.
    - **Fix:** Propagate `skip_gdn_state_update`, keep the model/collective tensor
      shape, but build GDN metadata with zero actual tokens/query length and
      null/PAD recurrent state indices. The runner, worker, and 310P paths all
      follow the same no-op contract.

12. [`82ca7b4c5`](https://github.com/vllm-project/vllm-ascend/commit/82ca7b4c550d638a167a290d6d793292b583b110) — **deferred reject synchronization**
    - **Problem:** `20f7bb3c3` avoided a default-stream D2H, but the proposer
      still waited for the host copy before building draft metadata, leaving a
      visible target-to-draft issuance bubble.
    - **Fix:** Copy `num_rejected_tokens_gpu` to pinned host memory on a side
      stream, publish optimistic lengths during metadata construction, and wait
      and subtract the per-request rejects exactly once at the first MLA draft
      FIA entry. This overlaps the copy with metadata build and early draft
      forward while preserving exact KV bounds.
    - **Additional startup fix in the same commit:** When materializing a
      draft-owned QuaRot boundary module, reuse its non-copyable communication
      group in the `deepcopy` memo while still cloning the parameters. This
      avoids attempting to pickle a distributed `ProcessGroup`.

13. [`be4dfdb69`](https://github.com/vllm-project/vllm-ascend/commit/be4dfdb69a9292b3aea38aa0dfcd51644cee7711) — **K3 MLA `context_proj` QuaRot gap**
    - **Problem:** The generic QuaRot path aligned a draft module named `fc`,
      but K3 MLA consumes five target aux hidden-state blocks through
      `context_proj`. That projection was skipped, so the draft received the
      wrong basis. In the matched 8K/256 reproduction, acceptance was 0.583%
      although the target token output was unchanged.
    - **Fix:** Detect `context_proj` and right-multiply each target-width input
      block by the global rotation. The matched run recovered acceptance to
      78.93% with the same target output token hash.

14. [`4fd628914`](https://github.com/vllm-project/vllm-ascend/commit/4fd628914ef8cd6de89bb38b63b419d4698a3a3f) — **Python 3.10 mypy failures in graph metadata**
    - **Problem:** Mypy could not prove that the optional rejected-token event,
      copy stream, host buffer, and padded request count were initialized on
      the guarded paths, producing four pre-commit failures.
    - **Fix:** Bind the optional resources to narrowed locals, assert their
      grouped initialization, and assert `num_reqs_padded` before the nested GDN
      metadata builder uses it. Runtime control flow and event order are
      unchanged.

15. [`214a56771`](https://github.com/vllm-project/vllm-ascend/commit/214a567714df7b658752255a9a23c1a39d98482c) — **AST-test type widening**
    - **Problem:** Assigning `node = self.generic_visit(node)` widened a known
      `ast.BoolOp` to the general `AST | list[AST] | None` return type, so mypy
      rejected the subsequent `node.values` access.
    - **Fix:** Let `generic_visit` mutate the BoolOp in place without assigning
      its general return value. The AST normalization performed by the test is
      unchanged.

16. [`840fe0dda`](https://github.com/vllm-project/vllm-ascend/commit/840fe0dda40a56c0b532116f032b0ad5b2f6b8c5) — **stale proposer signature test contract**
    - **Problem:** The A2 Eagle test reflectively asserted the old `_propose`
      parameter list and failed after `num_draft_tokens_cpu` was added.
    - **Fix:** Add the new host-metadata parameter to the expected signature;
      production code is unchanged.

The entries above intentionally separate feature enablement, runtime
correctness, operator liveness, performance synchronization, and CI-only
repairs so that validation evidence from one category is not treated as proof
for another.

Issue #13220, opened 2026-07-31
[RFC]: Refactor Ascend fused_moe to align with vLLM MoE architecture


weijinqian0:
### Motivation.

vllm: https://github.com/vllm-project/vllm/pull/32344

The Ascend fused MoE implementation has accumulated multiple responsibilities in the Ascend-specific runner and quant method path. Today `vllm_ascend.ops.fused_moe` handles routing, expert selection, quant method replacement, expert placement/EPLB, communication method selection, MLP execution, shared expert overlap, LoRA metadata, and runtime payload conversion across several tightly-coupled modules.

Meanwhile, upstream vLLM has evolved the MoE entry point into a clearer component model:

- `FusedMoE(...)` is a factory that builds `FusedMoEConfig`, `FusedMoEParallelConfig`, router, `RoutedExperts`, and `MoERunner`.
- Router implementations such as `FusedTopKRouter` own expert selection.
- `RoutedExperts` owns expert weights, weight loading, expert map metadata, and quant method creation.
- `MoERunner` orchestrates forward execution and shared expert behavior.
- `QuantizationConfig.get_quant_method(layer, prefix)` remains the quantization extension point.

Aligning Ascend fused MoE with this structure should reduce maintenance cost, make upstream rebases easier, isolate NPU-specific optimizations, and make future features such as new routers, dynamic EPLB, LoRA, and quantized MLP kernels easier to review independently.

<img width="2040" height="1118" alt="Image" src="https://github.com/user-attachments/assets/3a71e9a4-1997-49ef-a674-2cef1e26b19c" />

### Proposed Change.

This RFC proposes a staged refactor of `vllm_ascend/ops/fused_moe`.

#### Stage 1: Align Ascend fused MoE interfaces with upstream vLLM

1. Keep the Ascend patch of `FusedMoE(...)` as a thin factory wrapper. @weijinqian0 
   - Preserve upstream arguments such as `intermediate_pad`, `router`, `swiglu_alpha`, `swiglu_beta`, `activation_situ_beta`, `activation_situ_linear_beta`, `ckpt_names`, `apply_routed_scale_to_output`, `zero_expert_type`, `hash_indices_table`, `routed_experts_cls`, and `routed_experts_args`.
   - Limit Ascend-specific argument handling to NPU-only extensions such as `tid2eid` and Ascend EPLB capacity propagation.

[...]

PR #13282, opened 2026-07-31
[Refactor] Refactor the quant_apply_mlp method

zhangc10:

- vLLM version: v0.26.0


weijinqian0:
https://github.com/vllm-project/vllm-ascend/issues/13220, There is a RFC about Moe refactor.  May you can add a comment  about it.

PR #13289, opened 2026-07-31
Revert "[cherry-pick][v0.23.0][Performance] remove D2H sync in QLIMetadata builder for DSA_CP"

pisceskkk:
Reverts vllm-project/vllm-ascend#12537

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/attention/context_parallel/dsa_cp.py
+++ b/vllm_ascend/attention/context_parallel/dsa_cp.py
@@ -605,7 +605,7 @@ class AscendDSACPMetadataBuilder(AttentionMetadataBuilder[AscendDSAMetadata]):
         )
         local_seq_lens_q_cpu = local_query_start_loc_cpu[1 : num_reqs + 1] - local_query_start_loc_cpu[:num_reqs]
         max_local_query_len = max(1, int(local_seq_lens_q_cpu.max().item()))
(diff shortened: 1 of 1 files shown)


fourierr:
Will it cause garbled decode?  if we don't do the revert? We pulled the 0.23.0rc1 image on July 20th and have noticed intermittent garbled decode problems. Is this related to it? @pisceskkk

pisceskkk:
> Will it cause garbled decode? if we don't do the revert? We pulled the 0.23.0rc1 image on July 20th and have noticed intermittent garbled decode problems. Is this related to it? @pisceskkk

So far, we haven't observed this change causing garbled decode, and the accuracy on the AIME dataset is also normal.
However, the change reverted in this PR can lead to prefill node hangs or performance degradation in certain corner cases.

fourierr:
> > Will it cause garbled decode? if we don't do the revert? We pulled the 0.23.0rc1 image on July 20th and have noticed intermittent garbled decode problems. Is this related to it? @pisceskkk
> 
> So far, we haven't observed this change causing garbled decode, and the accuracy on the AIME dataset is also normal. However, the change reverted in this PR can lead to prefill node hangs or performance degradation in certain corner cases.

thanks for your reply

PR #13334, opened 2026-08-02
[BugFix][Ops] Rename custom SparseFlashAttention to avoid CANN collision

lxb007981:

CANN already provides a built-in sparse-attention operator. The vLLM Ascend custom implementation duplicated its names at two different integration layers:

- Both operators used `SparseFlashAttention` as their GE OpType, causing an operator-registration collision.
- Both implementations exposed ACLNN C symbols such as `aclnnSparseFlashAttention` and `aclnnSparseFlashAttentionGetWorkspaceSize`, causing a symbol-name collision.

When the built-in and custom operator packages are loaded together, these duplicate registration and symbol names can prevent the custom operator from being registered or resolved correctly.

Rename the custom GE OpType to `SparseFlashAttentionVllm` and its ACLNN interfaces to `aclnnSparseFlashAttentionVllm` so the built-in and custom implementations can coexist.

The changes include:

- Rename the OpDef, tiling, template-key, and kernel registrations.
- Rename the public and inner ACLNN interfaces.
- Rename the operator directory and update its build configuration.
- Update the `kv_quant_sparse_flash_attention` dependency paths.
- Update the documentation and ACLNN example.
- Keep the Torch API name `npu_sparse_flash_attention` unchanged for compatibility.

No. This only changes the internal CANN operator and ACLNN symbol names. The Torch API and operator behavior remain unchanged.

This patch was tested by running GLM5.2 inference.

- vLLM version: v0.26.0


lxb007981:
No it's not a documentation change. It is about a real bug that needs to be fixed.

PR #13340, opened 2026-08-02
[Performance]: Fuse multi-group slot mapping into a single 2D-grid Triton kernel with pre-cached params

frankie-ys:
### 1. Fused 2D-grid slot-mapping kernel (`ops/triton/slot_mapping.py`)
- New `compute_slot_mapping_fused_kernel` uses a 2D grid
  `(num_reqs + 1, num_groups)`.
- Each program handles exactly one `(req_idx, group_idx)` pair, so all KV
  cache groups execute concurrently within a single launch.
- Per-group parameters are accessed through pre-built pointer arrays
  (`block_table.data_ptr()`, `slot_mapping.data_ptr()`, `block_size`, ...),
  using the pointer-indirection pattern already used by `batch_memcpy`.
- Computation inside the normal path is byte-for-byte identical to the
  original single-group `_compute_slot_mapping_kernel`.

### 2. Zero-copy param pre-caching (`worker/block_table.py`)
- `MultiGroupBlockTable._build_fused_params()` builds the 6 per-group
  parameter tensors **once** during `__init__`.
- These values (`data_ptr`, `stride`, `block_size`, ...) are static after
  construction, so they are reused on every step — eliminating the ~5ms
  per-call tensor construction entirely.

### 3. Strategy dispatch & fallbacks
- `compute_slot_mapping` now prefers: fused kernel → multi-stream parallel
  launch → original sequential loop.
- Falls back to sequential when DCP is active or Triton is unavailable.
- Multi-stream path uses a reusable NPU stream pool.

- vLLM version: v0.27.1


linfeng-yuan:
PTAL this optimization @ZT-AIA

frankie-ys:
before fuse multi-group cc3a09a748bec64b42324a95e84bb40c
after fuse multi-group slot mapping 
2b93374ee391c9de52d4633d5574b794
  the compute_slot_mapping time from 2.13ms to 0.55ms

vllm_ascend/ops/triton/slot_mapping.py
```diff
@@ -0,0 +1,254 @@
...
+       ``torch.tensor(…)`` construction overhead (~5 ms / call).
+
+    Args:
+        block_tables: List of non-mamba ``BlockTable`` instances.
+        num_reqs: Number of requests in the batch.
+        query_start_loc: GPU tensor ``[num_reqs + 1]`` (int32).
+        positions: GPU tensor ``[num_tokens]`` (int64).
+    """
+    num_groups = len(block_tables)
+    if num_groups <= 1:
```
ckeyever: redundant conditional check, `if num_groups == 1:` is enough

vllm_ascend/ops/triton/slot_mapping.py
```diff
@@ -0,0 +1,254 @@
+# SPDX-License-Identifier: Apache-2.0
```
ckeyever, line 1: 1. Consider the following potential optimization points(could be applied in the next PR if necessary)
- if `num_reqs_plus_one * num_groups` exceed the real AICore number, it will introduce additional scheduling overhead
- reduce the call of tl.load by merging block_table_ptrs, block_table_strides, block_sizes, slot_mapping_ptrs, kv_cache_block_sizes into one or two tensors
2. add test for the new kernel
3. It's better to move launch_slot_mapping_fused from slot_mapping.py into test files
4. add doc for the new kernel, refer to https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/ops/triton/compute_slot_mapping.md

frankie-ys: I have made the modifications as suggested, but it seems there's an underlying issue with CI causing select-test to fail.

frankie-ys:
@ckeyever @ZT-AIA Hello, I have made the revisions based on your suggestions. Please review them again.

vllm_ascend/worker/block_table.py
```diff
@@ -422,13 +597,108 @@ def compute_slot_mapping(
...
+            On Ascend 910B, ``torch.npu.Stream()`` serves as its own
+            context manager (no ``StreamContext`` wrapper needed).
+            We use a dynamic-import fallback for portability.
+        """
+        if not self._non_mamba_tables:
+            return
+
+        num_groups = len(self._non_mamba_tables)
+        streams = self._get_stream_pool(num_groups)
+
```
weijinqian0: Plz use one stream.Too many streams cannot bring more revenue.

PR #13349, opened 2026-08-03
[v0.23.0][Misc] Add AscendStore PP save diagnostics

Pz1116:

Adds targeted AscendStore diagnostics for a GLM-5.1 `memcache + use_layerwise=false + PP4` issue where PP ranks 2 and 3 pass zero sizes to `batch_put_from_layers`.

The diagnostics:
- log global/local/TP/PP/PCP/DCP rank mapping once per worker
- log a compact KV cache registration summary once per worker
- consolidate non-layerwise puts into one log per request and cache group, including size counts, min/max size, invalid-key count, and at most three key-level samples

This is a diagnostic-only draft and does not change KV save behavior.

No functional change. It adds temporary warning-level diagnostics.

- Relevant pre-commit hooks passed, including ruff, formatting, codespell, and repository policy checks.
- Python syntax compilation passed for the changed modules.
- The targeted AscendStore unit test could not run in the local macOS environment because the workspace vLLM checkout is a dev/incompatible revision and Ascend runtime extensions are unavailable. NPU validation is required.
- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
@@ -371,51 +371,12 @@ class ChunkedTokenDatabase:
         if block_id is None:
             block_idx = start // group_block_size
             if block_idx >= len(block_ids):
-                group_metadata = self.metadata[kv_cache_group_id]
-                logger.warning(
-                    "KVSIZE_BLOCK_ID_MISSING pp=%d group=%d cache_role=%s "
-                    "start=%d end=%d group_block_size=%d block_idx=%d block_ids=%s",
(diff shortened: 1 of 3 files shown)


LCAIZJ:
1. Does this issue only occur with Memcache, and not with Mooncake?
2. Should this diagnostic be kept long-term, or is it only for temporary troubleshooting?

Pz1116:
> 1. Does this issue only occur with Memcache, and not with Mooncake?
> 2. Should this diagnostic be kept long-term, or is it only for temporary troubleshooting?

Currently this is just only for temporary troubleshooting, but for long-term, some logs of similar manner should be added in my opinon. When the loggers are finalized, I will make this PR ready for review (and maybe just for later branches not 23.0)

PR #13352, opened 2026-08-03
[BugFix] Restore MoE and DSA control buffers after level-2 wake

Ulrica111:
## Summary

Restore runtime control tensors through model buffers so that level-2 sleep/wake saves and restores them via `model.named_buffers()`.

- Register the DeepSeek-V4 DSA Hadamard matrices as shared, non-persistent buffers on their owning Attention modules.
- Register the ALLTOALL dispatcher's `expert_ids_per_ep_rank` tensor as a non-persistent buffer on `routed_experts`.

## Motivation

These tensors are created or owned by non-`nn.Module` runtime objects, so they are not discovered by the level-2 buffer backup path. After wake-up, the original owners could therefore retain invalid device storage.

This change keeps the same Tensor objects used by the runtime components, but makes an appropriate model module own each one as a buffer. `persistent=False` keeps this runtime-derived state out of model checkpoints.

## Scope

- Affects DeepSeek-V4 DSA attention Hadamard matrices and ALLTOALL dispatcher expert-ID mapping only.
- Does not change model weights, checkpoint contents, or ALLTOALL dispatch semantics.

## Testing

- `python -m compileall` for the four modified Python files.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/attention/dsa_v1.py
+++ b/vllm_ascend/attention/dsa_v1.py
@@ -22,6 +22,7 @@ from vllm_ascend.attention.attention_v1 import AscendAttentionState
 from vllm_ascend.attention.utils import (
     AscendCommonAttentionMetadata,
     enable_pcp,
+    get_or_register_attention_buffer,
     maybe_save_kv_layer_to_connector,
     notify_kv_cache_written,
     split_decodes_and_prefills,
@@ -329,7 +330,6 @@ def build_dspark_swa_indices(
 
 
 class AscendDSAMetadataBuilder(AttentionMetadataBuilder[AscendDSAMetadata]):
-    hadamard = None
     """
     NOTE: Please read the comment at the top of the file before trying to
     understand this class
@@ -391,7 +391,8 @@ class AscendDSAMetadataBuilder(AttentionMetadataBuilder[AscendDSAMetadata]):
         self.seq_lens: torch.Tensor = None
 
         self.compressor_ratio = getattr(kv_cache_spec, "compress_ratio", 0)
-        self._init_hadamard()
+        self.hadamard = None
+        self._init_hadamard(layer_names)
         self.start_pos_prefill: torch.Tensor = torch.zeros(
             scheduler_config.max_num_seqs, dtype=torch.int32, device=self.device
         )
@@ -409,10 +410,7 @@ class AscendDSAMetadataBuilder(AttentionMetadataBuilder[AscendDSAMetadata]):
         # [block_nums, block_size, head_num, head_dim]
         self.slot_mapping = torch.zeros(self.slot_mapping_shape, dtype=torch.int32, device=self.device)
 
-    def _init_hadamard(self) -> None:
-        if AscendDSAMetadataBuilder.hadamard is not None:
-            return
-
+    def _init_hadamard(self, layer_names: list[str]) -> None:
         hf_config = self.model_config.hf_config
         if hf_config.model_type != "deepseek_v4":
             return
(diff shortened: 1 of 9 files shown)


yjyang62 (approved):

Ulrica111:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E
- PR Create

vllm_ascend/ops/fused_moe/token_dispatcher.py
```diff
@@ -465,11 +466,21 @@ def __init__(self, **kwargs):
...
-            )
+            allocation_context = nullcontext()
+            if get_current_vllm_config().model_config.enable_sleep_mode:
+                # This is ALLTOALL dispatcher state, not model weight data. Keep
+                # it mapped across level-2 sleep/wake instead of restoring an
+                # empty discarded allocation.
+                from vllm_ascend.device_allocator.camem import CaMemAllocator
+
+                allocator = CaMemAllocator.get_instance()
+                allocation_context = allocator.use_allocation_tag(CaMemAllocator.sleep_persistent_tag)
```
Ronald1995: extract a method for these operations

vllm_ascend/attention/dsa_v1.py
```diff
@@ -394,20 +392,14 @@ def _init_hadamard(self) -> None:
...
-            allocator = CaMemAllocator.get_instance()
-            with allocator.use_allocation_tag(CaMemAllocator.sleep_persistent_tag):
-                AscendDSAMetadataBuilder.hadamard = torch.tensor(
-                    hadamard(dim_padded, dtype=float), dtype=torch.float, device=self.device
-                ).to(torch.bfloat16)
-        else:
-            AscendDSAMetadataBuilder.hadamard = torch.tensor(
-                hadamard(dim_padded, dtype=float), dtype=torch.float, device=self.device
-            ).to(torch.bfloat16)
+        self.hadamard = get_or_register_attention_buffer(
```
Ronald1995, line 425: when we set sleep_level=1, it won't clone sleep_buffer in `worker.wake_up`, hadamard  still be polluted.

vllm_ascend/worker/worker.py
```diff
@@ -220,10 +220,10 @@ def uninstall_static_kernel(self):
 
     def sleep(self, level: int = 1) -> None:
         free_bytes_before_sleep = torch.npu.mem_get_info()[0]
-        # Save the buffers before level 2 sleep
-        if level == 2:
-            model = self.model_runner.model
-            self._sleep_saved_buffers = {name: buffer.cpu().clone() for name, buffer in model.named_buffers()}
+        # Both level-1 and level-2 sleep need this CPU snapshot to restore
+        # registered control buffers after wake-up.
```
Ronald1995: it's better to use whitelist for level 1:
```python
# add comment to explain
allowed_names = (A, B)
if level == 1:
    self._sleep_saved_buffers = {name: buffer.cpu().clone() for name, buffer in model.named_buffers() if name in allowed_names }
```

Ulrica111:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

Ronald1995 (approved):
LGTM, use `name_buffer` which is the same method with vllm to fix dirty data problem in sleep mode.

Ulrica111:
/rerun
[Bot]: rerun completed.

**Failed:**

Ulrica111:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

Ulrica111:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

ZT-AIA (approved):

PR #13364, opened 2026-08-03
Modify README title with test demo note

MrYoung92:
Updated README to include a test demo note.it is just a demo

- vLLM version: 


MrYoung92:
it is just a test demo

PR #13365, opened 2026-08-03
[Feature][LoRA] 1.Support W8A8 dynamic multi-LoRA inference for Moe models 2.Generral moe quant frame work for Lora inference 3. quanted shared experted lora support

SkychenLee:
## What this PR does

This PR adds Ascend support for W8A8_DYNAMIC MoE inference with multiple LoRA adapters, targeting DeepSeek V4 Flash-style models containing both routed experts and shared experts.

It supports mixing base requests and requests using different LoRA adapters in the same batch, while preserving the existing optimized W8A8 path for base-only batches.

## Why this is needed

The existing Ascend W8A8 MoE path may quantize activations during the AllGather prepare or token dispatch stage and then execute fused GMM/SwiGLU kernels.

This is incompatible with MoE LoRA for two reasons:

1. LoRA A projections require the original BF16/FP16 activations. Once activations are converted to INT8 during prepare or dispatch, the original input needed by LoRA is lost.
2. The fused W8A8 MoE kernels do not provide an injection point for selecting LoRA weights by both expert ID and adapter slot.

For multi-LoRA serving, every routed row must select its LoRA weights using:

```text
(expert_id, lora_slot)
```

Different tokens in the same batch may use different adapters, while base tokens must produce exactly zero LoRA delta.

Merging LoRA weights into the quantized base weights cannot support per-token adapter selection and would also introduce significant memory overhead.

### 1. Quantized MoE LoRA backend abstraction

A new quantized MoE LoRA backend registry is introduced in:

```text
vllm_ascend/lora/quant_moe.py
```

The common MoE MLP and token dispatch code do not contain W8A8-specific execution logic. Instead, each quantization scheme can register:

- Its activation quantization policy.
- Input validation rules.
- Its quantized base GMM implementation.
- The floating-point boundaries where LoRA is injected.

The first registered backend supports `QuantType.W8A8`.

Unsupported quantization schemes fail explicitly instead of silently ignoring MoE LoRA.

### 2. Locally unfused W8A8 execution for active LoRA batches

When the current batch contains at least one LoRA token, the routed-expert MLP is expanded into:

```text
BF16/FP16 routed input
    -> dynamic INT8 quantization
    -> INT8 GMM1
    -> BF16/FP16 w13 LoRA injection
    -> SwiGLU / clipped SwiGLU
    -> preserve BF16/FP16 activation
    -> dynamic INT8 quantization
    -> INT8 GMM2
    -> BF16/FP16 w2 LoRA injection
```

The mathematical behavior is:

```text
gate_up = INT8_GMM1(input) + LoRA_w13(input)
hidden  = activation(gate_up)
output  = INT8_GMM2(quantize(hidden)) + LoRA_w2(hidden)
```

This preserves the correct LoRA inputs while keeping both base expert matmuls quantized.

For base-only batches, the original dispatch quantization and fused W8A8 path remain unchanged.

### 3. LoRA-aware activation quantization policy

On the current main branch, activations may be quantized in two places:

- Sequence-parallel AllGather prepare.
- Token dispatch.

The quantized MoE LoRA activation policy is therefore applied at both stages.

For active W8A8 LoRA batches, prepare and dispatch keep the routed activations in BF16/FP16. The W8A8 LoRA backend performs dynamic quantization immediately before each base GMM.

For base-only batches, prepare and dispatch continue using the existing quantized path.

The same policy is used by AllGather, AlltoAll, and MC2 dispatchers so unsupported communication and quantization combinations fail before losing the floating-point LoRA input.

### 8. Explicit support boundaries

The first quantized implementation supports:

- W8A8_DYNAMIC.
- AllGather.
- TP-only execution.
- Static multi-adapter loading.
- Eager execution.
- Routed experts, shared experts, and attention LoRA.

The following quantized combinations fail explicitly:

- Expert Parallel or AlltoAll.
- MC2 or FusedMC2.
- Dynamic EPLB.
- W4A8, FP8, and MXFP MoE LoRA.
- Derived W8A8FP or PDMix schemes.

The existing BF16 MoE LoRA AllGather and AlltoAll paths are unchanged.
## Verification
Here I verified this pr from deepseekv4 flash w8a8-int8, and the lora is trainied by mindspeed.
Since mindspeed does not support transfering dsv4flash lora checkpoint  from megatron to peft, so I solve this by my self https://github.com/SkychenLee/LoraAdapterTransformer/blob/main/README.md , you  can  directly use the scritpt to transfer the lora adapter for dsv4 flash or you can use the skill I provided to transfer any model's lora adapters from any code agent.
Here are the command for **start up** of the model
```
#!/usr/bin/bash
export OMP_PROC_BIND=false
export OMP_NUM_THREADS=10
export PYTORCH_NPU_ALLOC_CONF=expandable_segments:True

export TASK_QUEUE_ENABLE=1
export HCCL_BUFFSIZE=512

sysctl -w vm.swappiness=0
sysctl -w kernel.numa_balancing=0
sysctl kernel.sched_migration_cost_ns=50000
 
vllm serve "your deep seek v4 flash moel path" \
  --safetensors-load-strategy 'prefetch' \
  --max-model-len 10240 \
  --max-num-batched-tokens 10240 \
  --served-model-name ds \
  --max-num-seqs 32 \
  --data-parallel-size 1 \
  --tensor-parallel-size 8 \
  --quantization ascend \
  --port 7000 \
  --block-size 128 \
  --enable-chunked-prefill \
  --no-enable-prefix-caching \
  --enable-auto-tool-choice \
  --async-scheduling \
  --enable-lora \
  --lora-modules lora1="your lora adapter path"\
  --max-lora-rank 16 
```
## Tests

Unit tests were added for:

- LoRA-aware prepare and dispatch quantization.
- Base-only batch preservation of the existing W8A8 path.
- Two dynamic activation quantizations.
- Decode-time `no_lora` metadata refresh.
- Unsupported quantization and communication fail-fast behavior.
- W8A8 quantization method LoRA-context propagation.

Static validation completed:

- Python compilation checks.
- Ruff lint checks.
- Ruff formatting checks.
- Git diff whitespace checks.

NPU numerical comparison and DeepSeek V4 Flash end-to-end validation still need to be performed on an Ascend environment with `torch_npu` and the target hardware.

- vLLM version: v0.26.0


vllm_ascend/lora/fused_moe.py
```diff
@@ -58,6 +48,32 @@
...
+        raise ValueError(f"Expected 2 or 3 routed-expert LoRA B tensors, got {len(lora_b)}")
+
+    w1_lora_b, w2_lora_b, w3_lora_b = lora_b
+    w13_enabled = bool(torch.count_nonzero(w1_lora_b).item())
+    if w13_num_slices == 2:
+        w13_enabled = w13_enabled or bool(torch.count_nonzero(w3_lora_b).item())
+    return w13_enabled, bool(torch.count_nonzero(w2_lora_b).item())
+
+
+def is_moe_lora_active(lora_context) -> bool:
```
voidvelocity: The function implemention is not for moe only, so I think rename it as `has_lora` etc. is better.

vllm_ascend/lora/fused_moe.py
```diff
@@ -58,6 +48,32 @@
...
+    w13_num_slices: int,
+) -> tuple[bool, bool]:
+    """Return whether the routed w13 and w2 B factors are non-zero."""
+    if len(lora_b) == 2:
+        w13_lora_b, w2_lora_b = lora_b
+        return (
+            bool(torch.count_nonzero(w13_lora_b).item()),
+            bool(torch.count_nonzero(w2_lora_b).item()),
+        )
+    if len(lora_b) != 3:
```
voidvelocity: `!=3` but raise "Expected 2 or 3", not readable, it's suggested to set all raise at the beginning of functions. That is:

```
if len(lora_b) not in (2, 3):
    raise ValueError(f"xxx")
```

vllm_ascend/lora/fused_moe.py
```diff
@@ -338,11 +361,55 @@ def __init__(self, base_layer: nn.Module) -> None:
...
+            self._shared_experts = shared_experts
+
+    def create_lora_weights(
+        self,
+        max_loras,
+        lora_config,
+        model_config=None,
+    ) -> None:
+        super().create_lora_weights(max_loras, lora_config, model_config)
+        self.w13_adapter_enabled = torch.zeros_like(self.adapter_enabled)
```
voidvelocity: What's the difference of `self.adapter_enabled` and `self.w13_adapter_enabled / self.w2_adapter_enabled`? Why we need it, but the upstream vllm don't need it?

vllm_ascend/lora/punica_npu.py
```diff
@@ -9,6 +9,51 @@
 from vllm_ascend.utils import AscendDeviceType, get_ascend_device_type
 
 
+@torch.library.custom_op(
+    "vllm_ascend::moe_lora_bmm_expand_slice",
+    mutates_args={"y"},
+)
+def _moe_lora_bmm_expand_slice_op(
```
voidvelocity: If this op is necessary, move it to file `vllm_ascend/lora/lora_ops.py`.

vllm_ascend/lora/punica_npu.py
```diff
@@ -9,6 +9,51 @@
...
+    delta.masked_fill_((indices < 0).unsqueeze(-1), 0.0)
+    y_slice = y.narrow(1, y_offset, y_slice_size)[: delta.shape[0]]
+    if add_inputs:
+        y_slice.add_(delta)
+    else:
+        y_slice.copy_(delta)
+
+
+@_moe_lora_bmm_expand_slice_op.register_fake
+def _moe_lora_bmm_expand_slice_fake(
```
voidvelocity: If this op is necessary, move it to file `vllm_ascend/lora/lora_ops.py`.

vllm_ascend/lora/punica_npu.py
```diff
@@ -9,6 +9,51 @@
...
+def _moe_lora_bmm_expand_slice_op(
+    y: torch.Tensor,
+    x: torch.Tensor,
+    w_t_all: torch.Tensor,
+    indices: torch.Tensor,
+    y_offset: int,
+    y_slice_size: int,
+    add_inputs: bool,
+) -> None:
+    """Dense LoRA B projection for shapes unsupported by expand-slice."""
```
voidvelocity: Which shape is not supported by original expand-slice?

vllm_ascend/lora/punica_npu.py
```diff
@@ -144,6 +226,25 @@ def _expand_slice_decode(
             add_inputs,
         )
 
+    def _bmm_expand_slice(
```
voidvelocity: I don't think it's necessary to add new bmm_expand_slice. All relative code should be removed. If original bmm_expand_slice has bug, fix it instead create a new branch.

vllm_ascend/ops/fused_moe/moe_mlp.py
```diff
@@ -634,6 +634,13 @@ def unified_apply_mlp(*, mlp_compute_input: MoEMlpComputeInput) -> torch.Tensor:
             topk_ids=mlp_compute_input.topk_ids,
         )
 
+    from vllm_ascend.lora.fused_moe import is_moe_lora_active
+
+    if is_moe_lora_active(mlp_compute_input.lora_context):
+        from vllm_ascend.lora.quant_moe import apply_quant_moe_lora
+
+        return apply_quant_moe_lora(mlp_compute_input=mlp_compute_input)
```
voidvelocity: Keep function name style consistent. I think `quant_apply_mlp_with_moe_lora` is better.

vllm_ascend/ops/fused_moe/prepare_finalize.py
```diff
@@ -369,14 +370,23 @@ def _prepare_with_ep_group(
...
+        quantize_activations = True
+        if is_moe_lora_active(self.lora_context) and quant_type != QuantType.NONE:
+            quantize_activations = configure_quant_moe_lora_activation_quantization(
+                quant_type=quant_type,
+                hidden_states=hidden_states,
+                dynamic_scale=None,
+                with_quant=True,
+            )
+
+        if quantize_activations and quant_type == QuantType.W8A8:
```
voidvelocity: All the if branches add `quantize_activations`, will `quantize_activations=False` but `quant_type == QuantType.W8A8` happen? If happens, the new code has different logic as original code.

vllm_ascend/lora/quant_moe.py
```diff
@@ -0,0 +1,273 @@
...
+
+def apply_quant_moe_lora(
+    *,
+    mlp_compute_input: MoEMlpComputeInput,
+) -> tuple[torch.Tensor, torch.npu.Event | None]:
+    """Dispatch an active quantized MoE LoRA batch to its backend."""
+    return _get_quant_moe_lora_impl(mlp_compute_input.quant.quant_type).apply(mlp_compute_input)
+
+
+def configure_quant_moe_lora_activation_quantization(
```
voidvelocity: This function name doesn't reflect what it does. It name as `configure`, but `validate` and return a bool value to show if with quantization. If can't find a proper name for a function, normally means it need refactoring.

vllm_ascend/quantization/methods/w8a8_dynamic.py
```diff
@@ -234,6 +234,12 @@ def apply(
         mc2_mask: torch.Tensor | None = None,
         tid2eid: torch.Tensor | None = None,
     ) -> torch.Tensor:
+        lora_context = getattr(layer, "_ascend_moe_lora_context", None)
+        if lora_context is not None and type(self) is not AscendW8A8DynamicFusedMoEMethod:
+            raise NotImplementedError(
+                "Ascend quantized MoE LoRA currently supports only the "
+                "W8A8_DYNAMIC scheme, not derived W8A8FP/PDMix schemes."
```
voidvelocity: Just tell developer what type the code sees, print the `type(self)` in log.

vllm_ascend/ops/fused_moe/prepare_finalize.py
```diff
@@ -369,14 +370,23 @@ def _prepare_with_ep_group(
         self, hidden_states: torch.Tensor, router_logits: torch.Tensor, quant_type=QuantType.NONE
     ) -> MoEPrepareOutput:
         pertoken_scale = None
-        if quant_type == QuantType.W8A8:
+        quantize_activations = True
```
voidvelocity: It's suggested to set var name as `with_quant`, keep the same as latter code.

vllm_ascend/lora/punica_npu.py
SkychenLee: Agreed. I removed the custom BMM expand-slice op, the prefill/decode fallback branches, the shape-selection logic, and the related tests. The implementation now uses the existing `sgmv_expand_slice` and `bgmv_expand_slice` paths. Any unsupported shape should be fixed in the original operator instead of introducing a separate branch.

vllm_ascend/lora/punica_npu.py
SkychenLee: After rechecking the shared-expert path, there is no shared-expert-specific shape that should be handled by a separate fallback here. I removed the shape check and the BMM fallback. All dense LoRA expand-slice calls now use the existing native operators.

vllm_ascend/lora/punica_npu.py
SkychenLee: Agreed. The custom BMM expand-slice op and its fake implementation have been removed completely, so no new op needs to be added to `lora_ops.py`.

vllm_ascend/lora/punica_npu.py
SkychenLee: Agreed. The custom BMM expand-slice op and its fake implementation have been removed completely, so no new op needs to be added to `lora_ops.py`.

vllm_ascend/lora/fused_moe.py
SkychenLee: The original intent was to distinguish adapters that provide only w13 or only w2 weights, because `adapter_enabled` is adapter-level.

After reviewing the upstream loading behavior, the missing projection buffers are already zero-initialized/reset, so applying them naturally produces a zero delta. The projection-specific masks are therefore not required for correctness and do not actually avoid the kernel invocation in the current implementation.

I removed `w13_adapter_enabled`, `w2_adapter_enabled`, and the related helper, and now use the upstream `adapter_enabled` directly.

vllm_ascend/lora/fused_moe.py
SkychenLee: Thanks for the suggestion. After reviewing the projection-specific enable masks, I found that the helper itself was unnecessary and removed it together with those masks, so this validation is no longer needed.

vllm_ascend/lora/fused_moe.py
SkychenLee: Thanks for pointing this out. This helper only checks whether the current batch contains an active LoRA request and is not MoE-specific. I renamed it to `has_lora`.

vllm_ascend/ops/fused_moe/prepare_finalize.py
SkychenLee: Agreed. I renamed `quantize_activations` to `with_quant` to keep it consistent with the token-dispatch path and the existing quantization arguments.

vllm_ascend/quantization/methods/w8a8_dynamic.py
SkychenLee: Thanks. I updated the exception to include the actual quantization method type through `type(self)`, so developers can directly identify the unsupported derived implementation.

vllm_ascend/lora/quant_moe.py
SkychenLee: Agreed. The function mixed two responsibilities: validating the current activation state and deciding whether the next stage should quantize it.

I split it into:

- `validate_quant_moe_lora_activation_input`, which only validates the activation and scale.
- `should_quantize_moe_lora_activations`, which only returns the backend quantization policy.

The prepare and token-dispatch call sites now invoke these responsibilities explicitly.

vllm_ascend/ops/fused_moe/prepare_finalize.py
SkychenLee: Yes, `with_quant=False` while `quant_type == QuantType.W8A8` intentionally happens for an active quantized MoE LoRA batch.

The original path quantizes activations before AllGather, but the LoRA A projection requires the original BF16/FP16 activations. Therefore, the active-LoRA path preserves floating-point activations during prepare/dispatch and performs dynamic quantization immediately before GMM1 and GMM2 in the quantized MoE LoRA backend.

For a base-only batch, `has_lora(...)` is false and the original prepare-side W8A8 quantization path remains unchanged. I added a comment to make this intentional behavior explicit.

vllm_ascend/ops/fused_moe/moe_mlp.py
SkychenLee: Agreed. I renamed `apply_quant_moe_lora` to `quant_apply_mlp_with_moe_lora` to keep it consistent with the existing `quant_apply_mlp` and `unquant_apply_mlp` naming style. The call sites and tests have been updated accordingly.

vllm_ascend/ops/fused_moe/prepare_finalize.py
```diff
@@ -380,14 +384,28 @@ def _prepare_with_ep_group(
...
+        with_quant = True
+        if has_lora(self.lora_context) and quant_type != QuantType.NONE:
+            validate_quant_moe_lora_activation_input(
+                quant_type=quant_type,
+                hidden_states=hidden_states,
+                dynamic_scale=None,
+            )
+            # LoRA A needs the original floating-point activations. The active
+            # quantized MoE LoRA backend quantizes immediately before its GMMs.
+            with_quant = should_quantize_moe_lora_activations(
```
kunpengW-code: Not only in this one place, but many `with_quant` constructions and calls have been added. Essentially, when `has_lora` is present, communication quantization is not performed, and the corresponding code can be simplified. Currently, `with_quant` has increased the code's confusion.

Angazenn:
I see you have isolated quant moe mlp from normal no-lora path. But why do you directly edit shared_expert? what's the guiding principle?

SkychenLee:
> I see you have isolated quant moe mlp from normal no-lora path. But why do you directly edit shared_expert? what's the guiding principle?

In lora senario, the baseLayer of shared experted is in side "SharedExpertLoraLayer", it is a attribute in the LoraLayer, we can not directly obtain it from ".weight"

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -227,5 +227,6 @@ def _forward_impl(
                     swiglu_alpha=getattr(fused_moe_results, "swiglu_alpha", 1.0),
                     swiglu_beta=getattr(fused_moe_results, "swiglu_beta", 0.0),
                 ),
+                lora_context=getattr(self.routed_experts, "_ascend_moe_lora_context", None),
```
weijinqian0: SharedExperts set_lora.

vllm_ascend/ops/fused_moe/prepare_finalize.py
```diff
@@ -380,7 +381,15 @@ def _prepare_with_ep_group(
...
         pertoken_scale = None
-        if quant_type == QuantType.W8A8:
+        if has_lora(self.lora_context) and quant_type != QuantType.NONE:
+            validate_quant_moe_lora_activation_input(
+                quant_type=quant_type,
+                hidden_states=hidden_states,
+                dynamic_scale=None,
+            )
+            # LoRA A needs the original floating-point activations. The active
+            # quantized MoE LoRA backend quantizes immediately before its GMMs.
```
weijinqian0: plz remove this code.

vllm_ascend/quantization/methods/w8a8_dynamic.py
```diff
@@ -215,6 +216,17 @@ def apply(
...
+            raise NotImplementedError(
+                "Ascend quantized MoE LoRA currently supports only the AllGather TP path; "
+                "AlltoAll, MC2, and FusedMC2 are unsupported."
+            )
+        if lora_context is not None and type(self) is not AscendW8A8DynamicFusedMoEMethod:
+            raise NotImplementedError(
+                "Ascend quantized MoE LoRA currently supports only "
+                f"AscendW8A8DynamicFusedMoEMethod, but got {type(self)}."
+            )
         assert topk_ids is not None
```
weijinqian0, line 234: use select_moe_comm_method in ascend_forward_context.

SkychenLee: agree, I have delete these codes

vllm_ascend/ops/fused_moe/prepare_finalize.py
SkychenLee: agree, already remove

vllm_ascend/ops/fused_moe/fused_moe.py
SkychenLee: agree, I have adapt with lora_context

weijinqian0 (approved):

SkychenLee:
/rerun
[Bot]: rerun completed.

**Failed:**

SkychenLee:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

SkychenLee:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

SkychenLee:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

SkychenLee:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

SkychenLee:
/rerun

SkychenLee:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

paulyu12:
This PR completes W8A8 MoE LoRA support for AlltoAll + EP and correctly enables LoRA adaptation for shared-expert MoE models. The implementation looks good to me.

PR #13372, opened 2026-08-03
[BugFix][SpecDecode][Mamba] Fix DSpark slot mapping and mamba postprocess signature on hybrid targets

fg11991:

Two independent bugs that, together, make DSpark speculative decoding unusable on
hybrid Gated-DeltaNet targets (Qwen3.5 / Qwen3.6). Both are Ascend-only and both
are invisible on pure-attention targets such as Qwen3-8B, which is why CI has not
caught them — `tests/e2e/pull_request/one_card/spec_decode/test_dspark.py` only
covers `Qwen/Qwen3-8B`.

#### 1. `postprocess_mamba_fused_kernel` signature drifted from upstream vLLM

`vllm_ascend/patch/worker/patch_mamba_utils.py` installs the Ascend
`postprocess_mamba_fused_kernel` over
`vllm.v1.worker.mamba_utils.postprocess_mamba_fused_kernel`, but leaves the
launcher upstream. Upstream has since added `state_dim_row_count_ptr`,
`state_dim_row_stride_ptr`, `idx_mapping_ptr`, `CONV_STATE_DIM_FIRST`,
`HAS_IDX_MAPPING` and `PRECOMPUTED_NEW_COMPUTED`, so every launch raises
`TypeError` and hybrid models cannot be served at all.

This PR realigns the signature and implements the two modes that are only index
arithmetic:

* `HAS_IDX_MAPPING` — resolve `program_id(0)` from a batch index to a req-state
  slot, skipping negative entries.
* `PRECOMPUTED_NEW_COMPUTED` — read the post-step computed count directly.

`CONV_STATE_DIM_FIRST` (the `DS` conv-state layout) needs a different stride walk
in the copy loop, which this kernel does not implement. It is rejected with a
`tl.static_assert` rather than silently copying the wrong bytes. That layout is
only reachable via `VLLM_SSM_CONV_STATE_LAYOUT=DS`, which vllm-ascend never sets,
so the default path is unaffected.

#### 2. DSpark derives slot mappings from the KV-manager block size

`AscendDSparkProposer` builds the draft query-block slot mapping itself and passes
`kv_cache_spec.block_size` to
`copy_and_expand_dflash_and_dspark_inputs_kernel_single_grid`. But the block table
that kernel indexes, the physical KV cache and FIA all address memory in **kernel
blocks**:

* `AscendAttentionBackend.get_supported_kernel_block_sizes()` returns `[128]`.
* When the KV-manager block size differs, `BlockTable` sets
  `self.block_size = kernel_block_size` and expands every manager block id into
  `blocks_per_kv_block` kernel block ids (`vllm/v1/worker/block_table.py`).
* `_reshape_kv_cache_tensors` shapes the physical cache with
  `get_supported_kernel_block_sizes()[0]` whenever `use_hybrid_blocks` is set, and
  `AscendAttentionBackendImpl._get_fia_params` reads its block size back out of
  `key_cache.shape[1]`.

The two sizes coincide on pure-attention targets. On a hybrid target they do not:
vLLM enlarges the attention manager block size so the attention page size matches
the mamba page size (`unify_kv_cache_spec_page_size`). Note this happens **on top
of** whatever `--block-size` the user passes — the repro below explicitly sets
`--block-size 128` and still hits the bug.

The draft's query-block K/V is therefore written to wrong, out-of-range slots. The
draft then attends over uninitialized KV cache, its hidden states become `NaN`,
and the accepted length collapses.

The eagle/mtp path already solves exactly this in
`AscendSpecDecodeBaseProposer.attn_update_stack_num_spec_norm`:

```python
# NOTE: In vllm, `block_size = attn_metadata_builder.kv_cache_spec.block_size`.
# However, in vllm-ascend, the above value can be multiple of `kernel_block_size`,
# which is not correct for computing `slot_mapping` below.
if self.has_gdn:
    block_size = self.kernel_block_size
```

DFlash is unaffected because it uses the base proposer's `kernel_block_size`. Only
DSpark overrides it with the manager block size. This PR resolves the value from
`BlockTable.block_size`, which already holds the kernel block size when splitting
is active and the manager block size when it is not.

No. Both changes are bug fixes.

For (2), on pure-attention targets `BlockTable.block_size == kv_cache_spec.block_size`,
so the resolved value — and therefore the generated slot mapping — is bit-identical
to before.

**Unit tests (new):**

* `tests/ut/ops/test_mamba_postprocess_signature.py` — parses the upstream kernel
  definition out of the installed vLLM source and compares the parameter list
  against the Ascend kernel, so the next upstream signature change fails a cheap
  test instead of every hybrid-model launch.
* `tests/ut/spec_decode/test_dspark_proposer.py::TestKernelBlockSizeResolution` —
  covers the split case (manager 512 / kernel 128), the non-split case
  (128 / 128, asserting unchanged behaviour), the no-runner fallback, and that
  `set_inputs_first_pass` forwards the kernel block size to the inputs kernel.

```bash
pytest -sv tests/ut/spec_decode/test_dspark_proposer.py \
           tests/ut/ops/test_mamba_postprocess_signature.py
```

**Manual end-to-end, Atlas 800I A2 (910B) x4, TP=4:**

Target `Qwen3.6-27B` (hybrid Gated-DeltaNet), draft an internally trained
Qwen3.6-27B DSpark drafter with `block_size=7`.

```bash
export ASCEND_RT_VISIBLE_DEVICES=0,1,2,3
python -m vllm.entrypoints.openai.api_server \
  --model /path/to/Qwen3.6-27B \
  --tensor-parallel-size 4 \
  --data-parallel-size 1 \
  --trust-remote-code \
  --max-model-len 8192 \
  --max-num-seqs 1 \
  --dtype bfloat16 \
  --enable-prefix-caching \
  --block-size 128 \
  --enable-chunked-prefill \
  --enforce-eager \
  --speculative-config '{"method":"dspark","model":"/path/to/dspark-draft","num_speculative_tokens":7,"enforce_eager":"True"}'
```

Before this PR the server could not start at all until (1) was fixed. With (1)
applied but not (2), a single greedy request
(`"Write a Python function to compute the nth Fibonacci number iteratively."`,
`temperature=0`, `max_completion_tokens=512`) gave:

| | overall acceptance rate | notes |
|---|---|---|
| before (2) | **20%** | draft hidden states are `NaN`; the last three of the seven positions accept nothing |
| after (2)  | **97.5%** | the first four of the seven positions accept everything |

The same drafter scores an average accepted length of 6.117 over 7 steps in the
reference (non-vLLM) implementation, which is consistent with the post-fix number
and not with the pre-fix one.

**Not covered:** the end-to-end numbers above are from a single prompt at
`temperature=0`, and I have not re-run the Qwen3-8B DSpark e2e
(`tests/e2e/pull_request/one_card/spec_decode/test_dspark.py`) on hardware. That
path is a no-op by construction — see the user-facing-change section — and the new
unit test pins the non-split case, but a maintainer may want CI to confirm it.

- vLLM version: v0.26.0


vllm_ascend/spec_decode/dspark_proposer.py
```diff
@@ -99,9 +99,35 @@ def __init__(
...
+
+        The eagle/mtp path solves the same problem in
+        ``AscendSpecDecodeBaseProposer.attn_update_stack_num_spec_norm``.
+        """
+        gid = attn_group.kv_cache_group_id
+        input_batch = getattr(self.runner, "input_batch", None)
+        block_tables = getattr(input_batch, "block_table", None)
+        if block_tables is None:
+            return int(attn_group.kv_cache_spec.block_size)
+        return int(block_tables[gid].block_size)
```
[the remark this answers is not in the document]
fg11991: Thanks, but this suggestion would silently undo the fix, so I have not taken it.

`block_tables` here is a `MultiGroupBlockTable`, not a mapping. It defines only
`__getitem__(idx: int)` — no `__contains__`, `__iter__` or `__len__`
(`vllm/v1/worker/block_table.py`). `gid not in block_tables` therefore falls back to
the legacy iteration protocol, comparing `BlockTable` **objects** against an `int`,
so it is always `True`. The function would then always return
`kv_cache_spec.block_size`, which is exactly the value that writes the draft's
query-block K/V to out-of-range slots on hybrid targets — the NaN this PR fixes
would come back, silently.

That said, the underlying instinct is fair: there was a path that could quietly
return the wrong block size. I have addressed it in the opposite direction in
6a87ef0 — the `block_tables is None` fallback is gone and a missing runner is now
an assert. `initialize_attn_backend` is only ever driven by the model runner, which
rebuilds the input batch with the correct kernel block sizes
(`may_reinitialize_input_batch`) before calling it, so the runner is always present
in production; the fallback only existed to keep runner-less unit tests working.
Failing loudly beats falling back to a value known to corrupt the KV cache.

vllm_ascend/spec_decode/dspark_proposer.py
```diff
@@ -236,7 +266,8 @@ def set_inputs_first_pass(
             gid_block_table = self._per_group_block_table_buffers.get(gid)
             if gid_block_table is None:
                 continue
-            kv_block_size = int(attn_group.kv_cache_spec.block_size)
+            # Slot ids are expressed in kernel blocks, not KV manager blocks.
+            kv_block_size = self._per_group_kernel_block_size[gid]
```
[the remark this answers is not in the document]
fg11991: Not taken, for the same reason as the other comment: the proposed default is the
KV manager block size, which is the value that corrupts the KV cache on hybrid
targets. Swapping a `KeyError` for silent NaN hidden states and a collapsed
acceptance length is a bad trade — if this key were ever missing, that would mean
a broken invariant and we want it to be loud.

It also cannot happen by construction. `_per_group_kernel_block_size` is built by
a dict comprehension over `self.draft_attn_groups` in `initialize_attn_backend`,
and this loop iterates the same `self.draft_attn_groups`, so every `gid` reached
here is guaranteed to be a key.

drslark:
Good job.
Since there are currently no e2e tests for Qwen3.5 DSpark, please submit an issue describing the specific operating conditions.

tests/ut/ops/test_mamba_postprocess_signature.py
```diff
@@ -0,0 +1,67 @@
...
+import inspect
+
+from vllm.v1.worker import mamba_utils as upstream_mamba_utils
+
+from vllm_ascend.ops.triton.mamba.postprocess import postprocess_mamba_fused_kernel
+
+_KERNEL_NAME = "postprocess_mamba_fused_kernel"
+
+
+def _upstream_param_names() -> list[str]:
```
drslark: This test is redundant.
Please remove it.

tests/ut/spec_decode/test_dspark_proposer.py
```diff
@@ -691,4 +694,79 @@ def test_no_draft_attn_groups_raises(self, monkeypatch):
...
+        proposer = AscendDSparkProposer.__new__(AscendDSparkProposer)
+        proposer.runner = None
+        attn_group = SimpleNamespace(
+            kv_cache_group_id=0,
+            kv_cache_spec=SimpleNamespace(block_size=128),
+        )
+        with pytest.raises(AssertionError, match="model runner"):
+            proposer._resolve_kernel_block_size(attn_group)
+
+    def test_inputs_kernel_receives_kernel_block_size(self, monkeypatch):
```
drslark: This test is redundant.
Please remove it.

tests/ut/spec_decode/test_dspark_proposer.py
```diff
@@ -691,4 +694,79 @@ def test_no_draft_attn_groups_raises(self, monkeypatch):
...
+        # Hybrid target: manager block size is a multiple of the kernel one.
+        proposer, attn_group = self._make_proposer_with_block_table(512, 128)
+        assert proposer._resolve_kernel_block_size(attn_group) == 128
+
+    def test_unsplit_blocks_keep_manager_block_size(self):
+        # Pure-attention target: the two coincide, behaviour is unchanged.
+        proposer, attn_group = self._make_proposer_with_block_table(128, 128)
+        assert proposer._resolve_kernel_block_size(attn_group) == 128
+
+    def test_missing_runner_raises_instead_of_falling_back(self):
```
drslark: This test is redundant.
Please remove it.

vllm_ascend/spec_decode/dspark_proposer.py
```diff
@@ -99,9 +99,39 @@ def __init__(
         # per-gid context slot_mapping buffer
         self._per_group_context_slot_mapping_buffers: dict[int, torch.Tensor] = {}
 
+        # per-gid block size used to derive slot ids (see _resolve_kernel_block_size)
+        self._per_group_kernel_block_size: dict[int, int] = {}
+
         # per-layer context slot mappings as a flat list
         self._context_slot_mapping_buffers: list[torch.Tensor | None] | None = None
 
+    def _resolve_kernel_block_size(self, attn_group) -> int:
```
drslark: There is no need to extract a function.

vllm_ascend/spec_decode/dspark_proposer.py
```diff
@@ -161,7 +191,11 @@ def initialize_attn_backend(self, kv_cache_config, kernel_block_sizes=None) -> N
             )
 
         self.kv_cache_gid = self.draft_attn_groups[0].kv_cache_group_id
-        self.kernel_block_size = int(self.draft_attn_groups[0].kv_cache_spec.block_size)
+        self._per_group_kernel_block_size = {
```
drslark: Please follow the logic in `llm_base_proposer`.

```python
        # initialized for mamba models
        self.kernel_block_size = (
            draft_attn_layers_dict[self.attn_layer_names[0]].get_attn_backend().get_supported_kernel_block_sizes()[0]
        )
```

drslark: @slippersss Please check this pr.

drslark (changes requested):

tests/ut/ops/test_mamba_postprocess_signature.py
fg11991: Removed in 2085432.

One note for the record, since you also pointed out there is no e2e for Qwen3.5
DSpark: this bug *was* an upstream signature drift, and with the test gone there
is now nothing in the repo that catches the next one — the kernel is swapped in by
`patch_mamba_utils` while the launcher stays upstream, so a new upstream parameter
turns into a `TypeError` at every launch. I am fine with your call and have removed
it; if you would rather have a cheap guard than none, say so and I will put it back.

tests/ut/spec_decode/test_dspark_proposer.py
fg11991: Removed in 2085432 — following your other comment, `_resolve_kernel_block_size` is
gone entirely, so these cases lost their subject anyway.

I kept exactly one test, `test_inputs_kernel_receives_kernel_block_size`, which
asserts that `set_inputs_first_pass` forwards `self.kernel_block_size` rather than
`kv_cache_spec.block_size` to the inputs kernel. That is the actual defect, it needs
no NPU, and it is ~12 lines. Proposing it as a middle ground given AGENTS.md asks
bug fixes to carry a regression test — but if you consider it redundant too, I will
drop it and the PR will carry no new tests.

tests/ut/spec_decode/test_dspark_proposer.py
fg11991: Removed in 2085432 — following your other comment, `_resolve_kernel_block_size` is
gone entirely, so these cases lost their subject anyway.

I kept exactly one test, `test_inputs_kernel_receives_kernel_block_size`, which
asserts that `set_inputs_first_pass` forwards `self.kernel_block_size` rather than
`kv_cache_spec.block_size` to the inputs kernel. That is the actual defect, it needs
no NPU, and it is ~12 lines. Proposing it as a middle ground given AGENTS.md asks
bug fixes to carry a regression test — but if you consider it redundant too, I will
drop it and the PR will carry no new tests.

vllm_ascend/spec_decode/dspark_proposer.py
fg11991: Done in 2085432 — the helper is gone.

vllm_ascend/spec_decode/dspark_proposer.py
fg11991: Done in 2085432, and following that pointer turned out to shrink the fix a lot —
thanks.

`load_model` in `llm_base_proposer` already sets it from the draft backend:

```python
self.kernel_block_size = (
    draft_attn_layers_dict[self.attn_layer_names[0]].get_attn_backend().get_supported_kernel_block_sizes()[0]
)
```

and `drafter.load_model()` (model_runner_v1.py:3462) runs before
`drafter.initialize_attn_backend()` (model_runner_v1.py:3593). So the DSpark line
was not filling in a missing value — it was overwriting a correct one with the KV
manager block size. The fix is therefore just to stop overwriting it and to use
`self.kernel_block_size` in `set_inputs_first_pass`, exactly like
`dflash_proposer.py` does.

The production change is now two lines:

```diff
-        self.kernel_block_size = int(self.draft_attn_groups[0].kv_cache_spec.block_size)
+        # NOTE: do not overwrite self.kernel_block_size with kv_cache_spec.block_size ...

-            kv_block_size = int(attn_group.kv_cache_spec.block_size)
+            kv_block_size = self.kernel_block_size
```

fg11991:
@drslark thanks for the review — all five comments are addressed in 2085432, and
following your `llm_base_proposer` pointer shrank the production change to two lines
(`load_model` already sets `kernel_block_size` correctly and runs before
`initialize_attn_backend`, so the DSpark line was overwriting a correct value rather
than filling in a missing one). Details in the individual threads.

Filed #13380 with the operating conditions, as requested. It also notes that we can
provide a DSpark drafter checkpoint if CI needs one — comment there with what form is
most useful.

fg11991:
@211840040 Quite possibly, yes — but your symptom differs from ours in one detail
worth chasing, so I would confirm rather than assume. Two things you can check before
rebuilding anything.

**First, is it this bug?** The fact that your server *starts* tells us the mamba
kernel-signature half of this PR is not what you are hitting (that one raises
`TypeError` before the server is up), so only the DSpark slot-mapping half is
relevant to you. That half is two lines, and `releases/v0.25.1rc` carries the
identical code at the same line numbers, so you do not need to move to main:

```python
# vllm_ascend/spec_decode/dspark_proposer.py:164 -- delete this line
- self.kernel_block_size = int(self.draft_attn_groups[0].kv_cache_spec.block_size)

# vllm_ascend/spec_decode/dspark_proposer.py:239
- kv_block_size = int(attn_group.kv_cache_spec.block_size)
+ kv_block_size = self.kernel_block_size
```

`llm_base_proposer.load_model` already sets `self.kernel_block_size` correctly on
v0.25.1rc too (line 310), and it runs before `initialize_attn_backend`, so deleting
the override is all that is needed.

To confirm the bug applies to your configuration at all, add this next to line 164
and look at one startup log line:

```python
for g in self.draft_attn_groups:
    gid = g.kv_cache_group_id
    bt = self.runner.input_batch.block_table[gid]
    print(f"[dspark] manager_bs={g.kv_cache_spec.block_size} kernel_bs={bt.block_size} hybrid={bt.use_hybrid_blocks}")
```

If `manager_bs != kernel_bs` (`hybrid=True`), you are hitting it. If they are equal,
this PR will not help you and the cause is elsewhere.

**Second, "exactly 0" is suspicious.** With the slot-mapping bug we measured ~20%,
not 0 — a NaN draft still emits tokens, and a few get accepted by luck. Exactly 0
suggests something additional, so please check:

* `curl -s <host>/metrics | grep spec_decode_num_drafts` — if this is **0**, no
  drafting is happening at all, which is a different failure from "drafted and
  rejected". Worth knowing which one you have.
* What `num_speculative_tokens` are you passing? The config on
  `satgeze/Qwen3.6-27B-DSpark` declares `"block_size": 15`, i.e. the drafter was
  trained to emit a 15-token block. DSpark drafts the whole block in one
  non-causal pass, so running it at a different block length is not a
  configuration knob the way it is for EAGLE — try `"num_speculative_tokens": 15`
  if you were using something else.
* That checkpoint also sets `"enable_confidence_head": true`. vllm-ascend does not
  wire the confidence head yet (its weights are skipped at load), so dynamic
  proposal length is simply not applied. Not fatal, but it means you will not
  reproduce that checkpoint's reference numbers exactly.

If you can post your launch command and the `spec_decode_num_drafts` /
`spec_decode_num_accepted_tokens_per_pos` values, that would narrow it down quickly.
Our own repro conditions are written up in #13380 for comparison.

vllm_ascend/spec_decode/dspark_proposer.py
```diff
@@ -236,7 +241,8 @@ def set_inputs_first_pass(
             gid_block_table = self._per_group_block_table_buffers.get(gid)
             if gid_block_table is None:
                 continue
-            kv_block_size = int(attn_group.kv_cache_spec.block_size)
+            # Slot ids are expressed in kernel blocks, not KV manager blocks.
+            kv_block_size = self.kernel_block_size
```
slippersss: I guess this will cause error in dsv4. In dsv4 and most of other models, `kv_block_size` may be `attn_group.kv_cache_spec.block_size` strictly instead of `self.kernel_block_size` defined in llm_base_proposer. You may find something like `if self.has_gdn: ...` branch in llm_base_proposer, and that is specially designed for gdn-like models. Please double check #12000 carefully.

fg11991: Good catch, and it's worse than "may be" — `load_model` takes
`get_supported_kernel_block_sizes()[0]`, and for DSV4's DSA backend that list is
`[2, 4, 8, 16, 32, 64, 128]`, so the value would be 2.

Gated both sites on `self.has_gdn` in 550bafb, mirroring #12000. Non-GDN is
byte-identical to before. Test is parametrized over `has_gdn` so both branches
are covered.

211840040:
@fg11991 Thank you for your guidance. I added the logging after calling self.drafter.initialize_attn_backend in model_runner_v1.initialize_kv_cache around line 3600 (since draft_attn_groups are only properly set at that point). The output was [dspark] manager_bs=1536 kernel_bs=128 hybrid=True, confirming that the defect does indeed apply to my setup.

I then made the same changes in dspark_proposer.py as described in the PR, and the acceptance rate returned to normal — it is now largely consistent with the verification results the model author reported on vLLM GPU. This is great, and thank you again for your help!

One minor note: apologies for the confusion — I had moved the issue description to Issue #13380 before seeing your reply.

vllm_ascend/spec_decode/dspark_proposer.py
```diff
@@ -161,7 +161,11 @@ def initialize_attn_backend(self, kv_cache_config, kernel_block_sizes=None) -> N
             )
 
         self.kv_cache_gid = self.draft_attn_groups[0].kv_cache_group_id
-        self.kernel_block_size = int(self.draft_attn_groups[0].kv_cache_spec.block_size)
+        # On GDN targets the KV manager block size is a multiple of the kernel block
+        # size, so keep the value load_model derived from the draft backend. See the
+        # `has_gdn` branch in llm_base_proposer (#12000).
+        if not self.has_gdn:
```
drslark: If there is no gdn, do we still need `self.kernel_block_size`?
Maybe `self.kernel_block_size` from load_model is enough?

drslark (changes requested):

vllm_ascend/spec_decode/dspark_proposer.py
fg11991: Right, dropped it in 26cefc3.

Checked the readers first — nothing in the DSpark path reads
`self.kernel_block_size` when `has_gdn` is false:

* `dflash_proposer.py:116` is in `AscendDflashProposer.set_inputs_first_pass`, which
  DSpark overrides.
* `llm_base_proposer.py:326` (SlidingWindowAdapter) runs inside `load_model`, i.e.
  before `initialize_attn_backend`, so it already holds the base value.
* `llm_base_proposer.py:1615` is inside `attn_update_stack_num_spec_norm`, gated on
  `not self.parallel_drafting`, and DSpark sets `parallel_drafting = True`.

DSV4 is unaffected either way: the non-GDN branch reads
`attn_group.kv_cache_spec.block_size` directly and never consults the attribute.

One thing worth flagging separately, not for this PR: `load_model` derives
`self.kernel_block_size` from `get_supported_kernel_block_sizes()[0]`, which is the
weakest of the three derivations in the repo — `sfa_v1.py:236` uses
`select_common_block_size(...)` and `step3p5.py:780` takes the runner-provided
per-group `kernel_block_sizes`. `[0]` is fragile because the list can hold several
entries (DSA returns `[2, 4, 8, 16, 32, 64, 128]`) and, per `msa_m3.py`, entries may
be `MultipleOf` rather than `int`. DSpark's `initialize_attn_backend` already accepts
a `kernel_block_sizes` argument from the runner but ignores it, so switching to the
step3p5 pattern is possible — happy to do it here or in a follow-up, your call.

drslark (changes requested):
LGTM.
Thanks for your efforts.

fg11991:
> LGTM. Thanks for your efforts.

Thanks!

Tflowers-0129 (approved):

fg11991:
/rerun
[Bot]: rerun completed.

**Failed:**

fg11991:
CI status: `ci-gate` is green. The one remaining red is
`test_minimax_m3_gsm8k_one_case` on the a3-8 card shard, which I believe is
unrelated to this PR:

- The **same commit passes** on the `v0.26.0` job and fails only on the vllm-main
  job — identical code, only the vLLM version differs.
- The asserted value is not stable across runs (`actual='130'`, then `actual='3'`
  for the same greedy prompt), which points at nondeterminism rather than a logic
  error.
- That test constructs `LLM(...)` without a `speculative_config`, so neither
  changed path is reachable: `AscendDSparkProposer` is only instantiated for
  `method == "dspark"`, and `postprocess_mamba_fused_kernel` exists to reconcile
  mamba state after draft verification.
- With the three new constexpr flags all false (what the V1 launcher passes), the
  kernel body is byte-identical to before.

All review comments so far are addressed. Anything else you'd like changed?

fg11991:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

fg11991:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

fg11991:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

fg11991:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

fg11991:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

fg11991:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

PR #13391, opened 2026-08-03
[BugFix][MiniMax M3] Fix repeated-output drift in sparse attention

yejj710:

This PR fixes repeated-output drift in the MiniMax M3 Triton sparse attention implementation.

- Masks K/V loads for invalid pages and partially valid tail blocks in both prefill and decode kernels, preventing invalid cache data from affecting attention results.
- Limits the decode index block count to the actual block table width.
- Preserves the existing tensor-parallel block sharding constraints when calculating the effective block count.

Yes. MiniMax M3 sparse attention produces stable and repeatable outputs for requests containing tail blocks or invalid sparse-attention selections.

There are no API, configuration, or interface changes.

Not involved
- vLLM version: v0.26.0


linfeng-yuan:
1) PTAL @ZT-AIA 
2) Please verify this PR on A2 and A3. @yejj710

PR #12835, opened 2026-07-25
[Feature] layer and model runner support MiniMax M3

CXY-Katrina:
Migrate selected runtime changes from vllm-project/vllm-ascend#12734 into the synced main branch, covering Ascend config, ops, quantization, and model runner integration.

### How was this patch tested?

- vLLM version: v0.25.1
- vLLM main: https://github.com/vllm-project/vllm/commit/fe784ff22e630a31fd798f392b01e0a75c18f047

PR #13412, opened 2026-08-03
[BugFix][FusedMoE] Restore BF16 quant method initialization

weijinqian0:

PR #13268 moved Ascend quant-method selection into `AscendRoutedExperts._get_quant_method`. For unquantized models, this changes the method used during the upstream `RoutedExperts.__init__` lifecycle, including size roundup and expert-weight creation. Before the refactor, upstream created the BF16 weights first and the runner installed `AscendUnquantizedFusedMoEMethod` afterwards.

This change restores that ordering when `quant_config is None`:

- use the upstream unquantized method while `RoutedExperts` creates weights;
- install `AscendUnquantizedFusedMoEMethod` after the parent constructor completes;
- leave quantized method selection unchanged.

The issue was observed as a MiniMax M3 BF16 accuracy regression after #13268. The W8A8 configuration remains correct, and reverting #13268 restores BF16 accuracy. MiniMax M3 support was introduced in #12835.

Regression tests cover both the delayed BF16 replacement and preservation of the quantized path.

Yes. It restores the pre-refactor BF16 fused-MoE weight initialization behavior and is intended to fix the MiniMax M3 BF16 accuracy regression. There is no API or configuration change.

- `python -m py_compile vllm_ascend/ops/fused_moe/routed_experts.py tests/ut/ops/test_fused_moe.py`
- `python -m pytest -q tests/ut/ops/test_fused_moe.py tests/ut/_310p/fused_moe/test_shared_fused_moe_310.py` — 32 passed
- `git diff --check`
- `pre-commit run ruff-check --all-files --hook-stage manual`
- `pre-commit run ruff-format --all-files --hook-stage manual`

MiniMax M3 BF16 real-weight accuracy validation requires an Ascend NPU environment and is pending. The local full `format.sh ci` run passed Ruff, codespell, typos, clang-format, markdownlint, and the remaining static hooks; only the offline gitleaks hook could not run because this environment has neither `gitleaks` nor `wget`.

- vLLM version: v0.26.0


czydyy:
/nightly MiniMax-M3-BF16-A3
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/30815307515)

weijinqian0:
The current fixed path is as follows:
1. AscendRoutedExperts.__init__ first calls the upstream RoutedExperts.__init__.
2. In the BF16/non-quantization scenario, _get_quant_method() now returns the upstream UnquantizedFusedMoEMethod, allowing the upstream to complete maybe_roundup_sizes() and create_weights().
3. After the parent class construction is completed, self.quant_method is replaced with AscendUnquantizedFusedMoEMethod.
4. After the weights are loaded, the loader traverses the current module.quant_method. Therefore, the Ascend method's process_weights_after_loading() is called to perform transpose and NZ/layout conversion on w13_weight/w2_weight.
The corresponding code points are as follows:
- Ascend BF16 method is replaced after initialization: vllm_ascend/ops/fused_moe/routed_experts.py:268
- The upstream method is returned during the BF16 initialization phase: vllm_ascend/ops/fused_moe/routed_experts.py:384
- Ascend layout conversion after loading: vllm_ascend/ops/fused_moe/routed_experts.py:74
- The loader calls the post-loading processing of the current quant method: vllm_ascend/patch/worker/patch_process_weights_after_loading.py:27
Therefore, the more accurate causal chain is as follows:
After PR #13268, AscendUnquantizedFusedMoEMethod is used during the construction of the parent class RoutedExperts for BF16. The current change restores the original behavior, where the upstream method creates weights, and the Ascend method only takes over the execution and performs layout conversion after loading. The fact that your true weight test can be restored indicates that the deviation of MiniMax M3 BF16
comes from this initialization lifecycle change.

kunpengW-code (approved):

zzzzwwjj (approved):

PR #13414, opened 2026-08-03
[BugFix][MoE] Fix W8A8 scale handling for dynamic EPLB

muziyuhui666:

This PR fixes W8A8 MoE scale handling for the `swigluoai_uninterleave` path when dynamic EPLB provides list-form local physical expert weights and scales.

With dynamic EPLB enabled, a MoE layer may receive one weight tensor and one scale tensor per local physical expert. However, some MoE MLP paths previously passed all local expert weights while only using the first `w1_scale`. This made the weight and scale inputs inconsistent and could cause NPU operator shape errors.

This PR updates the scale preparation logic for the `swigluoai_uninterleave` path to:

* pass all local physical expert scales to grouped matmul in the All-to-All path
* stack multi-expert scales into the shape required by `npu_dequant_swiglu_quant` in the MC2 path
* convert the prepared scales to the dtype required by the corresponding NPU operator
* preserve the original tensor shape for single-element scale lists, avoiding regressions in the non-EPLB packed scale path
* keep the existing behavior unchanged for other activation paths

The issue was found while validating MiniMax M3 W8A8 with dynamic EPLB.

No.

This PR does not change public APIs, CLI arguments, configuration fields, routing policy, EPLB policy, or request behavior.

It only fixes the internal W8A8 MoE scale preparation for the `swigluoai_uninterleave` path and preserves the existing behavior of other activation paths and the non-EPLB packed scale path.

Targeted unit tests were added for:

* All-to-All with dynamic EPLB: verify that all local expert scales are passed to grouped matmul
* MC2 with dynamic EPLB: verify that multiple local expert scales are stacked into the expected shape and converted to `float32`
* MC2 without dynamic EPLB: verify that a single-element packed scale list preserves its original tensor shape

Related unit tests:

```bash
pytest -sv \
  tests/ut/ops/test_moe_mlp.py \
  tests/ut/ops/test_moe_runtime_args.py \
  tests/ut/ops/test_fused_moe.py \
  tests/ut/ops/test_moe_comm_method.py \
  tests/ut/quantization/methods/test_moe_logical_experts.py \
  tests/ut/quantization/methods/a2/test_w8a8_dynamic.py \
  tests/ut/eplb
```

Result:

```text
95 passed, 16 warnings in 15.13s
```

Real model validation on an 8-card Ascend NPU environment:

* MiniMax M3 W8A8 with dynamic EPLB enabled:

  * service startup passed
  * text request passed
  * image request passed
  * video request passed
  * EPLB expert hotness collection and expert weight update logs were observed

* MiniMax M3 W8A8 without EPLB:

  * service startup passed
  * text request passed
  * image request passed
  * video request passed
  * verified that the non-EPLB packed scale path was not regressed

* vLLM version: v0.26.0

* vLLM main: https://github.com/vllm-project/vllm/commit/0351e9aa1fdf1a51329d1906881528dfe61fc88e

### DIFF
--- a/vllm_ascend/ops/fused_moe/moe_mlp.py
+++ b/vllm_ascend/ops/fused_moe/moe_mlp.py
@@ -96,6 +96,68 @@ def _require_single_tensor_for_swiglu_quant(
     return tensor_or_list
 
 
+def _as_tensor_list(tensor_or_list: list[torch.Tensor] | torch.Tensor) -> list[torch.Tensor]:
+    return tensor_or_list if isinstance(tensor_or_list, list) else [tensor_or_list]
+
+
+def _validate_expert_tensor_lists(*, weights: list[torch.Tensor], scales: list[torch.Tensor], op_name: str) -> None:
+    if not weights:
+        raise ValueError(f"{op_name}: expert weight list must not be empty.")
+    if not scales:
+        raise ValueError(f"{op_name}: expert scale list must not be empty.")
+    if len(weights) != len(scales):
+        raise ValueError(
+            f"{op_name}: expected one scale per physical expert, "
+            f"but got weights={len(weights)} and scales={len(scales)}."
+        )
+
+
+def _prepare_dequant_swiglu_weight_scale(
+    w1: list[torch.Tensor] | torch.Tensor, w1_scale: list[torch.Tensor] | torch.Tensor
+) -> torch.Tensor:
+    if not isinstance(w1, list) and not isinstance(w1_scale, list):
+        weight_scale = w1_scale
+    else:
+        weights = _as_tensor_list(w1)
+        scales = _as_tensor_list(w1_scale)
+        _validate_expert_tensor_lists(weights=weights, scales=scales, op_name="npu_dequant_swiglu_quant")
+
+        if len(scales) == 1:
+            weight_scale = scales[0]
+        else:
+            scale_shape = scales[0].shape
+            for index, scale in enumerate(scales):
+                if scale.dim() not in (1, 2):
+                    raise ValueError(
+                        "npu_dequant_swiglu_quant: physical expert scales must be 1D or 2D, "
+                        f"but scale[{index}] has shape {tuple(scale.shape)}."
+                    )
+                if scale.shape != scale_shape:
+                    raise ValueError(
+                        "npu_dequant_swiglu_quant: all physical expert scales must have the same shape, "
+                        f"but scale[0] has shape {tuple(scale_shape)} "
+                        f"and scale[{index}] has shape {tuple(scale.shape)}."
+                    )
+
+            weight_scale = torch.stack([scale.reshape(-1) for scale in scales], dim=0)
+    if weight_scale.dtype != torch.float32:
+        weight_scale = weight_scale.to(torch.float32)
+    if weight_scale.dim() == 1:
+        weight_scale = weight_scale.reshape(1, -1)
+    return weight_scale
+
+
+def _prepare_grouped_matmul_scales(
+    weight: list[torch.Tensor] | torch.Tensor,
+    weight_scale: list[torch.Tensor] | torch.Tensor,
+    output_dtype: torch.dtype,
+) -> list[torch.Tensor]:
+    weights = _as_tensor_list(weight)
+    scales = _as_tensor_list(weight_scale)
+    _validate_expert_tensor_lists(weights=weights, scales=scales, op_name="npu_grouped_matmul")
+    return [scale.to(output_dtype) if scale.dtype != output_dtype else scale for scale in scales]
+
+
 def quant_apply_mlp(
     hidden_states: torch.Tensor,
     w1: list[torch.Tensor] | torch.Tensor,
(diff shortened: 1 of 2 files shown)


vllm_ascend/ops/fused_moe/moe_mlp.py
```diff
@@ -96,6 +96,68 @@ def _require_single_tensor_for_swiglu_quant(
...
+        )
+
+
+def _prepare_dequant_swiglu_weight_scale(
+    w1: list[torch.Tensor] | torch.Tensor, w1_scale: list[torch.Tensor] | torch.Tensor
+) -> torch.Tensor:
+    if not isinstance(w1, list) and not isinstance(w1_scale, list):
+        weight_scale = w1_scale
+    else:
+        weights = _as_tensor_list(w1)
```
Angazenn: Here you try to convert tensors to lists at first. But later you convert them back to tensors. Can we make this process simpler?

vllm_ascend/ops/fused_moe/moe_mlp.py
```diff
@@ -96,6 +96,68 @@ def _require_single_tensor_for_swiglu_quant(
     return tensor_or_list
 
 
+def _as_tensor_list(tensor_or_list: list[torch.Tensor] | torch.Tensor) -> list[torch.Tensor]:
+    return tensor_or_list if isinstance(tensor_or_list, list) else [tensor_or_list]
+
+
+def _validate_expert_tensor_lists(*, weights: list[torch.Tensor], scales: list[torch.Tensor], op_name: str) -> None:
```
Angazenn: I think there is no need to add this helper function.

vllm_ascend/ops/fused_moe/moe_mlp.py
```diff
@@ -96,6 +96,68 @@ def _require_single_tensor_for_swiglu_quant(
...
+    else:
+        weights = _as_tensor_list(w1)
+        scales = _as_tensor_list(w1_scale)
+        _validate_expert_tensor_lists(weights=weights, scales=scales, op_name="npu_dequant_swiglu_quant")
+
+        if len(scales) == 1:
+            weight_scale = scales[0]
+        else:
+            scale_shape = scales[0].shape
+            for index, scale in enumerate(scales):
```
Angazenn: These checks also seem to be redundant.

vllm_ascend/ops/fused_moe/moe_mlp.py
muziyuhui666: Thanks for the review. I agree this can be simpler.

The main issue we need to fix is that dynamic EPLB provides list-form local physical expert scales, but some paths only used `w1_scale[0]`. The extra helper and explicit validations are not strictly required for this bugfix.

I will simplify the patch by:
- removing `_as_tensor_list`
- removing `_validate_expert_tensor_lists`
- keeping only the scale preparation needed by grouped matmul and dequant swiglu
- preserving the single-element list scale shape for the non-EPLB packed path
- adjusting the UTs to focus on the actual behavior fixed by this PR

Angazenn:
Also make sure to only change branches that minimax really runs into.

muziyuhui666:
> Also make sure to only change branches that minimax really runs into.

Got it. I will narrow this PR to only the path confirmed by MiniMax M3 validation.

Specifically, I will keep the fix for the W8A8 + dynamic EPLB + `swigluoai_uninterleave` path, and remove the preventive changes/tests for adjacent branches such as `SWIGLUSTEP` that are not confirmed to be used by MiniMax M3.

muziyuhui666:
Updated. I narrowed the PR to the MiniMax M3 confirmed `swigluoai_uninterleave` path only.

The latest patch removes `_as_tensor_list`, `_validate_expert_tensor_lists`, and the extra defensive checks/tests. It also restores the unrelated `SWIGLUSTEP` path and removes the preventive `SWIGLUSTEP` coverage.

The remaining changes are limited to the two MiniMax M3 W8A8 dynamic EPLB scale handling cases we validated: passing all local physical expert scales to GMM1, and preparing `weight_scale` for `npu_dequant_swiglu_quant`.

vllm_ascend/ops/fused_moe/moe_mlp.py
```diff
@@ -226,7 +248,7 @@ def quant_apply_mlp(
                 hidden_states, act_quant_type=act_quant_type, use_mxfp_quant=use_mxfp_quant
             )
         else:
-            if w1_scale[0].dtype != torch.float32:
+            if not is_swigluoai_uninterleave and w1_scale[0].dtype != torch.float32:
```
Angazenn: move it to _prepare_dequant_swiglu_weight_scale

muziyuhui666: Sounds good. I will move this dtype preparation into `_prepare_dequant_swiglu_weight_scale`.

I will keep the non-`swigluoai_uninterleave` behavior unchanged: it still uses `w1_scale[0]` and only casts it to `float32` when needed. The list stacking / 1D reshape logic will remain limited to the `swigluoai_uninterleave` path.

muziyuhui666: Done. I moved the dtype preparation into `_prepare_dequant_swiglu_weight_scale`.

The non-`swigluoai_uninterleave` behavior is unchanged: it still uses `w1_scale[0]` and only casts it to `float32` when needed.

Angazenn (approved):

CXY-Katrina:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

weijinqian0 (approved):

PR #13421, opened 2026-08-03
[Misc][v0.23.0] Upgrade CANN, PTA, and Triton

MrZ20:

This PR upgrades the dependency stack used by the `releases/v0.23.0` branch:

- Upgrade CANN from 9.0.1 to 9.1.0.
  - Replace the 310P 9.1.0 beta images with the 9.1.0 final images.
  - Synchronize A2, A3, A5, and 310P images across Ubuntu, openEuler, wheel builds, CI, nightly tests, and CSRC cache jobs.
- Upgrade PTA (`torch-npu`) from 2.10.0.post2 to 2.10.0.post4.
- Upgrade Triton Ascend from 3.2.1 to 3.2.2.
- Update dependency metadata, installation documentation, Dockerfiles, CI image tags, and CANN package requirements accordingly.
- Fix the A5 wheel and CSRC cache configurations to use `ascend950dt_9582`, which is the SOC name recognized by the current build logic.
- Keep UVA disabled for Triton Ascend 3.2.2 because it is still affected by the pinned-memory pointer validation issue tracked in [triton-ascend#783](https://github.com/triton-lang/triton-ascend/issues/783).
- Temporarily skip three FlashComm1 E2E cases that are not yet compatible with the upgraded dependency stack. These cases require a follow-up fix and should not block the dependency upgrade.

The platform configuration propagation change in this PR is ported from [#13799](https://github.com/vllm-project/vllm-ascend/pull/13799), originally authored by @yiz-liu.

This change is needed because `NPUPlatform` may override `enable_npugraph_ex`, `enable_static_kernel`, and `recompute_scheduler_enable` according to the selected CUDAGraph mode or P/D role. Spawned EngineCore and worker processes reconstruct their Ascend configuration from `vllm_config.additional_config`; without propagating the effective values back to `additional_config`, those processes may restore stale defaults and enter an unsupported execution path. The fix from #13799 is therefore included in the v0.23.0 upgrade branch to keep the effective configuration consistent across process boundaries.

- vLLM version: v0.23.0


.github/workflows/dockerfiles/Dockerfile.buildwheel.a5
```diff
@@ -15,9 +15,9 @@
 # This file is a part of the vllm-ascend project.
 #
 ARG PY_VERSION=3.12 
-FROM quay.io/ascend/manylinux:9.0.1-950-manylinux_2_34-py${PY_VERSION}
+FROM quay.io/ascend/manylinux:9.1.0-950-manylinux_2_34-py${PY_VERSION}
 
-ARG SOC_VERSION="ascenda5_9582"
+ARG SOC_VERSION="ascend950dt_9582"
```
MrZ20, line 20: It is recommended to unify the three places to `ascend950dt_9582` as used in the existing Dockerfile/documents. cc @xqchen7 
The reasons are as follows:
- `ascenda5_9582` is used in two places: `Dockerfile.buildwheel.a5` (line 20) and `CSRC` matrix.
- The current `setup.py` (line 181) only recognizes values starting with `ascend950` as A5; otherwise, Undefined `soc_version` will be triggered.

MrZ20:
/rerun
[Bot]: rerun completed.

**Rerun:**
- Doc Test

yiz-liu (approved):

PR #13473, opened 2026-08-04
[Doc] Fix MiniMax-M2 launch_online_dp example in PD tutorial

Agoni-02:
Update the `launch_online_dp.py` example in the MiniMax-M2 multi-node PD separation tutorial (`docs/source/tutorials/models/MiniMax-M2.md`).

Changes:
- Add help text for CLI arguments to make the script easier to use.
- Start all local DP worker processes first, then `join()`, so multiple local DP ranks can run in parallel instead of sequentially.
- Check that `run_dp_template.sh` exists before launching processes, to fail fast with a clear error.

Why:
The previous example joined each process immediately after `start()`, which serializes local DP launches and does not match the intended multi-process launch behavior for multi-node PD deployment.

No. Documentation-only update.

- Reviewed the updated code block locally in `MiniMax-M2.md`.
- Ran pre-commit checks on the modified file (`codespell`, `typos` passed).
- No unit/e2e tests added because this PR only changes tutorial documentation.

- vLLM version: v0.26.0

### DIFF
--- a/docs/source/tutorials/models/MiniMax-M2.md
+++ b/docs/source/tutorials/models/MiniMax-M2.md
@@ -266,45 +266,7 @@ PD (Prefill-Decode) separation splits the Prefill and Decode phases across diffe
 
 First, prepare `launch_online_dp.py` on each node:
 
-```python
-import argparse
-import multiprocessing
-import os
-import subprocess
-import sys
-
-def parse_args():
-    parser = argparse.ArgumentParser()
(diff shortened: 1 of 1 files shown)


docs/source/tutorials/models/MiniMax-M2.md
voidvelocity: Use hyperlink to show the path of `launch_online_dp.py` instead of just copying code.

liqian79 (approved):

PR #13490, opened 2026-08-04
[Feature][LoRA] Isolate base and adapter ACL compile graphs

Liuchenbing-2026:
## What this PR does

Establishes independent ACL compile graphs for Base (multi-token), Base single-token, and LoRA requests, isolating graph events, workspace, attention parameters, and `BatchDescriptor` so that LoRA requests do not pollute the Base graph or force Base to fall back to eager.

- LoRA requests use a `/backbone` callable and an independent LoRA ACL Graph.
- Base multi-token requests use a `/base` callable; Base single-token uses `/base_one`.
- Base/LoRA use independent code objects, backends, and compile caches.
- Keeps speculative draft models out of dual-graph specialization.

## Scope

Based on `main`. Does not include the A3 single-adapter performance path (PR 3) or A2 enablement (PR 4).

## Validation

```
Ruff check/format: Pass
mypy Python 3.10/3.11/3.12: Pass
Focused pytest: 13 passed
git diff --check: Pass
```

Latest-head on-device matrix: pending.

## Related

- Stacked: PR 3 (A3 perf) → PR 4 (A2) depend on this PR.

### DIFF
--- a/tests/ut/compilation/test_lora_graph_params.py
+++ b/tests/ut/compilation/test_lora_graph_params.py
@@ -1,0 +1,58 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+from dataclasses import dataclass
+from types import SimpleNamespace
+from unittest.mock import patch
+
+from vllm.config import CUDAGraphMode
+
+from vllm_ascend.compilation.acl_graph import _GraphParamStore
+
+
+@dataclass(frozen=True)
+class _Descriptor:
+    num_tokens: int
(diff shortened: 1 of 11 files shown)


voidvelocity (changes requested):
Instead using `batch_descriptor: BatchDescriptor` as key, I think use `num_tokens` and `has_lora`  in `batch_descriptor` as key is enough for LoRA. Otherwise, it may capture too many graphs and use too much memory.

```python
class BatchDescriptor:
    num_tokens: int
    num_reqs: int | None = None
    uniform: bool = False
    has_lora: bool = False
    num_active_loras: int = 0
```

PR #13493, opened 2026-08-04
[Feature][LoRA][MoE] Support LoRA on models with shared experts

Liuchenbing-2026:
## What this PR does

Adds LoRA support for MoE models with shared experts (e.g. Qwen3.5-MoE, DeepSeek-V3). The shared-experts LoRA runs through vLLM's dense wrappers, whose expand-slice path does not match the `_C_ascend.sgmv_expand` stacked `lora_b` layout (and vLLM's `torch_ops` einsum fallback fails on the same layout).

This PR:
- Selects the shared-expert expand path automatically based on model type.
- Routes `PunicaWrapperNPU._expand_slice_prefill` / `_expand_slice_decode` through a `torch.bmm` fallback (per-token gather + bmm via a mutating custom op opaque to Dynamo fullgraph), bypassing `sgmv`/`bgmv_expand_slice`.
- Scopes the compatible expand-slice path to shared-experts models so models without shared experts (e.g. Qwen3-30B-A3B) keep the fused `_C_ascend` path and are unaffected.

## Scope

Based on `main`. Independent of the A3/A2 performance stack (PR 3 / PR 4).

## Validation

```
Ruff check/format: Pass
mypy Python 3.10/3.11/3.12: Pass
Focused pytest: 17 passed
git diff --check: Pass
```

- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/lora/test_lora.py
+++ b/tests/ut/lora/test_lora.py
@@ -10,7 +10,8 @@ from vllm_ascend.lora.fused_moe import (
     AscendFusedMoE3DWithLoRA,
     AscendFusedMoEWithLoRA,
 )
-from vllm_ascend.lora.punica_npu import PunicaWrapperNPU, _lora_bmm_expand_slice_op
+from vllm_ascend.lora.lora_ops import bmm_expand_slice
+from vllm_ascend.lora.punica_npu import PunicaWrapperNPU
 
 
 def test_ascend_fused_moe_3d_initializes_upstream_weight_state() -> None:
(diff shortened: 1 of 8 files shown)


paulyu12:
Plz add an e2e test. You can refer to the upstream vLLM testcase [tests/lora/test_qwen36_moe_lora.py](https://github.com/vllm-project/vllm/blob/main/tests/lora/test_qwen36_moe_lora.py)

paulyu12:
Maybe we can reuse the current sgmv_expand method. It already supports shared experts lora. You can refer to this PR: https://github.com/vllm-project/vllm-ascend/pull/13365. 

I close this PR first. If you have any problem, reopen it again.

PR #13498, opened 2026-08-04
[Bugfix][MRV1] Fix KV cache buffer corruption for extracting hidden states

PHOEBEMOON0802:
When a KV cache tensor's shared_by list contains both MambaSpec and HiddenStateCacheSpec layers, sharing one physical buffer causes float32 ssm_state writes to overwrite bfloat16 hidden-states data (same bytes, different interpretation). Allocate a separate tensor for the HiddenStateCacheSpec layers in that case.

Also:
- Route AscendExtractHiddenStatesProposer to cm_base when selecting spec_decode_common_attn_metadata so it uses the base KV cache group.
- Validate that extract_hidden_states drafter layers live in a single KV cache group during initialize_kv_cache.

Fix KV cache buffer corruption for hybrid model configurations that use both Mamba layers and extract-hidden-states layers in the same KV cache tensor group. Also wire AscendExtractHiddenStatesProposer into the spec-decode attention metadata plumbing.

- The added has_mamba / has_hidden scan runs only inside the existing hybrid/hidden-state branch, so the hot path for plain attention layers is untouched.
- Alignment handling (_align_memory with kv_transfer_config) is duplicated for the new tensor_hs; keeping the two allocations symmetric was intentional so KV-transfer setups behave identically for both buffers.
- No new public API. Behavior for non-hybrid configs is byte-for-byte identical.

No.

Signed-off-by: chenyue1122 oyoy7102@163.com

- vLLM version: v0.26.0

### DIFF
--- a/tests/e2e/pull_request/one_card/spec_decode/test_extract_hidden_states.py
+++ b/tests/e2e/pull_request/one_card/spec_decode/test_extract_hidden_states.py
@@ -129,6 +129,16 @@ def _verify_output(output, expected_shape, *, verify_nonzero, verify_token_ids):
         hidden_states = obj["hidden_states"]
         assert hidden_states.shape == expected_shape
 
+        assert not torch.isnan(hidden_states).any(), (
+            "hidden_states contains NaN — likely regression of PR #13498 "
+            "(KV cache buffer corruption between Mamba ssm_state and "
+            "HiddenStateCacheSpec)."
+        )
+        assert not torch.isinf(hidden_states).any(), (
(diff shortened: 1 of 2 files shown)


realliujiaxu:
LGTM. Upstream vLLM doesn't have this problem because its KV cache config builder isolates incompatible spec types into separate KVCacheTensor entries before allocation ever  runs, while the Ascend fork's rewritten _allocate_kv_cache_tensors lumps all non-attention layer types into a single shared buffer, breaking that isolation.

realliujiaxu (approved):

PR #12814, opened 2026-07-24
[Performance] Optimize AscendStore key construction and miss path

zmc1997:
### What this PR does / why we need it?

This supersedes #12136 and adapts the AscendStore key-construction and KVPool miss-path optimizations to `main` with a smaller, unconditional implementation.

- Generates serialized non-layerwise keys directly, caches immutable prefixes, and computes grouped hashes lazily.
- Centralizes token/chunk iteration and removes compatibility wrappers and duplicated key-building paths.
- Propagates HBM hit tokens through lookup, skips external lookup on a full HBM hit, and avoids querying or storing the HBM-resident prefix again.
- Applies hybrid-cache reachable, store, and load masks before key construction.
- Applies store masks before key construction and pre-shards filtered keys automatically when applicable.
- Uses asynchronous save by default and removes the global/per-request wait modes and their request-event lifecycle.
- Removes the environment-variable switches for these optimized paths; no `LOOKUP_FULL_GUARD` behavior is introduced.
- Keeps the layerwise `PoolKey` transfer path unchanged.

### Does this PR introduce _any_ user-facing change?

Yes. AscendStore now selects the optimized key construction, mask handling, pre-sharding, and asynchronous save paths automatically instead of exposing environment-variable switches. The cache key format remains unchanged. Because save is asynchronous, a following identical lookup may race with save visibility.

### How was this patch tested?

- 152 targeted AscendStore/Mooncake unit tests passed; 10 existing deprecated layerwise tests were skipped.
- Repository pre-commit hooks passed, including ruff, codespell, typos, custom Python checks, and DCO sign-off validation.
- `git diff --check` passed.
- NPU hardware testing was not available locally; CI/NPU validation is required.

- vLLM version: v0.25.1
- vLLM main: https://github.com/vllm-project/vllm/commit/fe784ff22e630a31fd798f392b01e0a75c18f047

PR #13602, opened 2026-08-05
[BugFix] Fix AscendStore lookup TypeError at ZMQ boundary

jichaofdu:

`LookupKeyClient` serializes vLLM `BlockHash` values as hexadecimal strings for
the ZMQ lookup RPC. After msgpack decoding, `LookupKeyServer` forwarded those
strings directly to `KVPoolWorker.lookup_scheduler()`, although downstream
grouped-hash code expects bytes-like `BlockHash` elements.

In a v0.23.0rc1-based AscendStore deployment with the grouped-hash changes from
#12814 and #13110 backported, this produced:

```text
TypeError: sequence item 0: expected a bytes-like object, str found
```

The lookup exception handler then returned zero external-cache hits.

This PR fixes the type mismatch at the ZMQ decode boundary, without changing
the existing RPC protocol:

- 64-character hex strings are restored with `bytes.fromhex()`.
- Other strings are encoded as bytes.
- Values that are already bytes are passed through unchanged.

As a result, all internal lookup paths receive bytes-like block hashes while
the scheduler-side serialization and wire format remain unchanged.

No public API, configuration, RPC framing, or cache-key format changes. The fix
prevents affected AscendStore lookups from failing and degrading to zero cache
hits.

- Added a `LookupKeyServer` regression test that feeds a 64-character hex hash
  through the decoded ZMQ request and verifies that `lookup_scheduler()`
  receives the corresponding 32-byte `BlockHash`.
- Verified the regression test fails before the production change because the
  worker receives `str`, and passes after the boundary conversion is added.
- Ran 50 targeted AscendStore unit tests covering the lookup client, lookup
  server, connector, and byte/string key construction paths.
- Ran `ruff check`, `ruff format --check`, and `git diff --check` on the changed
  files.

No NPU hardware is required for this RPC type-conversion fix.

- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/distributed/ascend_store/test_ascend_store_connector.py
+++ b/tests/ut/distributed/ascend_store/test_ascend_store_connector.py
@@ -25,6 +25,7 @@ from vllm.distributed.kv_events import KVCacheEvent
 from vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.ascend_store_connector import (
     AscendStoreConnector,
     AscendStoreKVEvents,
+    LookupKeyServer,
 )
 
 # isort: on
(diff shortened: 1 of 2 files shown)


jichaofdu:
Updated the implementation to normalize decoded hex hashes at the LookupKeyServer boundary. The previous client-side serialization change has been fully removed; the existing RPC wire format is unchanged.

PR #13669, opened 2026-08-06
[BugFix] Support virtual requests in Mooncake connector

xyy59516:

This PR adds support for the optional `do_virtual` flag in
`MooncakeConnector`.

When `do_virtual` is enabled for a decode-side KV receive request, this
change:

- propagates `do_virtual` from `kv_transfer_params` to `ReqMeta`;
- marks the request as completed through
  `KVCacheTaskTracker.add_not_transfer_request`;
- skips submitting the request to the remaining Mooncake KV receive and
  transfer path.

This is needed for virtual execution flows where the decode-side request
should follow the normal KV transfer completion path without performing
an actual remote KV cache receive operation.

The default value of `do_virtual` is `False`, so the existing Mooncake
KV transfer behavior remains unchanged when the flag is not specified.

Yes.

This PR adds an optional `do_virtual` field to the per-request
`kv_transfer_params` handled by `MooncakeConnector`.

When `do_virtual` is set to `True`, the decode-side KV receive request is
marked as completed through `add_not_transfer_request`, and the actual
Mooncake KV receive path is skipped.

When the field is omitted or set to `False`, the existing behavior is
unchanged.

- Ran Ruff lint checks:

```bash
ruff check \
  vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
```

- Formatted the modified file with Ruff:

```bash
ruff format \
  vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
```

- Ran a Python syntax check:

```bash
python -m py_compile \
  vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
```

- Manually tested the change in an Ascend 1P1D deployment with:
  - `MooncakeConnectorV1`
  - Prefill TP size: 2
  - Decode TP size: 2
  - Model: `Qwen3-30B-A3B-W8A8`

- Sent a decode-side request with `do_virtual=true` and dummy remote KV
  transfer metadata.

- Temporarily added a log in the `do_virtual` branch and verified that:
  - the request entered the virtual completion branch;
  - `add_not_transfer_request` was executed;
  - the subsequent Mooncake KV transfer scheduling path was skipped;
  - the decode server completed the request and returned a response
    successfully without performing an actual remote KV cache transfer.

The temporary validation log was removed from the final patch.

No automated test was added. The change was validated through the manual
Ascend 1P1D smoke test described above.

- vLLM version: v0.26.0


wxsIcey (approved):
LGTM!

wxsIcey (approved):
LGTM!

yiz-liu (approved):

PR #13672, opened 2026-08-06
[Refactor][device][1/N] Refactor Ascend hardware configuration and discovery

Tflowers-0129:

This PR introduces the first step of the Ascend hardware abstraction refactoring:

- Centralizes `AscendDeviceType` and SoC mappings in a lightweight module shared by runtime and build-time code.
- Adds a lazily initialized `DeviceConfig` as the unified device detection entry point.
- Removes duplicated device detection state from `utils.py`.
- Keeps `is_310p()`, `is_950()`, and other existing APIs as compatibility wrappers.
- Makes `setup.py` reuse the same SoC mapping instead of maintaining a separate mapping table.

This establishes a single source of truth for hardware identity and prepares for capability-oriented hardware profiles in subsequent PRs.

No intended user-facing behavior change.

Existing device detection APIs remain available and service startup behavior is unchanged. Device-type mismatch errors now use an explicit `RuntimeError` instead of an assertion.

- Added unit tests covering:
  - SoC-to-device mappings
  - build information detection
  - legacy `__soc_version__` compatibility
  - runtime device detection
  - unknown and mismatched device types
  - lazy initialization without import-time NPU probing

### Follow-up Plan

This PR only centralizes hardware identity and device detection. The remaining migration will be completed incrementally:

1. **Introduce hardware capabilities and profiles**
   - Define capability-oriented hardware profiles for supported Ascend device families.
   - Keep concrete device identities limited to detection and profile registration.

2. **Migrate platform and worker selection**
   - Move Worker, ModelRunner, and platform implementation selection behind hardware profiles.
   - Remove direct device checks from shared initialization logic.

3. **Migrate compilation and graph behavior**
   - Abstract device-specific graph backends, compilation options, fusion settings, and warm-up behavior.

4. **Migrate operators, quantization, and MoE**
   - Select weight layouts, attention implementations, quantization backends, MoE backends, and hardware-specific kernels through component providers or capabilities.

5. **Remove compatibility helpers**
   - Migrate existing call sites away from `is_310p()`, `is_950()`, and direct `AscendDeviceType` comparisons.
   - Deprecate or remove compatibility re-exports once no downstream users remain.
   - Add a static check to prevent new device-specific branches from being introduced into shared business logic.

The intended end state is that adding a new Ascend device primarily requires registering its SoC mapping and hardware profile, without modifying unrelated business modules.

- vLLM version: v0.26.0

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -24,7 +24,6 @@ import json
 import math
 import os
 from contextlib import nullcontext
-from enum import Enum
 from functools import lru_cache
 from typing import TYPE_CHECKING, Any
 
@@ -38,6 +37,13 @@ from vllm.sequence import IntermediateTensors
 
 import vllm_ascend.envs as envs_ascend
 from vllm_ascend.ascend_config import get_ascend_config
+from vllm_ascend.device.device_config import (  # noqa: F401
+    AscendDeviceType,
+    check_ascend_device_type,
+    get_ascend_device_type,
+    is_310p,
+    is_950,
+)
 
 if TYPE_CHECKING:
     from vllm.config import VllmConfig
@@ -137,10 +143,6 @@ def clear_enable_sp():
     _libc_getenv.cache_clear()
 
 
-def is_310p():
-    return get_ascend_device_type() == AscendDeviceType._310P
-
-
 _IS_RC_DEVICE: bool | None = None
 
 
@@ -165,10 +167,6 @@ def is_rc_device() -> bool:
     except (subprocess.CalledProcessError, FileNotFoundError):
         _IS_RC_DEVICE = False
     return _IS_RC_DEVICE
-
-
-def is_950():
-    return get_ascend_device_type() == AscendDeviceType.A5
 
 
 def _mark_op_side_effectful(op: Any) -> None:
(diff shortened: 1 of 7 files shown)


vllm_ascend/utils.py
```diff
@@ -38,6 +37,11 @@
 
 import vllm_ascend.envs as envs_ascend
 from vllm_ascend.ascend_config import get_ascend_config
+from vllm_ascend.device.device_config import (
+    AscendDeviceType,
+    get_ascend_device_type,
+    is_310p,
```
Alex-stack-hub, line 44: Could we also keep `is_950` re-exported from `utils` here? `vllm_ascend/attention/attention_v1.py` still imports it from `vllm_ascend.utils`

vllm_ascend/utils.py
```diff
@@ -38,6 +37,11 @@
...
+)
```
Alex-stack-hub, line 46: Could we keep `check_ascend_device_type` in this compatibility import as well? `vllm_ascend/worker/worker.py` still imports it from `utils`,

Alex-stack-hub (commented):
Thanks for centralizing the device detection logic. I noticed two compatibility exports that are still used by existing modules.

vllm_ascend/utils.py
Tflowers-0129: 当前 PR 保持兼容，调用方迁移留给后续 PR，不在 PR 1 扩大范围。

vllm_ascend/utils.py
Tflowers-0129: 当前 PR 保持兼容，调用方迁移留给后续 PR，不在 PR 1 扩大范围。

Tflowers-0129:
/rerun
[Bot]: rerun completed.

**Failed:**

Tflowers-0129:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

wangxiyuan (approved):

PR #13716, opened 2026-08-06
[Feature][Model] Use SparseFlashMla for DSV4 A5 BF16 KV

yjyang62:

Enable DeepSeek-V4 BF16 KV sparse attention on Ascend A5 (950) via SparseFlashMla.

- Route A5 BF16 DSA to the SparseFlashMla path via device adaptor (`device_op.py` + `ops/sparse_flash_mla.py`)
- Adapt DSA kwargs (`seqused_kv` → `seqused_ori_kv`, compressed KV lengths, etc.) to SparseFlashMla IR
- Fix A5 BF16 o_proj / weight layout (`npu_transpose_batchmatmul`, `wo_a` reshape to `[G,D,R]`)
- Keep A3 FP8 DSA path unchanged

Yes. DeepSeek-V4 on Ascend A5 can use BF16 KV sparse attention through the A5 SparseFlashMla path.

- Unit tests added/updated:
  - `tests/ut/device/test_device_op.py` (BF16 vs FP8 routing, kwargs adaptation)
  - `tests/ut/attention/test_dsa_v1.py` (BF16 projection / DSA path)
- Manual validation on Ascend A5 BF16 serve path
- Accuracy:
image

### DIFF
--- a/vllm_ascend/device/device_op.py
+++ b/vllm_ascend/device/device_op.py
@@ -31,9 +31,6 @@ from vllm_ascend.ops.triton.fused_gdn_gating import fused_gdn_gating_patch
 from vllm_ascend.quantization.quant_type import QuantType
 from vllm_ascend.quantization.utils import QUANT_DTYPES, SCALE_DTYPES, get_dynamic_mx_quant_scale_alg
 
-DSA_COMPRESSOR_SLOT_MAPPING_FLAT = 1
-DSA_COMPRESSOR_SLOT_MAPPING_BLOCK_OFFSET = 2
-
 if HAS_TRITON:
     from vllm_ascend.ops.triton.rms_norm import triton_q_rms  # noqa: F811
 else:
@@ -510,40 +507,6 @@ class BaseDeviceAdaptor:
 
         return context_layer
 
-    # ===== Sparse Attention Metadata & Op Selectors =====
-
-    @staticmethod
-    def get_dsa_sparse_attn_metadata_op():
-        """Returns the metadata-building operator for sparse attention."""
-        return torch.ops._C_ascend.npu_sparse_attn_sharedkv_metadata
-
-    @staticmethod
-    def get_dsa_sparse_attn_metadata_kwargs(device):
-        """Returns kwargs for sparse attention metadata builder."""
-        return {"device": str(device)}
-
-    @staticmethod
-    def get_dsa_sparse_attn_op():
-        """Returns the sparse attention operator."""
-        return torch.ops._C_ascend.npu_sparse_attn_sharedkv
-
-    @staticmethod
-    def get_dsa_sparse_attn_base_kwargs():
-        """Returns base kwargs for sparse attention (extended by caller)."""
-        return {}
-
-    @staticmethod
-    def get_dsa_compressor_slot_mapping_format():
-        """Slot mapping side output format consumed by the DSA scatter op."""
-        return DSA_COMPRESSOR_SLOT_MAPPING_BLOCK_OFFSET
-
-    # ===== SWA / Compressor KV Scatter =====
-
-    @staticmethod
-    def dsa_kv_compress_scatter(cache, x, slot_mapping):
-        """Scatter KV into cache. Non-A5: simple scatter of pre-quantized tensor."""
-        torch.ops._C_ascend.npu_scatter_nd_update_v2(cache, slot_mapping, x)
-
     # ===== Indexer Quant + Scatter =====
 
     @staticmethod
@@ -674,21 +637,10 @@ class BaseDeviceAdaptor:
         return slot_mapping
 
     @staticmethod
-    def format_dsa_slot_mapping(slot_mapping, block_size):
-        """Format slot_mapping for metadata storage.
-        Non-A5: 2D [block_idx, offset]; A5: 1D pass-through."""
-        return torch.stack([slot_mapping // block_size, slot_mapping % block_size], axis=-1)
-
-    @staticmethod
     def get_dsa_decode_cu_seqlens_cmp_kv(cmp_kv_tensor):
         """Non-A5: return the cached cu_seqlens_cmp_kv tensor.
         A5 override always returns None."""
         return cmp_kv_tensor
-
-    @staticmethod
-    def add_dsa_sparse_attn_extra_kwargs(extra_kwargs, **kwargs_to_add):
-        """Non-A5: add extra kwargs for sparse attention. A5: no-op."""
-        extra_kwargs.update(kwargs_to_add)
 
     @staticmethod
     def get_dsa_decode_cu_seqlens_ori_kv(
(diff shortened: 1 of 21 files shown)


linfeng-yuan (changes requested):
I think the feature direction is reasonable, but the current implementation needs a clearer boundary before we merge it.

The main issue is that “KV cache dtype” is being used as an implicit switch in too many places. DeepSeek-V4 has several different KV-like caches: attention/SWA KV, compressed KV, and indexer KV. These do not necessarily have the same dtype or layout. In this PR, the A5 BF16 SparseFlashMla path is selected by re-reading global `vllm_config.cache_config.cache_dtype` in the metadata builder / impl path. That is fragile, because the indexer cache is supposed to stay FP8 on A5 even when the attention KV is BF16.

I do not think we need a large DSA refactor in this PR, but we should at least make the BF16 path explicit and local.

Concretely, I would like to see:

1. Use `kv_cache_spec.dtype` / `kv_cache_spec.cache_dtype_str` as the source of truth inside DSA metadata builders, instead of re-resolving dtype from global `cache_config.cache_dtype`.

2. Distinguish attention KV dtype from indexer KV dtype in names and logic. For example, use `attn_kv_dtype` for the SparseFlashMla/DSA attention path, and keep `indexer_kv_dtype` explicitly FP8 on A5.

3. Centralize the A5 BF16 SparseFlashMla decision in one small helper or plan object, instead of scattering `kv_cache_dtype == torch.bfloat16` checks across model, metadata, and device adaptor code. That helper should own:
   - sparse attention op
   - metadata op
   - `layout_kv`
   - slot mapping format
   - base kwargs
   - whether block/offset slot mapping is required

4. Fix the BF16 scatter path to ignore invalid `-1` slot mappings instead of clamping them to zero. Clamping can overwrite valid cache entries and corrupt generation. (PS: Please make sure whether I make a mistake around slot_mapping here ) 

5. Add accuracy results of both `mxfp8` and `bf16` dtype with your PR. 

I am not asking for a global DSA rewrite in this PR. But after merge, the BF16 path should be clearly visible and easy to refactor later. Right now it looks like a functional patch layered on top of existing DSV4/A5 dtype behavior, and I worry it will make future cleanup harder.

linfeng-yuan:
8.17
1. Accuracy test with GPQA
2. blocksize mapping；
3. slot_mapping value clamping for -1, input shape 2dim?

vllm_ascend/device/device_op.py
```diff
@@ -662,29 +665,38 @@ def npu_flash_attention(query, key, value, seq_lens_cpu, head_num, scale_value,
...
-    def get_dsa_sparse_attn_metadata_op():
+    def build_dsa_attn_kv_plan(kv_cache_dtype: torch.dtype | None = None) -> DsaAttnKvPlan:
+        """A2/A3: sharedkv + PA_ND + block/offset slots (dtype does not switch path)."""
+        return build_base_dsa_attn_kv_plan(kv_cache_dtype)
+
+    @classmethod
+    def get_dsa_sparse_attn_metadata_op(cls, kv_cache_dtype: torch.dtype | None = None):
         """Returns the metadata-building operator for sparse attention."""
-        return torch.ops._C_ascend.npu_sparse_attn_sharedkv_metadata
+        return cls.build_dsa_attn_kv_plan(kv_cache_dtype).sparse_attn_metadata_op
```
lcfenglinwan: merge this logic to attn_kv_plan

yjyang62:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

yjyang62:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/models/layer/attention/layer.py
```diff
@@ -43,13 +44,28 @@ def get_dsv4_block_sizes():
...
         return _DSV4_BLOCK_SIZES_A5
     else:
         return _DSV4_BLOCK_SIZES
 
 
 DSV4_BLOCK_SIZES = get_dsv4_block_sizes()
+DSV4_BLOCK_SIZES_A5_BF16 = get_dsv4_block_sizes(use_a5_bf16_kv=True)
+
+
+def dsv4_block_sizes(vllm_config: VllmConfig):
```
Angazenn, line 65: I think these "double function wrapper" are strange. It can be simplified to a single `dsv4_block_sizes` function that use vllm_config as input argument and then dispatch by hardware && dtype. For example:
```suggestion
def dsv4_block_sizes(vllm_config: VllmConfig):
    # cache_config.block_size: [mla, swa, c4 state, c128 state], [page_size_padded_t1, page_size_padded_t2]
    _DSV4_BLOCK_SIZES = {
        128: [[128, 128, 8, 32], [16640, 131072]],
        64: [[64, 64, 4, 16], [8320, 65536]],
        32: [[32, 32, 2, 8], [4160, 32768]],
    }
    _DSV4_BLOCK_SIZES_A5 = {
        128: [[128, 128, 8, 16], [16896, 81920]],
        64: [[64, 64, 4, 8], [8448, 40960]],
        32: [[32, 32, 2, 4], [4224, 20480]],
    }
    _DSV4_BLOCK_SIZES_A5_BF16 = {
        128: [[128, 128, 8, 16], [16896, 131072]],
        64: [[64, 64, 4, 8], [8448, 65536]],
        32: [[32, 32, 2, 4], [4224, 32768]],
    }
    if get_ascend_device_type() in {AscendDeviceType.A5}:
        if uses_explicit_bf16_kv(vllm_config):
            return _DSV4_BLOCK_SIZES_A5_BF16
        else:
            return _DSV4_BLOCK_SIZES_A5
    else:
        return _DSV4_BLOCK_SIZES
```

Angazenn (changes requested):

vllm_ascend/models/deepseek_v4/model.py
```diff
@@ -587,7 +595,13 @@ def __init__(
...
 
         ascend_device_type = get_ascend_device_type()
-        k_dtype = torch.float8_e4m3fn if ascend_device_type == AscendDeviceType.A5 else torch.bfloat16
+        k_dtype = (
+            torch.bfloat16
+            if ascend_device_type == AscendDeviceType.A5 and uses_explicit_bf16_kv(vllm_config)
+            else torch.float8_e4m3fn
+            if ascend_device_type == AscendDeviceType.A5
+            else torch.bfloat16
+        )
```
lcfenglinwan: Here, using only `uses_explicit_bf16_kv` should be sufficient to distinguish the types.

vllm_ascend/models/deepseek_v4/model.py
```diff
@@ -108,10 +109,17 @@ def __init__(
...
+            self.dtype = torch.bfloat16
+        elif get_ascend_device_type() in {AscendDeviceType.A5}:
             self.dtype = torch.float8_e4m3fn
             vllm_config.cache_config.cache_dtype = "float8_e4m3fn"
-        cached_head_size = self.head_dim + 128 if get_ascend_device_type() in {AscendDeviceType.A5} else self.head_dim
+        cached_head_size = (
+            self.head_dim
+            if use_bf16_kv
+            else (self.head_dim + 128 if get_ascend_device_type() in {AscendDeviceType.A5} else self.head_dim)
+        )
```
lcfenglinwan: The judgment here is a bit too redundant. Please simplify it.

yjyang62:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

yjyang62:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/models/layer/attention/layer.py
```diff
@@ -178,14 +196,19 @@ def get_kv_cache_spec(self, vllm_config: VllmConfig) -> KVCacheSpec:
...
+            kv_cache_dtype = torch.bfloat16
+        elif get_ascend_device_type() in {AscendDeviceType.A5}:
             kv_cache_dtype = torch.float8_e4m3fn
             vllm_config.cache_config.cache_dtype = "float8_e4m3fn"
 
         cached_head_size = (
-            (self.head_size + 128) if get_ascend_device_type() in {AscendDeviceType.A5} else self.head_size
+            self.head_size
+            if use_bf16_kv
+            else ((self.head_size + 128) if get_ascend_device_type() in {AscendDeviceType.A5} else self.head_size)
```
lcfenglinwan: plz simplify here

lcfenglinwan (approved):
LGTM, and please submit another PR to address the issue with the slot mapping dimension.

Tflowers-0129 (approved):

linfeng-yuan (approved):

PR #13724, opened 2026-08-06
[Feature][Spec Decode] Add reduce_sample branch for dspark

drslark:

Add a dspark-specific branch under enable_reduce_sample that keeps the MarkovHead autoregressive correction loop but runs compute_logits and markov_bias as per-rank local vocab logits (AscendLogitsProcessor skips the all-gather) and uses greedy_sample to reduce the local top-1 across TP. markov_embed is unchanged (VocabParallelEmbedding already all-reduces). Guard the per-rank local add with an assert that lm_head and markov_w2 shard vocab identically.

The markov head runs in 4 steps:
* Compute original logits (no all-reduce).
* Compute logits_bias (with all-reduce).
* Compute final logits (no all-reduce).
* Sample tokens (reduce-sample).

N/A

- vLLM version: v0.26.0


zhenwenqi2024 (approved):

drslark:
Close this issue, as reduce_sample is about to be deprecated.

drslark:
Close this issue, as `reduce_sample` is about to be deprecated.

PR #13752, opened 2026-08-07
[BugFix][MRV2] Fix EAGLE3 startup crash and FlashComm1 FULL-graph capture under MRV2

xuchi-0808:
## Summary

Fixes the A2 MiniMax-M2.5-w8a8-QuaRot nightly startup crash (`202608051413.raw.log`: EAGLE3 draft forward fails during dummy profile when Model Runner V2 is used together with FlashComm1 and Eagle3 speculative decoding). Four independent defects on the same MRV2 + FlashComm1 path are fixed in this PR so the nightly configuration can start and complete requests. The Qwen3-30B-A3B-W8A8 setup below was used for hardware validation as an equivalent repro of the same path; fix 2 also covers the A3 `qwen3-30b-acc` TP4 MC2-init assert.

## Root causes and fixes

1. **Draft forwards wrongly enabled FlashComm1** (`ascend_forward_context.py`, `platform.py`, `register_custom_ops.py`, 3 speculator files)

   `NPUPlatform.set_additional_forward_context` hardcoded `is_draft_model=False` and decided FlashComm1 from the *target* model only. A dense drafter (Eagle/Eagle3) under a MoE target therefore ran with FlashComm1 enabled: `embed_tokens` did a TP reduce-scatter and returned `num_tokens / TP` rows, which no longer matched the full-sequence hidden states in `llama_eagle3`'s `torch.cat`, crashing during the dummy profile run (`Expected (s72//8) in dimension 0 but got s47`).

   Fix: mirror the V1 draft branch. A new `_MRV2_IS_DRAFT_FORWARD` ContextVar is set around the Ascend speculators' `_run_model` (Eagle/Eagle3/MTP via `AscendAutoRegressiveSpeculator`, DFlash via a new override, DSpark via MRO), the platform hook reads it, and FlashComm1 is disabled for dense drafters (MoE drafters keep it). `_maybe_pad_and_reduce_impl/fake` also guard the compile-pass SP backdoor so it cannot re-enable reduce-scatter for dense drafts.

2. **MRV2 runner never initialized `_potential_max_tokens`** (`worker/v2/model_runner.py`)

   `NPUModelRunner.__init__` (v2) missed `set_potential_max_tokens` that V1 calls. With EP>1 on the KV-consumer path, `TokenDispatcherWithMC2.__init__` → `get_potential_max_tokens()` hit `assert _potential_max_tokens is not None` and every worker crashed during model loading.

   Fix: call `set_potential_max_tokens(vllm_config)` next to the sibling `set_mc2_*` calls.

3. **FlashComm1 hidden-state all-gather used the unpadded token count** (`worker/v2/model_runner.py`)

   `_all_gather_hidden_states_and_aux` restored Eagle3 aux hidden states with `input_batch.num_tokens` (unpadded) while the speculator accesses the buffer with `num_tokens_after_padding`, causing an EZ1007 `12 vs 9` shape mismatch on the first request.

   Fix: use `num_tokens_after_padding`.

4. **FULL graph capture saw TP-local hidden states** (`worker/v2/aclgraph_utils.py`, `worker/v2/model_runner.py`)

   With FlashComm1 the target forward returns TP-local hidden states (`num_tokens / TP` rows), so upstream `ModelCudaGraphManager.capture` hit `self.hidden_states[:num_tokens] = hidden_states` shape mismatches for larger capture sizes (e.g. `100 vs 96`, desc 384 with TP4).

   Fix (vllm-ascend only, no upstream vLLM change): `ModelWithContext.forward` gathers the TP-local output back to the full sequence during FULL graph capture, so the all-gather is recorded into the graph and the capture buffers see full rows. `NPUModelRunner.execute_model` skips its outer all-gather when the rows are already full. PIECEWISE keeps the TP-local output and the outer all-gather path unchanged.

## Verification

- Hardware validation (2026-08-07, A3 90.90.97.37 container `xc_vllm_A00302_01`, Qwen3-30B-A3B-W8A8 + Eagle3 + `VLLM_ASCEND_ENABLE_FLASHCOMM1=1`, TP4/EP, FULL_DECODE_ONLY, MRV2, NPU 0-3):
  - 4 workers: `Graph capturing finished in 102 secs` — 42 FULL decode graphs + prefill/decode graphs all captured (the FULL-capture TP-local fix is effective);
  - `Application startup complete`, `GET /v1/models` → 200;
  - `POST /v1/chat/completions` → 200, 15 → 32 tokens, fingerprint `tp4-ep-...` — no EZ1007 (12 vs 9) and no `cat` shape error;
  - No traceback in the whole log.
  - The workspace was verified byte-by-byte against this PR head (`a427466a`): 5/8 files identical; the other 3 only differ by upstream base drift outside these hunks.
- Static checks: `py_compile` and `git diff --check` pass for all changed files.

## Related

- Closes the same issue class as #8569 (draft + FlashComm1 startup failure).

### DIFF
--- a/vllm_ascend/ascend_forward_context.py
+++ b/vllm_ascend/ascend_forward_context.py
@@ -31,6 +31,7 @@ class MoECommType(Enum):
 
 
 _MRV2_IN_PROFILE_RUN: ContextVar[bool] = ContextVar("_MRV2_IN_PROFILE_RUN", default=False)
+_MRV2_IS_DRAFT_FORWARD: ContextVar[bool] = ContextVar("_MRV2_IS_DRAFT_FORWARD", default=False)
 _CANN_MEGAMOE_SUPPORTED_QUANT_NAMES = {
     "w8a8",
     "w4a8",
@@ -64,6 +65,26 @@ def override_mrv2_in_profile_run(enabled: bool):
 
 def get_mrv2_in_profile_run() -> bool:
     return _MRV2_IN_PROFILE_RUN.get()
+
+
+@contextmanager
+def override_mrv2_is_draft_forward():
+    """Mark the enclosed forward path as a speculative-draft forward.
+
+    MRv2 builds the base forward context inside upstream vLLM, so Ascend's
+    platform hook cannot tell whether the current forward is the target or
+    the draft model. A ContextVar keeps this MRv2-only state scoped to the
+    draft forward without adding default fallback behavior.
+    """
+    token = _MRV2_IS_DRAFT_FORWARD.set(True)
+    try:
+        yield
+    finally:
+        _MRV2_IS_DRAFT_FORWARD.reset(token)
+
+
+def get_mrv2_is_draft_forward() -> bool:
+    return _MRV2_IS_DRAFT_FORWARD.get()
 
 
 def _cann_megamoe_supported_by_config(vllm_config: VllmConfig) -> bool:
--- a/vllm_ascend/worker/v2/aclgraph_utils.py
+++ b/vllm_ascend/worker/v2/aclgraph_utils.py
@@ -36,6 +36,7 @@ from vllm.v1.worker.utils import AttentionGroup
 
 from vllm_ascend.ascend_forward_context import _EXTRA_CTX
 from vllm_ascend.compilation.acl_graph import set_graph_params, update_full_graph_params
+from vllm_ascend.worker.v2.sp_utils import _all_gather_hidden_states_and_aux
 from vllm_ascend.worker.v2.utils import communicator_switch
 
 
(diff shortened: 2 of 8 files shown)


xuchi-0808:
## Hardware validation passed (2026-08-07, A3 90.90.97.37)

Validated the full PR on the nightly configuration: Qwen3-30B-A3B-W8A8 + Eagle3, TP4/EP, FULL_DECODE_ONLY, MRV2, `VLLM_ASCEND_ENABLE_FLASHCOMM1=1`, NPU 0-3, port 8099.

Key evidence (4 workers, log: `/tmp/pr13752_serve.log`):
- `Graph capturing finished in 102 secs` — 42 FULL decode graphs + prefill/decode graphs captured; the vllm-ascend-only FULL-capture TP-local fix (commit a427466a) is effective;
- `Application startup complete`; `GET /v1/models` → 200;
- `POST /v1/chat/completions` → 200, 15 → 32 tokens, fingerprint `vllm-0.26.1rc1.dev183+g0351e9aa1.d20260806-tp4-ep-a7290e11`;
- No `cat` shape error, no EZ1007 (12 vs 9), no traceback.

Workspace code was compared with PR head `a427466a`: 5/8 files byte-identical; the remaining 3 (`register_custom_ops.py`, `platform.py`, `model_runner.py`) differ only by upstream base drift outside the PR hunks.

xuchi-0808:
Clarification: the nightly case this PR fixes is **A2 MiniMax-M2.5-w8a8-QuaRot** (`202608051413.raw.log`, EAGLE3 draft + FlashComm1 startup crash). The Qwen3-30B-A3B-W8A8 validation above is an equivalent repro of the same MRV2 + FlashComm1 + Eagle3 path (matching `test_qwen3_30b_acc.py` TP4 parameters); fix 2 additionally covers the `qwen3-30b-acc` TP4 MC2-init assert.

vllm_ascend/worker/v2/aclgraph_utils.py
```diff
@@ -174,7 +175,29 @@ def forward(self, *args, **kwargs):
...
+        output = self.original_model(*args, **kwargs)
+
+        # Under MRV2 + FlashComm1 the target forward returns TP-local hidden
+        # states (num_tokens / TP rows). During FULL graph capture, gather them
+        # back to the full sequence inside the graph so the capture buffers see
+        # the same row count upstream vLLM expects; the runtime path then skips
+        # its outer all-gather when the rows are already full
+        # (NPUModelRunner.execute_model). PIECEWISE keeps the TP-local output
+        # and relies on the outer all-gather exactly as before.
+        try:
```
drslark, line 187: Please don't use try catch.

drslark (changes requested):

vllm_ascend/ascend_forward_context.py
```diff
@@ -66,6 +67,26 @@ def get_mrv2_in_profile_run() -> bool:
     return _MRV2_IN_PROFILE_RUN.get()
 
 
+@contextmanager
+def override_mrv2_is_draft_forward():
```
drslark, line 71: Change this function's name.
Please refer to other `speculator`s.

drslark (changes requested):

PR #13778, opened 2026-08-07
[Bugfix][P/D] Fix MTP PP layer routing in Mooncake

xyy59516:

Mooncake currently identifies index-cache metadata based solely on whether a synthetic layer index is greater than or equal to the target model's total number of layers.

MTP/Eagle draft layers may also use synthetic indices in this range. As a result, their indices can be misidentified as index-cache indices, translated back to parent transformer-layer indices, and incorrectly routed to the first prefill pipeline-parallel rank. However, the local drafter and its KV cache reside on the final PP stage.

This change identifies index-cache groups using their cache-layer names. Metadata indices are translated back to parent transformer-layer indices only for groups containing `.index_cache`. MTP/Eagle indices remain unchanged, ensuring that their KV-cache metadata is routed to the final prefill PP rank.

No. This change fixes Mooncake KV-cache transfer routing for pipeline-parallel deployments that use speculative decoding together with index-cache metadata.

- `ruff check vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py`
- `ruff format --check vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py`
- `bash format.sh ci`
- `pytest -sv tests/ut/kv_offload/test_mooncake_connector.py`

- vLLM version: v0.26.0


wxsIcey (approved):
LGTM!

yiz-liu (approved):

PR #13785, opened 2026-08-07
[BugFix] Add profile memory logger in worker

voidvelocity:

Restores `NPUWorker.profile_memory()` and its invocation at the start of
`execute_model()`, which were accidentally removed by a revert.

#### Background

- `profile_memory()` was originally added in
  [#9765](https://github.com/vllm-project/vllm-ascend/pull/9765)
  (`[Ops][Feature] Profiles the torch reserved memory/allocated memory in
  execute_model()`) to expose the torch reserved / allocated memory on the
  worker (`self.torch_reserved` / `self.torch_allocated`) and log it at DEBUG
  level.
- PR [#11960](https://github.com/vllm-project/vllm-ascend/pull/11960)
  (`[Ops][BugFix] revert "get torch_reserved_memory" in model forward`) was
  intended to remove only the `torch.npu.memory_reserved()` call in the model
  forward pass, which costs ~0.5 ms in TPOT. However, the revert commit
  unintentionally also deleted:
  - the `profile_memory()` method in `vllm_ascend/worker/worker.py`,
  - the `self.torch_reserved` / `self.torch_allocated` initialization in
    `NPUWorker.__init__`,
  - the `self.profile_memory()` calls in `execute_model()` and
    `execute_dummy_batch()`.

This PR restores the `profile_memory()` method and its call in
`execute_model()`. The method is guarded by a DEBUG-level check
(`logger.getEffectiveLevel() > logging.DEBUG`), so it has no effect on
performance in normal (non-DEBUG) runs.

- Fixes #<issue-number> (fill in after the issue is created)

No. The restored code only records memory counters on the worker and emits
DEBUG-level logs; it does not change any API or runtime behavior in default
configurations.

- CI passed with the existing test suite.
- Manual verification: start vLLM with `--log-level DEBUG` and confirm the
  following line appears on each `execute_model()` call:
  `torch reserved memory: X.XX GiB, torch allocated memory: X.XX GiB`.
- Verified that with the default log level (INFO) the method short-circuits and
  no profiling overhead is introduced.

- vLLM version: v0.26.0

### DIFF
--- a/vllm_ascend/worker/worker.py
+++ b/vllm_ascend/worker/worker.py
@@ -605,10 +605,23 @@ class NPUWorker(WorkerBase):
             )
         return int(available_memory)
 
+    def log_memory_stats(self) -> None:
+        """Profiles the torch reserved memory, torch allocated memory in execute_model()."""
+        if not logger.isEnabledFor(logging.DEBUG):
+            return
+        self.torch_reserved = torch.npu.memory_reserved()
+        self.torch_allocated = torch.npu.memory_allocated()
+        logger.debug(
+            "torch reserved memory: %.2f GiB, torch allocated memory: %.2f GiB",
+            self.torch_reserved / GiB_bytes,
+            self.torch_allocated / GiB_bytes,
+        )
+
     def execute_model(
         self,
         scheduler_output: "SchedulerOutput",
     ) -> ModelRunnerOutput | AsyncModelRunnerOutput | None:
+        self.log_memory_stats()
         # enable msMonitor to monitor the performance of vllm-ascend
         if get_ascend_config().msmonitor_use_daemon:
             dp.step()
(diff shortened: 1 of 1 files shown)


voidvelocity:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

vllm_ascend/worker/worker.py
```diff
@@ -605,10 +605,23 @@ def update_available_memory_for_sparse_kv_offload(self, available_memory):
             )
         return int(available_memory)
 
+    def profile_memory(self) -> None:
```
yiz-liu: Rename it to `log_mem_stats`

yiz-liu (approved):

vllm_ascend/worker/worker.py
voidvelocity: done

PR #13790, opened 2026-08-07
[BugFix][Config] Validate MC2 hierarchy and profiling timing

kk-ss1999:

This PR adds configuration guards for MC2 hierarchical communication and fixes a disabled profiling feature still causing timing side effects.

- Rejects `enable_mc2_hierarchy_comm` on devices other than A2 and A3.
- Rejects MC2 hierarchy configurations with more than 512 effective experts.
- Counts EPLB redundant experts in the authoritative runtime check using `len(expert_map) + global_redundant_expert_num`.
- Keeps an early validation for models whose logical expert count already exceeds 512.
- Requires both `profiling_chunk_config.enabled` and `need_timing` before the v1 model runner synchronizes the device or records execution timing.

The root causes were that the MC2 hierarchy configuration had no hardware/expert-count validation, and the v1 model runner checked only the child `need_timing` option without checking its parent feature switch.

Yes.

Invalid MC2 hierarchy configurations now fail with an actionable error before entering the unsupported operator path. Setting `profiling_chunk_config.enabled=false` now reliably disables timing behavior even when `need_timing=true` is present.

Added regression coverage for:

- A5 rejection and A2/A3 acceptance for MC2 hierarchy communication.
- The 512-expert boundary.
- EPLB redundant experts being included in the effective expert count.
- The hierarchy expert limit not applying when hierarchy communication is disabled.
- Disabled profiling chunk configuration suppressing v1 model-runner device synchronization and timing.

Local checks:

- `ruff format --check` on all changed Python files.
- `ruff check` on all changed Python files.
- Python syntax compilation for all changed Python files.
- `git diff --check`.
- Targeted static regression checks for the affected conditions.

The full pytest suite was not run locally because this Windows environment does not have the vLLM, Torch, or pytest runtime dependencies; CI validation is required.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -1784,13 +1784,14 @@ class NPUModelRunner(GPUModelRunner):
             if self.vllm_config.model_config.enable_return_routed_experts and self.routed_experts_initialized:
                 self.routed_experts_capturer.clear_buffer()
 
-        if self.ascend_config.scheduler_config.profiling_chunk_config.need_timing:
+        profiling_chunk_config = self.ascend_config.scheduler_config.profiling_chunk_config
+        if profiling_chunk_config.enabled and profiling_chunk_config.need_timing:
             # Check if the scheduler signaled that calibration is complete.
             # This flag is set cross-process via scheduler_output because
             # modifying the config singleton in the scheduler process does
             # not affect this worker process.
             if getattr(scheduler_output, "disable_profiling_timing", False):
-                self.ascend_config.scheduler_config.profiling_chunk_config.need_timing = False
+                profiling_chunk_config.need_timing = False
             else:
                 self._sync_device()
                 self._execution_start_time = time.perf_counter()
@@ -2205,6 +2206,7 @@ class NPUModelRunner(GPUModelRunner):
     def sample_tokens(
         self, grammar_output: "GrammarOutput | None"
     ) -> ModelRunnerOutput | AsyncModelRunnerOutput | IntermediateTensors:
+        profiling_chunk_config = self.ascend_config.scheduler_config.profiling_chunk_config
         kv_connector_output = self.kv_connector_output
         self.kv_connector_output = None
         pp = get_pp_group()
@@ -2365,8 +2367,10 @@ class NPUModelRunner(GPUModelRunner):
             cudagraph_stats=cudagraph_stats,
             routed_experts=None,
         )
-        if self.ascend_config.scheduler_config.profiling_chunk_config.need_timing and hasattr(
-            self, "_execution_start_time"
+        if (
+            profiling_chunk_config.enabled
+            and profiling_chunk_config.need_timing
+            and hasattr(self, "_execution_start_time")
         ):
             self._sync_device()
             model_runner_output.execution_time_ms = (time.perf_counter() - self._execution_start_time) * 1000.0
--- a/tests/ut/core/test_profiling_chunk.py
+++ b/tests/ut/core/test_profiling_chunk.py
@@ -76,8 +76,16 @@ class TestProfilingChunkConfig(TestBase):
     def test_default_values(self):
         cfg = ProfilingChunkConfig()
         self.assertFalse(cfg.enabled)
+        self.assertFalse(cfg.need_timing)
         self.assertAlmostEqual(cfg.smooth_factor, 1.0)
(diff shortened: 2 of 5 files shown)


vllm_ascend/platform.py
```diff
@@ -385,6 +389,21 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
...
         parallel_config = vllm_config.parallel_config
         cache_config = vllm_config.cache_config
+        if ascend_config.enable_mc2_hierarchy_comm:
+            from vllm_ascend.ascend_forward_context import MAX_MC2_HIERARCHY_COMM_EXPERTS
+
+            device_type = get_ascend_device_type()
+            if device_type not in MC2_HIERARCHY_COMM_SUPPORTED_DEVICES:
+                raise NotImplementedError(
+                    f"enable_mc2_hierarchy_comm is only supported on A2 and A3, but got {device_type.name}."
+                )
```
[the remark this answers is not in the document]
kk-ss1999: get_ascend_device_type() does not return None. When the cached value is unset, it calls _init_ascend_device_type(), which always initializes an AscendDeviceType enum value and falls back to A2 when build-time device information is unavailable. In addition, non-NPU configurations return before reaching this validation. Therefore, device_type.name is safe here.

vllm_ascend/platform.py
```diff
@@ -385,6 +389,24 @@ def check_and_update_config(cls, vllm_config: VllmConfig) -> None:
         model_config = vllm_config.model_config
         parallel_config = vllm_config.parallel_config
         cache_config = vllm_config.cache_config
+        if ascend_config.enable_mc2_hierarchy_comm:
```
linfeng-yuan: Please move this block to ascend_config

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1782,13 +1782,14 @@ def execute_model(
...
-        if self.ascend_config.scheduler_config.profiling_chunk_config.need_timing:
+        profiling_chunk_config = self.ascend_config.scheduler_config.profiling_chunk_config
+        if profiling_chunk_config.enabled and profiling_chunk_config.need_timing:
             # Check if the scheduler signaled that calibration is complete.
             # This flag is set cross-process via scheduler_output because
             # modifying the config singleton in the scheduler process does
             # not affect this worker process.
             if getattr(scheduler_output, "disable_profiling_timing", False):
-                self.ascend_config.scheduler_config.profiling_chunk_config.need_timing = False
+                profiling_chunk_config.need_timing = False
```
linfeng-yuan, line 1794: Please check this mechanism.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1782,13 +1782,14 @@ def execute_model(
             if self.routed_experts_initialized:
                 self.routed_experts_capturer.clear_buffer()
 
-        if self.ascend_config.scheduler_config.profiling_chunk_config.need_timing:
+        profiling_chunk_config = self.ascend_config.scheduler_config.profiling_chunk_config
+        if profiling_chunk_config.enabled and profiling_chunk_config.need_timing:
```
linfeng-yuan, line 1788: Also add constraint to ascend config if users disable this flag but set need_timing

linfeng-yuan (changes requested):

vllm_ascend/worker/model_runner_v1.py
kk-ss1999: done

vllm_ascend/ops/fused_moe/token_dispatcher.py
```diff
@@ -185,6 +185,12 @@ def get_dispatch_mc2_kwargs(
         else:
             quant_mode = 0
         self.moe_expert_num = len(expert_map) + global_redundant_expert_num
+        if self.need_comm_alg and self.moe_expert_num > 512:
```
linfeng-yuan: We've added this constraint in `ascend_config.py`, please remove this block.

linfeng-yuan (dismissed):

kk-ss1999:
/rerun

kk-ss1999:
/rerun

zhangxinyuehfad:
/cancel
[Bot]: cancel completed. No running workflow runs found for this commit.

kk-ss1999:
/rerun
[Bot]: rerun completed.

**Failed:**

linfeng-yuan (approved):

PR #13290, opened 2026-07-31
Revert "[Performance] remove D2H sync in QLIMetadata builder for DSA_CP"

pisceskkk:
Reverts vllm-project/vllm-ascend#12536

- vLLM version: v0.26.0
- vLLM main: https://github.com/vllm-project/vllm/commit/d02df748bf9efd99022f1a062597dc3cb3808485

PR #13123, opened 2026-07-29
[MRV2][BugFix] Fix token distribution in make_dummy for AscendInputBatch

zouzy5137:
### What this PR does / why we need it?

This PR aligns the dummy input generation logic in `AscendInputBatch.make_dummy` with the upstream vLLM implementation. Previously, any remaining tokens (`num_tokens % num_reqs`) were accumulated entirely onto the last request. This PR distributes the extra tokens evenly across the last `num_extra` requests, ensuring that each request receives either `base_tokens` or `base_tokens + 1` tokens. [49751](https://github.com/vllm-project/vllm/pull/49751)

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

CI passed with existing tests.

- vLLM version: v0.26.0
- vLLM main: https://github.com/vllm-project/vllm/commit/d02df748bf9efd99022f1a062597dc3cb3808485

PR #12536, opened 2026-07-22
[Performance] remove D2H sync in QLIMetadata builder for DSA_CP

pisceskkk:
### What this PR does / why we need it?

The DSA context-parallel QLI metadata builder previously calculated `max_seqlen_q` and `max_seqlen_k` from NPU tensors with `.max().item()`. When the `cp_qli` metadata cache was initialized, those scalar reads introduced two unnecessary device-to-host synchronizations on the request metadata-building path.

### Does this PR introduce _any_ user-facing change?

No API or functional behavior change.

- vLLM version: v0.25.1
- vLLM main: https://github.com/vllm-project/vllm/commit/54503ecec0f3ac31e5ecfc5f28652e4cc42307b5

PR #13810, opened 2026-08-07
[BugFix][Test] Restore NPU memory helper import

MrZ20:

Restore the `wait_until_npu_memory_free` import in the MRV2 basic E2E tests.

PR #13290 reverted the import together with the original decorator added by #12536. In the meantime, PR #13123 added three more uses of the same decorator. When #13290 was merged from its stale base, those uses remained without an import and caused the pre-commit ruff check to fail with `F821`.

No. This only fixes test linting and restores the intended test decorator import.

- `./.venv/bin/ruff check tests/e2e/pull_request/one_card/model_runner_v2/test_basic.py --select F821`
- `source .venv/bin/activate && SKIP=gitleaks-offline-scan pre-commit run --files tests/e2e/pull_request/one_card/model_runner_v2/test_basic.py`
- `git diff --check`

- vLLM version: v0.26.0

### DIFF
--- a/tests/e2e/pull_request/one_card/model_runner_v2/test_basic.py
+++ b/tests/e2e/pull_request/one_card/model_runner_v2/test_basic.py
@@ -22,7 +22,7 @@ import pytest
 from vllm import SamplingParams
 from vllm.v1.metrics.reader import Counter, Vector
 
-from tests.e2e.conftest import VllmRunner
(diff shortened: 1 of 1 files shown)


SkychenLee:
go and merge this quickly, ci is blocked

yiz-liu (approved):

PR #13816, opened 2026-08-07
[Feature][KV Offload][v0.26.0rc] Add layerwise prefill KV buffer reuse

ader47:

> [!NOTE]
> This is the `releases/v0.26.0rc` backport of #12852 and the first PR in a stacked series.
> It depends on #13711. The branch includes that prerequisite in its ancestry for integration validation, so its diff will contain the sparse offload changes until #13711 is merged.

Layerwise prefill offloading currently allocates a dedicated device KV cache buffer for every transformer layer. These buffers are not used concurrently during layerwise execution, resulting in unnecessary HBM reservation and limiting the available KV cache capacity.

This PR implements the prefill-side layerwise KV buffer reuse proposed in [RFC #48203](https://github.com/vllm-project/vllm/issues/48203). It allows compatible transformer layers to time-share a bounded set of physical device KV cache buffers while keeping the logical KV cache layout unchanged.

The main changes include:

- Reorganizing KV cache tensor descriptors before allocation so compatible transformer layers can share physical buffers.
- Keeping unsupported or explicitly independent layers on dedicated buffers.
- Updating KV cache memory accounting to expose the recovered HBM capacity.
- Reloading KV data before a shared buffer is reused.
- Reusing a buffer only after the previous layer has completed its save operation.
- Hardening layerwise MemCache allocation, readiness, lease, partial-hit, and failure handling.
- Fixing the SFA execution path so transfer and save hooks run consistently, including for layers without an indexer.

MTP, sparse C8, and heterogeneous main/indexer layouts are covered by the next PR in the stack.

Merge order: #13711 → this PR → the MTP/sparse-C8 PR → the SFA PD RD2H connector PR.

Yes. The supported layerwise prefill offload path can reserve fewer physical HBM KV cache buffers, leaving more device memory available for KV cache capacity without changing the connector interface.

- Added and updated unit tests for layout planning, physical buffer reuse, independent layers, connector scoping, transfer readiness, partial hits, and failure propagation.
- `bash format.sh ci` passed all repository hooks on the complete rebased stack.
- `git diff --check`, commit sign-off checks, duplicate patch checks, and Python syntax/import scans passed.
- Full pytest execution is pending CI because the local environment does not provide the `vllm` runtime package or NPU hardware.

- vLLM version: v0.26.0


ader47:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

ader47:
/rerun 

realliujiaxu (approved):
LGTM

PR #13828, opened 2026-08-08
[BugFix][Worker] Synchronize ACLGraph mode across MC2 DP ranks

### LINKED ISSUE #13826: [Bug]: Variable-length DP/EP requests can hang when MC2 skips ACLGraph mode sync
### Environment

- Hardware: Ascend A2, 8 nodes (4 prefill + 4 decode)
- Model: DeepSeek-V4-Flash-0731 W8A8 with DSpark speculative decoding
- Decode topology: DP32 / EP32 / TP1, PD KV consumer
- vLLM: `v0.25.1` (`752a3a504485790a2e8491cacbb35c137339ad34`)
- vLLM Ascend: `releases/v0.25.1rc` (`50e0ce60849a7b6fe6f94e1de085b6fc06e7006d`)
- Target compilation mode: `CUDAGraphMode.FULL_DECODE_ONLY`
- DSpark drafter: eager
- `max_num_batched_tokens=256`, `max_num_seqs=32`
- Async scheduling and expert parallelism enabled

The same missing synchronization is still present on current `main` at
`d6e8f1ad4bd67cc48a349cd7cf8efab98eecad13`.

### Problem

Fixed-length requests complete, but variable-length requests can leave the decode
service permanently stuck:

- Fixed 16K input / 1K output: completes.
- Uniform 1K-2K input / 1K-2K output, batch size 32, 128 requests: can hang.
- All NPUs remain busy while only a subset of DP ranks still report running requests.
- The service does not recover after the client run ends.

This is most visible at the variable-length tail, when some DP ranks become idle
before others.

### Root cause

`NPUModelRunner._sync_metadata_across_dp()` returns early when
`should_skip_allreduce_across_dp_group()` is true:

```python
if should_skip_allreduce_across_dp_group(self.vllm_config, is_draft_model):
    num_tokens_after_padding = torch.tensor(
        [num_tokens] * self.dp_size, device="cpu", dtype=torch.int32
    )
    return num_tokens, num_tokens_after_padding, cudagraph_mode
```

Skipping token-count synchronization is valid for the MC2 uneven-token path, but
the same early return also skips synchronization of the ACLGraph runtime mode.

At the variable-length tail this can produce:

- busy rank: `CUDAGraphMode.FULL`
- idle/dummy rank: `CUDAGraphMode.NONE`

The ranks then submit graph and eager work into the same EP collective. On the
reproduced A2 configuration this leaves the HCCL collective waiting indefinitely.

### Isolation and validation

A DSA metadata padding patch alone did not prevent the hang.

Making only the decode target globally eager isolated the issue to the target
ACLGraph path. The DSpark drafter was already eager.

| Configuration | Run 1 | Run 2 |
| --- | --- | --- |
| Target ACLGraph, without mode-sync fix | completed | hung |
| Target globally eager | 128/128 | 128/128 |
| Target ACLGraph, with mode-sync fix | 128/128 | 128/128 |

For both fixed runs:

- Failed requests: 0
- Service errors: 0
- All 32 decode endpoints remained healthy
- All ranks returned to `Running=0`, `Waiting=0`

The two ACLGraph fixed runs completed in 126.6 s and 115.7 s.

### Proposed fix

Keep the MC2 per-rank token counts, but synchronize only the runtime graph mode
for the main MoE model when ACLGraph is enabled:

```python
mode_tensor = torch.tensor(cudagraph_mode.value, device="cpu", dtype=torch.int32)
dist.all_reduce(
    mode_tensor,
    op=dist.ReduceOp.MIN,
    group=get_dp_group().cpu_group,
)
cudagraph_mode = CUDAGraphMode(mode_tensor.item())
```

Taking the minimum preserves the existing rule that if any rank requires eager
execution (`NONE`), every DP rank uses eager execution for that step.

The fix should not affect the dense/eager DSpark drafter and should not restore
token padding.

### A3 scope

Current `main` can also select MC2 or FUSED_MC2 on A3, so the same code path is
potentially affected when hierarchy communication is disabled. This report only
claims an A2 hardware reproduction; A3 still requires real-hardware validation.

QwertyJack:

When MC2 supports uneven token counts, `_sync_metadata_across_dp()` skips the
token-count all-reduce. The same early return also skipped ACLGraph mode
synchronization, allowing active and idle DP ranks to submit graph and eager
work into the same EP collective at a variable-length batch tail.

This patch preserves per-rank token counts and synchronizes only
`CUDAGraphMode` for the main MoE model when ACLGraph is enabled. It uses a
minimum reduction on the DP CPU group so that if any rank requires eager
execution, all ranks use eager execution for that step. The eager DSpark
drafter is unchanged.

Fixes #13826

No API or configuration change. Variable-length DP/EP requests no longer hang
when MC2 skips token-count synchronization and DP ranks choose different graph
modes.

- Paired vLLM main commit:
  `0351e9aa1fdf1a51329d1906881528dfe61fc88e`
- `pytest -q tests/ut/worker/a2/test_model_runner_v1.py` (`21 passed`)
- `ruff check`, `ruff format --check`, and `git diff --check`
- `SKIP=actionlint,gitleaks-offline-scan uv run --with pre-commit bash format.sh ci`
  (all remaining hooks passed; `actionlint` was skipped because its first-time
  Go toolchain download stalled, and the downloaded gitleaks binary was
  incompatible with the aarch64 test host)
- The same minimal mode-sync patch was validated on the reported A2
  `v0.25.1rc` service with the original variable-length workload: two
  consecutive runs completed `128/128`, with zero failed requests, zero
  service errors, all 32 decode endpoints healthy, and all ranks returning to
  `Running=0`, `Waiting=0`

- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/worker/a2/test_model_runner_v1.py
+++ b/tests/ut/worker/a2/test_model_runner_v1.py
@@ -38,7 +38,7 @@ class TestSyncMetadataAcrossDP(unittest.TestCase):
         runner.dp_size = 2
         runner.dp_rank = 0
         runner.vllm_config = MagicMock()
-        runner._use_aclgraph = MagicMock(return_value=True)
+        runner.use_aclgraph = True
         mock_get_dp_group.return_value.cpu_group = "cpu_group"
         mock_all_reduce.side_effect = lambda tensor, **_: tensor.fill_(CUDAGraphMode.NONE.value)
 
--- a/tests/ut/worker/a2/test_model_runner_v1.py
+++ b/tests/ut/worker/a2/test_model_runner_v1.py
@@ -4,6 +4,8 @@ from unittest.mock import MagicMock, call, patch
 
 import numpy as np
 import torch
+import torch.distributed as dist
+from vllm.config import CUDAGraphMode
 from vllm.model_executor.layers.attention import MLAAttention
 from vllm.model_executor.models.deepseek_v2 import DeepseekV32IndexerCache
 from vllm.v1.kv_cache_interface import (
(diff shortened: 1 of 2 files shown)


QwertyJack:
Closed by:
https://github.com/vllm-project/vllm-ascend/pull/14307
https://github.com/vllm-project/vllm-ascend/pull/14311
https://github.com/vllm-project/vllm-ascend/pull/14312

PR #12968, opened 2026-07-28
[Feature][MRV2] Support DSV4 Dspark

wxsIcey:
### What this PR does / why we need it?
Support DSV4 Dspark in MRV2

This pr also fix a bug by vllm:
  Fix draft quant config, introduced by https://github.com/vllm-project/vllm/pull/50330

### Does this PR introduce _any_ user-facing change?
N/A

### How was this patch tested?
```
python examples/test_dspark_acceptance.py  \
--model-dir ~/.cache/modelscope/hub/models/UploadWeight/DeepSeek-V4-Flash-DSpark-w4a8-test  \
--num-spec-tokens 5 \
--dataset-name hf \
--dataset-path openai/gsm8k \
--num-prompts 400 \
--temp 0 \
--output-len 256 \
--enforce-eager
```

**mrv1:**
```
------------------------------------------------------------
total_output_tokens: 66918
num_drafts: 15449
num_draft_tokens: 77245
num_accepted_tokens: 51483
mean acceptance length: 4.33
output token throughput: 877.16 tok/s
------------------------------------------------------------
  acceptance at token 0: 0.8951
  acceptance at token 1: 0.7780
  acceptance at token 2: 0.6576
  acceptance at token 3: 0.5516
  acceptance at token 4: 0.4502
```
**mrv2:**
```
------------------------------------------------------------
total_output_tokens: 67232
num_drafts: 15588
num_draft_tokens: 77940
num_accepted_tokens: 51685
mean acceptance length: 4.32
output token throughput: 1191.79 tok/s
------------------------------------------------------------
  acceptance at token 0: 0.8960
  acceptance at token 1: 0.7719
  acceptance at token 2: 0.6527
  acceptance at token 3: 0.5494
  acceptance at token 4: 0.4457
```

- vLLM version: v0.27.1
- vLLM main: https://github.com/vllm-project/vllm/commit/58d3918e3ea0a544ffedadad2ba84559e9c51d8f

Issue #11126, opened 2026-06-29
[RFC]: Add DSpark speculative decoding support for DeepSeek-V4


EanWang211123:
### Motivation.
DeepSeek released DSpark speculative decoding for DeepSeek-V4, with official checkpoints:
- https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-DSpark
- https://huggingface.co/deepseek-ai/DeepSeek-V4-Pro-DSpark
DeepSeek reports **60–85%** faster per-user generation vs MTP-1 at matched throughput (Flash).
vllm-ascend already supports V4 + MTP, but DSpark is not yet a stable spec-decode method. Draft work exists in [#11066](https://github.com/vllm-project/vllm-ascend/pull/11066).
References:
- https://github.com/deepseek-ai/DeepSpec
- Upstream: https://github.com/vllm-project/vllm/issues/46910

### Proposed Change.

Add DSpark support for DeepSeek-V4 on Ascend:
1. Load official `*-DSpark` checkpoints (Markov head + confidence head).
2. Implement `AscendDsparkProposer` (propose → dynamic verify → reject sample).
3. Expose via `--speculative-config` (prefer `method: "dspark"`, or extend existing MTP path).

QwertyJack:
## Design update: DeepSeek V4 DSpark architecture and four-PR delivery

This comment replaces the previous seven-PR plan. The end-to-end prototype in #11196 validated the main DeepSeek V4 DSpark functional and performance path, while #11765 has since added the generic Qwen/GLM DSpark proposer and shared input-preparation infrastructure to `main`.

The DeepSeek V4 implementation will therefore extend the generic DSpark foundation already present on the latest `main`, rather than introducing a second near-duplicate proposer. The remaining work is organized as four reviewable PRs covering the attention contract, model contract, eager correctness path, and graph optimization path.

### Goals

The implementation will provide:

- DSpark draft-model loading from the target checkpoint's `mtp.*` namespace.
- Target auxiliary hidden-state extraction from selected DeepSeek V4 layers.
- Anchor-first, noncausal block drafting followed by sequential Markov sampling.
- Greedy and probabilistic draft sampling with logits-space rejection recovery.
- Standard DeepSeek V4 DSA/SWA paged KV-cache integration.
- `SparseAttnSharedkv` SAS mode 4 with explicit `ori_sparse_indices`.
[...]

PR #13831, opened 2026-08-08
[BugFix][v0.26.0][SpecDecode] Avoid DSpark crash when causality API is absent

jianzs:

PR #13531 added model-reported draft attention causality for GLM-5.2, but the
v1 Ascend DSpark proposer calls `get_draft_attn_causal()` unconditionally.
vLLM treats this model method as optional. Legacy non-causal DSpark models such
as DeepSeek V4 do not expose it, so DSpark initialization/warmup fails with:

```text
AttributeError: 'DSparkDeepseekV4ForCausalLM' object has no attribute 'get_draft_attn_causal'
```

This patch handles the optional capability directly in the attention metadata
path:

- use the model-reported value when `get_draft_attn_causal()` exists;
- otherwise preserve the previous non-causal DSpark behavior (`False`).

The unit tests cover both the missing-method fallback and an explicitly causal
model, so the GLM-5.2 behavior introduced by #13531 remains intact.

Related to #13531 and #11126.

#### Duplicate check

Searched open PRs and issues for `get_draft_attn_causal`, `DSpark causality`,
and `DeepSeek V4 DSpark`.

- #13926 contains the same minimal runtime fallback but does not add regression
  coverage. This PR predates it and tests both the missing-API and explicit
  causal paths.
- #11066 adds GLM-5.2 DSpark support and supplies model-specific causality; it
  does not cover models without the optional API.
- #12968 and #11126 track DeepSeek V4 DSpark model support, rather than this v1
  proposer interface regression.

Yes. DSpark models without `get_draft_attn_causal()` no longer crash and use
the legacy non-causal attention behavior. Models that explicitly report
causality are unchanged. There is no CLI or configuration change.

The fallback behavior is covered by unit tests for both a model without the
optional API and a model that explicitly reports causal attention. Changed-file
pre-commit, `py_compile`, and `git diff --check` passed during development, and
the equivalent fallback behavior was tested on an Ascend dev0 node:

```bash
python3 -m pytest -q tests/ut/spec_decode/test_dspark_proposer.py
# 31 passed, 14 warnings
```

GitHub CI validates the current head.

DeepSeek V4 Flash end-to-end evidence for that equivalent fallback behavior:

- TP=8, PP=1, DSpark: the server became ready; three short requests and one
  200-token request returned HTTP 200. Speculative-decoding metrics confirmed
  the DSpark path was active.
- TP=2, PP=4, DSpark, DSA-CP, KV producer (the original crash topology): the
  server became ready and three short requests returned HTTP 200 with the
  expected output. The original `AttributeError` and engine traceback did not
  recur.

The PP=4 direct KV-producer run also exposed an unrelated long-response content
anomaly. This PR's validation scope is the DSpark startup and optional-interface
compatibility regression; full long-output correctness remains separate.

No documentation update is needed because this restores compatibility without
changing configuration or public APIs.

---

<details>
<summary>Essential Elements of an Effective PR Description Checklist</summary>

- [x] The purpose of the PR.
- [x] The test plan.
- [x] The test results and end-to-end evaluation.
- [x] Documentation impact considered.

</details>

- vLLM version: v0.26.0


tests/ut/spec_decode/test_dspark_proposer.py
```diff
@@ -660,6 +666,20 @@ def test_cad_rewritten_to_cross_attention_shape(self):
         assert cad.actual_seq_lengths_q == [block_size] * num_reqs
         assert cad.decode_token_per_req == block_size
 
+    def test_cad_uses_model_reported_causality(self):
```
wangbj127, line 667: Good.

vllm_ascend/spec_decode/dspark_proposer.py
```diff
@@ -122,6 +123,13 @@ def __init__(
         # per-layer context slot mappings as a flat list
         self._context_slot_mapping_buffers: list[torch.Tensor | None] | None = None
 
+    @staticmethod
+    def _resolve_draft_attn_causal(model: Any) -> bool:
```
wangbj127: Is it necessary to add a new method? I'm neutral.

vllm_ascend/spec_decode/dspark_proposer.py
```diff
@@ -122,6 +123,13 @@ def __init__(
...
+        get_draft_attn_causal = getattr(model, "get_draft_attn_causal", None)
+        # Legacy DSpark models such as DeepSeek V4 use non-causal attention
+        # without exposing this optional vLLM model API.
+        return get_draft_attn_causal()[0] if get_draft_attn_causal is not None else False
```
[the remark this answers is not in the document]
jianzs: Updated.

vllm_ascend/spec_decode/dspark_proposer.py
jianzs: Updated.

drslark (approved):
It's a mirror of https://github.com/vllm-project/vllm-ascend/pull/13935.
LGTM.

wangbj127:
/rerun
[Bot]: rerun command failed: you do not have permission. Only the PR author or users with triage+ permission can trigger /rerun.

lilinsiman:
/rerun
[Bot]: rerun completed.

**Failed:**

jianzs:
/rerun
[Bot]: rerun completed.

**Failed:**

PR #13836, opened 2026-08-08
[Performance][Ops] Support stride-aware recurrent KDA state updates

q664171689:

This PR aligns the recurrent KDA AscendC operator with the stride-aware implementation used by Kimi K3 and fixes the host/operator contract across the Torch, aclnn, tiling, and kernel layers.

The previous implementation converted a non-contiguous recurrent-state view into a contiguous temporary tensor and copied the updated state back afterward. Kimi K3 can pass cache views with a non-zero storage offset and padded outer strides, so this path adds an unnecessary full-state copy and can lose the original storage/stride semantics.

This PR:

- Preserves the recurrent-state storage offset and outer strides and lets the kernel read and update the original state view directly.
- Keeps the inner state matrix dense while supporting both V-first and K-first state layouts at the aclnn/kernel level.
- Adds K=128 with V=128 or V=256 support to the Kimi K3 Torch entry.
- Supports optional `cu_seqlens`, INT32/INT64 sequence metadata, and FP32/BF16/FP16 gate and beta inputs without host-side casts.
- Aligns the Torch adapter, pybind/aclnn API, operator definition, infer-shape, tiling data, and kernel argument/output order.
- Adds explicit `output_final_state` and `inplace_final_state` semantics. The vLLM integration updates the input cache in place and skips materializing an unused second full-state tensor.
- Registers independent ND formats with `FormatList`, avoiding invalid FE format/type combinations.
- Retains the optimized stride-aware state update path and avoids writes to unrelated cache slots or adjacent layer storage.

Yes. Kimi K3 recurrent KDA can consume non-contiguous cache views with non-zero storage offsets directly, supports V=128/256, and avoids the redundant full-state copy in the vLLM in-place path. No new user configuration is required.

- Added an Ascend NPU regression case for a non-contiguous Kimi K3 state pool with a non-zero storage offset.
- Compared the operator output and updated state against the PyTorch reference implementation.
- Verified that the state view keeps the same storage pointer and strides.
- Verified that only selected cache slots are updated and adjacent/unused storage remains unchanged.
- Existing recurrent KDA single-token decode and multi-step accuracy coverage is retained.

- vLLM version: v0.26.0

### DIFF
--- a/csrc/attention/recurrent_kda/README.md
+++ b/csrc/attention/recurrent_kda/README.md
@@ -32,7 +32,7 @@ out = torch.ops._C_ascend.recurrent_kda(
     - `safe_gate=False`：`gate = -exp(A_log) * softplus(g + dt_bias)`。
     - `safe_gate=True`：`gate = lower_bound * sigmoid(exp(A_log) * (g + dt_bias))`。
 - `use_beta_sigmoid_in_kernel=True` 时，kernel 使用 `sigmoid(beta)`；若 `allow_neg_eigval=True`，再乘 2。
-- `_C_ascend`/aclnn 入口支持非连续 state，并通过临时连续 tensor 回写原 view。
+- `_C_ascend`/aclnn 入口保留非连续 state 的 storage、stride 和 offset，kernel 按真实 stride 直接读写；仅允许 slot/head 外层维存在间隔，内部二维 state 矩阵必须行主序稠密。
 
 每个 token 的 recurrent 更新为：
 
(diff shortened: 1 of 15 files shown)


Tflowers-0129 (approved):

q664171689:
/rerun
[Bot]: rerun completed.

**Failed:**

maoxx241 (approved):

q664171689:
/rerun

q664171689:
/rerun

q664171689:
/rerun

linfeng-yuan (approved):

PR #13838, opened 2026-08-08
[Refactor][Config] Migrate AscendConfig to @config pydantic dataclass

massmass:

This PR migrates `AscendConfig` and its 12 sub-config classes to upstream vLLM's `@config` decorator (pydantic dataclass), which achieves two goals:

1. **Type validation**: Config fields use typed pydantic fields with coercion on the validated schema path (`"false"` to `False`, `"2"` to `2`) and reject values outside the declared type.

2. **Unknown key rejection (`extra="forbid"`)**: Undeclared configuration keys raise `ValidationError` instead of being silently ignored.

This PR also standardizes **canonical config consumption** across the in-scope runtime paths: `enable_dsa_cp`, `draft_window_size`, RFork EPLB, and graph-fusion settings now read from the validated `AscendConfig` singleton instead of independently re-reading raw `additional_config`.

### Key changes

- `AscendConfig` and its sub-configs use `@config` with `extra="forbid"`.
- `vllm_config` is not stored as an `AscendConfig` field. The factory passes it explicitly to `derive_and_validate()` and sub-config precondition checks.
- `init_ascend_config()` pre-constructs sub-configs that require precedence resolution or `vllm_config`, then publishes the singleton only after validation succeeds.

> **Scope note — balance scheduling:** `patch_balance_schedule.py` and its dedicated unit tests are unchanged from upstream `main`. Its Engine Core bootstrap reads `enable_balance_scheduling` before the process-local `AscendConfig` singleton is initialized; changing that bootstrap path is intentionally deferred to a dedicated balance-scheduling follow-up. This PR does not claim string coercion for that early raw read.

Yes:

- Typed fields on validated configuration paths coerce supported string representations, such as `"false"`, instead of relying on Python string truthiness.
- Invalid values for typed fields are rejected.
- Unknown top-level and nested keys raise validation errors instead of being silently ignored.

The balance-scheduling Engine Core bootstrap exception described above retains its existing `main` behavior.

- Unit tests added or updated under `tests/ut/test_ascend_config.py` and the affected consumer modules.
- Regression coverage includes explicit `False` values, unknown nested keys, invalid ranges, and full `init_ascend_config()` paths.
- Static checks: `ruff check`, `ruff format --check`, Python compilation, and `git diff --check`.
- Local focused pytest collection is currently blocked by the installed vLLM missing `vllm.model_executor.layers.minimax_rms_norm`; the failure occurs in the global `tests/ut/conftest.py` before target tests are collected.
- NPU runtime validation is left to CI.

- vLLM version: v0.27.1


ZT-AIA (dismissed):

ningjingbengxiaohai (dismissed):

ningjingbengxiaohai (approved):

vllm_ascend/ascend_config.py
```diff
@@ -1312,29 +1153,114 @@ def _is_ascend_config_initialized(config: AscendConfig | None) -> bool:
...
+        "_sparse_li_c8_layer_filter_enabled",
+        # SchedulerConfig-internal top-level legacy keys (resolved internally,
+        # then replaced by the typed scheduler_config passed above).
+        "enable_balance_scheduling",
+        "recompute_scheduler_enable",
+        "short_request_first_config",
+        "profiling_chunk_config",
+        "batch_job_sched_config",
+    }
+    kwargs = {k: v for k, v in additional_config.items() if k not in _NON_USER_INPUT_KEYS}
```
FayeSpica, line 1226: why flow additional_config into AscendConfig? it breaks downstream consumers(like vllm-omni)

PR #13848, opened 2026-08-08
[BugFix]Move batch_matmul_transpose to aclnn custom op path

ZT-AIA:
## Background

On `releases/v0.20.2rc`, `batch_matmul_transpose` still uses the legacy direct kernel launch path, which is harder to maintain and less portable across platforms. This PR migrates the op to an AscendC custom operator (aclnn path) and syncs the functional and accuracy fixes already made on the remote branch.

## Changes

- Add `csrc/moe/batch_matmul_transpose/`
  - `op_host`: op definition, infershape, tiling, and tiling data
  - `op_kernel`: AscendC kernel implementation
  - `batch_matmul_transpose_torch_adpt.h`: call through `aclnnBatchMatmulTranspose`
- Remove the legacy direct kernel implementation under `csrc/batch_matmul_transpose/`
- Update build and registration logic:
  - `CMakeLists.txt`
  - `csrc/build_aclnn.sh`
  - `csrc/ops.h`
  - `csrc/torch_binding.cpp`
  - `csrc/torch_binding_meta.cpp`
- Add `test_fused_mla_badcase_shape` to cover the bf16 shape used by fused MLA

## Key Fix

After migrating to the aclnn custom op path, the compiled kernel binary only registers the default tiling key (`_0`), while the host previously passed the real tiling key (e.g. `1168`) to the runtime, causing:

```text
BinaryGetFunctionByEntry failed, funcEntry=1168
```

Fix: set `context->SetTilingKey(0)` on the host so the runtime uses the default entry. The real tiling key is still written into the tiling data and dispatched inside the kernel via `tilingKey >> 2`.

## Verification

On the NPU environment :

```bash
pytest -sv tests/e2e/nightly/single_node/ops/singlecard_ops/test_batch_matmul_transpose.py
```

Result: `4 passed`

## Notes

- This PR is based on the latest commit of `releases/v0.20.2rc`
- Unrelated local changes (mrope, moe_comm, etc.) are intentionally excluded

- vLLM version: 


ApsarasX:
There is a `torch_npu.npu_transpose_batchmatmul` call in `sfa_v1.py`:

https://github.com/vllm-project/vllm-ascend/blob/main/vllm_ascend/attention/sfa_v1.py#L1131

Should we remove this custom op as well?

ZT-AIA:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E-Light

PR #13864, opened 2026-08-09
[BugFix][Worker] Preserve async accepted-token snapshot

### LINKED ISSUE #13863: [Bug]: Async MTP can reuse stale accepted-token counts after batch condense
### Your current environment

<details>
<summary>Environment used for the controlled reproduction</summary>

```text
Hardware: Ascend A3, 2 NPUs (TP2)
Model: /models/Qwen3.6-35B-A3B-w8a8 (real weights)
vLLM: 0.23.0
vLLM-Ascend: 0.23.0rc1 plus #12255
torch-npu: 2.10.0.post2
CANN: 9.0.1
```

</details>

### Describe the bug

Qwen3.5/Qwen3.6 hybrid models can intermittently return repeated, dropped, or otherwise garbled tokens when async scheduling, MTP speculative decoding, and prefix caching are enabled under high batch churn and KV-cache pressure.

This is a different failure mode from #12255. That PR fixed decode-graph dispatch for shape-ambiguous prefill rows and reduced the observed failure rate, but the controlled run after applying it still produced 2 corrupted outputs out of 512 requests at 100% KV-cache usage.

The original production request payloads are no longer available, so the numbers below are from a controlled pressure reproduction, not a byte-identical replay. The reproduction uses a fixed deterministic 8,196-token foreground request while 16 long-running 32,768-input/2,048-output requests create batch compression and KV pressure.

Relevant server options:

```bash
vllm serve /models/Qwen3.6-35B-A3B-w8a8 \
  --served-model-name qwen3.6 \
  --tensor-parallel-size 2 \
  --max-model-len 262144 \
  --quantization ascend \
  --max-num-seqs 128 \
  --enable-prefix-caching \
  --async-scheduling \
  --speculative-config '{"method":"qwen3_5_mtp","num_speculative_tokens":3}' \
  --num-gpu-blocks-override 288
```

Pressure probe shape:

```text
foreground: 512 requests, concurrency 16, 8,196 prompt tokens, temperature 0
background: 16 requests, 32,768 prompt tokens, 2,048 output tokens
```

Observed results:

```text
after #12255:             2/512 foreground outputs differed, KV peak 100%
async scheduling off:     0/1024 foreground outputs differed
accepted-count snapshot:  0/2048 foreground outputs differed
final current-code run:   0/512 differed, 512 HTTP 200, one output hash,
                          KV peak 98.82%
```

#### Root cause

The previous iteration's accepted-token counts are copied into `InputBatch.num_accepted_tokens_cpu`. Before the next `_prepare_inputs` remaps those counts, async `condense()` updates the persistent input-batch rows. A new request can therefore overwrite a previous row with the default value `1`.

For example:

```text
previous row 11 accepted 4 tokens
condense/new request overwrites persistent row 11 with 1
current row maps to previous row 11 through prev_positions
remap reads 1 instead of the previous iteration's 4
```

The incorrect value changes the Mamba/GDN state-copy offset, so a request can replay or drop fixed tokens. The problem is software state mapping, not an NPU hardware or numerical-precision failure.

The same vulnerable read exists in both `NPUModelRunner` and the corresponding vLLM core implementation. The earlier vLLM async hybrid fix (https://github.com/vllm-project/vllm/pull/38556) synchronized the copy, but did not keep the previous iteration's accepted counts independent from mutable `InputBatch` rows.

#### Expected behavior

Accepted-token counts from the previous iteration must survive input-batch condensation and be remapped by `prev_positions` before the current rows are written back to `InputBatch`.

#### Proposed fix

Use the model runner's independent CPU/GPU accepted-token buffer as the previous-iteration snapshot:

1. Copy postprocessed accepted counts into the independent CPU buffer.
2. Remap that snapshot through `prev_positions` after the async event completes.
3. Only then update the current `InputBatch` rows.

A regression test should cover a previous high row (for example row 11 with value 4) after that same persistent input-batch row has already been overwritten with 1.

QwertyJack:

Fixes #13863.

Async scheduling can condense and reuse `InputBatch` rows before the next model iteration remaps accepted-token counts through `prev_positions`. Because the previous counts are currently stored in those mutable rows, a new request can overwrite a previous value (for example, replacing `4` with `1`). The wrong count changes the Mamba/GDN state-copy offset and can produce repeated, dropped, or garbled tokens under MTP load.

This is separate from #12255, which fixed decode-graph dispatch for shape-ambiguous prefill rows. Applying that fix reduced the failure rate but did not prevent this state-mapping failure.

This PR:

- writes async postprocessed accepted-token counts to the runner's independent CPU/GPU buffer while preserving upstream `InputBatch` ownership for non-async scheduling;
- remaps that previous-iteration snapshot when a previous-row mapping exists and initializes all-new async rows to `1`;
- adds regression coverage for the postprocess destination, a reused high batch row, an all-new async batch, and a moved non-async survivor.

Yes. Hybrid models using async scheduling with speculative decoding no longer reuse overwritten accepted-token counts after batch condensation.

- `PYTHONPATH=<vLLM v0.26.0 worktree> pytest -q tests/ut/worker/a2/test_model_runner_v1.py` (`25 passed`)
- `ruff check vllm_ascend/worker/model_runner_v1.py tests/ut/worker/a2/test_model_runner_v1.py`
- `ruff format --check vllm_ascend/worker/model_runner_v1.py tests/ut/worker/a2/test_model_runner_v1.py`
- `python -m compileall -q vllm_ascend/worker/model_runner_v1.py tests/ut/worker/a2/test_model_runner_v1.py`
- `git diff --check`
- The same snapshot/remap algorithm was validated with real Qwen3.6-35B-A3B-W8A8 weights on Ascend A3 TP2. With async scheduling, MTP3, prefix caching, 16 concurrent 32k/2k background requests, and 512 deterministic 8.2k foreground probes, all 512 requests returned HTTP 200 with one output hash and zero corrupt outputs at 98.82% peak KV-cache usage.

- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/worker/a2/test_model_runner_v1.py
+++ b/tests/ut/worker/a2/test_model_runner_v1.py
@@ -20,6 +20,57 @@ from vllm_ascend.utils import AscendDeviceType
 from vllm_ascend.worker.model_runner_v1 import NPUModelRunner
 
 
+class TestNPUModelRunnerAcceptedTokens(unittest.TestCase):
+    @patch("vllm_ascend.worker.model_runner_v1.mamba_utils.postprocess_mamba_align_gpu")
+    def test_postprocess_writes_accepted_counts_to_independent_snapshot(self, mock_postprocess):
+        runner = NPUModelRunner.__new__(NPUModelRunner)
+        runner.speculative_config = object()
+        runner.model_config = SimpleNamespace(is_hybrid=True)
+        runner.cache_config = SimpleNamespace(mamba_cache_mode="align")
+        runner.num_accepted_tokens = SimpleNamespace(
+            cpu=torch.zeros(2, dtype=torch.int32),
+            gpu=torch.zeros(2, dtype=torch.int32),
+        )
+        persistent_counts = torch.ones(2, dtype=torch.int32)
+        runner.input_batch = SimpleNamespace(num_accepted_tokens_cpu_tensor=persistent_counts)
+        runner.kv_cache_config = object()
+        runner.compilation_config = SimpleNamespace(static_forward_context={})
+        runner.model = SimpleNamespace(get_mamba_state_copy_func=lambda: ())
+        runner.num_accepted_tokens_event = MagicMock()
+        runner._get_mamba_bufs = MagicMock()
+
+        runner._update_states_after_model_execute(
+            torch.tensor([[10, 11, -1], [20, -1, -1]]),
+            MagicMock(),
+        )
+
+        self.assertIs(
+            mock_postprocess.call_args.kwargs["num_accepted_tokens_cpu_tensor"],
+            runner.num_accepted_tokens.cpu,
+        )
+        self.assertIsNot(
+            mock_postprocess.call_args.kwargs["num_accepted_tokens_cpu_tensor"],
+            persistent_counts,
+        )
+        runner.num_accepted_tokens_event.record.assert_called_once_with()
+
+    def test_remap_uses_snapshot_after_persistent_row_is_overwritten(self):
+        runner = NPUModelRunner.__new__(NPUModelRunner)
+        previous_counts = np.ones(16, dtype=np.int32)
+        previous_counts[4] = 3
+        previous_counts[11] = 4
+        persistent_counts = np.ones(16, dtype=np.int32)
+
+        runner.num_accepted_tokens = SimpleNamespace(np=previous_counts)
+        runner.prev_positions = SimpleNamespace(np=np.array([11, -1, 4] + [-1] * 13, dtype=np.int64))
+        runner.input_batch = SimpleNamespace(num_accepted_tokens_cpu=persistent_counts)
+
+        runner._remap_num_accepted_tokens(num_reqs=3)
+
+        np.testing.assert_array_equal(previous_counts[:3], [4, 1, 3])
+        np.testing.assert_array_equal(persistent_counts[:3], [4, 1, 3])
+
+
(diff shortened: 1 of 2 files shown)


QwertyJack:
Closing this PR because its Ascend-side fix has already been incorporated into #14454, not because the underlying upstream issue no longer exists.

The root cause is in the vLLM GPUModelRunner contract and is tracked by https://github.com/vllm-project/vllm/issues/51571. The proposed upstream fix, https://github.com/vllm-project/vllm/pull/51599, uses the same essential approach as this PR: keep the previous iteration's accepted-token counts in a runner-owned buffer, remap them through prev_positions after synchronization, then publish the current-row values back to InputBatch. That upstream PR is still open, so CUDA's GPUModelRunner is not fixed yet. https://github.com/vllm-project/vllm/pull/53613 addresses a complementary ordering hazard and does not replace vllm#51599.

While this PR was waiting for review, the equivalent NPU implementation was unintentionally bundled into the much larger Kimi K3 PR #14454, merged as 0a023b094e9e88ffaca0b1fda02529cef6277f8e. It added the independent accepted-token snapshot, prev_positions remapping, new-request initialization, and regression coverage to NPUModelRunner. As a result, rebasing this PR would only duplicate and conflict with code already on vLLM Ascend main.

Therefore:

- vLLM Ascend main is already fixed by #14454;
- vLLM upstream still needs vllm#51599 or an equivalent fix;
- once the upstream fix is merged and included in the vLLM commit supported by vLLM Ascend, the downstream NPUModelRunner override should be removed.

Superseded by #14454.

PR #13865, opened 2026-08-09
[v0.23.0][BugFix][Worker] Preserve async accepted-token snapshot

### LINKED ISSUE #13863: [Bug]: Async MTP can reuse stale accepted-token counts after batch condense

QwertyJack:

Fixes #13863. Release-branch counterpart of #13864.

Async scheduling can condense and reuse `InputBatch` rows before the next model iteration remaps accepted-token counts through `prev_positions`. Because the previous counts are currently stored in those mutable rows, a new request can overwrite a previous value (for example, replacing `4` with `1`). The wrong count changes the Mamba/GDN state-copy offset and can produce repeated, dropped, or garbled tokens under MTP load.

This is separate from #12255, which fixed decode-graph dispatch for shape-ambiguous prefill rows. Applying that fix reduced the failure rate but did not prevent this state-mapping failure.

This PR:

- writes async postprocessed accepted-token counts to the runner's independent CPU/GPU buffer while preserving upstream `InputBatch` ownership for non-async scheduling;
- remaps that previous-iteration snapshot when a previous-row mapping exists and initializes all-new async rows to `1`;
- adds regression coverage for the postprocess destination, a reused high batch row, an all-new async batch, and a moved non-async survivor.

Yes. Hybrid models using async scheduling with speculative decoding no longer reuse overwritten accepted-token counts after batch condensation.

- `pytest -q tests/ut/worker/a2/test_model_runner_v1.py` (`20 passed`)
- `ruff check vllm_ascend/worker/model_runner_v1.py tests/ut/worker/a2/test_model_runner_v1.py`
- `ruff format --check vllm_ascend/worker/model_runner_v1.py tests/ut/worker/a2/test_model_runner_v1.py`
- `python -m compileall -q vllm_ascend/worker/model_runner_v1.py tests/ut/worker/a2/test_model_runner_v1.py`
- `git diff HEAD^ --check`
- The same snapshot/remap algorithm was validated with real Qwen3.6-35B-A3B-W8A8 weights on Ascend A3 TP2. With async scheduling, MTP3, prefix caching, 16 concurrent 32k/2k background requests, and 512 deterministic 8.2k foreground probes, all 512 requests returned HTTP 200 with one output hash and zero corrupt outputs at 98.82% peak KV-cache usage.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -742,6 +742,53 @@ class NPUModelRunner(GPUModelRunner):
 
         return super()._update_states(scheduler_output)
 
+    def _update_states_after_model_execute(
+        self, output_token_ids: torch.Tensor, scheduler_output: "SchedulerOutput"
+    ) -> None:
+        """Update cached hybrid-model states after model execution."""
+        if not self.speculative_config or not self.model_config.is_hybrid:
+            return
+
+        num_reqs = output_token_ids.size(0)
+        self.num_accepted_tokens.gpu[:num_reqs] = (output_token_ids != -1).sum(dim=1)
+
+        if self.cache_config.mamba_cache_mode == "align":
+            mamba_utils.postprocess_mamba_align_gpu(
+                bufs=self._get_mamba_bufs(),
+                num_reqs=num_reqs,
+                num_accepted_tokens_gpu=self.num_accepted_tokens.gpu,
+                num_accepted_tokens_cpu_tensor=self.num_accepted_tokens.cpu,
+                input_batch=self.input_batch,
+                kv_cache_config=self.kv_cache_config,
+                forward_context=self.compilation_config.static_forward_context,
+                mamba_state_copy_funcs=self.model.get_mamba_state_copy_func(),
+            )
+        else:
+            self.num_accepted_tokens.copy_to_cpu(num_reqs)
+
+            if self.cache_config.mamba_cache_mode == "all":
+                mamba_utils.postprocess_mamba_all(
+                    scheduler_output,
+                    self.kv_cache_config,
+                    self.input_batch,
+                    self.requests,
+                    self.mamba_state_idx,
+                    self.num_spec_tokens,
+                    num_reqs,
+                )
+
+        assert self.num_accepted_tokens_event is not None
+        self.num_accepted_tokens_event.record()
+
+    def _remap_num_accepted_tokens(self, num_reqs: int) -> None:
+        """Map the previous iteration's accepted counts to current rows."""
+        prev_idx = self.prev_positions.np[:num_reqs]
+        new_mask = prev_idx < 0
+        accepted = self.num_accepted_tokens.np
+        accepted[:num_reqs] = accepted[np.where(new_mask, 0, prev_idx)]
+        accepted[:num_reqs][new_mask] = 1
+        self.input_batch.num_accepted_tokens_cpu[:num_reqs] = accepted[:num_reqs]
+
     def _pad_query_start_loc_for_fia(
         self,
         query_start_loc: torch.Tensor,
@@ -1084,21 +1131,11 @@ class NPUModelRunner(GPUModelRunner):
             self.num_accepted_tokens_event.synchronize()
             # Async mode: condense() reordered indices, use prev_positions mapping
             if self.use_async_scheduling and prev_req_id_to_index:
-                prev_idx = self.prev_positions.np[:num_reqs]
-                new_mask = prev_idx < 0
-                self.num_accepted_tokens.np[:num_reqs] = (
-                    self.input_batch.num_accepted_tokens_cpu[
-                        np.where(new_mask, 0, prev_idx)
-                    ]
-                )
-                self.num_accepted_tokens.np[:num_reqs][new_mask] = 1
-                self.input_batch.num_accepted_tokens_cpu[:num_reqs] = (
-                    self.num_accepted_tokens.np[:num_reqs]
-                )
+                self._remap_num_accepted_tokens(num_reqs)
             else:
                 # Non-async mode: use values directly
-                self.num_accepted_tokens.np[:num_reqs] = (
-                    self.input_batch.num_accepted_tokens_cpu[:num_reqs]
+                self.input_batch.num_accepted_tokens_cpu[:num_reqs] = (
+                    self.num_accepted_tokens.np[:num_reqs]
                 )
             self.num_accepted_tokens.np[num_reqs:].fill(1)
             self.num_accepted_tokens.copy_to_gpu()
@@ -2100,10 +2137,8 @@ class NPUModelRunner(GPUModelRunner):
 
                 pad_attn = cudagraph_mode == CUDAGraphMode.FULL
 
-                # NOTE(Angazenn): According to https://github.com/vllm-project/vllm/pull/30877,
-                # there should be a corresponding 'postprocess_mamba'. However, it is called inside
-                # '_update_states_after_model_execute', which is not overridden in vLLM-Ascend.
-                # We simply utilize the implementation in vLLM.
+                # postprocess_mamba runs later in
+                # _update_states_after_model_execute.
                 if self.cache_config.mamba_cache_mode == "align":
                     # preprocess_mamba reads req_state.num_computed_tokens (CPU)
                     # to decide copy operations, so we must apply deferred
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -746,6 +746,11 @@ class NPUModelRunner(GPUModelRunner):
         self, output_token_ids: torch.Tensor, scheduler_output: "SchedulerOutput"
     ) -> None:
         """Update cached hybrid-model states after model execution."""
+        if not self.use_async_scheduling:
+            return super()._update_states_after_model_execute(
+                output_token_ids, scheduler_output
+            )
+
         if not self.speculative_config or not self.model_config.is_hybrid:
             return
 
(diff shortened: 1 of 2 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -742,6 +742,53 @@ def _update_states(self, scheduler_output: "SchedulerOutput") -> Callable | None
...
+
+        num_reqs = output_token_ids.size(0)
+        self.num_accepted_tokens.gpu[:num_reqs] = (output_token_ids != -1).sum(dim=1)
+
+        if self.cache_config.mamba_cache_mode == "align":
+            mamba_utils.postprocess_mamba_align_gpu(
+                bufs=self._get_mamba_bufs(),
+                num_reqs=num_reqs,
+                num_accepted_tokens_gpu=self.num_accepted_tokens.gpu,
+                num_accepted_tokens_cpu_tensor=self.num_accepted_tokens.cpu,
```
LoganJane, line 765: **[P1] Preserve accepted-count ownership for non-async scheduling**

This redirects the postprocessed count into the runner snapshot unconditionally, including when async scheduling is disabled. On the next iteration, `_update_states()` can remove requests and `condense()` the InputBatch before `_prepare_inputs()`. The runner snapshot still uses the previous row order, but the non-async prepare branch does not have a `prev_positions` remap. A survivor that moves from row 1 to row 0 can therefore receive the previous row 0 request's accepted count. Please keep the upstream/InputBatch destination for non-async mode (for example, delegate to `super()`), or maintain a request-identity mapping for all modes.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1084,21 +1131,11 @@ def _prepare_inputs(
...
-                self.num_accepted_tokens.np[:num_reqs][new_mask] = 1
-                self.input_batch.num_accepted_tokens_cpu[:num_reqs] = (
-                    self.num_accepted_tokens.np[:num_reqs]
-                )
+                self._remap_num_accepted_tokens(num_reqs)
             else:
                 # Non-async mode: use values directly
-                self.num_accepted_tokens.np[:num_reqs] = (
-                    self.input_batch.num_accepted_tokens_cpu[:num_reqs]
+                self.input_batch.num_accepted_tokens_cpu[:num_reqs] = (
```
LoganJane: **[P1] Do not treat a missing previous mapping as an identity mapping**

This `else` also covers the first iteration and an async batch with no surviving previous requests. `CpuGpuBuffer` starts at zero, while new InputBatch rows intentionally default `num_accepted_tokens_cpu` to `1`; after a full batch replacement this snapshot may instead contain stale counts from the old batch. Copying it into current rows can feed zero/stale accepted counts to Mamba/GDN. Please split this into: async + mapping => remap snapshot; async + no mapping => fill current rows with `1`; non-async => retain the old InputBatch-to-runner flow or otherwise remap by request identity.

LoganJane (changes requested):
The independent accepted-token snapshot is the right fix for #13863's async path, and the steady-state async implementation does not appear to add a meaningful performance cost: it reuses the existing `CpuGpuBuffer`, preserves the existing event synchronization, and redirects the same O(num_reqs) D2H/H2D traffic rather than adding a model kernel or a global device sync.

However, I think the current ownership change introduces correctness regressions in paths that do not have a previous-row mapping, so this should not merge as-is.

### Blocking correctness concerns

1. **Non-async batch changes lose request identity.** The new override stores the latest accepted counts only in `self.num_accepted_tokens.cpu` for all scheduling modes. Between model iterations, `_update_states()` can still remove requests and run `InputBatch.condense()` when async scheduling is disabled. The runner snapshot remains in the previous row order, while the InputBatch is now in the current row order. The non-async branch then writes the snapshot prefix directly into the current rows without remapping.

   For example, previous rows `[A=4, B=2]`; A finishes and B moves from row 1 to row 0; the new code writes snapshot row 0 (`4`, belonging to A) to B instead of `2`. This can change the Mamba/GDN state-copy offset and create the same silent repeated/dropped/garbled-token failure in an async-off configuration. The reported `async scheduling off: 0/1024` result predates this ownership reversal and therefore does not cover this regression.

2. **No previous mapping is not equivalent to an identity mapping.** `CpuGpuBuffer` is zero-initialized, while new InputBatch requests intentionally initialize `num_accepted_tokens_cpu` to `1`. On the first batch, or after all previous requests have left, `prev_req_id_to_index` is empty and the current `else` branch writes zero or stale snapshot values into new/current rows. This is often masked by prefill, but it violates the accepted-token invariant and can affect resumed/stateful or shape-ambiguous short chunks.

### Suggested implementation

The smallest safe release-branch fix is to scope the independent snapshot to async scheduling:

- for non-async scheduling, retain the upstream destination/ownership and the old `InputBatch -> runner` prepare direction, so `condense()` continues moving accepted counts with request rows;
- for async scheduling with a non-empty previous mapping, remap the independent snapshot as proposed here;
- for async scheduling without a previous mapping, initialize current rows explicitly to `1`;
- alternatively, maintain a previous `req_id -> row` snapshot and remap by request identity in every scheduling mode.

For example, `_update_states_after_model_execute()` could delegate to `super()` when `not self.use_async_scheduling`, while the prepare path handles the three cases above separately.

### Required regression coverage

Please add tests for:

- async off: a surviving request moves from a high row to a low row after another request finishes;
- async on with an empty previous mapping / all-new batch, verifying all current counts are `1`;
- initial runner buffer zeros versus InputBatch's default ones;
- preferably `mamba_cache_mode="all"` as well as `"align"`.

A real async-off churn run should also be repeated against the final patch. The current GitHub workflow passed lint/type-check selection, but its selected runtime test jobs were skipped; the PR description's local unit test and A3 async stress run do not exercise these branches.

### Non-blocking notes

- I do not see an obvious throughput/latency regression in the intended async path, but the PR does not include an A/B throughput, TPOT, or scheduler-step benchmark.
- The chained NumPy assignment is valid for the current `np.ndarray` view, although the single-`np.where` form suggested in the resolved thread would be clearer.
- This change should not inherently disable or deadlock async scheduling: it adds no locks, scheduler changes, streams, or new synchronization points. The primary risk here is silent state/accuracy corruption in the uncovered row-mapping branches.

vllm_ascend/worker/model_runner_v1.py
QwertyJack: Fixed in 537ea87f9. The sync logic now explicitly separates all three cases: async with a previous mapping remaps the independent snapshot; async without a previous mapping fills both current runner and InputBatch rows with `1`; non-async retains InputBatch-to-runner ownership. Added `test_async_without_previous_mapping_initializes_current_rows` with an initial zero plus stale snapshot values, while the existing overwritten-row test now exercises the mapped branch through the same helper. This sync path is shared by `align` and `all`; `postprocess_mamba_all` itself does not consume accepted counts. Release tests pass 20/20; main passes 25/25.

vllm_ascend/worker/model_runner_v1.py
QwertyJack: Fixed in 537ea87f9. Non-async postprocessing now delegates to `GPUModelRunner._update_states_after_model_execute`, preserving the upstream InputBatch ownership. During prepare, the non-async branch copies the already-condensed InputBatch rows back into the runner buffer, so a survivor keeps its request-associated accepted count after moving rows. Added `test_non_async_postprocess_delegates_to_upstream` and `test_non_async_sync_uses_condensed_input_batch_rows`. The release runner test file passes 20/20; the equivalent main fix is deab0a13e and passes 25/25.

QwertyJack:
@LoganJane Addressed both blocking ownership cases in 537ea87f9 and replied with the implementation and regression-test details in each inline thread. The equivalent main-branch update is deab0a13e. Please take another look when convenient.

ningjingbengxiaohai (approved):

PR #13870, opened 2026-08-09
[BugFix][P/D] Return KV params when P stops on EOS

hukongyi:

For the synthetic P-side one-token request used by PD deployment, an EOS token can cause vLLM to mark the request as `FINISHED_STOPPED`. MooncakeHybridConnector previously generated KV transfer parameters only for `FINISHED_LENGTH_CAPPED`, so the PD proxy could receive null `kv_transfer_params`.

This change accepts `FINISHED_STOPPED` as a valid completion status and adds a regression unit test.

Yes. This fixes KV transfer parameter generation for MooncakeHybridConnector in P/D deployment. No API changes are introduced.

- `python -m py_compile vllm_ascend\\distributed\\kv_transfer\\kv_p2p\\mooncake_hybrid_connector.py tests\\ut\\kv_offload\\test_mooncake_hybrid_connector.py`
- `git diff --check`
- The targeted pytest could not be collected in this environment because `torch` is not installed.
- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/kv_offload/test_mooncake_hybrid_connector.py
+++ b/tests/ut/kv_offload/test_mooncake_hybrid_connector.py
@@ -277,6 +277,24 @@ class TestMooncakeHybridConnectorScheduler(unittest.TestCase):
         self.assertEqual(params["num_prompt_blocks"], 2)
         self.assertIn("req1", scheduler._reqs_need_send)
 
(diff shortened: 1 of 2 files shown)


LCAIZJ:
Does mooncake_connector.py have a similar issue?

LCAIZJ:
Did the request have min_tokens=1 set when this happened?

PR #13882, opened 2026-08-09
[Refactor][Ops] Align routing replay capture with upstream

zjchenn:

Related to #13220.

Routing replay capture was still wired through an Ascend-specific model-runner override and a manual hook in `AscendRoutedExperts`. The override replaced the upstream binder on the vLLM 0.26 path, while the manual hook ran after Ascend `log2phy` mapping and could therefore export physical rather than logical expert IDs. Its broad exception handler could also hide incomplete capture results.

This PR aligns Ascend with the upstream MoE router capture contract:

- use the upstream binder to attach capture callbacks to Ascend routers;
- capture logical top-k expert IDs inside the router before EPLB or `log2phy` mapping;
- remove the legacy capturer plumbing and post-mapping manual capture path;
- retain the Ascend `RoutedExpertsCapturer.capture` patch for DP, SP, AllToAll, and MC2 token-layout normalization;
- add coverage for binder compatibility, logical-to-physical mapping order, and routing with no capture callback.

Yes. When routed expert export is enabled, Ascend now consistently returns logical expert IDs before EPLB or `log2phy` mapping, matching the upstream vLLM contract. Behavior is unchanged when routed expert export is disabled.

- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/ops/test_fused_moe.py
+++ b/tests/ut/ops/test_fused_moe.py
@@ -20,6 +20,7 @@ from vllm_ascend.ops.fused_moe.routed_experts import (
 )
 from vllm_ascend.ops.fused_moe.router import fused_topk_router as fused_topk_router_module
 from vllm_ascend.ops.fused_moe.router.fused_topk_router import AscendFusedTopKRouter
+from vllm_ascend.ops.fused_moe.router.grouped_topk_router import AscendGroupedTopKRouter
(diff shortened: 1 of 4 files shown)


zjchenn:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

weijinqian0 (approved):

Tflowers-0129 (approved):

PR #13929, opened 2026-08-10
[Bugfix][P/D] Release prefill KV on decode rejection

Yuli-yx:

In disaggregated prefill, the P node delays freeing KV blocks until the D node acknowledges the receive. If the D request is rejected before scheduler admission (for example, prompt tokens plus max_tokens exceeds max_model_len), update_state_after_alloc is never called and the P-side blocks remain allocated until the connector timeout. The load-balancing proxy also retried backend failures with the same KV-transfer metadata, even though the first rejected D request may already have released the P-side blocks.

This PR fixes both sides of the failure path:

- when request_finished observes an untouched do_remote_prefill request, enqueue an empty receive task;
- let the worker skip data transfer for the empty block list while still sending DONE to the P node;
- clear do_remote_prefill so the cleanup is scheduled only once;
- make each proxy P/D backend HTTP request a single attempt, without reusing stale KV-transfer metadata after an error;
- remove the obsolete max-retries and retry-delay proxy options;
- add scheduler and worker regression coverage for the empty-receive/DONE path.

This aligns the connector cleanup with vLLM Mooncake behavior introduced in https://github.com/vllm-project/vllm/pull/24718 and the pre-admission rejection notification path in https://github.com/vllm-project/vllm/pull/41269. It also matches the upstream toy and Mooncake proxies, which perform one P request and one D request per transaction.

The explicit stop_reason=recomputed path remains unchanged because it creates a fresh P-to-D transaction rather than retrying a failed backend request with stale transfer metadata. This PR intentionally does not address cancellation before the request reaches the D engine; that needs a separate proxy-to-D cancellation/cleanup design.

Yes. A D-side pre-admission rejection now triggers the Mooncake completion handshake so the P node releases the request KV cache immediately. The proxy no longer retries P or D backend HTTP failures with the same transaction metadata.

- cd vllm-ascend && uv run python .github/workflows/scripts/coverage.py
  - all 213 UT files covered
- cd vllm-ascend && uv run pytest -q tests/ut/kv_offload/test_mooncake_connector.py
  - 109 passed, 22 subtests passed
- cd vllm-ascend && SKIP=gitleaks-offline-scan uv run pre-commit run --files examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py tests/ut/kv_offload/test_mooncake_connector.py
  - all selected hooks passed; gitleaks was skipped because the local environment has neither gitleaks nor wget

- vLLM version: v0.26.0

### DIFF
--- a/examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
+++ b/examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
@@ -675,10 +675,6 @@ def parse_args() -> argparse.Namespace:
     parser.add_argument("--prefiller-ports", type=int, nargs="+", default=[8001])
     parser.add_argument("--decoder-hosts", type=str, nargs="+", default=["localhost"])
     parser.add_argument("--decoder-ports", type=int, nargs="+", default=[8002])
-    parser.add_argument("--max-retries", type=int, default=3, help="Maximum number of retries for HTTP requests")
-    parser.add_argument(
-        "--retry-delay", type=float, default=0.001, help="Base delay (seconds) for exponential backoff retries"
-    )
     parser.add_argument(
         "--max-waiting-retries", type=int, default=3, help="Maximum number of retries for waiting nodes to be started"
     )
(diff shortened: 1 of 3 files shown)


moonseeker1:
The requested retry-policy follow-up is ready in [`moonseeker1/vllm-ascend@89e35fb9`](https://github.com/moonseeker1/vllm-ascend/commit/89e35fb9): HTTP 400 is not retried; other HTTP statuses and transport failures retry with exponential backoff; streaming requests are not replayed after output begins. It also adds regression coverage.

Please cherry-pick `89e35fb9` onto `codex/mooncake-rejected-request-cleanup`. The currently configured account cannot push directly to the `Yuli-yx` fork (GitHub returns 403). The equivalent release-branch commit is already included in #14093.

PR #13949, opened 2026-08-10
[CI] Improve workflow cancellation behavior

MrZ20:
GitHub Actions reevaluates job and step conditions when a workflow is cancelled. Conditions using `always()` remain true during cancellation, which may keep dependent jobs queued or running, especially when custom runners or reusable workflows are involved. This can prevent `concurrency.cancel-in-progress` and normal UI cancellation from promptly terminating an old run.

This PR improves cancellation behavior for non-nightly workflows:

- Replaces necessary failure-handling conditions with `!cancelled()`, allowing jobs such as artifact uploads and failure reports to run after failures while skipping them after cancellation.
- Removes redundant `always()` conditions where jobs should only run after their dependencies succeed.
- Updates PR tests, E2E reporting, pre-commit cache handling, Main2Main, estimated-time updates, and weekly workflows.
- Leaves nightly workflows unchanged because some nightly jobs require unconditional cleanup or coordination.

cancel success: https://github.com/vllm-project/vllm-ascend/actions/runs/31394369115/job/93475456751

- vLLM version: v0.26.0

### DIFF
--- a/.github/workflows/_analyze_failure.yaml
+++ b/.github/workflows/_analyze_failure.yaml
@@ -114,7 +114,7 @@ jobs:
 
       # ---- Step 4: Upload report ----
       - name: Upload failure analysis report
-        if: always()
+        if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v7
         with:
           name: failure-analysis-report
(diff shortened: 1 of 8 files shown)


wangxiyuan (approved):

wangxiyuan (approved):

wangxiyuan:
CI failure doesn't relate to this change.

PR #13800, opened 2026-08-07
[Refactor] Remove MiniMax-M2 linear attn __init__/weight_loader patches

AceCoder0:
## Summary

- Remove the `MiniMaxText01RMSNormTP.__init__` and `weight_loader` monkey-patches from `vllm_ascend/patch/worker/patch_minimax_m2_linear_attn.py`, plus the now-unused imports (`partial`, `nn`, `CustomOp`, TP rank/world helpers).
- The NPU q/k norm fast path (`forward_qk`) monkey-patch is **retained** — this PR only drops the weight-sharding/init patches.

## Why removable

Both removed patches are redundant with upstream vLLM at the pinned commit (0351e9aa, `vllm/model_executor/layers/minimax_rms_norm/rms_norm_tp.py`):

- **`weight_loader`** is byte-for-byte identical to the upstream `@staticmethod MiniMaxText01RMSNormTP.weight_loader` (rms_norm_tp.py:336-350): same `loaded_weight.shape[0] // shard_world_size` slice-and-copy.
- **`__init__`** differs from upstream only by (a) an extra `hidden_size % weight_shard_world` divisibility check and (b) skipping the Lamport fused-allreduce workspace setup. On NPU the workspace setup is a no-op: it is gated by `_MINIMAX_FUSED_AR_RMS_QK = torch.ops._C.minimax_allreduce_rms_qk` (a CUDA csrc op, `None` on NPU builds), and even if reached it is wrapped in try/except with the `lamport_workspace` module importing `cuda` bindings. Weight-sharding semantics are identical to upstream.

The retained `forward_qk` patch is unaffected: it does not reference the removed helpers, and the linear-attention path uses `MiniMaxText01RMSNormTP._forward` (untouched) plus this NPU q/k fast path.

## Test plan

- Environment: Atlas 800I A2 (A3), CANN 9.1.0 snapshot, vllm + vllm-ascend from source.
- Model: `/mnt/weight/MiniMax-M2.7-w8a8-QuaRot`, TP8, `--quantization ascend`, EP enabled.
- With the patches removed: `vllm serve` starts cleanly (`Application startup complete`, no errors), chat-completion smoke returns correct output (`2+2 -> 4`, `finish_reason=stop`).

- vLLM version: v0.26.0
- vLLM main: https://github.com/vllm-project/vllm/commit/58d3918e3ea0a544ffedadad2ba84559e9c51d8f

PR #13992, opened 2026-08-11
[Refactor] Remove MiniMax-M2 linear attn forward_qk patch

AceCoder0:

Remove the `MiniMaxText01RMSNormTP.forward_qk` (NPU q/k norm fast path) monkey-patch from `vllm_ascend/patch/worker/patch_minimax_m2_linear_attn.py`, along with its now-dead detection scaffolding and the imports it alone used (`tensor_model_parallel_all_reduce`, `current_platform`, `logging`). Also drops the corresponding entry from the patch inventory in `vllm_ascend/patch/__init__.py`.

The `__init__` / `weight_loader` patches in the same file are **retained** — they are removed separately in #13800.

### Why removable

The patch is both **formula-equivalent** to upstream and **unreachable** under the target deployment:

- **Formula-equivalent**: `_patched_qk` computes `x * rsqrt(global_var + eps) * w`, where `global_var` is the TP all-reduced mean of local variances recovered from the NPU `npu_rms_norm` inv_rms. This is algebraically identical to upstream `_minimax_qk_norm_tp_eager` (`vllm/model_executor/layers/minimax_rms_norm/rms_norm_tp.py`), which all-reduces `qk_var` then applies `x * rsqrt(global_var + eps) * w`.
- **Unreachable**: `MiniMaxM2Attention.forward` (the patched path) routes all q/k normalization through `torch.ops.vllm.minimax_qk_norm_fusion` → `_minimax_qk_norm_tp_eager`; `forward_qk`/`forward_qkv` is only invoked when `tp_world == 1` (rms_norm_tp.py:406). Under TP8 + PP2 the NPU patch never runs, so removing it is behavior-neutral.

No. Patch removal is behavior-neutral for supported deployments (TP>1).

- Environment: Atlas 800I A2 (A3), CANN 8.5, vllm + vllm-ascend from source.
- Model: `/mnt/weight/MiniMax-M2.7-w8a8-QuaRot`, TP8, EP enabled, `--quantization ascend`, `--reasoning-parser minimax_m2`, `--compilation-config '{"cudagraph_mode": "FULL_DECODE_ONLY"}'`.
- With the patch removed: `vllm serve` starts cleanly (`Application startup complete`, no errors), chat-completion smoke returns correct output (`2+2 -> 4`, `finish_reason=stop`).

- vLLM version: v0.27.1


AceCoder0:
Retrigger CI: previous failures were self-hosted runner container errors.

wangxiyuan (approved):

PR #14036, opened 2026-08-11
[BugFix] Fix non-streaming response in load balance proxy server

billsysu:
The load balance proxy server always returns StreamingResponse for both streaming and non-streaming requests. For non-streaming requests (stream=false), this causes Content-Type mismatch: the response has application/json but the body is delivered via chunked transfer encoding instead of a complete JSON body with Content-Length header.

This PR fixes the issue by:
- Streaming requests (stream=true): Keep StreamingResponse with text/event-stream
- Non-streaming requests (stream=false): Collect all chunks from generate_stream() and return a single Response with application/json and proper Content-Length

The same fix is applied to both load_balance_proxy_server_example.py and load_balance_proxy_layerwise_server_example.py.

### DIFF
--- a/examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
+++ b/examples/disaggregated_prefill_v1/load_balance_proxy_layerwise_server_example.py
@@ -98,7 +98,7 @@ from contextlib import asynccontextmanager
 
 import httpx
 from fastapi import FastAPI, HTTPException, Request
-from fastapi.responses import StreamingResponse
+from fastapi.responses import Response, StreamingResponse
 from vllm.logger import init_logger
 
 from vllm_ascend.distributed.kv_transfer.kv_p2p.sfa_pd_rd2h.protocol import (
@@ -519,6 +519,8 @@ async def _handle_completions(api: str, request: Request):
         elif chat_flag:
             messages = req_data["messages"]
             origin_prompt = messages[0].get("content", "")
+            if isinstance(origin_prompt, list):
+                origin_prompt = origin_prompt[0].get("text", "")
         else:
             origin_prompt = ""
         # refer to vLLM sampling_params: max_token default value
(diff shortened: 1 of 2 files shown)


billsysu:
## Update: rewrote as a minimal, focused fix

Rewrote this PR to do **only** the non-streaming fix on top of the latest `main`, addressing the review feedback.

### What changed
- **Non-streaming response**: `generate_stream()` now accumulates chunks until the JSON is complete, then returns a single `Response` with `application/json` + `Content-Length` (instead of a chunked `StreamingResponse`).
- **Bot review (high-priority)**: the `json.JSONDecodeError` on partial chunks is fixed — chunks are accumulated before parsing when `stream=False`, so the `recomputed`/retry logic is no longer skipped.
- **Pre-open decode stream error propagation is PRESERVED** (`_replay_first_chunk` / `preopened_gen` / real HTTP status on initial decode 4xx/5xx/connection failure). It is compatible with the non-streaming `Response` collection: initial decode errors still return the real status; on success the full stream is collected into one JSON `Response`.

### Scope trimmed
The previous version also refactored retry logic, removed the pre-open stream, and changed `reassign_instances`'s signature. Those are reverted here so this PR does one thing. They can be proposed separately if needed.

### Tests (`tests/ut/proxy/test_load_balance_proxy_server.py`)
- Updated the non-streaming cases to use a realistic JSON decode response (`kind=json`) instead of SSE output (a non-stream request gets a single JSON from the decode backend).
- Added `test_nonstream_response_has_content_length` — asserts `Content-Length` present, not chunked, `application/json`, single parseable JSON.
- Added `test_nonstream_assembles_fragmented_json` — splits one JSON across chunks to directly cover the accumulation path.
- Added `test_nonstream_4xx_has_content_length` — asserts the pre-open path still returns the real status with `Content-Length`.

Rebased on latest `main` (no conflicts). The two stale inline review comments about per-chunk `JSONDecodeError` are resolved by the accumulation logic.

Issue #14013, opened 2026-08-11
[Bug]: PA_ND sparse_attn_sharedkv output depends on physical compressed-KV block allocation


QwertyJack:
### Your current environment

<details>
<summary>Sanitized reproducer environment</summary>

```text
Hardware: Ascend 910_93 (A3), TP8
vLLM Ascend: 50e0ce60849a7b6fe6f94e1de085b6fc06e7006d
vLLM: 752a3a504485790a2e8491cacbb35c137339ad34
CANN: 9.0.1, innerversion V100R001C10SPC002B220
torch: 2.10.0
torch_npu: 2.10.0.post2
Python: 3.12.13
Execution path: DeepSeek V4, non-PD, TP8, eager, EP disabled, MTP disabled, prefix caching disabled, DSA overlap disabled
```

The same long-context non-repeatability was also observed on Ascend 910B (A2) with BF16 weights. Customer-specific environment details and prompts are omitted.

</details>

## Summary

`sparse_attn_sharedkv` in the PA_ND compressed-KV path can produce different BF16 outputs when only the physical paged-KV block allocation changes. The logical query, K/V values, sparse indices, and attention metadata remain identical.

This makes DeepSeek V4 logits and selected tokens depend on allocator state. It reproduces without PD separation, graph mode, EP, MTP, prefix caching, or DSpark. Higher concurrency can exercise block reuse and fragmentation more often, but this issue does not make a quantitative failure-rate claim.

The operator defect and its fix are independently demonstrated below. Premature EOS and `stop_without_tool_call` are not used as correctness criteria for this issue because a separate frontend rendering defect in #14035 materially changes that boundary; see [Scope](#scope).

## Root cause

The first incorrect boundary is [`SASVectorBlock<SAST>::CopyInKv`](https://github.com/vllm-project/vllm-ascend/blob/50e0ce60849a7b6fe6f94e1de085b6fc06e7006d/csrc/attention/sparse_attn_sharedkv/op_kernel/arch32/sparse_attn_sharedkv_scfa_block_vector.h#L589-L643).

For a pair of compressed-KV blocks, the PA_ND fast path derives `srcStride` from the absolute difference between two physical GM offsets, starts from the lower physical address, and issues one strided DMA:

```cpp
int64_t blkTableSrcStride =
    Abs(keyOffset1 - keyOffset2) - sparseBlockSize * headDim;
int64_t startGmOffset = MinValid(keyOffset1, keyOffset2);
[...]

PR #14040, opened 2026-08-11
[BugFix] Fix CopyInKv DMA flipping KV block order when physical address order mismatches logical token order

JC-ut0:
## Problem

In paged KV cache architecture, when two KV blocks' physical addresses are in reverse order compared to their logical token order (`keyOffset2 < keyOffset1`), the `CopyInKv` fast path DMA copies blocks starting from the lower physical address, which flips the block order in Unified Buffer. This causes sparse attention to read KV cache in wrong order, producing incorrect logits.

## Symptoms

- Intermittent (~2% rate) infinite token repetition in structured output
- Model enters a loop outputting the same token until `max_tokens` is hit
- `finish_reason: length` instead of `stop`
- More visible with structured output (xgrammar bitmask prevents model from breaking out of the loop by constraining token space)

## Root Cause

The fast path in `CopyInKv` merges two block DMAs into one strided copy. It starts from `min(keyOffset1, keyOffset2)` to ensure contiguous read. However, when `keyOffset2 < keyOffset1`, block2 lands before block1 in UB, reversing their logical order. The attention computation then reads KV in wrong sequence, corrupting the output.

## Fix

Add a physical/logical order consistency check to the fallback condition. When both blocks are valid and `keyOffset2 < keyOffset1` (physical order reversed), force the fallback path (two separate `CopyInSingleKv` calls) to preserve correct logical token order.

## Files Changed

- `sparse_attn_sharedkv_scfa_block_vector.h` (A2/910B, arch32)
- `sparse_flash_attention_service_vector_mla.h` (MLA variant, arch22)
- `kv_quant_sparse_attn_sharedkv_scfa_block_vector.h` (A3/910C, arch35)

## Verification

- Deployed dsv4-flash-w8a8-mtp on A2 pod (8x 910B2C, PD separation)
- **Before fix**: 1/50 requests reproduced (`finish_reason=length`, token 223 infinite repetition)
- **After fix**: 0/500 requests reproduced (all `finish_reason=stop`)
- Debug logs confirmed grammar constraint was working correctly; root cause was attention kernel reading wrong KV order

Related issue: #14013

- vLLM version: v0.26.0

### DIFF
--- a/csrc/attention/sparse_attn_sharedkv/op_kernel/arch32/sparse_attn_sharedkv_scfa_block_vector.h
+++ b/csrc/attention/sparse_attn_sharedkv/op_kernel/arch32/sparse_attn_sharedkv_scfa_block_vector.h
@@ -610,11 +610,15 @@ __aicore__ inline void SASVectorBlock<SAST>::CopyInKv(int64_t &mte2Size, int64_t
         keySrcStride = ((keyOffset1 > keyOffset2 ? (keyOffset1 - keyOffset2) :
 	                    (keyOffset2 - keyOffset1)) - constInfo.sparseBlockSize) * constInfo.headDim * sizeof(KV_T);
     }
+    // PA_ND (paged KV cache) only: check if physical address order
+    // is inconsistent with logical token order
+    constexpr bool isPagedLayout = (KV_LAYOUT_T == SAS_LAYOUT::PA_ND);
     if (unlikely(keySrcStride >= INT32_MAX || keySrcStride < 0 ||
         realS2Idx1 + constInfo.sparseBlockSize >= s2IdLimit ||
-        realS2Idx2 + constInfo.sparseBlockSize >= s2IdLimit)) {
-        // stride溢出、stride为负数、s2超长等异常场景，还原成2条搬运指令
-        // 因为需要拷贝两块
+        realS2Idx2 + constInfo.sparseBlockSize >= s2IdLimit ||
+        (isPagedLayout && keyOffset1 >= 0 && keyOffset2 >= 0 && keyOffset2 < keyOffset1))) {
+        // stride溢出、stride为负数、s2超长、或（PA_ND模式下）物理顺序与逻辑顺序不一致等异常场景，
+        // 还原成2条搬运指令以保持逻辑 token 顺序
         CopyInSingleKv(mte2Size, mte3Size, mergeMte3Idx, realS2Idx1, keyOffset1, s2IdLimit, runInfo);
         CopyInSingleKv(mte2Size, mte3Size, mergeMte3Idx, realS2Idx2, keyOffset2, s2IdLimit, runInfo);
     } else {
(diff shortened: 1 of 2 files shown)


QwertyJack:
Thanks for taking this on. The reversed-physical-address diagnosis matches the operator evidence in #14013, and falling back to logical-order single-block copies is correct for the affected PA_ND path.

Before merging, could we address the following points?

1. **Keep the fix scoped to the paged path.** `sparse_attn_sharedkv_scfa_block_vector.h` supports PA_ND, BSND, and TND, but the new condition currently applies to all three layouts. Please guard it with `if constexpr (KV_LAYOUT_T == SAS_LAYOUT::PA_ND)`. Likewise, the arch22 SFA condition currently affects both `PAGE_ATTENTION=true` and `false`; it should be gated by `PAGE_ATTENTION`. #14013 proves allocation-dependent output for PA_ND. Changing the non-paged reduction order is a separate behavior and performance change that has not been validated here.

2. **Please reconsider the conditional dual/single-DMA implementation for the proven arch32 PA_ND path.** We tested an equivalent conditional strategy and an unconditional PA_ND two-`CopyInSingleKv` strategy with the same real layer-2 input, 10 warmups, 500 synchronized iterations, and independently fragmented CMP/SWA allocations. Both fixes were bitwise allocation-invariant on ranks 0 and 4, but the unconditional implementation was faster:

| Implementation | rank-0 baseline | rank-0 fragmented | rank-4 baseline | rank-4 fragmented |
| --- | ---: | ---: | ---: | ---: |
| Conditional dual/single DMA | 3.347 ms | 3.350 ms | 3.367 ms | 3.368 ms |
| Unconditional two-single-DMA for PA_ND | 3.207 ms | 3.208 ms | 3.208 ms | 3.213 ms |

The unconditional implementation was 4.2%-4.7% faster and is simpler. The complete replay and performance evidence is in the Validation and Performance sections of #14013. These numbers apply to the tested arch32 PA_ND kernel; the arch22 and arch35 changes still need their own build and operator validation.

3. **Please correct the hardware mapping and validation scope.** In the current build mapping, both `ascend910b` and `ascend910_93` use `arch32`; `arch35` maps to `ascend950`. The reported A2 service test validates the arch32 path, not arch35. The PR description should reflect that, and each additional changed kernel should at least be compiled and covered by an allocation-order replay.

4. `#14039` targets `releases/v0.20.2rc`; a separate backport is still needed for the affected `releases/v0.25.1rc` branch.

There are also analogous lower-physical-address-first copies in main's arch35 SFA and quantized SFA files. If this PR intends to fix the pattern generically across operators, those paths should be audited too. If the intended scope is #14013, narrowing this PR to the proven SAS PA_ND path is safer.

JC-ut0:
Thanks @QwertyJack for the thorough review! Updated this PR per your feedback:

**1. Scoped to paged path** — Done. The `keyOffset2 < keyOffset1` check is now guarded by `constexpr isPagedLayout = (KV_LAYOUT_T == SAS_LAYOUT::PA_ND)` in `sparse_attn_sharedkv`, and by `PAGE_ATTENTION` in `sparse_flash_attention`. BSND/TND and non-paged paths are unaffected.

**2. Conditional vs unconditional** — Kept the conditional approach for now. Your benchmark shows unconditional is ~4% faster for PA_ND, which is interesting. We prefer the conservative conditional approach to minimize behavioral change — the fast path is preserved for the common case where physical order is consistent. Happy to switch to unconditional in a follow-up if the community prefers.

**3. arch35 excluded** — Removed the `kv_quant_sparse_attn_sharedkv_scfa_block_vector.h` (arch35/ascend950) change from this PR. It needs its own build and allocation-order replay validation. This PR now only touches arch32 (A2/910B) and arch22 (MLA), both validated.

**4. Backport to v0.25.1rc** — Noted. Will do a separate backport PR after this one merges.

The verification was done on arch32 (A2/910B2C) with dsv4-flash-w8a8-mtp: 1/50 reproduction before fix, 0/500 after fix (tested on both D0 direct and a separate service via conductor).

PR #14084, opened 2026-08-12
[Feature] Warn when a deprecated env var conflicts with additional_config

### LINKED ISSUE #14067: [Feature]: Warn when a deprecated env var disagrees with its additional_config replacement
### 🚀 The feature, motivation and pitch

#9064 introduced the `additional_config` > deprecated env var > default precedence for the migrated options, and `AscendConfig._get_config_value` logs when the env var is used as a fallback. What it doesn't log is the opposite case: when both are set and they disagree, `additional_config` silently wins and the exported value disappears without a word.

That is the confusing case in practice, because the env var usually isn't where the user is looking — it is baked into an image or a launch script from before the migration. Someone who adds `--additional-config '{"weight_nz_mode": 0}'` on top of an image that still exports `VLLM_ASCEND_ENABLE_NZ=1` gets the behaviour they asked for, but a colleague reading the launch script sees `=1` and reasonably concludes NZ is on. The reverse is worse: people who set the env var expecting it to take effect get no hint that something else overrode it.

I ran into this while fixing three call sites that still read the env vars directly instead of the migrated keys (#14061, #14064), which is the same failure mode one layer down.

Proposal: in `_get_config_value`, when the env var is present in `os.environ` and its value differs from the `additional_config` value, log a warning naming both values and stating which one wins. No precedence change, no behaviour change — the existing info log and the quiet paths (values agree, or only one source set) stay exactly as they are, so it is a few lines plus tests.

I have this implemented with unit tests and am happy to open a PR, but I'd rather check the direction first — @lizy124 as the author of #9064, does this fit what you had in mind for the migration period, or would you prefer it stays quiet?

### Alternatives

Raising an error on conflict would be stricter but would break setups that are working today. Warning on every env-var use (even without a conflict) would be noisier than the current info log, which already covers the fallback case.

### Additional context

_No response_

MrlixiangWE:

#9064 set the precedence for the migrated options to `additional_config` > deprecated env var > default, and `_get_config_value` already logs when the env var is used as a fallback. The opposite case is silent: when both are set and they disagree, `additional_config` wins and the exported value disappears without a word.

That is the case people actually trip over, because the env var is usually not where they are looking — it is baked into an image or a launch script from before the migration. Adding `--additional-config '{"weight_nz_mode": 0}'` on top of an image that still exports `VLLM_ASCEND_ENABLE_NZ=1` does the right thing, but anyone reading the launch script concludes the opposite; and someone who sets only the env var expecting it to take effect gets no hint that something overrode it.

This logs a warning naming both values and which one wins, only when they disagree. No precedence change: the existing info log stays, and the quiet paths (values agree, or only one source is set) are untouched, so the existing migration tests pass unchanged.

Closes #14067. I wrote up the motivation there and asked whether this is wanted before sending code; opening the PR as well since that seems to be the usual entry point here. Happy to close it if you would rather not carry the extra log.

One additional warning line at startup, and only for setups where a deprecated env var and its `additional_config` replacement disagree. Behaviour is unchanged.

Two unit tests in `tests/ut/test_ascend_config.py`: one asserts the warning when `weight_nz_mode=0` meets `VLLM_ASCEND_ENABLE_NZ=1`, the other asserts nothing is logged when the two agree. The first fails without this change. `pytest -sv tests/ut/test_ascend_config.py` passes locally on CPU (42 tests). I don't have Ascend hardware, so I'm relying on the NPU CI for e2e coverage.
- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/test_ascend_config.py
+++ b/tests/ut/test_ascend_config.py
@@ -401,6 +401,36 @@ class TestAscendConfig(TestBase):
         mock_info_once.assert_any_call("AscendConfig.weight_nz_mode is set from additional_config with value 1.")
 
     @_clean_up_ascend_config
+    @patch("vllm_ascend.ascend_config.logger.warning_once")
+    @patch("vllm_ascend.platform.NPUPlatform.check_and_update_config")
+    def test_migrated_config_warns_when_env_conflicts(self, mock_fix_incompatible_config, mock_warning_once):
+        test_vllm_config = VllmConfig()
+        test_vllm_config.additional_config = {"weight_nz_mode": 0}
+        with patch.dict(os.environ, {"VLLM_ASCEND_ENABLE_NZ": "1"}, clear=True):
+            ascend_config = init_ascend_config(test_vllm_config)
+
+        self.assertEqual(ascend_config.weight_nz_mode, 0)
+        mock_warning_once.assert_any_call(
+            "AscendConfig.weight_nz_mode is set to 0 by additional_config, but environment variable "
+            "VLLM_ASCEND_ENABLE_NZ requests 1. additional_config takes precedence; unset VLLM_ASCEND_ENABLE_NZ "
+            "to silence this warning, because it will be removed in the next release."
+        )
+
+    @_clean_up_ascend_config
+    @patch("vllm_ascend.ascend_config.logger.warning_once")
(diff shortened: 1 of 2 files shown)


MrlixiangWE:
cc @linfeng-yuan since you reviewed and merged #9064 -- this only adds a warning on top of the precedence you set there, no behaviour change. Motivation is written up in #14067.

PR #14098, opened 2026-08-12
[BugFix][Triton] fix rejection sampler when target_logits are NaN

SOMEONEUNSEEN:

This PR fixes an issue in the rejection sampler where `target_logits` containing `NaN` or `Inf` values could cause errors or incorrect sampling. It cleans up `NaN` and `Inf` values in `target_logits` before applying logits processors.

No.

- vLLM version: v0.26.0

### DIFF
--- a/vllm_ascend/sample/rejection_sampler.py
+++ b/vllm_ascend/sample/rejection_sampler.py
@@ -230,10 +230,11 @@ class AscendRejectionSampler(RejectionSampler):
         if torch.isinf(target_logits).any():
             target_logits[target_logits == torch.inf] = info.max
             target_logits[target_logits == -torch.inf] = info.min
+            logger.warning("[sample/rejection_sampler] target_logits have infinite values.")
         nan_mask = torch.isnan(target_logits)
         if nan_mask.any():
             target_logits[nan_mask] = 0
-            logger.warning_once("[sample/rejection_sampler] target_logits have NaN.")
+            logger.warning("[sample/rejection_sampler] target_logits have NaN.")
 
         target_logits = self.apply_logits_processors(target_logits, sampling_metadata, metadata)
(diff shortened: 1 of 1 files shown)


SOMEONEUNSEEN:
[Bot]: cherry-pick completed successfully. New PR created: https://github.com/vllm-project/vllm-ascend/pull/14120

jianzs (changes requested):
"target_logits contains NaN or INF" — isn't that a precision issue?

SOMEONEUNSEEN:
> "target_logits contains NaN or INF" — isn't that a precision issue?

Thank you for your review. This PR is a preventive measure taken at the initial stage of sampling to intercept and prevent occasional NaNs from causing fatal infinite loop results. The root cause of the NaNs will be addressed in another PR for repair.

jianzs:
> > "target_logits contains NaN or INF" — isn't that a precision issue?
> 
> Thank you for your review. This PR is a preventive measure taken at the initial stage of sampling to intercept and prevent occasional NaNs from causing fatal infinite loop results. The root cause of the NaNs will be addressed in another PR for repair.

It's better to tackling the root cause directly instead of introducing an ostrich-like workaround that hides the problem.

SOMEONEUNSEEN:
> > > "target_logits contains NaN or INF" — isn't that a precision issue?
> > 
> > 
> > Thank you for your review. This PR is a preventive measure taken at the initial stage of sampling to intercept and prevent occasional NaNs from causing fatal infinite loop results. The root cause of the NaNs will be addressed in another PR for repair.
> 
> It's better to tackling the root cause directly instead of introducing an ostrich-like workaround that hides the problem.

You're right, this will mask other precision issues. I'll reconsider the fix on the main branch.

PR #14132, opened 2026-08-12
[BugFix][KVTransfer] Avoid co-queueing save with load in ascend store…

Wfd567:

This PR ports the root-cause fix from vllm-project/vllm#43371 to the
Ascend KV store (`vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store`).

**Root cause**

In `kv_both` mode, when a request can load its KV from the pool
(`load_spec.can_load == True`), `ReqMeta.from_request_tracker` used to
generate a ReqMeta that carried **both a save AND a load** for the same
request. As a result:

- The same `req_id` was queued into both the send (save) and recv (load)
  worker threads.
- The request's blocks were double delayed-freed, and the `req_id` could
  get stuck in the delayed-free whitelist forever — the blocks were never
  released, and the scheduler eventually crashed with
  `assert req_id in self.requests` on subsequent requests.

This matches the failures reported in vllm-project/vllm-ascend#13628 and
vllm-project/vllm#46240.

**Fix**

In `ReqMeta.from_request_tracker`, force `skip_save = True` whenever
`load_spec.can_load` is set, so a ReqMeta never carries both a save and a
load. This mirrors the invariant enforced upstream in #43371.

**Performance impact**

No regression. On a cache hit, the suppressed save was pure wasted work:
the bytes are being read out of the store right now, and re-saving them
would even hit a READABLE blob and fail with `MMC_UNMATCHED_KEY`.
Requests that cannot load, and pure-save nodes, are unaffected.
Partial-hit requests keep saving newly prefilled tokens through the later
`cached_reqs` path.

**Changes**

- `vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/metadata.py`:
  core fix in `ReqMeta.from_request_tracker`.
- `tests/ut/distributed/ascend_store/test_metadata.py`: new unit test
  `test_from_request_tracker_can_load_suppresses_save` asserting
  `can_load=True => can_save=False` with the `load_spec` kept.

No. The change is internal to the Ascend KV store's ReqMeta generation and
does not alter any API, CLI, or runtime configuration.

- Added unit test `test_from_request_tracker_can_load_suppresses_save`
  following the existing test style in
  `tests/ut/distributed/ascend_store/test_metadata.py`.
- Ran `python -m py_compile` on both modified files: passed.
- Branch-behavior simulation covering pure-save / full-hit / partial-hit /
  pure-load / skip_save combinations: all pass.
- (TODO before merge) `pytest tests/ut/distributed/ascend_store/test_metadata.py`
  will be run in the CI environment; the local machine has no
  torch/vllm installed, so the unit test could not be executed locally.


Wfd567:
@xueliangyang-oeuler

RonnieWang0223:
Great fix! This is a precise and minimal change that correctly addresses the root cause of the double delayed-free crash in kv_both mode. Forcing skip_save=True when can_load=True at the from_request_tracker entry point is the right place to enforce the mutual exclusivity of save and load, and the unit test covers the core assertion well.

One suggestion to strengthen the test coverage:

The inline comment mentions:
 Later cached_reqs steps save new tokens normally.
This implies that in subsequent steps (where load_spec is None or can_load=False ), can_save should resume being True so the incremental KV can still be saved to the pool. However, the current test only verifies the "suppress save" behavior in the load-hit step, and does not cover the "resume save" behavior in the follow-up step.

Angazenn (approved):

Wfd567:
Sorry to bother you both. Could you please take a look at this PR when you get a chance? Thanks! @LCAIZJ

Wfd567:
@LCAIZJ Could you please add the 'ready-precise' label to this PR? The CI gate requires it to run the selected tests. Thanks!

ningjingbengxiaohai (dismissed):

ningjingbengxiaohai (approved):

PR #14184, opened 2026-08-13
[Feature] Support MTP with PP in MRV2

LostFox11:

This PR adds generic Model Runner V2 pipeline-parallel support for MTP.

Upstream V2 creates the speculator only on the last pipeline stage. The Ascend runner reinitializes its speculator and previously did so on every stage. In addition, the target sampler output is broadcast before the last stage finishes generating the next-step MTP draft tokens, so non-last stages do not receive the draft state needed by their delayed PP request-state update.

This patch:

- preserves the upstream behavior of running an MTP speculator only on the last PP stage;
- defers the target-token broadcast until MTP drafting completes;
- transports accepted target tokens and next-step draft tokens in the same existing V2 `PPHandler` queue slot;
- updates draft state before the existing upstream `update_pp_decode_requests()` path postprocesses sampled tokens;
- reuses the existing PP queue, generation counter, and request filtering flow.

The behavior is gated only by `method == "mtp"` and `pipeline_parallel_size > 1`. It does not contain model-specific gates and does not enable Eagle3.

Yes. MTP models can be combined with pipeline parallelism under Model Runner V2.

- `python3 -m py_compile vllm_ascend/patch/worker/patch_v2/patch_spec_pp.py vllm_ascend/worker/v2/model_runner.py`
- `git diff --check`
- Ran a lightweight PPHandler behavior check in an Ascend container covering:
  - combined target/draft fixed-shape payload construction;
  - deferred broadcast ordering;
  - draft-state updates when requests are freed or excluded.
- The MTP PP path was previously validated with a 16-NPU TP8+PP2 DeepSeek-V4 service using two successful chat requests. Other MTP model families and the exact generic commit have not yet been rerun end to end.

- vLLM version: v0.27.1


LostFox11:
/rerun
[Bot]: rerun completed.

**Failed:**

zzzzwwjj (approved):

LostFox11:
/rerun
[Bot]: rerun completed.

**Failed:**

ZT-AIA (approved):

PR #14195, opened 2026-08-13
[CI] Add reproducible CPU-only build verification

ChenDAO1:

This PR adds a reproducible CPU-only package build and verification path for `vllm-ascend`. It validates that the package can be installed and built without Ascend NPU devices or the CANN toolkit, while preserving the normal NPU/runtime dependency contract.

Specifically, this PR:

- adds `Dockerfile.cpu`, based on openEuler 24.03 LTS, as a clean CPU-only build environment;
- adds `tools/cpu_only_build.sh` as the executable source of truth for dependency installation, package build, and verification;
- installs exact `+cpu` PyTorch wheels while resolving their generic dependencies from the main package index, then installs the complete `pyproject.toml` build-system requirements before building the package;
- builds with `COMPILE_CUSTOM_KERNELS=0`, an explicit `SOC_VERSION`, and `TORCH_DEVICE_BACKEND_AUTOLOAD=0`, so the verification path does not require CANN or NPU hardware;
- keeps `arctic-inference==0.1.1` in `requirements.txt` and therefore in the normal vLLM Ascend runtime dependency contract;
- excludes only the exact `arctic-inference==0.1.1` requirement from CPU-only verification because its source distribution is incompatible with the CPU-only build environment;
- filters Arctic Inference both when installing CPU-only dependencies and while `setup.py` generates editable-install package metadata, restoring the original `requirements.txt` through a shell trap on success or failure;
- verifies that no NPU device or CANN installation is visible, CPU PyTorch cannot access CUDA, Arctic Inference is absent only from the CPU-only environment, the package metadata is available, and `pip check` succeeds;
- adds a path-filtered GitHub Actions workflow on the `linux-amd64-cpu-4` self-hosted runner;
- uses the runner's BuildKit path through `docker/build-push-action@v7`, with CPU-only verification executed as a Dockerfile build layer so the job does not require a local Docker daemon or `docker run`;
- pulls the openEuler base image through the self-hosted runner's reachable public SWR proxy and routes openEuler DNF, PyPI, CPU PyTorch, and Ascend package traffic through CI caches, while retaining Docker Hub and public-index defaults for local builds;
- updates installation and suffix speculative decoding documentation, plus dependency-contract regression tests.

`arctic-inference` was removed only from `pyproject.toml` build-system requirements because it is not required to execute the package build backend. Normal installations still receive the exact runtime dependency through `requirements.txt` and `setup.py`.

No user-facing API or NPU runtime behavior changes are introduced.

Normal vLLM Ascend installations continue to include `arctic-inference==0.1.1`, and suffix speculative decoding keeps its existing dependency. The Arctic Inference exclusion is local to the CPU-only CI verification path and does not affect normal NPU installations.

The patch was validated with:

- `bash -n tools/cpu_only_build.sh`;
- `git diff --check`;
- actionlint for `.github/workflows/cpu_only_build.yaml`;
- `python -m unittest tests.ut.test_dependency_documentation -v`;
- CPU-only dependency, build, metadata, device/CANN absence, Arctic Inference absence, and `pip check` verification encoded in `Dockerfile.cpu` and `tools/cpu_only_build.sh`;
- the public SWR openEuler manifest and the DNF repository rewrite used by the self-hosted BuildKit path;
- GitHub Actions CPU-only BuildKit verification passed on the current PR head; the repository E2E selected tests were also triggered.

- vLLM version: v0.27.1


requirements.txt
```diff
@@ -34,6 +34,5 @@ numba
 memfabric_hybrid
 memcache_hybrid
 
-arctic-inference==0.1.1
```
MrZ20, line 37: Make sure the test case does not include it; otherwise, direct deletion is not recommended.

.github/workflows/cpu_only_build.yaml
```diff
@@ -0,0 +1,67 @@
...
+# clean Docker image rather than the CANN-based image used by NPU tests.
+name: CPU-only build verification
+
+on:
+  pull_request:
+    branches:
+      - main
+      - '*-dev'
+      - 'releases/v*'
+    paths:
```
MrZ20, line 11: Please determine the scope in which this test is triggered. At minimum, I believe the following items should not be included in the trigger scope; otherwise, this test may become a per‑pr test, while its intended purpose is to perform validation only when specific content is modified.
```yaml
      - 'vllm_ascend/**'
      - 'tests/ut/test_dependency_documentation.py'
      - '.github/workflows/scripts/test_config.yaml'
      - '.github/workflows/scripts/select_tests.py'
      - 'docs/source/installation.md'
      - 'docs/source/tutorials/features/suffix_speculative_decoding.md'
      - 'docs/source/user_guide/feature_guide/speculative_decoding.md'
```

ChenDAO1: Thanks for pointing this out. I agree that the current path filter is broader than necessary and may cause the CPU-only build verification to run for unrelated PRs. I'll narrow the trigger scope to dependency/build-related files. For vllm_ascend/**, I'll keep only the source files that are directly imported during package build, such as vllm_ascend/envs.py and vllm_ascend/device/hardware.py.

ChenDAO1:
/rerun
[Bot]: rerun completed.

**Failed:**

ChenDAO1:
/cancel
[Bot]: successfully force-cancelled the following runs:


https://github.com/vllm-project/vllm-ascend/actions/runs/32207228102

ChenDAO1:
/rerun
[Bot]: rerun completed.

**Failed:**

requirements.txt
ChenDAO1: I will Keeps arctic-inference==0.1.1 in the normal vLLM Ascend dependency contract, while excluding it only from the CPU-only build verification path because it is incompatible with the CPU-only PyTorch build environment.

.github/workflows/cpu_only_build.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - 'vllm_ascend/device/hardware.py'
+      - '.github/workflows/cpu_only_build.yaml'
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Build and smoke-test CPU image
+    runs-on: ubuntu-latest
```
zhangxinyuehfad: Plz use self-host runner, don't use github runner.

ChenDAO1: already fixed this

PR #14197, opened 2026-08-13
[Misc][Bugfix] Fix triton ops rope_forward_triton fails in CANN9.2 950PR

AuroraEmiya:
This PR fix a error in CANN9.2.0 950PR ,which fails in previous tiling size.

- vLLM version: v0.23.0


AuroraEmiya:
@ningjingbengxiaohai pls review this change

PR #14210, opened 2026-08-13
[BugFix][Model] Fix MiniMax-M3 shared vision tower pruning

muziyuhui666:

This PR fixes incorrect Vision Tower pruning for MiniMax-M3 when only one multimodal input type is enabled.

MiniMax-M3 image and video inputs share the same `MiniMaxVLVisionModel`. However, the current implementation registers image and video as two independent tower targets:

```python
tower_targets={
    "image": MiniMaxVLVisionModel,
    "video": MiniMaxVLVisionModel,
}
```

Since both modalities point to the same shared Vision Tower, disabling either modality can incorrectly mark the shared tower as missing even when the other modality is still enabled. For example:

```bash
--limit-mm-per-prompt '{"image":1,"video":0}'
```

should retain the Vision Tower because image input is still enabled, but the previous model marking could prune the shared `MiniMaxVLVisionModel`, which then causes Vision Tower weight loading to fail.

This PR represents the MiniMax-M3 model topology explicitly by marking the language model and the shared multimodal tower separately:

```python
with (
    self._mark_language_model(
        vllm_config,
        targets=MiniMaxM3SparseForCausalLM,
    ),
    self._mark_tower_model(
        vllm_config,
        {"image", "video"},
        targets=MiniMaxVLVisionModel,
    ),
):
```

With this change, the shared Vision Tower follows the expected behavior:

| Image | Video | Vision Tower |
| --- | --- | --- |
| enabled | enabled | retained |
| enabled | disabled | retained |
| disabled | enabled | retained |
| disabled | disabled | pruned |

The MiniMax-M3 documentation is also updated to explicitly disable unused modalities, for example:

```bash
# Image-only
--limit-mm-per-prompt '{"image":1,"video":0}'

# Video-only
--limit-mm-per-prompt '{"image":0,"video":1}'

# Image + video
--limit-mm-per-prompt '{"image":1,"video":1}'
```

This is preferable to omitting an unused modality because omitted modalities keep their default per-prompt limit and may still participate in multimodal capacity and profiling planning.

Yes.

There is no API change, but this PR fixes MiniMax-M3 multimodal serving behavior when image or video is explicitly disabled through `--limit-mm-per-prompt`.

Users can now independently configure:

```bash
--limit-mm-per-prompt '{"image":1,"video":0}'
```

or:

```bash
--limit-mm-per-prompt '{"image":0,"video":1}'
```

without incorrectly pruning the shared Vision Tower while the other modality remains enabled.

The MiniMax-M3 deployment and evaluation examples are updated accordingly.

Added the unit test:

```text
tests/ut/models/minimax_m3/test_minimax_m3_vit.py::TestMiniMaxM3VitProcessor::test_shared_vision_tower_pruning_follows_image_video_limits
```

The regression test covers all four image/video configurations:

```text
image=1, video=1 -> Vision Tower retained
image=1, video=0 -> Vision Tower retained
image=0, video=1 -> Vision Tower retained
image=0, video=0 -> Vision Tower pruned
```

The test uses lightweight mocked Vision Tower and language model modules, so it validates the model-marking and pruning behavior without loading the full MiniMax-M3 checkpoint.

The image-only configuration was also validated on Ascend with:

```bash
--limit-mm-per-prompt '{"image":1,"video":0}'
```

and the shared Vision Tower was retained and its weights were loaded correctly.

Documentation builds for both English and Chinese documentation also pass.

- vLLM version: v0.27.1

### DIFF
--- a/docs/source/tutorials/models/MiniMax-M3.md
+++ b/docs/source/tutorials/models/MiniMax-M3.md
@@ -131,7 +131,7 @@ Single-node deployment completes both Prefill and Decode within the same node. B
     --distributed_executor_backend "mp" \
     --gpu-memory-utilization 0.92 \
     --reasoning-parser minimax_m3 \
-    --limit-mm-per-prompt '{"image":1}' \
+    --limit-mm-per-prompt '{"image":1,"video":0}' \
     --compilation-config '{"cudagraph_mode":"FULL_DECODE_ONLY"}' \
     --additional-config '{
         "enable_cpu_binding": true,
@@ -167,7 +167,7 @@ Single-node deployment completes both Prefill and Decode within the same node. B
   --distributed_executor_backend "mp" \
   --gpu-memory-utilization 0.92 \
   --reasoning-parser minimax_m3 \
-  --limit-mm-per-prompt '{"image":1}' \
+  --limit-mm-per-prompt '{"image":1,"video":0}' \
   --speculative-config '{"model":"${EAGLE3_WEIGHT_PATH}", "method":"eagle3", "num_speculative_tokens":3}' \
   --compilation-config '{"cudagraph_mode":"FULL_DECODE_ONLY"}' \
   --additional-config '{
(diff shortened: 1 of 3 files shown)


linfeng-yuan (commented):
Please attach upstream modifcation pr.

muziyuhui666:
> Please attach upstream modifcation pr.

Thanks. The corresponding upstream vLLM PR is:

https://github.com/vllm-project/vllm/pull/45381

Upstream MiniMax-M3 already models image and video as modalities sharing the same vision tower via _mark_tower_model(vllm_config, {"image", "video"}).

This PR aligns the vllm-ascend MiniMax-M3 implementation with the upstream shared vision tower semantics.

linfeng-yuan (approved):

PR #14218, opened 2026-08-13
[Performance] Overlap hidden all-gather with local Q projection

jwkkkkkkdada:

This PR reorganizes the DSA-CP forward scheduling to overlap the
hidden-state all-gather with local Q projection.

Previously, the hidden-state all-gather completed before local Q
projection started, although the Q path only depends on
`hidden_states_local`.

This change:
- starts the hidden-state all-gather asynchronously;
- computes local Q projection while the collective is in flight;
- waits immediately before the first consumer of the gathered hidden
  states;
- preserves the original unpadding behavior;
- completes the hidden-state collective before starting the O-projection
  weight collective.

The numerical computation is unchanged. MindStudio traces show partial
communication/computation overlap. A residual communication wait remains
because the all-gather duration is longer than the local-Q computation
window.

No API, configuration, or numerical-output changes are introduced.
The internal DSA-CP execution schedule is changed to allow
communication/computation overlap.

- Tested on an 8-card 910B platform.
- Compared model outputs with the original synchronous implementation.
- Tested eager mode with `--enforce-eager`.
- Tested ACL Graph mode with
  `cudagraph_mode=FULL_DECODE_ONLY`.
- Verified repeated graph replay without collective hangs or output
  differences.
- Verified communication/computation overlap using MindStudio.
- Ran benchmark comparisons and observed no performance regression.


jwkkkkkkdada:
@weijinqian0 Hi, this PR is ready for review. Could you please take a look?

jwkkkkkkdada:
Could someone please add the ready label to trigger the CI tests? Thank you!

PR #14230, opened 2026-08-13
[CI] Hjp 813 verify

winson-00178005:

Temporary test PR to verify the custom ops verification fix in `schedule_daily_regression.yaml` on upstream NPU runners.

This PR adds a temporary `pull_request` trigger to the Daily workflow so it runs on the upstream repo's self-hosted NPU runners. The fix being verified:

- **Before (broken):** `from vllm_ascend import op` (ImportError - module name is `ops`, not `op`)
- **After (fixed):** `import vllm_ascend.vllm_ascend_C` (correct C extension verification, same pattern as `_selected_tests.yaml:250`)

Reference: https://github.com/vllm-project/vllm-ascend/actions/runs/31686379032/job/94405182982

### Does this PR introduce _any* user-facing change?

No. Test-only PR, will be closed after verification.

This IS the test.

- vLLM version: v0.27.1


winson-00178005:
Verification complete. verify-image build gate passed with the fix: `custom ops OK -> vllm_ascend_C loaded`. Closing test PR.

PR #14231, opened 2026-08-13
[BugFix][Model] Enable Kimi K3 projector rotation

Dawn952:

Kimi K3 ModelSlim checkpoints can carry an explicit `mm_projector.rot_proj` rotation matrix. The text path folds the rotation into its input projection, while multimodal vision embeddings require the projector rotation to be applied explicitly after `post_norm`.

The projector rotation is a checkpoint property, not a device-type property. This patch therefore always constructs `rot_proj` on A2, A3, and A5, then enables its forward path only when the checkpoint loader confirms that a `mm_projector.rot_proj.*` parameter was actually loaded. Checkpoints without the rotation tensor retain the original projector behavior.

This patch also adds regression coverage for unconditional layer construction, weight-gated activation, and the projector forward path.

Yes. Kimi K3 multimodal inference applies the ModelSlim projector rotation on A2, A3, and A5 whenever the checkpoint provides the corresponding weight.

- `ruff check vllm_ascend/models/kimi_k3.py tests/ut/models/test_kimi_k3.py`
- `ruff format --check vllm_ascend/models/kimi_k3.py tests/ut/models/test_kimi_k3.py`
- `python -m compileall -q vllm_ascend/models/kimi_k3.py tests/ut/models/test_kimi_k3.py`
- In an A5 development container, from an isolated worktree: `python -m pytest -q tests/ut/models/test_kimi_k3.py` (36 passed before the device-type restriction was removed; the updated regression no longer mocks a device type)

- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/models/test_kimi_k3.py
+++ b/tests/ut/models/test_kimi_k3.py
@@ -168,7 +168,9 @@ def test_kimi_k3_projector_applies_rotation_only_after_weight_load():
     projector.post_norm = nn.Identity()
     image_features = torch.tensor([[1.0, 2.0]])
 
+    projector.rot_proj = ScaleLinear()
     del projector.rot_proj
+    assert not hasattr(projector, "rot_proj")
     torch.testing.assert_close(projector(image_features), image_features)
 
     projector.rot_proj = ScaleLinear()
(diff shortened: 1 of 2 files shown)


Dawn952:
/rerun

linfeng-yuan (commented):
Please delete self.rot_proj if we do not use it

linfeng-yuan (approved):

PR #14250, opened 2026-08-14
[BugFix] allocate MTP hidden buffer only for speculative decoding

845473182:

DeepSeek-V4 currently allocates `_mtp_hidden_buffer` for every deployment, even when speculative decoding is disabled. The buffer is only consumed by the MTP draft path, so in regular inference it permanently occupies NPU memory without being used.

This PR:

- Allocates `_mtp_hidden_buffer` only when `vllm_config.speculative_config` is configured.
- Skips the corresponding hidden-state copy and FlashComm1 all-gather when speculative decoding is disabled.
- Preserves the existing buffer allocation and update behavior when speculative decoding is enabled.

The avoided persistent allocation per rank is:

```
max_num_batched_tokens * hc_mult * hidden_size * dtype_size
```

No API or configuration changes.

For DeepSeek-V4 deployments without speculative decoding, this reduces persistent NPU memory usage and avoids an unnecessary hidden-state copy/all-gather. Speculative decoding behavior is unchanged.

- Ran Python compilation validation for `vllm_ascend/models/deepseek_v4.py`.
- Ran `git diff --check`.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/models/deepseek_v4/model.py
+++ b/vllm_ascend/models/deepseek_v4/model.py
@@ -813,6 +813,7 @@ class DeepseekV4Model(nn.Module, EagleModelMixin):
         # Pre-hc_head residual stream buffer for the MTP draft. Only needed
         # when speculative decoding is enabled; allocating it unconditionally
         # would permanently cost max_num_batched_tokens * hc_dim per rank.
+        spec_config = vllm_config.speculative_config
         self._mtp_hidden_buffer = (
             torch.empty(
                 vllm_config.scheduler_config.max_num_batched_tokens,
@@ -820,7 +821,7 @@ class DeepseekV4Model(nn.Module, EagleModelMixin):
                 dtype=vllm_config.model_config.dtype,
                 device=self.device,
             )
-            if vllm_config.speculative_config is not None
+            if spec_config is not None and spec_config.method == "mtp"
             else None
         )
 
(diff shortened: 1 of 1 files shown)


slippersss (approved):

vllm_ascend/models/deepseek_v4.py
```diff
@@ -1091,14 +1091,18 @@ def make_empty_intermediate_tensors(
...
+        # when speculative decoding is enabled; allocating it unconditionally
+        # would permanently cost max_num_batched_tokens * hc_dim per rank.
+        self._mtp_hidden_buffer = (
+            torch.empty(
+                vllm_config.scheduler_config.max_num_batched_tokens,
+                hc_dim,
+                dtype=vllm_config.model_config.dtype,
+                device=self.device,
+            )
+            if vllm_config.speculative_config is not None
```
linfeng-yuan: Please check whether both mtp and dspark use this buffer?

845473182: it is confirmed that only mtp use this buffer. Also we can adjust this determine statement to make it initialize buffer only in mtp.

845473182:
/rerun
[Bot]: rerun completed.

**Failed:**

845473182:
/rerun
[Bot]: rerun completed.

**Failed:**

Tflowers-0129 (approved):

Tflowers-0129:
I left a non-blocking suggestion for PP + MTP, where the buffer could potentially be restricted to the last PP stage, as well as a suggestion to add regression coverage for the allocation condition.

HF-001:
@Tflowers-0129 @845473182 hi, it seems that this PR has introduced a bug. Please refer to: https://github.com/vllm-project/vllm-ascend/pull/14680

PR #14286, opened 2026-08-14
[MRV2] Support profiling chunk in model runner v2

dianWang325:

This PR adapts profiling-based chunk calculation for model runner v2.

The changes:

- Route CPP startup profiling through the model runner v2 dummy-run path.
- Propagate model execution timing through asynchronous model outputs.
- Force `CUDAGraphMode.NONE` during profiling dummy runs without mutating the captured graph state.
- Add a two-card profiling chunk performance test using PP=2 and TP=1.
- Keep coverage for the existing four-card PP=2 and TP=2 configuration.

Related to #5208.

This PR only adapts internal profiling and timing propagation for Model Runner V2. It does not change any user-facing API, configuration, CLI option, or default behavior.

CPP-related unit tests:

- 158 passed
- 1 skipped

Four-card E2E test:

```bash
pytest -sv tests/e2e/pull_request/four_card/test_profiling_chunk_performance.py

Result:
- Passed on 4 Ascend NPUs
- Request time was approximately 4.3 seconds
- Performance baseline: 5.2 seconds

Two-card E2E test:
```bash
pytest -sv tests/e2e/pull_request/two_card/test_profiling_chunk_performance.py

Result:
- Passed on 2 Ascend NPUs
- TTFT measurements: 7.02s, 7.05s, 7.03s, 7.05s, 7.04s
- Median TTFT: 7.04 seconds
- Performance baseline: 8.0 seconds


- vLLM version: v0.27.1

### DIFF
--- a/tests/ut/worker/test_model_runner_v2.py
+++ b/tests/ut/worker/test_model_runner_v2.py
@@ -18,7 +18,8 @@ def _make_runner(need_timing: bool = True):
     return runner
 
 
-def test_execute_model_records_profiling_time():
+@pytest.mark.parametrize("is_vllm_0_27_1", [True, False], ids=["v0.27.1", "newer"])
+def test_execute_model_records_profiling_time(is_vllm_0_27_1):
     runner = _make_runner()
     scheduler_output = SimpleNamespace(disable_profiling_timing=False)
 
@@ -28,6 +29,10 @@ def test_execute_model_records_profiling_time():
             "execute_model",
             return_value=None,
         ) as mock_execute_model,
+        patch(
+            "vllm_ascend.worker.v2.model_runner.vllm_version_is",
+            return_value=is_vllm_0_27_1,
+        ),
         patch("vllm_ascend.core.profiling_chunk_predictor.torch.npu.synchronize") as mock_synchronize,
         patch(
             "vllm_ascend.core.profiling_chunk_predictor.time.perf_counter",
@@ -39,13 +44,15 @@ def test_execute_model_records_profiling_time():
     assert output is None
     assert runner._cpp_execution_time_ms == pytest.approx(125.0)
     assert mock_synchronize.call_count == 2
-    mock_execute_model.assert_called_once_with(
-        scheduler_output,
-        intermediate_tensors=None,
-        dummy_run=False,
-        skip_attn_for_dummy_run=False,
-        is_profile=False,
-    )
+    expected_kwargs = {
+        "intermediate_tensors": None,
+        "dummy_run": False,
+        "skip_attn_for_dummy_run": False,
+        "is_profile": False,
+    }
+    if not is_vllm_0_27_1:
+        expected_kwargs["context_len"] = 0
+    mock_execute_model.assert_called_once_with(scheduler_output, **expected_kwargs)
 
 
 def test_execute_model_disables_profiling_timer_and_clears_stale_time():
--- a/tests/ut/worker/test_model_runner_v2.py
+++ b/tests/ut/worker/test_model_runner_v2.py
@@ -44,7 +44,7 @@ def test_execute_model_records_profiling_time(is_vllm_0_27_1):
     assert output is None
     assert runner._cpp_execution_time_ms == pytest.approx(125.0)
     assert mock_synchronize.call_count == 2
-    expected_kwargs = {
+    expected_kwargs: dict[str, object] = {
         "intermediate_tensors": None,
         "dummy_run": False,
         "skip_attn_for_dummy_run": False,
--- a/tests/ut/worker/test_model_runner_v2.py
+++ b/tests/ut/worker/test_model_runner_v2.py
@@ -28,10 +28,6 @@ def test_execute_model_records_profiling_time():
             "execute_model",
             return_value=None,
         ) as mock_execute_model,
-        patch(
-            "vllm_ascend.worker.v2.model_runner.enable_sp",
-            return_value=False,
-        ),
         patch("vllm_ascend.core.profiling_chunk_predictor.torch.npu.synchronize") as mock_synchronize,
         patch(
             "vllm_ascend.core.profiling_chunk_predictor.time.perf_counter",
@@ -62,10 +58,6 @@ def test_execute_model_disables_profiling_timer_and_clears_stale_time():
             GPUModelRunner,
             "execute_model",
             return_value=None,
-        ),
-        patch(
-            "vllm_ascend.worker.v2.model_runner.enable_sp",
-            return_value=False,
         ),
         patch("vllm_ascend.core.profiling_chunk_predictor.torch.npu.synchronize") as mock_synchronize,
         patch("vllm_ascend.core.profiling_chunk_predictor.time.perf_counter") as mock_perf_counter,
--- a/tests/ut/worker/test_model_runner_v2.py
+++ b/tests/ut/worker/test_model_runner_v2.py
@@ -77,7 +77,3 @@ def test_execute_model_disables_profiling_timer_and_clears_stale_time():
     assert runner._cpp_execution_time_ms is None
     mock_synchronize.assert_not_called()
     mock_perf_counter.assert_not_called()
-
-
-def test_sample_tokens_is_inherited_from_upstream():
-    assert NPUModelRunner.sample_tokens is GPUModelRunner.sample_tokens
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: vllm_ascend/worker/v2/model_runner.py [hunk 1]
```diff
@@ -228,6 +229,12 @@ def execute_model(
         skip_attn_for_dummy_run: bool = False,
         is_profile: bool = False,
     ):
+        if self.ascend_config.scheduler_config.profiling_chunk_config.need_timing:
```
weiguihua2: Is it necessary to check whether profiling_chunk_config is None?

vllm_ascend/worker/worker.py
```diff
@@ -857,16 +858,27 @@ def profile_prefill_latency(self, num_tokens: int) -> float:
...
-        # won't run, making profiling results inaccurate.
-        # _dummy_run handles PP internally (intermediate tensors, etc.)
-        self.model_runner._dummy_run(
-            num_tokens=num_tokens,
-            force_attention=True,  # Critical: ensure attention is executed
-            profile_cpp=True,
-        )
+        # Run a real model forward with attention enabled in eager mode.
+        # _dummy_run handles PP internally (intermediate tensors, etc.).
+        old_max_num_reqs = self.model_runner.max_num_reqs
```
weiguihua2, line 880: This kind of judgment only exists in the cpp scenario, right?

vllm_ascend/worker/v2/model_runner.py
dianWang325: SchedulerConfig.__init__ always initializes profiling_chunk_config as a ProfilingChunkConfig instance. When no configuration is provided, it is constructed from an empty dict, with enabled=False and need_timing=False, rather than being set to None.

This is also consistent with the existing Model Runner V1 implementation, which accesses need_timing directly.

vllm_ascend/worker/worker.py
dianWang325: Yes. This check is only used in the CPP startup profiling path.

profile_prefill_latency() is called by ProfilingChunkScheduler through collective_rpc. The signature check is used to keep compatibility between the two model runners: Model Runner V1 supports the profile_cpp argument, while Model Runner V2 does not. Therefore, we pass profile_cpp=True for V1 and omit it for V2.

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -263,6 +270,25 @@ def execute_model(
...
+        if self.ascend_config.scheduler_config.profiling_chunk_config.need_timing and hasattr(
+            self, "_execution_start_time"
+        ):
+            torch.npu.synchronize()
+            execution_time_ms = (time.perf_counter() - self._execution_start_time) * 1000.0
+
+            # MRV2 normally returns AsyncOutput on the last PP rank.
+            # Preserve the timing on its inner ModelRunnerOutput so it reaches
+            # the scheduler after get_output().
+            model_runner_output = getattr(output, "model_runner_output", output)
```
weiguihua2: Extract a method for the logic in cpp.

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -263,6 +274,16 @@ def execute_model(
 
         return output
 
+    @torch.inference_mode()
+    def sample_tokens(self, grammar_output):
```
gjc0824: Is this because the current vllm-ascend does not override the `sample_tokens` method in MRV2?

dianWang325: Yes. Before this PR, the Ascend MRV2 runner did not override sample_tokens() and inherited the implementation from the upstream GPUModelRunner.

This PR adds a thin override only to attach the CPP execution timing to that output; the sampling behavior itself is still delegated to super().sample_tokens().

gjc0824: I think you can also move this record_time function into the end of `execute_model`. This patch can be removed.

dianWang325:
/rerun
[Bot]: rerun completed.

**Failed:**

tests/ut/worker/test_model_runner_v2.py
```diff
@@ -0,0 +1,83 @@
...
+        runner.execute_model(scheduler_output)
+
+    profiling_config = runner.ascend_config.scheduler_config.profiling_chunk_config
+    assert not profiling_config.need_timing
+    assert runner._cpp_execution_time_ms is None
+    mock_synchronize.assert_not_called()
+    mock_perf_counter.assert_not_called()
+
+
+def test_sample_tokens_is_inherited_from_upstream():
```
weiguihua2: This use case is meaningless, remove it.

vllm_ascend/worker/worker.py
```diff
@@ -696,7 +700,16 @@ def execute_model(
 
     @torch.inference_mode()
     def sample_tokens(self, grammar_output: "GrammarOutput") -> ModelRunnerOutput | AsyncModelRunnerOutput:
-        return self.model_runner.sample_tokens(grammar_output)
+        if not self.use_v2_model_runner:
+            return self.model_runner.sample_tokens(grammar_output)
+
```
zhenwenqi2024, line 701: Why does v2 have one extra operation compared to v1 here?

vllm_ascend/worker/worker.py
```diff
@@ -866,16 +879,25 @@ def profile_prefill_latency(self, num_tokens: int) -> float:
...
-        self.model_runner._dummy_run(
-            num_tokens=num_tokens,
-            force_attention=True,  # Critical: ensure attention is executed
-            profile_cpp=True,
-        )
+        # Run a real model forward with attention enabled in eager mode.
+        # _dummy_run handles PP internally (intermediate tensors, etc.).
+        old_max_num_reqs = self.model_runner.max_num_reqs
+        try:
+            self.model_runner.max_num_reqs = 1
```
zhenwenqi2024, line 882: Why do we use max_num_seq=1 to probe the forward time in PP？Do we need to consider multi-batch scenarios?

vllm_ascend/worker/worker.py
```diff
@@ -866,16 +879,25 @@ def profile_prefill_latency(self, num_tokens: int) -> float:
...
+
+            dummy_run_kwargs = {
+                "num_tokens": num_tokens,
+                "force_attention": True,
+                "is_profile": False,
+                "cudagraph_runtime_mode": CUDAGraphMode.NONE,
+            }
+
+            if "profile_cpp" in inspect.signature(self.model_runner._dummy_run).parameters:
```
zhenwenqi2024, line 891: profile_cpp should be available by default in dummy_run, right? Do we need to add a check here?

vllm_ascend/worker/worker.py
dianWang325: This is not additional timing behavior for V2 compared with V1; the same timing propagation happens at a different layer.
MRV1 already sets execution_time_ms inside NPUModelRunner.sample_tokens(), so the worker returns its output directly to preserve the existing V1 behavior and avoid duplicate processing.
MRV2 inherits GPUModelRunner.sample_tokens() and has no Ascend-specific hook there. We stop and cache the CPP timer at the end of execute_model(), then attach the cached value after sample_tokens() creates the ModelRunnerOutput. Therefore, this worker-side post-processing is only required for V2.

vllm_ascend/worker/worker.py
dianWang325: `profile_cpp` is available in the MRv1 _dummy_run, but MRv2 inherits the upstream `_dummy_run`, whose signature does not include this argument. Since `profile_prefill_latency()` is shared by both runners, the `inspect.signature` check is used for compatibility: we pass `profile_cpp=True` only when the runner supports it.

vllm_ascend/worker/worker.py
dianWang325: Using `max_num_reqs=1` here is intentional. The startup profiling stage builds the baseline latency model from a single prefill request without history, so `num_tokens` directly represents the sequence length.
Multi-request batches and history lengths are handled later by the online calibration stage. It records the `(chunk_size, history_length)` of each request in the actual batch and uses the measured batch execution time to fit the history-aware model.
Therefore, the startup profiling does not need to simulate an arbitrary multi-request distribution; that workload-dependent behavior is refined using real runtime batches.

dianWang325:
/rerun
[Bot]: rerun completed.

**Failed:**

weiguihua2 (approved):

zhenwenqi2024 (approved):

PR #14292, opened 2026-08-14
[MRV2][BugFix][Kernel]Support apply_thinking_budget on Ascend

Liamup777:
This PR adds Ascend support for the Model Runner V2 thinking-token-budget kernels.

The upstream implementation exposes two independent Triton-Ascend compatibility issues:

1. `_update_committed_marker_cache_kernel`

   Triton-Ascend versions earlier than 3.6 cannot compile chained runtime boolean expressions such as `A and B and C`. The Ascend implementation expresses these conditions as nested two-operand conditions without changing the scanning algorithm.

   Triton-Ascend 3.6 already supports the upstream syntax. This compatibility implementation can be removed once 3.6 becomes the minimum supported version.

2. `_thinking_budget_kernel`

   Its `_load_effective_token` helper returns values from different runtime branches and pointer sources. During TTIR-to-Linalg conversion, this can produce an invalid zero-operand `func.return` and leave an unresolved `memref<?xi32>` to `!tt.ptr<i32>` materialization.

   The Ascend helper replaces the runtime branch returns with complementary masked loads followed by `tl.where`. The three effective-token loading call sites are the only functional changes to the thinking-budget kernel itself.

This PR also:

- Patches the two upstream kernel bindings to use their Ascend-compatible implementations.
- Keeps the upstream `apply_thinking_budget` wrapper unchanged.
- Ports all 12 upstream Model Runner V2 thinking-budget tests to the A2 NPU test suite.
- Documents the compiler issues, adaptations, and removal conditions in `vllm_ascend/patch/__init__.py`.

No 

```bash
pytest -q tests/ut/sample/a2/test_thinking_budget.py

result : 12 passed
```

- vLLM version: v0.27.2rc0
- vLLM main: https://github.com/vllm-project/vllm/commit/7f7a32cfec0f1bc5b73c37200b86631523a1ea8f

### DIFF
--- a/.github/workflows/scripts/test_config.yaml
+++ b/.github/workflows/scripts/test_config.yaml
@@ -881,6 +881,7 @@ estimated_times:
   tests/ut/quantization/methods/a2/test_w8a8_dynamic.py: 40
   tests/ut/quantization/methods/a2/test_w8a8_static.py: 40
   tests/ut/sample/a2/test_gumbel_sampling.py: 80
+  tests/ut/sample/a2/test_thinking_budget.py: 80
   tests/ut/spec_decode/a2/test_eagle_proposer.py: 50
   tests/ut/worker/a2/test_block_table.py: 30
   tests/ut/worker/a2/test_model_runner_v1.py: 30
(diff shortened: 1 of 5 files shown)


Liamup777:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

linfeng-yuan:
@Ronald1995 @taojiovo @ZT-AIA 
PTAL a triton kernel optimization for mrv2.

PR #14317, opened 2026-08-14
[v0.25.1rc][BugFix][Frontend] Align DeepSeek V4 parser thinking default

zhangshuai404:
Backport vLLM-Ascend PR #14074 from releases/v0.26.0rc to releases/v0.25.1rc.

The DeepSeek V4 Flash 0731 tokenizer backport defaults omitted thinking controls to thinking mode with high reasoning effort, while the vLLM v0.25.1 parser still starts in CONTENT mode. Because the rendered prompt already ends with <think>, generated reasoning may not emit another opening tag and is consequently returned in content.

Enable the parser thinking state only when both thinking and enable_thinking are omitted. Preserve explicit thinking disablement and reasoning_effort=none semantics, and add parser state and implicit-start regression coverage.

Related upstream changes: vLLM PR #50580 and PR #51296.

Validation: python -m pytest -q tests/ut/patch/platform/test_deepseek_v4_thinking.py (35 passed); two-node real-weight DeepSeek-V4-Pro-0813 W4A8 regression requests passed for default and explicit high reasoning effort.

(cherry picked from commit 761066ead262db0a0427ca143a14f1035cb8908c)

- vLLM version: v0.25.1

### DIFF
--- a/tests/ut/patch/platform/test_deepseek_v4_thinking.py
+++ b/tests/ut/patch/platform/test_deepseek_v4_thinking.py
@@ -2,6 +2,7 @@
 
 import pytest
 from vllm.entrypoints.openai.chat_completion.protocol import ChatCompletionRequest
+from vllm.parser.deepseek_v4 import DeepSeekV4Parser
 from vllm.tokenizers import deepseek_v4, deepseek_v4_encoding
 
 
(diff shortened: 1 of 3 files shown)


QwertyJack:
LGTM

PR #14321, opened 2026-08-15
[BugFix][Scheduler] Fix Mamba split call argument mismatch

zhao-stack:

Align the balance-scheduler call to `Scheduler._mamba_block_aligned_split()` with the pinned vLLM implementation.

The balance scheduling path still passed the removed `num_uncached_common_prefix_tokens` positional argument when `need_mamba_block_aligned_split=True` and `load_kv_async=False`. That path could raise a positional-argument `TypeError` at runtime.

This patch removes the obsolete final argument and its now-unused local hint assignments. It does not change the upstream method signature or scheduling behavior.

The regression is covered by a lightweight CPU scheduler-path smoke test. It constructs a real `BalanceScheduler`, `Request`, and KV cache, enables Mamba block alignment, disables the connector so `load_kv_async=False`, and executes `schedule()` through the waiting-to-running path. The test verifies the real inherited Mamba helper aligns a six-token request to a four-token block and completes without an argument mismatch. It does not parse source or hardcode the upstream signature.

No API change. It prevents a runtime `TypeError` on the affected balance-scheduler Mamba path.

```text
python -m compileall -q tests/ut/patch/platform/test_patch_balance_schedule.py vllm_ascend/patch/platform/patch_balance_schedule.py
# passed

git diff --check
# passed

python -m ruff format --check vllm_ascend/patch/platform/patch_balance_schedule.py tests/ut/patch/platform/test_patch_balance_schedule.py
2 files already formatted

python -m ruff check vllm_ascend/patch/platform/patch_balance_schedule.py tests/ut/patch/platform/test_patch_balance_schedule.py
All checks passed!
```

The local environment cannot execute the scheduler-path test because `torch` is not installed; the PR CI runs it in the repository test environment.


zhao-stack:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

weijinqian0 (approved):

zhao-stack:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

weijinqian0 (approved):

PR #14340, opened 2026-08-15
[Feature][KV Cache] Support non-contiguous KV cache in hybrid models like Qwen3.5 and GQA models

zhaochuang001:

This PR adds non-contiguous KV cache support for hybrid Attention/Mamba models on Ascend, including Qwen3.5 and GQA-based models.

The current Ascend implementation requires Attention and Mamba cache pages to use the same page size and materializes contiguous cache views. This wastes memory and prevents hybrid models from using independent logical page layouts.

This PR:

- Supports independent Attention and Mamba KV cache page sizes.
- Creates stride-based cache views for hybrid layouts instead of materializing padded cache regions.
- Preserves existing legacy paths for pure Mamba/SSM models, separate K/V allocations, KV-transfer layouts, shared-backing layouts, and legacy MLA paths.
- Adds optimized `gather_ssm_states` and `scatter_ssm_states_` Triton kernels for non-contiguous Mamba state access.
- Documents and validates the supported state layout contract:
  - Only the first state stride may contain physical page padding.
  - Inner state dimensions must remain contiguous.
  - State indices select physical rows and may be unsorted or non-contiguous.
  - `has_initial_state=False` produces a zero initial state without reading the cache.

Generic advanced indexing is inefficient and unsafe for padded non-contiguous state storage when used with the fused recurrent GDN path.

No new user-facing CLI option is introduced.

The change is internal to Ascend KV cache allocation, layout construction, and hybrid GDN state access.

#### Unit tests

- State gather/scatter correctness tests: 18 cases passed.
- MRV2 cache layout tests: 49 cases passed.
- Covered dense and padded layouts, block boundaries, multi-block tails, `int32`/`int64` indices, real GDN state shapes, unselected rows, padding preservation, empty-cache validation, and non-contiguous Attention K/V views.

#### End-to-end model tests

- Qwen3.5-35B-A3B W8A8:
  - V1 engine.
  - ModelRunner V2.
  - FULL_DECODE_ONLY graph mode.
  - Short, long-context, repeated, and concurrent requests.
- Qwen2.5-7B GQA:
  - 28 attention heads and 4 KV heads.
  - Runtime logs confirmed non-contiguous K/V views.
  - Short, long-context, repeated, and 32-way concurrent requests passed.
- All request matrix tests completed successfully with zero failed requests.

#### Performance before and after this PR

The pre-PR baseline used vLLM-Ascend commit `af277b2681443089e4d29684b1101707e2eb6426` with Custom GDN operators.

The post-PR configuration used vLLM-Ascend commit `fc4f6ffd97aafd7ad897c387031a46da05ce1d91` with FLA PR #559 operators and non-contiguous KV cache.

Both configurations used:

- Qwen3.5-35B-A3B W8A8.
- TP=2 and EP.
- MTP=3.
- FULL_DECODE_ONLY graph mode.
- `fused_mc2=0`.
- `gpu_memory_utilization=0.90`.
- GSM8K full dataset: 1319 requests.
- Concurrency: 64.
- Maximum output length: 512 tokens.

Two opposite-order runs were completed:

| Run | Configuration | TTFT | TPOT | Output throughput | Total throughput | Duration |
|---|---|---:|---:|---:|---:|---:|
| 1 | Pre-PR Custom | 627.1 ms | 35.8 ms | 1715.13 tok/s | 1954.16 tok/s | 393.008 s |
| 2 | Post-PR FLA | 612.3 ms | 36.0 ms | 1702.93 tok/s | 1940.23 tok/s | 395.881 s |

#### State I/O ablation

Under the same FLA, model, graph mode, and non-contiguous layout:
| Run | Configuration | TTFT | TPOT | Output throughput | Total throughput | Duration |
|---|---|---:|---:|---:|---:|---:|
| 1 | Old advanced-indexing path | 1491.2 ms | 87.8ms | 692.26 tok/s | 788.72 tok/s | 973.891 s |
| 2 | New gather/scatter path| 612.3 ms | 36.0 ms | 1702.93 tok/s | 1940.23 tok/s | 395.881 s |

The old advanced-indexing path caused both severe performance degradation and abnormal early termination under the non-contiguous layout.

#### Environment

- vLLM: `84030bbe3d74d99bad477a3d2e37a973ccd8865c`
- CANN: 9.2.0
- Ascend 910
- FLA
- TP=2, EP enabled
- `fused_mc2=0`
- FULL_DECODE_ONLY graph mode
- `gpu_memory_utilization=0.90`


vllm_ascend/patch/platform/patch_mamba_config.py
```diff
@@ -29,99 +25,16 @@ def _using_kv_store(vllm_config) -> bool:
...
-    Ensure that page size of attention layers is greater than or
-    equal to the mamba layers. If not, automatically set the attention
-    block size to ensure that it is. If the attention page size is
-    strictly greater than the mamba page size, we pad the mamba page size
-    to make them equal.
-
-    Args:
-        vllm_config: vLLM Config
-    """
+    """更新混合 Attention/Mamba 模型配置，不再强制两类 cache 的页大小一致。"""
```
Angazenn: Don't use chinese comments.

vllm_ascend/worker/block_table.py
```diff
@@ -377,12 +377,12 @@ def __init__(
                     max_num_batched_tokens,
                     pin_memory,
                     device,
-                    kernel_size_list,
+                    [kernel_size],
```
Angazenn: why do we need to change this "list" behavior? Is there only one supported kernel_size here?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -4153,6 +4166,60 @@ def _reshape_kv_cache_tensors(
...
+                            kv_cache_spec.page_size_bytes,
+                        )
+                    else:
+                        raise NotImplementedError(
+                            "Hybrid non-contiguous KV cache does not support "
+                            f"{type(kv_cache_spec).__name__}"
+                        )
+
+            if has_attn and has_mamba:
+                self._update_hybrid_attention_mamba_layout(
```
Angazenn: `_update_hybrid_attention_mamba_layout` calls `group.backend.get_kv_cache_block_dim` to find `block_dim` of kv_cache shapes. We don't need to change this?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -4153,6 +4166,60 @@ def _reshape_kv_cache_tensors(
...
+                            kv_cache_spec.dtypes,
+                            kv_cache_spec.page_size_bytes,
+                        )
+                    else:
+                        raise NotImplementedError(
+                            "Hybrid non-contiguous KV cache does not support "
+                            f"{type(kv_cache_spec).__name__}"
+                        )
+
+            if has_attn and has_mamba:
```
Angazenn: I think this condition is duplicate with `if self.hybrid_with_attn_and_mamba:`.

vllm_ascend/worker/model_runner_v1.py
Angazenn: Since we have supported strided attention kv now, I think this branch should be removed. MLA codes are still left here for legacy.

Angazenn:
Also remember to test with modelrunnnerV2 as we have already supported mamba on mrv2.

vllm_ascend/patch/platform/patch_mamba_config.py
zhaochuang001: Done, the docstring has been updated to English.

vllm_ascend/worker/model_runner_v1.py
zhaochuang001: Thanks for pointing this out. The condition was redundant because this branch is already guarded by `self.hybrid_with_attn_and_mamba`. I removed `has_attn`/`has_mamba` and now call `_update_hybrid_attention_mamba_layout()` directly.

vllm_ascend/worker/model_runner_v1.py
zhaochuang001: Thanks. The legacy hybrid KV branches are now unreachable because the new strided hybrid path returns earlier. I removed both obsolete branches while keeping the generic MLA reshape logic for legacy non-hybrid cases.

zhaochuang001:
> Also remember to test with modelrunnnerV2 as we have already supported mamba on mrv2.

Thanks. Tested with `VLLM_USE_V2_MODEL_RUNNER=1` using Qwen3.5. The server started successfully, and short, long-context, repeated, and concurrent requests completed without KV-cache layout or block-size errors.

vllm_ascend/patch/platform/patch_mamba_config.py
```diff
@@ -45,70 +46,56 @@ def _using_kv_store(vllm_config) -> bool:
...
-        qk_rope_head_dim = model_config.hf_text_config.qk_rope_head_dim
-        attn_single_token_k_page_size = kv_lora_rank * attn_num_kv_heads * get_dtype_size(kv_cache_dtype)
-        attn_rope_token_page_size = qk_rope_head_dim * attn_num_kv_heads * get_dtype_size(kv_cache_dtype)
-        attn_token_page_size = attn_single_token_k_page_size + attn_rope_token_page_size
-    else:
         attn_num_kv_heads = model_config.get_num_kv_heads(parallel_config)
-        attn_head_size = model_config.get_head_size()
-        attn_single_token_k_page_size = attn_head_size * attn_num_kv_heads * get_dtype_size(kv_cache_dtype)
-        attn_token_page_size = 2 * attn_head_size * attn_num_kv_heads * get_dtype_size(kv_cache_dtype)
+        if model_config.use_mla:
```
Angazenn, line 83: check whether use_mla should be placed outside `if index_kpool is not None:`

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -5301,6 +5381,43 @@ def _reshape_kv_cache_tensors(
...
+                            layer_name,
+                            current_kv_cache_spec.page_size_bytes,
+                            [tuple(tensor.shape) for tensor in state_tensors],
+                            [tensor.stride() for tensor in state_tensors],
+                            [tensor.storage_offset() for tensor in state_tensors],
+                            [tensor.is_contiguous() for tensor in state_tensors],
+                        )
+                        continue
+
                     state_tensors = []
```
Angazenn, line 5772: Check whether it is necessary to maintain these old layout codes

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -5081,15 +5118,25 @@ def _reshape_kv_cache_tensors(
...
+                        raw_k_tensor = raw_v_tensor = kv_cache_raw_tensors[layer_name]
+                        sum_page_size_bytes = raw_k_tensor.numel()
+                        raw_kv_is_combined = True
                     elif (
                         self.hybrid_with_attn_and_mamba
                         and "cache_only_layers" not in layer_name
                         and not is_hidden_state_cache_spec(current_kv_cache_spec)
                         and isinstance(kv_cache_raw_tensors[layer_name], torch.Tensor)
+                        and self.use_hybrid_blocks
                     ):
```
Angazenn, line 5487: `self.use_hybrid_blocks` should always be true if `self.hybrid_with_attn_and_mamba` is True. So these conditions can be simplified:


```suggestion
                    elif (
                        and "cache_only_layers" not in layer_name
                        and not is_hidden_state_cache_spec(current_kv_cache_spec)
                        and isinstance(kv_cache_raw_tensors[layer_name], torch.Tensor)
                    ):
```

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -4557,6 +4569,18 @@ def _allocate_kv_cache_tensors(self, kv_cache_config: KVCacheConfig) -> dict[str
         self.hybrid_with_attn_and_mamba = any(
             isinstance(spec, MambaSpec) for spec in layer_kv_cache_spec.values()
         ) and any(isinstance(spec, AttentionSpec) for spec in layer_kv_cache_spec.values())
+        non_contiguous_attention_layers: set[str] = set()
```
Angazenn: This variable name is a little bit confusing as we already have many attention backends that support stride view of caches. Maybe rename to `strided_gqa_cache_layers`?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -5152,23 +5199,56 @@ def _reshape_kv_cache_tensors(
...
+                    )
+                    kv_cache_shape = attn_backend.get_kv_cache_shape(
+                        num_blocks * block_size_chunk,
+                        kernel_block_size,
+                        current_kv_cache_spec.num_kv_heads,
+                        current_kv_cache_spec.head_size,
+                    )
+                    if (
+                        raw_kv_is_combined
+                        and len(kv_cache_shape) == 5
```
Angazenn, line 5564: Do we need `len(kv_cache_shape) == 5 and kv_cache_shape[0] == 2` to guard?

vllm_ascend/worker/model_runner_v1.py
Angazenn, line 5615: Check whether it is necessary to maintain this legacy branch for GQA contiguous cache.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -4087,6 +4100,60 @@ def _reshape_kv_cache_tensors(
             corresponding memory buffer for KV cache.
         """
         kv_caches: dict[str, torch.Tensor] = {}
+        if self.hybrid_with_attn_and_mamba:
+            has_attn, has_mamba = False, False
+            for group in self._kv_cache_spec_attn_group_iterator():
+                if group.kv_cache_group_id == len(kernel_block_sizes):
+                    # The last group may contain layers without KV cache.
+                    continue
```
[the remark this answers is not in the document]
zhaochuang001: Fixed in a later commit. The guard now uses >=, so this comment is outdated.

vllm_ascend/patch/platform/patch_mamba_config.py
zhaochuang001: index_kpool and use_mla control two different aspects of this logic.
index_kpool determines whether the special page-size alignment for the kpool/indexer, Attention, and Mamba caches is needed. use_mla only selects the Attention per-token page-size formula within that alignment logic:
- MLA uses kv_lora_rank + qk_rope_head_dim;
- standard Attention uses 2 * head_size * num_kv_heads.
Therefore, use_mla is intentionally checked inside the index_kpool branch. Models without index_kpool still support MLA, but they do not need this kpool-specific page alignment.

vllm_ascend/worker/model_runner_v1.py
zhaochuang001: This legacy layout path is still required.
It covers two cases:
1. Pure Mamba/SSM models, where all cache groups are MambaSpec and self.hybrid_with_attn_and_mamba is false.
2. Hybrid/cache-transfer layouts where multiple layers are views into the same backing tensor. The worker materializes per-layer slices from a shared allocation, and these layers must keep the existing contiguous slicing and offset semantics.
The new _adjust_kv_layout() path is used only for hybrid Attention/Mamba layers with independent raw-tensor backing. Removing the legacy path would break pure Mamba models and shared-backing cache groups, so it is intentionally retained.

vllm_ascend/worker/model_runner_v1.py
zhaochuang001: These two flags are intentionally kept separate. hybrid_with_attn_and_mamba is derived from the presence of both AttentionSpec and MambaSpec, while use_hybrid_blocks is derived from the number of attention groups. Some packed UniformType layouts, such as K3, can keep Mamba specs inside one group, so the two flags are not guaranteed to be equivalent. The additional use_hybrid_blocks guard is therefore retained for the layout path that requires multiple hybrid groups.

vllm_ascend/worker/model_runner_v1.py
zhaochuang001: The set currently contains all FullAttentionSpec layers, rather than GQA-only layers. Since FullAttentionSpec can also represent regular MHA, strided_gqa_cache_layers would be too narrow. I will rename it to strided_attention_cache_layers to describe the actual scope more accurately.

vllm_ascend/worker/model_runner_v1.py
zhaochuang001: This legacy path is still required for cache layouts where K and V are allocated as separate tensors, including PD/KV-transfer and backend-specific Attention paths. The new non-contiguous as_strided view is only used when K/V share one raw tensor and the backend exposes the expected 5D [K/V, ...] shape. Removing the legacy branch would break those separate K/V cache layouts.

vllm_ascend/worker/model_runner_v1.py
zhaochuang001: These guards are required for the special combined K/V view. raw_kv_is_combined only indicates that K and V share one backing tensor; it does not guarantee the shape layout. The as_strided construction below assumes a 5D shape whose first dimension is exactly 2, representing the K/V lanes, and uses the corresponding two-lane stride. The guards prevent other combined cache layouts from entering this path.

vllm_ascend/ops/triton/mamba/state_index.py
```diff
@@ -0,0 +1,163 @@
...
+        raise ValueError(f"state must have at least 2 dimensions, got {state.ndim}")
+    if indices.ndim != 1:
+        raise ValueError(f"indices must be 1D, got shape={tuple(indices.shape)}")
+    if indices.dtype not in (torch.int32, torch.int64):
+        raise TypeError(f"indices must be int32 or int64, got {indices.dtype}")
+    if indices.device != state.device:
+        raise ValueError("indices and state must be on the same device")
+    if not state[0].is_contiguous():
+        raise ValueError("each state row must be contiguous")
+
```
AuroraEmiya, line 93: Check if those validations are reasonable or any other missed according to your implementation

vllm_ascend/ops/triton/mamba/state_index.py
```diff
@@ -0,0 +1,163 @@
...
+    *,
+    output_dtype: torch.dtype | None = None,
+) -> torch.Tensor:
+    """Gather selected rows from a state cache with a padded batch stride.
+
+    Unlike ``state[indices]`` or ``index_select``, this kernel reads only the
+    requested rows. ``has_initial_state=False`` produces a zero row without
+    reading the state cache.
+    """
+    _validate_indices(state, indices)
```
AuroraEmiya: Since there already is a function to validate your inputs , why do you add more as below ?

vllm_ascend/ops/triton/mamba/state_index.py
```diff
@@ -0,0 +1,163 @@
...
+    batch_idx = tl.program_id(1)
+    offsets = block_idx * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < row_size
+
+    has_initial_state = tl.load(
+        has_initial_state_ptr + batch_idx * stride_has_initial_state
+    ).to(tl.int1)
+    state_idx = tl.load(indices_ptr + batch_idx * stride_indices).to(tl.int64)
+    state_idx = tl.where(has_initial_state, state_idx, 0)
+
```
AuroraEmiya, line 37: Check if tl.load has a mask shortcut

AuroraEmiya: See what you've done in L62 - L65

AuroraEmiya:
You may need to write  your pr description at the very first place : (

tests/ut/ops/test_mamba_state_index.py
```diff
@@ -0,0 +1,81 @@
...
+    not hasattr(torch, "npu") or not torch.npu.is_available(),
+    reason="Mamba state index tests require an Ascend NPU",
+)
+
+
+@pytest.mark.parametrize("padding", [0, 13])
+def test_gather_and_scatter_ssm_states(padding: int) -> None:
+    num_states = 5
+    state_shape = (2, 4, 8)
+    row_size = 64
```
AuroraEmiya: `STATE_IO_BLOCK_SIZE` is 1024, but this test uses `row_size = 64`, so `grid[0]` is always 1 and the `block_idx > 0` / multi-block path is never exercised. Could we cover at least `BLOCK_SIZE - 1`, `BLOCK_SIZE`, `BLOCK_SIZE + 1`, a multi-block + tail case, and one real GDN temporal-state shape? Also please parameterize both `int32` and `int64` indices, since the implementation explicitly supports both.

vllm_ascend/ops/triton/mamba/state_index.py
```diff
@@ -0,0 +1,163 @@
...
+
+def gather_ssm_states(
+    state: torch.Tensor,
+    indices: torch.Tensor,
+    has_initial_state: torch.Tensor,
+    *,
+    output_dtype: torch.dtype | None = None,
+) -> torch.Tensor:
+    """Gather selected rows from a state cache with a padded batch stride.
```
AuroraEmiya, line 131: Could we document the exact layout contract here? The kernel flattens all inner dimensions and only honors `state.stride(0)`, so the supported non-contiguous case is a padded state-row/batch stride with each row contiguous, not an arbitrary strided tensor. It would also help to state that `indices` select physical state rows and are not required to be contiguous/sorted, plus the `has_initial_state=False -> zero row` and `output_dtype` conversion semantics.

AuroraEmiya (commented):
This PR changes the production fused GDN prefill path from advanced indexing + `clear_ssm_states` + indexed write-back to the new Triton `gather_ssm_states` / `scatter_ssm_states_` implementation. Could we add old/new correctness parity and performance data for this replacement?

For correctness, please compare the gathered initial state and the full state cache after scatter, including unselected rows and padding, on both dense and padded-stride layouts and at least one real business shape.

For performance, please compare the previous indexing path with the new Triton path on representative GDN shapes/layouts. It would be useful to report gather/scatter latency and, if possible, the complete fused-prefill path. The code comment states that generic advanced indexing materializes the padded cache, so benchmark data should demonstrate the expected benefit and rule out regressions.

vllm_ascend/ops/triton/mamba/state_index.py
zhaochuang001: Thanks. I added checks for empty state caches and overlapping rows, and documented the supported layout: only stride(0) may include padding, while each state row must be contiguous.

Index range checks remain a caller precondition to avoid NPU synchronization on the hot path. Tests now cover these validations, both index dtypes, block boundaries, and a real GDN state shape.

vllm_ascend/ops/triton/mamba/state_index.py
zhaochuang001: Good suggestion. I will use a masked tl.load with other=0 for state_idx and remove the extra tl.where. The state load remains masked by both the row boundary and has_initial_state.

vllm_ascend/ops/triton/mamba/state_index.py
zhaochuang001: The additional checks are gather-specific, while the original helper only validated the shared state and indices contract. I split the validation into shared, gather-specific, and scatter-specific helpers to make the responsibilities explicit.

tests/ut/ops/test_mamba_state_index.py
zhaochuang001: Thanks. I expanded the tests to cover STATE_IO_BLOCK_SIZE - 1, STATE_IO_BLOCK_SIZE, and STATE_IO_BLOCK_SIZE + 1, including the multi-block tail path. I also added a real GDN state shape, both padded and dense layouts, and int32/int64 indices. All 18 cases pass.

vllm_ascend/ops/triton/mamba/state_index.py
zhaochuang001: Documented. The state layout is [num_state_rows, *inner_dims]: all inner dimensions must be contiguous, and only state.stride(0) may contain physical page padding. Arbitrary inner-dimension striding is not supported. The docstring also clarifies that indices select physical rows and may be unsorted or non-contiguous, has_initial_state=False produces a zero row without reading the cache, and output_dtype controls the gathered output dtype.

ZT-AIA (commented):
For the ascendc operator, switch from the customized repository to the fla. Therefore, delete the code related to custom operator and replace the operator. The single operator test cases are not required because the use cases in the fla repository are monitored. If the triton operator is added, write the MD document and test case. The test case can be verified locally. Then, place the test case in the https://github.com/vllm-project/vllm-ascend/blob/main/tests/e2e/nightly/single_node/ops/singlecard_ops/triton/

Issue #14273, opened 2026-08-14
[Bug]: DeepSeek V4 PD output corruption with CANN 9.1.0 (works with CANN 9.0.1)


jianzs:
### Your current environment

<details>
<summary>Sanitized environment information</summary>

The output is limited to public, relevant fields because the full environment report contains deployment-specific addresses and paths.

```text
Platform: Linux aarch64, glibc 2.38
Python: 3.12.13
PyTorch: 2.10.0+cpu
vLLM: 0.26.0
vLLM GitHub base commit: 568afb3a13806beb53bb2e6bd518269357b237c0
vLLM Ascend GitHub base commit: 561bd00f952cb28ff57ffda75acfe728f948aa4f
Model: DeepSeek-V4-Flash-W8A8
Deployment: Prefill/Decode disaggregation with MooncakeConnector

Affected image:
  CANN: 9.1.0
  torch-npu: 2.10.0.post4

Working comparison image:
  CANN: 9.0.1
  torch-npu: 2.10.0.post2
```

GitHub commits:

- vLLM: https://github.com/vllm-project/vllm/commit/568afb3a13806beb53bb2e6bd518269357b237c0
- vLLM Ascend: https://github.com/vllm-project/vllm-ascend/commit/561bd00f952cb28ff57ffda75acfe728f948aa4f

</details>

#### Description

A DeepSeek V4 W8A8 Prefill/Decode-disaggregated deployment produces corrupted output immediately when run with the CANN 9.1.0 image. An otherwise equivalent deployment using the same model, vLLM/vLLM Ascend code, launch arguments, and request works correctly with the CANN 9.0.1 image.

Both requests return HTTP 200 and complete without a Python traceback or a Mooncake transfer error. The affected response contains multilingual garbage in `reasoning_content`, has `content: null`, and reaches `finish_reason: "length"`.

The comparison image also has a different torch-npu patch version, so this report does not claim that CANN alone is the root cause. It reports the reproducible environment regression.

#### Relevant deployment configuration

```text
Prefill:
  TP=2, DP=1, PP=4, expert parallel enabled
  max_num_batched_tokens=4096, max_num_seqs=12, block_size=32
  eager mode, async scheduling, prefix caching enabled

Decode:
  TP=2, DP=4, PP=1, expert parallel enabled
  max_num_batched_tokens=768, max_num_seqs=48, block_size=32
  async scheduling, prefix caching disabled
  FULL_DECODE_ONLY graph mode

Common:
[...]

PR #14358, opened 2026-08-15
[v0.26.0rc][BugFix][MoE] Fix MegaMoe prefill buffer sizing

jianzs:

CANN MegaMoe allocates its symmetric communication buffer once from the MC2 token capacity. When `enable_prefill_mc2` is disabled, that capacity was derived from the decode graph capture size even though eager prefill could still select MegaMoe through fused MC2. A larger prefill batch could therefore exceed the allocated capacity and produce invalid outputs.

This PR:

- Sizes the MegaMoe buffer from `max_num_batched_tokens` when MegaMoe is actually eligible.
- Centralizes MegaMoe eligibility in `use_cann_megamoe` so communication selection, buffer sizing, DP token handling, and W4A8/W8A8 weight preparation use the same decision, while preserving the `dispatch_ffn_combine` fallback.
- Rejects intermediate sizes outside the CANN 9.1.0 MegaMoe range of 1024 to 3072 or not aligned to 512.
- Keeps DP token counts uniform for MegaMoe, without disabling the uneven-token optimization for `dispatch_ffn_combine`.
- Raises an explicit error if a MegaMoe input exceeds the allocated per-rank capacity.

Yes. MegaMoe now allocates enough symmetric-buffer capacity for eager prefill batches when prefill MC2 is disabled. Unsupported model configurations continue to use `dispatch_ffn_combine`, and an invalid capacity is reported explicitly instead of producing incorrect output.

- `python3 -m py_compile vllm_ascend/ascend_forward_context.py vllm_ascend/ops/fused_moe/moe_comm_method.py vllm_ascend/quantization/methods/w4a8.py vllm_ascend/quantization/methods/w8a8_dynamic.py vllm_ascend/utils.py`
- `ruff check vllm_ascend/ascend_forward_context.py vllm_ascend/ops/fused_moe/moe_comm_method.py vllm_ascend/quantization/methods/w4a8.py vllm_ascend/quantization/methods/w8a8_dynamic.py vllm_ascend/utils.py`
- `python3 -m pytest tests/ut/test_ascend_forward_context.py -q` (`26 passed`)
- End-to-end DeepSeek V4 W8A8 inference on Ascend A3 with TP8/DP2, eager prefill, fused MC2 enabled, and prefill MC2 disabled. A prefill batch larger than the decode graph capture capacity completed without non-finite output and produced the expected response. This validated the buffer fix before the eligibility helper was centralized; the final helper refactor and rebase were verified with the static checks above.

No new test case was added because the buffer failure depends on CANN MegaMoe symmetric-buffer allocation and collective execution across an NPU communication group. The existing config fixture was updated to cover the new intermediate-size eligibility requirement.

- vLLM version: v0.26.0

### DIFF
--- a/vllm_ascend/ascend_forward_context.py
+++ b/vllm_ascend/ascend_forward_context.py
@@ -93,6 +93,20 @@ def _cann_megamoe_supported_by_config(vllm_config: VllmConfig) -> bool:
     return quant_name in _CANN_MEGAMOE_SUPPORTED_QUANT_NAMES
 
 
+def use_cann_megamoe(vllm_config: VllmConfig) -> bool:
+    # TODO: drop the EP-size guard when MegaMoe supports larger EP sizes.
+    return (
+        _MEGA_MOE_SUPPORTED
+        and get_ascend_device_type() == AscendDeviceType.A3
+        and get_ascend_config().enable_fused_mc2 == 1
+        and is_moe_model(vllm_config)
+        and vllm_config.parallel_config.enable_expert_parallel
+        and 1 < get_ep_group().world_size <= 64
+        and getattr(vllm_config, "lora_config", None) is None
+        and _cann_megamoe_supported_by_config(vllm_config)
+    )
+
+
 @contextmanager
 def set_ascend_forward_context(
     attn_metadata: Any,
@@ -141,6 +155,7 @@ def set_ascend_forward_context(
 
         forward_context.moe_comm_type = moe_comm_type
         forward_context.moe_comm_method = get_moe_comm_method(moe_comm_type)
+        forward_context.use_mega_moe = use_cann_megamoe(vllm_config)
 
         tp_world_size = get_tensor_model_parallel_world_size()
 
@@ -243,7 +258,11 @@ def set_mc2_tokens_capacity(vllm_config, max_num_reqs, uniform_decode_query_len)
     global _mc2_tokens_capacity
     if _mc2_tokens_capacity is not None:
         return
-    if get_ascend_config().enable_prefill_mc2:
+
+    ascend_config = get_ascend_config()
+    use_mega_moe = use_cann_megamoe(vllm_config)
+
+    if ascend_config.enable_prefill_mc2 or use_mega_moe:
         max_num_tokens = vllm_config.scheduler_config.max_num_batched_tokens
     elif vllm_config.compilation_config.cudagraph_capture_sizes:
         max_num_tokens = vllm_config.compilation_config.max_cudagraph_capture_size
@@ -254,8 +273,8 @@ def set_mc2_tokens_capacity(vllm_config, max_num_reqs, uniform_decode_query_len)
     # Use integer arithmetic for ceiling division.
     num_tokens_per_tp_rank = (max_num_tokens + tp_size - 1) // tp_size
     # keep the num_tokens_per_tp_rank less than fused_mc2 (mega_moe) tokens per rank limit
-    if get_ascend_config().enable_fused_mc2:
-        if _MEGA_MOE_SUPPORTED:
+    if ascend_config.enable_fused_mc2:
+        if use_mega_moe:
             num_tokens_per_tp_rank = min(num_tokens_per_tp_rank, _MEGA_MOE_TOKENS_PER_RANK_LIMIT)
         else:
             num_tokens_per_tp_rank = min(num_tokens_per_tp_rank, _DISPATCH_FFN_COMBINE_TOKENS_PER_RANK_LIMIT)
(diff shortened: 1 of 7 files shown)


jianzs:
/rerun
[Bot]: rerun completed.

**Failed:**

vllm_ascend/ascend_forward_context.py
```diff
@@ -93,6 +93,20 @@ def _cann_megamoe_supported_by_config(vllm_config: VllmConfig) -> bool:
     return quant_name in _CANN_MEGAMOE_SUPPORTED_QUANT_NAMES
 
 
+def use_cann_megamoe(vllm_config: VllmConfig) -> bool:
+    # TODO: drop the EP-size guard when MegaMoe supports larger EP sizes.
+    return (
+        _MEGA_MOE_SUPPORTED
+        and get_ascend_device_type() == AscendDeviceType.A3
```
nanshen3000, line 110: Why have this device_type=A3 judegement? It seems that you call this func in _select_a3_moe_comm_method, when using this func, get_ascend_device_type() == AscendDeviceType.A3 must be true. Maybe you can delete this judegement?

vllm_ascend/ascend_forward_context.py
```diff
@@ -93,6 +93,20 @@ def _cann_megamoe_supported_by_config(vllm_config: VllmConfig) -> bool:
...
+        and get_ascend_config().enable_fused_mc2 == 1
+        and is_moe_model(vllm_config)
+        and vllm_config.parallel_config.enable_expert_parallel
+        and 1 < get_ep_group().world_size <= 64
+        and getattr(vllm_config, "lora_config", None) is None
+        and _cann_megamoe_supported_by_config(vllm_config)
```
nanshen3000, line 116: You can add more judegement for this func, because currently MegaMoe does not support model for moe_intermediate_size < 1024 or > 3072. You can refer to the latest main branch. But this is not so important

vllm_ascend/ops/fused_moe/moe_comm_method.py
```diff
@@ -367,6 +367,14 @@ def _apply_cann_mega_moe(
         # TokenDispatcherWithMC2 carries global_bs (used below for the mc2_mask
         # branch); assert the subtype so mypy resolves it off the base class.
         assert isinstance(self.token_dispatcher, TokenDispatcherWithMC2)
+        num_tokens = fused_experts_input.hidden_states.shape[0]
+        num_max_tokens = self.token_dispatcher.max_num_tokens_per_rank
+        if num_tokens > num_max_tokens:
+            raise ValueError(
```
nanshen3000, line 373: You make a judgemnet here, if MgeaMoe buffer size is setted wrong, it will raise an error. But if you can know the correct buffer size, maybe you can try to set the num_tokens correctly for self.token_dispatcher.max_num_tokens_per_rank? In this way, you can run this model instead of just raise an error and shut down the vllm process

jianzs: We don’t expect any errors here. If one does occur, it means our underlying assumptions have been broken—that’s a bug. The more reasonable approach is to find and fix the issue, rather than adding a safety net at this point.

vllm_ascend/ascend_forward_context.py
jianzs: This function isn't just used in `_select_a3_moe_comm_method`.

vllm_ascend/ascend_forward_context.py
jianzs: Updated in `_cann_megamoe_supported_by_config`

jianzs:
/rerun
[Bot]: rerun completed.

**Failed:**

QwertyJack:
I reproduced #14273 on the v0.25 release code and validated the same buffer-sizing strategy used by this PR on two A3 nodes.

With `TP=4`, `DP=4`, `EP=16`, `max_num_seqs=16`, seven DSpark draft tokens, and `enable_prefill_mc2=false`, the original MegaMoe buffer was only 32 tokens per rank. A 477-token prefill request exceeded that capacity. Under concurrency the first corruption signals were token `0`, 904960 NaN target-logit values, and all-`-1` recovered tokens, followed by NPU `507035` and EngineDead.

After deriving capacity from `max_num_batched_tokens`, all 16 ranks allocated and observed 2048/2048 tokens per rank. The short request and single 477-token request passed, followed by three concurrency-16 rounds: 48/48 HTTP 200, 48 valid JSON responses, and no NaN, token-0, all-`-1`, `507035`, traceback, or EngineDead signature during the request window.

The controls also passed: CANN 9.1 forced to legacy `dispatch_ffn_combine` completed 48/48 requests, and the 32-rank HCCL all-reduce smoke test returned the expected `528.0` on every rank. These results support this PR's MegaMoe capacity diagnosis and exclude a generic HCCL failure.

The fail-fast check is necessary in addition to correct sizing because the CANN 9.1 wrapper currently forwards `x.size(0)` and `num_max_tokens_per_rank` without checking that the runtime input fits the allocated symmetric buffer.

jianzs:
/rerun
[Bot]: rerun completed.

**Failed:**

jianzs:
/rerun
[Bot]: rerun completed.

**Failed:**

jianzs:
/rerun
[Bot]: rerun completed.

**Failed:**

pgzddxx:
The issue in this PR also occurs on MAIN. Please merge this PR into MAIN as well.

jianzs:
> The issue in this PR also occurs on MAIN. Please merge this PR into MAIN as well.

@pgzddxx  https://github.com/vllm-project/vllm-ascend/pull/14439

PR #14368, opened 2026-08-16
[Feature][Spec Decode] Support probabilistic DSpark sampling on V1

jianzs:

- Enables probabilistic draft sampling for DSpark on Model Runner V1.
- Samples each draft token after applying its Markov correction and retains the resulting conditional draft probabilities for standard rejection sampling.
- Forwards cached draft probabilities through the V1 runner for both single-stage and pipeline-parallel deployments, including request reordering before verification.
- Rejects reduced-vocabulary DSpark configurations during V1 proposer initialization because the Markov bias and remapped target logits have incompatible vocabulary widths.
- Keeps the existing all-greedy path unchanged. Model Runner V2 is outside the scope of this PR.

Yes. Full-vocabulary DSpark can use `draft_sample_method="probabilistic"` with Model Runner V1. Reduced-vocabulary DSpark on V1 now fails during initialization with a clear error.

Validated on Ascend dev1 with the vLLM commit recorded by `.github/vllm-main-verified.commit` (`58d3918e3ea0a544ffedadad2ba84559e9c51d8f`):

- `pytest -q tests/ut/spec_decode/test_dspark_proposer.py`: 28 passed
- `pytest -q tests/ut/worker/a2/test_model_runner_v1.py`: 22 passed
- `pytest -q tests/ut/spec_decode`: 157 passed, 13 skipped
- `python3 -m py_compile` for all changed Python files
- `bash format.sh ci`
- `git diff --check`
- End-to-end mix experiment on Ascend dev0: the average accepted length increased from 2.6 with greedy DSpark to 3.2 with probabilistic DSpark (+0.6, approximately +23.1%).

- vLLM version: v0.27.1


drslark:
V1 will be deprecated and phased out in the future. To maintain architectural consistency, vLLM does not support DSpark and its probabilistic inference in V1. 
Furthermore, we have no plans to backport V2 features to V1, including documentation updates and ongoing maintenance. 
Therefore, developing this feature in V1 is not recommended. You can host this in your own repo instead.

jianzs:
> V1 will be deprecated and phased out in the future. To maintain architectural consistency, vLLM does not support DSpark and its probabilistic inference in V1. Furthermore, we have no plans to backport V2 features to V1, including documentation updates and ongoing maintenance. Therefore, developing this feature in V1 is not recommended. You can host this in your own repo instead.

I wonder if the low acceptance ratio is actually a bug?

kunpengW-code:
> > V1 will be deprecated and phased out in the future. To maintain architectural consistency, vLLM does not support DSpark and its probabilistic inference in V1. Furthermore, we have no plans to backport V2 features to V1, including documentation updates and ongoing maintenance. Therefore, developing this feature in V1 is not recommended. You can host this in your own repo instead.
> 
> I wonder if the low acceptance ratio is actually a bug?

We are consistent with the upstream community. In MRV1, the draft sampling of dspark supports only the greedy mode. For details, see the preceding description.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2434,8 +2438,53 @@ def propose_draft_token_ids(sampled_token_ids):
         )
         return async_output
 
+    def _get_spec_decode_draft_probs(
+        self,
+        spec_decode_metadata: SpecDecodeMetadata,
+        num_invalid_spec_tokens: dict[str, int] | None = None,
```
lilinsiman: why this parameter is necessary in vllm-ascend. Maybe give a note is better for clarifying the diff between the vllm

slippersss:
Could you kindly provide the acceptance rate with and without this pull request? Curious about the improvement.

drslark:
I checked it again.

Though the comments below show that v1 does not support `probabilistic`.

```python
# NOTE(woosuk): Currently, the below code is not used and we always use argmax
# to sample the draft tokens. We will use this after we find a way to manage
# the draft prob tensor.
# Refer to https://github.com/vllm-project/vllm/pull/16899 for the details.
# FIXME(woosuk): The logic here is duplicated with the main sampling code.
# We should refactor this to reuse the same sampling implementation.
def compute_probs_and_sample_next_token(
```
I found a pr to enable `probabilistic`. https://github.com/vllm-project/vllm/pull/40269.

So, `v1` actually supports `probabilistic`.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2434,8 +2438,53 @@ def propose_draft_token_ids(sampled_token_ids):
...
+    ) -> torch.Tensor | None:
+        if not num_invalid_spec_tokens:
+            return super()._get_spec_decode_draft_probs(spec_decode_metadata)
```
drslark, line 2460: Is `super()._get_spec_decode_draft_probs` enough?
Why do you need below codes?

tests/ut/worker/a2/test_model_runner_v1.py
```diff
@@ -13,13 +13,185 @@
...
 )
+from vllm.v1.spec_decode.metadata import SpecDecodeMetadata
 
 from vllm_ascend.attention.utils import get_sfa_qsfa_packed_head_dim
 from vllm_ascend.core.kv_cache_interface import AscendMLAAttentionSpec, AscendSFAIndexerCacheSpec
 from vllm_ascend.utils import AscendDeviceType
 from vllm_ascend.worker.model_runner_v1 import NPUModelRunner
 
 
+class TestProbabilisticDraftProbHandoff(unittest.TestCase):
```
drslark, line 24: Are these uts all necessary?

vllm_ascend/worker/model_runner_v1.py
jianzs: super() is still used when there are no invalid speculative tokens. However, in a mixed PD batch, a newly transferred request can have only padded -1 draft tokens and therefore no cached probability row. The parent implementation treats that expected absence as an error, logs the warning, and returns None, which makes the entire batch fall back to legacy rejection behavior. This override only inserts zero rows for requests whose draft tokens are all invalid padding, so valid co-batched requests keep their cached probabilities. A missing row for any real draft token still preserves the warning and fallback behavior.

vllm_ascend/worker/model_runner_v1.py
jianzs: Thanks for pointing this out. This information is needed by the Ascend V1 PD mixed-batch path. A newly transferred request can be scheduled with only -1 draft placeholders before it has a cached draft-probability row. The upstream helper treats any missing row as an error and returns None, which also causes valid co-batched requests to lose their cached probabilities.

I updated the implementation to derive fully_padded_draft_req_ids directly from the upstream SchedulerOutput.scheduled_spec_decode_tokens field instead of relying on num_invalid_spec_tokens. This avoids depending on additional scheduler population logic and avoids inspecting device-side draft tokens, which would introduce a device-to-host synchronization. Only fully padded rows are filled with zeros; a missing row for real draft tokens still preserves the warning and fallback behavior. I also added a code comment explaining this difference.

jianzs:
> Could you kindly provide the acceptance rate with and without this pull request? Curious about the improvement.

image

tests/ut/worker/a2/test_model_runner_v1.py
jianzs: Good point. The three behavioral cases are needed, but the two padding-related tests did not need to exercise the full sampling path. I simplified them to call _get_spec_decode_draft_probs directly, removing unrelated sampler, logits, sampling-metadata, and lmhead mocks. This removes 25 lines while retaining coverage for:

1. The PP=1 cache, request-reordering, and rejection-sampler handoff.
2. The expected all-padding row in a mixed PD batch.
3. The warning and fallback for a genuinely missing probability row.

The full runner setup is now used only by the PP=1 handoff regression.

drslark:
@SOMEONEUNSEEN
Please review this pr.
Compare it with https://github.com/vllm-project/vllm-ascend/pull/14485.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2438,8 +2451,53 @@ def propose_draft_token_ids(sampled_token_ids):
...
+                    "Missing cached draft probabilities for request %s; "
+                    "falling back to legacy speculative rejection behavior.",
+                    req_id,
+                )
+                return None
+            draft_probs_rows.append(self._draft_probs[row_idx, :num_draft])
+
+        if not draft_probs_rows:
+            return None
+        return torch.cat(draft_probs_rows, dim=0).contiguous()
```
SOMEONEUNSEEN: This fix for the PD batch is a good idea; it might be better to submit an issue or suggestion to the vLLM community.

PR #14373, opened 2026-08-16
[Performance](DCP, GLM5.2): accelerate SFA sparse-index remap with Ascend C

pisceskkk:
## What changed

- add an Ascend C implementation of SFA DCP sparse-index remapping and stable compaction;
- route the SFA CP backend through the custom operator on NPU while retaining the existing reference behavior;
- fuse owner filtering, exact int32 DCP-local remapping, stable packing, and trailing `-1` fill into one AIV launch;
- use exact int32 shift/mask remapping for power-of-two production configurations;
- use exact Vector magic division for non-power-of-two DCP/interleave configurations instead of a Scalar fallback;
- process TopK rows in UB-sized chunks while preserving global row order and fixed-width output;
- support dynamic TopK through 8192 and add focused coverage for int32 precision and chunk boundaries.

## Why

`_remap_sparse_indices` runs once per SFA layer during DCP execution. The replaced path launches a multi-kernel PyTorch chain, converts int32 indices through float32, sorts packing keys, and gathers the result. This adds decode launch overhead, becomes expensive for the 8K-32K query-token rows used by chunked prefill, and loses integer precision above 16,777,216.

The fused operator preserves TopK order, dtype, fixed-width output shape, trailing `-1` padding, and exact int32 semantics while replacing the chain with one native launch. Wide rows remain on AIV and are processed in UB-sized chunks rather than falling back to a per-element Scalar loop.

### Build provenance

- parent branch: `85acaa6f68cd53968d50cae0a33195d0c5022ebe`;
- final Vector commit: `70f9b473ea034003eadd9a533f7aebf7608fb09d`;
- rebuilt the exact combined source on Ascend A3 with CANN 9.1;
- serving A/B runtime: the same internal A3 daily container for both states, vLLM `0.23.0+empty`, PyTorch `2.10.0+cpu`, and torch-npu `2.10.0.post4`;
- rebuilt package SHA256: `3c0beac290149d5ec371823d41a31fbec305a08a06fa7c4e76b6fdc8bd6b2e5f`;
- kernel source SHA256: `69196dad358b36f5968ce1b51fe1b63556d2ef70291e893b00bacf17da9c1193`.

### Operator correctness

- parent PR validation: eager/compile matrix 9/9 bit-exact, isolated NPU e2e 22/22 exact comparisons, and SFA CP precision UT 3/3;
- exact rebuilt final package: 365/365 standalone bit-exact cases;
- rows 1/5/16/128 and TopK 1 through 8192, including 31/32/33, 63/64/65, 127/128/129, 511/512, 2047/2048/2049, and 8191/8192 boundaries;
- DCP/interleave configurations 2/1, 3/3, 8/64, 16/128, and 16/257;
- negative sentinels, `INT32_MIN`/`INT32_MAX`, and values above the float32 exact-integer range;
- every performance shape below matched the replaced PyTorch implementation element-for-element.

### Operator performance

Against the replaced upstream PyTorch method on one idle Ascend A3 logical device. Each sample measures one complete method call with `torch.npu.synchronize()` immediately before and after it, including launches and output allocation. The PyTorch and CANN calls were alternated within each shape. Configuration: TopK 2048, DCP16/rank7, interleave128.

| Rows | PyTorch P50 | Vector CANN P50 | Vector CANN P90 | P50 reduction | Speedup |
|---:|---:|---:|---:|---:|---:|
| 1 | 417.665 us | 76.180 us | 97.400 us | 81.8% | 5.48x |
| 2 | 326.855 us | 64.455 us | 74.710 us | 80.3% | 5.07x |
| 4 | 319.370 us | 63.155 us | 73.610 us | 80.2% | 5.06x |
| 8 | 330.190 us | 67.170 us | 83.600 us | 79.7% | 4.92x |
| 16 | 527.410 us | 77.240 us | 90.600 us | 85.4% | 6.83x |
| 32 | 348.350 us | 70.670 us | 85.820 us | 79.7% | 4.93x |
| 8192 | 2332.325 us | 278.875 us | 324.730 us | 88.0% | 8.36x |
| 16384 | 6926.720 us | 462.465 us | 486.310 us | 93.3% | 14.98x |
| 32768 | 14175.505 us | 833.860 us | 862.330 us | 94.1% | 17.00x |

Rows 1-32 cover decode/multi-request shapes; rows 8192-32768 cover the query-row range used by chunked prefill.

As an incremental check against the parent branch's per-element Scalar CANN kernel, the Vector kernel was 1.28x-1.56x faster for rows 1-32 and 19.80x/22.96x/25.40x faster for rows 8192/16384/32768 in a separate same-device run. The comparison used the same synchronized complete-call timer for both custom-op implementations.

### End-to-end TPOT without MTP

Speculative decoding was disabled to remove acceptance-rate noise. Both states used the same A3 container and logical devices 0-7. Only the sparse-index remap implementation changed.

Configuration: GLM-5.2 W4A8C8, TP8/DCP8, `ag_rs`, `FULL_DECODE_ONLY`, cudagraph capture size `[1]`, SP off, DSA-CP off, prefix cache off, `max-num-seqs=8`, concurrency 1 (online batch size 1), nominal 4096 input / 1000 output tokens.

The final comparison uses two groups and eight requests per implementation in `baseline -> final -> final -> baseline` order. Every group started a fresh service and used zero benchmark warmup requests. All four groups read the same pre-generated prompt JSONL (`SHA256 7c5d332f8bf0be5925c638b4846528dae93d70be2f2bdc0b237d493e12d9474e`) rather than independently regenerating synthetic inputs.

| Metric | PyTorch | Final Vector CANN | Change |
|---|---:|---:|---:|
| Mean TPOT (exact timestamps) | 53.004 ms | 50.709 ms | -4.33% |
| Median TPOT | 53.012 ms | 50.735 ms | -4.30% |
| P90 TPOT | 53.092 ms | 51.142 ms | -3.67% |
| TPOT population CV | 0.148% | 0.827% | - |
| Mean TTFT | 1522.3 ms | 1658.5 ms | +8.95% |
| Median TTFT | 1511.6 ms | 1510.5 ms | -0.08% |
| P90 TTFT | 1566.8 ms | 2024.1 ms | +29.18% |
| Mean E2EL | 54526.2 ms | 52367.3 ms | -3.96% |
| Combined output throughput | 18.3396 tok/s | 19.0956 tok/s | +4.12% |
| Combined input throughput | 78.3100 tok/s | 81.5382 tok/s | +4.12% |
| Combined prefill throughput | 2804.99 tok/s | 2574.59 tok/s | -8.21% |
| Combined total-token throughput | 96.6495 tok/s | 100.6339 tok/s | +4.12% |
| Combined request throughput | 0.018340 req/s | 0.019096 req/s | +4.12% |
| Combined benchmark duration | 436.215 s | 418.945 s | -3.96% |
| Combined external wall time | 496.335 s | 478.004 s | -3.69% |
| Successful requests | 8/8 | 8/8 | - |

Repeat stability:

| Group | Implementation | Mean TPOT | TPOT CV | Output throughput |
|---|---|---:|---:|---:|
| baseline_g1 | PyTorch | 52.955 ms | 0.126% | 18.3552 tok/s |
| final_g1 | Vector CANN | 51.124 ms | 0.115% | 18.9258 tok/s |
| final_g2 | Vector CANN | 50.294 ms | 0.126% | 19.2684 tok/s |
| baseline_g2 | PyTorch | 53.052 ms | 0.107% | 18.3240 tok/s |

The server-tokenized input-length vector was `[4262, 4287, 4292, 4239]` in every group, proving that the paired states used identical prompts. The nominal 4096-token generator input expands slightly after the model's server-side tokenization; this expansion is identical in all four groups. Output length was exactly 1000 for every request. Exact TPOT is calculated from the stored per-token timestamps as `(last timestamp - first-token timestamp) / output_tokens`; AISBench logs, timestamp databases, per-request JSON, service logs, manifests, and wall-clock files are retained.

Across the eight prompt-matched request pairs, the mean TPOT change was -4.33%, with a request-level bootstrap 95% CI of `[-4.98%, -3.68%]` (20,000 deterministic resamples). Every paired request improved; the individual changes ranged from -5.22% to -3.44%. This interval describes these eight repeated requests only.

The two final groups showed first-request TTFT outliers (2337.1 ms and 1889.9 ms), while their remaining six TTFT samples were 1499.0-1512.5 ms. Consequently, mean/P90 TTFT and combined prefill throughput regress, whereas median TTFT is unchanged. These cold-start samples are retained rather than discarded. Decode TPOT, E2EL, output throughput, benchmark duration, and wall time all improve despite those outliers.

The previous single-group result in this PR reported TPOT -3.54% and output throughput +3.65%. The fresh fixed-prompt ABBA run reports -4.33% and +4.12%, respectively. Absolute latency differs between runs, but the direction and magnitude of the end-to-end decode benefit are consistent.

No. This is an internal performance optimization with unchanged operator output semantics and no new configuration or environment variable.

## Pre-submit checks

The exact custom-op build and A3 hardware validations above passed. The runtime serving image uses an internal daily vLLM package, so repository-wide tests remain delegated to the upstream-matched CI image:

```bash
pytest -q tests/e2e/nightly/single_node/ops/singlecard_ops/test_sfa_remap_sparse_indices.py
bash format.sh ci
```

- vLLM version: v0.27.1


Pz1116 (approved):
LGTM, need another approve from kernal maintainers

Issue #13466, opened 2026-08-04
[Bug][P/D]: Proxy recompute retries corrupt chat token history and reasoning streams


QwertyJack:
### Your current environment

<details>
<summary>Relevant versions</summary>

```text
vllm-ascend: releases/v0.23.0 @ d17a99cd51d535c4026ffb596bafa0a23fa87064
vLLM:         v0.23.0 @ 0fc695fc6d1d82e9a5ac6835ac8e4e1c83703665
API:          /v1/chat/completions
Feature:      recompute_scheduler_enable=true on the decode side
```

The minimal protocol reproduction below does not require an NPU or model execution. It uses the response schema emitted by the vLLM v0.23 chat serving layer and the retry logic in the release proxy.

</details>

### 🐛 Describe the bug

The P/D load-balancing proxy does not resume a recomputed request from its exact generated-token history. It reconstructs a new request from parsed response text, which silently changes the chat token history and can combine output from independent generations.

This is a correctness issue, not only a precision risk. It affects requests that trigger decode-side recomputation. Normal requests that do not trigger recomputation do not enter this retry path.

#### Current control flow

The decode scheduler captures the recomputed request's `output_token_ids`:

https://github.com/vllm-project/vllm-ascend/blob/d17a99cd51d535c4026ffb596bafa0a23fa87064/vllm_ascend/core/recompute_scheduler.py#L339-L344

However, the output sent to the API layer discards that history and contains only an empty token delta plus the marker:

```python
EngineCoreOutput(
    request_id=req_info.request_id,
    finish_reason=FinishReason.STOP,
    new_token_ids=[],
    stop_reason="recomputed",
)
```

https://github.com/vllm-project/vllm-ascend/blob/d17a99cd51d535c4026ffb596bafa0a23fa87064/vllm_ascend/core/recompute_scheduler.py#L868-L878

The proxy then accumulates only parsed `content` text:

```python
content = (
    delta.get("content")
    or message.get("content")
    or choice.get("text")
    or ""
)
generated_token += content
```

On `stop_reason == "recomputed"`, chat requests are rebuilt by mutating the first message:

```python
messages[0]["content"] = origin_prompt + generated_token
```

https://github.com/vllm-project/vllm-ascend/blob/d17a99cd51d535c4026ffb596bafa0a23fa87064/examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py#L973-L1058

This reconstruction is not equivalent to continuation from the original token history.

[...]

PR #14388, opened 2026-08-17
[BugFix][P/D] Preserve exact token history across recompute retries

Zhaocen:

Ports #13524 to `main` while preserving the original PR's private retry-token transport and renderer replacement as closely as the current APIs allow.

When the decode-side recompute scheduler returns `stop_reason="recomputed"`, the proxy must resume from the exact original prompt token IDs plus every emitted output-token delta. Reconstructing the retry from parsed response text can change the token sequence, corrupt chat roles, and omit reasoning output.

This port:

- requests prompt and output token IDs internally from the decode side;
- resumes from the exact token prefix carried in the private `_vllm_ascend_recompute_token_ids` field;
- keeps the renderer input replacement from #13524, adapted from the removed v0.23 `OpenAIServingRender` API to the current `OnlineRenderer` API;
- removes the private field after rendering so it does not enter KV-transfer metadata;
- parses streaming responses by SSE event framing;
- preserves one client-visible response identity and assistant stream across retries;
- accounts for remaining generation budget and usage from token counts;
- preserves the current cached-token usage reporting behavior.

Although the vLLM revision used by `main` has a native `prompt_token_ids` reuse path, this port intentionally retains the original PR's private transport contract to minimize behavioral differences from #13524.

Related: #13466

Yes. Chat and completion requests reassigned after decode-side recomputation continue from their exact generated-token history. Internal token fields and retry-only chunks remain hidden unless the client requested token IDs.

Added `tests/ut/test_recompute_proxy.py` covering SSE framing, exact token history, repeated recomputes, non-stream response merging, renderer input replacement, response identity, usage accounting, and token-field visibility.

Local checks:

```bash
python3 -m py_compile \
  examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py \
  vllm_ascend/recompute_proxy.py \
  vllm_ascend/patch/platform/patch_recompute_chat.py \
  tests/ut/test_recompute_proxy.py

git diff --check origin/main..HEAD
```

Direct state-machine and renderer-replacement smoke assertions also passed. Full pytest and Ruff were not run locally because this environment does not have vLLM, pytest, or Ruff installed; PR CI is required. The original v0.23 patch's real-model NPU A/B validation is documented in #13524.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/patch/platform/patch_recompute_chat.py
+++ b/vllm_ascend/patch/platform/patch_recompute_chat.py
@@ -1,0 +1,32 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Ascend project
+
+from functools import wraps
+
+from vllm.renderers.online_renderer import OnlineRenderer
+
+from vllm_ascend.patch.recompute_proxy import replace_rendered_chat_inputs
+
+_ORIGINAL_RENDER_CHAT_ATTR = "_ascend_original_recompute_render_chat"
+
+if not hasattr(OnlineRenderer, _ORIGINAL_RENDER_CHAT_ATTR):
+    setattr(
+        OnlineRenderer,
+        _ORIGINAL_RENDER_CHAT_ATTR,
+        OnlineRenderer.render_chat,
+    )
+
+
+@wraps(getattr(OnlineRenderer, _ORIGINAL_RENDER_CHAT_ATTR))
+async def _render_chat_with_recompute_tokens(
+    self: OnlineRenderer,
+    request,
+    *,
+    skip_mm_cache: bool = False,
+):
+    original = getattr(type(self), _ORIGINAL_RENDER_CHAT_ATTR)
+    result = await original(self, request, skip_mm_cache=skip_mm_cache)
+    return replace_rendered_chat_inputs(request, result)
+
+
+OnlineRenderer.render_chat = _render_chat_with_recompute_tokens
--- a/vllm_ascend/patch/recompute_proxy.py
+++ b/vllm_ascend/patch/recompute_proxy.py
@@ -1,0 +1,154 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM-Ascend project
+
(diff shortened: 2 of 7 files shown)


Pz1116 (approved):
LGTM

Zhaocen:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

vllm_ascend/patch/recompute_proxy.py
```diff
@@ -0,0 +1,154 @@
+# SPDX-License-Identifier: Apache-2.0
```
wangxiyuan, line 1: this file is only used for patch. So please move it to patch module

vllm_ascend/patch/platform/patch_recompute_chat.py
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: Apache-2.0
```
wangxiyuan, line 1: this patch only works if `recompute scheduler` is enabled, right? Please enable the patch if `recompute_scheduler_enable=True`

wangxiyuan (changes requested):

vllm_ascend/patch/platform/patch_recompute_chat.py
Zhaocen: enable patch when is_pd_decode_recompute_scheduler_enabled(vllm_config) returns True

vllm_ascend/patch/recompute_proxy.py
Zhaocen: moved recompute_proxy.py to vllm_ascend/patch module

vllm_ascend/patch/platform/patch_recompute_chat.py
Zhaocen: the patch is required on both the prefill and decode nodes during a recompute retry. `recompute_scheduler_enable` may only be enabled on the decode node. It is a no-op for normal requests because it only changes the rendered input when the internal recompute token-ID key is present.

Zhaocen:
Since the recompute scheduling mechanism is planned to be deprecated, this PR will not be merged into the main branch.

PR #14392, opened 2026-08-17
[Feature][Model] Add Qwen3-Omni DSpark support on Ascend

Zhou248:

Depends on vllm-project/vllm#52560.

This PR completes the Ascend/NPU registration required for Qwen3-Omni DSpark framework support:

- registers the checkpoint architecture `Qwen3OmniDSparkModel`;
- maps it to the existing `AscendQwen3DSparkForCausalLM` runtime;
- reuses the existing Ascend Qwen3 DSpark confidence-head, FC-rotation, and weight-loading implementation selected by the dependent vLLM support.

The Qwen3-Omni checkpoint keeps its dedicated architecture name and configuration contract. No behavior-free Ascend wrapper is added because the Ascend execution and weight-loading path is the same as the existing Qwen3 DSpark runtime.

This is a framework registration PR. A compatible trained Qwen3-Omni DSpark checkpoint is not currently available, so this PR does not claim Qwen3-Omni acceptance-rate, accuracy, performance, eager-mode, or ACLGraph results.

AI assistance disclosure: this patch was prepared with OpenAI Codex assistance and reviewed against the dependent vLLM changes.

Yes. After the dependent vLLM support is available, a draft checkpoint whose architecture is `Qwen3OmniDSparkModel` can resolve to the existing Ascend Qwen3 DSpark runtime. Existing Qwen3 DSpark behavior and command-line options are unchanged.

Local static checks:

```bash
ruff check vllm_ascend/models/__init__.py tests/ut/model_executor/test_qwen3_dspark.py
ruff format --check vllm_ascend/models/__init__.py tests/ut/model_executor/test_qwen3_dspark.py
git diff --check
```

Results:

- Ruff check: passed.
- Ruff format check: passed.
- `git diff --check`: passed.

The focused Qwen3 DSpark test cannot be collected in this local macOS environment because `torch_npu` is unavailable. It remains covered by the standard vLLM-Ascend CI environment.

No Qwen3-Omni DSpark E2E test is added in this PR because no compatible trained checkpoint is currently available. A converted Qwen3 checkpoint would only be a shape/loading surrogate and would not provide a valid Qwen3-Omni DSpark behavioral E2E. Eager/ACLGraph correctness, acceptance, memory, and throughput coverage should be added with the real checkpoint in a follow-up.

- dependent vLLM PR: https://github.com/vllm-project/vllm/pull/52560

### DIFF
--- a/vllm_ascend/models/__init__.py
+++ b/vllm_ascend/models/__init__.py
@@ -49,6 +49,10 @@ def register_model():
     )
     ModelRegistry.register_model("Qwen3DSparkModel", "vllm_ascend.models.qwen3_dspark:AscendQwen3DSparkForCausalLM")
     ModelRegistry.register_model(
+        "Qwen3OmniDSparkModel",
+        "vllm_ascend.models.qwen3_dspark:AscendQwen3DSparkForCausalLM",
+    )
+    ModelRegistry.register_model(
         "DFlash2DraftModel",
         "vllm_ascend.models.qwen3_dflash2:DFlash2Qwen3ForCausalLM",
     )


tests/ut/model_executor/test_qwen3_dspark.py
```diff
@@ -19,16 +19,46 @@
...
+
+        with patch.object(
+            qwen3_dspark.VllmConfig,
+            "get_quantization_config",
+            return_value=target_quant_config,
+        ) as mock_get_quantization_config:
+            result = qwen3_dspark.get_rotation_path(vllm_config)
+
+        mock_get_quantization_config.assert_called_once_with(target_model_config, vllm_config.load_config)
+        assert result == Path("/models/qwen3-omni/rotation.safetensors")
```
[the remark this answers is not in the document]
Zhou248: This test intentionally patches VllmConfig.get_quantization_config to verify that rotation metadata is resolved from the target ModelConfig and target load context. Setting target_model_config.quantization_config directly would test the user-facing quantization field rather than the resolved runtime QuantizationConfig, and would no longer guard the target-quantized/draft-unquantized regression addressed by this PR.

Zhou248:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/models/qwen3_dspark.py
```diff
@@ -7,10 +8,25 @@
...
+    if getattr(target_model_config, "quantization", None) is None:
+        return None
+
+    target_quant_config = VllmConfig.get_quantization_config(target_model_config, vllm_config.load_config)
+    try:
+        quant_description = target_quant_config.quant_description
+        rotation_relative_path = quant_description["optional"]["quarot"]["rotation_map"]["global_rotation"]
+    except (AttributeError, KeyError, TypeError):
+        return None
+    return Path(target_model_config.model) / rotation_relative_path
```
slippersss: Is this necessary? Why do we ignore original `get_rotation_path` and write  another one here?

vllm_ascend/models/qwen3_dspark.py
```diff
@@ -7,7 +8,22 @@
...
```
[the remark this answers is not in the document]
Zhou248: This is necessary for the standalone DSpark case. The existing helper in llama_eagle3.py reads vllm_config.quant_config, which may represent the draft model or be None when the draft model is unquantized. However, the DSpark FC layer consumes hidden states produced by the target model, so its weights must use the target model's QuaRot rotation metadata. Therefore, this helper resolves the runtime quantization config explicitly from the target ModelConfig and LoadConfig.  
I agree that defining another function with the same name is confusing. I will rename it to make the target-specific behavior explicit and add a comment explaining why the existing helper cannot be used directly.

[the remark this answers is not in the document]
slippersss: So here we need quant config of target model, right? If yes, does it mean that `get_rotation_path` in llama_eagle3.py may be problematic? I think both should read target model quant config instead of draft model.

vllm_ascend/models/qwen3_dspark.py
```diff
@@ -7,10 +8,25 @@
...
 
-from vllm_ascend.models.llama_eagle3 import get_rotation_matrix, get_rotation_path
+from vllm_ascend.models.llama_eagle3 import get_rotation_matrix
 from vllm_ascend.utils import vllm_version_is
 
 
+def get_rotation_path(vllm_config: VllmConfig) -> Path | None:
+    """Resolve QuaRot metadata from the target rather than the draft config."""
+    target_model_config = vllm_config.model_config
+    if getattr(target_model_config, "quantization", None) is None:
```
drslark: `getattr` is bad.
Don't use it.

vllm_ascend/models/qwen3_dspark.py
```diff
@@ -7,10 +8,25 @@
 from vllm.model_executor.models.qwen3_dspark import Qwen3DSparkForCausalLM
 from vllm.model_executor.models.utils import AutoWeightsLoader, maybe_prefix
 
-from vllm_ascend.models.llama_eagle3 import get_rotation_matrix, get_rotation_path
+from vllm_ascend.models.llama_eagle3 import get_rotation_matrix
 from vllm_ascend.utils import vllm_version_is
 
 
+def get_rotation_path(vllm_config: VllmConfig) -> Path | None:
```
drslark: Why do you need to add a func instead of modifying the old one?

vllm_ascend/models/qwen3_dspark.py
```diff
@@ -7,10 +8,25 @@
...
+    """Resolve QuaRot metadata from the target rather than the draft config."""
+    target_model_config = vllm_config.model_config
```
drslark: Please provide the error message about it.

drslark (changes requested):

vllm_ascend/models/qwen3_dspark.py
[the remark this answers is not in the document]
slippersss: I think that original `get_rotation_path ` is indeed reading target model quant config. We should check out what is wrong in your scenario.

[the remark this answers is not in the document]
Zhou248: Yes, the rotation metadata must come from the target model because the DSpark FC consumes target hidden states. I retraced the DSpark loading path and found that the existing helper already does this: vllm_config.quant_config is initialized from vllm_config.model_config, which remains the target model config when the standalone draft is loaded. Draft quantization is resolved separately through get_draft_quant_config().  
Therefore, get_rotation_path() in llama_eagle3.py is not problematic. My previous explanation that vllm_config.quant_config could represent the draft model was incorrect. I will remove the duplicate helper and reuse the existing one.

tests/ut/model_executor/test_qwen3_dspark.py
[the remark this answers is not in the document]
Zhou248: Addressed in 666b403. After retracing the loading path, I confirmed that the shared `llama_eagle3.get_rotation_path()` already consumes the target runtime `vllm_config.quant_config`; draft quantization is resolved separately by `get_draft_quant_config()`. I removed the duplicate helper and this test, whose target/draft regression premise was therefore incorrect.

vllm_ascend/models/qwen3_dspark.py
Zhou248: Addressed in 666b403. The duplicate helper has been removed and `qwen3_dspark.py` now reuses the original `llama_eagle3.get_rotation_path()` unchanged.

vllm_ascend/models/qwen3_dspark.py
Zhou248: Addressed in 666b403. The duplicate helper, including this `getattr`, has been removed; the file now reuses the existing shared helper.

vllm_ascend/models/qwen3_dspark.py
Zhou248: After retracing the loading path, I could not provide a valid error caused by the original helper: `vllm_config.quant_config` already belongs to the target model, while draft quantization is resolved separately. The previously observed tensor-split error was an attention/KV-head shape mismatch and was unrelated to rotation lookup. I removed the unsupported workaround in 666b403.

vllm_ascend/models/qwen3_dspark.py
Zhou248: You are right. After retracing the configuration flow, I confirmed that the old helper already reads the target runtime quantization config. Commit 666b403 removes the additional function and restores the shared implementation.

Zhou248:
Updated the PR in commit `666b403` based on the latest review feedback.

After retracing the standalone DSpark loading path, I confirmed that the existing `llama_eagle3.get_rotation_path()` already reads the target runtime `vllm_config.quant_config`; draft quantization is resolved separately by `get_draft_quant_config()`. The duplicate Qwen3 DSpark helper and its incorrect regression test have therefore been removed. The PR is now reduced to the Qwen3-Omni DSpark architecture registration and its focused registry test (2 files, 14 added lines relative to upstream).

Local checks:
- `ruff check`: passed
- `py_compile`: passed
- `git diff --check`: passed

The focused pytest could not be collected in the local macOS environment because the available editable vLLM checkout and vLLM-Ascend branch expose incompatible `FusedMoEFactory` APIs; PR CI should run it against the matched dependency environment.

Ronald1995:
please confirm if we need to implement it in vllm-omini

Ronald1995 (approved):

Zhou248:
Fixed the CI rebase failure in `3bbc7b83c`.

Root cause: the `select-tests` job rebases the PR onto the moving `main` branch before selecting tests. The previous branch history still contained the old `daaeac4` implementation commit, so that rebase conflicted in `vllm_ascend/models/__init__.py` and `vllm_ascend/models/qwen3_dspark.py`, even though the final PR diff had already removed the latter change.

I rebuilt the PR branch from the latest upstream `main` as one minimal commit and force-pushed it with a lease. The PR still contains only the intended architecture registration and registry unit test (2 files, 14 added lines), and is now directly based on `80c833fb8`.

Local checks passed: `ruff check`, `ruff format --check`, `py_compile`, and `git diff --check`.

Zhou248:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

Zhou248:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

Zhou248:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

tests/ut/model_executor/test_qwen3_dspark.py
```diff
@@ -23,12 +23,22 @@
...
 import torch
 
+import vllm_ascend.models as ascend_models
 import vllm_ascend.models.qwen3_dspark as qwen3_dspark
 
 
 class TestQwen3DSparkWeightLoading:
     """Tests for Qwen3 DSpark weight loading."""
 
+    def test_qwen3_omni_uses_qwen3_dspark_runtime(self) -> None:
```
drslark: Why do you need this ut?
There are no register uts for "Qwen3DSparkModel".

drslark (changes requested):

drslark:
Does it need an e2e?

Zhou248:
Regarding E2E coverage: a Qwen3-Omni DSpark E2E would be valuable once a compatible trained checkpoint is available. There is currently no such checkpoint; converting a Qwen3 checkpoint would only exercise a shape/loading surrogate and would not be a valid Qwen3-Omni DSpark behavioral E2E. I have therefore kept this PR limited to the architecture registration and updated the PR description accordingly. Eager/ACLGraph correctness, acceptance, memory, and throughput coverage can follow with the real checkpoint.

tests/ut/model_executor/test_qwen3_dspark.py
Zhou248: Removed in `9329bcbd9`. You are right: this test only mocked `ModelRegistry.register_model`, while the existing `Qwen3DSparkModel` registration has no equivalent UT, so it did not add meaningful behavioral coverage. The existing Qwen3 DSpark weight-loading test remains unchanged.

drslark (approved):
LGTM.

Zhou248:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

slippersss (approved):

ningjingbengxiaohai (approved):

PR #14420, opened 2026-08-17
[Feat] Add A5 prefill MoE communication method override

zzzzzz198:

Adds a new optional `additional_config` option:

`moe_comm_method_for_a5_prefill`

Used to explicitly control the A5 MoE prefill communication method.

Behavior:

- `None` / unset: keep the existing A5 selection logic unchanged.
- `ALLGATHER`: force A5 prefill to use `ALLGATHER`.
- `ALLTOALL`: force A5 prefill to use `ALLTOALL`.
- Other values are rejected at startup.

The validation and normalization are performed in `AscendConfig`, so invalid values fail fast instead of being detected later in the forward path.

This override is applied after the existing MC2 branch, so decode batches still use MC2 and only prefill behavior is affected.

Yes, it adds a new optional `additional_config` field:

- Name: `moe_comm_method_for_a5_prefill`
- Type: `str | None`
- Default: `None`
- Valid values: `ALLGATHER`, `ALLTOALL`, or unset/`None`

Default behavior is unchanged. Setting this field changes the A5 prefill MoE communication method. Invalid values now fail at startup.

## How was this patch tested?

- Added unit tests in `tests/ut/test_ascend_config.py`:
  - default value is `None`
  - lowercase/whitespace value is normalized
  - invalid value is rejected

- Added unit tests in `tests/ut/test_ascend_forward_context.py`:
  - `ALLGATHER` override is selected for A5 prefill
  - `ALLTOALL` override is selected for A5 prefill
  - decode path still uses `MC2` when token count is within MC2 capacity

- Test command:

```bash
pytest -q tests/ut/test_ascend_config.py tests/ut/test_ascend_forward_context.py

- vLLM version: v0.27.1


zzzzzz198:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

PR #14426, opened 2026-08-17
[Feature][Ops] Add Kimi K3 AscendC operators

maoxx241:

- Adds AscendC KDA kernels for chunk prefill, state-layout conversion, and recurrent decode.
- Extends chunk gated delta rule for the Kimi K3 execution shapes.
- Adds Dequant-SiTU for A2/A3 and MX-SiTU for A5.
- Registers build targets, `_C_ascend` bindings, meta functions, and nightly operator coverage.

### KDA and GDN boundary

KDA reuses the shared chunk gated-delta-rule primitive, but it is not an alias for the existing GDN execution path. Kimi K3 additionally needs gate cumulative-sum preprocessing, KDA-specific recurrent state/layout conversion, and the ordered `KdaLayoutSwap12` device-copy boundary used between split forward stages. The KDA adapters own those K3 tensor and layout contracts; Dequant-SiTU and MX-SiTU are independent activation-quantization operators.

### Dependencies

None. This is the device-operator foundation for the Kimi K3 runtime PRs.

- `git diff --check`
- Python syntax compilation for the repository-level operator tests
- Ascend operator builds for KDA, recurrent KDA, Dequant-SiTU, and chunk gated delta rule
- Direct NPU coverage for KDA prefill/state conversion, recurrent KDA, Dequant-SiTU, and chunk gated delta rule
- MX-SiTU A5 cross-compilation and packaging

No direct API change. The Kimi K3 runtime consumes these operators.

- vLLM version: v0.27.1


csrc/attention/chunk_kda_fwd/op_host/op_api/aclnn_chunk_kda_fwd.cpp
```diff
@@ -0,0 +1,906 @@
...
+    return numel;
+}
+
+aclnnStatus KdaFwdCopyMaybeCastAfter(const aclTensor *src, const aclTensor *dependency,
+                                     const aclTensor *dst, aclOpExecutor *executor)
+{
+    const aclTensor *castSrc = KdaFwdMaybeCast(src, dst->GetDataType(), executor);
+    CHECK_RET(castSrc != nullptr, ACLNN_ERR_INNER_NULLPTR);
+    const aclTensor *linearSrc = l0op::Reshape(castSrc, KdaFwdMakeShape({1, 1, 1, KdaFwdNumel(castSrc)}), executor);
+    CHECK_RET(linearSrc != nullptr, ACLNN_ERR_INNER_NULLPTR);
```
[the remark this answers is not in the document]
maoxx241: This path intentionally calls the internal `l0op::KdaLayoutSwap12`, not the public `aclnnKdaLayoutSwap12` wrapper. The split-forward intermediates already use the destination BNSD layout; flattening to `[1, 1, 1, numel]` makes the swap kernel an identity and reuses its dependency input as an ordered device-copy barrier. Reshaping to the swapped destination shape would perform an extra transpose and corrupt the already-correct layout. The current head adds an inline comment documenting this distinction.

maoxx241:

#### KDA operator pipeline

- `csrc/attention/chunk_kda_fwd/` implements chunk-prefill KDA. The host side validates tensor contracts and selects tiling; the kernel computes the chunked recurrent update and writes both attention output and the final recurrent state.
- `csrc/attention/kda_gate_cumsum/` converts the per-token decay gate into the cumulative form consumed by chunk KDA.
- `csrc/attention/kda_layout_swap12/` converts the recurrent-state layout at the boundary between the framework representation and the recurrent kernel.
- `csrc/attention/recurrent_kda/` implements the single-step decode path. Its tiling and state update are separated from chunk prefill so decode does not pay the chunk setup cost.

The runtime data flow is:

`gate -> gate_cumsum -> chunk_kda_fwd` for prefill, and
`gate -> recurrent_kda` for decode, with `kda_layout_swap12` used at the state-layout boundary.

#### Chunk gated delta rule

`csrc/moe/chunk_gated_delta_rule_fwd_h/` extends the existing operator to the Kimi K3 shapes and recurrent-state update contract. The arch20/arch22/arch35 schedulers and epilogues keep the same external API while selecting the architecture-specific block scheduling and state writeback.

#### SiTU operators

- `csrc/moe/dequant_situ_quant/` fuses dequantization, SiTU activation, and output quantization for the A2/A3 W4A8/W8A8 paths.
- `csrc/moe/situ_mx_quant/` performs SiTU plus MX quantization for A5, including the E4M3FN output descriptor used by the following grouped matmul.
- Both operators consume `beta` and `linear_beta`, so the framework does not need to reconstruct SiTU with separate elementwise kernels.

#### Build and registration

- `csrc/build_aclnn.sh` and the per-operator CMake files add the new targets to the custom-op package.
- `csrc/torch_binding.cpp` exposes the runtime bindings.
- `csrc/torch_binding_meta.cpp` defines output shape/dtype contracts used during graph construction.

#### Coverage

The nightly operator cases cover chunk/recurrent KDA, state-layout conversion, chunk gated delta rule, Dequant-SiTU, MX-SiTU integration shapes, non-contiguous recurrent state, and output comparison against reference implementations.

ZT-AIA (commented):
Outstanding issue: Supplement the operator documentation, especially highlighting the differences from GDN.

ZT-AIA (commented):
Having redundant function templates can subsequently lead to increased compilation and installation times. Therefore, it is necessary to delete these redundant function templates later @Foriv

csrc/torch_binding.cpp
```diff
@@ -1881,6 +1884,320 @@ at::Tensor chunk_fwd_o(
...
+    at::Tensor final_state = output_final_state.value_or(false) ?
+        final_state_work : at::empty({0}, q.options().dtype(at::kFloat));
+    at::Tensor empty = at::empty({0}, q.options());
+    at::Tensor g = gk.scalar_type() == at::kFloat ? gk : gk.to(at::kFloat);
+    at::Tensor initial_state_out = initial_state_tensor.defined() ? initial_state_tensor : empty;
+    (void)return_intermediate;
+    return std::make_tuple(o, final_state, g, aqk, akk, w, u, qg, kg, v_new, h, initial_state_out);
+}
+
+at::Tensor kda_gate_cumsum(
```
ZT-AIA: If possible, place these functions under the operator directory in a newly created file, rather than placing them inside torch_binding.

csrc/moe/situ_mx_quant/op_kernel/inc/platform.h
```diff
@@ -0,0 +1,81 @@
...
+#else
+    return false;
+#endif
+}
+
+/**
+ * Get the block size of unified buffer in bytes
+ */
+__aicore__ inline constexpr uint32_t GetUbBlockSize() { return 32U; }
+
```
ZT-AIA, line 48: Why do I remember that it uses `ascendcPlatform.GetCoreMemSize(platform_ascendc::CoreMemType::UB, ubSize));`? This system function is used to obtain the size of the UB.

csrc/attention/kda_layout_swap12/op_host/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+if (BUILD_OPEN_PROJECT)
+    target_sources(op_host_aclnnExc PRIVATE
+        kda_layout_swap12_def.cpp
+    )
+endif()
+
+add_modules_sources(OPTYPE kda_layout_swap12 ACLNNTYPE aclnn_exclude)
+add_ops_compile_options(
+    OP_NAME KdaLayoutSwap12
+    OPTIONS --cce-auto-sync=off
```
ZT-AIA, line 19: This switch seems to have different settings for each operator. Please check whether there are any risks.

csrc/attention/kda_layout_swap12/op_host/kda_layout_swap12_def.cpp
```diff
@@ -0,0 +1,48 @@
...
+ * the BSD 3-Clause License (the "License"). Please refer to the License for details.
+ * THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND.
+ */
+
+#include "register/op_def_registry.h"
+
+namespace ops {
+class KdaLayoutSwap12 : public OpDef {
+public:
+    explicit KdaLayoutSwap12(const char *name) : OpDef(name)
```
ZT-AIA, line 15: What does the 12 in this operator name mean?

ZT-AIA (dismissed):

csrc/attention/kda_layout_swap12/op_host/CMakeLists.txt
maoxx241: Checked the kernel: it explicitly orders MTE2/MTE3 with SetFlag and WaitFlag in both copy paths. Disabling compiler auto-sync avoids duplicating those barriers. I documented the reason next to the compile option in 0e8d3a2ef.

csrc/torch_binding.cpp
maoxx241: Moved the ChunkKDA, KdaGateCumsum, and KdaLayoutSwap12 PyTorch adapters out of torch_binding.cpp into operator adapter headers in 0e8d3a2ef. torch_binding.cpp now only includes and registers them.

csrc/attention/kda_layout_swap12/op_host/kda_layout_swap12_def.cpp
maoxx241: The name means swapping logical axes 1 and 2 for rank-4 tensors. The equivalent TND/NTD rank-3 path swaps axes 0 and 1. This is now documented in the op definition and adapter in 0e8d3a2ef.

csrc/moe/situ_mx_quant/op_kernel/inc/platform.h
maoxx241: GetUbBlockSize returns the 32-byte UB data-movement alignment unit, not the total UB capacity. The total capacity is already queried by the host tiling code with GetCoreMemSize in situ_mx_quant_tiling_arch35.cpp. I clarified that distinction in 0e8d3a2ef.

maoxx241:
Added a dedicated KDA/GDN boundary section to the PR description. It now states which chunk gated-delta-rule primitive is reused and which gate-cumsum, recurrent-state/layout, and ordered layout-swap contracts remain KDA-specific. This keeps the operator documentation in the PR without restoring the csrc README/test directories removed from this change.

maoxx241:
Re-evaluated the template-compilation concern against the current operator sources. Commit 8bd56a392 removes the duplicate local `IntegralConstant`/`IsSame` definitions from `situ_mx_quant_common.h` and reuses `ops::IsSame` from the already-included `kernel_utils.h`. I retained the remaining KDA/SiTU kernel templates after checking their call sites: they select distinct input/output/state dtypes, gate formats, linear-beta variants, or architecture paths, so deleting them would remove compiled operator variants rather than only reduce duplicate instantiation.

maoxx241:

PR #14428, opened 2026-08-17
[Performance] Fuse index+copy into single out= gather in get_cos_and_sin_dsa

ivyilike:
fuse the two-op "index-then-copy" pattern into a single torch.gather(..., out=...) that writes straight into the persistent runtime buffer. Per (cos, sin) pair this drops 2 GPU dispatches and 2 intermediate tensor allocations (4 ops -> 2 ops), eliminating host-side allocation/GC.

No. This is an internal performance optimization; there is no API, interface, or behavior change.

get_cos_and_sin_dsa single-call wall time on the decode path: 814us -> 701us
(~14% faster).


vllm_ascend/ops/rope_dsv4.py
```diff
@@ -119,19 +116,24 @@ def get_cos_and_sin_dsa(
                 buf_cos, buf_sin = group_buffers
                 num_tokens = pos_tensor.size(0)
 
+                gather_idx = (
```
maoxx241, line 130: Please add a brief comment explaining the logic used to construct `gather_idx`, especially why reshaping and expanding `pos_tensor` produces the index expected by `torch.gather`. Please also state that this is semantically equivalent to the previous `full_rope_cos[pos_tensor]` / `full_rope_sin[pos_tensor]` indexing followed by `copy_`; the change only combines the indexing and the write into the preallocated output buffers. Please update this before the PR is merged.

maoxx241 (changes requested):
Please address the inline comment and add the requested explanation before this PR is merged.

ivyilike:
> Please address the inline comment and add the requested explanation before this PR is merged.
@maoxx241 , I've added the requested comment for gather_idx. If it's acceptable, could you please add the ready-to-merge label and re-trigger the cancelled CI jobs? Thanks!

yiz-liu (approved):

PR #14402, opened 2026-08-17
[v0.26.0rc][Bugfix][DCP] Use local block table for replicated metadata and avoid double metadata build

coder-fny:
### What this PR does / why we need it?
This PR fixes a prefix-cache performance regression in SFA DCP on v0.26.
image

The root cause had two parts:
- DCP replicated metadata was sized from the global BlockTable width instead of the rank-local DCP width, which made the replicated view and compact KV-gather path scan too many block columns.
- `AscendSFADCPMetadataBuilder` in the v0.26 path routed through the parent `build()`/`build_for_drafting()` implementation, which introduced a nested metadata-view wrapper and caused the DCP path to see an already replicated block table.

The patch:
- limits SFA DCP replicated metadata to the local DCP block-table width,
- crops the input block table before building replicated slot mapping and compact KV-gather metadata,
- and calls `_build()` directly inside the DCP metadata-view wrapper to avoid nested wrapping.

Fixes #13973 

### Does this PR introduce _any_ user-facing change?
Yes, but only in runtime behavior and performance.

There is no API or interface change. The visible effect is that prefix-cache-enabled prefill throughput in v0.26 no longer regresses due to the DCP metadata path.

### How was this patch tested?
Validated on the v0.26 pod with prefix cache enabled.
image

- vLLM version: v0.26.0
- vLLM main: https://github.com/vllm-project/vllm/commit/d02df748bf9efd99022f1a062597dc3cb3808485

PR #14435, opened 2026-08-17
[main][Bugfix][DCP] Use local block table for replicated metadata and avoid double metadata build

coder-fny:
Partially fix PR #14402   changes on the mainline version

DCP replicated metadata was sized from the global BlockTable width instead of the rank-local DCP width, which made the replicated view and compact KV-gather path scan too many block columns.

test result：

image

- vLLM version: v0.27.1


pisceskkk (approved):
LGTM, thanks for the work!

kunpengW-code (approved):

PR #14451, opened 2026-08-17
[BugFix] Fix GLM5.2 MTP bug that dose not utilize shared index

lijiahang226:

GLM5.2 MTP (multi-token prediction) with index sharing (`index_share_for_mtp_iteration`) was broken on Ascend because the runtime index-cache state was not correctly exposed or initialized:

1. The actual indexer gate (`skip_topk`) and the shared top-k buffer (`topk_indices_buffer`) live in the inner MLA implementation (e.g. `AscendSFAImpl`), which is not an `nn.Module` and is therefore invisible to `named_modules()`. Upstream `DeepSeekMultiTokenPredictor` hooks call `set_skip_topk` / `compact_topk_indices` on the module, so those calls never reached the inner implementation on Ascend.
2. MTP/nextn layers could be initialized with `skip_topk=True` from the backbone skip pattern. They must never start in skip mode: they compute their own indices at draft step 0 and toggle at runtime via `set_skip_topk` (`index_share_for_mtp_iteration`).
3. After draft step 0, the shared top-k buffer rows were not compacted to request-aligned positions, so steps 1+ could read the wrong rows.

This PR fixes the Ascend MTP index-cache path by:

- Adding `skip_topk` and `topk_indices_buffer` properties on `AscendMultiHeadLatentAttention` that forward to the inner MLA impl, so upstream proposer hooks keep working on Ascend.
- Masking the initial `skip_topk` with `not is_mtp_layer` in `_deepseek_v2_mla_attention_init`, matching upstream `deepseek_v2.py` behavior for MTP layers.
- Calling `compact_topk_indices(token_indices_to_sample)` after draft step 0 in `AscendSpecDecodeBaseProposer` when the draft model supports it, mirroring the upstream proposer. Draft predictors without this method are left untouched.

Related to #15864

No user-facing API or configuration change. This is an internal fix for GLM5.2 MTP index-cache sharing on Ascend.

- Unit tests: `pytest -sv tests/ut/ops/test_mla.py tests/ut/spec_decode/a2/test_eagle_proposer.py`

### DIFF
--- a/tests/ut/ops/test_mla.py
+++ b/tests/ut/ops/test_mla.py
@@ -211,7 +211,8 @@ class TestAscendMultiHeadLatentAttention(TestBase):
         mock_mla_attn.impl = MagicMock()
         mock_mla_attn.impl.process_weights_after_loading = MagicMock()
 
-        with patch("vllm_ascend.ops.mla.MLAAttention", return_value=mock_mla_attn):
+        with patch("vllm_ascend.ops.mla.MLAAttention",
+                   return_value=mock_mla_attn) as mock_mla_attn_cls:
             mock_tp_size.return_value = 2
             mock_vllm_config = MagicMock(spec=VllmConfig)
             mock_vllm_config.model_config.hf_text_config = MagicMock(num_hidden_layers=32, first_k_dense_replace=True)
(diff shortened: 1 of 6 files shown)


lijiahang226:
/rerun

lijiahang226:
/rerun

lijiahang226:
/rerun

lijiahang226:
/rerun

lijiahang226:
/rerun

PR #14513, opened 2026-08-18
[v0.26.0rc][BugFix][MoE] Restore FlashComm1 shared expert DP default

maoxx241:
## What changed

- Restore the historical behavior that FlashComm1 defaults shared experts to data parallelism when `enable_shared_expert_dp` is not set.
- Preserve the explicit override: users can set `enable_shared_expert_dp: false` to use shared-expert TP.

## Why

#12950 decoupled the two options and changed the implicit default to `false`. Existing FlashComm1 deployments that relied on the previous default therefore silently switched their shared experts from DP to TP.

This PR only restores the compatibility default. It does not change the shared-expert execution or stream ordering paths.

## Validation

- `python -m pytest -q tests/ut/test_ascend_config.py`: 41 passed
- Added coverage for the FlashComm1 default and the explicit `false` override.
- A5 8-card startup and 2k/1k serving validation completed with shared-expert DP enabled.

- vLLM version: v0.26.0

### DIFF
--- a/tests/e2e/pull_request/four_card/context_parallel/test_accuracy.py
+++ b/tests/e2e/pull_request/four_card/context_parallel/test_accuracy.py
@@ -68,8 +68,8 @@ DEEPSEEK_V4_PROMPTS = [
     "What is the meaning of life?",
 ]
 
-# The routed-expert SwiGLU clamp changes the second continuation (#14397).
-DEEPSEEK_V4_GOLDEN = ["Hello, my name is {name} and I", "What is the meaning of life?', 'What is the"]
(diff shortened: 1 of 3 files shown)


maoxx241:
/rerun
[Bot]: rerun completed.

**Failed:**

maoxx241:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

maoxx241:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

maoxx241:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

kunpengW-code (approved):

PR #14439, opened 2026-08-17
[BugFix][MoE] Fix MegaMoe prefill buffer sizing

jianzs:
### What this PR does / why we need it?

CANN MegaMoe allocates its symmetric communication buffer once from the MC2 token capacity. When `enable_prefill_mc2` is disabled, that capacity was derived from the decode graph capture size even though eager prefill could still select MegaMoe through fused MC2. A larger prefill batch could therefore exceed the allocated capacity and produce invalid outputs.

This PR:

- Sizes the MegaMoe buffer from `max_num_batched_tokens` when MegaMoe is actually eligible.
- Centralizes MegaMoe eligibility in `use_cann_megamoe` so communication selection, buffer sizing, DP token handling, and BF16/W4A8/W8A8 weight preparation use the same decision, while preserving the `dispatch_ffn_combine` fallback.
- Keeps DP token counts uniform for MegaMoe, without disabling the uneven-token optimization for `dispatch_ffn_combine`.
- Raises an explicit error if a MegaMoe input exceeds the allocated per-rank capacity.

Fixes #14273.

Related release-branch PR: #14358.

### Does this PR introduce _any_ user-facing change?

Yes. MegaMoe now allocates enough symmetric-buffer capacity for eager prefill batches when prefill MC2 is disabled. Non-MegaMoe configurations continue to use their existing communication path, and an invalid capacity is reported explicitly instead of producing incorrect output.

### How was this patch tested?

- `git diff --check`
- `python3 -m py_compile vllm_ascend/ascend_forward_context.py vllm_ascend/ops/fused_moe/moe_comm_method.py vllm_ascend/ops/fused_moe/routed_experts.py vllm_ascend/quantization/methods/w4a8.py vllm_ascend/quantization/methods/w8a8_dynamic.py vllm_ascend/utils.py`
- `ruff check vllm_ascend/ascend_forward_context.py vllm_ascend/ops/fused_moe/moe_comm_method.py vllm_ascend/ops/fused_moe/routed_experts.py vllm_ascend/quantization/methods/w4a8.py vllm_ascend/quantization/methods/w8a8_dynamic.py vllm_ascend/utils.py`
[...]

PR #14519, opened 2026-08-18
[v0.25.1rc][BugFix][MoE] Fix MegaMoe prefill buffer sizing

QwertyJack:

CANN 9.1 makes `cann_ops_transformer` available, so the v0.25 fused-MC2 path can select CANN MegaMoe instead of the legacy `dispatch_ffn_combine` implementation. When `enable_prefill_mc2` is disabled, MegaMoe was still selected for eager prefill, but its symmetric buffer capacity was derived from the decode graph capture size. A larger prefill batch could therefore exceed the allocated capacity and produce invalid output instead of failing explicitly.

This PR backports the fix from #14358 to `releases/v0.25.1rc`:

- Size the MegaMoe buffer from `max_num_batched_tokens` whenever MegaMoe is eligible.
- Centralize MegaMoe eligibility so communication selection, buffer sizing, DP token handling, and W4A8/W8A8 weight preparation use the same decision while preserving the `dispatch_ffn_combine` fallback.
- Keep DP token counts uniform for MegaMoe without disabling the uneven-token optimization for `dispatch_ffn_combine`.
- Raise an explicit error if the runtime input exceeds the allocated per-rank MegaMoe capacity.
- Add a focused regression test for buffer sizing when prefill MC2 is disabled.

Refs #14273.

Related #14358 and #14439.

Yes. Eligible MegaMoe configurations now allocate enough symmetric-buffer capacity for eager prefill batches when prefill MC2 is disabled. Non-MegaMoe configurations retain their existing communication path, and an invalid capacity is reported explicitly instead of causing silent corruption.

- `git diff --check upstream/releases/v0.25.1rc...HEAD`
- `python3 -m py_compile vllm_ascend/ascend_forward_context.py vllm_ascend/ops/fused_moe/moe_comm_method.py vllm_ascend/quantization/methods/w4a8.py vllm_ascend/quantization/methods/w8a8_dynamic.py vllm_ascend/utils.py tests/ut/test_ascend_forward_context.py`
- `ruff check tests/ut/test_ascend_forward_context.py vllm_ascend/ascend_forward_context.py vllm_ascend/ops/fused_moe/moe_comm_method.py vllm_ascend/quantization/methods/w4a8.py vllm_ascend/quantization/methods/w8a8_dynamic.py vllm_ascend/utils.py`
- `python3 -m pytest tests/ut/test_ascend_forward_context.py -q` (`26 passed`)

The underlying capacity fix was also validated on the affected v0.25 stack with a two-node DeepSeek V4 W8A8 PD deployment. The original path reproduced invalid target logits and EngineDead after a 477-token prefill exceeded the 32-token-per-rank MegaMoe buffer. With capacity derived from `max_num_batched_tokens`, a short request, the 477-token request, and three concurrency-16 rounds completed successfully (`48/48` HTTP 200 with valid JSON and no NaN, token-0, all-`-1`, `507035`, traceback, or EngineDead signature). A separate 32-rank HCCL all-reduce smoke test also returned the expected value on every rank.

The final backport and focused regression test were verified with the unit and static checks above; the full two-node run was not repeated after packaging the commits for this PR.

- vLLM version: v0.25.1

### DIFF
--- a/tests/ut/test_ascend_forward_context.py
+++ b/tests/ut/test_ascend_forward_context.py
@@ -118,6 +118,25 @@ def test_set_mc2_tokens_capacity_prefill_mc2_uses_max_num_batched_tokens(monkeyp
     assert afc.get_mc2_tokens_capacity() == 520
 
 
+def test_set_mc2_tokens_capacity_megamoe_uses_max_num_batched_tokens(monkeypatch):
+    monkeypatch.setattr(
+        afc,
+        "get_ascend_config",
+        lambda: SimpleNamespace(enable_prefill_mc2=False, enable_fused_mc2=1),
+    )
+    monkeypatch.setattr(afc, "use_cann_megamoe", lambda _: True)
+    vllm_config = _make_vllm_config(
+        tensor_parallel_size=8,
+        cudagraph_capture_sizes=[128],
+        max_cudagraph_capture_size=128,
+        max_num_batched_tokens=513,
+    )
+
+    afc.set_mc2_tokens_capacity(vllm_config, max_num_reqs=16, uniform_decode_query_len=1)
+
+    assert afc.get_mc2_tokens_capacity() == 520
+
+
 def test_select_moe_comm_method_returns_none_for_non_moe(monkeypatch):
     _patch_select_moe_comm_method_deps(
         monkeypatch,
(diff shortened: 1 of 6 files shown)


QwertyJack:
Superseded by #14530. For v0.25, the lower-risk release policy is to keep fused MC2 on the legacy `dispatch_ffn_combine` implementation rather than backport the MegaMoe buffer and path refactors. Main and v0.26 retain the relevant MegaMoe fixes in #14439 and #14358.

Issue #2575, opened 2025-08-27
[Bug]:Qwen3-30B-A3B推理速度与N卡差距过大


zdfb:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
Your output of above commands here
```

</details>

### 🐛 Describe the bug

Vllm-Ascend推理Qwen3-30B-A3B，速度与Nvidia下同样使用VLLM相差10倍

![Image](https://github.com/user-attachments/assets/6aa912a2-bd62-4d0b-bf26-37ef1a315244)

Nvidia-A800下的推理速度

![Image](https://github.com/user-attachments/assets/3e6b8589-7bd8-4915-b706-28997cf7cd7e)

910B3下的推理速度

lilinsiman:
Please try to use the latest vllm and vllmascend code for testing.
vllm v0.20.2
vllm_ascend latest main branch

PR #14528, opened 2026-08-19
[Refactor][Worker] Retire redundant worker patches

kokomicx:
Retire two worker-level monkey patches whose behavior is now provided by the pinned upstream dependencies:

- Remove `patch_npugraph_ex_triton.py` and use the upstream Triton `ValuePack` handling introduced by [TorchAir #2575](https://gitcode.com/Ascend/torchair/pull/2575).
- Remove `patch_process_weights_after_loading.py` and use the shared post-load weight-processing lifecycle introduced by [vLLM #49519](https://github.com/vllm-project/vllm/pull/49519).
- Remove the corresponding patch loading and inventory entries.
- Add focused regression tests for DSA attention, Ascend custom loaders, and `npugraph_ex` `ValuePack` handling.

This PR retires the following worker patches:

1. `patch_npugraph_ex_triton.py`

   This patch previously extended `npugraph_ex` argument unpacking so that Triton `ValuePack` arguments could be used with ACL Graph capture. The supported `npugraph_ex`/TorchAir dependency now handles both metadata and NPU tensor components directly, so the local monkey patch is no longer required.

2. `patch_process_weights_after_loading.py`

   This patch previously replaced vLLM's post-load weight-processing helper to recognize `DSAAttention` and updated references imported by Ascend custom loaders.

   The pinned vLLM revision now provides a shared `AttentionLayerBase` post-load lifecycle through vLLM #49519. Since `DSAAttention` implements this interface, the default loader, layerwise loading paths, NetLoader, and RFork loader can use the upstream implementation directly.

Removing these patches reduces global runtime mutation and avoids maintenance drift when the corresponding upstream implementations change.

No.

This is an internal cleanup. Existing DSA post-load weight processing and Triton `ValuePack` handling are preserved through the upstream implementations.

#### Unit tests

The focused regression tests cover both retired worker patches:

DSA post-load processing

Verify that DSAAttention inherits from AttentionLayerBase.

Verify that the upstream process_weights_after_loading() dispatch recognizes DSAAttention and invokes its post-load hook with the expected activation dtype.

Verify that the Ascend NetLoader and RForkLoader directly use the upstream post-load processing function without monkey patching it.

npugraph_ex Triton ValuePack handling

Construct a ValuePack containing separate meta and NPU values.

Pass it through nested positional and keyword arguments.

Verify that the upstream _unpack_meta() extracts the meta values correctly.

Verify that the upstream _NpuGraphConverter._unpack_npu() extracts the NPU values correctly.

This confirms that the current npugraph_ex/TorchAir implementation provides the behavior previously supplied by patch_npugraph_ex_triton.py.

#### DSA end-to-end validation (Ascend 910/A3)

Model: DeepSeek-V3.2 W8A8 DSA validation checkpoint

DSA configuration: index_topk=2048

Parallelism: TP4 / EP4

All four ranks completed model weight loading successfully.

KV cache profiling, KV cache creation, model warmup, and real generation completed successfully.

The request generated 8 output tokens with finish_reason="length".

Process exit code: 0

Result: PR14528_DSA_V32_E2E_VALIDATION_OK

The validation host did not provide the optional aclnnAddRmsNormBias operator, so the test used the existing torch_npu fallback path. No PR source code was modified for this validation.

image


tests/ut/patch/worker/test_worker_patch_sunsets.py
```diff
@@ -0,0 +1,75 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+import importlib
+from types import SimpleNamespace
```
zhenwenqi2024: 这个test文件的看护点在哪里？

kokomicx: It verifies functionality rather than guarding a specific change, so I’ll remove it.

Ronald1995 (approved):

weiguihua2 (approved):

Issue #14378, opened 2026-08-16
[Bug][0.23.0]: Accuracy Fluctuations with GLM-5.2-W4A8C8 on Ascend


underfituu:
### Description

  We observed accuracy fluctuations when serving `GLM-5.2-W4A8C8` with Prefill-Decode (PD) disaggregation and enabling Sparse Flash Attention C8:

  ```bash
  --additional-config '{"enable_sparse_sfa_c8": true}'
  ```

  When GPQA is evaluated repeatedly with the same model, dataset, and decoding configuration, the accuracy fluctuates by approximately 2–3 percentage points.

  GLM-5.2-W4A8C8 should currently be considered experimental. Although it reduces memory usage, its accuracy stability may not match GLM-5.2-W8A8C8, particularly in PD deployments with enable_sparse_sfa_c8 enabled.

  ### Recommendation for users

  For production and accuracy-sensitive workloads, we recommend using GLM-5.2-W8A8C8 by default.

  Use GLM-5.2-W4A8C8 only for experimental or memory-constrained deployments, and validate its accuracy on the target workload before production use. In PD disaggregation, enable_sparse_sfa_c8 should remain disabled unless it is required to meet memory constraints and the resulting accuracy has been carefully evaluated.

  ### Expected behavior

  Repeated evaluations with the same configuration should produce stable accuracy within normal evaluation variance.

  ### Actual behavior

  With GLM-5.2-W4A8C8 and enable_sparse_sfa_c8 enabled in PD disaggregation, GPQA accuracy varies by approximately 2–3 percentage points across runs.

slippersss:
@wangzhishenghw @bowgneo Please take a look. This issue is about glm5.2 with pd disaggregation and c8 quantization. Thanks.

GoCHug:
https://github.com/vllm-project/vllm-ascend/pull/14536 加上该pr首请求还是存在乱码

PR #14536, opened 2026-08-19
[BugFix][SFA] Use DCP-sharded slot mapping in prolog v3 KV write

bnbryan:
#### How it was found

Found while debugging garbled decode output on **GLM5.2** in a PD-disaggregated deployment with **w4a8** weights and `enable_sparse_sfa_c8` enabled. The prefill node was unaffected; only the decode node produced wrong results.

Four conditions must hold at once for the bug to trigger:

1. `enable_sfa_prolog_v3` requires `is_kv_consumer`, so only the **D node** of a PD-disaggregated deployment takes this path;
2. it also requires `enable_sparse_sfa_c8`;
3. `_get_sfa_prolog_v3_unsupported_reasons` requires w8a8-dynamic `fused_qkv_a_proj` and `q_proj`, which the w4a8 checkpoint satisfies;
4. the two slot mappings only diverge when `decode_context_parallel_size > 1` (see `enable_sfa_dcp_replicated_indexer` in `vllm_ascend/utils.py`).

#### Root cause

Under DCP, `AscendSFADCPMetadataBuilder` maintains two different slot mappings, as documented by the "SFA DCP replicated-indexer layout" comment in `vllm_ascend/attention/context_parallel/sfa_cp.py`:

- `attn_metadata.slot_mapping` is the **replicated** view. The LightningIndexer cache is replicated on every DCP rank so index selection runs against the full sequence and keeps the same sparse top-k semantics as non-DCP SFA.
- `attn_metadata.dcp_context.slot_mapping` is the **DCP-local** view. The SFA KV cache stays sharded across DCP ranks to preserve the KV memory saving.

`AscendSFAImpl.forward` already derives `slot_mapping_sfa` for exactly this purpose and uses it on the native path — `exec_kv`, the packed-C8 `npu_scatter_nd_update_`, and `reshape_and_cache`. The SFA Prolog V3 decode branch was the only KV-writing path still passing the replicated `attn_metadata.slot_mapping` into `_sfa_preprocess_with_prolog_v3`, which forwards it to `npu_mla_prolog_v3` as `cache_index`. The prolog therefore wrote SFA KV into indices belonging to the replicated layout, so subsequent decode steps read back KV from the wrong slots.

#### Changes

- Pass `slot_mapping_sfa` instead of `attn_metadata.slot_mapping` to `_sfa_preprocess_with_prolog_v3`.
- Update the assert above the call to validate the same tensor that is actually passed. Both tensors are sliced to `num_input_tokens`, so this does not change which batches pass the assert; it only keeps the check and its error message consistent with the value in use.

No API or configuration change. It is a correctness fix: with SFA Prolog V3 and DCP both enabled, decode output was previously incorrect due to mismatched KV cache indices, and is now correct.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/attention/sfa_v1.py
+++ b/vllm_ascend/attention/sfa_v1.py
@@ -1500,7 +1500,7 @@ class AscendSFAImpl(MLAAttentionImpl):
 
         cos = attn_metadata.cos
         sin = attn_metadata.sin
-        slot_mapping = attn_metadata.slot_mapping
+        slot_mapping_li = attn_metadata.slot_mapping
         slot_mapping_sfa = self._get_sfa_kv_slot_mapping(attn_metadata)
 
         # Inputs and outputs may be padded for CUDA graphs
@@ -1551,7 +1551,7 @@ class AscendSFAImpl(MLAAttentionImpl):
                     kv_cache=kv_cache,
                     cos=cos,
                     sin=sin,
-                    slot_mapping=slot_mapping,
+                    slot_mapping=slot_mapping_sfa,
                     num_input_tokens=num_input_tokens,
                 )
         # native
@@ -1642,7 +1642,7 @@ class AscendSFAImpl(MLAAttentionImpl):
             else:
                 torch_npu.npu_scatter_nd_update_(
                     kv_cache[dsa_k_cache_idx].view(-1, k_li.shape[-1]),
-                    slot_mapping.view(-1, 1),
+                    slot_mapping_li.view(-1, 1),
                     k_li.view(-1, k_li.shape[-1]),
                 )  # b, s, n, d
             if self.enable_sparse_li_c8:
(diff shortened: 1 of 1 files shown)


bnbryan:
This may be the root cause behind #13983 and #14378 — both report accuracy problems on GLM-5.2-W4A8C8 with `enable_sparse_sfa_c8`, which is exactly the configuration that selects the path fixed here. `releases/v0.23.0` carries the same defect (`sfa_v1.py:1699`), so both reports are on affected code.

bnbryan:
@wangzhishenghw @bowgneo PTAL — this touches the GLM-5.2 W4A8C8 + `enable_sparse_sfa_c8` path discussed in #13983 and #14378.

pisceskkk (approved):
LGTM

pisceskkk (approved):

weijinqian0 (approved):

bnbryan:
[Bot]: cherry-pick command failed. Please check the [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/32924220164) for details.

PR #14549, opened 2026-08-19
[Bugfix][PD] Support P-side pipeline parallelism in MooncakeHybridConnector

AceCoder0:
## Description

MooncakeHybridConnector (the HMA path for hybrid SSM/attention models, e.g. Qwen3.5) could not serve as the PD prefill instance with `pipeline-parallel-size > 1`. This PR fixes three stacked bugs; the second commit is an independent MRV2 crash fix found while testing the same configuration.

### Commit 1: hybrid connector PP fixes (`mooncake_hybrid_connector.py`)

1. **Startup crash** — the connector lacks `set_xfer_handshake_metadata_pp_aware`; with PP>1 the engine core passes `(pp_rank, tp_rank)`-keyed handshake metadata and the upstream base class raises `does not support PP-disaggregated KV transfer`. Fixed by storing worker metadata keyed by handshake port offset (`pp_rank * tp_size + tp_rank`), matching how the recv thread resolves peers in `_get_remote_host_info_by_port`.
2. **Missing transfers** — the `use_mamba` branch of `start_load_kv` issued a single pull (first peer only), so PP stages > 0 never transferred their KV and decode read stale memory. Now iterates all peers like the non-mamba branch; per-stage pull count is derived from `_get_remote_ranks_for_req` (one peer per stage for equal-TP GQA) rather than the mamba `tp_num_need_pulls = tp_size` convention, which mismatches that rank array.
3. **Wrong destination groups** — `_transfer_kv_cache_all_groups` zipped D's full group-major address list against the remote stage worker's shorter list, so pulls from stage 1 landed in D's groups 0..N/2-1, overwriting stage-0 data at wrong layer positions → garbage output. Now the local address list (plus aligned `block_len`/`block_stride`/`addr_group_idx` arrays) is offset to the remote stage's global group range before zipping.

### Commit 2: duplicate `is_prefilling` kwarg (`attn_utils.py`)

`build_attn_metadata` passes `is_prefilling` explicitly and also forwards `model_specific_attn_metadata.get_extra_common_attn_kwargs()`, which for `MambaHybridAttnMetadata` (vllm `58d3918e3`, the current verified commit) already returns `is_prefilling` → `TypeError` on the first forward of a mamba-hybrid model under MRV2. Can be split into a separate PR if maintainers prefer.

## Test Plan

- [x] Ascend 910, vllm @ verified commit `58d3918e3`, Qwen3.5-27B (hybrid GatedDeltaNet + attention, 64 layers, 4-layer HMA cache groups)
- [x] Topology: P = PP2 × TP2 (`kv_producer`) + D = TP2 (`kv_consumer`), `MooncakeHybridConnector`, MRV2 on both sides, single-node proxy
- [x] Before: startup ValueError (bug 1); after bypassing it: decode outputs garbage (bugs 2+3)
- [x] After: 3 distinct prompts decode correctly, matching the `MooncakeConnectorV1` baseline on the same topology; per-request logs show 4 cross-stage pulls (2 stages × 2 TP ranks) with 24↔24 address pairs each
- [ ] CI

## Notes

- Decode side still must be PP=1 (pre-existing hard constraint, unchanged).
- Non-uniform `prefill.pp_layer_partition` that does not align with HMA group boundaries is rejected with an assert.

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_hybrid_connector.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_hybrid_connector.py
@@ -667,6 +667,41 @@ class KVCacheRecvingThread(threading.Thread):
             remote_transfer_port = self.remote_te_port[remote_engine_id][remote_handshake_port]
         session_id = f"{remote_host}:{remote_transfer_port}"
 
+        # With P-side PP the remote worker only owns its stage's
+        # kv_cache_tensors. Offset the local (D) group-major flat addr list to
+        # the remote stage's global group range before zipping, so stage-1
+        # data lands in D's groups 8-15 instead of overwriting groups 0-7.
+        local_addrs = local_kv_caches_base_addrs
+        block_len_arr = self.block_len_per_addr
+        block_stride_arr = self.block_stride_per_addr
+        addr_group_arr = self.addr_group_idx
+        if self._prefill_pp_size > 1:
+            tp_num_need_pulls = req_meta["tp_num_need_pulls"]
+            remote_pp_rank = req_meta["offset"] // tp_num_need_pulls
+            num_groups = len(self.kv_cache_config.kv_cache_tensors)
+            groups_per_stage = num_groups // self._prefill_pp_size
+            addrs_per_group, rem = divmod(len(local_addrs), num_groups)
+            if rem != 0:
+                raise ValueError("non-uniform per-group address counts unsupported")
+            if num_groups % self._prefill_pp_size != 0:
+                raise ValueError("pp split not group-aligned")
+            start = remote_pp_rank * groups_per_stage * addrs_per_group
+            end = (remote_pp_rank + 1) * groups_per_stage * addrs_per_group
+            local_addrs = local_addrs[start:end]
+            block_len_arr = self.block_len_per_addr[start:end]
+            block_stride_arr = self.block_stride_per_addr[start:end]
+            if addr_group_arr:
+                addr_group_arr = self.addr_group_idx[start:end]
+            logger.debug(
+                "Cross-PP pull alignment: remote_port=%d -> pp_rank=%d local_addrs[%d:%d]=%d remote=%d",
+                remote_handshake_port,
+                remote_pp_rank,
+                start,
+                end,
+                len(local_addrs),
+                len(remote_kv_caches_base_addrs),
+            )
+
         req_start_time = time.perf_counter()
         src_list, dst_list, length_list = [], [], []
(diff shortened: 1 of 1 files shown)


AceCoder0:
All author-side checks are green now (pre-commit, select-tests, DCO). The remaining ci-gate failure is only waiting for the `ready` label to dispatch the selected tests.

For reference, the connector changes were validated on Ascend 910 hardware with Qwen3.5-27B (hybrid GatedDeltaNet): P = PP2×TP2 producer + D = TP2 consumer, MRV2 enabled — decode output was garbage before, correct (matching MooncakeConnectorV1 baseline) after. Happy to add a unit test if maintainers point me at the preferred pattern for connector coverage.

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_hybrid_connector.py
```diff
@@ -1177,6 +1210,10 @@ def set_xfer_handshake_metadata(self, metadata: dict[int, KVConnectorHandshakeMe
         assert self.connector_scheduler is not None
         self.connector_scheduler.set_xfer_handshake_metadata(metadata)
 
+    def set_xfer_handshake_metadata_pp_aware(self, metadata) -> None:
```
LostFox11: 这里没必要分成三个函数调用 嵌套太多了

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_hybrid_connector.py
```diff
@@ -1468,6 +1505,32 @@ def request_finished_all_groups(
             num_prompt_blocks=num_prompt_blocks,
         )
 
+    def _port_offset_from_handshake_metadata(self, rank_metadata: KVConnectorHandshakeMetadata, metadata_key) -> int:
```
LostFox11: 这几个判断条件真的都能走到吗 确认一下是否可以进行删减

AceCoder0:
Both assert-related review comments are addressed in 66cebdba9 — replaced with explicit `raise ValueError` checks so they survive `python -O`.

AceCoder0:
Friendly ping — the verified vLLM commit has since moved to `cdc4824a2`, and the `is_prefilling` duplicate-kwarg TypeError this PR fixes (commit 2) is still reproducible on current main with MRV2 + mamba-hybrid models (it also blocks smoke-testing of other PRs, e.g. #14733 on Qwen3.5). If it helps, that commit could also land standalone.

nwpu-zxr (approved):

linfeng-yuan (approved):

PR #14555, opened 2026-08-19
[Refactor][310P] MRv2 Qwen3/Qwen3.5 TP+ACLGraph+w8a8/w8a8sc adaptation

Thiagor2002:

310P cannot run the shared Ascend MRv2 path as-is: it has no Triton runtime, Attention KV cache must be allocated as `ACL_FORMAT_FRACTAL_NZ`, and ACL Graph capture must record direct NPU operators rather than mainline graph-task handles.

This PR adds a 310P MRv2 stack under `vllm_ascend/_310p/worker/v2/` plus narrow shared MRv2 extension points so the following acceptance surface works on 310P:

> Qwen3 / Qwen3.5 dense, VL, MoE, and hybrid models with **TP + ACL Graph (`FULL_DECODE_ONLY`) + W8A8 / W8A8SC / W8A8-Dynamic**, without prefix cache or MTP. Qwen3 dense accuracy must not regress relative to MRv1 on the same checkpoint and TP size.

Prefix cache, full sampling postprocessing, and MTP stay rejected at startup and are deferred to a follow-up 310P MRv2 release. Disabling `VLLM_USE_V2_MODEL_RUNNER` must leave V1 behavior unchanged.

**Support matrix** 

Status: **verified** = real-weight TP + graph serve evidence was supplied from the 310P server; **known failing** = real-weight serve currently fails and must not be advertised as supported; **code-compatible** = the path exists but still needs 310P acceptance.

| Model | TP | ACL Graph `FULL_DECODE_ONLY` | Quantization boundary |
| --- | --- | --- | --- |
| Qwen3-8B | verified TP + graph | verified | W8A8 verified; W8A8SC verified TP1/TP2 + graph (`sharded_state`) |
| Qwen3-VL-2B-Instruct | verified TP1/TP2 | verified (encoder eager, decode graph) | quantized VL checkpoint pending |
| Qwen3-VL-8B-Instruct | verified TP1/TP2 | verified (encoder eager, decode graph) | W8A8SC TP2 + graph verified (`sharded_state`) |
| Qwen3-30B-A3B | verified TP2 | verified | W8A8-Dynamic verified TP2 + graph |
| Qwen3.5-2B | verified TP + graph | verified | Qwen3.5-2B-W8A8 verified TP2 + graph (310P dynamic linear uses ND fp16 dequant) |
| Qwen3.5-4B | FP16 code-compatible; post-§31 hardware re-validation required | FP16 hardware re-validation required | Qwen3.5-4B-W8A8 verified with TP + graph |
| Qwen3.5-9B | — | — | Qwen3.5-9B-W8A8 verified TP2 + graph |

Yes. Setting `VLLM_USE_V2_MODEL_RUNNER=1` on 310P selects `NPUModelRunner310V2` instead of `NPUModelRunner310` for supported Qwen3 and Qwen3.5 workloads. The default remains MRv1.

- **Unit tests added/updated:** `tests/ut/_310p/test_model_runner_v2_310p.py`,
  `tests/ut/_310p/quantization/test_w8a8_dynamic_310.py`,
  `tests/ut/_310p/quantization/test_w8a8_dynamic_tp_310.py`,
  `tests/ut/_310p/quantization/test_w8a8sc_310.py`,
  `tests/ut/_310p/quantization/test_modelslim_config_310.py`,
  `tests/ut/_310p/attention/test_attention_v1_310.py`,
  `tests/ut/_310p/ops/test_qwen3vl_310.py`,
  `tests/ut/worker/test_attn_utils_v2.py`, and
  `tests/ut/quantization/test_modelslim_config.py`.
- **E2E tests added:** `tests/e2e/pull_request/one_card/_310p/test_model_runner_v2_310p.py`,
  `tests/e2e/pull_request/four_card/_310p/test_model_runner_v2_310p.py`, and
  `tests/e2e/pull_request/four_card/_310p/test_model_runner_v2_moe_310p.py`.
- **Local review (this workspace):** static code review and diff reconciliation
  against current main; no 310P hardware in this environment.
- **310P server validation (required before merge):** run the unit and E2E commands

- vLLM version: v0.27.1


tests/e2e/pull_request/four_card/_310p/test_model_runner_v2_moe_310p.py
```diff
@@ -0,0 +1,139 @@
...
+    ) as vllm_model:
+        outputs = vllm_model.generate_greedy(["Hello, my name is"], max_tokens=5)
+        follow_up_outputs = vllm_model.generate_greedy(["Count to two."], max_tokens=2)
+
+    assert outputs[0][0]
+    assert follow_up_outputs[0][0]
+
+
+@patch.dict(os.environ, {"VLLM_USE_V2_MODEL_RUNNER": "1"})
+def test_model_runner_v2_qwen3_5_moe_tp4_eager():
```
ShySummer: The TP2 tests above cover Qwen3-30B-A3B, while the newly added Qwen3.5-35B-A3B E2E cases only cover TP4.
If Qwen3.5-35B-A3B is documented as verified with TP2, should we also add a TP2 regression case for this model, or update the documented validation configuration to TP4?

tests/ut/worker/test_attn_utils_v2.py
```diff
@@ -299,3 +299,23 @@ def test_mrv2_builds_shared_dsa_metadata_for_each_execution_mode(
...
+
+
+def test_asymmetric_head_size_keeps_existing_v_shape_behavior():
+    kv_cache_shape = (2, 10, 16, 8, 128)
+    kv_cache_spec = SimpleNamespace(head_size=128, head_size_v=64)
+
+    k_shape, v_shape = attn_utils._get_non_mla_kv_cache_shapes(kv_cache_shape, kv_cache_spec)
+
+    assert k_shape == (10, 16, 8, 128)
+    assert v_shape == (10, 16, 8, 64)
```
Tflowers-0129: wrong place, use only 310p

tests/e2e/pull_request/four_card/_310p/test_model_runner_v2_310p.py
```diff
@@ -0,0 +1,210 @@
...
+from tests.e2e.pull_request.utils_310p import (
+    FULL_DECODE_ONLY_GRAPH,
+    hybrid_runner_kwargs,
+    run_vl_model_test,
+)
+
+
+@patch.dict(os.environ, {"VLLM_USE_V2_MODEL_RUNNER": "1"})
+@pytest.mark.parametrize("model", ["Qwen/Qwen3-8B", "Qwen/Qwen3.5-2B", "Qwen/Qwen3.5-4B"])
+def test_model_runner_v2_tp2_chunked_prefill_aclgraph(model: str) -> None:
```
YangShuai52: MRV2的测试用例对齐MRV1现有的，不要新增这么多测试用例，单卡和4卡的都是

vllm_ascend/_310p/worker/v2/kernel_registry.py
```diff
@@ -0,0 +1,58 @@
...
+shared ``kernel[grid](...)`` call site.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+
+from vllm.logger import logger
+
+try:
```
YangShuai52: 这里源神的PR还没合入，没合入的就当不存在

vllm_ascend/_310p/worker/v2/kernel_registry.py
```diff
@@ -0,0 +1,58 @@
...
+* ``Ascend310PRopeState`` builds multimodal positions on CPU;
+* ``Ascend310PGreedySampler`` avoids the upstream sampling kernels.
+
+This module keeps the dispatcher path alive for the day that PR lands and the
+relevant upstream kernel is marked pluggable: add its fully qualified name and
+310P implementation to ``KERNEL_IMPLS``, then the override takes effect at the
+shared ``kernel[grid](...)`` call site.
+"""
+
+from __future__ import annotations
```
YangShuai52: 全文都检查一下，triton_dispatcher 没合入就不要体现

vllm_ascend/_310p/worker/v2/model_runner.py
```diff
@@ -0,0 +1,572 @@
...
+from vllm_ascend._310p.attention.attention_v1 import AscendAttentionBackend310
+from vllm_ascend._310p.worker.v2.aclgraph import ModelAclGraphManager310
+from vllm_ascend._310p.worker.v2.block_table import Ascend310PBlockTables
+from vllm_ascend._310p.worker.v2.feature_support import (
+    FIRST_RELEASE_FEATURE_SUPPORT,
+    MRv2FeatureSupport,
+)
+from vllm_ascend._310p.worker.v2.kernel_registry import register_310p_kernels
+from vllm_ascend._310p.worker.v2.kv_block_zeroer import AscendKVBlockZeroer310V2
+from vllm_ascend._310p.worker.v2.sampler import Ascend310PGreedySampler
```
YangShuai52: 虽然目前还没适配topk\topp等后处理但是也不能直接起名校Greedy后续是要适配的

mkdocs.yml
```diff
@@ -309,6 +309,9 @@ nav:
         - developer_guide/Design_Documents/dynamic_chunked_pipeline_parallel.md
         - developer_guide/Design_Documents/quantization.md
         - developer_guide/Design_Documents/npugraph_ex.md
+        - developer_guide/Design_Documents/ModelRunner_v2_310P_adaptation.md
+        - developer_guide/Design_Documents/ModelRunner_v2_310P_code_changes.md
+        - developer_guide/Design_Documents/ModelRunner_v2_310P_pr_notes.md
```
csoulnd: Delete along with the document

vllm_ascend/_310p/worker/v2/feature_support.py
```diff
@@ -0,0 +1,39 @@
...
+            )
+        if vllm_config.speculative_config is not None and not self.qwen3_5_mtp:
+            raise NotImplementedError(
+                "Qwen3.5 MTP/speculative decoding is deferred to a later 310P "
+                "Model Runner V2 release; the first-release feature boundary "
+                "keeps it disabled."
+            )
+
+
+FIRST_RELEASE_FEATURE_SUPPORT = MRv2FeatureSupport()
```
csoulnd, line 24: The temporary validation logic used in development mode shall be removed.

vllm_ascend/_310p/quantization/methods/w8a8_dynamic.py
```diff
@@ -143,31 +166,15 @@ def apply(
...
+        scale = layer.weight_scale.data if hasattr(layer.weight_scale, "data") else layer.weight_scale
+        weight_fp = layer.weight.data.to(x.dtype) * scale.to(x.dtype).view(-1, 1)
+        bias_term = bias if (tp_rank is None or tp_rank == 0) else None
+        return torch.nn.functional.linear(x, weight_fp, bias_term)
 
     def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
-        # cast quantized weight tensors in NZ format for higher inference speed
-        layer.weight.data = maybe_trans_nz(layer.weight.data).transpose(0, 1)
+        layer.weight.data = layer.weight.data.contiguous()
+        layer._310p_w8a8_dynamic_fp16_fallback = _needs_fp16_quant_matmul_fallback(layer, layer.weight.data)
```
csoulnd: The callback will definitely be triggered under current conditions, which seems ill‑ogical.

mkdocs.yml
Thiagor2002: it has completed.

vllm_ascend/worker/v2/model_states/mamba_hybrid.py
```diff
@@ -104,3 +111,25 @@ def prepare_attn(
...
+        del num_computed_tokens
+
+        if isinstance(num_sampled, int):
+            self.num_accepted_tokens_gpu.index_fill_(0, idx_mapping, max(num_sampled, 1))
+            return
+
+        valid = idx_mapping >= 0
+        valid_indices = idx_mapping.masked_select(valid)
+        accepted = torch.clamp(num_sampled.masked_select(valid), min=1).to(self.num_accepted_tokens_gpu.dtype)
+        self.num_accepted_tokens_gpu.index_copy_(0, valid_indices, accepted)
```
csoulnd: I think this part should be overridden via inheritance under the 310p folder, instead of adding branches in the mainline.

vllm_ascend/_310p/fused_moe/moe_mlp.py
```diff
@@ -17,8 +17,59 @@
...
+    the cast happens inside this opaque custom op.
+    """
+    weight_nz = torch_npu.npu_format_cast(quantized_weight.contiguous(), ACL_FORMAT_FRACTAL_NZ)
+    return torch_npu.npu_quant_grouped_matmul_dequant(
+        x=x,
+        quantized_weight=weight_nz,
+        weight_scale=weight_scale,
+        group_list=group_list,
+        quant_mode="pertoken",
+    )
```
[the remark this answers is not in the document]
Thiagor2002: it has completed.

Alex-stack-hub (commented):
Thanks for this big 310P MRv2 PR! The overall structure is really clean - keeping the 310P-specific Triton-free paths, NZ layout handling, and ACL Graph capture under the `_310p/` subtree while leaving MRv1 as the default is nice, and the test coverage looks thorough. I left a few gentle comments after a close read - none of them are blocking, just things worth considering. Thanks!

vllm_ascend/_310p/fused_moe/moe_mlp.py
```diff
@@ -17,8 +17,67 @@
...
+
+    ModelSlim stores ND ``[E, N, K]``. 310P WeightNZ accepts FRACTAL_NZ of
+    that layout. torch.compile/GE strips format-29 from 3D Parameters, so
+    the cast happens inside this opaque custom op and is cached by data_ptr.
+    """
+    key = quantized_weight.data_ptr()
+    weight_nz = _WEIGHT_NZ_CACHE.get(key)
+    if weight_nz is None:
+        weight_nz = torch_npu.npu_format_cast(quantized_weight.contiguous(), ACL_FORMAT_FRACTAL_NZ)
+        _WEIGHT_NZ_CACHE[key] = weight_nz
```
Alex-stack-hub, line 46: A small thought: using `data_ptr()` as the global cache key means a stale entry could be picked up if a weight tensor is freed and a new one happens to reuse the same address. It's probably fine for inference where weight lifetimes span the whole process, but would a `WeakKeyDictionary` (or `id()` with cleanup) be a safer fallback? Not blocking - just something worth keeping in mind.

vllm_ascend/_310p/worker/v2/model_runner.py
```diff
@@ -0,0 +1,584 @@
...
+        """Return query lengths for real requests, excluding graph padding."""
+        # A FULL graph may pad idx_mapping to its capture bucket while
+        # query_start_loc still contains boundaries for real requests only.
+        # Padding entries are trailing -1 sentinels and must not participate in
+        # the num_computed_tokens update.
+        num_query_lens = min(idx_mapping.shape[0], query_start_loc.shape[0] - 1)
+        query_lens = query_start_loc[1 : num_query_lens + 1] - query_start_loc[:num_query_lens]
+        return query_lens.masked_select(idx_mapping[:num_query_lens] >= 0)
+
+    def postprocess_num_computed_tokens(self, input_batch: AscendInputBatch) -> None:
```
Alex-stack-hub, line 584: Here `index_add_` uses `input_batch.idx_mapping` directly, and it doesn't seem to filter out the `-1` graph padding sentinels the way `postprocess_sampled` does. If this path can also be reached after a FULL graph replay, could the padding entries corrupt `num_computed_tokens`? If the call ordering guarantees no negative values here, a short comment stating that assumption would be reassuring.

vllm_ascend/_310p/worker/v2/model_runner.py
```diff
@@ -0,0 +1,584 @@
...
+        """Drain NPU stream on layout-change steps to prevent 310P ACLGraph precision regressions.
+
+        Mirrors vLLM-Ascend PR #9727: `condense()` rewrites `block_table.np`
+        (move_row). For 310P MRv2, the previous step's ACLGraph replay may
+        still be in-flight when the condensed CPU layout is uploaded and
+        consumed by paged/splitfuse attention metadata.
+        """
+        deferred = super()._update_states(scheduler_output)
+        if scheduler_output.finished_req_ids:
+            torch.npu.current_stream().synchronize()
```
Alex-stack-hub, line 95: I see this mirrors the precedent in PR #9727. Just a small performance question: `finished_req_ids` being non-empty is quite common (almost every decode window has requests finishing), so would a full `synchronize()` on each of those steps cause a noticeable pipeline stall? Could the sync be limited to steps where a `condense()` / layout change actually happened? Feel free to ignore this if the measured impact is small~

vllm_ascend/_310p/worker/v2/aclgraph.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(
+        self,
+        vllm_config: VllmConfig,
+        device: torch.device,
+        cudagraph_mode: CUDAGraphMode,
+        decode_query_len: int,
+        model_runner,
+        lora_capture_cases: list[int] | None = None,
+    ) -> None:
+        ModelCudaGraphManager.__init__(
```
Alex-stack-hub, line 32: Just curious: calling `ModelCudaGraphManager.__init__` directly instead of `super().__init__` - is this intentional, to skip the graph-task handle setup in `ModelAclGraphManager.__init__`? If so, it might be worth a one-line comment explaining why, so future maintainers (including future us) don't accidentally change it back to `super()`.

vllm_ascend/_310p/quantization/methods/w8a8_dynamic.py
```diff
@@ -143,31 +155,21 @@ def apply(
...
-        layer.weight.data = maybe_trans_nz(layer.weight.data).transpose(0, 1)
+        layer.weight.data = layer.weight.data.contiguous()
         layer.weight_scale.data = layer.weight_scale.data.flatten()
         layer.weight_offset.data = layer.weight_offset.data.flatten()
+        # Dequant once at load: scales are static and redoing int8->fp16 * scale
+        # on every forward is pure hot-path overhead.
+        params_dtype = getattr(layer, "params_dtype", torch.float16)
+        dtype = params_dtype if isinstance(params_dtype, torch.dtype) else torch.float16
+        scale = layer.weight_scale.data.to(dtype).view(-1, 1)
+        layer.weight_fp = torch.nn.Parameter(layer.weight.data.to(dtype) * scale, requires_grad=False)
```
Alex-stack-hub, line 175: One heads-up: dequantizing at load time keeps both the int8 weight and the fp16 `weight_fp` resident, and together with the MoE NZ cache the peak memory is roughly 3x the quantized size, which could get tight for larger deployments. Would it make sense to offer a switch (e.g. keep int8 + dequantize at runtime as a low-memory fallback)? Also, if `weight` is no longer referenced after dequant, `del layer.weight` could save one copy. Totally up to your memory budget, of course.

vllm_ascend/_310p/worker/v2/states.py
```diff
@@ -0,0 +1,122 @@
...
+            self._dirty_indices.clear()
+            return
+        indices = sorted(self._dirty_indices)
+        indices_device = torch.tensor(indices, dtype=torch.int64, device=self.device)
+        values = self.cpu[indices].to(self.device, non_blocking=True)
+        self.gpu.index_copy_(0, indices_device, values)
+        self._dirty_indices.clear()
+
+
+class Ascend310PRequestState(AscendRequestState):
```
Alex-stack-hub, line 70: `Ascend310PRequestState.__init__` manually duplicates the full field initialization of the parent `AscendRequestState`. If either class gains or drops a field later, the two can silently drift apart. Would it work to reuse `super().__init__(...)` and then only replace the buffers that need to be Triton-free? Of course, if the parent constructor has Triton dependencies that force this workaround, a brief comment explaining that would be great - thanks!

vllm_ascend/_310p/worker/v2/model_runner.py
Thiagor2002: 采用全局 synchronize() 同步可能会阻断npu流，vLLM-ascend主线 MRv2 重构对该问题采用了更复杂的 流分离机制，从而减小阻塞，允许并发。之后会进一步改进。

Thiagor2002: Using global `synchronize()` may block the NPU stream. The vLLM-ascend main branch MRv2 refactoring addresses this issue by adopting a more sophisticated stream separation mechanism to reduce blocking and allow concurrency. This will be further improved in follow-ups.

PR #14619, opened 2026-08-20
[Cherry-pick][releases/v0.26.0rc][BugFix][P/D] Clean up rejected requests and retry non-400 failures (from #14142)

moonseeker1:

Backports #14142 to `releases/v0.26.0rc`.

This fixes rejected-request cleanup in the disaggregated prefill path and applies the requested backend retry policy:

- release prefill KV and roll back scheduler accounting when decode rejects a request;
- retry transport failures and every HTTP failure except status 400, using exponential backoff;
- fail HTTP 400 immediately without retry;
- restore `--max-retries` and `--retry-delay` for prefill and decode requests;
- treat non-positive `--max-retries` values as one backend attempt with no retry;
- never replay a streaming request after response bytes have already been emitted, avoiding duplicate output.

Yes. Rejected decode requests no longer leave stale prefill KV or scheduler state. Backend requests retry failures except HTTP 400, while streaming requests are retried only before output begins.

- Added unit coverage for Mooncake rejected-request cleanup.
- `git range-diff` confirmed the five backported commits match the five substantive commits from #14142.
- `git diff --check` passed.
- Python syntax compilation passed for all three modified files.
- Full unit tests were not run locally because the required vLLM/NPU development dependencies are unavailable in the Windows workspace; CI verification is required.

- vLLM version: v0.26.0


kunpengW-code:
The current modification has been fixed through PR https://github.com/vllm-project/vllm-ascend/pull/14687, and this PR can be closed.

PR #14633, opened 2026-08-20
[Refactor][Attention] Use CANN LightningIndexer for non-quantized SFA

yanhao5023:

This PR removes the bundled non-quantized LightningIndexer custom operator and uses the CANN-provided `torch_npu.npu_lightning_indexer` implementation for the non-C8 SFA path.

It also:

- removes the deleted operator from custom-op build and registration lists;
- adds a unit regression test that verifies the non-quantized path calls the torch-npu operator.

This avoids maintaining a duplicate LightningIndexer implementation now that the operator is provided by CANN.

Yes. Non-quantized SFA models now require a CANN/torch-npu version that provides `torch_npu.npu_lightning_indexer`. There is no API change.

Validation results:

- Unit regression test: **passed** (`1 passed, 36 deselected`).
- Build and installation on Atlas A3: **passed**.
- Four-layer BF16 GLM-5 request with TP=2 and EP enabled: **passed**. The server returned HTTP 200 and generated tokens.
- Operator dispatch: **confirmed**. Both TP workers invoked `torch_npu.npu_lightning_indexer` on the non-C8 path.
- Operator accuracy: **exact match** between the removed custom operator and the CANN operator across FP16/BF16, decode/prefill, three random seeds, and `sparse_count` 64/256/2048. All **44,544 indices in 24 cases matched**, with **0 mismatches**.
- Operator performance: **no material regression observed**. Median synchronized latency improved in all measured BF16 PA_BSND cases:

| Case | Custom operator | CANN operator | Result |
| --- | ---: | ---: | ---: |
| Decode, KV=4K | 0.0815 ms | 0.0761 ms | 6.7% faster |
| Decode, KV=16K | 0.0901 ms | 0.0850 ms | 5.7% faster |
| Batched prefill, KV=4K/8K | 0.1146 ms | 0.1106 ms | 3.4% faster |

Known gaps:

- DeepSeek-V3.2 graph mode is not yet validated. The former dual-operator implementation existed because this model had graph-compilation issues with `torch_npu.npu_lightning_indexer`.
- The model-level result covers GLM-5 eager mode only; it does not cover DeepSeek-V3.2 `torch.compile` / ACLGraph startup, capture, replay, or request accuracy.
- DeepSeek-V3.1 does not use LightningIndexer and therefore is not a validation target for this change.
- The full 78-layer non-quantized GLM-5 model was not tested because it requires two A3 nodes.
- Full-model dataset accuracy evaluation has not been completed.


ZT-AIA:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

maoxx241 (approved):

PR #14641, opened 2026-08-20
[BugFix][KV Offload] Revert aligned Mamba state block normalization

qijiajin:

This PR reverts #14254.

In a real Kimi-K3 P/D disaggregated deployment using DP=4, TP=16, Expert Parallelism, DSpark speculative decoding, aligned Mamba cache, prefix caching, and Mooncake KV transfer, the decode side may receive multiple Mamba state blocks.

The validation introduced by #14254 requires exactly one normalized remote state block and fails with:

```text
Mooncake Mamba transfer requires exactly one normalized remote state block;
remote_block_count=8, local_block_count=8
```

This failure causes the affected request to enter the invalid-block handling path. The EngineCore subsequently exits, and the remaining DP ranks fail after their Gloo peers are closed.

The problem was reproduced with 32 requests containing 8K inputs at concurrency 8 across four prefill nodes and four decode nodes.

This revert restores the behavior before #14254 while a complete fix for aligned Mamba metadata and hybrid KV-cache invalid-block handling is investigated.

This reverts commit 33761777c8b16741456f43a5420f0fb2af941371.

Yes.

This PR restores the previous Mooncake Mamba KV-transfer behavior. It removes the strict single-block validation introduced by #14254.

No public API or command-line interface is changed.

The revert was applied to four prefill nodes and four decode nodes.

The following checks were performed:

- Verified that only the two files changed by #14254 were reverted.
- Ran `git diff --check`.
- Verified identical reverted file hashes across all eight nodes.
- Verified that the `exactly one normalized remote state block` validation was removed.
- Verified the Python syntax of `mooncake_connector.py`.
- Ran the existing Mooncake connector unit tests:

```bash
pytest -q tests/ut/kv_offload/test_mooncake_connector.py \
  --disable-warnings \
  --maxfail=1
```

Result:

```text
111 passed, 14 warnings
```

No new tests are added because this PR is a direct revert and removes the tests introduced by #14254.
- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/kv_offload/test_mooncake_connector.py
+++ b/tests/ut/kv_offload/test_mooncake_connector.py
@@ -933,15 +933,6 @@ class TestCoreFunctionality(unittest.TestCase):
         self.thread.remote_te_port = {"remote_engine": {6666: 7777}}
         self.thread.remote_block_stride_per_addr["remote_engine"][6666] = [[1024]]
 
-    def _configure_mock_mamba_transfer(self):
-        self.thread.kv_group2layeridx = {0: ({"kv_cache_spec_type": "MambaSpec"}, [0])}
-        self.thread.kv_caches_base_addr["local_engine"][5555] = [[0x1000, 0x2000]]
-        self.thread.kv_caches_base_addr["remote_engine"] = {6666: [[0x3000, 0x4000]]}
-        self.thread.block_len_per_addr = [[100, 200]]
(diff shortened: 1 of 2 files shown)


qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun:**
- E2E

PR #14649, opened 2026-08-20
[BugFix][Mamba] Fix staged-write kernel and duplicate is_prefilling argument

Lethobenthos20:
## Summary

This PR fixes two issues in the Ascend V2 Runner inference path:

1. Fixes incorrect index/address offset calculation in the block-table staged-write Triton kernel, which caused:

   ```text
   PtroffsetInfo::AxisInfo::operator[]: Assertion `idx < size()` failed
2. Fixes the duplicate is_prefilling argument passed to AscendCommonAttentionMetadata, which caused:
    ```text
     got multiple values for keyword argument 'is_prefilling'
     ```
   After these changes, V2 Runner inference works correctly with curl and aisbench.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -532,12 +532,6 @@ def adapt_patch(is_global_patch: bool = False):
     else:
         from vllm_ascend.patch import worker  # noqa: F401
 
-    from vllm.v1.sample.ops import topk_topp_sampler
-    from vllm_ascend.sample.sampler import _apply_top_k_top_p_torch_npu
-
-    topk_topp_sampler.apply_top_k_top_p = _apply_top_k_top_p_torch_npu
-
-
 def setup_ascend_local_comm_res(local_rank: int, kv_transfer_config: Any | None) -> None:
     """Load the local A5 endpoint config into ASCEND_LOCAL_COMM_RES."""
     if kv_transfer_config is None:
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -532,6 +532,7 @@ def adapt_patch(is_global_patch: bool = False):
     else:
         from vllm_ascend.patch import worker  # noqa: F401
 
+
 def setup_ascend_local_comm_res(local_rank: int, kv_transfer_config: Any | None) -> None:
     """Load the local A5 endpoint config into ASCEND_LOCAL_COMM_RES."""
     if kv_transfer_config is None:
--- a/vllm_ascend/utils.py
+++ b/vllm_ascend/utils.py
@@ -532,6 +532,11 @@ def adapt_patch(is_global_patch: bool = False):
     else:
         from vllm_ascend.patch import worker  # noqa: F401
 
+    from vllm.v1.sample.ops import topk_topp_sampler
+    from vllm_ascend.sample.sampler import _apply_top_k_top_p_torch_npu
+
+    topk_topp_sampler.apply_top_k_top_p = _apply_top_k_top_p_torch_npu
+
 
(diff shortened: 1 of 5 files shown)


vllm_ascend/utils.py
```diff
@@ -535,6 +535,11 @@ def adapt_patch(is_global_patch: bool = False):
     else:
         from vllm_ascend.patch import worker  # noqa: F401
 
+    from vllm.v1.sample.ops import topk_topp_sampler
+    from vllm_ascend.sample.sampler import _apply_top_k_top_p_torch_npu
+
+    topk_topp_sampler.apply_top_k_top_p = _apply_top_k_top_p_torch_npu
```
zouzy5137: apply_top_k_top_p has been fixed on ascend. No need to do patch here.

Lethobenthos20: Thanks for pointing this out. I removed the redundant patch and pushed a new commit.

vllm_ascend/worker/v2/attn_utils.py
```diff
@@ -227,6 +220,10 @@ def build_attn_metadata(
             if model_specific_attn_metadata is not None
             else {}
         )
+        common_is_prefilling = common_attn_metadata_extra_kwargs.pop(
+            "is_prefilling",
+            is_prefilling,
+        )
         common_attn_metadata = AscendCommonAttentionMetadata(
```
weijinqian0, line 234: Please confirm the reason why the value is repeatedly transferred.

vllm_ascend/worker/v2/attn_utils.py
```diff
@@ -187,7 +181,6 @@ def build_attn_metadata(
     num_actual_tokens: int | None = None,
     num_input_tokens: int | None = None,
     is_prefilling: torch.Tensor | None = None,
-    pcp_context: "AscendPCPAttentionContext | None" = None,
     model_specific_attn_metadata: ModelSpecificAttnMetadata | None = None,
```
slippersss, line 191: Maybe we should check out why there is `is_prefilling` in both here and model_specific_attn_metadata.

vllm_ascend/worker/v2/attn_utils.py
Lethobenthos20: Thanks for pointing this out. This follows the same pattern already used upstream in vllm/v1/worker/gpu/attn_utils.py around line 545:
```
group_is_prefilling = common_attn_metadata_extra_kwargs.pop(
    "is_prefilling", is_prefilling
)
```
It allows the model-specific, padding-aware is_prefilling value to take precedence over the default, so I believe this change is still necessary.

vllm_ascend/worker/v2/attn_utils.py
Lethobenthos20: Thanks for pointing this out. This follows the same pattern already used upstream in vllm/v1/worker/gpu/attn_utils.py around line 545:
```
group_is_prefilling = common_attn_metadata_extra_kwargs.pop(
    "is_prefilling", is_prefilling
)
```
It allows the model-specific, padding-aware is_prefilling value to take precedence over the default, so I believe this change is still necessary.

slippersss: LGTM, thanks for your effort. CC @weijinqian0.

slippersss (approved):

weijinqian0 (dismissed):

vllm_ascend/worker/v2/attn_utils.py
slippersss: BTW, we'd better change `common_is_prefilling` to `group_is_prefilling`, to align with vllm.

weijinqian0 (approved):

PR #14680, opened 2026-08-20
[Bugfix] fix DeeSeek-v4  and spec decode(other than mtp)  pre_hc_hidden_states none error in MRV2

HF-001:
In MRV2, when the DeepSeek-v4 model enables speculative decode(like dpsark、dflash、eagle3 ，etc) other than MTP, the following error may occur：

```
tests/e2e/pull_request/four_card/model_runner_v2/test_deepseek_v4.py::test_dspark_spec_decoding

File "/usr/local/python3.12.13/lib/python3.12/site-packages/vllm/v1/executor/multiproc_executor.py", line 1021, in worker_busy_loop
      output = func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/python3.12.13/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
      return func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
    File "/__w/vllm-ascend/vllm-ascend/vllm_ascend/worker/worker.py", line 523, in determine_available_memory
      self.model_runner.profile_run()
    File "/usr/local/python3.12.13/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
      return func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
    File "/__w/vllm-ascend/vllm-ascend/vllm_ascend/worker/v2/model_runner.py", line 279, in profile_run
      self._dummy_run(mc2_tokens_capacity, skip_attn=True, skip_eplb=True, is_profile=True)
    File "/usr/local/python3.12.13/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 124, in decorate_context
      return func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/python3.12.13/lib/python3.12/site-packages/vllm/v1/worker/gpu/eplb_utils.py", line 27, in wrapper
      result = fn(self, *args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/python3.12.13/lib/python3.12/site-packages/vllm/v1/worker/gpu/model_runner.py", line 685, in _dummy_run
      spec_hidden_states = pre_hc_hidden_states[: hidden_states.shape[0]]  # type: ignore[union-attr]
                           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  TypeError: 'NoneType' object is not subscriptable
```

- vLLM version: v0.27.1


slippersss:
`_mtp_hidden_buffer` will be allocated when `spec_config.use_eagle() or spec_config.uses_draft_model()` in vllm, while in vllm-ascend, condition is restricted to `spec_config.method == "mtp"`. Since model runner v2 will reuse vllm, so we have to consider that:

1. if `_mtp_hidden_buffer` is indeed functioning only in mtp, we should modify vllm
2. otherwise, we should relax condition to be the same as vllm

CC @Tflowers-0129 @845473182.

HF-001:
@slippersss Thank you for your reply. I will evaluate and adjust the upstream VLLM as soon as possible to see if it is more reasonable

PR #14701, opened 2026-08-21
[BugFix][KV Cache] Align A5 C8 MLA cache blocks with Mamba

Yaphets24:

Fixes the hybrid MLA/Mamba KV-cache block alignment used by A5 C8 Kimi K3. The Mamba configuration originally derives the attention block size before automatic quantization has loaded the ModelSlim C8 settings. It therefore sees BF16 latent-K bytes instead of the actual FP8 storage size, leaving the MLA manager-block stride at half the Mamba state stride and allowing cross-ID physical aliasing.

This PR uses the real FP8 latent-K byte size while retaining the positional cache in the configured KV-cache dtype. It also reruns the hybrid MLA/Mamba configuration immediately after `maybe_auto_detect_quantization(vllm_config)` and before KV-cache planning. At that point the C8 configuration is available, so the final MLA and Mamba manager-block strides agree before cache allocation. It complements #13286 and is validated by the diagnostic introduced in #14693.

Yes. A5 C8 Kimi K3 hybrid cache initialization now uses a safe MLA/Mamba block layout instead of one that can alias distinct logical blocks.

- `python -m py_compile vllm_ascend/platform.py tests/ut/test_platform.py vllm_ascend/patch/platform/patch_mamba_config.py tests/ut/patch/platform/test_patch_mamba_config.py`
- `git diff --check`
- Added unit regression coverage for BF16/C8 block alignment and for refreshing only hybrid model configurations after quantization detection.

The local Windows environment does not include pytest or ruff, so the targeted test requires CI/NPU validation.

- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/patch/platform/test_patch_mamba_config.py
+++ b/tests/ut/patch/platform/test_patch_mamba_config.py
@@ -1,0 +1,88 @@
+# SPDX-License-Identifier: Apache-2.0
+
+from types import SimpleNamespace
+
+import pytest
+import torch
+
+from vllm_ascend.patch.platform import patch_mamba_config
+from vllm_ascend.utils import AscendDeviceType
+
+
+class _MambaModel:
+    @staticmethod
+    def get_mamba_state_shape_from_config(_vllm_config):
+        return [(786432,), (1,)]
+
+    @staticmethod
+    def get_mamba_state_dtype_from_config(_vllm_config):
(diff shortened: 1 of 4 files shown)


Yaphets24:
Superseded by #13286. The two signed cache-alignment commits have been cherry-picked into #13286, and its description now includes this fix and its regression coverage.

PR #14716, opened 2026-08-21
[Ops][Feature] Add prefill abort side-channel support for disaggregated prefill

UpDown9:

This PR introduces a prefill abort side-channel mechanism for disaggregated prefill. It allows the load balance proxy to notify prefillers to abort and force-release resources when a decoder stream is cancelled or encounters an error. This is achieved by introducing a new `ABORT_REQUEST_MSG` message type in the `Mooncake_hybrid_connector` and scheduling abort requests via ZeroMQ (`zmq`).

Feedback on the changes:
- In `_send_prefill_abort_sync`, creating and terminating a new `zmq.Context` on every abort request is inefficient and can cause resource leaks. It is recommended to use `zmq.Context.instance()` instead.
- In `assign_instances`, `kv_transfer_params` should be safely defaulted to prevent potential `AttributeError` or `TypeError` if it is `None`.

No, this is an internal optimization and error-handling mechanism for disaggregated prefill.

### DIFF
--- a/examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
+++ b/examples/disaggregated_prefill_v1/load_balance_proxy_server_example.py
@@ -152,6 +152,12 @@ from fastapi.responses import JSONResponse, Response, StreamingResponse
 
 logger = logging.getLogger(__name__)
 
+
+class UpstreamHTTPError(Exception):
+    def __init__(self, response: httpx.Response):
+        self.response = response
+        super().__init__(f"upstream returned HTTP {response.status_code}")
+
 try:
     import uvloop  # type: ignore[import-not-found]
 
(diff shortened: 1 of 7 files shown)


vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_hybrid_connector.py
```diff
@@ -161,6 +162,15 @@ def update_done_task_count(self, request_id: str):
                     request_id,
                 )
 
+    def abort_request(self, request_id: str) -> None:
+        """Force-release a prefiller request aborted by the decoder."""
```
nu11ptr7: One thing about `abort_request`: it ACKs and marks the request finished immediately, without checking whether the KV transfer is still in flight. A later DONE_RECVING then hits the "finish req not in reqs to process" error path, and the prefiller can free pages the recv thread is still writing. Maybe track an in-flight/cancelled flag per request and only ACK once the transfer observes the abort.

PR #14717, opened 2026-08-21
[BugFix] Guard dp device ids patch for missing vLLM API

ink0410:
## Summary

Guard access to `get_physical_gpu_ids_for_local_dp_rank` in
`patch_dp_device_ids.py` for compatibility with vLLM versions where
the API is unavailable.
get_physical_gpu_ids_for_local_dp_rank is an internal vLLM helper and is not guaranteed to exist across vLLM revisions. The current patch accesses it unconditionally, which can make vLLM-Ascend fail during module import when the helper is absent. This change guards the whole monkey-patch installation with hasattr, preserving the existing behavior when the helper exists and safely skipping the patch otherwise.
## Testing

Verified on `releases/v0.23.0`.

- vLLM version: v0.23.0

### DIFF
--- a/vllm_ascend/patch/platform/patch_dp_device_ids.py
+++ b/vllm_ascend/patch/platform/patch_dp_device_ids.py
@@ -37,8 +37,6 @@ if not vllm_version_is("0.23.0"):
     from vllm.platforms import current_platform
     from vllm.v1.engine import utils as _engine_utils
 
-    _original_get_physical_gpu_ids = _engine_utils.get_physical_gpu_ids_for_local_dp_rank
-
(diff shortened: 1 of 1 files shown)


nu11ptr7 (commented):
test

weijinqian0 (approved):

vllm_ascend/patch/platform/patch_dp_device_ids.py
slippersss, line 34: We should check out this version check.

slippersss: Have you reinstalled vllm? Normally, vllm will be version 0.23.0 by default.

PR #14730, opened 2026-08-21
 [v0.26.0rc][Ops][Feature] Use mega_moe_max_tokens for CANN MegaMoe symmetric buffer allocation

coder-fny:

This PR updates the CANN MegaMoe symmetric buffer allocation to use the
configurable `mega_moe_max_tokens` instead of calculating an absolute safe
maximum. The absolute safe maximum is typically too large and causes
excessive device memory usage.

Using a configurable reference value helps optimize memory consumption, while
a warning is logged to alert users of potential precision degradation if the
actual token count exceeds this limit.

Yes. Users can configure `mega_moe_max_tokens` through `additional_config`.
The default value is set to `65536`, and its documentation is updated to
clarify its usage and tuning.

A warning is also logged when the MegaMoe symmetric buffer is initialized,
including the configured capacity and the potential precision risk when the
actual token count exceeds the configured limit.

Unit tests were added or updated in
`tests/ut/ops/test_moe_comm_method.py` and
`tests/ut/test_ascend_config.py` to verify that:

- `mega_moe_max_tokens` defaults to `65536`;
- configuration values are validated correctly;
- the symmetric buffer uses the configured `mega_moe_max_tokens`;
- the relevant MegaMoe buffer capacity arguments are forwarded correctly.

- vLLM version: v0.26.0


vllm_ascend/ascend_forward_context.py
```diff
@@ -35,6 +35,28 @@ class MoECommType(Enum):
 _MC2_TOKENS_PER_RANK_LIMIT = 512
 
 
+def _is_decode_only_node(vllm_config: VllmConfig) -> bool:
+    kv_transfer_config = getattr(vllm_config, "kv_transfer_config", None)
+    if kv_transfer_config is None:
+        return False
+
+    is_decode_bench = getattr(kv_transfer_config, "kv_connector", None) == "DecodeBenchConnector"
```
ningjingbengxiaohai: 为什么会有这个判断？

vllm_ascend/ops/fused_moe/moe_comm_method.py
```diff
@@ -305,11 +306,14 @@ def _init_mega_moe_symm_buffer(
         self,
         dispatch_quant_mode: int = 0,
         dispatch_quant_out_dtype: torch.dtype | None = None,
+        is_decode_only_node: bool | None = None,
```
kunpengW-code: The place where this method is called has not been modified, and there are no new parameters added unless necessary.

kunpengW-code (approved):

PR #13946, opened 2026-08-10
[Refactor] remove FlashComm v1

jiaqi-lee:
### What this PR does / why we need it?
The upstream vLLM community has already implemented the SP (Sequence Parallel) logic. To keep the code simple and maintainable, the downstream SP feature needs to be refactored.
### Does this PR introduce _any_ user-facing change?
Delete flashcomm1.
upstream SP is activated with DP>1, TP>1 and all2allbackend set to allgather_reducescatter

### vLLM bench serve command
Bench config
```bash
vllm bench serve \
  --backend openai \
  --base-url http://127.0.0.1:8010 \
  --endpoint /v1/completions \
  --served-model-name qwen \
  --dataset-name random \
  --random-input-len 4096 \
  --random-output-len 2048 \
  --num-prompts 50 \
  --num-warmups 5 \
  --max-concurrency 32 \
  --metric-percentiles 50,90,99 \
  --seed 0
```
Serve config
```bash
vllm serve \
    --model /mnt/weight/Qwen3.5-35B-A3B \
    --served-model-name qwen \
    --host 0.0.0.0 \
    --port 8010 \
    --data-parallel-size 2 \
    --tensor-parallel-size 2 \
    --gpu-memory-utilization 0.9 \
    --max-num-seqs 32 \
    --enable-expert-parallel \
```

## Model & setup 1

Qwen3-30B-A3B
DP2, TP2, EP
Chat verification: passed

### Benchmark

```
PR
============ Serving Benchmark Result ============
Successful requests:                     50        
Failed requests:                         0         
Maximum request concurrency:             32        
Benchmark duration (s):                  101.71    
Total input tokens:                      204800    
Total generated tokens:                  102400    
Request throughput (req/s):              0.49      
Output token throughput (tok/s):         1006.81   
Peak output token throughput (tok/s):    1312.00   
Peak concurrent requests:                38.00     
Total token throughput (tok/s):          3020.42   
---------------Time to First Token----------------
Mean TTFT (ms):                          2354.92   
Median TTFT (ms):                        1549.68   
P50 TTFT (ms):                           1549.68   
P90 TTFT (ms):                           4923.33   
P99 TTFT (ms):                           5666.07   
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          24.96     
Median TPOT (ms):                        27.29     
P50 TPOT (ms):                           27.29     
P90 TPOT (ms):                           27.38     
P99 TPOT (ms):                           27.47     
[...]

PR #14733, opened 2026-08-21
[Feature][Core] Opt-in patch: enable sequence-parallel MoE with pipeline parallelism

AceCoder0:
## Description

Upstream vLLM disables SP-MoE (`use_sequence_parallel_moe`) when `pipeline_parallel_size > 1` — the PP boundary cannot carry sequence-sharded hidden states (see the TODO in vllm's `deepseek_v2.py`). This PR adds an **opt-in** patch (`VLLM_ASCEND_ENABLE_SP_WITH_PP=1`) that enables the combination for qwen3_next / qwen3_5 / deepseek_v2 on Ascend, until the upstream boundary work lands.

### How it works (`vllm_ascend/patch/platform/patch_sp_with_pp.py`)

1. **Layer construction with a PP=1 config view** (`_wrap_layer_init`): the SP flag is consumed at construction time (attention o_proj is built with `reduce_results=not flag`), so post-hoc flag overrides cannot work. The wrapped `__init__` hands the original constructor a shallow-copied config view with `pipeline_parallel_size=1\); the caller's config object is never mutated. A per-instance `_sp_with_pp_active` marker gates everything downstream.

2. **Patched decoder-layer forwards with small-batch downgrade**: the layer forwards are replaced (dispatching on `_sp_with_pp_active`; non-activated instances fall through to the original forward byte-identically). Two behaviours are added:
   - **Runtime downgrade**: steps with `num_tokens < tp_size` (decode) fall back from reduce-scatter to an explicit all-reduce — SP pads rows up to the TP size and the pad rows produce NaN through norm/experts. SP gains are prefill-dominated, so the downgrade costs nothing where it matters.
   - **PP boundary gather** (`_wrap_model_forward`, signature-preserving via `functools.wraps`): `IntermediateTensors` handed across stages are gathered back to full-token tensors. Triggered by an explicit SP flag, not a shape check — after SP padding a shard can equal `full_num_tokens` (T=1, tp=2), which would silently skip the gather and leak a pad row across the boundary.

3. **Deferred installation**: patches install at the first `ModelRegistry.resolve_model_cls` call, not at plugin registration — importing vllm model modules too early leaves the MoE router on a CUDA-only dispatch path (`topk_softmax` AttributeError on NPU).

Upstream-drift sentinels (`inspect.getsource` anchors) refuse to install when the copied forwards no longer match, so a vLLM pairing bump fails loudly instead of silently.

## Test plan

- [x] Unit tests (`tests/ut/patch/test_patch_sp_with_pp.py`, CPU, mocked TP comms): enable predicate matrix, boundary gather + pad trim, model-forward wrapper, default-off no-op, PP=1 view construction, caller-config non-mutation — **10/10 pass** in the NPU container
- [x] Ascend 910 smoke, Qwen3.5-35B-A3B, PP2 × DP2 × TP2 + EP + `allgather_reducescatter` (PR #13946 base):
  - env on: prefill runs SP (per-layer reduce-scatter kernels), decode runs the downgraded all-reduce path, outputs correct across prompts — matches the pre-patch SP-off baseline
  - env off: outputs correct (default behaviour unchanged; patch not loaded)
- [x] ruff format + check clean

### Note: pre-existing PP=1+SP issue (not introduced here)

While matrix-testing we found **PP=1 + native SP-MoE decode produces garbled output** on the PR #13946 base with a clean vLLM — the same NaN failure mode (T=1 padding through norm) that the runtime downgrade in this patch fixes for PP>1. This is orthogonal to this PR (with the feature off, PP=1 behaviour is untouched) but likely deserves a separate investigation; the downgrade idea here may be the fix.

DeepSeek-V2 path is implemented symmetrically but only Qwen3.5-35B-A3B was hardware-verified.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/patch/platform/patch_sp_with_pp.py
+++ b/vllm_ascend/patch/platform/patch_sp_with_pp.py
@@ -1,0 +1,381 @@
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+# This file is a part of the vllm-ascend project.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Enable sequence-parallel MoE together with pipeline parallelism.
+
+Upstream vLLM disables SP-MoE when PP > 1 because the PP boundary could not
+carry sequence-sharded hidden states (see the TODO in deepseek_v2.py). This
+opt-in patch (VLLM_ASCEND_ENABLE_SP_WITH_PP=1) re-enables the combination
+for qwen3_next / qwen3_5 / deepseek_v2 by:
+
+1. overriding the per-layer SP flag after model construction (dropping the
+   pipeline_parallel_size == 1 condition),
+2. replacing the decoder-layer forwards with versions that downgrade SP to
+   an explicit all-reduce on steps whose token count is below the TP size
+   (decode steps; SP padding rows otherwise produce NaN that spreads
+   through every subsequent gather),
+3. wrapping the model forwards so IntermediateTensors handed across PP
+   stages are gathered back to full-token tensors first. The trigger is an
+   explicit SP flag rather than a shape check: after SP padding a shard can
+   equal full_num_tokens (T=1, tp=2), which would silently skip the gather
+   and leak a pad row across the boundary.
+
+Verified on Ascend 910 with Qwen3.5-35B-A3B (PP2 x DP2 x TP2 + EP +
+allgather_reducescatter): prefill runs SP (per-layer reduce-scatter
+kernels), decode runs the downgraded all-reduce path, outputs match the
+PP=1-SP and PP>1-SP-off baselines.
+"""
+
+import functools
+import inspect
+from typing import TYPE_CHECKING
+
+import torch
+from vllm.distributed import (
+    get_tensor_model_parallel_world_size,
+    tensor_model_parallel_all_gather,
+    tensor_model_parallel_all_reduce,
+    tensor_model_parallel_reduce_scatter,
+)
+from vllm.model_executor.models.utils import sequence_parallel_chunk
+from vllm.sequence import IntermediateTensors
+
+from vllm_ascend.envs import VLLM_ASCEND_ENABLE_SP_WITH_PP
+
+if TYPE_CHECKING:
+    # runtime-injected in _install_patches; imported here only for type
+    # checkers (never at module import time, see apply())
+    from vllm.model_executor.models.deepseek_v2 import (
+        DeepseekAttention,
+        DeepseekV2MLP,
+    )
+
+
+def _qwen3next_layer_forward(
+    self,
+    hidden_states: torch.Tensor,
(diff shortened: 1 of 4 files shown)


AceCoder0:
All author-side checks are green (pre-commit incl. mypy, select-tests, coverage, DCO). The remaining ci-gate failure is the `ready` label gate waiting on maintainer dispatch, same as the repo's standard flow.

Hardware smoke (PR #13946 base, Ascend 910, Qwen3.5-35B-A3B, PP2×DP2×TP2 + EP + allgather_reducescatter):
- env on: prefill SP active (per-layer reduce-scatter kernels), decode on the downgraded all-reduce path, outputs correct across prompts
- env off: outputs correct, default behaviour unchanged (patch not loaded)
- UT 10/10 in the NPU container

Also documented in the description: PP=1 + native SP-MoE decode produces garbled output on this base **with the feature off** — pre-existing upstream issue (SP padding rows through norm on decode steps), orthogonal to this PR; the runtime-downgrade mechanism here is likely the fix for it. Happy to split that into a separate PR if maintainers prefer.

vllm_ascend/patch/platform/patch_sp_with_pp.py
```diff
@@ -0,0 +1,370 @@
...
+def _qwen3next_layer_forward(
+    self,
+    hidden_states: torch.Tensor,
+    residual: torch.Tensor | None,
+    positions: torch.Tensor = None,
+    **kwargs: object,
+):
+    full_num_tokens = positions.shape[-1]
+    input_is_sequence_parallel = (
+        self.use_attn_reduce_scatter_for_moe and residual is not None and hidden_states.shape[0] != full_num_tokens
```
nu11ptr7: The docstring says the trigger is "an explicit SP flag rather than a shape check", but `input_is_sequence_parallel` here is still `hidden_states.shape[0] != full_num_tokens` - exactly the shape check the docstring warns can misread a padded shard (T=1, tp=2). If that case ever reaches a layer with SP active, the gather gets skipped and a pad row flows into attention. Maybe drive this from the same explicit flag so code and docstring agree.

AceCoder0:
Both review comments addressed:

1. `DeepseekAttention` is now injected in `_install_patches()` alongside the other runtime-resolved symbols (real catch — it would have raised NameError on the DeepSeek FP16 path).

2. `input_is_sequence_parallel` in both copied layer forwards now also requires `hidden_states.shape[0] >= get_tensor_model_parallel_world_size()`, so a padded shard that happens to equal `full_num_tokens` (T=1, tp=2) is never treated as sequence-parallel input. Code and docstring now agree; the safety no longer depends on the downgrade path and the entry check staying in sync implicitly.

Re-verified after the changes: UT 10/10, and Ascend 910 smoke (PP2×DP2×TP2, env on) still produces correct output on both test prompts.

AceCoder0:
Rebased onto current main (post-#13946 merge) and adapted to the new verified vLLM commit `cdc4824a2`: upstream reshaped `Qwen3NextModel.forward` (model-level `sequence_parallel_chunk` entry + `assert residual is None`, which only holds on the first stage). The patch now installs a copy of that forward with the entry block dropped and the boundary gather folded in; the decoder-layer copies keep the verified semantics. A new drift sentinel anchors the upstream entry block.

Re-verified on the new base (vllm-ascend @ `9794b38fa` + vLLM @ `cdc4824a2`, Ascend 910, Qwen3.5-35B-A3B, PP2×DP2×TP2 + EP): UT 10/10, smoke outputs correct on both test prompts.

AceCoder0:
Update: the gate now covers **PP=1 as well** — the patched layer forwards' small-batch downgrade also fixes the upstream PP=1 + SP-MoE decode NaN disclosed earlier (garbled output with the feature off). Verified on the same hardware/base: PP=1 previously garbled, now correct; PP=2 unchanged (re-verified). Commit message carries the details; happy to split into a separate PR if maintainers prefer.

PR #14738, opened 2026-08-21
[BugFix][Performance] Skip identity d2t mapping for full-vocab drafts

drslark:

When the draft model uses the full target vocabulary (`draft_vocab_size == vocab_size`), the `draft_id_to_target_id` mapping is the identity. The draft's `compute_logits` still builds a full-vocab `-inf` buffer and scatters logits through this identity mapping on every call, which is pure overhead: measured ~1.4 ms per call at (112, 151936) bf16 — ~55x the cost of a plain copy, and comparable to the lm_head matmul itself.

This drops the mapping right after the draft model loads when the vocab sizes are equal, so `compute_logits` takes the fast path and returns logits unchanged.

No.

1. Verified with a Qwen3-8B target and the EAGLE3 draft `Qwen3-8B-speculator-eagle3`: the live `draft_id_to_target_id` table dumped from a real run equals `arange` (identity), and the scatter output is bit-identical to its input for real tensors, so skipping it is a no-op transform.
2. A micro-benchmark of the removed op shows `logits_new[:, targets] = logits` at (112, 151936) bf16 takes ~1.42 ms/op, vs 0.019 ms for a plain `copy_`.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/spec_decode/llm_base_proposer.py
+++ b/vllm_ascend/spec_decode/llm_base_proposer.py
@@ -259,6 +259,15 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
                 "disable_padded_drafter_batch in the speculative_config."
             )
 
+    def _maybe_remove_d2t(self, draft_model: nn.Module) -> None:
+        """Drop the identity d2t mapping of a full-vocab EAGLE3 draft."""
+        if self.method != "eagle3":
+            return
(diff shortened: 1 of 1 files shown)


xuyucheng220:
对glm无效

slippersss (approved):

drslark:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

weijinqian0 (approved):

PR #14742, opened 2026-08-21
[CI]Update the test case regex to handle rename scenarios. Restrict rename handling and precision testing to product code only (Python files).

shenhui-cli:
Updates the test case matching regex to properly handle file rename scenarios. Restricts the rename detection and handling logic to Python product code files only.

This PR makes the following changes:
Updates the test case matching regex to properly handle file rename scenarios.
*Restricts the rename detection and handling logic to Python product code files only (e.g., .py). Non-Python files are excluded from rename processing.
No. This PR does not introduce any user-facing changes.
This patch was tested through the following approaches:
Test the rename scenario for test cases, as well as other scenarios.

- vLLM version: v0.27.1

### DIFF
--- a/.github/workflows/scripts/test_selector.py
+++ b/.github/workflows/scripts/test_selector.py
@@ -59,10 +59,13 @@ def _get_test_files_from_pr_diff(diff_file: str) -> list[str]:
         return test_files_found
 
     # Pattern to match test file paths: tests/e2e/pull_request/ or tests/ut/ directory
-    # In diff output: +++ b/tests/ut/core/test_xxx.py
+    # In diff output:
+    #   - +++ b/tests/ut/core/test_xxx.py (new/modified test file)
+    #   - rename to tests/ut/attention/test_xxx.py (renamed test file)
     # Test files must be in tests/e2e/pull_request/ or tests/ut/ directory and start with test_
     test_file_pattern = re.compile(
-        r"^\+\+\+ [ab]/(tests/e2e/pull_request(?:/.+)?/test_\w+\.py|tests/ut(?:/.+)?/test_\w+\.py)", re.MULTILINE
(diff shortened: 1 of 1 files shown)


.github/workflows/scripts/test_selector.py
```diff
@@ -459,7 +462,9 @@ def detect_renames(self, diff_output: str) -> dict[str, str]:
                     # Remove a/ or b/ prefix if present
                     old_path = current_old_path[2:] if current_old_path.startswith("a/") else current_old_path
                     new_path = current_new_path[2:] if current_new_path.startswith("b/") else current_new_path
-                    renames[old_path] = new_path
+                    # Only record product code renames (vllm_ascend/)
+                    if old_path.startswith(f"{REPO_NAME}/"):
```
shenhui-cli, line 466: Filter out changes to non-product code files.

.github/workflows/scripts/test_selector.py
```diff
@@ -59,10 +59,13 @@ def _get_test_files_from_pr_diff(diff_file: str) -> list[str]:
...
 
     # Pattern to match test file paths: tests/e2e/pull_request/ or tests/ut/ directory
-    # In diff output: +++ b/tests/ut/core/test_xxx.py
+    # In diff output:
+    #   - +++ b/tests/ut/core/test_xxx.py (new/modified test file)
+    #   - rename to tests/ut/attention/test_xxx.py (renamed test file)
     # Test files must be in tests/e2e/pull_request/ or tests/ut/ directory and start with test_
     test_file_pattern = re.compile(
-        r"^\+\+\+ [ab]/(tests/e2e/pull_request(?:/.+)?/test_\w+\.py|tests/ut(?:/.+)?/test_\w+\.py)", re.MULTILINE
+        r"^(?:\+\+\+ [ab]/|rename to )((?:tests/e2e/pull_request(?:/.+)?/test_\w+\.py|tests/ut(?:/.+)?/test_\w+\.py))",
```
shenhui-cli, line 67: Enhance the regex to support test case rename scenarios.

wangxiyuan (approved):

PR #14748, opened 2026-08-21
[v0.26.0][Feature] dspark: always replicate the Markov head via ReplicatedGroup

Levi-JQ:

The DSpark Markov head (`markov_w1` VocabParallelEmbedding + `markov_w2` ParallelLMHead) serves the draft-decoding causal-correction loop, which runs serially per draft position. Under TP sharding every step pays a TP all-reduce (w1) and all-gather (w2), and these collectives are latency-bound: measured on a TP16 Kimi-K3 DSpark deployment, each costs ~240 us regardless of payload, which dominates the markov loop.

This PR replicates the Markov head on every rank instead of sharding it: any `AscendVocabParallelEmbedding` whose prefix contains "markov" is routed to a `ReplicatedGroup` — a pure world_size=1 stand-in (no `hcclCommInitRootInfoConfig`) — so each rank holds the full V x r / r x V table (~150 MB extra HBM per rank) and the per-step collectives disappear. This matches upstream vLLM, where the markov head is replicated unconditionally (vllm#49731; deepseek_v4_dspark in #15496 likewise replicates it), so **no new config option is introduced**: the prefix routing itself is the switch, and it is checked before the lmhead prefix match because "markov_head" contains "head".

- before
image-20260826144751584

- after
image-20260826144547733

No new config or API. For DSpark models under TP > 1, the Markov head is now replicated on every rank instead of TP-sharded, matching upstream vLLM behavior: draft outputs are numerically equivalent (any difference is limited to floating-point reduction order), while the per-draft-step TP collectives are removed at the cost of ~150 MB extra HBM per rank.

End-to-end PD-separated deployment — kimi-k3, 8P+8D (16×Ascend910A3/node), `markov_tensor_parallel_size=1` (P+D), dspark num_spec=7 greedy.

- random 40k/700

| percentile | metric | markov=1 | markov=0 | diff |
| --- | --- | --- | --- | --- |
| P50 | ITL | 98.71ms | 101.19ms | -2.5% |
| P90 | ITL | 102.03ms | 105.07ms | -2.9% |

- online_20k_40k

| percentile | metric | markov=1 | markov=0 | diff |
| --- | --- | --- | --- | --- |
| P50 | ITL | 98.19ms | 100.60ms | -2.4% |
| P99 | ITL | 123.65ms | 126.39ms | -2.2% |

### DIFF
--- a/.github/workflows/scripts/test_config.yaml
+++ b/.github/workflows/scripts/test_config.yaml
@@ -534,8 +534,10 @@
   base: spec_decode_base
   source_file_dependencies:
     - vllm_ascend/spec_decode/dspark_proposer.py
+    - vllm_ascend/ops/vocab_parallel_embedding.py
   tests:
     - tests/e2e/pull_request/one_card/spec_decode/test_dspark.py
+    - tests/e2e/pull_request/four_card/spec_decode/test_dspark_qwen3.py
 
 - name: spec_decode_extract_hidden_states
   optional: false
@@ -809,6 +811,7 @@ estimated_times:
   tests/e2e/pull_request/four_card/context_parallel/test_prefix_caching_cp.py: 210
   tests/e2e/pull_request/four_card/spec_decode/test_mtp_qwen3_next.py: 450
   tests/e2e/pull_request/four_card/spec_decode/test_dspark_deepseekv4.py: 0
+  tests/e2e/pull_request/four_card/spec_decode/test_dspark_qwen3.py: 360
   tests/e2e/pull_request/four_card/spec_decode/test_mtp_step3p5.py: 30
   tests/e2e/pull_request/eight_card/test_glm5_2.py: 3600
   tests/e2e/pull_request/four_card/test_data_parallel_tp2.py: 20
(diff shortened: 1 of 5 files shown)


vllm_ascend/utils.py
```diff
@@ -846,6 +846,10 @@ def oproj_tp_enable() -> bool:
     return get_ascend_config().finegrained_tp_config.oproj_tensor_parallel_size > 0
 
 
+def markov_tp_enable() -> bool:
+    return get_ascend_config().finegrained_tp_config.markov_tensor_parallel_size > 0
+
```
Levi-JQ: （1）增加e2e
（2）默认打开即可 不需要finegrained_tp_config

drslark (changes requested):
Add an e2e about it.

vllm_ascend/utils.py
```diff
@@ -846,6 +846,10 @@ def oproj_tp_enable() -> bool:
     return get_ascend_config().finegrained_tp_config.oproj_tensor_parallel_size > 0
 
 
+def markov_tp_enable() -> bool:
```
drslark: Follow https://github.com/vllm-project/vllm/pull/49731 and make it disabled default.

drslark (changes requested):

drslark (approved):
LGTM.
I think current e2e is enough.

weiguihua2 (approved):

Levi-JQ:
> Add an e2e about it.

done

lijiahang226 (approved):

lijiahang226:
@slippersss, is there anything else needed before we can move this toward merge?

PR #14749, opened 2026-08-21
[Performance] Fuse the Qwen3.5 GDN decode pipeline on Ascend

nodeeeeee:

Qwen3.5-9B GDN decode currently materializes full intermediate tensors and launches separate operators for causal convolution, Q/K normalization, gating, the recurrent update, and gated RMSNorm. For single-token decode, these operators repeatedly read and write the same head-local data, so launch and memory-traffic overhead dominate their small amount of arithmetic.

This PR adds an opt-in Triton kernel that preserves the same GDN equations but decomposes the work by GQA group. Each program processes one Q/K head together with its two V/Z heads and directly continues through convolution state update, Q/K L2 normalization, gate construction, recurrent-state update, and gated RMSNorm. This avoids the full-tensor materialization barriers between those steps.

The patch is based on the current `releases/v0.23.0` branch at `d68bb5e`. Its diff is limited to the GDN implementation, Qwen3.5 GDN dispatch, the GDN environment switch, and focused GDN tests. It does not include the separate RMSNorm--MatMul experiment, its AscendC sources, bindings, build changes, or Qwen3.5 MLP patch.

The integration also keeps runtime decode selection behind vLLM's existing `qwen_gdn_attention_core` custom-op boundary. This is required for graph mode: the initial compile/profile pass does not carry attention metadata, and tracing the Python branch directly would otherwise permanently capture the original path.

The optimized path is deliberately narrow and falls back to the existing implementation unless all of the following hold:

- `VLLM_ASCEND_ENABLE_GDN_DECODE_TILE_PIPELINE=1`;
- Qwen3.5-9B GDN layout, BF16, and convolution width 4;
- TP=1 or TP=2;
- non-speculative decode with no prefill tokens;
- real decode batch size from 1 through 16.

Prefill, speculative decoding, unsupported dtypes/shapes, and other models keep the original path.

Yes, it adds the optional environment variable `VLLM_ASCEND_ENABLE_GDN_DECODE_TILE_PIPELINE`. It defaults to `0`, so existing behavior is unchanged unless users explicitly enable the optimization.

#### Unit tests

```bash
pytest -q tests/ut/ops/test_gdn_attn_builder.py
```

Result: `22 passed`. New cases cover graph-safe custom-op dispatch, PIECEWISE projection/metadata padding, and fallback for unsupported dtype and TP modes.

#### Testbed

- Ascend 910B3, BF16, Qwen3.5-9B, vLLM 0.23.0;
- exclusive NPU allocation for each A/B implementation;
- original and optimized paths compared with identical model and workload configuration.

The exhaustive results below were collected on the pre-cleanup experimental branch. The GDN patch carried into this PR is semantically identical: cleanup only replaced repeated literals with named constants and applied formatting, the base-branch update did not modify any of the five GDN PR paths, and the separate RMSNorm--MatMul switch remained disabled throughout the measurements. A reduced confirmation on the exact PR head is reported separately below.

#### Exact PR-head reduced confirmation

The clean PR commit `59b249e6` was deployed independently and its five changed file hashes were checked against the local worktree before the run. The reduced matrix used B=1/16, prefill=16/1024, decode=256, TP=1/2, and eager, PIECEWISE, and FULL+FX execution. Original and optimized processes were interleaved in ABBA/BAAB order. TP1 used two independent device replicas; every shape had at least 12 samples per implementation for TP1 and six for TP2.

|   TP | Mode      | Cases | Geomean wall-time speedup |         Range | Faster cases | Mean TPOT reduction |
| ---: | --------- | ----: | ------------------------: | ------------: | -----------: | ------------------: |
|    1 | eager     |     4 |                    1.496x | 1.477--1.533x |          4/4 |              33.73% |
|    1 | PIECEWISE |     4 |                    1.687x | 1.599--1.729x |          4/4 |              42.40% |
|    1 | FULL + FX |     4 |                    1.078x | 1.063--1.087x |          4/4 |               7.58% |
|    2 | eager     |     4 |                    1.371x | 1.342--1.405x |          4/4 |              27.81% |
|    2 | PIECEWISE |     4 |                    1.784x | 1.717--1.841x |          4/4 |              44.91% |
|    2 | FULL + FX |     4 |                    1.095x | 0.806--1.455x |          3/4 |               8.59% |

All 36 independent benchmark processes and 144 measured shape records completed without failure. The small FULL+FX matrix showed high TP2 variance, including a slower B=16/prefill=16 shape, while its geomean and mean TPOT remained improved; the larger original campaign below observed 29/30 faster TP2 FULL+FX cases.

Probability precision was then checked on the same exact commit with FULL+FX, decode=64, B=1/16, prefill=16/1024, TP=1/2, and captured top-20 log-probabilities. Each TP used an Original/Fused/Original-repeat sequence. Only positions whose preceding token context remained identical in all compared runs were included; 7,403 such comparison positions were processed overall.

|   TP | Comparison                  | Paired positions | Argmax agreement | Mean top-1 probability error | P99 top-1 probability error |
| ---: | --------------------------- | ---------------: | ---------------: | ---------------------------: | --------------------------: |
|    1 | Original vs Fused           |            1,654 |          99.516% |                     0.002016 |                    0.030316 |
|    1 | Original vs Original repeat |            1,654 |          99.758% |                     0.001579 |                    0.029818 |
|    2 | Original vs Fused           |            1,627 |          99.262% |                     0.001900 |                    0.029081 |
|    2 | Original vs Original repeat |            1,627 |         100.000% |                            0 |                           0 |

For TP1, the A/B mean top-1 error above the independent-run A/A baseline was 0.000437; for TP2 it was 0.001900. These bounded differences are expected from the changed BF16 operation order and are not evidence of a different real-number algorithm. The test does not claim full-vocabulary KL because only top-20 probabilities were captured.

#### Original validation campaign: operator path and numerical checks

- GDN decode path event latency improved by 5.86--6.26x for TP1/TP2 and B=1/2/4/8/16.
- Kernel-profiler attribution measured 14.13 us per optimized GDN layer versus 78.30 us for the selected original subchain, approximately 5.54x.
- Batched and per-request optimized executions were bit-identical, excluding cross-request state mixing.
- Maximum operator output difference was 0.11914062 and maximum recurrent-state difference was 0.02343604. These differences come from BF16 operation-order changes; the real-number equations are unchanged.

#### Original validation campaign: eager end-to-end Cartesian sweep

The sweep covered the Cartesian product of:

- batch: 1, 2, 4, 8, 16;
- prefill tokens: 16, 64, 256, 1024;
- decode tokens: 64, 256, 1024, 4096;
- TP: 1 and 2.

|   TP | Cases | Geomean wall-time speedup | Faster cases | Mean TPOT reduction |
| ---: | ----: | ------------------------: | -----------: | ------------------: |
|    1 |    80 |                     1.59x |        79/80 |              38.68% |
|    2 |    80 |                     1.63x |        80/80 |              37.87% |

#### Original validation campaign: graph-mode end-to-end sweep

Each row contains 30 shape cases and at least six samples per implementation. There were no failed case records.

|   TP | Graph mode | Geomean speedup | Faster cases | Mean TPOT reduction |
| ---: | ---------- | --------------: | -----------: | ------------------: |
|    1 | PIECEWISE  |          1.682x |        30/30 |              43.62% |
|    2 | PIECEWISE  |          1.722x |        30/30 |              43.65% |
|    1 | FULL + FX  |          1.072x |        30/30 |               7.61% |
|    2 | FULL + FX  |          1.111x |        29/30 |              10.06% |

FULL graph mode has a smaller gain because graph replay already removes much of the launch overhead; PIECEWISE mode retains the larger memory-traffic and synchronization benefit.

#### Original validation campaign: compatibility and quality

- Prefix caching: 12 cases per implementation completed.
- MTP: 4 cases per implementation completed and correctly used the original fallback path.
- Long case: B=16, prefill=1024, decode=4096 completed for both paths.
- Probability comparison across tested execution modes retained 99.57--99.67% top-1 agreement; the additional mean top-1 probability error was approximately 0.0011--0.0020.
- GSM8K 5-shot, 1,319 examples: strict exact match changed from 87.491% to 87.263% (-0.227 percentage points), below the reported single-run standard error of approximately 0.91 percentage points.
- WikiText-2 perplexity was identical: word perplexity 9.608194782 and byte perplexity 1.526718360 for both paths.

The implementation is therefore not claimed to be token-exact over long greedy generation. The observed divergence is consistent with BF16 rounding after the mathematical reassociation; probability-, task-, and perplexity-level checks do not show a material quality regression.

- vLLM version: v0.23.0


vllm_ascend/ops/gdn.py
```diff
@@ -75,6 +84,32 @@ def forward(
         2. Core attention (custom op)
         3. Output projection
         """
+        if self._supports_qwen35_decode_tile_pipeline(hidden_states):
+            num_tokens = hidden_states.size(0)
```
nu11ptr7, line 88: Quick one on the fast path: `_supports_qwen35_decode_tile_pipeline` only checks shapes/config, not that this is actually a one-token non-spec decode (`num_prefills == 0`, spec masks None). With the env flag on, a prefill batch whose hidden dim matches would get routed into this branch and skip the conv/SSM state path entirely. `_try_qwen35_decode_tile` has those metadata checks - maybe gate the fast path on the same conditions.

nodeeeeee: Thanks for raising this. Currently there are two layers of gates to ensure that prefill is routed to the right path, not skipping any prefill path:
1. `_supports_qwen35_decode_tile_pipeline`: checks tensor layout and dtype, etc.   These metadata are visible at compile time.
2. `_try_qwen35_decode_tile`: This checks runtime information including prefill/decode, num of tokens, etc.

The logic here is:
```
if _supports():
    ...
    ...
    # inside `qwen_gdn_attention_core`
    _try_qwen35()  
        if or layer_metadata.num_prefills != 0:   # here we check prefill/decode
            return None     -> return to _frward_core_ascend, then fused_decode_output is None, it will fall back to the original path
        -> gdn_decode_tile()
    else:
        fallback
else:
    fallback
```
So if it is a prefill it will eventually be captured by try_qwen35...(), and fallback to the original path.

However, this is truly a bit confusing. I can try to make the logic clearer if you want.

nodeeeeee:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #14755, opened 2026-08-22
[Ops][Misc] Bundle sparse KV CPU helpers and register with _C_ascend

ader47:

- Moves the Sparse KV offload CPU helpers from the Python package directory to `csrc/kv_transfer`.
- Registers the helpers with the existing `_C_ascend` extension instead of compiling a separate extension at service startup.
- Uses the package build's A3 and OpenMP configuration and removes runtime `CC`/`CXX` environment mutation.
- Updates the Sparse KV offload guide to distinguish package-build and runtime OpenMP requirements.

This keeps native code in the repository's native-source tree, ensures it is included in the built package, and removes runtime compiler dependencies from Decode startup.

Yes. Sparse KV offload no longer JIT-compiles its CPU helpers when the manager is initialized. There is no API or configuration change, but A3 source builds must provide OpenMP when building the native extension, and prebuilt installations must retain the corresponding OpenMP runtime library.

- `bash format.sh ci`
- Python static checks for `sparse_kv_offload_manager.py`
- C++17 syntax compilation of `csrc/kv_transfer/sparse_kv_offload.cpp`
- Standalone loading and functional smoke validation of both registered CPU operators
- Full package build and end-to-end Sparse KV offload inference validation on A3 hardware


vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
```diff
@@ -1014,7 +973,7 @@ def _onload_topk_kv_cpu(self, args):
             num_tokens_buffer,
             layer_id,
         ) = args
-        self.sparse_kv_offload_cpp.lru_resident_compact(
+        torch.ops._C_ascend.sparse_kv_lru_resident_compact(
             lru_req_ids_ptr,
```
nu11ptr7, line 1103: Heads-up on the fallback removal: with `_build_cpp()` gone, these ops only exist when the package is built for SOC_VERSION `ascend910_93.*` (the CMake gate). On any other build, this line raises a missing-op error at runtime instead of building on demand like before. Maybe add an init-time check so it fails with a clear message.

ader47: I added an initialization-time check in `SparseKVOffloadManager`. It first triggers the lazy custom-op loading, then verifies that both Sparse KV native operators are registered before allocating the CPU KV pool. Non-A3 builds now fail early with a clear message explaining the `SOC_VERSION=ascend910_93.*` requirement.

I also added unit coverage for both available and missing operator registrations. Fixed in commit `192d0bd834`.

maoxx241 (commented):
Please consider the inline comment about where native-op availability should be validated.

vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
```diff
@@ -313,6 +333,8 @@ def __init__(
         kv_cache_config: KVCacheConfig,
         sparse_kv_offload_config: SparseKVOffloadConfig,
     ):
+        _require_sparse_kv_offload_ops()
```
maoxx241: I do not think this validation should happen in `SparseKVOffloadManager.__init__`. On `main`, `enable_custom_op()` is already the centralized lazy-loading boundary: it bootstraps the custom-op environment, imports `vllm_ascend_C` to register `torch.ops._C_ascend`, registers the meta implementations, and caches the result. More specifically, after `vllm_ascend_C` is imported, please check whether the current SoC is A3; on A3, load/verify the two Sparse KV operators as part of custom-op initialization, while non-A3 paths should skip this A3-only requirement. Then `SparseKVOffloadManager` can rely on the operators already being available, and `_require_sparse_kv_offload_ops()` does not need to run from its constructor.

ader47:
Rebased this PR onto the latest `main` to pick up the vLLM PCP import compatibility fix. The new CI run now passes `pre-commit`, all `mypy` checks, coverage recommendation, and `select-tests`.

`ci-gate` is now blocked only because selected tests require `ready-precise` or `ready-all`, and the PR author account does not have permission to add labels. Could a maintainer please add `ready-precise` to trigger the selected native/test jobs?

CI run: https://github.com/vllm-project/vllm-ascend/actions/runs/32946987639

ader47:
Updated this PR to match the reviewer-refined implementation in #15016, rebased it onto the latest `main`, and added the new A3 Sparse KV native-op tests plus their `estimated_times` entry.

The new CI run passes pre-commit/mypy, coverage recommendation, and select-tests. The PR is mergeable; `ci-gate` is blocked only because selected tests require a maintainer-added `ready-precise` or `ready-all` label. Could a maintainer please add the appropriate ready label?

CI run: https://github.com/vllm-project/vllm-ascend/actions/runs/33153507204

.github/workflows/configs/nightly_config.yaml
```diff
@@ -171,6 +171,11 @@ a3:
         size: 2
   single_node:
     test_config:
+      # pytest-driven tests
+      - name: sparse-kv-offload-ops
```
ZT-AIA, line 175: Delete this directly.

csrc/torch_binding.cpp
```diff
@@ -2121,6 +2502,25 @@ TORCH_LIBRARY_EXPAND(CONCAT(_C, _ascend), ops)
...
+            "int miss_count_ptr, int miss_tokens_ptr, int miss_slots_ptr, int token_mark_workspace_ptr, "
+            "int token_pos_workspace_ptr, int slot_workspace_ptr, int miss_position_workspace_ptr, "
+            "int epochs_ptr, int num_reqs, int topk, int capacity, int max_token, int workspace_threads, "
+            "int requested_threads) -> ()");
+    ops.impl("sparse_kv_lru_resident_compact", c10::DispatchKey::CompositeExplicitAutograd,
+             &vllm_ascend::sparse_kv_lru_resident_compact);
+
+    ops.def("sparse_kv_compute_lru_resident_addrs(Tensor miss_count, Tensor miss_tokens, Tensor miss_slots, "
+            "Tensor block_table, int block_size, int token_size_bytes_k, int token_size_bytes_v, "
+            "int gvas_k_base, int gvas_v_base, int addr_k_base, int addr_v_base, int resident_capacity, "
```
ZT-AIA, line 2517: It also needs to be registered in. `torch_binding_meta.cpp`

PR #14758, opened 2026-08-22
[Bugfix] Synchronize NPU worker device

Ronald1995:
## Summary

- Provide `NPUWorker.synchronize_device()`, which blocks on `torch.npu.synchronize()`, so the pause-completion barrier that vLLM added in [#52914](https://github.com/vllm-project/vllm/pull/52914) waits for in-flight NPU work. Carrying the method on the worker also keeps `wake_up()` correct on vLLM releases that predate #52914 (the verified `v0.28.0` tag does not define `WorkerBase.synchronize_device`).
- Synchronize at the end of `NPUWorker.wake_up()` so callers can safely begin same-device work (for example RLHF weight sync) as soon as wake-up returns.
- Add unit coverage and real-hardware coverage of the contract.

## Upstream context

Ports the Ascend backend requirement from vLLM [#52914](https://github.com/vllm-project/vllm/pull/52914): a resolved pause promises an idle device, and backends outside `torch.accelerator` must provide their own wait.

## What this PR adds test-wise

Unit tests (`tests/ut/worker/test_worker_v1.py`): `wake_up()` ends with the barrier, and the barrier waits through `torch.npu.synchronize()`.

NPU system tests:

- `tests/e2e/pull_request/one_card/rlhf/state_transitions/test_npu_device_synchronize.py` — on real hardware, `synchronize_device()` and `wake_up()` must leave the NPU stream idle.
- `tests/e2e/pull_request/two_card/rlhf/state_transitions/test_pause_device_idle.py` — the Ascend port of upstream's `test_dp_pause_completion_implies_device_idle`: inflate the DP dummy batches, then require an idle device the moment `pause_generation()` returns. Skipped when the installed vLLM predates #52914.
- `tests/e2e/pull_request/one_card/rlhf/state_transitions/npu_stream_busy.py` — the device-busy helper both files share.

`torch.cuda._sleep(200_000_000)`, which the upstream test uses to keep a CUDA stream busy, cannot run on NPU: `torch.npu` exposes no `_sleep`, and `torch.cuda._sleep` raises `AttributeError: module 'torch._C' has no attribute '_cuda_sleep'`. `enqueue_npu_busy_work()` queues real matmuls without waiting instead, and `torch.npu.current_stream().query()` reports the same "is the stream quiet" fact as `torch.cuda.current_stream().query()`.

## How was this patch tested?

On Ascend 910, 2 chips, vLLM main `84030bbe3d` (the commit pinned by `.github/vllm-main-verified.commit`):

| suite | result |
|---|---|
| `pytest tests/ut/worker/test_worker_v1.py -k "wake_up or synchronize"` | 5 passed |
| `pytest tests/e2e/pull_request/one_card/rlhf/state_transitions/test_npu_device_synchronize.py` | 2 passed |
| `pytest tests/e2e/pull_request/two_card/rlhf/state_transitions/test_pause_device_idle.py` | 1 passed |

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/worker/worker.py
+++ b/vllm_ascend/worker/worker.py
@@ -272,6 +272,8 @@ class NPUWorker(WorkerBase):
         if cleanup_enabled:
             self.sleep_wakeup_manager.wakeup(tags)
 
+        self.synchronize_device()
+
     def _check_weight_transfer_engine(self) -> None:
         if self.weight_transfer_engine is None:
             raise RuntimeError(
@@ -1044,6 +1046,10 @@ class NPUWorker(WorkerBase):
 
     def reset_encoder_cache(self) -> None:
         self.model_runner.reset_encoder_cache()
+
+    def synchronize_device(self) -> None:
+        """Block until all in-flight NPU work has completed."""
+        torch.npu.synchronize()
 
     def execute_dummy_batch(self) -> None:
         self.log_memory_stats()
--- a/tests/ut/worker/a2/test_worker_v1.py
+++ b/tests/ut/worker/a2/test_worker_v1.py
@@ -374,9 +374,10 @@ class TestNPUWorker(TestBase):
         mock_allocator_class.get_instance.return_value.sleep.assert_called_once_with(offload_tags=("weights",))
         self.assertEqual(mock_mem_get_info.call_count, 2)
 
+    @patch("vllm_ascend.worker.worker.NPUWorker.synchronize_device")
     @patch("vllm_ascend.worker.worker.CaMemAllocator")
     @patch("vllm_ascend.worker.worker.get_ascend_config")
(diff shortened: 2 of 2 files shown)


vllm_ascend/worker/worker.py
```diff
@@ -1045,6 +1047,10 @@ def pin_lora(self, lora_id: int) -> bool:
     def reset_encoder_cache(self) -> None:
         self.model_runner.reset_encoder_cache()
 
+    def synchronize_device(self) -> None:
+        """Block until all in-flight NPU work has completed."""
+        torch.npu.synchronize()
```
nu11ptr7, line 1052: `torch.npu.synchronize()` without args syncs whatever the *current* device is on the calling thread, and `wake_up()` can be invoked from a thread that never set a device - so this could silently barrier on the wrong NPU. Since `self.device` is already there, maybe pass it explicitly.

PR #14763, opened 2026-08-22
[BugFix][KV Pool] Fix GVA layerwise waiter state lookup

ader47:

Fixes an initialization-time AttributeError when AscendMultiConnector wires a layerwise buffer-reuse waiter into AscendStoreConnector. A previous refactor removed use_gva_layerwise from the outer connector, while set_external_slot_release_waiter continued to access it there.

Use KVPoolWorker.use_gva_layerwise as the source of truth and keep scheduler and non-GVA paths returning False. Add regression coverage for missing workers, non-GVA workers, and enabled GVA layerwise workers.

No API or configuration change. This restores startup for the affected AscendMultiConnector layerwise configuration.

- ruff check: passed
- ruff format --check: passed
- python compileall for the changed files: passed
- Focused pytest was attempted, but the local environment is missing psutil/cbor2 and stopped during collection before test execution. The added regression test is covered by CI.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/ascend_store_connector.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/ascend_store_connector.py
@@ -196,8 +196,8 @@ class AscendStoreConnector(KVConnectorBase_V1, SupportsHMA):
     # Worker Side Methods
     ############################################################
     def set_external_slot_release_waiter(self, waiter: Callable[[int], None]) -> bool:
-        connector_worker = getattr(self, "connector_worker", None)
-        if connector_worker is None or not connector_worker.use_gva_layerwise:
+        connector_worker = self.connector_worker
+        if not connector_worker.use_gva_layerwise:
             return False
         connector_worker.set_external_slot_release_waiter(waiter)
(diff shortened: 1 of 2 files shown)


vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/ascend_store_connector.py
```diff
@@ -196,9 +196,10 @@ def take_events(self) -> Iterable["KVCacheEvent"]:
     # Worker Side Methods
     ############################################################
     def set_external_slot_release_waiter(self, waiter: Callable[[int], None]) -> bool:
-        if not self.use_gva_layerwise or getattr(self, "connector_worker", None) is None:
+        connector_worker = getattr(self, "connector_worker", None)
```
maoxx241: Is there any valid path where `connector_worker` can be absent or `None` when this worker-side method is called? If not, `getattr` is unnecessary here; please access `self.connector_worker` directly, with an assertion if the invariant needs to be made explicit.

maoxx241 (commented):
Please check the inline comment about the worker invariant.

PR #14764, opened 2026-08-22
[BugFix] Fix Kimi K3 C8 DSpark and MiniMax M3 follow-up issues

HaoxinZong:

This PR rebases and combines the applicable changes from #14612 and #14704 onto the latest `vllm-ascend` main.

For Kimi K3, it:

- includes the K3 A3 MLA C8 prerequisite from #13225;
- fixes KV-cache initialization when A3 MLA C8, hybrid MLA + Mamba cache, and K3 DSpark speculative decoding with a BF16 draft cache are used together;
- keeps target and draft layers in one logical KV-cache group while allocating heterogeneous physical cache tensors safely;
- accounts for heterogeneous page sizes in cache capacity and memory calculations;
- preserves EAGLE metadata and per-layer cache layout selection;
- restores the Kimi K3 SiTU activation path.

For MiniMax M3 and shared experts, it:

- tolerates quantization configs without `quant_description` in RMSNorm;
- clamps the zero-prefix sparse-attention score range to prevent negative block access;
- uses the Triton sparse decode implementation on A5;
- defers shared-expert quantization imports to avoid loading unavailable optional implementations.

The old #14704 changes for `SequenceRowParallelOp` and `matmul_and_reduce*` are intentionally omitted. FlashComm v1 and those custom-op paths were removed from main by #13946. The upstream SP path passes pre-quantized MXFP8 tuples into the quantization method and performs communication on the resulting tensor, so the FlashComm-specific tuple reduction fix is no longer required.

This PR is intended to supersede #14612 and #14704.

Yes.

- Kimi K3 deployments using A3 MLA C8 with K3 DSpark can initialize and use heterogeneous target/draft KV-cache layouts.
- MiniMax M3 MXFP8 and sparse attention avoid the covered configuration and index-bound failures.
- There are no new public CLI or configuration-format changes.

Validated in the current workspace with:

```bash
python3 -m compileall -q <all changed Python files>
git diff --check upstream/main...HEAD
```

Both checks passed.

Ruff, pytest, and NPU end-to-end tests were not run in this workspace because the corresponding Python packages and NPU runtime are unavailable. CI should run the repository checks; before merge, an A3 Kimi K3 W4A8C8 + DSpark smoke test and a MiniMax M3 MXFP8 test with upstream SP (`DP > 1`, `TP > 1`, EP, and `allgather_reducescatter`) are recommended.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -759,10 +759,15 @@ class AscendMLAImpl(MLAAttentionImpl):
         if not self.use_mla_rope and self.enable_mlapo and device_type != AscendDeviceType.A5:
             logger.warning_once("MLAPO is disabled for MLA layers with RoPE disabled outside A5.")
             self.enable_mlapo = False
-        if not self.use_mla_rope and self.fa_quant_layer and device_type not in {
-            AscendDeviceType.A3,
-            AscendDeviceType.A5,
-        }:
+        if (
+            not self.use_mla_rope
+            and self.fa_quant_layer
+            and device_type
+            not in {
+                AscendDeviceType.A3,
+                AscendDeviceType.A5,
+            }
+        ):
             logger.warning_once("FA quant for no-RoPE MLA layers is supported only on A3/A5; falling back to BF16.")
             self.fa_quant_layer = False
         if self.fa_quant_layer:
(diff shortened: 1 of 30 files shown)


vllm_ascend/attention/mla_v1.py
```diff
@@ -742,8 +742,36 @@ def __init__(
...
+        # normally takes the fused MLA-prolog path. No-RoPE MLA layers bypass
+        # that prolog because their positional slice must not be reordered.
+        if (
+            not self.fa_quant_layer
+            and not self.use_mla_rope
+            and device_type == AscendDeviceType.A3
+            and self.vllm_config.quant_config is not None
+            and getattr(self.vllm_config.quant_config, "enable_fa_quant", False)
+        ):
+            self.fa_quant_layer = self.vllm_config.quant_config.is_fa_quant_layer(self.layer_name)
```
nu11ptr7, line 756: The A3 fallback gates on `getattr(quant_config, "enable_fa_quant", False)` but then calls `quant_config.is_fa_quant_layer(...)` unguarded - a config that has the flag but no such method (non-ModelSlim) would AttributeError. Maybe getattr it too, or check the class first.

PR #14601, opened 2026-08-20
[Feature][Spec Decode] Add Kimi K3 DSpark execution

maoxx241:
### What this PR does / why we need it?

- Registers the upstream Kimi K3 DSpark model as a hidden-state drafter.
- Normalizes supported DSpark configuration fields for the proposer.
- Invokes an optional model-owned shared-layer preparation hook before the generic proposer shares target embedding or LM-head modules; QuaRot loading and matrix conversion remain in modeling.
- Prepares proposals, verification metadata, multimodal placeholders, and corrected host sequence lengths through the existing Ascend DSpark runtime.
- Adds focused proposer, configuration, metadata, and utility tests.

### Checkpoint

[`RadixArk/Kimi-K3-DSpark`](https://huggingface.co/RadixArk/Kimi-K3-DSpark) is the public GQA draft checkpoint used by this compatibility path. The MLA path requires a Kimi K3 MLA draft checkpoint matching the target model.

### Dependencies

Merge after #14600 and #14839. The Kimi hybrid cache layout is provided by #14765.

### How was this patch tested?

- `git diff --check`
- Python syntax compilation for changed implementation and test files
- Focused proposer, speculative-config, model-owned shared-layer hook, multimodal-placeholder, and metadata tests
- Integrated GQA/MLA DSpark, QuaRot, ACLGraph, Prefix Cache, C64/C128, and finite non-empty output coverage

### Does this PR introduce any user-facing change?

Yes. Kimi K3 can use DSpark speculative decoding on Ascend.

- vLLM version: v0.27.1
- vLLM main: https://github.com/vllm-project/vllm/commit/ba07e4a48fc951300d97eb506217dd530583dea3

PR #14765, opened 2026-08-22
[Performance][KV Cache] Group Kimi K3 DSpark caches

maoxx241:

- Groups Kimi K3 target attention and DSpark draft attention by shared scheduler block ownership while retaining each layer's exact physical cache spec.
- Splits recurrent layers into the minimum balanced group count derived from the runtime layer layout.
- Supports TP8 and TP16, plus both GQA and MLA DSpark drafts, without hard-coding block or page sizes.
- Carries logical and physical block sizes through scheduler, block-table, model-runner, Mamba, and DSpark metadata paths.
- Preserves upstream MLA compression fields and separates causal target MLA from non-causal multi-token draft MLA.
- Preserves Mamba request capacity for Prefix Cache and speculative blocks by forwarding UniformTypeKVCacheSpecs capacity calculation to its inner specs when upstream still inherits the generic formula.
- Allocates block tables from the resolved per-rank capacity without multiplying replicated Mamba tables by DCP again. For a 133120-token limit, 384-token blocks and seven speculative slots, scheduler and worker both retain 354 columns at DCP1 and DCP4.

For the Kimi K3 layout, 24 target attention layers plus five draft layers form one group and 69 recurrent layers form three balanced groups: **29/23/23/23**. Both GQA and MLA DSpark create **five exact-spec metadata builders**: two attention builders plus three recurrent builders.

### Dependencies

Merge after #14597 and #14601.

- 31 targeted CPU block-table and Prefix Cache tests pass on vLLM 0.27.1; two coordinator-API cases introduced after v0.27.1 are skipped. Tests check raw/wrapped Mamba specs, DCP1/DCP4, scheduler/worker agreement, and writes to the last speculative slots. `bash format.sh ci` passes. Full-model NPU serving was not rerun for these capacity changes.

- `git diff --check`
- Python bytecode compilation for all changed implementation and test files
- Focused TP8/TP16, GQA/MLA, scheduler, Mamba-wrapper, block-table, compressed-cache, capacity, fallback, and proposer tests
- Integrated four-node DSpark coverage for known accuracy, Prefix Cache, long-context, concurrency, and hang-trigger cases

Yes. Compatible Kimi K3 DSpark deployments use four topology-derived KV groups for TP8 and TP16.

- vLLM version: v0.27.1

### DIFF
--- a/tests/ut/patch/platform/test_prefix_cache_cp_patches.py
+++ b/tests/ut/patch/platform/test_prefix_cache_cp_patches.py
@@ -38,6 +38,7 @@ from vllm_ascend.patch.platform.patch_kv_cache_utils import (
     group_and_unify_kv_cache_specs,
 )
 from vllm_ascend.patch.platform.patch_mamba_manager import AscendMambaManager
+from vllm_ascend.utils import vllm_version_is
 
 
 def _make_hybrid_kv_cache_config(
@@ -353,6 +354,8 @@ def test_kimi_k3_mla_dspark_uses_four_groups_and_five_builders() -> None:
 def test_kimi_k3_gqa_mixed_groups_preserve_scheduler_and_mamba_contracts() -> None:
     groups = _get_kimi_k3_dspark_mixed_kv_cache_groups(_make_kimi_k3_dspark_kv_cache_specs())
     assert groups is not None
+    max_model_len = 133120
+    vllm_config = _make_vllm_config(enable_prefix_caching=True, dcp=1, block_size=384)
     worker_config = KVCacheConfig(
         num_blocks=100,
         kv_cache_tensors=[],
@@ -361,17 +364,19 @@ def test_kimi_k3_gqa_mixed_groups_preserve_scheduler_and_mamba_contracts() -> No
 
     assert worker_config.has_mamba_layers
     assert worker_config.needs_kv_cache_zeroing
-    assert (
-        groups[1].kv_cache_spec.max_num_blocks_per_req(
-            _make_vllm_config(enable_prefix_caching=True, dcp=1, block_size=384),
-            3840,
-        )
-        == 17
-    )
+    worker_mamba_widths = {
+        group.kv_cache_spec.max_num_blocks_per_req(vllm_config, max_model_len) for group in groups[1:]
+    }
+    assert worker_mamba_widths == {354}
 
     scheduler_config = generate_scheduler_kv_cache_config([worker_config])
     assert isinstance(scheduler_config.kv_cache_groups[0].kv_cache_spec, MLAAttentionSpec)
     assert all(isinstance(group.kv_cache_spec, MambaSpec) for group in scheduler_config.kv_cache_groups[1:])
+    scheduler_mamba_widths = {
+        group.kv_cache_spec.max_num_blocks_per_req(vllm_config, max_model_len)
+        for group in scheduler_config.kv_cache_groups[1:]
+    }
+    assert scheduler_mamba_widths == worker_mamba_widths
     assert scheduler_config.needs_kv_cache_zeroing
 
 
@@ -399,10 +404,9 @@ def test_kimi_k3_gqa_mixed_groups_use_expected_physical_layout(monkeypatch) -> N
     assert sum(tensor.size for tensor in tensors) == available_memory
 
 
-def test_kimi_k3_gqa_mixed_grouping_falls_back_on_partial_signature() -> None:
+def test_kimi_k3_gqa_mixed_grouping_falls_back_on_unrecognized_layer() -> None:
     specs = _make_kimi_k3_dspark_kv_cache_specs()
-    draft_layer = "model.layers.93.self_attn.attn"
-    specs.pop(draft_layer)
+    specs["unrecognized.layer"] = next(iter(specs.values()))
 
     assert _get_kimi_k3_dspark_mixed_kv_cache_groups(specs) is None
 
@@ -560,6 +564,10 @@ def test_get_kv_cache_coordinator_delegates_hybrid_without_caching(monkeypatch) 
 
 
 @pytest.mark.parametrize("num_prefill_lookahead", [0, 8])
+@pytest.mark.skipif(
+    vllm_version_is("0.27.1"),
+    reason="num_prefill_lookahead was added to the coordinator after v0.27.1",
+)
 def test_get_kv_cache_coordinator_forwards_prefill_lookahead(
     monkeypatch,
     num_prefill_lookahead: int,
--- a/tests/ut/patch/platform/test_prefix_cache_cp_patches.py
+++ b/tests/ut/patch/platform/test_prefix_cache_cp_patches.py
@@ -1,12 +1,14 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import math
+from dataclasses import replace
 from types import SimpleNamespace
 from unittest.mock import MagicMock
 
 import pytest
 import torch
 from vllm.v1.core.block_pool import BlockPool
+from vllm.v1.core.kv_cache_utils import generate_scheduler_kv_cache_config
 from vllm.v1.core.single_type_kv_cache_manager import (
     FullAttentionManager,
     SlidingWindowManager,
@@ -30,6 +32,8 @@ from vllm_ascend.patch.platform.patch_kv_cache_coordinator import (
 )
 from vllm_ascend.patch.platform.patch_kv_cache_utils import (
     _ascend_resolve_kv_cache_block_sizes,
+    _get_kimi_k3_dspark_mixed_kv_cache_groups,
+    _get_kv_cache_config_deepseek_v4,
     group_and_unify_kv_cache_specs,
 )
 from vllm_ascend.patch.platform.patch_mamba_manager import AscendMambaManager
@@ -62,6 +66,65 @@ def _make_hybrid_kv_cache_config(
             KVCacheGroupSpec(layer_names=["mamba"], kv_cache_spec=mamba_spec),
         ],
     )
+
+
+def _make_kimi_k3_dspark_kv_cache_specs(
+    *,
+    block_size: int = 384,
+    page_size: int = 488448,
+    target_layer_count: int = 24,
+    draft_layer_count: int = 5,
+    mamba_layer_count: int = 69,
+    draft_uses_mla: bool = False,
+) -> dict:
+    target_mla_spec = MLAAttentionSpec(
+        block_size=block_size,
+        num_kv_heads=1,
+        head_size=576,
+        dtype=torch.bfloat16,
+        page_size_padded=page_size,
+        cache_dtype_str="auto",
+    )
+    if draft_uses_mla:
+        draft_attention_spec = MLAAttentionSpec(
+            block_size=block_size,
+            num_kv_heads=1,
+            head_size=576,
+            dtype=torch.bfloat16,
+            page_size_padded=page_size,
+            cache_dtype_str="auto",
+            non_causal_multi_token_decode=True,
+        )
+    else:
+        draft_attention_spec = FullAttentionSpec(
+            block_size=block_size,
+            num_kv_heads=1,
+            head_size=64,
+            dtype=torch.bfloat16,
+            page_size_padded=page_size,
+        )
+    mamba_spec = MambaSpec(
+        block_size=block_size,
+        shapes=((10, 2304), (6, 128, 128)),
+        dtypes=(torch.bfloat16, torch.float32),
+        page_size_padded=page_size,
+        mamba_cache_mode="align",
+        num_speculative_blocks=7,
+    )
+    specs = {
+        f"language_model.model.layers.{layer_idx}.self_attn.attn": target_mla_spec
+        for layer_idx in range(target_layer_count)
+    }
+    specs.update(
+        {
+            f"model.layers.{layer_idx}.self_attn.attn": draft_attention_spec
+            for layer_idx in range(93, 93 + draft_layer_count)
+        }
+    )
+    specs.update(
+        {f"language_model.model.layers.{layer_idx}.self_attn": mamba_spec for layer_idx in range(mamba_layer_count)}
+    )
+    return specs
 
 
 def _make_deepseek_v4_kv_cache_config() -> KVCacheConfig:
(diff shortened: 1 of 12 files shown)


maoxx241:

#### Kimi K3 grouping rule

patch_kv_cache_utils.py applies the optimized grouping only when the complete compatible Kimi signature is present: causal target full-attention layers, DSpark draft full-attention layers, aligned speculative Mamba layers, and compatible runtime block/page sizes.

Target and draft attention layers share one scheduler ownership group while retaining their exact inner KVCacheSpec and physical tensors. The recurrent group count is ceil(mamba_layers / attention_layers), and recurrent layers are distributed evenly. Partial or incompatible signatures use vLLM generic grouping.

For the current layout this yields one attention group with 24 target plus 5 draft layers and three recurrent groups with 23 layers each, with group widths 29/23/23/23.

#### Exact metadata-builder count

Builders are keyed by backend and exact inner spec, not only Python type. GQA DSpark uses target MLA plus draft GQA plus three Mamba builders; MLA DSpark uses causal target MLA plus non-causal multi-token draft MLA plus three Mamba builders. Both modes therefore have four scheduler groups and five exact metadata builders.

#### Logical versus physical cache layout

core/kv_cache_interface.py keeps scheduler block size separate from physical storage block size. AscendMLAAttentionSpec.merge delegates common fields to upstream, preserving compress_ratio, model version, quantization, alignment, and block-stride semantics, then reattaches Ascend layout fields. block_table.py and model_runner_v1.py reshape physical pages into the logical kernel block shape and pass each group kernel block size to DSpark.

#### Mamba grouping and capacity

Worker configs use the upstream UniformTypeKVCacheSpecs wrapper so each recurrent layer physical layout remains available; no K3-specific wrapper class is introduced. The Mamba helpers unwrap recurrent specs only when discovering state groups. Group capacity continues to use the largest recurrent requirement, including Prefix Cache and speculative lookahead.

#### Coverage

Focused tests cover TP8/TP16 runtime sizes, GQA/MLA drafts, exact builder counts, Mamba discovery, block-table remapping, compressed MLA page size, DeepSeek compression fields, Prefix Cache capacity, generic fallback signatures, and DSpark per-group slot mappings.

vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
```diff
@@ -590,6 +590,7 @@ def get_kv_cache_coordinator(  # type: ignore[misc]
         # compatibility; platform validation guarantees that it is one.
         del pcp_world_size
         token_budget = _select_kv_token_budget(max_model_len, max_in_flight_tokens, max_num_batched_tokens)
+        prefill_lookahead = 0 if num_prefill_lookahead is None else num_prefill_lookahead
```
Angazenn: what do we need to set prefill_lookahead to 0?

vllm_ascend/core/kv_cache_interface.py
```diff
@@ -53,7 +53,7 @@ def storage_block_size(self) -> int:
         return self.block_size // self.compress_ratio
 
     @property
-    def page_size_bytes(self) -> int:
+    def real_page_size_bytes(self) -> int:
```
Angazenn, line 56: where do we need `real_page_size_bytes`? Will this affect other codes that require `spec.page_size_bytes`?

vllm_ascend/core/kv_cache_interface.py
```diff
@@ -65,46 +65,42 @@ def merge(cls, specs: list[Self]) -> Self:
         assert all(isinstance(spec, MLAAttentionSpec) for spec in specs), (
             "All attention layers in the same KV cache group must be MLAAttentionSpec."
         )
-        layout_set = {
+        ascend_layouts = {
```
Angazenn, line 68: please also check whether these changes are necessary for `merge`. At least I think we should preserve `block_size` for `merge`. otherwise it might break dsv4.

tests/ut/patch/platform/test_prefix_cache_cp_patches.py
```diff
@@ -64,6 +68,65 @@ def _make_hybrid_kv_cache_config(
...
+def _make_kimi_k3_dspark_kv_cache_specs(
+    *,
+    block_size: int = 384,
+    page_size: int = 488448,
+    target_layer_count: int = 24,
+    draft_layer_count: int = 5,
+    mamba_layer_count: int = 69,
+    draft_uses_mla: bool = False,
+) -> dict:
+    target_mla_spec = MLAAttentionSpec(
```
Angazenn: use `AscendMLAAttentionSpec` in ut

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3667,10 +3666,7 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
         # NOTE(cmq): initialize_attn_backend must before using self.attn_groups
         self.initialize_attn_backend(kv_cache_config)
         self.use_hybrid_blocks = len(self.attn_groups) > 1
-        # NOTE: Currently, we determine whether we need `num_accepted_tokens` through `MambaSpec`.
-        self.need_accepted_tokens = any(
-            [isinstance(attn_group[0].kv_cache_spec, MambaSpec) for attn_group in self.attn_groups]
-        )
+        self.need_accepted_tokens = kv_cache_config.has_mamba_layers
```
zhenwenqi2024, line 3671: I remember this parameter already existed, right?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3693,10 +3689,26 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
                 AscendEagleProposer | AscendDflashProposer | AscendDSparkProposer | AscendDraftModelProposer,
             )
-            block_size = (self.kernel_block_sizes[0] if isinstance(
-                self.kernel_block_sizes, list) else self.kernel_block_sizes)
-            self.drafter.initialize_attn_backend(kv_cache_config, block_size)
-        
+            if isinstance(self.drafter, AscendDSparkProposer):
+                if isinstance(self.kernel_block_sizes, list):
+                    draft_kernel_block_sizes = [
+                        int(sizes[0] if isinstance(sizes, (list, tuple)) else sizes)
```
zhenwenqi2024: Maybe this is better ?
if isinstance(self.drafter, AscendDSparkProposer):
    sizes = self.kernel_block_sizes
    if not isinstance(sizes, list):
        sizes = [sizes]
    draft_kernel_block_sizes= [int(s[0] if isinstance(s, (list, tuple)) else s) for s in sizes]
else:
    draft_kernel_block_sizes= self.kernel_block_sizes[0] if isinstance(self.kernel_block_sizes, list) else self.kernel_block_sizes

self.drafter.initialize_attn_backend(kv_cache_config, draft_kernel_block_sizes)

vllm_ascend/patch/platform/patch_kv_cache_utils.py
```diff
@@ -56,6 +74,133 @@ def _ascend_resolve_kv_cache_block_sizes(
     return _orig_resolve_kv_cache_block_sizes(kv_cache_config, vllm_config)
 
 
+def _is_kimi_k3_target_attention_layer(layer_name: str, spec: KVCacheSpec) -> bool:
```
Angazenn: These helper function (`_is_kimi_k3_xxx_layer`) are over-defensive. Since they are only used for Kimi-K3 now, I think draft/target layers can be distinguished by layer prefix, while mamba can be selected by spec (MambaSpec)

vllm_ascend/patch/platform/patch_kv_cache_utils.py
```diff
@@ -56,6 +74,133 @@ def _ascend_resolve_kv_cache_block_sizes(
...
+            layer_names=list(mixed_attention_specs),
+            kv_cache_spec=mixed_attention_spec,
+        )
+    ]
+    mamba_layer_names = list(mamba_specs)
+    mamba_group_count = cdiv(len(mamba_layer_names), len(mixed_attention_specs))
+    for group_idx in range(mamba_group_count):
+        layer_names = mamba_layer_names[group_idx::mamba_group_count]
+        group_specs = {name: mamba_specs[name] for name in layer_names}
+        uniform_mamba_spec = _MambaUniformKVCacheSpecs.from_specs(group_specs)
```
Angazenn: From my perspective, we do not need to wrap mamba spec into such a `_MambaUniformKVCacheSpecs`. `UniformKVCacheSpecs` is mainly used for specs that cannot be merged but can use the same block table. Since MambaSpec usually forms one single group, it is not necessary to introduce `_MambaUniformKVCacheSpecs`.

vllm_ascend/patch/worker/patch_mamba_utils.py
```diff
@@ -23,6 +27,31 @@ def _can_launch_triton_batch_memcpy() -> bool:
     return not is_310p()
 
 
+def _get_mamba_groups(
```
Angazenn, line 30: ditto. I think we can remove it as we do not actually need `_MambaUniformKVCacheSpecs`

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3693,10 +3689,26 @@ def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
...
+                        for sizes in self.kernel_block_sizes
+                    ]
+                else:
+                    draft_kernel_block_sizes = [int(self.kernel_block_sizes)]
+                self.drafter.initialize_attn_backend(
+                    kv_cache_config,
+                    draft_kernel_block_sizes,
```
Angazenn: seems to rely on dspark support?

vllm_ascend/core/kv_cache_interface.py
maoxx241: real_page_size_bytes describes the actual Ascend page payload, including the optional scale storage. Upstream unpadded_page_size_bytes and page_size_bytes consume it, so allocation is intentionally affected. I kept this override and added direct page-size coverage in ccd38a2b0.

vllm_ascend/worker/model_runner_v1.py
maoxx241: Applied this simplification in ccd38a2b0 and kept a single initialize_attn_backend call. Only AscendDSparkProposer receives the per-group list; other drafters keep the scalar block size.

vllm_ascend/patch/platform/patch_kv_cache_utils.py
maoxx241: Removed the custom _MambaUniformKVCacheSpecs subclass in ccd38a2b0 and use the upstream UniformTypeKVCacheSpecs. A standard wrapper is still required because each K3 Mamba group carries 23 per-layer specs that share one block table and participate in the packed tensor tuple layout.

tests/ut/patch/platform/test_prefix_cache_cp_patches.py
maoxx241: Updated the K3 KV grouping tests to construct AscendMLAAttentionSpec in ccd38a2b0.

vllm_ascend/patch/platform/patch_kv_cache_utils.py
maoxx241: Simplified the classifiers in ccd38a2b0: target and draft attention are selected by their layer prefixes and attention spec type, while recurrent layers are selected by MambaSpec.

vllm_ascend/worker/model_runner_v1.py
maoxx241: Yes, the per-group block-size list is DSpark-specific. ccd38a2b0 makes that branch explicit; non-DSpark proposers still receive the original scalar block size.

vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
maoxx241: Matched the upstream contract in ccd38a2b0: num_prefill_lookahead is now int with default 0 and is forwarded directly. The extra None-to-zero normalization was removed.

vllm_ascend/worker/model_runner_v1.py
maoxx241: Yes, need_accepted_tokens already existed. This change only switches detection from scanning the first visible attention spec to kv_cache_config.has_mamba_layers, because the packed worker layout keeps Mamba specs inside UniformType groups.

vllm_ascend/patch/worker/patch_mamba_utils.py
maoxx241: The custom wrapper is gone, but the standard UniformType wrapper remains for the 23-layer Mamba groups. Upstream get_mamba_groups only recognizes a direct MambaSpec, so this small unwrapping patch is still required for worker copy buffers and contexts.

vllm_ascend/core/kv_cache_interface.py
maoxx241: Checked the merge path. super().merge preserves block_size, compress_ratio, model_version, cache dtype, and block-stride indexing; the final replace only adds Ascend layout fields. Added a DeepSeek-V4-shaped merge test in ccd38a2b0.

vllm_ascend/patch/platform/patch_kv_cache_utils.py
```diff
@@ -56,6 +64,107 @@ def _ascend_resolve_kv_cache_block_sizes(
...
+            layer_names=list(mixed_attention_specs),
+            kv_cache_spec=mixed_attention_spec,
+        )
+    ]
+    mamba_layer_names = list(mamba_specs)
+    mamba_group_count = cdiv(len(mamba_layer_names), len(mixed_attention_specs))
+    for group_idx in range(mamba_group_count):
+        layer_names = mamba_layer_names[group_idx::mamba_group_count]
+        group_specs = {name: mamba_specs[name] for name in layer_names}
+        uniform_mamba_spec = UniformTypeKVCacheSpecs.from_specs(group_specs)
```
Angazenn, line 156: We should at least leave comments here for why we must use UniformTypeKVCacheSpecs to wrap mamba spec. It seems we finally walks into `_get_kv_cache_config_deepseek_v4` that is patched in vLLM-Ascend. This looks quite tricky. I think it might be better to patch the whole `get_kv_cache_config_from_groups` so that we can have a more general group function, and do not require "all `UniformTypeKVCacheSpecs`"

Angazenn: Introducing `UniformTypeKVCacheSpecs` for `MambaSpec` also result in additional patches like ` _get_mamba_groups`. My concern is whether these further changes affect dsv4 and Qwen3.x .

maoxx241:

PR #14768, opened 2026-08-22
[v0.26.0rc][Feature][SpecDecode] Support Kimi K3 GQA and MLA DSpark

maoxx241:

This PR supports Kimi K3 GQA DSpark and the newer causal MLA Block5 DSpark checkpoint on `releases/v0.26.0rc`.

KV-cache grouping for GQA DSpark:

- Group the 24 target MLA layers and 5 draft GQA layers into one `UniformType` scheduler group because they share full-sequence block ownership.
- Split the 69 align-mode Mamba layers into three groups of 23.
- The resulting scheduler layout is `[29, 23, 23, 23]`; model-runner metadata is built for MLA, GQA, and the three Mamba groups instead of the generic approximately 20-group layout.
- Derive block and page sizes from the matched cache specs, so the implementation is not limited to TP16 or a hard-coded page size.

QuaRot support for GQA DSpark:

- Load the descriptor-selected global rotation before the Qwen3 DSpark early return.
- Keep the existing Qwen3 loader responsible for FC anti-rotation.
- Give the draft unrotated embedding and LM-head copies instead of aliasing the rotated target tensors.

New MLA Block5 checkpoint support:

- Read draft-attention causality from the checkpoint's `dflash_config.causal`, falling back to `full_attention_causal`; legacy checkpoints that omit both fields remain bidirectional.
- Preserve explicit speculative-decoding graph sizes when every size is already aligned to both the speculative query length and TP size. This allows the intended `[48, 96]` sizes for Block5 + TP16 without changing vLLM's generic graph-size resolver.

This PR does not add a runtime restriction between checkpoint `block_size` and `num_speculative_tokens`, and it does not bring back the four-plane physical KV layout.

No API change. Matching Kimi K3 GQA DSpark deployments use the mixed scheduler groups, QuaRot GQA drafts use correctly aligned shared boundaries, and causal MLA DSpark checkpoints use their serialized attention mode. Other model signatures continue through the existing generic paths.

- 107 focused unit tests passed on the v0.26 environment, covering KV grouping, Mamba grouping, block-table updates, DSpark proposer behavior, QuaRot sharing, explicit graph alignment, and MLA causality.
- `pre-commit` passed for all changed files.
- Four-node full-expert GQA and MLA Block5 validation is in progress. Single-node 16-expert runs are startup/function smoke only and are not used as accuracy evidence.

- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/models/test_kimi_k3_dspark.py
+++ b/tests/ut/models/test_kimi_k3_dspark.py
@@ -4,9 +4,21 @@ from vllm_ascend.models.kimi_k3_dspark import K3DSparkForCausalLM, K3DSparkModel
 
 
 def test_k3_mla_draft_reports_non_causal_attention_per_layer():
-    draft_model = SimpleNamespace(layers=[object(), object(), object()])
+    draft_model = SimpleNamespace(config=SimpleNamespace(), layers=[object(), object(), object()])
 
     assert K3DSparkModel.get_draft_attn_causal(draft_model) == [False, False, False]
+
+
+def test_k3_mla_draft_uses_checkpoint_attention_causality():
+    draft_model = SimpleNamespace(
+        config=SimpleNamespace(
+            full_attention_causal=True,
+            dflash_config={"causal": True},
+        ),
+        layers=[object(), object(), object()],
+    )
+
+    assert K3DSparkModel.get_draft_attn_causal(draft_model) == [True, True, True]
 
 
 def test_k3_mla_causal_lm_forwards_attention_causality():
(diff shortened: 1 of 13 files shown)


maoxx241:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

maoxx241:
> Superseded: the commands and results previously shown here used a single-node 16-expert checkpoint. That checkpoint is suitable only for startup/function smoke and is not valid accuracy evidence. This comment will be replaced with the four-node full-expert GQA/MLA launch and test commands after validation.

PR #14774, opened 2026-08-22
[BugFix][Attention] Use physical block size for C8 NZ cache writes

EricLingRui:

Hybrid models may use a scheduler block size that differs from the physical attention KV cache block size. For example, the scheduler/Mamba-aligned block size can be 1536 while the C8 attention cache tensor uses 128.

The C8 PA_NZ write path previously reshaped the cache with `cache_config.block_size`, while FIA read paths derive the block size from the cache tensor. This makes writes and reads use different layouts and corrupts C8 KV cache contents.

This PR derives the PA_NZ write block size from `self.key_cache.shape[1]` and adds a regression test covering scheduler block size 1536 with physical cache block size 128.

Yes. C8 KV cache inference remains correct when a hybrid model uses different logical scheduler and physical attention cache block sizes. No API changes are introduced.

- Added `test_c8_nz_cache_write_uses_physical_block_size`, which verifies the 5D PA_NZ cache view uses physical block size 128 even when the scheduler block size is 1536.
- `uvx ruff format --check vllm_ascend/attention/attention_v1.py tests/ut/attention/a2/test_attention_v1.py`
- `uvx ruff check vllm_ascend/attention/attention_v1.py tests/ut/attention/a2/test_attention_v1.py`
- `git diff --check`
- Manually verified on Ascend 910B3 with Qwen3.8-27B W8A8C8: TP1 and TP2 inference passed 1K/8K/32K/64K/128K prompt cases with 512/1024 generated tokens; TP1 also passed 32K/128K/256K marker retrieval and multimodal correctness checks.

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -2207,7 +2207,10 @@ class AscendC8AttentionBackendImpl(AscendAttentionBackendImpl):
             encoder_decoder = self.attn_type == AttentionType.ENCODER_DECODER
 
             # NZ write path: 5D view + npu_scatter_pa_kv_cache
-            block_size = self.vllm_config.cache_config.block_size
+            # Hybrid models can use a scheduler block size that differs from
+            # the physical attention cache block size. PA_NZ packing must use
+            # the physical cache dimension so the write layout matches FIA.
+            block_size = self.key_cache.shape[1]
             k_cache_layer = self._nz_5d_view(self.key_cache, block_size)
             v_cache_layer = self._nz_5d_view(self.value_cache, block_size)
 
--- a/tests/ut/attention/test_attention_v1.py
+++ b/tests/ut/attention/test_attention_v1.py
@@ -588,6 +588,29 @@ class TestAscendAttentionBackendImpl(TestBase):
         self.assertIs(returned[2], value)
         self.assertIs(returned[3], output)
 
+    @patch("vllm_ascend.attention.attention_v1.notify_kv_cache_written")
+    @patch("torch_npu.npu_scatter_pa_kv_cache", create=True)
+    def test_c8_nz_cache_write_uses_physical_block_size(self, mock_scatter_pa_kv_cache, _mock_notify):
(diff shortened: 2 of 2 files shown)


vllm_ascend/attention/attention_v1.py
```diff
@@ -2215,7 +2215,10 @@ def _reshape_and_cache(
             encoder_decoder = self.attn_type == AttentionType.ENCODER_DECODER
 
             # NZ write path: 5D view + npu_scatter_pa_kv_cache
-            block_size = self.vllm_config.cache_config.block_size
+            # Hybrid models can use a scheduler block size that differs from
+            # the physical attention cache block size. PA_NZ packing must use
+            # the physical cache dimension so the write layout matches FIA.
+            block_size = self.key_cache.shape[1]
```
nu11ptr7, line 2213: `_nz_5d_view` treats the cache as `[num_blocks, num_kv_heads, block_size, head_size]`, where dim 1 is the head count - so `key_cache.shape[1]` here reads heads, not the physical block size. If that's the layout on this path, the write layout breaks whenever heads != block size. Maybe take the block size from the cache spec instead.

EricLingRui: Thanks for checking this. On this backend the input cache shape is explicitly defined by `AscendAttentionBackend.get_kv_cache_shape` as `(2, num_blocks, block_size, num_kv_heads, head_size)`. `_reshape_and_cache` stores `kv_cache[0]` in `self.key_cache`, so its logical shape is `(num_blocks, block_size, num_kv_heads, head_size)` and dimension 1 is the physical block size. `_nz_5d_view` documents the resulting 5D NZ view, not the input layout. The existing C8 decode and chunked-prefill read paths also use `num_block, block_size, _, _ = self.key_cache.shape`. The regression test uses an input cache of `(24, 128, 8, 64)` and verifies the NZ view is `(24, 8, 2, 128, 32)`. Therefore `self.key_cache.shape[1]` is 128 here, rather than the head count.

PR #14776, opened 2026-08-22
[Feature]: add hybrid AscendStore transfer planner

lizy124:
 Adds a hybrid transfer supervisor/planner for AscendStore KV transfer, with safer queue accounting, startup failure detection, stale-task cleanup, and graceful shutdown. This improves transfer-process reliability and avoids hangs during shutdown or worker failures.

No user-facing API or configuration changes. The changes are internal to AscendStore KV transfer management.

- Python compileall: passed

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/transfer_planner.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/transfer_planner.py
@@ -1,0 +1,304 @@
+from __future__ import annotations
+
+import multiprocessing as mp
+import queue
+import threading
+from collections.abc import Callable, Iterable
+from typing import Any
+
+
+class PlannerError(RuntimeError):
+    """Raised when the CPU planning process cannot return a plan."""
+
+
+def _hash_to_string(value: bytes | str) -> str:
+    return value.hex() if isinstance(value, bytes) else value
+
+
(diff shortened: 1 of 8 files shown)


vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/transfer_planner.py
```diff
@@ -0,0 +1,304 @@
...
+                hash_block_size = int(request["hash_block_size"])
+                token_len = int(request["token_len"])
+                mask_num = int(request.get("mask_num", 0))
+                block_ids = request.get("block_ids")
+                allowed = request.get("allowed_chunk_indices")
+                allowed_set = set(allowed) if allowed is not None else None
+                factor = max(block_size // max(hash_block_size, 1), 1)
+                grouped_hashes = [hashes[(idx + 1) * factor - 1] for idx in range(len(hashes) // factor)]
+                num_chunks = min(len(grouped_hashes), (token_len + block_size - 1) // block_size)
+                block_id_offset = max(num_chunks - len(block_ids), 0) if block_ids is not None else 0
```
nu11ptr7, line 38: When `block_ids` is shorter than `num_chunks`, `block_id_offset` skips the *first* chunks (`block_index = chunk_index - block_id_offset`). If the ids list aligns with the tail that's right, but if it's a prefix or sparse list the wrong chunks get dropped silently - maybe add a comment or assert on which end the ids align.

PR #14783, opened 2026-08-23
[Performance] skip per-layer indexer cache for GLM-5.1 IndexCache S layers

### LINKED ISSUE #15866: [Performance]: GLM-5.1 IndexCache S layers allocate/write unused per-layer indexer cache
﻿### Your current environment

- Model: GLM-5.1
- Feature: IndexCache
- Branch: main / v0.26.0rc / v0.27.1rc

---

### 🐛 Describe the issue

GLM-5.1 IndexCache keeps an `Indexer` module on static S layers for weight-loading compatibility, but those layers reuse another layer's top-k indices at runtime. Previously they still allocated and wrote a per-layer indexer KV cache even though the computed indexer keys were never consumed.

This causes unnecessary per-layer memory usage and extra NPU compute/scatter work.

### Expected behavior

Static S layers should behave like GLM-5.2 shared-indexer layers at runtime:

- Do not allocate a per-layer indexer KV cache.
- Do not write per-layer indexer cache data.
- Skip indexer pre-processing for static S layers.
- Keep MTP/nextn layers on the existing per-layer indexer path.

### Actual behavior / current limitation

- Static S layers still allocate and write a per-layer indexer KV cache.
- Indexer pre-processing is still performed even though the computed indexer keys are never consumed.
- This increases memory usage and causes unnecessary NPU compute/scatter work.

### Proposed change

- Add `_is_mtp_layer()` to keep MTP/nextn layers on the existing per-layer indexer path.
- Add `runtime_has_indexer` to distinguish "has an Indexer module for weight loading" from "actually computes/owns a runtime indexer cache".
- Skip indexer pre-processing and indexer KV cache writes for static S layers.
- Skip indexer KV cache allocation in `NPUModelRunner.get_kv_cache_spec()` for those layers.

### Related PRs

- Main: https://github.com/vllm-project/vllm-ascend/pull/14783
- v0.26.0rc: https://github.com/vllm-project/vllm-ascend/pull/15756
- v0.27.1rc: https://github.com/vllm-project/vllm-ascend/pull/15758

lijiahang226:

GLM-5.1 IndexCache retains an `Indexer` module on static S layers for weight loading, although those layers reuse another layer's top-k indices. They previously allocated and wrote an indexer KV cache whose keys were never consumed.

`runtime_has_indexer` distinguishes the weight-loading module from runtime cache ownership. Static S layers skip indexer preprocessing, cache writes and cache allocation in both model runner versions. MTP/nextn layers retain their indexer cache and writes even when they reuse top-k indices. GLM-5.2 shared-indexer behavior is preserved.

The cache composition path also preserves NoPE latent-cache normalization, including removal of an empty RoPE view. Regression coverage checks this together with native/fused preprocessing, independent SFA/LI C8 settings, and MRV2 DCP indexer replication.

Fixes #15866

Expected lower indexer cache memory use and less redundant indexer work for GLM-5.1 IndexCache deployments. No API or configuration changes. MTP keeps its existing cache behavior.

- Local isolated control-flow validation: 78 tests and 6 subtests passed using actual source methods, CPU tensors and mocked framework/kernel boundaries. This checks static S/MTP ownership, SFA cache composition and forward dispatch, and MRV1/MRV2 cache-spec decisions; it does not replace full module-import or NPU integration tests.
- Regression coverage: `tests/ut/attention/test_sfa_v1.py`, `tests/ut/attention/test_sfa_cp.py`, `tests/ut/test_ascend_config.py`, `tests/ut/worker/test_model_runner_v1.py`, and `tests/ut/worker/test_attn_utils_v2.py`. Full UT collection in the local Windows environment is blocked by a missing upstream `regex` dependency.
- Ruff check/format and `git diff --check` passed for the changed files. `bash format.sh ci` was attempted: repository-wide Ruff, codespell, typos and several source checks passed, but the complete run was not green on Windows due to unavailable Unix tools and hook encoding issues.
- NPU correctness and before/after memory/latency measurements have not been rerun because the remote validation host timed out.
- Nightly test removal (`13753accb`): `git diff --check` passed and no references to the removed test remain. `bash format.sh ci` could not start in this environment because `pre-commit` is unavailable.


lijiahang226:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/attention/sfa_v1.py
```diff
@@ -261,6 +262,19 @@ def _get_config_bool(configs: tuple[Any, ...], attr: str) -> bool:
     return False
 
 
+def _is_mtp_layer(hf_config: Any, layer_name: str | None) -> bool:
```
pisceskkk: please check if there is same functions in `vllm_ascend.utils` or move this function into `vllm_ascend.utils`.

lijiahang226:
Addressed in f24e4af90.

1. Moved the helper to `vllm_ascend/utils.py` as the public `is_mtp_layer(hf_config, layer_name)`, next to `parse_layer_idx` which it builds on. `vllm_ascend/core/kv_cache_placement.py` already imports from `utils`, so `utils` is the side of the dependency that keeps the graph acyclic; the reverse move would have closed a `utils -> core -> utils` cycle.
2. `sfa_v1.py` no longer defines it locally — it imports the shared helper. `parse_layer_idx` was the only local use, so that import was dropped too.
3. There was a second, pre-existing copy of the same predicate: `patch/worker/patch_deepseek_v2.py` open-coded `layer_id >= num_hidden_layers` twice (in `_should_skip_indexer_init` and in the runtime `skip_topk` mask). Both now call the shared helper, so this PR does not leave a third copy of the MTP decision behind.
4. Unit tests moved to `tests/ut/test_utils.py` (next to the function's new home) and widened with the boundary cases: last backbone layer, `num_hidden_layers` that is not an `int`, and a `None` layer name. The three original cases and the `_should_skip_indexer_init` MTP test are preserved unchanged.

On `find_mtp_layers` in `vllm_ascend/core/kv_cache_placement.py`: I did not reuse it here. It is not a layer predicate but a set-over-names API that needs a full `vllm_config` and is gated on `speculative_config.method == mtp`, and it uses the bounded range `[num_hidden_layers, num_hidden_layers + num_nextn_predict_layers)`. The SFA ownership decision runs before a speculative config is guaranteed, and the `.mtp.` naming branch must be caught too, so the two predicates genuinely differ. Happy to unify them on the bounded form in a follow-up if you prefer.

I also dropped stale-base noise while rebasing onto current `main` (`nightly_config.yaml` and `tests/e2e/pull_request/four_card/context_parallel/test_deepseek_v4.py` were reverting newer upstream changes); the diff is now 13 files, all on-topic.

vllm_ascend/attention/sfa_v1.py
lijiahang226: That makes sense. I haved moved this function into `vllm_ascend.utils`.

lijiahang226:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

weijinqian0 (dismissed):

linfeng-yuan (dismissed):

linfeng-yuan:
Please split your nightly test to another PR~

lijiahang226:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

linfeng-yuan (approved):

PR #14794, opened 2026-08-24
[Test][Feature] Add Qwen3.5-27B multi-LoRA MTP eager serving E2E

Agoni-02:
This PR adds an end-to-end (E2E) test for Qwen3.5-27B serving with multi-LoRA and built-in Multi-Token Prediction (MTP) in eager mode (without ACL graph). It covers the serving path that preloads two named LoRA modules and ensures that the LoRA routing works correctly by verifying the chat completions output of both adapters.

PR CI is Hugging Face offline (`HF_HUB_OFFLINE=1`) and uses ModelScope. There is no ModelScope-hosted Qwen3.5-27B PEFT adapter, so the test synthesizes two rank-8 adapters with the 27B MLP shapes. Local `MIX_LORA_A` / `MIX_LORA_B` still override that path.

No.

Added `tests/e2e/pull_request/four_card/lora/test_qwen35_27b_multi_lora_mtp_eager.py` which starts `RemoteOpenAIServer` with TP=4, `--enforce-eager`, `qwen3_5_mtp`, fully-sharded LoRA, and two `--lora-modules` names, then checks `/v1/models` and `/v1/chat/completions` for `mix-lora-a` vs `mix-lora-b`.

- vLLM version: v0.27.1

### DIFF
--- a/.github/workflows/scripts/test_config.yaml
+++ b/.github/workflows/scripts/test_config.yaml
@@ -278,6 +278,7 @@
     - tests/ut/lora
     - tests/e2e/pull_request/one_card/lora
     - tests/e2e/pull_request/two_card/lora
+    - tests/e2e/pull_request/four_card/lora
 
 # CPU Weight Offloading
 - name: cpu_weight_offload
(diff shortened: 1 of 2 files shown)


voidvelocity (approved):
Add a test for e2e lora test to keep lora function correct.

zzzzwwjj (approved):

PR #14822, opened 2026-08-24
[Feature][MRV2] Support Eagle3 and DSpark with PP

LostFox11:

This PR extends the existing Model Runner V2 speculative-decoding pipeline-parallel path to support:

- MiniMax M3 with EAGLE3 + PP
- DeepSeek V4 with DSpark + PP

The implementation reuses the existing MRV2 `PPHandler`, request-state update, and sampled/draft-token broadcast flow. It adds only the model data and loader handling required by these methods:

- a typed PP transport utility for auxiliary hidden states;
- ordered auxiliary-hidden-state propagation across PP stages;
- receive-buffer construction for transported auxiliary tensors;
- complete local draft-model initialization on the last PP stage with draft `pipeline_parallel_size=1`;
- EAGLE-specific draft config isolation: EP and EPLB are disabled for the dense EAGLE draft, while other speculative methods retain the target configuration;
- DSpark loading support for target quantization inheritance and a local draft embedding when the target embedding is absent on the last PP stage.

The EAGLE3 guard is bypassed only for the registered MiniMax M3 architectures, and DSpark PP is enabled only for `DeepseekV4ForCausalLM`. Existing MTP and non-PP paths remain unchanged.

PIECEWISE graph mode is outside the scope of this PR. Eager mode is covered by the E2E test in this PR. MiniMax M3 was also manually validated with `FULL_DECODE_ONLY`; the unrelated ACLGraph backend-selection fix needed by that mode is tracked separately in #15405.

Yes. Model Runner V2 users can combine pipeline parallelism with MiniMax M3 EAGLE3 and DeepSeek V4 DSpark in the supported configurations above.

- Added an eight-card E2E accuracy test for DeepSeek V4 W8A8, DSpark, PP2/TP4/EP, eager mode, chunked prefill, and 5 draft tokens.
- Real-weight curl smoke tests completed with coherent deterministic output and repeated requests without hangs.
- `python3 -m py_compile` on changed Python files.
- `git diff --check`.

#### Manual GPQA-Diamond accuracy

AISBench evaluation used all 198 GPQA-Diamond samples (`b1ed2c`) with a 100% answer extraction rate.

| Model and speculative method | Parallelism | Execution mode | Accuracy |
| --- | --- | --- | ---: |
| MiniMax M3 W8A8 + EAGLE3 | PP2 / TP8 / EP | Target `FULL_DECODE_ONLY`, draft eager | 92.42% |
| DeepSeek V4 W8A8 + DSpark | TP8 / PP1 / EP (pure TP) | Eager | 85.35% (169/198) |
| DeepSeek V4 W8A8 + DSpark | PP2 / TP4 / EP | Eager | 83.84% (166/198) |

The DeepSeek V4 pure-TP baseline combines two shards: 85.00% on 100 samples and 85.71% on 98 samples. The PP result combines 85.00% on 100 samples and 82.65% on 98 samples. The two configurations differ by only three correct answers over the full 198-sample set, so no material accuracy regression was observed. This PR extends PP transport and draft initialization without changing target/draft model computation or verification semantics, and does not affect model accuracy in the supported modes.


vllm_ascend/patch/worker/patch_v2/patch_dspark.py
```diff
@@ -49,13 +57,35 @@ def _load_dspark_model_with_target_quant(target_model, vllm_config):
...
+    original_get_pp_group = dspark_utils.get_pp_group
+    original_should_share = eagle_utils._should_share
     if inherits_target_quant:
         model_utils.get_draft_quant_config = lambda _vllm_config: vllm_config.quant_config
+    if bypass_pp_guard:
+        # The upstream loader checks the live PP group independently of config.
+        single_rank_pp_group = SimpleNamespace(world_size=1)
+        dspark_utils.get_pp_group = lambda: single_rank_pp_group
+
+        def should_share(eagle, flag, draft, target):
```
wxsIcey, line 71: `should_share` should be a public method applicable to various speculative methods, please move it externally.

Regarding the speculative modifications, LGTM!

vllm_ascend/worker/v2/aclgraph_utils.py
```diff
@@ -62,9 +62,12 @@ def _get_graph_update_backend(
     for groups in attn_groups:
         for group in groups:
             backend = group.backend
-            if backend.get_impl_cls() is not None:
+            impl_cls = backend.get_impl_cls()
```
wxsIcey: Please revert this change.

vllm_ascend/patch/worker/patch_v2/patch_dspark.py
LostFox11: In the upstream VLLM, both eagle3 and dflash have skipped embed tokens shares in parallel pp. Currently, only dspark requires this modification.

vllm_ascend/worker/v2/aclgraph_utils.py
LostFox11: Done

freyfwt (approved):

vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
```diff
@@ -97,9 +97,16 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
 
     def _create_draft_vllm_config(self) -> VllmConfig:
         """Build the runtime config used while executing the draft model."""
+        parallel_config = replace(
+            self.vllm_config.parallel_config,
+            pipeline_parallel_size=1,
+            enable_expert_parallel=False,
+            enable_eplb=False,
```
freyfwt: Do not completely disable ep in the base class here. If eagle3 does not support ep, disable it separately in the eagle3 hierarchy.

xuchi-0808:
Hi @LostFox11, we hit the same MiniMax-M3 full-graph crash and extracted the `_get_graph_update_backend` selection change from this PR into a standalone bugfix: #15405.

One finding from live testing on A3 (M3 + MRV2 + FULL_DECODE_ONLY, TP4xPP2): the raise-on-miss variant still fails on a PP stage without full-attention layers — M3 keeps GQA layers only in 0-2, so the second stage (layers 30-59) has no backend owning `update_graph_params` at all, and the RuntimeError path triggers there. #15405 handles this by returning None and skipping the update for such stages (they legitimately have no replayed attention parameters to update).

If this matches what you saw, feel free to rebase on #15405 once it lands so this PR doesn't need to carry the hunk. Thanks for the original diagnosis!

vllm_ascend/patch/platform/patch_use_v2_model_runner.py
```diff
@@ -21,7 +23,17 @@ def _patched_use_v2_model_runner(self) -> bool:
     return False
 
 
+def _patched_get_unsupported_features(self) -> list[str]:
```
wxsIcey, line 26: The upstream startup mrv2 and vllm-ascend solutions are inconsistent. Don‘t change this file. Please revert it.

lcfenglinwan:
The accuracy of the DSV4 model is lower than expected. Please check whether this is due to feature introduction.

LostFox11:
CI follow-up: the A3 eight-card runner is offline and does not currently cache `Inferact/MiniMax-M3-EAGLE3-GQA`, so the MiniMax M3 E2E cannot start. This PR now declares the model in `.github/workflows/misc/model_dataset_list.json`. A maintainer with label permission please add `model-dataset-download` once to populate the A3 PVC; the author's token cannot add that label. The DSpark failure in the previous run was a test-parser issue (the model emitted the correct first `answer:71` and then continued generating examples); commit e690e4543 makes the check use the first formatted answer.

LostFox11:
Final result of the latest CI run: 25 checks passed. The DeepSeek V4 + DSpark PP accuracy E2E passed after the parser stabilization. The only real failure is MiniMax M3 startup because the offline A3 PVC does not contain `Inferact/MiniMax-M3-EAGLE3-GQA`; it fails in `SpeculativeConfig` before engine initialization. The cache list change is already in e690e4543. @lcfenglinwan, could you or another maintainer please add the `model-dataset-download` label? The downloader workflow has no manual-dispatch path and my token is not permitted to add this label. After the A3 download completes, rerunning the failed job should be sufficient.

vllm_ascend/patch/platform/patch_use_v2_model_runner.py
wxsIcey: My fault. We must change it because vllm don‘t support eagle + pp. LGTM!

linfeng-yuan (approved):

PR #14825, opened 2026-08-24
[BugFix][Quantization] Cache MX scale alg and warm it during AscendConfig init

lijiahang226:

`get_dynamic_mx_quant_scale_alg()` previously used `get_current_vllm_config()`
when no `vllm_config` was passed. This works during model loading, but fails
during ACL graph capture because the forward path executed for capture is not
wrapped in `set_current_vllm_config()`. As a result, starting a service on
Ascend A5 with graph capture enabled could crash with:

`Current vLLM config is not set.`

This PR fixes the crash and avoids recomputing the value on every forward:

1. Warm the MX scale algorithm cache during `init_ascend_config()` using the
   explicit `vllm_config`.
2. Cache the resolved `scale_alg` process-wide in `quantization/utils.py`.
3. `DeviceOperator.npu_dynamic_quant()` reuses the cached value during
   forward / graph capture instead of calling `get_dynamic_mx_quant_scale_alg()`
   on every invocation.
4. Reset the cache when `AscendConfig` is re-initialized, so a stale value
   from a previous model/config cannot be reused.

This keeps MiniMax M3 on Ascend A5 using `scale_alg=1` even during graph
capture, while avoiding the crash in `DeviceOperator.npu_dynamic_quant()`.

No user-facing API or configuration change. It fixes an internal crash during
ACL graph capture on Ascend A5 and preserves the existing MiniMax M3 MXFP8
behavior.

- Updated unit tests in `tests/ut/quantization/test_utils.py`:
  - verifies the resolved `scale_alg` is cached and reused across calls
  - verifies explicit-config and no-arg calls share the same cached value

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/quantization/utils.py
+++ b/vllm_ascend/quantization/utils.py
@@ -16,6 +16,7 @@
 #
 
 import json
+from functools import lru_cache
 from pathlib import Path
 
 import torch_npu  # noqa: F401
@@ -56,9 +57,11 @@ def get_dynamic_mx_quant_scale_alg(vllm_config=None) -> int:
             if scale_alg is not None:
                 return scale_alg
 
-        from vllm.config import get_current_vllm_config
+        from vllm.config import get_current_vllm_config_or_none
 
-        vllm_config = get_current_vllm_config()
+        vllm_config = get_current_vllm_config_or_none()
+        if vllm_config is None:
+            return 0
 
     model_config = vllm_config.model_config
     architectures = getattr(model_config, "architectures", None) or ()
@@ -72,6 +75,11 @@ def get_dynamic_mx_quant_scale_alg(vllm_config=None) -> int:
     ):
         return 1
     return 0
+
+
+@lru_cache(maxsize=1)
+def get_dynamic_mx_quant_scale_alg_cached() -> int:
+    return get_dynamic_mx_quant_scale_alg()
 
 
 def get_model_file(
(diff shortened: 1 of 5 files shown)


vllm_ascend/quantization/utils.py
```diff
@@ -49,9 +49,15 @@ def get_dynamic_mx_quant_scale_alg(vllm_config=None) -> int:
...
     if vllm_config is None:
-        from vllm.config import get_current_vllm_config
+        from vllm.config import get_current_vllm_config_or_none
 
-        vllm_config = get_current_vllm_config()
+        vllm_config = get_current_vllm_config_or_none()
+        if vllm_config is None:
+            from vllm_ascend.ascend_config import get_init_vllm_config
+
+            vllm_config = get_init_vllm_config()
```
nu11ptr7: During graph capture the current config is None, so this falls back to the init-time config - if the config was rebuilt or mutated since init, the scale alg silently comes from a stale one. Maybe cache the config object and assert identity.

vllm_ascend/quantization/utils.py
kunpengW-code, line 41: We should cache the return value of this method instead of regenerating it every time during forward computation. @HaoxinZong cc

vllm_ascend/quantization/utils.py
lijiahang226: Fixed: we now warm the cache in init_ascend_config() with the explicit config and reset it on re-init, so graph capture never falls back to a stale init-time config.

vllm_ascend/quantization/utils.py
lijiahang226: Fixed: get_dynamic_mx_quant_scale_alg() now caches the return value process-wide; forward/graph-capture paths reuse it instead of recomputing.

lijiahang226:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #14840, opened 2026-08-24
[Performance][Ops] Add Kimi K3 attention residual fusion

maoxx241:

- Adds the Triton attention-residual fusion used by Kimi K3.
- Preserves the upstream preallocated residual-buffer contract while specializing the launch shape for Ascend.
- Covers valid-block counts and residual layouts with a nightly operator test.

### Dependencies

- No child PR is required to review this operator in isolation.
- The Kimi K3 model adapter PR consumes this fusion.

- `git diff --check`
- Python bytecode compilation for changed implementation and test files
- Nightly operator coverage in `tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_kimi_k3_fusions.py`

No direct API change. Kimi K3 uses the fusion internally.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/ops/triton/kimi_k3/attention_residual.py
+++ b/vllm_ascend/ops/triton/kimi_k3/attention_residual.py
@@ -31,6 +31,7 @@ def _apply_attn_res_kernel(
     NUM_CORES: tl.constexpr,
     NB: tl.constexpr,
 ):
+    tl.static_assert(NB >= B + 1, "NB must include all block residuals and prefix_sum")
     block_size = (N - 1) // NUM_CORES + 1
     pid = tl.program_id(0)
     tok0 = pid * block_size
--- a/vllm_ascend/ops/triton/kimi_k3/attention_residual.py
+++ b/vllm_ascend/ops/triton/kimi_k3/attention_residual.py
@@ -1,0 +1,111 @@
+"""Fused Kimi K3 attention-residual mixture.
+
+This ports the validated v0.26 computation to vLLM 0.27's preallocated
+residual-buffer contract.
+"""
+
+import torch
+from vllm.triton_utils import tl, triton
+
+from vllm_ascend.ops.triton.triton_utils import (
+    get_vectorcore_num,
+    init_device_properties_triton,
+)
+
+
+@triton.jit
+def _apply_attn_res_kernel(
+    block_residual_ptr,
+    prefix_sum_ptr,
+    norm_w_ptr,
+    proj_w_ptr,
+    out_ptr,
+    N: tl.constexpr,
+    H: tl.constexpr,
+    B: tl.constexpr,
+    BLOCK_CAPACITY: tl.constexpr,
+    EPS: tl.constexpr,
+    NUM_CORES: tl.constexpr,
+    NB: tl.constexpr,
+):
+    block_size = (N - 1) // NUM_CORES + 1
+    pid = tl.program_id(0)
+    tok0 = pid * block_size
+    if tok0 >= N:
+        return
+    tok1 = tl.minimum(tok0 + block_size, N)
+
+    cols = tl.arange(0, H)
+    s_idx = tl.arange(0, NB)
+    block_residual_stride = BLOCK_CAPACITY * H
+
+    norm_w = tl.load(norm_w_ptr + cols).to(tl.float32)
+    proj_w = tl.load(proj_w_ptr + cols).to(tl.float32)
+    w = norm_w * proj_w
+
+    for tok in range(tok0, tok1):
+        scores = tl.full([NB], -float("inf"), dtype=tl.float32)
+        for s in range(B + 1):
+            if s < B:
+                v = tl.load(block_residual_ptr + tok * block_residual_stride + s * H + cols).to(tl.float32)
+            else:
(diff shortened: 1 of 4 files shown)


maoxx241:

`apply_attn_res()` computes Kimi K3's learned mixture of attention residual streams in one Triton kernel.

For each token:

1. read the `num_valid_blocks` residual streams from the preallocated `block_residual` buffer;
2. append `prefix_sum` as the final candidate stream;
3. RMS-normalize every candidate in FP32;
4. score each candidate with `norm.weight * proj.weight`;
5. apply a softmax across the valid candidates;
6. form the weighted residual sum and cast it back to the model dtype.

`block_residual.shape[1]` is treated as allocation capacity, while `num_valid_blocks` controls the streams participating in the computation. Unused preallocated slots are never scored or accumulated.

The launch uses one program per vector-core partition of the token dimension. Hidden-state columns and the next-power-of-two score vector are compile-time specialized, while the valid block count is passed as the model-level value.

The nightly regression compares the Triton result with the FP32 PyTorch formula for:

- a partial-capacity buffer, where valid blocks are fewer than allocated slots;
- the profile-run shape with the full residual capacity.

Both cases use the K3 hidden size and verify BF16 output within the expected tolerance.

vllm_ascend/ops/triton/kimi_k3/attention_residual.py
```diff
@@ -0,0 +1,111 @@
...
+        scores = tl.full([NB], -float("inf"), dtype=tl.float32)
+        for s in range(B + 1):
+            if s < B:
+                v = tl.load(block_residual_ptr + tok * block_residual_stride + s * H + cols).to(tl.float32)
+            else:
+                v = tl.load(prefix_sum_ptr + tok * H + cols).to(tl.float32)
+            ms = tl.sum(v * v) / H
+            rstd = tl.rsqrt(ms + EPS)
+            k = v * rstd
+            scores = tl.where(s_idx == s, tl.sum(k * w), scores)
```
nu11ptr7, line 60: The prefix stream is selected at index `s == B`, so `NB` must be at least B+1 - if a caller sets NB == B, the prefix bucket silently never matches and contributes zero weight. Worth an assert.

vllm_ascend/ops/triton/kimi_k3/attention_residual.py
ZT-AIA, line 1: Supplement the operator description file.

ZT-AIA (dismissed):

vllm_ascend/ops/triton/kimi_k3/attention_residual.py
maoxx241: Added vllm_ascend/ops/triton/kimi_k3/attention_residual.md in 0d5385109 with the formula, inputs, preallocated-buffer contract, stream-capacity invariant, and launch behavior.

vllm_ascend/ops/triton/kimi_k3/attention_residual.py
maoxx241: Added a compile-time tl.static_assert for NB >= B + 1 in 663c1332d. The launch side continues to derive NB as next_power_of_2(num_valid_blocks + 1).

vllm_ascend/ops/triton/kimi_k3/attention_residual.py
```diff
@@ -0,0 +1,111 @@
...
+        out,
+        N=num_tokens,
+        H=hidden_size,
+        B=num_valid_blocks,
+        BLOCK_CAPACITY=block_capacity,
+        EPS=eps,
+        NUM_CORES=num_vectorcore,
+        NB=num_streams,
+        multibuffer=True,
+    )
```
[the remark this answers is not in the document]
maoxx241: Thanks. I kept H=hidden_size rather than adding BLOCK_H=next_power_of_2(hidden_size). On Ascend Triton the exact constexpr extent is supported and is the intended specialization. Padding 7168 to 8192 would increase the vector width and require masking, adding about 14% extra lanes and register pressure without fixing a backend limitation.

vllm_ascend/ops/triton/kimi_k3/attention_residual.py
```diff
@@ -0,0 +1,111 @@
...
+        out = tl.zeros([H], dtype=tl.float32)
+        for s in range(B + 1):
+            if s < B:
+                v = tl.load(block_residual_ptr + tok * block_residual_stride + s * H + cols).to(tl.float32)
+            else:
+                v = tl.load(prefix_sum_ptr + tok * H + cols).to(tl.float32)
+            w_s = tl.sum(tl.where(s_idx == s, weights, 0.0))
+            out += w_s * v
+
+        tl.store(out_ptr + tok * H + cols, out.to(out_ptr.dtype.element_ty))
```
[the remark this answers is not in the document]
maoxx241: Thanks. No code change is needed for the Ascend Triton backend. It supports compile-time arange extents that are not powers of two; existing Ascend kernels use the same pattern with model dimensions. This kernel deliberately specializes H to the actual K3 hidden size, and the nightly numerical case uses H=7168. Padding the vector to 8192 is therefore unnecessary.

maoxx241:

PR #14841, opened 2026-08-24
[MRV2][BugFix] fix rejection sampler on mrv2 with upstream commits

SOMEONEUNSEEN:
This PR implements synthetic rejection sampling support on NPU (Ascend) within the speculative decoding framework. It adds the `SYNTHETIC_MODE` path in `_probabilistic_rejection_kernel` using a 1-element block reduction pattern to work around the lack of scalar `tl.rand` in NPU Triton. Additionally, it improves numerical stability in `_resample_kernel` by using `tldevice.log1p` and adds explicit `tl.int64` casts to several index variables.

Feedback: Two potential compilation issues on NPU were identified due to mixed-type operations (comparing or subtracting `tl.int32` and `tl.int64`). It is recommended to also cast `end_idx` to `tl.int64` in both `_resample_kernel` and `_probabilistic_rejection_kernel`.

Yes, it enables synthetic rejection sampling on NPU, which was previously unsupported and raised a `NotImplementedError`.

The changes should be verified using existing speculative decoding integration tests on NPU.

### DIFF
--- a/vllm_ascend/worker/v2/spec_decode/rejection_sampler_utils.py
+++ b/vllm_ascend/worker/v2/spec_decode/rejection_sampler_utils.py
@@ -51,6 +51,10 @@ def _npu_gumbel_block_argmax(
     vocab_size,
     APPLY_TEMPERATURE: tl.constexpr,
 ):
+    # Convert token_idx to int64 before pointer arithmetic so the offset
+    # does not overflow int32 at large token counts (matches the other
+    # kernels that cast tl.program_id(0) up front).
+    token_idx = token_idx.to(tl.int64)
     req_state_idx = tl.load(expanded_idx_mapping_ptr + token_idx).to(tl.int64)
     temp = tl.load(temp_ptr + req_state_idx).to(tl.float32)
     if temp != 0.0 and APPLY_TEMPERATURE:
--- a/vllm_ascend/worker/v2/spec_decode/rejection_sampler_utils.py
+++ b/vllm_ascend/worker/v2/spec_decode/rejection_sampler_utils.py
@@ -553,6 +553,10 @@ def rejection_sample(
         # kernel signatures receive valid pointers/strides. The kernels
         # will never read from it when HAS_DRAFT_LOGITS=False.
         draft_logits = target_logits.new_empty(1, 1, 1)
+    else:
+        # In some cases (e.g. MiMo v2.5 Pro + DFlash) the target model's
+        # vocab size is larger than the draft's due to padding.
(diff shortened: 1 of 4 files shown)


vllm_ascend/worker/v2/spec_decode/rejection_sampler_utils.py
```diff
@@ -46,7 +46,7 @@ def _npu_gumbel_block_argmax(
     vocab_size,
     APPLY_TEMPERATURE: tl.constexpr,
 ):
-    req_state_idx = tl.load(expanded_idx_mapping_ptr + token_idx)
+    req_state_idx = tl.load(expanded_idx_mapping_ptr + token_idx).to(tl.int64)
```
nu11ptr7, line 75: Here `expanded_idx_mapping_ptr + token_idx` uses `token_idx` as-is and only converts the loaded value after - the other kernels in this PR convert `program_id(0)` to int64 first because pointer arithmetic overflows at large offsets. Same fix probably needed here.

SOMEONEUNSEEN: Thanks, your suggestion is reasonable, and I have made the changes.

vllm_ascend/worker/v2/spec_decode/rejection_sampler_utils.py
drslark, line 514: Can our `rejection_sample` align to `vllm`?

vllm_ascend/worker/v2/spec_decode/rejection_sampler_utils.py
```diff
@@ -40,38 +45,74 @@ def _npu_gumbel_block_argmax(
...
-        # NPU: use tl.rand (float32) instead of tl_rand64 (float64 not supported)
-        r = tl.rand(gumbel_seed, block).to(tl.float32)
-        gumbel_noise = -tl.log(-tl.log(r + 1e-20) + 1e-20)
+        # NPU: tl.rand (fp32) clamped away from zero, matching upstream
+        # tl_rand32(includes_zero=False). Draw the large-noise tail (which
+        # decides the argmax winner) from u -> 0, where fp32 has fine
+        # resolution, instead of u -> 1 (see upstream gumbel.py for the
+        # full log1p rationale).
+        u = tl.rand(gumbel_seed, block).to(tl.float32)
+        u = tl.maximum(u, 4.6566127342e-10)
```
wxsIcey: Do not hardcode `4.6566127342e-10`, please use a const constant.

PR #14852, opened 2026-08-24
[Feature][Quantization] Serve native block-wise FP8 checkpoints for qwen3.8-27b-fp8

### LINKED ISSUE #14467: [Bug]: Serving native FP8 checkpoints fails with AttributeError: 'Qwen3_5Config' object has no attribute 'o_groups'
### Your current environment

Docker image: quay.io/ascend/vllm-ascend:qwen3.8-a5
Hardware: Ascend 950PR, single card
vLLM: 0.23.0 / vLLM Ascend: 0.23.0
Model: Qwen/Qwen3.8-27B-FP8 (official native FP8 checkpoint, local copy)

(EngineCore pid=122161) INFO 08-18 10:30:53 [model_runner_v1.py:3718] Starting to load model /workspace/user_data/Qwen3.8-27B-FP8...
(EngineCore pid=122161) INFO 08-18 10:30:53 [fp8_config.py:57] Using the vLLM Ascend fp8 Quantization now!
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195] EngineCore failed to start.
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195] Traceback (most recent call last):
...
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]   File "/vllm-workspace/vllm/vllm/model_executor/models/qwen3_5.py", line 578, in __init__
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]     self.visual = Qwen3_VisionTransformer(
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]   File "/vllm-workspace/vllm/vllm/model_executor/models/qwen3_vl.py", line 581, in __init__
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]     self.merger = Qwen3_VisionPatchMerger(
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]   File "/vllm-workspace/vllm/vllm/model_executor/models/qwen3_vl.py", line 490, in __init__
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]     self.linear_fc1 = ColumnParallelLinear(
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]   File "/vllm-workspace/vllm-ascend/vllm_ascend/ops/linear.py", line 407, in __init__
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]     AscendLinearBase.__init__(
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]   File "/vllm-workspace/vllm-ascend/vllm_ascend/ops/linear.py", line 137, in __init__
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]     self.quant_method = quant_config.get_quant_method(self, prefix=prefix)
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]   File "/vllm-workspace/vllm-ascend/vllm_ascend/quantization/fp8_config.py", line 125, in get_quant_method
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]     scheme = create_scheme_for_layer(self.quant_description, prefix, "ds_linear", self.packed_modules_mapping)
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]   File "/vllm-workspace/vllm-ascend/vllm_ascend/quantization/fp8_config.py", line 63, in create_scheme_for_layer
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]     return scheme_cls(quant_description)
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]   File "/vllm-workspace/vllm-ascend/vllm_ascend/quantization/methods/fp8.py", line 45, in __init__
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195]     self.n_groups = hf_config.o_groups
(EngineCore pid=122161) ERROR 08-18 10:30:54 [core.py:1195] AttributeError: 'Qwen3_5Config' object has no attribute 'o_groups'
(APIServer pid=121784) RuntimeError: Engine core initialization failed. See root cause above. Failed core proc(s): {}

Root cause
detect_quantization_method() returns FP8_METHOD for any checkpoint whose config.json declares quantization_config.quant_method == "fp8" (vllm_ascend/quantization/utils.py:137-140), and FP8_METHOD is listed in NPUPlatform.supported_quantization (vllm_ascend/platform.py:90-95). So the model is routed to AscendFp8Config.

That config, however, only implements the DeepSeek-V4 native FP8 layout. In the v0.23.0 image it dispatches every LinearBase to the ds_linear scheme, whose constructor reads DeepSeek-specific config fields:

# vllm_ascend/quantization/methods/fp8.py (main: lines 56-58)
self.n_groups = hf_config.o_groups
self.n_local_groups = self.n_groups // tp_size
self.o_lora_rank = hf_config.o_lora_rank
Qwen3_5Config has neither o_groups nor o_lora_rank, hence the AttributeError. On current main the same call path instead hits a bare raise NotImplementedError (vllm_ascend/quantization/fp8_config.py:34-40), which is equally opaque to users.

A second problem is visible at the crash site: the failing layer is visual.merger.linear_fc1, which this checkpoint keeps in BF16 and lists in ignored_layers / modules_to_not_convert. The FP8 path never consults that list, so unquantized layers are treated as quantized. AscendCompressedTensorsConfig handles this correctly and can serve as the reference (vllm_ascend/quantization/compressed_tensors_config.py:161-173, returning AscendUnquantizedLinearMethod() when no scheme matches).

### 🐛 Describe the bug

```markdown
### What happens

Serving an official native FP8 checkpoint (`Qwen/Qwen3.8-27B-FP8`, architecture
`Qwen3_5ForConditionalGeneration`) on Ascend 950 fails with an `AttributeError`
during model construction, before any weight is loaded.

### Reproduce

```shell
vllm serve /workspace/user_data/Qwen3.8-27B-FP8 \
    --host 0.0.0.0 \
    --port 8000 \
    --tensor-parallel-size 1 \
    --served-model-name qwen3.8 \
    --max-num-seqs 32 \
    --max-model-len 131072 \
    --trust-remote-code \
    --enable-prefix-caching \
    --gpu-memory-utilization 0.85 \
    --speculative-config '{"method": "qwen3_5_mtp", "num_speculative_tokens": 3, "enforce_eager": true}' \
    --compilation-config '{"cudagraph_mode":"FULL_DECODE_ONLY"}'

### PR DESCRIPTION
**Pull request by yiminghub2024**

Native HF checkpoints with `quant_method: "fp8"` and `weight_block_size` (e.g. Qwen/Qwen3.8-27B-FP8) currently hit `AscendFp8Config` and either raise `NotImplementedError` or fall through to DeepSeek-V4 `ds_linear` (`o_groups`). This PR dispatches them in `get_quant_method` (same pattern as DSV4):

- Linear/MoE: new `fp8`/`linear` and `fp8`/`moe` schemes that resolve `float8_e4m3fn` + `weight_scale_inv` tiles
- Ascend 950: re-quantize to MXFP8 at load; other generations: resolve to BF16
- Honour both `ignored_layers` and `modules_to_not_convert` (e.g. `visual.merger.*`)
- Per-tensor / static activation: actionable error, not AttributeError

Fixes #14467

Refs #14466

Yes. Official block-wise FP8 checkpoints can be served as published:

```bash
vllm serve /model/Qwen3.8-27B-FP8 \
  --host 0.0.0.0 \
  --port 8000 \
  --served-model-name qwen3.8-27b-fp8 \
  --tensor-parallel-size 1 \
  --max-model-len 128000 \
  --max-num-seqs 8 \
  --gpu-memory-utilization 0.85 \
  --trust-remote-code \
  --speculative-config '{"method": "qwen3_5_mtp", "num_speculative_tokens": 3, "enforce_eager": true}'
```

### Performance (Ascend 950PR, from internal 950PR test notes)

Workload unless noted: `vllm bench serve --backend openai-chat --dataset-name random --random-input-len 64000 --random-output-len 3000 --num-prompts 10 --request-rate 10`.

#### Qwen3.8-27B family (this PR)
| Checkpoint | Path | Quant | MTP | KV cache | Bench output tok/s | Bench duration | Mean TTFT | Mean TPOT | Accept. rate / length | Notes |
|---|---|---|---|---|---:|---:|---:|---:|---|---|
| BF16 | `Qwen3.8-27B` | none | `mtp` ×3 | — | **45.31** | 662.11 s | 143551 ms | 157.69 ms | 55.82% / 2.67 | baseline |
| W8A8 (ModelSlim) | `Qwen3.8-27B-w8a8` | `--quantization ascend` | `qwen3_5_mtp` ×3 | 883,762 tok (6.74× @ 131k) | **92.26** | 325.16 s | 112931 ms | 66.21 ms | 64.21% / 2.93 | first serve |
| W8A8 + cpu_binding | same | ascend | same | 1,076,763 tok (4.11× @ 262k) | **96.40** | 311.21 s | 96627 ms | 66.96 ms | 64.21% / 2.93 | handbook also lists logger **62.9** tok/s |
| W8A8-MXFP8 | `Qwen3.8-27B-w8a8-mxfp8` | ascend | same | 1,011,932 tok (3.86× @ 262k) | **68.02** | 441.03 s | 142280 ms | 93.44 ms | 63.24% / 2.90 | handbook logger **69** tok/s |
| **Native FP8 (this PR)** | `Qwen3.8-27B-FP8` | **none** (`quant_method: fp8`) | `qwen3_5_mtp` ×3 | **913,333 tok** | **66.83** | 448.87 s | 143909 ms | 81.87 ms | **51.30% / 2.54** | serve `--max-model-len 128000`; handbook logger **128** tok/s |

Native FP8 vs same-card W8A8-MXFP8: bench output tok/s is in the same band (**66.8 vs 68.0**). vs BF16 it is **+47%** output tok/s (66.8 vs 45.3). vs ModelSlim W8A8 it is still behind on the **bench average** (66.8 vs 92–96), mainly because MTP acceptance is lower (51% / AL 2.54 vs ~64% / AL 2.93) and TTFT is similar to MXFP8 (~144 s mean under 10-way 64k prefill).

### Accuracy

Evaluated with [AISBench](https://github.com/AISBench/benchmark) on GPQA Diamond
(198 questions, 0-shot, `gpqa_gen_0_shot_str`), served by this PR on
**1x Ascend950PR, TP1, vLLM v0.27.1**, no speculative decoding.
Sampling: `temperature=0.6, top_k=10, top_p=0.95, repetition_penalty=1.03`,
`max_out_len=32768`, concurrency 32.

| GPQA_diamond | 07ea2e | accuracy | gen | 87.88 |

All 198 requests finished with `finish_reason=stop`; zero `length`,
`repetition` or `error` terminations, i.e. no truncation and no degenerate
repetition loops.

For context, the reference numbers published in the Qwen3.8-27B tutorial
(measured on **vllm-ascend v0.23.0rc1**, different hardware, so not a strictly
apples-to-apples comparison) are BF16 90.40, W8A8 89.90, W8A8-MXFP8 89.39.
With n=198 the 1-sigma sampling error is ~2.3 points, so the 87.88 observed
here is ~1.1 sigma from the BF16 reference and is not a statistically
significant regression. The gap corresponds to 5 questions out of 198.

- vLLM version: v0.27.1


vllm_ascend/quantization/fp8_config.py
```diff
@@ -31,13 +46,80 @@ def __init__(self, *args, **kwargs):
...
         raise NotImplementedError('Ascend hardware dose not support "get_min_capability" feature.')
 
+    @classmethod
+    def from_config(cls, config: dict[str, Any]) -> "AscendFp8Config":
+        quant_config = super().from_config(config)
+        # Checkpoints spell the exclusion list either way, and upstream Fp8Config
+        # only reads `ignored_layers`. Qwen3.8-27B-FP8 keeps `visual.merger.*` in
+        # `modules_to_not_convert`, so both keys have to be honoured.
+        ignored_layers = list(quant_config.ignored_layers or [])
+        for key in ("ignored_layers", "modules_to_not_convert"):
```
nu11ptr7: `modules_to_not_convert` entries are merged into `ignored_layers` verbatim - if they use globs like `visual.merger.*`, they only work if the skip check glob-matches. Worth confirming `is_layer_skipped` handles those patterns.

yiminghub2024: Good catch. Confirmed against Qwen/Qwen3.8-27B-FP8: modules_to_not_convert is 882 exact module names (e.g. visual.blocks.0.mlp.linear_fc1, visual.deepstack_merger_list.0.linear_fc1). There are no * / re: globs.

is_layer_skipped is called with match_mode="exact" (same default as upstream Fp8Config). That matches this checkpoint. We also added a UT for the official visual layer name.

Not adding glob matching here: it would change which layers get skipped and could incorrectly leave quantized weights unquantized. If a later checkpoint actually ships glob/re: entries, we can extend skip matching then.  @nu11ptr7

vllm_ascend/quantization/methods/fp8_block.py
```diff
@@ -0,0 +1,355 @@
...
+            f"({block_n}, {block_k}) needs a scale of shape {expected_scale_shape}, but the "
+            f"checkpoint provided {tuple(scale_inv.shape)}. The weight and its "
+            "`weight_scale_inv` are mismatched or unpaired."
+        )
+
+    resolved = torch.empty((out_features, in_features), dtype=out_dtype, device=weight.device)
+    rows_per_step = max(block_n, _ROWS_PER_DEQUANT_STEP // block_n * block_n)
+    for row_start in range(0, out_features, rows_per_step):
+        row_end = min(row_start + rows_per_step, out_features)
+        row_scales = scale_inv[row_start // block_n : cdiv(row_end, block_n)].to(torch.float32)
```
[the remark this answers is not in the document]
yiminghub2024: done

vllm_ascend/quantization/methods/fp8_block.py
```diff
@@ -0,0 +1,355 @@
...
+            self.model_dtype,
+        )
+        del layer.weight_scale_inv
+
+        if self.mxfp8_method is not None and not _supports_mx_regroup(resolved.shape[1], self.mxfp8_method.group_size):
+            logger.warning_once(
+                "Reduction dim %d of %s is not a multiple of the MXFP8 group size %d; serving this "
+                "layer in %s instead of MXFP8.",
+                resolved.shape[1],
+                layer.prefix,
```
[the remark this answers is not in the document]
yiminghub2024: done

vllm_ascend/quantization/methods/fp8_block.py
```diff
@@ -0,0 +1,355 @@
...
+        resolved,
+        dst_type=BLOCK_FP8_WEIGHT_DTYPE,
+        scale_alg=scale_alg,
+    )
+    out_features, in_features = resolved.shape
+    return quantized, _as_linear_mx_scale(mx_scale, out_features, in_features, group_size)
+
+
+def _supports_mx_regroup(in_features: int, group_size: int) -> bool:
+    return in_features % group_size == 0
```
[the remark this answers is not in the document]
yiminghub2024: Leaving the asserts as-is. This matches `AscendDeepseekV4FP8Config` in the same file, and the `fp8/linear` and `fp8/moe` schemes are registered at import time, so `get_scheme_class` is not None on the serve path.
vLLM Ascend is not run with `python -O`. If we convert asserts here we should do it for the whole file in a follow-up, not only this dispatch.

vllm_ascend/quantization/fp8_config.py
```diff
@@ -31,13 +46,80 @@ def __init__(self, *args, **kwargs):
...
+        logger.info_once("Using the vLLM Ascend block-wise fp8 Quantization now!")
+
+        if is_linear:
+            scheme_class = get_scheme_class(FP8_METHOD, "linear")
+            assert scheme_class is not None, f"No scheme registered for {FP8_METHOD}/linear"
+            return AscendLinearMethod(scheme_class(self.weight_block_size))
+
+        scheme_class = get_scheme_class(FP8_METHOD, "moe")
+        assert scheme_class is not None, f"No scheme registered for {FP8_METHOD}/moe"
+        return AscendFusedMoEMethod(
```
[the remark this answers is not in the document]
yiminghub2024: Leaving the asserts as-is. This matches `AscendDeepseekV4FP8Config` in the same file, and the `fp8/linear` and `fp8/moe` schemes are registered at import time, so `get_scheme_class` is not None on the serve path.
vLLM Ascend is not run with `python -O`. If we convert asserts here we should do it for the whole file in a follow-up, not only this dispatch.

vllm_ascend/quantization/methods/fp8_block.py
```diff
@@ -0,0 +1,367 @@
...
+
+    def __init__(self, weight_block_size: tuple[int, int], moe_config):
+        self.block_n, self.block_k = weight_block_size
+        # AscendFusedMoEMethod tags a scale as group-wise when the scheme reports
+        # a group size, which is how the upstream loader learns to narrow
+        # `*_weight_scale_inv` per expert shard.
+        self.group_size = self.block_k
+        self.model_dtype = get_current_vllm_config().model_config.dtype
+        self.moe_config = moe_config
+        self.mxfp8_method = AscendW8A8MXFP8DynamicFusedMoEMethod() if is_950() else None
```
kunpengW-code, line 206: The FP8 weight is actually available only in the 950 scenario. In the current implementation, floating-point calculation is directly used in non-950 scenarios, which is meaningless and increases the code complexity. The A2/A3 can directly use the BF16 weight. It is recommended that the floating-point branch be deleted.

vllm_ascend/quantization/methods/fp8_block.py
```diff
@@ -0,0 +1,367 @@
...
+        bias: torch.Tensor | None = None,
+        tp_rank: int | None = 0,
+    ) -> torch.Tensor:
+        if self.mxfp8_method is not None:
+            return self.mxfp8_method.apply(layer, x, bias, tp_rank)
+        return torch.ops.vllm.unquantized_gemm(x, layer.weight, bias)
+
+
+@register_scheme(FP8_METHOD, "moe")
+class AscendFp8BlockFusedMoEMethod(AscendMoEScheme):
```
kunpengW-code, line 189: After removing the floating-point branch, this class can directly inherit from MXFP8 and only override methods other than `apply`, just like the implementation in dsv4. With this modification, this class can be migrated to the `w8a8_mxfp8.py` file for unified management, eliminating the need to create a new `fp8_block.py` file.

vllm_ascend/quantization/methods/fp8_block.py
```diff
@@ -0,0 +1,367 @@
...
+        mx_scale = torch.empty(
+            (num_experts, weight.shape[1], in_features // mxfp8_method.group_size),
+            dtype=torch.uint8,
+            device=weight.device,
+        )
+        for expert in range(num_experts):
+            resolved = resolve_block_scales(
+                weight[expert], scale_inv[expert], self.block_n, self.block_k, self.model_dtype
+            )
+            quantized, expert_scale = _mx_quantize(resolved, scale_alg, mxfp8_method.group_size)
```
kunpengW-code: Requantizing weights from the fp8 format to the mxfp8 format may have an impact on precision. If possible, could you provide some test results for the precision dataset?

yiminghub2024: done ,thanks @kunpengW-code

vllm_ascend/quantization/methods/fp8_block.py
yiminghub2024: ok ,after merged , i will open new pr to fix it later

vllm_ascend/quantization/methods/fp8_block.py
yiminghub2024: ok ,i will remove this with a new pr  later thanks

wangxiyuan (dismissed):

vllm_ascend/quantization/methods/fp8_block.py
```diff
@@ -0,0 +1,367 @@
...
+            device=weight.device, dtype=torch.float32
+        )
+        row_scales = row_scales.repeat_interleave(block_n, dim=0)[: row_end - row_start]
+        row_scales = row_scales.repeat_interleave(block_k, dim=1)[:, :in_features]
+        resolved[row_start:row_end] = weight[row_start:row_end].to(torch.float32) * row_scales
+    return resolved
+
+
+def _as_linear_mx_scale(mx_scale: torch.Tensor, out_features: int, in_features: int, group_size: int) -> torch.Tensor:
+    mx_scale = mx_scale.view(torch.uint8)
```
lcfenglinwan: The data type of the scale output by the `npu_dynamic_mx_quant` interface is `torch.uint8`. Therefore, no further type conversion is required here.

vllm_ascend/quantization/methods/fp8_block.py
```diff
@@ -0,0 +1,367 @@
...
+
+def _mx_quantize(resolved: torch.Tensor, scale_alg: int, group_size: int) -> tuple[torch.Tensor, torch.Tensor]:
+    """Re-quantize a dense matrix to MXFP8, returning the weight and uint8 scale."""
+    quantized, mx_scale = torch_npu.npu_dynamic_mx_quant(
+        resolved,
+        dst_type=BLOCK_FP8_WEIGHT_DTYPE,
+        scale_alg=scale_alg,
+    )
+    out_features, in_features = resolved.shape
+    return quantized, _as_linear_mx_scale(mx_scale, out_features, in_features, group_size)
```
lcfenglinwan: Why is it necessary to process mx_scale here? Could there be situations where it cannot be used directly?

vllm_ascend/quantization/methods/fp8_block.py
yiminghub2024: fixed ,thanks @lcfenglinwan

vllm_ascend/quantization/methods/fp8_block.py
yiminghub2024: fixed ,thanks @lcfenglinwan

vllm_ascend/quantization/fp8_config.py
```diff
@@ -31,13 +50,89 @@ def __init__(self, *args, **kwargs):
     def get_min_capability(cls) -> int:
         raise NotImplementedError('Ascend hardware dose not support "get_min_capability" feature.')
 
+    @classmethod
+    def from_config(cls, config: dict[str, Any]) -> "AscendFp8Config":
```
lcfenglinwan: Why is it necessary to rewrite this function? Are there any differences caused by hardware?

vllm_ascend/quantization/fp8_config.py
```diff
@@ -31,13 +50,89 @@ def __init__(self, *args, **kwargs):
...
+
+        self._verify_block_quantization()
+        logger.info_once("Using the vLLM Ascend block-wise fp8 Quantization now!")
+
+        if is_linear:
+            scheme_class = get_scheme_class(FP8_METHOD, "linear")
+            assert scheme_class is not None, f"No scheme registered for {FP8_METHOD}/linear"
+            return AscendLinearMethod(scheme_class(self.weight_block_size))
+
+        scheme_class = get_scheme_class(FP8_METHOD, "moe")
```
lcfenglinwan: add `if is_moe:` before L129

vllm_ascend/quantization/fp8_config.py
```diff
@@ -31,13 +50,89 @@ def __init__(self, *args, **kwargs):
...
+        assert scheme_class is not None, f"No scheme registered for {FP8_METHOD}/moe"
+        return AscendFusedMoEMethod(
+            scheme_class(self.weight_block_size, layer.moe_config),
+            layer.moe_config,
+            tid2eid=tid2eid,
+        )
```
lcfenglinwan, line 116: add `return None` at the end

vllm_ascend/quantization/fp8_config.py
```diff
@@ -31,13 +50,89 @@ def __init__(self, *args, **kwargs):
...
+            AscendFusedMoEMethod,
+            AscendLinearMethod,
+        )
+
+        is_linear = isinstance(layer, LinearBase)
+        is_moe = _is_fused_moe_layer(layer)
+        if not is_linear and not is_moe:
+            # Embeddings and the lm_head stay in the model dtype in every native
+            # FP8 release published so far.
+            if isinstance(layer, VocabParallelEmbedding):
```
lcfenglinwan: I think this processing for VocabParallelEmbedding can be removed. plz check

vllm_ascend/quantization/fp8_config.py
yiminghub2024: done ,thanks

vllm_ascend/quantization/fp8_config.py
yiminghub2024: done ,thanks

vllm_ascend/quantization/fp8_config.py
yiminghub2024: done ,thanks

vllm_ascend/quantization/fp8_config.py
yiminghub2024: done ,thanks  @lcfenglinwan

wangxiyuan (approved):

wangxiyuan:
CI failure doesn't relate to this change. Merged now.

PR #14863, opened 2026-08-24
[BugFix][310P] Fix Moe aclgraph error caued by routing to aclop on 310p

YangShuai52:
ACLGraph Compatibility: Implemented a pre-casting mechanism for router weights to prevent unsupported forward-time Cast operations on 310P hardware.
No

- vLLM version: v0.27.1


vllm_ascend/_310p/fused_moe/fused_moe.py
```diff
@@ -116,6 +116,10 @@ def __init__(
         tid2eid=None,
         n_shared_experts: int = 0,
     ):
+        if not vllm_version_is("0.27.1") and gate is not None:
+            # Pre-cast the internal router weight during model loading. A
+            # forward-time Cast cannot be captured by ACLGraph on 310P.
+            gate.precast_fp32_weight = True
```
nu11ptr7: This only skips vLLM 0.27.1 - any other version (older or newer) silently gets the fp32 precast. A capability check would survive future upgrades better.

YangShuai52: Thank you for your suggestion, I have made the revisions.

Tflowers-0129 (approved):

PR #14747, opened 2026-08-21
[BugFix][v0.26.0rc] Fix MegaMoe hang when main and draft models have inconsistent MoE quantization

kunpengW-code:
### What this PR does / why we need it?
If the quantization modes of the main model and draft model are inconsistent, for example, the main model is a quantized model and the draft model is a floating-point model, the MEGAMOE operator reports an error, indicating that this format is not supported.
The error message is as follows:
```
RuntimeError: call aclnnMegaMoe failed, detail:[PID: 2799] 2026-08-20-22:15:59.473.813 Execution_Error(EZ1008): Failed to execute operator MegaMoe_6081. Reason: Failed to execute tiling function.
        Solution: 1.If the tiling function does not exist, check whether the tiling function is registered successfully. 2.If the tiling function fails to be executed, check the implementation logic of the tiling function.
TraceBack (most recent call last):
        weight1[0] must be 2-dimension, but got 3 dim.[FUNC:CheckWeight1Input][FILE:mega_moe_tiling_arch22.cpp][LINE:648]
        CheckWeight1Input failed.[FUNC:MegaMoeA2A3CheckShapeAndSetTiling][FILE:mega_moe_tiling_arch22.cpp][LINE:860]
        MegaMoeA2A3 CheckShapeAndSetTiling Failed[FUNC:MegaMoeA2A3TilingFuncImpl][FILE:mega_moe_tiling_arch22.cpp][LINE:1139]
        Failed to execute operator MegaMoe_6081. Reason: Failed to execute tiling function.
        Check NnopbaseExecutorDoTiling(executor) failed
        Check NnopbaseExecutorTilingAndUpdateBinInfo(executor) failed
        Check NnopbaseExecutorMatchCache(executor) failed
        Check NnopbaseRunForWorkspace(*executor, workspaceSize) failed
```

**Modification plan:**
When it is identified that the current weight is in this hybrid format, the draft model will be rolled back to the small operator path.

This PR is only a workaround for version 0.26.0rc. The main branch will address this issue through the megamoe operator.

### Does this PR introduce _any_ user-facing change?
no

### How was this patch tested?
nightly test case: `glm-4.7-w8a8`

- vLLM version: v0.26.0
- vLLM main: https://github.com/vllm-project/vllm/commit/d02df748bf9efd99022f1a062597dc3cb3808485

PR #14864, opened 2026-08-24
[v0.26.0rc][BugFix] Fix dispatch_v2 tokens capacity to follow normal MC2 branch logic

kunpengW-code:
Follow-up modifications for PR #14747
This PR fixes the calculation of `_dispatch_v2_tokens_capacity` (normal MC2 tokens capacity) to follow the normal MC2 branch logic.
Previously, when `enable_prefill_mc2` or `use_mega_moe` was true, `max_num_tokens` was based on `max_num_batched_tokens` (prefill+decode unified batch), which is too large for the normal MC2 decode path.
This PR recomputes the token count from the decode-only configuration so that `_dispatch_v2_tokens_capacity` is computed independently from the fused MC2 path.

no

/nightly Qwen3.5-397B-A17B-w8a8-mtp --branch releases/v0.26.0rc

- vLLM version: v0.26.0

### DIFF
--- a/vllm_ascend/ascend_forward_context.py
+++ b/vllm_ascend/ascend_forward_context.py
@@ -246,7 +246,18 @@ def set_mc2_tokens_capacity(vllm_config, max_num_reqs, uniform_decode_query_len)
         else:
             num_tokens_per_tp_rank = min(num_tokens_per_tp_rank, _DISPATCH_FFN_COMBINE_TOKENS_PER_RANK_LIMIT)
         global _dispatch_v2_tokens_capacity
-        _dispatch_v2_tokens_capacity = min(num_tokens_per_tp_rank, _MC2_TOKENS_PER_RANK_LIMIT) * tp_size
+        # Compute dispatch_v2 (normal MC2) tokens capacity independently from the
+        # fused_mc2 path above. When enable_prefill_mc2 or use_mega_moe is true,
+        # max_num_tokens is based on max_num_batched_tokens (prefill+decode unified
+        # batch), which is too large for the normal MC2 decode path. Recompute the
+        # token count from the decode-only configuration so that
+        # _dispatch_v2_tokens_capacity follows the normal MC2 branch logic.
+        if vllm_config.compilation_config.cudagraph_capture_sizes:
+            dispatch_v2_max_num_tokens = vllm_config.compilation_config.max_cudagraph_capture_size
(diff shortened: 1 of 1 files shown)


kunpengW-code:
/nightly Qwen3.5-397B-A17B-w8a8-mtp --branch releases/v0.26.0rc
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/32743592187)

kunpengW-code:
/nightly Qwen3.5-397B-A17B-w8a8-mtp --branch releases/v0.26.0rc
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/32796677204)

vllm_ascend/ascend_forward_context.py
```diff
@@ -246,7 +246,18 @@ def set_mc2_tokens_capacity(vllm_config, max_num_reqs, uniform_decode_query_len)
...
         global _dispatch_v2_tokens_capacity
-        _dispatch_v2_tokens_capacity = min(num_tokens_per_tp_rank, _MC2_TOKENS_PER_RANK_LIMIT) * tp_size
+        # Compute dispatch_v2 (normal MC2) tokens capacity independently from the
+        # fused_mc2 path above. When enable_prefill_mc2 or use_mega_moe is true,
+        # max_num_tokens is based on max_num_batched_tokens (prefill+decode unified
+        # batch), which is too large for the normal MC2 decode path. Recompute the
+        # token count from the decode-only configuration so that
+        # _dispatch_v2_tokens_capacity follows the normal MC2 branch logic.
+        if vllm_config.compilation_config.cudagraph_capture_sizes:
+            dispatch_v2_max_num_tokens = vllm_config.compilation_config.max_cudagraph_capture_size
```
nu11ptr7, line 256: The capacity now comes from the decode-only path (max cudagraph capture size). If prefill and decode share this buffer but a prefill batch can exceed the decode graph size, the buffer is sized for decode only - is that safe?

kunpengW-code: The capacity limit here is used to determine whether to use MC2/ALL2ALL. Any excess will be uniformly handled through ALL2ALL, and there is no security risk.

weijinqian0 (approved):

PR #14883, opened 2026-08-25
[BugFix][Spec Decode] Correct rejected token seq_lens in MRV1

jiajinzhu2:

ModelRunner V1 does not subtract rejected target tokens from the device-side `seq_lens` before building the later one-token draft decode steps. SFA and DSA consume this tensor as the exact KV sequence length. Consequently, their later draft Attention steps use an overestimated sequence length that still includes rejected token positions.

The issue was observed by printing the rejected-token count and the device-side sequence lengths used by three consecutive draft steps:

```text
[V1_DEVICE_SEQ_LENS] rejected=[3] steps=[[47], [48], [49]]
[V1_DEVICE_SEQ_LENS] rejected=[2] steps=[[56], [57], [58]]
[V1_DEVICE_SEQ_LENS] rejected=[3] steps=[[98], [99], [100]]
```

The sequence lengths advance by one for each draft step, but their baseline is not reduced by the corresponding rejected-token count. Therefore, the device-side `seq_lens` is overestimated by 3, 2, and 3 tokens in these examples. SFA-style model backends pass this incorrect value to both the sparse Indexer and the Attention kernel as the exact KV sequence length. As a result, both the Indexer and Attention execute with an overestimated KV length that includes the rejected token positions.

This PR preserves the unadjusted sequence lengths used by merged step 0, then copies them to an independent NPU buffer and subtracts rejected tokens for the remaining one-token draft decode steps. The CPU-side sequence lengths remain an optimistic upper bound to avoid an NPU-to-CPU synchronization.

N/A

```
========================================================
Successful requests:                      100
Failed requests:                          0
Maximum request concurrency:              4
Benchmark duration (s):                   1272.46
Total input tokens:                       6882
Total generated tokens:                   25501
Request throughput (req/s):               0.08
Output token throughput (tok/s):          20.04
Peak output token throughput (tok/s):     14.00
Peak concurrent requests:                 7.00
Total token throughput (tok/s):           25.45
Mean TTFT (ms):                           788.87
Median TTFT (ms):                         798.56
P99 TTFT (ms):                            934.36
Mean TPOT (ms):                           194.55
Median TPOT (ms):                         193.15
P99 TPOT (ms):                            219.91
Mean ITL (ms):                            440.91
Median ITL (ms):                          449.26
P99 ITL (ms):                             480.29
Acceptance rate (%):                      43.11
Acceptance length:                        2.29
Drafts:                                   11113
Draft tokens:                             33339
Accepted tokens:                          14374
Per-position acceptance (%):
  Position 0:                             86.72
  Position 1:                             32.79
  Position 2:                              9.84
========================================================
```


vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -1011,6 +1011,19 @@ def _propose(
...
         )
 
+        # Merged step 0 is the draft model's first-pass/prefill forward, not a
+        # one-token draft decode step. It reuses the target-model layout and
+        # therefore keeps the unadjusted sequence lengths. Before building the
+        # remaining one-token draft decode steps, remove the rejected target
+        # tokens from an independent NPU buffer. This mirrors upstream vLLM's
+        # rejection-correction boundary while preserving step 0 metadata and
+        # the model runner's shared seq_lens.
+        if not self.parallel_drafting and self.num_speculative_tokens > 1 and num_rejected_tokens_gpu is not None:
```
drslark, line 1109: It is a bug just in mrv1.
But it influence the alignment between v2 and v1.
So, I think it is necessary.

drslark (approved):
LGTM.

slippersss (approved):

weijinqian0 (approved):

PR #14900, opened 2026-08-25
[v0.26.0][BugFix][Model Runner] Skip compilation during MC2 profiling

yiz-liu:

The pre-KV MC2 dummy run can be the first shape eligible for npugraph_ex static-kernel compilation. At that time, attention metadata is not built yet, so the early return makes the compiled kernel miss attention operators.

Skip the compiled model only for this profiling run, allowing later graph capture with complete attention metadata to trigger static-kernel compilation.

No.

Added regression coverage for the target and non-target paths.

- vLLM version: v0.26.0


weiguihua2 (approved):

yiz-liu:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

yiz-liu:
/rerun
[Bot]: rerun completed.

**Failed:**

PR #14920, opened 2026-08-25
[BugFix][P/D] Derive DSV4 Mooncake block size from KV spec

Yuli-yx:

This is a clean resubmission of #11620 on the current main branch. The original fix was authored by @liziyu179.

DeepSeek-V4 uses heterogeneous KV cache groups. For logical block sizes 32, 64, and 128, the smallest generated physical group block sizes are 2, 4, and 8 respectively. During KV cache initialization, EngineCore writes that minimum group size back to `vllm_config.cache_config.block_size`.

`MooncakeConnectorScheduler` previously copied this mutated global value into transfer metadata. With the default DSV4 layout, the producer therefore sent `remote_block_size=2`, while the decoder attention path used kernel size 32 and failed with:

    remote_block_size(2) not divisible by kernel_size(32)

Instead of resetting the value to 32, this change derives the scheduler transfer block size from the already-resolved DSV4 `AscendSlidingWindowMLASpec`. That SWA spec retains the logical block size selected by the user, so 32, 64, and 128 are preserved without a hardcoded fallback or mutation of `vllm_config`.

The change is limited to MooncakeConnectorV1 scheduler metadata. Non-DSV4 models keep the existing `cache_config.block_size` path. If DSV4 supports another logical block size in the future and exposes it through the resolved SWA spec, Mooncake will consume it without adding another allowlist entry.

No new option or behavior is introduced. It fixes DeepSeek-V4 P/D execution with MooncakeConnectorV1, including explicit `--block-size 64` and `--block-size 128` configurations.

- Added a scheduler regression test covering resolved DSV4 layouts `2 -> 32`, `4 -> 64`, and `8 -> 128`.
- The test verifies that Mooncake uses the SWA logical block size and does not modify the global `cache_config.block_size`.
- Added a non-DSV4 fallback test.
- `uvx ruff==0.14.0 check` passed for both changed files.
- `uvx ruff==0.14.0 format --check` passed for both changed files.
- `python -m py_compile` passed for both changed files.
- Unit tests were not executed locally because the Windows development host does not provide `torch_npu`; selected-test CI is expected to run them.

Refs #11620


vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
```diff
@@ -1691,6 +1691,8 @@ class MooncakeConnectorScheduler:
     def __init__(self, vllm_config: VllmConfig, engine_id: str, kv_cache_config: KVCacheConfig):
         self.vllm_config = vllm_config
         self.kv_cache_config = kv_cache_config
+        if vllm_config.model_config.hf_config.model_type == "deepseek_v4":
+            refresh_block_size(vllm_config)
```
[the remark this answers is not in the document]
Yuli-yx: Addressed in 6a51e3b0f with defensive getattr access for model_config and hf_config. This keeps the DeepSeek-V4-only scope without introducing AttributeError for minimal configs.

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
```diff
@@ -1691,6 +1691,12 @@ class MooncakeConnectorScheduler:
     def __init__(self, vllm_config: VllmConfig, engine_id: str, kv_cache_config: KVCacheConfig):
         self.vllm_config = vllm_config
         self.kv_cache_config = kv_cache_config
+        model_config = getattr(vllm_config, "model_config", None)
```
weiguihua2: Add test cases for monitoring.

Yuli-yx: Added two scheduler regression tests in 1c45aafda: the DSV4 path verifies block size is restored from 2 to 32 before init_ascend_config and before the scheduler snapshot, while the non-DSV4 path verifies the refresh is not called and block size remains unchanged.

vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
```diff
@@ -1691,6 +1691,12 @@ class MooncakeConnectorScheduler:
...
+        hf_config = getattr(model_config, "hf_config", None)
+        if getattr(hf_config, "model_type", None) == "deepseek_v4":
+            # KV cache initialization replaces the global block size with the
+            # smallest DSV4 cache-group block size (2 for the C4 state group).
+            refresh_block_size(vllm_config)
```
weiguihua2: Hardcoding 32 in this block might be problematic; let's discuss it further.

Yuli-yx: Updated in 56f5e35ca. I removed the `refresh_block_size()` call instead of hardcoding a fallback value.

The scheduler now reads its transfer block size from the already-resolved DeepSeek-V4 `AscendSlidingWindowMLASpec`. This spec retains the logical block size selected by the user, while `cache_config.block_size` may already contain the smallest physical cache-group size. As a result, `--block-size 32/64/128` produces transfer block sizes 32/64/128 respectively, without mutating `vllm_config` or maintaining a Mooncake-side allowlist.

The production diff is one lookup helper plus the scheduler assignment. The regression test covers `2 -> 32`, `4 -> 64`, and `8 -> 128`, and verifies the global cache block size remains unchanged. Non-DSV4 behavior still falls back to the existing value.

PR #14921, opened 2026-08-25
[Bugfix][KV Offload] Bound sparse KV offload memory by NPU, DRAM and workload limits

Abstrey:

Sparse KV offload (SFA with `enable_sparse_li_c8`) had two problems in memory planning:

1. `SparseKVOffloadManager` always sized the CPU KV pool at the **full configured `dram_size_per_dp_GB`**, regardless of how many KV blocks the engine actually planned to use. On large-DRAM configs (e.g. 180 GiB per DP group) this allocates far more DRAM than needed and makes pool initialization very slow (tens of minutes of hugepage-backed allocation), even when the active workload only needs a fraction of it.
2. The worker-side logic (`update_available_memory_for_sparse_kv_offload`) only enlarged total available memory by a host/device page-size ratio. The number of KV blocks was therefore never bounded by DRAM capacity or by active workload demand, and when the blocks implied by free NPU memory exceeded the configured DRAM, startup failed with a generic "please increase `sparse_kv_offload_config.dram_size_per_dp_GB`" error.

This PR replaces that logic with an explicit memory plan:

- New `plan_sparse_kv_offload_memory()` returns a `SparseKVOffloadMemoryBudget` that bounds the block count by the minimum of three limits and reports the limiting factor:
  - **NPU capacity** — free device memory (counting the full page size when `keep_device_kv_cache=True`, device portion only otherwise);
  - **DRAM limit** — `dram_size_per_dp_GB` minus a per-layer alignment reserve (each host layer may lose up to 3 × 2 MiB to 2 MiB-aligned allocations);
  - **Workload demand** — `max_blocks_per_request × max_num_seqs` plus the null block.
- `NPUWorker.determine_available_memory()` now applies the budget through `_apply_kv_offload_decode_memory_constraints()` and returns the planned bytes, with a one-shot log line showing every limit, the final block count, the limiting factor, and the host/device/alignment byte breakdown.
- `SparseKVOffloadManager` sizes the actual CPU pool from the planned block count (`num_blocks × host page sizes + alignment reserve`, via the new `get_sparse_kv_offload_cpu_pool_size_bytes()`) instead of the full configured DRAM, and raises an actionable error (planned vs. limit vs. `num_blocks`) if the planned pool still exceeds the DRAM limit.

As a result, the offload pool is allocated only as large as needed: DRAM footprint and CPU pool initialization time drop accordingly, and undersized `dram_size_per_dp_GB` configurations fail early with a clear message.

Yes. With sparse KV offload enabled:

- The CPU offload pool is no longer always allocated at the full `dram_size_per_dp_GB`; it is sized from the planned block count, so the number of KV cache blocks can be smaller than before when DRAM or workload demand is the limiting factor. The limiting factor and full plan are logged at startup.
- The previous error "Needed dram size ... Please increase `sparse_kv_offload_config.dram_size_per_dp_GB`" is replaced by a new "planned CPU pool exceeds DRAM limit after alignment" error carrying planned/limit/num_blocks values.
- Log messages around sparse KV offload memory planning have changed.

Unit tests added:

- `tests/ut/distributed/kv_transfer/sparse_kv_offload/test_sparse_kv_offload_manager.py` — memory plan limited by workload / DRAM / NPU (each asserting block counts, planned bytes and the limiting factor); non-positive NPU or DRAM capacity producing zero blocks; invalid spec layouts rejected (missing host or device specs, mixed block sizes); CPU pool size including the per-layer alignment reserve; `UniformTypeKVCacheSpecs` support; manager initializing the backend with the planned pool size (rank 0 allocates, other ranks 0) with a TP barrier; and rejection when the planned pool exceeds the DRAM limit.
- `tests/ut/worker/a2/test_worker_v1.py` — constraints disabled returns memory unchanged; planner budget returned by `_apply_kv_offload_decode_memory_constraints`; `kv_cache_spec` fetched when not cached; explicit `kv_cache_memory_bytes` path and the normal `determine_available_memory` path both apply the constraints.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
+++ b/vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
@@ -169,6 +169,8 @@ class SparseKVOffloadMemoryBudget:
 def _split_host_device_kv_specs(
     kv_cache_spec: dict[str, KVCacheSpec],
 ) -> tuple[list[KVCacheSpec], list[KVCacheSpec]]:
+    if not kv_cache_spec:
+        raise ValueError("Sparse KV offload requires a non-empty kv_cache_spec")
     host_specs: list[KVCacheSpec] = []
     device_specs: list[KVCacheSpec] = []
     for spec in kv_cache_spec.values():
--- a/vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
+++ b/vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
@@ -1,7 +1,7 @@
 import contextlib
 import os
-from dataclasses import dataclass
 import typing
+from dataclasses import dataclass
 from zlib import adler32
 
 import numpy as np
@@ -184,10 +184,7 @@ def _split_host_device_kv_specs(
         raise ValueError("Sparse KV offload requires at least one device KV cache spec")
     block_sizes = {spec.block_size for spec in host_specs + device_specs}
     if len(block_sizes) != 1:
-        raise ValueError(
-            "Sparse KV offload memory planning requires one shared block size, "
-            f"got {sorted(block_sizes)}"
-        )
+        raise ValueError(f"Sparse KV offload memory planning requires one shared block size, got {sorted(block_sizes)}")
     return host_specs, device_specs
 
 
@@ -204,15 +201,11 @@ def plan_sparse_kv_offload_memory(
     device_page_size_bytes = sum(spec.page_size_bytes for spec in device_specs)
     total_page_size_bytes = host_page_size_bytes + device_page_size_bytes
 
-    host_alignment_reserve_bytes = (
-        len(host_specs) * _CPU_CACHE_MAX_ALIGNMENT_OVERHEAD_PER_LAYER
-    )
+    host_alignment_reserve_bytes = len(host_specs) * _CPU_CACHE_MAX_ALIGNMENT_OVERHEAD_PER_LAYER
     usable_dram_bytes = max(dram_limit_bytes - host_alignment_reserve_bytes, 0)
     dram_limit_blocks = usable_dram_bytes // host_page_size_bytes
 
-    npu_page_size_bytes = (
-        total_page_size_bytes if keep_device_kv_cache else device_page_size_bytes
-    )
+    npu_page_size_bytes = total_page_size_bytes if keep_device_kv_cache else device_page_size_bytes
     npu_limit_blocks = max(available_device_memory_bytes, 0) // npu_page_size_bytes
 
     max_blocks_per_request = max(
@@ -222,10 +215,7 @@ def plan_sparse_kv_offload_memory(
         )
         for spec in host_specs + device_specs
     )
-    workload_limit_blocks = (
-        max_blocks_per_request * vllm_config.scheduler_config.max_num_seqs
-        + _VLLM_NULL_BLOCK_COUNT
-    )
+    workload_limit_blocks = max_blocks_per_request * vllm_config.scheduler_config.max_num_seqs + _VLLM_NULL_BLOCK_COUNT
 
     limits = {
         "npu": npu_limit_blocks,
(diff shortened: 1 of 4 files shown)


vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
```diff
@@ -150,6 +153,130 @@ def empty_aligned_int8_cpu_tensors(
...
+    host_alignment_reserve_bytes: int
+    limiting_factor: str
+
+
+def _split_host_device_kv_specs(
+    kv_cache_spec: dict[str, KVCacheSpec],
+) -> tuple[list[KVCacheSpec], list[KVCacheSpec]]:
+    host_specs: list[KVCacheSpec] = []
+    device_specs: list[KVCacheSpec] = []
+    for spec in kv_cache_spec.values():
```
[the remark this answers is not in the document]
Abstrey: Thanks for the review! Addressed in 4db067f.

Note that an empty dict was already rejected downstream by the `if not host_specs` check ("Sparse KV offload requires at least one host KV cache spec"); the added guard mainly prevents an `AttributeError` when `kv_cache_spec` itself is `None`.

vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
```diff
@@ -119,6 +120,8 @@ def allocate_kv_offload_topk_profile_buffers(
 
 
 _CPU_CACHE_ALIGNMENT = 2 * 1024 * 1024
+_CPU_CACHE_MAX_ALIGNMENT_OVERHEAD_PER_LAYER = 3 * _CPU_CACHE_ALIGNMENT
```
Abstrey, line 113: TODO：评估下对齐大小，避免魔鬼数字

Abstrey: 补充说明 _CPU_CACHE_MAX_ALIGNMENT_OVERHEAD_PER_LAYER = 3 × 2MiB 的取值依据：

每层 host 侧的 k_cache_cpu / v_cache_cpu 由 empty_aligned_int8_cpu_tensors 从一块 raw 缓冲区中切出，每个张量的起始地址强制对齐到 2MiB 边界。每层恰好有三处独立的对齐浪费，每处上界 2MiB：

基地址取整——raw 缓冲区起始地址向上取整到 2MiB 边界（分配代码中 total_chunk_num = 1 + sum(chunk_nums) 多出的那 1 块就是为此预留的）；
k_cpu 尾部补齐——k 的字节数向上取整到整数个 2MiB 块；
v_cpu 尾部补齐——同理。
因此保守上界为每层 3 × 2MiB。这是最坏情况界而非运行时精确值（精确浪费取决于当时的尺寸与地址零头），容量规划必须使用不会低估的上界。

该预留量在代码中被计入两次、方向相反：规划 DRAM 块数上限时先减去（防止把对齐浪费当作可用容量），get_sparse_kv_offload_cpu_pool_size_bytes 计算池子大小时加回（保证实际对齐分配装得下）。两者配合，从构造上保证 块数 × host 页宽 + 对齐浪费 ≤ DRAM 配额 恒成立。

vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
zhangsicheng5: Consider to delete the `get_host_device_memory_usage_ratio` function and its related unit tests if it's no longer used now.

vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
```diff
@@ -150,6 +157,111 @@ def empty_aligned_int8_cpu_tensors(
...
+        for spec in host_specs + device_specs
+    )
+    workload_limit_blocks = max_blocks_per_request * vllm_config.scheduler_config.max_num_seqs + _VLLM_NULL_BLOCK_COUNT
+
+    limits = {
+        "npu": npu_limit_blocks,
+        "dram": dram_limit_blocks,
+        "workload": workload_limit_blocks,
+    }
+    limiting_factor = min(limits, key=lambda name: limits[name])
```
zhangsicheng5, line 226: I think's it better to add a warning here if `dram_limit_blocks` is smaller than `npu_limit_blocks`, since in this case the available npu memory (and thus available batch_size or max_model_len) is restricted by host memory size and not fully utilized. We should warn the user that we can further enlarge bs/seqlen by giving more `dram_size_per_dp_GB` if available, in case that it's default value (128GB now) is too small and the user fail to notice this.

vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
Abstrey: Done in a629481 — removed `get_host_device_memory_usage_ratio`, the `TestHostDeviceMemoryRatio` tests, and the imports that became unused (`torch`, `KVCacheSpec`, `AscendMLAAttentionSpec`, `AscendSFAIndexerCacheSpec`).

vllm_ascend/distributed/kv_transfer/sparse_kv_offload/sparse_kv_offload_manager.py
Abstrey: Done in a629481 — added a `logger.warning_once` right after the limits are computed when `dram_limit_blocks < npu_limit_blocks`, including both block counts and the current DRAM budget in GiB, and suggesting to increase `sparse_kv_offload_config.dram_size_per_dp_GB`. Covered by two new unit tests (`test_warning_when_dram_budget_caps_npu_utilization` fires, `test_no_warning_when_dram_budget_not_below_npu` stays silent).

Regarding the alignment add/subtract: both directions are pinned by unit tests — the subtraction side via `test_memory_plan_is_limited_by_dram_capacity` / `test_cpu_pool_size_includes_per_layer_alignment_reserve` (raw = blocks × host page + reserve), and the bound itself is documented above `_CPU_CACHE_MAX_ALIGNMENT_OVERHEAD_PER_LAYER`.

Abstrey:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

zhangsicheng5:
lgtm

weiguihua2 (approved):

PR #14715, opened 2026-08-21
[Feature][Attention] Add TurboQuant 4-bit MLA latent KV cache for SFA

chenweiqiang11:
### What this PR does / why we need it?

This PR adds TurboQuant (TQ4), a 4-bit storage format for the MLA latent KV
cache used by sparse flash attention (SFA) on Ascend. Users enable it with:

```bash
vllm serve <model> --kv-cache-dtype turboquant_4bit_nc ...
```

For `kv_lora_rank=512` and `qk_rope_head_dim=64`, each fused TQ4 KV slot uses
386 bytes:

```text
0                         256          384   386
+--------------------------+------------+-----+
| int4 MLA latent (512 dim) | bf16 rope  |scale|  386 B
+--------------------------+------------+-----+
```

The corresponding C8 slot is 656 bytes and the bf16 K/V pair is 1152 bytes.
TQ4 targets long-context or high-concurrency deployments where KV capacity is
the binding constraint. It is not intended to improve single-request decode
latency.

#### Operator source

| Operator | Source | Source commit |
|---|---|---|
| `turbo_quant_compress_latent` | [cann/ops-nn PR 8222](https://gitcode.com/cann/ops-nn/pull/8222) | `02b3c80a76a9f55fd42d48637dd098dbba36468b` |
| `turbo_quant_sparse_flash_attention` | [cann/ops-transformer PR 9890](https://gitcode.com/cann/ops-transformer/pull/9890) | `4043903c647d76b829be4cefe61249a9c6764032` |

#### Implementation

TurboQuant is inference-only and does not provide autograd or backward
support. The integration:

- validates `turboquant_4bit_nc` before initializing `AscendConfig`;
- uses the packed SFA layout while keeping lightning-indexer C8 independent;
- computes the 386-byte KV-cache specification in Model Runner V1;
- compresses KV writes through `turboquant_kv_rmsnorm_rope`;
- consumes the compressed cache directly in sparse flash attention;
- supports non-DCP and DCP prefill/decode paths;
- registers PrivateUse1 and Meta implementations for both operators.

#### KV capacity

On GLM-5.2 W4A8C8, 78 layers, 8x Ascend 910B4, TP8 + EP, with 6.87 GiB
reserved for KV cache:

| KV format | Slot size | KV pool | Relative to bf16 |
|---|---:|---:|---:|
| bf16 | 1152 B | 76,288 tokens | 1.00x |
| C8 | 656 B | 128,256 tokens | 1.68x |
[...]

PR #14929, opened 2026-08-25
[Test] Supplement UT for EPLB

Spicy-Stick:

This PR adds comprehensive unit tests for `AscendEPLBController` and `_unwrap_moe` in `tests/ut/worker/v2/test_eplb_controller.py` to improve test coverage and ensure correctness of the EPLB controller logic on Ascend.

No.

This patch adds unit tests which can be run via pytest/unittest.

- vLLM version: v0.27.1

### DIFF
--- a/tests/ut/worker/v2/test_eplb_controller.py
+++ b/tests/ut/worker/v2/test_eplb_controller.py
@@ -1,0 +1,259 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM Ascend project
+
+import unittest
+from types import SimpleNamespace
+from unittest.mock import MagicMock, patch
+
+import torch
+import torch.nn as nn
+from vllm.model_executor.models.interfaces import SupportsMultiModal
+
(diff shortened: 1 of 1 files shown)


weijinqian0 (approved):

chenweiqiang11:
The two `AscendEPLBController.step()` tests added here are now failing on `main` and are blocking unrelated PR CI (for example, #14715 CPU job 99043645428: 2843 passed, only these two failed).

This appears to be a merge-order regression: #14099 removed the Ascend `step()` override and moved phase/window gating to graph-safe `prepare_forward()`, while this PR later merged tests written against the old override. The current failing expectations are:

- phase mismatch is forwarded as `is_dummy=True`;
- a closed window clears `expert_load_pass` in `step()`.

Neither behavior exists after #14099, and the new `prepare_forward()` semantics are already covered by `test_eplb_load_collection_phase.py::test_prepare_forward_combines_window_and_phase_device_gates`. The implementation and test blobs are identical between #14715's base and head, so this is reproducible from the current base rather than caused by that PR.

Could these two stale tests be removed or rewritten on `main`? Restoring the old `step()` override would conflict with #14099's async/graph-safe design.

PR #14932, opened 2026-08-25
[Performance][Attention] Reuse DSV4 compressor metadata across layers

maoxx241:

DeepSeek V4 currently launches `compressor_metadata` before every compressor invocation, even though layers in the same cache group use identical compressor inputs within one speculative substep.

This PR syncs main with the `releases/v0.26.0rc` #14994 implementation:

- lazily computes compressor metadata at the first real compressor consumer and reuses it across layers in that cache group;
- keeps the helper in `dsa_v1` so regular DSA, DSA-CP, and `Compressor.forward` share one cache;
- keys the cache by cache-group, metadata type, and input tensor identity so prefill/decode stay isolated (v0.26 used separate dataclasses; main unifies them into `AscendDSAReqMetadata`);
- explicitly resets the cache when a merged MTP/DSpark forward changes speculative substeps;
- keeps the metadata op inside model forward so ACLGraph captures each substep's real operator invocation.

Compressor kernel + FP32 RMSNorm at init is #15002.

No. This is an internal DSV4 performance optimization with unchanged cache geometry and model outputs.

- Unit regression covering same-phase reuse, prefill/decode isolation (including unified-type isolation), group isolation, substep reset, and forward-context lifetime.
- C4/C128 physical storage block geometry coverage on the shared helper.
- `bash format.sh ci` expected from PR CI.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/spec_decode/llm_base_proposer.py
+++ b/vllm_ascend/spec_decode/llm_base_proposer.py
@@ -1401,6 +1401,13 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
             forward_context.attn_metadata = (
                 multi_steps_attn_metadata[draft_index + 1] if multi_steps_attn_metadata else None
             )
+            if self.use_compress:
+                # A merged speculative forward reuses one ForwardContext while
+                # each substep has different compressor inputs. Drop the prior
+                # substep's outputs before entering the next model invocation.
+                from vllm_ascend.attention.dsa_v1 import reset_compressor_metadata_cache
+
+                reset_compressor_metadata_cache()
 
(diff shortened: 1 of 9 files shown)


vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -1401,6 +1401,15 @@ def _run_merged_draft(
             forward_context.attn_metadata = (
                 multi_steps_attn_metadata[draft_index + 1] if multi_steps_attn_metadata else None
             )
+            if self.use_compress:
```
lcfenglinwan, line 1404: no compressor in mtp layer

PR #14933, opened 2026-08-25
[Performance][MoE] Reuse MC2 expert scales for combine

GDzhu01:

MC2 dispatch converts `topk_weights` to FP32 before passing it as `expert_scales`. The original lower-precision tensor was then stored in combine metadata, so MC2 combine converted the same `[num_tokens, top_k]` values to FP32 a second time.

This change stores the already-converted dispatch tensor in the combine metadata when it is available. Combine therefore reuses the same FP32 tensor and avoids the duplicate Cast kernel. Non-hierarchical paths retain the existing behavior.

A unit test verifies that dispatch and combine receive the same FP32 tensor object.

No. This is an internal MoE performance optimization with no API or output behavior change.

- `git diff --check`
- `python -m py_compile vllm_ascend/ops/fused_moe/token_dispatcher.py tests/ut/ops/a2/test_token_dispatcher.py`
- Added `test_token_dispatch_reuses_fp32_expert_scales_for_combine`

NPU runtime tests and profiling comparison were not run locally because they require an Ascend environment. The expected profiling change is removal of the second `Cast [num_tokens, top_k]` immediately before `MoeDistributeCombineV2` when dispatch already supplies FP32 `expert_scales`.

- vLLM version: v0.26.0

### DIFF
--- a/tests/ut/ops/a2/test_token_dispatcher.py
+++ b/tests/ut/ops/a2/test_token_dispatcher.py
@@ -263,6 +263,34 @@ class TestTokenDispatcherWithMC2(TestBase):
             self.assertEqual(output.group_list_type, 0)  # group_list_type == 0
             self.assertIsInstance(output.combine_metadata, MoEMC2CombineMetadata)
 
+    def test_token_dispatch_reuses_fp32_expert_scales_for_combine(self):
+        hidden_states = torch.randn(10, 128, dtype=torch.bfloat16)
+        topk_weights = torch.randn(10, 1, dtype=torch.bfloat16)
+        topk_ids = torch.randint(0, 8, (10, 1))
+        expert_map = torch.tensor([0, 1, 2, 3, 4, 5, 6, 7])
+        self.dispatcher.enable_dispatch_v2 = True
+        self.dispatcher.need_expert_scale = True
+
+        with patch(
+            "torch_npu.npu_moe_distribute_dispatch_v2",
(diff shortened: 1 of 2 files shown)


tests/ut/ops/a2/test_token_dispatcher.py
```diff
@@ -263,6 +263,34 @@ def test_token_permutation_dispatch(self):
             self.assertEqual(output.group_list_type, 0)  # group_list_type == 0
             self.assertIsInstance(output.combine_metadata, MoEMC2CombineMetadata)
 
+    def test_token_dispatch_reuses_fp32_expert_scales_for_combine(self):
+        hidden_states = torch.randn(10, 128, dtype=torch.bfloat16)
+        topk_weights = torch.randn(10, 1, dtype=torch.bfloat16)
+        topk_ids = torch.randint(0, 8, (10, 1))
+        expert_map = torch.tensor([0, 1, 2, 3, 4, 5, 6, 7])
+        self.dispatcher.enable_dispatch_v2 = True
```
GDzhu01, line 271: [Testability] The production change is before the enable_dispatch_v2 branch, so it affects both npu_moe_distribute_dispatch and npu_moe_distribute_dispatch_v2. This regression test forces only V2, leaving legacy dispatch unprotected. Please parameterize enable_dispatch_v2 and patch or assert the matching operator for both branches; also keep a no-expert-scale case to verify fallback to the original topk_weights.

PR #14939, opened 2026-08-25
[Feature][Ops] Add custom GatherPaKvCache operator

yanhao5023:

- Ports every non-Ascend-950 GatherPaKvCache path from [ops-transformer at c157c4b](https://github.com/hicann/ops-transformer/tree/c157c4b561c4e5c92551038800ec5aeb124862ee/attention/gather_pa_kv_cache):
  - tiling key 619: Norm FP16/BF16
  - tiling key 618: Norm INT8
  - tiling key 577: PA_NZ INT8
- Registers `_C_ascend::gather_pa_kv_cache` as a direct AscendC custom operator, following the `batch_matmul_transpose` launch path without invoking ACLNN.
- Routes the A2/A3 device adaptor and Mooncake cache-loading call sites through the custom operator.
- Explicitly excludes Ascend 950. Ascend 310P remains on the existing fallback because the upstream kernel does not support that platform.

### Upstream code preservation

The non-950 kernel code is intentionally kept as close to upstream as the direct vLLM launch path permits:

- `gather_pa_kv_cache_nd.h` is source-equivalent to upstream apart from whitespace and line wrapping. Its class layout, initialization, UB tiling, core split, block-table lookup, ND data movement, cumulative-length handling, sequence-offset handling, dtype template, and first-axis cache-stride support are preserved.
- `gather_pa_kv_cache_nz.h` preserves the upstream class/process structure, UB layout, per-core loop, block-table lookup, PA_NZ copy layout, and the 4095-block split-copy path.
- The original 11-argument device-kernel ABI is preserved. The unused upstream `key`, `value`, and `workspace` slots are populated by the direct launcher without changing the non-950 kernel inputs/outputs.
- The ten upstream tiling fields retain their original names, types, order, and element-based units: `blockSize`, `numTokens`, `numblkTabCol`, `tokenSizeK`, `tokenSizeV`, `typeByte`, `hasSeqStarts`, `isSeqLensCumsum`, `kCacheBlockStride`, and `vCacheBlockStride`.
- The upstream dtype/path mapping is preserved: ND INT8 -> 618, ND FP16/BF16 -> 619, PA_NZ INT8 -> 577. `cache_mode` only selects between the original Norm and PA_NZ paths.

The NZ header is deliberately not described as byte-for-byte identical. It has four scoped adaptations required by the supported vLLM inputs:

- use the upstream tiling stride fields for first-dimension-strided PA_NZ cache views;
- reserve and copy the extra cumulative-sequence-length endpoint;
- initialize the first cumulative start before the first UB copy;
- index `seq_offset` by the global sequence and convert the token offset to a block-table offset.

The GE/gert host tiler and compile-time `GET_TILING_DATA` / `TILING_KEY_IS` / `ORIG_DTYPE_KEY_CACHE` channel cannot be used by a monolithic direct custom operator. They are replaced with an ATen host adapter plus one extra `tilingKey` field for runtime dispatch. The operator schema/meta registration, direct launcher, device-adaptor routing, CMake integration, and UT are vllm-ascend-specific integration code.

Yes. Supported non-Ascend-950 platforms now use the vLLM Ascend custom GatherPaKvCache operator. The operator exposes `cache_mode` and `is_seq_lens_cumsum` keyword arguments.

- Built `vllm_ascend_kernels` and `vllm_ascend_C` successfully on Ascend 910B1 with CANN 9.0 after the upstream-preservation revision.
- Added `tests/e2e/nightly/single_node/ops/singlecard_ops/test_gather_pa_kv_cache.py`.
- Ran the final standalone NPU suite: 11 passed.
- Coverage includes ND FP16/BF16/INT8, PA_NZ INT8, normal/cumulative sequence lengths, optional sequence offsets, first-dimension-strided caches, the PA_NZ >4095-block copy split, and validation errors.
- Verified codespell, typos, Python repository checks, Python byte compilation, and `git diff --check`. The remote pre-commit job also passed.

`bash format.sh ci` could not complete in this container because the markdownlint npm environment selected an incompatible Node runtime. The local gitleaks download also selected an incompatible binary, and shellcheck is not installed; the remaining configured checks passed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,16 +58,19 @@ set(VLLM_ASCEND_CUSTOM_OP
     ${KERNEL_FILES}
     ${CMAKE_CURRENT_SOURCE_DIR}/csrc/mla_preprocess/op_kernel/mla_preprocess_kernel.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/csrc/batch_matmul_transpose/op_kernel/batch_matmul_transpose_kernel.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/csrc/gather_pa_kv_cache/op_kernel/gather_pa_kv_cache_kernel.cpp
 )
 
 set(VLLM_ASCEND_CUSTOM_OP_EXCLUDE_ASCEND950
     ${CMAKE_CURRENT_SOURCE_DIR}/csrc/mla_preprocess/op_kernel/mla_preprocess_kernel.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/csrc/batch_matmul_transpose/op_kernel/batch_matmul_transpose_kernel.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/csrc/gather_pa_kv_cache/op_kernel/gather_pa_kv_cache_kernel.cpp
 )
 
 if(SOC_VERSION MATCHES "ascend950")
     message(STATUS "A5 hardware detected: disabling MLAPO operators")
     message(STATUS "A5 hardware detected: excluding batch_matmul_transpose operators")
+    message(STATUS "A5 hardware detected: excluding gather_pa_kv_cache operators")
     list(REMOVE_ITEM VLLM_ASCEND_CUSTOM_OP ${VLLM_ASCEND_CUSTOM_OP_EXCLUDE_ASCEND950})
 endif()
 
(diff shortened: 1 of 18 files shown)


yanhao5023:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

yanhao5023:
CI selected required tests, but ci-gate needs the ready-precise or ready-all label before they can run. The contributor account cannot add repository labels (GitHub API returned 403). Maintainers: please add ready-precise to trigger the selected tests; add ready-a5 as well only if the selected A5 tests should run. The final local Ascend 910B1 single-op suite passes: 11 passed.

PR #14941, opened 2026-08-25
[v0.26.0rc][Attention][Feature] Rewrite SFA DCP sparse index remap with Triton

coder-fny:

  Rewrite AscendSFADCPImpl._remap_sparse_indices (the DCP sparse-attention topk index remap, which runs once per SFA layer on the decode path) with Triton
  kernels:

  - New vllm_ascend/ops/triton/sparse_index_remap.py:
    - a fused kernel (vectorized fp32 remap + per-chunk compaction + output tail pre-fill, grid over (chunk, row));
    - a vectorized gather kernel that concatenates chunk results in top-k order;
  - vllm_ascend/attention/context_parallel/sfa_cp.py routes _remap_sparse_indices to the Triton path on NPU (HAS_TRITON and topk_indices.is_npu) and keeps the previous torch implementation as the fallback.
  - Extended tests/ut/attention/a2/test_sfa_cp_precision.py with an independent torch reference and new cases: interleave_size == 1, parametrized 2D cross-checks, and the real 3D DCP shape [dcp_size, 1, topk_count].

  Why: on GLM5.2 DCP decode the previous implementation dispatches ~30 small host-side ops per SFA layer per step (profiling:
  Cast/Floor/SelectV2/Sort/Gather chain with [6,1,2048] tensors), which contributes to the host-bound behavior on Ascend. The rewrite reduces this to 1-2 kernels.

  Measured on Ascend 910B (triton-ascend 3.2.0), standalone single-op benchmark vs the torch implementation:

image

  Numerical semantics are unchanged: the Triton kernels use the same fp32 index math as the previous implementation and match it bit-exactly (indices are far below the 2^24 fp32 exactness bound).

  Development notes for future maintainers (documented in the kernel comments): on triton-ascend 3.2.0 / CANN 9.1, tl.cumsum + tl.sum based compaction miscompiles (nondeterministic garbage values), tl.cumsum alone carries a large per-program fixed cost, an in-kernel literal -1 tail fill writes wrong values, and tl.sort is unavailable. The current design deliberately avoids all of these; re-validate if the triton-ascend version changes.

  No. Internal change to the DCP SFA decode path only; no API, config, or behavior change. The previous torch implementation is retained as the fallback for non-NPU / no-Triton environments.

  - Unit tests: extended tests/ut/attention/a2/test_sfa_cp_precision.py — existing cases plus interleave=1, parametrized 2D cross-checks (dcp_size 2/4 ×
  interleave 1/2/4 × topk 1/8/48/256 × all ranks) and the 3D [6,1,2048] case against an independent torch reference (requires NPU CI; not yet run on the v0.26 CI runner).
  - On-device single-op verification (Ascend 910B, torch 2.10 + torch_npu 2.10.0.post4 + triton-ascend 3.2.0): correctness 144/144 bit-exact vs the torch reference including -1 padding and out-of-bounds lanes; 10x repeated runs on the previously failing input patterns; 200 back-to-back unsynchronized launches all bit-exact (regression guard against the miscompile class found during development).
  - End-to-end model precision on a DCP>1 environment is still pending and will be done separately.
 
before：
image
after：
image

- vLLM version: v0.26.0


vllm_ascend/ops/triton/sparse_index_remap.py
AuroraEmiya, line 1: You may need to add a md for it.

vllm_ascend/ops/triton/sparse_index_remap.py
```diff
@@ -0,0 +1,200 @@
...
+    DCP-local index and the owner-validity mask for this chunk; then the
+    chunk's valid entries are compacted to the front of its ``chunk_out``
+    slot in order via scalar ``get_element`` operations, the chunk's valid
+    count is written to ``chunk_count``, and the chunk's region of ``out``
+    is pre-filled with -1 (the gather kernel later overwrites the compacted
+    front).
+
+    NOTE: keep the compaction serial. ``tl.cumsum``/``tl.sum`` based
+    compaction miscompiles or carries a large fixed cost on triton-ascend
+    3.2.0 / CANN 9.1 on 910B (verified); scalar ops are the reliable
```
AuroraEmiya: Could your optimization apply for all general devices like a2 ,a3 or a5? Or it just a specific situation?

vllm_ascend/ops/triton/sparse_index_remap.py
```diff
@@ -0,0 +1,200 @@
...
+    chunk = tl.program_id(0)
+    row = tl.program_id(1)
+    offsets = chunk * BLOCK + tl.arange(0, BLOCK)
+    in_bounds = offsets < topk_count
+
+    idx = tl.load(indices_ptr + row * topk_count + offsets, mask=in_bounds, other=-1)
+    idx_f = idx.to(tl.float32)
+    il_f = interleave_size.to(tl.float32)
+    dcp_size_f = dcp_size.to(tl.float32)
+    dcp_rank_f = dcp_rank.to(tl.float32)
```
AuroraEmiya: I have no idea of the  point of changing all the type into f32. Which might cause unnecessary workload.

vllm_ascend/ops/triton/sparse_index_remap.py
```diff
@@ -0,0 +1,200 @@
...
+        for i in range(BLOCK):
+            if get_element(valid_i32, (i,)) == 1:
+                tl.store(out_ptr + row * topk_count + cnt, get_element(remapped, (i,)))
+                cnt += 1
+        tl.store(chunk_count_ptr + row * NUM_CHUNKS + chunk, cnt)
+    else:
+        for i in range(BLOCK):
+            if get_element(valid_i32, (i,)) == 1:
+                tl.store(chunk_out + cnt, get_element(remapped, (i,)))
+                cnt += 1
```
AuroraEmiya: The scalar get_element loop here may serialize the compaction. Could we try a vectorized compaction, e.g. computing destination offsets from a prefix sum of valid_i32 and then using a masked tl.store? It may reduce the scalar/control overhead, especially for larger BLOCK; a benchmark would be useful to confirm.

vllm_ascend/ops/triton/sparse_index_remap.py
```diff
@@ -0,0 +1,200 @@
...
+        remapped = remapped_f.to(tl.int32)
+
+    valid_i32 = valid.to(tl.int32)
+    chunk_out = chunk_out_ptr + (row * NUM_CHUNKS + chunk) * BLOCK
+    # Pre-fill this chunk's output region with -1 first; the serial stores
+    # below overwrite the compacted front in program order.
+    tl.store(out_ptr + row * topk_count + offsets, -1, mask=in_bounds)
+    cnt = 0
+    if NUM_CHUNKS == 1:
+        # Fast path: with a single chunk the row order is already final, so
```
AuroraEmiya: It's that necessary for this fast path in real models? Which may make this code unreadable and specific.

vllm_ascend/ops/triton/sparse_index_remap.py
coder-fny: Thanks. I benchmarked both versions on-device and they perform the same (the kernel is dominated by the compaction loop). The fp32 cast was only for bit-exact parity with the torch fallback; since the int32 variant is also bit-exact, I'll change it back to int32. commit: [7283d7a](https://github.com/vllm-project/vllm-ascend/pull/14941/commits/7283d7ad553a3d9742c2b9b40ef187a22388ea53)

vllm_ascend/ops/triton/sparse_index_remap.py
coder-fny: Thanks for the suggestion — I re-verified the vectorized compaction (exclusive prefix sum + masked tl.store) on triton-ascend 3.2.0. Single-shot results are 105/105 bit-exact vs the torch reference, but under repeated launches the generated kernel deterministically traps the vector core (aivec timeout,  all cores stuck at the same PC) and kills the process — reproduced twice at the same config, consistent with the earlier 910B garbage-value finding from tl.cumsum. The one timing point we could collect shows no measurable difference, so we keep the bounded serial loop for now and will re-validate once triton-ascend is upgraded.

vllm_ascend/ops/triton/sparse_index_remap.py
coder-fny: Agreed — real models (topk=2048) always span multiple chunks, so the fast path was dead code in practice. Removed the NUM_CHUNKS == 1 branch; the fused kernel now always writes to chunk_out and the gather kernel always runs. Unit tests pass 30/30, and performance is unchanged for topk >= 512 (1.00x); only small topk (<= 128) pays one extra gather launch (~0.06ms). commit: [e31a260](https://github.com/vllm-project/vllm-ascend/pull/14941/commits/e31a260cea06402746c345816691d203e5f69c59)

vllm_ascend/ops/triton/sparse_index_remap.py
coder-fny: They're plain Triton kernels gated by HAS_TRITON and is_npu — no device-specific code, so they apply wherever triton-ascend runs. Verified on a2 and a3 (bit-exact vs the torch reference, ~2.2x faster); a5 is not validated yet.

vllm_ascend/ops/triton/sparse_index_remap.py
coder-fny: ok, i will commit soon

vllm_ascend/ops/triton/sparse_index_remap.py
AuroraEmiya: If so , clean those redundant comments and notes would make it clean and readable : )

coder-fny: done

vllm_ascend/ops/triton/sparse_index_remap.py
coder-fny: #15197 Opened a cleaner PR against main — please re-review when you have time. Thanks!

PR #14952, opened 2026-08-25
[v0.26.0rc][BugFix][Frontend] Select DeepSeek V4 effort mapping by checkpoint config

### LINKED ISSUE #14951: [Bug]: DeepSeek V4 Preview checkpoints receive post-Preview reasoning-effort prompts
### Your current environment

- vLLM Ascend branch: `releases/v0.26.0rc` at `820a21d6593b74211863c7787e1ba9b4deaf9126`
- vLLM version: `v0.26.0`
- Regression introduced by #13993 and still present after #14624
- Reproduction is tokenizer-only; the user-visible impact was found by the DeepSeek V4 Flash GPQA smoke test

### Describe the bug

The DeepSeek V4 thinking patch added by #13993 unconditionally applies the post-Preview reasoning-effort mapping to every DeepSeek V4 checkpoint.

The official DeepSeek V4 Preview encoders and the encoders shipped with later checkpoints such as DeepSeek-V4-Flash-0731 and DeepSeek-V4-Pro-0813 use different meanings for the same effort names:

| Requested effort | V4 Preview prompt | Post-Preview prompt |
|---|---|---|
| `high` (also the API default) | no effort prefix | `Reasoning Effort: Absolute maximum...` |
| `max` / `xhigh` | `Reasoning Effort: Absolute maximum...` | `max` uses `Reasoning Effort: Beyond maximum...`; `xhigh` maps to post-Preview high |

After #13993, a V4 Preview request with omitted effort or explicit `high` therefore receives the prompt that its own encoder associates with maximum effort. In the GPQA smoke test with thinking enabled and `max_tokens=131072`, this produces many truncated answers and lowers accuracy.

This is a checkpoint compatibility regression in the reasoning-effort patch. It is separate from the system/tools placement bug reported in https://github.com/vllm-project/vllm/issues/51829.

A tokenizer-only probe on the current release branch shows the incorrect behavior before the fix:

```text
DeepSeek-V4-Flash (Preview):
  default: absolute-maximum
  high: absolute-maximum
  max: beyond-maximum
```

The checkpoint generations can be distinguished from `config.json` without relying on model path names:

- DeepSeek-V4-Flash and DeepSeek-V4-Pro Preview have no `dspark_*` fields.
- DeepSeek-V4-Flash-0731 and DeepSeek-V4-Pro-0813 contain `dspark_block_size`, `dspark_markov_rank`, `dspark_noise_token_id`, and `dspark_target_layer_ids`.

Expected behavior:

1. If `config.json` contains any `dspark_*` field, use the post-Preview mapping added by #13993.
2. Otherwise, preserve the V4 Preview mapping: default/high has no effort prefix, while max/xhigh uses the original `Absolute maximum` prefix.
3. Explicit thinking disablement must continue to override the requested effort.

Regression coverage should verify both mapping tables and the exact prompt prefix for Preview default/high/max and post-Preview high/max.

QwertyJack:

Fixes #14951.

#13993 added the post-Preview DeepSeek V4 reasoning-effort mapping to the v0.26 release stack, but applied it to every DeepSeek V4 checkpoint. The official V4 Preview encoders use different semantics: default/high has no effort prefix, while max/xhigh uses the original `Absolute maximum` prefix. Applying the post-Preview mapping makes a Preview model interpret the API default `high` as its maximum effort, causing many GPQA responses to exhaust `max_tokens=131072` and be truncated.

This PR reads the checkpoint `config.json` when wrapping the tokenizer. Checkpoints containing a `dspark_*` field, including DeepSeek-V4-Flash-0731 and DeepSeek-V4-Pro-0813, use the post-Preview `low/high/max` mapping. Checkpoints without those fields use the V4 Preview compatibility mapping. Explicit thinking disablement continues to override the requested effort.

The system/tools placement problem in https://github.com/vllm-project/vllm/issues/51829 is separate and remains covered by #14624.

Yes. On the v0.26 release stack, DeepSeek V4 Preview checkpoints no longer receive post-Preview high/max prompts. DeepSeek-V4-Flash-0731 and DeepSeek-V4-Pro-0813 behavior remains unchanged.

```bash
PYTHONPATH=/path/to/vllm-v0.26.0 python -m pytest -q \
  tests/ut/patch/platform/test_deepseek_v4_thinking.py \
  tests/ut/patch/platform/test_deepseek_v4_tool_streaming.py
# 61 passed

ruff check \
  tests/ut/patch/platform/test_deepseek_v4_thinking.py \
  vllm_ascend/patch/__init__.py \
  vllm_ascend/patch/platform/patch_deepseek_v4_thinking.py
# All checks passed!

ruff format --check \
  tests/ut/patch/platform/test_deepseek_v4_thinking.py \
  vllm_ascend/patch/__init__.py \
  vllm_ascend/patch/platform/patch_deepseek_v4_thinking.py
# 3 files already formatted
```

A tokenizer-only probe using local checkpoint directories produced:

```text
DeepSeek-V4-Flash (Preview)
  default: no effort prefix
  high: no effort prefix
  max: Absolute maximum
DeepSeek-V4-Flash-0731
  default: Absolute maximum
  high: Absolute maximum
  max: Beyond maximum
DeepSeek-V4-Pro-0813
  default: Absolute maximum
  high: Absolute maximum
  max: Beyond maximum
```

The Flash-0731 and Pro-0813 `config.json` files both contain `dspark_*` fields, while the Flash and Pro Preview configs do not. `python -m py_compile` on all changed Python files and `git diff --check` also passed. No NPU service was required because the regression is fully determined during prompt rendering.

- vLLM version: v0.26.0

### DIFF
--- a/vllm_ascend/patch/platform/patch_deepseek_v4_thinking.py
+++ b/vllm_ascend/patch/platform/patch_deepseek_v4_thinking.py
@@ -20,6 +20,7 @@ from typing import Any
 from vllm.entrypoints.chat_utils import ChatCompletionMessageParam
 from vllm.parser.deepseek_v4 import DeepSeekV4Parser
 from vllm.tokenizers import deepseek_v4, deepseek_v4_encoding
+from vllm.transformers_utils.repo_utils import get_hf_file_to_dict
 
 REASONING_EFFORT_PROMPTS = {
     "low": "",
@@ -52,6 +53,15 @@ _original_get_deepseek_v4_tokenizer = deepseek_v4.get_deepseek_v4_tokenizer
 _original_deepseek_v4_parser_init = DeepSeekV4Parser.__init__
 
 
+def _uses_preview_reasoning_effort_mapping(tokenizer: deepseek_v4.HfTokenizer) -> bool:
+    model_name_or_path = getattr(tokenizer, "name_or_path", None)
+    if not model_name_or_path:
+        return True
+
+    config = get_hf_file_to_dict("config.json", model_name_or_path)
+    return not (config and any(key.startswith("dspark_") for key in config))
+
+
 def _patched_render_message(
     index: int,
     messages: list[dict[str, Any]],
@@ -78,6 +88,7 @@ def _patched_render_message(
 
 
 def _patched_get_deepseek_v4_tokenizer(tokenizer: deepseek_v4.HfTokenizer):
+    uses_preview_mapping = _uses_preview_reasoning_effort_mapping(tokenizer)
     dsv4_tokenizer = _original_get_deepseek_v4_tokenizer(tokenizer)
     tokenizer_cls = type(dsv4_tokenizer)
 
@@ -110,16 +121,24 @@ def _patched_get_deepseek_v4_tokenizer(tokenizer: deepseek_v4.HfTokenizer):
 
         reasoning_effort = kwargs.get("reasoning_effort")
         if not isinstance(reasoning_effort, str):
-            reasoning_effort = "high" if thinking_enabled else None
+            if thinking_enabled:
+                reasoning_effort = "low" if uses_preview_mapping else "high"
+            else:
+                reasoning_effort = None
         elif reasoning_effort == "none":
             thinking_mode = "chat"
             reasoning_effort = None
-        elif reasoning_effort == "max":
-            reasoning_effort = "max"
-        elif reasoning_effort in ("low", "minimal", "medium"):
-            reasoning_effort = "low"
-        else:
+        elif not uses_preview_mapping:
+            if reasoning_effort == "max":
+                reasoning_effort = "max"
+            elif reasoning_effort in ("low", "minimal", "medium"):
+                reasoning_effort = "low"
+            else:
+                reasoning_effort = "high"
+        elif reasoning_effort in ("max", "xhigh"):
             reasoning_effort = "high"
+        else:
+            reasoning_effort = "low"
 
         prompt_str = deepseek_v4.encode_messages(
             messages,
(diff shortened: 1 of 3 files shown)


vllm_ascend/patch/platform/patch_deepseek_v4_thinking.py
```diff
@@ -52,6 +53,15 @@
 _original_deepseek_v4_parser_init = DeepSeekV4Parser.__init__
 
 
+def _uses_0731_reasoning_effort_mapping(tokenizer: deepseek_v4.HfTokenizer) -> bool:
+    model_name_or_path = getattr(tokenizer, "name_or_path", None)
+    if not model_name_or_path:
+        return False
+
+    config = get_hf_file_to_dict("config.json", model_name_or_path)
+    return bool(config and any(key.startswith("dspark_") for key in config))
```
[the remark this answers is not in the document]
QwertyJack: Thanks for flagging this. `get_hf_file_to_dict` already handles offline mode, missing repositories/revisions/files, local-cache misses, and Hugging Face HTTP errors by returning `None`. A broad `except Exception` here would additionally hide malformed local `config.json` and other programming errors, then silently select the Preview mapping. I am keeping those errors visible. I did rename the predicate to `_uses_preview_reasoning_effort_mapping` because the post-Preview mapping is also used by DeepSeek-V4-Pro-0813, not only Flash-0731.

vllm_ascend/patch/platform/patch_deepseek_v4_thinking.py
```diff
@@ -52,6 +53,15 @@
 _original_deepseek_v4_parser_init = DeepSeekV4Parser.__init__
 
 
+def _uses_preview_reasoning_effort_mapping(tokenizer: deepseek_v4.HfTokenizer) -> bool:
```
jianzs, line 56: If it's the preview version, just skip applying the `_patched_get_deepseek_v4_tokenizer` patch—will it work with the preview version?

QwertyJack: It would not work. `_patched_render_message` is installed globally, so the original tokenizer would still use the post-Preview mapping: `high` -> `Absolute maximum` and `max` -> `Beyond maximum`. The wrapper is also needed to keep omitted thinking options aligned with the parser defaults. The Preview checkpoint therefore still needs the wrapper, with the Preview mapping selected inside it.

PR #14961, opened 2026-08-26
[Performance][v0.26.0rc] Preserve top-k weights to avoid redundant casts

dragondream-chen:
This PR removes a redundant dtype round trip for MoE top-k routing weights in the W8A8 Dynamic MC2 path.
MoeGatingTopKHash produces FP32 routing weights, but they were previously converted to the model dtype (BF16) before MC2 dispatch and then converted back to FP32 before MoeDistributeCombineV2, which requires FP32 expert scales:
FP32 → BF16 → FP32
This introduced two unnecessary cast kernels and caused avoidable precision loss.
no


dragondream-chen:
/rerun

[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

dragondream-chen:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #14992, opened 2026-08-26
[v0.26.0rc][bugfix] correct _is_decode_only_node check so megaMoE buffer is sized by num_batched_tokens

coder-fny:
The original implementation returned
  `not bool(recompute_scheduler_enable)`. The `not` inversion made the result
  contradict the semantics above: the case where the decode node executes prefill
  locally was misclassified as decode-only, so the megaMoE buffer size no longer
  matched the prefill load the node actually handles.

  | recompute_scheduler_enable | Actual behavior | Old result | Fixed result |
  |---|---|---|---|
  | true | Preemption sends requests back to the P node to redo prefill; decode node is decode-only | not decode-only ❌ | decode-only ✅ |
  | false | Prefill runs locally on the decode node | decode-only ❌ | not decode-only ✅ |

- vLLM version: v0.26.0


kunpengW-code:
The UT has been completed. The fused_mc2 is not enabled in the e2e test case and will be incorporated first.

PR #15001, opened 2026-08-26
[Feature][v0.26.0rc] Sync DeepSeek V4 compressor operator

maoxx241:

- Rebases onto latest `releases/v0.26.0rc` (`85625bd77`, includes #14994 compressor metadata reuse).
- Syncs only the DeepSeek V4 **compressor** kernel from `ops-transformer/experimental/attention/compressor`.
- Creates compressor RMSNorm in float32 at module init. The new compressor kernel only accepts FP32 `norm_weight`.
- Drops the quant lightning indexer / QLI-metadata operator updates. Isolation on A3 showed QLI-only reproduces the full-network GPQA collapse; compressor-only does not. End-to-end 8k/32k/128k A/B also showed no QLI v2 win over compressor-only.
- Does not include `scatter_nd_update_asc`; DSV4 continues to use `scatter_nd_update_v2`.

No public Python or CLI API change. The bundled DeepSeek V4 compressor custom operator is updated.

- Isolation GPQA-Diamond frozen 107 (same IDs as the 93/107=86.92% vanilla 0.26.0rc baseline), A3 DP4xTP4 EP MTP1 `FULL_DECODE_ONLY`:
  - Combo (compressor + QLI): 61/107 = 57.01%.
  - QLI-only: 62/107 = 57.94%.
  - Compressor-only (this PR): 96/107 = 89.72%, 0 errors.
- 8k/32k/128k in/1k out, single and multi concurrency, ~95% prefix cache for ≥16k: compressor-only matches or slightly beats vanilla; QLI v2 has no e2e advantage.
- Source/layout check: PR diff vs `releases/v0.26.0rc` contains only `csrc/attention/compressor/**` plus the FP32 RMSNorm dtype at compressor init.

- vLLM version: v0.26.0


vllm_ascend/device/device_op.py
```diff
@@ -702,14 +701,26 @@ def get_dsa_sparse_attn_base_kwargs():
...
+            x.view(-1, head_dim),
+        )
 
     # ===== SWA / Compressor KV Scatter =====
 
     @staticmethod
     def dsa_kv_compress_scatter(cache, x, slot_mapping):
         """Scatter KV into cache. Non-A5: simple scatter of pre-quantized tensor."""
-        torch.ops._C_ascend.npu_scatter_nd_update_v2(cache, slot_mapping, x)
+        DeviceOperator._scatter_nd_update_asc(cache, x, slot_mapping)
```
lcfenglinwan: It can be directly called using `DeviceOperator._scatter_nd_update_asc`.

vllm_ascend/device/device_op.py
```diff
@@ -843,8 +854,8 @@ def pad_dsa_decode_slot_mapping(slot_mapping, num_decode_tokens, compress_ratio,
     @staticmethod
     def format_dsa_slot_mapping(slot_mapping, block_size):
```
lcfenglinwan, line 844: this can be remove because the implementation is the same

maoxx241:
## QLI v2 landed on this PR + A3 8k/1k perf A/B

Pushed `04003d906` (`feat(ops): vendor CANN QuantLightningIndexer v2 for DSV4`).

### Accuracy (already collected)
GPQA-Diamond frozen 107, DeepSeek-V4-Flash-w8a8-mtp, DP4xTP4, EP, MTP=1, `FULL_DECODE_ONLY`: **94/107** (acc 87.85%, missing 7, errors 0). Non-regression vs the ~93/107 baseline; compressor-only on this PR was 96/107.

### Perf protocol (machine 154, A3)
Same serve flags as GPQA (`max-num-batched-tokens=10240`, `max-num-seqs=64`, DP4xTP4, MTP=1). Exact **8192-in / 1024-out** token-id prompts, `ignore_eos`, seeded unique prompts (no prefix-cache hit). Warmup 8k/64, then **c=1 n=4** and **c=16 n=16**. All requests finished `length` with exact 8192/1024 tokens.

Fair pair is two **fresh** serves: `v026-src` (this PR before QLI v2, in-tree QLI) vs `v026-rebuild` (QLI v2).

| conc | metric | before (in-tree QLI) | after (QLI v2, fresh) | delta |
|---:|---|---:|---:|---:|
| 1 | e2e p50 | 21.32 s | 24.69 s | **+15.8%** |
| 1 | TTFT p50 | 1.22 s | 1.39 s | **+13.6%** |
| 1 | TPOT p50 | 19.64 ms | 22.82 ms | **+16.2%** |
| 1 | output tok/s | 47.82 | 41.61 | **-13.0%** |
| 16 | e2e p50 | 28.64 s | 31.88 s | **+11.3%** |
| 16 | TTFT p50 | 5.02 s | 5.52 s | **+9.9%** |
| 16 | TPOT p50 | 23.40 ms | 26.41 ms | **+12.9%** |
| 16 | output tok/s | 513.32 | 486.53 | **-5.2%** |

**Result: this is a regression, not a speedup**, on both single-concurrency and c=16. Decode TPOT is the main gap; prefill TTFT is also worse.

A3 QLI v2 tiling rejected non-C-contiguous K/scale (including axis 0), so the Python path currently does `.contiguous()` on indexer K and key_scale every QLI call. That is a suspect extra copy vs the old in-tree QLI, which allowed non-contiguous axis 0. Not proven as the whole gap.

JSON: `/workspace/dsv4-ops-154/perf-8k1k/{before,after-fresh,after}.json` on 154.

maoxx241:
## Follow-up: PA axis-0 non-contiguous on A3 (`e3c537436`)

A3 QLI v2 was going through the arch22 kernel, which assumed a compact block stride and forced a `.contiguous()` copy of indexer K/scale. Ported the in-tree QLI `stride` / `scaleStride` addressing and allowed PA_BBND axis-0 non-contiguous in tiling.

Same 8k-in / 1k-out protocol, fresh serves, DP4xTP4.

| conc | metric | compressor-only | QLI v2 + `.contiguous()` | QLI v2 + axis-0 stride |
| ---: | --- | ---: | ---: | ---: |
| 1 | e2e p50 | 21.12 s | 24.69 s | **21.56 s** |
| 1 | TPOT p50 | 19.49 ms | 22.82 ms | **19.76 ms** |
| 1 | output tok/s | 47.54 | 41.61 | **47.76** |
| 16 | e2e p50 | 28.23 s | 31.88 s | **28.53 s** |
| 16 | TPOT p50 | 23.05 ms | 26.41 ms | **23.17 ms** |
| 16 | output tok/s | 561.4 | 486.5 | **541.9** |

vs the contiguous QLI v2 path this is **~13% better decode TPOT** and **+12–15% output tok/s**. vs compressor-only (in-tree QLI), decode TPOT is flat; TTFT is still ~8–10% higher.

PR #15031, opened 2026-08-26
[BugFix][Spec Decode] Skip unused MTP draft outputs on prefill-only workers

ader47:

Pure KV-producer workers still need the single-step MTP forward pass to populate the draft-layer KV cache, but they do not consume MTP draft tokens or draft logits.

During dummy execution, the synthetic draft batch can be much larger than max_num_seqs. Computing and TP-all-gathering its full-vocabulary logits can therefore request several GiB of unnecessary HBM.

For the observed GLM-5.2 failure with BF16, TP16, 32,768 dummy tokens, and 16,384 synthetic sample rows, the early return avoids these per-rank allocations:

| Avoided allocation | Approximate HBM per rank |
| --- | ---: |
| MTP hidden-state sequence-parallel all-gather | 384 MiB |
| Sample hidden states | 192 MiB |
| Local vocabulary logits | 302.5 MiB |
| Full-vocabulary logits TP all-gather | 4.73 GiB |
| Position all-gather | 0.25 MiB |
| Total direct allocations | about 5.58 GiB |

The exact saving scales with dummy token count, vocabulary size, dtype, and parallel configuration. The 4.73 GiB full-vocabulary logits allocation was the immediate OOM request in this failure.

This PR:

- keeps the single-step MTP forward pass on pure KV producers so draft KV caches are populated;
- returns before sequence-parallel output gathering, LM Head, logits gathering, and draft sampling;
- applies the same behavior to the Step3.5 proposer;
- converts the empty draft result to None so no draft-token CPU copy is scheduled.

The optimization is restricted to MTP with one speculative token on a producer-only PD worker. Multi-step MTP and decode workers keep their existing behavior.

No API or configuration change. Pure prefill workers use less peak HBM and avoid computing unused MTP draft logits.

- Added a unit regression that verifies the MTP forward still runs while output all-gather is skipped on a pure prefill worker.
- Added a Step3.5 source-contract regression for the same early-return behavior.
- Ran the source-level regression tests successfully.
- Ran bash format.sh ci successfully; all checks passed.
- The local host has no Ascend NPU runtime, so full GLM-5.2 TP16 startup validation is left to CI and NPU hardware testing.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/spec_decode/llm_base_proposer.py
+++ b/vllm_ascend/spec_decode/llm_base_proposer.py
@@ -539,6 +539,9 @@ class AscendSpecDecodeBaseProposer(SpecDecodeBaseProposer):
             return self.model.unwrap()
         return self.model
 
+    def _is_prefill_only_mtp(self) -> bool:
+        return self.method == "mtp" and self.num_speculative_tokens == 1 and self.runner._is_pd_prefill_worker() is True
+
     def shallow_copy_metadata(self, attn_metadata):
         # Currently, new objects will be assigned to the lists in attn_metadata
         # when update. So we can use the shallow copy.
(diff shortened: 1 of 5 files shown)


vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -539,6 +539,9 @@ def get_model(self) -> nn.Module:
             return self.model.unwrap()
         return self.model
 
+    def _is_prefill_only_mtp(self) -> bool:
+        return self.method == "mtp" and self.num_speculative_tokens == 1 and self.runner._is_pd_prefill_worker() is True
```
lijiahang226, line 543: Please consider removing `and self.num_speculative_tokens == 1`

ader47: This condition is intentional to keep the fix scoped to the validated single-step MTP case. With K=1, the MTP forward count remains unchanged and we only skip unused output processing. For K>1, removing it would reduce multiple per-step draft forwards to one and may skip required per-step/layer KV-cache side effects, particularly for Step3.5. I would prefer to keep it until multi-step MTP is verified separately.

ader47:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #10276, opened 2026-06-10
[Attention][BugFix] Fix NaN issue in Lightning Attention NPU kernel caused by unmasked padding

ghphotoframe:
### What this PR does / why we need it?

This PR fixes a NaN issue in the Lightning Attention NPU kernel on Ascend hardware. On Ascend, `tl.load(..., other=0.0)` does not reliably clear out-of-bound data due to hardware-specific vector-to-cube loading behavior. When the sequence length is not a multiple of `BLOCK_SIZE`, trailing blocks contain garbage values, leading to NaNs during dot-product computation and model instability.

To resolve this, we explicitly apply `tl.where` masking for query (`q`) and key (`k`) tensors after loading to zero out padding elements in the last block. Additionally, redundant debug logging in `lightning_attention_npu` is removed to reduce runtime overhead.

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

Tested on Atlas 910C with model `ling-2.6-flash`. Verified that models no longer produce NaNs when sequence lengths are not aligned with `BLOCK_SIZE`, and attention outputs match expected numerical ranges.
image

---
testcase:
```
====================================================================================== test session starts =======================================================================================
platform linux -- Python 3.11.15, pytest-8.3.2, pluggy-1.6.0 -- /usr/local/python3.11.15/bin/python3.11
cachedir: .pytest_cache
rootdir: /a3_inference/itask/workdir/jwx02384838/40m3hh8ksyh20pw3/code/gemma_script/code/vllm-ascend
configfile: pyproject.toml
plugins: xdist-3.6.1, anyio-4.13.0
collected 29 items                                                                                                                                                                               

code/vllm-ascend/tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_lightning_attn.py::test_lightning_attention_npu_single_chunk[b1-h4-n32-d128-e64-bfloat16] [13:56:10]
PASSED
code/vllm-ascend/tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_lightning_attn.py::test_lightning_attention_npu_single_chunk[b1-h4-n100-d128-e128-bfloat16] [13:56:10]
PASSED
code/vllm-ascend/tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_lightning_attn.py::test_lightning_attention_npu_single_chunk[b1-h4-n256-d128-e128-float16] [13:56:11]
PASSED
[...]

PR #15051, opened 2026-08-27
[Doc][Ops] Add the lightning_attn operator doc

MrMC-Sqz:
## Why

The four Triton kernels of `vllm_ascend/ops/triton/mamba/lightning_attn.py` —
`_fwd_diag_kernel`, `_fwd_kv_parallel`, `_fwd_kv_reduce`, `_fwd_none_diag_kernel` —
have no spec of any kind: no formula, no parameter list, no record of the
constraints a caller has to respect. Reading the four of them together is
currently the only way to find out what the operator computes.

## What

Adds `vllm_ascend/ops/triton/docs/lightning_attn.md`. One document for all four
kernels, because they have no individual dispatch site — they only ever run as one
pipeline inside `_attention.forward`.

Contents:

- **Description**: what the operator computes, the closed-form formula of the tiled
  algorithm (intra-block and cross-block terms, and the KV state update), a
  step-by-step table of what each kernel contributes, the dispatch chain from
  `AscendBailingMoELinearAttention` down to the kernels, and the grid layout of each
  kernel.
- **Parameters**: separate tables for the two Python entry points
  (`jit_linear_forward_prefix`, `lightning_attention_npu`) and for each kernel, with
  `tl.constexpr` versus runtime parameters marked.
- **Constraints**: see the notes below.
- **Test notes**: per kernel, what a case has to check to lock that kernel down;
  plus what is deliberately not testable today and why.

## Notes

Findings from reading the kernels, **recorded in the doc but not changed here**, so
the owner can decide whether any deserves its own fix PR:

1. Cross-block token pairs are weighted `exp(-s*(t-j-1))` while pairs inside one
   BLOCK get `exp(-s*(t-j))`. `_fwd_kv_parallel` decays a block state to the last
   token of its block and `_fwd_none_diag_kernel` replays it with `exp(-s*t_local)`,
   so the two paths disagree by `exp(s)` for the same token distance. This matches
   the upstream vLLM / MiniMax GPU implementation, so it is recorded as a constraint
   rather than changed. A consequence worth knowing: segmented prefill is only
   equivalent to one-shot prefill on a 256-aligned split.
2. `d > 128` is broken: `lightning_attention_npu` chunks `d` by 128 but hands each
   chunk the whole `[b, h, d, e]` `kv_history`, which the kernel then addresses with
   `d = 128`, and it returns the `kv` of the last chunk only. BailingMoE uses
   `head_dim = 128` and never reaches the path.
3. `block_size` is accepted by both public entry points and silently ignored;
   `_attention.forward` pins `BLOCK = 256`.

Also recorded: `_attention` has no `backward` despite subclassing
`autograd.Function`; `_fwd_kv_parallel` relies on the masked load alone where #10276
added a `tl.where` re-mask to `_fwd_diag_kernel`; the returned `kv` is an *exclusive*
scan, so entry 0 is the incoming history and the final state is the trailing entry;
and no parameter carries `do_not_specialize` while `b/h/n/d/e` are all `constexpr`,
so every new shape costs four recompilations.

The test section documents the existing `test_lightning_attn.py` as it stands — its
13 test functions and 29 cases, its tolerance table, and which kernels each group of
cases reaches — so the doc and the test file agree. No test change in this PR.

Note on placement: the doc sits under `vllm_ascend/ops/triton/docs/`, the same
directory as the other Triton operator docs. Its section layout follows
`csrc/attention/kv_quant_sparse_attn_sharedkv/README.md`, with an extra
`Supported Products` and `Change Log` section.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/ops/triton/docs/lightning_attn.md
+++ b/vllm_ascend/ops/triton/docs/lightning_attn.md
@@ -10,11 +10,11 @@ They have no individual entry point — they only ever run as one pipeline dispa
 | Product | Supported |
 | ------------------------------------------------------------ | :------: |
 |<term>Ascend 950PR/Ascend 950DT</term>|      √     |
-|<term>Atlas A3 训练系列产品/Atlas A3 推理系列产品</term>|      √     |
-|<term>Atlas A2 训练系列产品/Atlas A2 推理系列产品</term>|      √     |
-|<term>Atlas 200I/500 A2 推理产品</term>|      ×     |
-|<term>Atlas 推理系列加速卡产品</term>|      ×     |
-|<term>Atlas 训练系列产品</term>|      ×     |
+|<term>Atlas A3 Training Series Product/Atlas A3 Inference Series Product</term>|      √     |
+|<term>Atlas A2 Training Series Product/Atlas A2 Inference Series Product</term>|      √     |
+|<term>Atlas 200I/500 A2 Inference Product</term>|      ×     |
+|<term>Atlas Inference Series Accelerator Card Product</term>|      ×     |
+|<term>Atlas Training Series Product</term>|      ×     |
 
 > The operator does not go through `DeviceOperator`: A2/A3 and Ascend 950 receive
 > identical launch parameters and there is no adaptor branch.
--- a/vllm_ascend/ops/triton/doc/lightning_attn.md
+++ b/vllm_ascend/ops/triton/doc/lightning_attn.md
@@ -1,0 +1,296 @@
+# LightningAttention
+
+This document covers the four Triton kernels in `vllm_ascend/ops/triton/mamba/lightning_attn.py`:
+`_fwd_diag_kernel`, `_fwd_kv_parallel`, `_fwd_kv_reduce` and `_fwd_none_diag_kernel`.
+They have no individual entry point — they only ever run as one pipeline dispatched by
+`_attention.forward` — so they share a single document.
+
+## Supported Products
+
+| Product | Supported |
+| ------------------------------------------------------------ | :------: |
+|<term>Ascend 950PR/Ascend 950DT</term>|      √     |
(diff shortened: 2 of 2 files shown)


tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_lightning_attn.py
```diff
@@ -1,49 +1,26 @@
+# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
MrMC-Sqz: 已解决

vllm_ascend/ops/triton/docs/lightning_attn.md
```diff
@@ -0,0 +1,296 @@
...
+This document covers the four Triton kernels in `vllm_ascend/ops/triton/mamba/lightning_attn.py`:
+`_fwd_diag_kernel`, `_fwd_kv_parallel`, `_fwd_kv_reduce` and `_fwd_none_diag_kernel`.
+They have no individual entry point — they only ever run as one pipeline dispatched by
+`_attention.forward` — so they share a single document.
+
+## Supported Products
+
+| Product | Supported |
+| ------------------------------------------------------------ | :------: |
+|<term>Ascend 950PR/Ascend 950DT</term>|      √     |
```
weiguihua2, line 12: Convert to English document

herizhen (approved):

freyfwt (approved):

AuroraEmiya (approved):

vllm_ascend/ops/triton/docs/lightning_attn.md
MrMC-Sqz: done

liqian79 (approved):

PR #15053, opened 2026-08-27
[BugFix][DSA-CP] Stabilize lightning indexer TopK across query slices

pisceskkk:

GLM-5.2 DSA-CP shards a chunked-prefill Query chunk across TP ranks. For a 1024-token chunk with TP=8, each rank evaluates 128 Query tokens against the shared Key history. The quantized Lightning Indexer could return different TopK indices for a Query row depending on whether that row was evaluated in the full Query chunk or in its rank-local slice. The selected indices feed compressed sparse attention, so this shape-dependent result can affect model output.

The mismatch occurs when TopK cuts through a large equal-score group. `MrgSort` with exhausted suspension prefers a later source for equal values, while the S2 source lists were registered in increasing order. Full and sliced Query shapes can use different merge groupings, making tie resolution depend on the Query tiling.

This PR:

- registers S2 merge sources in reverse order for both four-way merges and two/three-way tail merges, giving equal scores a stable absolute-Key ordering;
- adds source-copy and binary-build dependencies so nested kernel-header changes trigger recompilation;
- adds an NPU regression covering exact TopK-order equivalence between full and rank-sliced Query execution.

Yes. GLM-5.2 DSA-CP chunked prefill now produces shape-invariant Lightning Indexer TopK results for equal-score ties.

- Added `test_equal_score_topk_is_invariant_to_query_slicing` with a 1024-token Query, eight 128-token slices, and an equal-zero-score group crossing `sparse_count=2048`.
- Ran the regression on Ascend 910_93: all 8 ranks matched the full-Query TopK indices in exact order.
- Replayed captured GLM-5.2 layer 0/1 inputs: 2048/2048 Query rows matched exactly across all 8 ranks; minimum TopK intersection was 2048/2048.
- Checked per-token Query quantization: quantized values and scales matched exactly for all 8 slices.
- Checked fused shared-KV attention in sliding-window and compressed-sparse modes: all 8 slices matched the corresponding full-Query rows with `max_abs=0`.
- Started GLM-5.2-W4A8 with TP=8, DCP=8, eager mode, `enable_dsa_cp=true`, and `long_prefill_token_threshold=1024`; a 12019-token prompt completed successfully with readable output.

Validation limitations:

- Exact online repeatability remains inconclusive because repeated greedy requests diverged in both DCP=8 and DCP=1 controls.
- Graph mode and performance were not measured.
- The NPU container lacked the optional `modelscope` package required by the repository-wide pytest conftest, so the added regression function was executed directly; CI should collect it through the standard test environment.


weiguihua2:
/nightly multi-node-GLM-5.1-W8A8C8-A3_128k_90_50
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/33855708596)

weiguihua2:
/nightly multi-node-GLM-5.1-W8A8C8-A3_128k_90_50
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/34181619910)

PR #9465, opened 2026-05-22
[CI]repair custom ops ci

ZT-AIA:
### What this PR does / why we need it?
Fix the nightly custom ops test cases; this is mainly caused by changes in vllm and inherent defects in the test cases themselves.

### Does this PR introduce _any_ user-facing change?
no

### How was this patch tested?
nightly custom op

- vLLM version: v0.20.2
- vLLM main: https://github.com/vllm-project/vllm/commit/1ac10f159a09897baada01b14b6a0dd6442aefd6

PR #15066, opened 2026-08-27
[Doc][Test] Add topk_log_softmax operator doc and re-enable its accuracy tests

MrMC-Sqz:

Two related changes around the `_topk_log_softmax_kernel` Triton operator (`vllm_ascend/worker/v2/sample/logprob.py`).

**1. Re-enable the skipped accuracy tests.**

All four tests in `tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_compute_token_logprobs.py` are currently skipped with `@pytest.mark.skip("UB overflow, zengtian needs to fix it later")`, added by #9465. The UB overflow they refer to is the `tl.maximum` NaN-handling regression after the NPU-IR upgrade, which #9193 worked around by pinning `propagate_nan=tl.PropagateNan.ALL` in the kernel — merged four days *before* #9465. That workaround is still in main, and `test_compute_topk_logprobs.py` already exercises the same kernel through `compute_topk_logprobs` at the same production `BLOCK_SIZE=12944` without being skipped. The skip markers therefore no longer describe a live defect, and the operator is currently uncovered in the nightly job.

This PR only removes the four markers; the test bodies are untouched.

**2. Add the operator document.**

`_topk_log_softmax_kernel` had no spec of any kind. `vllm_ascend/ops/triton/docs/topk_log_softmax.md` is structured as Description / Parameters / Constraints / Origin and Differences / Test Cases.

Beyond the formula and the two-pass reduction flow, it records the constraints a caller cannot see from the signature:

- `token_ids` and the output are addressed as `req_idx * topk`, so a column-sliced or otherwise non-contiguous `token_ids` is silently mis-addressed; the caller must pass a contiguous tensor.
- Token ids are not bounds-checked against `vocab_size`.
- `BLOCK_SIZE=12944` and `multibuffer=False` are sized against the UB budget, and the `propagate_nan=tl.PropagateNan.ALL` form of `tl.maximum` is the NPU-IR compatibility workaround from #9193 — neither is a free tuning knob.
- `PADDED_TOPK` is `constexpr`, so every distinct `next_power_of_2(topk)` costs one recompilation.
- The grid is one program per request, so a single-request call uses one vector core no matter how large `vocab_size` is.
- The output is fp32 regardless of the `logits` dtype, and the reduction is accumulated in fp32.

The Origin section records that the kernel is adapted from vLLM's GPU `vllm/v1/worker/gpu/sample/logprob.py` and installed by `patch_v2/patch_triton.py`, that the companion `_ranks_kernel` is parallelized over vector cores rather than one program per request, and that the upstream `logprob_token_ids_state` path is accepted for signature compatibility but not implemented.

> Note on the doc location: the kernel lives in `vllm_ascend/worker/v2/sample/logprob.py` rather than under `vllm_ascend/ops/triton`, but the doc is placed in `vllm_ascend/ops/triton/docs/` so all Triton operator docs stay in one place, with the source path stated at the top.

No. Test-only and documentation-only changes; no operator behavior is modified.

The kernel itself is unchanged; this only removes the skip markers.

```bash
pytest -sv tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_compute_token_logprobs.py
```

Verified locally on NPU hardware; all four re-enabled cases pass.

- vLLM version: v0.27.1

### DIFF
--- a/tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_compute_token_logprobs.py
+++ b/tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_compute_token_logprobs.py
@@ -43,7 +43,6 @@ VOCAB_SIZES = [
 TOPK_VALUES = [1, 2, 5, 10, 32, 64]
 
 
-@pytest.mark.skip("UB overflow, zengtian needs to fix it later")
 @pytest.mark.parametrize(
     "batch_size, vocab_size, topk",
     [(random.randint(1, 64), vocab_size, topk) for vocab_size in VOCAB_SIZES for topk in TOPK_VALUES],
@@ -87,7 +86,6 @@ def test_topk_log_softmax_kernel(batch_size, vocab_size, topk):
     )
 
 
-@pytest.mark.skip("UB overflow, zengtian needs to fix it later")
 @pytest.mark.parametrize("vocab_size", VOCAB_SIZES)
 def test_topk_log_softmax_edge_cases(vocab_size):
     """
@@ -131,7 +129,6 @@ def test_topk_log_softmax_edge_cases(vocab_size):
     )
 
 
-@pytest.mark.skip("UB overflow, zengtian needs to fix it later")
 @pytest.mark.parametrize(
     "batch_size, vocab_size, topk",
     [
@@ -166,7 +163,6 @@ def test_topk_log_softmax_deterministic(batch_size, vocab_size, topk):
         assert torch.equal(results[0], results[i]), f"Non-deterministic results detected in run {i}"
 
 
-@pytest.mark.skip("UB overflow, zengtian needs to fix it later")
 @pytest.mark.parametrize("dtype", [torch.float32, torch.float16])
 def test_topk_log_softmax_dtypes(dtype):
     """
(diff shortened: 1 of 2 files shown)


herizhen (approved):

freyfwt (approved):

tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_compute_token_logprobs.py
```diff
@@ -131,7 +129,6 @@ def test_topk_log_softmax_edge_cases(vocab_size):
     )
 
 
-@pytest.mark.skip("UB overflow, zengtian needs to fix it later")
```
AuroraEmiya, line 134: @ZT-AIA check this. Ensure fixes are applied.

ZT-AIA: https://github.com/vllm-project/vllm-ascend/pull/9193 This PR fixes it.

AuroraEmiya (approved):

liqian79 (approved):

ZT-AIA (approved):

PR #15105, opened 2026-08-27
[BugFix][Spec Decode] Remove obsolete sequence length update

jiajinzhu2:

This PR removes the unused `_ascend_update_seq_lens` method and its invocation in `_run_model` within `AscendAutoRegressiveSpeculator`.

The method runs after the draft model execution has completed, and the generated `seq_len_list` is not consumed by the subsequent speculative decoding flow. Sequence lengths needed by draft attention are already prepared through the existing attention metadata update path, so this additional update is unnecessary.

No.

- `python -m py_compile vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py`
- `git diff --check`

### DIFF
--- a/vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
+++ b/vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
@@ -256,7 +256,6 @@ class AscendAutoRegressiveSpeculator(AutoRegressiveSpeculator):
             cudagraph_runtime_mode,
             mm_inputs,
         )
-        self._ascend_update_seq_lens(attn_metadata)
         return last_hidden_states, hidden_states
 
     def _generate_draft(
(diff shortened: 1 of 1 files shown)


vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
```diff
@@ -383,14 +382,6 @@ def build_draft_attn_metadatas(self, num_reqs_padded, is_draft_model_prefill):
 
         return draft_attn_metadatas
 
-    def _ascend_update_seq_lens(self, attn_metadata: dict[str, Any] | None) -> None:
```
drslark, line 386: Clear.
It is dead code.
LGTM.

drslark (approved):

lijiahang226 (approved):
LGTM

weijinqian0 (approved):

PR #15111, opened 2026-08-27
[Performance] Fuse shrink and expand into sgmv_lora/bgmv_lora

### LINKED ISSUE #15110: [Feature] Fuse LoRA shrink+expand into a single sgmv_lora/bgmv_lora kernel
## Feature request

Fuse per-slice LoRA `sgmv_shrink` + `sgmv_expand` into one vector kernel `y += (x @ A) @ B` so decode does not round-trip the rank intermediate through GM.

- Rank 8/16/32/64; otherwise keep shrink/expand.
- Fully-sharded LoRA: all-gather A at `set_lora`.
- Measured on Qwen3.5-27B TP=4 FSL rank 8: TPOT 21.0 -> 20.3 ms vs shrink+expand.

This issue tracks the PR into `main`.

Agoni-02:

Linear LoRA currently launches `sgmv_shrink` then `sgmv_expand` per slice, writing the rank buffer to GM. This PR adds one vector kernel `sgmv_lora` / `bgmv_lora` that computes `y += scale * (x @ A) @ B` with the rank intermediate kept in UB.

- Supported ranks: 8 / 16 / 32 / 64. Other ranks keep shrink+expand.
- Fully-sharded LoRA: all-gather A on the rank dim at `set_lora` only.
- Devices without `HardwareCapability.LORA_CUSTOM_OPS` (including 310P) and `max_lora_rank >= 128` stay on the PyTorch punica path.
- Do **not** patch `apply()`, `_mcp_apply`, or `_apply_lora_to_output`. FSL QKV/gate_up stays on official shrink → all_gather → expand. Replacing that path made `/v1/chat/completions` look like the base model.
- Fusion is attempted in `add_lora_linear` for layers that already call it. Fallback is shrink+expand.

Rebased onto current `main` (includes the hardware-profile refactor in #15256). The previous PR head targeted `453f409` and would not merge.

Fixes #15110

No API change. Linear LoRA with supported ranks uses the fused kernel automatically when `add_lora_linear` is on the hot path.

- Single-card kernel vs shrink+expand (rank=8, hidden=5120): T=1 and T=32 `max_abs_err=0` after `SHRINK_TILE` / `W_IN_TILE_NUM_ELEMENTS` = 8192.
- Official main `875cb55` + this kernel + `HardwareCapability` gate, 910B4 TP=4, FSL off, prefix-cache off: `/v1/chat/completions` adapter identity OK. Fused and official no-FSL decode throughput are comparable; this PR does not claim an e2e speedup.
- Production 0.19.1 overlay numbers are a different stack and are not the merge target.
- Unit tests were not added because the kernel requires NPU + CANN.

**CheckList**
- [x] Pre-commit: ruff check/format, codespell, typos, long-functions, symbolic-meta, forbidden imports. Windows cannot run bash gitleaks (`/bin/bash`); Linux CI will.
- [x] One commit, DCO `Signed-off-by`, no Cursor co-author.
- [x] Docs/assets: not involved.
- [x] User-facing API: not involved.

### DIFF
--- a/vllm_ascend/lora/punica_npu.py
+++ b/vllm_ascend/lora/punica_npu.py
@@ -37,9 +37,11 @@ class PunicaWrapperNPU(PunicaWrapperBase):
             from vllm_ascend.lora.lora_ops import (
                 bgmv_expand,
                 bgmv_expand_slice,
+                bgmv_lora,
                 bgmv_shrink,
                 sgmv_expand,
                 sgmv_expand_slice,
+                sgmv_lora,
                 sgmv_shrink,
             )
         self.bgmv_expand = bgmv_expand
@@ -48,6 +50,14 @@ class PunicaWrapperNPU(PunicaWrapperBase):
         self.sgmv_expand = sgmv_expand
         self.sgmv_expand_slice = sgmv_expand_slice
         self.sgmv_shrink = sgmv_shrink
+        if not get_current_hardware_profile().supports(HardwareCapability.LORA_CUSTOM_OPS) or (
+            self.lora_config is not None and self.lora_config.max_lora_rank >= 128
+        ):
+            self.sgmv_lora = None
+            self.bgmv_lora = None
+        else:
+            self.sgmv_lora = sgmv_lora
+            self.bgmv_lora = bgmv_lora
 
     def update_metadata(
         self,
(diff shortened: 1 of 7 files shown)


vllm_ascend/lora/punica_npu.py
```diff
@@ -330,6 +340,10 @@ def add_lora_linear(
 
         assert len(lora_a_stacked) == len(lora_b_stacked) == len(output_slices)
 
+        a_use = kwargs.get("full_rank_lora_a") or lora_a_stacked
```
nu11ptr7, line 343: If `full_rank_lora_a` was never set (hook didn't run, import failed), this silently falls back to the sharded `lora_a_stacked` and the fused kernel computes wrong results instead of falling back to shrink+expand.

PR #15123, opened 2026-08-27
[Performance][Spec Decode] Shard DeepSeek V4 DSpark main projection across TP ranks

jianzs:

DeepSeek V4 DSpark currently implements `main_proj` with `ReplicatedLinear`, so every tensor-parallel rank stores the full projection weight and executes the full projection.

This PR replaces it with `ColumnParallelLinear`. The output features are sharded across TP ranks, and `gather_output=True` restores the full hidden state before RMSNorm and the downstream draft layers.

For DeepSeek V4 Pro, the BF16 `main_proj` weight has shape `[7168, 21504]`. With TP=8, this change reduces persistent weight memory on each rank from 294 MiB to 36.75 MiB, saving 257.25 MiB per rank and about 2.01 GiB across the TP group, excluding temporary communication buffers.

It also reduces the projection weight read on each rank to one eighth of the replicated implementation. This is expected to provide a significant memory-access benefit during decode, where the small-M projection is more bandwidth-sensitive. Decode end-to-end performance has not been quantified yet.

Performance profiling on DeepSeek V4 Pro:

| Configuration | Before | After |
| --- | ---: | ---: |
| TP=8, `--max-num-batched-tokens 8192` | 9 ms | 2 ms |

No. This only changes the internal tensor-parallel execution strategy of the DSpark main projection.

- Profiled DeepSeek V4 Pro with TP=8 and `--max-num-batched-tokens 8192`; `main_proj` latency decreased from 9 ms to 2 ms.
- `python3 -m py_compile vllm_ascend/models/deepseek_v4/dspark.py`

No new unit test was added because the change uses the existing `ColumnParallelLinear` output sharding and AllGather path without introducing a new operator or interface.


vllm_ascend/models/deepseek_v4/dspark.py
```diff
@@ -153,13 +153,14 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = "") -> None:
...
         )
-        self.main_proj = ReplicatedLinear(
+        self.main_proj = ColumnParallelLinear(
             config.hidden_size * len(self.target_layer_ids),
             config.hidden_size,
             bias=False,
             return_bias=False,
             quant_config=_main_proj_qconfig,
             prefix=maybe_prefix(prefix, f"layers.{self.mtp_start_layer_idx}.main_proj"),
+            gather_output=True,
```
nu11ptr7, line 164: With `gather_output=True` every rank rebuilds the full hidden output each step, so the shard only saves compute, not communication. If the next layer could consume the shard directly, the gather is pure overhead - required?

jianzs: Yes, it does reduce compute, but it also adds communication overhead. The net benefit comes from the compute savings exceeding the communication cost. Adding gather_output here is purely to keep the behavior fully aligned with ReplicatedLinear. If we skip the all‑gather at this layer, the downstream pipeline might need additional adjustments—though I'm not entirely sure about that.

jianzs: By the way, during the decoding phase, it can reduce the cost of memory access.

jianzs:
[Bot]: cherry-pick command failed. Please check the [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/33076802535) for details.

ningjingbengxiaohai (dismissed):

ningjingbengxiaohai (approved):

PR #15122, opened 2026-08-27
[Performance][KDA] Compose and overlap K3 gate projections on main

Dawn952:
### What this PR does / why we need it?

Ports #14497 to current `main` on top of the Kimi K3 integration in #14454.

The vLLM 0.27 K3 adapter in #14454 differs from the release-branch implementation used by #14497, so this PR adapts the optimization to the main-specific mixed KDA projection path:

- compose `f_b_proj @ f_a_proj` while loading and pack beta/F/output-gate into one FLOAT BFG projection;
- split MXFP QKV DynamicQuant from its matmul and overlap it with the BFG projection on a dedicated NPU stream;
- join both stages with explicit events and stream-lifetime tracking;
- route global/local F shards and reloads through the model's v0.27 packed loader;
- leave the ordinary upstream `in_proj_qkvgfab` path unchanged.

The branch merges the current #14454 head into the latest `main`. Its three main conflicts were resolved by retaining both current-main and K3 behavior: `sparse_attention_score` plus `mla_prolog_v3`, current v2 Mamba precopy ownership plus the K3 state-copy hook, and the current source-level Mamba regression test.

### How was this patch tested?

- `ruff format` and `ruff check` on all six changed Python files.
- Python syntax compilation.
- In `zjb_0827_0822-main` on `green-52-183`:
  - `tests/ut/ops/test_kimi_kda.py`
  - `tests/ut/models/test_kimi_k3_adapter.py`
  - `tests/ut/quantization/test_modelslim_config.py`
  - Result: **90 passed**.

### Does this PR introduce any user-facing change?

No API change. Mixed-precision Kimi K3 KDA projection preprocessing is reorganized for lower runtime overhead.

Related: #14454, #14497
- vLLM version: v0.27.1
- vLLM main: https://github.com/vllm-project/vllm/commit/ba07e4a48fc951300d97eb506217dd530583dea3

PR #15124, opened 2026-08-27
[Performance][KDA] Integrate K3 and overlap gate projections on main

Dawn952:

This is the conflict-free `main` replacement for #14497, based on the current Kimi K3 integration in #14454.

To keep CI and maintainers' rebases deterministic, the full integration is represented as one signed-off commit on the latest `main` (`23c8bf712`). It contains the current #14454 tree plus the main-specific #14497 adaptation:

- compose `f_b_proj @ f_a_proj` while loading and pack beta/F/output-gate into one FLOAT BFG projection;
- split MXFP QKV DynamicQuant from its matmul and overlap it with BFG on a dedicated NPU stream;
- join both stages with explicit events and stream-lifetime tracking;
- route global/local F shards and reloads through the v0.27 packed loader;
- leave the ordinary upstream `in_proj_qkvgfab` path unchanged;
- preserve the default SiTU beta in the unquantized path, fixing the inherited mypy error.

This supersedes #15122, whose multi-commit integration was GitHub-mergeable but could not pass the workflow's per-commit rebase step.

- Ruff format/check on all 68 changed Python files.
- `git diff --check` and Python syntax compilation.
- In `zjb_0827_0822-main` on `green-52-183`:
  - `tests/ut/ops/test_kimi_kda.py`
  - `tests/ut/models/test_kimi_k3_adapter.py`
  - `tests/ut/quantization/test_modelslim_config.py`
  - Result: **90 passed**.

Kimi K3 support from #14454 is added on main. The #14497 portion changes only mixed-precision KDA projection preprocessing and does not change its public API.

Related: #14454, #14497

- vLLM version: v0.27.1


Dawn952:
All self-service gates now pass: merge-conflict labeler, DCO, pre-commit/mypy, coverage recommendation, and select-tests. The final CI gate requires the ready-precise label because selected tests were produced. I attempted to add it, but this contributor account does not have label permission. @wenjun91 could you please add ready-precise so the recommended tests can run?

PR #14620, opened 2026-08-20
[Refactor][Ops] Remove causal_conv1d_update_npu

maoxx241:
### What this PR does / why we need it?

This PR removes the obsolete Ascend Triton causal_conv1d_update_npu implementation.

- Delete the Triton kernel and Python wrapper.
- Stop monkey-patching the vLLM causal_conv1d_update function, leaving the upstream implementation in place.
- Remove the two skipped tests that only exercised the deleted operator.
- Keep the shared extract_last_width helper and the existing AscendC and 310P causal-conv paths unchanged.

The removed tests documented an overflow case and a probabilistic failure, so they did not provide active coverage for this implementation.

### Does this PR introduce _any_ user-facing change?

No public API or configuration change. Internally, vLLM-Ascend no longer replaces the vLLM causal_conv1d_update function with this Triton implementation.

### How was this patch tested?

- Full repository checks: bash format.sh ci
- Python syntax compilation for the three changed files
- Verified there are no remaining references to causal_conv1d_update_npu or its Triton kernel

NPU operator tests were not run locally because the operator and its skipped tests are removed by this change.

- vLLM version: v0.27.1
- vLLM main: https://github.com/vllm-project/vllm/commit/58d3918e3ea0a544ffedadad2ba84559e9c51d8f

PR #15127, opened 2026-08-27
[Feature][Model] Support GLM-5.3-Flash on Ascend 950

yiminghub2024:

Adapts vllm-ascend for GLM-5.3-Flash (hybrid KDA linear attention + NoPE
sparse MLA, native FP8) on Ascend 950.

The architecture itself now lives in this repository, under
`vllm_ascend/models/glm5next/`, so the PR no longer depends on the unmerged
upstream GLM-5.3-Flash PR (vllm-project/vllm#53906). Every `vllm` symbol the
ported code imports — 139 of them — resolves against vLLM main. What the port
required on the Ascend side is described in *Porting the architecture
downstream* below.

Three defects made the model emit fluent-looking but wrong output, or fail
only in environments other than the author's, so they are worth calling out:

1. `MHCPreOp` / `MHCFusedPostPreOp` accept `norm_weight` / `norm_eps` in
   `forward_native`, but `mhc_pre_torch` has no such parameters. Out-of-tree
   platforms reach `forward_native` through `forward_oot`, so all 45 layers
   ran without their input RMSNorm. No NaNs, plausible activation magnitudes,
   incoherent text.
2. In the chunked-context prefill path, MLA-NoPE leaves both the rope cache
   and its gather output zero-width, and `npu_gather_pa_kv_cache` then returns
   without filling the latent output. Any prompt longer than
   `max_num_batched_tokens` therefore attended over uninitialised KV. GSM8K
   dropped to 0.040.
3. Replacing an op by rebinding its defining module does not reach callers
   that already did a from-import. The GLM vision tower is imported during
   model registration, before `adapt_patch` runs, so it kept the CUDA fused
   Q/K RMSNorm kernel and crashed on `tl.extra.cuda.gdc_wait`. The patch now
   sweeps `sys.modules`.

Full list of defects this PR fixes, each with the symptom that identifies it:

| Symptom | Root cause | Where |
| --- | --- | --- |
| Fluent but incoherent text; no NaNs, plausible activations | `mhc_pre_torch` has no `norm_weight` / `norm_eps`, so `forward_native` dropped every layer's input RMSNorm on out-of-tree platforms | `patch/worker/patch_triton.py` |
| Prompts longer than `max_num_batched_tokens` return only `!`; GSM8K 0.040 | rope cache and its gather output are both zero-width, so `npu_gather_pa_kv_cache` leaves the latent output unfilled | `attention/mla_v1.py` |
| Vision tower raises `module 'triton.language.extra.cuda' has no attribute 'gdc_wait'` | the module was imported during model registration, before the patch ran, so its from-import kept the CUDA kernel | `patch/worker/patch_triton.py` |
| FIA rejects `input_layout BNSD_NBSD`; reshape of a 0-element tensor | `qk_rope_head_dim == 0` leaves the rope operands empty and FIA falls back to GQA | `attention/mla_v1.py`, `ops/mla.py`, `ops/rotary_embedding.py` |
| `ValueError: too many values to unpack (expected 2)` while loading weights | CANN 9.1 `npu_dynamic_mx_quant` returns `[out, in // 64, 2]`, the MXFP8 method unpacks `[out, in // 32]` | `quantization/methods/w8a8/fp8_block.py` |
| `AssertionError` on `state.is_cuda` in the KDA path | `gather_initial_states` / `scatter_states` are CUDA-only and their kernels reference `gdc_wait` | `patch/worker/patch_triton.py` |
| `TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'` at scheduler init | the hybrid coordinator receives `num_prefill_lookahead=None` while the non-hybrid path guards it | `patch/platform/patch_kv_cache_coordinator.py` |
| `block_size must be divisible by hash_block_size` | the GLM kpool tail uses `block_size=index_kpool` and opts out of prefix caching | `patch/platform/patch_kv_cache_coordinator.py` |
| `'NoneType' object has no attribute 'decode'` during ACL graph update | hybrid KDA+MLA models register every layer but only MLA layers contribute a captured op | `attention/mla_v1.py`, `worker/model_runner_v1.py` |
| MTP fails to start (four separate interface mismatches) | the spec-decode path assumes a DeepSeek-shaped, text-only target model | `spec_decode/llm_base_proposer.py` |
| ACL graph capture aborts with 107027 / stalls at decode-FULL | the 310P `causal_conv1d_update` fallback calls `.item()`, a host sync | `patch/worker/patch_triton.py` |

### Porting the architecture downstream

The model definition, config classes, multimodal processor and MTP module are
carried under `vllm_ascend/models/glm5next/` and registered through
`ModelRegistry.register_model`. Three helpers that upstream satisfies with CUDA
kernels have Ascend equivalents in `vllm_ascend/models/glm5next/ops/`:

| Helper | Upstream | Here |
| --- | --- | --- |
| KDA chunk / recurrent | CUDA Triton in `third_party/flash_linear_attention` | the existing NPU Triton entry points in `vllm_ascend.ops.triton.kda`, imported directly |
| `fwht128_quant_fp8` | fused Hadamard-128 rotation + block-128 ue8m0 FP8 quantization | torch against a cached normalized Hadamard matrix, keeping the bf16 materialization before quantizing; device-side only, so ACL-graph safe |
| `gather_initial_states` / `scatter_states` | Triton kernels asserting `state.is_cuda` | plain `index_select` / `index_copy_`, no host sync |

`causal_conv1d` routes to the Ascend kernels through a local wrapper that drops
the CUDA-side cache-management kwargs once instead of at each call site.

`KpoolTailSpec` / `KpoolTailManager` register through vLLM's
`register_custom_kv_cache_specs` platform hook, which runs after the built-in
specs, so the kpool tail layers keep their own spec without patching the
registry.

Config registration still needs a patch, because the architecture is
downstream: `_CONFIG_REGISTRY` learns the three GLM-5.3-Flash config classes,
and `is_deepseek_mla` is widened because upstream resolves it from a hard-coded
`model_type` whitelist. Both are documented in `vllm_ascend/patch/__init__.py`
with a removal condition.

Two consequences worth calling out:

- The monkey patches that rebound module attributes on
  `vllm.models.glm5next.nvidia.kda` are now unreachable and have been removed.
  The remaining patches in `patch_triton.py` target upstream modules that the
  ported code still goes through (mHC, the vision tower's fused Q/K RMSNorm,
  the mamba state ops, FLA's KDA entry points), so they stay.
- The kpool sparse indexer keeps its `CustomOp` layer so the index-K and tail
  caches still build, but the scoring path raises on Ascend. Upstream
  implements it as fused CUDA kernels (DeepGEMM block-FP8 MQA logits, paged MQA
  logits, radix top-k over a device workspace) with no NPU equivalent yet.
  GLM-5.3-Flash only enables it when a checkpoint sets `index_topk`; the
  configuration validated here is dense NoPE MLA plus KDA.

No new flags. GLM-5.3-Flash requires `--block-size 512`, since the kpool
indexer uses `index_kpool * 32`. Serving no longer requires a vLLM branch
carrying #53906; vLLM main plus this branch is enough.

Ascend 950PR x8, TP=8, native FP8 checkpoint, `--max-model-len 131072`,
`--block-size 512`, ACLGraph enabled.

Everything below was re-run in a **brand-new container**, with vllm-ascend
installed from this branch as it stood before the port and vLLM pinned to a
fixed commit, to make sure no local state from development was involved.
ACLGraph captured cleanly (piecewise 4/4, decode FULL 4/4 in 20 s) and prefill
reached 6399.8 tokens/s, matching the development environment.

That run therefore resolved the model from a vLLM branch carrying #53906, on
the same checkpoint and the same kernels. The port relocates the model
definition and swaps the three helpers listed above, leaving the dense NoPE
MLA + KDA execution path unchanged, so a re-run on the ported tree is the
remaining thing to confirm.

Accuracy, GSM8K 5-shot via lm-eval `local-chat-completions`, 250 questions,
greedy:

| Configuration | flexible-extract | strict-match |
| --- | ---: | ---: |
| before the chunked-gather fix, no MTP | 0.040 | 0.036 |
| after the fix, no MTP | 0.896 | 0.860 |
| after the fix, MTP `num_speculative_tokens=3` | 0.880 | 0.860 |
| after the fix, MTP=5, clean container rebuild | 0.892 | 0.856 |

The first row is the regression signature of defect 2. The last three agree
within one standard error (±0.02).

Correctness of the chunked-context path, as a single deterministic request:
a 28282-token prompt asking for the last number in the text answers `1200`
exactly. Before the fix the same prompt returned only `!` tokens.

AIME2026:

aime26 report table:
┌───────────────┬───────────┬────────────┬──────────┬───────┬─────────┐
│ Model         │ Dataset   │ Metric     │ Subset   │   Num │ Score   │
├───────────────┼───────────┼────────────┼──────────┼───────┼─────────┤
│ glm-5.3-flash │ AIME-2026 │ Accuracy ↑ │ default  │    30 │ 86.7%   │
└───────────────┴───────────┴────────────┴──────────┴───────┴─────────┘ 

gpqa_diamond report table:
┌───────────────┬──────────────┬────────────┬──────────┬───────┬─────────┐
│ Model         │ Dataset      │ Metric     │ Subset   │   Num │ Score   │
├───────────────┼──────────────┼────────────┼──────────┼───────┼─────────┤
│ glm-5.3-flash │ GPQA-Diamond │ Accuracy ↑ │ default  │   198 │ 77.8%   │

Both numbers are lower bounds, in the same direction and for the same reason.
Generations that exhaust the 40960-token budget while still inside the reasoning
phase emit no final answer and score as wrong: 4 of 30 on AIME 2026, 36 of 198 on
GPQA-Diamond. Their reasoning traces are not degenerate — no repetition loops,
still advancing new case analysis at the cut — so this is the model's reasoning
length on hard items, not a generation defect. I kept the cap rather than raising
it, since capping and scoring over-length generations as wrong is the standard
convention; scoring only the completed subset would be selection-biased upward and
is not reported here.

Performance:

- Single request, real prompt, greedy: 55.6 tokens/s, MTP mean acceptance
  length 2.1-3.1.
- 4 concurrent real prompts: acceptance 53% greedy, 30-44% at temperature=1.0
  / top_p=0.95.
- `vllm bench serve`, random dataset, 64K input / 3000 output, 10 requests:
  aggregate prefill 6400 tokens/s, MTP acceptance 46.21%, acceptance length
  2.39. Acceptance on this dataset is a sensitive indicator of chunked-prefill
  correctness: before the fix it was 0%.

Unit test added: `tests/ut/attention/test_mla_nope_chunked_gather.py` covers
defect 2 and asserts the rope-present path is unchanged.

Known dependencies:

  `Glm5NextForConditionalGeneration` and `Glm5NextMTPModel` are registered from
  this repository, so neither a vLLM fork nor a cherry-pick of #53906 is
  needed.
- ACLGraph needs `causal_conv1d_update_npu`, which is not in vllm-ascend main
  yet. The binding added here is guarded and falls back when the kernel is
  absent, but decode-FULL capture then stalls on a host sync, so the kernel is
  a hard requirement for ACLGraph on KDA models.


vllm_ascend/ops/triton/kda/kda.py
```diff
@@ -132,6 +140,21 @@ def fused_recurrent_kda(
...
+            raise ValueError("compute_gate requires a_log.")
+        g = apply_kda_gate(
+            g,
+            a_log,
+            g_bias,
+            safe_gate=True,
+            lower_bound=DEFAULT_KDA_LOWER_BOUND if lower_bound is None else lower_bound,
+        )
+    if sigmoid_beta:
+        beta = beta.float().sigmoid()
```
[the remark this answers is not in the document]
yiminghub2024: This is deliberate. `beta` feeds the decay term of the recurrent gate, where
accumulating in the activation dtype loses precision over long sequences, so
it is kept in fp32 and the kernel consumes it as fp32.

No dtype mismatch shows up in practice: the configuration in this PR was run
end to end on Ascend 950PR x8 and scores 0.896 on GSM8K 5-shot (250 questions,
greedy), which would not hold if the kernel were rejecting or misreading the
operand.

Casting back to `beta.dtype` would change the numerics of a validated path to
guard against a failure that is not occurring, so I would rather leave it and
revisit if a concrete mismatch is reported.

vllm_ascend/ops/triton/kda/gate.py
```diff
@@ -0,0 +1,55 @@
...
+        )
+    a = a_log.reshape(*([1] * (g.dim() - 2)), heads, 1).to(device=g.device, dtype=torch.float32)
+    if g_bias is not None:
+        bias = g_bias.to(device=g.device, dtype=torch.float32).reshape(
+            *([1] * (g.dim() - 2)), heads, g.shape[-1]
+        )
+        g = g + bias
+    if safe_gate:
+        return (lower_bound * torch.sigmoid(torch.exp(a) * g)).to(dtype=torch.float32)
+    return ((-torch.exp(a)) * F.softplus(g)).to(dtype=torch.float32)
```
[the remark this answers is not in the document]
yiminghub2024: Same reasoning as the `beta` comment above. The gate output is an exponential
decay factor that is accumulated across timesteps, so it is computed and
returned in fp32 on purpose; downcasting it to the activation dtype at this
point is where precision is actually lost.

The consumers accept fp32 here, and the full configuration was validated at
0.896 on GSM8K 5-shot on Ascend 950PR x8. I would prefer not to change the
numerics of that path without a reproducer showing an actual dtype error.

vllm_ascend/attention/mla_v1.py
```diff
@@ -1222,6 +1242,27 @@ def _forward_prefill(
...
+        # so the scatter would silently write nowhere.
+        cache = kv_cache[0]
+        if not cache.is_contiguous():
+            raise RuntimeError(
+                f"mla_nope_kv_write: kv cache not contiguous, shape={tuple(cache.shape)} stride={cache.stride()}"
+            )
+        flat = cache.view(-1, cache.shape[-1])
+        token = k_nope.reshape(-1, cache.shape[-1]).to(flat.dtype)
+        idx = slots.to(torch.int64)
+        flat.index_copy_(0, idx, token)
```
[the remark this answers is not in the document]
yiminghub2024: Thanks, good catch on the failure mode. I have replaced the hard error with a
fallback rather than taking the suggested code as-is, because the suggestion
reads `block_size` from `cache.shape[-2]`, which is the head dimension in this
layout, not the block dimension.

The existing flat path (`cache.view(-1, cache.shape[-1])` indexed directly by
the slot mapping) only lines up if the cache is laid out as
`(num_blocks, block_size, num_heads, dim)`, so `block_size` is `shape[1]`.
Indexing with `shape[-2]` would scatter the tokens to the wrong slots, which is
exactly the class of silent corruption the guard was there to catch.

The updated version keeps the fast flat scatter when the cache is contiguous
and falls back to indexed assignment when it is not, so the padded-page case
you describe is handled without a crash.

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -497,9 +499,28 @@ def _maybe_share_lm_head(self, model: nn.Module) -> None:
...
+                )
+            else:
+                # mtp_head_tied_share: GLM-5.3-Flash omits shared_head.head from
+                # the checkpoint and ties it to the target lm_head, so the
+                # equality test never fires and the draft head stays
+                # uninitialised. Share whenever the shapes line up.
+                for _, layer_module in self.model.model.layers.items():
+                    draft_head = layer_module.shared_head.head
+                    if draft_head.weight.shape == target_lm_head.weight.shape:
+                        layer_module.shared_head.head = target_lm_head
```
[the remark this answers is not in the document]
yiminghub2024: Agreed, applied. Simplified slightly by relying on `getattr(None, "head", None)`
returning `None`, so a single guard covers both a missing `shared_head` and a
missing `head`.

yiminghub2024:
> This pull request has conflicts, please resolve those before we can evaluate the pull request.

fixed

Liears:
Thanks for this comprehensive GLM-5.3-Flash adaptation! 🙏

We validated it on **Atlas A2 (910B3)** against the **open-source vLLM main baseline** (the glm-release branch of vllm#53906) and hit ~10 compatibility gaps where the code assumed the vendored vLLM tree (PCP import location, `indexes_kv_by_block_stride` / `storage_block_size` / `shared_by` field differences, the aliasing KV-cache layout, bind_kv_cache dispatch skipping Mamba state unpacking, and ModelSlim quant-description lookups for the draft model).

Fixes are up in #15276 — happy to split/rebase however works best for you. Also validated end-to-end with a W8A8-converted BF16 checkpoint (msmodelslim 26.1.0 doesn't support glm5_next yet, so we wrote a converter; can contribute it separately if useful).

Liears:
Gemini Code Assist left two defensive-coding suggestions on this PR's diff that look reasonable — passing them along in case they're useful:

1. Wrap the `sys.modules` attribute rebinding sweep in `patch_triton.py` in try/except so a read-only or frozen module can't abort startup.
2. Wrap the `inspect.signature` usage in try/except to tolerate C-bound or mock objects.

(Context: we're running this PR's tree against open-source vLLM main on 910B3 — see follow-up fixes #15284–#15288.)

yiminghub2024:
> Gemini Code Assist 在这个 PR 的差异分析中留下了两条防御性编码建议，看起来比较合理——如果有用的话，就分享给大家：
> 
> 1. 将`sys.modules`属性重新绑定扫描操作包装`patch_triton.py`在 try/except 语句中，以防止只读或冻结的模块中止启动。
> 2. 将使用语句包裹`inspect.signature`在 try/except 语句中，以容忍 C 绑定对象或模拟对象。
> 
> （背景：我们正在针对开源 vLLM 主分支的 910B3 版本运行此 PR 的代码树——请参阅后续修复#15284 – #15288。）

fIxed , thanks @Liears

yiminghub2024:
> 感谢您提供的如此全面的 GLM-5.3-Flash 适配！🙏
> 
> **我们在Atlas A2 (910B3)**上针对**开源 vLLM 主基线**（vllm#53906 的 glm-release 分支）进行了验证，发现存在约 10 个兼容性差距，其中代码假定了供应商提供的 vLLM 树（PCP 导入位置、 `indexes_kv_by_block_stride`/ `storage_block_size`/`shared_by`字段差异、别名 KV 缓存布局、bind_kv_cache 分发跳过 Mamba 状态解包以及草案模型的 ModelSlim 量化描述查找）。
> 
> [修复程序已在#15276](https://github.com/vllm-project/vllm-ascend/pull/15276)中发布——乐意根据您的需要进行拆分/重新基于。此外，我们还使用 W8A8 转换的 BF16 检查点进行了端到端验证（msmodelslim 26.1.0 尚不支持 glm5_next，因此我们编写了一个转换器；如果需要，我们可以单独贡献它）。

Thanks for taking this to real hardware against open-source vLLM main — that is exactly the validation this PR needed, and the vendored-tree assumptions you found are real. Also thanks for passing along the Gemini suggestions.

Two updates from my side, then a proposal on sequencing.

The model architecture is now ported into this repo. Following the earlier review feedback, GLM-5.3-Flash no longer depends on vllm#53906. The config, model, KDA, MTP and multimodal code now live under vllm_ascend/models/glm5next/, with the CUDA-only pieces reimplemented for NPU (KDA chunked recurrence, the FWHT/UE8M0 quantized path, and the Mamba-style state gather/scatter), and the architectures plus Glm5NextConfig are registered from vllm-ascend. All 139 vLLM symbols the port imports resolve against current vLLM main.

To be clear about what that does and does not fix: it removes the dependency on the unmerged model definition only. It does not cover the gaps you found — the PCP import location, the indexes_kv_by_block_stride / storage_block_size / shared_by field differences, the aliasing KV-cache layout, the bind_kv_cache dispatch skipping Mamba state unpacking, and the ModelSlim quant-description lookup for the draft model are all assumptions in the vllm-ascend adaptation layer rather than in the model files. So your fixes are still needed on top; the two efforts are complementary, not overlapping.

Applied both defensive suggestions. The two sys.modules rebinding sweeps in patch_triton.py now handle each module independently, so a frozen, read-only or lazily loaded module is skipped instead of aborting the sweep at import time — I confirmed the unguarded version does abort with AttributeError on such a module, and that modules after it still get rebound with the guard in place. The inspect.signature call in _draft_embed_accepts_mm now falls back to treating the head as text-only, which is the safe direction: not forwarding the multimodal kwargs only loses multimodal support, whereas assuming support and forwarding them to a head that rejects them raises TypeError. Unrelated exceptions still propagate. Both follow the existing except (TypeError, ValueError) precedent in mooncake_backend.py, and there are unit tests for the fallback.

On sequencing — would you mind rebasing the fixes onto this PR and keeping only your incremental changes? I would like to land #15127 first, for two reasons:

Each of #15284–#15288 currently carries the entire #15127 diff in addition to its own fix, rather than just the fix. Whichever one lands first would leave #15127 essentially empty and the remaining four conflicting with each other.
They are based on a snapshot of #15127 from before the downstream port, and on an older main — for example they patch vllm_ascend/quantization/modelslim_config.py, which has since moved to vllm_ascend/quantization/configs/modelslim_config.py.
Rebasing onto the current #15127 head should also shrink each of them considerably, which makes the individual fixes much easier to review on their own merit. Happy to help resolve anything that conflicts with the ported files.

A few last things:

Yes please on the W8A8 converter as a separate contribution. msmodelslim not supporting glm5_next yet is a real blocker for anyone trying this model, so that would be genuinely useful on its own.
Heads-up while you are testing: the kpool sparse indexer is currently a skeleton that raises explicitly if index_topk is enabled, so only the dense path is covered in this PR.
My accuracy numbers (0.896 on GSM8K) were measured before the port, on a branch carrying #53906, so they need a re-run on the ported tree. If your 910B3 setup is still handy, a cross-check there would be very welcome. @Liears

docs/source/tutorials/models/GLM5.3-Flash.md
```diff
@@ -0,0 +1,285 @@
+# GLM-5.3-Flash
```
wangxiyuan: remove this guide fisrt. Let's add it back once the model is well tested.

vllm_ascend/models/glm5next/ops/causal_conv1d.py
```diff
@@ -0,0 +1,72 @@
...
+kernel is unavailable; ``has_npu_triton_conv1d_update()`` lets the caller report
+that up front instead of hanging during capture.
+"""
+
+import torch
+
+from vllm_ascend._310p.ops.causal_conv1d import (
+    causal_conv1d_fn as _torch_causal_conv1d_fn,
+)
+from vllm_ascend._310p.ops.causal_conv1d import (
```
wangxiyuan: why import 310 related code?

mkdocs.yml
```diff
@@ -211,6 +211,7 @@ nav:
         - tutorials/models/GLM4.x.md
         - tutorials/models/GLM5.md
         - tutorials/models/GLM5.2.md
+        - tutorials/models/GLM5.3-Flash.md
```
wangxiyuan: ditto

docs/source/user_guide/support_matrix/supported_models.md
```diff
@@ -120,6 +120,7 @@ Get the latest info here: <https://github.com/vllm-project/vllm-ascend/issues/16
     |-----------------|----------|--------|------|------|------|---------|-------|------|------|--------|-------|--------|--------|-------|-------|--------|----------|---------|----------|-----|
     |Qwen3.5-397B-A17B|✅        |       |✅    | Ascend 950DT |✅|✅|✅||✅|✅|✅||✅|✅|✅|✅|✅|1010000|| [Qwen3.5-397B-A17B](../../tutorials/models/Qwen3.5-397B-A17B.md) |
     |Qwen3.6-27B      |✅        |       |✅    | Ascend 950 Products |✅|✅|✅||✅|✅|✅||✅|✅|✅|✅|✅|262144|| [Qwen3.5-27B / Qwen3.6-27B](../../tutorials/models/Qwen3.5-27B-Qwen3.6-27B.md) |
+    |GLM-5.3-Flash    |🔵        |Native FP8; KDA + sparse MLA| | Ascend 950PR (128GB × 8) | |✅|🟡||🔵|✅|✅||🔵|✅|🟡|🟡|🟡|128k|| [GLM-5.3-Flash](../../tutorials/models/GLM5.3-Flash.md) |
```
wangxiyuan: ditto

tests/e2e/models/configs/GLM-5.3-Flash.yaml
```diff
@@ -0,0 +1,25 @@
+model_name: "zai-org/GLM-5.3-Flash"
```
wangxiyuan: remove this file if it's useless

vllm_ascend/patch/worker/patch_triton.py
```diff
@@ -116,3 +223,146 @@ def _l2norm(t):
...
+# npu_kda_causal_conv1d_triton: the 310P PyTorch fallback calls .item() per
+# request, which aborts ACL graph capture. The upstream NPU Triton kernel has
+# no host sync and accepts the spec-decode kwargs the fallback had to drop.
+try:
+    from vllm_ascend.ops.triton.mamba.causal_conv1d import (  # type: ignore[attr-defined]
+        causal_conv1d_update_npu as _cc1d_update_npu,
+    )
+
+    _cc1d.causal_conv1d_update = _cc1d_update_npu
+    print("[npu_kda_causal_conv1d_triton] bound NPU Triton causal_conv1d_update", flush=True)
```
wangxiyuan: do not use `print`

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -4051,6 +4052,15 @@ def _get_layer_kv_cache_specs(self, kv_cache_config: KVCacheConfig) -> dict[str,
...
         return layer_kv_cache_spec
 
+    def _is_glm5_next_kpool_layer(self, layer_name: str, spec=None) -> bool:
+        compilation_config = getattr(self, "compilation_config", None)
+        if compilation_config is None:
+            compilation_config = getattr(self.vllm_config, "compilation_config", None)
+        static_ctx = getattr(compilation_config, "static_forward_context", {}) if compilation_config else {}
+        if is_glm5_next_kpool_cache(static_ctx.get(layer_name)):
+            return True
+        return spec is not None and type(spec).__name__ == "KpoolTailSpec"
```
wangxiyuan: `type(spec).__name__` -> `isinstance`?

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -524,9 +526,33 @@ def _maybe_share_lm_head(self, model: nn.Module) -> None:
...
+                # uninitialised. Share whenever the shapes line up.
+                for _, layer_module in self.model.model.layers.items():
+                    shared_head = getattr(layer_module, "shared_head", None)
+                    draft_head = getattr(shared_head, "head", None)
+                    if (
+                        shared_head is not None
+                        and draft_head is not None
+                        and draft_head.weight.shape == target_lm_head.weight.shape
+                    ):
+                        shared_head.head = target_lm_head
```
wangxiyuan: This is safe for DeepSeek V3/R1 (their checkpoints ship a shared_head.head identical to lm_head by design), but if an MTP checkpoint has an independently trained draft head (same shape, different values), it would be silently overwritten with the target lm_head — acceptance rate would drop with no way to trace the cause.  Please consider this case?

vllm_ascend/ops/rotary_embedding.py
```diff
@@ -91,13 +91,19 @@ def set_cos_and_sin(vllm_config, max_num_reqs, decode_token_per_req, dtype, devi
...
+        return _cos_mla[:num_tokens], _sin_mla[:num_tokens]
     cos = _cos_cache[positions].unsqueeze(1).unsqueeze(2)
     sin = _sin_cache[positions].unsqueeze(1).unsqueeze(2)
     if not use_cache:
         return cos, sin
-    global _cos_mla
-    global _sin_mla
-    num_tokens = positions.size(0)
+    if _cos_mla is None or _cos_mla.shape[-1] != cos.shape[-1]:
+        return cos, sin
```
wangxiyuan, line 119: Existing rope models never hit this branch (it only affects cases that would previously have crashed outright), but callers with use_cache=True rely on the return value being a fixed-address slice of the persistent buffer for ACL graph replay. If this branch is ever reached inside a graph capture region, the cos/sin pointers would be stale at replay time, producing wrong results. Please either guarantee this branch is only reachable on the eager path, or add an explicit warning.

docs/source/tutorials/models/GLM5.3-Flash.md
yiminghub2024: Removed in 2fca6679.

Agreed on the reasoning. Publishing the guide before the model is validated across the A2/A3/A5 version pairings points users at a path we cannot support yet, and "it is documented so it must work" is a fair expectation for them to have. Happy to add it back, together with the version matrix, once the model is properly tested.

vllm_ascend/models/glm5next/ops/causal_conv1d.py
yiminghub2024: Good catch. Fixed in 2fca6679: both entry points now import from `vllm_ascend.ops.causal_conv1d`.

Some context on how it got that way, since the answer turned out to be more than a wrong import path. The module is a plain PyTorch reference implementation with no 310P-specific logic in it, and nothing inside `_310p/` actually consumes it -- the 310P GDN path goes through `torch.ops._C_ascend.npu_causal_conv1d_*`. This branch already carried a 100% rename moving the file out of `_310p/ops/` into `ops/`, but these import sites were still pointing at the old location, which made this PR the only reverse dependency from a non-310P code path into that directory.

The imports now follow the moved file, and I also dropped the two stale "for 310P" mentions from its docstrings so the next reader is not misled the same way.

mkdocs.yml
yiminghub2024: Removed in 2fca6679, together with the guide itself.

docs/source/user_guide/support_matrix/supported_models.md
yiminghub2024: Removed in 2fca6679 -- the whole row, not just the doc link.

Dropping the guide while keeping a support claim in the matrix would still invite users onto the same path, which defeats the purpose. It is also worth noting that row only ever covered Ascend 950PR: the A2/A3 table below it had no GLM-5.3-Flash entry at all, which is exactly the version-pairing gap you raised. I checked the 950 table is still well-formed after the removal (6 rows, consistent column count).

tests/e2e/models/configs/GLM-5.3-Flash.yaml
yiminghub2024: It was useless, so it is removed in 2fca6679.

For the record, in case the same question comes up for other model configs: it was not listed in `tests/e2e/models/configs/accuracy.txt`, which is the config list that actually drives the eval runs, and `conftest.py` only picks up configs via `--config-list-file` or an explicit `--config` (defaulting to `Qwen3-8B.yaml`) rather than globbing the directory. Nothing else referenced the filename either, and its only metric was a `0.0` placeholder. So no job ever executed it, and if one had, it would not have asserted anything.

vllm_ascend/patch/worker/patch_triton.py
yiminghub2024: Fixed in 2fca6679: both calls go through `logger` now.

Worth flagging what those prints were masking. `causal_conv1d_update_npu` was removed upstream in #14620, so on current main this `try` always fails and the `FAILED:` line was written to stdout on every single start -- looking like an error, while telling nobody what it actually meant. The failure branch is now a `logger.warning` that names the consequence instead of dumping a bare `repr`:

```python
logger.warning(
    "NPU Triton causal_conv1d_update is unavailable (%s); falling back to the"
    " PyTorch implementation, which syncs per request and therefore stalls ACL"
    " graph capture at decode-FULL.",
    _cc1d_err,
)
```

That fallback cost is real and currently unavoidable on main, so I would rather have it visible in the log than silent.

vllm_ascend/worker/model_runner_v1.py
yiminghub2024: Agreed, changed to `isinstance(spec, KpoolTailSpec)` in 2fca6679.

I checked the import direction before making the change, since a string comparison is sometimes there to dodge a cycle: `glm5next/kv_cache.py` imports only from vllm, and `models/__init__.py` registers the architectures by string path rather than importing them, so pulling in `KpoolTailSpec` adds neither a cycle nor any import-time cost. It also lines up with the `isinstance(spec, AscendSFAIndexerCacheSpec)` check a few lines above in the same function.

vllm_ascend/spec_decode/llm_base_proposer.py
yiminghub2024: You are right, and this was a regression I introduced. Fixed in 2fca6679.

Upstream compared weight *values* (`torch.equal`), which is inherently safe: it only ever replaces the head when the values already match, so nothing numerical changes. I relaxed it to a shape match because GLM-5.3-Flash omits `shared_head.head` from its checkpoint and ties it to the target `lm_head`, so the equality test never fires and the draft head would stay uninitialised. But as you point out, this branch is gated on `method == "mtp" and is_deepseek_mla`, not on GLM, so the relaxation applied to every checkpoint reaching it.

The shape-only match is now gated on the tied architecture, and everything else keeps the value-equality test:

```python
model_type = str(getattr(self.vllm_config.model_config.hf_config, "model_type", ""))
draft_head_is_tied = model_type.startswith("glm5_next")
...
if draft_head_is_tied or torch.equal(draft_head.weight, target_lm_head.weight):
    shared_head.head = target_lm_head
```

`startswith` covers both `glm5_next` and `glm5_next_text`. I added unit tests for the tied architecture sharing despite differing values, an independently trained head of the same shape being left untouched (the case you described), the preserved value-equality behaviour, a shape mismatch, and a layer with no `shared_head`.

vllm_ascend/ops/rotary_embedding.py
yiminghub2024: Correct, and this guard is new in this PR -- upstream crashed here instead, so I turned a crash into a silent degradation, which inside a capture region is the worse of the two. Added an explicit warning in 2fca6679.

I went with the warning rather than claiming eager-only reachability, because proving that statically across all callers is not something I can do with confidence. It is `warning_once` since this runs every forward step, and it reports both the buffer's rope_dim and the requested rope_dim so the mismatch is diagnosable:

```python
logger.warning_once(
    "MLA rope cache is unusable (buffer rope_dim=%s, requested rope_dim=%d);"
    " returning a temporary cos/sin pair instead of the persistent buffer."
    " This is unsafe inside ACL graph capture.",
    "unset" if _cos_mla is None else _cos_mla.shape[-1],
    cos.shape[-1],
)
```

On the sibling branch just above it: the 0-width early return hands back a slice of `_cos_mla` itself, so that path does keep the fixed-address guarantee and is safe for replay.

lijiahang226:
Thank you for the comprehensive PR. Could we have the latest comprehensive accuracy data, including AIME2026 and GPQA? The reported 0.896 accuracy on GSM8K looks suspicious.

yiminghub2024:
> 感谢您提供的全面公关稿。能否提供最新的综合准确率数据，包括 AIME2026 和 GPQA？报告中提到的 GSM8K 准确率 0.896 看起来有些可疑。

i will add  AIME2026 和 GPQA some days later asap ,but really test many times for  gsm8k  :0.896 ,thanks  @lijiahang226

wangxiyuan:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/attention/mla_v1.py
```diff
@@ -2085,3 +2151,20 @@ def exec_kv_prefill(
         ]
 
         return prefill_k_pe, prefill_k_c_normed
+
+
+# mla_nope_zero_rope_cache: the paged key_rope operand is cache-sized
+# (num_blocks x block_size x 64), so rebuilding it on every MLA layer of every
+# step would allocate hundreds of MB per call. It is read-only zeros, so one
+# buffer per shape is shared across all layers.
+_MLA_NOPE_ZERO_ROPE: dict = {}
```
wangxiyuan: this buffer will be always exist in memory. please init it only when it's used by the target model.

yiminghub2024: Moved off module scope in aaceeec.

The cache now lives on `AscendMLAMetadataBuilder` and reaches the impl through `AscendMLADecodeMetadata.nope_zero_rope_cache`, so it is allocated only when `hf_text_config.qk_rope_head_dim == 0` and is released with the model instead of staying resident for the lifetime of the process. For every other model the field is `None` and nothing is allocated at all.

One thing I deliberately did not do is move it onto each `AscendMLAImpl`. The `key_rope` operand FIA needs here is shaped like the paged KV cache (`num_blocks x num_kv_heads x block_size x 64`), i.e. an eighth of that layer's latent KV cache at `kv_lora_rank=512`, and it is read-only zeros that every MLA layer can share. A per-layer copy would multiply it by the number of MLA layers. The metadata builder is created per KV-cache group, so keeping it there gives one buffer per model while still tying the lifetime to the model.

The impl asserts the cache is present rather than indexing it blindly, so if `hf_text_config.qk_rope_head_dim` ever disagrees with the layer's own `qk_rope_head_dim` the failure names the cause instead of surfacing as a `NoneType` subscript.

Also added two unit tests in `tests/ut/attention/a2/test_mla_v1.py`: the builder creates the cache only for NoPE models, and the helper hands back the same buffer for a repeated shape.

yiminghub2024:
> Thank you for the comprehensive PR. Could we have the latest comprehensive accuracy data, including AIME2026 and GPQA? The reported 0.896 accuracy on GSM8K looks suspicious.

aime2026 updated as your wish ,in pr main text ,thanks ~ @lijiahang226

lijiahang226:
> > Thank you for the comprehensive PR. Could we have the latest comprehensive accuracy data, including AIME2026 and GPQA? The reported 0.896 accuracy on GSM8K looks suspicious.
> 
> aime2026 updated as your wish ,in pr main text ,thanks ~ @lijiahang226

Thank you! The accuracy looks good to me now.

yiminghub2024:
> > > Thank you for the comprehensive PR. Could we have the latest comprehensive accuracy data, including AIME2026 and GPQA? The reported 0.896 accuracy on GSM8K looks suspicious.
> > 
> > 
> > aime2026 updated as your wish ,in pr main text ,thanks ~ @lijiahang226
> 
> Thank you! The accuracy looks good to me now.

add gpqa tested result also

vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
```diff
@@ -159,9 +159,13 @@ def __init__(
             # can be a multiple of hash_block_size.
             self.hash_block_size = hash_block_size
             if enable_caching:
+                # The GLM kpool tail spec uses block_size=index_kpool and opts
+                # out of prefix caching, so it is not bound by the MLA hash
+                # block size.
                 assert all(
                     self._get_effective_block_size(g.kv_cache_spec) % hash_block_size == 0
                     for g in kv_cache_config.kv_cache_groups
+                    if getattr(g.kv_cache_spec, "participates_in_prefix_caching", True)
```
lijiahang226, line 168: Follow-up (not blocking): Track upstream release and remove related patch.

vllm_ascend/models/glm5next/attention.py
```diff
@@ -0,0 +1,558 @@
...
+            kv_a_layernorm=self.kv_a_layernorm,
+            kv_b_proj=self.kv_b_proj,
+            rotary_emb=self.rotary_emb,
+            o_proj=self.o_proj,
+            fused_qkv_a_proj=self.fused_qkv_a_proj if self.q_lora_rank is not None else None,
+            kv_a_proj_with_mqa=self.kv_a_proj_with_mqa if self.q_lora_rank is None else None,
+            q_a_layernorm=self.q_a_layernorm if self.q_lora_rank is not None else None,
+            q_b_proj=self.q_b_proj if self.q_lora_rank is not None else None,
+            q_proj=self.q_proj if self.q_lora_rank is None else None,
+            indexer=self.indexer,
```
lijiahang226, line 533: Unfortunately, the indexer cannot be invoked correctly due to a limitation in the current framework. MLAAttention and the Indexer are tightly coupled. This architectural issue needs to be addressed in the future.

vllm_ascend/ops/triton/kda/kda.py
lijiahang226, line 37: K3 has already integrated the KDA implementation. Consider using the existing interface.

vllm_ascend/models/glm5next/kv_cache.py
```diff
@@ -0,0 +1,168 @@
...
+
+from vllm.config import VllmConfig
+from vllm.v1.core.block_pool import BlockPool
+from vllm.v1.core.kv_cache_utils import BlockHashList, KVCacheBlock
+from vllm.v1.core.single_type_kv_cache_manager import FullAttentionManager
+from vllm.v1.kv_cache_interface import KVCacheSpec, SlidingWindowSpec
+from vllm.v1.request import Request
+
+
+class KpoolTailManager(FullAttentionManager):
```
lijiahang226, line 30: Please consider using AscendSlidingWindowMLASpec, which has already been used in DeepSeek-V4.

vllm_ascend/models/glm5next/ops/mhc_ops.py
```diff
@@ -0,0 +1,20 @@
...
+
+Upstream keeps these next to the MHC ops in ``vllm.model_executor.layers.mhc``.
+They are plain shape ops with no backend-specific behavior, so vLLM Ascend
+carries its own copy while the GLM-5.3-Flash architecture lives downstream.
+"""
+
+import torch
+
+
+def hc_expand(x: torch.Tensor, n: int) -> torch.Tensor:
```
lijiahang226, line 13: Please consider using torch.ops._C_ascend.npu_hc_pre_v2 and torch.ops._C_ascend.npu_hc_post directly, as they are already used in DeepSeek-V4.

vllm_ascend/models/glm5next/ops/causal_conv1d.py
```diff
@@ -0,0 +1,72 @@
...
+
+_UPDATE_FALLBACK_UNSUPPORTED_KWARGS = (*_UNSUPPORTED_KWARGS, "max_query_len", "out")
+
+
+def has_npu_triton_conv1d_update() -> bool:
+    """Whether the host-sync-free NPU Triton update kernel is available."""
+    return _HAS_NPU_TRITON_CONV1D_UPDATE
+
+
+def causal_conv1d_fn(*args, **kwargs) -> torch.Tensor:
```
lijiahang226, line 58: Please consider directly using the existing causal_conv1d implementation in the repository.

vllm_ascend/models/glm5next/attention.py
```diff
@@ -0,0 +1,558 @@
...
+        self.mla_attn = MultiHeadLatentAttentionWrapper(
+            self.hidden_size,
+            self.num_local_heads,
+            self.scaling,
+            self.qk_nope_head_dim,
+            self.qk_rope_head_dim,
+            self.v_head_dim,
+            self.q_lora_rank,
+            self.kv_lora_rank,
+            mla_modules,
```
Biuapha, line 548: MultiHeadLatentAttentionWrapper will not use the 'forward()' of mla_modules.indexer

vllm_ascend/models/glm5next/kv_cache.py
Biuapha, line 30: The state_cache of dsv4 is SlidingWindowSpec，there is no need to add a new one.

vllm_ascend/models/glm5next/attention.py
```diff
@@ -0,0 +1,558 @@
...
+            q_a_layernorm=self.q_a_layernorm if self.q_lora_rank is not None else None,
+            q_b_proj=self.q_b_proj if self.q_lora_rank is not None else None,
+            q_proj=self.q_proj if self.q_lora_rank is None else None,
+            indexer=self.indexer,
+            indexer_rotary_emb=self.indexer_rope_emb,
+            is_sparse=self.is_v32,
+            topk_indices_buffer=topk_indices_buffer,
+        )
+
+        self.mla_attn = MultiHeadLatentAttentionWrapper(
```
Biuapha, line 539: In MultiHeadLatentAttentionWrapper,  kernel 'npu_sparse_flash_attention' still not support rope_dim=0

vllm_ascend/models/glm5next/kda.py
```diff
@@ -0,0 +1,595 @@
+# SPDX-License-Identifier: Apache-2.0
```
Biuapha, line 1: Reuse Kimi K3’s KDA , don't add a new one

vllm_ascend/models/glm5next/ops/mhc_ops.py
```diff
@@ -0,0 +1,20 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""Hyper-connection width helpers used by the GLM-5.3-Flash decoder layers.
+
+Upstream keeps these next to the MHC ops in ``vllm.model_executor.layers.mhc``.
```
Biuapha, line 5: why not reuse dsv4's MHC? Don't add a new one

vllm_ascend/patch/platform/patch_kv_cache_coordinator.py
yiminghub2024: Agreed, and thanks for marking it non-blocking. The patch only exists because the coordinator
change has not been released upstream yet; it should be deleted, not maintained. I will track the
upstream release and remove it once the pinned vLLM carries the change, rather than leaving it to
rot in `patch/platform/`.

vllm_ascend/models/glm5next/attention.py
yiminghub2024: Confirmed, and thanks for pinning down the framework reason rather than just the symptom.

To be explicit about the current state: the indexer is not merely mis-wired, it has no Ascend
implementation at all. `SparseAttnIndexerKpool.forward_oot` / `forward_native` both raise
NotImplementedError, because upstream's scoring path is a set of fused CUDA kernels (block-FP8
MQA logits through DeepGEMM, paged MQA logits, radix top-k over a device workspace) with no NPU
equivalent yet. So the sparse path is unreachable by construction: a checkpoint that sets
`index_topk` fails at load with an actionable message, and the dense NoPE MLA + KDA path is the
only configuration this PR claims.

That means the coupling you describe does not currently cause a wrong result, but it does mean
this PR is carrying scaffolding for a path that cannot run. I would rather remove that
scaffolding than leave a half-wired indexer in the tree, so I will strip the kpool indexer
skeleton and everything that exists only to serve it in a follow-up:

- `models/glm5next/sparse_attn_indexer_kpool.py`
- `models/glm5next/kv_cache.py` (`Glm5NextIndexerCache`, `Glm5NextTailCache`, `KpoolTailSpec`,
  `KpoolTailManager`) and its registration in `core/kv_cache_interface.py`
- `models/glm5next/ops/kpool_compress.py` (the FWHT/UE8M0 quantized compress path)
- the `is_glm5_next_kpool_cache` branches in `worker/model_runner_v1.py`,
  `worker/v2/attn_utils.py` and `attention/utils.py`

That resolves this comment together with the `kv_cache.py` and `npu_sparse_flash_attention`
comments below, without touching the dense path that the accuracy numbers were measured on.
Re-landing the indexer is then a self-contained PR that can be reviewed against a working
kernel, and the MLAAttention/Indexer decoupling you raise is the right thing to settle first.

vllm_ascend/models/glm5next/attention.py
yiminghub2024: Right — the wrapper never dispatches into `mla_modules.indexer.forward()`, so the `Indexer`
this PR constructs would never run even if it had an implementation. See my reply on the
indexer comment above: the whole kpool skeleton is coming out in a follow-up, which removes
this dead construction rather than trying to wire it into the wrapper.

vllm_ascend/models/glm5next/attention.py
yiminghub2024: Agreed, and this is worth stating clearly for anyone reading the PR later: GLM-5.3-Flash is
NoPE (`qk_rope_head_dim == 0`), and `npu_sparse_flash_attention` does not support rope_dim=0,
so the sparse path could not work on this model even with the indexer implemented. That is
independent of the wiring issue.

The dense path is unaffected: it goes through FIA with a shared read-only zero `key_rope`
buffer (the one moved onto `AscendMLAMetadataBuilder` in aaceeec), not through
`npu_sparse_flash_attention`. So this is another reason the kpool scaffolding should come out
rather than be repaired here — re-landing it needs kernel-side rope_dim=0 support first.

vllm_ascend/models/glm5next/kv_cache.py
yiminghub2024: Partly already the case, and the rest is moot once the kpool path comes out.

`KpoolTailSpec` is a `SlidingWindowSpec` subclass, so it is not a parallel spec hierarchy. It
overrides three things: `max_admission_blocks_per_request`/`max_num_blocks_per_req` return 1,
and `participates_in_prefix_caching` is False. The reason is that the tail cache is not a
sliding window over a prefix — it is one block per request holding the incomplete pool's raw K
plus gate score, overwritten in place by `pos % kpool`. `SlidingWindowManager.remove_skipped_blocks`
would evict that in-progress pool's earlier tokens mid-pool, before the pool completes and
before PD transfer, which is why `KpoolTailManager` never prunes.

`AscendSlidingWindowMLASpec` specifically would not fit: it describes an MLA latent cache
(single latent vector per state, no separate V, `real_page_size_bytes` computed from
`storage_block_size * num_kv_heads * head_size`), whereas the tail holds raw K concatenated
with a gate score.

That said — none of this is load-bearing today, because the only consumer is the sparse indexer
that raises NotImplementedError. So rather than argue the spec design, I am removing
`KpoolTailSpec`/`KpoolTailManager` along with the rest of the kpool skeleton in the follow-up
(details in my reply on `attention.py:533`). If the indexer is re-landed later with a working
kernel, the spec question is worth revisiting then, with the sliding-window option as the
starting point.

yiminghub2024: Same point as lijiahang226's comment on this line, answered there: `KpoolTailSpec` is already a `SlidingWindowSpec` subclass, and the divergence is in the manager (`KpoolTailManager` never prunes, because `SlidingWindowManager.remove_skipped_blocks` would evict the in-progress pool mid-pool). More importantly it is all dead weight -- the only consumer is the indexer that raises NotImplementedError -- so the spec and the manager are coming out with the rest of the kpool skeleton in a follow-up.

vllm_ascend/ops/triton/kda/kda.py
yiminghub2024: You are right, and the overlap is larger than it looks from the diff. I checked the AscendC KDA
ops against what this PR adds:

- `chunk_kda_fwd` already takes `use_gate_in_kernel`, `A_log`, `dt_bias`, `safe_gate` and
  `lower_bound`, i.e. exactly the fused-gate computation that `ops/triton/kda/gate.py`
  (`apply_kda_gate`) and `chunk_kda_with_fused_gate` add on the host side. `kda_gate_cumsum`
  covers the same computation standalone.
- `recurrent_kda` already takes `use_beta_sigmoid_in_kernel` and `num_accepted_tokens`, which
  are the two things the PR's `fused_recurrent_kda` parameters were added for.
- `ops/kimi_kda.py` (`AscendKimiK3DeltaAttention`) is the working reference for driving both.

So `ops/triton/kda/gate.py` should not exist, and the `ops/triton/kda/kda.py` additions should
collapse into the existing kernel flags.

I would rather not do that swap in this PR. The KDA path is what the reported GSM8K / AIME2026 /
GPQA numbers were measured on, and moving from the Triton gate to the in-kernel gate changes the
numerics of the decay accumulation (this is the same fp32-vs-activation-dtype question raised
earlier on `beta` and the gate output). That needs a re-run of the full accuracy set to be
defensible, not a same-day rewrite.

Follow-up plan, in one PR: route GLM-5.3 KDA through `chunk_kda_fwd` / `recurrent_kda` with
`use_gate_in_kernel=True` and `use_beta_sigmoid_in_kernel=True`, delete `ops/triton/kda/gate.py`
and the `chunk_kda_with_fused_gate` wrapper, and post GSM8K + AIME2026 + GPQA before/after. If
you would rather see that land before #15127, say so and I will reorder — but the dense path
here works today, so I would prefer to unblock it first.

vllm_ascend/models/glm5next/kda.py
yiminghub2024: Agreed, same as lijiahang226's comment on `ops/triton/kda/kda.py`, answered there in detail. Short version: `chunk_kda_fwd` already takes `use_gate_in_kernel` / `A_log` / `dt_bias` / `safe_gate` / `lower_bound`, and `recurrent_kda` already takes `use_beta_sigmoid_in_kernel` / `num_accepted_tokens`, so `ops/triton/kda/gate.py` and the `chunk_kda_with_fused_gate` wrapper duplicate kernel functionality that Kimi K3 already drives via `ops/kimi_kda.py`. I want to do the swap in a follow-up rather than here, because it changes the numerics of the decay accumulation on the path the reported GSM8K / AIME2026 / GPQA numbers were measured on, so it needs a full accuracy re-run to be defensible.

vllm_ascend/models/glm5next/ops/mhc_ops.py
yiminghub2024: The underlying point is right, but it lands on the wrong file — worth separating the two, because
the real problem is bigger than this one.

`ops/mhc_ops.py` is not an MHC op. It is two shape helpers, `hc_expand` (replicate
`[s, hidden]` to `[s, n, hidden]`) and `hc_contract` (mean over the stream dim), used only at the
mHC block boundaries in `model.py` (the first layer expands, the last contracts). They have no
counterpart in `npu_hc_pre_v2` / `npu_hc_post`, which are the fused pre/post mixing ops.

The actual duplication is in `patch/worker/patch_triton.py`, which rebinds upstream's
`MHCPreOp.forward_oot` and `MHCFusedPostPreOp.forward_oot` to `mhc_pre_torch` / `mhc_post_torch`
— pure-PyTorch implementations, including a hand-rolled RMSNorm, while
`torch.ops._C_ascend.npu_hc_pre_v2` / `npu_hc_post` already exist and are what
`deepseek_v4/model.py` uses. That is the real performance gap, and it is on the hot path of every
decoder layer.

I checked feasibility: GLM-5.3's config fits the kernel constraints (`hidden_size=4096`,
`mhc_num_residual_streams=4`, so `hc_mult=4` and `mix_hc=(2+4)*4=24`), and `npu_hc_pre_v2` takes
`hc_sinkhorn_iters` and both eps values, matching `mhc_sinkhorn_iterations` / `mhc_tau`.

Two things make it a follow-up rather than an edit here. First, `npu_hc_pre_v2` fuses the input
RMSNorm, so the current `MHCFusedPostPreOp` inter-layer fusion (one call doing the previous
layer's post plus this layer's pre) has to be split into `npu_hc_post` + `npu_hc_pre_v2`, and
whether that is a net win needs measuring, not assuming. Second, like the KDA swap it changes
numerics on the path the accuracy numbers came from.

Follow-up: replace the pure-torch mHC rebinding in `patch_triton.py` with the AscendC ops, with
before/after decode throughput and the accuracy set. `hc_expand`/`hc_contract` stay as they are;
they are plain shape ops with no kernel to reuse.

vllm_ascend/models/glm5next/ops/mhc_ops.py
yiminghub2024: Agreed on reusing dsv4's MHC, answered in detail on lijiahang226's comment on this file. Short version: the file you are pointing at is not the MHC op -- it is `hc_expand` / `hc_contract`, two shape helpers used only at the mHC block boundaries, with no counterpart in `npu_hc_pre_v2` / `npu_hc_post`. The real duplication is in `patch/worker/patch_triton.py`, which rebinds upstream's `MHCPreOp` / `MHCFusedPostPreOp` to pure-PyTorch `mhc_pre_torch` / `mhc_post_torch` (including a hand-rolled RMSNorm) instead of the AscendC ops that `deepseek_v4/model.py` uses. That is the follow-up worth doing, and it needs measuring because `npu_hc_pre_v2` fuses the input RMSNorm and so breaks up the current inter-layer `MHCFusedPostPreOp` fusion.

vllm_ascend/models/glm5next/ops/causal_conv1d.py
yiminghub2024: Agreed. `torch.ops._C_ascend.npu_causal_conv1d_custom` covers what this file does, including the
two things I would have expected to be blockers: it takes `num_accepted_tokens_opt` (needed for
MTP/spec-decode) and `activation_mode` / `run_mode`, and it handles GLM's `width=4` conv.

Note this is a different concern from wangxiyuan's earlier comment on line 23 of the same file,
which was about the import pointing into `_310p/` — that one is already fixed in 2fca6679 (both
entry points now import from `vllm_ascend.ops.causal_conv1d`). This one is about dropping the
PyTorch reference path entirely in favour of the AscendC operator, which is the right end state.

The one real difference is weight layout: the AscendC op expects the conv weight in a different
shape than the reference path, so the swap needs a reshape at load time plus a correctness check
against the current path. I will do it in the same follow-up as the KDA kernels, since both touch
the GLM-5.3 KDA layer and should be validated in one accuracy run rather than two.

yiminghub2024:
Rebased onto latest main — the conflict flag should clear.

Also fixed the CI failure, which was a regression from this PR rather than a flake.
`run-selected-tests / a3-8 card-(part 1-1)` was failing in
`tests/e2e/pull_request/eight_card/test_glm5_2.py::test_glm_5_2_dspark_acceptance_tp8` with:

    File "vllm_ascend/worker/model_runner_v1.py", in _reshape_kv_cache_tensors
        raw_k_tensor, raw_v_tensor = raw_cache
    ValueError: not enough values to unpack (expected 2, got 1)

Cause: this PR set `indexes_kv_by_block_stride = True` as a class-level default on
`AscendMLAAttentionSpec`. Upstream's default is False, and the flag is what
`unify_kv_cache_spec_page_size` uses to decide whether a layer whose page does not evenly divide
the maximum may be padded (`page_size_padded`) instead of having its block size scaled. Setting
it on the shared spec class opted *every* MLA model on the v1 runner into padded-page
unification, not just GLM-5.3. On GLM-5.2 + dspark at TP8 that let the SFA indexer spec and the
main MLA spec land in one KV cache tensor, and since `_allocate_kv_cache_tensors` assigns the
indexer's single-tensor tuple to every layer sharing that tensor, the MLA layer received a
1-tuple where `_reshape_kv_cache_tensors` expects `(k, v)`. GLM-5.2 on the v2 runner was
unaffected, which matches both v2 GLM-5.2 cases passing in the same job.

Fix: restore upstream's False default and opt in explicitly only where the kpool layout needs it.
Added `model_uses_kpool_indexer()`, which also replaces the duplicated `index_kpool` attribute
probes that were in `enable_sfa` and `model_uses_sfa_sparse`.

AI assistance was used for the CI triage and this fix; I have reviewed every changed line.

yiminghub2024:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

wangxiyuan:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/ops/causal_conv1d.py
```diff
@@ -64,7 +64,7 @@ def causal_conv1d_fn(
     pad_slot_id: int = PAD_SLOT_ID,
```
wangxiyuan, line 64: @Tflowers-0129  Please check this change .

yiminghub2024:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

yiminghub2024:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

wangxiyuan (approved):

Issue #15106, opened 2026-08-27
[Bug]: v0.26.0rc版本高并发下面decode性能劣化


coder-fny:
### Your current environment

A3 v0.26.0

### 🐛 Describe the bug

机器：A3
并行策略：dp4dcp8
## 并发为4的时候：

<img width="741" height="893" alt="Image" src="https://github.com/user-attachments/assets/ba0e881d-b046-4a75-884f-3f98a627bd46" />

<img width="693" height="324" alt="Image" src="https://github.com/user-attachments/assets/29e4b2f2-6ce5-47e7-89d7-32f0e1237c7f" />

## 高并发的时候

<img width="690" height="964" alt="Image" src="https://github.com/user-attachments/assets/5aa6e95e-38a6-44d6-a519-ca991b39fea7" />

<img width="1083" height="452" alt="Image" src="https://github.com/user-attachments/assets/9760a153-eca1-40d5-bc5b-a4e364fb6868" />

## stack调用栈：
<img width="1013" height="423" alt="Image" src="https://github.com/user-attachments/assets/de07381d-571d-4af3-8178-53543730343f" />

Celdrick:
大佬，请教下这个是怎么监测的？有什么软件吗？

PR #15130, opened 2026-08-27
[v0.26.0rc][bugfix] skip generate_mtp_attention_mask_for_decode with sparse attention to avoid hostbound

coder-fny:
GLM5.2 uses sparse attention and therefore doesn't need a mask. However, the current mask generation path has host-side operations that take too long at high concurrency, which makes the MTP stage host-bound and degrades performance.

Fixed issues: #15106 
- vLLM version: v0.26.0


pisceskkk (approved):
LGTM, thanks for fixing!

PR #15137, opened 2026-08-27
[Feature][xlite] Add preliminary support for GLM-5.2 with shared DSA indexer in the xlite backend

SijieFu:

Extends xlite support for GLM-5.2, which introduce a **shared indexer topk** scheme where only a subset of layers run a dedicated sparse indexer while the rest rely on a shared top-k selection. The previous adapter assumed every layer carried its own indexer weights, which breaks weight extraction and KV-cache layout for these new models.

The xlite backend was updated to `0.2.0rc1` to adapt for the changes already.

More specifically,

- `kv cache` registration is now handled per model architecture
- indexer normalization weights and biases are handled in their original precision (for GLM-5.1, BF16; for GLM-5.2, FP32)
- The `XliteModelRunner` is refactored to better incorporate `NPUModelRunner` as the fallback implementation and `XliteModelRunner`'s `XliteWrapper` as the default xlite implementation.

No, not on the API level. However, users must upgrade from `xlite==0.2.0rc0` to `xlite==0.2.0rc1` per `requirements-dev.txt`.

We benchmarked model accuracies with `GLM-5.2-w8a8` and `GLM-5.1-w8a8` using `aisbench`. The [batch_aisbench.py](https://atomgit.com/openeuler/GVirt/blob/master/xlite/tests/e2e/batch_aisbench.py) script was used with 416 questions from the `ceval` dataset on one Atlas A3:

```bash
python batch_aisbench.py "/root/benchmark" \
--num-prompts 8 \
--model-dir "/mnt/sdb/models" \
--models "GLM-5.2-w8a8" "GLM-5.1-w8a8" \
--tps 8 8 \
--dps 2 2 \
--eps 1 1 \
-MNS 64 \
-MML 8192 \
--xlite 2 1 0 \
--broadcast-xlite
```

Each model is exercised across three backends — `xlite full` (prefill + decode on xlite), `xlite decode-only` (prefill on native aclgraph, decode on xlite), and `aclgraph` (native baseline).

| Model | Dataset | TP | EP | DP | MTP | Backend | Metric | Accuracy |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| GLM-5.1-w8a8 | ceval-weighted | 8 | Y | 2 | 0 | xlite full | weighted_average | 93.19 |
| GLM-5.1-w8a8 | ceval-weighted | 8 | Y | 2 | 0 | xlite decode-only | weighted_average | 92.72 |
| GLM-5.1-w8a8 | ceval-weighted | 8 | Y | 2 | 0 | aclgraph | weighted_average | 93.42 |
| GLM-5.2-w8a8 | ceval-weighted | 8 | Y | 2 | 0 | xlite full | weighted_average | 84.67 |
| GLM-5.2-w8a8 | ceval-weighted | 8 | Y | 2 | 0 | xlite decode-only | weighted_average | 89.71 |
| GLM-5.2-w8a8 | ceval-weighted | 8 | Y | 2 | 0 | aclgraph | weighted_average | 93.19 |

No accuracy degradation observed for `GLM-5.1`. However, for `GLM-5.2`, the accuracy from the current `xlite` implementation is observably lower. This is due to a **known indexer top-k accuracy issue** by the upstream top-k compute bug at <https://atomgit.com/openeuler/GVirt/issues/27>. The fix is tracked in the xlite backend and will be picked up via an `xlite` release bump; no change is expected in this vllm-ascend docking.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -21,7 +21,7 @@ scipy
 soundfile
 pytest_mock
 mindstudio-probe>=8.3.0
-xlite==0.2.0rc0
+xlite==0.2.0rc1
 uc-manager
 ninja
 coverage
--- a/vllm_ascend/xlite/utils.py
+++ b/vllm_ascend/xlite/utils.py
@@ -33,6 +33,24 @@ from vllm_ascend.attention.sfa_v1 import AscendSFAMetadata
 
 _MISSING = object()
 """Unique sentinel for missing attributes in this module."""
+_DUMMY_TENSOR = torch.empty(0)
+"""Placeholder tensor for padding missing attributes in :meth:`get_layer_weights`."""
+
+
+def set_dummy_tensor(tensor: torch.Tensor) -> None:
+    """Rebind the module-level :data:`_DUMMY_TENSOR` placeholder.
+
+    :func:`get_layer_weights` reads :data:`_DUMMY_TENSOR` as a bare name, which resolves against this module's globals
+    at call time. Rebinding must therefore mutate *this* module's namespace: a ``from ... import _DUMMY_TENSOR``
+    followed by a plain reassignment in another module only rebinds that module's local copy and
+    leaves the placeholder here unchanged.
+
+    Args:
+        tensor (torch.Tensor): The new placeholder tensor, typically a zero-element tensor allocated on the target
+            device and in the model dtype.
+    """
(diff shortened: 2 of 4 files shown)


linfeng-yuan (approved):

lulina:
LGMT

lulina (approved):

PR #15139, opened 2026-08-27
[BugFix] Reduce Ascend opc compilation parallelism from CPU×2 to nproc to fix OOM

JavaPythonAIForBAT:

Fix OOM during nightly image build on ARM buildkitd pods (16 CPU, 24Gi memory).

The root cause is in `csrc/build.sh`'s `get_cpu_num()` function:
```bash
CPU_NUM=$(($(cat /proc/cpuinfo | grep "^processor" | wc -l)*2))
```
It defaults parallelism to **CPU count × 2** (32 on 16-core pods). 

`build_aclnn.sh` calls `build.sh --pkg --ops=... --soc=...` without passing `-j`, so the default `-j32` is used. 32 parallel `opc` (Ascend operator compiler) processes consume more memory than the 24Gi limit, causing OOM kills.

Fix: pass `-j$(nproc)` to use CPU count (=16) instead of CPU×2 (=32).

No. CI build fix only.

- Code review: verified `build.sh`'s `get_cpu_num()` and `JOB_NUM` assignment logic
- Verified the call chain: `build_aclnn.sh:252` → `build.sh --pkg` → `get_cpu_num` → `JOB_NUM="-j${CPU_NUM}"`

### DIFF
--- a/csrc/build_aclnn.sh
+++ b/csrc/build_aclnn.sh
@@ -247,9 +247,9 @@ log_selected_ops
   : "${SOC_VERSION:?SOC_VERSION is not set}"
   : "${SOC_ARG:?SOC_ARG is not set}"
 
-  log "build command: bash build.sh --pkg --ops=\"${CUSTOM_OPS}\" --soc=\"${SOC_ARG}\""
+  log "build command: bash build.sh --pkg -j$(nproc) --ops=\"${CUSTOM_OPS}\" --soc=\"${SOC_ARG}\""
   log "building custom ops ${CUSTOM_OPS} for ${SOC_VERSION}"
-  bash build.sh --pkg --ops="${CUSTOM_OPS}" --soc="${SOC_ARG}"
+  bash build.sh --pkg -j$(nproc) --ops="${CUSTOM_OPS}" --soc="${SOC_ARG}"
   log "build.sh finished"
 
   custom_ops_install_dir="${ROOT_DIR}/vllm_ascend/_cann_ops_custom"
--- a/csrc/build_aclnn.sh
+++ b/csrc/build_aclnn.sh
@@ -221,17 +221,54 @@ fi
 
 log_selected_ops
 
-# Determine optimal build parallelism.
-# - Use nproc if available, fall back to /proc/cpuinfo, then default to 1.
-# - Cap at 16 to prevent OOM on memory-constrained environments (e.g. CI buildkitd pods).
+# Determine optimal build parallelism, cgroup-aware (v1 and v2).
+# - CPU: respect cgroup quota (cpu.max or cpu.cfs_quota_us) and cpuset affinity.
+# - Memory: conservative ~1.5 GiB per opc compiler process, based on the
+#   observed OOM boundary (32 processes × 1.5 GiB ≈ 48 GiB > 24 GiB CI limit).
 # - build.sh's get_cpu_num() further caps the value via OPS_CPU_NUMBER/MAX_JOBS.
 get_build_jobs() {
-    local cpu_count
-    cpu_count=$(nproc 2>/dev/null || grep -c ^processor /proc/cpuinfo 2>/dev/null || echo 1)
-    if [ "${cpu_count}" -gt 16 ]; then
-        cpu_count=16
-    fi
-    echo "${cpu_count}"
+    local cpu_count mem_bytes mem_limit
+
+    # --- cgroup-aware CPU count ---
+    {
+        local quota= period=100000 affinity=
+        if [ -r /sys/fs/cgroup/cpu.max ]; then
+            read -r quota _ < /sys/fs/cgroup/cpu.max
+        elif [ -r /sys/fs/cgroup/cpu/cpu.cfs_quota_us ]; then
+            quota=$(cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us)
+            period=$(cat /sys/fs/cgroup/cpu/cpu.cfs_period_us)
+            [ "$quota" = "-1" ] && quota=max
+        fi
+        affinity=$(nproc 2>/dev/null || grep -c ^processor /proc/cpuinfo 2>/dev/null || echo 1)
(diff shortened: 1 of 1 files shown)


csrc/build_aclnn.sh
```diff
@@ -247,9 +247,9 @@ log_selected_ops
   : "${SOC_VERSION:?SOC_VERSION is not set}"
   : "${SOC_ARG:?SOC_ARG is not set}"
 
-  log "build command: bash build.sh --pkg --ops=\"${CUSTOM_OPS}\" --soc=\"${SOC_ARG}\""
+  log "build command: bash build.sh --pkg -j$(nproc) --ops=\"${CUSTOM_OPS}\" --soc=\"${SOC_ARG}\""
```
nu11ptr7, line 250: `-j$(nproc)` on a many-core machine still launches one compile job per core - the OOM is mitigated, not eliminated. Maybe cap it or derive it from available memory.

JavaPythonAIForBAT:
Merged into #15218

happyaron:
What about getting the CPU and RAM resources from inside the container??

```
#!/bin/sh                                                                                                                                                                                                                                                                   
  # Effective CPU + RAM available to this process, cgroup-aware (v1 and v2).                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                              
  cg_cpus() {                                                                                                                                                                                                                                                                 
      quota=max; period=100000                                                                                                                                                                                                                                                
      if [ -r /sys/fs/cgroup/cpu.max ]; then                      # cgroup v2                                                       
          read -r quota period < /sys/fs/cgroup/cpu.max                                                                             
      elif [ -r /sys/fs/cgroup/cpu/cpu.cfs_quota_us ]; then       # cgroup v1                                                       
          quota=$(cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us)                                                                          
          period=$(cat /sys/fs/cgroup/cpu/cpu.cfs_period_us)                                                                        
          [ "$quota" = "-1" ] && quota=max                                                                                          
      fi                                                                                                                            
      affinity=$(nproc)                                           # honours cpuset/taskset                                           
      [ "$quota" = max ] && { echo "$affinity"; return; }                                                                           
      n=$(( (quota + period - 1) / period ))                      # ceil                                                            
      [ "$n" -lt 1 ] && n=1                                                                                                         
      [ "$n" -lt "$affinity" ] && echo "$n" || echo "$affinity"                                                                     
  }                                                                                                                                 
                                                                                                                                     
  cg_mem_bytes() {                                                                                                                  
      lim=                                                                                                                          
      [ -r /sys/fs/cgroup/memory.max ] && lim=$(cat /sys/fs/cgroup/memory.max)                                                              
      [ -z "$lim" ] && [ -r /sys/fs/cgroup/memory/memory.limit_in_bytes ] &&                                                        
          lim=$(cat /sys/fs/cgroup/memory/memory.limit_in_bytes)                                                                    
      host=$(awk '/^MemTotal:/ {print $2 * 1024}' /proc/meminfo)                                                                    
      [ -z "$lim" ] || [ "$lim" = max ] && { echo "$host"; return; }                                                                
      [ "$lim" -gt "$host" ] && echo "$host" || echo "$lim"       # catches v1's unlimited sentinel                                 
  }
```

zhangxinyuehfad (approved):

happyaron:
为什么是写死最大16个job，在container内是否能获取到分配的资源数？

csrc/build_aclnn.sh
```diff
@@ -247,9 +247,9 @@ log_selected_ops
...
   log "building custom ops ${CUSTOM_OPS} for ${SOC_VERSION}"
-  bash build.sh --pkg --ops="${CUSTOM_OPS}" --soc="${SOC_ARG}"
+  bash build.sh --pkg -j$(nproc) --ops="${CUSTOM_OPS}" --soc="${SOC_ARG}"
```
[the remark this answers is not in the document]
JavaPythonAIForBAT: Good catch. Fixed in the latest commit: we now use a proper `get_build_jobs()` function with a three-level fallback chain (`nproc 2>/dev/null || grep -c ^processor /proc/cpuinfo 2>/dev/null || echo 1`), so the build won't fail even if `nproc` is missing in minimal containers.

csrc/build_aclnn.sh
JavaPythonAIForBAT: Addressed in the latest commit. We now use a cgroup-aware (v1/v2) `get_build_jobs()` function that:

1. Reads `cpu.max` (v2) or `cpu.cfs_quota_us` (v1) to detect the actual container CPU quota, falling back to `nproc` then `/proc/cpuinfo`.
2. Reads `memory.max` (v2) or `memory.limit_in_bytes` (v1) to derive a memory-based cap (~1.5 GiB per opc compiler process, based on the observed OOM boundary: 32 processes x 1.5 GiB ~ 48 GiB > 24 GiB CI limit).
3. Takes `min(cpu_count, mem_limit)` as the final parallelism.

This ensures OOM is eliminated, not just mitigated, regardless of host core count.

linfeng-yuan (approved):

JavaPythonAIForBAT:
@happyaron Good question. The hardcoded cap of 16 was from an intermediate commit and has been replaced in the latest commit (1d06654c7). The current implementation reads actual container resources from cgroup files:

- CPU: reads `/sys/fs/cgroup/cpu.max` (v2) or `/sys/fs/cgroup/cpu/cpu.cfs_quota_us` (v1) to get the container's CPU limit, falling back to `nproc` then `/proc/cpuinfo`.
- Memory: reads `/sys/fs/cgroup/memory.max` (v2) or `/sys/fs/cgroup/memory/memory.limit_in_bytes` (v1) to derive a safe parallelism cap (~1.5 GiB per opc process).
- Final: `min(cpu_count, mem_limit)`.

So on a container with e.g. limits.cpu=32, limits.memory=128Gi, it computes `min(32, 128/1.5=85) = -j32`. No hardcoded cap. Your cgroup-aware code snippet was adopted. Thanks!

PR #15085, opened 2026-08-27
[Refactor][CI] Align AscendKVBlockZeroer with vllm and add test for kernel

slippersss:
### What this PR does / why we need it?
This PR aims to refactor AscendKVBlockZeroer for aligning it with vllm and add ci test for zeroing kernel. The key strategy is to construct fake attention groups to meet vllm upstream interface. We also migrate vllm test and add vllm-ascend test, and warmup is also introduced as other triton kernels.

### Does this PR introduce _any_ user-facing change?
N/A

### How was this patch tested?
by ci

- vLLM version: v0.27.1
- vLLM main: https://github.com/vllm-project/vllm/commit/ba07e4a48fc951300d97eb506217dd530583dea3

PR #15152, opened 2026-08-27
[BugFix][Worker] Align compatible Full/SWA logical blocks in hybrid caches

EricLingRui:

Ascend's hybrid cache reshaper stores contiguous conv/K/V columns, while cache groups share backing buffers by logical pool block ID. Equal padded `page_size_bytes` alone does not make different K/V block strides safe: a write by one group can overlap a different, simultaneously live pool ID owned by another group.

The observed Qwen3.8-27B/DFlash2 TP2 geometry uses full-attention blocks of 1536 tokens with `2×256` K/V values per token, and draft SWA blocks of 128 tokens with `4×128` values. The flattened widths match, but the logical strides do not. A CPU probe using the actual reshaper reproduced a target V write at pool ID 20 changing 65,536 BF16 draft K elements at pool ID 0; aligning the logical widths eliminates this overlap.

For Mamba hybrid configurations, align a dense SWA logical block width to an unambiguous larger, divisible FullAttention width only when flattened K/V dimensions, dtype and quantization layout match. Preserve the attention window, retained-token policy and existing kernel-block splitting. Specialized layouts, mismatches and ambiguous full-attention widths are unchanged.

Only the model-runner implementation and its regression tests are changed.

It prevents cross-group corruption for the compatible dense hybrid layout and can change logical cache-allocation granularity. It does not increase the attention window or speculative K. Unsupported layouts are not claimed to be repaired.

#### Current rebase validation (2026-09-10)

Patch commit: `52a822a4f0fe2a2c10b554809524b136252cf91d`, rebased onto Ascend `5e062cbc85c3c9a1dc6297eb31c4884e9a660e1b`, with the base's verified vLLM main `b2f685834a6456197e7033966fdef52a23f1abcd`.

- Resolved the import conflict with GLM-Next KV management (#15913), retaining its MLA/indexer imports and tests. Preserved current upstream runner behavior and migrated the tests from `tests/ut/worker/a2/` into `tests/ut/worker/`. The cross-pool regression now exercises the real allocator and reshaper through the current standardized KV descriptors, with the existing descriptor helper preserving the v0.28.0 lane.
- `python3 -m pytest -q tests/ut/worker/test_model_runner_v1.py`: **56 passed**, no failures or skips. This is the complete module count, including the eight regression tests, not 56 new tests.
- A separate model-free NPU probe using the same allocator/reshaper and full TP2 geometry passed both cases. FullAttention uses 1536 tokens with 2x256 K/V values, SWA initially uses 128 tokens with 4x128 values, and 24 shared pages of 4 MiB are allocated on one Ascend 910B3. Writing Full V at pool ID 20 overwrites **65,536 BF16 SWA K elements at pool ID 0 with alignment disabled**, and **zero with alignment enabled**. The SWA logical block becomes 1536 while its attention window remains 2048. This is an alignment-disabled control of the same path, not a separate unpatched-upstream image run.
- Complete module plus the two NPU probe cases in one process: **58 passed**. The standalone probe must be collected with the UT module to initialize the shared test fixtures; collecting it alone outside the UT directory exposed an import-order cycle.
- CPU-only diagnostic: **55 passed, 1 failed**. The unchanged upstream `test_mtp3_placeholder_metadata_is_preserved_before_sanitizing_forward` calls `pin_memory()` and fails in this local torch-npu runtime without registered PrivateUse1 hooks. It passes in the NPU-enabled complete run above. This is not a claim of a clean local CPU-only module run.
- Complete manual pre-commit suite (including ShellCheck with `format.sh` options), `git diff --check`, and the PR commit secret scan passed.
- Runtime: CANN 9.1.0, PyTorch 2.10.0+cpu, torch-npu 2.10.0.post4. Source paths/hashes were checked. The dedicated image used vLLM's empty-device build and Ascend `COMPILE_CUSTOM_KERNELS=0`; the probe executes native NPU tensor allocation/write/count operations, not custom kernels. No model checkpoint or runtime source overlay was used. Existing service containers were not stopped or restarted.

These results verify current module behavior and the targeted NPU memory-isolation regression. They do not repeat the historical model/throughput validation below or establish arbitrary-layout, C4 or async correctness. Upstream hardware CI still needs a maintainer's `ready-precise` or `ready-all` label.

#### Historical integration validation (before this rebase)

The regression tests are in:

```bash
pytest -q tests/ut/worker/a2/test_model_runner_v1.py
```

- The repository's complete manual pre-commit suite passed locally with `pre-commit==4.6.2` and the ShellCheck options used by `format.sh`, including secret scanning. No unrelated files were changed by the checks.
- Eight focused tests cover property preservation/idempotence, no-Mamba behavior, exclusion of specialized MLA/SFA layouts, matching flattened dimensions with different head shapes, value-width mismatch, incompatible formats/block widths, ambiguous widths and cross-ID isolation through the actual reshaper. The reshaper unit test uses a scaled geometry; the separate full-geometry probe reproduced 65,536 overwritten BF16 elements before alignment and zero after it.
- After rebasing onto current `main`, the complete target module passed 45 tests in a separate Ascend test container on NPU 4. Before the rebase, the functional patch was also included in a 104-test suite of the complete attention, model-runner and GDN modules in a freshly rebuilt Ascend image. These are combined suite counts, not claims of 45 or 104 new tests.
- Built and tested Ascend base: `72059e41b239e2ea469a0fb2d9eb0cb7ee4e3085`; vLLM: `ba07e4a48fc951300d97eb506217dd530583dea3`. Full sources and custom extensions were rebuilt; installed imports and source hashes were checked, with no source overlays.
- Hardware/configuration: 2 × Ascend 910B3 64GB, Qwen3.8-27B + DFlash2, TP2/C1, BF16 weights, FP32 Mamba state, K7/block8, `FULL_DECODE_ONLY`/capture8, explicit synchronous scheduling, draft window 2048 and prefix caching enabled.
- Integrated cold/prime/first-hot checks, natural 64K/128K requests, both multimodal 5×2 grids, public GSM8K 10/300 questions, three text/tool API checks and the 262144-token success/overflow boundary passed. Public300 outputs and all speculative metrics matched the frozen optimized reference: 288/300 correct, the same 12 wrong answers and two capped responses.
- Public300 used GSM8K revision `3101c7d5072418e28b9008a6636bde82a006892c`, the first five training examples and first 300 test questions, T=0, seed 42, no thinking, max 512, natural EOS and same-request one-token priming.

The integrated service also contains separate non-causal attention, zeroer/H2D and Mamba-snapshot safeguards, plus a separately evaluated GDN metadata candidate. These are **not included in this PR**. These checks do not establish an isolated performance gain, arbitrary-layout compatibility or C4/async correctness. The capacity probe is synthetic text, not a 262K visual-quality claim; private request texts and logs are not attached.

### Related work

- #14340 proposes a broader non-contiguous hybrid-cache reshape and kernel-block interface refactor. This PR addresses the current contiguous-column implementation; its necessity should be reassessed if that refactor lands.
- #14543 fixes K+V page-size derivation in platform configuration, rather than this Full/SWA cross-group stride mismatch.
- #14693 detects MLA/Mamba cross-ID overlap; this PR repairs the compatible dense Full/SWA case.
- #15085 overlaps the separately documented H2D/zeroer prerequisite, which is not part of this patch.

The historical validation above followed a rebase onto `4b25eb34793efa9f00201d654eb2f107ecebfbbd`; the current rebase and test results are listed separately above.

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -77,11 +77,13 @@ from vllm.v1.core.sched.output import SchedulerOutput
 from vllm.v1.kv_cache_interface import (
     AttentionSpec,
     EncoderOnlyAttentionSpec,
+    FullAttentionSpec,
     HiddenStateCacheSpec,
     KVCacheConfig,
     KVCacheGroupSpec,
     KVCacheSpec,
     MambaSpec,
+    SlidingWindowSpec,
     UniformTypeKVCacheSpecs,
 )
 from vllm.v1.outputs import (
@@ -5710,6 +5712,7 @@ class NPUModelRunner(GPUModelRunner):
                 if spec := mamba_module.get_kv_cache_spec(self.vllm_config):
                     kv_cache_spec[layer_name] = spec
                     mamba_page_size_padded = spec.page_size_bytes
+            _align_hybrid_sliding_window_block_sizes(kv_cache_spec)
             # align attn_page_size to mamba_page_size_padded
             for layer_name in attn_layer_names:
                 if kv_cache_spec[layer_name].page_size_bytes < mamba_page_size_padded:  # type: ignore[attr-defined]
@@ -5854,6 +5857,46 @@ class NPUModelRunner(GPUModelRunner):
             runner_only_attn_layers=self.runner_only_attn_layers,
             static_forward_context=(self.compilation_config.static_forward_context),
         )
+
+
+def _align_hybrid_sliding_window_block_sizes(specs: dict[str, KVCacheSpec]) -> None:
+    """Keep dense SWA and full-attention pages on the same hybrid byte layout.
+
+    Ascend stores hybrid caches as contiguous conv/K/V columns, rather than
+    interleaving the components of each logical page. Padding page_size_bytes
+    alone therefore cannot make different K/V block strides safe to share:
+    a full-attention block can overlap a *different* SWA block ID. Match the
+    logical block width when the two dense attention formats have the same
+    per-token K/V widths. Head counts and dimensions may factor differently.
+    The backend still splits these pages into its supported kernel blocks.
+    """
+    if not any(isinstance(spec, MambaSpec) for spec in specs.values()):
+        return
+
+    def layout_key(spec: AttentionSpec) -> tuple:
+        return (
+            spec.num_kv_heads * spec.head_size,
+            spec.num_kv_heads * (getattr(spec, "head_size_v", None) or spec.head_size),
+            spec.dtype,
+            getattr(spec, "cache_dtype_str", None),
+            getattr(spec, "kv_quant_mode", None),
+        )
+
+    full_block_sizes: dict[tuple, set[int]] = defaultdict(set)
+    for spec in specs.values():
+        # Restrict this helper to the exact dense layouts validated below.
+        # MLA and other subclasses can use different physical cache layouts.
+        if type(spec) is FullAttentionSpec:
+            full_block_sizes[layout_key(spec)].add(spec.block_size)
+    for name, spec in specs.items():
+        if type(spec) is not SlidingWindowSpec:
+            continue
+        matching = full_block_sizes.get(layout_key(spec), set())
+        if len(matching) != 1:
+            continue
+        block_size = next(iter(matching))
+        if block_size > spec.block_size and block_size % spec.block_size == 0:
+            specs[name] = replace(spec, block_size=block_size)
 
 
 def _post_process_cudagraph_mode(tensor: torch.Tensor) -> int:
--- a/tests/ut/worker/test_model_runner_v1.py
+++ b/tests/ut/worker/test_model_runner_v1.py
@@ -1,6 +1,7 @@
 import unittest
 from collections import deque
 from contextlib import nullcontext
+from dataclasses import replace
 from types import SimpleNamespace
 from unittest.mock import MagicMock, call, patch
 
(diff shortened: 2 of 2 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -4931,6 +4934,45 @@ def _init_kv_zero_meta(self) -> None:
...
+            spec.num_kv_heads * (getattr(spec, "head_size_v", None) or spec.head_size),
+            spec.dtype,
+            getattr(spec, "cache_dtype_str", None),
+            getattr(spec, "kv_quant_mode", None),
+        )
+
+    full_block_sizes: dict[tuple, set[int]] = defaultdict(set)
+    for spec in specs.values():
+        # MLA and other specialized cache layouts need their own handling.
+        if type(spec) is FullAttentionSpec:
```
nu11ptr7, line 5258: `type(spec) is FullAttentionSpec` only matches the exact base class - an Ascend subclass of FullAttentionSpec would be silently skipped and never aligned. `isinstance` would be safer.

EricLingRui: Thanks for raising this. The exact-type check is intentional: `MLAAttentionSpec` also inherits from `FullAttentionSpec`, but its storage layout is outside this helper's scope. The current layout key does not capture all of those differences, so a blanket `isinstance()` check would admit unvalidated layouts.

I agree that compatible subclasses should not be overlooked. I'd prefer to keep this fix narrowly scoped, make the restriction explicit, and add regression coverage for the excluded types. If there is a particular dense Ascend subclass that should be covered, I'm happy to check its layout and consider explicit support.

PR #15157, opened 2026-08-28
[Misc] remove compatibility with dispatch_v1/combine_v1 interface

zzzzwwjj:

Remove compatibility with `torch_npu.npu_moe_distribute_dispatch`/`torch_npu.npu_moe_distribute_combine` interface, which is only used in very old version.

None.

- vLLM version: v0.27.1

### DIFF
--- a/vllm_ascend/ops/fused_moe/token_dispatcher.py
+++ b/vllm_ascend/ops/fused_moe/token_dispatcher.py
@@ -119,7 +119,6 @@ class TokenDispatcherWithMC2(MoETokenDispatcher[MoEMC2CombineMetadata]):
         self.moe_all_to_all_group_name = backend.get_hccl_comm_name(local_rank)
         self.ep_rank_id = get_mc2_group().rank_in_group
         self.ep_world_size = get_mc2_group().world_size
-        self.enable_dispatch_v2 = hasattr(torch_npu, "npu_moe_distribute_dispatch_v2")
         self.need_extra_args = get_ascend_device_type() in [AscendDeviceType.A3, AscendDeviceType.A5]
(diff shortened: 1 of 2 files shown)


vllm_ascend/ops/fused_moe/token_dispatcher.py
```diff
@@ -237,11 +231,7 @@ def token_dispatch(
         token_dispatch_input: MoETokenDispatchInput,
     ):
         kwargs_mc2 = self.get_dispatch_mc2_kwargs(token_dispatch_input)
-        output = (
-            torch_npu.npu_moe_distribute_dispatch_v2(**kwargs_mc2)
-            if self.enable_dispatch_v2
-            else torch_npu.npu_moe_distribute_dispatch(**kwargs_mc2)
-        )
+        output = torch_npu.npu_moe_distribute_dispatch_v2(**kwargs_mc2)
```
nu11ptr7, line 234: The `hasattr(torch_npu, "npu_moe_distribute_dispatch_v2")` guard and the RuntimeError were removed - on older PTA/CANN without dispatch_v2, this now raises AttributeError mid-request instead of a clear upgrade message. If v2 is a hard requirement now, an init-time check would keep the error actionable.

ningjingbengxiaohai (approved):

vllm_ascend/ops/fused_moe/token_dispatcher.py
zzzzwwjj: `torch_npu.npu_moe_distribute_dispatch_v2` has been supported since 25Q3, and this compatibility check has been maintained for a sufficient period of time.

PR #15113, opened 2026-08-27
[0.26.0][feature] add param additional_config.mc2_comm_alg

zzzzwwjj:
### What this PR does / why we need it?

main pr: https://github.com/vllm-project/vllm-ascend/pull/15094
add param additional_config.mc2_comm_alg, support `""/"fullmesh"/"hierarchy"/"fullmesh_v2"` four types, and "fullmesh_v2" may get better performance in some cases.

### Does this PR introduce _any_ user-facing change?

+ Add new param: additional_config.mc2_comm_alg;
+ additional_config.enable_mc2_hierarchy_comm will be deprecated and be replaced by mc2_comm_alg = "hierarchy";

### How was this patch tested?

- vLLM version: v0.26.0
- vLLM main: https://github.com/vllm-project/vllm/commit/d02df748bf9efd99022f1a062597dc3cb3808485

PR #15162, opened 2026-08-28
[Performance] Stabilize Qwen3.8 Flash-Next and optimize QSA on Ascend

shaopeng-666:
# Problem

Qwen3.8 Flash-Next could not complete profile, `FULL_DECODE_ONLY` capture, and
replay on Ascend A3 when using the hybrid QSA/PLE cache layout. The failures
were independent but surfaced sequentially:

- packed KV-cache descriptors were allocated as independent physical buffers,
  multiplying profile-run HBM usage;
- a non-contiguous PLE cache row selection materialized almost the entire
  convolution cache, and the short depthwise convolution was not graph-safe;
- PLE n-gram context used a temporary device address, causing invalid-GM replay
  failures (holding the captured tensor alone would instead replay stale data);
- the QSA Triton pre-indexer and dynamic cache-update path were not portable to
  this graph/compiler path;
- Qwen4Exp MTP did not handle both vLLM hybrid-cache group representations or
  the model's full hyper-connection hidden width.

# Baseline

- Target repository/base: `vllm-project/vllm-ascend:releases/v0.26.0rc`
- Original audited A snapshot:
  `85625bd772f27a4a13bc8e548f131af5ce3ff734`
- Live base head at PR creation:
  `9c01df962a245346c904c9d224c5ad77e1212b0a` (the release branch advanced by
  the unrelated #15113 after the original audit)
- Prerequisite B: #15072, head
  `5d5bf084323cec9a4bc47d6dca1dfbfe0928ec58`
- GitHub test merge for #15072:
  `10c7bc0c433118b82991b85a7e0aba199615097b`

#15072 is not yet merged into the target release branch, so the GitHub PR diff
necessarily contains B + C. The commits and diffstat under "C-only commits"
below are the self-contained increment of this PR measured from an exact local
live-base + B merge (`15aae5d5e7cff3358ff4dd3a62760f31da075242`).
This PR does not duplicate or rewrite #15072.
All C commits carry the required DCO sign-off. The two prerequisite commits are
preserved byte-for-byte; their existing DCO findings must be resolved in
#15072 rather than by rewriting another author's commits here.

# Verified software and runtime

The real-weight Ascend A3 validation used the following exact vLLM build:

- installed package version: `vllm==0.26.0+empty`
- source installation command: `VLLM_TARGET_DEVICE=empty pip install --no-deps -v -e .`
  `9065212a215e5f244aa7a8e4d8c0d468e7950aba`

The validated real-weight topology was Ascend A3 DP2 x TP8 x EP16 with 16
workers, `FULL_DECODE_ONLY` for the main model, eager MTP with three draft
tokens, `max_model_len=71680`, `max_num_batched_tokens=4096`, and
`max_num_seqs=32`.

# Changes (C only)

1. Reuse one physical backing allocation for packed KV-cache descriptors while
   retaining a separate hidden-state backing for mixed Mamba/hidden owners.
   Logical offsets, views, and cache-sizing metadata remain unchanged.
2. Make PLE state access and convolution graph-safe:
   - gather/scatter through the contiguous storage layout;
   - replace the tiny dilated convolution with explicit contiguous slices;
   - reserve speculative state capacity and remove the runtime-owned
     `query_start_loc` from model dynamic arguments.
3. Keep PLE n-gram input address-stable and value-fresh with a runner-lifetime
   int32 buffer updated by a blocking per-step copy. Also install the current
   vLLM config while preparing async-DP full-graph metadata.
4. Replace only the QSA pre-indexer with the portable
   `project_qk -> update_and_compress -> select` chain. Avoid host-sync equality
   checks and data-dependent `masked_select` cache writes; keep the main QSA
   attention backend unchanged.
5. Support Qwen4Exp MTP hybrid cache groups in both direct and uniform forms,
   forward every group block table, allocate the full hyper-connection hidden
   buffer, and normalize `MultipleOf` block-size constraints.
6. Align the QSA/GDN/PLE physical cache storage as six independently contiguous
   slabs while retaining the logical cache groups. QSA main and compressed
   caches share a composite group and block table (`B=768`, 192 compressed
   rows); the raw circular index cache remains a separate `C=8` group.
7. Load sparse QSA keys in source-contiguous `[token, head_dim]` order and
   transpose in UB, avoiding the scalarized scattered-dimension lowering.
8. Bound PLE short-conv window materialization to 16 output positions per
   chunk, and classify prefill/decode from authoritative request-phase metadata
   rather than query length. Graph-capture dummy MTP inputs now carry the same
   accepted/draft-token metadata contract as real execution.
9. Add focused allocator, PLE, QSA, MTP, six-slab, phase-classification, and
   graph-capture regression tests.

# Why FULL_DECODE_ONLY main graph + MTP eager previously failed

The main runner needs stable graph inputs and graph-compatible PLE/QSA cache
updates. At the same time, the speculative runner must honor
`enforce_eager=true` without inheriting the main runner's graph capture, while
still receiving all Qwen4Exp hybrid cache groups and the full HC hidden state.
Before these changes, the main graph could fail during capture/replay and the
MTP proposer had incomplete cache/hidden metadata. The changes isolate those
two execution modes: the main runner remains `FULL_DECODE_ONLY`; MTP stays out
of `ACLGraphWrapper` and runs eager.

# Single-host reproduction command

**Validated with dummy weights; this does not establish real-weight
accuracy.**

The recorded A3 functional pass used a single host with 16 NPUs. One
`vllm serve` process created both data-parallel ranks with topology DP2 x TP8
and expert parallel enabled (effective EP16). The exact sanitized command was:

```bash
unset CPLUS_INCLUDE_PATH CPATH C_INCLUDE_PATH

export RUNTIME_OVERLAY="<RUNTIME_OVERLAY>"
export MODEL_PATH="<MODEL_PATH>"
export LOCAL_IP="<LOCAL_IP>"
export PYTHONPATH="${RUNTIME_OVERLAY}${PYTHONPATH:+:$PYTHONPATH}"
export PYTHONUNBUFFERED=1

exec vllm serve "$MODEL_PATH" \
  --host 0.0.0.0 \
  --port 8010 \
  --served-model-name qwen3.8-flash-next \
  --trust-remote-code \
  --load-format dummy \
  --tensor-parallel-size 8 \
  --data-parallel-size 2 \
  --data-parallel-size-local 2 \
  --data-parallel-start-rank 0 \
  --data-parallel-address "$LOCAL_IP" \
  --data-parallel-rpc-port 29550 \
  --enable-expert-parallel \
  --max-model-len 131072 \
  --max-num-seqs 1 \
  --max-num-batched-tokens 256 \
  --gpu-memory-utilization 0.95 \
  --no-enable-prefix-caching \
  --language-model-only \
  --speculative-config '{"method":"qwen3_5_mtp","num_speculative_tokens":3,"enforce_eager":true}' \
  --additional-config '{"enable_cpu_binding":true,"ascend_compilation_config":{"fuse_norm_quant":false}}'
```

`--async-scheduling` is intentionally absent: the successful
`FULL_DECODE_ONLY` capture/replay baseline used synchronous scheduling. Prefix
caching was disabled, and no GPU-block override or explicit KV-cache-memory
override was used.

Substitutions:

- `<LOCAL_IP>`: the single host's routable data-parallel address.
- `<MODEL_PATH>`: the model directory visible inside the prepared runtime.
- `<RUNTIME_OVERLAY>`: the directory containing this PR's installed Python
  overlay; omit the `PYTHONPATH` line when running from an installed wheel.

The command above is retained as the historical dummy-weight reproduction.
Real Qwen3.8-Flash-Next weights were subsequently validated by removing
`--load-format dummy` and using the software/runtime configuration documented
in **Verified software and runtime** above.

# Validation

Historical A3 functional validation used dummy weights with DP2 x TP8 x EP16,
16 workers, 128K maximum context, `gpu_memory_utilization=0.95`, and no block
override.

- Main graph without MTP: 1 frontend, 2 EngineCore processes, 16 workers and
  all 16 EP ranks; `torch.compile` completed; FULL decode graph capture was
  `1/1` (about 13.32 s and 0.66 GiB graph memory); `/v1/models` became ready;
  a short request returned HTTP 200 and 8 non-empty tokens; all workers entered
  graph replay.
- Main `FULL_DECODE_ONLY` + MTP eager: only the speculative configuration was
  enabled (`qwen3_5_mtp`, 3 draft tokens, `enforce_eager=true`); main capture
  completed (about 14.23 s); source/config/log evidence confirmed MTP did not
  enter `ACLGraphWrapper`; API became ready and a short request returned HTTP
  200 and 8 non-empty tokens with EngineCore/worker replay evidence.
- Isolated historical A3 checks covered PLE convolution equivalence, persistent
  n-gram address plus cross-request value refresh, QSA eager equivalence, and
  two QSA NPU graph replays.
- Repository checks in this checkout: `git diff --check`, conflict-marker scan,
  120-column scan, Git connectivity, and bounded secret/debug-artifact scans
  pass.
- Real-weight A3 validation: the 16-worker DP2 x TP8 x EP16 service completed
  graph capture and reached `/health=200`; a deterministic streaming request
  asking for `193 + 386` returned the correct final answer `579`, and the
  service remained healthy afterwards.
- QSA Variant-A validation covered every context length 1 through 4096 for both
  decode and 20-row shapes (8192 baseline comparisons, zero BF16 failures),
  plus FP32-reference checks at 72 boundary/representative points with zero
  failures.
- Targeted repository validation completed with `compileall`, Ruff 0.14.0
  checks, and 47 passing unit tests covering six-slab layout/group semantics,
  QSA lowering, bounded PLE convolution, authoritative P/D classification, and
  MTP capture metadata. The exact `[D4,D4,D4,P4084]` regression yields three
  decode requests and one prefill request without hard-coding the decode width.

# Limitations / pending full evaluation

- A complete, uninterrupted GPQA-Diamond accuracy result is still pending; no
  full-dataset score is claimed here.
- At approximately 15K generated tokens under eight concurrent long-running
  requests, HCCL `allGather` latency increased sharply at about 94.5% KV-cache
  utilization. QSA kernel latency itself remained stable; this communication
  long-tail is not addressed by this PR.
- A5 MXFP/W8A8 experiments, transfer/launch scripts, logs, response artifacts,
  credentials, environment-specific diagnostics, and vLLM-main changes are
  intentionally excluded.

# C-only commits

- `da4cb3b355b589afa9ebc7b1a30fd5b5abfd7728` - reuse packed KV-cache backing
- `cc66df1a311f6862e3c29164dd2aa6b438450721` - make Qwen4Exp PLE graph-safe
- `73f9a70c29e360fba8fc005ac0e01c92a32e4ac6` - support Qwen4Exp hybrid MTP caches
- `1110ffe51b6fae3b3bd24cf8114b1dd75aa816b9` - use graph-safe Ascend QSA indexer fallback
- `3804775939e75afe4474f1c013f7c728fee7af97` - align Qwen3.8 hybrid KV cache layout
- `9065212a215e5f244aa7a8e4d8c0d468e7950aba` - stabilize PLE and optimize QSA

# Exact C-only diffstat (relative to A + #15072)

```text
24 files changed, 2495 insertions(+), 353 deletions(-)
```

- Validated vLLM: [`v0.26.0`](https://github.com/vllm-project/vllm/commit/568afb3a13806beb53bb2e6bd518269357b237c0)


vllm_ascend/patch/platform/patch_speculative_config.py
```diff
@@ -11,6 +12,13 @@
 
 _orig_post_init = SpeculativeConfig.__post_init__
 
+# vLLM 0.26 predates Qwen4Exp MTP. Extend its runtime validation list before
+# SpeculativeConfig normalizes the draft model returned by hf_config_override.
+if "qwen4_exp_mtp" not in get_args(vllm_speculative.MTPModelTypes):
+    vllm_speculative.MTPModelTypes = Literal.__getitem__(
```
nu11ptr7, line 18: Rebinding `MTPModelTypes` at import only fixes modules that read it after this patch - anything that already did `from vllm.config.speculative import MTPModelTypes` keeps the old Literal and still fails validation. Mutating a Literal at runtime is also fragile; better to handle the check explicitly.

Issue #14347, opened 2026-08-15
[v0.23.0][Bug]: In the graph mode scenario, the operators `torch.ops._C_ascend.mla_preprocess` and `torch.ops._C_ascend.batch_matmul_transpose` are used. In this case, plog will produce the following error.


ZT-AIA:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
Your output of above commands here
```

</details>

### 🐛 Describe the bug

In the graph mode scenario, the operators `torch.ops._C_ascend.mla_preprocess` and `torch.ops._C_ascend.batch_matmul_transpose` are used. In this case, plog will produce the following error, **but it will not affect the actual usage**.
[ERROR] RUNTIME(212914,):2026-03-26-14:23:18.204.183 [api_c.cc:1097]212914 rtMemcpy:ErrCode=107030, desc=[the current capture mode does not support this operation], InnerCode=0x703001c
[ERROR] RUNTIME(212914,):2026-03-26-14:23:18.204.187 [error_message_manager.cc:61]212914 FuncErrorReason:rtMemcpy execution failed, reason=the current capture mode does not support this operation

AbarnaaSree:
I'd like to investigate this issue and see whether the rtMemcpy call from mla_preprocess / batch_matmul_transpose can be avoided or handled correctly during graph capture. Since the error does not currently affect functionality, I can first trace where the unsupported rtMemcpy is triggered and add a regression test if appropriate. Would it be okay for me to work on this and open a PR?

slippersss:
@ZT-AIA Thank you for pointing out this issue. Could you please help take a look at it and resolve it?

PR #15181, opened 2026-08-28
[BugFix][Kernel] Fix missing V-to-S synchronization in MLA preprocess

Foriv:

The BF16 `PER_TOKEN_SYMM_QUANT` paths compute the per-token maximum with `ReduceMaxCustom` on the Vector pipeline and then read `max.GetValue(0)` on the Scalar pipeline. `PipeBarrier<PIPE_V>()` only orders operations within the Vector pipeline; it does not establish visibility from Vector to Scalar. The raw launcher timing could mask this race, while validation through the standard ACLNN launch path exposed stale and nondeterministic Quant1 results.

This PR adds an explicit `SET_FLAG(V, S, EVENT_ID0)` / `WAIT_FLAG(V, S, EVENT_ID0)` pair before each of the four Scalar reads in the normal and qdown BF16 templates.

The patch does not change the quantization formula, tensor shapes, tiling, launch configuration, supported modes, or public Torch API. It is the correctness prerequisite for a separate MLA ACLNN/NPU Graph refactor.

Refs #14347.

Yes. It prevents stale or nondeterministic output in the existing per-token symmetric quantization path. There is no public API or supported-mode change.

Current commit: `2a9a5780ff1a505bf8b8a53eac8427f421c40e1b`, rebased onto upstream `main` at `2617fb6abf541145ea7f12d1dd39c757f1180b66`.

NPU correctness validation was performed on pre-rebase commit `6554393787f02f1a85f02b311da4158337009618`. Its stable patch ID is identical to the current commit, so the rebase changed commit ancestry but not the two-file, eight-line kernel patch.

Environment: Ascend 910B4, CANN 9.1.0, Python 3.11.10, and torch-npu 2.10.0.post4.

- `bash format.sh ci` passed all hooks on a clean worktree at the current commit.
- The existing `tests/e2e/nightly/single_node/ops/singlecard_ops/test_mla_preprocess.py` suite passed on the patch-equivalent pre-rebase commit: `13 passed, 32 warnings in 19.52s`. Torch-NPU internal formats were enabled by the external test launcher, matching the vLLM-Ascend model runner configuration.
- An external instrumented A/B probe confirmed that all 24,576 bytes entering Quant1 were identical before the affected reduction. Without the V-to-S event, repeated output comparisons mismatched 24,575, 22,648, and 22,516 of 24,576 bytes. With only this synchronization change, all 24,576 bytes matched in each of three runs.
- External fixed-input checks for both normal and qdown paths matched the raw launcher's stable hot-state output in three of three runs each.
- `git diff --check upstream/main...HEAD` passed after the rebase; the PR still changes only the two intended kernel files.
- GitHub [E2E run 33148586547](https://github.com/vllm-project/vllm-ascend/actions/runs/33148586547) passed pre-commit, coverage recommendation, and test selection. The selected tests did not start because this Ready PR does not yet have the maintainer-controlled `ready-precise` or `ready-all` label; the resulting `ci-gate` failure is the expected missing-label gate rather than a source or test failure.

No test files are added or modified. The deterministic A/B probe requires an instrumented NPU build and is kept as external validation rather than a timing-sensitive CI assertion. The full repository test suite was not run.

- vLLM version: v0.27.1

### DIFF
--- a/csrc/mla_preprocess/op_kernel/mla_preprocess_mix_bf16.hpp
+++ b/csrc/mla_preprocess/op_kernel/mla_preprocess_mix_bf16.hpp
@@ -513,6 +513,8 @@ public:
                 AscendC::PipeBarrier<PIPE_V>();
                 ReduceMaxCustom(max, abs, work, num_col_ - input_stride_);
                 AscendC::PipeBarrier<PIPE_V>();
+                SET_FLAG(V, S, EVENT_ID0);
+                WAIT_FLAG(V, S, EVENT_ID0);
                 float scaleOut = max.GetValue(0) / 127;
                 SET_FLAG(S, V, EVENT_ID0);
                 WAIT_FLAG(S, V, EVENT_ID0);
@@ -817,6 +819,8 @@ public:
                 AscendC::PipeBarrier<PIPE_V>();
                 ReduceMaxCustom(max, abs, work, num_col_ - input_stride_);
                 AscendC::PipeBarrier<PIPE_V>();
+                SET_FLAG(V, S, EVENT_ID0);
+                WAIT_FLAG(V, S, EVENT_ID0);
                 float scaleOut = max.GetValue(0) / 127;
                 SET_FLAG(S, V, EVENT_ID0);
                 WAIT_FLAG(S, V, EVENT_ID0);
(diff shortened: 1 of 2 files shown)


Foriv:
Superseded by #15182, which now targets main directly and carries the final standalone solution: the ACLNN two-stage/graph-safe Host Tiling migration plus compiler auto-sync. The hand-written V-to-S synchronization patch is therefore no longer needed. The replacement has passed the operator build, repository lint hooks, the existing MLA test suite, graph replay checks, and fixed-input golden comparisons.

PR #15192, opened 2026-08-28
 [v0.26.0][Feature][Kimi K3 DSPark] Enable TP for context_proj

Levi-JQ:
Switch context_proj from ReplicatedLinear to ColumnParallelLinear (gather_output=True) so the projection weight [7168, 35840] is sharded along the output dim (7168) across TP ranks instead of being replicated.

This is not just a memory saving. During decode the context_proj GEMM is memory-bandwidth bound: profiling shows aic_mte2_ratio ~94% vs aic_mac_ratio ~8% (weight-load bound, batch=8). With the weight replicated, every rank re-reads the full ~490MB weight each step and the cluster bandwidth is not distributed. Sharding the output dim cuts the per-rank HBM weight read by TPx and spreads the traffic across ranks. gather_output=True reassembles [8, 7168] for context_norm, leaving the combine_hidden_states caller unchanged. Numerically equivalent to the replicated path; checkpoint layout unchanged.

idea from ：https://github.com/vllm-project/vllm-ascend/pull/15144

### DIFF
--- a/vllm_ascend/models/kimi_k3_dspark.py
+++ b/vllm_ascend/models/kimi_k3_dspark.py
@@ -33,7 +33,6 @@ from vllm.model_executor.layers.layernorm import RMSNorm
 from vllm.model_executor.layers.linear import (
     ColumnParallelLinear,
     MergedColumnParallelLinear,
-    ReplicatedLinear,
     RowParallelLinear,
 )
 from vllm.model_executor.layers.logits_processor import LogitsProcessor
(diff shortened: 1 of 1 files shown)


drslark (approved):
Same as https://github.com/vllm-project/vllm-ascend/pull/15144/changes.
LGTM.

drslark:
Add [v0.26.0] to title.

Levi-JQ:
> Add [v0.26.0] to title.
done

PR #15196, opened 2026-08-28
[feature] reduce cases that required DP padding in mrv1

zzzzwwjj:

Reducing the cases that required DP padding.

Before this pr, if one of the three cases is met, padding is required:
1. cudagraphmode != None;
2. enable_sp == True;
3. is_draft_model == True;

This is not reasonable, some cases do not require padding across dp.

And after this pr, there are three cases where padding between DPs is required:
1. comm_method == ALLGATHER;
2. comm_method == MC2, reduce communication and computation through active_mask to enhance performance;
3. when finegrained_tp is open, we need to ensure num_tokens remains consistent within finegrained_tp_group.

None

### DIFF
--- a/vllm_ascend/spec_decode/extract_hidden_states_proposer.py
+++ b/vllm_ascend/spec_decode/extract_hidden_states_proposer.py
@@ -97,7 +97,6 @@ class AscendExtractHiddenStatesProposer(ExtractHiddenStatesProposer):
                 num_tokens=num_tokens_padded,
                 is_draft_model=True,
                 cudagraph_mode=cudagraph_mode,
-                allow_dp_padding=use_cudagraphs,
             )
 
             if num_tokens_across_dp is not None:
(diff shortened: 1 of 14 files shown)


vllm_ascend/spec_decode/extract_hidden_states_proposer.py
```diff
@@ -138,10 +137,12 @@ def dummy_run(
         # llm_base_proposer.dummy_run); otherwise the DP cpu_group collectives
         # desynchronize and the group deadlocks.
         (
-            num_tokens,
+            _,
```
Levi-JQ: Is this a redundant edit?

yiz-liu (approved):
Well, I think this is acceptable since it’s more aligned with the actual code paths that require padding, rather than following the upstream logic. But we do need to thoroughly test this change.

weijinqian0 (approved):

slippersss (approved):

PR #15197, opened 2026-08-28
[main][Attention][Feature] Rewrite SFA DCP sparse index remap with Triton

coder-fny:

  Rewrite AscendSFADCPImpl._remap_sparse_indices (the DCP sparse-attention topk index remap, which runs once per SFA layer on the decode path) with Triton
  kernels:

  - New vllm_ascend/ops/triton/sparse_index_remap.py:
    - a fused kernel (vectorized fp32 remap + per-chunk compaction + output tail pre-fill, grid over (chunk, row));
    - a vectorized gather kernel that concatenates chunk results in top-k order;
    - a single-kernel fast path (constexpr NUM_CHUNKS == 1) when topk_count <= 128 that skips the gather kernel entirely.
  - vllm_ascend/attention/context_parallel/sfa_cp.py routes _remap_sparse_indices to the Triton path on NPU (HAS_TRITON and topk_indices.is_npu) and keeps the previous torch implementation as the fallback.
  - Extended tests/ut/attention/a2/test_sfa_cp_precision.py with an independent torch reference and new cases: interleave_size == 1, parametrized 2D cross-checks, and the real 3D DCP shape [dcp_size, 1, topk_count].

  Why: on GLM5.2 DCP decode the previous implementation dispatches ~30 small host-side ops per SFA layer per step (profiling:
  Cast/Floor/SelectV2/Sort/Gather chain with [6,1,2048] tensors), which contributes to the host-bound behavior on Ascend. The rewrite reduces this to 1-2 kernels.

  Measured on Ascend 910B (triton-ascend 3.2.0), standalone single-op benchmark vs the torch implementation:

image

  Numerical semantics are unchanged: the Triton kernels use the same fp32 index math as the previous implementation and match it bit-exactly (indices are far below the 2^24 fp32 exactness bound).

  Development notes for future maintainers (documented in the kernel comments): on triton-ascend 3.2.0 / CANN 9.1, tl.cumsum + tl.sum based compaction miscompiles (nondeterministic garbage values), tl.cumsum alone carries a large per-program fixed cost, an in-kernel literal -1 tail fill writes wrong values, and tl.sort is unavailable. The current design deliberately avoids all of these; re-validate if the triton-ascend version changes.

  No. Internal change to the DCP SFA decode path only; no API, config, or behavior change. The previous torch implementation is retained as the fallback for non-NPU / no-Triton environments.

  - Unit tests: extended tests/ut/attention/a2/test_sfa_cp_precision.py — existing cases plus interleave=1, parametrized 2D cross-checks (dcp_size 2/4 ×
  interleave 1/2/4 × topk 1/8/48/256 × all ranks) and the 3D [6,1,2048] case against an independent torch reference (requires NPU CI; not yet run on the v0.26 CI runner).
  - On-device single-op verification (Ascend 910B, torch 2.10 + torch_npu 2.10.0.post4 + triton-ascend 3.2.0): correctness 144/144 bit-exact vs the torch reference including -1 padding and out-of-bounds lanes; 10x repeated runs on the previously failing input patterns; 200 back-to-back unsynchronized launches all bit-exact (regression guard against the miscompile class found during development).
  - End-to-end model precision on a DCP>1 environment is still pending and will be done separately.
 
before：
image
after：
image


coder-fny:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

ningjingbengxiaohai (approved):

PR #15212, opened 2026-08-28
[MRV2][Performance][Kernel] Optimize V2 slot mappings on Ascend

Liamup777:

MRv2 `AscendBlockTables` inherits the upstream slot-mapping algorithm. On
Ascend, its per-token indirect block-table loads and integer remainder
calculation cause high scalar overhead, especially as the number of request
programs or token tiles grows.

This PR keeps the existing V2 interface and moves the Ascend-specific Triton
kernel to `vllm_ascend/ops/triton/v2/block_table/compute_slot_mappings.py`.
`AscendBlockTables` remains responsible for deriving launch metadata and
launching the kernel. The kernel:

- use int32 position arithmetic;
- replace the int32 remainder with multiply/subtract;
- load each request's block-table row contiguously once and use `tl.gather`
  for token-to-block lookup; and
- derive a power-of-two compile-time row bound from the allocated block-table
  stride, while masking the runtime load by each group's actual stride.

The row bound uses the allocated stride rather than
`max_num_blocks_per_group`, because one KV block can map to multiple kernel
blocks.

No. This is an internal Ascend MRv2 kernel optimization.

- `bash format.sh`
- `bash format.sh ci`
- `SHELLCHECK_OPTS="--exclude=SC2046,SC2006,SC2086" pre-commit run --all-files --hook-stage manual --show-diff-on-failure`
- NPU test:

  ```bash
  /usr/local/python3.12.13/bin/python -m pytest -q -s \
    tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_compute_slot_mapping.py
  ```

  Result: `4 passed`, covering CP=1/2/4 numerical comparison with the
  upstream kernel and `AscendBlockTables.compute_slot_mappings(..., out=...)`.

- Direct-kernel performance on Ascend 910. Each implementation and case was
  profiled independently with `msprof op`; values are `Task Duration(us)`,
  not host-side Event timing. The baseline is the original vLLM kernel from
  `6e448d0ea9bf3d88d898b65449ca6dc2aec170ac`.

| Case | Request x tokens | Block-table width | Grid | Upstream (us) | Ascend (us) | Speedup |
| --- | --- | ---: | --- | ---: | ---: | ---: |
| `tile-1` | 1 x 1024 | 8 | 1 x 2 | 114.48 | 2.86 | 40.03x |
| `tile-64` | 64 x 1024 | 512 | 1 x 65 | 230.46 | 8.58 | 26.86x |
| `tile-1024` | 1024 x 1024 | 8192 | 1 x 1025 | 2944.54 | 78.62 | 37.45x |
| `long-64x8192` | 64 x 8192 | 4096 | 1 x 65 | 1574.28 | 13.10 | 120.17x |
| `long-64x32768` | 64 x 32768 | 16384 | 1 x 65 | 6183.24 | 36.12 | 171.19x |
| `cp2-tile-64-rank0` | 64 x 1024, CP=2 | 512 | 1 x 65 | 497.50 | 43.60 | 11.41x |
| `cp4-tile-64-rank2` | 64 x 1024, CP=4 | 512 | 1 x 65 | 492.56 | 43.74 | 11.26x |

Full CI was not run.

### DIFF
--- a/vllm_ascend/worker/v2/block_table.py
+++ b/vllm_ascend/worker/v2/block_table.py
@@ -17,7 +17,13 @@
 # This file is a part of the vllm-ascend project.
 #
 import torch
+from vllm.triton_utils import triton
+from vllm.v1.attention.backends.utils import PAD_SLOT_ID
 from vllm.v1.worker.gpu.block_table import BlockTables
+
+from vllm_ascend.ops.triton.v2.block_table.compute_slot_mappings import (
+    _compute_slot_mappings_kernel,
+)
 
 
 class AscendBlockTables(BlockTables):
@@ -48,6 +54,13 @@ class AscendBlockTables(BlockTables):
             cp_rank,
             cp_interleave,
         )
+        # The kernel block-table row can be wider than
+        # max_num_blocks_per_group when one KV block maps to multiple kernel
+        # blocks. Use the allocated row stride so the staged row is complete.
+        max_block_table_stride = max(block_table.gpu.stride(0) for block_table in self.block_tables)
+        # tl.arange needs a compile-time power-of-two size. This value is
+        # passed as a constexpr and covers every KV cache group's row.
+        self._block_table_pad_size = triton.next_power_of_2(max_block_table_stride)
         # because we will override these attribute, delete these attribute to
         # make sure it's collected by python gc immediately.
         del self.slot_mappings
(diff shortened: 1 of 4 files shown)


vllm_ascend/worker/v2/block_table.py
```diff
@@ -17,7 +17,76 @@
...
+        return
+
+    block_table_ptr = _load_ptr(block_table_ptrs + group_id, tl.int32)
+    block_table_stride = tl.load(block_table_strides + group_id)
+    block_size = tl.load(block_sizes + group_id)
+
+    req_state_idx = tl.load(idx_mapping + batch_idx)
+    start_idx = tl.load(query_start_loc + batch_idx)
+    end_idx = tl.load(query_start_loc + batch_idx + 1)
+    for i in range(start_idx, end_idx, TRITON_BLOCK_SIZE):
```
AuroraEmiya: move offset out of loop would save temporary memory

vllm_ascend/worker/v2/block_table.py
```diff
@@ -17,7 +17,76 @@
...
+        offset = i + tl.arange(0, TRITON_BLOCK_SIZE)
+        positions = tl.load(pos + offset, mask=offset < end_idx, other=0).to(tl.int32)
+
```
AuroraEmiya: It seems that only tail block needs mask

vllm_ascend/worker/v2/block_table.py
AuroraEmiya, line 1: move the file to the correct place.

Liamup777: done

chenlongxiao (approved):

Liamup777:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

AuroraEmiya (approved):
LGTM

weijinqian0 (approved):

Pz1116 (approved):
LGTM, let's submit an RFC as a substitude for the document and add the doc later when doc refactor is done

Liamup777:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

weijinqian0 (approved):

PR #13713, opened 2026-08-06
[Refactor][quantization]Remove deprecated W4A8Linear/W8A8PDMixMoE schemes and sync W4A8 MoE refactor

kunpengW-code:
### What this PR does / why we need it?

- Part of https://github.com/vllm-project/vllm-ascend/issues/13318

Some quantization modes are sunset in this PR, including

- W4A8 Linear
- W4A8 Per-Group MoE(2 weight formats)
- W8A8 PDMix MoE

The reasons are as follows:
-    It was a temporary solution and will not have any new evolution plan in the future.
-    The performance is poor, and there are better alternatives.
-    The application scope is small, and the maintenance difficulty is increased.

### Does this PR introduce _any_ user-facing change?

no

### How was this patch tested?

- vLLM version: v0.26.0
- vLLM main: https://github.com/vllm-project/vllm/commit/0351e9aa1fdf1a51329d1906881528dfe61fc88e

PR #15227, opened 2026-08-28
[Feature][Attention] Support Kimi K3 A3 MLA C8 on main

q664171689:

Ports the focused Kimi K3 no-RoPE MLA C8 capability from #13225 onto the latest `main` architecture and restores the checkpoint-specific W4A8 shared-expert dependency exposed by real-weight startup.

- Enables standalone A3 FA quant for selected no-RoPE MLA layers.
- Stores latent KV as INT8 PA-NZ while keeping positional KV in the model dtype.
- Dynamically quantizes latent Q for decode and supplies FIA descales.
- Restores referenced PA-NZ blocks to compact logical-ND views for chunked-prefill gather, without a device-to-host `.item()` synchronization or a cache-sized remap tensor.
- Accounts for mixed latent/positional dtypes in normal and hybrid KV-cache allocation and reshape.
- Detects ModelSlim FA quant before `quant_config` initialization and sizes the hybrid MLA/Mamba cache from the real C8 bytes, keeping Kimi K3's 768-token C8 blocks aligned with Mamba state pages.
- Preserves the existing FA-quant allocation path for models that do not use Kimi K3's A3 mixed INT8/BF16 cache.
- Pads per-head query descales together with non-power-of-two query heads.
- Adds a Kimi K3-only per-channel W4A8 linear scheme for shared experts and executes their packed INT4 weights through the single-expert grouped-matmul path.
- Keeps the deprecated generic `W4A8_DYNAMIC + linear` scheme disabled for all other models and projections.

The Kimi-specific W4A8 path is required because the real Kimi K3 ModelSlim checkpoint describes shared-expert `gate_up_proj` and `down_proj` as `W4A8_DYNAMIC` linear layers. Current `main` supports W4A8 routed MoE weights, but intentionally removed generic W4A8 linear support in #13713.

Related: #13225

Yes. Kimi K3 ModelSlim checkpoints can load their per-channel W4A8 shared experts and use the A3 INT8 latent-KV C8 path on the V1 model runner.

Added unit coverage for:

- standalone A3 no-RoPE C8 capability detection;
- PA-NZ latent/positional cache writes;
- compact block-table remapping and chunked-prefill latent-KV dequantization;
- A3 latent-query dynamic quantization;
- query-descale head padding;
- mixed INT8/BF16 cache byte accounting and reshape;
- pre-`quant_config` ModelSlim FA-quant detection;
- 768-token hybrid MLA/Mamba page alignment;
- nested `hf_text_config` FA-quant metadata;
- scoped ModelSlim selection of Kimi K3 W4A8 shared-expert linears;
- packed INT4 weight conversion and single-expert grouped-matmul execution;
- fused SiTU execution with already-dequantized W4A8 shared-expert projections;
- Kimi K3 E2E dummy metadata matching the real checkpoint's W4A8 shared-expert layout.

Local checks completed after rebasing:

- `python3 -m compileall` on the changed Python sources and tests;
- `git diff --check`.

Targeted pytest, ruff and NPU execution were not run locally because this host has no `torch`, `vllm`, `torch_npu`, Ascend device environment, or project lint environment. PR CI is used for formatting/static checks. A3 real-checkpoint startup, first-request output and accuracy validation remain required before merge.


q664171689:
### Kimi K3 real-checkpoint startup fix

The reported startup failure was reproduced from the stack trace as a missing `W4A8_DYNAMIC + linear` method for `block_sparse_moe.shared_experts`. The real checkpoint stores Kimi K3 shared experts as per-channel W4A8, while the existing main-branch dummy E2E described those projections as W8A8 and therefore did not cover the checkpoint layout.

Commit `fdcc8d4ea` adds a scoped fix:

- selects W4A8 linear only for `kimi_k3`/`kimi_linear` shared-expert projections;
- keeps deprecated generic W4A8 linear support disabled;
- loads the packed INT4 weights into a single-expert WeightNZ/GMM layout;
- uses grouped matmul around fused SiTU without double dequantization;
- updates UT and Kimi K3 E2E quantization metadata to cover the real layout.

CI status:

- DCO: passed
- pre-commit: passed
- mypy for Python 3.10/3.11/3.12: passed
- coverage recommendation and test selection: passed
- selected tests: waiting for a maintainer to add `ready-precise` (the contributor account cannot add repository labels)

An A3 real-checkpoint startup plus first inference request is still required before merge.

AI-assisted workflow summary: inspected the first fatal stack, compared main with #13225, isolated the removed linear scheme dependency, implemented a Kimi-only compatibility path, added regression coverage, rebased to current main, and iterated on repository CI until all contributor-runnable checks passed.

q664171689:
A3 real-weight follow-up (commit `fa3c9e253`):

The previous W4A8 shared-expert failure is resolved; loading then reached the next real-checkpoint-only failure:

`ValueError: There is no module or parameter named mm_projector.rot_proj`

Root cause: the main-port created `mm_projector.rot_proj` only when an external global QuaRot file was configured. The checkpoint-provided projector rotation is independent of that file, so the module was absent even though `mm_projector.rot_proj.weight` existed in the checkpoint.

This update restores the validated checkpoint-driven behavior from the 0.26 line:
- construct `rot_proj` before streaming weight loading;
- retain and execute it when its checkpoint weight is loaded;
- remove it after loading when the checkpoint does not provide that weight, preserving non-rotated checkpoint behavior;
- handle a wrapped tower placeholder when cleaning up the unused layer.

Regression tests cover unconditional construction and both checkpoint-present/checkpoint-absent paths. Local Ruff, compileall, and diff checks pass. Targeted pytest cannot run on this host because pytest/vLLM/NPU dependencies are unavailable. Please rerun A3 real-weight startup and the first text plus multimodal request on this head.

q664171689:
A3 DSpark real-startup follow-up (commit `037bf58f1`):

The latest failure is a load-order regression, not evidence that Kimi K3 DSpark lacks embedding support. The older successful path shares the target token embedding into a DSpark draft that has no private embedding. Main commit #14454 added a multimodal capability probe immediately after constructing the draft, before `_maybe_share_embeddings()`. For Kimi K3 DSpark that calls `embed_input_ids()` while `model.embed_tokens` is still `None`, causing the reported assertion.

This update moves the existing multimodal probe after embedding/top-k/LM-head sharing. No embedding is duplicated, multimodal support is not disabled, and the existing text-only fallback remains unchanged. A regression test asserts `share` occurs before `probe`. It also fixes the mypy narrowing error from the preceding `rot_proj` update. Local Ruff, compileall, and diff checks pass; A3 real-weight startup and first text/multimodal requests still require rerun.

q664171689:
Scope correction: the DSpark load-order change from commit `037bf58f1` has been withdrawn at the user request. Current head `a2ec7343a` restores `llm_base_proposer.py` and its tests to their pre-change state; PR #15227 is again limited to the Kimi K3 C8 path and real ModelSlim checkpoint-loading dependencies. The reported DSpark issue is deferred and should be avoided for now by starting without the DSpark `--speculative-config` / draft configuration.

PR #15229, opened 2026-08-28
[Performance][Attention] Avoid draft metadata device copies

Dawn952:

Parallel drafting keeps `seq_lens` on device for attention, but the attention metadata builder also converted that device tensor to a Python list. This introduced a synchronous D2H copy in every draft metadata build. The builder additionally recreated `query_start_loc` on device even though the common metadata already maintains a device copy.

This patch:

- derives Python sequence-length lists from the maintained CPU mirror;
- reuses the existing device `query_start_loc`;
- preserves device `seq_lens` for parallel drafting and cross attention;
- pads both CPU and device sequence lengths for graph/FIA padding;
- retains the original copy fallbacks for callers without CPU/device mirrors.

No. This is an internal performance optimization and keeps compatibility fallbacks for legacy metadata producers.

- `ruff format --check vllm_ascend/attention/attention_v1.py tests/ut/attention/a2/test_attention_v1.py`
- `ruff check vllm_ascend/attention/attention_v1.py tests/ut/attention/a2/test_attention_v1.py`
- `python -m py_compile vllm_ascend/attention/attention_v1.py tests/ut/attention/a2/test_attention_v1.py`
- `pytest -q tests/ut/attention/a2/test_attention_v1.py tests/ut/spec_decode/test_dspark_proposer.py` (59 passed)
- `pytest -q tests/ut/spec_decode/test_llm_base_proposer.py` (23 passed)

Ascend NPU validation on a single-node DP1/TP8/PP1/EP8 Kimi-K3 service with DCP=1, main-model `FULL_DECODE_ONLY`, and eager DSpark speculative decoding:

- `/v1/models` returned HTTP 200 and all 8 ranks became ready;
- 3/3 deterministic precision requests returned HTTP 200 with non-empty valid content;
- rank-0 2-second Decode profiling parsed successfully offline;
- within `AscendAttentionMetadataBuilder.build`, 852 CPU `tolist` calls had a maximum duration of 8.63 us, with 0 `aten::_to_copy` and 0 `aten::copy_` events;
- the previous device `seq_lens.tolist()` path was about 2.1 ms per call.

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -327,11 +327,14 @@ class AscendAttentionMetadataBuilder(AttentionMetadataBuilder[AscendMetadata]):
         # Prefer _seq_lens_cpu (always available, updated during draft
         # iterations) over seq_lens_cpu (None in async spec decode mode).
         if common_attn_metadata._seq_lens_cpu is not None:
-            seq_lens = common_attn_metadata._seq_lens_cpu[:num_reqs]
+            seq_lens_cpu = common_attn_metadata._seq_lens_cpu[:num_reqs]
         elif common_attn_metadata.seq_lens_cpu is not None:
-            seq_lens = common_attn_metadata.seq_lens_cpu[:num_reqs]
+            seq_lens_cpu = common_attn_metadata.seq_lens_cpu[:num_reqs]
         else:
-            seq_lens = common_attn_metadata.seq_lens[:num_reqs].to("cpu")
+            seq_lens_cpu = common_attn_metadata.seq_lens[:num_reqs].to("cpu")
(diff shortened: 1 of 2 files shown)


vllm_ascend/attention/attention_v1.py
```diff
@@ -376,7 +382,12 @@ def build(
         if len(seq_lens_list) < num_reqs_fia:
             padding_len = num_reqs_fia - len(seq_lens_list)
             seq_lens_list = seq_lens_list + [1] * padding_len
-            seq_lens = torch.cat([seq_lens, seq_lens.new_ones(padding_len)])
+            seq_lens_uses_cpu_mirror = seq_lens is seq_lens_cpu
```
Dawn952: 这是为了验证什么？

PR #15246, opened 2026-08-28
[BugFix][Attention] Fix causal C8 multi-query verification and FULL replay tables

EricLingRui:

Fix the C8 FIA V1 BNSD path when one request verifies multiple query tokens.
This path flattens queries into the batch dimension with `S=1`, so it needs
one block-table row and one visible KV length per query, not per request.

For a causal request with final KV length `L` and `Q` verification queries,
the visible lengths must be `L-Q+1, ..., L`. Repeating `L` for every query
would allow earlier queries to attend to unverified future tokens.

The patch:

- Expands block-table rows and computes per-query causal lengths in C8 decode,
  the decode portion of chunked prefill, FULL capture, and FULL replay.
- Preserves full KV visibility for non-causal metadata and handles the dummy
  padding request without generating negative KV lengths.
- Allocates layer-owned expanded table buffers before Model Runner V1 graph
  capture. Replay copies current tables into these buffers before updating FIA
  tasks. A temporary expansion must not become an unowned graph input, and
  capture must not record a dummy-table copy that overwrites replay metadata.
- Records C8 layer names for replay metadata lookup instead of relying on
  dictionary insertion order. Graph sizes share one maximum-sized allocation
  per layer, while different layers retain independent storage.
- Retains explicit ownership when single-query dummy requests are captured
  but replay verifies multiple queries for one request at the same total size.
- Keeps ordinary single-query decode-only graphs without speculation on the
  existing table storage; mixed FULL graphs reserve expanded storage.

Related work: #10862 addresses C8/EAGLE length-list shape mismatches, and
#10399 explores C8/MTP GQA support. This PR is a focused alternative for the
existing flattened `BNSD, S=1` path, with causal and persistent-storage
regressions. It does not supersede or incorporate those unmerged PRs.

It corrects C8 multi-query attention metadata and Model Runner V1 FULL replay.
There are no new flags or configuration defaults, and the BF16 attention path
is unchanged.

This is **not** a complete W8A8C8 + DFlash2 support or performance claim.
Mixed INT8/BF16 KV allocation, checkpoint loading, D256 operator support,
Mamba prefix-cache correctness, Model Runner V2, and end-to-end long-context
model validation are outside this patch. The tests below use the native D128
kernel and require no model checkpoint or private custom operator.

Current validation (2026-09-10): rebased onto vllm-ascend
`6c30de40cdd0194e2bcc8a125f8a87cd3323a485`, with the verified vLLM main
`b2f685834a6456197e7033966fdef52a23f1abcd` (release reference: v0.28.0).
Patch head: `57fe622370e6833fb24b4a6900875fd626da371b`.

The rebased patch preserves upstream layer-aware FIA replay and workspace
handling. The native NPU regression now lives in the one-card hardware suite,
because upstream routes `tests/ut` to CPU runners. Its estimated runtime is
registered in `test_config.yaml`; the explicit CI selector selects the
`linux-aarch64-a2b3-1` runner with one NPU.

Runtime: Ascend 910B3 (one device), CANN 9.1.0, PyTorch 2.10.0+cpu,
torch-npu 2.10.0.post4. Isolated images install the complete verified vLLM
source and complete Ascend checkout (`COMPILE_CUSTOM_KERNELS=0`).
No model checkpoint, runtime source overlay, or private custom operator was used.

```bash
python3 -m pytest -q \
  tests/ut/attention/test_attention_v1.py \
  tests/ut/worker/test_model_runner_v1.py \
  tests/e2e/pull_request/one_card/spec_decode/test_c8_spec_decode_npu.py
```

- Combined run in one pytest process: **97 passed**, including all five real-NPU
  cases, without failures or skips.
- Separate CPU-only container, with no NPU devices and
  `TORCH_DEVICE_BACKEND_AUTOLOAD=0`: both UT modules **92 passed**.
- Full manual pre-commit checks, diff whitespace validation, and secret scan
  of all four commits: passed.
- Negative control: unpatched upstream `6c30de40` plus only the native test file,
  using the same runtime and `-k 'not full_replay'`: **1 passed, 3 failed,
  1 deselected**. Single-query decode passes; eight-query, variable-query,
  and future-value-isolation assertions fail numerically, not at import or
  kernel launch. FULL is excluded because it uses the new preparation API.
- The unpatched future-value sentinel returns `0.2578125` for the first query;
  the patched result is exactly zero, as required by the causal reference.

The model-free tests use native D128 FIA and a CPU per-query causal reference.
The FULL regression performs actual capture/update/replay for two layers,
reverses metadata insertion order, changes KV lengths and non-contiguous page
tables, and switches graph sizes 8 -> 4 -> 8. Size-8 capture uses eight
single-query dummy requests; replay verifies eight queries for one request.
The mocked UT fixtures restore streams/events so they do not leak into the
native graph test in the combined run.

Historical validation on vllm-ascend `6953f266922e3ec2c33a04db4febb00e15893cec`
and vLLM `ba07e4a48fc951300d97eb506217dd530583dea3` was **81 passed** on the
former test paths, with the same negative-control outcome. This historical
count is separate from the current 97-test run above.

These are focused regression results, not full-model inference or throughput
claims. Upstream hardware CI still requires a maintainer's `ready-precise` or
`ready-all` label; local validation does not replace that CI gate.

AI assistance: OpenAI Codex assisted with implementation, tests, and PR preparation.

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -75,6 +75,54 @@ SWA_INT_MAX = 2147483647
 _ATTN_KEYS_BUFFER = None
 
 
+def _expand_c8_decode_metadata(
+    block_tables: torch.Tensor,
+    seq_lens: list[int],
+    actual_seq_qlen: list[int],
+    num_requests: int,
+    *,
+    causal: bool = True,
+    num_actual_tokens: int | None = None,
+    expand_tables: bool = True,
+) -> tuple[torch.Tensor, list[int]]:
+    """Map flattened BNSD queries to paged rows and their visible KV lengths.
+
+    The C8 path uses S=1, including when a request verifies multiple tokens.
+    Repeating its final KV length would let earlier queries see future tokens.
+    Lengths are already host metadata; do not read device tensors here.
+    """
+    if not 0 < num_requests <= min(len(seq_lens), len(actual_seq_qlen), block_tables.shape[0]):
+        raise ValueError("C8 decode requires matching request metadata and block-table rows")
+    query_lens = []
+    expanded_seq_lens = []
+    previous_q_end = 0
+    for request_idx in range(num_requests):
+        q_end = int(actual_seq_qlen[request_idx])
+        query_len = q_end - previous_q_end
+        kv_len = int(seq_lens[request_idx])
+        is_padding = num_actual_tokens is not None and previous_q_end >= num_actual_tokens
+        if query_len <= 0 or (not is_padding and (kv_len <= 0 or (causal and kv_len < query_len))):
+            raise ValueError(f"Invalid C8 decode lengths: query={query_len}, kv={kv_len}")
+        query_lens.append(query_len)
+        if is_padding:
+            # The builder uses a null block row for the dummy padding request.
+            # Its output is discarded; avoid negative lengths for Q > KV = 1.
+            expanded_seq_lens.extend([1] * query_len)
+        elif causal:
+            expanded_seq_lens.extend(range(kv_len - query_len + 1, kv_len + 1))
+        else:
+            expanded_seq_lens.extend([kv_len] * query_len)
+        previous_q_end = q_end
+    tables = block_tables[:num_requests]
+    if not expand_tables or all(length == 1 for length in query_lens):
+        return tables, expanded_seq_lens
+    if len(set(query_lens)) == 1:
+        tables = tables.repeat_interleave(query_lens[0], dim=0)
+    else:
+        tables = torch.cat([tables[i : i + 1].expand(length, -1) for i, length in enumerate(query_lens)], dim=0)
+    return tables, expanded_seq_lens
+
+
 @register_backend(AttentionBackendEnum.CUSTOM, "ASCEND")
 class AscendAttentionBackend(AttentionBackend):
     accept_output_buffer: bool = True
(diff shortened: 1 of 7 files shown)


vllm_ascend/attention/attention_v1.py
```diff
@@ -74,6 +74,54 @@
...
+        elif causal:
+            expanded_seq_lens.extend(range(kv_len - query_len + 1, kv_len + 1))
+        else:
+            expanded_seq_lens.extend([kv_len] * query_len)
+        previous_q_end = q_end
+    tables = block_tables[:num_requests]
+    if not expand_tables or all(length == 1 for length in query_lens):
+        return tables, expanded_seq_lens
+    if len(set(query_lens)) == 1:
+        tables = tables.repeat_interleave(query_lens[0], dim=0)
```
nu11ptr7, line 119: `repeat_interleave` / `expand` create shared-memory views of the block tables. If anything downstream writes into the expanded table in place, it would corrupt multiple rows at once - worth confirming the expanded tables are read-only.

EricLingRui: Thanks for pointing this out — this is an important lifetime/aliasing concern. I double-checked both branches: `repeat_interleave` materializes a new tensor, while the variable-length branch uses `expand` only for the temporary inputs to `torch.cat`, whose result is also materialized. The resulting table is passed to FIA as a read-only input; there are no in-place writes to it downstream. For FULL replay, updates use `copy_` only on the separately preallocated, layer-owned persistent buffer, not on the expanded source table. So shared-row corruption should not occur here. I would be happy to add a defensive comment or an explicit non-aliasing assertion in the regression test if you think that would make this clearer.

PR #15252, opened 2026-08-28
[Performance][MoE] Overlap shared-expert collectives and routed projections

Dawn952:
## What this PR does

For the SP-only shared-expert path with MoE multistream enabled, this PR adds/restores three overlaps:

1. shared-input TP `all_gather` with routed latent-down
2. shared down projection with routed combine
3. shared-output `reduce_scatter` with routed latent-up

The Ascend Kimi-K3 adapter reaches this path through `AscendMoERunner.apply_routed_input_transform`; other parallel modes and multistream-off behavior remain unchanged.

## Ordering and correctness

```text
input:  main record(input_ready) -> latent-down -> wait(all_gather_done)
        aux  wait(input_ready)   -> all_gather -> record(all_gather_done)

down:   main record(before_combine) -> routed combine
        aux  wait(before_combine)   -> shared down projection

output: main routed finalize -> record(after_finalize) -> latent-up -> wait(aux) -> add
        aux  wait(after_finalize) -> reduce_scatter
```

The waits keep shared TP collectives outside the routed collective interval. The change also:

- keeps the gathered tensor as an explicit custom-op input and gives graph tracing the real local-token output shape;
- retains TP padding only across the custom-op boundary, while the shared MLP trims padding before computation;
- preserves the fused RMSNorm FP32 router-input path introduced by #15426;
- joins the shared output before the upstream FP16 in-place scaling path;
- restores the upstream custom-op entry when 310P disables multistream;
- preserves A3/A5 W8A8 and W4A8MXFP operator argument contracts.

### Asynchronous shared-expert stages

`AscendSharedExperts` exposes the collective lifecycle without deciding which model should use the overlap:

- `start_input_all_gather` records input readiness on the main stream, starts the SP gather on the shared-expert stream, and returns its completion event;
- the shared down projection waits on `before_combine`, allowing it to run alongside routed combine rather than waiting for routed finalization;
- `defer_output_wait` and `wait_for_output` separate launching the shared-output reduce-scatter from joining it, while non-latent paths keep the original synchronization point.

### MoE runner orchestration

`AscendMoERunner` composes those stages into the three overlaps shown above. The optimized path is enabled only when a routed transform and shared experts are present, multistream is enabled, and the shared-expert mode is SP-only. The runner also:

- selects a dedicated `ascend_moe_forward_shared_sp` custom-op entry for this path;
- starts the gather before routed latent-down and delays the shared-output join until after routed latent-up;
- trims TP padding only for shared-MLP computation;
- supplies a fake implementation whose output token dimension remains SP-local even though the explicit shared input is gathered;
- preserves #15426 by reusing a supplied FP32 router input and retaining `shared_hidden_states.float()` only as its existing fallback.

## Commit structure

The branch is linear on the current `main` snapshot and contains four signed-off commits grouped by responsibility:

1. `[Performance][MoE] Add asynchronous shared-expert collective stages`
   - exposes the early SP input gather and deferred shared-output synchronization primitives;
2. `[Performance][MoE] Overlap shared collectives with routed projections`
   - schedules the three overlaps, keeps them capability-gated, and preserves the fused FP32 router input;
3. `[Compatibility][310P] Preserve the fused MoE fallback`
   - isolates the overlap-specific path from the 310P single-stream fallback;
4. `[Test][MoE] Cover shared-expert overlap paths`
   - consolidates MoE, Kimi-K3, graph-shape, quantized-contract, event-ordering, and 310P coverage.

## Evidence and tests

The supplied rank-7 Kimi-K3 profile contains 2,116 serial occurrences of each target pattern. Mean kernel times are 18.086 us for latent-down, 11.982 us for the following all-gather, 9.145 us for shared down, 17.979 us for reduce-scatter, and 13.905 us for the following latent-up. These identify overlap opportunities; they are not end-to-end performance claims.

Host-only tests cover the Ascend Kimi call contract, SP-only on/off gating, SP+shared-DP and non-SP exclusion, padding and shared-MLP trimming, graph fake shapes, A3/A5 quantized paths, all three event boundaries, and the 310P fallback.

Local Ruff lint, Ruff format, and changed-file compilation pass, and every commit carries a matching `Signed-off-by`. The Windows host cannot collect these tests because it lacks the Linux vLLM/torch-NPU environment. NPU correctness, whole-network validation, and post-fix profiling remain pending.

### DIFF
--- a/vllm_ascend/ops/fused_moe/fused_moe.py
+++ b/vllm_ascend/ops/fused_moe/fused_moe.py
@@ -24,6 +24,8 @@ from vllm.distributed import (
 )
 from vllm.model_executor.layers.fused_moe import FusedMoEConfig, FusedMoERouter
 from vllm.model_executor.layers.fused_moe.layer import MoERunner
+from vllm.model_executor.layers.fused_moe.runner.moe_runner import _moe_forward_shared
+from vllm.utils.torch_utils import direct_register_custom_op
 
 from vllm_ascend.ascend_forward_context import _EXTRA_CTX, MoECommType
 from vllm_ascend.distributed.parallel_state import get_mc2_group
@@ -32,6 +34,45 @@ from vllm_ascend.ops.fused_moe.routed_experts import AscendRoutedExperts
 from vllm_ascend.ops.fused_moe.shared_experts import (
     AscendSharedExperts,
     SharedExpertParallelMode,
+)
+
+
+def _ascend_moe_forward_shared_sp_fake(
+    hidden_states: torch.Tensor,
+    router_logits: torch.Tensor,
+    shared_experts_input: torch.Tensor | None,
+    input_ids: torch.Tensor | None,
+    layer_name: object,
+    hidden_dim_unpadded: int,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Infer the local SP outputs of the shared-expert custom op.
+
+    The explicit shared input contains all gathered tokens, while the real
+    shared output has already been reduce-scattered back to the routed input's
+    local token count. The upstream fake uses the shared input's token count,
+    which is only correct before the early all-gather optimization.
+    """
+    del router_logits, input_ids, layer_name
+    assert shared_experts_input is not None
+    shared_out = shared_experts_input.new_empty(
+        (*hidden_states.shape[:-1], shared_experts_input.shape[-1]),
+    )
+    if hidden_dim_unpadded > 0:
+        fused_out = hidden_states.new_empty(
+            (*hidden_states.shape[:-1], hidden_dim_unpadded),
+        )
+    else:
+        fused_out = torch.empty_like(hidden_states)
+    return shared_out, fused_out
+
+
+# Keep the gathered shared input as an explicit graph dependency, but expose
+# the SP-local shared output shape to torch.compile/ACL graph tracing.
+direct_register_custom_op(
+    op_name="ascend_moe_forward_shared_sp",
+    op_func=_moe_forward_shared,
+    fake_impl=_ascend_moe_forward_shared_sp_fake,
+    tags=(torch.Tag.needs_fixed_stride_order,),
 )
 
 
@@ -84,6 +125,8 @@ class AscendMoERunner(MoERunner):  # type: ignore[no-redef]
                 self.quant_type,
                 self._quant_method,
             )
+            if self._can_overlap_sp_shared_with(self.routed_input_transform):
+                self._forward_entry = torch.ops.vllm.ascend_moe_forward_shared_sp
 
         setup_moe_comm_method(self.moe_config)
         alltoall_comm = get_moe_comm_method(MoECommType.ALLTOALL)
@@ -108,6 +151,21 @@ class AscendMoERunner(MoERunner):  # type: ignore[no-redef]
         """Ascend uses its own forward_impl path, not the FlashInfer Cutlass
         chunked path. Always return False to stay on forward_impl."""
         return False
+
+    def _can_overlap_sp_shared_with(self, routed_transform: object | None) -> bool:
+        """Limit SP shared-expert overlap changes to routed transforms.
+
+        Kimi-K3 is currently the only Ascend model that supplies these latent
+        transforms. Keep the guard capability-based so every model without
+        them stays on the original single-stream synchronization path.
+        """
+        shared_experts = getattr(self, "ascend_shared_experts", None)
+        return (
+            routed_transform is not None
+            and shared_experts is not None
+            and shared_experts.multistream_overlap
+            and shared_experts.parallel_mode() is SharedExpertParallelMode.SEQUENCE_PARALLEL_ONLY
+        )
 
     @property
     def _fused_output_is_reduced(self) -> bool:
@@ -208,6 +266,32 @@ class AscendMoERunner(MoERunner):  # type: ignore[no-redef]
         if self.ascend_shared_experts is not None:
             self.ascend_shared_experts.set_lora_context(lora_context)
 
+    def apply_routed_input_transform(
+        self,
+        hidden_states: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor | None]:
+        """Start the SP shared-input gather before the latent down projection."""
+        gathered_input = hidden_states
+        all_gather_done = None
+        shared_experts = self.ascend_shared_experts
+        if self._can_overlap_sp_shared_with(self.routed_input_transform):
+            assert shared_experts is not None
+            gathered_input, all_gather_done = shared_experts.start_input_all_gather(hidden_states)
+        routed_input, shared_input = super().apply_routed_input_transform(hidden_states)
+        if all_gather_done is not None:
+            torch.npu.current_stream().wait_event(all_gather_done)
+            shared_input = gathered_input
+        return routed_input, shared_input
+
+    def apply_routed_output_transform(self, fused_output: torch.Tensor) -> torch.Tensor:
+        """Run the latent up projection before joining the SP shared output."""
+        fused_output = super().apply_routed_output_transform(fused_output)
+        shared_experts = self.ascend_shared_experts
+        if self._can_overlap_sp_shared_with(self.routed_output_transform):
+            assert shared_experts is not None
+            shared_experts.wait_for_output()
+        return fused_output
+
     def _forward_impl(
         self,
         hidden_states: torch.Tensor,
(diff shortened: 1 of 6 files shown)


vllm_ascend/ops/fused_moe/shared_experts.py
```diff
@@ -266,11 +266,10 @@ def forward(
...
         local_dp_metadata = None
-        down_projection_ready = (
-            fused_moe_evts.after_routed_finalize
-            if self.multistream_overlap and mode is SharedExpertParallelMode.SEQUENCE_PARALLEL_ONLY
-            else fused_moe_evts.before_combine
-        )
+        # The down projection only depends on the shared-expert activation and
+        # should overlap the routed-expert combine.  The later shared-output
+        # collective is the operation that must wait for routed finalization.
+        down_projection_ready = fused_moe_evts.before_combine
```
nu11ptr7, line 278: `down_projection_ready` now always waits on `before_combine` - for SEQUENCE_PARALLEL_ONLY with multistream overlap it no longer waits for `after_routed_finalize`. If the shared down projection and the routed combine touch the same buffer, the earlier wait point could race - worth confirming there's no aliasing.

Dawn952: Thanks for raising this. I confirmed that the two paths do not alias the same tensor buffer:

- routed combine consumes the routed-expert `mlp_output` produced by `_apply_mlp` and returns a separate `combined_output`;
- the shared down projection consumes its own activation/quantized intermediate and returns a separate `shared_out`.

The results remain separate when `AscendMoERunner` returns `(shared_out, routed_out)`. The shared-output reduce-scatter still waits for `after_routed_finalize`, so the collective ordering is unchanged. This also restores the down-projection/combine overlap that existed before `2d7a2e3f6`; that commit intended to order the SP shared-output collective, but extended the wait to the down projection as well.

wxsIcey:
cc @linfeng-yuan @weijinqian0 PTAL.

lcfenglinwan:
Check whether it has an impact on other models.

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -213,6 +266,51 @@ def set_lora_context(self, lora_context):
...
+    def apply_routed_output_transform(self, fused_output: torch.Tensor) -> torch.Tensor:
+        """Run the latent up projection before joining the SP shared output."""
+        fused_output = super().apply_routed_output_transform(fused_output)
+        shared_experts = self.ascend_shared_experts
+        if self._can_overlap_sp_shared_with(self.routed_output_transform):
+            assert shared_experts is not None
+            shared_experts.wait_for_output()
+        return fused_output
+
+    def _maybe_apply_routed_scale_to_output(
```
Angazenn: please check whether we need to rewrite this function? are there any cases that `fused_output.dtype == torch.float16` is True?

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -114,6 +157,16 @@ def use_dp_chunking(self) -> bool:
         chunked path. Always return False to stay on forward_impl."""
         return False
 
+    def _can_overlap_sp_shared_with(self, routed_transform: object | None) -> bool:
+        """Limit SP shared-expert overlap changes to routed transforms."""
```
Angazenn: left some comments here to remind when `routed_transform is not None`. It seems for Kimi-K3 only now?

vllm_ascend/ops/fused_moe/shared_experts.py
```diff
@@ -243,34 +243,56 @@ def _pad_and_reduce_scatter(self, shared_out: torch.Tensor) -> torch.Tensor:
...
+            and hidden_states.shape[0] > local_num_tokens
+        ):
+            return hidden_states
+        shard_start = self.moe_config.tp_group.rank_in_group * local_num_tokens
+        return hidden_states.narrow(0, shard_start, local_num_tokens)
+
+    def wait_for_output(self) -> None:
+        """Join a deferred SP shared-output collective on the current stream."""
+        if self.multistream_overlap and self.parallel_mode() is SharedExpertParallelMode.SEQUENCE_PARALLEL_ONLY:
+            torch.npu.current_stream().wait_stream(shared_experts_calculation_stream())
```
Angazenn: I think this condition is duplicated in many places. My suggestion is to keep this function simply:
```suggestion
    def wait_for_output(self) -> None:
        """Join a deferred SP shared-output collective on the current stream."""
        torch.npu.current_stream().wait_stream(shared_experts_calculation_stream())
```
And make sure the condition is sufficient when calling `self.wait_for_output()`.

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -230,17 +328,26 @@ def _forward_impl(
...
+                # The runner input transform provides a padded gathered tensor
+                # in SP multistream mode. Trim only the shared-MLP view; retain
+                # the padded tensor for local internal-router reconstruction.
+                shared_input_is_gathered = self._can_overlap_sp_shared_with(self.routed_input_transform)
+                defer_shared_output_wait = self._can_overlap_sp_shared_with(self.routed_output_transform)
+                shared_expert_input = (
+                    shared_hidden_states[: _EXTRA_CTX.num_tokens] if shared_input_is_gathered else shared_hidden_states
                 )
                 if self.is_internal_router:
+                    router_hidden_states = self.ascend_shared_experts.local_input_from_gathered(
```
Angazenn: would the `narrow` here introduce operators like `split`? Or it is just `view`?

vllm_ascend/ops/fused_moe/fused_moe.py
```diff
@@ -213,6 +266,51 @@ def set_lora_context(self, lora_context):
...
+        """Start the SP shared-input gather before the latent down projection."""
+        gathered_input = hidden_states
+        all_gather_done = None
+        shared_experts = self.ascend_shared_experts
+        if self._can_overlap_sp_shared_with(self.routed_input_transform):
+            assert shared_experts is not None
+            gathered_input, all_gather_done = shared_experts.start_input_all_gather(hidden_states)
+        routed_input, shared_input = super().apply_routed_input_transform(hidden_states)
+        if all_gather_done is not None:
+            torch.npu.current_stream().wait_event(all_gather_done)
```
Angazenn, line 287: It seems wait for `shared_input_all_gather_done` is moved before router gate and this could potentially lose overlap between allgather with router gate. Please check the influence on performance.

vllm_ascend/ops/fused_moe/fused_moe.py
Dawn952: After rechecking only the execution paths currently reachable on Ascend NPU, I have revised my earlier conclusion. Ascend does have FP16 MoE workloads, including some 300I and Mixtral paths, but those FP16 models do not currently provide routed latent transforms. Kimi-K3 is the only current model that provides routed down/up transforms, and its supported configurations use BF16. Therefore, FP16 and routed-transform multistream overlap cannot occur together in the current NPU implementation, so this extra guard has no reachable call path. Commit 73f09aca6 removes the override and the artificially constructed FP16 overlap test.

vllm_ascend/ops/fused_moe/fused_moe.py
Dawn952: Confirmed: Kimi-K3 is currently the only Ascend model that supplies these routed latent transforms. Commit 1a6dc8b65 documents that explicitly. The guard remains capability-based; when either transform is absent, other models retain the original synchronization path and do not start the extra SP overlap.

vllm_ascend/ops/fused_moe/shared_experts.py
Dawn952: Addressed in 1a6dc8b65. wait_for_output is now only the stream join. Its callers own the conditions: apply_routed_output_transform and the FP16 scaling guard both require _can_overlap_sp_shared_with(...), while AscendSharedExperts.forward only defers the join for multistream SP-only with a routed output transform.

vllm_ascend/ops/fused_moe/fused_moe.py
Dawn952: torch.Tensor.narrow returns a view sharing the original storage; it does not launch a split/copy operator. Commit 1a6dc8b65 clarifies the zero-copy contract in the docstring and adds a unit assertion that the recovered local shard shares storage with the gathered tensor.

Dawn952:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

Dawn952:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

Dawn952:
[Bot]: cherry-pick completed successfully. New PR created: https://github.com/vllm-project/vllm-ascend/pull/15874

linfeng-yuan (dismissed):

Dawn952:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

Dawn952:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

linfeng-yuan (approved):

vllm_ascend/ops/fused_moe/fused_moe.py
Dawn952: The wait is intentionally after super().apply_routed_input_transform(...), so all-gather still overlaps with Kimi-K3’s latent-down projection. Router-gate overlap is not targeted in this PR; the wait preserves the gathered shared-input dependency.

PR #15253, opened 2026-08-29

maoxx241:

Data-parallel ranks with no scheduled requests still execute a one-token dummy batch to participate in graph replay and collectives. For GDN/Mamba layers, treating that synchronization-only token as a real decode token can advance causal-conv and KDA recurrent state through block-table entries left by an earlier request.

This PR keeps the model/collective dummy shape unchanged, but builds zero-length GDN metadata for idle DP dummies. Both captured recurrent branches become no-ops while other layers and distributed communication continue to execute normally.

The change restores the behavior previously validated on the v0.26 Kimi K3 path. It adds no device-wide synchronization and does not change GDN kernels.

No API change. Multi-node GDN/Mamba inference no longer depends on whether another DP rank was idle between requests.

## How was this patch tested?

- `python -m compileall` for the changed runtime files
- `git diff --check`
- Existing unit expectations updated for both idle-dummy entry points; the existing GDN builder regression covers zero-length recurrent metadata
- Real-weight A/B on two A3 nodes, TP16 x DP2 / EP32, Kimi K3 GQA DSpark, async scheduling, prefix cache, FULL_DECODE_ONLY graph, 131072-input + 1024-output:
  - before: a cache-hit repeat diverged from token 0 and collapsed into a repeated `1.1.1...` output while speculative acceptance remained 80.55%
  - after: no corrupted output in five cold/warm repeats; the last four cold/warm pairs were token-for-token identical, with per-token acceptance between 95.81% and 97.49%
- Real-weight four-node smoke on TP16 x DP4 / EP64 with full QuaRot weights and the same 128K-to-1K case:
  - repetition 1: aggregate acceptance 85.40%, mean accepted length 6.98, per-DP acceptance 88.11% / 89.03% / 85.91% / 79.16%
  - repetition 2: aggregate acceptance 87.00%, mean accepted length 7.09, per-DP acceptance 87.40% / 84.08% / 87.09% / 89.56%
  - both repetitions reported a 94.63% prefix-cache hit rate and completed without the cross-DP rotating collapse seen before the fix
  - repetition 3 was not counted: the DP2 EngineCore exited concurrently with an external Docker exec ending with status 137. The host reported no OOM or device fault, but the initiating command was not captured, so this is recorded as an interrupted run rather than attributed to the patch. The shared nodes were reclaimed and no restart was performed.

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2139,8 +2139,14 @@ class NPUModelRunner(GPUModelRunner):
                         # 0, and has_unfinished_requests in the outer loop
                         # returns True. before returning early here we call
                         # dummy run to ensure coordinate_batch_across_dp
-                        # is called into to avoid out of sync issues.
-                        self._dummy_run(1, skip_gdn_state_update=True)
+                        # is called into to avoid out of sync issues. Match the
+                        # active decode descriptor so an idle rank does not
+                        # downgrade graph mode across the DP group.
+                        self._dummy_run(
+                            self.uniform_decode_query_len,
+                            uniform_decode=True,
+                            skip_gdn_state_update=True,
+                        )
                     if not has_kv_transfer_group():
                         # Return empty ModelRunnerOutput if no work to do.
                         return EMPTY_MODEL_RUNNER_OUTPUT
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -3247,17 +3247,22 @@ class NPUModelRunner(GPUModelRunner):
                 GDNAttentionMetadataBuilder,
             )
             if is_gdn_noop:
-                assert num_reqs_padded is not None
-                # An idle DP dummy has a real tensor shape for model and
-                # collective execution, but no GDN tokens. Keep both captured
-                # recurrent branches inert instead of advancing cached state.
+                # Dummy-run callers coalesce this; fall back to the unpadded
+                # batch size instead of asserting in the execute path.
+                gdn_num_reqs = (
+                    num_reqs if num_reqs_padded is None else num_reqs_padded
+                )
+                # Idle DP dummy: keep captured GDN tensor ranks for graph
+                # replay/collectives. num_actual_tokens=0 is the kernel no-op
+                # (ops slice mixed_qkv[:0]); query_start_loc is a zero-filled
+                # prefix sized to the graph, not a real token schedule.
                 common_attn_metadata = replace(
                     common_attn_metadata,
                     query_start_loc=self.gdn_query_start_loc.gpu[
-                        : num_reqs_padded + 1
+                        : gdn_num_reqs + 1
                     ],
                     query_start_loc_cpu=self.gdn_query_start_loc.cpu[
-                        : num_reqs_padded + 1
+                        : gdn_num_reqs + 1
                     ],
                     num_actual_tokens=0,
                     max_query_len=0,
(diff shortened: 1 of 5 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3045,12 +3046,39 @@ def _build_attn_group_metadata(
...
+                        : num_reqs_padded + 1
+                    ],
+                    num_actual_tokens=0,
+                    max_query_len=0,
+                    is_prefilling=(
+                        torch.zeros_like(common_attn_metadata.is_prefilling)
+                        if common_attn_metadata.is_prefilling is not None
+                        else None
+                    ),
+                )
```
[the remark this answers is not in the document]
maoxx241: Addressed in ef7c1c576ccbf90513aa75721513839d663108b0: switched to dataclasses.replace(common_attn_metadata, ...) so this works even when CommonAttentionMetadata has no instance replace() (older vLLM). Current vLLM also implements replace() via dataclasses.replace, so behavior is the same.

maoxx241:
@vllm-project/vllm-ascend-maintainers could someone please add the `ready-precise` label on this PR?

`ci-gate` failed solely because selected tests are required (`HAS_READY_LABEL` / `ready-precise` / `ready-all` are all false). `pre-commit` and `select-tests` already passed; `run-selected-tests` / a5 / upstream were skipped for that reason. I do not have permission to add labels (`AddLabelsToLabelable`).

This is a small GDN dummy-run fix: idle DP ranks skip recurrent state updates, plus a follow-up using `dataclasses.replace(...)` for dummy attention metadata so older vLLM pins without `CommonAttentionMetadata.replace()` still work. `ready-precise` is enough; `ready-all` is not needed.

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -3045,12 +3046,41 @@ def _build_attn_group_metadata(
         ) -> None:
             attn_group = self.attn_groups[kv_cache_gid][attn_gid]
             builder = attn_group.get_metadata_builder(ubid or 0)
+            is_gdn_noop = skip_gdn_state_update and isinstance(
+                builder,
+                GDNAttentionMetadataBuilder,
+            )
+            if is_gdn_noop:
+                assert num_reqs_padded is not None
```
nu11ptr7: An `assert` in the inference hot path - if any dummy-run path ever reaches here without `num_reqs_padded`, the engine crashes with AssertionError instead of a graceful fallback. Also `num_actual_tokens=0` with a non-empty `query_start_loc` relies on kernels treating 0 specially; worth a comment or a defensive check.

maoxx241: Addressed in 260cd9d44846ecbd0ee33080fc74f5e902b3914a: dropped the dummy-run assert and fall back to `num_reqs` if `num_reqs_padded` is missing so the execute path cannot AssertionError. Idle dummy still keeps captured GDN tensor ranks; `num_actual_tokens=0` is the kernel no-op (`mixed_qkv[:0]`), and `query_start_loc` is the zero-filled graph-sized prefix filled by `_dummy_run` (not a real token schedule).

maoxx241:
cpu-0 (`run-selected-tests` job 99073063226) failed on `tests/ut/test_ascend_config.py::TestTopLevelSwitchTypeValidation::test_fused_mc2_is_disabled_for_unsupported_megamoe_config` with `AssertionError: 1 != 0`.

This is a product/test mismatch after the MegaMoe rollback (#15267): `enable_fused_mc2=1` forces `_MEGA_MOE_SUPPORTED=False` (dispatch_ffn_combine), so `derive_and_validate` no longer zeros the switch for unsupported MegaMoe model configs.

Aligned the unit test in `31bb358b297d9bb2d2aed796c511198299137f59` to expect `enable_fused_mc2=1` on that path.

maoxx241:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

changanla (approved):

maoxx241:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #15267, opened 2026-08-29
[bugfix] roll back megamoe op

zzzzwwjj:

Currently, there are many problems with the megamoe op. We will first roll back the megamoe internally and keep `enable_fused_mc2=2` to enable the megamoe for testing capabilities.
These codes will be removed after megamoe is ready.

Now when enable_fused_mc2=1, we will use dispatch_ffn_combine instead of mega_moe to ensure that there are no issues with the functionality.

### DIFF
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -329,6 +329,17 @@ class AscendConfig:
         # TODO(zzzzwwjj): remove it after deprecating `enable_mc2_hierarchy_comm`.
         if self.enable_mc2_hierarchy_comm:
             self.mc2_comm_alg = "hierarchy"
+        # TODO(zzzzwwjj): Currently, there are many problems with the megamoe op.
+        # We will first roll back the megamoe internally and keep `enable_fused_mc2=2`
+        # to enable the megamoe for testing capabilities.
+        # These codes will be removed after megamoe is ready.
+        global _MEGA_MOE_SUPPORTED
+        if self.enable_fused_mc2 == 1:
+            # When enable_fused_mc2=1, roll back to dispatch_ffn_combine.
+            _MEGA_MOE_SUPPORTED = False
+        elif self.enable_fused_mc2 == 2:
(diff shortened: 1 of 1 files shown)


ningjingbengxiaohai (approved):

chenweiqiang11:
This change now causes the full CPU UT suite on current `main` to fail and is blocking unrelated PRs. In #14715 run 33242168568, job 99074322375 finished with 2847 passed and only this failure:

```text
tests/ut/test_ascend_config.py::TestTopLevelSwitchTypeValidation::test_fused_mc2_is_disabled_for_unsupported_megamoe_config
AssertionError: 1 != 0
```

Job: https://github.com/vllm-project/vllm-ascend/actions/runs/33242168568/job/99074322375

The new before-validator overwrites the test's `_MEGA_MOE_SUPPORTED=True` patch when the input is `enable_fused_mc2=1`. Under this PR's intended contract, mode 1 now selects `dispatch_ffn_combine`, so the old test no longer reaches the unsupported-MegaMoE downgrade. #14715 does not change this test or any `enable_fused_mc2` / `_MEGA_MOE_SUPPORTED` logic.

The minimal semantic test fix is to exercise the new MegaMoE test mode instead:

```python
@patch("vllm_ascend.ascend_config.importlib.util.find_spec", return_value=object())
@patch.object(AscendConfig, "_is_megamoe_supported_by_config", return_value=False)
@patch("vllm_ascend.platform.NPUPlatform.check_and_update_config")
def test_fused_mc2_is_disabled_for_unsupported_megamoe_config(
    self, mock_fix, mock_megamoe_supported, mock_find_spec
):
    vc = VllmConfig()
    vc.additional_config = {"enable_fused_mc2": 2}
    self.assertEqual(init_ascend_config(vc).enable_fused_mc2, 0)
```

This preserves coverage of the unsupported-MegaMoE downgrade while aligning the test with the new mode-1/mode-2 contract. Simply changing the expected mode-1 result to 1 would lose that coverage.

chenweiqiang11:
Tracked as #15308 so the main-branch CPU regression can be handled independently of #14715.

PR #15271, opened 2026-08-29
[Test] System-level HTTP tests for load_balance_proxy_server + decode error fix

### LINKED ISSUE #12166: [load_balance_proxy] Decode backend errors are swallowed into an empty HTTP 200 response

xlshaoscu:
## Description

This PR combines:
1. The decode error forwarding fix (issue #12166, previously in #12168) — decode 4xx/5xx/connection errors are returned with the real HTTP status code instead of an empty 200.
2. A new **system-level test file** (`tests/ut/kv_offload/test_load_balance_proxy_server.py`) that verifies the proxy's HTTP behavior.

## Why system-level tests (not unit tests)?

Unit tests that import/call internal proxy functions break when the proxy is refactored (function rename, class extraction, etc.). System-level tests only check **HTTP input → output** (status code + body), which is the proxy's stable external contract. They survive any internal refactoring.

## Test approach

- **Self-contained**: the test file includes a mock ASGI backend (prefill + decode) with a scriptable response queue, started as subprocesses via `__main__`.
- **Real proxy**: the actual `load_balance_proxy_server_example.py` runs as a subprocess, pointing at the mock backends.
- **Real client**: `httpx` sends real HTTP requests and asserts status + body.
- **No internal imports**: the test never imports or calls proxy internals.

## Coverage (11 scenarios)

| Scenario | Mock decode returns | Expected proxy output |
|---|---|---|
| Normal non-stream | 200 stream + content | 200 + content |
| Normal stream | 200 stream + SSE + [DONE] | 200 + SSE + [DONE] |
| Over-limit 4xx non-stream | 400 `{"error":{...}}` | 400 + error body |
| Over-limit 4xx stream | 400 `{"error":{...}}` | 400 + error body |
| Decode 5xx (retry exhausted) | 500 ×3 | 500 |
| Decode connection failure | dead port | 502 |
| /v1/completions (prompt) | 200 stream | 200 |
| /v1/models | — | 200 |
| /healthcheck | — | 200 |
| Recompute | stream(recomputed) → stream(normal) | 200 + content |
| Mid-stream error | stream partial + abort | 200 (proxy doesn't crash) |

## Verification

```
pytest tests/ut/kv_offload/test_load_balance_proxy_server.py -v --noconftest
# 11 passed in 7.68s
```

- ruff format/check: pass
- DCO: Signed-off-by

## Supersedes #12168

This PR supersedes #12168 (fix-only) by adding the system-level test suite on top of the same fix.

Fixes #12166.


tests/ut/proxy/__init__.py
Yuli-yx, line 1: Why do we need this file? Can we remove it?

tests/ut/proxy/test_load_balance_proxy_server.py
Yuli-yx, line 1: Maybe we need more cases, like more prefill and decode nodes to test load balance function. Anyway, the test file is a good start. Let's merge it ASAP.

Yuli-yx (approved):

tests/ut/proxy/__init__.py
xlshaoscu: Kept intentionally: 23 of the 27 subdirectories under `tests/ut/` have an (empty) `__init__.py` (e.g. `kv_offload/`, `ops/`, `attention/`), and `tests/__init__.py` / `tests/ut/__init__.py` exist too, so this follows the existing convention. The pytest config in `pyproject.toml` uses the default `prepend` import mode, so the package chain makes the module name `tests.ut.proxy.test_load_balance_proxy_server` instead of the bare basename, which prevents `import file mismatch` errors if a same-named test file ever appears elsewhere. Happy to drop it if you'd rather be the exception, but I'd prefer to keep it consistent with the rest of `tests/ut/`.

Yuli-yx: OK. lets keep it.

Yuli-yx (approved):

wangxiaoteng888 (approved):

xlshaoscu:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

lcfenglinwan (approved):

xlshaoscu:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

xlshaoscu:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

Tflowers-0129 (approved):

PR #15284, opened 2026-08-29
[BugFix][Misc] Adapt imports to upstream vLLM module moves and API changes

Liears:
### What this PR does / why we need it?

Upstream vLLM main moved/changed two interfaces that the Ascend plugin references:

- The PCP helper `_gather_prefelch_cache_inputs` moved from `vllm.model_executor.layers.attention.pcp` (vendored trees) to `vllm.v1.attention.ops.pcp`. The version-gated import in `attention_v1` / `mla_v1` / `context_parallel/sfa_cp` picks the vendored location on 0.27.1 and raises `ModuleNotFoundError` on open-source main. Import from either location (try/except).
- `LogitsProcessor._get_logits` gained a `skip_gather` argument that upstream now passes positionally; `AscendLogitsProcessor._get_logits/_get_logits_normal` reject it with `TypeError: takes 3 to 4 positional arguments but 5 were given`. Accept the argument and honor it by skipping the TP gather.

Required to run the GLM-5.3-Flash adaptation #15127 on open-source vLLM main.

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

Manual: GLM-5.3-Flash serve on 8x Atlas 910B3 with open-source vLLM main (glm-release branch of vllm#53906) — the two TypeErrors/ModuleNotFoundError above were the first two startup blockers; both are gone and serving proceeds to weight loading.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- vLLM main: https://github.com/vllm-project/vllm/commit/ba07e4a48fc951300d97eb506217dd530583dea3

PR #15285, opened 2026-08-29
[BugFix][Core] Tolerate upstream KV cache spec field differences

Liears:

Upstream vLLM main (required by the GLM-5.3-Flash adaptation #15127) differs from the vendored tree in the KV cache spec API; four Ascend code paths break:

1. `MLAAttentionSpec` now declares `storage_block_size` as a dataclass field, so the generated `__init__` assigns it — `AscendMLAAttentionSpec`'s read-only property aborts with `AttributeError: property 'storage_block_size' has no setter`. Add a setter (stores an explicit override, keeps the computed Ascend layout) and declare the `compress_ratio` field the computed layout reads.
2. Upstream specs do not carry the Ascend-only `indexes_kv_by_block_stride` flag; `replace(spec, ...)` in `model_runner_v1`/`attn_utils` raises `TypeError` — fall back to the page-padding-only replace.
3. Upstream renamed `KVCacheTensor.shared_by` to `layers`; expose `shared_by` as a property alias so Ascend allocation code keeps working.
4. `unify_kv_cache_spec_page_size` refuses to pad MLA-family layers (`NotImplementedError: page size is not divisible ...`); patch a copy that also pads kpool indexer/tail caches — Ascend binds KV as block-first views, so padded indexer pages are addressed by runtime block stride.

No.

Manual: GLM-5.3-Flash (hybrid KDA + sparse MLA + kpool indexer) serve on 8x Atlas 910B3, open-source vLLM main: the four errors above were sequential startup blockers; all pass and serving reaches KV cache allocation.


Liears:

PR #15286, opened 2026-08-29
[BugFix][Worker] Dispatch bind_kv_cache to each layer's own override

Liears:

The patched `bind_kv_cache` assigns the raw allocation directly to `forward_context[layer].kv_cache`. Upstream vLLM binds through each layer's `bind_kv_cache()` override so subclasses can unpack the raw buffer — Mamba-family layers split it into conv/ssm state views, MLA layers squeeze the head dim. Direct assignment skips that unpacking: on GLM-5.3-Flash (#15127) every KDA layer then fails at the first forward with

```
ValueError: too many values to unpack (expected 2)
```

(`Glm5NextLinearAttention.forward` unpacking `(conv_state, recurrent_state) = self.kv_cache` from a single 4-D tensor).

The base `AttentionLayerBase.bind_kv_cache` stores the view as-is, so plain attention layers keep the previous behavior.

No.

Manual: GLM-5.3-Flash serve on 8x Atlas 910B3 — before the fix every chat request died with the ValueError above; after it, inference returns coherent output (verified Chinese reasoning chains). `MambaBase.bind_kv_cache` is invoked (24 bindings across 8 workers x 3 KDA layers in the mock model).


Liears:

PR #15287, opened 2026-08-29
[BugFix][Worker] Support the upstream aliasing KV cache layout

Liears:

Upstream vLLM main lays out hybrid KV caches as ONE backing allocation that every `KVCacheTensor` views into (all tensors share the same size; groups overlay each other because a block ID is owned by one group at a time). The legacy Ascend path in `_allocate_kv_cache_tensors` allocates each tensor independently, which multiplies the backing size by the tensor count and OOMs — on GLM-5.3-Flash (#15127) a 26.76 GiB budget became one 26.76 GiB allocation per tensor (indexer.k_cache, indexer.tail_cache, MLA, mamba, ...).

Detect the aliasing layout in `initialize_kv_cache_tensors` (all `kv_cache_tensors` share one size) and build the views through upstream's `create_kv_cache_views` with `kernel_block_size=None`. Two Ascend specifics remain:

- MLA main layers keep a dense contiguous `(key_cache, value_cache)` tuple of `[num_blocks, block_size, H, D]` tensors: the scatter/FIA kernels reject the interleaved aliasing views (`key tensor must be contiguous`, `dim2 of keyCache should be same as numHead`), so they get their own allocation like the legacy allocator produced.
- NoPE MLA (`qk_rope_head_dim == 0`) and the kpool indexer K-only caches have no V part; keep a single full-size tensor (sparse-C8 pattern) instead of asserting `k_dim > 0 and v_dim > 0`.

Mamba/indexer/tail layers use the aliasing views and are unpacked by each layer's `bind_kv_cache` (see #15286).

No.

Manual: GLM-5.3-Flash W8A8 serve (TP=8, block-size 512) on 8x Atlas 910B3 with open-source vLLM main — fixed the OOM above; weights + KV caches allocate, inference verified.


Liears:

PR #15288, opened 2026-08-29
[BugFix][Quantization] Tolerate GLM-5.3-Flash draft models in ModelSlim config

Liears:

Three lookups in the ModelSlim quant config assume every layer appears in `quant_model_description.json` under the exact vLLM module name. The GLM-5.3-Flash (glm5_next, #15127) multimodal wrapper and its MTP draft predictor violate that:

1. The wrapper registers the inner text model with `hf_config=text_config`, so `get_quant_method` sees `model_type='glm5_next_text'`, not `'glm5_next'` — without the `*_text` key in `packed_modules_model_mapping` the fused-module shard lookup raises `KeyError: 'language_model.model.layers.0.mlp.gate_up_proj.weight'`. Mirrors the existing `qwen3_5`/`qwen3_5_moe_text` pattern.
2. The MTP `SharedHead` placeholder reuses the main lm_head and has no checkpoint entry; layers absent from the description now fall back to FLOAT (`get_linear_quant_type` uses `.get`, `is_layer_skipped_ascend` treats zero hits as skipped) instead of `KeyError` / `NotImplementedError: quant_type=FLOAT for layer_type=linear`.
3. The draft module queries `model.layers.N.*` while descriptions are keyed under `language_model.model.layers.N.*`: `quant_prefix_mapper` retries the wrapper alias (same fallback pattern as `step3p5_mtp`).

No — with a ModelSlim W8A8 GLM-5.3-Flash checkpoint the draft MTP path starts instead of erroring.

Unit tests added: `tests/ut/quantization/test_modelslim_glm5_next.py` (5 tests, all pass; run with `VLLM_VERSION` set to a non-0.27.1 value so the pcp import takes the open-source branch):

```
pytest -sv tests/ut/quantization/test_modelslim_glm5_next.py
```

Manual: W8A8 GLM-5.3-Flash serve with `--speculative-config '{"method":"mtp",...}'` on 8x Atlas 910B3 — the three errors above were the sequential MTP blockers; the draft model now loads and runs.


Liears:

PR #15289, opened 2026-08-29
[Performance] Fuse DSV4 multi-group slot mapping

pisceskkk:

This PR merges repeated DSV4 slot-mapping calculations for multiple attention KV-cache groups into one Triton launch.

- Build device-side pointer and metadata tables for eligible attention groups.
- Address each request row inside the fused kernel.
- Route balanced, uneven, short-prefill, long-prefill, and decode-like request shapes through fixed or adaptive fused kernels.
- Keep fused-kernel selection in `vllm_ascend/ops/triton/compute_slot_mapping.py`.

Fusion eligibility follows the slot-mapping semantics:

- At least two non-Mamba attention groups share the launch.
- DCP groups use rank-local virtual-block/interleave metadata and retain their existing computation.
- Speculative decoding inputs with group-specific compressed position/request-index lists retain independent computation.

All otherwise eligible request counts and token lengths use the fused route.

The independent single-group kernel-path optimization is in #15397.

### Performance

Measurements were collected on one Ascend A3 device at 1850 MHz with CANN 9.1.0. Each shape used `msprof op` kernel replay with 20 warmups and 20 recorded replays per state.

For the heterogeneous profile-derived DSV4 six-group case, fusing six launches reduces aggregate kernel time from **26.321 us to 4.570 us**, an **82.6%** improvement.

The following generalization matrix uses six identical attention KV-cache groups. The separate value is the aggregate device time of six independent kernels; the fused value is one kernel launch.

| Per-group request shape | Separate aggregate median (us) | Fused median (us) | Improvement |
| --- | ---: | ---: | ---: |
| 1 x 4096 | 16.920 | 4.690 | 72.3% |
| 1 x 128 | 18.960 | 3.690 | 80.5% |
| 2 x 2048 | 17.040 | 4.710 | 72.4% |
| 8 x 512 | 17.400 | 6.490 | 62.7% |
| 8 x 16 | 15.780 | 6.030 | 61.8% |
| 64 x 1 | 37.741 | 18.170 | 51.9% |

All rows use 20 recorded samples. Per-mode CV is 2.6%–6.1%. The results cover long and short prefill, multi-request batches, and decode-like shapes.

Representative `PipeUtilization` results show 8 x 512 is Scalar-dominated after fusion (74.2% Scalar ratio), while 64 x 1 has higher MTE2 activity (41.1%) and wait time (7.99 us per vector-core row). The fused kernel remains 51.9% faster for the latter case.

No API or configuration change. This is an internal Triton launch-consolidation optimization.

- Split-branch A3 validation: five representative shapes in eager and ACL Graph modes, 10/10 passed exactly.
- The split matrix covers a profile-derived six-group long request, 8-request balanced and uneven batches, 64-request decode, and short multi-request prefill.
- Earlier generalized matrix for the same fused implementation: 8/8 eager and 8/8 ACL Graph cases passed exactly.
- Ruff, Python syntax, and `git diff --check` passed.

Evidence scope: kernel-level correctness, ACL Graph capture/replay, latency, and pipeline attribution on Ascend A3.

Co-authored-by: frankie <wangyongsheng686@gmail.com>

### DIFF
--- a/vllm_ascend/ops/triton/compute_slot_mapping.py
+++ b/vllm_ascend/ops/triton/compute_slot_mapping.py
@@ -82,3 +82,291 @@ def _compute_slot_mapping_kernel(
         if TOTAL_CP_WORLD_SIZE != 1:
             slot_ids = tl.where(is_local, slot_ids, PAD_ID)
         tl.store(slot_mapping_ptr + offsets, slot_ids, mask=mask)
+
+
+@triton.jit
+def _compute_slot_mapping_request(
+    start_idx,
+    end_idx,
+    req_idx,
+    positions_ptr,
+    block_table_ptr,
+    block_table_stride,
+    block_size,
+    slot_mapping_ptr,
+    KV_CACHE_BLOCK_SIZE: tl.constexpr,
+    BLOCKS_PER_KV_BLOCK: tl.constexpr,
+    TOTAL_CP_WORLD_SIZE: tl.constexpr,
+    TOTAL_CP_RANK: tl.constexpr,
+    CP_KV_CACHE_INTERLEAVE_SIZE: tl.constexpr,
+    PAD_ID: tl.constexpr,
+    TILE_BLOCK_SIZE: tl.constexpr,
+    BLOCK_TABLE_WINDOW_SIZE: tl.constexpr,
+):
+    row_offset = req_idx * block_table_stride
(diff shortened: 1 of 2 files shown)


pisceskkk:
[Bot]: cherry-pick completed successfully. New PR created: https://github.com/vllm-project/vllm-ascend/pull/15733

ningjingbengxiaohai (approved):

Issue #14974, opened 2026-08-26
[Bug]:  Online torch profiler (`/start_profile` / `/stop_profile`) segfaults on Ascend NPU


Sunshine-llh:
## Environment

- vLLM: `0.23.0`
- vLLM-Ascend: `0.23.1.dev15+g90ce6b11d` (and stock v0.23.0)
- torch_npu: `2.10.0.post4`
- Hardware: Atlas 800I A2 (Ascend 910B), single NPU
- CANN: 9.1.0

## Summary

On Ascend NPU (vllm-ascend + torch_npu), starting the online torch profiler via the HTTP endpoints
`POST /start_profile` / `POST /stop_profile` triggers a hard `Segfault` in the EngineCore process.

The profiler is started/stopped from an **asyncio worker thread** (`asyncio.to_thread`), while the
Ascend NPU runtime requires the profiler's `registerClient` to run on the **same thread that
initialized the device context**. The thread mismatch corrupts the profiler state and crashes.

## Environment

- vLLM: `0.23.0`
- vLLM-Ascend: `0.23.1.dev15+g90ce6b11d` (and stock v0.23.0)
- torch_npu: `2.10.0.post4`
- Hardware: Atlas 800I A2 (Ascend 910B), single NPU
- CANN: 9.1.0

## Repro

```bash
# 1. Start the server with torch profiler enabled
vllm serve <model> \
  --profiler-config '{"profiler":"torch","torch_profiler_dir":"./profile_out"}' \
  --enable-prefix-caching

# 2. During traffic, trigger the online profiler
curl -X POST http://127.0.0.1:8000/start_profile
# wait a few seconds (a small script that sleeps 5-10s works)
curl -X POST http://127.0.0.1:8000/stop_profile
```

### Observed log

```
ERROR: External init callback must run in same thread as registerClient (402649344 != 1108545664)
INFO: Profiler started.
INFO: POST /start_profile HTTP/1.1" 200 OK
INFO: Stopping profiler...
WARNING: Incorrect schedule: Stop profiler while current state is RECORD which may result in incomplete parsed data.
!!!!!!! Segfault encountered !!!!!!!
Fatal Python error: PyGILState_Release: thread state 0xffff... must be current when releasing
```

The server crashes with a `Segfault` after `stop_profile`.

## Root Cause

In `vllm/v1/engine/async_llm.py`, the profiler lifecycle is dispatched to a worker thread:

```python
[...]

PR #15294, opened 2026-08-29
[BugFix] Keep the torch profiler lifecycle on the client-registering thread

zihanlin-ai:

Refs #14974. **Blocked on vllm-project/vllm#54322** — that PR adds the flag this one sets;
without it this is an inert class attribute. Draft until it lands.

`AsyncLLM.start_profile` / `stop_profile` dispatch the frontend profiler lifecycle with
`asyncio.to_thread`. torch_npu's Kineto client is bound to the thread that registered it, so
`POST /start_profile` is rejected:

```
ERROR: External init callback must run in same thread as registerClient (1088942336 != -1328559392)
```

and the frontend profiler is left in a broken state. #14974 and vllm-project/vllm#39603
both report the serving process crashing on `stop_profile` afterwards.

The upstream PR adds `Platform.profiler_needs_client_thread`, defaulting to `False` so CUDA
keeps running the lifecycle off the event loop — `profiler.stop()` exports the trace inline
and costs ~4.5 µs per recorded span, so inlining it everywhere would stall the API server.
This PR is the Ascend half: set the flag to `True` on `NPUPlatform`.

Yes, once the upstream change lands: `POST /start_profile` and `POST /stop_profile` work on
NPU instead of corrupting the profiler.

Atlas A3 (`Ascend910_9392`), CANN 9.0.1, torch_npu 2.10, `vllm serve Qwen3-0.6B
--profiler-config '{"profiler":"torch","torch_profiler_dir":...}'`, then `POST /start_profile`
→ traffic → `POST /stop_profile`, with both this change and vllm-project/vllm#54322 applied:

| | `same thread as registerClient` errors in the server log |
|---|---|
| before | 1 |
| after | **0** |

Traces are written in both cases.

`tests/ut/test_platform.py` extends the existing `test_class_variables` assertions, since
that test already pins the platform's class attributes one by one: 52 passed, 1 skipped.
Deleting the attribute makes it fail, so it is a real guard rather than a restatement.
(That file needs `TORCH_DEVICE_BACKEND_AUTOLOAD=0` to run in my container; it fails the same
way on unmodified `main` there, so that is an environment quirk and not from this change.)

Two things I would rather state than overclaim:

- I could **not** reproduce the hard segfault #14974 describes. On my stack the server stayed
  alive and both endpoints returned 200; the reporter was on Atlas A2 with CANN 9.1.0 and
  vLLM 0.23.0. What is demonstrably fixed here is the thread violation that the issue
  identifies as the root cause.
- A separate `Incorrect schedule: Stop profiler while current state is RECORD` warning comes
  from the **EngineCore** process. It appears identically before and after, so it is a
  different problem and out of scope.

vLLM version: v0.27.1
vLLM main: https://github.com/vllm-project/vllm/commit/ba07e4a48fc951300d97eb506217dd530583dea3

### DIFF
--- a/tests/ut/test_platform.py
+++ b/tests/ut/test_platform.py
@@ -281,6 +281,9 @@ class TestNPUPlatform(TestBase):
         ):
             _validate_eplb_config(vllm_config)
         self.assertEqual(NPUPlatform.dispatch_key, "PrivateUse1")
+        # torch_npu's Kineto client is bound to its registering thread, so the
+        # profiler lifecycle must not be dispatched to a worker thread.
+        self.assertTrue(NPUPlatform.profiler_needs_client_thread)
         self.assertEqual(NPUPlatform.supported_quantization, [ASCEND_QUANTIZATION_METHOD, COMPRESSED_TENSORS_METHOD])
 
     def test_get_recompute_scheduler_cls(self):
--- a/vllm_ascend/platform.py
+++ b/vllm_ascend/platform.py
@@ -89,6 +89,9 @@ class NPUPlatform(Platform):
         "RAY_EXPERIMENTAL_NOSET_ASCEND_RT_VISIBLE_DEVICES",
     ]
     dispatch_key: str = "PrivateUse1"
+    # torch_npu's Kineto client is bound to the thread that registered it, so
+    # the profiler lifecycle must not be dispatched to a worker thread.
+    profiler_needs_client_thread: bool = True
 
     supported_quantization: list[str] = [
         ASCEND_QUANTIZATION_METHOD,


zihanlin-ai:
Closing: not needed after all.

I assumed this was Ascend-specific and that vLLM would need a platform flag for it. It isn't. The same failure reproduces on plain CPU torch and on CUDA — starting the frontend profiler from a worker thread records nothing on any platform, and start/stop landing on different pool threads segfaults. The message comes from torch's bundled Kineto, not torch_npu.

So the fix belongs entirely upstream and is unconditional: vllm-project/vllm#54322. No change is required in vllm-ascend.

PR #15303, opened 2026-08-29
[Bugfix][Config][Test] Fix ineffective megamoe rollback from stale _MEGA_MOE_SUPPORTED import

coder-fny:

Follow-up to #15267 (`[bugfix] roll back megamoe op`), which rebinds the module global `_MEGA_MOE_SUPPORTED = False` in `AscendConfig._validate_user_input_ranges` so that `enable_fused_mc2=1` routes to `dispatch_ffn_combine` instead of `mega_moe`.

**The rollback was ineffective.** `_MEGA_MOE_SUPPORTED` is a module-level bool, and every consumer imports it directly at the top of its file:

- `vllm_ascend/ascend_forward_context.py`
- `vllm_ascend/ops/fused_moe/routed_experts.py`
- `vllm_ascend/ops/fused_moe/moe_comm_method.py`
- `vllm_ascend/quantization/methods/w4a8/w4a8.py`
- `vllm_ascend/quantization/methods/w8a8/w8a8_dynamic.py`

`from module import var` binds a snapshot of an immutable (bool) at import time. #15267's `global _MEGA_MOE_SUPPORTED; _MEGA_MOE_SUPPORTED = False` only rebinds `ascend_config`'s own globals; the consumers keep reading their stale snapshot. In an environment where `cann_ops_transformer` is installed (initial value `True`), every `if _MEGA_MOE_SUPPORTED:` in those consumers still sees `True`, so they keep routing to `mega_moe` — the rollback to `dispatch_ffn_combine` is completely bypassed. Only `ascend_config` itself (e.g. `derive_and_validate`) reads the rebound value, so the symptom looks correct inside that module while the real op-selection points never observe the rollback.

**Fix:** expose a dynamic accessor and switch the consumers to it:

```python
# vllm_ascend/ascend_config.py
def is_mega_moe_supported() -> bool:
    return _MEGA_MOE_SUPPORTED
```

All five consumers now `from vllm_ascend.ascend_config import is_mega_moe_supported` and call `is_mega_moe_supported()` at each op-selection point, so #15267's rollback is actually observed. `ascend_config.py` itself keeps reading `_MEGA_MOE_SUPPORTED` directly (same module → always current, no stale issue), including the `derive_and_validate` disable branch.

**Test:** `test_fused_mc2_is_disabled_for_unsupported_megamoe_config` was broken by #15267 — `init_ascend_config`'s validator overrides the test's `@patch(_MEGA_MOE_SUPPORTED, True)` (a `global x = ...` rebind wins over `mock.patch` for a module global), so `derive_and_validate`'s "disable to 0" branch short-circuits and `enable_fused_mc2` stays `1` (test asserted `0` → `AssertionError: 1 != 0`). It is renamed to `test_fused_mc2_rolls_back_even_when_config_supported` and now asserts that **even when `_is_megamoe_supported_by_config()` returns `True`** (model config supports megamoe), the rollback still short-circuits `is_mega_moe_supported()` to `False` and `enable_fused_mc2` stays `1` — proving the rollback is unconditional and the fused path routes to `dispatch_ffn_combine` regardless of config support.

Yes (behavior fix). With `cann_ops_transformer` installed and `enable_fused_mc2=1`, the fused MoE path previously still routed to `mega_moe` (stale `True`), so #15267's intended rollback did not take effect. After this PR the consumers observe the rollback and route to `dispatch_ffn_combine`, matching #15267's documented intent ("when enable_fused_mc2=1, we will use dispatch_ffn_combine instead of mega_moe").

- `python -m py_compile` on all 7 modified files → OK
- Logic trace: with `@patch(_MEGA_MOE_SUPPORTED, True)` + `enable_fused_mc2=1` + `_is_megamoe_supported_by_config()=True`, the validator sets the module global to `False` (overriding the patch); `is_mega_moe_supported()` returns `False` and `enable_fused_mc2` stays `1` — both assertions hold, demonstrating the rollback short-circuits even for a supported config.
- Full run pending the CI cpu test runner (requires torch_npu/Linux):
```
pytest tests/ut/test_ascend_config.py::TestTopLevelSwitchTypeValidation::test_fused_mc2_rolls_back_even_when_config_supported -v
```


realliujiaxu (dismissed):

coder-fny:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

realliujiaxu (approved):

PR #15314, opened 2026-08-29
[Ops][Feature][MRV2] Improve performance of dflash_prepare_inputs

AuroraEmiya:
## Summary

This PR optimizes the Ascend Triton implementation of DFlash `prepare_dflash_inputs` for the vLLM 0.27.1 path currently supported by vLLM-Ascend.

The original implementation effectively serialized most per-request work in a single Triton program. This PR replaces it with a VectorCore-aware 2-D launch and distributes Context, Query, Sample, and graph-padding work across multiple programs.

The functional behavior and vLLM 0.27.1 ABI remain unchanged.

## Changes

* Move the optimized Triton implementation to:
  `vllm_ascend/ops/triton/v2/spec_decode/prepare_dflash_inputs.py`.
* Use a 2-D launch grid:
  `(num_reqs, workers_per_req)`.
* Distribute Context work across request-local workers using quotient/remainder partitioning.
* Independently distribute Query and Sample work across the same workers.
* Distribute graph-safety padding across the complete launch grid instead of serializing it in one program.
* Select `workers_per_req` according to:

  * available Ascend VectorCore parallelism;
  * Context vector width;
  * Query vector width;
  * Sample vector width.
* Keep the Context vector block size dynamic and capped at 256.
* Use a minimum Context block size of 2 to avoid a Triton-Ascend lowering issue on 1-element vector pointer arithmetic.
* Keep the implementation generic across request counts, context lengths, block sizes, and speculative-token counts instead of specializing for the captured business shape.
* Preserve chunk-prefill handling through `next_prefill_tokens`.
* Preserve rejected-token handling used by the vLLM 0.27.1 DFlash path.
* Preserve graph-visible padding behavior for CUDA/NPU graph replay.

## Scope

This optimization targets the DFlash `prepare_dflash_inputs` ABI used by vLLM 0.27.1, which is the version currently adapted by vLLM-Ascend for this path.

Upstream vLLM later changed the shared DFlash-family input-preparation ABI in `vllm-project/vllm#52188` while adding DSpark + Decode Context Parallelism support.

That newer ABI and its corresponding functionality are not adapted by this PR. Support can be added separately if needed in vLLM-Ascend.

## Correctness

The Triton output is compared bit-exactly against an independent Python reference.

The UT covers:

* captured business shape: batch 8 / 1626 context tokens;
* captured business shape: batch 64 / 576 context tokens;
* minimum-context compilation boundary;
* ragged request lengths;
* non-default KV block size;
* mixed chunk-prefill and rejected-token requests;
* Query/Sample sizes crossing the 16-element vector boundary;
* query-position clamping at `max_model_len`;
* full-capacity execution with minimal graph padding.

All cases pass with `rtol=0` and `atol=0`.

Test command:

```bash
pytest -sv --noconftest \
  tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_prepare_dflash_inputs.py
```

## Performance

Representative DFlash business shape:

```text
num_reqs              = 8
total_context_tokens   = 1626
max_num_reqs           = 64
max_num_tokens         = 8192
block_size             = 128
num_query_per_req      = 9
num_speculative_steps  = 8
```

Latest msprof result:

```text
Op Name: _prepare_dflash_inputs_kernel
Op Type: vector
Task Duration(us): 19.719999
Block Dim: 40
Current Freq: 1650 MHz
Rated Freq: 1650 MHz
```

The optimized implementation uses all 40 launched VectorCore programs for this workload and removes the previous request-local serialization bottleneck.

## Notes

The implementation intentionally avoids hard-coding the business shape or a fixed device core count.

The business cases are used as validation points rather than defining the supported input range.


Liamup777:
<html><head></head><body><h3>Performance</h3><p>Tested <code inline="">_prepare_dflash_inputs_kernel_ascend</code> with the real workload in the end-to-end model.</p>

  | Count | Total Time (us) | Avg Time (us) | Min Time (us) | Max Time (us) | Ratio
-- | -- | -- | -- | -- | -- | --
Before | 15 | 17192.88 | 1146.192 | 1082.56 | 1204.48 | 1.638%
After | 15 | 462.06 | 30.804 | 19.52 | 45.16 | 0.045%

<p>The average kernel latency is reduced from <strong>1146.192 us to 30.804 us</strong>, achieving about <strong>37.2x speedup</strong>.</p><p>The kernel's end-to-end execution time ratio is reduced from <strong>1.638% to 0.045%</strong>.</p></body></html>

vllm_ascend/ops/triton/v2/spec_decode/prepare_dflash_inputs.py
```diff
@@ -0,0 +1,1097 @@
...
+    valid_ctx_end = ctx_end - num_rejected
+    num_valid_ctx = valid_ctx_end - ctx_start
+
+    num_sampled = tl.load(num_sampled_ptr + req_idx)
+    if num_sampled > 0:
+        bonus_token = tl.load(last_sampled_ptr + req_state_idx).to(tl.int32)
+    else:
+        bonus_token = tl.load(next_prefill_tokens_ptr + req_state_idx).to(tl.int32)
+
+    last_valid_pos = tl.load(target_positions_ptr + valid_ctx_end - 1)
```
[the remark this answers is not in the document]
AuroraEmiya: Good point. This load requires each request to retain at least one valid context token, i.e. `0 <= num_rejected < num_ctx`.

Instead of clamping inside the kernel, we plan to enforce this invariant at the upper layer where the request metadata is prepared, preferably while the values are still available on the host to avoid adding a device-host synchronization to this hot path. The UT also keeps this constraint explicit.

chenlongxiao (approved):

AuroraEmiya:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

AuroraEmiya:
/rerun

[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

wxsIcey (approved):

vllm_ascend/ops/triton/v2/spec_decode/prepare_dflash_inputs.py
```diff
@@ -0,0 +1,331 @@
...
+
+from vllm_ascend.ops.triton.triton_utils import get_vectorcore_num, init_device_properties_triton
+
+_MAX_CONTEXT_BLOCK_SIZE = 256
+_QUERY_BLOCK_SIZE = 16
+_SAMPLE_BLOCK_SIZE = 16
+
+
+@triton.jit
+def _prepare_dflash_inputs_kernel(
```
wxsIcey: Do not place this ops in the spec_decode directory; we can create a new triton directory.

vllm_ascend/ops/triton/v2/spec_decode/prepare_dflash_inputs.py
```diff
@@ -0,0 +1,331 @@
...
+    sample_pad_start = num_reqs * num_speculative_steps
+    sample_pad_count = (max_num_reqs - num_reqs) * num_speculative_steps
+    sp_base = sample_pad_count // total_workers
+    sp_extra = sample_pad_count % total_workers
+    sp_begin = sample_pad_start + global_worker_idx * sp_base + tl.minimum(global_worker_idx, sp_extra)
+    sp_count = sp_base + tl.where(global_worker_idx < sp_extra, 1, 0)
+    sp_end = sp_begin + sp_count
+
+    for i in range(sp_begin, sp_end, 64):
+        sp_off = i + tl.arange(0, 64)
```
wxsIcey: Please do not use hard-coded values.

AuroraEmiya: Addressed in 11b7a171f. Replaced the inline graph-padding tile widths with named constexpr parameters and documented that they are per-iteration widths, not buffer-size limits. Kept the existing values (16/64/256) and launch behavior unchanged. Also factored the seven duplicated quotient/remainder partition calculations into one Triton helper, preserving disjoint ownership and fused sample-buffer writes. Validation: all 15 PR accuracy cases passed on Ascend910_9362 with exact reference comparisons (rtol=atol=0), including request counts 39/40/41/48 and context lengths 1024/16384. Used pytest -sv --noconftest with a small loading plugin to exercise the exact updated source without changing the container installation. Ruff check/format and git diff --check passed. The full format.sh ci check could not run because pre-commit is unavailable.

vllm_ascend/ops/triton/v2/spec_decode/prepare_dflash_inputs.py
```diff
@@ -0,0 +1,336 @@
...
+# Each loop covers the full worker partition regardless of the chosen tile width.
+_REQUEST_PADDING_BLOCK_SIZE = 16
+_SAMPLE_PADDING_BLOCK_SIZE = 64
+_QUERY_PADDING_BLOCK_SIZE = 256
+
+
+@triton.jit
+def _partition_work(num_items, worker_idx, num_workers):
+    # Contiguous, disjoint partitions; the first remainder workers get one extra item.
+    base = num_items // num_workers
```
drslark, line 40: Why not using `(num_items + (num_workers - 1)) // num_workers`?

drslark:
I think it is also necessary to test the cases that `num_context = num_speculative_steps + 1`.

drslark:
Add a comment that multimodel is not supported.
Refer to https://github.com/vllm-project/vllm-ascend/pull/9340.

vllm_ascend/ops/triton/v2/spec_decode/prepare_dflash_inputs.py
AuroraEmiya: Ceil-div would cover the items, but the quotient/remainder split keeps worker spans within one item of each other: 10 items / 4 workers gives 3,3,2,2 instead of 3,3,3,1. Each worker still owns a disjoint contiguous range. I kept this helper for Context, Query, Sample, and graph padding to balance the last workers. The request-count and long-context boundary cases pass bit-exactly on Ascend 910B4 (17/17 in the full operator test file).

AuroraEmiya:
Addressed the two remaining review notes from @drslark:

- [Context boundary](https://github.com/vllm-project/vllm-ascend/pull/15314#issuecomment-5725359205): added an explicit `num_context = num_speculative_steps + 1` case with a rejected suffix. I also added a fully rejected context after another request, which caught the `valid_ctx_end - 1` cross-request read. The optimized kernel now restarts from that request's first position, following the current shared-path semantic.
- [“Multimodel” limitation](https://github.com/vllm-project/vllm-ascend/pull/15314#issuecomment-5725412768): #9340 describes multimodal/M-RoPE positions. The operator documentation now states that this optimized path does not support multimodal inputs, since its KV slot mapping uses scalar positions instead of separate linear token indices. The v0.29 wrapper also documents that DFlash+DCP slot remapping is not implemented.

The branch is rebased on current main and all 17 operator correctness cases passed on an idle Ascend 910B4, including both new boundary cases.

drslark (approved):
LGTM.

vllm_ascend/ops/triton/v2/spec_decode/prepare_dflash_inputs.py
drslark: Good.

PR #14632, opened 2026-08-20
[BugFix][Frontend] Align DeepSeek V4 frontend behavior

QwertyJack:
### What this PR does / why we need it?

Ports the remaining DeepSeek V4 frontend behavior from the v0.25 release branch to `main` as temporary compatibility patches:

- #14317: default the parser to thinking when neither thinking control is supplied;
- #14035: attach request-level tools to an existing system message instead of inserting a second system message;
- #14521: stream long schema-typed string tool arguments incrementally before the closing parameter tag.

The upstream status does not yet make these patches removable from the supported vLLM matrix:

- https://github.com/vllm-project/vllm/pull/51296 is merged and the rebased main2main #14131 already targets a commit containing it, but #14131 is still open and review-required; the current vLLM-Ascend pin `58d3918e3` predates the fix, so the compatibility wrapper remains necessary for the supported matrix and is behaviorally idempotent after the pin advances;
- https://github.com/vllm-project/vllm/pull/51856 and https://github.com/vllm-project/vllm/pull/52255 are both still open for system/tools placement;
- https://github.com/vllm-project/vllm/pull/52865 is still open for incremental tool-argument streaming.

The streaming fast path is installed only for the affected upstream contract, `arg_structural_chars=frozenset('>')`. It automatically skips installation after upstream removes that filter and fails fast on an unknown non-null contract.

### Does this PR introduce _any_ user-facing change?

Yes. DeepSeek V4 requests preserve their system prompt when tools are present, omitted thinking controls no longer leak reasoning/tool markup into content, and long string tool arguments stream incrementally rather than being buffered until parameter close.

### How was this patch tested?

Tested against both vLLM sources supported by the current `main` branch:

```bash
PYTHONPATH=$VLLM_MAIN_SOURCE pytest -q \
  tests/ut/patch/platform/test_deepseek_v4_thinking.py \
  tests/ut/patch/platform/test_deepseek_v4_tool_streaming.py
# 22 passed

VLLM_VERSION=0.27.1 PYTHONPATH=$VLLM_RELEASE_SOURCE pytest -q \
  tests/ut/patch/platform/test_deepseek_v4_thinking.py \
  tests/ut/patch/platform/test_deepseek_v4_tool_streaming.py
# 22 passed

[...]

PR #15317, opened 2026-08-29
[BugFix][Frontend] Handle DeepSeek V4 trailing system messages

### LINKED ISSUE #11951: [Bug]: [main] DSv4 returns empty response when receiving mid-conversation system messages (system -> user -> system)
### Your current environment

npu & gpu

### 🐛 Describe the bug

A similar root cause has been observed in the vLLM ecosystem when running DSv4. As detailed in [vllm-project/vllm#40854 (comment-4864596205)](https://github.com/vllm-project/vllm/issues/40854#issuecomment-4864596205), presenting a system -> user -> system sequence to the vLLM engine causes prompt template rendering failures, confirming that multi-turn system message structures currently break the inference pipeline for this model variant across different setups.

### Description
Describe the bug
The DSv4 model currently lacks support for mid-conversation system messages. When an input payload contains multiple system messages interspersed within the conversation history (e.g., a sequence of system -> user -> system), the model fails to parse the chat template correctly and returns an empty response.

### To Reproduce

* Send a chat completion request to the DSv4 model with a multi-turn conversation.

* Insert a second system message after a user or assistant turn.

* See the empty string or null response returned by the API/model.

### Suggested Workarounds (Temporary Fix)
If you are encountering this issue before an official fix is released, you can try the following workarounds:

* Downgrade Client Tools (e.g., Claude Code): If you are using an agent framework like Claude Code that automatically injects mid-conversation system messages, downgrade to an older version that does not utilize this multi-system message behavior.

* Adjust Agent Logic: Modify your application or agent logic to avoid sending system roles mid-conversation. Instead, concatenate new instructions into the user prompt or maintain a single system prompt at the beginning.

* Await vLLM/Framework Updates: Monitor the upstream community (such as vLLM or SGLang) for upcoming pull requests that patch the DSv4 chat template parsing logic.

QwertyJack:

Backport the DeepSeek V4 trailing-system prompt-rendering fix from https://github.com/vllm-project/vllm/pull/51262 for the vLLM version currently supported by vLLM-Ascend.

For a message sequence such as `system -> user -> system`, the existing renderer leaves the prompt at a bare system-message boundary. This patch appends the existing assistant/thinking transition after a trailing system message or before an assistant history turn, while preserving `system -> user` and `system -> latest_reminder` behavior.

The platform monkey patch uses a behavior probe and is skipped automatically once the supported vLLM version contains the upstream fix.

Fixes #11951

Depends on #14632

Yes. DeepSeek V4 requests ending in a system message now start generation from the correct assistant boundary instead of potentially returning empty content or raw tags.

Regression tests cover trailing system messages in chat and thinking modes, `system -> assistant`, and a non-trailing mid-conversation system message.

```bash
VLLM_PLUGINS="" PYTHONPATH=/path/to/vllm:/path/to/vllm-ascend \
pytest -q -p no:cacheprovider --confcutdir=tests/ut/patch/platform --noconftest \
  tests/ut/patch/platform/test_deepseek_v4_trailing_system.py
```

- Supported release (`v0.27.1`): `4 passed`
- Verified source (`ba07e4a`): `4 passed`
- Current vLLM source containing upstream #51262: behavior probe kept the monkey patch disabled and both chat/thinking prompts rendered correctly
- `ruff check`: passed
- `git diff --check`: passed

No additional NPU inference run was performed for this frontend-only monkey patch. The upstream PR reports 12/12 production-router requests passing after applying the same renderer behavior.

### DIFF
--- a/vllm_ascend/patch/platform/patch_deepseek_v4_trailing_system.py
+++ b/vllm_ascend/patch/platform/patch_deepseek_v4_trailing_system.py
@@ -24,13 +24,18 @@ _original_render_message = deepseek_v4_encoding.render_message
 _render_message_signature = inspect.signature(_original_render_message)
 
 
-def _upstream_handles_trailing_system() -> bool:
-    prompt = _original_render_message(
-        0,
-        [{"role": "system", "content": ""}],
-        thinking_mode="chat",
-    )
-    return prompt.endswith(deepseek_v4_encoding.ASSISTANT_SP_TOKEN + deepseek_v4_encoding.thinking_end_token)
+def _needs_trailing_system_patch() -> bool:
+    try:
+        bound_args = _render_message_signature.bind(
+            index=0,
+            messages=[{"role": "system", "content": ""}],
+            thinking_mode="chat",
+        )
+    except TypeError:
+        return False
+    bound_args.apply_defaults()
+    prompt = _original_render_message(*bound_args.args, **bound_args.kwargs)
+    return not prompt.endswith(deepseek_v4_encoding.ASSISTANT_SP_TOKEN + deepseek_v4_encoding.thinking_end_token)
 
 
 @wraps(_original_render_message)
@@ -41,7 +46,7 @@ def _patched_render_message(*args: Any, **kwargs: Any) -> str:
     index = bound_args.arguments["index"]
     messages = bound_args.arguments["messages"]
     thinking_mode = bound_args.arguments["thinking_mode"]
-    drop_thinking = bound_args.arguments["drop_thinking"]
+    drop_thinking = bound_args.arguments.get("drop_thinking", True)
     message = messages[index]
     if message.get("role") != "system" or message.get("task") is not None:
         return prompt
@@ -59,5 +64,5 @@ def _patched_render_message(*args: Any, **kwargs: Any) -> str:
     return prompt
 
 
-if not _upstream_handles_trailing_system():
+if _needs_trailing_system_patch():
     deepseek_v4_encoding.render_message = _patched_render_message
--- a/vllm_ascend/patch/platform/patch_deepseek_v4_trailing_system.py
+++ b/vllm_ascend/patch/platform/patch_deepseek_v4_trailing_system.py
@@ -14,12 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import inspect
 from functools import wraps
 from typing import Any
 
 from vllm.tokenizers import deepseek_v4_encoding
 
 _original_render_message = deepseek_v4_encoding.render_message
+_render_message_signature = inspect.signature(_original_render_message)
 
 
 def _upstream_handles_trailing_system() -> bool:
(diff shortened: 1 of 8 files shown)


vllm_ascend/patch/platform/patch_deepseek_v4_trailing_system.py
```diff
@@ -0,0 +1,65 @@
...
+        "last_user_idx",
+        args[5] if len(args) > 5 else None,
+    )
+    if last_user_idx is None:
+        last_user_idx = deepseek_v4_encoding.find_last_user_index(messages)
+    if thinking_mode == "thinking" and (not drop_thinking or index >= last_user_idx):
+        prompt += deepseek_v4_encoding.thinking_start_token
+    else:
+        prompt += deepseek_v4_encoding.thinking_end_token
+    return prompt
```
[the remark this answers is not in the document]
QwertyJack: Addressed. The wrapper now caches the original `render_message` signature at module load time and uses `bind()`/`apply_defaults()` to read arguments by name, removing positional assumptions such as `args[3]` and `args[5]`. Validation passed: 3 regression tests on v0.27.1, 6 tests on verified commit `ba07e4a`, and the monkey patch remains a no-op with the upstream #51262 implementation. Ruff, format, and diff checks also passed. I kept the current title because this change targets frontend prompt rendering, not the attention module.

vllm_ascend/patch/platform/patch_deepseek_v4_trailing_system.py
```diff
@@ -0,0 +1,63 @@
...
+from functools import wraps
+from typing import Any
+
+from vllm.tokenizers import deepseek_v4_encoding
+
+_original_render_message = deepseek_v4_encoding.render_message
+_render_message_signature = inspect.signature(_original_render_message)
+
+
+def _upstream_handles_trailing_system() -> bool:
```
nu11ptr7: This probe runs at import time with a hard-coded 3-arg call to `render_message`; if the supported vLLM version's signature drifts, the probe itself raises and vLLM fails to start instead of skipping the patch. A try/except or signature check around the probe would keep the drift case graceful.

QwertyJack: Addressed in 8acec9366. The import-time probe now binds named probe inputs against the cached upstream signature before invoking `render_message`. If the signature is incompatible, `_needs_trailing_system_patch()` returns `False`, so the monkey patch is skipped instead of making vLLM fail during import. I also added a regression test covering an upstream signature with a new required parameter. Ruff, formatting, and diff checks pass; the full local pytest run is currently blocked during global test setup by an unrelated installed-vLLM API mismatch (`UniformTypeKVCacheSpecs.max_num_blocks_per_req`) before this test is collected.

kunpengW-code (dismissed):

QwertyJack:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #15324, opened 2026-08-29
[Performance] Add scatter_nd_update_asc custom op and replace scat…

ZT-AIA:
Introduce the new `scatter_nd_update_asc` custom operator (supports 2D-4D `var` and K=1/2 indices), remove the legacy `scatter_nd_update_v2` custom operator, and switch all call sites to `npu_scatter_nd_update_asc`.

The legacy `scatter_nd_update_v2` custom operator maintained inside vllm-ascend only supports a narrow input range (2D `var` with 1D indices) and forces call sites to pre-reshape their tensors. This PR replaces it with a new, more general custom operator and removes the duplicated legacy code.

1. **Adds the `scatter_nd_update_asc` custom operator** (AscendC kernel + host tiling/infershape + op_api wrapper + C++ binding), supporting:
   - `var` tensors of 2D–4D shapes (ND format)
   - `indices` with K = 1 or 2 (last dim 1 or 2), covering both flat slot mapping and `[block_idx, offset]` block-level mapping
   - In-place update semantics: `var[indices[k]] = updates[k]`
   - Skipping out-of-bounds / negative indices, so `-1` padding slots in slot mapping never write into the cache

2. **Removes the legacy `scatter_nd_update_v2` operator**:
   - `csrc/moe/scatter_nd_update_v2/` (AscendC kernel, host tiling/infershape, op_api wrappers, per-SOC configs)
   - Its bindings in `csrc/torch_binding.cpp` and `csrc/torch_binding_meta.cpp`
   - Its entry in `CUSTOM_OPS_ARRAY` in `csrc/build_aclnn.sh`

3. **Switches all call sites to `npu_scatter_nd_update_asc`**:
   - `vllm_ascend/device/device_op.py`: `dsa_kv_compress_scatter`, `indexer_quant_scatter`, `warmup_indexer_quant_scatter` (7 calls)
   - `vllm_ascend/models/minimax_m3/minimax_m3.py`: `_insert_kv` index-cache scatter (1 call)

**Benefits:**
- Broader shape/index support removes the need for shape conversion at call sites.
- Single source of truth: drops the duplicated legacy operator and reduces maintenance burden.
- Keeps the same in-place, last-write-wins semantics for KV/index-cache scatter.

No. All changes are internal: the internal custom operator used by the DSA and MiniMax-M3 cache-scatter paths is replaced; no public API, configuration, or model interface is changed.

- CI passed with existing tests.
- Validated on NPU (Ascend): exercised the DSA scatter and MiniMax-M3 index-cache scatter paths, confirming numeric equivalence with the previous `scatter_nd_update_v2` behavior, including negative/out-of-bounds index skipping.
- Verified no residual references to `scatter_nd_update_v2`, `ScatterNdUpdateV2`, or `aclnnScatterNdUpdateV2` remain in the repository.


vllm_ascend/models/minimax_m3/minimax_m3.py
```diff
@@ -287,7 +287,7 @@ def _insert_kv(
         flat = idx_cache.view(-1, self.idx_head_dim)
         # Scatter ND update ignores indices outside the cache bounds, so graph
         # padding slots set to -1 do not write into the last cache row.
-        torch.ops._C_ascend.npu_scatter_nd_update_v2(
+        torch.ops._C_ascend.npu_scatter_nd_update_asc(
```
nu11ptr7, line 290: Main has since moved this scatter into `_scatter_index_cache`, which still calls `npu_scatter_nd_update_v2` (and its UT mocks that name). This branch is based on an older main and only renames the call here, so after merge the M3 index-cache path would hit a deleted op - worth rebasing and updating the helper + its test.

PR #15325, opened 2026-08-29
[Feat][MoE] Wire A3 GroupedMatmulSituQuant into W4A8 SiTU path

MaybeChz:

Vendors the A3 fused operator `GroupedMatmulSituQuant` (fake-A8W8 GMM1 + SiTU + per-token INT8, single launch) and wires it into the main-branch Kimi-k3 W4A8 SiTU MoE path.

- Kernel/pybind: `csrc/gmsq_situ_quant/` + `torch.vllm_ascendC.grouped_matmul_situ_quant`
- Framework: in `quant_apply_mlp`, on A3 (non-MX) when `quantize_situ_output`, replace `npu_grouped_matmul` (GMM1) + `dequant_situ_quant` with the fused op; **GMM2 stays separate**
- Gate: `VLLM_ASCEND_ENABLE_GMSQ_SITU` (default `1`); set `0` to force the legacy split path for A/B

Adapted on top of current `main` (SiTU uses `MoEActivation.SITU` rather than the v0.26 `SituActivationConfig` helper path).

- New optional env: `VLLM_ASCEND_ENABLE_GMSQ_SITU` (default enabled when `vllm_ascendC` is present on A3)
- No public Python API change for model users

- Unit tests updated/added in `tests/ut/ops/test_moe_mlp.py` (legacy path with GMSQ disabled; A3 fused path with mocked op)
- End-to-end accuracy/perf on Ascend A3: to be verified by author

### DIFF
--- a/csrc/gmsq_situ_quant/grouped_matmul_situ_quant.cpp
+++ b/csrc/gmsq_situ_quant/grouped_matmul_situ_quant.cpp
@@ -1,0 +1,381 @@
+/*
+ * Copyright (c) 2026 Huawei Technologies Co., Ltd.
+ * This program is free software, you can redistribute it and/or modify it under the terms and conditions of
+ * CANN Open Software License Agreement Version 2.0 (the "License").
+ * Please refer to the License for details. You may not use this file except in compliance with the License.
+ * THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND, EITHER EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
+ * See LICENSE in the root of the software repository for the full text of the License.
+ */
+
+// GroupedMatmulSituQuant op_host.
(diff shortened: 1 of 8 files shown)


csrc/gmsq_situ_quant/grouped_matmul_situ_quant.cpp
```diff
@@ -0,0 +1,381 @@
...
+            fullWPtrs[static_cast<size_t>(e)] = reinterpret_cast<uintptr_t>(weight[e].data_ptr());
+            fullSPtrs[static_cast<size_t>(e)] = reinterpret_cast<uintptr_t>(weight_scale[e].data_ptr());
+            int64_t wfmt = at_npu::native::get_npu_format(weight[e]);
+            if (wfmt != ACL_FMT_ND && wfmt != ACL_FMT_FRACTAL_NZ) {
+                ptrsOk = false;
+            }
+        }
+
+        FusedMetaCache &fc = g_fusedCache;
+        bool hit = ptrsOk && fc.valid && (fc.K == K) && (fc.N == N) && (fc.C == C) &&
```
nu11ptr7, line 227: This is a single-slot cache keyed on the full weight-pointer set plus C, so in a multi-layer MoE model every layer's forward misses and rebuilds (per-expert format cast + H2D table copy), and a changing capacity misses even on the same layer. A small map keyed on pointers/shapes would let the steady-state zero-copy path actually hit.

PR #15332, opened 2026-08-29
[Feature][Kernel] Extend split_qkv_rmsnorm_rope with value normalization for Gemma4

dwoai22:

Gemma4 runs its pre-attention chain, qkv split, per-head q/k RMSNorm, RoPE, and a weight-less per-head **v** RMSNorm, as separate ops on Ascend, so it cannot use the existing fused `split_qkv_rmsnorm_rope` kernel, which copies V straight out of the fused QKV tensor.

The existing `qknorm_rope_fusion_pass` cannot cover it either: it has no notion of the V norm, and Gemma4 alternates sliding and full attention layers with different `head_dim`, kv head counts and RoPE type, while `torch.compile` matches a single pattern against every attention layer.

This PR adds a `split_qkv_rmsnorm_rope_vnorm` Triton kernel and wires Gemma4 to it.

**Kernel.** A variant of `split_qkv_rmsnorm_rope` that normalizes V instead of copying it. Because V now needs the same float32 reduction scratch as Q and K, the separate wide-tile V loop that the base kernel uses for the plain copy no longer pays for itself, so V is folded into the q/k loop:

- one `tl.load` covers the whole fused row (`total_hidden_size`) instead of two separate column segments;
- one RMSNorm reduction covers every q, k **and** v head of the tile, reshaped to `(tile * (q_head_num + 2*kv_head_num), HEAD_DIM)`;
- Q, K and V are then cut out with three `extract_slice` calls at head offsets `0`, `q_head_num` and `q_head_num + kv_head_num`. Q and K apply their weights and RoPE; V has neither, so it is a cast and a store that reuses the K output index and mask.

This is valid because Gemma4's `v_norm` is the same per-head RMSNorm with the same `eps` and no learnable scale, the weight application is what moves after the shared reduction.

The kernel keeps everything else from `split_qkv_rmsnorm_rope`: the 1-D grid over vector cores, the batched UB tiling, the per-tile cos/sin gather shared by all heads, and partial-RoPE support. The UB tiling factor grows by `4 * kv_hidden_size` to account for the V columns now living in the tile.

Full attention layers use proportional RoPE, which `Gemma4RotaryEmbedding` implements by zero-padding `inv_freq` and passing `rotary_dim=head_size` to the base class, so its cos/sin cache is an ordinary full-width cache and the kernel needs no layer-type knowledge.

`Gemma4Attention.__init__` is patched as well to resolve the fusion decision at build time and log it. Model construction never runs inside a compiled region, so this keeps the branch out of the traced graph and keeps the outcome observable in every compilation mode, unlike a check inside `forward`, which `torch.compile` evaluates once at trace time and an ACL graph then replays without Python.

Layers fall back to the upstream unfused chain, with the reason logged, when: the layer is KV-shared; `v_norm` carries a learnable scale; any q/k/v norm carries a bias; the three norms use different epsilons; weights or the cos/sin cache are not bfloat16; the rotary embedding is not neox-style; or one token's tile exceeds the vector core unified buffer (which excludes full attention layers below tensor parallel size 4).

No API or interface change. Gemma4 on Ascend picks up the fused pre-attention path automatically; every layer that does not qualify keeps the existing behavior and logs why.

Tested on 910B1, TP=4, dtype bfloat16 with gemma4-31b-it.

**Performance**

perf_v2

**Accuracy**

| Benchmark | Unfused (baseline) | Fused (this PR) |
|-----------|-------------------:|----------------:|
| HumanEval | 96.74 | 96.74 |
| GPQA      | 76.09 | 76.28 | 

Averaged over 3 runs.

**Correctness**

pytest -sv tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_split_qkv_rmsnorm_rope.py


dwoai22:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

dwoai22:
@zzzzwwjj @realliujiaxu @weijinqian0  This is ready for review, thanks in advance for taking a look.

`pre-commit`, `DCO`, and `select-tests` pass. `ci-gate` is red only because `run-selected-tests` is gated behind the `ready-precise` / `ready-all` label, which I don't have permission to add. Could one of you add `ready-precise` so the selected tests can run?

linfeng-yuan:
> @zzzzwwjj @realliujiaxu @weijinqian0 This is ready for review, thanks in advance for taking a look.
> 
> `pre-commit`, `DCO`, and `select-tests` pass. `ci-gate` is red only because `run-selected-tests` is gated behind the `ready-precise` / `ready-all` label, which I don't have permission to add. Could one of you add `ready-precise` so the selected tests can run?

@ZT-AIA  PTAL

Ronald1995:
please add a doc for this operator, such as:

image

vllm_ascend/patch/worker/patch_gemma4.py
```diff
@@ -0,0 +1,153 @@
...
+            scope="global",
+        )
+
+
+def _patched_attention_init(self, *args, **kwargs) -> None:
+    _original_attention_init(self, *args, **kwargs)
+    _configure_fused_preattention(self)
+
+
+def _patched_attention_forward(
```
Ronald1995, line 100: it's better to do like this
```python
_original_attention_forward = Gemma4Attention.forward

if self.is_kv_shared_layer:
    return _original_attention_forward(
        self,
        positions,
        hidden_states,
        **kwargs,
    )
else:
    if self.use_fused_preattention:
        pass
    else:
         return _original_attention_forward(
        self,
        positions,
        hidden_states,
        **kwargs,
    )

    

```

dwoai22:
@Ronald1995 I have added a document for the operator: `vllm_ascend/ops/triton/docs/split_qkv_rmsnorm_rope_vnorm.md`

 Let me know if you want further changes in the doc.

vllm_ascend/patch/worker/patch_gemma4.py
dwoai22: Changed accordingly. Now, unfused path uses original forward method directly.

Ronald1995 (approved):

weijinqian0:
plz realize it by pass.

dwoai22:
> plz realize it by pass.

@weijinqian0 Sure, I can work on it.

One thing that shapes the design: Gemma4 has two attention shapes. Sliding layers use `head_dim` 256 with 16 KV heads; full attention layers use `head_dim` 512 with 4 KV heads (`global_head_dim` / `num_global_key_value_heads`, resolved per layer by `gemma4_layer_config`). The cos/sin cache width differs as well, 256 vs 512.

### Plan

Add one new pattern class to `qknorm_rope_fusion_pass.py` rather than a new file, matching how `norm_quant_fusion_pass.py` groups its patterns, and register it from a second block that keeps its own conditions:

```python
# existing block unchanged: base patterns, still registered only for head_size == 128

# new block: V-norm variant, one registration per attention shape
eps = vllm_config.model_config.hf_text_config.rms_norm_eps
for head_dim, num_heads, num_kv_heads in sorted(
    {(l.head_size, l.num_heads, l.num_kv_heads) for l in attn_layers.values()}
):
    if not qkv_rmsnorm_rope_vnorm_fits_ub(
        q_hidden_size=num_heads * head_dim,
        kv_hidden_size=num_kv_heads * head_dim,
        head_dim=head_dim,
        rope_dim=head_dim,
    ):
        continue
    QKVNormRopeFusionPattern(
        vllm_config, head_dim, num_heads, num_kv_heads, eps
    ).register(self.pattern_match_passes)
```

`rope_dim == head_dim` for both layer types, because `Gemma4RotaryEmbedding` passes `rotary_dim=head_size` to the base class, so the cos/sin cache is full-width in both cases. The new pattern derives it locally and `get_rope_dim()` stays untouched. The UB check is the same one the current patch uses, so full-attention layers still fall back below TP4.

### Registering per shape

Both shapes need their own registration, for the same reason the pass already loops over `eps`: `register_replacement` traces the pattern once and the constants become literals in the matched graph. Sliding layers emit `split_with_sizes(qkv, [2048, 1024, 1024], -1)`, full attention layers emit `[4096, 512, 512]`, so one registration cannot cover both.

That runs into the dedup key in `BasePattern.register`:

```python
# base_pattern.py:43
pattern_id = f"{self.__class__.__name__}_{self.eps}"
if pattern_id in _registered_patterns:
    return
```

It includes `eps` because `eps` varies between registrations of the same class. It omits shape only because no model has needed two shapes until now, so the second registration would silently return and only 50 of 60 layers would fuse, with no error.

I'd add a small seam so subclasses can widen that identity, leaving the dedup logic itself untouched:

```diff
 class BasePattern(ABC):
+    def pattern_key(self) -> str:
+        """Identity for dedup. Subclasses registered per shape widen this."""
+        return f"{self.__class__.__name__}_{self.eps}"
+
     def register(self, pm_pass: PatternMatcherPass) -> None:
-        pattern_id = f"{self.__class__.__name__}_{self.eps}"
+        pattern_id = self.pattern_key()
         if pattern_id in _registered_patterns:
             return
         ...
         _registered_patterns.add(pattern_id)
```

The new pattern then extends the base key with the constants that get baked into its traced graph:

```python
class QKVNormRopeFusionPattern(BasePattern):
    def pattern_key(self) -> str:
        return (
            f"{super().pattern_key()}"
            f"_hd{self.head_dim}_nh{self.num_heads}"
            f"_nkv{self.num_kv_heads}_rd{self.rope_dim}"
        )
```

On gemma-4-31B at TP=4 that yields two distinct keys, so both shapes register:

```
QKVNormRopeFusionPattern_1e-06_hd256_nh8_nkv4_rd256
QKVNormRopeFusionPattern_1e-06_hd512_nh8_nkv1_rd512
```

Three additive lines, and both existing patterns keep byte-identical keys. If you'd rather not touch `base_pattern.py` at all, I can generate a per-shape subclass inside the pass file instead, just let me know which you prefer.

### Leaving the head_size guard alone

I'd keep `head_size != 128` on the existing block as it is. If I relax it, the existing `QKNormRopeFusionPattern` also becomes eligible for Gemma4, and since its pattern is a prefix of the V-norm one it could match first and leave `v_norm` unfused. Giving the new pattern its own block avoids that and changes nothing for existing models. Also, I am not sure what effect it would have to relax that restriction.

`patch_gemma4.py` and its registry entry then go away. KV-shared layers need no special handling either, since they simply won't match the pattern.

**Do you think this design is OK?**

PR #15333, opened 2026-08-29
[Misc]feat: adapt to vLLM main (574e6a00)

wjunLu:

Adapt vllm-ascend to vLLM main commits up to August 21.

### Changes

| Files | Upstream vLLM change | vllm-ascend adaptation |
|-------|---------------------|------------------------|
| `vllm_ascend/models/qwen3_dflash2.py` | [574e6a00](https://github.com/vllm-project/vllm/commit/574e6a00efd3d594c4d7f72e766c3e605316a118) — In the accumulated range ba07e4a48..574e6a00ef, PR #52816 changed the DFlash draft-model constructors in vllm/model_executor/models/qwen3_dflash.py to build layers / the draft via the class attributes `self.decoder_layer_cls` and `self.model_cls` instead of referencing the module-global `DFlashQwen3DecoderLayer` / `DFlashQwen3Model`. vllm-ascend's DFlash2 model monkey-patched those module globals before calling `super().__init__()`, which became a silent no-op on main (parent now reads the class attribute, not the patched global). | In vllm_ascend/models/qwen3_dflash2.py, added `decoder_layer_cls = DFlash2Qwen3DecoderLayer` on `DFlash2Qwen3Model` and `model_cls = DFlash2Qwen3Model` on `DFlash2Qwen3ForCausalLM` so the main parent constructors pick up the DFlash2 classes; the module-global monkey-patch is now guarded with `vllm_version_is("0.27.1")` (only needed on the release branch, whose parent still reads the module global). Added the `vllm_version_is` import. No other changed upstream symbols affect vllm-ascend (FlashInfer LSE, DraftModelSpeculator hook, LogitsProcessor top-k, SupportsMultiModal flag, NCCL/XPU/b12x changes are additive or platform-specific). |
| `tests/ut/test_ascend_config.py` | [4b25eb34](https://github.com/vllm-project/vllm-ascend/commit/4b25eb34793efa9f00201d654eb2f107ecebfbbd) — vllm-ascend PR #15267 "roll back megamoe op" modified `AscendConfig` in vllm_ascend/ascend_config.py so `enable_fused_mc2=1` now forces `_MEGA_MOE_SUPPORTED=False` (falling back to dispatch_ffn_combine), and reserved `enable_fused_mc2=2` as the megamoe-testing path that sets `_MEGA_MOE_SUPPORTED = importlib.util.find_spec("cann_ops_transformer") is not None` and resets `enable_fused_mc2` back to 1. This broke `test_fused_mc2_is_disabled_for_unsupported_megamoe_config`, which patched `_MEGA_MOE_SUPPORTED` to True with `enable_fused_mc2=1` and expected the unsupported-model-config downgrade to 0 to fire. | Updated the test in tests/ut/test_ascend_config.py to drive the megamoe-testing path: replaced the `_MEGA_MOE_SUPPORTED` patch with `@patch("importlib.util.find_spec", return_value=object())` (so `cann_ops_transformer` resolves as present) and set `vc.additional_config = {"enable_fused_mc2": 2}`, so the unsupported-model config still gets downgraded to `enable_fused_mc2=0`. Added the `mock_find_spec` fixture parameter and a comment documenting the post-rollback semantics (the `enable_fused_mc2=1` path can no longer reach the downgrade because it forces `_MEGA_MOE_SUPPORTED=False`). |
| `tests/ut/test_ascend_config.py` | [9cf7a0a2](https://github.com/vllm-project/vllm-ascend/commit/9cf7a0a23944b032d512040f4da542777042cc3e) — vllm-ascend PR #13838 migrated `AscendConfig` and its 12 sub-config classes to upstream vLLM's `@config` pydantic dataclass with typed fields and `extra="forbid"`. The UT suite deliberately constructs these configs with string values (e.g. `enabled="true"`, `threshold="512"`) to verify lax coercion, and with undeclared keys to verify rejection. Because CI runs mypy over the `tests` tree (`tools/mypy.sh`), those typed-violating constructor calls now surface as mypy errors. | Added per-call `# type: ignore[arg-type]` (string-to-typed-field) and `# type: ignore[call-arg]` (undeclared keyword) comments to the intentional type-violating constructor invocations in tests/ut/test_ascend_config.py (RlConfig, ShortRequestFirstConfig, AscendFusionConfig, AscendCompilationConfig, ProfilingChunkConfig, DyntraLBConfig, EplbConfig), keeping the CI mypy check green without weakening the lax-coercion / forbidden-key regression coverage. |


wjunLu:
Superseded by #15348.

Issue #1526, opened 2025-06-30
[Usage]: Is FP8 Quantization Version(Qwen/Qwen3-235B-A22B-FP8) for Qwen3 Supported?


zkryakgul:
### Your current environment

I want to ask about is Qwen/Qwen3-235B-A22B-FP8 supported by vllm-ascend?

### How would you like to use vllm on ascend

I want to run inference of a [specific model](put link here). I don't know how to integrate it with vllm.

KevinXyy:
Seems like FP8 quantization is not suppoprted

yiz-liu:
Well for FP8 you will need the A5, for A2/A3 it's a no go.

Issue #1383, opened 2025-06-24
[Bug]: [v0.9.1rc1] Use 310P3 Not Start. [Log] Optype [TransData] of Ops kernel [AIcoreEngine] is unsupported. Reason: [tbe-custom]:op type TransData is not found in this op store.


LwengGitHub:
### Your current environment

Use 310P3 Not Start.

[Log] 
Optype [TransData] of Ops kernel [AIcoreEngine] is unsupported. Reason: [tbe-custom]:op type TransData is not found in this op store.[tbe-custom]:op type TransData is not found in this op store.[Dynamic shape check]: The format and dtype is not precisely equivalent to format and dtype in op information library[Static shape check]:The format and dtype is not precisely equivalent to format and dtype in op information library.

### 🐛 Describe the bug

LogInfo:
[root@74dae9b20914 workspace]# export VLLM_USE_MODELSCOPE=true
[root@74dae9b20914 workspace]# vllm serve Qwen/Qwen2.5-0.5B-Instruct &
[1] 332
[root@74dae9b20914 workspace]# INFO 06-23 10:49:40 [importing.py:17] Triton not installed or not compatible; certain GPU-related functions will not be available.
WARNING 06-23 10:49:40 [importing.py:29] Triton is not installed. Using dummy decorators. Install it via `pip install triton` to enable kernel compilation.
INFO 06-23 10:49:42 [__init__.py:39] Available plugins for group vllm.platform_plugins:
INFO 06-23 10:49:42 [__init__.py:41] - ascend -> vllm_ascend:register
INFO 06-23 10:49:42 [__init__.py:44] All plugins in this group will be loaded. Set `VLLM_PLUGINS` to control which plugins to load.
INFO 06-23 10:49:43 [__init__.py:235] Platform plugin ascend is activated
WARNING 06-23 10:49:46 [_custom_ops.py:22] Failed to import from vllm._C with ModuleNotFoundError("No module named 'vllm._C'")
WARNING 06-23 10:49:51 [registry.py:401] Model architecture DeepSeekMTPModel is already registered, and will be overwritten by the new model class vllm_ascend.models.deepseek_mtp:CustomDeepSeekMTP.
WARNING 06-23 10:49:51 [registry.py:401] Model architecture Qwen2VLForConditionalGeneration is already registered, and will be overwritten by the new model class vllm_ascend.models.qwen2_vl:AscendQwen2VLForConditionalGeneration.
[...]

PR #15348, opened 2026-08-30
[Misc]feat: adapt to vLLM main (574e6a00)

wjunLu:

Adapt vllm-ascend to vLLM main commits up to August 21.

### Changes

| Files | Upstream vLLM change | vllm-ascend adaptation |
|-------|---------------------|------------------------|
| `vllm_ascend/models/qwen3_dflash2.py` | [574e6a00](https://github.com/vllm-project/vllm/commit/574e6a00efd3d594c4d7f72e766c3e605316a118) — PR 52816 (b389ac2946) refactored `DFlashQwen3Model`/`DFlashQwen3ForCausalLM` to build layers and the draft model via class attributes `decoder_layer_cls`/`model_cls` instead of the module-global `DFlashQwen3DecoderLayer`/`DFlashQwen3Model`. vllm-ascend's DFlash2 model monkeypatched those module globals around `super().__init__()`, which became a silent no-op on the new upstream and would degrade the DFlash2 draft to DFlash1. | In `vllm_ascend/models/qwen3_dflash2.py`, added `decoder_layer_cls = DFlash2Qwen3DecoderLayer` to `DFlash2Qwen3Model` and `model_cls = DFlash2Qwen3Model` to `DFlash2Qwen3ForCausalLM` so the new upstream parent ctors build DFlash2 components directly; kept the module-global swap in the `if vllm_version_is("0.27.1")` branch (v0.27.1 parent still reads module globals) and imported `vllm_version_is` from `vllm_ascend.utils`. All other files in the range (b12x MoE, NVIDIA/XPU quant, FlashInfer LSE, gumbel, LoRA `supports_tower_connector_lora`, mrope `forward_xpu`) are no-ops for vllm-ascend — verified no imports/call sites/subclasses. |
| `tests/ut/test_ascend_config.py` | [PR #1383](https://github.com/Ascend/vllm-ascend/pull/13838) — test-only mypy cleanup — added `# type: ignore[arg-type]` / `# type: ignore[call-arg]` comments on every pydantic config ctor call in `tests/ut/test_ascend_config.py` that intentionally passes coercion-test literals (`"false"`, `"4096"`, `"2.5"`, unknown keys like `{"foo": 1}`, ...) whose types don't match the declared pydantic field types. No runtime behavior change; assertions are unchanged. | — |
| `tests/ut/test_ascend_config.py` | [PR #1526](https://github.com/Ascend/vllm-ascend/pull/15267) — vllm-ascend rolled back the megamoe op by default: `enable_fused_mc2=1` now forces `_MEGA_MOE_SUPPORTED = False` and falls back to `dispatch_ffn_combine`, while `enable_fused_mc2=2` (the megamoe testing path) sets `_MEGA_MOE_SUPPORTED` from `importlib.util.find_spec("cann_ops_transformer")`. The old test, which used `enable_fused_mc2=1` with `_MEGA_MOE_SUPPORTED` patched True, no longer exercised the config-downgrade branch because `1` short-circuits to False. | In `tests/ut/test_ascend_config.py`, `test_fused_mc2_is_disabled_for_unsupported_megamoe_config` now sets `vc.additional_config = {"enable_fused_mc2": 2}` (instead of 1), patches `vllm_ascend.ascend_config.importlib.util.find_spec` to return a truthy object (so `_MEGA_MOE_SUPPORTED` becomes True), and updates the test signature to accept the new `mock_find_spec` param; the `_is_megamoe_supported_by_config` return_value=False patch is kept so the assert `init_ascend_config(vc).enable_fused_mc2 == 0` still verifies the downgrade-to-0 path. |


wjunLu:
Superseded by #15711.

PR #15354, opened 2026-08-30
[BugFix][MRV2] Isolate Eagle draft parallel config

jiajinzhu2:

ModelRunner V2 builds the EAGLE draft runtime config from the target model's `VllmConfig`. When an MoE target enables expert parallelism, those parallel flags are inherited by the dense EAGLE draft model. `VllmConfig` then validates the draft as an expert model and fails because the draft has no experts:

```text
Value error, Number of experts in the model must be greater than 0 when expert parallelism is enabled.
```

This change overrides the EAGLE draft runtime config to disable EP and EPLB only for the draft model. The target model keeps its original EP/EPLB configuration. A source comment documents why these flags must not be inherited.

No.

Clean A/B startup validation with MiniMax-M3 BF16 + EAGLE3, TP16/DP1 and target EP enabled:

- Before: startup fails while validating the dense draft model with inherited expert parallelism.
- After: the target and draft weights load and the service starts successfully in both `FULL_DECODE_ONLY` and `PIECEWISE` graph modes.

Also verified with `git diff --check`.

### DIFF
--- a/vllm_ascend/worker/v2/spec_decode/eagle/speculator.py
+++ b/vllm_ascend/worker/v2/spec_decode/eagle/speculator.py
@@ -16,13 +16,28 @@
 # limitations under the License.
 # This file is a part of the vllm-ascend project.
 #
+from vllm.config import VllmConfig, replace
 from vllm.v1.worker.gpu.spec_decode.eagle.speculator import EagleSpeculator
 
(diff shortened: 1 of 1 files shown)


lilinsiman (approved):

slippersss:
LGTM, but we need to delete it later.

slippersss (approved):

linfeng-yuan (approved):

PR #14643, opened 2026-08-20
[Performance] Remove redundant deepcopy of scheduler_output in execute_model

xiayingqing:
### What this PR does / why we need it?

Remove the redundant `deepcopy(scheduler_output)` in
`NPUModelRunner.execute_model` on the async-scheduling + spec-decode path.

This deepcopy was removed upstream in vllm-project/vllm#29821, which
replaced the in-worker mutation of `scheduler_output` with
`_copy_draft_token_ids_to_cpu()` / `take_draft_token_ids()` /
`Scheduler.update_draft_token_ids_in_output()` (executed in EngineCore).

It was accidentally re-introduced by  vllm-project/vllm-ascend#6043, which was
developed against v0.13.0. The code it protected (in-worker mutation of
`scheduler_output`) has never existed in this repo — nothing below
mutates `scheduler_output` in the worker, so the copy is pure overhead.

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

- vLLM version: v0.27.1
- vLLM main: https://github.com/vllm-project/vllm/commit/58d3918e3ea0a544ffedadad2ba84559e9c51d8f

PR #15365, opened 2026-08-31
[Performance][KV Pool] Avoid redundant KV pool metadata copies

bowgneo:

Backport two CPU-overhead reductions to the 0.26 release branch:

- Retain prompt token IDs in AscendStore request metadata only when KV cache events are enabled, avoiding prompt-sized list copies on the default path.
- Remove the redundant async-scheduling `deepcopy(scheduler_output)`. Draft token IDs are propagated through the current EngineCore path, so the worker does not mutate the scheduler output. This change is already present on `main` via #14643.

No.

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_scheduler.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_scheduler.py
@@ -98,6 +98,8 @@ class KVPoolScheduler:
             "consumer_is_to_put", False
         )
         self.load_async = vllm_config.kv_transfer_config.kv_connector_extra_config.get("load_async", False)
+        kv_event_config = vllm_config.kv_events_config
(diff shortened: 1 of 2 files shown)


vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_scheduler.py
```diff
@@ -98,6 +98,8 @@ def __init__(
             "consumer_is_to_put", False
         )
         self.load_async = vllm_config.kv_transfer_config.kv_connector_extra_config.get("load_async", False)
+        kv_event_config = vllm_config.kv_events_config
+        self.enable_kv_events = bool(kv_event_config and kv_event_config.enable_kv_cache_events)
```
bowgneo: use getattr(vllm_config, "kv_events_config", None) instead.

PR #15366, opened 2026-08-31
[BugFix][KVTransfer] Split prefill PP keys on precise layer boundaries for heterogeneous-layer models

lsy0214:

`ChunkedTokenDatabase.decode_adaptor_prefill_pp` splits the KV key/addr/size lists across pipeline-parallel ranks for the non-layerwise AscendStore PD path. The current implementation assumes every physical layer contributes the same number of cache entries and splits by a uniform `caches_per_layer` count:

```python
caches_per_layer = len(addr_list) // group_num_layers
end = start + part * caches_per_layer
```

For heterogeneous-layer models this cuts the split point in the middle of a layer and corrupts the `@pp_rank` key mapping. DeepSeek-V4 is affected on three counts:

- `compress_ratios: [0, 0, 4, 128, 4, 128, ...]` - dense / compressed / indexer layers are interleaved, so cache entry counts differ per layer;
- DSA sparse-attention layers each carry 5 cache specs (attn, indexer.k_cache, compressor.state_cache, swa_cache, ...);
- `num_nextn_predict_layers: 1` - the MTP tail layer makes the real layer count 44 while `partitions` is computed over `num_hidden_layers`=43.

This PR makes the split prefer the precise per-physical-layer offsets already tracked in `group_layer_cache_entry_offsets`: each partition boundary lands exactly on a layer boundary, and trailing layers not covered by `partitions` (e.g. MTP) are folded into the last partition. Homogeneous layouts without an offsets table keep the legacy uniform split, so existing deployments are unaffected.

No behavior change for homogeneous-layer models with PP=1 or models that do not use the AscendStoreConnector non-layerwise PD path. For heterogeneous-layer models (DeepSeek-V4) with `prefill_pp_size > 1`, the PP key mapping is now correct instead of corrupted.

- Unit tests added: `tests/ut/kv_transfer/test_decode_adaptor_prefill_pp.py` (7 cases)
  - DSV4-like layout (44 physical layers, 126 cache entries, `partitions=[21,22]`): split lands exactly at the layer-21 boundary (entry 59) instead of the legacy entry 42 (mid layer-14); MTP tail folded into the last partition; short addr lists are clamped; sizes split consistently with addrs; keys tagged with correct `@pp_rank`.
  - Homogeneous layout without offsets keeps the legacy uniform split (regression guard).
  - `partitions=[N]` single-partition passthrough.
- Manual testing on 8x910B3 (npu2):
  - 8 ranks joined the memcache store (rankId 0-7), zero KV transfer errors, 50/50 requests succeeded.
  - Throughput 244.24 tok/s vs 146.70 tok/s TP8 baseline (+66%), P99 latency 10.54s vs 29.69s.

### DIFF
--- a/tests/ut/kv_transfer/test_decode_adaptor_prefill_pp.py
+++ b/tests/ut/kv_transfer/test_decode_adaptor_prefill_pp.py
@@ -1,0 +1,182 @@
+"""Unit tests for ``decode_adaptor_prefill_pp`` layer-boundary splitting.
+
+Covers the PP (pipeline parallel) key/addr/size re-splitting logic of
+``ChunkedTokenDatabase`` used by the non-layerwise AscendStore PD path:
+
+- Heterogeneous-layer models (DeepSeek-V4: dense + DSA multi-cache-spec
+  layers + MTP tail layer) must be split exactly on physical-layer
+  boundaries via ``group_layer_cache_entry_offsets``.
+- Homogeneous-layer models without the offsets table keep the legacy
+  uniform ``caches_per_layer`` split.
+"""
+
+from typing import Any
+
+import pytest
+
+from vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.metadata import (
+    ChunkedTokenDatabase,
+    KeyMetadata,
+)
+
+BASE_KEY = (
+    "DSV4@pcp0@dcp0@head_or_tp_rank:0@pp_rank:0"
+    "@group:0@cache_role:kv@cache_family:default@hash_abcd"
+)
+
+
+def _make_db(
+    partitions: list[int],
+    num_layers: int,
+    offsets: list[int] | None,
+    total_entries: int,
+) -> tuple[ChunkedTokenDatabase, Any]:
+    meta = KeyMetadata(
+        model_name="DSV4",
+        head_or_tp_rank=0,
+        pcp_rank=0,
+        dcp_rank=0,
+        pp_rank=0,
+        kv_cache_group_id=0,
+    )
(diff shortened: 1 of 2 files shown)


lsy0214:
Addressed the review feedback on the surplus-entry clamping: the last partition's end bound is now always len(addr_list), so entries beyond the registered offsets are folded into the last partition instead of being silently dropped. Covered by the new unit test 	est_long_addr_list_surplus_goes_to_last_partition (8/8 UT pass).

lsy0214:
Closing this PR after re-verification.

**Verdict: the non-layerwise PP path natively supports DeepSeek-V4 after all - the patch turned out to be unnecessary.**

What we verified (8x Ascend 910B3, DeepSeek-V4-Flash-w8a8-mtp, PP=2 x TP=4, kv_both + memcache host_shm):

1. Re-ran the end-to-end benchmark on **pure upstream code** (the four files this PR touches restored to pristine `e6a133f`): 50/50 requests succeeded, 79.84 tok/s, zero KV transfer errors - byte-for-byte identical to the run with this patch applied (79.81 tok/s).
2. Root cause of my initial misdiagnosis: `decode_adaptor_prefill_pp` is only invoked in the `kv_consumer + consumer_is_to_put` (decode-side re-put) configuration - it is dead code in the standard kv_both / producer-put PD paths, which rely on the native pool-key mechanism (`@pp_rank` embedded per stage, lookup expands across ranks). The uniform-split imprecision this PR fixed therefore never manifests in standard deployments.

Thanks to the maintainer feedback for prompting the re-check. The two layerwise PRs (#15442, #15507) have been rebased to no longer include the commits from this PR.

PR #15291, opened 2026-08-29
[Attention][BugFix] Restore use_gva_layerwise dropped by #14465

lizy124:
### What this PR does / why we need it?
When #14465 simplified AscendStoreConnector.__init__, it accidentally dropped self.use_gva_layerwise, which set_external_slot_release_waiter() still references → MultiConnector (PD disaggregation) crashes at init. This PR restores it with the same semantics as KVPoolWorker: use_gva_layerwise = use_layerwise and backend == 'memcache'.

### Does this PR introduce _any_ user-facing change?
No.

### How was this patch tested?
PD disaggregation (MultiConnector, proxy 9004): 400/400 requests succeed; external hit rate 89.95% (prefill) / 99.99% (decode) on GSM8K.

- vLLM main: https://github.com/vllm-project/vllm/commit/ba07e4a48fc951300d97eb506217dd530583dea3

PR #15367, opened 2026-08-31
[Refactor][AscendStore] Converge layerwise GVA protocol and gating into backend

lizy124:

Part 1 of the layerwise GVA refactor (series tracked in #15277). The layerwise transfer path is generic (mooncake layerwise will reuse it); the only backend-specific pieces are the key format and the opt-in gate. This PR moves those into the memcache backend — currently the only backend opting into layerwise transfer — so generic code (worker, scheduler, transfer threads) no longer mentions GVA or MemcacheBackend. Behavior is unchanged.

- `backend/__init__.py`: registry resolves the layerwise protocol via the backend's `path`; a boolean marker opts a backend in (pinned by a cross-backend agreement test)
- `backend/memcache_backend.py`: key construction (`make_full_key` / `make_partial_key` / `make_hit_check_keys`) and `extract_layout_config` behind a module-level protocol; key strings byte-identical (snapshot tests)
- `pool_worker.py` / `pool_scheduler.py`: gate renamed to `use_layerwise_transfer`, derived from `get_layerwise_protocol(...) is not None`
- `kv_transfer.py`: layerwise transfer threads made backend-agnostic (assert/import dropped)
- `set_external_slot_release_waiter` is now a pure forwarder; the gate lives in `KVPoolWorker`, which stores the waiter before handing it to the backend on `on_worker_ready` (supersedes the connector-side flag from #15291)

No. The gate is derived from the same backend registry and the key format is byte-identical; existing memcache layerwise and non-layerwise (e.g. mooncake) deployments behave exactly as before.

- UT: `tests/ut/distributed/ascend_store` — 314 passed (registry truth table + exclusivity interlock, lazy protocol resolution, key snapshots, waiter gate forwarding/handover)
- CPU CI: ruff + mypy (green on the rebased head `63be9e03b`)
- Server smoke: MultiConnector PD (4xTP + D consumer), memcache layerwise long-prefix, mooncake non-layerwise — all three re-verified on the rebased head `63be9e03b` after the review rework; also validated earlier on the pre-rework head `2a239d18a` (two rounds, same results)


lizy124:
Thanks for the review! Addressing the two substantive points:

**1. Normalize `backend_name` in `backend_supports` — adopted** (16bdb52fd). The function now strips and lowercases internally, so callers may pass the raw config string and future call sites cannot silently miss the gate on case variants (the layout call site does not consult `backend_map`, so a missed `.lower()` there would fail silently — exactly the #14465 failure mode). The truth-table case `(True, "MEMCACHE")` now expects `True`, plus a `" Memcache "` case and a new `test_backend_supports_normalizes_name`. All three existing call sites already lower the name, so production behavior is unchanged.

**2. Replace `assert` with `if not isinstance(...): raise TypeError` — not adopted**, for three reasons:
- `assert` is the established invariant idiom in this module: 74 existing `assert` statements across ascend_store (including untouched code next to these threads). Switching only these two would create local style inconsistency.
- Under `python -O` the base-class stubs still raise `NotImplementedError` on the first GVA store call, so the failure is delayed, never silent — no incorrect behavior can result.
- The check's mypy-narrowing purpose (typing `m_store` to `MemcacheBackend` for the GVA store calls in the method body) works identically with `assert`.

Note the review was generated against an earlier iteration that has since been rewritten (the suggested code references a `GVALayerwiseCapable` ABC that no longer exists); re-running the review on the current head.

lizy124:
Follow-up on the review reply above: since GVA is a memcache-exclusive protocol, the gate was reshaped from a capability registry into a bare `use_gva_layerwise()` living in `backend/gva_protocol.py` (the memcache protocol module) — the adopted name normalization is kept, and the exclusivity is now asserted directly by a test (only `MemcacheBackend` overrides the five GVA store methods). `backend/__init__.py` keeps only `backend_map`.

lizy124:
The single failed check (`a3-16 card-(part 1-1)`) looks like an environment flake, not related to this PR:

- `test_k3_mla_pd_tp8` uses `MooncakeConnectorV1` (upstream vLLM connector); this PR only touches `ascend_store`, which that test never imports — it was selected by the test-mapping, not by causality.
- The failure is a 600s server-health readiness timeout (`RuntimeError: Timeout: these nodes did not become ready: []`), not an assertion failure. The empty node list is a reporting artifact in `tests/e2e/conftest.py::_wait_for_multiple_servers`: the `ready` dict is keyed by host, and the P/D servers share one host, so the prefill server's 200 marks the host ready while the decode server's connection failures keep the loop alive — the timeout then reports `[]` instead of the decode server.
- The same job passed on this PR's earlier head `735065fe1` (same suite, green), and every other check on this head (ruff, mypy, full ascend_store UT, other e2e jobs) is green.

Could a maintainer re-run the failed job, or shall I push an empty commit to retrigger? Happy to do either.

lizy124:
Correction to the comment above (mechanism unchanged, one detail reversed): per the captured log, the server that repeatedly reported ready was the decode one (port 41187); the **prefill** server (port 50971) never became healthy within 600s — its process stayed alive (no `exited unexpectedly`), it just never finished startup, consistent with model-load/runner contention rather than a code fault.

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/backend/gva_protocol.py
```diff
@@ -0,0 +1,97 @@
...
+    Called by the generic layout layer through the backend registry; the
+    protocol itself owns the opt-in check so the layout layer never spells
+    out the gate.
+    """
+    if extra_config.get("use_layerwise", False):
+        return extra_config
+    return None
+
+
+class GVAKeyFactory:
```
Pz1116: Can we move this part to memcache backend as well instead a adding a new file under backend folder?

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/ascend_store_connector.py
```diff
@@ -198,10 +196,18 @@ def take_events(self) -> Iterable["KVCacheEvent"]:
     # Worker Side Methods
     ############################################################
     def set_external_slot_release_waiter(self, waiter: Callable[[int], None]) -> bool:
-        if not self.use_gva_layerwise or getattr(self, "connector_worker", None) is None:
+        """Pure forwarder: the GVA-layerwise gate is evaluated by the worker.
```
Pz1116: we should remove the comments here since gva is backend-specific

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/kv_transfer.py
```diff
@@ -1329,6 +1332,11 @@ def build_shared_data(self, task: LayerTransferTask) -> SharedBlockData | None:
     def _handle_request(  # type: ignore[override]
         self, transfer_tasks: list[LayerTransferTask]
     ):
+        # Layerwise threads only run when the use_gva_layerwise gate is on,
+        # i.e. the memcache backend; the assert fails fast (at thread entry)
+        # if that gate invariant is ever broken, instead of surfacing as a
+        # NotImplementedError from the base-class stubs on the first call.
+        assert isinstance(self.m_store, MemcacheBackend)
```
Pz1116: we should not assert and check memcache backend here, mooncake layerwise will also use a similar path

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/kv_transfer.py
```diff
@@ -1481,6 +1489,11 @@ def _stagger_h2d_submit(self, layer_id: int) -> None:
     def _handle_request(  # type: ignore[override]
         self, data: LayerLoadTask
     ):
+        # Layerwise threads only run when the use_gva_layerwise gate is on,
+        # i.e. the memcache backend; the assert fails fast (at thread entry)
+        # if that gate invariant is ever broken, instead of surfacing as a
+        # NotImplementedError from the base-class stubs on the first call.
+        assert isinstance(self.m_store, MemcacheBackend)
```
Pz1116: same here

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_scheduler.py
```diff
@@ -158,9 +159,15 @@ def __init__(
...
         self.tp_mismatch = tp_mismatch_info.enabled
 
-        backend_name = vllm_config.kv_transfer_config.kv_connector_extra_config.get("backend", "mooncake")
+        backend_name = str(vllm_config.kv_transfer_config.kv_connector_extra_config.get("backend", "mooncake"))
         self.backend_name = backend_name.lower()
-        self.use_gva_layerwise = self.use_layerwise and self.backend_name == "memcache"
+        # Resolve the backend's layerwise protocol (if any) once through the
+        # registry; generic code never imports the protocol module by name.
+        self.layerwise_protocol = get_layerwise_protocol(self.backend_name)
+        self.use_gva_layerwise = self.use_layerwise and self.layerwise_protocol is not None
```
Pz1116: use gva layerwise still exists here

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
```diff
@@ -154,7 +156,13 @@ def _init_kv_transfer_config(self, vllm_config, extra_config, use_layerwise, kv_
         self.consumer_is_to_put = extra_config.get("consumer_is_to_put", False)
         self.backend = extra_config.get("backend", "mooncake")
         self.backend_name = self.backend.lower()
-        self.use_gva_layerwise = self.use_layerwise and self.backend_name == "memcache"
+        # Resolve the backend's layerwise protocol (if any) once through the
+        # registry; generic code never imports the protocol module by name.
+        self.layerwise_protocol = get_layerwise_protocol(self.backend_name)
+        self.use_gva_layerwise = use_layerwise and self.layerwise_protocol is not None
+        self._layerwise_key_factory = (
```
Pz1116: self.layerwise_protocol.GVAKeyFactory is also backend specific

Pz1116 (approved):

lizy124:
@Pz1116 以上意见已按讨论方向全部处理，连同 gemini-code-assist 的 2 条，共 4 个 commit（`f057d367c..359876493`，fast-forward）：

1. `GVAKeyFactory` / `extract_layout_config` 已迁入 memcache backend，改为模块级中性函数（`make_full_key` / `make_partial_key` / `make_hit_check_keys` / `extract_layout_config`），`gva_protocol.py` 已删除；key 字符串逐字节不变，快照测试迁入 `test_backend.py` 的 memcache 区块。
2. `kv_transfer.py` 两处 `assert isinstance(self.m_store, MemcacheBackend)` 已删，`MemcacheBackend` import 及相关注释一并清除——layerwise 线程保持后端无关，mooncake 后续可直接复用。gemini 那条"assert 改 TypeError"的建议随之消解（入口不再有类型断言）。
3. `use_gva_layerwise` 改名 `use_layerwise_transfer`，仍由 registry 派生（`get_layerwise_protocol(...) is not None`；registry 字段改为布尔标记，resolver 经 backend `path` 解析协议模块）。
4. `pool_worker.py` / `pool_scheduler.py` 不再出现 `GVA` 字样，key 构造全部走协议函数。
5. connector `set_external_slot_release_waiter` docstring 去 GVA 表述，保持纯转发；worker 侧 waiter 初始化顺序加固（构造期先存 waiter 引用，`on_worker_ready` 时移交 backend，新增交接测试 `test_worker_ready_waiter_handover`）。
6. 测试同步：layerwise 线程 fixture 去掉 `spec=MemcacheBackend` 的 mock；真值表并入 `TestLayerwiseProtocolRegistry`；新增 `TestLayerwiseProtocolMemcacheExclusivity` 锁定"协议函数 ⇔ registry 标记 ⇔ 类覆写"三方一致，并负向断言其余 backend 不携带协议。

验证：服务器全量 `tests/ut/distributed/ascend_store/` **314 passed**（较此前 313 +1，为新增 waiter 交接测试）；ruff 0.14.0 check + format 全绿。

weijinqian0 (approved):

weijinqian0 (approved):

PR #15395, opened 2026-08-31
[Doc][Attention] Correct misleading attention docstrings

jiangkuaixue123:

Correct misleading and stale attention class docstrings:

- Describe `DSAAttention` as the DeepSeek sparse attention layer that selects the C4, C128, or sliding-window backend, instead of using the copied generic MLA description.
- Correct `DSAModules`, which was incorrectly documented as an SFA V2 container.
- Replace stale `MLACommon` and nonexistent file-level-comment references with concise descriptions for the MLA, SFA, DSA, and DSA-CP metadata, builders, and implementations.

No runtime behavior is changed.

No. This is a documentation-only correction.

- Targeted manual-stage pre-commit hooks passed for all seven changed Python files, including Ruff, formatting, codespell, typos, and repository-specific checks.
- `python3 -m compileall` passed for all changed files.
- `git diff --check` passed.
- No unit tests were added because the patch changes docstrings only.

The full `bash format.sh ci` run also passed Ruff, formatting, codespell, typos, clang-format, markdownlint, Actionlint, and repository-specific checks. Its gitleaks and shellcheck hooks could not run on this macOS host because the downloaded gitleaks binary was for x86-64 Linux and shellcheck was not installed.

https://github.com/vllm-project/vllm/commit/ba07e4a48fc951300d97eb506217dd530583dea3

### DIFF
--- a/vllm_ascend/attention/context_parallel/dsa_cp.py
+++ b/vllm_ascend/attention/context_parallel/dsa_cp.py
@@ -122,10 +122,7 @@ class AscendDSAReqMetadata:
 
 @dataclass
 class AscendDSAMetadata:
-    """Metadata for MLACommon.
-    NOTE: Please read the comment at the top of the file before trying to
-    understand this class
-    """
+    """Runtime metadata for context-parallel DeepSeek sparse attention."""
 
     num_actual_tokens: int  # Number of tokens excluding padding.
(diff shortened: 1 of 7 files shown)


jiangkuaixue123:
cc @weijinqian0 @realliujiaxu @zzzzwwjj

PR #15398, opened 2026-08-31
[Perf][Ops] Fused mHC and AscendC causal_conv1d for GLM/KDA hybrid models on A2

Liears:
### What does this PR do?

Replaces the two biggest host-time sinks in the GLM-5 hybrid-attention (KDA + mHC) decode path with fused AscendC kernels, and fixes an import-order bug that silently kept the upstream decomposition active.

**1. Fused mHC (`vllm_ascend/ops/mhc_ascendc.py`)**
`mhc_pre_torch` costs ~340 eager dispatches per layer (117 of them from the 20-round Sinkhorn normalisation), which py-spy measured at 43% of worker host time on GLM-5.3-Flash. This adds wrappers for the existing `npu_hc_pre_v2` / `npu_hc_post` AscendC kernels (mix projection + RMS-normalised gates + 20-round Sinkhorn + stream contraction in ONE launch; 4 launches per layer instead of ~340). The upstream `MHCPreOp`/`MHCFusedPostPreOp` return contract is preserved and the layer input RMSNorm is fused in. Automatic fallback to the torch path when the operator is unavailable or shapes are outside the envelope (hc_mult<=4, d in {4096, 7168}, bf16 x / fp32 hc params).

**2. AscendC causal_conv1d_update with small-pool staging**
The GLM/KDA decode path hands `causal_conv1d_update` a non-contiguous DS-layout view of the aliasing mamba pool. The previous wrapper transposed and copied the whole pool per call and wrote it back with a full-pool `copy_` through the strided view — measured **3.7ms per call** (a per-row write costs 0.02ms). With 34 conv layers this dominated the step. The new wrapper stages only the rows selected by `cache_indices` into a small contiguous SD pool, runs `npu_causal_conv1d_custom` on it, and scatters the few updated rows back (per-row slice assignment in eager; tensorized `index_copy_` under ACL-graph capture). Weight/bias are cast to the cache dtype as required by the aclnn kernel.

**3. KDA from-import sweep**
`patch_triton.py` imported the GLM model module (for the conv rebinding) *before* rebinding `fla_kda.fused_recurrent_kda`, so the model's `from fla_kda import fused_recurrent_kda` kept the upstream object — the NPU implementation was silently never called. The sweep rewrites stale aliases in every loaded vllm/vllm_ascend module (identity check, existing attrs only).

**4. MTP uniform-decode guard (aligns with the GLM-Next A3 branch fix)**
Tightens the uniform-decode guard from `num_computed > 0` to `num_computed >= num_prompt` and removes the spec-sized prefill chunk folding, which misdispatched graph metadata and garbled speculative output (same fix as the A3 branch's garbled-replay bug).

### Performance

GLM-5.3-Flash W8A8, 8x Atlas A2 (910B3), TP8, eager, single stream:

| config | TPOT |
|---|---|
| before | 613.9 ms |
| after | **282.3 ms** (2.2x) |

py-spy: mHC host share 43% -> <4%. Output parity verified on math/code/Chinese prompts; unit tests cover numeric parity vs the torch reference (29 checks).

No API change. Two env switches gate the new paths (`GLM53_HC_ASCENDC`, `GLM53_ASCENDC_CONV`, both default to the fused path with automatic fallback).


Liears:
Closing in favor of single-topic PRs (easier to review/land independently):

- #15429 — fused mHC via AscendC npu_hc_pre_v2/npu_hc_post (ops + tests)
- #15430 — AscendC causal_conv1d_update with small-pool staging
- #15431 — bind fused ops + from-import sweep + recursion fix
- Liears/vllm-ascend#3 — spec conv slot semantics fix (stacked on #15430, will retarget to upstream when base merges)

All commits here are contained in the split PRs; nothing is lost. The original 25-commit bundle also carried the GLM bring-up fixes already tracked by #15284-#15288 (pending #15127).

PR #15409, opened 2026-08-31
[BugFix][Worker] Invalidate dummy slots before attention metadata build

zhenwenqi2024:

This PR fixes stale KV slot metadata in `NPUModelRunner._dummy_run()`.

A DP rank with no real requests can still execute a dummy batch for DP alignment. The previous order invalidated the generic `slot_mapping` only after `_build_attention_metadata()`. Attention backends such as DSA may derive or copy backend-specific slot metadata during that build, so the later `fill_(-1)` cannot invalidate the already-built metadata. Dummy graph execution can then write through slots left by a completed request into logical KV blocks that have already been released and may be assigned to a new request.

The fix moves the existing per-KV-group `slot_mapping.gpu.fill_(-1)` before attention metadata construction. It adds no new synchronization, NPU operator, allocation, or mathematical operation.

A production one-shot probe captured the causal transition inside one `_dummy_run` on DP2 in FULL graph mode:

- stale source slots before metadata build: `1498..1505`;
- corresponding persistent SWA KV rows: block 11, offsets 90..97;
- before dummy forward: 4,096/4,096 finite BF16 values and 0 NaNs;
- after the same dummy forward: 0 finite values and 4,096 NaNs;
- the generic slot buffer was already `-1` after the late clear, proving that the clear happened after backend metadata had consumed the stale slots.

The previous `record_stream` changes and tests have been removed from the final diff because the captured corruption is logical KV-block reuse inside a persistent KV cache tensor, not caching-allocator reuse of a temporary tensor.

AI assistance was used during investigation, instrumentation, branch adaptation, test development, and PR preparation. Every changed line, the captured evidence, and the final causal explanation were manually reviewed by the submitter.

No API or configuration change. It prevents dummy execution from corrupting released logical KV blocks.

- Added `TestDummyRunSlotInvalidation::test_backend_metadata_sees_invalidated_dummy_slots` in `tests/ut/worker/a2/test_model_runner_v1.py`. It initializes two KV cache groups with stale slots and asserts that both mappings are `-1` when `_build_attention_metadata()` is entered.
- `python -m py_compile vllm_ascend/worker/model_runner_v1.py tests/ut/worker/a2/test_model_runner_v1.py` — passed.
- `git diff --check` — passed.
- Focused pytest collection in the local Windows checkout is blocked because a matching vLLM environment and `cbor2` are not installed; branch CI remains required.
- On the original v0.25.1 Ascend PD-separated workload, the minimal ordering fix passed 70 rounds / 17,920 requests with 0 request failures and 0 NaN signals.
- The additional stress run completed 100/100 rounds and 25,600/25,600 successful requests with 0 failures, 0 NaN signals, and 0 stop conditions. Combined minimal-fix validation is 170 rounds / 43,520 requests.


ningjingbengxiaohai (approved):

kunpengW-code:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #15416, opened 2026-08-31
[Performance][KDA] Compose and overlap gate projections on main

Dawn952:

This recreates #15168 without the unrelated `mla_v1.py` head-padding change and its corresponding test changes.

For the existing mixed-precision Kimi K3 KDA layout, this change:

- composes `f_proj.weight = f_b_proj.weight @ f_a_proj.weight` after checkpoint loading and later source-weight reloads;
- packs beta, the composed F projection, and the output gate into one floating-point BFG projection;
- runs a two-stage auxiliary-stream schedule:
  1. main-stream MXFP DynamicQuant overlaps the auxiliary BFG GEMM;
  2. main-stream QKV GEMM overlaps auxiliary B/F/G split, beta FP32 sigmoid, and gate reshaping;
- marks mixed-path beta as preprocessed so the v0.27 dispatch path only slices it, while ordinary upstream raw beta retains its existing FP32 sigmoid path;
- preserves the configured W8A8 MXFP8 `scale_alg` when DynamicQuant is split from the linear method;
- routes global or local F shards through the v0.27 packed loader;
- leaves the ordinary upstream `in_proj_qkvgfab` path unchanged.

Compared with #15168, this PR does not modify:

- `vllm_ascend/attention/mla_v1.py`;
- `tests/ut/attention/a2/test_mla_v1.py`.

### ACL graph and stream-order assessment

The auxiliary stream records `bfg_ready` as its final node, and the main stream waits for that exact tail event. No auxiliary operation is queued after the join point.

The four-event schedule is intentional:

1. `hidden_states_ready` forks BFG from main;
2. `bfg_projection_ready` serializes the BFG and QKV Cube matmuls;
3. `quant_ready` starts auxiliary beta vector work only after main DynamicQuant;
4. `bfg_ready` joins the complete auxiliary tail before KDA consumes its outputs.

The live-token slices for `mixed_qkv`, `g1`, and `g2` remain at the eager dispatch boundary. They are host-side basic views: after QKV GEMM and the auxiliary tail wait are enqueued, Python creates those views while the asynchronous device work is still running. Moving `num_actual_tokens` into the projection path would not add an NPU kernel overlap and would weaken the runtime metadata boundary.

Static assessment: the stream is fully joined for ACLGraph capture and mixed beta is transformed exactly once. A real NPU ACLGraph capture/replay has not been run for this main port.

No additional local or NPU validation was run for this replacement PR. The retained KDA implementation and focused tests are unchanged from #15168; this replacement only excludes the MLA implementation and test diffs.

No API or configuration change. Runtime behavior changes only for Kimi K3 KDA layers using the existing mixed-precision full-rank gate layout.


maoxx241 (approved):

linfeng-yuan (approved):

Dawn952:
[Bot]: cherry-pick completed successfully. New PR created: https://github.com/vllm-project/vllm-ascend/pull/15869

PR #15425, opened 2026-08-31
[BugFix][Quantization] Propagate auto-detected quantization to MTP draft

ZYang6263:

Quantization auto-detection runs from the Ascend platform hook after SpeculativeConfig has already created its MTP draft ModelConfig. For ModelSlim checkpoints started without an explicit --quantization ascend argument, the target config is updated to ascend but the same-checkpoint MTP draft remains unquantized. Quantized GLM-5 MTP weights then fail to load because the draft layers were constructed without their quantization parameters.

This change propagates the auto-detected method to an MTP draft only when it shares the target checkpoint and has no explicit quantization method. Independent or explicitly configured drafts remain unchanged. It also adds regression coverage for all three cases.

Yes. Quantized GLM-5-style same-checkpoint MTP models can rely on quantization auto-detection and no longer need --quantization ascend solely to load their MTP weights correctly.

- python -m pre_commit run ruff-check --files vllm_ascend/quantization/utils.py tests/ut/quantization/test_utils.py
- python -m pre_commit run ruff-format --files vllm_ascend/quantization/utils.py tests/ut/quantization/test_utils.py
- python -m pre_commit run codespell --files vllm_ascend/quantization/utils.py tests/ut/quantization/test_utils.py
- python -m compileall -q vllm_ascend/quantization/utils.py tests/ut/quantization/test_utils.py
- The focused pytest suite was not run locally because this Windows environment does not provide torch or torch_npu; the added unit tests are included for CI execution.

### DIFF
--- a/tests/ut/quantization/test_utils.py
+++ b/tests/ut/quantization/test_utils.py
@@ -199,6 +199,51 @@ class TestMaybeAutoDetectQuantization(TestBase):
         self.assertIn(ASCEND_QUANTIZATION_METHOD, call_args)
         self.assertIn("/fake/quant_model", call_args)
 
+    @patch("vllm.config.VllmConfig._get_quantization_config", return_value=MagicMock())
+    @patch("vllm_ascend.quantization.utils.detect_quantization_method", return_value=ASCEND_QUANTIZATION_METHOD)
+    def test_auto_detect_sets_quantization_on_same_checkpoint_mtp(self, mock_detect, mock_get_quant_config):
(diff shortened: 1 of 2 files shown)


tests/ut/quantization/test_utils.py
kunpengW-code, line 1: In addition to the UT, you can also modify the startup command of the GLM5 E2E test case or nightly test case to delete '--quantization ascend'. In this way, you can monitor this function in an end-to-end manner.

PR #15429, opened 2026-08-31
[Perf][Ops] Fused mHC via AscendC npu_hc_pre_v2/npu_hc_post

Liears:
### What does this PR do?

Replaces ~340 eager dispatches per layer of `mhc_pre_torch`/`mhc_post_torch` (117 of them from the 20-round Sinkhorn normalisation) with two fused AscendC kernels:

- `npu_hc_pre_v2`: mix projection + RMS-normalised gates + 20-round Sinkhorn + stream contraction in ONE launch
- `npu_hc_post`: comb @ residual + post ⊙ x

4 launches per layer instead of ~340. py-spy measured `mhc_pre_torch` at 43% of worker host time on GLM-5.3-Flash.

Keeps the upstream `MHCPreOp`/`MHCFusedPostPreOp` return contract, fuses the layer input RMSNorm, and falls back to the torch path when the operator is unavailable or shapes are outside the envelope (hc_mult<=4, d in {4096,7168}, bf16 x / fp32 hc params). Env switch `GLM53_HC_ASCENDC` (default on).

New file only (ops module + unit tests); the patch wiring is a separate PR.

### Performance

GLM-5.3-Flash W8A8, 8x Atlas A2, TP8, eager: mHC host share 43% → <4%. Unit tests cover numeric parity vs the torch reference (29 checks, within 0.5 bf16 ulp).

No API change; opt-out env switch preserved.

### DIFF
--- a/tests/ut/ops/test_mhc_ascendc.py
+++ b/tests/ut/ops/test_mhc_ascendc.py
@@ -15,11 +15,16 @@ Run on one card:  ASCEND_RT_VISIBLE_DEVICES=0 python3 test_mhc_ascendc.py
 
 from __future__ import annotations
 
-import json
 import os
 import sys
 
-sys.path.insert(0, "/vllm-workspace/vllm-ascend")
+# Run from any checkout location; falls through to the installed package
+# when the repo layout is absent.
+_REPO_ROOT = os.path.dirname(
+    os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+)
+if os.path.isdir(os.path.join(_REPO_ROOT, "vllm_ascend")) and _REPO_ROOT not in sys.path:
+    sys.path.insert(0, _REPO_ROOT)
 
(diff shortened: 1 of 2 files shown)


Liears:
Addressed the review findings in 17546395d:

- **Test path hardcoding** (tests/ut/ops/test_mhc_ascendc.py): the config constants are now inlined (verified equal to the shipped GLM-5.3-Flash config.json) and the repo root is derived from `__file__`, so the test is self-contained and CI-safe.
- **`causal_conv1d_update_npu_DISABLED` naming / import mismatch**: no longer present in the current tree — the branch was rebased past #14620 (which removed `causal_conv1d_update_npu`), and no references remain anywhere in `vllm_ascend/` or `tests/`.
- **`.item()` host syncs in kda.py**: that spec-decode fallback decomposition was superseded by the AscendC `recurrent_kda` path; no `.item()` calls remain in `vllm_ascend/ops/triton/kda/kda.py`.

PR #15431, opened 2026-08-31
[Fix][Patch] Bind fused KDA ops and sweep stale from-imports

Liears:
### What does this PR do?

1. Binds the fused mHC / AscendC conv / AscendC KDA recurrent implementations into the runtime patch layer (each behind its env switch with torch fallback preserved).

2. **From-import sweep**: `patch_triton.py` imported the GLM model module (for the conv rebinding) BEFORE rebinding `fla_kda.fused_recurrent_kda`, so the model's `from fla_kda import fused_recurrent_kda` kept the upstream object — the NPU implementation was silently never called. The sweep rewrites stale aliases in every loaded vllm/vllm_ascend module (identity check, existing attrs only).

3. **Recursion fix**: after the sweep, the AscendC wrapper's fallback call resolved back to itself (infinite recursion, RecursionError swallowed as silent decode corruption — measured 0.000 acceptance across 63 samples). Fallbacks now go through a frozen module-level reference snapshotted before any patching.

### Validation

- Before sweep: `[kda-ascendc] active` never appeared in logs; after: appears on all 8 workers
- GLM-5.3-Flash, 8x Atlas A2, TP8: with PRs #15429 + conv-staging applied, TPOT 613.9ms → 282.3ms with correct output

No.


Liears:
Closing — this change depends on the GLM bring-up chain (#15127 + #15284-15288) which is not on main yet; a PR against main cannot express that stack. Will reopen as a stacked PR once the base lands. The mHC (#15429) and conv-staging (#15430) PRs stand alone and stay open.

PR #15441, opened 2026-09-01
[Feature][Quant] Introduce combine_quant_mode to force MC2 combine stage communication quantization

lcfenglinwan:
### What does this PR do / why we need it?

  Introduces a new top-level Ascend config option `combine_quant_mode` (default `0`, disabled) that lets a model force the MC2 **combine** stage's communication quantization mode regardless of its `quant_type`.

  By default the combine `comm_quant_mode` passed to`torch.moe_distribute_combine_v2` is derived from the model's quant type (e.g. `W8A8MXFP` → `4`). When a user sets `combine_quant_mode` to a non-zero value, the combine comm quant mode is forced to that value instead, so non-MXFP models can opt into a specific combine communication quant mode (e.g. `4` = MXFP communication quantization) without changing the model weights' quantization.

  Precedence in `get_combine_mc_kwargs`:
  1. explicit `comm_quant_mode` set by a deeper per-layer quant config still wins, so the switch never silently overrides a deliberate quant decision;
  2. `additional_config.combine_quant_mode` when non-zero;
  3. otherwise the existing `quant_type`-based default.

  Yes — a new optional `additional_config` key `combine_quant_mode` (int). It is documented in the additional-configuration table. Default behavior is unchanged (`0` = use the model's own quant).

  - New unit tests:
    - `tests/ut/test_ascend_config.py`: default is `0`, int-string `"4"` is lax-coerced to `4`, non-integer values are rejected.
    - `tests/ut/ops/a2/test_token_dispatcher.py`: non-zero `combine_quant_mode` forces the combine comm quant mode for a non-MXFP model; `0` keeps the default; an explicit `comm_quant_mode` still takes precedence.


docs/source/user_guide/configuration/additional_config.md
```diff
@@ -79,6 +79,7 @@ The following table lists additional configuration options available in vLLM Asc
...
 | `enable_reduce_sample`              | bool | `False` | Whether to enable reduce sample optimization to reduce communication and computation overheads in the tensor parallelism scenario. When enabled, logits are kept partitioned across TP ranks and only the small set of top-k candidate values/indices is communicated, instead of performing a full-vocabulary all-to-all/all-gather. **Note**: This is an experimental feature. **Limitations**: (1) Not supported on PD-disaggregated scenario. (2) Must be disabled when sampling logprobs are requested. When reduce sample is enabled, logprobs are silently computed over partitioned logits instead of the full vocabulary, producing incorrect logprob values and top-k rankings. (3) Cannot be enabled together with lmhead TP.|
+| `combine_quant_mode`                | int  | `0`     | Fused MC2 configuration. This configuration will be passed as the `comm_quant_mode` argument for the `torch.moe_distribute_combine_v2` operator. Please refer to the operator documentation for the valid value range. |
```
kunpengW-code: The operator interface needs to be changed to the name of the operator that is actually called.

lcfenglinwan: fixed

kunpengW-code (dismissed):

lcfenglinwan:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

kunpengW-code (approved):

Issue #15280, opened 2026-08-29
[RFC]: Asynchronous Device Metadata Construction for vLLM Ascend


freyfwt:
### Motivation.

DeepSeek V4 inference constructs several kinds of device-side metadata before or during each forward pass, including compressor, Quant Lightning Indexer (QLI), Sparse Attention Selector (SAS), DSpark sliding-window attention, and DSA context-parallel metadata.

The current metadata-builder path mixes three different responsibilities:

1. Host-side scheduling and control-flow decisions.
2. Binding tensor views and scalar attributes.
3. Launching device operators that construct metadata tensors.

This causes two classes of performance overhead:

- The same metadata may be reconstructed by multiple layers even when its semantic inputs are identical. [PR #14994](https://github.com/vllm-project/vllm-ascend/pull/14994) and its main-branch counterpart [PR #14932](https://github.com/vllm-project/vllm-ascend/pull/14932) demonstrate that compressor metadata can be reused across layers when cache group, execution phase, and speculative substep are correctly isolated.
- Device scalar reads or metadata operators executed on the model stream can block host enqueue-ahead and serialize metadata construction with model execution. For example, [PR #15042](https://github.com/vllm-project/vllm-ascend/pull/15042) reports that 13 QLI metadata builds caused 26 host-device synchronizations per sampled rank because of device-side `max().item()` reads.

Removing redundant work is necessary but does not address the remaining serialization. This RFC proposes a small framework-level mechanism in vLLM Ascend that can asynchronously construct eligible device metadata on a dedicated NPU stream and synchronize only at the actual consumption points.

The expected benefit is bounded by:

```text
net gain =
    min(metadata device time, independent model execution window)
    - event overhead
    - device resource contention
    - host enqueue delay
```

Therefore, moving metadata to another stream is not assumed to be beneficial by definition. The change will only be merged if it demonstrates a statistically significant end-to-end TPOT improvement.

#### Goals

This RFC proposes an internal, model-extensible execution mechanism with DeepSeek V4 as the first implementation.

The initial implementation will:

- Modify only vLLM Ascend.
- Support DeepSeek V4 on Model Runner V1.
- Construct eligible device metadata asynchronously on one dedicated metadata stream per worker.
- Preserve every graph mode and feature combination already supported by DeepSeek V4.
- Support arbitrary `num_speculative_tokens`; the design must not specialize for DSpark step count 5.
- Preserve the legacy synchronous path for ineligible phases or components.
[...]

PR #15443, opened 2026-09-01
[Performance][Worker] Build DeepSeek V4 metadata asynchronously

freyfwt:

Refs #15280.

DeepSeek V4 currently constructs compressor, Quant Lightning Indexer (QLI), Sparse Attention Selector (SAS), DSpark sliding-window, and local DSA-CP metadata on the model stream. These device operators serialize metadata construction with otherwise independent model work.

This PR adds a small worker-owned asynchronous execution mechanism and adopts it only for DeepSeek V4 on Model Runner V1:

- A structural `DeviceMetadataTaskProvider` capability opts eligible attention builders into one `DeviceMetadataExecutor` per worker. Models without this capability retain the existing path and allocate no metadata stream or events.
- The executor uses one dedicated NPU stream, stage/group-specific readiness events, and a reuse fence for persistent output buffers.
- Consumers wait only at their first compressor, indexer, or attention dependency. Repeated per-layer waits are deduplicated within a submission.
- Eligible QLI, SAS, compressor, DSpark SWA, and communication-free local DSA-CP metadata are submitted asynchronously. Existing synchronous behavior remains for prefill or topology combinations that are not eligible and for pure `FULL` graph compressor metadata whose active extent is not capture-stable.
- Compressor metadata uses an explicit-output operator so graph-visible buffer addresses remain stable. Partial submission failures fail closed instead of retrying into partially modified buffers.

The implementation is independent of `num_speculative_tokens`, introduces no user-facing flag, and follows the effective graph mode after existing compatibility fallback.

No API or configuration change. Eligible DeepSeek V4 Model Runner V1 executions use the asynchronous path automatically. Other models and Model Runner V2 continue to use the existing implementation.

### Performance

A closed-loop A/B run used DeepSeek V4 W8A8C8 on Ascend A3 with DP4/TP4/EP16, DSA-CP, DSpark with five speculative tokens, async scheduling, concurrency 288, 12K input tokens, fixed 5K output tokens, and a 90% prefix-cache hit ratio. Both variants completed 1,728 requests with zero failures.

| Metric | Baseline | Async metadata | Change |
| --- | ---: | ---: | ---: |
| Mean TPOT | 33.111 ms | 30.917 ms | -6.63% |
| p99 TPOT | 41.419 ms | 39.489 ms | -4.66% |
| Output throughput | 7,794.29 tok/s | 8,335.89 tok/s | +6.95% |
| Total-token throughput | 26,500.61 tok/s | 28,342.03 tok/s | +6.95% |
| Mean ITL | 196.645 ms | 183.514 ms | -6.68% |
| Mean TTFT | 9,136.30 ms | 8,953.28 ms | -2.00% |
| Benchmark duration | 1,108.50 s | 1,036.48 s | -6.50% |
| Speculative acceptance | 98.846% | 98.776% | -0.069 pp |

The performance run was collected on implementation commit `895a0b189`. Later commits rebased the feature onto newer `main` revisions and refreshed compatibility fixtures; no performance result from the rebased HEAD is implied.

- `bash format.sh ci`
- 153 focused tests on the exact rebased source covering executor lifecycle, provider gating, DSA/DSA-CP staging, compressor reuse, DSpark metadata, failure handling, graph fallbacks, and current PCP integration.
- Full mypy runs for Python 3.10, 3.11, and 3.12: zero issues in `vllm_ascend`, `examples`, and `tests` for each version.
- Real NPU validation of all 20 compressor-metadata cases, including the fixed-output operator, cross-stream event ordering, stable output addresses, buffer reuse, and buffer rewriting.

### DIFF
--- a/.github/workflows/configs/nightly_config.yaml
+++ b/.github/workflows/configs/nightly_config.yaml
@@ -250,6 +250,9 @@ a3:
   multi_card:
     test_config:
       # pytest-driven tests
+      - name: compressor-metadata-cross-stream
+        os: linux-aarch64-a3-2
+        tests: tests/e2e/nightly/single_node/ops/singlecard_ops/test_compressor_metadata.py
       - name: qwen3-30b-acc
         os: linux-aarch64-nightly-a3-4
         tests: tests/e2e/weekly/single_node/models/test_qwen3_30b_acc.py
--- a/tests/e2e/nightly/single_node/ops/singlecard_ops/test_compressor_metadata.py
+++ b/tests/e2e/nightly/single_node/ops/singlecard_ops/test_compressor_metadata.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from dataclasses import dataclass
+from functools import partial
 from typing import Any
 
 import pytest
@@ -11,6 +12,12 @@ from vllm_ascend.utils import bootstrap_custom_op_env
 
 bootstrap_custom_op_env(include_vendor_lib=True)
 import vllm_ascend.vllm_ascend_C  # type: ignore[import-untyped] # noqa: E402,F401
+
+from vllm_ascend.worker.device_metadata import (  # noqa: E402
+    DeviceMetadataExecutor,
+    DeviceMetadataStage,
+    DeviceMetadataTask,
+)
 
 KV_BLOCK_SIZE = 128
 SLOT_MAPPING_FLAT = 1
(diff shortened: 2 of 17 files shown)


maoxx241 (approved):

yiz-liu (dismissed):

yiz-liu (commented):
1. Please see if we can remove some test cases?
2. Building compressor metadata needs AI Vector core, so we'd better overlap it with an Cube op like MatMul.

freyfwt:
> 1. Please see if we can remove some test cases?
> 2. Building compressor metadata needs AI Vector core, so we'd better overlap it with an Cube op like MatMul.

1. I removed some test cases. If I removed any more, the test coverage would decrease.
2. I reviewed the profiling results again. The compressor metadata building operator always finishes before `model_execute` begins.

yiz-liu (approved):

Issue #15204, opened 2026-08-28
[RFC]: DeepSeek-V4 compressor partial-state memory architecture


ShwStone:
## Motivation.

A DeepSeek-V4-Flash workload was observed to reject long prefill chunks before any request could be scheduled.

The reproduced environment was:

- vLLM-Ascend: `3281a5fc44ec344ba304c9161a3959d8649471f4`
- vLLM: `d02df748bf9efd99022f1a062597dc3cb3808485` (`v0.26.0`)
- Model: DeepSeek-V4-Flash W8A8, 43 layers
- Hardware: Ascend 910C, 64 GiB HBM per rank
- Parallelism: DP2 × TP8
- Eager mode, prefix caching disabled, EPLB disabled

With:

- `max_num_batched_tokens=65536`
- `gpu_memory_utilization=0.8`
- prompt lengths of 32,850 / 51,976 / 63,778 tokens

the requests remained in `waiting=capacity` and never started. One admission check reported:

```text
need=11094 blocks
free=6958 blocks
KV cache usage=0%
```

The 0% metric is expected here because admission fails before blocks are allocated. With `max_num_batched_tokens=32768` and async scheduling disabled, prompts above 50K tokens showed the same class of failure.

Baseline logs and reproduction details are available here:

https://github.com/ShwStone/vllm-ascend/pull/1

Only the baseline failure in that PR is relevant to this RFC. Its accounting-only patch and one-token output tests do not establish a correct solution.

The current compressor state cache stores FP32 partial states indexed by raw token position. The Ascend kernel receives a raw-token state block table, uses `CONTINUOUS` cache mode, and calculates cache offsets from the raw sequence index:

- Compressor state definition:
  https://github.com/vllm-project/vllm/blob/d02df748bf9efd99022f1a062597dc3cb3808485/vllm/models/deepseek_v4/compressor.py#L152-L200
- Current Ascend state cache and compressor invocation:
  https://github.com/vllm-project/vllm-ascend/blob/cfea50dde154800657197740b3e6db53dbe69086/vllm_ascend/models/deepseek_v4/compressor.py#L44-L71
  https://github.com/vllm-project/vllm-ascend/blob/cfea50dde154800657197740b3e6db53dbe69086/vllm_ascend/models/deepseek_v4/compressor.py#L201-L230
- Kernel state indexing:
[...]

PR #15452, opened 2026-09-01
[BugFix][Model] Reuse DeepSeek-V4 compressor tail workspace

jiangkuaixue123:

DeepSeek-V4 compressor partial states are currently represented as token-indexed paged KV cache. For a large prefill chunk, the scheduler reserves state pages for the full in-flight chunk across every compressor layer, although only a small incomplete compression tail must survive across forwards.

This PR:

- adds `AscendCompressorTailSpec` and `CompressorTailManager` so scheduler-managed compressor state remains constant per request;
- reserves one C4 tail block and four C128 tail blocks per request on the A3 block layout;
- enables compressor `cache_mode=2` (`CYCLE`), mapping absolute token positions into a persistent multi-block ring;
- keeps transient current-chunk partial states in the existing CANN operator workspace;
- extends the arch32 and arch35 kernels and tiling validation for the multi-block C128 ring;
- keeps the existing `CONTINUOUS` path as a gated fallback.

For a 4096-token in-flight chunk, compressor-state reservations change from 512 to 1 C4 block and from 128 to 4 C128 blocks per request. This removes the `chunk size x compressor layers` contribution from compressor-state block occupancy.

#### Current limitations

The optimized path is opt-in and disabled by default. Stage 1 is enabled only when all of the following are true:

- the model is DeepSeek-V4;
- the device is Atlas A3;
- eager mode is enforced; ACL graph capture/replay is not supported yet;
- `multistream_dsv4_dsa_overlap` is disabled;
- prefix caching is disabled;
- KV transfer and prefill/decode disaggregation are disabled;
- speculative decoding is disabled;
- decode and prefill context parallel sizes are both 1.

Unsupported configurations are not silently changed. They log the incompatibility and fall back to the existing `cache_mode=1` continuous state-cache path.

Prefix-cache tail checkpoints, P/D tail transfer, multistream synchronization, ACL graph support, and A5 support are follow-up work.

Yes. It adds the opt-in `additional_config.enable_dsv4_shared_compressor_workspace` configuration. The default remains `false`, and existing behavior is unchanged unless explicitly enabled on a supported configuration.

- Added `tests/ut/core/test_compressor_tail_manager.py` covering fixed C4/C128 capacity, chunk-length independence, ring address mapping, manager registration, block-table sizing, kernel source invariants, and feature-gate fallback reasons.
- Added host-side CANN validation for ring width and physical block capacity.
- Rebased onto the current `releases/v0.26.0rc` head and passed `git diff --check`.
- The targeted UT was not run in the PR preparation environment because `pytest` is unavailable; CI validation is pending.
- Full A3 NPU accuracy, peak-memory, and performance validation is pending and should be completed before merge.

### DIFF
--- a/csrc/attention/compressor/README.md
+++ b/csrc/attention/compressor/README.md
@@ -70,6 +70,7 @@
 | coff | 可选属性 | 默认值1，支持1/2。当coff=1时，无需进行overlap数据重排。当coff=2时，需要进行overlap数据重排。  | INT32          | -         |
 | norm\_eps | 可选属性 | 表示RmsNorm计算的权重系数。默认值1e-6。 | FLOAT32          | -         |
 | rotary\_mode | 可选属性 | 表示Rop计算的模式。默认值1，支持1/2。rotary\_mode为1时，代表half模式。rotary\_mode为2时，代表interleave模式。 | INT32          | -         |
+| cache\_mode | 可选属性 | 状态缓存模式。1为按绝对token连续分页；2为persistent tail循环分页，当前chunk中间结果使用算子workspace。默认值1。 | INT32          | -         |
 | enabled\_grad | 可选属性 | 训练场景使用，表示是否参与反向更新。默认值false，支持false/true。**目前暂不支持输入true**。 | BOOL          | -         |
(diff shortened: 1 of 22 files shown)


ShwStone:
Refs #15204

PR #15473, opened 2026-09-01
[BugFix] MooncakeConnector: MTP draft layer expansion wastes ~1 GiB KV cache on prefill

### LINKED ISSUE #15467: [Bug] MooncakeConnector MTP draft layer expansion wastes ~1 GiB on prefill, forcing max_model_len down from 256K to 200K
## Summary

In MTP + PD separation, `pp_layer_indices` unconditionally expands the layer range by `num_draft_layers`, forcing prefill nodes to load draft model weights (~1 GiB). This reduces KV cache from 4.59 GiB → 3.28 GiB (−28.5%), capping `max_model_len` at ~200K instead of 256K.

MTP draft layers share KV cache with the main model and are computed locally on the decode side — they do not need to be transferred from prefill. The existing code already acknowledges this (line 549: `"all MTP layer use the same kv cache layer, so only need to transfer once"`), but `pp_layer_indices` does not differentiate MTP from other speculative methods.

## Environment

| Item | Value |
|---|---|
| vllm-ascend | v0.23.0 (commit `5cb98caaa`) and `main` (commit `1cfec6e41` as of 2026-09-01) |
| vllm | commit `0fc695f` |
| Hardware | 4 × Ascend 910C A2 (64 GiB HBM each) |
| Model | Intern-S2-Preview-397B, PD separation (DP2×TP8) |
| GPU memory utilization | 0.95 (prefill) / 0.93 (decode) |

## Root Cause

**File:** `vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py`

**`pp_layer_indices` (line ~835 on `main`, line 822 on v0.23.0):**

```python
def pp_layer_indices(layer_indices, prefill_pp_rank, group_spec):
    first_layer_index, end_layer_index = self.pp_layer_indices[prefill_pp_rank]
    if self.vllm_config.speculative_config is not None and prefill_pp_rank == self._prefill_pp_size - 1:
        end_layer_index += self.num_draft_layers  # ← unconditional!
    ...
```

`self.pp_layer_indices` covers only main model layers (0 to `num_layers - 1`). The unconditional `end_layer_index += self.num_draft_layers` extends the range to include MTP draft layers (index ≥ `num_layers`), even though MTP draft KV cache is computed locally on decode and does not need prefill transfer.

Because prefill's KV cache layer count must match decode's (otherwise Mooncake transfer raises `IndexError`), prefill is forced to load `--speculative-config` → draft model weights loaded → ~1 GiB wasted → KV cache shrinks.

## Steps to Reproduce

1. Deploy a model with MTP (`num_speculative_tokens ≥ 1`) in PD separation mode (DP2×TP8 or similar)
2. Set `--max-model-len 256000` on both prefill and decode
3. Observe prefill engine core error:

```
ValueError: To serve at least one request with the model's max seq len (256000),
(4.16 GiB KV cache is needed, which is larger than the available KV cache memory (3.28 GiB).
Based on the available memory, the estimated maximum model length is 200192.
```

## Measured Impact

| Scenario | KV Cache | GPU KV Cache Tokens | max_model_len |
|---|---|---|---|
| no-mtp (no speculative-config) | 4.59 GiB | 311,652 | 256K |
| spec3 without fix (with speculative-config) | 3.28 GiB | 200,000 | 200K |
| spec3 with fix (no speculative-config on prefill) | 4.59 GiB | 311,652 | 256K |

## Fix Direction

The fix is to add a `method` check in `pp_layer_indices`: only expand `end_layer_index` for non-MTP speculative methods (e.g. EAGLE) whose draft models have independent KV caches. For MTP, skip the expansion since draft layers share KV cache with the main model and are computed locally on the decode side. This is consistent with the existing code at line 549 which already treats MTP specially (`num_draft_layers = 1`).

iceflysnow:
## Summary

Fixes #15467: MTP draft layer expansion in `pp_layer_indices` wastes ~1 GiB KV cache on prefill nodes, forcing `max_model_len` down from 256K to 200K.

## Root Cause

In `mooncake_connector.py`, `pp_layer_indices` unconditionally expands the layer range by `num_draft_layers` when speculative config is present:

```python
if self.vllm_config.speculative_config is not None and prefill_pp_rank == self._prefill_pp_size - 1:
    end_layer_index += self.num_draft_layers  # unconditional!
```

MTP draft layers share KV cache with the main model and are computed locally on the decode side — they do **not** need to be transferred from prefill. The existing code already acknowledges this (line 549: `"all MTP layer use the same kv cache layer"`), but `pp_layer_indices` does not differentiate MTP from other speculative methods.

## Fix

Add a `method != "mtp"` guard so only non-MTP speculative methods (e.g. EAGLE) whose draft models have independent KV caches expand the layer range:

```python
if self.vllm_config.speculative_config is not None and prefill_pp_rank == self._prefill_pp_size - 1:
    if self.vllm_config.speculative_config.method != "mtp":
        end_layer_index += self.num_draft_layers
```

## Measured Impact

| Scenario | KV Cache | max_model_len |
|---|---|---|
| Without fix | 3.28 GiB | ~200K |
| With fix | 4.59 GiB | 256K |
| Improvement | +1.31 GiB (+40%) | +56K (+28%) |

## Why This Is Safe

| Concern | Analysis |
|---|---|
| EAGLE/EAGLE3 | `method != "mtp"` → expansion logic preserved → unaffected |
| Non-speculative | `speculative_config is None` → branch not entered |
| PP > 1 | Expansion only affects the last PP rank; all ranks remain consistent |
| MTP correctness | MTP draft KV cache is local to decode; no correctness impact |

## Tested

- 4-node PD separation deployment (DP2×TP8) with Intern-S2-Preview-397B
- 24-scenario benchmark (random dataset, 1K-32K input, 1K-1.5K output, 1-64 concurrency)
- All 24/24 scenarios pass with 256K max_model_len
- KV cache verified at 4.59 GiB on both prefill and decode nodes

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
@@ -42,6 +42,7 @@ from vllm.distributed.parallel_state import (
 )
 from vllm.distributed.utils import get_pp_indices
 from vllm.logger import logger
+from vllm.model_executor.layers.mamba.mamba_utils import is_conv_state_dim_first
 from vllm.utils.math_utils import cdiv
 from vllm.utils.network_utils import get_ip, make_zmq_path, make_zmq_socket
 from vllm.v1.core.sched.output import SchedulerOutput
@@ -49,25 +50,31 @@ from vllm.v1.kv_cache_interface import (
     FullAttentionSpec,
     KVCacheConfig,
     MambaSpec,
+    MLAAttentionSpec,
     SlidingWindowSpec,
     UniformTypeKVCacheSpecs,
 )
 from vllm.v1.request import RequestStatus
 
-from vllm_ascend import envs as ascend_envs
 from vllm_ascend.ascend_config import get_ascend_config, init_ascend_config
+from vllm_ascend.core.kv_cache_interface import AscendSFAIndexerCacheSpec, AscendSlidingWindowMLASpec
 from vllm_ascend.distributed.kv_transfer.utils.mooncake_transfer_engine import global_te
 from vllm_ascend.distributed.kv_transfer.utils.utils import (
     RegisterRegions,
     collect_storage_merged_register_regions,
     get_transfer_timeout_value,
+    tensor_storage_key,
     validate_register_region_count,
 )
 from vllm_ascend.distributed.utils import (
     get_decode_context_model_parallel_rank,
     get_decode_context_model_parallel_world_size,
 )
-from vllm_ascend.utils import enable_custom_op
+from vllm_ascend.utils import (
+    enable_custom_op,
+    enable_sfa_dcp_replicated_indexer,
+    model_uses_sfa_sparse,
+)
 
 # isort: off
 if TYPE_CHECKING:
@@ -85,6 +92,7 @@ DONE_RECVING_MSG = b"done_recving_msg"
 # number of peers is larger than max_workers. Yield after a small FIFO batch so
 # other peers already waiting in the global executor queue can make progress.
 MAX_REQUESTS_PER_PEER_HANDLER = 5
+KV_CACHE_BUFFER_ALIGNMENT = 2 * 1024 * 1024
 
 
 class RemotePortInfo(TypedDict):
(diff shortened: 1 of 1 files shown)


iceflysnow:
Closes #15467

PR #15477, opened 2026-09-01
[Bugfix] Add extra judgement about moe intermediate size

nanshen3000:
For some models like Minimax, it does not have moe_intermediate_size field. Instead, it only has intermediate_size in its config.json. This pr use intermediate_size if moe_intermediate_size is not found in config.json. In this way, it can support models without moe_intermediate_size field.

### DIFF
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -739,6 +739,8 @@ class AscendConfig:
 
         moe_intermediate_size = getattr(hf_text_config, "moe_intermediate_size", None)
         if moe_intermediate_size is None:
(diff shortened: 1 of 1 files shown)


CXY-Katrina:
/nightly Minimax_m2.7_w8a8_A3
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/33483554861)

czydyy:
/nightly Minimax_m2.7_w8a8_A3
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/33600800412)

PR #15488, opened 2026-09-01
[Performance] add swiglu_group_quant groupindex input

lcfenglinwan:
### What does this PR do / why we need it?

  In the W4A8 MXFP quantization path of `npu_grouped_matmul_swiglu_quant`
  (`vllm_ascend/device/device_op.py`), the call to
  `torch.ops._C_ascend.npu_swiglu_group_quant` now passes a `group_index`
  instead of `None`.

  `npu_swiglu_group_quant`'s `group_index` input currently only supports the
  `count` type, and internally the operator uses `group_index` only for
  summation. We therefore pass the last value of the cumsum tensor
  (`group_list[-1:]`, i.e. the total token count across groups) as the
  `group_index`, which provides the operator with the count it needs.

  No. 

  Tested with DSv4 Flash


Pz1116 (approved):
LGTM, @ZT-AIA please also take a look

ZT-AIA (dismissed):

linfeng-yuan (approved):

PR #15497, opened 2026-09-01
[v0.26.0rc][Bugfix] Replace scatter_nd_update_v2 with scatter_nd_update_sk

ZT-AIA:

`npu_scatter_nd_update_sk` is based on the latest `npu_scatter_nd_update_` version on CANN, and delivers better performance than `npu_scatter_nd_update_v2`. The default branch keeps the duplicate-index handling behavior, and a high-performance branch is also provided to further improve performance. It adapts to non-contiguous first-axis input; building on CANN, the split core count is rounded up to an even number to support superkernel.

### How was this patch tested

Validated against `torch_npu.npu_scatter_nd_update` and CPU reference with duplicate-index and non-contiguous `var` cases (`tests/e2e/nightly/single_node/ops/singlecard_ops/test_scatter_nd_update_sk.py`).


ZT-AIA:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #15121, opened 2026-08-27
[Feature]Add Forced Load‑Balancing Feature

shunqinlu:
### What this PR does / why we need it?

Add the forced load‑balancing feature for alignment with cann‑recipe, and remove the rand + argsort overhead within `model_forward`.

### Does this PR introduce _any_ user-facing change?

Yes, it introduces a new environment variable `VLLM_ASCEND_FORCE_EPLB_POLICY` which can be set to `cann_round_robin` to enable this deterministic load balancing policy.

### How was this patch tested?

- vLLM version: v0.27.1
- vLLM main: https://github.com/vllm-project/vllm/commit/ba07e4a48fc951300d97eb506217dd530583dea3

PR #15499, opened 2026-09-01
[Feature] add force EPLB Feature through additional config

weijinqian0:
## Motivation
   PR: #15121 

  The previous implementation had force-EPLB logic distributed across `NPUModelRunner` and `AscendRoutedExperts`:

  - Policy parsing was based on an environment variable.
  - The model runner traversed all model modules to find force-EPLB helpers.
  - Each routed-expert layer owned its own cache and helper methods.
  - Cache construction, policy checks, and expert-parameter derivation were
    duplicated across different components.

  This made the feature harder to configure and increased coupling between the model runner and individual MoE layers.

  Using `additional_config` provides a typed, validated configuration path that is included in the vLLM configuration lifecycle and can be represented together with the rest of the Ascend options.

  ## Changes

  ### `vllm_ascend/ascend_config.py`

  - Add the following configuration option:

    ```json
    {
      "enable_force_eplb": false
    }

  - enable_force_eplb=true enables the Round-Robin force-EPLB path.
  - Add validation to reject configurations that enable both enable_force_eplb and eplb_config.dynamic_eplb.

  Example:

  vllm serve <model> \
    --additional-config '{"enable_force_eplb": true}'

  ### vllm_ascend/ops/fused_moe/force_eplb.py

  Add a centralized implementation for force-EPLB table handling:

  - _build_or_get_topk() manages table construction and caching.
  - prebuild_force_eplb_topk() builds tables before ACLGraph capture.
  - apply_force_eplb_topk() obtains the deterministic expert IDs at runtime.
  - build_round_robin_topk() contains the deterministic Round-Robin mapping algorithm.

  The cache is associated with the active MoE communication method and uses the following values as part of its key:

  - Number of tokens
  - Top-k value
  - Number of logical experts
  - EP size
  - EP rank
  - Tensor dtype
  - Device

  This prevents tables with different shapes, ranks, dtypes, or devices from being incorrectly reused.

  The helper functions also preserve the previous behavior of returning without force-EPLB work when no MoE communication method is available.

  ### vllm_ascend/ops/fused_moe/routed_experts.py

  - Move the force-EPLB enable check to the routing call site.
  - Use apply_force_eplb_topk() when get_ascend_config().enable_force_eplb is enabled.

  - Preserve the existing random expert selection fallback when force EPLB is disabled and enable_force_load_balance is enabled.

  The normal routing path is therefore unchanged when enable_force_eplb=false.

  ### vllm_ascend/worker/model_runner_v1.py

  - Move force-EPLB prebuild invocation into the non-capturing dummy-run path.
  - Avoid performing force-EPLB table construction while is_graph_capturing=true.

  - Pass the current device and max_num_tokens to the centralized prebuild function.

  This keeps the first table construction outside the ACLGraph capture path.

  ### Tests

  Update the Ascend configuration unit test to verify that additional_config.enable_force_eplb=true is correctly parsed and exposed through AscendConfig.

  ## User-facing change

  Yes.

  Force EPLB should now be enabled through:

  {
    "enable_force_eplb": true
  }

 
  ## How was this patch tested?

  The following static checks passed:

  python -m py_compile \
    tests/ut/test_ascend_config.py \
    vllm_ascend/ascend_config.py \
    vllm_ascend/ops/fused_moe/force_eplb.py \
    vllm_ascend/ops/fused_moe/routed_experts.py \
    vllm_ascend/worker/model_runner_v1.py

  git diff --check

  NPU runtime and ACLGraph validation were not performed in the current environment because the required NPU runtime/test dependencies were unavailable.

Co-authored-by: [shunqinlu](https://github.com/shunqinlu)

### DIFF
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -257,6 +257,7 @@ class AscendConfig:
             "enable_mc2_hierarchy_comm": false,
             "enable_reduce_sample": false,
             "enable_dsa_cp": false,
+            "enable_force_eplb": false,
             "draft_window_size": null,
             "mix_placement": false,
             "pa_shape_list": [],
@@ -391,6 +392,7 @@ class AscendConfig:
     enable_mc2_hierarchy_comm: bool = False  # deprecated, will be replaced by mc2_comm_alg = "hierarchy"
     enable_reduce_sample: bool = False
     enable_dsa_cp: bool = False
+    enable_force_eplb: bool = False
     draft_window_size: int | None = None
     mix_placement: bool = False
     pa_shape_list: list[Any] = dataclasses.field(default_factory=list)
@@ -479,6 +481,8 @@ class AscendConfig:
     # the max_num_batched_tokens that sequence-parallel writeback corrected).
     def derive_and_validate(self, vllm_config: VllmConfig) -> AscendConfig:
         vc = vllm_config
+        if vc.model_config.is_moe and self.enable_force_eplb and self.eplb_config.dynamic_eplb:
+            raise ValueError("enable_force_eplb cannot be mixed with dynamic_eplb.")
         self._check_mooncake_c8_kv_cache_quant(vc)
 
         # profiling_chunk vs min_chunk clamp
--- a/vllm_ascend/ascend_config.py
+++ b/vllm_ascend/ascend_config.py
@@ -481,7 +481,7 @@ class AscendConfig:
     # the max_num_batched_tokens that sequence-parallel writeback corrected).
     def derive_and_validate(self, vllm_config: VllmConfig) -> AscendConfig:
         vc = vllm_config
-        if vc.model_config is not None and self.enable_force_eplb and self.eplb_config.dynamic_eplb:
+        if vc.model_config.is_moe and self.enable_force_eplb and self.eplb_config.dynamic_eplb:
             raise ValueError("enable_force_eplb cannot be mixed with dynamic_eplb.")
         self._check_mooncake_c8_kv_cache_quant(vc)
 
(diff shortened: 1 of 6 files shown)


vllm_ascend/ascend_config.py
```diff
@@ -472,6 +474,8 @@ def _validate_user_input_ranges(self):
     # the max_num_batched_tokens that sequence-parallel writeback corrected).
     def derive_and_validate(self, vllm_config: VllmConfig) -> AscendConfig:
         vc = vllm_config
+        if self.enable_force_eplb and self.eplb_config.dynamic_eplb:
+            raise ValueError("enable_force_eplb cannot be mixed with dynamic_eplb.")
```
ningjingbengxiaohai, line 490: 是不是还要约束下，moe模型才可以生效？

vllm_ascend/ops/fused_moe/routed_experts.py
```diff
@@ -479,10 +480,9 @@ def _select_experts(
         # fp32, which is what npu_moe_token_unpermute expects for its `probs` arg.
         if hidden_states.dtype not in [torch.uint8, torch.float8_e4m3fn]:
             topk_weights = topk_weights.to(hidden_states.dtype)
-        # This is a naive implementation for experts load balance so as to
-        # avoid accumulating too much tokens on a single rank. It is only
-        # activated when doing profile runs.
-        if enable_force_load_balance:
+        if get_ascend_config().enable_force_eplb:
+            topk_ids = get_force_eplb_topk(topk_ids, num_logical_experts)
+        elif enable_force_load_balance:
```
ningjingbengxiaohai, line 485: 原有的分支删除吧

vllm_ascend/ascend_config.py
ningjingbengxiaohai, line 490: 是不是还要约束下，moe模型才可以生效？

vllm_ascend/ops/fused_moe/routed_experts.py
weijinqian0: 之前的没法删除，之前是动态变化的，是为了dummy_run 采集显存用的。

vllm_ascend/ascend_config.py
weijinqian0: ok

Pz1116 (approved):

ningjingbengxiaohai (approved):

PR #15502, opened 2026-09-01
 [v0.26.0][Performance][Communicator] Replace per-layer F.pad with cat of a persistent zero block in MoE prepare

Levi-JQ:

`PrepareAndFinalizeWithMC2.prepare()` (and the identical site in `PrepareAndFinalizeWithAll2All`) pads `hidden_states` and `router_logits` to `padded_num_tokens` with `nn.functional.pad` on **every MoE layer, every step**, before the TP-split that feeds `MoeDistributeDispatchV2` (which requires a uniform token count across ranks). On NPU each `F.pad` lowers to two device kernels (`constant_pad_nd` → MemSet + PadV3) while the padded data is only a few KB in decode, so the four launches are pure overhead: ~44 us per MoE layer (~5.6% of a decode layer's kernel time in a Kimi K3 TP16 profile). With speculative decoding (`num_speculative_tokens=7`, minimum graph tier of 8 tokens), odd batch sizes always pad 8 rows to 16 and hit this path. The padded rows are dead data (dropped via `x_active_mask` in dispatch/combine and sliced off in finalize) — only the shape matters, so the padding can be made cheap.

This PR replaces both pads with `torch.cat([x, zero_block[:pad]], dim=0)` against a cached per-`(width, dtype)` zero block that is never written after allocation:

- one ConcatD kernel per tensor instead of MemSet + PadV3 (2 kernels per layer instead of 4);
- value-equivalent to `F.pad` — the zero tail is structural, not maintained by any cross-call bookkeeping;
- cudagraph-replay-safe by construction: the cat output is a fresh tensor per call, so each captured graph owns its output buffer and no mutable state is shared across capture sizes (a copy-into-persistent-buffer scheme breaks its zero-tail invariant when graphs of different sizes alternate at replay, and was rejected for that reason);
- covers eager and any capture size; applies to both MC2 and All2All paths.

### DIFF
--- a/vllm_ascend/ops/fused_moe/prepare_finalize.py
+++ b/vllm_ascend/ops/fused_moe/prepare_finalize.py
@@ -31,6 +31,7 @@ from vllm.model_executor.layers.fused_moe import FusedMoEConfig
 
 from vllm_ascend.ascend_forward_context import _EXTRA_CTX
 from vllm_ascend.lora.fused_moe import prepare_lora_indices
+from vllm_ascend.ops.fused_moe.comm_utils import _pad_tokens_with_cat
 from vllm_ascend.ops.fused_moe.moe_runtime_args import MoEPrepareOutput
 from vllm_ascend.quantization.quant_type import QuantType
 from vllm_ascend.utils import enable_sp, enable_sp_by_pass
@@ -164,8 +165,8 @@ class PrepareAndFinalizeWithAll2All(PrepareAndFinalize):
                 )
 
             if pad_size > 0:
-                hidden_states = nn.functional.pad(hidden_states, (0, 0, 0, pad_size))
-                router_logits = nn.functional.pad(router_logits, (0, 0, 0, pad_size))
+                hidden_states = _pad_tokens_with_cat(hidden_states, self.num_tokens + pad_size)
+                router_logits = _pad_tokens_with_cat(router_logits, self.num_tokens + pad_size)
                 padded_hidden_states_shape = hidden_states.shape
 
             if self.tp_size > 1:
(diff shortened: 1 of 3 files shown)


vllm_ascend/ops/fused_moe/prepare_finalize.py
```diff
@@ -35,6 +35,46 @@
...
+    zero_block = _PAD_ZERO_BLOCKS.get(key)
+    if zero_block is None or zero_block.shape[0] < padded_len - x.shape[0]:
+        zero_block = torch.zeros(
+            (max(_PAD_ZERO_BLOCK_ROWS, padded_len - x.shape[0]), *x.shape[1:]),
+            dtype=x.dtype,
+            device=x.device,
+        )
+        _PAD_ZERO_BLOCKS[key] = zero_block
+    return torch.cat([x, zero_block[: padded_len - x.shape[0]]], dim=0)
+
```
weijinqian0: Please move the function to the util.py file located under the moe directory.

vllm_ascend/ops/fused_moe/prepare_finalize.py
```diff
@@ -35,6 +35,46 @@
...
+# < tp_size (<= 64), and any smaller pad is a slice view of the block.
+#
+# Memory footprint is bounded: the key drops the token dim, so the cache holds
+# one entry per distinct (trailing shape, dtype, device) — a set fixed by the
+# model config, independent of batch size and cudagraph capture sizes (two
+# entries for Kimi K3 TP16: hidden 3584, experts 896). Each entry is
+# _PAD_ZERO_BLOCK_ROWS rows, at most 64 * width * dtype.itemsize bytes
+# (~0.5MB for the 3584-wide hidden states in bf16), so the whole cache peaks
+# at O(1) MB.
+_PAD_ZERO_BLOCK_ROWS = 64
```
weijinqian0: Why not simply use TP_size?，It can be read from vllm_config.

Levi-JQ: done

vllm_ascend/ops/fused_moe/prepare_finalize.py
Levi-JQ: > Please move the function to the util.py file located under the moe directory.

done

PR #15504, opened 2026-09-01
[CI] Added the following cases for offload precision verification

jiangli221:
Added the Qwen3-30B-A3B-W8A8-offload-test-acc case for offload precision verification.
Modified the verification logic of input_throughput.


wenjun91:
/weekly Qwen3-30B-A3B-W8A8-offload-test-acc
weekly command triggered.

PR #15506, opened 2026-09-01
[Bugfix][KVPool] Cover physical DCP shards in scheduler lookup

### LINKED ISSUE #15505: [Bug][KVPool][DCP] Scheduler lookup checks only DCP0 and can report false cache hits
## Your current environment

<details>
<summary>Environment summary</summary>

```text
vLLM Ascend: v0.26.0rc
Hardware: Ascend NPU
Deployment: P/D disaggregation with AscendStore + Mooncake KV pool
Topology: 2P2D, DP=4, TP=8, PCP=1, DCP=8, PP=1
Model layout: MLA (one KV head)
Speculative decoding: MTP, one speculative token
KV pool backend: Mooncake
```

The original environment is a restricted multi-node cluster, so the full
`collect_env.py` output cannot be attached publicly. The failure was reproduced
with engine processes remaining healthy; this is a request-level KV load error,
not an NPU or process crash.

</details>

## Describe the bug

`KVPoolWorker.lookup_scheduler()` performs an all-rank existence check before
advertising an external prefix-cache hit. With `TP=8`, `DCP=8`, `PCP=1`, and an
MLA model, each block has eight DCP-specific keys:

```text
...@pcp0@dcp0@head_or_tp_rank:0@pp_rank:0@...@<hash>
...
...@pcp0@dcp7@head_or_tp_rank:0@pp_rank:0@...@<hash>
```

However, `_expand_lookup_keys_by_rank()` currently rewrites only
`head_or_tp_rank` and `pp_rank`. For MLA,
`get_group_tp_size(group_id) == 1`, so a key built by the local DCP0 worker is
not expanded to DCP1-7. The scheduler can therefore treat DCP0 existence as an
all-rank hit, while the later rank-specific load requires every DCP key.

Relevant current code:

```python
def _expand_lookup_keys_by_rank(self, keys, group_id):
    expanded = []
    for pp_rank in range(self.pp_size):
        for tp_rank in range(self.get_group_tp_size(group_id)):
            for key in keys:
                tp_key = self._replace_key_field(
                    key, "head_or_tp_rank", tp_rank
                )
                expanded.append(
                    self._replace_key_field(tp_key, "pp_rank", pp_rank)
                )
    return expanded
```

### Observed request timeline

For one 44,032-token request with 2,048 local HBM-hit tokens:

```text
05:40:22.604 UTC  lookup builds 43 keys; every submitted key is @dcp0
05:40:22.606 UTC  exists(@dcp0 target block) = 1
05:40:22.606 UTC  scheduler reports raw_hit_tokens=4096
05:40:22.606 UTC  scheduler reports final_hit_tokens=4096
05:40:24.293 UTC  DCP4 loads the corresponding @dcp4 key
05:40:24.293 UTC  get(@dcp4 target block) = -704 (object not found)
05:40:25.993 UTC  the request returns HTTP 500
```

Across the reproduction logs:

```text
exists calls:              44
submitted exists keys:   2217
exists-key DCP coverage: DCP0 only
rank-specific get calls:  152
failed get calls:           1
```

The EngineCore and worker processes remained alive and both P/D health
endpoints returned HTTP 200.

### Expected behavior

An external KV block should count as an all-rank hit only when every physical
DCP shard required by that head/TP shard exists. If any DCP-specific key is
missing, the reported prefix hit must stop before that block so the request can
compute it instead of loading a nonexistent key.

### Proposed fix

Expand each base lookup key over the physical DCP ranks associated with its
`head_or_tp_rank`, while preserving complete block/layer rows for the existing
`[rank_shard][block]` hit reduction. For the reproduced MLA topology this means
checking `8 * 43 = 344` unique keys instead of 43 DCP0-only keys.

After applying this change, one validation run produced:

```text
requests:                         500/500 successful
exists calls:                     500
exists-key DCP coverage:          DCP0-DCP7 on every call
queried keys:                     210480
duplicate queried keys:           0
queries with DCP0 present but
  another DCP shard missing:       6 (23 block observations)
rank-specific get results:        4008 successful, 0 x -704
```

One captured query had DCP0-3 and DCP6 present but DCP4, DCP5, and DCP7 absent.
The old logic would have advertised an additional 1,024-token hit from DCP0;
the all-DCP reduction correctly returned no external hit for that block and the
request completed normally.

### Reproduction outline

1. Start a P/D deployment with AscendStore + Mooncake KV pool, `TP=8`,
   `DCP=8`, `PCP=1`, and an MLA model.
2. Use a small KV pool to create eviction pressure.
3. Send concurrent long-context requests.
4. Instrument the original `m_store.exists(keys)` call and later per-rank
   `m_store.get(keys)` calls.
5. Observe that scheduler existence checks contain only `@dcp0`, while actual
   loads use `@dcp0` through `@dcp7`; a non-DCP0 key can return `-704` after the
   scheduler has already advertised a hit.

This is a concrete false-hit mechanism related to #15060, but it does not claim
to explain every `-704`. Eviction between a complete all-rank `exists()` and the
subsequent `get()` remains a separate race.

zmc1997:

Fixes #15505.

`KVPoolWorker.lookup_scheduler()` must check every rank-specific key before
reporting an external prefix-cache hit. The previous expansion covered PP and
`head_or_tp_rank`, but not every DCP shard, so a DCP0 hit could be reported
even when another DCP key required by the subsequent load was missing.

This PR:

- uses the same `field:value` format for PCP and DCP key fields as the other
  rank fields;
- expands lookup keys over every DCP rank for each PP and head/TP rank;
- keeps each rank shard's block/layer keys contiguous for the existing
  `[rank_shard][block]` hit reduction.

The all-rank lookup still assumes `PCP=1`; this change does not add unvalidated
PCP expansion behavior.

Yes. DCP deployments now report an external cache hit only when every required
DCP shard exists.

The serialized key format changes from `@pcp0@dcp0` to
`@pcp:0@dcp:0`. Workers should be rolled together, and existing KVPool entries
written with the old format will not be reused.

```text
PYTHONPATH=. python3 tests/ut/distributed/ascend_store/test_config_data.py
Ran 52 tests ... OK

python3 -m py_compile +  vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py +  vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py

git diff --check
```

The fix was also validated in a multi-rank DCP deployment under KVPool
eviction pressure. No dedicated DCP lookup test file is added.

### DIFF
--- a/tests/ut/distributed/ascend_store/test_config_data.py
+++ b/tests/ut/distributed/ascend_store/test_config_data.py
@@ -66,12 +66,10 @@ class TestPoolKey(unittest.TestCase):
     def test_to_string(self):
         k = PoolKey(self.meta, "hash1")
         s = k.to_string()
-        self.assertIn("llama", s)
-        self.assertIn("@pcp2", s)
-        self.assertIn("@dcp3", s)
-        self.assertIn("@head_or_tp_rank:1", s)
-        self.assertIn("@pp_rank:0", s)
-        self.assertIn("hash1", s)
+        self.assertEqual(
+            s,
+            "llama@pcp:2@dcp:3@head_or_tp_rank:1@pp_rank:0@group:0@cache_role:kv@cache_family:default@hash1",
+        )
 
     def test_pp_ranks_use_distinct_keys(self):
         other_pp_meta = KeyMetadata("llama", 1, 2, 3, 1)
@@ -103,6 +101,7 @@ class TestLayerPoolKey(unittest.TestCase):
         meta = KeyMetadata("model", 0, 0, 0, 0)
         k = LayerPoolKey(meta, "h1", 5)
         s = k.to_string()
+        self.assertIn("@pcp:0@dcp:0", s)
         self.assertIn("@layer_id:5", s)
         self.assertIn("model", s)
         self.assertTrue(s.endswith("@h1"))
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/config_data.py
@@ -114,7 +114,7 @@ class PoolKey:
     def to_string(self):
         return (
             f"{self.key_metadata.model_name}"
-            f"@pcp{self.key_metadata.pcp_rank}@dcp{self.key_metadata.dcp_rank}"
+            f"@pcp:{self.key_metadata.pcp_rank}@dcp:{self.key_metadata.dcp_rank}"
             f"@head_or_tp_rank:{self.key_metadata.head_or_tp_rank}"
             f"@pp_rank:{self.key_metadata.pp_rank}"
             f"@group:{self.key_metadata.kv_cache_group_id}"
@@ -161,7 +161,7 @@ class LayerPoolKey(PoolKey):
     def to_string(self):
         return (
             f"{self.key_metadata.model_name}"
-            f"@pcp{self.key_metadata.pcp_rank}@dcp{self.key_metadata.dcp_rank}"
+            f"@pcp:{self.key_metadata.pcp_rank}@dcp:{self.key_metadata.dcp_rank}"
             f"@head_or_tp_rank:{self.key_metadata.head_or_tp_rank}"
             f"@group:{self.key_metadata.kv_cache_group_id}"
             f"@cache_role:{self.key_metadata.cache_role}"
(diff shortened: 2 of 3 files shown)


zmc1997:
The code-side pre-commit failure has been fixed, and pre-commit/select-tests now pass. This workflow checks for the exact ready label; the current ready-all label does not satisfy that condition, so selected tests remain skipped and ci-gate fails. Maintainers, please add ready to run the selected tests.

Pz1116 (approved):

LCAIZJ (approved):

PR #15442, opened 2026-09-01
[Feature][KVTransfer] Support multi-main-cache-spec layers in layerwise reuse layout

lsy0214:
### What this PR does / why we need it?

`build_layerwise_reuse_layout` required each multi-spec physical layer to have **exactly one main spec and one indexer spec**, so DeepSeek-V4 DSA layers (5 cache specs: `attn`, `indexer.k_cache`, `indexer.compressor.state_cache`, `compressor.state_cache`, `swa_cache`) raised `ValueError` and blocked the layerwise KV-cache-reuse path entirely for the model:

```
ValueError: Physical layer N with multiple cache specs must have exactly one main spec and one '.indexer.k_cache' spec
```

This PR relaxes the constraint:

- `'.attn'` is selected as the main spec (falls back to the first main spec if absent);
- the remaining non-indexer specs are preserved in a new `extra_main_specs` field of `LayerwiseLayerCacheSpecs` (currently informational; per-spec buffer planning is tracked by the existing TODO);
- the indexer spec becomes optional (`None` when absent);
- single-spec layers keep the existing fast path; multi-spec layers whose specs are all indexer specs still raise.

### Does this PR introduce _any_ user-facing change?

Yes for DeepSeek-V4 + `use_layerwise=true`: engine init no longer fails, layerwise KV reuse works. No change for models whose layers have at most 1 main + 1 indexer spec, or for layerwise-disabled deployments.

### How was this patch tested?

- Unit tests added: `tests/ut/kv_transfer/test_layerwise_multi_cache_spec.py` (7 cases)
  - DSV4-like 5-spec DSA layer builds without error; `.attn` chosen as main; `swa_cache`/`state_cache` preserved in `extra_main_specs`; indexer never lands in `extra_main_specs`;
  - indexer spec optional (layer without indexer works);
  - legacy 1-main + 1-indexer layout unchanged (regression guard);
  - multi-indexer-only layer still raises; single-spec fast path preserved.
- Manual end-to-end on 8x910B3: DeepSeek-V4-Flash-w8a8-mtp, vLLM v0.27.1 + vllm-ascend main, layerwise three-scenario comparison runs end to end with this patch (OFF 146.70 / ON 53.76 / ON+prefetch 55.37 tok/s, 50/50 requests each), while `use_layerwise=true` fails at engine init without it.

- vLLM main: https://github.com/vllm-project/vllm/commit/e6bfe03ad73a3330cb427885aa90d97a12e1c704

PR #15507, opened 2026-09-01
[Feature][KVTransfer] Global layer addressing for layerwise pool keys under pipeline parallel

lsy0214:

The layerwise KV pool key space previously used **per-stage local layer ids**. With `pipeline_parallel_size > 1` on the prefill side:

- Both pipeline stages emitted `LayerPoolKey` with layer ids `0..num_layers-1` (where `num_layers = model_config.get_num_layers()` is the **per-stage** count), so the two stages wrote to **the same keys** and overwrote each other's entries;
- A decode instance (PP=1) reads the full layer range `0..total_layers-1`, which cannot match either stage's keys - the layerwise PD path was effectively unusable with PP.

This PR introduces **global layer addressing** for layerwise pool keys:

- `KVPoolWorker` derives `pp_layer_offset` from `model_config.get_layers_start_end_indices()` and passes it to every `split_layers()` call site (save threads, receive threads, lookup);
- `PoolKey.split_layers(num_layers, layer_offset=0)` emits global layer ids: for DeepSeek-V4-Flash (43 layers + 1 MTP) with PP=2, stage 0 writes global layers 0..20, stage 1 writes 21..43 (MTP tail folds naturally into the last stage's range);
- The producer key sets are **disjoint** and their **union equals exactly** the PP=1 decode key space (0..43) - verified by unit test `test_producer_stages_cover_decode_keyspace_exactly`;
- Non-PP deployments are unaffected: the offset defaults to 0 everywhere and the emitted keys are byte-identical to previous releases.

Yes for layerwise PD with `--pipeline-parallel-size > 1` on the producer: layer KV entries from different pipeline stages no longer collide, and decode-side layerwise lookup can hit them. No change for PP=1 deployments (default offset 0).

- Unit tests added: `tests/ut/kv_transfer/test_layerwise_pp_global_layers.py` (8 cases)
  - default offset backward compatibility (keys identical to prior releases);
  - offset shifts layer ids correctly;
  - **PP=2 producer union == PP=1 decode key space** (stage0: 21 layers @ offset 0, stage1: 23 layers @ offset 21 incl. MTP@43, decode: 44 layers @ offset 0);
  - no key overlap between stages; key string contains global layer id; hash/equality semantics across constructions.
- All prior unit tests still pass: `test_decode_adaptor_prefill_pp.py` (8/8), `test_layerwise_multi_cache_spec.py` (7/7).


lsy0214:
Real-machine verification completed on 8x Ascend 910B3 (npu2, DeepSeek-V4-Flash-w8a8-mtp):

- PP structure: 8 workers (PP0_TP0-3 + PP1_TP0-3) under PP=2 x TP=4, kv_both + memcache host_shm, layerwise ON (prefetch_layers=3).
- **Global layer addressing confirmed at runtime** (log evidence in `pp_verify/lwpp_pp_keyspace_evidence.log`):
  ```
  Layerwise PP key space: stage 0/2 writes global layers [0, 22) (total 43).
  Layerwise PP key space: stage 1/2 writes global layers [21, 42) (total 43).
  ```
  The two stage key spaces are complementary, exactly as the unit test predicts - no collisions, decode-side key space fully covered.
- Benchmark: 50/50 requests succeeded, zero KV transfer errors (results in `pp_verify/lwpp_bench_result.json`).

Verification scripts and evidence added in commit a0c2885 (`pp_verify/`). Running the layerwise path with PP>1 was previously impossible (stage key collision); with this patch it runs end to end.

lsy0214:
Branch rebased: removed the two cherry-picked commits from the closed PR #15366 (the non-layerwise PP split fix turned out to be unnecessary - see #15366 for the re-verification). This PR now contains exactly the layerwise enablement for PP:

1. `feat(kv_pool): support multi-main-cache-spec layers in layerwise reuse layout` - DSV4 DSA layers (5 cache specs) no longer fail at engine init;
2. `feat(kv_pool): global layer addressing for layerwise pool keys under pipeline parallel` - stage-local layer ids previously collided across PP stages; now stage 0 writes global layers [0, 22) and stage 1 writes [21, 42), union == the PP=1 decode key space;
3. verification scripts + runtime evidence (`pp_verify/`).

lsy0214:
Rebased with two fixes: PR title prefix [Feat] -> [Feature] (pre-commit title check), and a one-line ruff-format fix in the test file (4f87e26). The E2E workflow is in action_required state (fork PR approval) - could a maintainer please approve the workflow run? This PR also depends on #15442 (same multi-cache-spec commit is the first commit here).

lsy0214:
Superseded by #16208 — clean rebase onto latest main with only the PP+layerwise changes (3 files). This PR had a stale base and mixed unrelated changes from other branches.

PR #15510, opened 2026-09-01

qijiajin:
> Rebased resubmission of #15253 onto current `main` (`f69831343`). The original review fixes are included in this branch.

Data-parallel ranks with no scheduled requests still execute a one-token dummy batch to participate in graph replay and collectives. For GDN/Mamba layers, treating that synchronization-only token as a real decode token can advance causal-conv and KDA recurrent state through block-table entries left by an earlier request.

This PR keeps the model/collective dummy shape unchanged, but builds zero-length GDN metadata for idle DP dummies. Both captured recurrent branches become no-ops while other layers and distributed communication continue to execute normally.

The change restores the behavior previously validated on the v0.26 Kimi K3 path. It adds no device-wide synchronization and does not change GDN kernels.

No API change. Multi-node GDN/Mamba inference no longer depends on whether another DP rank was idle between requests.

## How was this patch tested?

- `python -m compileall` for the changed runtime files
- `git diff --check`
- Existing unit expectations updated for both idle-dummy entry points; the existing GDN builder regression covers zero-length recurrent metadata
- Real-weight A/B on two A3 nodes, TP16 x DP2 / EP32, Kimi K3 GQA DSpark, async scheduling, prefix cache, FULL_DECODE_ONLY graph, 131072-input + 1024-output:
  - before: a cache-hit repeat diverged from token 0 and collapsed into a repeated `1.1.1...` output while speculative acceptance remained 80.55%
  - after: no corrupted output in five cold/warm repeats; the last four cold/warm pairs were token-for-token identical, with per-token acceptance between 95.81% and 97.49%
- Real-weight four-node smoke on TP16 x DP4 / EP64 with full QuaRot weights and the same 128K-to-1K case:
  - repetition 1: aggregate acceptance 85.40%, mean accepted length 6.98, per-DP acceptance 88.11% / 89.03% / 85.91% / 79.16%
  - repetition 2: aggregate acceptance 87.00%, mean accepted length 7.09, per-DP acceptance 87.40% / 84.08% / 87.09% / 89.56%
  - both repetitions reported a 94.63% prefix-cache hit rate and completed without the cross-DP rotating collapse seen before the fix
  - repetition 3 was not counted: the DP2 EngineCore exited concurrently with an external Docker exec ending with status 137. The host reported no OOM or device fault, but the initiating command was not captured, so this is recorded as an interrupted run rather than attributed to the patch. The shared nodes were reclaimed and no restart was performed.

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -2150,7 +2150,7 @@ class NPUModelRunner(GPUModelRunner):
                         # returns True. before returning early here we call
                         # dummy run to ensure coordinate_batch_across_dp
                         # is called into to avoid out of sync issues.
-                        self._dummy_run(1)
+                        self._dummy_run(1, skip_gdn_state_update=True)
                     if not has_kv_transfer_group():
                         # Return empty ModelRunnerOutput if no work to do.
                         return EMPTY_MODEL_RUNNER_OUTPUT
@@ -3125,6 +3125,7 @@ class NPUModelRunner(GPUModelRunner):
         num_scheduled_tokens: dict[str, int] | None = None,
         num_scheduled_tokens_np: np.ndarray | None = None,
         cascade_attn_prefix_lens: list[list[int]] | None = None,
+        skip_gdn_state_update: bool = False,
         cudagraph_runtime_mode: CUDAGraphMode | None = None,
         batch_descriptor: BatchDescriptor | None = None,
     ) -> tuple[PerLayerAttnMetadata, CommonAttentionMetadata | None]:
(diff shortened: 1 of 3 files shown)


qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2138,8 +2138,14 @@ def execute_model(
...
-                        # is called into to avoid out of sync issues.
-                        self._dummy_run(1)
+                        # is called into to avoid out of sync issues. Match the
+                        # active decode descriptor so an idle rank does not
+                        # downgrade graph mode across the DP group.
+                        self._dummy_run(
+                            self.uniform_decode_query_len,
+                            uniform_decode=True,
+                            skip_gdn_state_update=True,
+                        )
```
weijinqian0: Why is eager mode rolled back in image mode?

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/worker/model_runner_v1.py
qijiajin: The issue where a dummy run causes a certain step to fall back to eager mode may be related to `uniform_decode`. If `uniform_decode` is `true` for the other steps but defaults to `false` during the dummy run, the graph descriptors may become inconsistent, causing that step to fall back to eager mode. I’ll add some logging later to verify this. I’ve already reverted this change for now, and I’ll submit the PR again once the root cause is confirmed.

weijinqian0 (dismissed):

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E
- Merge Conflict Labeler

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (full workflow):**
- E2E

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

weijinqian0 (approved):

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #15518, opened 2026-09-01
[v0.26.0rc][Bugfix] Fix CPU binding device mapping

zouyida2052:

vLLM v0.24 changed worker device assignment so a local rank is not always the device ID that CPU binding should use. Indexing the discovered NPU list by `local_rank` can therefore select the wrong CPU/NUMA/IRQ pool and amplify cross-rank waiting at communication barriers.

This patch keeps the change narrow:

- `NPUWorker` resolves its assigned device through `current_platform.device_id_to_physical_device_id(local_rank)` when CPU binding runs.
- `bind_cpus` and `CpuAlloc` require that resolved NPU ID and use it for the current worker binding.
- Existing hardware-specific allocation behavior for A2, A3, 310P, and A5/Ascend 950 is unchanged.
- No worker helper, worker state, or 310P-specific worker path is added.

No CLI or public configuration is added. CPU binding now follows the worker assigned device mapping.

- `git diff --check`
- `ruff check .`
- `ruff format --check .`
- `PYTHONDONTWRITEBYTECODE=1 python3 -m compileall -q` for changed Python files
- Focused CPU binding pytest was attempted locally, but collection requires the unavailable `vllm` package.
- `pre-commit` is unavailable locally; GitHub pre-commit CI is used.

### DIFF
--- a/tests/ut/device_allocator/test_cpu_binding.py
+++ b/tests/ut/device_allocator/test_cpu_binding.py
@@ -27,6 +27,7 @@ from vllm_ascend.utils import AscendDeviceType
 def make_cpu_alloc(rank_id=0):
     cpu_alloc = object.__new__(CpuAlloc)
     cpu_alloc.rank_id = rank_id
+    cpu_alloc.current_npu = 0
     cpu_alloc.device_info = SimpleNamespace(
         running_npu_list=[0],
         all_logic_npus=[0],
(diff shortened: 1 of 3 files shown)


zouyida2052:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

kunpengW-code (approved):

PR #15555, opened 2026-09-02
Fix/310p dflash full and piecewise route

HaonanWu1:


lvturner:
Independent validation on **Ascend 310P (Orange Pi AIStudio Pro / "Orange Pi AI Pro")** — the branch works end-to-end on this hardware. Sharing the build recipe, two 310P findings, and benchmark numbers in case they help the merge.

## Hardware

- Orange Pi AIStudio Pro (Orange Pi AI Pro), amd64 host
- 2× **Ascend 310P1** inference SoCs (Atlas 300I Duo-class), ~93 GiB LPDDR4X per die, HCCS die-pair, TP=2
- CANN 9.1.0-310P ops package, torch 2.10.0 / torch_npu 2.10.0.post4, Python 3.12

## How we built it (amd64)

Upstream publishes `nightly-main-310p` as arm64-only (single-arch), so we built an amd64 replica on the official `v0.23.0-310p` base: **vLLM core v0.24.0 + this PR's head (`7fe925c`)**, csrc rebuilt for ascend310p. Two notes from the pairing work:

1. The PR header's "vLLM main ba07e4a" appears stale — 0.28-era core imports `triton.experimental.gluon`, which triton-ascend 3.2.x doesn't provide. v0.24.0 (the in-branch compat target) builds and runs cleanly.
2. On this branch `COMPILE_CUSTOM_KERNELS=0` also disables the **torch extension** — a python-only `pip install -e` leaves the base image's stale `vllm_ascend_C.so`, and serving then fails at request time with `AttributeError: ... _C_ascend ... npu_copy_and_expand_dflash_inputs`. We added an explicit `COMPILE_CUSTOM_KERNELS=1 python setup.py build_ext --inplace` after the aclnn build.

One 310P patch we needed (not on A2/A3): `aclnnMatmulWeightNz` rejects `mat2` with k == 1 in FRACTAL_NZ — the drafter's `[1, hidden]` projection trips it. We keep such weights ND via a small guard in `_should_trans_nz` (`if 1 in weight.shape[-2:]: return False` — vLLM weights are `[out, in]`). Happy to send it as a commit if useful.

## Validation (Qwen3.6-35B-A3B W8A8 + z-lab/Qwen3.6-35B-A3B-DFlash, TP2, 262k ctx, fp16)

- Correct answers (incl. arithmetic), reasoning parser, `qwen3_xml` tool calls (finish_reason=tool_calls), greedy clean
- FULL_DECODE_ONLY: graph capture + replay stable, zero errors in multi-stream soaks
- PIECEWISE failed for us at capture on TP2: an `all_reduce` inside the captured region invalidates the stream (`rtStreamEndCapture` 507903). Not chased further since FDO works; can provide full logs if helpful.

Decode throughput (our streaming bench, prose workload, temperature 0.6):

| Config | single-stream | 2-stream agg | 3-stream agg | TTFT |
|---|---|---|---|---|
| DFlash K=7, FDO | **61.6 tok/s** | 83.1 | 55.9 | 0.17s |
| DFlash K=3, FDO | 50-53 | **89.6** | — | 0.17s |
| DFlash K=15, FDO | 42 | 65.7 | — | 0.20s |
| DFlash K=15, eager | 36-37 | 51.1 | — | 0.21s |

Acceptance length ~3.2-3.5 on short prose prompts (your 6.9/15 on GSM8K + xhigh thinking matches the workload dependence). Prefill 1.3-1.6k tok/s at 2k-16k prompts.

Thanks for this excellent work — the AscendC ops (copy_and_expand, compute-WY, fused rejection sampling) are exactly what 310P needs given it has no usable Triton path.

PR #15568, opened 2026-09-02
[Performance][Attention] Fuse local SFA LSE formation into DCP pack

Ruiqiu-Zheng:

This PR removes a local SparseFlashAttention DCP post-processing boundary by forming LSE inside the existing pack kernel instead of materializing a standalone FP32 LSE tensor before packing.

Before:

- native FP32 `softmax_max` / `softmax_sum`
- materialize `LSE = softmax_max + log(softmax_sum)`
- permute / reshape
- DCP pack
- HCCL All2All
- existing stable combine

After:

- native FP32 `softmax_max` / `softmax_sum`
- max/sum-aware DCP pack, which forms the same LSE while writing the packed layout
- same HCCL All2All
- unchanged stable combine

The change preserves the existing row/head/token mapping, packed payload semantics, HCCL communication pattern, and downstream stable-combine ABI.

The legacy materialized-LSE custom-op path is kept for compatibility and regression coverage.

Invalid local statistics are handled explicitly. Non-finite `softmax_max`, non-finite `softmax_sum`, or `softmax_sum <= 0` produce an invalid (`-inf`) LSE contribution, which is then handled by the existing combine semantics. Focused tests cover these cases.

### Performance evidence

The performance experiment used a preregistered paired/interleaved A/B protocol with fixed inputs and ordering:

- protocol seed: `20260916`
- BF16 producer seed: `20260913`
- FP16 correctness seed: `20260914`
- 10 warmup iterations
- 5 timed samples per arm/block
- 8 paired A/B blocks
- fixed base/candidate order
- pre/post baseline-only null controls
- no post-hoc performance samples

The positive decision rule was fixed before measurement:

`at least 75% of paired blocks favor candidate AND paired median improvement > max(1%, 2 * relative_noise_floor)`

Matched measurements on the recorded Ascend test environment were:

| Interval | Baseline median | Candidate median | Paired median effect | Noise floor | Result |
| --- | ---: | ---: | ---: | ---: | --- |
| local composition: precomputed output/max/sum -> pack | 310.342 us | 242.750 us | 21.655% faster | 4.237% | positive, 8/8 pairs favor candidate |
| world-size-one SparseFlashAttention producer -> pack | 435.846 us | 375.199 us | 12.945% faster | 5.847% | positive, 8/8 pairs favor candidate |
| same-host two-rank producer -> HCCL -> stable combine | 5.043 ms | 5.068 ms | 0.385% paired effect in favor of candidate | 4.410% | neutral, 4/8 pairs favor candidate |

For the two-rank lane, the standalone arm medians and paired median point in slightly different directions, and both effects are far below the 8.819% decision threshold. It is therefore classified as **neutral**, not as either a speedup or a regression.

These numbers are scoped to the measured local/laboratory intervals. They are **not** end-to-end model, serving-throughput, or production-performance claims.

A sanitized reproducibility packet is available on an auxiliary branch of the fork. It is not part of this production PR and is not a CI dependency:

- protocol, seeds, fixture, raw paired/null-control measurements, and correctness sentinels:
  https://github.com/Ruiqiu-Zheng/vllm-ascend/tree/repro/sfa-local-lse-r016-20260902/repro/sfa-local-lse-r016

- README and reproduction instructions:
  https://github.com/Ruiqiu-Zheng/vllm-ascend/blob/repro/sfa-local-lse-r016-20260902/repro/sfa-local-lse-r016/README.md

- independent reducer that recomputes the table above from raw JSON:
  https://github.com/Ruiqiu-Zheng/vllm-ascend/blob/repro/sfa-local-lse-r016-20260902/repro/sfa-local-lse-r016/reduce_results.py

The exact binary fixture is:

`SHA256 63785fa8f4592b7f8f9e06d345a20cf29a4ac8d19adc1e89dc8f90e60eb7db19`

The packet also contains the exact candidate reconstruction patch and source/runtime lineage.

Absolute latency is not expected to reproduce bit-for-bit on a different hardware/runtime state. The intended reproduction target is the paired effect relative to a freshly observed null-control noise floor under matched lineage and the same protocol.

No.

This is an internal DCP post-processing optimization. It does not add or change a public API, user configuration, environment variable, or user-visible model behavior.

Focused post-rebase regression on the PR branch:

- host / fake-op / routing tests: **13 passed**
- single-device NPU tests: **32 passed**
- same-host two-rank HCCL eager + NPUGraph tests: **2 passed**
- Ruff check: passed
- Ruff format check: passed
- `py_compile`: passed
- `git diff --check`: passed

NPU correctness coverage includes:

- BF16 and FP16
- `scatter_dim = 0` and `scatter_dim = 1`
- head dimension 512 on the measured specialization
- non-contiguous output / max / sum strides
- zero, negative, NaN, and Inf softmax sums
- non-finite softmax maxima
- all-invalid rows
- legacy materialized-LSE regression

The two-rank graph regression covers eager execution, graph capture, first replay, and changed-input replay while keeping input/output storage addresses fixed. This checks that replay consumes updated tensor contents rather than returning stale captured results.

The PR was rebased as upstream `main` advanced. Before each rebase, the decision-relevant Local-SFA callsite/pack seam was checked for upstream drift. No relevant source drift was observed, and the focused post-rebase regressions remained passing.


Ruiqiu-Zheng:
CI triage: `pre-commit`, coverage recommendation, and `select-tests` all passed. The selected test jobs were skipped only because this PR does not yet have a ready label, so `ci-gate` failed without any test case actually failing.

Could a maintainer please add `ready-precise` to run the selected relevant test set? `ready-all` is not needed at this stage.

Ruiqiu-Zheng:
Current-main follow-up: this PR is still conflict-free against `main` at `e5118d151314ae18e56c0b63aa8dd00d294adc22`, and the max/sum-to-pack Local-LSE seam has no equivalent implementation on current main. The existing `pre-commit`, coverage recommendation, and `select-tests` checks passed; the relevant selected tests remain blocked only by the missing ready label.

Could a maintainer please add `ready-precise` to run the focused selected tests? `ready-all` is not needed at this stage.

PR #15579, opened 2026-09-02
[Performance] Reuse PA workspace across layers in graph param update

solin28825:
# perf(attention): reuse PA workspace across layers in graph param update

## Motivation

In `AscendAttentionBackendImpl` (graph mode, `_npu_paged_attention` path), the
graph-param update loop calls `torch_npu._npu_paged_attention_get_workspace`
**once per layer**. On a 28-layer model this is 28 calls per decode step, each
backed by a ~36-54MB buffer allocation and a host-side "Setup" round trip.

Profiling on Ascend 910B (vLLM v0.26.0, vllm-ascend v0.19.1rc1 base, TP=1,
FULL_DECODE_ONLY, steady-state c=32):

| Metric | Before | After |
|---|---|---|
| get_workspace Setup / Execute ratio | 2.000 (22232/11116) | **1.036** (14355/13860) |
| update_graph_params avg | 2.4-2.95 ms | **1.65-1.67 ms** (-1.1~1.3 ms/step) |
| Host ACL API total (5.1s window) | 1335 ms | **1040 ms** (-22%) |
| QPS c=1 / c=32 / c=64 | 1.27 / 16.73 / 21.49 | **1.35 / 17.13 / 21.90** (+5.9% / +2.4% / +1.9%) |

Functional outputs are unchanged (same md5 on the full functional suite).

## Change

Hoist the workspace out of the per-layer loop: query it once per update pass
and reuse across layers, re-querying only when the seq_lens tensor identity
changes (`(data_ptr, shape)` key). This is the same shape as the FIA path,
which already reuses `graph_params.workspaces[num_tokens]`.

## Why per-pass (not cross-step) reuse is safe

- The workspace size depends only on shapes **and seq_lens**, and all layers
  within one update pass share the same step's `seq_lens` tensor, so one call
  is sufficient.
- Cross-step reuse is deliberately **avoided**: the workspace size also
  depends on the *values* of seq_lens (observed 36MB for lens<=80 vs 54MB for
  lens<=400), which grow as decode progresses — a cached workspace from a
  previous step can be too small.
- If any layer's seq_lens tensor differs (key mismatch), the code falls back
  to a fresh `get_workspace` call — conservative by construction.

## Environment

- Ascend 910B, CANN 8.x, torch_npu 2.10.0.post2
  the touched code is identical on main as of 2026-09-02)

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -538,6 +538,16 @@ class AscendAttentionBackendImpl(AttentionImpl):
             else:
                 graph_params = get_graph_params()
             with torch.npu.stream(update_stream):
+                # The workspace size depends only on shapes and seq_lens, which
+                # are shared by all layers within one graph-param update pass,
+                # so one get_workspace call per pass is sufficient. Reuse the
+                # workspace across layers and re-query only when any
+                # size-relevant input changes. This mirrors the FIA path, which already
+                # reuses graph_params.workspaces[num_tokens], and removes the
+                # redundant per-layer get_workspace calls (each backed by a
+                # ~36-54MB buffer allocation) per decode step.
+                step_ws_key = None
+                workspace = None
                 for key, param, handle, event in zip(
                     forward_context.attn_metadata,
                     graph_params.attn_params[num_tokens],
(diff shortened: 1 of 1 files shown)


yiz-liu (approved):

solin28825:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

solin28825:
/rerun
[Bot]: rerun completed.

**Failed:**

PR #15585, opened 2026-09-02
[Bugfix] Support standardized KV-cache layouts on current vLLM main

ShuhaoZhangTony:
## What this PR does

Updates the Ascend attention backend for the standardized strided KV-cache contract on current vLLM main.

- delegates KV-cache binding to the current core helper instead of replacing it;
- allocates the standardized strided cache through the core allocator;
- advertises `LHBNC` and packs each attention spec as two dense, contiguous K/V planes;
- propagates backend `customize_spec()` in the v1 and v2 spec collection paths;
- unpacks the standardized per-layer view into the 4-D K/V tensors expected by Ascend kernels;
- retains compatibility paths for legacy K/V pairs and keeps Ascend compressed-MLA metadata consistent with current core fields.

The patch preserves the PCP refactor already present on vLLM-Ascend main; it does not reintroduce a product-specific launcher or monkey patch.

## Reproduction and root cause

With current vLLM KV-cache layout APIs, vLLM-Ascend still bound and interpreted cache tensors using the older pair/list contract. The backend-specific packing was not applied in every spec collection path, and the old binding replacement discarded current cache-group metadata.

On Ascend graph-mode startup this surfaced as:

```text
Standardized Ascend KV cache must contain two dense K/V planes, got 1 slots
```

The failure occurred before serving because the allocated per-layer view did not match the dense K/V layout required by the Ascend native attention kernels.

## Compatibility boundary

- Tested vLLM integration commit: `7c82451c03818d23746c962c173ffb1dbc78891b`
- vLLM-Ascend base: `8a4ea1478b2eefcc516ce687674c7bef3db1f0eb`
- Python 3.12
- PyTorch 2.13.0
- torch-npu 2.13.0rc1
- CANN 9.1

The current standardized path is selected by the presence of the new cache tensor descriptors. Older release-line descriptors continue through the existing compatibility path.

## Tests

Unit tests in the current-core environment:

```text
87 passed
```

Covered files:

```text
tests/ut/attention/test_attention_v1.py
tests/ut/worker/test_model_runner_v1.py
tests/ut/worker/test_attn_utils_v2.py
```

Static checks:

```text
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

NPU E2E:

- Ascend 910B2, 4 devices, tensor parallel size 4
- `zai-org/GLM-4-32B-0414`
- graph mode (`enforce_eager=False`)
- cold startup and graph compilation succeeded
- `/health`, `/v1/models`, non-streaming completion, streaming completion, two-request concurrency, client cancellation, and post-restart completion passed
- observed streaming TTFT: 0.23 s in the validation run

## Risk

The change affects KV-cache allocation, layout selection, and view binding. The tests exercise both standardized tensor views and compatibility pair/list inputs. The implementation rejects non-contiguous or malformed standardized layouts early instead of allowing native operators to consume an ambiguous cache view.


ShuhaoZhangTony:
Pre-commit and the Python 3.10/3.11/3.12 mypy gate now pass on head 58741b323. Coverage recommendation and test selection also complete successfully. The remaining ci-gate is expected because contributors cannot apply repository labels (the API returns 403). Maintainers: please add `ready-precise` to run the selected test set.

PR #15589, opened 2026-09-02
[Performance][Ops] Split fused qkvz projection into separate qkv and z linears in GDN

leijie-ww:

In `vllm_ascend/ops/gdn.py`, `AscendGatedDeltaNetAttention.forward` previously computed the fused `qkvz` input projection and then split the output tensor into `mixed_qkv` and `z`.

For the non-interleaved GQA layout used by Qwen3.5 (`gqa_interleaved_layout=False`), this PR adds a fast path that splits the projection weight up front and computes the two parts with independent `torch.nn.functional.linear` calls:

```python
qkv_size = (self.key_dim * 2 + self.value_dim) // self.tp_size
z_size = self.value_dim // self.tp_size

use_weight_split = (
    getattr(self, "quant_config", None) is None
    and isinstance(
        getattr(self.in_proj_qkvz, "quant_method", None),
        UnquantizedLinearMethod,
    )
)

if use_weight_split:
    qkvz_weight = self.in_proj_qkvz.weight
    mixed_qkv = torch.nn.functional.linear(
        hidden_states, qkvz_weight[:qkv_size]
    )
    z = torch.nn.functional.linear(
        hidden_states, qkvz_weight[qkv_size:]
    )
else:
    mixed_qkvz, _ = self.in_proj_qkvz(hidden_states)
    mixed_qkv, z = mixed_qkvz.split([qkv_size, z_size], dim=-1)
```

The weight-split path is restricted to genuinely unquantized models. Quantized models continue to use the projection module's normal `forward` path before splitting the result.

This guard is required for the quantized Qwen3.5 regression identified during review. ModelSlim checkpoints can leave Qwen3.5 GDN projections marked as `FLOAT` while quantizing surrounding MoE layers as `W8A8_DYNAMIC`. Even in that case, `AscendUnquantizedLinearMethod` may pack the stored weight in `FRACTAL_NZ` format. Applying `F.linear` directly to the stored weight would bypass the required Ascend weight handling and produce incorrect results.

Why the unquantized fast path is useful:

- Avoids materializing the full `mixed_qkvz` intermediate tensor.
- Avoids the explicit `tensor.split` on the projected output.
- Produces `z` as a fresh contiguous tensor, allowing the following reshape to remain a view.
- Avoids tuple/bias handling in `MergedColumnParallelLinear.forward` when it is safe to do so.
- Preserves the quantization method and custom-op path for quantized configurations.

The interleaved-layout Qwen3-Next branch and the separate `in_proj_qkv` / `in_proj_z` branch are unchanged.

No. This is an internal projection refactor.

For unquantized models, the new path is numerically equivalent to the previous fused-projection-then-split implementation. For quantized models, the original quantization-aware projection path is preserved to avoid accuracy regressions.

- Added `tests/ut/ops/test_gdn_qkvz_split.py`.
- Verified numerical equivalence across four shape and tensor-parallel configurations, including `tp_size=2`.
- Added a regression test confirming that the unquantized fast path bypasses `in_proj_qkvz.forward`.
- Added packed INT8 coverage confirming that quantized weights use the module `forward` path.
- Added coverage for the Qwen3.5 mixed configuration where GDN projections are `FLOAT` but the model has an active quantization configuration.

```text
pytest -q tests/ut/ops/test_gdn_qkvz_split.py
7 passed
```

Focused GDN unit tests also pass:

```text
pytest -q \
  tests/ut/ops/test_gdn_qkvz_split.py \
  tests/ut/ops/test_gdn_attn_builder.py \
  tests/ut/ops/test_gdn_chunk_meta.py

35 passed
```

- E2E greedy-generation coverage:
  `tests/e2e/pull_request/four_card/test_qwen3_5.py`
  for Qwen3.5-27B dense and Qwen3.5-35B-A3B MoE models.
- TP2 hardware smoke test completed successfully with
  `Eco-Tech/Qwen3.6-35B-A3B-w8a8` on NPU 8 and NPU 9.
- GPQA test result on `Eco-Tech/Qwen3.6-35B-A3B-w8a8`:

image
image
image

- `py_compile` and `git diff --check` pass.
  https://github.com/vllm-project/vllm/commit/ba07e4a48fc951300d97eb506217dd530583dea3


vllm_ascend/ops/gdn.py
```diff
@@ -200,11 +200,11 @@ def forward(
             a = a.contiguous()
         else:
             if not self.gqa_interleaved_layout:
-                mixed_qkvz, _ = self.in_proj_qkvz(hidden_states)
-                num_tokens = mixed_qkvz.size(0)
+                qkvz_weight = self.in_proj_qkvz.weight
                 qkv_size = (self.key_dim * 2 + self.value_dim) // self.tp_size
-                z_size = self.value_dim // self.tp_size
-                mixed_qkv, z = mixed_qkvz.split([qkv_size, z_size], dim=-1)
+                mixed_qkv = torch.nn.functional.linear(hidden_states, qkvz_weight[:qkv_size])
```
maoxx241: [P1] Preserve the quantized projection path

Directly slicing `in_proj_qkvz.weight` and calling `F.linear` bypasses `in_proj_qkvz.quant_method.apply`. Qwen3.5 registers this fused projection for quantization; for Ascend W8A8, the quant method quantizes the activations, consumes the scale parameters, calls `npu_quant_matmul`, and transposes/NZ-converts the stored weight after loading. Feeding that processed weight to `F.linear` therefore uses the wrong dtype/layout and drops the required scale handling, so quantized Qwen3.5 can fail or return incorrect results. The new CPU stub only models a plain floating-point 2-D weight and cannot catch this. Please retain the module-forward path for quantized methods (or implement a quantization-aware split), add a quantized regression, and run the selected Qwen3.5 W8A8 E2E test.

maoxx241 (changes requested):
Blocking on the quantized Qwen3.5 regression described in the inline comment.

vllm_ascend/ops/gdn.py
leijie-ww: Updated: F.linear is now used exclusively for the non-quantized in_proj_qkvz.weight under normal ND layout, while all other cases retain the original execution path.

Ronald1995 (commented):
it's better to validate the e2e result

vllm_ascend/ops/gdn.py
```diff
@@ -200,11 +237,35 @@ def forward(
...
+                        use_weight_split = isinstance(
+                            getattr(self.in_proj_qkvz, "quant_method", None), UnquantizedLinearMethod
+                        ) and _is_standard_nd_weight(self.in_proj_qkvz.weight)
+                        self._use_weight_split = use_weight_split
+                if use_weight_split:
+                    qkvz_weight = self.in_proj_qkvz.weight
+                    mixed_qkv = torch.nn.functional.linear(hidden_states, qkvz_weight[:qkv_size])
+                    z = torch.nn.functional.linear(hidden_states, qkvz_weight[qkv_size:])
+                else:
+                    mixed_qkvz, _ = self.in_proj_qkvz(hidden_states)
```
big-surprise, line 274: The current method yields no performance improvement for quantified weights.

maoxx241 (approved):

leijie-ww:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

leijie-ww:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/ops/gdn.py
leijie-ww: only work for unquantified weight at this stage.

leijie-ww:
> it's better to validate the e2e result

already discussed

PR #15590, opened 2026-09-02
[Performance][Ops] Cache packed GDN conv1d weight instead of repacking each forward

leijie-ww:

`AscendGatedDeltaNetAttention._forward_core` repacked the conv1d weight on **every forward** call: it did `conv_weights = self.conv1d.weight.view(D, W)` and then called `conv_weights.transpose(0, 1).contiguous()` again in **four** separate branches (spec / PCP-prefill / non-PCP-prefill / decode). For a hybrid GDN model with many such layers and a long decode loop, that is a wasted transpose + contiguous allocation per layer per step, with no functional benefit since the weight is immutable.

This PR centralizes that into a single helper `_get_packed_conv_weights` that materializes the `[W, D]` transposed weight **once** and caches it on the layer:

- Mirrors `kimi_kda._pack_conv_weights` (`AscendKimiK3DeltaAttention` already does the same packing-once pattern).
- Cached as a **plain attribute** (`setattr`, name `ascend_conv1d_weight`) rather than a registered parameter, because this class is patched onto `QwenGatedDeltaNetAttention` via method copy so `__init__` never runs on the patched instance — packing must therefore happen lazily on first use.
- Meta-device weights (meta-init / checkpoint loading) are handled specially: a non-cached view is returned so packing is deferred until a real (materialized) weight exists, avoiding storage allocation on meta tensors.
- The four per-branch `conv_weights_T = conv_weights.transpose(0, 1)` lines are removed in favor of a single `conv_weights_T = _get_packed_conv_weights(self)` at the top of the conv section.
- `width = conv_weights.shape[1]` becomes `width = self.conv_kernel_size` — equivalent (that index was always the kernel width `W`), and the `conv_weights` local is no longer needed.

Behavior is unchanged; only redundant work is removed.

No. This is an internal performance refactor with identical numerics — the packed `[W, D]` weight passed to `npu_causal_conv1d_custom` is byte-for-byte the same tensor that was previously computed inline. No public API, config option, flag, model interface, or output changes. No environment variables are added or modified.

- New unit tests added in `tests/ut/ops/test_gdn.py` covering `_get_packed_conv_weights`:
  - `test_get_packed_conv_weights_packs_in_kernel_layout` — output is `[W, D]`, contiguous, correct dtype, values equal `weight[:, 0, :].transpose(0, 1)`.
  - `test_get_packed_conv_weights_caches_after_first_call` — second call reuses the cached tensor (same `data_ptr`); stored as a plain attribute, not a `Parameter`; later mutation of the source weight does not invalidate the cache.
  - `test_get_packed_conv_weights_meta_weight_returns_view_without_caching` — meta-device weights return a non-cached view and do not populate the cache attribute.
- Ran locally on the CPU UT runner (conftest mocks `torch_npu`):
  ```
  pytest tests/ut/ops/test_gdn.py tests/ut/ops/test_kimi_kda.py
  => 10 passed
  ```
  (the `kimi_kda` suite is included to confirm no regression in the analogous packing path.)
- The `_forward_core` refactor (branch consolidation + `width = self.conv_kernel_size`) is an equivalent substitution validated by code inspection; the existing GDN functional/e2e suites (`test_gdn_*`) cover the forward paths end-to-end on NPU.
- GPQA test result on Eco-Tech/Qwen3.6-35B-A3B-w8a8
image

---

### DIFF
--- a/vllm_ascend/ops/gdn.py
+++ b/vllm_ascend/ops/gdn.py
@@ -15,13 +15,17 @@
 # limitations under the License.
 #
 
+from functools import wraps
+
 import torch
 import torch_npu
 from einops import rearrange
+from torch import nn
 from vllm.distributed import get_pcp_group
 from vllm.forward_context import get_forward_context
 from vllm.model_executor.layers.mamba.gdn.base import GatedDeltaNetAttention
 from vllm.model_executor.layers.mamba.mamba_utils import MambaStateShapeCalculator
+from vllm.model_executor.utils import replace_parameter
 from vllm.third_party.flash_linear_attention.ops.l2norm import l2norm_fwd
 from vllm.triton_utils import triton
 from vllm.v1.attention.backend import AttentionBackend, AttentionMetadata  # type: ignore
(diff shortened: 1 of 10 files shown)


vllm_ascend/ops/gdn.py
```diff
@@ -40,6 +40,29 @@
 from vllm_ascend.ops.triton.fla.utils import clear_ssm_states
 from vllm_ascend.ops.triton.mamba.causal_conv1d import extract_last_width
 
+_PACKED_CONV_WEIGHT_NAME = "ascend_conv1d_weight"
+
+
+def _get_packed_conv_weights(layer: "AscendGatedDeltaNetAttention") -> torch.Tensor:
```
nu11ptr7, line 46: The packed weight is cached on first use and never invalidated, so any later conv1d.weight update (LoRA, draft-weight swapping, reloading) is silently ignored and the stale packed copy keeps being used. If this layer's weights can change at runtime, the cache needs invalidation keyed on the weight's version/pointer.

maoxx241 (approved):

vllm_ascend/ops/gdn.py
leijie-ww: Thank you for your comment. Currently vllm-ascend does not fully support weight update functionality, and we would refactor this part once needed.

weijinqian0: Need check if there is some error in RL scenario.

leijie-ww: Introduces functions to manage the cached weight (initialize_packed_conv_weight, _pack_conv_weights) and integrates this with the layer's weight-loading and update process, ensuring the cache is updated when base weights change.

lijiahang226 (approved):
LGTM

vllm_ascend/ops/gdn.py
leijie-ww: Has updated code to involve weight update mechanism. Pls check, thanks

weijinqian0 (approved):

PR #15591, opened 2026-09-02
[Bugfix][KVPool] Cover physical DCP shards in scheduler lookup

zmc1997:

Main-branch counterpart of #15506. Addresses #15505.

`KVPoolWorker.lookup_scheduler()` must check every rank-specific key before
reporting an external prefix-cache hit. The previous expansion covered PP and
`head_or_tp_rank`, but not every DCP shard, so a DCP0 hit could be reported
even when another DCP key required by the subsequent load was missing.

This PR:

- uses the same `field:value` format for PCP and DCP key fields as the other
  rank fields;
- expands lookup keys over every DCP rank for each PP and head/TP rank;
- keeps each rank shard's block/layer keys contiguous for the existing
  `[rank_shard][block]` hit reduction.

The all-rank lookup still assumes `PCP=1`; this change does not add unvalidated
PCP expansion behavior.

Yes. DCP deployments now report an external cache hit only when every required
DCP shard exists.

The serialized key format changes from `@pcp0@dcp0` to
`@pcp:0@dcp:0`. Workers should be rolled together, and existing KVPool entries
written with the old format will not be reused.

```text
PYTHONPATH=. python3 tests/ut/distributed/ascend_store/test_metadata.py
Ran 58 tests ... OK

python3 -m py_compile +  vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/metadata.py +  vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py

git diff --check
```

The same implementation was validated in a multi-rank DCP deployment under
KVPool eviction pressure. No dedicated DCP lookup test file is added.

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
@@ -2211,12 +2211,18 @@ class KVPoolWorker:
         return f"{key[:value_start]}{value}{key[value_end:]}"
 
     def _expand_lookup_keys_by_rank(self, keys: list[str], group_id: int) -> list[str]:
+        # All-rank KV pool lookup currently assumes PCP=1.
         expanded: list[str] = []
+        num_head_or_tp_ranks = self.get_group_tp_size(group_id)
+        # Keep each rank shard's block/layer keys contiguous to match
+        # lookup_scheduler()'s [rank_shard][block] result slicing.
         for pp_rank in range(self.pp_size):
-            for tp_rank in range(self.get_group_tp_size(group_id)):
-                for key in keys:
-                    tp_key = self._replace_key_field(key, "head_or_tp_rank", tp_rank)
-                    expanded.append(self._replace_key_field(tp_key, "pp_rank", pp_rank))
+            for dcp_rank in range(self.dcp_size):
+                for head_or_tp_rank in range(num_head_or_tp_ranks):
+                    for key in keys:
+                        rank_key = self._replace_key_field(key, "dcp", dcp_rank)
+                        rank_key = self._replace_key_field(rank_key, "head_or_tp_rank", head_or_tp_rank)
+                        expanded.append(self._replace_key_field(rank_key, "pp_rank", pp_rank))
         return expanded
 
     def _expand_lookup_key_variants(self, key: str, group_id: int, include_all_ranks: bool) -> list[str]:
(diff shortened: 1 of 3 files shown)


zmc1997:
The code-side pre-commit failure has been fixed, and pre-commit/select-tests now pass. Selected tests require the repository ready label, but the PR author does not have permission to add it. Maintainers, please add ready to run the selected tests.

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
```diff
@@ -2211,12 +2211,18 @@ def _replace_key_field(key: str, field: str, value: int) -> str:
...
         expanded: list[str] = []
+        num_head_or_tp_ranks = self.get_group_tp_size(group_id)
+        # Keep each rank shard's block/layer keys contiguous to match
+        # lookup_scheduler()'s [rank_shard][block] result slicing.
         for pp_rank in range(self.pp_size):
-            for tp_rank in range(self.get_group_tp_size(group_id)):
-                for key in keys:
-                    tp_key = self._replace_key_field(key, "head_or_tp_rank", tp_rank)
-                    expanded.append(self._replace_key_field(tp_key, "pp_rank", pp_rank))
+            for dcp_rank in range(self.dcp_size):
```
nu11ptr7, line 2220: The key format change (`@pcp2` to `@pcp:2`) means keys written by an older version - or a mixed-version peer during a rolling upgrade - never match the new lookups, so every remote entry misses. Fine if the pool isn't expected to survive upgrades, otherwise a version tag or keeping the old layout readable would help.

Pz1116 (approved):

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
LCAIZJ: > The key format change (`@pcp2` to `@pcp:2`) means keys written by an older version - or a mixed-version peer during a rolling upgrade - never match the new lookups, so every remote entry misses. Fine if the pool isn't expected to survive upgrades, otherwise a version tag or keeping the old layout readable would help.

I think your suggestion is reasonable. The KVPool key should indeed be backward compatible, which will require a proper design proposal. However, the issue this PR aims to fix feels more like an initial design flaw—the arbitrary design of the keys led to inconsistent formats. We probably need to address this first to prevent further impact.

LCAIZJ (approved):

PR #15592, opened 2026-09-02
[BugFix][Mamba] Overlap-safe state copies for align-mode block moves

Liears:
## Problem

The NPU path binds upstream's pointer-based `collect_mamba_copy_meta` and executes align-mode mamba state moves through the Ascend triton `batch_memcpy` port, which **dropped upstream's `is_left_overlap` guard** (`vllm/v1/worker/mamba_utils.py:639-652`). Same-pool moves can have overlapping src/dst ranges; without the guard the kernel is a program-order race whose corruption lands in the KV/mamba pools and is consumed by later kernels — the EZ9999 *"MTE accesses an invalid GM address"* crash family:
- faulting kernel **drifts** between runs (GatherPaKvCache / aiv_all_reduce) — the corruption is read by whoever runs next
- **delayed** manifestation (seconds after the corrupting write, often at the next host sync)
- **block-reuse dependence** (second round / next request after a long sequence completes)
- any host sync (per-step, prep-end, launch-blocking) masks it — classic async write-vs-read race

## Fix

1. Collect tensor pairs alongside the pointer buffers (bind the torch collector on the NPU path too) and copy via `dst.copy_(src.clone())` — memmove semantics identical to upstream's guarded kernel. `GLM53_BATCH_MEMCPY=1` restores the raw pointer kernel for A/B.
2. Add upstream's `is_left_overlap` + `tl.debug_barrier()` guard to the fused postprocess copy loops (defensive; today's conv-state shift fits in a single load-then-store block iteration, so it is safe by construction).

## Verification (GLM-5.3-Flash, hybrid KDA + spec=3 + align mode, TP8)

| Case | Before | After |
|---|---|---|
| 33.5k+15k concurrent, round-2 block reuse | crash (6/6 runs, EZ9999) | **pass** |
| 3-way concurrent (33.5k+15k+30k) | crash | **pass** |
| 2000-token decode at high context | crash | **pass** |
| prefix caching | had to be disabled | **re-enabled** |

### DIFF
--- a/vllm_ascend/patch/worker/patch_mamba_utils.py
+++ b/vllm_ascend/patch/worker/patch_mamba_utils.py
@@ -1,6 +1,7 @@
 # mypy: ignore-errors
 
 import itertools
+import os
 from typing import Any
 
 import torch
@@ -18,9 +19,9 @@ from vllm.v1.worker.gpu_input_batch import CachedRequestState
 from vllm.v1.worker.lora_model_runner_mixin import GPUInputBatch
 from vllm.v1.worker.mamba_utils import MambaCopyBuffers
 
-from vllm_ascend.device.hardware_profile import HardwareCapability, get_current_hardware_profile
 from vllm_ascend.ops.triton.batch_memcpy import batch_memcpy_kernel
 from vllm_ascend.ops.triton.mamba.postprocess import postprocess_mamba_fused_kernel
+from vllm_ascend.utils import is_310p
 
 # Upstream uses 16 temporal-copy tiles to saturate H100/GB200. K3 already
 # exposes 138 independent state programs per request, while Triton-Ascend
@@ -32,7 +33,7 @@ mamba_utils._TEMPORAL_TILES = 1
 
 
 def _can_launch_triton_batch_memcpy() -> bool:
-    return get_current_hardware_profile().supports(HardwareCapability.TRITON_BATCH_MEMCPY)
+    return not is_310p()
 
 
 def _get_mamba_groups(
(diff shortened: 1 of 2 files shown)


vllm_ascend/patch/worker/patch_mamba_utils.py
```diff
@@ -207,10 +227,7 @@ def _postprocess_mamba_align_gpu_cpu_fallback(
     # block. Preserve that default so the next preprocess keeps the right
     # accept_token_bias when multiple draft tokens were accepted.
     num_accepted_tokens_cpu_tensor[:num_reqs].copy_(num_accepted_tokens_gpu[:num_reqs])
-    # InputBatch rows may be condensed/reused by async scheduling before this
-    # fallback consumes the snapshot. Keep this step's accepted counts
-    # independent from those mutable request rows.
-    num_accepted_tokens = num_accepted_tokens_cpu_tensor
+    num_accepted_tokens = input_batch.num_accepted_tokens_cpu
```
nu11ptr7: This replaces the CPU snapshot (copied before the fallback runs) with a direct read of `input_batch.num_accepted_tokens_cpu` - the removed comment warned that input_batch rows can be condensed/reused by async scheduling before this fallback consumes them. Is the new field guaranteed to hold this step's values?

Liears: Good catch — no, it is not guaranteed. Async scheduling can condense/reuse input_batch rows before this fallback consumes them, so reading the live field was wrong.

Fixed in 87906e2ba: the fallback now reads the per-step snapshot staged into `num_accepted_tokens_cpu_tensor` (copied from the GPU accepted counts right above), which is the same source of truth the fused kernel consumes. The `preprocess_mamba` reads of `input_batch.num_accepted_tokens_cpu` are untouched — those run before the scheduler can mutate the rows, matching upstream semantics.

PR #15600, opened 2026-09-02
[BugFix][DCP] Fix MLA chunked context lengths for DSpark drafting

q664171689:

Fixes MLA DCP chunked-context metadata construction during DSpark drafting.

`num_computed_tokens_of_dcp` may carry the full draft KV length in speculative metadata, while MLA chunked KV-cache reorganization must use the already-computed context length. This mismatch can cause DCP ranks to build inconsistent sequence-length metadata and fail during MLA KV-cache reorganization.

This PR derives DCP-local chunked-context lengths from `context_lens_cpu`, imports the shared helper explicitly, and adds focused coverage for non-uniform context lengths.

No new API or configuration is introduced.

It fixes a runtime failure for MLA + DCP + DSpark speculative decoding workloads.

- Ran `git diff --check`.
- Ran Python syntax checks for:
  - `vllm_ascend/attention/context_parallel/mla_cp.py`
  - `tests/ut/attention/test_mla_cp.py`
- Verified the shared DCP sequence-length helper imports successfully.
- Manually validated a four-node Kimi K3 deployment (`TP=16`, `DP=4`, `DCP=16`, DSpark): a `/v1/chat/completions` request completed successfully after the fix, with no MLA-DCP assertion, `NameError`, NPU error, or engine failure.

`pytest` was not available in the deployment container, so the focused unit test was not executed there.

### DIFF
--- a/vllm_ascend/attention/context_parallel/mla_cp.py
+++ b/vllm_ascend/attention/context_parallel/mla_cp.py
@@ -88,10 +88,7 @@ class AscendMlaDCPMetadataBuilder(
         if chunked_context_metadata is None:
             return None
 
-        local_context_lens_allranks = self._get_dcp_context_lens(
-            common_attn_metadata,
-            start=self.num_decodes,
-        )
+        local_context_lens_allranks = self._get_dcp_local_context_lens_allranks()
         padded_local_context_lens_cpu = (
             cdiv(self.context_lens_cpu, self.cp_virtual_block_size) * self.cp_local_block_size
         )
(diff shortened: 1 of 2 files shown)


vllm_ascend/attention/context_parallel/mla_cp.py
```diff
@@ -139,6 +137,23 @@ def build_chunked_metadata(
             chunk_size=padded_local_max_context_chunk_across_ranks,
         )
 
+    def _get_dcp_local_context_lens_allranks(self) -> torch.Tensor:
```
nu11ptr7, line 140: The new helper derives local lengths from `context_lens_cpu`, while the old path started at `self.num_decodes` - if context_lens_cpu also covers in-flight decode requests, their context would fold into the chunked-prefill layout. Worth confirming the tensor only holds the requests being loaded.

zhenwenqi2024 (approved):

PR #15613, opened 2026-09-03
[BugFix][MoE] Keep topk weights in float dtype when hidden states arrive quantized (sequence-parallel MoE + W4A8/W8A8)

dsxsteven:

The MoE sequence-parallel prepare step (`_prepare_with_ep_group`) quantizes `hidden_states` for dynamic-quant schemes **before routing**: `W8A8 -> int8`, `W8A8MXFP/W4A8MXFP -> float8_e4m3fn`, `W4A4MXFP -> uint8` (fp4 packs). `AscendRoutedExperts._select_experts` then propagated that dtype to `topk_weights`, which:

- breaks `topk_weights * mask` in the token dispatcher (fp8 x bool promotion is unsupported),
- silently truncates int8 weights (0..1 values become 0) — a live correctness bug on main for SP-MoE + W8A8,
- violates `npu_moe_token_unpermute`'s fp32 `probs` expectation.

Changes:
- `routed_experts.py`: only follow `hidden_states.dtype` for floating compute dtypes (fp32/fp16/bf16); otherwise fall back to `float32`. This whitelist subsumes the previous uint8/float8_e4m3fn skip added for MXFP4 and covers the int8 case the skip missed.
- `grouped_topk_router.py`: apply the same guard — without it the renormalization and `routed_scaling_factor` multiply would run in a quantized dtype.
- Add a UT covering the dtype contract for fp32/fp16/bf16 and quantized hidden_states dtypes.

No API/config change. Fixes a crash and a silent-precision bug when activations arrive pre-quantized from the MoE sequence-parallel prepare step.

```
pytest tests/ut/ops/test_routed_experts.py
# 16 passed (10 existing + 6 new parametrized dtype cases)
```

Validated on A5 nodes with dsa_cp + W4A8/W8A8 quantized models.

Signed-off-by: dsxsteven <dsxsteven@sina.com>

### DIFF
--- a/vllm_ascend/ops/fused_moe/routed_experts.py
+++ b/vllm_ascend/ops/fused_moe/routed_experts.py
@@ -476,10 +476,18 @@ class AscendRoutedExperts(RoutedExperts):  # type: ignore[no-redef]
             )
             topk_ids = torch.cat([topk_ids, shared_expert_ids], dim=1)
             topk_weights = torch.cat([topk_weights, shared_expert_weights], dim=1)
-        # MXFP4 packs activations as uint8; skip the cast so topk_weights stays
-        # fp32, which is what npu_moe_token_unpermute expects for its `probs` arg.
-        if hidden_states.dtype not in [torch.uint8, torch.float8_e4m3fn]:
+        # Routing weights must stay in a floating compute dtype: the MoE
+        # sequence-parallel prepare (`_prepare_with_ep_group`) quantizes
+        # hidden_states for dynamic-quant schemes before routing (W8A8 ->
+        # int8, W8A8MXFP/W4A8MXFP -> float8_e4m3fn, W4A4MXFP -> uint8 fp4
+        # packs). Propagating that dtype to topk_weights breaks downstream
+        # ops (e.g. `topk_weights * mask` in the token dispatcher: fp8 x bool
+        # promotion is unsupported), and npu_moe_token_unpermute expects fp32
+        # `probs`.
+        if hidden_states.dtype in (torch.float32, torch.float16, torch.bfloat16):
             topk_weights = topk_weights.to(hidden_states.dtype)
+        else:
+            topk_weights = topk_weights.to(torch.float32)
         if get_ascend_config().enable_force_eplb:
             topk_ids = get_force_eplb_topk(topk_ids, num_logical_experts)
         elif enable_force_load_balance:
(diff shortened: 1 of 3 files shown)


yiz-liu (dismissed):

dsxsteven:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/ops/fused_moe/routed_experts.py
```diff
@@ -476,10 +476,18 @@ def _select_experts(
...
+        # Routing weights must stay in a floating compute dtype: the MoE
+        # sequence-parallel prepare (`_prepare_with_ep_group`) quantizes
+        # hidden_states for dynamic-quant schemes before routing (W8A8 ->
+        # int8, W8A8MXFP/W4A8MXFP -> float8_e4m3fn, W4A4MXFP -> uint8 fp4
+        # packs). Propagating that dtype to topk_weights breaks downstream
+        # ops (e.g. `topk_weights * mask` in the token dispatcher: fp8 x bool
+        # promotion is unsupported), and npu_moe_token_unpermute expects fp32
+        # `probs`.
+        if hidden_states.dtype in (torch.float32, torch.float16, torch.bfloat16):
             topk_weights = topk_weights.to(hidden_states.dtype)
```
lcfenglinwan: if the conversion here can probably be removed?

vllm_ascend/ops/fused_moe/router/grouped_topk_router.py
```diff
@@ -119,7 +119,14 @@ def _compute_routing(
             topk_weights = topk_weights + self.e_score_correction_bias
 
         topk_weights, topk_ids = topk_weights.topk(self.top_k, dim=-1)
-        topk_weights = topk_weights.to(hidden_states.dtype)
+        # Keep routing weights in a floating compute dtype: the MoE
+        # sequence-parallel prepare (`_prepare_with_ep_group`) may have already
+        # quantized hidden_states (int8 / float8_e4m3fn / uint8), and
+        # quantized weights break downstream ops and the renormalization below.
+        if hidden_states.dtype in (torch.float32, torch.float16, torch.bfloat16):
+            topk_weights = topk_weights.to(hidden_states.dtype)
```
lcfenglinwan: same as before

Issue #15353, opened 2026-08-30
[RFC]: Refactor shared expert multistream overlap scheduling


zhenwenqi2024:
### Motivation.

The current shared-expert multistream implementation overlaps shared-expert computation with routed-expert execution, but the scheduling contract is expressed through communication-specific event names and implicit layout assumptions. This makes the code difficult to reason about and leaves different overlap opportunities on the AllGather and AllToAll paths.

The main problems are:

1. **Ambiguous synchronization semantics.** Events such as “before/after routed experts” or “before dispatch” describe implementation positions rather than the data dependency they guarantee. Shared-expert code must know routed-expert internals to select the correct wait point.
2. **Different overlap windows for AllGather and AllToAll.**
   - On the AllGather path, routed-input gathering happens during routed-expert preparation. Starting shared Gate-Up only at dispatch means that the AllGather latency cannot be covered.
   - On the AllToAll path, the forward expert exchange begins after dispatch. The existing schedule already covers much of this communication, but it does not cover local preparation, top-k processing, and dispatch preprocessing.
3. **Quantization scheduling is mixed with stream orchestration.** W8A8/W4A8, W4A8 MXFP, other MXFP/FP formats, non-quantized execution, and active LoRA do not have the same operator pipeline or Cube/Vector resource profile.
4. **Input/output layout is inferred indirectly.** Sequence-parallel and shared-expert-DP layouts are currently coupled to event availability and call order instead of being represented explicitly.
5. **Overly broad synchronization.** Whole-stream waits can serialize unrelated work and make the actual dependency unclear.
6. **Collective ordering is fragile.** Shared-expert output collectives must not overtake routed-expert collectives on the same communication domain, otherwise performance regressions or deadlocks are possible.
7. **AIV resource contention is not explicit.** With AIV enabled by default on Ascend, communication may consume Vector resources. Scheduling shared Vector-heavy quantization directly against AIV communication can reduce or reverse overlap benefits.
8. **Version-dependent router handling is duplicated.** Similar control flow is repeated for supported upstream vLLM interfaces, which increases maintenance cost.

This RFC is intentionally narrower than #13220. That RFC discusses the broader Ascend fused-MoE architecture alignment; this proposal focuses on the shared-expert multistream dependency model, communication-backend behavior, and quantization-aware scheduling. It also contributes to the shared-expert multistream optimization item tracked by #13452.

#### 1. Introduce backend-neutral routed-MoE milestones

[...]

PR #15352, opened 2026-08-30
[Refactor][MoE] Clarify shared expert multistream scheduling

zhenwenqi2024:
### What this PR does / why we need it?

Fixes #16446
Refs #15353
Related fix: #15500

This PR refactors shared-expert multistream overlap around backend-neutral routed-MoE milestones instead of communication-specific wait points, and fixes the SP collective ordering regression exposed by that refactor.

The existing implementation can overlap shared-expert computation with routed experts, but its event semantics are tied to internal code positions. This makes the dependency chain difficult to maintain and creates different behavior for AllGather and AllToAll:

- **AllGather:** routed-input gathering occurs during routed preparation. Waiting until dispatch to start eligible shared Gate-Up work leaves that communication exposed.
- **AllToAll:** the forward exchange starts around dispatch and is already substantially overlapped, but local preparation, top-k processing, and dispatch preprocessing remain outside the shared Gate-Up overlap window.

#### Main changes

- Introduce explicit `PreparedSharedExpertInput` and `RoutedMoEMilestones` data classes.
- Replace positional event semantics with explicit milestones:
  - shared input ready;
  - router output ready;
  - routed dispatch start;
  - routed GMM2 start;
  - routed combine start;
  - routed finalize done.
- Start eligible MXFP and registered-linear-wrapper Gate-Up work after router output so it can cover routed AllGather preparation or AllToAll preprocessing and dispatch.
- Preserve dedicated W8A8/W4A8 and W4A8 MXFP shared-expert pipelines.
- Keep other quantization schemes and active LoRA on their registered linear wrappers.
- Keep SP-only shared collectives ordered after routed finalization.
- Keep the deferred Kimi SP output join graph-safe with stream synchronization across the opaque custom-op boundary.
- Consolidate duplicated internal-router handling across supported vLLM interfaces.
- Add fail-fast checks for incomplete or invalid milestone/input contracts.

#### SP + multistream deadlock fix

With FlashComm1 enabled, shared-expert DP disabled, multistream overlap enabled, and DP greater than one, the no-routed-input-transform path could start the TP-group shared-input AllGather on the auxiliary shared-expert stream while the default stream entered routed EP-group collectives. Different ranks could therefore enqueue cross-group HCCL collectives in incompatible orders and hang during the startup profile run, eventually surfacing as error 507034.

[...]

PR #15616, opened 2026-09-03
[Feature][Model][MOE] Align shared activation with routed GMM2

slipfre:

Related to #15353.

This PR aligns shared-expert activation with routed-expert GMM2 in the existing multistream schedule:

- record the routed `before_gmm2` event in the unquantized MoE MLP path after activation and before the down projection;
- propagate an explicit recording flag so the event is created only when shared-expert multistream overlap is enabled;
- split the generic shared-expert path into Gate-Up, activation, and Down stages;
- wait for `before_gmm2` before shared activation so its Vector work can overlap routed GMM2 Cube work;
- keep the shared Down projection behind the existing combine/finalize boundary;
- make the MLP tuple return contract explicit in its type annotations;
- update coverage for event placement, disabled-path behavior, and shared-stage ordering, including the 310P contract.

This is a deliberately narrow alternative to #15352. That PR introduces new prepared-input and semantic-milestone contracts across the routed/shared MoE implementation. This PR preserves the existing `FusedMoEEvents` and call contracts and changes only the scheduling boundary needed for activation/GMM2 overlap. It does not change the dedicated W8A8/W4A8 or W4A8 MXFP pipelines; quantization schemes and LoRA configurations that use the generic linear-wrapper path inherit the same split scheduling.

No NPU performance improvement is claimed without device profiling.

Duplicate-work checks searched open PRs for `shared activation`, `before_gmm2`, and `shared expert overlap`. #15352 is the only PR that contains the same overlap goal, with the broader implementation difference described above.

No. There is no intended API, configuration, checkpoint-format, or numerical behavior change. The change only adjusts internal stream scheduling when shared-expert multistream overlap is enabled.

Static checks completed locally:

- Python syntax compilation passed for all changed Python files.
- `git diff --check` passed.

Tests were added or updated in:

- `tests/ut/ops/test_moe_mlp.py`
- `tests/ut/ops/test_moe_comm_method.py`
- `tests/ut/ops/test_fused_moe.py`
- `tests/ut/_310p/fused_moe/test_shared_fused_moe_310.py`
- `tests/e2e/nightly/single_node/ops/singlecard_ops/test_fused_moe.py`

The pytest cases were not executed on this Windows host because the workspace has no project pytest environment. Ascend NPU correctness, timeline, and performance validation have not been run and remain required before merge.

### AI assistance

AI assistance was used to prepare this change. The human submitter is responsible for reviewing and understanding every changed line and for completing the required NPU validation.

### DIFF
--- a/tests/ut/_310p/fused_moe/test_shared_fused_moe_310.py
+++ b/tests/ut/_310p/fused_moe/test_shared_fused_moe_310.py
@@ -205,18 +205,17 @@ class _Gate(nn.Module):
 @pytest.mark.parametrize("with_gate", [False, True])
 def test_shared_experts_part2_310_applies_optional_gate(with_gate):
     shared_experts_layer = SimpleNamespace(
-        act_fn=nn.Identity(),
         down_proj=_Projection(),
         expert_gate=_Gate() if with_gate else None,
     )
     shared_experts = AscendSharedExperts.__new__(AscendSharedExperts)
     shared_experts.layer = shared_experts_layer
     hidden_states = torch.randn(3, 4)
-    shared_gate_up = torch.randn(3, 4)
+    shared_act = torch.randn(3, 4)
 
-    output = shared_experts.part2(hidden_states, shared_gate_up)
+    output = shared_experts.part2(hidden_states, shared_act)
 
-    expected = shared_gate_up * 2.0 + 1.0
+    expected = shared_act * 2.0 + 1.0
     if with_gate:
         expected = expected * 0.5
     torch.testing.assert_close(output, expected)
--- a/tests/ut/ops/test_fused_moe.py
+++ b/tests/ut/ops/test_fused_moe.py
@@ -859,18 +859,17 @@ class _Gate(nn.Module):
 @pytest.mark.parametrize("with_gate", [False, True])
 def test_shared_experts_part2_applies_optional_gate(with_gate):
     shared_experts_layer = SimpleNamespace(
-        act_fn=nn.Identity(),
         down_proj=_Projection(),
         expert_gate=_Gate() if with_gate else None,
     )
     shared_experts = AscendSharedExperts.__new__(AscendSharedExperts)
     shared_experts.layer = shared_experts_layer
     hidden_states = torch.randn(3, 4)
-    shared_gate_up = torch.randn(3, 4)
+    shared_act = torch.randn(3, 4)
 
-    output = shared_experts.part2(hidden_states, shared_gate_up)
+    output = shared_experts.part2(hidden_states, shared_act)
 
-    expected = shared_gate_up * 2.0 + 1.0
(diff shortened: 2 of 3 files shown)


vllm_ascend/ops/fused_moe/moe_mlp.py
```diff
@@ -768,7 +770,7 @@ def unquant_apply_mlp(
             silu_out=gate_up_out,
             lora_routing=lora_routing,
         )
-    return hidden_states, None
+    return hidden_states, before_gmm2_evt
```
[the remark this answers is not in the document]
slipfre: Fixed in a17540212. Updated `quant_apply_mlp`, `unquant_apply_mlp`, `unified_apply_mlp`, and `MoECommMethod._apply_mlp` to use the `tuple[Tensor, Event | None]` contract. Also corrected the existing nightly caller to unpack the tuple. The event is now explicitly gated so the unquantized path does not create it when shared-expert multistream overlap is disabled.

vllm_ascend/ops/fused_moe/routed_experts.py
```diff
@@ -47,7 +46,9 @@ class AscendUnquantizedFusedMoEMethod(UnquantizedFusedMoEMethod):
     def __init__(self, moe: FusedMoEConfig = None, tid2eid=None):
         super().__init__(moe=moe)
         vllm_config = get_current_vllm_config()
-        self.dynamic_eplb = False if vllm_config.use_v2_model_runner else get_ascend_config().eplb_config.dynamic_eplb
+        ascend_config = get_ascend_config()
+        self.dynamic_eplb = False if vllm_config.use_v2_model_runner else ascend_config.eplb_config.dynamic_eplb
```
Angazenn: This line of change seems redundant

vllm_ascend/ops/fused_moe/routed_experts.py
```diff
@@ -47,7 +46,9 @@ class AscendUnquantizedFusedMoEMethod(UnquantizedFusedMoEMethod):
...
+        self.record_before_gmm2 = ascend_config.multistream_overlap_shared_expert
```
Angazenn: I think we do not need to add this flag? if multistream_overlap_shared_expert is disabled, shared_expert is normally after routed_expert

vllm_ascend/ops/fused_moe/routed_experts.py
slipfre: fixed

vllm_ascend/ops/fused_moe/routed_experts.py
slipfre: yes, the redundant flag is removed.

Angazenn (approved):

realliujiaxu (approved):
LGTM

PR #15636, opened 2026-09-03
[Performance][AscendStore] Reduce PP save synchronization and deepcopy overhead

LCAIZJ:

This PR optimizes the PP + AscendStore execution path:

1. Replaces the full `deepcopy` of `SchedulerOutput` with targeted `dataclasses.replace`, avoiding unnecessary copies of large fields such as KV connector metadata.
2. Removes the same-step `request_queue.join()` from `wait_for_save()`. KV saves are queued asynchronously and synchronized at the next step before KV cache blocks can be reused, preserving KV cache correctness.

No.

Existing AscendStore unit tests passed:

```text
124 passed, 41 subtests passed

```

Co-authored-by：@zmc1997 


vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
```diff
@@ -1732,10 +1736,31 @@ def save_kv_layer(self, connector_metadata: AscendConnectorMetadata) -> None:
...
+    def wait_for_previous_save(self) -> None:
+        save_batch = self._previous_save_batch
+        if save_batch is None:
+            return
+
+        assert self.kv_send_thread is not None
+        send_thread = self.kv_send_thread
+        wait_start = time.perf_counter()
+        while True:
+            send_thread.raise_if_failed()
```
Pz1116, line 1748: Is this a fatal error that will raise when kv put failed? If so should this be non-fatal?

LCAIZJ: `raise_if_failed()` only propagates an unexpected exception that escapes the request handler and terminates the sending thread.

Pz1116 (approved):

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
Pz1116: LGTM

weijinqian0 (approved):

PR #15640, opened 2026-09-03
[Ops][Feature] Add overflow guards and fallback to all2all for MegaMoe and dispatch_ffn_combine

coder-fny:
**What this PR does / why we need it?**

The fused-MC2 MoE path on A3 currently has two buffer overflow points that can cause failures, neither of which is handled well:

- **mega_moe** (`enable_fused_mc2 == 2`): The symm buffer size is set by the reference value `mega_moe_max_tokens` (default 65536). When the worst-case post-dispatch received token count per rank exceeds this value, tokens are silently truncated — no error, just precision degradation.
- **dispatch_ffn_combine** (`enable_fused_mc2 == 1`): The HCCL communication buffer size is determined by the `HCCL_BUFFSIZE` environment variable. The C++ tiling code explicitly checks `M×topK×K×3 > HCCL_BUFFSIZE` and raises an error (`GRAPH_FAILED`), which aborts the run.

This PR adds framework-side guards in `_select_a3_moe_comm_method` that, before the operator is ever invoked, fall back to ALLTOALL for both cases (the all2all path uses the dynamic `torch.distributed` buffer, which needs neither `HCCL_BUFFSIZE` nor `mega_moe_max_tokens`):

- **mega_moe** → a new `mega_moe_threshold_ratio` config knob (float, default 1.0): when `worst_case_recv × ratio > mega_moe_max_tokens`, fall back to all2all. Decode-only nodes are exempt (they size the symm buffer to the absolute safe bound). Set `< 1` to accept truncation risk; set `> 1` to require additional margin.
- **dispatch_ffn_combine** → replicates the C++ `HCCL_BUFFSIZE` check on the framework side (`M×topK×K×3 + 10MB > HCCL_BUFFSIZE`, where `M = ceil(num_tokens/tp)` is the per-TP-rank operator input, verified against the C++ `info.M`): turns the hard error into a graceful fallback.

This PR also includes:

- **Bug fix**: The `use_cann_megamoe` EP guard condition was `<= 64`; it should be `<= 32` (both dispatch_ffn_combine and mega_moe support ep `<= 32`). This aligns it with the merged selection guard, so that the `use_mega_moe` flag and `should_skip_allreduce_across_dp_group` are consistent with the selection result (eliminating the leftover-flag edge case for `32 < ep <= 64`).
- **Refactor**: Merges the two redundant `FUSED_MC2` selection branches into one, with a sub-branch that selects the matching guard based on buffer type (mega_moe = symm buffer → ratio check; dispatch_ffn_combine = HCCL buffer → HCCL check). The execution distinction between mega_moe and dispatch_ffn_combine is still preserved in `FusedMC2CommImpl.fused_experts` via `_EXTRA_CTX.use_mega_moe`.
- `HCCL_BUFFSIZE` is centralized in `vllm_ascend/envs.py` per the environment-variable convention.

**Does this PR introduce any user-facing change?**

Yes. A new `additional_config` option `mega_moe_threshold_ratio` (float, default 1.0) is added. Users on mega_moe (`enable_fused_mc2 == 2`) experiencing silent precision issues can now choose to increase `mega_moe_max_tokens` or decrease `mega_moe_threshold_ratio`, without having to manually compute safe bounds. For the `dispatch_ffn_combine` path, a previously set-too-small `HCCL_BUFFSIZE` would hard-error at runtime; it now falls back to all2all gracefully with a warning.

**How was this patch tested?**

- Unit tests (`tests/ut/test_ascend_config.py`): default value, override, range validation (`0` / `-1` / `-0.5` → `ValueError`), and lenient float coercion (`"0.5"` → `0.5`) for the new field.
- Unit tests (`tests/ut/test_ascend_forward_context.py`): `_hccl_buffsize_would_overflow` (within limit / over limit / `None` token / per-TP-rank `M = ceil(num_tokens/tp)` for `tp > 1`), `_select_mega_moe_or_all2all` (ratio guard: no overflow / fallback / ratio `< 1` keeps mega_moe / decode-only exempt), and the integrated `_select_a3` routing (mega_moe → ratio, dispatch_ffn_combine → HCCL fallback, no overflow → `FUSED_MC2`).
- `ruff check`, `ruff format --check`, and `py_compile` all pass.

| 路径 | 测试配置 | 结果 |
| --- | --- | --- |
| HCCL 无溢出→FUSED_MC2 (dispatch_ffn_combine) | =1, HCCL_BUFFSIZE=50, decode(8 tok) | m=1, actual=10.1MB<50MB → FUSED_MC2 ✓ |
| HCCL 溢出→ALLTOALL 回退 | =1, HCCL_BUFFSIZE=50, prefill(2720 tok) | m=340, actual=57.8MB>50MB → ALLTOALL, warning 触发 ✓ |
| HCCL 溢出→ALLTOALL 回退 (warmup) | =1, HCCL_BUFFSIZE=50, warmup(16384 tok) | m=2048, actual=298MB>50MB → ALLTOALL ✓ |
| 合并分支结构 (单 FUSED_MC2 + sub-branch) | eager + graph | 服务稳定运行 ✓ |
| EP guard ≤32 | ep=32 | 分支正常进入 ✓ |
| mega_moe ratio 检查逻辑 | =2, cap=4096 | wc=524288>4096 → overflow=True → ALLTOALL ✓ |
| mega_moe 路径激活 (use_cann_megamoe=True) | =2, CANN PYTHONPATH | cann_ops 找到, mega_moe 路径走通 ✓ |


coder-fny:
image
image
mega_moe path (`enable_fused_mc2=2`) builds a dedicated per-expert int8+NZ weight list (`cann_mega_moe_w13_weight_list`) at load time and **deletes the original single-tensor `layer.w13_weight`** (`w4a8.py:354-369`); the per-forward `apply()` then selects this mega_moe-specific weight format based on `_EXTRA_CTX.use_mega_moe`, which is a **config-level flag** (set once from `use_cann_megamoe(vllm_config)`, not re-evaluated per-forward comm method). When the `mega_moe_threshold_ratio` check triggers an ALLTOALL fallback, `use_mega_moe` remains `True`, so `apply()` still hands the mega_moe-formatted per-expert list to the non-fused GMM (`unified_apply_mlp` → `aclnnGroupedMatmulSwigluQuantWeightNzV2`), which expects a single `[hidden, 2×intermediate]` tensor — causing the `AclNN_Parameter_Error` shape mismatch (`[6144, 2048]` vs `[6144, 4096]`). The regular weight format no longer exists (deleted at load), so the ALLTOALL path has no compatible weights to fall back to, making the mega_moe→ALLTOALL fallback fundamentally incompatible with the current weight lifecycle. Properly fixing this incompatibility depends on first landing a weight-format-unification PR that makes the mega_moe and non-fused paths share a single weight format, so the fallback can work without format-specific preparation.

PR #15646, opened 2026-09-03
[BugFix][P/D] Sync Mooncake block size from workers

moonseeker1:

DeepSeek-V4 heterogeneous KV cache initialization rewrites `cache_config.block_size` in EngineCore to the smallest physical group block size. The Mooncake scheduler is created afterwards and snapshots that value, while workers were already spawned with the original logical block size. As a result, a producer configured with block size 128 can send `remote_block_size=8`, and the decoder rejects it because it is not divisible by the attention kernel size.

Mooncake workers already report their effective block size through `MooncakeAgentMetadata` after KV cache registration. This change treats that worker-reported value as the authoritative P/D transfer block size during the existing local worker handshake and validates that every worker reports the same positive value.

The scheduler now keeps `transfer_block_size` separate from `block_size`: the former is used only for `remote_block_size` and `num_prompt_blocks`, while `group_transfer_info` continues to use each KV cache group's own block size. This prevents the worker handshake from changing heterogeneous group geometry such as `tokens_per_block` and SWA `blocks_per_window`.

No P/D wire schema or `cache_config` mutation is introduced. Refs #14920.

No new option or API is introduced. It fixes MooncakeConnectorV1 P/D transfers when EngineCore's scheduler-facing cache block size differs from the block size used by workers.

- Added a scheduler regression test with scheduler block size 8, heterogeneous cache groups using block sizes 8 and 32 (including an SWA group), and worker transfer block size 128.
- Verified the handshake produces `remote_block_size=128` and computes a 129-token prompt as two transfer blocks, while group `tokens_per_block` and `blocks_per_window` remain unchanged.
- Added a failure-mode test for inconsistent worker block sizes.
- `ruff 0.14.0 check` passed for both changed files.
- `ruff 0.14.0 format --check` passed for both changed files.
- `python -m py_compile` passed for both changed files.
- The unit test was not executed on the Windows development host because the environment does not provide pytest, torch, or torch_npu; selected-test CI is expected to run it.

### DIFF
--- a/tests/ut/kv_offload/test_mooncake_connector.py
+++ b/tests/ut/kv_offload/test_mooncake_connector.py
@@ -19,6 +19,7 @@ from vllm.v1.kv_cache_interface import (
     FullAttentionSpec,
     KVCacheGroupSpec,
     MLAAttentionSpec,
+    SlidingWindowSpec,
     UniformTypeKVCacheSpecs,
 )
 from vllm.v1.request import RequestStatus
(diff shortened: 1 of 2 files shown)


moonseeker1:
Addressed the group_transfer_info concern by separating the two block-size semantics instead of rebuilding group geometry. self.block_size remains the scheduler/cache-group value, while the worker handshake updates a dedicated self.transfer_block_size used only by 
um_prompt_blocks and 
emote_block_size. The regression test now uses real heterogeneous KVCacheGroupSpecs (block sizes 8 and 32, including SWA) and verifies their 	okens_per_block and locks_per_window remain unchanged when workers report transfer block size 128.

Issue #14143, opened 2026-08-13
[Contribution] 任务 #2：[Arch/Perf] 传输路径改 IPC，消除 GIL 瓶颈


chengduxiaowu:
## 背景

`ascend_store` 的 save/load 当前通过进程内 daemon 线程完成（`KVCacheStore*Sending/RecvingThread`），主线程与 send/recv 线程之间用 `queue.Queue` + `threading.Event` 协作。每进程最多 1 个 send 线程 + 1 个 recv 线程（`_start_kv_transfer_threads` 在三种模式间二选一）。部分热路径在 Python 层执行：`PoolKey.to_string`、`_handle_stored_request` 内 addrs/sizes 构建、key 生成器循环，与 forward 的 Python 层代码共享同一进程的 GIL。

随着多 rank 并行传输、独立崩溃隔离、NPU stream/event 跨进程协作等需求出现，单进程多线程架构在扩展性上存在瓶颈。

相关代码路径：`vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store`。

## 任务

将传输路径由"进程内多线程 + queue.Queue"改造为"独立传输子进程 + IPC"，使 key 构建、buffer 调度、I/O 提交在独立子进程执行，不再与 worker 主进程共享 GIL。

* 范围：三种传输模式（非 layerwise / key layerwise / GVA layerwise）
* 改造范围较大，建议分阶段推进，阶段划分与方法由开发者自定

### 1. 功能正确性

* 三种传输模式功能与改造前一致
* greedy / non-greedy 输出一致（精度无回归）
* 子进程异常不拖垮 worker 主进程，且能被重启或降级回进程内线程模式

### 2. 性能验证

* 附改造前后对比数据：吞吐 / 延迟 / key 构建 CPU 占比 / IPC 开销

### 3. 交付件

* 设计文档（含方案选型、风险清单、回退预案）
* 原型对比报告
* PR + 设计说明 + 单测（覆盖子进程启停、异常恢复、跨进程 buffer 共享、IPC 通道断连）

## 环境约定
- vllm-ascend：最新 main
- 硬件：Ascend NPU（注明型号 + 卡数 + TP/CP/PP 配置）
- 关联任务池：#9079
- 验收人：@赵鹏博

## 重点关注

* NPU KV cache buffer 跨进程共享是核心难点
* layerwise 模式的 NPU Event 跨进程语义需重新设计（`AttentionComputeStartGate`）
[...]

PR #15652, opened 2026-09-03
[Refactor][AscendStore] Run AscendStore KV cache services in an independent multiprocess server

ChenZhuo888:
Implements the direction of #14143: move the KV transfer hot path (key
building, buffer scheduling, I/O submission) out of the engine process so
it no longer contends for the forward path's GIL.

AscendStore's Scheduler/Worker services currently run inside the vLLM engine
process as daemon threads coordinated by queues and events. This PR adds an
opt-in multiprocess mode where they run in a dedicated KV cache server
process, decoupling the cache service lifecycle from engine lifetime and
letting multiple engine instances attach to one external pool service.

- `mp/rpc`: ZMQ transport layer — framing, routing, deadlines, per-lane
  execution, server health and shutdown. Owns transport-level errors only;
  carries no KVCache semantics.
- `mp/service`: service lifecycle — sessions, heartbeat-driven leases,
  expiry cleanup, forced-shutdown reporting.
- `mp/kv_cache`: business layer — wire protocol, client, server and manager
  (Lookup / Store / Retrieve orchestration and degradation policy compatible
  with in-process AscendStore), registration, scheduler view; pool
  scheduler/worker reuse the in-process `KVPoolScheduler` / `KVPoolWorker`
  semantics via inheritance; Mooncake / Memcache / YuanRong backend adapters
  own their own resource lifecycle; NPU IPC path for cross-process buffer
  sharing.
- `AscendStoreMPConnector`: translates vLLM v1 scheduler/worker callbacks
  into MP client operations. Connector registration routes
  `AscendStoreConnector` / `MooncakeConnectorStoreV1` to the MP class when
  `VLLM_ASCEND_STORE_MULTIPROCESS=1`; default `0` keeps in-process behavior
  unchanged (rollback path).
- Standalone server CLI: `vllm-ascend kv-cache-server [--bind-url ...]`,
  default URL from `VLLM_ASCEND_STORE_SERVER_URL`
  (default `tcp://127.0.0.1:5555`). Exit codes: 0 clean shutdown, 130 forced
  termination, 1 startup/config errors.
- Rebased onto current main and adapted to upstream refactors
  (`get_layerwise_reuse_config` rename, `use_layerwise_transfer`,
  `get_group_cache_family`, `AscendConnectorMetadata` signature, scheduler
  `__init__` changes).

Part of #14143. Follow-up deliverables under that RFC (design doc and
before/after performance data) will land separately.

Yes, opt-in:

- New env switch `VLLM_ASCEND_STORE_MULTIPROCESS` (default `0`, in-process
  behavior unchanged when off).
- New env `VLLM_ASCEND_STORE_SERVER_URL` and new CLI subcommand
  `vllm-ascend kv-cache-server`.
- User guide for the standalone server will follow with the design doc.

All items below on real Ascend hardware (910B4 / A3).

- Unit tests: 323 passed (`tests/ut/distributed/ascend_store/` mp suite +
  connector + layerwise layout), including explicit cross-session isolation
  cases (`test_lookup_isolated_by_engine_and_data_parallel_rank`,
  `test_new_worker_session_accepts_a_new_cache_mapping`,
  `test_manager_expires_idle_worker_while_lookup_renews_scheduler`,
  `test_lookup_returns_miss_after_coordinator_unregisters`).
- e2e: one-card IPC, two-card cross-chip IPC (`multi_npu`), memcache backend
  e2e, and a real-model
  external hit + retrieve run through `vllm serve` (external prefix cache
  hit rate ~47%).
- Standalone CLI: `vllm serve` full chain with env-based mode selection and
  URL fallback; multi-instance shared-pool behavior spot-checked (two
  engines on one server, cross-instance reuse verified correct for
  identical model + layout).
- Exit codes: 0 / 130 verified manually, 1 covered by `test_server_cli`.

Not run: YuanRong e2e and RFC before/after performance benchmarks; both
tracked as follow-ups below.

### Known Limitations

1. **External pool entries follow engine sessions** (by design of the
   service lifecycle): the pool is shared among coexisting instances, and
   entries are released when the worker session that owns them ends —
   restarting an engine or the server clears its entries. In-process mode
   leaves entry lifetime to the backend store instead; decoupling entry
   lifetime from sessions (e.g. backend-managed expiry) is a future
   evolution direction.
2. **A3 serve×MP fails with error 507001** (NPU Event synchronization in
   serve mode; identical topology green on A2). This is the cross-process
   event-semantics area called out in #14143; under investigation.

### Follow-ups

1. **Design doc and before/after performance data** (throughput / latency /
   key-build CPU share / IPC overhead, in-process vs multiprocess) per the
   #14143 acceptance criteria.
2. **YuanRong backend e2e validation**.


Pz1116:
Thanks for contributing! I think the currently this PR modifies ascend_store a little bit too much. The goal is to change KVCacheStore Sending/RecvingThread into sub-processes to reduce GIL overhead, so the modification should be limited to kv transfer and pool worker only, with a few extra helpers. Can we simplify this PR a little bit?

ChenZhuo888:
> Thanks for contributing! I think the currently this PR modifies ascend_store a little bit too much. The goal is to change KVCacheStore Sending/RecvingThread into sub-processes to reduce GIL overhead, so the modification should be limited to kv transfer and pool worker only, with a few extra helpers. Can we simplify this PR a little bit?

@Pz1116 Thanks for the clear direction — agreed, let's scope this back to the RFC's core goal.

Plan:

- I'm converting this PR to Draft and keeping it as a full-architecture prototype (standalone multi-process KV cache server, validated on 910B4/A3 hardware), for a possible future discussion on multi-instance deployments — not for review now.

- A new minimal PR will follow, limited to kv_transfer.py + pool_worker.py + a few helpers: the KVCacheStore Sending/RecvingThreads move into a worker subprocess owned by the engine, key building/buffer scheduling/IO submission included; the in-process thread mode stays the default with a switch as the fallback path, per the acceptance criteria.

One design question before I start the rework: for the engine ↔ subprocess channel, do you prefer stdlib multiprocessing primitives (Queue/Pipe/Event — minimal review surface), or is a small dedicated transport acceptable (e.g. ZMQ-based; pyzmq is already a dependency via the mooncake connectors)? The channel only carries metadata — KV data moves via NPU IPC — but it would still set the shape of the helpers.

Happy to share the prototype's measurements (external-hit behavior, cross-process event semantics) for the design doc once the minimal PR is up.

PR #15655, opened 2026-09-03

qijiajin:

Backport of #15510 to `releases/v0.27.1rc`.

Data-parallel ranks with no scheduled requests still execute a one-token dummy batch to participate in DP coordination and collectives. For GDN/Mamba-family recurrent layers, treating that synchronization-only token as a real decode token can advance causal-conv and KDA recurrent state through stale block-table/state indices from an earlier request.

This backport keeps the existing one-token dummy execution semantics unchanged, but makes GDN metadata zero-length for the idle-DP dummy so recurrent state is not updated. Normal requests, non-GDN attention backends, kernels, and user-facing APIs are unchanged.

This PR intentionally does **not** include the separate `uniform_decode` / graph-dispatch change that was previously explored; it only carries the recurrent-state correctness fix from the current #15510.

## Scope

- idle external-launcher DP dummy remains `_dummy_run(1)`
- only that path requests `skip_gdn_state_update=True`
- GDN dummy metadata uses zero `query_start_loc`, `num_actual_tokens=0`, and `max_query_len=0`
- normal GDN/spec-decode metadata behavior is unchanged
- no kernel change and no additional device-wide synchronization

## Validation

The corresponding main-branch fix is #15510 and has been validated with real-weight Kimi K3 multi-node DP/EP workloads as described there. The release branch shares the same idle-DP and GDN metadata code paths, so this is a minimal backport of the same fix.

- Backport target: `releases/v0.27.1rc`
- Main PR: #15510


weijinqian0 (approved):

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

qijiajin:
/rerun
[Bot]: rerun completed.

**Failed:**

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

ningjingbengxiaohai (approved):

PR #15666, opened 2026-09-03
[BugFix][Model] Fix DeepSeek V4 weight update in graph mode

Debonex:

`AscendUnquantizedLinearMethod.process_weights_after_loading` handles the `precast_fp32_weight` path for DeepSeek V4 by assigning a newly created fp32 weight tensor via `layer.weight_fp32 = new_fp32`.

In graph mode (ACLGraph capture), the captured graph holds weight tensors by reference. When the weights are updated (e.g., in RL weight-update scenarios) and `process_weights_after_loading` is called again, replacing the reference means the graph still holds the old tensor, so it keeps using stale weights and produces incorrect results.

This PR updates the `precast_fp32_weight` branch so that, when a `weight_fp32` tensor already exists , the new values are copied in-place via `copy_` instead of replacing the reference. This preserves tensor identity so that graphs captured in graph mode continue to reference the updated weights. All other cases fall back to the original create/assign logic.

No. 

### DIFF
--- a/vllm_ascend/ops/linear.py
+++ b/vllm_ascend/ops/linear.py
@@ -44,6 +44,7 @@ from vllm_ascend.device.hardware_profile import HardwareCapability, WeightLayout
 from vllm_ascend.ops.linear_op import get_parallel_op, get_replicated_op
 from vllm_ascend.utils import (
     maybe_trans_nz,
+    update_tensor_inplace,
 )
 from vllm_ascend.weight_switch import WeightSwitchGatherSpec, WeightSwitchMixin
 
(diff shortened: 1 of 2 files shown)


vllm_ascend/ops/linear.py
```diff
@@ -94,7 +94,18 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
...
+            if (
+                isinstance(old_fp32, torch.Tensor)
+                and old_fp32.data_ptr() != new_fp32.data_ptr()
+                and old_fp32.shape == new_fp32.shape
+                and old_fp32.dtype == new_fp32.dtype
+            ):
+                # RL weight-update path: copy the new weight to the existing fp32 weight tensor
+                old_fp32.copy_(new_fp32)
+            else:
+                layer.weight_fp32 = new_fp32
```
Ronald1995: we should extract a method for this operation. what's more it's better to check `old_fp32.device== new_fp32.device`

```python
def update_tensor_inplace(layer, name, old, new):
    if (
                isinstance(old, torch.Tensor)
                and old.data_ptr() != new.data_ptr()
                and old_fp32.shape == new.shape
                and old.dtype == new.dtype
            ):
                # RL weight-update path: copy the new weight to the existing fp32 weight tensor
                old.copy_(new)
            else:
               setattr( layer, name) = new

```

Debonex: resolved

Ronald1995 (approved):
LGTM. fix error of update weights.

vllm_ascend/utils.py
```diff
@@ -712,6 +712,29 @@ def dispose_tensor(x: torch.Tensor):
     x.set_(torch.empty((0,), device=x.device, dtype=x.dtype))
 
 
+def update_tensor_inplace(layer: torch.nn.Module, name: str, new: torch.Tensor) -> None:
```
linfeng-yuan: Please check whether we already have this sort of function to keep address invariance for RL.

linfeng-yuan (commented):
Please check whether we already have this sort of function to keep address invariance for RL.

vllm_ascend/utils.py
lcfenglinwan: check if there are similar functions

Debonex: resolved, using `replace_parameter` from the upstream vLLM instead.

lcfenglinwan (approved):

Debonex:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

Debonex:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

linfeng-yuan (approved):

PR #15668, opened 2026-09-03
[Feature] Add global expert pool EPLB policy

sokosad:
Add MRv1 policy 4, which keeps the base expert placement immutable and dynamically assigns a small per-rank expert pool across MoE layers based on predicted prefill benefit.\n\nReuse the existing EPLB worker, adaptor, D2D loader, and token dispatcher contracts while keeping the initial per-layer redundant expert count at zero. Document the experimental eager prefill scope and add planner and migration regression coverage.

### DIFF
--- a/tests/ut/eplb/adaptor/test_vllm_adaptor.py
+++ b/tests/ut/eplb/adaptor/test_vllm_adaptor.py
@@ -4,6 +4,7 @@ from unittest.mock import MagicMock, call, patch
 import torch
 from transformers import DeepseekV2Config
 
+from vllm_ascend.ascend_config import EplbConfig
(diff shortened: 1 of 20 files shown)


sokosad:
Static checks, formatting, and mypy now pass. Could a maintainer please add the `ready-precise` label to run the selected tests? The CI gate requires this label to be applied by a maintainer.

sokosad:
CI code checks (`pre-commit`, `main`, `PR create action`, and DCO) pass on c6dc92764. The remaining `ci-gate` failure is only the repository policy requiring a maintainer label for source changes. Could a maintainer please add `ready-precise` to run the selected precision tests? TP16 runtime startup, migration, request completion, and deterministic output equivalence were also validated on Ascend.

PR #15669, opened 2026-09-03
[Refactor][Model] Decouple indexer from SFA with unified forward pipeline

Biuapha:

Decouples the sparse-attention indexer from the SFA attention impl so indexer variants (different cache layout / compression / top-k kernel) can be plugged in without touching SFA:

- Introduces `AscendSFAIndexerBackend` (`vllm_ascend/attention/indexer.py`): owns the whole indexer pipeline in one unified `forward` - k-path (`forward_k`) -> parallel-layout gather (`_gather_cache_inputs`) -> `write_cache` -> top-k selection. `compute_topk=False` keeps the skip_topk layers' cache up to date while reusing shared top-k indices.
- The indexer builds its own attention metadata via `AscendSFAIndexerMetadataBuilder` (slot_mapping / block_table / seq lens / rope tables), instead of borrowing SFA's `attn_metadata`, so a variant indexer with a compressed cache layout gets its own slot mapping.
- Parallel-layout transforms are internal to the backend: PCP all-gathers the prefill region (reordering slot mapping to the gathered layout), DSA-CP all-gathers the indexer k across the TP group. SFA no longer carries indexer-specific gather/write hooks.
- `IndexerWrapper` (`vllm_ascend/ops/mla.py`) instantiates the backend directly, registers the indexer weights on the wrapper so module-tree paths keep the pre-refactor layout (`...indexer.<name>`) that weight loading and quant name mapping key off, and delegates the unified `forward` to the backend.
- Removes the now-dead `ops/indexer/base.py` / `ops/indexer/lightning.py`.

No. Pure internal refactor; no API, config, or behavior change.

- Unit tests updated/added: `tests/ut/attention/test_indexer.py` (metadata builder, unified forward, PCP/DSA-CP gather), `tests/ut/attention/test_sfa_v1.py`, `tests/ut/attention/test_sfa_cp.py`, `tests/ut/ops/test_mla.py`.
- Equivalence review against pre-refactor flow (k-path ordering, cache write timing, skip_topk/index-cache reuse, CP/PCP/DSA-CP gather layouts).
- Two-node deployment on Ascend 950 (PP2TP8EP8, GLM-5.3 W8A8C8 MXFP8): service up with `enable_sparse_sfa_c8`, `enable_sparse_li_c8`, `enable_dsa_cp`, flashcomm1, multistream-overlap; smoke prompts correct in both eager and `FULL_DECODE_ONLY` ACLGraph modes (graph capture passes with DSA-CP enabled).
- GSM8K zero-shot CoT full-set eval (1319 questions, max_tokens=10240) on the dual-node PP2TP8EP8 deployment with MTP5 (`deepseek_mtp`, num_speculative_tokens=5) and `enable_sparse_sfa_c8`/`enable_sparse_li_c8`/`enable_dsa_cp` on:
  - pre-refactor baseline: 1283/1319 = 97.27%
  - with this PR: 1286/1319 = 97.50%
  Accuracy is consistent (delta +0.23%, within run-to-run variance); MTP mean acceptance length 4.00/5.


vllm_ascend/attention/indexer.py
```diff
@@ -61,9 +141,308 @@ def get_kv_cache_shape(
...
+            tensors = (k_li,) if k_li_scale is None else (k_li, k_li_scale)
+            gathered_tensors, slot_mapping = _gather_prefill_cache_inputs(tensors, slot_mapping, indexer_metadata.num_decode_tokens)
+            k_li = gathered_tensors[0]
+            assert slot_mapping.numel() == k_li.shape[0], (
+                "PCP indexer cache write requires one slot per gathered token: "
+                f"tokens={k_li.shape[0]}, slots={slot_mapping.numel()}."
+            )
+            if k_li_scale is not None:
+                k_li_scale = gathered_tensors[1]
+        elif self._dsa_cp_active:
```
lijiahang226, line 326: The DSA-CP all-gather in _gather_cache_inputs() is now synchronous and immediately waited, losing the previous overlap with the main KV all-gather. Consider returning handles and waiting at a later unified point, or explicitly document the intentional serialization.

vllm_ascend/attention/indexer.py
```diff
@@ -61,9 +141,308 @@ def get_kv_cache_shape(
...
+        self.q_lora_rank: int = vllm_indexer.q_lora_rank  # 1536
+        self.wq_b = vllm_indexer.wq_b
+        self.wk_weights_proj = vllm_indexer.wk_weights_proj
+        self.k_norm = vllm_indexer.k_norm
+        self.softmax_scale = vllm_indexer.softmax_scale
+        self.k_cache = getattr(vllm_indexer, "k_cache", None)
+        self.qk_rope_head_dim = qk_rope_head_dim
+        vllm_indexer.topk_indices_buffer = None  # delete topk_indices_buffer
+
+        self.enable_sparse_li_c8 = get_ascend_config().is_sparse_li_c8_layer(self.k_cache.prefix)
```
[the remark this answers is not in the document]
Biuapha: Fixed - the backend now raises an explicit `RuntimeError` at construction if the vLLM indexer module does not expose `k_cache`, instead of a cryptic AttributeError at `.prefix`.

vllm_ascend/attention/indexer.py
Biuapha: Good catch - this serialization is intentional for this PR: with the unified interface the indexer owns its whole cache-write pipeline, so it can no longer join SFA's fused collective (pre-refactor k_li was cat'ed into the main KV gather). Restoring that fusion would re-couple SFA and the indexer through a collective plan, which is exactly what this refactor removes. As a middle ground the two DSA-CP gathers (k and scale) are now launched back-to-back asynchronously and waited together so they overlap each other, and a TODO documents the re-fusion option (passing a collective plan via the indexer metadata) if DSA-CP throughput becomes a concern.

vllm_ascend/attention/indexer.py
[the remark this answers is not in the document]
Biuapha: done

vllm_ascend/attention/indexer.py
Biuapha: done

pisceskkk (approved):

vllm_ascend/device/device_op.py
```diff
@@ -1286,23 +1288,25 @@ def get_dsa_kernel_block_sizes():
...
         attn_metadata,
         actual_seq_lengths_query: torch.Tensor,
         actual_seq_lengths_key: torch.Tensor,
         enable_sparse_li_c8: bool,
         use_torch_npu_lightning_indexer: bool,
     ) -> torch.Tensor:
-        indexer_cache_idx = sfa_impl.kv_cache_indexer_k_idx
-        indexer_scale_cache_idx = sfa_impl.kv_cache_indexer_scale_idx
+        indexer_cache_idx = indexer_k_cache_idx
+        indexer_scale_cache_idx = indexer_scale_cache_idx
```
lcfenglinwan: There is no need to perform another assignment operation here. You can directly use the variables `indexer_scale_cache_idx` or `indexer_k_cache_idx` afterward.

vllm_ascend/ops/mla.py
```diff
@@ -32,33 +32,67 @@
...
+    """Model-facing wrapper owning the per-layer indexer backend.
+
+    Mirrors the wrapper/backend split of AscendMultiHeadLatentAttention: the
+    wrapper wires the upstream weight module into the model tree and
+    dispatches; all compute and cache persistence live in the
+    ``AscendSFAIndexerBackend`` instance it owns.
     """
 
-    def __init__(self, vllm_indexer: nn.Module) -> None:
+    def __init__(self, vllm_indexer: nn.Module, qk_rope_head_dim: int) -> None:
```
lcfenglinwan: vllm_indexer module should have `qk_rope_head_dim` attribute, no need to pass this parameter

vllm_ascend/attention/sfa_v1.py
```diff
@@ -1602,12 +1391,15 @@ def forward(
...
+            # indexer.forward_k below and gathered at cache-write time, so
+            # the fused gather below only carries the main KV.
+            _, _, fused_kv_no_split, kv_ag_handles = self._prepare_kv_for_parallel(
                 k_pe,
                 k_nope,
                 knope_scale,
-                k_li,
-                k_li_scale,
+                None,
+                None,
```
lcfenglinwan: When the `k_li` parameter is no longer used, it is recommended to modify the definition of the `_prepare_kv_for_parallel` method here and remove the related parameters.

lijiahang226 (approved):
LGTM. Thank you for the contribution.

weijinqian0 (dismissed):

vllm_ascend/attention/sfa_v1.py
```diff
@@ -1602,12 +1392,15 @@ def forward(
...
```
weijinqian0: if k_li and k_li_scale is removed, may its' parameter can be removed.

vllm_ascend/attention/sfa_v1.py
```diff
@@ -1622,12 +1415,12 @@ def forward(
...
+                _,
             ) = self._store_parallel_kv(
                 k_pe,
                 k_nope,
                 knope_scale,
-                k_li,
+                None,
                 fused_kv_no_split,
                 kv_ag_handles,
                 kv_cache,
```
weijinqian0, line 1422: Same as the last comment.

weijinqian0 (dismissed):

zzzzwwjj (approved):

PR #15672, opened 2026-09-03
[Feat][Worker] Pass lease TTL to memcache batch_alloc

tyy0829:

The memcache_hybrid `batch_alloc` interface supports a `leaseTtlMs` parameter, but vllm-ascend currently only passes `keys` and `sizes`, so allocated blocks are not covered by a lease at allocation time.

- Add `lease_ttl_ms` parameter to `Backend.batch_alloc` and `MemcacheBackend.batch_alloc`, forwarded to the underlying memcache_hybrid store as `leaseTtlMs` (default `0` keeps existing behavior for other backends)
- Pass `LAYERWISE_READ_LEASE_TTL_MS` (5 min, the same constant already used by the layerwise load path) at both GVA allocation call sites in `_alloc_gvas_for_save` so newly allocated blocks are covered by a lease
- Update the UT assertion to cover the new argument

No

- Unit test updated: `tests/ut/distributed/ascend_store/test_pool_worker.py::test_evicted_allocated_gva_is_reallocated` now asserts `batch_alloc` is called with the lease TTL
- `pytest -sv tests/ut/distributed/ascend_store/` (302 passed locally; remaining failures are pre-existing on unmodified main and caused by the local Windows environment lacking `torch.npu` / real vllm)
- `ruff check` / `ruff format --check` pass

### DIFF
--- a/tests/ut/distributed/ascend_store/test_pool_worker.py
+++ b/tests/ut/distributed/ascend_store/test_pool_worker.py
@@ -1445,6 +1445,10 @@ class TestKVPoolWorkerProcessLayerData(unittest.TestCase):
         )
 
     def test_evicted_allocated_gva_is_reallocated(self):
+        from vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.pool_worker import (
+            LAYERWISE_READ_LEASE_TTL_MS,
+        )
+
         worker = self._make_gva_worker()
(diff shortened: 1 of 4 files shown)


tyy0829:
Closing in favor of #15830, which now contains this commit (plus a layerwise MTP final-block trim fix) rebased on the latest main, so only one PR needs review. cc @reviewers

PR #15700, opened 2026-09-03
[Refactor][Device] Complete hardware profile migration and enforce HAL boundary

Tflowers-0129:

This PR completes the Device HAL / Hardware Profile migration and makes the HAL boundary enforceable.

1. **Complete the remaining business-layer migration**
   - Migrates the remaining direct device-identity consumers to precise capability, policy, or implementation-family selectors.
   - Removes hidden device-name helpers from shared runtime/model paths.
   - Keeps device identity confined to detection, type mapping, and profile registration.

2. **Audit HAL semantics and naming**
   - Capabilities describe positive hardware/runtime functions.
   - Policies and families describe implementation, algorithm, layout, or default-behavior choices.
   - Replaces ambiguous unscoped names and device/model codenames with names that state scope, action, and object.
   - Documents the naming contract and updates the exact profile matrix and focused tests.

3. **Add a permanent regression guard**
   - Adds an AST checker that rejects production uses of direct device identity outside the audited HAL allowlist.
   - Detects direct calls/comparisons, aliases, imported aliases, dynamic attribute access, and scoped shadowing cases.
   - Wires the checker into pre-commit/required CI and adds positive, negative, and adversarial unit tests.

**Merge order:** #15695 fixes the independent PR5 DSA-CP o_proj TP behavior regression. This PR is opened now for review and CI, but #15695 must merge first; this branch will then be rebased so the obsolete DSA gate is removed by the dedicated BugFix rather than mixed into this refactor. Auto-merge remains disabled until that rebase.

No intended runtime behavior change. Shared code selects the same behavior through explicit HAL contracts instead of A2/A3/A5/310P identity branches. The developer workflow gains a CI error for new production device-identity branching outside the HAL boundary.

- Rebased onto `main@3546357838389aa7201d9b44e234f831e98b2fd4`; both commits remain patch-equivalent under `git range-diff`.
- Both commits include `Signed-off-by`.
- Passed Ruff check, Ruff format --check, compileall, the full-tree hardware-identity AST guard, adversarial diagnostics, and `git diff --check`.
- The available host does not provide pytest, mypy, or vLLM; focused matrix/consumer/checker unit tests and Python 3.10/3.11/3.12 mypy comparison are left to CI.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -121,6 +121,12 @@ repos:
     entry: python3 tools/check_boolean_context_manager.py
     language: python
     types: [python]
+  - id: check-hardware-identity
+    name: Forbid hardware identity use outside the HAL boundary
+    entry: python3 tools/check_hardware_identity.py
+    language: python
+    pass_filenames: false
+    always_run: true
   - id: check-long-functions
     name: Check that new functions over 100 lines have comments
     entry: python3 tools/check_long_functions.py
--- a/docs/source/developer_guide/contribution/index.md
+++ b/docs/source/developer_guide/contribution/index.md
@@ -71,6 +71,31 @@ git commit -sm "your commit info"
 
 You can refer to [Testing](./testing.md)  to set up a testing environment and running tests locally.
 
+## Hardware-specific behavior
+
+Shared runtime and model code must select hardware-specific behavior through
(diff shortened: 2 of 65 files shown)


Tflowers-0129:
split this PR to some small PRs

PR #15707, opened 2026-09-03
[BugFix][Spec Decode] Fix GDN graph padding and Qwen MTP MRoPE inputs

Tame21:

This PR fixes graph-padding metadata handling on the v2 model runner and an input-layout mismatch in Qwen MTP's fused MRoPE path.

1. Preserve GDN graph padding metadata during full-graph capture and replay.
- Keep the padded request count aligned with the captured batch descriptor for uniform decode graphs.
- Mark padded Mamba requests as speculative dummy requests when their query lengths match the speculative decode length.
- Keep capture and replay on the same pure-speculative GDN metadata path so persistent state indices are refreshed correctly.

2. Validate actual query lengths before applying uniform decode padding.
- Require every real request's query length to equal decode_query_len.
- Avoid classifying prefill or mixed batches as uniform decode solely because the total token count matches the graph descriptor.
- Preserve the appropriate padding layout for non-uniform batches.

3. Provide the required three-plane input layout to Qwen's fused MRoPE kernel.
- The v2 MTP path supplies one-dimensional text positions. Indexing the rotary cache directly with these positions produces [T, D], while the fused Triton kernel reads contiguous T/H/W planes as [3, T, D].
- Expand one-dimensional positions to three identical axes before cache lookup, allowing the lookup to materialize the required contiguous planes.
- Preserve existing three-axis positions without changing their values.

These changes prevent capture/replay metadata mismatches and out-of-bounds H/W-plane reads that can corrupt MTP attention and reduce draft acceptance.

Yes. It fixes GDN full-graph metadata handling and the Qwen MTP acceptance regression caused by the fused MRoPE input-layout mismatch.

No public API, configuration option, or command-line interface changes are introduced.

Added or extended regression coverage for:
- Captured request-shape preservation across full-graph modes.
- Padded speculative-request metadata for Mamba hybrid models.
- Non-uniform queries, including prefill, mixed batches, and batches whose average query length matches the decode length.
- One-dimensional and three-axis MRoPE positions, checking the resulting cos/sin shape, contiguity, and values.

Run the relevant tests with:
```
pytest -sv \
  tests/ut/worker/test_model_runner_v2.py \
  tests/ut/worker/test_model_runner_v2_mamba.py \
  tests/ut/patch/worker/test_patch_qwen3_5_mtp.py
```

### Model accuracy and speculative acceptance validation
Evaluated both model runners on **GPQA_diamond**
Model | Runner | Accuracy (%) | Position 1 acceptance | Position 2 acceptance | Position 3 acceptance | Mean acceptance length
-- | -- | -- | -- | -- | -- | --
Qwen3.6-35B-A3B | MRV1 | 83.84 | 76.7534% | 58.6171% | 45.4967% | 2.808672
Qwen3.6-35B-A3B | MRV2 | 83.84 | 76.5456% | 58.2090% | 44.9781% | 2.797328
Qwen3.6-27B | MRV1 | 85.35 | 83.6819% | 67.5557% | 54.3134% | 3.055510
Qwen3.6-27B | MRV2 | 85.86 | 83.8894% | 68.0398% | 55.0173% | 3.069465

### DIFF
--- a/tests/ut/worker/test_model_runner_v2.py
+++ b/tests/ut/worker/test_model_runner_v2.py
@@ -96,10 +96,14 @@ def test_full_decode_only_keeps_graph_descriptor_request_count():
     np.testing.assert_array_equal(actual[:5], np.array([0, 1, 2, 3, 4], dtype=np.int32))
 
 
-def test_uniform_full_graph_preserves_descriptor_request_shape():
+@pytest.mark.parametrize(
+    "cudagraph_mode",
+    [CUDAGraphMode.FULL, CUDAGraphMode.FULL_DECODE_ONLY, CUDAGraphMode.FULL_AND_PIECEWISE],
+)
+def test_uniform_full_graph_preserves_descriptor_request_shape(cudagraph_mode):
     runner = NPUModelRunner.__new__(NPUModelRunner)
     runner.decode_query_len = 4
-    runner.compilation_config = SimpleNamespace(cudagraph_mode=CUDAGraphMode.FULL)
+    runner.compilation_config = SimpleNamespace(cudagraph_mode=cudagraph_mode)
     query_start_loc = np.full(8, 8, dtype=np.int32)
     query_start_loc[:3] = [0, 4, 8]
 
@@ -114,6 +118,40 @@ def test_uniform_full_graph_preserves_descriptor_request_shape():
 
     assert num_reqs_padded == 4
     np.testing.assert_array_equal(padded_query_start_loc[:5], [0, 4, 8, 12, 16])
+
+
+@pytest.mark.parametrize(
+    "decode_query_len, query_lens, num_tokens_padded, descriptor_num_reqs, expected_query_start_loc",
+    [
+        (1, [4], 8, 8, [0, 4, 8]),
+        (4, [4, 5], 16, 4, [0, 4, 9, 16]),
+        (4, [2, 6], 16, 4, [0, 2, 8, 16]),
+        (4, [2, 4], 8, 2, [0, 2, 6, 8]),
+    ],
+    ids=["non-mtp-prefill", "mtp-mixed", "mtp-uniform-average", "mtp-no-request-padding"],
+)
+def test_full_graph_non_uniform_queries_use_mixed_padding(
+    decode_query_len, query_lens, num_tokens_padded, descriptor_num_reqs, expected_query_start_loc
+):
+    runner = NPUModelRunner.__new__(NPUModelRunner)
+    runner.decode_query_len = decode_query_len
+    runner.compilation_config = SimpleNamespace(cudagraph_mode=CUDAGraphMode.FULL)
+    num_reqs = len(query_lens)
+    query_start_loc = np.full(descriptor_num_reqs + 2, sum(query_lens), dtype=np.int32)
+    query_start_loc[: num_reqs + 1] = np.cumsum([0, *query_lens])
+
+    padded_query_start_loc, num_reqs_padded = runner._pad_query_start_loc_for_fia(
+        num_tokens_padded=num_tokens_padded,
+        num_reqs_padded=descriptor_num_reqs,
+        num_reqs=num_reqs,
+        query_start_loc_np=query_start_loc,
+        cudagraph_runtime_mode=CUDAGraphMode.FULL,
+        batch_desc_num_reqs=descriptor_num_reqs,
+    )
+
+    assert num_reqs_padded == num_reqs + 1
+    np.testing.assert_array_equal(padded_query_start_loc[: num_reqs_padded + 1], expected_query_start_loc)
+    assert padded_query_start_loc[num_reqs_padded] == num_tokens_padded
 
 
 def test_sample_tokens_restores_replicated_draft_hidden_states():
--- a/tests/ut/worker/test_model_runner_v2.py
+++ b/tests/ut/worker/test_model_runner_v2.py
@@ -96,6 +96,26 @@ def test_full_decode_only_keeps_graph_descriptor_request_count():
     np.testing.assert_array_equal(actual[:5], np.array([0, 1, 2, 3, 4], dtype=np.int32))
 
 
+def test_uniform_full_graph_preserves_descriptor_request_shape():
+    runner = NPUModelRunner.__new__(NPUModelRunner)
+    runner.decode_query_len = 4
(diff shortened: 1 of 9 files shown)


jiajinzhu2:
Please include a performance comparison between MRV1 and MRV2 in the description.

jiajinzhu2 (changes requested):
Please include the acceptance rate comparison between MRV1 and MRV2 in the PR description.

vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
```diff
@@ -553,14 +560,21 @@ def _update_decode_attn_metadata(
...
+        next_seqs_cpu[:num_reqs] = torch.clamp(
+            next_seqs_cpu[:num_reqs] - self.num_rejected_cpu[:num_reqs] + step,
+            min=0,
+            max=self.max_model_len,
+        )
         next_seqs_cpu[num_reqs:].fill_(0)
         return next_seqs_cpu
 
+    def _copy_num_rejected_to_cpu(self, num_rejected: torch.Tensor, num_reqs: int) -> None:
+        # FIA graph parameters consume host sequence lengths. This blocking copy
```
drslark: This synchronization will make performance bad.
So, in previous version, we sacrifice acceptance to get performance.
Why we need synchronization here?

vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
```diff
@@ -553,14 +560,21 @@ def _update_decode_attn_metadata(
             metadata.seq_lens_cpu.copy_(next_seq_lens_cpu)
 
     def _calc_next_seq_lens_cpu(self, seq_lens_cpu, num_reqs, num_reqs_padded, step):
-        # NOTE(drslark) to achieve fully alignment with vllm, `num_rejected` should be subtracted from `seq_lens`
```
drslark, line 570: @slippersss 
Please review it, does here need a zero filler?

wxsIcey:
Please add e2e tests

liaoqingqing677:
| dataset       | version  | metric   | mode  | vllm-api-general-chat |
|---------------|----------|----------|-------|------------------------|
| GPQA_diamond  | b1ed2c   | accuracy | gen   | 83.84                  |

vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
Tame21: You’re right that a blocking D2H copy on every drafting iteration adds overhead. I’ve reverted that change in [1bb910956](https://github.com/Tame21/vllm-ascend/commit/1bb910956821ea7a354d6e454b7dbdf5be4c4f35), preserving the existing optimistic-length strategy.

tests/ut/worker/test_model_runner_v2.py
```diff
@@ -100,6 +100,64 @@ def test_full_decode_only_keeps_graph_descriptor_request_count():
...
+    "decode_query_len, query_lens, num_tokens_padded, descriptor_num_reqs, expected_query_start_loc",
+    [
+        (1, [4], 8, 8, [0, 4, 8]),
+        (4, [4, 5], 16, 4, [0, 4, 9, 16]),
+        (4, [2, 6], 16, 4, [0, 2, 8, 16]),
+        (4, [2, 4], 8, 2, [0, 2, 6, 8]),
+    ],
+    ids=["non-mtp-prefill", "mtp-mixed", "mtp-uniform-average", "mtp-no-request-padding"],
+)
+def test_full_graph_non_uniform_queries_use_mixed_padding(
```
drslark, line 109: Just keep `test_full_graph_non_uniform_queries_use_mixed_padding` and `test_prepare_attn_marks_uniform_full_graph_padding_as_spec`. I think others are redundant.

drslark (changes requested):

drslark (approved):
LGTM.

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -702,15 +953,25 @@ def _pad_query_start_loc_for_fia(
         """
         # TODO: need refactor later, related to vllm PR #34043 this pr delete func
         # relax_for_mixed_batch_cudagraphs, num_reqs no longer equals the actual number of requests.
+        descriptor_num_reqs = batch_desc_num_reqs if batch_desc_num_reqs is not None else num_reqs_padded
+        is_uniform_decode_batch = np.all(np.diff(query_start_loc_np[: num_reqs + 1]) == self.decode_query_len)
```
realliujiaxu: what if there is only a prefill request and its length equals to `decode_query_len`?

Tame21: You’re right: a single prefill request whose query length equals decode_query_len would also make this condition true. The condition only checks query lengths.
After merging the latest main, the pinned vLLM version (b2f6858) already handles this before graph selection:
1. [gather_batch_req_state()](https://github.com/vllm-project/vllm/blob/b2f685834a6456197e7033966fdef52a23f1abcd/vllm/v1/worker/gpu/model_runner.py#L1121) identifies prefills using num_computed_prefill_tokens < prefill_len and sets has_prefill if any scheduled request is still prefilling.
2. [get_uniform_decode_token_count()](https://github.com/vllm-project/vllm/blob/b2f685834a6456197e7033966fdef52a23f1abcd/vllm/v1/worker/utils.py#L704) returns None whenever has_prefill is true, even when all query lengths equal decode_query_len.
3. Graph dispatch receives this value, and [_is_compatible()](https://github.com/vllm-project/vllm/blob/b2f685834a6456197e7033966fdef52a23f1abcd/vllm/v1/worker/gpu/cudagraph_utils.py#L95) rejects descriptors requiring a uniform decode token count.
Therefore, the short prefill in your example cannot select a uniform decode graph merely because its query length matches. This also covers mixed batches containing prefills.
I’ve renamed the variable to has_uniform_decode_query_lens and added a comment clarifying that it checks shape, while graph dispatch handles request-phase eligibility. I’ve also removed the obsolete legacy dispatch workaround now that the upstream check covers this case, and retained the padding fix needed to preserve the captured request shape for valid uniform decode batches.

realliujiaxu (approved):
LGTM

jiajinzhu2 (approved):

PR #15745, opened 2026-09-04
[BugFix][Scheduler] Record balance prefix cache metrics

### LINKED ISSUE #15744: [Bug]: BalanceScheduler does not report prefix-cache metrics
## Your current environment

<details>
<summary>Environment summary</summary>

```text
Hardware: Ascend 910B/C
vLLM Ascend: main at b4b04c5eb (vLLM v0.27.1 pin)
Model: Eco-Tech/GLM-5.3-w8a8c8 (glm_moe_dsa)
Parallelism: TP=8, DP=2
Key options:
  --enable-prefix-caching
  --enable-chunked-prefill
  --async-scheduling
  --block-size 128
  --additional-config '{"enable_balance_scheduling":true,...}'
```

</details>

## Describe the bug

When both prefix caching and Ascend balance scheduling are enabled, prefix-cache
lookups work, but the local prefix-cache metrics are never reported.

I sent the same prompt longer than 4K tokens five times to DP rank 0. The first
request was cold and the following requests hit 5,248 cached tokens:

```text
run 0 [COLD]  4.01s  prompt_tokens=5457  cached_tokens=0
run 1 [WARM]  1.93s  prompt_tokens=5457  cached_tokens=5248
run 2 [WARM]  2.09s  prompt_tokens=5457  cached_tokens=5248
run 3 [WARM]  1.98s  prompt_tokens=5457  cached_tokens=5248
run 4 [WARM]  1.95s  prompt_tokens=5457  cached_tokens=5248
```

Image

Despite the confirmed cache hits, both engines continue to report zero:

```text
vllm:prefix_cache_queries_total{engine="0",model_name="glm-5.3"} 0.0
vllm:prefix_cache_queries_total{engine="1",model_name="glm-5.3"} 0.0
vllm:prefix_cache_hits_total{engine="0",model_name="glm-5.3"} 0.0
vllm:prefix_cache_hits_total{engine="1",model_name="glm-5.3"} 0.0
```

The periodic log also reports `Prefix cache hit rate: 0.0%`.

The issue is specific to `enable_balance_scheduling: true`. When balance
scheduling is disabled, `BalanceScheduler.schedule()` delegates to upstream
`Scheduler.schedule()` and the metrics are updated normally.

Image

### Reproduction

1. Start a TP=8 / DP=2 GLM-5.3 deployment with the options listed above.
2. Send an identical prompt longer than 4K tokens five times. Pin all requests
   to the same DP rank (for example, `X-data-parallel-rank: 0`).
3. Confirm that later responses report cached tokens and lower latency.
4. Inspect the metrics:

   ```bash
   curl -s http://127.0.0.1:8000/metrics \
     | grep -E 'prefix_cache_(queries|hits)_total\{'
   ```

### Root cause

`BalanceScheduler.schedule()` contains a copied scheduler body. Upstream vLLM
PR [#48860](https://github.com/vllm-project/vllm/pull/48860) moved local
prefix-cache metric recording out of `KVCacheManager.get_computed_blocks()` and
into the scheduler admission path. The copied balance scheduler still performs
the cache lookup, so requests get cache hits, but it never calls
`kv_cache_manager.record_prefix_cache_stats(...)`.

The custom Mamba hybrid lookup is a related edge case: it currently records
directly during lookup. Once the admission-time call is restored, that direct
recording must be removed to avoid double-counting, and the lookup must retain
the normal `prefix_cache_lookup_enabled()` guard.

### Expected behavior

For each admitted request that performed a prefix-cache lookup,
`prefix_cache_queries_total` should increase by the queried token count and
`prefix_cache_hits_total` should increase by the local hit count. Deferred or
allocation-failed requests must not be counted.

### Validated fix

Restore the vLLM v0.27.1 admission-time logic in `BalanceScheduler.schedule()`:

- track whether the request performed a local prefix-cache lookup;
- call `record_prefix_cache_stats()` only after slot allocation succeeds;
- remove direct Mamba lookup-time recording to prevent double-counting;
- preserve the prefix-cache lookup guard in the Mamba custom path.

With this change, the same validation produced non-zero metrics on DP rank 0:

```text
vllm:prefix_cache_queries_total{engine="0",model_name="glm-5.3"} 55245.0
vllm:prefix_cache_hits_total{engine="0",model_name="glm-5.3"} 42496.0
```

Image

The periodic log reported a prefix-cache hit rate of approximately 69-77%.

Image

[prefix_cache_test_6k.py](https://github.com/user-attachments/files/31818146/prefix_cache_test_6k.py)

mashuiping:

Fixes #15744.

When `enable_balance_scheduling` is enabled, `BalanceScheduler.schedule()`
performs local prefix-cache lookups but does not report them through
`KVCacheManager.record_prefix_cache_stats()`. As a result, cached requests are
served faster and expose non-zero `cached_tokens`, while
`prefix_cache_queries_total`, `prefix_cache_hits_total`, and the logged hit rate
remain zero.

This change aligns the copied balance-scheduler admission path with upstream
vLLM PR [#48860](https://github.com/vllm-project/vllm/pull/48860):

- remember whether a waiting request performed a local prefix-cache lookup;
- record local cache statistics only after slot allocation succeeds, so
  deferred or allocation-failed requests are not counted;
- remove lookup-time recording from the custom Mamba hybrid path to avoid
  double-counting;
- preserve `prefix_cache_lookup_enabled()` and `shared_prefix_boundary`
  semantics in that custom path;
- add a regression guard for the admission-time recording contract.

Manual validation on an Ascend 910B/C TP=8 / DP=2 GLM-5.3 deployment changed
the affected DP rank from zero metrics to:

```text
vllm:prefix_cache_queries_total{engine="0",model_name="glm-5.3"} 55245.0
vllm:prefix_cache_hits_total{engine="0",model_name="glm-5.3"} 42496.0
```

The periodic prefix-cache hit rate also changed from `0.0%` to approximately
`69-77%`.

Yes. With balance scheduling enabled, prefix-cache Prometheus counters and the
logged prefix-cache hit rate now reflect admitted requests and actual local
cache hits.

- Added
  `tests/ut/patch/platform/test_patch_balance_schedule.py::test_prefix_cache_stats_recorded_only_at_admission`.
- `ruff format --check vllm_ascend/patch/platform/patch_balance_schedule.py tests/ut/patch/platform/test_patch_balance_schedule.py`
- `ruff check vllm_ascend/patch/platform/patch_balance_schedule.py tests/ut/patch/platform/test_patch_balance_schedule.py`
- `python -m py_compile vllm_ascend/patch/platform/patch_balance_schedule.py tests/ut/patch/platform/test_patch_balance_schedule.py`
- Manual before/after validation with repeated >4K-token prompts on Ascend
  910B/C, TP=8, DP=2, GLM-5.3, prefix caching and balance scheduling enabled.

The targeted pytest module could not be executed in the local host environment
because PyTorch is not installed (`ModuleNotFoundError: No module named
'torch'`); CI should execute it in the project test image.


mashuiping:
@zzzzwwjj  @yiz-liu  Could you please help review this BalanceScheduler prefix-cache metrics fix? The issue includes before/after [validation on Ascend 910B](https://github.com/vllm-project/vllm-ascend/issues/15744). CI also requires a maintainer to add the ready-precise label. Thanks!

PR #15747, opened 2026-09-04
[Feature][MRV2] Support GLM5.2 DSpark with PP

LostFox11:

Enable GLM-5.2 DSpark + pipeline parallelism in Model Runner V2 while supporting both released and mainline vLLM builds.

| Paired vLLM version | Speculative PP implementation |
| --- | --- |
| 0.28.x / 0.29.x | Install the vLLM 0.30 PP sampled-token protocol (pure-function participation gate, immediate broadcasts, receive-side generation-counter filtering) onto the release PPHandler. This replaces the legacy deferred-broadcast transport that deadlocked under KV saturation with async EPLB. Delete once the paired vLLM ships the protocol natively. |
| 0.30+ | Use upstream PP initialization, draft broadcast, request-state writeback, and aux relay natively. No Ascend protocol patches are installed. |

- Centralize version selection in `pp_utils.use_legacy_spec_pp()`. All call sites are annotated with the retirement condition.
- Add GLM to the DSpark PP support map. On the native path, connect its DeepSeekV2 decoder to the upstream `EagleModelMixin` slot bookkeeping and capability check.
- Capture aux states on the stage producing them, including `end_layer`, so a boundary state is transmitted exactly once. Dual-path support: legacy `pp_transport` buffers (0.28/0.29) and upstream relay (0.30+).
- Construct the Ascend speculator only on the last PP rank on both paths.
- Mask the target's manual PP partition during DSpark draft loading on both paths, restoring it even on exceptions.
- Retain GLM's local MoE layer count for MRV2+PP EPLB maps.
- DeepseekV2 MLA attention init with Ascend IndexCache and top-k skip-pattern support.

> **Note for 0.28/0.29 release trains:** when using this feature at high load, backport vllm-project/vllm#49675 (~19-line scheduler check that stops zero-progress preemption cascades for deferred KV frees). Without it, KV saturation with async EPLB triggers a preemption storm and potential engine stall.

GLM-5.2 DSpark can use MRV2 PP with either supported release lane or the upstream PP implementation. No new switches or environment variables are introduced.

For the previously tested GLM-5.2 W4A8 + EPLB configuration, use `additional_config.enable_fused_mc2=1` to avoid the independently identified non-fused grouped-matmul bias-list issue.

### Hardware validation (169 server, 16× Ascend 910B)

**vLLM 0.30 (upstream protocol, `84030bbe3`)**

| Test | Config | Result |
| --- | --- | --- |
| GPQA full 198 questions | c24, u0.85, mns12, KV pool 365K | **176/198 = 88.89%**, 0 failures, 0 stalls, 40 preemptions |
| GPQA c32 partial | c32, u0.80, KV pool 306K | 109/198 before voluntary stop, 0 failures, 95 preemptions |

**vLLM 0.28 + this PR's protocol + #49675 backport**

| Test | Config | Result |
| --- | --- | --- |
| GPQA partial (terminated at 97/198) | c32, u0.80, KV pool 306K, async EPLB | **91/99 = 91.92%** on completed questions, 0 failures, **0 stalls** (vs. deadlocks without fix), 78 preemptions |
| Same-question comparison vs 0.30 | — | 91.92% vs 91.92% (exact match), 94.95% per-question answer agreement |

**vLLM 0.29 legacy protocol (before this PR's fix, for reference)**

| Test | Config | Result |
| --- | --- | --- |
| GPQA c32 (stalled) | c32, u0.80, KV pool 306K, async EPLB | Deadlocked at 66/198, 44,202 preemptions, 20+ min zero-progress stall |

### CPU tests

- **143 pytest cases passed**: version routing, partition restoration, draft-loader isolation, PP2/PP4 aux equivalence to unpartitioned execution, V1 setter preservation, local EPLB layer counts, protocol broadcast ordering and purity, model-runner/PCP regressions.

### Smoke test (169 native path)

GLM-5.2 W4A8, PP2×DP2×TP4, DSpark7, async scheduling, prefix cache, chunked prefill, EPLB. Loading, single request, and 10 concurrent requests all pass with correct outputs.


LostFox11:
Added `44dd069b2398cf675c8aa8a90776e9b95c114c61`: a minimal shared MRV2 speculative-PP CPU sequence-length fix.

Non-last PP stages do not own a speculator, but PP feedback still corrects their device-side computed-token counts after rejection. Before building CPU attention metadata, these stages now refresh the existing D2H snapshot and use the exact counters instead of scheduler upper bounds. Copying at this point also includes non-final prefill chunks, which advance without sampled-token feedback. The existing shared CPU/NumPy state update is preserved. PP broadcast, auxiliary-state transport, sampling, MRV1, and non-PP behavior are unchanged.

Validation:
- Eight NumPy/mock regression cases added. Isolated execution of the production method and test bodies, without Torch/NPU imports: before the fix, 5 failed / 3 passed; after the fix, 8 passed.
- Python syntax and `git diff --check` passed.
- The same fix was previously exercised with real-weight M3 PP2TP8 + Eagle3, target FULL_DECODE_ONLY and draft eager: 16/16 chunked-prefill smoke requests completed correctly after diagnostic probes were removed.
- The GLM-5.2 + DSpark combination has not been rerun with this appended commit. This fixes a confirmed metadata inconsistency; it does not claim that the remaining DP/PP GPQA accuracy differences are resolved. D2H synchronization overhead has not been benchmarked.
- Full local pytest and `bash format.sh ci` could not run because pytest/pre-commit are absent; no environment was reinstalled. CI validation is still required.

LostFox11:
Updated in bfbb0e278 to follow upstream MRV2's separation between optimistic CPU bounds and exact device sequence lengths (the semantics documented in vllm-project/vllm#40654 and vllm-project/vllm#42202).

This supersedes the blanket non-last-stage D2H refresh introduced in 44dd069b2:
- Ordinary non-last speculative PP stages preserve the scheduler CPU counters as upper bounds. They no longer enqueue an unconditional counter copy and immediately wait on its event in the runner.
- AscendInputBatch marks those lengths as upper bounds; default and hybrid model states forward this distinction and seq_lens_cpu_upper_bound separately to attention metadata construction.
- Exact CPU metadata is left unset on this path. FIA uses its existing fallback to obtain rejection-corrected device lengths. DSA's MRV2 caller supplies a CPU bound only for the planning maximum; per-request kernel lengths remain on device.
- The PCP path still needs exact CPU counts for partitioning, so its refresh is retained. Existing last-stage/non-PP draft and capture behavior is preserved. The optional DSA argument is passed only by MRV2; MRV1 callers are unchanged.

Validation:
- 29 isolated CPU regression cases passed using the existing CPU Torch/pytest in the 169 pp container, including actual production function bodies for the runner, metadata construction, DSA and FIA. Hardware initialization and external metadata containers were stubbed; this is not a full import/integration or NPU runtime test.
- Ruff check, Ruff format --check, Python syntax checks and git diff --check passed.
- Full format.sh ci could not run because pre-commit is not installed locally; no environment was installed or rebuilt.
- Existing services were not restarted. Prior NPU/GPQA measurements apply to the previous implementation, not to this revision. NPU accuracy and performance remain to be revalidated.

This removes the blanket runner barrier, not every possible host wait: backends requiring exact CPU lengths (including FIA) can still synchronize on demand, and PCP retains its existing correctness dependency. No zero-bubble or throughput claim is made.

LostFox11:
/rerun
[Bot]: rerun completed.

**Failed:**

tests/e2e/pull_request/eight_card/model_runner_v2/test_glm5_2_spec_pp.py
```diff
@@ -0,0 +1,162 @@
...
+        "VLLM_PP_LAYER_PARTITION": GLM52_PP_LAYER_PARTITION,
+        "HCCL_BUFFSIZE": "2048",
+    },
+)
+@wait_until_npu_memory_free(target_free_percentage=0.8)
+def test_glm5_2_dspark_pp_full_decode_only() -> None:
+    """GLM-5.2 W4A8 DSpark + PP2/TP8/EP in graph mode with all features on."""
+    with VllmRunner(
+        GLM52_MODEL,
+        quantization="ascend",
```
weiguihua2: Currently, community resources are insufficient to support the 8-card use case.

vllm_ascend/patch/worker/patch_v2/patch_spec_pp.py
```diff
@@ -2,12 +2,29 @@
...
 
+from dataclasses import replace
+
 import numpy as np
 from vllm.v1.worker.gpu.buffer_utils import async_copy_to_gpu
 
 _BROADCAST_PATCHED = "_vllm_ascend_spec_pp_broadcast_patched"
 
 
+def _make_spec_pp_comm_batch(input_batch):
```
LostFox11, line 13: todo：fall down after main to main

vllm_ascend/patch/worker/patch_deepseek_v2.py
```diff
@@ -290,25 +292,45 @@ def _deepseek_v2_mla_attention_init(
 DeepseekV2MLAAttention.__init__ = _deepseek_v2_mla_attention_init
 
 
+_original_deepseek_v2_model_init = DeepseekV2Model.__init__
+
+
```
LostFox11, line 306: fall down after m2m

vllm_ascend/patch/worker/patch_deepseek_v2.py
```diff
@@ -290,25 +292,45 @@ def _deepseek_v2_mla_attention_init(
...
+def _patched_deepseek_v2_model_init(self, *args, **kwargs):
```
LostFox11, line 307: todo: fall down after new m2m

vllm_ascend/worker/v2/pp_utils.py
freyfwt, line 143: It seems the 'layer_idx <= model.start_layer' here should be 'layer_idx < model.start_layer'. Please double check.

freyfwt (approved):

LostFox11:
/rerun
[Bot]: rerun completed.

**Failed:**

nwpu-zxr (approved):

yiz-liu (approved):

Issue #15462, opened 2026-09-01
[Roadmap][Model] DeepSeek V4 Flash Vision support on Ascend


GDzhu01:
## Summary

Track end-to-end enablement of
[DeepSeek-V4-Flash-Vision-Exp](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-Vision-Exp)
on vLLM Ascend `main`.

Related work:

- Initial feature request: #15419
- Implementation PR: #15457
- Upstream behavior reference: vllm-project/vllm#54566

The initial implementation PR #15457 merged into `main` on 2026-09-04
(merge commit `63233e64a6027f9b664657595cf3cec0c1a2fee4`) after the required CI gate passed.
This roadmap remains open until numerical accuracy, coverage, performance, and
documentation satisfy release criteria.

## Validated baseline

- vLLM commit: `6e448d0ea9bf3d88d898b65449ca6dc2aec170ac`
- Runtime-validation vLLM Ascend PR commit: `485a9b7bb57cdf2bebac50bdbc8129250fe6675c`
- Final implementation head: `a0f4947bbb72942806a8ac3fcdb2772a0ef70ec7`
- Main merge commit: `63233e64a6027f9b664657595cf3cec0c1a2fee4`
- Checkpoint: DeepSeek-V4-Flash-Vision-Exp W8A8 with all BF16 `bias_vl`
  tensors present
- Release topology tested: TP4 / DP4 / EP16 on 16 Ascend NPUs
- Eager launch: `--max-num-seqs 1 --enforce-eager`
- Graph launch: `--max-num-seqs 4`; VLLM_COMPILE plus PIECEWISE prefill/decode
  and FULL decode graph capture

### Phase 0 — Compatibility and checkpoint contract

- [x] Port the processor, ViT, and aligner using vLLM v0.27 extension points.
- [x] Route vision checkpoints to `DeepseekV4ForConditionalGeneration`.
- [x] Confirm the five image sentinel embeddings, placeholder expansion,
  leading-pad alignment, and image-span metadata path.
- [x] Confirm the W8A8 tensor namespace and add multimodal ModelSlim prefix
  mappings.
- [x] Include and load all 46 BF16 `bias_vl` tensors in the W8A8 checkpoint
  overlay.
- [x] Freeze the validated vLLM/vLLM Ascend commit pair above.
- [x] Land the initial implementation on `main` via #15457 with the required
  CI gate passing.

[...]

PR #15457, opened 2026-09-01
[Feature][Model] Support DeepSeek V4 Flash Vision on Ascend

GDzhu01:
### What this PR does / why we need it

Adds `DeepSeek-V4-Flash-Vision-Exp` support to vLLM Ascend `main`, paired with
the vLLM v0.27.x API line.

Roadmap and remaining validation tracking: #15462

Upstream behavior reference: vllm-project/vllm#54566

The paired vLLM release does not contain the DeepSeek-V4 vision processor and
vision tower from the upstream PR. This change therefore carries the required
v0.27-compatible implementation in vLLM Ascend without importing unmerged
vLLM modules.

### Implementation

- registers `DeepseekV4ForConditionalGeneration` and routes vision
  checkpoints to an Ascend multimodal wrapper;
- provides the image processor, ViT, aligner, sentinel embeddings, placeholder
  expansion, and multimodal weight mapping required by vLLM v0.27;
- preserves raw input token IDs through prefill for DeepSeek-V4 MoE routing;
- implements hybrid `bias_vl` routing: text tokens retain hash/correction-bias
  routing while image sentinel tokens use dynamic routing with `bias_vl`;
- builds paged original-KV sparse indices that make each complete image span
  bidirectionally visible while keeping causal SWA outside image spans;
- passes original-KV image indices and compressed-KV top-k indices to the
  Ascend sparse-attention operator independently and simultaneously;
- supports ModelSlim W8A8 checkpoints whose language quantization description
  remains in the causal-LM `model.*` / `lm_head.*` namespace;
- keeps config-convertor registration safe in spawned DP engine processes.
- enables DeepSeek-V4 Vision with a text-only DSpark drafter: image inputs are
  encoded by the target during prefill, while DSpark proposes subsequent text
  decode tokens; restores the draft architecture after VL config conversion,
  exposes the EAGLE3 hidden-state interface, and ignores draft-only vision
  router bias tensors without changing text routing.

### Runtime validation completed

Validated in a 16-NPU Ascend container with:

- vLLM: `6e448d0ea9bf3d88d898b65449ca6dc2aec170ac` (v0.27.1 API line)
- checkpoint: DeepSeek-V4-Flash-Vision-Exp W8A8, with all BF16 `bias_vl`
  tensors included in the checkpoint overlay
- release topology: TP4 / DP4 / EP16, block size 32

Results:

- targeted processor, ModelSlim, MoE router, and DSA tests pass (90 tests in
[...]

PR #15751, opened 2026-09-04
[Feature][Ops] Support original sparse indices in SCFA

QwertyJack:

This operator follow-up is stacked directly on #15457. DeepSeek V4 Vision builds `ori_sparse_indices` for bidirectional attention inside image spans, but the arch32 SCFA kernel currently reads original K/V through the contiguous paged-attention window. As a result, passing the indices alone does not make compressed SCFA layers consume the selected physical PA slots.

With #15457 plus this PR, the core DeepSeek-V4-Flash-Vision-Exp multimodal SCFA path is available. 

Related roadmap: #15462.

### What changed

- Detect the valid original sparse-index length from the first negative sentinel in each query row.
- Gather original K for MM1 and V for MM2 through the same physical PA slots with `DataCopyPABySlots`.
- Use the TND cumulative query-token offset when selecting each sparse-index row.
- Preserve the existing contiguous causal/SWA path when `ori_sparse_indices` is absent.
- Fall back to two PA copies when compressed paged blocks descend physically, preserving their logical order.
- Add dense-reference NPU coverage for FP16/BF16, compression ratios 1/4/128, non-contiguous PA blocks, K/V gather, and the production block-32 image geometry.

Yes. On arch32 with `layout_q=TND` and `layout_kv=PA_ND`, SCFA now applies the per-query original-KV physical slots supplied by the runtime. Image tokens can therefore attend bidirectionally within their image span while surrounding text retains causal sliding-window attention.

The rebased PR contains one operator commit whose parent is the current #15457 head `a0f4947bb`.

Current branch checks:

```text
python -m py_compile tests/e2e/nightly/single_node/ops/singlecard_ops/test_sparse_attn_sharedkv_scfa_indices.py
ruff format --check tests/e2e/nightly/single_node/ops/singlecard_ops/test_sparse_attn_sharedkv_scfa_indices.py
ruff check tests/e2e/nightly/single_node/ops/singlecard_ops/test_sparse_attn_sharedkv_scfa_indices.py
git diff --check a0f4947bb..HEAD
# passed
```

The three modified SCFA kernel sources are byte-identical to the sources used to build the retained Ascend 910B OPP package. That package passed the new dense-reference suite:

```text
9 passed, 15 warnings in 8.40s
```

The covered operator configurations are:

- query layout: `TND`;
- KV layout: `PA_ND`;
- KV heads: 1;
- index width: 512;
- PA block sizes: 16 and the service geometry 32;
- dtype: FP16 and BF16;
- compression ratio: 1, 4, and 128;
- absent-index compatibility path;
- non-contiguous and descending physical PA blocks.

The same operator sources were also used in the full W8A8 integration stack that aligned the fixed official-API teacher sequences and content-token top-1 results. That service evidence included additional runtime hardening from #15740, so it is integration evidence rather than an isolated #15457-plus-operator comparison.

### Scope boundaries

- This PR intentionally targets the DeepSeek V4 production combination `TND + PA_ND` on arch32.
- The existing host tiling restriction `kv_head_num == 1` remains unchanged.
- DSA context parallel is not changed.
- No Python runtime, frontend, RNG, or performance changes are included.

### DIFF
--- a/csrc/attention/sparse_attn_sharedkv/README.md
+++ b/csrc/attention/sparse_attn_sharedkv/README.md
@@ -78,7 +78,7 @@
 - 目前暂不支持返回`softmax_lse`，`return_softmax_lse`仅支持输入False，返回值`softmax_lse`为无效值。
 - ori_mask_mode及cmp_mask_mode所表示的mask模式的详细介绍见[sparse_mode参数说明](../../../docs/zh/context/sparse_mode参数说明.md)。
 - 目前暂不支持指定`q`中参与运算的token数，因此设置`seqused_q`无效。
-- 目前暂不支持对`ori_kv`进行稀疏计算，因此设置`ori_sparse_indices`无效。
+- `ori_sparse_indices`中的非负值是`ori_kv`的physical slot，按顺序参与计算；每行从首个负值开始为padding。
 - 目前所有输入不支持传入空tensor。
 - `q`、`ori_kv`、`cmp_kv`数据排布格式支持从多种维度解读，B（Batch）表示输入样本批量大小、S（Seq-Length）表示输入样本序列长度、H（Hidden-Size）表示隐藏层的大小、N（Head-Num）表示多头数、D（Head-Dim）表示hidden层最小的单元尺寸，且满足D=H/N、T表示所有Batch输入样本序列长度的累加和。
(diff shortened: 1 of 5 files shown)


ZT-AIA (approved):

QwertyJack:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

QwertyJack:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

QwertyJack:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #15755, opened 2026-09-04
[v0.26.0rc][Feature][Quantization] Support NZ for A5 MX quantization

lijiahang226:

Backports the A5 MX quantization FRACTAL_NZ support to v0.26.0rc.

A5 requires FRACTAL_NZ for MX quantized W8A8/W4A4 weights. Previously the code always kept the non-contiguous transpose layout, which is not compatible with NZ-enabled execution on A5.

This PR:
- Extends `maybe_trans_nz` to forward `customize_dtype` / `input_dtype` to `npu_format_cast` so FP8/FP4 MX weights can be converted to FRACTAL_NZ with correct dtype hints.
- Updates W8A8 MXFP8 and W4A4 MXFP4 weight processing to convert to FRACTAL_NZ when NZ is enabled, while preserving the old non-contiguous layout when NZ is disabled.
- Avoids double NZ conversion for layers processed by fused SFA preprocess (MLAPO/PROLOG_V3).
- Refactors SFA preprocess type resolution so the fused preprocess decision can be queried before submodule weight processing.

Related to #15865

No new user-facing API. Behavior is affected only when NZ is enabled on supported hardware; with NZ disabled, existing MX layout behavior is preserved.

- Unit tests: `pytest -sv tests/ut/attention/a2/test_sfa_v1.py tests/ut/ops/test_mla.py tests/ut/quantization/methods/test_w4a4_mxfp4.py tests/ut/quantization/methods/test_w8a8_mxfp8.py tests/ut/test_utils.py`

### DIFF
--- a/tests/ut/attention/a2/test_sfa_v1.py
+++ b/tests/ut/attention/a2/test_sfa_v1.py
@@ -1089,17 +1089,17 @@ class TestAscendSFAImpl(TestBase):
         path = self.impl._resolve_preprocess_type(torch.bfloat16)
         self.assertEqual(path, PreprocessType.PROLOG_V3)
 
-    def test_resolve_path_unquantized_c8_goes_prolog_v3(self):
-        """Unquantized + is_kv_consumer + C8 → PROLOG_V3 (blocked by reasons)."""
+    def test_resolve_path_unquantized_c8_goes_native(self):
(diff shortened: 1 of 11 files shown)


lijiahang226:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #15758, opened 2026-09-04
[v0.27.1rc][Performance] Skip per-layer indexer cache for GLM-5.1 IndexCache S layers

lijiahang226:

Backports the GLM-5.1 IndexCache optimization to v0.27.1rc.

GLM-5.1 IndexCache keeps an `Indexer` module on static S layers for weight-loading compatibility, but those layers reuse another layer's top-k indices at runtime. Previously they still allocated and wrote a per-layer indexer KV cache even though the computed indexer keys were never consumed.

This PR:
- Adds `_is_mtp_layer()` so MTP/nextn layers stay on the existing per-layer indexer path.
- Adds `runtime_has_indexer` to distinguish weight-loading-only modules from runtime indexer owners.
- Skips indexer pre-processing and indexer KV cache writes for static S layers.
- Skips indexer KV cache allocation in `NPUModelRunner.get_kv_cache_spec()` for those layers.
- Also fixes the sparse LI C8 quant-description suffix to use `.indexer.wq_b.weight`.

Related to #15866

No user-facing API or configuration change. Internal memory/compute savings for GLM-5.1 IndexCache deployments; MTP layers keep existing behavior.

- Unit tests: `pytest -sv tests/ut/attention/test_sfa_v1.py tests/ut/test_ascend_config.py tests/ut/worker/test_model_runner_v1.py`

### DIFF
--- a/tests/ut/attention/test_sfa_v1.py
+++ b/tests/ut/attention/test_sfa_v1.py
@@ -26,6 +26,7 @@ from vllm_ascend.attention.sfa_v1 import (
     AscendSFAMetadata,
     AscendSFAMetadataBuilder,
     PreprocessType,
+    _is_mtp_layer,
     custom_kv_rmsnorm_rope,
 )
 from vllm_ascend.attention.utils import get_sfa_qsfa_packed_head_dim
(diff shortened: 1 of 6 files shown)


lijiahang226:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

lijiahang226:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #15759, opened 2026-09-04
[v0.27.1rc][BugFix] Fix GLM5.2 MTP index cache

lijiahang226:

Backports the GLM5.2 MTP shared-index fix to v0.27.1rc.

GLM5.2 MTP with shared index was broken on Ascend because the runtime index-cache state was not correctly exposed or initialized:
- The actual indexer gate (`skip_topk`) and shared top-k buffer (`topk_indices_buffer`) live in the inner MLA impl, which is not visible to `named_modules()`, so upstream proposer hooks never reached it.
- MTP/nextn layers could start in skip mode incorrectly.
- After draft step 0, the shared top-k buffer rows were not compacted to request-aligned positions.

This PR fixes the Ascend MTP index-cache path by adding the missing `skip_topk` / `topk_indices_buffer` forwarding, masking initial `skip_topk` for MTP layers, and calling `compact_topk_indices` after draft step 0 when supported.

Related to #15864

No user-facing API or configuration change. This is an internal fix for GLM5.2 MTP index-cache sharing on Ascend.

- Unit tests: `pytest -sv tests/ut/ops/test_mla.py tests/ut/spec_decode/test_eagle_proposer.py`

### DIFF
--- a/tests/ut/ops/test_mla.py
+++ b/tests/ut/ops/test_mla.py
@@ -210,8 +210,7 @@ class TestAscendMultiHeadLatentAttention(TestBase):
         mock_mla_attn.impl = MagicMock()
         mock_mla_attn.impl.process_weights_after_loading = MagicMock()
 
-        with patch("vllm_ascend.ops.mla.MLAAttention",
-                   return_value=mock_mla_attn) as mock_mla_attn_cls:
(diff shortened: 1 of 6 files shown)


lijiahang226:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #15765, opened 2026-09-04
[CI] Update dspark mrv2 nightly

dragondream-chen:

Add five DeepSeek V4 nightly configurations for Model Runner V2:

- Flash single-node
- Flash PD-prefix
- Pro DSpark PD-prefix
- Pro DSpark 1M-PD
- Pro MTP PD-prefix

The PR also prevents the Chinese Read the Docs build from resolving tools/bisect instead of Python standard-library bisect when tools/generate_zh_docs.py is executed directly. This fixes the deterministic ImportError reported by build 34403297.

The MRV2 slot-mapping runtime fix is reviewed separately in #15817.

No.

- All five configurations passed their configured smoke, accuracy, and/or performance validation.
- Flash single-node GPQA: 88.89; three-run performance mean: 234.6171 requests/s.
- Flash PD-prefix AIME: 96.67.
- Pro DSpark PD-prefix AIME: 93.33; three-run performance mean: 471.8498 requests/s.
- Pro DSpark 1M-PD three-run performance mean: 16.9460 requests/s.
- Pro MTP PD-prefix AIME: 93.33; three-run performance mean: 471.1520 requests/s.
- Verified the Read the Docs failure mode with tools/ first on sys.path; after the fix, random imports the standard-library bisect.py module.
- git diff --check passes.


dragondream-chen:
Superseded by #15818. The replacement uses the no-codex branch ci/main_dsv4_mrv2_nightly and includes the Read the Docs fix for build 34403297.

PR #15769, opened 2026-09-04
[Feature][Model] Support DeepSeek V4 Vision with DSA-CP

GDzhu01:

This PR adds DeepSeek V4 Vision support to the legacy MRV1 DSA context-parallel path after the base Vision support landed in #15457.

The regular DSA metadata builder already creates bidirectional sliding-window indices for image spans. The DSA-CP builder did not carry those indices into its rank-local metadata, so Vision requests with `additional_config.enable_dsa_cp=true` lost the image-aware attention mask.

This change:

- builds the global Vision SWA indices from `mm_req_doc_ranges` in the legacy DSA-CP metadata builder;
- pads and slices them to the token range owned by each TP rank;
- forwards the rank-local indices to the DSA attention operator;
- fixes MRV1 DP graph padding so padded request slots reuse the last cumulative token offset instead of repeating the real scheduled-token vector. The old behavior can produce a source/destination shape mismatch while capturing a padded FULL_DECODE graph.

Yes. DeepSeek V4 Vision can now be served with the legacy MRV1 DSA-CP path by setting `additional_config.enable_dsa_cp=true`.

- `python -m py_compile` on the four changed Python files.
- On an Ascend 800I A3 container:

  ```bash
  PYTHONPATH=/vllm-workspace/dsv4-vision-dsa-cp-pr:/vllm-workspace/vllm \
    pytest -q tests/ut/attention/test_dsa_v1.py \
      tests/ut/worker/test_model_runner_v1.py
  ```

  Result: `87 passed`.

- Started `DeepSeek-V4-Flash-Vision-Exp-w8a8` with MRV1, DP=4, TP=4, EP enabled, DSpark, `enable_dsa_cp=true`, and `cudagraph_mode=FULL_DECODE_ONLY` on 16 Ascend devices. All 29 graph-capture buckets completed; `/health` and `/v1/models` returned successfully, and a real image chat-completion request returned HTTP 200 with the expected answer.

### DIFF
--- a/tests/ut/attention/test_dsa_v1.py
+++ b/tests/ut/attention/test_dsa_v1.py
@@ -506,6 +506,76 @@ def test_dsa_cp_device_metadata_tasks(
         builder._build_qli_metadata.assert_not_called()
 
 
+def test_dsa_cp_builds_and_slices_vision_swa_indices():
+    builder = _make_cp_builder(compressor_ratio=1)
+    builder.num_prefills = 1
+    builder.num_actual_tokens = 4
+    builder.seq_lens = torch.tensor([4], dtype=torch.int32)
+    builder.seq_lens_cpu = builder.seq_lens
+    builder.block_table = torch.tensor([[1, 2]], dtype=torch.int32)
+    builder.slot_mapping = torch.zeros((4, 2), dtype=torch.int32)
+    builder.model_config.hf_config.vision_n_layers = 1
+    builder.model_config.hf_config.vision_max_n_token = 3
+    builder.model_config.hf_config.sliding_window = 2
+    builder.common_ratio_to_sas_metadata = {
+        "_cpu_local": {
+            "qsl_cpu": torch.tensor([0, 3], dtype=torch.int32),
+            "sl_cpu": torch.tensor([3], dtype=torch.int32),
+        }
+    }
+    builder._ensure_device_local_metadata = MagicMock(
+        return_value=(
+            3,
+            6,
+            3,
+            6,
+            torch.tensor([0, 3], dtype=torch.int32),
+            torch.tensor([3], dtype=torch.int32),
+            None,
+        )
+    )
+    builder._get_cmp_seqlens_for_metadata = MagicMock(return_value=None)
+    builder._build_sas_metadata = MagicMock(return_value=builder.req_sas_metadata)
+    global_indices = torch.arange(12, dtype=torch.int32).view(4, 1, 3)
+    common_attn_metadata = SimpleNamespace(
+        num_reqs=1,
+        query_start_loc=torch.tensor([0, 4], dtype=torch.int32),
+        query_start_loc_cpu=torch.tensor([0, 4], dtype=torch.int32),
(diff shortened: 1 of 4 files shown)


vllm_ascend/attention/context_parallel/dsa_cp.py
```diff
@@ -905,6 +907,41 @@ def build_local_metadata() -> None:
...
+        vision_swa_indices = None
+        mm_ranges = getattr(common_attn_metadata, "mm_req_doc_ranges", None)
+        hf_config = self.model_config.hf_config
+        max_image_tokens = (
+            getattr(hf_config, "vision_max_n_token", 0) if getattr(hf_config, "vision_n_layers", 0) > 0 else 0
+        )
+        if has_prefill and max_image_tokens > 0 and mm_ranges:
+            assert num_actual_reqs is not None
+            assert self.num_actual_tokens is not None
+            global_vision_indices, _ = build_vision_bidirectional_swa_indices(
```
GDzhu01, line 919: [Efficiency] This metadata builder already shares device and local results through common_ratio_to_sas_metadata across KV-cache groups, but the vision indices are rebuilt and padded for every group. build_vision_bidirectional_swa_indices scales with token and window metadata and can add repeated work on multimodal prefills. Please cache the global or local vision indices under the shared cache, keyed by inputs that can differ, as is done for _device_local and _rope_local.

tests/ut/attention/test_dsa_v1.py
```diff
@@ -506,6 +506,76 @@ def test_dsa_cp_device_metadata_tasks(
         builder._build_qli_metadata.assert_not_called()
 
 
+def test_dsa_cp_builds_and_slices_vision_swa_indices():
```
GDzhu01, line 509: [Testability] This test mocks build_vision_bidirectional_swa_indices and therefore verifies only plumbing and slicing; it cannot catch an off-by-one between real global image ranges and the CP local slice. Please add one unmocked case with at least two requests and an image span crossing local_start or local_end, plus a padded tail, then assert the exact local sparse indices.

vllm_ascend/attention/context_parallel/dsa_cp.py
```diff
@@ -1932,6 +1970,8 @@ def _forward(
             kv_plan.add_dsa_sparse_attn_extra_kwargs(extra_attn_kwargs, cu_seqlens_ori_kv=local_seq_lengths_query)
         if swa_req_metadata.dspark_swa_indices is not None:
             extra_attn_kwargs["ori_sparse_indices"] = swa_req_metadata.dspark_swa_indices
+        if swa_req_metadata.vision_swa_indices is not None:
```
GDzhu01, line 1973: 【可靠性、多模态】当 DSpark 非因果索引与 vision 双向 SWA 索引同时存在时，这里会覆盖上一分支设置的 dspark_swa_indices，导致一种约束静默丢失。建议明确两类场景是否互斥并 fail-fast；若可并存，应定义合并/优先规则并增加多模态 speculative decode 用例。

PR #15801, opened 2026-09-04
[v0.27.1rc][Bugfix][Attention] Gate parallel-drafting seq-lens override

Dawn952:

This PR ports only the `attention_v1.py` parallel-drafting sequence-length behavior from #15681 to the v0.27.1 release branch and puts it behind an opt-in speculative-config switch.

- Add `skip_parallel_drafting_seq_lens_override` to `SpeculativeConfig`, defaulting to `False`.
- Keep the existing device-side `seq_lens` override when the option is omitted or disabled.
- Preserve the previously selected host-side sequence lengths when the option is enabled.
- Do not include the unrelated Kimi KDA early-return change from #15681.

The vLLM config schema is rebuilt after adding the plugin-owned field so the option is accepted inside `--speculative-config`, including through the nested `VllmConfig` schema.

Yes. Users can opt in with:

```json
{"skip_parallel_drafting_seq_lens_override": true}
```

> [!WARNING]
> Enabling `skip_parallel_drafting_seq_lens_override` makes parallel-drafting attention preserve host-side sequence lengths instead of using the device-side sequence-length buffers. Enable it only when the draft model explicitly requires host-side metadata; otherwise leave the option disabled.

The default is `False`, so existing behavior is unchanged.

- Added a unit test covering both switch states in the Ascend attention metadata builder.
- Added a unit test verifying that the new speculative-config field defaults to `False` in both the dataclass and Pydantic schemas.
- `ruff check` passed for all modified Python files.
- `python -m compileall` passed for all modified Python files.
- Full pytest execution was not run locally because the local Python environment does not contain vLLM or torch-npu; repository CI is expected to run the added tests.

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -334,7 +334,11 @@ class AscendAttentionMetadataBuilder(AttentionMetadataBuilder[AscendMetadata]):
         if isinstance(self.kv_cache_spec, CrossAttentionSpec):
             seq_lens = common_attn_metadata.seq_lens
             slot_mapping = common_attn_metadata.slot_mapping.to(torch.int32)
-        elif self.speculative_config and self.speculative_config.parallel_drafting:
+        elif (
+            self.speculative_config
+            and self.speculative_config.parallel_drafting
+            and not self.speculative_config.skip_parallel_drafting_seq_lens_override
+        ):
             seq_lens = common_attn_metadata.seq_lens
 
         attn_state = common_attn_metadata.attn_state
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -336,8 +336,8 @@ class AscendAttentionMetadataBuilder(AttentionMetadataBuilder[AscendMetadata]):
             slot_mapping = common_attn_metadata.slot_mapping.to(torch.int32)
         elif (
             self.speculative_config
-            and self.speculative_config.parallel_drafting
-            and not self.speculative_config.skip_parallel_drafting_seq_lens_override
+            and getattr(self.speculative_config, "parallel_drafting", False)
+            and not getattr(self.speculative_config, "skip_parallel_drafting_seq_lens_override", False)
         ):
             seq_lens = common_attn_metadata.seq_lens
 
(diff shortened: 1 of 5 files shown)


vllm_ascend/attention/attention_v1.py
```diff
@@ -334,7 +334,11 @@ def build(
         if isinstance(self.kv_cache_spec, CrossAttentionSpec):
             seq_lens = common_attn_metadata.seq_lens
             slot_mapping = common_attn_metadata.slot_mapping.to(torch.int32)
-        elif self.speculative_config and self.speculative_config.parallel_drafting:
+        elif (
+            self.speculative_config
+            and self.speculative_config.parallel_drafting
+            and not self.speculative_config.skip_parallel_drafting_seq_lens_override
+        ):
```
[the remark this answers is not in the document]
Dawn952: Addressed in 84ddd6b8e: both speculative-config accesses now use getattr(..., False), and the test covers a config object without the new field.

zhenwenqi2024 (approved):

Dawn952:
CI selection completed successfully, but `ci-gate` requires the selected tests to run. I do not have permission to add labels on this repository. Could a maintainer please add `ready-precise` and `ready-a5`? The selector reported both regular and A5 tests for this PR.

vllm_ascend/attention/attention_v1.py
```diff
@@ -334,7 +334,11 @@ def build(
         if isinstance(self.kv_cache_spec, CrossAttentionSpec):
             seq_lens = common_attn_metadata.seq_lens
             slot_mapping = common_attn_metadata.slot_mapping.to(torch.int32)
-        elif self.speculative_config and self.speculative_config.parallel_drafting:
+        elif (
+            self.speculative_config
+            and getattr(self.speculative_config, "parallel_drafting", False)
+            and not getattr(self.speculative_config, "skip_parallel_drafting_seq_lens_override", False)
+        ):
```
slippersss, line 341: Warning or documentation is needed for illustrating `skip_parallel_drafting_seq_lens_override`. Please choose to add a note about the impact.

slippersss (changes requested):

vllm_ascend/attention/attention_v1.py
Dawn952: Addressed in ac193eac0. When the option is enabled, SpeculativeConfig post-init now emits a warning that parallel-drafting attention preserves host-side sequence lengths instead of the device-side buffers and that the option should only be used when the draft model requires host-side metadata. A unit test verifies the warning text.

slippersss (approved):

PR #15806, opened 2026-09-04
[BugFix][Gemma4] initialize _enable_mm_lora for Gemma4 multimodal

yzlbig:

This PR initializes `_enable_mm_lora` in `AscendGemma4ForConditionalGeneration`.

The Ascend Gemma4 multimodal implementation overrides the upstream model initialization to support quantized vision tower adaptation. Since the parent initializer is not called, the `_enable_mm_lora` attribute is not initialized.

During multimodal inference (e.g. MMMU-Pro evaluation), the runtime accesses this attribute, which may cause runtime failures.

This patch adds:

`self._enable_mm_lora = False`

to keep the default behavior when multimodal LoRA is disabled and align with the upstream initialization logic.

---

No.

This change only fixes the internal initialization of the Gemma4 multimodal model. It does not introduce any API, interface, or behavior changes for users.

---

Tested with:

- Gemma4 multimodal model serving with vLLM-Ascend.
- MMMU-Pro evaluation through EvalScope.

The model can successfully initialize and run multimodal inference after this patch.

### DIFF
--- a/vllm_ascend/models/gemma4_mm.py
+++ b/vllm_ascend/models/gemma4_mm.py
@@ -80,6 +80,12 @@ class AscendGemma4ForConditionalGeneration(Gemma4ForConditionalGeneration):
         self.model_dtype = vllm_config.model_config.dtype
         self.vllm_config = vllm_config
 
+        # Keep consistent with Gemma4ForConditionalGeneration.
+        # _process_image_input accesses self._enable_mm_lora for multimodal LoRA handling.
+        # Initialize the attribute here to prevent AttributeError when processing image inputs.
+        lora_config = vllm_config.lora_config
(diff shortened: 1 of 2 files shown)


vllm_ascend/models/gemma4_mm.py
```diff
@@ -79,6 +79,8 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
         self.multimodal_config = multimodal_config
         self.model_dtype = vllm_config.model_config.dtype
         self.vllm_config = vllm_config
+        lora_config = vllm_config.lora_config
```
kunpengW-code, line 86: Add notes, explaining the reason for adding

kunpengW-code (approved):

PR #15817, opened 2026-09-05
[Bugfix][MRV2] Bound slot mapping block table staging

dragondream-chen:

MRV2 stages a complete request block-table row in the Ascend Triton slot-mapping kernel and converts it to fp32 before gathering physical block numbers. For long rows (for example, a 131072-entry padded row), the staged tensor plus per-token temporaries can exceed UB capacity during compilation, causing DeepSeek V4 Flash graph warmup/capture to hang before the service becomes ready.

This patch bounds the optimized staged-row path to power-of-two row sizes up to 16384 entries, the largest size already validated on A3. Larger rows use indexed GM loads in the same kernel. Slot-mapping semantics, CP ownership, padding, and the existing fast path are unchanged.

This is split from the DeepSeek V4 MRV2 nightly change in #15765 so the runtime fix can be reviewed independently.

No API or configuration change. Long block-table rows no longer overflow compiler UB resources; smaller rows retain the staged fast path.

- Added NPU regression coverage at the 16384-entry staged boundary and for a 131072-entry direct-load row. Both are compared with the upstream MRV2 reference kernel.
- Existing CP slot-mapping cases remain bit-for-bit aligned with the reference implementation.
- Ran the complete slot-mapping test file successfully (6 tests).
- Validated DeepSeek V4 Flash W8A8 MRV2 with graph mode and enable_npugraph_ex: service startup, curl smoke requests, full GPQA evaluation (88.89), and three stable performance runs (mean 234.6171 requests/s).

### DIFF
--- a/vllm_ascend/ops/triton/v2/block_table/compute_slot_mappings.py
+++ b/vllm_ascend/ops/triton/v2/block_table/compute_slot_mappings.py
@@ -22,6 +22,7 @@ def _compute_slot_mappings_kernel(
     PAD_ID: tl.constexpr,
     TRITON_BLOCK_SIZE: tl.constexpr,
     BLOCK_TABLE_PAD_SIZE: tl.constexpr,
+    USE_BLOCK_TABLE_STAGING: tl.constexpr,
 ):
     group_id = tl.program_id(0)
     batch_idx = tl.program_id(1)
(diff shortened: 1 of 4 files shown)


vllm_ascend/worker/v2/block_table.py
```diff
@@ -100,5 +107,6 @@ def compute_slot_mappings(
             PAD_ID=PAD_SLOT_ID,
             TRITON_BLOCK_SIZE=1024,
             BLOCK_TABLE_PAD_SIZE=self._block_table_pad_size,
+            USE_BLOCK_TABLE_STAGING=(self._block_table_pad_size <= _MAX_STAGED_BLOCK_TABLE_PAD_SIZE),
```
Liamup777, line 110: if _block_table_pad_size too large (> 16384 then once load a row ub will overflow 128k), add USE_BLOCK_TABLE_STAGING to fallback to load block number discretely

Liamup777 (approved):

vllm_ascend/ops/triton/v2/block_table/compute_slot_mappings.py
```diff
@@ -42,14 +43,17 @@ def _compute_slot_mappings_kernel(
...
-        mask=block_table_offsets < block_table_stride,
-        other=0,
-    ).to(tl.float32)
+    if USE_BLOCK_TABLE_STAGING:
+        # BLOCK_TABLE_PAD_SIZE is a compile-time upper bound for every group's
+        # row. The runtime stride mask keeps loads within the current group's
+        # row. Large rows deliberately skip this allocation because staging a
+        # complete fp32 row can exceed UB during Triton compilation.
+        block_table_offsets = tl.arange(0, BLOCK_TABLE_PAD_SIZE)
+        block_table_values = tl.load(
```
Liamup777: load the whole row from block_table if it wont cause ub overflow then use tl.gather to avoid scalar operations

vllm_ascend/ops/triton/v2/block_table/compute_slot_mappings.py
```diff
@@ -42,14 +43,17 @@ def _compute_slot_mappings_kernel(
...
-    # BLOCK_TABLE_PAD_SIZE is a compile-time upper bound for every group's row.
-    # The runtime stride mask keeps loads within the current group's row.
-    block_table_offsets = tl.arange(0, BLOCK_TABLE_PAD_SIZE)
-    block_table_values = tl.load(
-        block_table_ptr + req_state_idx * block_table_stride + block_table_offsets,
-        mask=block_table_offsets < block_table_stride,
-        other=0,
-    ).to(tl.float32)
+    if USE_BLOCK_TABLE_STAGING:
+        # BLOCK_TABLE_PAD_SIZE is a compile-time upper bound for every group's
```
AuroraEmiya, line 47: Would it cause any performance fallback?

zzzzwwjj (approved):

PR #15826, opened 2026-09-05
[Feature][Spec Decode] Evict DSpark KV cache outside the draft window

Tflowers-0129:
## Summary

Make `draft_window_size` a physical Qwen3 DSpark KV-cache window instead of only limiting the KV range read by draft attention.

- Publish `SlidingWindowSpec` for Qwen3 DSpark draft layers before cache groups are built.
- Keep target MLA/indexer caches and DSpark draft caches in separate native groups.
- Size block capacity and async in-flight-token accounting for the native layout.
- Allow the scheduler to recycle only draft-model blocks outside the committed window.
- Register separate K/V storages correctly in Mooncake while preserving packed-layout registration.
- Document window semantics and the fact that the target-model cache is not evicted.

This PR is independent of enabling DSpark ACLGraph; it provides the scheduler/cache ownership needed by eager and graph execution paths.

## Validation

- Qwen3 DSpark, KV grouping/allocation, and Mooncake tests: **140 passed, 2 deselected**
- The two deselected Mooncake thread timing tests fail identically on the unmodified upstream base in this environment; the new Mooncake registration tests pass.
- `ruff check`, `compileall`, and `git diff --check` passed.


vllm_ascend/models/qwen3_dspark.py
```diff
@@ -24,7 +25,38 @@
...
+    if isinstance(window_size, bool) or not isinstance(window_size, int) or window_size <= 0:
+        raise ValueError("draft_window_size must be a positive integer")
+
+    draft_config = vllm_config.speculative_config.draft_model_config.hf_config
+    dflash_config = dict(getattr(draft_config, "dflash_config", None) or {})
+    dflash_config.update(
+        use_swa=True,
+        swa_window_size=window_size,
+    )
+    draft_config.dflash_config = dflash_config
```
nu11ptr7, line 55: `_configure_dspark_draft_window` rebuilds `dflash_config` as a plain dict and overwrites the attribute, also forcing `use_swa=True` regardless of the model's own setting. Any reader that accesses `dflash_config.use_swa` via attributes (or the original object held extra fields) would break or silently lose settings. Consider updating in place and only overriding when the user actually set draft_window_size.

PR #15827, opened 2026-09-05
[BugFix][MoE] Align MC2 mask capacity with tensor parallelism

Tflowers-0129:
## Problem

MC2 mask buffers were sized from the local token count even when distributed execution padded the working token capacity to a tensor-parallel-aligned size.

## Change

Use the TP-aligned token capacity for MC2 mask allocation and preserve the real-token boundary used by consumers.

## Validation

- `tests/ut/test_ascend_forward_context.py`: **30 passed**
- `ruff check`, `compileall`, and `git diff --check` passed.

### DIFF
--- a/tests/ut/test_ascend_forward_context.py
+++ b/tests/ut/test_ascend_forward_context.py
@@ -15,6 +15,7 @@ from vllm_ascend.device.hardware_profile import get_hardware_profile
 @pytest.fixture(autouse=True)
 def reset_mc2_tokens_capacity(monkeypatch):
     monkeypatch.setattr(afc, "_mc2_tokens_capacity", None)
+    monkeypatch.setattr(afc, "_reserved_mc2_mask", None)
     monkeypatch.setattr(
         afc,
         "get_ascend_config",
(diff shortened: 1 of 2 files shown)


tests/ut/test_ascend_forward_context.py
```diff
@@ -243,6 +244,35 @@ def test_is_decode_only_node_false_without_recompute_scheduler(monkeypatch):
     assert afc._is_decode_only_node(vllm_config) is False
 
 
+@pytest.mark.parametrize(
```
ShySummer, line 247: Could we add a regression test that exercises the forward-context consumer as well, rather than only checking the reserved buffer shape?

The actual invariant we need to preserve here is that `mc2_mask.shape[0] == padded_num_tokens` for the TP-aligned forward batch. For example, with 90 tokens and TP=8, the forward path should produce a 96-element mask with the first 90 entries active and the padded tail inactive.

Testing this invariant directly would better capture the original failure mode and protect against future changes to either the allocation or forward-padding logic.

PR #15830, opened 2026-09-05
[BugFix][KV Pool] Align layerwise MTP final-block trim to lcm boundary and pass lease TTL to memcache batch_alloc

tyy0829:

This PR combines two KV Pool changes and supersedes #15672 (its commit is included here so a single review is enough).

**1. Align the layerwise MTP final-block trim to the lcm block boundary**

- Fixes an over-conservative trim in `KVPoolScheduler.get_num_new_matched_tokens` for the layerwise + EAGLE/MTP path.
- The old condition compared the external hit against `num_tokens - lcm_block_size`, treating it as the final granularity block's start. That equality only holds when the prompt length is a multiple of `lcm_block_size`. For unaligned prompts the threshold is lower than the real (lcm-aligned) block start, so a quantized hit that stops exactly on the final block's start boundary - without entering the block - is trimmed by one whole lcm block. This contradicts the code comment stating that partial hits stopping on an interior block boundary can be loaded as-is.
- The trim now compares against the lcm-aligned start of the block containing the last token (`(num_tokens - 1) // lcm_block_size * lcm_block_size`), so only hits that genuinely reach into the final block are trimmed. Behavior for lcm-aligned prompts is unchanged (covered by the existing unit test).
- Measured impact (DeepSeek-V4-Flash, `block_size=32`, `lcm_block_size=4096`, input length 16000, 90% prefix repetition): the quantized hit is 12288, exactly the start of the final block `[12288, 16384)`. The old threshold `16001 - 4096 = 11905` trimmed the hit to 8192, cutting the pool hit rate to 51.19%. With this fix the full 12288 tokens are loaded and the measured hit rate is 76.78% (expected `floor(0.9 * 16000 / 4096) * 4096 / 16000 = 76.8%`).

**2. Pass lease TTL to memcache batch_alloc** (from #15672)

The memcache_hybrid `batch_alloc` interface supports a `leaseTtlMs` parameter, but vllm-ascend currently only passes `keys` and `sizes`, so allocated blocks are not covered by a lease at allocation time.

- Add `lease_ttl_ms` parameter to `Backend.batch_alloc` and `MemcacheBackend.batch_alloc`, forwarded to the underlying memcache_hybrid store as `leaseTtlMs` (default `0` keeps existing behavior for other backends)
- Pass `LAYERWISE_READ_LEASE_TTL_MS` (5 min, the same constant already used by the layerwise load path) at both GVA allocation call sites in `_alloc_gvas_for_save` so newly allocated blocks are covered by a lease
- Update the UT assertion to cover the new argument

No API or configuration change. For layerwise KV pool + MTP workloads with lcm-unaligned prompt lengths, up to one additional granularity block can now be loaded from the pool instead of being recomputed (higher pool hit rate, shorter prefill). Memcache KV pool allocations are now covered by a lease at allocation time.

- Unit tests on the combined branch (both commits, rebased on latest main): `pytest tests/ut/distributed/ascend_store/` - 338 passed, including:
  - existing `test_layerwise_mtp_hit_uses_safe_load_extent` (aligned case: trim still applies, unchanged);
  - new regression test `test_layerwise_mtp_hit_on_final_block_boundary_not_trimmed` (unaligned case: hit stopping on the final block boundary is not trimmed; fails on the old condition);
  - `tests/ut/distributed/ascend_store/test_pool_worker.py::test_evicted_allocated_gva_is_reallocated` now asserts `batch_alloc` is called with the lease TTL.
- `ruff check` / `ruff format --check` pass.
- Manual verification on DeepSeek-V4-Flash with a layerwise memcache KV pool + MTP (input 16000, 90% prefix repetition): pool hit rate 51.19% -> 76.78%, consistent with the granularity model.

### DIFF
--- a/tests/ut/distributed/ascend_store/test_pool_scheduler.py
+++ b/tests/ut/distributed/ascend_store/test_pool_scheduler.py
@@ -163,6 +163,33 @@ class TestKVPoolScheduler(unittest.TestCase):
         self.assertTrue(load_spec.can_load)
 
     @patch("vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.pool_scheduler.LookupKeyClient")
+    def test_layerwise_mtp_hit_on_final_block_boundary_not_trimmed(self, mock_client_cls):
+        scheduler = KVPoolScheduler(
+            self._make_config(block_size=16, extra_config={"backend": "memcache"}),
+            use_layerwise=True,
+        )
+        scheduler.use_eagle = True
+        scheduler.cache_transfer_granularity = 16
+        scheduler._get_layerwise_hit_tokens = MagicMock(return_value=96)
+
+        request = MagicMock()
+        request.prompt_token_ids = list(range(100))
+        request.num_tokens = 101
+        request.request_id = "r1"
+        request.block_hashes = [b"h"] * 7
+
+        need, is_async = scheduler.get_num_new_matched_tokens(request, 0)
+
+        # The hit (96) stops exactly on the final block's start boundary
+        # ((101 - 1) // 16 * 16 = 96) and must be loaded as-is instead of
+        # being trimmed back by one whole lcm block.
+        self.assertEqual(need, 96)
+        self.assertFalse(is_async)
+        load_spec = scheduler.load_specs["r1"]
+        self.assertEqual(load_spec.kvpool_cached_tokens, 96)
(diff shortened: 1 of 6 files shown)


tyy0829:
Closing in favor of #15854, which contains both commits of this PR (e7550cc7834a3394ca9cb4084db688144143ab09 is an ancestor of its head) plus the layerwise reachable-store change that depends on them. Further review/testing will continue there.

PR #15834, opened 2026-09-05
[BugFix] Keep unsupported DeepSeek V4 weights in ND format

GDzhu01:

Keep DeepSeek V4 weights in ND format when they are consumed by execution paths that bypass the regular Linear method:

- add a per-layer `skip_weight_nz_conversion` flag to unquantized Linear weight post-processing;
- preserve ND for both the original weight and the optional FP32 copy;
- mark compressor `wkv` and `wgate` weights, which are passed directly to the custom compressor op;
- mark DSA `wo_a` when OTP is enabled or OLoRA TP is disabled, because those paths call `npu_transpose_batchmatmul` directly.

This is the `main`-branch counterpart of #15684. It is adapted to the refactored DeepSeek V4 package layout and the hardware-profile-based main branch. In particular, the DSA decision follows the active execution path rather than hard-coding a device type; the A5 quantized path uses a separate quantization method and is unaffected.

No API change. It fixes DeepSeek V4 runtime compatibility by preventing unsupported ND-to-NZ weight conversion on custom execution paths.

- Added a unit test covering the per-layer bypass, including the FP32 weight-copy path.
- `ruff check` and `ruff format --check` on all changed Python files.
- Python `compileall` on all changed Python files.
- Project codespell, forbidden-import, boolean-context-manager, and long-function checks on the changed files.
- Full device coverage is delegated to the vLLM Ascend PR CI matrix.

### DIFF
--- a/tests/ut/ops/test_linear.py
+++ b/tests/ut/ops/test_linear.py
@@ -59,6 +59,7 @@ class TestAscendUnquantizedLinearMethod(TestBase):
         mock_is_meta = mock.PropertyMock(return_value=False)
         type(self.layer.weight.data).is_meta = mock_is_meta
         self.layer.precast_fp32_weight = False
+        self.layer.skip_weight_nz_conversion = False
 
     @patch("vllm_ascend.utils.get_ascend_config")
     @mock.patch("torch_npu.npu_format_cast")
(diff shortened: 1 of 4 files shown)


GDzhu01:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

ningjingbengxiaohai (approved):

PR #15844, opened 2026-09-05
 [Performance][Cherry-pick][DSV4] Compute compressor metadata once per cache group per sub…

ZT-AIA:
Backport of #14994 to main.

DSV4 runs one AscendCompressor forward per attention layer, but every layer
in a cache group consumes the same compressor metadata. This PR:

- Adds get_or_compute_compressor_metadata in dsa_v1.py, memoizing the
  device-side compressor_metadata operator outputs on the current forward
  context keyed by cache group, so it runs only once per group per substep.
- Adds reset_compressor_metadata_cache and calls it in _run_merged_draft when
  a merged speculative draft forward advances to its next substep.
- Threads a cache_group_key (derived from the builder's first layer name)
  through AscendDSAReqMetadata in both dsa_v1.py and the CP variant dsa_cp.py,
  and routes both eager fallback paths through the shared helper.

Note: main has already refactored the v0.26 prefill/decode metadata split
into a unified AscendDSAReqMetadata and moved the eager fallback into
AscendCompressor._compute_metadata, so the cache key no longer needs the
metadata-type component from the original PR; everything else is a 1:1 port.

Test: new UT tests/ut/attention/test_dsv4_compressor_metadata_cache.py
verifies same-group reuse, cross-group isolation, reset between substeps,
and fresh-context isolation (operator invoked exactly 4 times).

### DIFF
--- a/vllm_ascend/attention/dsa_v1.py
+++ b/vllm_ascend/attention/dsa_v1.py
@@ -8,6 +8,7 @@ import torch.distributed as dist
 import torch_npu
 from vllm.config import CUDAGraphMode, VllmConfig
 from vllm.distributed import get_tensor_model_parallel_world_size
+from vllm.forward_context import get_forward_context
 from vllm.triton_utils import HAS_TRITON
 from vllm.v1.attention.backend import (
     AttentionBackend,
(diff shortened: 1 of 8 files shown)


vllm_ascend/attention/dsa_v1.py
```diff
@@ -76,6 +77,69 @@
 CompressorForwardOutput = tuple[torch.Tensor, torch.Tensor]
 CompressorOverlapOutput = tuple[CompressorForwardOutput, torch.npu.Event]
 
+_COMPRESSOR_METADATA_CACHE_KEY = "dsv4_compressor_metadata_cache"
+
+
+def reset_compressor_metadata_cache() -> None:
```
nu11ptr7, line 83: The metadata is memoized on the current forward context, but the reset is only wired into the draft loop (`_run_merged_draft`). If a plain DSV4 prefill/decode step reuses the same forward context across steps, the first step's cached compressor metadata (stale start_pos/block_table/cos/sin) would be returned. Worth confirming non-draft paths get a fresh context per step or resetting there too.

maoxx241 (approved):

ningjingbengxiaohai (approved):

PR #15855, opened 2026-09-05
[Feature] support Kimi K3 DSpark with DCP

weiguihua2:
This PR adds support for running Kimi K3 DSpark with Decode Context Parallelism (DCP) on Ascend.

The main changes include:

- Add DCP-aware DSpark query slot mapping:
  - Map global KV positions to the owning DCP rank and compact local KV positions.
  - Mark query slots as `-1` on non-owner ranks.
  - Support configurable DCP size, rank, and KV-cache interleave size.

- Add dedicated DCP metadata preparation for DSpark's parallel first-pass query block:
  - Build local sequence lengths from the complete draft query width.
  - Treat the DSpark draft query as decode metadata.
  - Skip the autoregressive MTP DCP path for parallel drafting.

- Align MLA and GDN metadata:
  - Slice the DCP MTP attention mask to the decode sub-batch.
  - Fix the DCP MTP causal mask before a rank's first local key.
  - Use `1 + num_speculative_tokens` as the target verification reorder threshold.
  - Preserve prefill classification for spec-width prompt chunks under DCP.

- Fix DCP dummy/FULL-graph metadata:
  - Synthesize isolated sequence and computed-token state for uniform-decode dummy runs.
  - Avoid inheriting metadata from the previous real batch.
  - Require the full prompt to be computed before selecting a DCP uniform-decode graph.

- Keep Mamba prefix-cache lookup on replicated state by excluding it from DCP KV sharding.

No

### DIFF
--- a/tests/e2e/pull_request/one_card/test_model_runner_v1_with_device.py
+++ b/tests/e2e/pull_request/one_card/test_model_runner_v1_with_device.py
@@ -435,6 +435,7 @@ def test_stateful_handoff_preserves_decode_graph(
 ):
     # Exercise the real dispatcher and DP synchronization using CPU metadata only.
     runner = NPUModelRunner.__new__(NPUModelRunner)
+    runner.dcp_size = 1
     runner.dp_size = dp_size
     runner.dp_rank = 0
     runner.parallel_config = SimpleNamespace(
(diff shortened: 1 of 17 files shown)


drslark (approved):
LGTM.

drslark:
A nightly should be added in future pr.

weiguihua2:
> A nightly should be added in future pr.

I will add nightly later.

pisceskkk (approved):
LGTM

linfeng-yuan (approved):

PR #15859, opened 2026-09-05
[BugFix][Attention] Backport DSA-CP sequence-parallel FP8 fix to v0.27.1rc

zhenwenqi2024:

SFA DSA-CP computes a complete o_proj result for each rank's local token shard after temporarily gathering the full TP weight. When the upstream decoder enables sequence-parallel MoE, o_proj.reduce_results is false because the decoder owns the following reduce-scatter. Gathering every local result into a replicated full tensor before that reduce-scatter causes the already-complete values to be summed once more across TP ranks.

This PR:

- teaches AscendFp8BlockLinearMethod to participate in TP full-weight switching while keeping the FP8 wrapper as the layer's execution method;
- reuses the internal A5 MXFP8 implementation's TP weight-layout specifications, including weight scales;
- stages each rank's complete local projection only in its token slot when the upstream decoder owns the reduce-scatter, avoiding the redundant SFA all-gather and duplicate summation;
- preserves the existing replicated-output path when reduce_results is true;
- adds regression coverage for the sequence-parallel handoff and FP8 TP weight-switch specifications.

Yes. DSA context parallelism can use block FP8 o_proj weights together with upstream sequence-parallel MoE without duplicate TP reduction. No configuration or public API changes are introduced.

    pytest -q tests/ut/attention/test_sfa_o_proj_tp.py tests/ut/quantization/methods/test_fp8_block.py
    # 30 passed, 14 warnings in 0.33s

    python -m compileall -q vllm_ascend/attention/context_parallel/sfa_cp.py vllm_ascend/quantization/methods/w8a8/fp8_block.py tests/ut/attention/test_sfa_o_proj_tp.py tests/ut/quantization/methods/test_fp8_block.py

The issue and duplicate TP reduction were reproduced on a two-node Ascend A5 setup with DP=2, TP=8, EP=16, block FP8 weights, and enable_dsa_cp=true. A final two-node run of this exact patch is pending.

### DIFF
--- a/tests/ut/attention/test_sfa_o_proj_tp.py
+++ b/tests/ut/attention/test_sfa_o_proj_tp.py
@@ -54,6 +54,7 @@ class TestAscendSFAOProjTPParams(TestBase):
             self.weight = torch.nn.Parameter(torch.randn(4, 3), requires_grad=False)
             self.weight_scale = torch.nn.Parameter(torch.randn(2, 3), requires_grad=False)
             self.quant_method = linear_method
+            self.reduce_results = True
 
     def setUp(self):
         AscendSFADSACPImpl.o_proj_full_pools.clear()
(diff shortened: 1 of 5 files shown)


zhenwenqi2024:
/rerun
[Bot]: rerun completed.

**Failed:**
- Image Build and Push: still in progress, retry /rerun after completion

PR #15860, opened 2026-09-05
[BugFix][MTP] Preserve hybrid cache layout for draft configs

bowgneo:

With Model Runner V2 and Qwen3.5 MTP, reconstructing `VllmConfig` for the attention-only draft model reruns platform configuration. `refresh_block_size()` then treats the draft as a regular non-hybrid model and changes the shared target cache block size from its resolved hybrid layout (1536 in the reproduction) to 128, causing a KV-cache/hash block-size assertion during startup.

Skip block-size refresh only when the active model config is a separate speculative draft model config. The draft shares the target model's `CacheConfig` and must retain the target's resolved cache layout. Model-free proposers such as n-gram alias `draft_model_config` to `target_model_config`, so the target still follows the normal block-size refresh path.

Fixes hybrid-model startup with Model Runner V2 and speculative decoding. No new flags or configuration changes are required.

- Direct configuration check: an n-gram target cache is refreshed from block size 16 to 128, while a separate draft config retains the target's resolved block size 1536.
- Hardware smoke test: Qwen3.5-27B W8A8, TP=4, Model Runner V2 and MTP enabled, with AscendStoreConnector/Mooncake. All four workers retained `hash_block_size=1536` and `lcm_block_size=1536`; `/health` and `/v1/completions` returned HTTP 200. Speculative-decoding metrics reported 12 drafted tokens and 5 accepted tokens.


jiajinzhu2 (approved):
LGTM.

lilinsiman (approved):

lilinsiman:
good

vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
```diff
@@ -125,15 +125,17 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
 
     def _create_draft_vllm_config(self) -> VllmConfig:
         """Build the runtime config used while executing the draft model."""
-        parallel_config = replace(
+        # This is a runtime view of the already configured model. Rebuilding
+        # VllmConfig re-runs platform defaults for the draft and can overwrite
+        # the target's shared cache config (e.g. Qwen3.5 MTP resets 1536 to 128).
+        draft_config = copy(self.vllm_config)
```
drslark: Add a comment explaining that the purpose is to skip the `__init__` and `__post_init__` methods of `vllm_config`.

drslark (changes requested):

vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
bowgneo: done

drslark (approved):
LGTM.

slippersss:
It seems that, this is not just the issue in the mrv2. If target is hybrid, `refresh_block_size` overrides `cache_config.block_size` when `replace(...)` is triggered and caller is draft. We should consider to fix `refresh_block_size`, since the mrv1 will also encounter the issue.

vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
```diff
@@ -125,15 +125,18 @@ def __init__(self, vllm_config: VllmConfig, device: torch.device):
...
+        draft_config.cache_config = copy(self.vllm_config.cache_config)
+        draft_config.parallel_config = replace(
             self.vllm_config.parallel_config,
             pipeline_parallel_size=1,
         )
-        return replace(
-            self.vllm_config,
-            model_config=self.draft_model_config,
-            parallel_config=parallel_config,
-        )
```
slippersss, line 136: Perheps mrv1 is also problematic. Maybe `refresh_block_size` should be checked.

slippersss (changes requested):

lilinsiman:
Final conclusion: A draft model judgment is added at the beginning of the refresh cache config.

weijinqian0 (dismissed):

slippersss:
LGTM.

slippersss (approved):

realliujiaxu (approved):

PR #15870, opened 2026-09-06
[v0.27.1rc][BugFix][Attention] Keep selected seq lens for parallel drafting

Dawn952:

Backport only the parallel-drafting sequence-length change from #15681 to releases/v0.27.1rc.

This removes the two-line override that replaced the sequence lengths selected earlier in the metadata builder with common_attn_metadata.seq_lens. The unrelated Kimi KDA change from #15681 is intentionally excluded.

Yes. Parallel drafting keeps the sequence lengths selected earlier in the metadata builder.

- Python AST syntax parsing
- git diff --check


linfeng-yuan (approved):
Note that we only guarantee Kimi-K3 and MinimaxM3 for this release candidate. So  we can remove the `seq_lens` for GQA paralllel-drafting

PR #15879, opened 2026-09-06
[BugFix] Fix duplicate SFA-CP prefill output reduction

MmMmaru:
This PR is independently based on the latest `main` and fixes the V3.2 SFA-CP full-weight prefill output being summed again by the upstream sequence-parallel decoder. After upstream MoE SP was introduced by `#13946`, the decoder applies SUM reduce-scatter to attention output. SFA-CP full-weight prefill already produces the same complete output on every TP rank, so passing it to that SUM scales it by TP size and changes the attention/residual balance.

The fix follows `o_proj.reduce_results`: when the caller performs the outer reduction, clear the full-shaped output buffer and write only this rank's token interval; otherwise retain all-gather for a complete output. Decode, DeepSeek V4/DSA-CP, V2 metadata, SFA-DCP, and MLA-CP communication paths are outside this PR.

No API change. SFA-CP sequence-parallel prefill now returns the output layout expected by the upstream decoder, restoring numerical behavior.

- Added unit coverage for TP sizes 2/4/16, token counts 1/3/16/17, both `reduce_results` modes, output-buffer reuse, and TP weight restoration on projection failure.
- Added a four-card V3.2 E2E matrix covering V1/V2 runners and eager/FULL_DECODE_ONLY, with CP on/off comparison while SP remains enabled. Prompt lengths are 1/3/4095/4096/4097 and each request generates eight tokens.
- Local static validation passed: `ruff-check`, `ruff-format`, `markdownlint`, Python AST parsing, and `git diff --check`.

### DIFF
--- a/tests/e2e/pull_request/four_card/context_parallel/test_sfa_cp_sp.py
+++ b/tests/e2e/pull_request/four_card/context_parallel/test_sfa_cp_sp.py
@@ -1,0 +1,69 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+import os
+from unittest.mock import patch
+
+import pytest
+from vllm import SamplingParams
+
+from tests.e2e.conftest import DPVllmRunner, wait_until_npu_memory_free
+
+SFA_CP_MODEL = os.environ.get("SFA_CP_TEST_MODEL", "vllm-ascend/DeepSeek-V3.2-W8A8-Pruning")
+
+
+@pytest.mark.e2e_model(SFA_CP_MODEL)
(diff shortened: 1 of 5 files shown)


czydyy:
/nightly multi-node-deepseek-v3.2-W8A8-EP
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/34077518081)

czydyy:
/nightly multi-node-deepseek-v3.2-W8A8-EP
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/34082285428)

PR #15894, opened 2026-09-07
[0.27.1rc][Feature] Support Kimi K3 DCP with replicated GQA DSpark cache

pisceskkk:

Backport Kimi K3 decode context parallelism with a replicated GQA DSpark draft KV cache to `releases/v0.27.1rc`.

- Shard target KV cache across DCP ranks and allocate replicated draft KV cache with matching block tables and slot mappings.
- Preserve KV cache spec padding and isolate target and draft physical layouts.
- Handle variable speculative query lengths and merge DCP attention results with the SFA Triton LSE implementation.
- Correct DSpark GDN decode thresholds and DCP chunked-prefill metadata handling.
- Support both the legacy three-field `KVCacheConfig` constructor and schemas exposing `prefix_cache_retention_interval`.

The release branch's verified vLLM pairing remains `ba07e4a48fc951300d97eb506217dd530583dea3`. The constructor compatibility coverage is scoped to that API boundary; end-to-end support for the native vLLM v0.27.1 tag still requires runtime validation.

Yes. Enables Kimi K3 target DCP with a GQA DSpark draft whose KV cache is replicated within the DCP group. Existing platform block-size selection policy remains unchanged.

Completed on this release backport:

- Ruff lint: all 24 changed Python files passed.
- Ruff format check: all 24 files passed.
- Python syntax compilation: passed.
- Git whitespace checks: passed.

Added coverage includes legacy/current KVCacheConfig constructors, physical cache allocation, replicated block tables and slot mappings, model-runner metadata, and Kimi K3 TP8 DCP1/2/4/8 E2E scenarios.

This is a draft pending execution of the unit and NPU E2E tests on the release branch. Static checks do not establish inference correctness or performance.


.github/vllm-main-verified.commit
pisceskkk: do not change this pin commit.

vllm_ascend/attention/context_parallel/attention_cp.py
```diff
@@ -123,20 +123,31 @@ def _build_backend_metadata(
         dcp_metadata = self._require_dcp_metadata(common_attn_metadata)
         prefill_metadata = None
         if num_prefills > 0:
+            # Parallel drafting keeps seq_lens on the accelerator, while
+            # query_lens is derived from query_start_loc_cpu. Copy query_lens
+            # asynchronously and keep the metadata arithmetic on device.
+            num_prefill_tokens = int(query_lens[num_decodes:].sum().item())
```
pisceskkk: please add `_cpu` for cpu tensor.

vllm_ascend/attention/context_parallel/mla_cp.py
```diff
@@ -71,7 +71,7 @@ def __init__(
         metadata_cls: type[AscendMLAMetadata] | None = None,
         supports_dcp_with_varlen: bool = False,
     ):
-        super().__init__(kv_cache_spec, layer_names, vllm_config, device, metadata_cls, supports_dcp_with_varlen)
```
pisceskkk: Why we change this? Is mla dcp impl supports dcp with varlen?

vllm_ascend/attention/context_parallel/common_cp.py
```diff
@@ -124,15 +125,13 @@ def _merge_dcp_attention_output(
...
-        return _npu_attention_update(
-            head_size,
-            _process_attn_out_lse(
-                attn_output,
-                softmax_lse,
-                dcp_size=self.dcp_size,
-                dcp_device_group=self.dcp_device_group,
-            ),
-            dcp_size=self.dcp_size,
+        del head_size
```
pisceskkk: do we still need `head_size` as param for this function? could we just remove it from params list?

vllm_ascend/models/kimi_k3.py
```diff
@@ -765,6 +765,8 @@ def forward(
             self.output_attn_res_norm,
             attn_res_block_num,
         )
+        if self.dspark_aux_capture_materialized and self.end_layer in self.aux_hidden_state_layers:
+            aux_hidden_states.append(hidden_states)
```
pisceskkk: Why do we need this?

vllm_ascend/models/qwen3_dspark.py
```diff
@@ -80,8 +81,10 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = "") -> None:
                 config=config,
                 prefix=maybe_prefix(model_prefix, "confidence_head"),
             )
-        self.rotation_path = get_rotation_path(vllm_config) if vllm_config.quant_config is not None else None
-        self.target_model_path = Path(vllm_config.model_config.model)
+        target_model_config = vllm_config.speculative_config.target_model_config or vllm_config.model_config
+        target_vllm_config = replace(vllm_config, model_config=target_model_config)
+        self.rotation_path = get_rotation_path(target_vllm_config) if vllm_config.quant_config is not None else None
+        self.target_model_path = Path(target_model_config.model)
```
pisceskkk: Do we need this? Or if the model weights config file has problem?

vllm_ascend/patch/platform/patch_kv_cache_utils.py
pisceskkk: we have resolve dcp branch here, why change the below codes?

vllm_ascend/spec_decode/dspark_proposer.py
```diff
@@ -99,11 +117,174 @@ def __init__(
...
+            # GQA backend setup normalizes its cache block size to 128.
+            # Keep the parent hybrid model's page geometry unchanged.
+            cache_config=copy.deepcopy(base.cache_config),
+            # The target runner owns the PD connector and transfers all cache
+            # groups. The model-only draft config must not validate that
+            # connector's target topology against its local DP/DCP settings.
+            kv_transfer_config=None,
+        )
+
+    def _build_replicated_block_table(
```
pisceskkk: could we move `_build_replicated_block_table` and `_build_replicated_context_slot_mapping` into dcp_manager?

.github/vllm-main-verified.commit
pisceskkk: Restored in 4b1b224ea. `.github/vllm-main-verified.commit` now matches the release base (`ba07e4a48fc951300d97eb506217dd530583dea3`), with no remaining PR diff in this file. The earlier `6e448d0ea9` pin was brought in from the experiment and should not have changed the release pairing. The earlier NPU results used that experimental pairing; they are not validation of the restored release pin.

vllm_ascend/attention/context_parallel/attention_cp.py
pisceskkk: Updated in 4b1b224ea: the CPU input is explicitly named `query_lens_cpu`, and the token-count calculation uses that tensor. Only the prefill slice is then copied to `seq_lens.device` as `prefill_query_lens`. The inherited `query_lens` keyword is retained at the method boundary to preserve the base-class call interface.

vllm_ascend/attention/context_parallel/mla_cp.py
pisceskkk: You are right to question this. The forced `True` overstated the capability: `_forward_decode` still reshapes packed queries with `view(num_decodes, -1, ...)`, which assumes a rectangular per-request query width. Passing per-request sequence lengths and an MTP mask does not establish support for arbitrary ragged decode query lengths. In 4b1b224ea, I restored forwarding `supports_dcp_with_varlen` (default `False`) and added regression cases for both argument values. Fixed-width speculative verification should not bypass the general varlen capability guard. The new regression cases have been syntax/lint checked, but not executed on the release runtime yet.

vllm_ascend/attention/context_parallel/common_cp.py
pisceskkk: Removed in 4b1b224ea. `_merge_dcp_attention_output` now accepts only `attn_output` and `softmax_lse`; both GQA and MLA callers, the fused-merge test, and the MLA test stubs have been updated. The fused operator obtains the dimensions from the tensors, so `head_size` was redundant.

vllm_ascend/models/kimi_k3.py
pisceskkk: This covers the requested auxiliary state at `end_layer`. With materialized DSpark capture, the loop captures layer inputs at indices in `[start_layer, end_layer)`, so it never captures index `end_layer`. The final state must be appended after `_apply_ascend_attn_res` materializes the output residual. Otherwise a draft requesting the last target state receives an incomplete auxiliary-state list. The existing adapter regression requests `(1, 2)` in a two-layer model and checks both intermediate and final materialized states. I kept this change and documented the boundary in 4b1b224ea; it only runs when `end_layer` is explicitly requested.

vllm_ascend/models/qwen3_dspark.py
pisceskkk: This is model ownership rather than a malformed weight config. The replicated drafter is constructed with the draft `ModelConfig`, so `vllm_config.model_config.model` now points to the draft checkpoint. However, `load_weights` rotates the draft FC projection into the target hidden-state basis and, when the draft omits its own embedding/lm_head, loads those tensors from the target checkpoint. Both operations therefore need `speculative_config.target_model_config`, not the draft path. Using the draft path can select the wrong rotation asset or look for fallback target weights in the draft directory. I retained this and added an ownership comment in 4b1b224ea.

vllm_ascend/patch/platform/patch_kv_cache_utils.py
pisceskkk: The earlier branch returns for `dcp != 1`; this fallback specifically handles DCP=1, which is also used by one side of P_DCP1→D_DCP2 or P_DCP2→D_DCP1. With prefix caching and aligned Mamba groups, the upstream resolver can return a hash unit that does not divide every attention block size. This code preserves the upstream scheduler block size and adjusts only that incompatible hash unit, using an explicit `prefix_match_unit` when valid, otherwise the GCD. It leaves the upstream result unchanged when already compatible. I kept the narrowly guarded fallback and made its DCP=1 scope explicit in 4b1b224ea.

vllm_ascend/spec_decode/dspark_proposer.py
pisceskkk: These operations are related to DCP, but I would keep them on the proposer for this PR. They translate the target block table into the draft’s full-context replicated layout and own per-draft-group buffers. The target `DCPManager` is created only when target DCP>1, whereas this layout path also runs at DCP=1, and the replicated draft deliberately uses a DCP=1 model configuration. Moving them directly onto the target manager would require introducing a manager for DCP=1 or maintaining a second path, and would mix target-sharded metadata with draft-replicated buffer ownership. A dedicated replicated-draft layout helper would be a reasonable separate extraction; the current methods and buffer ownership are retained here. No code relocation was made for this comment.

PR #15895, opened 2026-09-07
[MRV2][BugFix]Fix backend input string path for sfa.

ZhangwenTaoHW:

### DIFF
--- a/vllm_ascend/worker/worker.py
+++ b/vllm_ascend/worker/worker.py
@@ -1133,9 +1133,8 @@ class NPUWorker(WorkerBase):
             and speculative_config.method == "eagle3"
             and speculative_config.num_speculative_tokens > 1
         )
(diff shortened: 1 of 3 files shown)


vllm_ascend/worker/worker.py
```diff
@@ -1042,8 +1042,9 @@ def initialize_from_config(self, kv_cache_config: KVCacheConfig) -> None:
             and speculative_config.method == "eagle3"
             and speculative_config.num_speculative_tokens > 1
         )
+        needs_mrv2_mamba_zeroing = self.use_v2_model_runner and kv_cache_config.has_mamba_layers
```
nu11ptr7: KV zeroing for the v2 runner is now gated on `has_mamba_layers`. If `needs_kv_cache_zeroing` also covers non-mamba v2 caches (sleep/wake reuse for DSV4/MLA), those would silently stop being zeroed; worth confirming the flag is only set for mamba.

weijinqian0 (approved):

PR #15907, opened 2026-09-07
[0.27.1rc][BugFix][MTP] Preserve hybrid cache layout for draft configs

bowgneo:

This applies the Model Runner V2 draft block-size fix from #15860 to the `releases/v0.27.1rc` branch.

With Qwen3.5 MTP, reconstructing `VllmConfig` for the attention-only draft model reruns platform configuration. `refresh_block_size()` then treats the draft as a regular non-hybrid model and changes the shared target cache block size from its resolved hybrid layout (1536 in the reproduction) to 128, causing a KV-cache/hash block-size assertion during startup.

Skip block-size refresh only when the active model config is a separate speculative draft model config. The draft shares the target model's `CacheConfig` and must retain the target's resolved cache layout. Model-free proposers such as n-gram alias `draft_model_config` to `target_model_config`, so the target still follows the normal block-size refresh path.

Fixes hybrid-model startup with Model Runner V2 and speculative decoding. No new flags or configuration changes are required.

- Direct configuration check: an n-gram target cache is refreshed from block size 16 to 128, while a separate draft config retains the target's resolved block size 1536.
- Hardware smoke test: Qwen3.5-27B W8A8, TP=4, Model Runner V2 and MTP enabled, with AscendStoreConnector/Mooncake. All four workers retained `hash_block_size=1536` and `lcm_block_size=1536`; `/health` and `/v1/completions` returned HTTP 200. Speculative-decoding metrics reported 12 drafted tokens and 5 accepted tokens.


lilinsiman (approved):

bowgneo:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

realliujiaxu (approved):

weiguihua2 (approved):

PR #15912, opened 2026-09-07
[Feature][KV Transfer] Add D2RH host cache with local HBM prefix hits

wangxiaoteng888:

- Add `MooncakeD2RHConnectorV1`, retaining D-side Host staging/cache while reusing local HBM prefixes.
- Keep the existing opt-in partial-group prefix lookup for DeepSeek-V4 remote-prefill consumers (`D2RH_EXTERNAL_PARTIAL_PREFIX_CACHE=1`).
- Refresh the model block size at Mooncake scheduler initialization, before snapshotting it. Resolve the DSV4 transfer block size from its sliding-window MLA spec, rather than the hybrid allocator quantum or the largest state-cache spec.
- Remove the late D2RH `request_finished` block-size rewrite: the scheduler must use consistent geometry throughout the request lifecycle, not just in its final response.
- Add connector scheduler regression tests and focused D2RH PP mapping/component/lifetime tests. No new per-request INFO logging is introduced by these updates.

Hybrid cache initialization can replace the logical DSV4 block size with a smaller physical allocation quantum. Using that value for connector metadata/hash indexing loses valid D-side HBM prefix hits. The initialization fix aligns scheduler metadata with the worker's logical KV layout.

Yes: an opt-in D2RH connector with Host-cache reuse, local-HBM prefix reuse, and prefill-PP transfer support. This PR does not modify the example proxy or add a rank-affinity option. Routing policy is outside its scope.

The normal D2RH path retains the D Host cache. The diagnostic benchmark that skipped P/KV handoff is not the production implementation and is not included in this PR.

### Historical targeted tests before proxy separation

- Repository-pinned Ruff **0.14.0** check and format check on the changed Python files: passed.
- Static compilation of the changed Python files: passed.
- Full `bash format.sh ci` was attempted but stopped while initializing the Markdown hook's dependencies on Windows; the complete all-files hook suite is **not** reported as passed.
- `git diff --check`: passed.
- In an isolated source tree on the A5 day20260904 container, the Mooncake connector + proxy test files: **140 passed, 1 failed**. The sole failure is the existing `TestKVCacheRecvingThreadBasic::test_submit_request_serializes_same_peer_fifo` one-second thread-start wait. The same test also fails on the unchanged PR head `09a366ef477337d6caa01ddf49ad35a557ee3ed6` in the same environment. It is not reported as a green full suite.
- `tests/ut/test_compressed_prefix_cache.py`: **12 passed**.
- At that earlier revision, the scheduler and proxy tests reported **16 passed** (125 unrelated tests deselected). The proxy-specific tests have since been removed with the proxy change; this is a historical result, not the current test count. Existing review threads are not automatically marked resolved.

### Recorded hardware validation of the runtime fixes now ported into this update

Image: `vllm-ascend:dev-26.2.0.day20260904-A5-py311-Ubuntu24.04-lts-aarch64`; model: DeepSeek-V4-Flash. P and D each use DP4/TP1 (four NPUs each), both prefix caches enabled, CPU binding disabled, MTP=1. D uses the recompute scheduler. Normal PD benchmarks use matching-rank routing. These are previously completed runtime measurements using an experiment-specific matching-rank proxy, not a fresh performance run of the final PR commit with the unmodified example proxy. That routing modification is no longer included in this PR.

Prefill-oriented case: 128K input, 1 output token, 98% repeated prefix, concurrency 32, 320 requests.

- PD proxy: **3.5694 req/s**; direct P: **3.4910 req/s** (PD +2.25%).
- Both: **320/320 successful**, zero failures, identical actual input-length distributions, all outputs exactly 1 token.
- Actual P/D local prefix hit rates for PD: approximately **96.87%**. Dataset repetition is not the same as measured cache hit rate.

Decode-oriented case: 128K input, 1024 output tokens, 100% repeated data, concurrency 40, 320 requests.

- PD proxy: **2087.7728 output token/s**, average TTFT **3638.1 ms**, average end-to-end **18823.9 ms**.
- Direct D (mixed/prefill+decode path): **2167.4741 output token/s**, average TTFT **1589.2 ms**, average end-to-end **18445.4 ms**.
- PD output throughput is **3.68% lower**, not identical. Both: **320/320 successful**, all outputs exactly 1024 tokens, no D preemptions.
- D local prefix hit rate: **99.996948%** (131072/131076). P still computes approximately 4100 prompt tokens/request; it is not a P-side 100%-hit test.
- D2RH Host-cache logs show hits=304/misses=0 per formal PD request. This is not proof that every hybrid KV group needs zero H2D traffic; no isolated H2D-byte/timing measurement is claimed.

At concurrency 32, the earlier PD result was 1568.9433 output token/s, but a same-configuration recheck reached 1880.9528 without production code changes. Consequently, the earlier approximately 20% gap is not presented as a fixed connector transfer penalty. Runs are not a randomized multi-repeat significance study, and overall throughput comparisons do not imply equal per-NPU efficiency (PD uses eight NPUs, direct P/D four).

The prior description's unconditional “formal Host-to-HBM transfers: 0” statement has been removed: the new-image validation supports local-HBM and Host-cache reuse, not a universal zero-H2D claim. Production services were not replaced/restarted while preparing this PR update.

## Proxy separated from this PR

The example proxy is restored exactly to the PR base, and `tests/ut/examples/test_pd_proxy_rank_affinity.py` is removed. D2RH Host/HBM cache and PP connector changes remain unchanged. The removal received Ruff 0.14.0 check/format, Python compilation, base-file equality, and diff-whitespace checks. The complete format hook suite remains unverified because Markdown environment installation hangs. Running services and the experiment proxy were not modified or restarted. No new accuracy or performance claim for disabling rank affinity is made.


Yuli-yx (changes requested):
这轮复审在当前 HEAD 上确认了 6 组 correctness / lifecycle 问题。我把可以安全表达的修复都写成了 GitHub suggested changes；其中 TP 部分失败与 Hop 1 异常各自包含两条必须配套应用的建议。建议补充对应 CPU/mock 回归测试后再合入。

Yuli-yx (approved):
Withdrawing my previous CHANGES_REQUESTED review and all inline comments. This approval only clears that review state; it is not a merge-readiness endorsement. I moved the proposed fixes into an optional stacked PR for the author to evaluate independently: https://github.com/wangxiaoteng888/vllm-ascend/pull/5

Tflowers-0129 (approved):

PR #15940, opened 2026-09-07
[v0.26.0rc][Refactor][Ops]  Unify negate_sin handling in inplace_partial_rotation

ZT-AIA:
### What this PR does / why we need it:
Two call sites of inplace_partial_rotary_mul pass -sin to apply the inverse rotation to attention outputs ( dsa_v1.py o_proj_input and dsa_cp.py local_attn_output ), while the other 11 call sites pass sin directly for the forward rope. This mixes sign semantics into the callers.

This PR moves the negation into the torch binding via a new negate_sin flag (default False ):

- csrc/torch_binding.cpp : schema and impl add bool negate_sin=False ; the binding negates r2 internally before invoking aclnnInplacePartialRotaryMul
- csrc/torch_binding_meta.cpp : meta signature updated accordingly
- The two inverse-rotation call sites now pass the original sin plus negate_sin=True ; all other call sites are unchanged (default False )
No functional change; the schema default keeps all existing calls (including UT mocks) compatible.

### Does this PR introduce any user-facing change?:
No.

### How was this patch tested?:
CI verification (existing DSA/CP attention UTs cover both forward and inverse rotation paths).

### DIFF
--- a/csrc/torch_binding.cpp
+++ b/csrc/torch_binding.cpp
@@ -1643,8 +1643,7 @@ void inplace_partial_rotary_mul_npu(at::Tensor & x, const at::Tensor &r1, const 
     }
     auto origin_dim_num = x.dim();
     TORCH_CHECK(origin_dim_num == BSND_DIM_NUM, "Input tensor x's dim num should be 4, actual ", origin_dim_num, ".");
-    const at::Tensor &sin = negate_sin ? at::neg(r2) : r2;
-    EXEC_NPU_CMD(aclnnInplacePartialRotaryMul, x, r1, sin, it->second, partial_slice);
+    EXEC_NPU_CMD(aclnnInplacePartialRotaryMul, x, r1, r2, it->second, partial_slice, negate_sin);
 }
 
 std::tuple<at::Tensor, at::Tensor> npu_rms_norm_dynamic_quant_npu(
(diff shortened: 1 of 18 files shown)


csrc/torch_binding.cpp
```diff
@@ -1626,7 +1626,7 @@ std::tuple<at::Tensor, at::Tensor, at::Tensor> npu_hc_pre_sinkhorn_npu(
     return std::tuple<at::Tensor, at::Tensor, at::Tensor>(y, post, comb_frag);
 }
 
-void inplace_partial_rotary_mul_npu(at::Tensor & x, const at::Tensor &r1, const at::Tensor &r2, c10::string_view rotary_mode, at::IntArrayRef partial_slice)
+void inplace_partial_rotary_mul_npu(at::Tensor & x, const at::Tensor &r1, const at::Tensor &r2, c10::string_view rotary_mode, at::IntArrayRef partial_slice, bool negate_sin)
```
yiz-liu, line 1629: Use `inverse` instead of `negate_sin`.

ZT-AIA: More intuitively, and emphasizing that only sin is negate.

zzzzwwjj (approved):

PR #15943, opened 2026-09-07
[Model][Performance] Optimize DeepSeek V4 vision operators

GDzhu01:

This PR optimizes the DeepSeek V4 vision transformer by replacing hand-written operator compositions with standard vLLM layers:

- use `ApplyRotaryEmb` for vision RoPE while retaining FP32 computation;
- use `MMEncoderAttention` for vision attention;
- use `SiluAndMul` for the gated MLP;
- use vLLM `RMSNorm` with FP32 weights;
- keep the DeepSeek V4 multimodal processor override compatible with the base-class call signature;
- add numerical-equivalence coverage for attention, a complete vision block, and the full 32-layer real-weight ViT.

The change reduces Python-level small-operator composition in the ViT path while preserving the original numerical behavior.

No API or configuration change. DeepSeek V4 vision serving uses the same model and launch arguments, with a faster ViT implementation.

- DeepSeek V4 multimodal preprocessing unit tests: `4 passed`.
- ViT numerical-equivalence tests: `7 passed` across 4x4, 16x24, and 48x72 image-token grids.
- Full 32-layer real-weight ViT comparison:
  - max absolute error: `0.01953125`
  - mean absolute error: `0.0007206165`
  - cosine similarity: `0.9999631643`
- Real serving smoke with DP4/TP4 and expert parallelism: service initialized successfully in `FULL_DECODE_ONLY` graph mode and passed a real-image request.
- MMMU-Pro image subset: `1730/1730` requests passed at concurrency 128; no indexer out-of-bounds error, worker failure, or traceback was observed.
- Controlled 128-request A/B runs at concurrency 128, with identical service configuration and warmup:
  - baseline: `23.8163 s`, `23.1940 s` (mean `23.5052 s`)
  - candidate: `22.2728 s`, `16.6807 s` (mean `19.4767 s`)
  - mean elapsed time reduced by `17.1%`; mean per-run request throughput increased by `23.2%`
- `ruff check`, `ruff format --check`, and `git diff --check` passed.

### DIFF
--- a/tests/e2e/pull_request/one_card/test_deepseek_v4_vision_precision.py
+++ b/tests/e2e/pull_request/one_card/test_deepseek_v4_vision_precision.py
@@ -1,0 +1,218 @@
+import json
+import os
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+import torch
+import torch.nn.functional as F
+import torch_npu  # noqa: F401
+from safetensors.torch import load_file
+from vllm.config import VllmConfig, set_current_vllm_config
+
+from vllm_ascend.models.deepseek_v4.vision import (
+    DeepseekV4VisionAttention,
+    DeepseekV4VisionBlock,
+    DeepseekV4ViT,
+    get_vision_cos_sin,
+)
+from vllm_ascend.utils import register_ascend_customop
+
+VISION_DIM = 1024
+VISION_HEADS = 16
+VISION_INTER_DIM = 2816
+RMS_NORM_EPS = 1e-6
+
+
+@pytest.fixture(scope="module", autouse=True)
+def register_ascend_vision_ops():
+    register_ascend_customop()
+    with set_current_vllm_config(VllmConfig()):
+        yield
+
+
+@pytest.fixture
+def vision_config():
+    return SimpleNamespace(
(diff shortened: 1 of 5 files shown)


tests/e2e/pull_request/one_card/test_deepseek_v4_vision_precision.py
```diff
@@ -0,0 +1,218 @@
...
+        max_abs_error=0.0625,
+        mean_abs_error=0.004,
+        min_cosine_similarity=0.9999,
+    )
+
+
+def test_real_weights_full_vit_matches_original_formula():
+    model_path = os.getenv("DEEPSEEK_V4_VISION_MODEL_PATH")
+    if not model_path:
+        pytest.skip("DEEPSEEK_V4_VISION_MODEL_PATH is not set")
```
GDzhu01, line 178: [Portability / Testability] Silently skipping when DEEPSEEK_V4_VISION_MODEL_PATH is absent allows a misconfigured nightly runner to report green without executing the real-weight full-ViT comparison. Please make the scheduled CI variant fail fast when its model fixture is missing, while retaining an explicitly marked optional path for local development. That keeps the same test meaningful across runner environments.

PR #15952, opened 2026-09-07
[0.27.1rc][Performance][KDA] Drop redundant Kimi KDA padding mask

MaybeChz:
Cherry-pick of PR #15553 onto `releases/v0.27.1rc`.

Original PR: #15553

---

Kimi KDA previously masked `core_attn_out` after `o_norm` with `query_start_loc[-1]`. That device mask is redundant:

- `num_actual_tokens` is the unpadded live token count from the scheduler.
- Kernel output already overwrites `[:num_actual_tokens]`.
- Dummy seqs with `seqLen=0` live in request-dimension metadata, not extra rows inside the live token slice.

Keep GDN-style `torch.zeros` for the graph window. `o_proj` still runs on `hidden_states.size(0)` (`T_pad`), so `[num_actual_tokens:]` must stay zero instead of `torch.empty` garbage.

Merge still copies the full kernel tensors. Do not Python-slice index buffers to a live count; FULL ACLGraph inlines `_forward` and would bake that length, which is the CANN IndexCheck crash (`1065369473..1065369479`).

No. Internal KDA padding handling only.

- Unit tests in `tests/ut/ops/test_kimi_kda.py`:
  - Reproduces capture-length `index_copy_` against DSpark garbage indices.
  - Decode / spec / mixed `_forward` overwrite live rows.
- Please re-run the Kimi K3 DSpark + ACLGraph accuracy case.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/vllm_ascend/ops/kimi_kda.py
+++ b/vllm_ascend/ops/kimi_kda.py
@@ -150,28 +150,6 @@ class _KDAFusedBFGLinear(MergedColumnParallelLinear):
         param_shard.copy_(fused_weight)
 
 
-def _zero_padded_output(
-    output: torch.Tensor,
-    num_live_tokens: torch.Tensor,
-) -> torch.Tensor:
-    """Clear graph-padding rows using a device-side live-token count."""
-    token_indices = torch.arange(
-        output.shape[1],
-        dtype=num_live_tokens.dtype,
-        device=output.device,
-    )
-    valid_tokens = token_indices < num_live_tokens
-    return torch.where(valid_tokens.view(1, -1, 1, 1), output, 0.0)
-
-
-def _zero_padded_recurrent_output(
-    output: torch.Tensor,
-    query_start_loc: torch.Tensor,
-) -> torch.Tensor:
-    """Clear graph-padding rows skipped by recurrent KDA."""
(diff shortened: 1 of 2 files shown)


vllm_ascend/ops/kimi_kda.py
```diff
@@ -164,7 +142,9 @@ def forward(
             g1 = self.f_b_proj(f_a)[0]
             g1 = rearrange(g1, "n (h d) -> 1 n h d", d=self.head_dim)
             g2 = rearrange(g_proj_states, "... (h d) -> ... h d", d=self.head_dim)
-            core_attn_out = torch.empty(
+            # o_proj consumes the full T_pad window. zeros keep
+            # [num_actual_tokens:] from leaking uninitialized values.
+            core_attn_out = torch.zeros(
```
nu11ptr7, line 248: With FULL ACLGraph, `forward` runs once at capture, so the `torch.zeros` initialization also happens once - padding rows stay zero only as long as nothing writes them on replay. The index copies now use the full capture length; worth confirming no replay path can write garbage into rows beyond `num_actual_tokens` before o_proj consumes the padded window.

PR #15953, opened 2026-09-07
[BugFix] Apply expert gate in quantized shared expert paths

Sunshine-llh:
## What this PR does

  Apply the optional model-specific `expert_gate` after the quantized
  shared-expert fast paths for W8A8, W4A8, and W4A8MXFP.

  The non-quantized path applies this gate in `part2()`, but the quantized
  fast paths bypass `part2()`. As a result, models that use the optional
  shared-expert gate, such as Qwen3-Next, can produce incorrectly scaled
  shared-expert outputs and corrupted generation results.

  ## Root cause

  The expected shared-expert computation is:

  ```python
  gate_out, _ = self.layer.expert_gate(hidden_states)
  shared_out = F.sigmoid(gate_out) * shared_out
  ```

  The quantized fast paths execute the gate/up projection, activation, and
  down projection directly, bypassing `part2()` where this gate is normally
  applied.

  ## Changes

  - Preserve the original shared-expert input before the quantized path
    overwrites `hidden_states`.
  - Apply `expert_gate` after the W8A8, W4A8, and W4A8MXFP fast paths.
  - Keep other quantization types unchanged because they fall back to
    `part2()` and already apply the gate.
  - Add a W8A8 regression test that verifies:
    - `expert_gate` receives the original shared-expert input.
    - A zero gate logit scales the output by `sigmoid(0) = 0.5`.

   ## Validation

  Test environment:

  ```text
  vLLM: 0.26.0
  System fingerprint: vllm-0.26.0-1bb1d043
  Model: qwen36-gptq-w8a8
  Quantization: GPTQ W8A8
  ```

  The following request was used before and after applying the fix:

  ```bash
  curl -sS http://127.0.0.1:8327/v1/chat/completions \
    -d '{
      "model": "qwen36-gptq-w8a8",
      "messages": [
        {
          "role": "user",
          "content": "请用简洁的语言介绍一下人工智能。"
        }
      ],
      "temperature": 0,
      "max_tokens": 512,
      "stream": false
    }'
  ```

  Before applying the fix, the request generated corrupted mixed-language
  tokens and exhausted the configured output length:

  ```text
  Output:
  " i心里gm. oraf_ChCh,�曾ki,\\u it.� was is心里dibe ..."
  finish_reason: "length"
  completion_tokens: 512
  ```

  After applying the fix, the same request generated a normal answer and
  stopped naturally:

  ```text
  人工智能（AI）是计算机科学的一个分支，旨在开发能够模拟人类智能行为的系统。
  其核心能力包括学习（从数据中获取知识）、推理（解决问题）、感知（识别图像
  或语音）以及自然语言处理。

  简单来说，AI 让机器能够像人一样“看”、“听”、“说”和“思考”，广泛应用于
  医疗、交通、金融、教育等领域，旨在提高效率、辅助决策并创造新价值。
  ```

  ```text
  finish_reason: "stop"
  completion_tokens: 91
  ```

  An additional generation regression run passed 20/20 cases after the fix.
  ## Checks

  - `pre-commit`: passed
  - `cpu-ut`: passed
  - `ruff check`: passed
  - `py_compile`: passed
  - `git diff --check`: passed


vllm_ascend/ops/fused_moe/shared_experts.py
```diff
@@ -397,6 +400,22 @@ def maybe_wait_event(evt: torch.npu.Event | None):
...
+                and self.quant_type
+                in (
+                    QuantType.W8A8,
+                    QuantType.W4A8,
+                    QuantType.W4A8MXFP,
+                )
+                and hasattr(self.layer, "expert_gate")
+                and self.layer.expert_gate is not None
+            ):
+                gate_out, _ = self.layer.expert_gate(shared_expert_input)
```
nu11ptr7, line 416: The gate is applied only when `has_quantized_shared_without_lora` and quant_type is in (W8A8, W4A8, W4A8MXFP). If another quantized branch bypasses part2 (e.g. a W8A8MXFP or MXFP fast path), it would skip the gate just like the bug this fixes - worth confirming the type list covers every fast path that bypasses part2.

Sunshine-llh:
Could a maintainer please add the `ready-precise` label to run the selected tests? Pre-commit and CPU unit tests have passed. Thank you.

vllm_ascend/ops/fused_moe/shared_experts.py
Sunshine-llh: Confirmed. These are all quantized fast paths that currently bypass
  `part2()` in this method: W8A8/W4A8 share the first branch, and
  W4A8MXFP uses the second branch. Other quantization types fall back to
  the `else` path and apply `expert_gate` through `part2()`. There is
  currently no W8A8MXFP fast path here.

Sunshine-llh:
Gentle ping for the `ready-precise` label when convenient. The pre-commit and CPU unit tests have passed. Thank you. @nu11ptr7

Sunshine-llh:
The CPU unit-test process ended abruptly without a pytest failure summary or traceback. The job reports Executing the custom container implementation failed, so this appears to be a self-hosted runner/container issue. Could a maintainer please rerun the failed job?

PR #15964, opened 2026-09-07
CI: tolerate pending pods in multi-node nightly readiness loop

pgzddxx:
## What this fixes

The multi-node nightly pod-ready loop runs with set -euo pipefail. Immediately after applying a LeaderWorkerSet, worker or leader pods may not exist yet. In that normal pending window, the ready-field kubectl get pod exits non-zero and aborts the step instead of polling.

This was reproduced in workflow run 34126577054, job 101758309384: the step exited about one second after kubectl apply, before any pod/model work began.

## Change

Allow the two readiness lookups to return an empty value while pods are pending. Phase lookup already maps NotFound, and the existing loop then sleeps and retries until the 1200-second deadline.

## Validation

- YAML parse passed
- git diff --check passed
- Existing phase/timeout/error semantics are unchanged
- The change only affects the expected transient NotFound path


pgzddxx:
Closing this standalone workflow proposal. The associated changes were prepared only to diagnose an infrastructure race observed while running a temporary performance experiment; they are not intended for merge as part of the experiment.

PR #15969, opened 2026-09-07
[Feature][MRV2] Adapt PCP for data parallelism

wzx0726:

Add Ascend MRV2 handling for PCP combined with DP, reusing the upstream PCP dispatch sizing and input partitioning paths.

When a DP replica becomes idle, its dummy batch bypasses PCP partitioning. Saved PCP attention state may be absent or belong to the preceding real batch. Runtime dummy inputs use runner buffers, while decode graphs capture persistent PCP buffers.

- Build dummy PCP attention contexts from the current batch, block tables and rank-local slot mappings, including the DSA metadata path.
- Refresh persistent PCP buffers before dummy execution. Keep the implementation in `AscendPCPManager`; `NPUModelRunner.prepare_dummy_attn` delegates or takes the existing non-PCP path.
- Synchronize replicated speculative drafts independently of the target's PCP-local DP state, with every DP replica participating, including idle and decode replicas.
- Restrict PCP+DP graph configuration to eager or `FULL_DECODE_ONLY`.

Upstream validation dependency: https://github.com/vllm-project/vllm/pull/54523, merged as [7c2f1ff4958eaf0818405e9192c71608fe4a16b1](https://github.com/vllm-project/vllm/commit/7c2f1ff4958eaf0818405e9192c71608fe4a16b1), moves the PCP+DP restriction from common `ParallelConfig` validation to CUDA/ROCm platform checks. This PR relies on that change and contains no Ascend validator bypass or Pydantic schema rebuilding.

Related upstream PR: https://github.com/vllm-project/vllm/pull/53867 ([Feature][PCP] Support decode-only FULL CUDA graphs).

Once the paired vLLM includes #53867, adapt its PCP `prepare_inputs_to_capture` entry point to create `AscendInputBatch` directly in persistent PCP buffers. After capture and real-to-idle replay validation, remove `AscendPCPManager.prepare_dummy_attn` and the runner override, as tracked by `TODO(wzx0726)`. The Ascend dummy attention-context handling and replicated-draft DP synchronization remain necessary.

Adds Ascend-side MRV2 handling for PCP combined with DP. No new CLI options or environment variables are introduced.

Rebased onto Ascend main `ad86348b0cb2324d643df6a496f2d9c3879e4481`. The draft conflict resolution preserves vLLM 0.28.0 `num_tokens_across_dp` forwarding and applies fresh synchronization for replicated drafts on the main2main `dp_sync` interface.

**Compatibility and remaining validation:**

- The previously tested baseline was Ascend `b1c91857c16bde10c2fa7f5d7548b7666a86d4bb` with vLLM `e6bfe03ad73a3330cb427885aa90d97a12e1c704`.
- Removing the validator bypass requires vLLM #54523. Ascend main currently pins `b2f685834a6456197e7033966fdef52a23f1abcd`, which predates that change and still rejects PCP+DP. The local vLLM checkout remains at the user-selected `7c2f1ff4958eaf0818405e9192c71608fe4a16b1`. This PR does not change the main2main pin or reintroduce the bypass.
- Compatibility work against that newer revision is still pending, including the added `prepare_dummy_attn` argument and removal of `InputBatch.max_seq_len_np`. The current branch is not yet validated to run against that revision.
- Multi-card Attention/MLA/DSA inference, real-to-idle DP transitions during `FULL_DECODE_ONLY` replay, speculative-decoding numerical correctness, and EP/TP combinations require model-level validation on the final paired sources.

The latest rebase onto `ad86348b0` resolves an append-location conflict in `test_pcp_manager_v2.py`, retaining both the main DCP padding regression and this PR's tests. AST comparison confirms that all main tests and the final PR tests are preserved, and the other changed files match the automatic merge. Ruff lint/format, syntax parsing, and `git diff --check` passed. Runtime tests were not rerun for this conflict-only update.

For the preceding rebase onto `f8481287a`, **6 targeted candidate-method cases passed** in an isolated process in the Ascend container: five main2main draft-sync cases and one mocked v0.28.0 argument-routing case. The exact candidate `propose` method was loaded into the installed runtime classes; native dependencies and the rest of the runtime remained at their installed versions. This is method-level regression evidence, not full rebased-source or dual-version runtime validation.

All eight changed Python files passed Ruff lint/format checks, syntax parsing, and `git diff --check`. Full unit-suite and model validation on the final paired sources remain pending.

The runtime adaptation suite previously passed on the earlier baseline in an isolated Ascend-container test directory, using CPU tensors and mocks for device kernels:

```bash
python3 -m pytest -q tests/ut/worker/test_pcp_manager_v2.py tests/ut/worker/test_model_runner_v2.py tests/ut/worker/test_attn_utils_v2.py tests/ut/worker/test_mtp_pcp_speculator_v2.py
```

- **75 passed** (14 existing `torch.jit` deprecation warnings). This is prior-baseline evidence, not a test result for vLLM `7c2f1ff`.
- Covers missing/stale dummy PCP state, rank-local mappings, persistent buffer contents and storage, DSA metadata forwarding, non-PCP fallback, graph restrictions, and replicated draft DP synchronization.
- The removed validator bypass's dedicated tests are also removed; the existing unrelated platform test is retained.
- For the removal, targeted Ruff lint/format checks, Python syntax checks and `git diff --check` passed. No runtime tests were rerun against the upgraded vLLM. The local `format.sh ci` entry point remains unavailable because its shell lacks `pre-commit`.

### DIFF
--- a/tests/ut/worker/test_attn_utils_v2.py
+++ b/tests/ut/worker/test_attn_utils_v2.py
@@ -647,8 +647,7 @@ def test_dsv4_backends_declare_role_specific_logical_sizes(
         ("model_state", CUDAGraphMode.NONE, False, 1, 5),
         ("model_state", CUDAGraphMode.FULL, False, 1, 8),
         ("pcp_capture", CUDAGraphMode.NONE, True, 2, 8),
-        ("pcp_dummy", CUDAGraphMode.NONE, False, 2, 8),
-        ("pcp_dummy", CUDAGraphMode.FULL, True, 2, 8),
+        ("pcp_runtime", CUDAGraphMode.NONE, False, 2, 8),
     ],
 )
 def test_mrv2_builds_shared_dsa_metadata_for_each_execution_mode(
--- a/tests/ut/worker/test_attn_utils_v2.py
+++ b/tests/ut/worker/test_attn_utils_v2.py
@@ -647,6 +647,8 @@ def test_dsv4_backends_declare_role_specific_logical_sizes(
         ("model_state", CUDAGraphMode.NONE, False, 1, 5),
         ("model_state", CUDAGraphMode.FULL, False, 1, 8),
         ("pcp_capture", CUDAGraphMode.NONE, True, 2, 8),
+        ("pcp_dummy", CUDAGraphMode.NONE, False, 2, 8),
+        ("pcp_dummy", CUDAGraphMode.FULL, True, 2, 8),
     ],
 )
 def test_mrv2_builds_shared_dsa_metadata_for_each_execution_mode(
@@ -666,7 +668,7 @@ def test_mrv2_builds_shared_dsa_metadata_for_each_execution_mode(
         [2, 1, 0, 0],
         dtype=torch.int32,
     )
-    pcp_context = object() if caller == "pcp_capture" else None
+    pcp_context = object() if pcp_size > 1 else None
     pcp_manager = (
         SimpleNamespace(
             build_attention_context=MagicMock(return_value=pcp_context),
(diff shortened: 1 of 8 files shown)


zhenwenqi2024 (approved):

Ronald1995 (approved):
LGTM

weijinqian0 (approved):

PR #15980, opened 2026-09-07
fix: retain async speculative metadata copies

taojiovo:

Cherry-picks the fix from #15719 onto releases/v0.26.0rc.

No.

- Added/ported the unit test from #15719.
- Verified the changed Python files compile successfully.

### DIFF
--- a/vllm_ascend/worker/model_runner_v1.py
+++ b/vllm_ascend/worker/model_runner_v1.py
@@ -595,6 +595,13 @@ class NPUModelRunner(GPUModelRunner):
         self.mamba_state_idx: dict[str, int] = {}
         self._mamba_bufs: Any | None = None
         self._mamba_copy_bufs: Any | None = None
+        # Keep the pinned CPU sources for speculative-decode metadata alive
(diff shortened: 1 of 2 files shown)


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -1370,6 +1377,23 @@ def _sanitize_placeholder_input_ids_for_forward(
...
+        )
+        if self.device.type == "cpu":
+            return device_metadata
+
+        pending_copies = self._pending_spec_decode_metadata_copies
+        while pending_copies and pending_copies[0][1].query():
+            pending_copies.popleft()
+        copy_done = torch.npu.Event()
+        copy_done.record(torch.npu.current_stream())
+        pending_copies.append((cpu_metadata, copy_done))
```
nu11ptr7, line 1394: The FIFO cleanup only inspects the head and only runs when the method is called again, and entries are never drained on teardown. Copies complete in stream order so this is normally bounded, but a stuck/erroring stream would grow the deque across steps - worth popping all completed entries each call and/or capping the length.

PR #15987, opened 2026-09-08
[CI] Add --profile switch to /nightly command for profiling artifact export

zhangxinyuehfad:

Introduce a --profile flag to the /nightly slash command that force-enables the torch profiler and uploads profiling output as artifacts for performance debugging.

Key changes:
- Add --profile argument parsing in pr_nightly_command.yml, propagated as profile_enabled through all SOC nightly workflows (A2, A3, A3-560T, A5, 310P) and their reusable workflows
- Inject --profiler-config into server commands via VLLM_ASCEND_FORCE_PROFILE env var in test_single_node.py, multi_node_config.py, and runtime.py
- Package profiling output as tar.gz and upload to both OBS and GitHub Artifacts with 14-day retention
- Support multi-node profiling via PVC fetch (same pattern as benchmark results)
- Pass VLLM_ASCEND_FORCE_PROFILE and VLLM_ASCEND_PROFILE_DIR env vars to pods via lws.yaml.jinja2

### DIFF
--- a/.github/workflows/_e2e_nightly_multi_node.yaml
+++ b/.github/workflows/_e2e_nightly_multi_node.yaml
@@ -83,6 +83,11 @@ on:
         type: boolean
         default: false
         description: enable Pod AOP pipeline on failure
+      profile_enabled:
+        required: false
+        type: boolean
+        default: false
+        description: force-enable the torch profiler and export profiling artifacts
       bisect_good_commit:
         required: false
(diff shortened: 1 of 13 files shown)


czydyy:
/nightly kimi-k2.5 Qwen3.5-27B-w8a8-A2 --profile_enabled
command triggered.

PR #15995, opened 2026-09-08
[Feature][Kernel] Add opt-in Ascend950 AddRmsNormBias fusion

Wyz-134:

Add an opt-in Ascend950 implementation of AddRmsNormBias. The existing residual RMSNorm path issues a separate bias addition when norm bias is present; the new AscendC row kernel performs residual addition, RMS normalization, weight multiplication, and optional bias addition in one launch.

- Build and register the operator for Ascend950, with dedicated FP16, BF16, and FP32 tiling keys.
- Support aligned hidden widths up to 6144 that fit the UB allocation. Preserve the residual output and FP32 reciprocal RMS output.
- Round the residual sum to the input dtype before the FP32 reduction. Perform normalization, weight multiplication, and bias addition in FP32 before the final cast; this is not bitwise equivalent to the separate-op path.
- Avoid private shape-helper symbols when loading the standalone custom operator package.
- Add guarded RMSNorm/GemmaRMSNorm dispatch without enabling all A5 custom operators globally.
- Add dispatch unit tests and an A5-specific numerical reference to the existing operator system tests.

The implementation follows the row scheduling and fused affine calculation in [sgl-kernel-npu AddRMSNormBias](https://github.com/sgl-project/sgl-kernel-npu/blob/69d392bd11f280babcef113fcef9cc4756a95fc6/python/sgl_kernel_npu/sgl_kernel_npu/norm/add_rmsnorm_bias.py). Its optional quantized-output branch is not included here.

`VLLM_ASCEND_ENABLE_ADD_RMS_NORM_BIAS=1` opts into the A5 path. It defaults to `0`; unsupported widths, non-A5 devices, and batch-invariant execution retain the existing dispatch. An explicitly enabled but missing extension reports a load error.

This PR does not change MC2 selection or include machine-specific examples, benchmark scratch scripts, Claude settings, or fleet-monitor files.

The kernel and host implementation were built and installed for Ascend950 on the original development baseline. Eighteen dtype/row-count cases and twenty-four dtype/width/optional-bias cases passed (42 total), checking the normalized result, exact rounded residual output, and FP32 reciprocal RMS. The main correctness matrix also compared against `torch_npu.npu_add_rms_norm` followed by bias addition within dtype-specific tolerances.

BF16, hidden width 6144, NPU-event timings around graph replay (median microseconds):

| Rows | Separate operators | Fused operator |
| ---: | ---: | ---: |
| 1 | 3.701 | 3.407 |
| 32 | 8.039 | 6.636 |
| 128 | 9.531 | 8.632 |
| 1024 | 25.782 | 35.067 |
| 4096 | 114.544 | 147.439 |

**Draft:** large-row performance regresses, so the feature remains disabled by default. End-to-end accuracy comparison is still running. The source changes have been ported to current upstream main; full integration/NPU testing of that rebased tree remains pending. The container-specific `maybe_chunk_residual` call was not carried into upstream main.

Ruff and Python syntax checks cover the changed Python files. The repository-wide `bash format.sh ci` was attempted but stalled while installing its markdownlint environment; full lint and the newly integrated pytest cases remain pending.

Reproduction on an Ascend950 development environment:

```bash
cd csrc
bash build.sh --pkg --ops=add_rms_norm_bias --soc=ascend950
cd ..
# Install the generated custom operator package for this checkout first.
pytest tests/ut/ops/test_layernorm.py -k a5_add_rms_norm_bias
pytest tests/e2e/nightly/single_node/ops/singlecard_ops/test_add_rms_norm_bias.py -k a5_add_rms_norm_bias
```

### DIFF
--- a/csrc/moe/add_rms_norm_bias/op_host/add_rms_norm_bias_infershape.cpp
+++ b/csrc/moe/add_rms_norm_bias/op_host/add_rms_norm_bias_infershape.cpp
@@ -13,7 +13,6 @@
  * \brief
  */
 #include "tiling_base/error_log.h"
-#include "util/shape_util.h"
 #include "register/op_impl_registry.h"
 
 static constexpr int IDX_0 = 0;
@@ -21,7 +20,6 @@ static constexpr int IDX_1 = 1;
 static constexpr int IDX_2 = 2;
 
 using namespace ge;
-using namespace Ops::Base;
 
 namespace ops {
 
@@ -47,8 +45,12 @@ static ge::graphStatus InferShape4AddRmsNormBias(gert::InferShapeContext* contex
     size_t xDimNum = x1Shape->GetDimNum();
     size_t gammaDimNum = gammaShape->GetDimNum();
 
-    if (IsUnknownRank(*x1Shape) || IsUnknownRank(*gammaShape)) {
-        SetUnknownRank(*rstdShape);
+    // Keep standalone custom packages independent of private Ops::Base symbols.
+    const bool xUnknown = xDimNum == 1 && x1Shape->GetDim(0) == -2;
+    const bool gammaUnknown = gammaDimNum == 1 && gammaShape->GetDim(0) == -2;
+    if (xUnknown || gammaUnknown) {
+        rstdShape->SetDimNum(1);
+        rstdShape->SetDim(0, -2);
         OP_LOGD(context, "End to do InferShape4AddRmsNormBias with unknown rank.");
         return GRAPH_SUCCESS;
     }
--- a/csrc/moe/add_rms_norm_bias/op_kernel/add_rms_norm_bias_a5.h
+++ b/csrc/moe/add_rms_norm_bias/op_kernel/add_rms_norm_bias_a5.h
@@ -1,0 +1,97 @@
+// Copyright (c) 2026 Huawei Technologies Co., Ltd.
+// Licensed under CANN Open Software License Agreement Version 2.0.
+// Row-strided fusion follows sgl-kernel-npu/norm/add_rmsnorm_bias.py:
+// round the residual addition, reduce in FP32, then fuse affine and bias.
+#ifndef ADD_RMS_NORM_BIAS_A5_H_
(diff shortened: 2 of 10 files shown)


csrc/moe/add_rms_norm_bias/op_host/add_rms_norm_bias_infershape.cpp
```diff
@@ -13,15 +13,13 @@
  * \brief
  */
 #include "tiling_base/error_log.h"
-#include "util/shape_util.h"
```
chenlongxiao, line 16: 原来A3需要这个文件不？删掉去对A3有影响吗

Wyz-134:
/nightly GLM-5.1-w8a8c8-128k-1k-90-50-PD --aop_enabled --good-commit 182b75ea84ff13ece27b57391331ef8935410aea --bad-commit 219a7e7efd0b5363be0d5cc198995e9417b2de13

csrc/moe/add_rms_norm_bias/op_kernel/add_rms_norm_bias_a5.h
```diff
@@ -0,0 +1,97 @@
...
+        if(hasBias) {
+            DataCopy(a,bias,cols); Sync<HardEvent::MTE2_V>();
+            ToFloat(b,a); Sync<HardEvent::V_MTE2>();
+        }
+        for(uint32_t row=GetBlockIdx(); row<rows; row+=GetBlockNum()) {
+            uint64_t offset=static_cast<uint64_t>(row)*cols;
+            DataCopy(a,in1[offset],cols); DataCopy(r,in2[offset],cols);
+            Sync<HardEvent::MTE2_V>();
+            ToFloat(sum,a); ToFloat(other,r);
+            Add(sum,sum,other,cols); PipeBarrier<PIPE_V>();
```
lijiahang226, line 64: Could we clarify whether this rounding order is part of the intended AddRmsNormBias contract on A5?
In particular, I think we should have model-level accuracy evidence for BF16.

PR #15996, opened 2026-09-08
[BugFix][Ops] Zero-init KDA chunked-prefill pipeline buffers

Liears:

The KDA chunked-prefill path allocates intermediate buffers with `torch.empty*`/`new_empty`:
`chunk_delta_h`'s `h` / `final_state` / `v_new`, `chunk_local_cumsum`'s output (x2), and `recompute_w_u`'s `w` / `u` / `kg`.

The triton kernels write **tile-shaped** regions and leave tail elements untouched whenever the sequence length is not tile-aligned. With a pooled caching allocator those unwritten bytes hold whatever a previous launch left there, so:

- **two identical calls can return different recurrent states** (measured on 910B3: 99.3% of final-state elements differing, max |delta| 0.207),
- the final recurrent state — the hand-off that decode builds on — depends on unrelated request history, silently corrupting long-context generation quality.

This PR zero-initializes all eight buffers. The cost is one memset per call, negligible next to the kernel runtimes.

Yes — long-context generations on KDA-based hybrid models become deterministic and no longer depend on request history (a silent accuracy fix).

- New UT: `tests/ut/ops/test_kda_chunk_buffer_init.py`
  - `test_chunk_kda_is_deterministic` (771/968/1303-token prefills, 3 identical calls → bitwise-identical output + final state)
  - `test_chunk_kda_survives_allocator_churn` (pool-sized garbage allocated/freed between calls → results unchanged)
  - All 4 tests pass on Atlas 910B3.
- End-to-end on a 48-layer hybrid GLM/KDA model: chunked prefill becomes bitwise deterministic; a 12-problem GSM8K-level gate scores 12/12 after the fix (nondeterministic outputs before it).

### DIFF
--- a/tests/ut/ops/test_kda_chunk_buffer_init.py
+++ b/tests/ut/ops/test_kda_chunk_buffer_init.py
@@ -1,0 +1,110 @@
+# SPDX-License-Identifier: Apache-2.0
+# FileCopyrightText: Copyright contributors to the vLLM project
+"""Regression tests for the KDA chunked-prefill pipeline's buffer init.
+
+The chunked-prefill kernels (``chunk_gated_delta_rule_fwd_h_kda``,
+``chunk_local_cumsum``, ``recompute_w_u_fwd``) write tile-shaped regions and
+leave tail elements of their output buffers untouched whenever the sequence
+length is not tile-aligned. With ``torch.empty*`` those bytes hold whatever
+the caching allocator last served there, so two identical calls can produce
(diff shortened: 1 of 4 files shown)


Liears:
Addressed (33d716867): the test module now uses `pytest.importorskip("torch_npu")` instead of a bare import, so collection skips gracefully on hosts without torch_npu (CPU-only environments, lint, docs builds) instead of erroring. Applied the same pattern to the other NPU-bound test modules in the companion PR (#16005).

PR #16005, opened 2026-09-08
[Feat][Model] GLM-5 kpool sparse attention: indexer, caches, backend, graph-safe top-k

Liears:

GLM-5.3-Flash devotes **11 of its 45 layers** to NoPE sparse-MLA attention: each query selects top-k 4-token pools via a trained indexer, expands them to token indices, appends the un-pooled causal tail, and attends only that set. On Ascend the vendored CUDA path (fwht + fp8) cannot compile (triton-ascend rejects fp8 stores), so these layers **silently ran dense attention behind a no-op Indexer wrapper**.

This adds a BF16 port of the full pipeline (traced from the vendor's production branch):

| Component | File |
|---|---|
| Orchestrator (state scatter → pool assembly → triton compress → top-k → expansion → causal tail → buffer) | `ops/glm5_kpool_indexer_ascend.py` |
| Top-k scorers: chunked eager (prefill) + static-shape sort-based (graph capture; aclnnTopk caps k at 256) | `ops/indexer_kpool_topk.py` |
| BF16 pool-granular indexer K cache + token-granular compressor-state cache with manager specs | `ops/glm5_kpool_caches.py` |
| MLA backend: 3-role metadata (kv / indexer_state / indexer), pytorch golden-path sparse attention, persistent block-table buffer sized for SFA kernel-block expansion | `attention/indexer_kpool_mla_v1.py` |
| Model-level wiring behind `GLM53_KPOOL_INDEXER=1` (default off) | `patch/worker/patch_glm5_kpool_indexer.py` |

Two fidelity details from the trained model's contract:
- **causal tail**: the ≤ kpool-1 most recent tokens (incl. the current decode token) are forced into the attended set — without it 3 of 4 decode steps lose the most recent tokens and multi-digit arithmetic degrades (GSM8K-level gate: **0/4 → 12/12**)
- **pool budget**: `topk // pool_size` pools (vendored `history_group_budget_for_topk`), so expanding yields exactly `topk` tokens

Also fixes an off-by-one in the persistent indexer block-table buffer: upstream expands at `cdiv(max_len, scheduler_bs) * split` columns, which exceeds `cdiv(max_len, kernel_bs)` by one at any non-scheduler-block-aligned `max_model_len` (e.g. 200000 → 1564 vs 1563).

Stacked on the GLM bring-up chain (#15592 and earlier PRs); will show those commits until the base merges.

Yes — GLM-5.3 hybrid models get their trained sparse attention on NPU (opt-in via `GLM53_KPOOL_INDEXER=1`), fixing a silent dense-attention fallback that degraded multi-step arithmetic.

- UTs (all pass on Atlas 910B3): `tests/ut/ops/test_glm5_kpool_tail.py` (causal-tail oracle 10/10), `tests/ut/ops/test_indexer_kpool_topk_static.py` (static-vs-eager top-k 4/4)
- End-to-end on a 48-layer hybrid with official W8A8 weights: GSM8K-level 12-problem gate **12/12**; FULL_DECODE_ONLY graph capture 4/4 sizes; single-stream TPOT 365 ms (eager) → 40 ms (graph).


Liears:
Addressed all three findings (e3a7f6df0..c33122d54):

1. **kpool compress fallback invalid slots (critical)** — confirmed real: a `loc` of `-1` produced `block_id=-1` / `offset=cache_block_size-1`, i.e. negative indexing into the last cache block. The production NPU path was unaffected (the triton kernel masks `flat_slot`), but the fallback violated the same contract. Fixed by masking `(loc >= 0) & (loc < num_slots)` and indexing with the filtered rows, mirroring the kernel. Added `tests/ut/ops/test_glm5_next_kpool_compress.py` covering the slot contract (valid rows write, invalid rows write nothing, empty batch no-op) — CPU tensors keep those tests on the fallback path deterministically.

2. **stale `_tensor_copy_pairs`** — confirmed: a partially collected list survived into the next step's collection and corrupted the `len == n` completeness check. Now cleared whenever the tensor-pair path is not taken.

3. **`_pinned_keepalive` leak** — confirmed: appends are unconditional in `_pinned_h2d` but the clear only runs under the async-scheduling event protocol, so sync scheduling + spec decode grew the list by five pinned buffers per step. Fixed in `_pinned_h2d` itself: without the event protocol it now issues a blocking copy (completes before return, no keepalive needed) instead of the leaking async form.

Also switched the NPU-bound test modules to `pytest.importorskip("torch_npu")` so collection skips gracefully on hosts without torch_npu.

PR #16006, opened 2026-09-08
[v0.27.1rc][Performance][Kimi K3] Avoid attention-residual token-count specialization

Dawn952:

Kimi K3's attention-residual kernel currently declares `N = prefix_sum.shape[0]` as `tl.constexpr`, so changing the token count creates another compiled variant even when the model dimensions and launch options are unchanged.

Rename `N` to `num_tokens`, make it a runtime scalar, and add `@triton.jit(do_not_specialize=["num_tokens"])`. The runtime parameter and `do_not_specialize` are both needed: declaring `num_tokens` as `tl.constexpr` still specializes on its value. `num_tokens` only controls the per-core token range; it does not define a tile shape. `B` is the number of valid residual blocks, not batch size. Keep `H`, `B`, `BLOCK_CAPACITY`, `EPS`, `NUM_CORES`, and `NB` compile-time parameters.

The NPU model path in `vllm_ascend/models/kimi_k3.py` calls this operator through `apply_attn_res`. Target branch: `releases/v0.27.1rc`.

No API or mathematical changes. Avoid token-count-driven compilation variants for this kernel. Changes to model dimensions, dtype, or compilation options can still require compilation.

- Added a regression test that executes 1, 7, 16, 17, and 512 tokens with fixed model dimensions, compares each output against the existing PyTorch reference, and asserts all launches return the same compiled kernel object. This covers the value-one case, integer alignment changes, partial cores, and multiple rows per core.
- Passed `ruff check`, `ruff format --check`, Python AST parsing for both changed files, and `git diff --check`.
- NPU execution is pending: the registered test container was stopped on both checked test hosts. No containers or services were started or stopped. Keeping this PR as a draft until hardware validation is available.
- The latest local `bash format.sh ci` attempt was stopped during actionlint environment download; the full local lint run did not complete.

Triton reference: https://triton-lang.org/main/python-api/generated/triton.jit.html

### DIFF
--- a/tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_kimi_k3_fusions.py
+++ b/tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_kimi_k3_fusions.py
@@ -22,7 +22,7 @@ import torch_npu  # noqa: F401
 from vllm.triton_utils import HAS_TRITON
 
 if HAS_TRITON:
-    from vllm_ascend.ops.triton.kimi_k3.attention_residual import apply_attn_res
+    from vllm_ascend.ops.triton.kimi_k3.attention_residual import _apply_attn_res_kernel, apply_attn_res
 
 
 pytestmark = [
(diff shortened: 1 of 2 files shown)


AuroraEmiya (approved):
LGTM

linfeng-yuan (approved):

PR #16058, opened 2026-09-08
[BugFix][Misc] Skip NZ conversion for weights with k=1 or n=1

ckeyever:

`aclnnMatmulWeightNZ` does not support matrices whose reduction/output dimension is one (k=1 or n=1). Previously such weights were still converted to FRACTAL_NZ, so the subsequent linear/matmul dispatch selected the NZ-only path and failed at runtime.

This PR adds a guard in `_should_trans_nz` so that weights with a 1 in either of the lowest two dimensions are kept in ND format. The guard applies unconditionally, including on 310P (`WeightLayoutPolicy.FORCE_NZ`) hardware, matching the intent of the original change.

No. This is an internal weight-layout dispatch fix; no API, config, or behavior contract changes for users.

- Unit tests added in `tests/ut/test_utils.py::TestUtils::test_maybe_trans_nz`:
  - non-310P (A2, CONFIGURABLE) with NZ mode enabled: shapes `(32, 1)`, `(1, 64)`, `(2, 32, 1)` are not cast to NZ
  - 310P (FORCE_NZ): the same shapes are not cast to NZ (validates guard precedence over the always-convert policy)
- `ruff check` and `ruff format --check` pass on the changed files


Ronald1995 (approved):
LGTM

vllm_ascend/utils.py
```diff
@@ -272,6 +278,14 @@ def _should_trans_nz(weight: torch.Tensor) -> bool:
     if weight.is_meta:
         return False
 
+    # aclnnMatmulWeightNZ does not support matrices whose reduction/output
+    # dimension is one (k=1 or n=1). Keep these weights in ND format so the
+    # subsequent linear/matmul dispatch does not select the NZ-only path.
```
lcfenglinwan, line 283: What kind of model would have such a weight shape?

ckeyever: Qwen3.6-35B-A3B

weijinqian0 (dismissed):

weijinqian0 (approved):

PR #16075, opened 2026-09-08
[BugFix][Spec Decode] Recompute draft DP counts during profiling

jiajinzhu2:

Fix DP token-count reuse during the initial memory profiling pass of the V2 DSpark and DFlash speculators. A target-model DP synchronization describes the token counts of that particular forward pass. Reusing it for a draft forward with a different input length causes engine initialization to fail.

Upstream vLLM already fixes the profiling argument in [#54856](https://github.com/vllm-project/vllm/pull/54856) (`facd9a74a1`). However, Ascend main currently pins `.github/vllm-main-verified.commit` to `b2f685834a6456197e7033966fdef52a23f1abcd`, which predates that fix and still forwards the target counts. This adaptation covers the current verified baseline and older affected vLLM revisions. With the upstream fix present, the parent already ignores the profiling DP state.

**TODO (main2main cleanup):** Once vLLM-Ascend's main2main integration and verified vLLM baseline include upstream [vllm-project/vllm#54856](https://github.com/vllm-project/vllm/pull/54856) (`facd9a74a1cd1b9fed324cdc2cceb8d54fdad3d0`), remove the temporary profiling guards from both `AscendDSparkSpeculator.propose()` and `AscendDFlashSpeculator.propose()` and rely on the upstream profiling path.

The call chain is:

1. `profile_run()` calls `_dummy_run(..., skip_attn=True, is_profile=True)`.
2. The target forward coordinates its per-rank token counts and produces `dp_sync`.
3. The runner passes this state to the draft `propose()` with `dummy_run=True` and `skip_attn_for_dummy_run=True`.
4. The parallel drafter computes its own query length. DFlash uses `num_reqs * (K + 1)`; DSpark with `sample_from_anchor=True` uses `num_reqs * K`.
5. The upstream profiling branch calls `_generate_draft()` with that draft length but forwards `dp_sync.num_tokens_across_dp` from the target.
6. `set_forward_context()` passes these inconsistent values to `DPMetadata.make()`, whose local token-count assertion fails.

This was observed on A5 with DP4 and DSpark: 64 requests and K=5 produced 320 draft tokens while the reused target count was 384, resulting in `AssertionError: 384 320`. DFlash can hit the same issue even though its query length is K+1: a 4096-token profiling batch with 64 requests and K=5 has only 384 draft query tokens. Matching lengths in a uniform decode batch do not guarantee matching lengths during memory profiling.

Both Ascend wrappers now pass `None` as the synchronization state only when `dummy_run and skip_attn_for_dummy_run`. This lets `set_forward_context()` coordinate the actual draft token counts across DP ranks. It does not disable DP synchronization or mutate the target state. Normal inference and graph warmup keep their existing state forwarding.

Eagle3 and MTP do not need this adjustment: their shared autoregressive path preserves the target batch shape for the first draft step and performs a fresh DP synchronization for the later one-token decode steps. This change does not address graph-padding index handling.

V2 DSpark/DFlash DP deployments can complete memory profiling when the target and draft input lengths differ. No CLI or configuration changes are required.

- Reproduced the DSpark `384 320` assertion on A5, then verified that the profiling guard allows startup to proceed. The local DeepSeek-V4-Flash-DSpark V2 DP4/EP deployment subsequently completed a real chat request and all 1319 GSM8K requests with zero failed requests. That runtime checkout also contained separate FP8 operator compatibility fixes; those are outside this PR.
- Executed the upstream DFlash and autoregressive `propose()` functions in an isolated CPU check with model kernels and collectives mocked. With 64 requests and K=5, a 4096-token target forwarded stale counts to DFlash's 384-token draft; Eagle3/MTP reused counts only for their shape-preserving first step and synchronized the later step independently. This is not a DFlash end-to-end NPU test.
- No test files are included in this PR.

Local runtime validation used vLLM `e6bfe03ad73a3330cb427885aa90d97a12e1c704`.

Local guard/existing DSpark checks: **26 passed** on Ascend `3d27bd866` with its verified vLLM revision `b2f685834a6456197e7033966fdef52a23f1abcd`. These checks cover both wrappers, target counts 384/4096, with/without DP state, and profiling versus ordinary/graph-warmup calls. They were run locally and are not part of the PR. The isolated test environment used the image's NPU Triton import compatibility shim.

The two changed files pass Ruff lint and formatting checks. Full repository hook validation remains incomplete locally because the actionlint environment is still downloading its Go toolchain; this is separate from the Python checks above.

### DIFF
--- a/.github/workflows/scripts/test_config.yaml
+++ b/.github/workflows/scripts/test_config.yaml
@@ -196,6 +196,7 @@ estimated_times:
   tests/e2e/pull_request/four_card/context_parallel/test_deepseek_v4.py: 520
   tests/e2e/pull_request/four_card/context_parallel/test_prefix_caching_cp.py: 210
   tests/e2e/pull_request/four_card/model_runner_v2/test_deepseek_v4.py: 520
+  tests/e2e/pull_request/four_card/model_runner_v2/test_spec_decode_data_parallel.py: 600
   tests/e2e/pull_request/four_card/rlhf/consistency/test_batch_invariant_tp4.py: 210
   tests/e2e/pull_request/four_card/spec_decode/test_mtp_qwen3_next.py: 730
   tests/e2e/pull_request/four_card/spec_decode/test_dspark_deepseekv4.py: 1280
--- a/tests/e2e/pull_request/four_card/model_runner_v2/test_spec_decode_data_parallel.py
+++ b/tests/e2e/pull_request/four_card/model_runner_v2/test_spec_decode_data_parallel.py
@@ -59,8 +59,8 @@ PROMPTS = [
 def test_deepseek_v4_dspark_spec_decoding_dp4_full_graph() -> None:
     with DPVllmRunner(
         DEEPSEEK_V4_DSPARK_MODEL,
-        data_parallel_size=4,
-        tensor_parallel_size=1,
+        data_parallel_size=2,
+        tensor_parallel_size=2,
         max_model_len=4096,
         max_num_seqs=2,
         max_num_batched_tokens=512,
--- a/tests/e2e/pull_request/four_card/model_runner_v2/test_spec_decode_data_parallel.py
+++ b/tests/e2e/pull_request/four_card/model_runner_v2/test_spec_decode_data_parallel.py
@@ -1,0 +1,93 @@
+# Copyright (c) 2026 Huawei Technologies Co., Ltd. All Rights Reserved.
+# Copyright 2026 The vLLM team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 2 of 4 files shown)


Ronald1995 (approved):
LGTM. it's fine to fix the bug by follow upstream, but you need to resolve conflicts and it's better to add a test case to cover it.

jiajinzhu2:
> LGTM. it's fine to fix the bug by follow upstream, but you need to resolve conflicts and it's better to add a test case to cover it.

Done

drslark (approved):
LGTM.

jiajinzhu2:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

jiajinzhu2:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

jiajinzhu2:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

jiajinzhu2:
/rerun
[Bot]: rerun completed.

**Failed:**

jiajinzhu2:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

jiajinzhu2:
/cancel
[Bot]: successfully force-cancelled the following runs:


https://github.com/vllm-project/vllm-ascend/actions/runs/34470285936

jiajinzhu2:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

jiajinzhu2:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

weiguihua2 (approved):

PR #16104, opened 2026-09-09
[Test][DSA] Cover the indexer quant_mode selection on the device adaptor

yiminghub2024:

Follow-up to #16224, which fixed the DSA lightning indexer `quant_mode` on Ascend 950 but landed without
unit coverage. This PR adds the tests, and nothing else — after rebasing on main, `vllm_ascend/` here is
identical to main.

`QuantLightningIndexerV2` derives the accepted query/key, dequant scale and weights dtypes from its
`quant_mode` attribute. Per the operator docs in this repo
(`csrc/attention/quant_lightning_indexer_v2/README.md`), on Ascend 950PR/950DT `query`/`key` only support
`FLOAT8_e4m3fn` under `quant_mode` 1/3 and `INT8` under `quant_mode` 2, while the dequant scales and
weights are `FLOAT32` under mode 1 versus `FLOAT16` under mode 2. A2/A3 support `quant_mode` 2 only.

So the mode is not a free-standing constant: it is a description of the dtypes the device adaptor
produces, and the op rejects the call when the two disagree. That is exactly how DeepSeek-V4 used to abort
during warmup on Ascend 950 (`Invalid_Argument_Tensor_Dtype(EZ0021)`, `The dtype of q and k must match
quant_mode`) — the adaptor emitted FP8 while the mode said int8. `get_dsa_indexer_quant_mode()` and the
`prepare_dsa_indexer_*` helpers therefore have to move together, and these tests are what keeps them from
drifting apart again.

No. Tests only; no change under `vllm_ascend/`.

Two unit tests:

- `tests/ut/device/test_device_op.py`: parametrized over `BaseDeviceAdaptor`, `Ascend310PDeviceAdaptor` and
  `A5DeviceAdaptor`, pinning each adaptor's `get_dsa_indexer_quant_mode()` to the dtypes its own
  `prepare_dsa_indexer_weights` / `prepare_dsa_indexer_key_scale` helpers produce — int8 mode with
  `float16` scales on non-A5, fp8 mode with `float32` scales on A5.
- `tests/ut/models/test_deepseek_v4_indexer.py`: asserts `AscendIndexerOps.select_topk` forwards the
  adaptor's mode to the op instead of a hardcoded literal (checked against `A5DeviceAdaptor`, the case
  #16224 fixed), and the existing `select_topk` test now compares against the adaptor's value rather than
  a bare `2`.

```bash
pytest -sv tests/ut/device/test_device_op.py tests/ut/models/test_deepseek_v4_indexer.py
```


yiminghub2024:
https://github.com/vllm-project/recipes/pull/942

yiminghub2024:
@wangxiyuan please take a look ,thanks

tests/ut/models/test_deepseek_v4_indexer.py
```diff
@@ -392,6 +405,7 @@ def test_quantize_scatter_then_select_topk(self):
         )
 
         with (
+            _on(AscendDeviceType.A2),
```
wangxiyuan: we won't support the code style like `if a2, if a5...`, please move this kind of hardware check into `device` module

yiminghub2024: Thanks, fixed. I removed the hardware check entirely instead of moving the branch around: `quant_mode` is
now a `dsa_indexer_quant_mode` attribute on the device adaptors in `vllm_ascend/device/device_op.py`, in
the same `Lightning Indexer Dtype Prep` section as `indexer_quantize_query` and the
`prepare_dsa_indexer_*` helpers that produce the dtypes the op validates it against
(`BaseDeviceAdaptor` = 2/int8, `A5DeviceAdaptor` = 1/fp8-e4m3).

Call sites just read `DeviceOperator.dsa_indexer_quant_mode`, so there is no `if a2 / if a5` anywhere, the
helper in `dsa_attn_kv_plan.py` is gone, and the tests no longer patch the hardware profile - they assert
per-adaptor values in `tests/ut/device/test_device_op.py` and that `select_topk` forwards the adaptor's
value.

PR #16110, opened 2026-09-09
[Bugfix] Fix Gemma4 MoE W4A4 MXFP4 loading with TP1, TP2, and TP4

yzlbig:
# [Bugfix] Fix MXFP4 MoE scale layout and dtype handling

This PR fixes several general compatibility issues in the Ascend W4A4 MXFP4 path.

#### Explicit MXFP4 dtype interpretation for grouped matmul

MXFP4 activations and weights are physically stored as `torch.uint8`. Without explicit logical dtype arguments, `npu_grouped_matmul` may interpret them as an unsupported `DT_UINT8`/`DT_UINT8` pair.

This PR explicitly passes:

~~~python
x_dtype=torch_npu.float4_e2m1fn_x2
weight_dtype=torch_npu.float4_e2m1fn_x2
~~~

The tensors are not converted or copied. These arguments tell the NPU operator to interpret the packed `uint8` data as MXFP4.

#### Support odd MXFP4 scale group counts

The existing scale conversion assumes that the final scale dimension is even before reshaping it into pairs:

~~~text
[G, N, K] -> [G, N, K // 2, 2]
~~~

Some valid tensor shapes produce an odd number of scale groups. Reshaping these scales directly causes a tensor shape mismatch.

This PR pads the final scale dimension with one zero when necessary before applying the existing paired layout conversion.

Existing checkpoints with an even number of scale groups continue to use the original path.

#### Support uneven row-parallel MX scale shards

Row-parallel MX scale parameters use a ceil-divided local group count. When the global number of scale groups is not divisible by the tensor-parallel size, the final shard can contain fewer valid groups than the allocated local parameter.

This PR adds an Ascend MX scale loader wrapper that pads the global scale tensor before delegating to the existing row-parallel weight loader.

Padding is applied only when:

~~~python
0 < padding < tp_size
~~~

This limits the behavior to the natural remainder introduced by tensor-parallel ceil division. Unexpected checkpoint shape mismatches are not hidden and continue to fail through the existing loader.

The compatibility logic applies only to:

- Ascend row-parallel linear layers
- MX quantization methods
- Per-group scale parameters

The generic vLLM weight loader remains unchanged.

Yes.

W4A4 MXFP4 models can now load and execute correctly when they use:

- Packed MXFP4 tensors requiring explicit logical dtype metadata
- An odd number of scale groups
- Scale dimensions that are not evenly divisible across tensor-parallel ranks

There are no API, command-line interface, or configuration changes.

Unit tests were added for:

- Odd MXFP4 MoE scale group padding
- Explicit MXFP4 logical dtype arguments passed to grouped matmul
- Uneven tensor-parallel MX scale padding

Static validation was performed with:

~~~bash
git diff --check
~~~

The unit tests could not be executed in the local development environment because PyTorch and an Ascend NPU runtime are unavailable.

The relevant tests can be run in a configured vLLM-Ascend development environment with:

~~~bash
pytest -q tests/ut/quantization/methods/test_w4a4_mxfp4.py
pytest -q tests/ut/quantization/test_method_adapters.py \
  -k mx_scale_weight_loader
~~~

End-to-end validation should cover W4A4 MXFP4 model loading and inference with multiple tensor-parallel configurations, including configurations where scale groups cannot be divided evenly across ranks.

### DIFF
--- a/vllm_ascend/quantization/method_adapters.py
+++ b/vllm_ascend/quantization/method_adapters.py
@@ -63,34 +63,6 @@ def _make_mx_scale_weight_loader(
     return mx_scale_weight_loader
 
 
-def _make_padded_mx_scale_weight_loader(weight_loader, tp_size: int):
-    """Pad MX scales so the upstream loader can shard them evenly.
-
-    Some MX schemes (for example W8A8 MXFP8) create one local scale for
-    every partial group, so the sum of the local scale counts can be a few
-    entries larger than the unsharded checkpoint tensor.  Pad those missing
-    tail entries before delegating to the standard TP loader.
-    """
-
-    def mx_scale_weight_loader(param: torch.nn.Parameter, loaded_weight: torch.Tensor):
-        input_dim = getattr(param, "input_dim", None)
-        if input_dim is not None:
-            expected_size = param.shape[input_dim] * tp_size
-            loaded_size = loaded_weight.shape[input_dim]
-            padding = expected_size - loaded_size
-            if 0 < padding < tp_size:
-                padded_shape = list(loaded_weight.shape)
-                padded_shape[input_dim] = expected_size
-                padded_weight = loaded_weight.new_zeros(padded_shape)
-                index = [slice(None)] * loaded_weight.ndim
-                index[input_dim] = slice(0, loaded_size)
-                padded_weight[tuple(index)] = loaded_weight
-                loaded_weight = padded_weight
-        return weight_loader(param, loaded_weight)
-
-    return mx_scale_weight_loader
-
-
 class AscendLinearMethod(LinearMethodBase):
     """Linear method for Ascend quantization.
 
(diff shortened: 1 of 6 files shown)


Alex-stack-hub (approved):

vllm_ascend/quantization/methods/w4a4/w4a4_mxfp4.py
```diff
@@ -150,6 +155,22 @@ def process_weights_after_loading(self, layer):
 
         _rename_packed_weight_parameter(layer, "weight")
 
+        prefix_padding = suffix_padding = 0
```
kunpengW-code, line 168: Add notes to explain why padding is applied to weights and activations.

kunpengW-code (dismissed):

vllm_ascend/quantization/methods/w8a8/w8a8_mxfp8.py
```diff
@@ -208,17 +228,19 @@ def restore_weights_for_rl_loading(self, layer):
         orig_shapes = layer._mxfp8_original_shapes
         orig_scale_shape = orig_shapes["weight_scale"]
 
-        # Restore weight: (input_size, output_size) -> (output_size, input_size)
-        target_weight = layer.weight.data.transpose(0, 1).contiguous()
-        layer.weight.data = layer.weight.data.transpose(0, 1)
-        layer.weight.data.copy_(target_weight)
+        # Restore the unpadded checkpoint shape for the next weight load.
```
kunpengW-code: For now, do not modify the RL code logic. Adapt it separately later if any issues arise with the RL.

vllm_ascend/quantization/method_adapters.py
```diff
@@ -147,6 +171,16 @@ def create_weights(
                 or is_mx_quant_type(self.quant_method)
             ):
                 param.input_dim = 1
+            if isinstance(layer, RowParallelLinear) and getattr(
+                self.quant_method, "supports_unaligned_tp_groups", False
+            ):
+                assert hasattr(self.quant_method, "group_size")
+                param.weight_loader = _make_mx_scale_weight_loader(
```
lijiahang226, line 185: Since this bypasses the original weight_loader, I wonder if an already-sharded input could be handled incorrectly.

yzlbig: Good catch. I updated the wrapper to distinguish already-partitioned inputs from full checkpoint tensors using the local parameter shape. Already-sharded scales are copied directly, while full checkpoint scales continue through the phase-aware slicing path to preserve MX group alignment across TP boundaries. I also added a CPU unit test covering the pre-sharded case.

lijiahang226 (approved):
LGTM. Thanks for the contribution.

yzlbig:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

yzlbig:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

yzlbig:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

yzlbig:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

yzlbig:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

kunpengW-code (approved):

PR #16114, opened 2026-09-09
[Doc][Getting Started] Make setup and first inference reproducible

sugm0521:

Correct specific setup and first-inference ambiguities while preserving the existing documentation structure:

- Pair the main development checkout with its verified vLLM commit, and pair release CPU test checkouts with matching release tags.
- Identify commands that run inside the CPU test container and use the correct nightly image names for NPU examples.
- Clarify that the manual 910B Ops installation example applies to A2 and refer other hardware to the official CANN guide.
- Use `max_tokens` for completion requests, report HTTP failures with plain curl, and describe the expected nonempty completion.
- Fix the shared hardware table indentation in Quick Start so each product series renders as a separate row.

The existing chapters, hardware tabs, background-serving flow, startup/shutdown log examples, and four doctest stages are retained. No test infrastructure changes or new performance sections are included.

Only the documented commands and explanations change. There are no runtime/API changes. The examples retain the existing background-serving and process-lookup commands.

- `bash format.sh ci`: passed.
- Strict English documentation build via `tools/rtd_build.sh`: passed.
- 27 doctest helper tests passed; all eight online blocks extract and pass Bash syntax checks.
- Verified that the generated Quick Start hardware table contains four separate product-series rows. The original PID lookup and stop commands are retained.
- `git diff --check`: passed.

AI assistance was used for editing and local validation. NPU inference, CANN installation, source builds, and hardware-dependent tests have not been run on this macOS host. No performance measurements are claimed.

### DIFF
--- a/docs/source/developer_guide/contribution/testing.md
+++ b/docs/source/developer_guide/contribution/testing.md
@@ -23,7 +23,11 @@ The fastest way to set up a test environment is to use the main branch's contain
         -v $(pwd):/vllm-project \
         -v ~/.cache:/root/.cache \
         -ti $IMAGE bash
+    ```
 
+    Run the remaining commands inside the container:
+
+    ```bash
     # (Optional) Configure mirror to speed up download
     sed -i 's|ports.ubuntu.com|mirrors.huaweicloud.com|g' /etc/apt/sources.list
     pip config set global.index-url https://mirrors.huaweicloud.com/repository/pypi/simple/
@@ -40,7 +44,7 @@ The fastest way to set up a test environment is to use the main branch's contain
     apt-get install -y python3-pip git vim wget net-tools gcc g++ cmake libnuma-dev curl gnupg2
 
     git clone -b {{ vllm_ascend_version }} --depth 1 https://github.com/vllm-project/vllm-ascend.git
-    git clone --depth 1 https://github.com/vllm-project/vllm.git
+    git clone -b {{ vllm_version }} --depth 1 https://github.com/vllm-project/vllm.git
 
     # vllm
     cd $SRC_WORKSPACE/vllm
@@ -63,8 +67,8 @@ The fastest way to set up a test environment is to use the main branch's contain
 
     # Update DEVICE according to your device (/dev/davinci[0-7])
     export DEVICE=/dev/davinci0
-    # Update the vllm-ascend image
-    export IMAGE=quay.io/ascend/vllm-ascend:main
+    # A2 Ubuntu image; use nightly-main-a3 for A3 and add -openeuler for openEuler.
+    export IMAGE=quay.io/ascend/vllm-ascend:nightly-main
     docker run --rm \
         --name vllm-ascend \
         --shm-size=1g \
@@ -98,8 +102,8 @@ The fastest way to set up a test environment is to use the main branch's contain
 === "Multi-cards"
 
     ```bash
-    # Update the vllm-ascend image
-    export IMAGE=quay.io/ascend/vllm-ascend:main
+    # A2 Ubuntu image; use nightly-main-a3 for A3 and add -openeuler for openEuler.
+    export IMAGE=quay.io/ascend/vllm-ascend:nightly-main
     docker run --rm \
         --name vllm-ascend \
         --shm-size=1g \
--- a/docs/source/getting_started/installation/base_environment.inc.md
+++ b/docs/source/getting_started/installation/base_environment.inc.md
@@ -6,7 +6,13 @@
 
     #### Install CANN manually {: #installation-base-environment-install-cann }
 
-    Please refer to [CANN Installation Resources](https://www.hiascend.com/cann/download) or the following code to complete the installation.
+    The official [CANN Installation Resources](https://www.hiascend.com/cann/download) are the authoritative installation reference.
+
+    ???+ important "The example commands are for A2 only"
+
+        The commands below are for reference and install 910B Ops, which applies
+        only to A2. For other hardware, including Ascend 950DT, use the matching
+        Ops package according to the official guide.
 
     The commands below use the default CANN and NNAL installation paths. If you install either component in a non-default directory, source the corresponding `set_env.sh` from the actual installation directory.
 
--- a/docs/source/getting_started/quick_start.md
+++ b/docs/source/getting_started/quick_start.md
@@ -8,7 +8,8 @@ This guide uses Qwen3-0.6B as an example to help you run your first offline infe
 - Python: {{ release_python_version }}
 - Docker
 - Supported hardware:
-{% include "getting_started/installation/supported_hardware.inc.md" %}
+
+{% filter indent(4, true) %}{% include "getting_started/installation/supported_hardware.inc.md" %}{% endfilter %}
 
 ??? note "Software stack included in the vLLM Ascend image"
 
--- a/docs/source/getting_started/quick_start/online/qwen3-0.6b-310p.inc.md
+++ b/docs/source/getting_started/quick_start/online/qwen3-0.6b-310p.inc.md
@@ -30,22 +30,25 @@ You can query the model list:
 
 <!-- doctest: quickstart-300i-duo-online-model-list -->
 ```bash
-curl http://localhost:8000/v1/models | python3 -m json.tool
+curl --fail http://localhost:8000/v1/models
 ```
 
 You can also send a prompt to the model:
 
 <!-- doctest: quickstart-300i-duo-online-completion -->
 ```bash
-curl http://localhost:8000/v1/completions \
+curl --fail http://localhost:8000/v1/completions \
     -H "Content-Type: application/json" \
     -d '{
         "model": "Qwen/Qwen3-0.6B",
         "prompt": "Beijing is a",
-        "max_completion_tokens": 5,
+        "max_tokens": 5,
         "temperature": 0
-    }' | python3 -m json.tool
+    }'
 ```
+
+Confirm that the response contains nonempty generated text in `choices[0].text`.
+An error response or an empty completion does not confirm successful inference.
 
 vLLM is running as a background process. You can use `kill -2 $VLLM_PID` to stop it gracefully, which is similar to pressing `Ctrl+C` for a foreground vLLM process:
 
--- a/docs/source/developer_guide/contribution/index.md
+++ b/docs/source/developer_guide/contribution/index.md
@@ -22,7 +22,7 @@ python3 -m venv .venv
 source ./.venv/bin/activate
 
 # Clone vllm-ascend and install
-git clone https://github.com/vllm-project/vllm-ascend.git
+git clone --branch main https://github.com/vllm-project/vllm-ascend.git
 cd vllm-ascend
 
 # Install lint requirement and enable pre-commit hook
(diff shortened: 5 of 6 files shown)


docs/source/getting_started/installation/base_environment.inc.md
```diff
@@ -4,6 +4,16 @@
 
     This path is intended for advanced users who need to manage the user-space software stack themselves. You can install on an existing Linux host or start from a minimal Linux container.
 
+    ???+ important "The manual commands below are for A2 only"
+
+        The example installs the **910B** operator package. Do not use it unchanged
+        for A3, Atlas 300I DUO, Atlas 200I Pro, or 950DT. For those devices, choose
```
ella1107: Ascend 950DT

docs/source/getting_started/quick_start/online/qwen3-0.6b-310p.inc.md
```diff
@@ -1,69 +1,78 @@
...
-INFO:     Shutting down FastAPI HTTP server.
-INFO:     Shutting down
-INFO:     Waiting for application shutdown.
-INFO:     Application shutdown complete.
-```
+The check succeeds only when the request returns a nonempty completion for the
+requested model. HTTP errors, invalid/error JSON, empty completions, startup
+failure, and timeouts produce a nonzero exit status. Inspect `vllm-serve.log` on
+failure. The exit trap sends SIGINT only to the server process started by this
+example, on both success and failure. It does not search for other vLLM processes.
```
MrZ20: This feels too heavy for a Quick Start. The main goal here should be to show users how to start the server and send a request.

Port checks, readiness polling, timeout handling, PID cleanup, and response assertions are more suitable for doctests/CI. I’d prefer to keep the user-facing example simple, similar to [vLLM's Quick Start](https://docs.vllm.ai/en/latest/getting_started/quickstart/#online-serving).

docs/source/developer_guide/contribution/testing.md
```diff
@@ -4,7 +4,43 @@ This document explains how to write unit tests, E2E tests, and nightly tests to
...
+    ```
+
+For prebuilt NPU environments, select the corresponding hardware image in
+[Quick Start](../../getting_started/quick_start.md#quick-start-installation).
+For main development, replace the release tag with `main` and retain its hardware
+and OS suffix (for example, `main-a3` for A3 on Ubuntu). Record the image digest
+and both source revisions before testing; a mutable tag can change independently
+of a local checkout. The commands below test the sources installed inside the
+image at `/vllm-workspace`, not an unrelated host checkout. To test local changes,
+use the source installation path and verify the imported module location.
```
MrZ20: The current main-branch image publishing path uses the `nightly-` prefix, and our current documentation uses `nightly-main-a3` for the main-branch A3 image.

docs/source/developer_guide/contribution/testing.md
```diff
@@ -39,8 +77,18 @@ The fastest way to set up a test environment is to use the main branch's contain
     apt-get update -y
     apt-get install -y python3-pip git vim wget net-tools gcc g++ cmake libnuma-dev curl gnupg2
 
-    git clone -b {{ vllm_ascend_version }} --depth 1 https://github.com/vllm-project/vllm-ascend.git
-    git clone --depth 1 https://github.com/vllm-project/vllm.git
+    # In this container, run one source-pair selection block above first.
+    : "${VLLM_ASCEND_REF:?Select a source pair first}"
+    git clone -b "$VLLM_ASCEND_REF" --depth 1 https://github.com/vllm-project/vllm-ascend.git
+    if [ "$VLLM_ASCEND_REF" = main ]; then
+        VLLM_REF=$(tr -d '[:space:]' < vllm-ascend/.github/vllm-main-verified.commit)
```
MrZ20: Using `.github/vllm-main-verified.commit` makes sense, but this block feels more like CI scripting than documentation.

Checks like:

`: "${VLLM_ASCEND_REF:?Select a source pair first}"
[[ "$VLLM_REF" =~ ^[0-9a-f]{40}$ ]] || ...`

seem unnecessary if users follow the documented steps in order. Can we keep the version pairing but simplify the commands?

docs/source/developer_guide/contribution/testing.md
```diff
@@ -4,7 +4,43 @@ This document explains how to write unit tests, E2E tests, and nightly tests to
...
+
+For the CPU source-build path below, choose one pair in the container before
+cloning the repositories:
+
+=== "Release"
+
+    Use the paired tags from the [release compatibility matrix](../../community/versioning_policy.md#release-compatibility-matrix).
+
+    ```bash
+    VLLM_ASCEND_REF={{ vllm_ascend_version }}
```
MrZ20: The source-pair selection is shown before entering the container, while the clone commands are inside the container.

This means users need to scroll back and run the selection block again. It would be clearer to put the source selection directly before the clone commands.

docs/source/developer_guide/contribution/testing.md
```diff
@@ -63,8 +111,8 @@ The fastest way to set up a test environment is to use the main branch's contain
 
     # Update DEVICE according to your device (/dev/davinci[0-7])
     export DEVICE=/dev/davinci0
-    # Update the vllm-ascend image
-    export IMAGE=quay.io/ascend/vllm-ascend:main
+    # Set IMAGE to the selected release/main and hardware variant first.
+    : "${IMAGE:?Select the matching image as described above}"
```
MrZ20: Instead of:

`: "${IMAGE:?Select the matching image as described above}"`

I’d prefer to show a concrete export `IMAGE=...` here.

Users should be able to copy and run the example directly. We can add a short note for different hardware/OS suffixes.

docs/source/getting_started/installation/base_environment.inc.md
```diff
@@ -4,6 +4,16 @@
 
     This path is intended for advanced users who need to manage the user-space software stack themselves. You can install on an existing Linux host or start from a minimal Linux container.
 
+    ???+ important "The manual commands below are for A2 only"
```
MrZ20: The A2-only note makes sense. But the current note is also a bit long and repeats the CANN installation link below. 

You may write — the official CANN installation guide serves as the authoritative installation reference. The commands we provide below are for reference only; this example installs 910B Ops and thus is only applicable to A2. For other hardware, please replace it with the corresponding Ops package in accordance with the official guide.

docs/source/getting_started/quick_start/online/qwen3-0.6b-310p.inc.md
```diff
@@ -1,69 +1,78 @@
...
+
+with socket.socket() as sock:
+    sock.bind(("0.0.0.0", 8000))
+PYTHON
+    vllm serve Qwen/Qwen3-0.6B \
+        --dtype float16 \
+        --max-num-seqs 8 \
+        --compilation-config '{"cudagraph_mode":"FULL_DECODE_ONLY","cudagraph_capture_sizes":[1,2,4,8]}' \
+        --additional-config '{"ascend_compilation_config":{"enable_npugraph_ex":false}}' \
+        >vllm-serve.log 2>&1 &
```
MrZ20: I’d prefer to keep the documented command close to the canonical usage:
`vllm serve ...`
Adding `$!`, `VLLM_PID`, and `trap` makes the basic command harder to read.

docs/source/getting_started/quick_start/online/qwen3-0.6b-310p.inc.md
```diff
@@ -1,69 +1,78 @@
...
+        if ! kill -0 "$VLLM_PID" 2>/dev/null || (( SECONDS >= deadline )); then
+            cat vllm-serve.log >&2
+            echo "Server exited or startup timed out" >&2
+            exit 1
+        fi
+        sleep 1
+    done
+    kill -0 "$VLLM_PID"
+
+    curl --fail --silent --show-error --connect-timeout 5 --max-time 120 \
```
MrZ20: Do we really need `| python3 -m json.tool` here?

I think a plain `curl` request is enough and also matches the upstream Quick Start style. Keeping `max_tokens` is useful, but the extra formatting step doesn’t seem necessary.

docs/source/getting_started/quick_start.md
```diff
@@ -18,6 +18,35 @@ This guide uses Qwen3-0.6B as an example to help you run your first offline infe
 
     For the exact validated versions, see [Installation Guide > Hardware and software stack](installation.md#installation-hardware-software-stack).
 
+### Record a reproducible first run
```
MrZ20: I’m not sure `Record a reproducible first run` belongs here.

Image digest, model snapshot, cache state, and cold/warm timing are useful for benchmarking or reproducing performance results, but they don’t help much with the main Quick Start goal: getting the first inference running.

I’d suggest moving this to a benchmarking/reproducibility section, or keeping only a short optional note here.

MrZ20 (changes requested):

docs/source/getting_started/installation/base_environment.inc.md
sugm0521: Thanks. Updated the wording to use the full product name `Ascend 950DT` in the hardware-scope note.

docs/source/getting_started/quick_start/online/qwen3-0.6b-310p.inc.md
sugm0521: Agreed. I restored the existing Quick Start structure and the simple background `vllm serve` / `curl` flow. The port precheck, readiness loop, timeout handling, embedded Python assertions, and custom PID cleanup were removed. The existing doctest stages remain unchanged. Only `curl --fail`, the correct `max_tokens` request field, and a short explanation of the expected nonempty completion remain.

docs/source/developer_guide/contribution/testing.md
sugm0521: Fixed. The A2 examples now use `nightly-main`; the accompanying note identifies `nightly-main-a3` for A3 and says to add the `-openeuler` suffix for openEuler.

docs/source/developer_guide/contribution/testing.md
sugm0521: Agreed. I removed the defensive parameter guards and SHA-format check. The contribution guide now reads `.github/vllm-main-verified.commit` and fetches that commit directly, while the testing guide keeps the release pairing as adjacent explicit branch names.

docs/source/developer_guide/contribution/testing.md
sugm0521: Adjusted. The detached source-selection block was removed. The paired vLLM Ascend and vLLM release refs now appear directly on the two adjacent clone commands inside the container. I also split the code block after `docker run` so it is clear that the remaining commands run inside the container.

docs/source/developer_guide/contribution/testing.md
sugm0521: Fixed. Both NPU examples now provide the concrete `export IMAGE=quay.io/ascend/vllm-ascend:nightly-main` value, with a short note for the A3 image and openEuler suffix.

docs/source/getting_started/installation/base_environment.inc.md
sugm0521: Updated as suggested. The note now points to the official CANN installation guide once, states that the commands are for reference, identifies the installed 910B Ops package as A2-only, and directs other hardware, including Ascend 950DT, to the matching Ops package.

docs/source/getting_started/quick_start/online/qwen3-0.6b-310p.inc.md
sugm0521: Agreed. I restored the original `vllm serve ... &` command and the existing `pgrep`-based stop block. The added `$!`, launch-time `VLLM_PID` assignment, and trap were removed.

docs/source/getting_started/quick_start/online/qwen3-0.6b-310p.inc.md
sugm0521: Updated. Both examples now use plain `curl --fail`; `python3 -m json.tool` was removed, and `max_tokens` is retained for the completions request.

docs/source/getting_started/quick_start.md
sugm0521: Agreed. The `Record a reproducible first run` section was removed entirely, and this PR does not add a new benchmarking or reproducibility section. Quick Start remains focused on first inference; its only other current change is an indentation fix so the shared supported-hardware table renders one product series per row.

liqian79 (approved):

PR #16119, opened 2026-09-09
[Performance][KV Pool] Remove delayed block free tracking

weijinqian0:

AscendStore waits for all queued save work to complete in `wait_for_save()` before returning model output. This makes the request-level delayed block-free handshake redundant.

- Free request blocks immediately after synchronous AscendStore saves.
- Remove `delayed_free_req_ids` metadata and request-level `finished_sending` filtering.
- Clear worker-local synchronous completion records without forwarding them to the Scheduler.
- Remove the now-unused AscendStore SWA clipping helper.
- Preserve Hybrid Mamba `sending_events`/`completed_events` block-reference protection.

No. This simplifies the internal AscendStore block-release lifecycle after synchronous saves.

- `python -m py_compile` on the changed implementation and test files
- `git diff --check upstream/main...HEAD`
- Updated AscendStore Scheduler and Worker unit tests
- Focused pytest execution was not available in the local environment because `pytest` and `numpy` are not installed; CI verification is required.

### DIFF
--- a/tests/ut/distributed/ascend_store/test_pool_scheduler.py
+++ b/tests/ut/distributed/ascend_store/test_pool_scheduler.py
@@ -235,7 +235,7 @@ class TestKVPoolScheduler(unittest.TestCase):
         self.assertEqual(result, (False, None))
 
     @patch("vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.pool_scheduler.LookupKeyClient")
-    def test_request_finished_with_saved_tokens(self, mock_client_cls):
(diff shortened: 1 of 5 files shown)


LCAIZJ:
Can I understand it this way: Synchronous saving in AscendStore allows KV cache space to be freed as early as possible, which helps improve QPS but increases TTFT? On the other hand, asynchronous saving helps reduce TTFT, but results in lower QPS?

LCAIZJ (dismissed):
LGTM

nwpu-zxr:
LGTM.

nwpu-zxr (approved):

yiz-liu (approved):

PR #16121, opened 2026-09-09
[BugFix][Feat] Fix  _prepare_prefill_topk_scores_kernel op

AuroraEmiya:

This PR fixes the MiniMax-M3 prefill index top-k score preparation path on Ascend A3.

On A3 with Triton-Ascend 3.2.2, the previous `_prepare_prefill_topk_scores_kernel` could fail during sustained MiniMax-M3 serving workloads such as TerminalBench.

Two device-side failure modes were observed during investigation:

* The original query-tiled 2-D masked tail-store implementation could trigger an MTE/SMMU fault with `The DDR address of the MTE instruction is out of range`.
* A follow-up implementation that converted the stores to query-direction 1-D stores could trigger a Vector Core timeout/trap during real serving workloads.

The original implementation vectorized score updates across the query dimension, while the score tensor layout is:

```text
[num_index_heads, total_query_tokens, score_blocks]
```

where `score_blocks` is the contiguous dimension.

This PR changes the prefill score preparation to follow the tensor layout directly:

* Initialize the score buffer to `-inf`, so invalid/padded score blocks no longer need a separate tail-fill path.
* Keep the existing prefill QK score computation unchanged.
* Treat one `(query, index_head)` row as the logical preparation unit.
* Apply init/local forced priorities using block-contiguous stores along the last dimension.
* Keep query handling as task scheduling rather than using query as the vectorized memory-access dimension.
* Use an AIV-oriented program mapping so large/ragged prefills do not create one launched program per `(query, head)` and exceed the Ascend `coreDim <= 65535` launch limit.
* Preserve the existing public operator APIs and priority semantics:

  * init blocks: `1e30`
  * local blocks: `1e29`
  * local priority overrides init priority when they overlap
  * invalid blocks remain `-inf`

This removes the dynamic invalid-tail store path and avoids both the previous `[query, block]` masked GM store pattern and query-direction strided Vector stores.

The issue was reproduced on MiniMax-M3 running on Ascend A3 with DP=2 / TP=8 across 16 devices under sustained TerminalBench workloads.

The workaround is based on behavior observed with Triton-Ascend 3.2.2. The exact compiler/backend root cause has not been proven, so the implementation should be re-evaluated after Triton-Ascend upgrades.

No.

The public APIs, model behavior, output layout, top-k semantics, and sparse-attention semantics are unchanged.

This PR only changes the internal Triton implementation and task mapping used by MiniMax-M3 prefill index-score preparation on Ascend.

The patch was validated with standalone correctness cases, TerminalBench-style workload sweeps, and real MiniMax-M3 serving workloads on Ascend A3.

Correctness coverage includes:

* sparse-block boundaries: query lengths `127`, `128`, and `129`
* ragged query batches
* long invalid score tails
* init/local overlap
* `2048` / `2049` prefill boundary cases
* large ragged prefill workloads
* block counts matching the production workload range, including `480` and `528`

The regression tests verify the complete prefill index path:

```text
index score
→ score preparation
→ torch.topk
→ invalid-index masking
```

Representative TerminalBench-style end-to-end results with 528 score blocks:

```text
q=2048, prefix=0:
baseline  p50 = 2392.265 us
patched   p50 =  803.281 us

q=2048, prefix=32768:
baseline  p50 = 2006.960 us
patched   p50 =  847.873 us

q=2048, prefix=65512:
baseline  p50 = 1628.907 us
patched   p50 = 1537.166 us

q=4096, prefix=32768:
baseline  p50 = 1362.001 us
patched   p50 = 1315.651 us

q=8192, prefix=32768:
baseline  p50 = 2416.103 us
patched   p50 = 2336.767 us

q=32768, prefix=0:
baseline  p50 = 4760.416 us
patched   p50 = 4365.349 us
```

A large ragged TerminalBench-style workload also passed:

```text
query lengths:
[2048, 2048, 4096, 8192, 16384]

prefix lengths:
[0, 8192, 16384, 32768, 0]

total query tokens:
32768

score blocks:
528

baseline  p50 = 4809.989 us
patched   p50 = 4567.960 us
```

This case previously exposed the `coreDim` limitation for the direct one-query-per-program A5-style launch (`81920 > 65535`). The patched task mapping completes successfully.

Additional regression guards were added for the failure boundaries and large/ragged workloads. Large TerminalBench-style cases are kept as nightly/stability guards rather than regular lightweight CI cases.

### DIFF
--- a/vllm_ascend/models/minimax_m3/ops/msa_m3_triton.py
+++ b/vllm_ascend/models/minimax_m3/ops/msa_m3_triton.py
@@ -35,6 +35,16 @@ else:
     get_aicore_num = _get_aicore_num
     init_device_properties_triton = _init_device_properties_triton
 
+get_vectorcore_num: Callable[[], Any] | None
+try:
+    from vllm_ascend.ops.triton.triton_utils import (
+        get_vectorcore_num as _get_vectorcore_num,
+    )
+except ImportError:
+    get_vectorcore_num = None
+else:
+    get_vectorcore_num = _get_vectorcore_num
+
 # Data-layout constants.
 SPARSE_BLOCK_SIZE = 128
 SCORE_BLOCK_STRIDE_ALIGNMENT = 16
@@ -46,6 +56,11 @@ PREFILL_SCORE_QUERY_TILE_SIZE = 96
 # available; 32 is only a fallback when device-property discovery is unavailable.
 DECODE_SCORE_FALLBACK_PROGRAM_COUNT = 32
 DECODE_SCORE_MAX_CHUNK_COUNT = 256
+
+# Prefill prepare is a Vector/AIV workload. Keep the logical grid close to the
+# detected Vector Core count; each program processes a contiguous query range
+# for one (batch, head) group and only writes contiguous block segments.
+PREFILL_PREPARE_FALLBACK_VECTORCORE_COUNT = 64
 
 _FP8_DTYPES = (torch.float8_e4m3fn, torch.float8_e5m2)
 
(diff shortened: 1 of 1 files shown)


chenlongxiao (approved):

weiguihua2 (approved):

yingyingzizi:
The updated _prepare_prefill_topk_scores_kernel has passed an initial stability validation. Under a sustained, concurrent Terminal-Bench workload, the service completed 81 tasks and handled 6,700 inference requests without reproducing the previously observed Vector Core timeouts, AIVEC errors, or service crashes.
The attached screenshots show the aggregate results from this Terminal-Bench validation run. At the time of the snapshot, 81 tasks had produced complete evaluation results, with 55 tasks passing, corresponding to an interim accuracy of 67.90%. These results were obtained under a sustained concurrent workload using the updated _prepare_prefill_topk_scores_kernel, and provide additional evidence for the stability regression validation of this fix. Note that this was a retry run, so the screenshots represent an interim summary of the completed tasks rather than the final score for the full benchmark.
result1
result2

PR #16137, opened 2026-09-09
[Feature][KV Pool] Add AscendStore delayed-release and load metrics

bowgneo:

Adds AscendStore metrics for two operational questions:

- `vllm:ascend_store_delayed_release_requests`: current finished requests whose KV blocks remain allocated while asynchronous save completion is pending.
- `vllm:ascend_store_delayed_release_blocks`: current KV block references retained by those requests.
- `vllm:ascend_store_load_get_duration_seconds`: histogram of each worker's non-layerwise backend `get` duration.
- `vllm:ascend_store_load_get_keys_total`: total keys fetched by completed backend `get` calls.

The scheduler remains the source of truth for delayed-release membership. A request-to-block-count map and running total update both Gauges in constant time when a finished request enters delayed release and when worker completion clears it.

GET telemetry covers synchronous loading (`load_async=false`), asynchronous loading, and TP-mismatch loading. Each instrumented backend GET takes two `perf_counter` readings and updates interval statistics once per backend batch, not once per key.

The hot path adds no extra RPC, collective, per-key traversal, NPU synchronization, request-ID metric label, or metric log. Layerwise GET timing remains out of scope.

Periodic vLLM `KV Transfer metrics` summaries use `ascend_store_`-prefixed field names so their connector source is explicit. This only affects the existing vLLM stats log; Prometheus metric names are unchanged.

### Upstream design reference

The implementation follows [vLLM #43392 — Add metrics for MooncakeStoreConnector operations](https://github.com/vllm-project/vllm/pull/43392):

- the connector implements vLLM's `get_kv_connector_stats`, `build_kv_connector_stats`, and `build_prom_metrics` hooks;
- workers record one sample per backend batch under a short lock;
- `get_stats` swaps the accumulator so samples do not grow in worker memory;
- vLLM's existing connector pipeline performs worker aggregation and Prometheus observation;
- MultiConnector routes the AscendStore child through the upstream `MultiKVConnectorStats` and `MultiKVConnectorPromMetrics` implementations.

AscendStore additionally exposes delayed-release request and block Gauges because this live state belongs to its scheduler and has no MooncakeStore equivalent.

Yes. The four AscendStore metric families above are exposed through the standard `/metrics` endpoint.

- `python -m pytest -q tests/ut/distributed/ascend_store/test_metrics.py tests/ut/distributed/ascend_store/test_ascend_store_connector.py tests/ut/distributed/ascend_store/test_pool_scheduler.py tests/ut/distributed/ascend_store/test_pool_worker.py`: 143 passed.
- Tests use the repository-pinned vLLM main commit below.


Pz1116 (approved):

Ronald1995 (approved):
LGTM

yiz-liu (approved):

PR #16139, opened 2026-09-09
[Ops][Feature] Support AddRmsNormBias on Ascend A5

Foriv:

Add Ascend 950 (A5) support for the existing AddRmsNormBias operator, with upstream-derived FullLoad and SplitD implementations. Optional beta is added to the FP32 normalized output before the final output cast.

Wire the implementation through the A5 build list, `ascend950` OpDef configuration, Host tiling and APT entry. The ordinary `AscendRMSNorm` residual path lazily imports the existing extension, calls `enable_custom_op()`, and directly invokes `torch.ops._C_ascend.npu_add_rms_norm_bias`. Its return value is not used as a gate. `AscendGemmaRMSNorm` retains its pre-PR implementation, including its original custom-op gate and native fallback. The generic capability gate is unchanged; no separate A5 availability helper or new DeviceOperator method is introduced.

Keep the no-beta tiling when beta is present: FullLoad replaces one residual-output buffer with a single beta vector; SplitD replaces one gamma buffer with a single beta tile. Both inputs remain double-buffered. Output vector functions remain specialized for beta presence, so the no-beta specialization contains no beta loads or additions. Both optional-beta cases remain supported.

A5 ordinary RMSNorm residual calls use the custom fused operator with FP16, BF16 and FP32 and optional beta. This ordinary-class path requires the corresponding extension and OPP to be installed: loading or execution failures propagate instead of falling back to native AddRMSNorm. It calls the custom interface even when `enable_custom_op()` returns false, including when batch-invariant mode is enabled; this PR does not establish a batch-invariance guarantee for that path. This direct-call policy applies wherever the ordinary implementation is used, rather than being an A5-only Python condition.

GemmaRMSNorm, the non-residual paths and public signatures retain their pre-PR behavior. The existing A2/A3 kernel implementations are retained.

Atlas 300I DUO / 310P follows the existing compatibility-op registration: `register_ascend_customop()` selects `AscendRMSNorm310` and `AscendGemmaRMSNorm310`. These classes override `forward_oot`, so their normal registered path does not enter the ordinary-class direct-call branch changed here. RMSNorm310 uses `torch_npu.npu_add_rms_norm` followed by optional bias addition; GemmaRMSNorm310 explicitly adds the residual and calls `torch_npu.npu_rms_norm`. This is a source-level routing statement; no new 310P hardware run was performed.

Rebased onto main `ce9e24a28131b59c6df27470ac42aee742a711ad`; the new PR head is `df1502c8dabce2541a4c87b99f8f9d3f4aa00988`. This includes the upstream KV-cache CPU-test fixture fix. The rebase was conflict-free, and `git range-diff` confirmed all seven PR patches are unchanged. The ordinary RMSNorm integration, AddRmsNormBias implementation and retained operator tests are unchanged from the previous head.

For the rebased head, `git diff --check` passed and all 18 pre-commit hooks passed on server 107. The latter used the same hook revisions and arguments as `format.sh ci`, with an external configuration selecting the cached system Node/Go toolchains to avoid toolchain downloads. The tracked worktree remained clean. CPU/NPU test results for this rebased head are not yet available.

Before this rebase, the restored source tree (`d4a7a10fc`) was identical to `b9633abc6` (Git tree `a14bcceec973a55ab837657bb3034307ad48ec38`). The results below were previously obtained for `b9633abc6`, which was validated in the 107 `zrr_dev` container with CANN 9.1, PyTorch 2.10, torch-npu 2.10.0.post4, Python 3.11.10 and vLLM v0.28.0 source. Real-NPU tests used physical A5 device 5 after confirming it had no running processes.

- All 18 repository pre-commit hooks passed through `bash format.sh ci`.
- 108 temporary CPU checks passed: 81 dispatch/argument/output cases, nine non-residual cases, 16 cold/warm fullgraph captures, and two missing-extension error cases. These use the source functions with CPU/Meta operator substitutes and synthetic A2/A3/A5 profiles.
- All 48 retained direct-operator e2e cases passed, with no failures or skips: FP16/BF16/FP32, optional beta, FullLoad/SplitD, non-aligned tails, multidimensional normalization, large rows, beta rounding and invalid beta rejection.
- 37 real-NPU layer checks passed: one cold fullgraph capture with the real extension first imported during capture, 18 eager cases, six warm fullgraph cases with changed inputs, three routing/output checks with the batch-invariant setting enabled, and nine non-residual cases. Ordinary RMSNorm with and without beta and the restored Gemma path were covered. Constructors were bypassed; compilation used `backend="eager", fullgraph=True, dynamic=True`, not the production compiler backend. The batch-invariant setting checks verify routing and outputs, not a batch-invariance guarantee.
- The tested source was `b9633abc6` with a clean tracked tree, identical in content to the restored revision. Existing compiled artifacts were verified against the unchanged sources and archived hashes, including all three dtype kernels and both Host tiling library copies. CANN logs confirmed 69 custom kernel launches (23 layer, 46 direct) from the isolated validation OPP package; the two invalid-beta cases were rejected before kernel launch.
- GemmaRMSNorm was compared with the pre-PR source and is unchanged. The restoration does not change the C++/kernel sources.

No tests or performance measurements were rerun for this source-identical restoration. The 112 checks on server 111 were for the separate shared-gate revision `38344f0e4`, which has been reverted; they are not presented as validation of the restored routing.

Run the retained direct-operator test in a configured A5 environment:

```bash
python -m pytest --noconftest -q tests/e2e/nightly/single_node/ops/singlecard_ops/test_add_rms_norm_bias_a5.py
```

Earlier optimized-kernel graph-replay measurements gave these representative FP16 medians in microseconds; these are not new measurements for the latest Python revision:

| Shape | Beta | Native baseline | Custom |
| --- | --- | ---: | ---: |
| `(1024, 11264)` | present | 33.79 | 20.41 |
| `(1024, 32768)` | present | 180.08 | 128.27 |
| `(1024, 32768)` | absent | 128.35 | 129.92 |

The beta baseline is `torch_npu.npu_add_rms_norm` followed by `y.add_(beta)`; without beta it is `npu_add_rms_norm` alone. Measurements used five rounds of NPU-event timing, repeated graph replay and reused buffers. These checks and microbenchmarks do not establish full-model accuracy or end-to-end performance.


csrc/moe/add_rms_norm_bias/op_kernel/arch35/add_rms_norm_bias_regbase_split_d.h
```diff
@@ -0,0 +1,342 @@
...
+            ComputeMultiLevelReduce(level1Local, level2Local, level3Local, level1, level2, level3);
+        }
+        // stage3: 处理主块逻辑
+        for (uint32_t repeat = 0; repeat < masterLoop; repeat++) {
+            ComputeFormerHandle(level1Local, offset, level1, ubFactor, ubFactor);
+            offset += ubFactor;
+            level1 += 1;
+            ComputeMultiLevelReduce(level1Local, level2Local, level3Local, level1, level2, level3);
+        }
+        ComputeMultiLevelRstd<false>(dstLocal, position, level1Local, level2Local, level3Local, level1, level2, level3);
```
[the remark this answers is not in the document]
Foriv: Thanks for pointing this out. The final `level3` argument does bind to `avgFactor`; it is redundant in this instantiation. The call is valid C++ because the helper has [seven required parameters and two defaulted parameters](https://github.com/vllm-project/vllm-ascend/blob/603aedeabbc9dba993d2a1ffdea8a91b0af2d441/csrc/moe/add_rms_norm_bias/op_kernel/arch35/deps/reduce_common_regbase.h#L700-L710).

Here, every branch reaches `LevelMergeRstd<false>`. Both `avgFactor` and `epsilon` are used only inside [`if constexpr (IS_RSTD)`](https://github.com/vllm-project/vllm-ascend/blob/603aedeabbc9dba993d2a1ffdea8a91b0af2d441/csrc/moe/add_rms_norm_bias/op_kernel/arch35/deps/reduce_common_regbase.h#L672-L679), so the value passed as the eighth argument does not affect the reduction result in this `<false>` instantiation. The actual rstd calculation happens afterward in [`ComputeRstd(rstdLocal, epsilon, avgFactor, calRowNum)`](https://github.com/vllm-project/vllm-ascend/blob/603aedeabbc9dba993d2a1ffdea8a91b0af2d441/csrc/moe/add_rms_norm_bias/op_kernel/arch35/add_rms_norm_bias_regbase_split_d.h#L114).

The same call and helper signature are present in the pinned upstream ops-nn `v9.2.0-beta.2` (`30ef7dd563c8a4b74c3161835c8e47d1d96f87b6`), in `norm/add_rms_norm/op_kernel/arch35/add_rms_norm_regbase_split_d.h` and `norm/norm_common/op_kernel/reduce_common_regbase.h`.

Removing the extra argument would be a readability cleanup with unchanged results, rather than a high-severity correctness fix for the current instantiation. We are keeping the imported call unchanged in this PR for now.

vllm_ascend/ops/layernorm.py
```diff
@@ -69,7 +69,12 @@ def forward_oot(
         import torch_npu
 
         if residual is not None:
-            if enable_custom_op():
+            if is_950():
+                use_custom = enable_a5_add_rms_norm_bias()
```
lijiahang226, line 73: Could we avoid introducing a separate availability mechanism for this single A5 op, or share more of the existing custom-op capability detection?

Foriv: Thanks, I agree we can share more of the initialization logic. The operator-specific availability check is needed because A5 currently does not advertise `RUNTIME_CUSTOM_OPS`; enabling the generic gate would also enable unrelated custom-op paths. For AddRmsNormBias, we check that the packaged Ascend 950 implementation and its `Meta`/`PrivateUse1` registrations are available.

The current helper already reuses `bootstrap_custom_op_env()`. I can also extract the common lazy C++ extension-loading and `libcust_opapi.so` retry logic into a shared helper, while keeping the generic Python Meta registration in the existing path and retaining the operator-specific checks. This would reduce duplication without changing the global custom-op policy.

Wyz-134:
After verification, the performance has improved, and there is no impact on accuracy.

vllm_ascend/ops/layernorm.py
```diff
@@ -69,14 +69,12 @@ def forward_oot(
         import torch_npu
 
         if residual is not None:
-            if enable_custom_op():
-                x, _, residual = torch.ops._C_ascend.npu_add_rms_norm_bias(
-                    x, residual, self.weight, self.bias, self.variance_epsilon
-                )
-            else:
```
linfeng-yuan, line 76: Does 300I DUO execute this block now?

Foriv: No, Atlas 300I DUO / 310P does not enter this block through the normal operator registration path. The existing [compatibility registration](https://github.com/vllm-project/vllm-ascend/blob/d4a7a10fc28ed22b02376f297678000b9aed66c8/vllm_ascend/utils.py#L809-L844) maps RMSNorm and GemmaRMSNorm to `AscendRMSNorm310` and `AscendGemmaRMSNorm310`.

Both classes override `forward_oot` without calling the parent implementation. [RMSNorm310](https://github.com/vllm-project/vllm-ascend/blob/d4a7a10fc28ed22b02376f297678000b9aed66c8/vllm_ascend/_310p/ops/layernorm.py#L9-L24) continues to call `torch_npu.npu_add_rms_norm` followed by optional bias addition; [GemmaRMSNorm310](https://github.com/vllm-project/vllm-ascend/blob/d4a7a10fc28ed22b02376f297678000b9aed66c8/vllm_ascend/_310p/ops/layernorm.py#L27-L41) explicitly adds the residual and calls `torch_npu.npu_rms_norm`.

The latest revision leaves these 310P implementations and their registration unchanged.

vllm_ascend/ops/layernorm.py
Foriv: Following up on the current revision (`d4a7a10fc`): I have removed the dedicated A5 availability helper and its separate checks/cache. This supersedes the helper-refactoring proposal in my earlier reply.

The [ordinary RMSNorm residual path](https://github.com/vllm-project/vllm-ascend/blob/d4a7a10fc28ed22b02376f297678000b9aed66c8/vllm_ascend/ops/layernorm.py#L71-L78) now lazily imports the existing C++ extension, calls `enable_custom_op()`, and directly invokes the existing `torch.ops._C_ascend.npu_add_rms_norm_bias` interface. The generic enable implementation and hardware capability profiles are unchanged, so this does not globally enable other A5 custom-op paths.

For this ordinary-class path, the enable return value is not used as a gate, including when batch-invariant mode is set. The extension and corresponding OPP must be installed; loading or execution failures propagate instead of taking the native fallback. This direct-call policy is not restricted by an A5-only Python condition.

[GemmaRMSNorm](https://github.com/vllm-project/vllm-ascend/blob/d4a7a10fc28ed22b02376f297678000b9aed66c8/vllm_ascend/ops/layernorm.py#L87-L106) retains its original gate and fallback. The 310P/300I DUO compatibility overrides also remain unchanged and do not enter this ordinary-class branch.

ZT-AIA (approved):

PR #16146, opened 2026-09-09
[BugFix] Fix kimi k3 dspark kv grouping when --no-enable-prefix-caching

slippersss:
This PR aims to fix kimi k3 dspark kv grouping when not enabling prefix cache. The problem is that patch_mamba_config.py aligns ssm to k, which is from only target model without draft model. Since kimi k3 dspark may use gqa, kv cache layout thus can be problematic. That is, when aligning, mamba considers mla without gqa, so finally v in gqa can overlap ssm in mamba. Example is shown below:

```
Mamba: |   Conv   |               SSM               | Padding |
MLA:   | Padding  |            K(latent)            |  RoPE   |
GQA:   |    Padding     |        K         |        V         |
```

When enabling prefix cache, kimi k3 has its own kv grouping logic where no tensor is shared by mamba and gqa. So we just fall back into it.

N/A

by ci

### DIFF
--- a/vllm_ascend/patch/platform/patch_kv_cache_utils.py
+++ b/vllm_ascend/patch/platform/patch_kv_cache_utils.py
@@ -145,7 +145,10 @@ def _get_kimi_k3_dspark_mixed_kv_cache_groups(
     all_specs = [*attention_specs, *mamba_specs.values()]
     # Only attention layers share a scheduler block table. Mamba keeps its
     # own groups and may use max_model_len as block_size when cache_mode=none.
-    if len({spec.block_size for spec in attention_specs}) != 1 or len({spec.page_size_bytes for spec in all_specs}) != 1:
+    if (
+        len({spec.block_size for spec in attention_specs}) != 1
+        or len({spec.page_size_bytes for spec in all_specs}) != 1
+    ):
         return None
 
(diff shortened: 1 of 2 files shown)


maoxx241 (approved):

vllm_ascend/patch/platform/patch_kv_cache_utils.py
```diff
@@ -141,8 +139,14 @@ def _get_kimi_k3_dspark_mixed_kv_cache_groups(
     ):
         return None
 
-    all_specs = [*target_attention_specs.values(), *draft_attention_specs.values(), *mamba_specs.values()]
-    if len({spec.block_size for spec in all_specs}) != 1 or len({spec.page_size_bytes for spec in all_specs}) != 1:
+    attention_specs = [*target_attention_specs.values(), *draft_attention_specs.values()]
+    all_specs = [*attention_specs, *mamba_specs.values()]
+    # Only attention layers share a scheduler block table. Mamba keeps its
+    # own groups and may use max_model_len as block_size when cache_mode=none.
+    if (
```
weiguihua2, line 148: Need to supplement use case guardianship

slippersss: Done.

weiguihua2 (approved):

PR #16080, opened 2026-09-08
[Feature][MRV2][P/D] Support PCP KV transfer

li1how:
### What this PR does / why we need it?

Replace PCP KV sharding with per-request selection of a complete prefill replica. Reuse TP/group routing and existing completion tracking to release KV.

### Does this PR introduce _any_ user-facing change?

Enable prefill-side PCP with decode-side PCP disabled for MRV2 P/D disaggregation using MooncakeConnectorV1.

### How was this patch tested?

GPQA Diamond first 100 questions, concurrency 100. Qwen3-8B BF16 uses TP2 on both sides; DeepSeek-V4-Flash-w4a8 (DSA) uses TP4 on both sides with MooncakeHybridConnector.

| Model | Scenario | P PCP | D PCP | Accuracy | Valid answers | Output cap hits + failed requests (limit) |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| Qwen3-8B BF16 | PD baseline | 1 | 1 | 42/100 | 92 | 8 (4096) |
| Qwen3-8B BF16 | PCP + PD | 2 | 1 | 44/100 | 95 | 6 (4096) |
| DeepSeek-V4-Flash-w4a8 (DSA) | PD baseline | 1 | 1 | 72/100 | 98 | 2 (8192) |
| DeepSeek-V4-Flash-w4a8 (DSA) | PCP + PD | 2 | 1 | 81/100 | 99 | 1 (8192) |

Both Qwen3 runs passed 18 short/long/batched probes. Remote KV reads and cleanup were verified, with no duplicate DONEs or timeout reclamations.

- vLLM main: https://github.com/vllm-project/vllm/commit/b2f685834a6456197e7033966fdef52a23f1abcd

PR #16149, opened 2026-09-09
[BugFix][P/D] Fix garbled output in PD disaggregation with DCP + PP parallelism

fuzhihong699:

model: GLM-5.2

This PR fixes a KV‑cache transfer issue in the Mooncake connector when Pipeline Parallelism (PP) and Decode Context Parallel (DCP) are used together in a PD‑disaggregated service. 

When both PP > 1 and DCP > 1 are enabled, the D‑side workers only pull KV from P‑side PP rank 0. Layers belonging to PP rank 1+ are silently skipped, resulting in garbled output because the D‑side model misses half of its KV cache.

Root cause:
The get_cp_group_meta function in mooncake_connector.py did not iterate over the PP dimension when generating P‑side handshake ports. The generated ports only covered the range for PP rank 0: [port_base, port_base + tp_size * pcp_size). Ports for PP rank 1+ (located at offset pp_rank * pcp_size * tp_size) were never created. As a result, the D‑side never connected to PP rank 1+ workers, and the KV cache for those layers was never transferred.

Fixes (5 changes):

get_cp_group_meta (L3005): Added a pp_size parameter and an outer for pp_rank loop. Port generation now includes pp_rank_offset = pp_rank * pcp_size * tp_size to cover all PP ranks.

get_local_remote_block_port_mappings call sites (L3025‑3028): Pass self._prefill_pp_size to the P‑side call and self._decode_pp_size to the D‑side call.

Mapping logic in get_local_remote_block_port_mappings (L3032‑3052): Replaced the old round‑robin selection (where each D CP group consumed only one P CP group, silently dropping PP rank 1+) with a PP‑aware mapping: each D CP group now maps to one P CP group per PP rank.

Non‑HMA assertion (L3332): Changed tp_num_need_pulls == len(remote_handshake_port_list[0]) to len(remote_handshake_port_list[0]) % tp_num_need_pulls == 0. When PP > 1, each shard has pp_size * tp_num_need_pulls ports, so the strict equality would cause a crash.

PP rank derivation in _get_cp_shard_pulls (L3358): Replaced 0 if remote_pcp_size > 1 else (port - base) // tp_size with (port - base) // (remote_pcp_size * prefill_tp_size). The old formula always returned 0 or ignored the PCP dimension; the new one correctly divides by the stride pcp_size * tp_size (ports per PP rank).

No. When pp_size == 1 (the default and most common case), all changes become no‑ops:

The PP‑rank loop runs once with offset 0.

The PP‑aware mapping falls back to the original round‑robin behaviour (because p_cp_groups_per_pp == len(p_cp_groups)).

PP rank derivation always returns 0.

The assertion len % tp_num == 0 degenerates to tp_num == len (same as before).

Manual testing with a PD‑disaggregated deployment using the glm‑5 model (P: PP=2, TP=8; D: TP=2, DCP=8), verifying that the output is no longer garbled.

Regression tests under PP=1, DCP=1 and PP=2, DCP=1 configurations to ensure no regressions.

### DIFF
--- a/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
+++ b/vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
@@ -3014,7 +3014,7 @@ class MooncakeConnectorWorker:
                     kv_head_groups.append(tuple([kv_head_ids_]))
                 return kv_head_groups
 
-        def get_cp_group_meta(tp_size, pcp_size, dcp_size, port_base):
+        def get_cp_group_meta(tp_size, pcp_size, dcp_size, port_base, pp_size=1):
             # key is kv_head_group, value is cp_groups and which cp_groups to select
             cp_group_meta: dict = {}
             kv_head_groups = get_kv_head_groups(tp_size)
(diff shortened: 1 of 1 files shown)


vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py
```diff
@@ -3353,7 +3365,7 @@ def _get_cp_shard_pulls(self, remote_handshake_port_list, prefill_tp_size, remot
                 pulls = []
                 port_tp = (port - remote_base_port) % prefill_tp_size
                 # PCP and PP are mutually exclusive; when PCP > 1, pp_rank is always 0.
-                pp_rank = 0 if remote_pcp_size > 1 else (port - remote_base_port) // prefill_tp_size
+                pp_rank = (port - remote_base_port) // (remote_pcp_size * prefill_tp_size)
                 # The first attn_num ports of each shard (i.e., the original ports with randomly substituted TPs).
                 if port_idx < attn_num:
```
lijiahang226, line 3377: Does _get_cp_shard_pulls() also need to be made PP-aware?

With PP > 1, ports can now contain multiple groups of attn_num attention ports, but this path still assumes a single group (port_idx < attn_num, remote_tp_offset=port_idx). Should the TP offset be computed per PP group, e.g. port_idx % attn_num?

fuzhihong699: The rollback has been performed, and the following changes are used.

https://github.com/vllm-project/vllm-ascend/pull/16080

PR #16153, opened 2026-09-09
[v0.26.0rc][BugFix][SpecDecode] Guard DSpark at max model length

zhenwenqi2024:

On the v0.26 DSpark path, `set_inputs_first_pass` always expands a fixed parallel draft block. When

```
effective_seq_len + num_query_per_req > max_model_len
```

the expansion kernel computes a block-table address for query tokens beyond the model limit. Clamping only the tail is not correct for DSpark because the draft block is evaluated jointly (and is non-causal for DeepSeek V4); an invalid tail can contaminate the drafts that appear to remain in range.

This patch adds a request-level boundary fallback:

- compute device and host validity masks for the complete DSpark query window;
- if any query would exceed `max_model_len`, disable the whole DSpark block for that request;
- use safe metadata for the disabled row (position `0`, sequence length `1`);
- write `PADDING_SLOT_ID` (`-1`) for all of its draft KV slots, selecting block-table column zero before masked pointer arithmetic;
- replace every returned draft token for that request with `-1`, so the scheduler uses the target-model token only;
- leave the shared DFlash kernel path unchanged unless the new DSpark-only check is enabled.

Related to #15880, which addresses broader DSpark pointer/window validation on v0.27.1rc. This PR is a focused v0.26.0rc fix for the max-model-length boundary.

No API or configuration change. DSpark now falls back to target-only decoding for a request when its complete parallel draft window cannot fit within `max_model_len`.

- `python -m pytest -q tests/ut/spec_decode/test_dspark_proposer.py`
  - 44 passed on an Ascend A3 runtime using `quay.io/ascend/vllm-ascend:nightly-releases-v0.26.0rc-a3`.
- NPU regression input from `tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_parallel_draft_max_model_len.py`:
  - exact-boundary request (`3 + 5 == 8`) kept slots `[43, 80, 81, 82, 83]`;
  - over-limit request (`6 + 5 > 8`) produced positions `[0, 0, 0, 0, 0]` and slots `[-1, -1, -1, -1, -1]`;
  - PASS on the same Ascend A3 container.
- Shared-kernel DFlash compatibility smoke on NPU, omitting all new optional arguments: PASS.
- `pre-commit run ruff-check --files ...`: passed.
- `pre-commit run ruff-format --files ...`: passed.
- `python -m compileall -q ...`: passed.

### Short-sequence performance validation

Tested on Ascend A3 with DeepSeek V4 Flash P/D (P: DP4/TP4; D: DP16/TP1), DSpark5, DSA-CP, Mooncake, task queue enabled. The workload used 512 requests at concurrency 256, random input/output lengths 128 (211 actual input tokens after the chat template), infinite request rate, and one warmup followed by five measured runs.

Five-run medians:

| Metric | Without this PR | Optimized guard | Delta |
| --- | ---: | ---: | ---: |
| Output throughput | 4368.50 tok/s | 4417.94 tok/s | +1.13% |
| Mean TPOT | 36.940 ms | 36.988 ms | +0.13% |
| Median TPOT | 37.039 ms | 37.285 ms | +0.67% |
| Mean ITL | 78.916 ms | 79.375 ms | +0.58% |

The sub-1% latency differences are within run-to-run variance; the optimized guard shows no measurable short-sequence regression.

A same-process Ascend kernel microbenchmark (batch 256, draft width 5, 1,000 launches per sample, 14 interleaved samples) measured 212.724 us with the CHECK_REQUEST_WINDOW=False specialization, versus 212.738 us for the earlier duplicated fast kernel (-0.007%). The guarded specialization measured 220.662 us (+3.731%), so the CPU batch predicate and tl.constexpr specialization keep that cost off ordinary short-sequence batches.

### DIFF
--- a/vllm_ascend/spec_decode/dspark_proposer.py
+++ b/vllm_ascend/spec_decode/dspark_proposer.py
@@ -8,6 +8,7 @@ from vllm.config import CUDAGraphMode, VllmConfig, get_layers_from_vllm_config
 from vllm.model_executor.layers.attention_layer_base import AttentionLayerBase
 from vllm.v1.attention.backends.utils import CommonAttentionMetadata
 from vllm.v1.kv_cache_interface import UniformTypeKVCacheSpecs
+from vllm.v1.spec_decode.utils import PADDING_SLOT_ID
 from vllm.v1.worker.utils import AttentionGroup
 
 from vllm_ascend.ascend_forward_context import set_ascend_forward_context
@@ -119,6 +120,11 @@ class AscendDSparkProposer(AscendDflashProposer):
 
         # per-layer context slot mappings as a flat list
         self._context_slot_mapping_buffers: list[torch.Tensor | None] | None = None
+        # DSpark runs a request's draft block as one non-causal forward. If
+        # any query in that block would cross max_model_len, the whole block
+        # must be disabled rather than keeping only its in-range prefix.
+        self._request_window_ok = torch.ones(self.max_batch_size, dtype=torch.bool, device=device)
+        self._request_window_ok_cpu = torch.ones(self.max_batch_size, dtype=torch.bool)
 
     @staticmethod
     def _resolve_kernel_block_size(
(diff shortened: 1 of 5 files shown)


vllm_ascend/spec_decode/dspark_proposer.py
```diff
@@ -333,20 +394,24 @@ def set_inputs_first_pass(
...
-        if has_num_rejected:
-            effective_seq_lens = effective_seq_lens - num_rejected_tokens_gpu
-
         cad.query_start_loc = self.arange_dflash[: batch_size + 1] * self.num_query_per_req
-        cad.seq_lens = effective_seq_lens + self.num_query_per_req
+        expanded_seq_lens = effective_seq_lens + self.num_query_per_req
+        cad.seq_lens = torch.where(
+            request_window_ok,
+            expanded_seq_lens,
+            torch.ones_like(expanded_seq_lens),
```
drslark, line 413: Do our operators support 0?

PR #16087, opened 2026-09-08
[BugFix][KDA] Allocate Post-WU workspace for padded chunks

maoxx241:
### What this PR does / why we need it?

`chunk_kda_fwd` allocates Post-WU scratch using the packed token count, but `WScratchOffset` addresses it using `total_chunks * chunk_size`. When sequences contain tails, full-chunk Cube writes can overrun this allocation and corrupt the U seed buffer that is still in use. This produces NaN outputs/states and can lead to all-invalid speculative sampling rows (`valid_sampled_tokens_count=0`).

For example, 127 sequences alternating between 16, 32 and 64 tokens contain 4,720 tokens. With six heads and K=128, the old allocation is 14,499,840 bytes, while the padded layout requires 24,969,216 bytes. Allocate scratch using the padded chunk count, matching both the generic and A5 addressing. This also covers fixed-length inputs with a partial final chunk.

Add regression coverage for 127/128 variable-length sequences and fixed lengths 80/129, with both recompute modes, comparing outputs and final states against an independent CPU reference.

### Does this PR introduce _any_ user-facing change?

Fixes incorrect KDA outputs and downstream sampling errors for affected sequence batches. No API or configuration change.

### How was this patch tested?

On Ascend A3 with CANN 9.1.0:

- Replayed the same saved failing operator inputs twice before and after rebuilding only the tiling library. Before: U contained 22,612 NaNs and 316 Infs. After: all 12 returned tensors were finite; inputs were unchanged.
- New regression matrix: 8 passed. Restoring the old library makes `[127-None-True]` fail; the fixed library was restored afterward.
- Full `test_chunk_kda_aclnn.py`: 21 passed, 5 A5-specific cases skipped.
- `bash format.sh ci` and `git diff --check` passed.

[...]

PR #16180, opened 2026-09-09
[BugFix][KDA] Allocate Post-WU workspace for padded chunks

MaybeChz:

`chunk_kda_fwd` allocates Post-WU scratch using the packed token count, but `WScratchOffset` addresses it using `total_chunks * chunk_size`. When sequences contain tails, full-chunk Cube writes can overrun this allocation and corrupt the U seed buffer that is still in use. This produces NaN outputs/states and can lead to all-invalid speculative sampling rows (`valid_sampled_tokens_count=0`).

For example, 127 sequences alternating between 16, 32 and 64 tokens contain 4,720 tokens. With six heads and K=128, the old allocation is 14,499,840 bytes, while the padded layout requires 24,969,216 bytes. Allocate scratch using the padded chunk count, matching both the generic and A5 addressing. This also covers fixed-length inputs with a partial final chunk.

Add regression coverage for 127/128 variable-length sequences and fixed lengths 80/129, with both recompute modes, comparing outputs and final states against an independent CPU reference.

This is a re-submission of the same fix from #16087, authored from this account.

Fixes incorrect KDA outputs and downstream sampling errors for affected sequence batches. No API or configuration change.

On Ascend A3 with CANN 9.1.0 (original validation of this change):

- Replayed the same saved failing operator inputs twice before and after rebuilding only the tiling library. Before: U contained 22,612 NaNs and 316 Infs. After: all 12 returned tensors were finite; inputs were unchanged.
- New regression matrix: 8 passed. Restoring the old library makes `[127-None-True]` fail; the fixed library was restored afterward.
- Full `test_chunk_kda_aclnn.py`: 21 passed, 5 A5-specific cases skipped.
- `bash format.sh ci` and `git diff --check` passed.

Full-service validation on a single A3 node uses TP16/EP16, a 16-expert Kimi K3 target, GQA DSpark with 7 speculative tokens, prefix caching disabled, and `max_num_seqs=256`. All three phases reached 128 concurrent running requests and passed: 128 short requests with 256 output tokens, 512 mixed 16/32/64-token prompts with 128 output tokens, and another 256 mixed requests after removing the diagnostic hooks. Total: 896/896 HTTP 200 responses and 131,072 output tokens, with no missing/nonfinite selected logprobs or service errors.

### DIFF
--- a/csrc/attention/chunk_kda_fwd/op_host/chunk_kda_fwd_tiling.cpp
+++ b/csrc/attention/chunk_kda_fwd/op_host/chunk_kda_fwd_tiling.cpp
@@ -268,7 +268,11 @@ ge::graphStatus Tiling4ChunkKdaFwd(gert::TilingContext *context)
 
     const uint64_t postWuScratchOffset = AlignWorkspace(cursor);
     if (!arch35Options.fusePostWu && !arch35Options.fusePostWuIntoFwdH) {
-        cursor = postWuScratchOffset + tokenHeads * shape.kDim * sizeof(float);
+        // WScratchOffset stores each chunk with chunkSize rows, including
+        // padding for short sequences. Packed token counts underallocate this
+        // region and let full-chunk writes overwrite the following workspace.
+        cursor = postWuScratchOffset + hChunkCount * shape.vHeads * chunkSize *
+            shape.kDim * sizeof(float);
     }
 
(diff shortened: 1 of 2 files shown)


lijiahang226:
@ZT-AIA Please review this PR.

ZT-AIA (approved):

linfeng-yuan (approved):

PR #16194, opened 2026-09-10
[Feature][Platform]: Patch use sp moe

MmMmaru:
Currently, SP uses upstream SP moe function to judge SP is enabled or not. However, upstream SP option only supports DP>1.
When DP=1, SP is functionable
Tested on A5, with SP on.


MmMmaru:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #16198, opened 2026-09-10
[Performance][Attention] Use npu_transpose_batchmatmul for SFA q up-projection

31-ryougishiki:

`AscendSFAImpl._q_proj_and_k_up_proj` computes the SFA Q up-projection with
`transpose -> torch.bmm -> transpose`:

```python
q_nope = q_nope.transpose(0, 1)
ql_nope = torch.bmm(q_nope, self.W_UK_T)
return ql_nope.transpose(0, 1), q_pe
```

This PR folds both permutes into the matmul: a single
`torch_npu.npu_transpose_batchmatmul` call (`perm_x1=(1, 0, 2)`,
`perm_x2=(0, 1, 2)`, `perm_y=(1, 0, 2)`) does the layout conversion inside the
NPU op, following the pattern already used by `_v_up_proj` in the same file.

The original `transpose + torch.bmm` path is kept as a fallback for torch_npu
builds that do not provide the fused op, so behavior elsewhere is unchanged.

Note: #16071 also targets this op, with a larger change (padded `W_UK_T`, which
additionally removes the `split`); this PR intentionally keeps the change minimal.

No. Numerics are unchanged; only the op dispatch on the SFA q up-projection
path changes.

- Local latency of the SFA q up-projection on GLM-5.2-w4a4c8:
  `59.5us -> 37.7us` (~37% lower, ~21.8us saved per call).
- Accuracy: GSM8K on GLM-5.2-w4a4c8 shows no accuracy issue with the fused path.
- CI.

### DIFF
--- a/vllm_ascend/attention/sfa_v1.py
+++ b/vllm_ascend/attention/sfa_v1.py
@@ -65,6 +65,9 @@ if TYPE_CHECKING:
 # token count limits within bmm_transpose operator
 BMM_TRANS_MAX_SUPPORTED_TOKENS = 1024
 
+# npu_transpose_batchmatmul rejects operand dimensions >= 65536
+TRANSPOSE_BMM_MAX_SUPPORTED_DIM = 65536
+
 
 class PreprocessType(enum.Enum):
     NATIVE = "native"
@@ -864,7 +867,11 @@ class AscendSFAImpl(MLAAttentionImpl):
             .split([self.qk_nope_head_dim, self.qk_rope_head_dim], dim=-1)
         )
 
-        if hasattr(torch_npu, "npu_transpose_batchmatmul"):
+        if (
+            q_nope.dtype in [torch.float16, torch.bfloat16]
+            and hasattr(torch_npu, "npu_transpose_batchmatmul")
+            and q_nope.shape[0] < TRANSPOSE_BMM_MAX_SUPPORTED_DIM
+        ):
             # Convert from (B, N, P) to (N, B, P) and multiply
             # (N, B, P) x (N, P, L) -> (B, N, L)
             ql_nope = torch_npu.npu_transpose_batchmatmul(
@@ -875,6 +882,8 @@ class AscendSFAImpl(MLAAttentionImpl):
                 perm_y=(1, 0, 2),
             )
         else:
+            # Fallback for torch_npu builds without the fused op, unsupported
+            # dtypes, or a token dim beyond the operand limit.
             # Convert from (B, N, P) to (N, B, P)
             q_nope = q_nope.transpose(0, 1)
             # Multiply (N, B, P) x (N, P, L) -> (N, B, L)
--- a/vllm_ascend/attention/sfa_v1.py
+++ b/vllm_ascend/attention/sfa_v1.py
@@ -864,12 +864,24 @@ class AscendSFAImpl(MLAAttentionImpl):
             .split([self.qk_nope_head_dim, self.qk_rope_head_dim], dim=-1)
         )
 
-        # Convert from (B, N, P) to (N, B, P)
-        q_nope = q_nope.transpose(0, 1)
-        # Multiply (N, B, P) x (N, P, L) -> (N, B, L)
-        ql_nope = torch.bmm(q_nope, self.W_UK_T)
-        # Convert from (N, B, L) to (B, N, L)
-        return ql_nope.transpose(0, 1), q_pe
+        if hasattr(torch_npu, "npu_transpose_batchmatmul"):
+            # Convert from (B, N, P) to (N, B, P) and multiply
+            # (N, B, P) x (N, P, L) -> (B, N, L)
+            ql_nope = torch_npu.npu_transpose_batchmatmul(
+                q_nope,
+                self.W_UK_T,
+                perm_x1=(1, 0, 2),
+                perm_x2=(0, 1, 2),
+                perm_y=(1, 0, 2),
+            )
+        else:
+            # Convert from (B, N, P) to (N, B, P)
+            q_nope = q_nope.transpose(0, 1)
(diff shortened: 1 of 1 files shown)


31-ryougishiki:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/attention/sfa_v1.py
```diff
@@ -864,12 +864,24 @@ def _q_proj_and_k_up_proj(self, x):
             .split([self.qk_nope_head_dim, self.qk_rope_head_dim], dim=-1)
         )
 
-        # Convert from (B, N, P) to (N, B, P)
-        q_nope = q_nope.transpose(0, 1)
-        # Multiply (N, B, P) x (N, P, L) -> (N, B, L)
-        ql_nope = torch.bmm(q_nope, self.W_UK_T)
-        # Convert from (N, B, L) to (B, N, L)
-        return ql_nope.transpose(0, 1), q_pe
+        if hasattr(torch_npu, "npu_transpose_batchmatmul"):
```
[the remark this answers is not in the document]
31-ryougishiki: changed

vllm_ascend/attention/sfa_v1.py
```diff
@@ -864,12 +864,24 @@ def _q_proj_and_k_up_proj(self, x):
...
+            # Convert from (B, N, P) to (N, B, P) and multiply
+            # (N, B, P) x (N, P, L) -> (B, N, L)
+            ql_nope = torch_npu.npu_transpose_batchmatmul(
+                q_nope,
+                self.W_UK_T,
+                perm_x1=(1, 0, 2),
+                perm_x2=(0, 1, 2),
+                perm_y=(1, 0, 2),
+            )
```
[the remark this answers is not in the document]
31-ryougishiki: Thanks for the review.

`q_nope` is a strided view from `split(..., dim=-1)`, yes — and that is intentional here:
folding the permute into `npu_transpose_batchmatmul` is what removes the two explicit
transposes, and the latency in the PR description (`59.5us -> 37.7us`) was measured with this
exact non-contiguous input on our stack: torch_npu <x.y.z>, CANN <a.b>, GLM-5.2-w4a4c8 with
GSM8K accuracy verified.

Adding an explicit `.contiguous()` would materialize a `(B, N, P)` copy right before the op,
which is exactly the kind of copy this change is meant to avoid, so we would rather not add it
unconditionally. The `hasattr(torch_npu, "npu_transpose_batchmatmul")` guard keeps the original
`transpose + torch.bmm` path for builds without the fused op. If there is a known torch_npu
version that rejects non-contiguous input for this op, we can version-gate the fused path.

vllm_ascend/attention/sfa_v1.py
lijiahang226: Should we also guard the supported dtype/shape range here? hasattr() alone may select npu_transpose_batchmatmul for unsupported dimensions instead of falling back to torch.bmm.

31-ryougishiki: Thanks for the review — I'd prefer to keep this PR as is. Reasoning:

1. This call does not add a new failure mode. Later in the very same `forward()`, `_v_up_proj`
   (sfa_v1.py:1506 -> sfa_v1.py:901) already calls the same op with the same token dimension,
   and its npu branch has no dtype/shape guard either:

       if hasattr(torch_npu, "npu_transpose_batchmatmul"):
           x = x.view(-1, self.local_num_heads, self.kv_lora_rank)
           x = torch_npu.npu_transpose_batchmatmul(x, self.W_UV, perm_x1=(1, 0, 2), perm_y=(1, 0, 2))

   So for any batch where the new call could exceed the operand limits, the existing call in the
   same layer hits the same limit with the same `B` — those limits are already an implicit
   precondition of this path, not something introduced here.

2. The same unguarded pattern is used at the other call sites of this op in the repo:
   `mla_v1._v_up_proj` calls it unconditionally, and the two DSA o_proj sites rely on an
   upper-level fp8/bf16 branch. Guarding only the new call site would be inconsistent with the
   rest of the code; the proper fix is a shared capability check applied to all call sites.

3. For dtype: `q_nope` comes from `self.q_proj(x)`, and SFA weights/activations are fp16/bf16
   (`process_weights_after_loading` asserts `kv_b_proj.weight.dtype == act_dtype`), so a dtype
   check here would be dead code on this path.

So I'd rather keep this PR minimal (single-purpose). If you think the op-level guard is worth
adding, I'm happy to do it for all call sites in a follow-up PR.

31-ryougishiki: Updated as suggested — the fused branch now checks dtype and shape, mirroring `_v_up_proj`:

    # npu_transpose_batchmatmul rejects operand dimensions >= 65536
    TRANSPOSE_BMM_MAX_SUPPORTED_DIM = 65536
    ...
    if (
        q_nope.dtype in [torch.float16, torch.bfloat16]
        and hasattr(torch_npu, "npu_transpose_batchmatmul")
        and q_nope.shape[0] < TRANSPOSE_BMM_MAX_SUPPORTED_DIM
    ):
        ...
    else:
        # unchanged transpose + torch.bmm fallback

Only the token dim can approach the limit here (local_num_heads <= 128, qk_nope_head_dim <= 192,
kv_lora_rank = 512). The checks are Python-level dtype/shape comparisons, so the latency numbers
in the description are unchanged; unsupported cases now fall back to `torch.bmm` instead of raising.

lijiahang226:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

lijiahang226 (approved):

pisceskkk (approved):
LGTM

ZT-AIA (approved):

PR #16200, opened 2026-09-10
[BugFix][Model] DeepSeek-V4 Vision: skip vision SWA index build for text-only batches

ivyilike:
For `is_mm_prefix_lm` models (DeepSeek-V4 vision), the metadata builder in `NPUModelRunner` unconditionally inserts an entry into `mm_req_doc_ranges` for every scheduled request — text-only requests get an empty list. A dict of empty lists is truthy, so the gate

    if has_prefill and max_image_tokens > 0 and mm_ranges:

in the DSA metadata builder passes, and `build_vision_bidirectional_swa_indices` runs on every text-only prefill even though there are no image spans to apply. Profiling shows this costs ~1.1 ms per step (multiple torch kernel launches) on pure-text traffic.
Fix: only insert entries for requests that actually carry image spans. A text-only batch now yields an empty dict, so the gate short-circuits and the vision SWA index build is skipped. Mixed batches are unchanged: `build_vision_bidirectional_swa_indices` iterates `mm_prefix_ranges.items()` and empty range lists contributed nothing, so dropping them is semantically a no-op.
Related roadmap: https://github.com/vllm-project/vllm-ascend/issues/15462.

No. Text-only batches now skip an unnecessary computation; indices for
image-bearing requests are identical.

On-device profiling of a text-only request before/after:  `build_vision_bidirectional_swa_indices` no longer appears in the
attention metadata build path.

Profile comparison
(only text request，DeepSeek-V4 Vision Exp W8A8, 8TP):
Before: `build_attention_metadata` costs 21ms

After: `build_attention_metadata` costs 13ms and `build_vision_bidirectional_swa_indices` no longer appears


weiguihua2 (approved):

lijiahang226 (approved):

lijiahang226:
@linfeng-yuan, would it be possible to continue moving this PR forward toward merge? Thanks!

PR #11692, opened 2026-07-09
[Misc][Platform]enable model runner v2 by default via whitelists

wxsIcey:
### What this PR does / why we need it?
Replace the `use_v2_model_runner` monkey-patch with an Ascend-native property on `VllmConfig,` applied in `NPUPlatform.check_and_update_config`.

- Model whitelist: default-V2 architectures (Qwen3ForCausalLM) .
- Feature whitelist: whitelisted models may stack eagle/mtp/dflash spec decoding.
- V2 requires Triton; fall back to V1 otherwise.

### Does this PR introduce _any_ user-facing change?
N/A

### How was this patch tested?
CI passed with new added/existing test.

Co-authored-by: zouzy

- vLLM main: https://github.com/vllm-project/vllm/commit/e6bfe03ad73a3330cb427885aa90d97a12e1c704

PR #16203, opened 2026-09-10
[Misc][MRv2]enable model runner v2 by default via whitelists

yjyang62:

Rebase https://github.com/vllm-project/vllm-ascend/pull/11692 onto current `main`.

`main` currently overrides `VllmConfig.use_v2_model_runner` with an env-only gate: unless `VLLM_USE_V2_MODEL_RUNNER` is set, Ascend always uses Model Runner V1. Upstream vLLM instead defaults V2 from its own architecture / Triton / feature checks. Following those GPU defaults on NPU can crash, because the Ascend V2 runner does not yet cover the same model and feature set.

This PR replaces that env-only property with an Ascend-owned default in `vllm_ascend/mrv2_utils.py`, installed wherever `VllmConfig` is created or unpickled (frontend `NPUPlatform.check_and_update_config`, EngineCore via `vllm_ascend/__init__.py`, and worker init).

Default V2 when all of the following hold; otherwise V1:

- Model architecture whitelist: `Qwen3ForCausalLM` (generate, not hybrid, not attention-free).
- Feature whitelist: no speculative decoding, or `eagle3` / `mtp` / `dflash`.
- Platform is not 310P and Triton is available.

`VLLM_USE_V2_MODEL_RUNNER=0/1` still wins over the whitelist.

Related changes in the same patch file:

- Neutralize upstream `_validate_v2_model_runner`. GPU V2 feature/Triton checks must not reject an Ascend enablement decision (including an explicit env override).
- Keep the existing V2 unsupported-feature exceptions: drop `prefill context parallelism` on v0.28.0, and honor `resolve_spec_pp_support` for spec+PP.
- Keep the V1 unsupported-feature exceptions for `dspark speculative decoding` and `dflash2 drafts`, but install that patch with `hasattr` instead of `vllm_version_is("0.28.0")` so SHA / `__version__="dev"` CI checkouts do not patch a missing API.

Yes. `vllm serve` of `Qwen3ForCausalLM` (for example Qwen3 dense) now defaults to Model Runner V2 on A2/A3/A5 when Triton is present and speculative decoding is absent or `eagle3`/`mtp`/`dflash`. Models not on the architecture whitelist (DeepSeek, GLM, Qwen3-MoE, hybrid models, 310P) still default to V1. Explicit `VLLM_USE_V2_MODEL_RUNNER` is unchanged.

- Unit tests added/updated:
  - `tests/ut/test_mrv2_utils.py` (architecture / feature / env / 310P / Triton gates, validation no-op)
  - `tests/ut/patch/platform/test_patch_use_v2_model_runner.py` (property wiring, V1 feature exceptions, v0.28.0 PCP exception)
- `tests/e2e/pull_request/one_card/test_attention_v1_precision.py`: mock TP/DP groups because Qwen3 now defaults to V2 and `NPUPlatform.set_additional_forward_context` reads those groups.
- CI: cpu-ut / e2e on this PR.


vllm_ascend/xlite/xlite.py
```diff
@@ -661,7 +661,7 @@ def extract_kv_cache(self, kv_caches: list[tuple[torch.Tensor, ...]], /) -> list
         mla_caches: list[tuple[torch.Tensor, ...]] = []
 
         idx = 0
-        index_mask = self.xlite_config.index_full_mask or [True] * self.xlite_config.n_layers
+        index_mask = getattr(self.xlite_config, "index_full_mask", None) or [True] * self.xlite_config.n_layers
```
SijieFu, line 664: This seems unnecessary. Update `xlite` with `pip install xlite==0.2.0rc1` and the lint issue should be gone.

PR #16211, opened 2026-09-10
[Performance][Spec Decode] Avoid cloning unused raw target logits

zzy-6666:

This PR avoids an unnecessary FP32 target-logits clone in the vLLM Ascend
speculative decoding path.

`AscendRejectionSampler.forward()` preserves the original target logits
before applying logits processors because these processors may modify the
tensor in place. However, this copy is unnecessary when raw logprobs are not
requested and Ascend entropy verification is disabled.

Inspired by upstream vLLM PR
[https://github.com/vllm-project/vllm/pull/53489](https://github.com/vllm-project/vllm/pull/53489),
this change makes the clone conditional while accounting for the
Ascend-specific entropy verification path.

vLLM Ascend passes the original logits to `rejection_sample()` through
`ori_target_logits`. Therefore, the clone is preserved when:

- raw logprobs are requested; or
- Ascend entropy verification is enabled, regardless of the logprobs mode.

Otherwise, the redundant full-vocabulary tensor copy is skipped.
`max_num_logprobs=0` continues to be treated as a logprobs request, and
processed-logprobs behavior remains unchanged.

No.

This is an internal performance optimization. The API, generated results,
and configuration behavior remain unchanged. Original logits are still
preserved whenever they are required for raw logprobs or Ascend entropy
verification.

The clone path was instrumented and manually verified on 4 × Ascend 910B
with TP=4 before and after the change:

| Entropy verification | Logprobs requested | Before | After |
| --- | --- | --- | --- |
| Disabled | No | Clone executed | Clone skipped |
| Disabled | Yes | Clone executed | Clone executed |
| Enabled | No | Clone executed | Clone executed |
| Enabled | Yes | Clone executed | Clone executed |

A parameterized regression test,
`test_forward_clones_raw_target_logits_only_when_needed`, was also added to
`tests/ut/sample/test_rejection_sampler.py`. It covers:

- raw-logprobs mode with `max_num_logprobs=None`, `0`, and a positive value;
- processed-logprobs mode;
- entropy verification in both logprobs modes.

The test uses an in-place logits processor and verifies that processed logits
are updated while the original logits remain unchanged whenever required.
In particular, it ensures that entropy verification preserves the original
logits independently of `is_processed_logprobs_mode`.

Local checks:

- `ruff check`: PASS
- `ruff format --check`: PASS
- Python syntax compilation: PASS
- `git diff --check`: PASS

The regression test passed in the PR CPU UT environment.

### DIFF
--- a/tests/ut/sample/test_rejection_sampler.py
+++ b/tests/ut/sample/test_rejection_sampler.py
@@ -979,6 +979,102 @@ def _replace(obj, **kwargs):
     return SimpleNamespace(**data)
 
 
+@pytest.mark.parametrize(
+    (
+        "is_processed_logprobs_mode",
+        "max_num_logprobs",
+        "entropy_verify",
+        "expected_clone",
+    ),
+    [
+        pytest.param(False, None, False, False, id="raw-without-consumer"),
+        pytest.param(False, 5, False, True, id="raw-logprobs"),
+        pytest.param(False, 0, False, True, id="raw-zero-logprobs"),
+        pytest.param(False, None, True, True, id="raw-entropy-verify"),
+        pytest.param(True, 5, False, False, id="processed-logprobs"),
+        pytest.param(True, None, True, True, id="processed-entropy-verify"),
+    ],
+)
+def test_forward_clones_raw_target_logits_only_when_needed(
+    monkeypatch,
+    is_processed_logprobs_mode,
+    max_num_logprobs,
+    entropy_verify,
+    expected_clone,
+):
+    rejection_sampler = AscendRejectionSampler.__new__(AscendRejectionSampler)
+    rejection_sampler.sampler = MagicMock()
+    rejection_sampler.sampler.logprobs_mode = "processed_logprobs" if is_processed_logprobs_mode else "raw_logprobs"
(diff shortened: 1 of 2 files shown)


lijiahang226 (approved):

lijiahang226:
@slippersss Could you please check if there is anything else we should be concerned about before we move this toward merge?

wxsIcey (approved):

linfeng-yuan (approved):

czydyy:
@zzy-6666 Your PR affected the performance of the nightly test case Qwen3-32B-QuaRot. Under test [PR16653](https://github.com/vllm-project/vllm-ascend/pull/16653), there are verified results showing that performance recovered from 329.56 to 391.
It also affected the accuracy of the nightly test case Qwen3.5-397B-A17B-w8a8-mtp. Verification results are available under [PR16665](https://github.com/vllm-project/vllm-ascend/pull/16665), where the accuracy recovered from 73.33 to 90.
I will comment /revert to roll back your PR.

czydyy:
/revert
[Bot]: revert completed successfully. New PR created: https://github.com/vllm-project/vllm-ascend/pull/16678

PR #12808, opened 2026-07-24
[Performance][Model] Reduce DSV4 hidden-state copies

caozuoba:
### What this PR does / why we need it?

This PR eliminates redundant DeepSeek-V4 hidden-state copies by:

- Reusing mHC residual tensors because `hc_pre` and `hc_post` are out-of-place operators.
- Allocating and refreshing `_mtp_hidden_buffer` only when MTP is enabled, avoiding an unused TP all-gather and copy in non-MTP serving.

### Does this PR introduce _any_ user-facing change?

No. No API or configuration changes are introduced.

### How was this patch tested?

Tested with DeepSeek-V4-Flash W8A8 on Atlas A3, TP8 + EP:

| Metric | Main | PR | Improvement |
|---|---:|---:|---:|
| Output throughput | 302.82 tok/s | 323.76 tok/s | +6.91% |
| Mean TTFT | 1010.43 ms | 902.80 ms | -10.65% |
| P99 TTFT | 1988.60 ms | 1632.69 ms | -17.90% |
| Mean TPOT | 56.34 ms | 53.06 ms | -5.82% |
| Mean E2EL | 6588.14 ms | 6155.30 ms | -6.57% |
| P99 E2EL | 7378.16 ms | 6576.29 ms | -10.87% |

Accuracy result:

```text
dataset    version    metric    mode      vllm-api-general-chat
---------  ---------  --------  ------  -----------------------
gsm8k      271d0b     accuracy  gen                       96.97
- vLLM version: v0.25.1
- vLLM main: https://github.com/vllm-project/vllm/commit/fe784ff22e630a31fd798f392b01e0a75c18f047

PR #16230, opened 2026-09-10
[Performance][Model] Elide DeepSeek-V4 mHC residual clones

Ruiqiu-Zheng:
## What this PR does

This PR removes two redundant `hidden_states.clone()` calls when preserving the mHC residual in `DeepseekV4DecoderLayer.forward`:

```python
residual = hidden_states
```

The change is intentionally limited to the two residual-copy sites. It does not include MTP-buffer or other DeepSeek-V4 changes.

## Why

`hc_pre` and `hc_post` return separate outputs, so retaining the input tensor as the residual avoids two full hidden-state copies per decoder layer while preserving the residual value required by `hc_post`.

Historical measurements on the same DeepSeek-V4 mHC path showed:

- standalone NPU component: about 8.45% lower latency at T=1024 and 15.26% at T=4096, with the clone allocation removed;
- matched 4096-token full-model prefill: p50 TTFT 469.080 ms -> 448.763 ms (-4.33%), mean -3.82%.

These measurements are historical evidence for the mechanism; this PR does not claim broader production gains.

## Correctness / tests

Current-main focused regression coverage is included in this PR and passes:

- `pytest --noconftest tests/ut/models/test_deepseek_v4_mhc_residual.py`: **6 passed**
- `ruff check` on the two changed files: **passed**

The focused tests exercise both regular and sequence-parallel control flow, verify residual storage/value liveness across both attention and FFN mHC halves, and include a negative control where a mutating `hc_pre` view demonstrates why aliasing would be unsafe if the input were modified.

Separately, a TP8 eager NPU tensor-level attribution run on the same residual-alias mechanism covered all 8 ranks and 43 layers (1376 mHC-half observations) and observed:

- zero original-input value mutations after `hc_pre`;
- zero clone-vs-alias `hc_post` tensor divergences;
- `first_divergence = null`.

A current-main exact-commit TP8 rerun has not completed: the available historical test runtime requires reconstructing its Python/native packaging around the current source tree, and the bounded reconstruction attempts stopped before HC execution / TP8 model execution. Those failures are test-environment/harness failures, not observed candidate correctness failures. Current-main exact-commit TP8 remains unavailable in this local runtime reconstruction; the PR therefore keeps that limitation explicit and relies on the historical tensor-level NPU attribution plus current-main host/CI coverage. Maintainer-triggered selected hardware tests remain pending.

## Prior art

This is a current-main, clone-only, correctness-hardened rework of the residual-copy part of #12808, which was closed without merge. #12808 also contained separate MTP-buffer changes; those are deliberately excluded here.

### DIFF
--- a/tests/ut/models/test_deepseek_v4_mhc_residual.py
+++ b/tests/ut/models/test_deepseek_v4_mhc_residual.py
@@ -1,0 +1,188 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM Ascend project
+
+"""CPU storage-flow tests for DeepSeek-V4 mHC residual aliasing.
+
+Extract the real methods to avoid importing vLLM/NPU runtime dependencies.
+Can run independently with pytest --noconftest on CPU-only installations.
+"""
+
+import ast
+import copy
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+import torch
+
+MODEL_PATH = Path(__file__).resolve().parents[3] / "vllm_ascend/models/deepseek_v4/model.py"
+
+
+def _decoder_methods(clone_residual: bool = False) -> ast.Module:
+    tree = ast.parse(MODEL_PATH.read_text())
+    layer = next(n for n in tree.body if isinstance(n, ast.ClassDef) and n.name == "DeepseekV4DecoderLayer")
+    methods = copy.deepcopy(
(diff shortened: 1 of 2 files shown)


Ruiqiu-Zheng:
CI classification: pre-commit and CPU UT are passing. The current E2E gate is failing only because this PR touches source code while selected tests were skipped with no run-enabling label. Could a maintainer please add `ready-precise` to trigger the recommended selected tests? The PR is intentionally clone-only and keeps the current-main exact-commit TP8 limitation explicit in the description.

PR #16244, opened 2026-09-10
[CI] Speed up E2E dependency installation with uv wheel cache

zhangdiago:

Speed up PR gating CI by caching uv wheels across runs in the selected-tests workflow, restricted to runner classes whose storage can actually sustain the cache I/O.

#### Problem

1. **Dependency install dominates job time.** Every matrix shard runs a cold `uv pip install` (~7 min each). A uv wheel cache makes repeat runs hit cache and skip downloads.
2. **The save step can hang on self-hosted runners.** `runs-on/cache/save@v5` packs everything under `/root/.cache/uv`, including `builds-v0/` — uv's ephemeral build-isolation envs (GB-scale for packages like torch). Observed on the A2 runner: tar keeps grinding until the **2-hour job timeout** kills the whole job → shards report `cancelled` → `CI-gate` fails. (The ~100-min "save duration" in logs is the job timeout wall, not actual work.)
3. **The slow small-file I/O is not specific to A2.** On the amd64 CPU runner, `uv cache prune --ci` alone (pure unlink/metadata ops over the wheel cache) also exceeded 30 minutes. Any runner whose storage behaves this way will fail restore (unpack) and save (pack) the same way — the cache must be enabled per runner class, not just excluded per known-bad class.
4. **`hashFiles()` returned empty in the container job**, producing cache keys with no hash suffix (e.g. `uv-wheels-v1-9.1.0-910b-ubuntu22.04-py3.12-`), so the key never changed with requirements.

#### Fix — three layers

1. **Shrink the payload before save** (new `Prune uv cache for CI` step):
   - `uv cache prune --ci` (uv's official CI command) + remove `builds-v0/` explicitly.
   - Keeps `sdists-v9/` / `wheels-v0/` / `simple-v9/` (the reusable parts); drops ephemeral build dirs that are rebuilt on every install anyway.
   - Includes a `du -sh` debug line so the pruned cache size is visible in logs for verification.
2. **Skip save on cache hit** (`if: steps.uv-cache.outputs.cache-hit != 'true'`): steady-state runs (the common case) pay zero save cost; save runs only on cold cache.
3. **Bound the worst case** (`timeout-minutes: 10` + `continue-on-error: true`): even if save degrades again, it is capped at 10 min and cannot skip the test run or fail the gate — the same best-effort pattern already used by the upload steps in this workflow (`Upload timing data`, `Upload coverage data`, etc.).

**Runner allowlist**: the restore, prune, and save steps are gated on an allowlist of runner classes with proven storage — `a3` and `310p`. A2 (tar packing ground to the 2h job timeout; prune alone >30 min), CPU (prune alone >30 min), and A5 jobs stay on the cold-install baseline, identical to the pre-PR behavior and no longer at risk of the 2h-timeout cancellation. Runners can be added to the allowlist once their storage proves able to complete prune/save quickly.

Supporting fixes: the cache key hash is computed by an explicit `sha256sum` step (over `requirements-dev.txt`, `requirements.txt`, `pyproject.toml`) instead of `hashFiles()`, the cache path uses the absolute `/root/.cache/uv`, and the key includes `inputs.vllm` so cache invalidates across vLLM versions.

#### Expected behavior

| Scenario | Before | After |
|---|---|---|
| Cold cache (reqs changed, A3/310P) | ~7 min install, no cache written | ~7 min install + ≤3 min save (one-time) |
| Warm cache (typical PR push, A3/310P) | ~7 min install every run | ~2 min restore + ~2 min install; save skipped |
| A2 / CPU / A5 jobs | ~7 min install every run | unchanged: ~7 min cold install (cache skipped by design) |
| Save degraded on a runner | job hangs to 2h timeout → cancelled → gate red | capped at 10 min, warning only, tests still run |

No. CI-workflow-only change.

- CI on this PR (all matrix groups: A2 / A3 / 310P).
- The `du -sh` output in the prune step shows the pruned cache size; the save step duration in the run summary verifies it completes well within the 10-min cap.
- A2 and CPU jobs verify the three cache steps are skipped and the jobs complete on the cold-install baseline.

## Gate rerun-rate statistics (cache ROI justification)

> Sample: the 100 most recently merged PRs. Gate runs (the `E2E` / run-selected-tests workflow) were matched to each PR via its own commit SHAs, which excludes interference from same-name branches.

### Overview

| Metric | Value |
|---|---|
| Average gate runs per PR | **2.83** |
| Median | 2 |
| Passed on the first run | 37 / 100 (37%) |
| Required at least 2 runs | 63 / 100 (**63%**) |

### Distribution

| Gate runs | PRs | Share | |
|---:|---:|---:|---|
| 1 | 37 | 37% | █████████████████████ |
| 2 | 26 | 26% | █████████████ |
| 3 | 14 | 14% | ███████ |
| 4 | 7 | 7% | ████ |
| 5 | 5 | 5% | ███ |
| 6-9 | 9 | 9% | █████ |
| 16 | 1 | 1% | █ |
| 20 | 1 | 1% | █ |

### Cache ROI estimate

| Item | Value |
|---|---|
| Extra cost of the first run (cold install + save) | ~3 min |
| Saving per subsequent run with the same key (uv cache hit, A3/310P shards) | ~7 min |
| Break-even point | ~1.4 runs |
| **Expected net gain at the measured 2.83 runs per PR** | **~+10 min / PR** |

Conclusion: 63% of PRs need at least one gate rerun, so the cache optimization is a net win for the majority of PRs. PRs that pass on the first run pay a one-time cost of ~3 min.


zhangdiago:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

zhangdiago:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

zhangdiago:
Closing this PR: measured negative ROI.

Post-merge-style verification runs on A3/310P shards showed the uv wheel cache saves only ~3s on the install step (noise-level) while warm runs still pay ~20s for restore, and cold runs pay prune (~2.5 min) + save (capped at 10 min, most shards hit the cap). Root cause: the ~7 min install step is dominated by source builds (vllm-ascend editable compile, and the upstream vllm empty wheel build), not by dependency downloads — the in-cluster nginx pypi mirror already makes downloads fast, so the bottleneck this PR targeted does not exist on the allowlisted runners.

Thanks to reviewers for the pushback that led to the measurement. A follow-up targeting the real bottleneck (caching the built vllm empty wheel, keyed on the pinned vllm commit) is being prepared.

PR #16246, opened 2026-09-10
[Performance][GDN] Share group-invariant metadata across kv-cache groups

Wfd567:

**Background**

Hybrid models such as Qwen3.8 / Qwen4-Exp (e.g. Qwen3.8-27B) interleave
GDN (linear-attention) layers and full-attention layers at 3:1. The hybrid KV cache
manager aligns kv-cache group sizes to the least-numerous layer type, which splits
the GDN layers across **3 kv-cache groups**. As a result, `_build_attn_group_metadata`
invokes the GDN metadata builder's `build()` once per group on **every step**.

Within a single `_build_attention_metadata` call, every GDN group receives an
identical `common_attn_metadata` **except its own `block_table_tensor` /
`slot_mapping`**. However, the full metadata pipeline is executed 3x per step:

- spec sequence classification + fold (1 `.item()` host sync per group)
- batch statistics (4 `.item()` host syncs per group)
- stable argsort + `repeat_interleave` token reorder
- spec / non-spec `query_start_loc` cumsums and accepted-token reorder
- chunked-prefill chunk indices/offsets (6 CPU `prepare_*` calls + H2D per group)
- causal conv1d metadata and `has_initial_state`

**What this PR changes**

- Add `GDNGroupInvariantCache` to `gdn_attn_builder.py`.
- `AscendGDNAttentionMetadataBuilder.build()` accepts an optional
  `group_invariant_cache` argument.
- The model runner creates one cache before the kv-cache-group loop and passes it
  to every GDN group's build. The **first group (fingerprint miss) performs the
  full build and fills the cache**; subsequent groups **reuse the group-invariant
  intermediates and only recompute the state-index tensors derived from their own
  block table**.

Effect per step: host-device syncs drop from 3x5 to 5; the argsort,
chunked-prefill metadata and conv1d metadata drop from 3x to 1x.

**Safety properties**

- Only block-table-independent results are cached. `spec_state_indices_tensor`,
  `non_spec_state_indices_tensor` and `prefill_state_indices` are always
  recomputed from each group's own block table.
- Cached tensors are shared read-only; per-group full-graph padding still writes
  each builder's own persistent buffers, so captured graph addresses are unchanged.
- An O(1) shape-level fingerprint falls back to a full build on any mismatch.
- Callers that pass no cache (drafting, cudagraph capture, external callers)
  keep the original path unchanged.
- Guarded to non-ubatched calls (`ubid is None`): the shape-level fingerprint
  cannot distinguish same-shaped ubatches, so ubatched GDN builds fall back to
  full computation instead of silently reusing stale entries.

This mirrors the existing DSA precedent (`shared_dsa_draft_cache`).

### Benchmark

**GDN metadata `build()` per-step cost** (single forward):
- **Before**: all 3 kv-cache groups each run the full build, ~`2.1 ms/group`, total `6.3 ms`.
- **After**: the first group still builds fully (`2.1 ms`); the 2nd and 3rd groups
  reuse the cache (`1.2 ms` each), total `4.5 ms`.
- Saves **~1.8 ms / step (≈28.6%)** per forward. With speculative decoding
  (MTP, k=3) this is another forward per step, so the wall-clock benefit compounds.

**End-to-end benchmark** — Qwen3.6-35B-A3B, TP=2, 8k-token random input, 350 output tokens:

| Metric                | Before   | After    | Δ        |
|-----------------------|----------|----------|----------|
| Throughput (tokens/s) | 222.28   | 227.36   | +2.29%   |
| TTFT (ms)             | 1162.65  | 1156.24  | -0.55%   |
| TPOT (ms)             | 23.21    | 22.74    | -2.03%   |
| E2E Latency (ms)      | 9262.48  | 9094.04  | -1.82%   |

Throughput improves by **+2.29%**, and TTFT / TPOT / E2E latency all drop
(by roughly 0.6% ~ 2%). The win comes entirely from removing the redundant
host-device syncs (`.item()`), the repeated NPU argsort, and the duplicated
chunked-prefill / conv1d metadata across the 3 GDN groups.

**Accuracy regression check** — GPQA-diamond benchmark:

| Dataset       | Before | After |
|--------------|--------|-------|
| GPQA-diamond | 0.8622 | 0.868 |

Accuracy is preserved (0.8622 → 0.868), confirming the metadata reuse does not
degrade output quality.

No. This is a pure internal optimization of attention metadata construction.
No API, configuration or behavior change. The produced metadata is field-by-field
identical to the original full build (verified by UTs).

### DIFF
--- a/vllm_ascend/ops/gdn_attn_builder.py
+++ b/vllm_ascend/ops/gdn_attn_builder.py
@@ -125,6 +125,75 @@ class GDNSpecDecodeMetadata:
     actual_seq_lengths: torch.Tensor
 
 
+@dataclass
+class GDNGroupInvariantCache:
+    """Group-invariant build intermediates shared across KV cache groups.
+
+    Hybrid models (e.g. Qwen3-Next / Qwen4-Exp style GDN:attention = 3:1)
+    split their GDN layers across multiple kv cache groups, so the model
+    runner calls ``build`` once per group within a single
+    ``_build_attention_metadata`` call. Every group receives identical
+    ``common_attn_metadata`` except ``block_table_tensor`` / ``slot_mapping``,
+    which means most intermediates (spec classification, batch statistics,
+    token reordering, chunked-prefill metadata, conv1d metadata) are
+    recomputed identically per group, including several host-device
+    synchronizations (``.item()``) and an NPU argsort.
+
+    The model runner creates one cache before the kv-cache-group loop and
+    drops it after the loop, so entries are naturally isolated per step /
+    per draft substep / per dummy run and no explicit invalidation is
+    needed. The first group (fingerprint miss) computes and fills the cache;
+    later groups reuse the group-invariant intermediates and only redo the
+    block-table-derived parts.
+
+    Correctness contract: only results that never read the block table may
+    be stored here; anything derived from a group's own block table MUST
+    stay per-group. Cached tensors are shared read-only across groups:
+    consumers may index / copy-from them but must not mutate them in place.
+    """
+
+    # O(1) defensive fingerprint (shape-level scalars, no content compare).
+    fingerprint: tuple | None = None
+
+    # Spec sequence classification (post-fold; holds the only
+    # sum().item() sync of the classification block).
+    spec_sequence_masks_cpu: torch.Tensor | None = None
+    spec_sequence_masks: torch.Tensor | None = None
+    spec_sequence_indices: torch.Tensor | None = None
+    non_spec_sequence_indices: torch.Tensor | None = None
+    num_spec_decodes: int = 0
+
+    # Batch statistics (post-fixup final values):
+    # (num_decodes, num_prefills, num_decode_tokens, num_prefill_tokens,
+    #  num_spec_decode_tokens)
+    batch_stats: tuple[int, int, int, int, int] | None = None
+
+    # Token reorder (stable-partition result slices).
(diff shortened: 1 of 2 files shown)


vllm_ascend/ops/gdn_attn_builder.py
```diff
@@ -160,6 +161,76 @@ class GDNSpecDecodeMetadata:
...
+
+    # Prefill derived.
+    prefill_query_start_loc: torch.Tensor | None = None
+    prefill_query_start_loc_cpu: torch.Tensor | None = None
+    prefill_has_initial_state: torch.Tensor | None = None
+    chunked_prefill_metadata: GDNChunkedPrefillMetadata | None = None
+
+    # Causal conv1d metadata / initial state.
+    has_initial_state: torch.Tensor | None = None
+    nums_dict: dict[int, dict[str, object]] | None = None
```
realliujiaxu: `nums_dict`, `batch_ptr`, token_chunk_offset_ptr are never used or updated?

Wfd567: Those three fields (`nums_dict` ,`batch_ptr` ,`token_chunk_offset_ptr` ) were leftover placeholders from the planned conv1d-metadata caching component. They were never assigned nor read in`build()` , so they carried no runtime meaning. I've removed them from`GDNGroupInvariantCache` . Functionality is unaffected

Wfd567:
@zzzzwwjj please take a look

PR #16247, opened 2026-09-10
[Performance][v0.26.0rc] Switch MoE shared+routed combine to torch._foreach_add in AscendMoE

qigangc:
For the MoE model with shared experts, when superkernel is enabled, aclnnAdd is used to add shared_output and fused_output.
However, superkernel can be added to the graph only by using the torch._foreach_add operator. Otherwise, the graph will be disconnected.
No, this is an improvement change. To incresing add in MoE model(sk)
- Replace vllm-ascend/vllm_ascend/ops/fused_moe/fused_moe.py
- Enable static kernel option and superkernel option
- Run vllm bench


lijiahang226 (approved):
LGTM

yiz-liu (approved):

PR #16249, opened 2026-09-10
[BugFix][Worker] Select local communication endpoints by bound NPU

Yuli-yx:

The NPU visible-to-physical device mapping from #16256 is merged in upstream main as `9750add07ee64aa966ce2d03e86c8e070b46e9f6`. This PR builds on that merged implementation and contains only the endpoint-selection fix.

DP workers can share `local_rank=0` while binding different visible NPUs. The old `setup_ascend_local_comm_res()` indexed `ASCEND_RT_VISIBLE_DEVICES` with this rank, which could select the wrong `ub_endpoint_npu_<id>.json`. Containers exposing only part of a host's NPUs additionally require runtime-to-host conversion.

Pass the bound `visible_device_index` from the worker into the helper, then resolve its host physical ID through `current_platform.visible_device_id_to_physical_device_id()` merged by #16256. Select the endpoint file using that physical ID and propagate mapping errors. Calls without endpoint configuration still return without performing conversion.

Rebased onto upstream main `e2f7e3110` after #16256 merged. The PR now contains one endpoint-fix commit and four files: `vllm_ascend/utils.py`, `vllm_ascend/worker/worker.py`, `tests/ut/test_local_comm_res.py`, and `tests/ut/worker/test_worker_v1.py`. No prerequisite platform or native-binding commits remain in this PR.

Endpoint files configured through `ascend_local_comm_res_path` are selected by the bound NPU's host physical ID. This uses the native conversion binding already merged by #16256 and requires an extension built with a supporting CANN version.

- 12 focused unit tests passed for the combined stack, including endpoint selection, error propagation, platform delegation, and passing the bound visible ordinal instead of local rank.
- Built and installed the native extension with CANN 9.1.0 for Ascend950DT (`ascend950dt_9582`).
- All 13 independent process checks passed across non-privileged container mounts `0,1,2,3`, `2,3`, `1,3`, and `3`. Each used vLLM local rank 0 and checked device count, host physical ID, real endpoint-file selection, and invalid IDs. Visibility was unset or restricted to the last container logical device.
- The hardware and focused-UT results above describe the earlier combined stack. No model-serving or P/D transfer workload was run. For the latest rebase, the endpoint patch is unchanged (`git range-diff`), Python syntax checks, Ruff 0.14.0 lint/format checks on all four changed files, and `git diff --check` passed. The merged platform interface and worker device-binding order were reviewed. Hardware tests were not rerun on this rebased checkout; CI will validate the rewritten commit.

Runtime validation used vLLM `2cf0a6915ce544dc493a0990f2ea38d81601128a` and vLLM-Ascend `f8481287a07cd7c407a3142f5d4188d71d964467`, existing connector changes, and the combined patch (SHA256 `6509d3bdd0f3ab437e505329bd5fc01124d7010a1b73ed0d76c789c557005db6`). The installed torch_npu requires ascending visibility lists; descending lists fail before the binding.

### Additional partial-container validation (2026-09-11)

Tested on Ascend950DT with CANN 9.1.0, torch 2.10.0+cpu, and torch_npu 2.10.0.post4, using the previously validated image `vllm-ascend:npu-id-fixed-v2-20260910`. The non-privileged container mounted only host NPUs `3,4,5,6,7` and set `ASCEND_RT_VISIBLE_DEVICES=2`. Device count was 1. The test set vLLM's assigned device list to `[2]`, resolved local rank 0 through the platform binding interface, and called `torch.npu.set_device(0)`.

| Mapping / check | Before: environment-only mapping and old endpoint helper | After: CANN conversion and fixed endpoint helper |
| --- | --- | --- |
| Mounted host physical IDs | `3,4,5,6,7` | `3,4,5,6,7` |
| Container device numbering before visibility filtering | `0,1,2,3,4` | `0,1,2,3,4` |
| `ASCEND_RT_VISIBLE_DEVICES` | `2` | `2` |
| vLLM local rank / bound visible ordinal | `0` / `0` | `0` / `0` |
| Expected host physical ID | `5` | `5` |
| Returned physical ID | `2` (incorrect) | `5` (correct) |
| Endpoint file actually loaded | `ub_endpoint_npu_2.json` | `ub_endpoint_npu_5.json` |
| Correct-device / endpoint check | **FAIL: wrong device, despite successful file read** | **PASS** |

The before case executed the original `setup_ascend_local_comm_res()` extracted from commit `c607f1075`, and separately checked the inherited platform conversion. The after case executed the installed compiled binding and fixed helper. Both endpoint contents were compared against the real `/etc/hixlep` files, and the old and new contents differed. This verifies device mapping and endpoint selection, not model serving or P/D transfer. The temporary test container exited with code 0 and was removed.

Full test log: `/mnt/share/z00931220/npu-id-20260910/results-exact/npu-id-test-3-4-5-6-7-20260911.log`.

### DIFF
--- a/tests/ut/worker/test_worker_v1.py
+++ b/tests/ut/worker/test_worker_v1.py
@@ -15,7 +15,7 @@ from vllm.v1.kv_cache_interface import (
 
 from tests.ut.base import TestBase
 from vllm_ascend.device.hardware import AscendDeviceType
-from vllm_ascend.device.hardware_profile import get_hardware_profile
+from vllm_ascend.device.hardware_profile import HardwareCapability, get_hardware_profile
 from vllm_ascend.utils import vllm_version_is
 
 init_cached_hf_modules_path = "vllm.utils.import_utils.init_cached_hf_modules"
@@ -667,7 +667,11 @@ class TestNPUWorker(TestBase):
 
         # Setup mock
         mock_mem_get_info.return_value = (1000, 2000)
-        mock_get_device_type.return_value = get_hardware_profile(AscendDeviceType.A2)
+        profile = get_hardware_profile(AscendDeviceType.A2)
+        mock_get_device_type.return_value = MagicMock(wraps=profile)
+        mock_get_device_type.return_value.supports.side_effect = (
+            lambda capability: capability == HardwareCapability.LOCAL_KV_COMM_RESOURCE or profile.supports(capability)
+        )
 
         # Mock MemorySnapshot
         mock_snapshot = MagicMock()
@@ -676,7 +680,7 @@ class TestNPUWorker(TestBase):
         mock_snapshot_cls.return_value = mock_snapshot
 
         # Mock current_platform for v0.24.0 init_device path
-        mock_current_platform.logical_device_id_to_visible_device_id.return_value = 0
+        mock_current_platform.logical_device_id_to_visible_device_id.return_value = 1
         mock_current_platform.device_type = "npu"
 
         # Create worker mock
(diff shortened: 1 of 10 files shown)


tests/ut/worker/test_worker_v1.py
```diff
@@ -662,7 +666,7 @@ def test_init_device(
         mock_snapshot_cls.return_value = mock_snapshot
 
         # Mock current_platform for v0.24.0 init_device path
```
ZT-AIA: v0.24?

lijiahang226 (approved):
LGTM

tests/ut/worker/test_worker_v1.py
Yuli-yx: This comment was left from the v0.24.0 main2main adaptation, where this device-mapping path was introduced. It is not a v0.24-only path and is unrelated to a missing PTA API. The mock intentionally maps local_rank=0 to visible NPU ordinal 1, so the test verifies that local communication setup uses the actually bound device rather than local_rank. We will update the comment to clarify this.

ZT-AIA (approved):

PR #15818, opened 2026-09-05
[CI][MRV2] Add DeepSeek V4 MRV2 nightly cases

dragondream-chen:
### What this PR does / why we need it?

Add five DeepSeek V4 nightly configurations for Model Runner V2:

- Flash single-node
- Flash PD-prefix
- Pro DSpark PD-prefix
- Pro DSpark 1M-PD
- Pro MTP PD-prefix

The PR also prevents the Chinese Read the Docs build from resolving tools/bisect instead of Python standard-library bisect when tools/generate_zh_docs.py is executed directly. This fixes the deterministic ImportError reported by build 34403297.

This PR supersedes #15765. GitHub does not permit reopening that PR after its force-pushed head branch was closed. The MRV2 slot-mapping runtime fix remains separate in #15817.

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

- All five configurations passed their configured smoke, accuracy, and/or performance validation.
- Flash single-node GPQA: 88.89; three-run performance mean: 234.6171 requests/s.
- Flash PD-prefix AIME: 96.67.
- Pro DSpark PD-prefix AIME: 93.33; three-run performance mean: 471.8498 requests/s.
- Pro DSpark 1M-PD three-run performance mean: 16.9460 requests/s.
- Pro MTP PD-prefix AIME: 93.33; three-run performance mean: 471.1520 requests/s.
- Verified the Read the Docs failure mode with tools/ first on sys.path; after the fix, random imports the standard-library bisect.py module.
- git diff --check passes.

- vLLM main: https://github.com/vllm-project/vllm/commit/a97dacb7106ee49f39f3d1fc6ae1800ff724e01d

PR #16255, opened 2026-09-10
[CI][MRV2] Compare original and short DeepSeek V4 Pro nightly names

sugm0521:

Add paired original-name and short-name variants of the three DeepSeek V4 Pro MRV2 four-node nightly cases from #15818 to investigate name-related CI failures. All six entries use `size: 4`.

Each pair differs only in its matrix name, config filename/reference, and `test_name`. The model, deployment parameters, datasets, workload, and thresholds are identical. The restored long-name configs match the original configs byte for byte; the existing short-name configs are unchanged.

Flash MRV2 cases remain excluded. The `generate_zh_docs.py` standard-library `bisect` import fix carried by #15818 remains included.

This change adds no concurrency limit: selecting all six variants increases total resource demand. Controlled CI comparisons should account for concurrent resource availability.

No.

- Parsed the nightly matrix and MRV2 case YAML files.
- Verified each long-name config exactly matches the original commit and each short-name config is unchanged.
- Verified each pair is byte-identical after normalizing only `test_name`.
- Verified all six matrix entries resolve to their matching configs with `size: 4`; the matrix delta contains only the three added original-name entries.
- `git diff --cached --check` passed.
- `bash format.sh ci` could not run because local `pre-commit` is not installed.

NPU/CI execution has not been performed for this appended commit; this PR is for CI comparison.


czydyy:
/nightly DSV4-Pro-W4A8-1M-PD-MRV2 DSV4-Pro-W4A8-MTP-PC-PD-MRV2
DSV4-Pro-W4A8-PC-PD-MRV2
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/34470195245)

czydyy:
/nightly DSV4-Pro-W4A8-1M-PD-MRV2
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/34474652943)

czydyy:
/nightly DSV4-Pro-W4A8-PC-PD-MRV2
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/34478354936)

PR #16256, opened 2026-09-10
[Feature][Platform] Implement NPU visible-to-physical device mapping

Yuli-yx:

The inherited `visible_device_id_to_physical_device_id()` implementation indexes the device visibility environment variable, or returns the input unchanged when the variable is unset. On Ascend, neither operation resolves container device numbering to host physical NPU IDs. For example, host devices `3,4,5,6,7` are numbered `0,1,2,3,4` inside the tested container. With `ASCEND_RT_VISIBLE_DEVICES=2`, application-visible device `0` refers to host physical device `5`, but the inherited implementation returns `2`.

#### CANN API and device ID semantics

```cpp
aclError aclrtGetPhyDevIdByUserDevId(
    int32_t userDevId, int32_t *phyDevId);
```

- `userDevId`: application-visible ordinal after visibility filtering, as passed to `torch.npu.set_device()`.
- Logic ID in the table below: container device number before visibility filtering. It is distinct from vLLM's worker-local rank.
- `phyDevId`: host physical NPU ID returned by CANN.
- The function returns an ACL status code, checked by the binding before returning the physical ID.

The table's logic IDs describe the tested container enumeration and visibility selection; they are not separately measured return values from `aclrtGetLogicDevIdByUserDevId`.

`aclrtGetPhyDevIdByLogicDevId` takes a CANN logical ID rather than an application-visible ordinal. In our CANN 9.1.0 validation, composing `aclrtGetLogicDevIdByUserDevId` with `aclrtGetPhyDevIdByLogicDevId` returned `107001` under visibility filtering. The direct user-to-physical API passed the same cases, so this implementation uses it directly.

#### Fix

Override the existing `NPUPlatform.visible_device_id_to_physical_device_id()` interface and delegate to a scalar C++ binding in `torch_binding.cpp`. The platform method lazily loads the extension after device binding. The binding validates the input range, calls `aclrtGetPhyDevIdByUserDevId`, and checks both its status and the returned ID.

CMake detects API availability. Older CANN builds remain possible, while requesting conversion without API support raises an explicit error rather than guessing a device ID.

This PR also adds a standalone `examples/device_id_mapping_demo.py`, following the example pattern used by `device_print_demo.py`, and registers a schema-compatible Meta implementation for tracing and export paths.

The existing platform conversion interface now returns host physical NPU IDs correctly for partially mounted containers. Using it requires rebuilding the extension with CANN supporting the conversion API.

- A focused platform unit test covers delegation to the native binding.
- Built the native extension with CANN 9.1.0 for Ascend950DT (`ascend950dt_9582`). Runtime: torch 2.10.0+cpu and torch_npu 2.10.0.post4.
- The installed torch_npu requires ascending visibility lists; descending lists fail before this binding.

#### Device mapping validation

Loaded the complete source archive of PR commit `70d8c072396e507e21f6f8f8dd99ff3a8d65a1f6` into a clean container directory and asserted that the platform class was imported from that directory. Reused the previously compiled native extension only after byte-for-byte checks of `csrc/torch_binding.cpp` and `CMakeLists.txt` against the PR source, plus verification of the extension SHA256 (`79e55542fb0f0a4a5ca3a5ebf860a7b9420389203c8b948535d9de7c045a064a`). The runtime dependencies remain the previously documented CANN 9.1.0 image.

**27 independent process checks passed. Both before and after columns below are actual function return values:** each process bound its user device, called the inherited platform method, then called the PR platform override backed by the compiled CANN binding. Each process also checked device count and rejection of invalid input IDs.

| Combination | Mounted host physical IDs | Container numbering before visibility filtering | ASCEND_RT_VISIBLE_DEVICES | userDevId | Logic ID (container, before filtering) | Expected host physical ID | Physical ID before PR | Physical ID after PR |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1 | `0,1` | `0,1` | unset | `0,1` | `0,1` | `0,1` | `0,1` PASS | `0,1` PASS |
| 2 | `0,1` | `0,1` | `1` | `0` | `1` | `1` | `1` PASS | `1` PASS |
| 3 | `0,1` | `0,1` | `0` | `0` | `0` | `0` | `0` PASS | `0` PASS |
| 4 | `1,3,5,7` | `0,1,2,3` | unset | `0,1,2,3` | `0,1,2,3` | `1,3,5,7` | `0,1,2,3` FAIL | `1,3,5,7` PASS |
| 5 | `1,3,5,7` | `0,1,2,3` | `2` | `0` | `2` | `5` | `2` FAIL | `5` PASS |
| 6 | `1,3,5,7` | `0,1,2,3` | `0,2` | `0,1` | `0,2` | `1,5` | `0,2` FAIL | `1,5` PASS |
| 7 | `3,4,5,6,7` | `0,1,2,3,4` | unset | `0,1,2,3,4` | `0,1,2,3,4` | `3,4,5,6,7` | `0,1,2,3,4` FAIL | `3,4,5,6,7` PASS |
| 8 | `3,4,5,6,7` | `0,1,2,3,4` | `2` | `0` | `2` | `5` | `2` FAIL | `5` PASS |
| 9 | `3,4,5,6,7` | `0,1,2,3,4` | `0,2,4` | `0,1,2` | `0,2,4` | `3,5,7` | `0,2,4` FAIL | `3,5,7` PASS |
| 10 | `4,6` | `0,1` | unset | `0,1` | `0,1` | `4,6` | `0,1` FAIL | `4,6` PASS |
| 11 | `4,6` | `0,1` | `1` | `0` | `1` | `6` | `1` FAIL | `6` PASS |
| 12 | `4,6` | `0,1` | `0` | `0` | `0` | `4` | `0` FAIL | `4` PASS |
| 13 | `7` | `0` | unset | `0` | `0` | `7` | `0` FAIL | `7` PASS |
| 14 | `7` | `0` | `0` | `0` | `0` | `7` | `0` FAIL | `7` PASS |

Lists in a row correspond position by position. Logic ID denotes the selected pre-filter container ordinal, not a separately queried CANN logical-ID return value. The single-device `visibility=0` case was run twice as a repeated-process check. No model-serving workload was run.

Full logs and device mount manifests: `/mnt/share/z00931220/npu-id-20260910/results-pr16256/`.

#### Latest-head validation (`0a0cd0ead`, 2026-09-12)

- Rebuilt and installed the native extension successfully with the new Meta implementation.
- Confirmed `_C_ascend::get_physical_device_id` has a registered `Meta` kernel.
- Focused platform test: `1 passed, 86 deselected`.

The demo's NPU execution was not repeated in this run because all eight devices on the available host were occupied by existing vLLM workers. The 27 mapping checks above remain the runtime validation for the CANN binding.

#### A3 partial-device container validation (`0a0cd0ead`, 2026-09-14)

Validated on host 165 with a fresh container using the locally cached September 10 `quay.io/ascend/vllm-ascend:nightly-main-a3` image (digest `sha256:b4fde70ae7bf994587f785b2e8bb79724e91bc92a631ee3f38dccee357a024c3`). Runtime: Ascend A3 (`ascend910_9391`), CANN 9.1.0, Python 3.12.13, torch 2.10.0+cpu, torch_npu 2.10.0.post4, and image-provided vLLM commit `2cf0a6915ce544dc493a0990f2ea38d81601128a`.

Applied the complete image-source → PR-source diff. The resulting source tree matched PR HEAD `0a0cd0ead3dac8afa15ae0a9e048a436f29b9e57` exactly (tree `073c73449d748a29e3f0b62c39803e0291026439`), with no unstaged source differences. Native differences from the image were limited to this PR's `CMakeLists.txt`, `torch_binding.cpp`, and `torch_binding_meta.cpp`; unchanged ACLNN artifacts were reused. CMake detected `aclrtGetPhyDevIdByUserDevId` support. Incremental compilation and installation of the C++ extension completed successfully; every runtime check verified the loaded extension's path and SHA256 (`810fe715735b49223df2ef40a899686bc9daa8eea98098f636a0e7a8d978759a`).

**15/15 independent-process mapping checks passed.** Only host physical devices `8,10,12,14` were mounted, enumerated as container devices `0,1,2,3` before visibility filtering. Both before and after columns are actual function return values; lists correspond position by position.

| ASCEND_RT_VISIBLE_DEVICES | User device ID | Physical ID before PR | Physical ID after PR / expected |
| --- | --- | --- | --- |
| unset | `0,1,2,3` | `0,1,2,3` | `8,10,12,14` |
| `0` | `0` | `0` | `8` |
| `1` | `0` | `1` | `10` |
| `3` | `0` | `3` | `14` |
| `0,2` | `0,1` | `0,2` | `8,12` |
| `1,3` | `0,1` | `1,3` | `10,14` |
| `0,1,2,3` | `0,1,2,3` | `0,1,2,3` | `8,10,12,14` |

Each process also checked device count, rejection of `-1`, `2**31`, and `device_count` as invalid user IDs, and registration of the Meta kernel.

- The PR demo passed: `ASCEND_RT_VISIBLE_DEVICES=2 python examples/device_id_mapping_demo.py` reported user device `0` → host physical device `12`.
- Container unit test: `TORCH_DEVICE_BACKEND_AUTOLOAD=0 python -m pytest -q tests/ut/test_platform.py -k visible_device_id_to_physical_device_id` → `1 passed, 86 deselected`.
- Local CPU/mock preflight against the archived PR source, using the workspace's checked-in `.venv` through `uv run pytest`, also passed: `1 passed, 86 deselected`.

Build limitation: repeating the full CMake build encountered an existing AscendC preprocessing target's `ld.lld: unknown file type` error. Recompiling the two binding files with `cmake --build /validation/build --target vllm_ascend_C/fast -j 8`, followed by `cmake --install /validation/build`, succeeded with exit code 0. This validates the rebuilt PR extension; it does not establish that the full build is repeatable. No model-serving, performance, or accuracy workload was run.

Preserved container: `vaws-zhuyixiang-pr16256-20260914`. Full build logs, per-process logs, device mount manifest, source hashes, test script, and structured results: `/mnt/pr16256-165-20260914/evidence/` on host 165.

#### Explicit implementation registration (`3364ac71e`, 2026-09-14)

Aligned both the 310P and other-platform branches with `device_print`: the schema is declared with `ops.def(...)`, and the runtime implementation is registered separately with `ops.impl(..., c10::DispatchKey::CompositeExplicitAutograd, ...)`. This supports the integer-only call without tensor-based device dispatch. The existing Meta registrations are unchanged.

- Incremental native rebuild and installation passed on the same A3 container.
- Repeated the complete partial-device mapping matrix: **15/15 independent-process checks passed**, including explicit `CompositeExplicitAutograd` and Meta registration checks, invalid-ID rejection, import paths, and the loaded extension hash.
- Extension SHA256: `a0d426e40ef0e68d9cf3e1d610c0107ff64423e0e9f09136cc024006efad2680`.
- Local CPU/mock preflight through the workspace `.venv` using `uv run pytest`: `1 passed, 86 deselected`. Repository `bash format.sh ci` checks also passed.
- Follow-up build log, test script, and per-process results: `/mnt/pr16256-165-20260914/impl-results/` on host 165. The earlier tables remain the historical validation of `0a0cd0ead`; all 15 mapping results were reproduced with this new registration.


lijiahang226 (approved):

lijiahang226:
@ZT-AIA Could you please take a loot at this PR? It includes changes to torch_binding.

csrc/torch_binding.cpp
```diff
@@ -2722,6 +2742,7 @@ at::Tensor restore_tensor(uintptr_t ptr_val, const std::vector<int64_t>& shape,
 // Pybind on Ascend 310P
 TORCH_LIBRARY_EXPAND(CONCAT(_C, _ascend), ops)
 {
+    ops.def("get_physical_device_id(int user_device_id) -> int", &vllm_ascend::get_physical_device_id);
```
ZT-AIA, line 2745: Has it been verified on the 310P?

Yuli-yx: The CANN api `aclrtGetPhyDevIdByUserDevId` should support 310p or all other npu products according to ascend cann[ api doc](https://www.hiascend.com/document/detail/zh/CANNCommunityEdition/910/API/runtimeapi/aclcppdevg_03_2286.html), but it is not verified on these platforms yet due to lock of test devices.

ZT-AIA (commented):
Please also supplement the implementation here in torch_binding_meta.cpp.

tests/ut/test_platform.py
```diff
@@ -26,6 +26,28 @@
 )
 
 
+@pytest.mark.parametrize("failure", [None, "missing_op", "runtime_error"])
+def test_visible_device_id_to_physical_device_id(failure):
```
ZT-AIA: ut runs on the CPU. Is it unnecessary here?

vllm_ascend/platform.py
```diff
@@ -112,6 +112,20 @@ def pass_key(self) -> str:
     def manual_seed_all(cls, seed: int) -> None:
         pass
 
+    @classmethod
+    def visible_device_id_to_physical_device_id(cls, device_id: int) -> int:
```
ZT-AIA, line 116: Does this place actually not provide any protective function either?

ZT-AIA (commented):
Will this operator have a practical use case, or is it intended as a debugging tool? Could we consider writing a demo similar to `device_print` in the following path: `vllm-ascend\examples\device_print_demo.py`?

vllm_ascend/platform.py
Yuli-yx: The check provides a clear rebuild message when the extension lacks this operator. CANN availability and input validation are handled in C++. Do we need to provide a falling-back ?

tests/ut/test_platform.py
Yuli-yx: These tests mock the CANN call and verify Python delegation and error handling. Actual device mapping is tested separately on NPU hardware. We can remove it if u think this is necessary.

Yuli-yx:
> Please also supplement the implementation here in torch_binding_meta.cpp.

This is an integer-only runtime query used outside the compiled model, so there is no shape inference to perform. What scenario would require a Meta implementation here?

Yuli-yx:
@ZT-AIA We have addressed the three suggestions in `0a0cd0ead`:

1. Added `examples/device_id_mapping_demo.py`, following `device_print_demo.py`.
2. Added the Meta implementation and registered it for both 310P and other platforms.
3. Removed the redundant operator-existence check and simplified the related unit test.

The native extension rebuild, Meta registration check, pre-commit, and CPU UT have passed. Could you please take another look? Thanks.

tests/ut/test_platform.py
```diff
@@ -26,6 +26,18 @@
...
 
 
+def test_visible_device_id_to_physical_device_id():
+    with (
+        patch("vllm_ascend.platform.bootstrap_custom_op_env"),
+        patch("vllm_ascend.platform.import_module") as load_extension,
+        patch.object(torch.ops, "_C_ascend") as ops,
+    ):
+        ops.get_physical_device_id.return_value = 6
+        assert NPUPlatform.visible_device_id_to_physical_device_id(0) == 6
```
ZT-AIA, line 36: ut is not running on npu, and it may be necessary to add ut later.

ZT-AIA (dismissed):

Yuli-yx:
/rerun
[Bot]: rerun completed.

**Failed:**

ZT-AIA (approved):

PR #16258, opened 2026-09-10
[BugFix][MRV2] Recognize wrapped Mamba cache groups

Tame21:

MRV2 can fail on the first real request with
`AssertionError: no mamba layers in the model` when Ascend's KV cache
grouping wraps Mamba specs in `UniformTypeKVCacheSpecs`. The inherited
lookup only checks the outer spec type, so it misses the recurrent states
inside these groups.

Use the shared Ascend Mamba group resolver in
`AscendMambaHybridModelState._get_mamba_group_info()`. This recognizes
both plain and wrapped Mamba specs while preserving their original
group indices and caching the lookup result.

Add regression coverage for plain, wrapped, mixed, missing, and
inconsistent cache groups.

Yes. MRV2 can process requests with wrapped Mamba cache groups without
crashing during align-state preprocessing, including the affected
Qwen3.6 + Eagle3 configuration.

- Five isolated regression cases passed against both vLLM v0.28.0 and
  the current local vLLM source. These checks exercised the cache-spec,
  group-resolution, and preprocessing code with hardware dependencies
  stubbed.
- Before the fix, the wrapped-group case reproduced the reported
  assertion, and the mixed-group case exposed missing group indices.
  Both passed after the fix.
- Ruff lint and formatting checks, and `git diff --check`, passed.

### DIFF
--- a/tests/ut/worker/test_model_runner_v2_mamba.py
+++ b/tests/ut/worker/test_model_runner_v2_mamba.py
@@ -1,55 +1,19 @@
-import ast
 from dataclasses import replace
-from pathlib import Path
-from types import SimpleNamespace
-from unittest.mock import MagicMock, patch
+from unittest.mock import patch
 
-import numpy as np
 import pytest
 import torch
-from vllm.config.compilation import CUDAGraphMode
 from vllm.v1.kv_cache_interface import (
     FullAttentionSpec,
     KVCacheConfig,
     KVCacheGroupSpec,
-    KVCacheTensor,
     MambaSpec,
     UniformTypeKVCacheSpecs,
 )
-from vllm.v1.worker.gpu.model_states.mamba_hybrid import MambaHybridModelState
 
-from vllm_ascend.core.kv_cache_interface import AscendMLAAttentionSpec
-from vllm_ascend.utils import vllm_version_is
-from vllm_ascend.worker.v2.attn_utils import (
-    _allocate_kv_cache,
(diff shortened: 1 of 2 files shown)


tests/ut/worker/test_model_runner_v2_mamba.py
```diff
@@ -94,6 +97,68 @@ def test_mamba_model_state_inherits_upstream_state_management():
     assert AscendMambaHybridModelState.postprocess_state is MambaHybridModelState.postprocess_state
 
 
+@pytest.mark.parametrize("wrapped_group_ids", [(), (1, 3), (1,)], ids=["plain", "uniform", "mixed"])
```
drslark, line 27: Remove this ut.

drslark (changes requested):

drslark:
Add an e2e.

vllm_ascend/worker/v2/model_states/mamba_hybrid.py
```diff
@@ -41,6 +41,28 @@ class AscendMambaHybridModelState(MambaHybridModelState, AscendModelState):
     base so cooperative ``super()`` calls retain the Ascend model-state MRO.
     """
 
+    def _get_mamba_group_info(self, kv_cache_config: KVCacheConfig) -> tuple[list[int], MambaSpec]:
```
drslark, line 44: It is correct.

PR #16259, opened 2026-09-10
[BugFix][Model] Skip unused hash-router bias when loading DeepSeek-V4 weights

yiminghub2024:

Serving a native DeepSeek-V4 checkpoint whose config sets `num_hash_layers > 0` fails during weight loading:

```
KeyError: 'model.layers.0.mlp.gate.e_score_correction_bias'
  File "vllm_ascend/models/deepseek_v4/model.py", line 1319, in load_weights
    param = params_dict[name]
```

`DeepseekV4MoE` gives hash-router layers a `tid2eid` lookup table and deliberately leaves `e_score_correction_bias` unset, because those layers route text tokens purely through the table rather than through router scores:

```python
if self.hash:
    self.gate.tid2eid = nn.Parameter(...)
```

The checkpoints, however, still ship a `gate.bias` tensor for every MoE layer. `load_weights` renames it to `gate.e_score_correction_bias` and then indexes `params_dict` unconditionally, so the unused hash-layer bias aborts startup.

This PR skips the tensor when the model does not own the target parameter. The existing `gate.bias_vl` guard directly above is unaffected.

Reproduced on `deepseek-ai/DeepSeek-V4-Flash-Vision-Exp` (`num_hash_layers: 3`, `n_routed_experts: 256`) on Ascend 950PR. The checkpoint carries all three tensors for the hash layers:

```
layers.0.ffn.gate.tid2eid
layers.0.ffn.gate.bias
layers.0.ffn.gate.bias_vl
layers.0.ffn.gate.weight
```

No new flags or interfaces. Native DeepSeek-V4 checkpoints with hash-router layers now finish loading instead of raising `KeyError` at startup.

Unit test added: `tests/ut/models/test_deepseek_v4_moe.py::test_hash_layer_router_bias_is_skipped_when_unused`. It feeds a router bias for one hash layer and one dense layer and asserts only the dense one is consumed, so the previous `KeyError` is a regression failure.

```bash
pytest -sv tests/ut/models/test_deepseek_v4_moe.py
```

Manual check: `vllm serve` of `DeepSeek-V4-Flash-Vision-Exp` on one Ascend 950PR node reaches `Application startup complete` and answers `/v1/completions` requests, where it previously failed in `WorkerProc` init on every rank.


yiminghub2024:
pre-commit, cpu-ut, and DCO are green. `ci-gate` is blocked because this PR touches source and has no run-enabling label.

Could a maintainer add `ready-precise` (preferred) so the selected tests and ci-gate can run? I do not have permission to add that label.

yiminghub2024:
A3 `prepare-csrc-cache` failed after ~93m in `Install vllm-ascend` ([job](https://github.com/vllm-project/vllm-ascend/actions/runs/34477067641/job/102875397420)). This is the cache-builder CMake probe, not the DeepSeek-V4 weight-load change:

```
CMake Error at cmake/utils.cmake:15: Failed to locate torch path
RuntimeError: Found 0 concrete subclasses of DriverBase in triton.backends.ascend
RuntimeError: Failed to load the backend extension: torch_npu
```

Same A3 cache job on #15885 succeeded, so this looks like a runner/triton-ascend import flake (`npu-smi` is also missing on the CPU cache builder). Selected e2e is still running; I will `/rerun` the failed cache job once this E2E run finishes so we do not cancel in-progress device tests.

yiminghub2024:
/rerun

wangxiyuan (approved):

PR #16287, opened 2026-09-10
[BugFix] Fix KV pool NPU device binding and avoid Memcache scheduler contexts

zouyida2052:

KV pool clients can create NPU contexts on the wrong device when a worker's process-group local rank differs from its bound visible device, such as DP=2 / TP=4. Memcache's metadata-only scheduler client can also create an unnecessary EngineCore NPU context.

This combines #15968 (head `393ba9455ace8b4cd3aa6182088460fa5e213472`, by @Pz1116) with the additional Memcache scheduler-context fix into one commit, rebased onto main at `c5055c808`:

- Capture the owning worker's current device for Memcache and Mooncake initialization and transfer threads.
- Resolve scheduler device mapping from assigned physical IDs. Memcache passes the resolved device explicitly and skips `set_device()` when `init_bm=False`; Mooncake retains explicit scheduler binding.
- Preserve main's QoS configuration and batch-result validation, and update the affected constructor test.
- Add regression coverage for device mapping, backend setup, and the complete Memcache scheduler factory path.

Original author Pz1116 is credited with a `Co-authored-by` commit trailer. The original PR is left unchanged.

Fixes unintended NPU contexts in KV pooling without adding configuration flags. The internal MemcacheBackend constructor argument changes from `local_rank` to `device_id`.

The no-context scheduler path uses `assigned_physical_gpu_ids`; without assigned IDs, the existing `torch.npu.current_device()` fallback remains.

On the final submitted source tree, in an isolated Linux ARM64 container with a dedicated Python environment and the test suite's mocked heavy dependencies:

```bash
python -m pytest -q --confcutdir=tests/ut/distributed/ascend_store \
  tests/ut/distributed/ascend_store/test_backend.py
```

**102 passed.** The `--confcutdir` option isolates the self-contained backend tests from the parent conftest that requires an installed vLLM runtime. All 3,858 source-file SHA256 hashes matched the local staged source snapshot before testing.

Also passed: `git diff --check`, Ruff 0.14.0 lint and format checks for all five changed files, codespell, forbidden-import checks, boolean-context-manager checks, package `__init__.py` checks, and symbolic-meta checks.

Full `bash format.sh ci` was attempted. The archived script first required CRLF normalization for Bash; the equivalent normalized script then stalled fetching the ruff-pre-commit hook repository from GitHub and was terminated. The complete pre-commit suite is therefore **not verified** locally and remains for CI.

Prior hardware evidence (2026-09-08) applies specifically to #15968 at `393ba945` plus the original scheduler-context patch, before this rebase: A3 DP=2 / TP=4 startup completed, all eight workers appeared only on their assigned physical NPUs 8-15, four duplicate 113 MiB worker contexts disappeared, and neither EngineCore PID appeared in stable NPU process samples. This historical run was startup/context validation only; no inference requests or benchmarks were run. Hardware startup was not repeated on the rebased tree.

### DIFF
--- a/tests/ut/distributed/ascend_store/_mock_deps.py
+++ b/tests/ut/distributed/ascend_store/_mock_deps.py
@@ -56,6 +56,10 @@ if "torch" not in sys.modules and importlib.util.find_spec("torch") is None:
     sys.modules["torch"] = _torch
     sys.modules["torch.distributed"] = _torch.distributed  # type: ignore[attr-defined]
 
+_torch_module = sys.modules.get("torch") or importlib.import_module("torch")
+if not hasattr(_torch_module, "npu"):
+    _torch_module.npu = MagicMock()  # type: ignore[attr-defined]
+
 if "torch_npu" not in sys.modules:
     sys.modules["torch_npu"] = MagicMock()
     sys.modules["torch_npu._inductor"] = MagicMock()
@@ -84,6 +88,7 @@ _vllm_mock_modules = [
     "vllm.model_executor.layers.linear",
     "vllm.model_executor.layers.quantization",
     "vllm.platforms",
+    "vllm.platforms.interface",
     "vllm.utils",
     "vllm.utils.hashing",
     "vllm.utils.math_utils",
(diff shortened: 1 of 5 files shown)


Pz1116 (approved):

lijiahang226 (approved):
LGTM

weiguihua2 (approved):

Issue #15665, opened 2026-09-03
[RFC]: GLM-5.3-Flash Adaptation Roadmap for vLLM Ascend


lijiahang226:
# Motivation.

GLM-5.3-Flash combines KDA linear attention, NoPE sparse-MLA, MoE, multimodal
input, and an MTP draft layer, with a maximum context length of 1,048,576
tokens (see the [upstream vLLM recipe][vllm-recipe]).

The hybrid architecture exposes framework gaps in indexer integration, cache
ownership, prefix caching, graph execution, parallelism, and quantization.

Merged PR [#15127][pr-15127] provides the initial native-FP8 baseline on
Ascend 950 (TP=8): KDA, dense NoPE MLA prefill/decode/KV write, basic MTP
execution, ACLGraph, and the model-loading and compatibility fixes required by
the native-FP8 checkpoint. Validated on eight Ascend 950PR devices up to
131,072 tokens.

The baseline does **not** include a working kpool sparse indexer. The PR
review confirms three gaps:

- the MLA wrapper does not invoke the indexer through a usable path
  ([discussion][indexer-review-comment]);
- the kpool scoring path has no NPU implementation
  ([discussion][indexer-discussion]); and
- the sparse-attention operator does not support NoPE `rope_dim=0`
  ([discussion][nope-discussion]).

# Proposed Change.

This is a living roadmap. Implementation status, owners, and linked PRs may be
updated continuously. Material changes to the agreed framework design will be
flagged explicitly and require renewed review.

## Goals

- Generalize sparse-MLA so lightning and kpool indexers share a common interface.
- Connect prefix caching with pooling/indexer state.
- Support graph-safe asynchronous scheduling.
- Support DP, TP, EP, SP, PP, KVPP, and PCP.
- Add W8A8C8, W4A8C8, and W4A4C8 coverage across applicable INT, FP, and MXFP
  paths.
- Reuse existing vLLM Ascend operators wherever their contracts match.
- Keep existing models backward compatible.

## Non-goals

- Finalize every operator or parallelism design in this RFC.
- Treat private prototypes as publicly supported capabilities.
- Guarantee every hardware or quantization combination in the first milestone.
- Track unrelated GLM-5.x features.

## Status conventions

| Status | Meaning |
|---|---|
| **Merged** | Merged and documented in a public target branch. |
| **Public WIP** | Available in an open public PR. |
| **Planned** | Roadmap work without a public implementation. |
[...]

PR #16291, opened 2026-09-10
[BugFix][Spec Decode] Preserve prefill tails during speculative decoding

lijiahang226:

Refs #15665

Keep unfinished prompt chunks on the GDN prefill path even when their query length matches the speculative decode width. Remove the conversion to speculative metadata and its forced all-token acceptance, which can incorrectly commit convolution and recurrent state.

This is an independent GDN metadata fix. It contains no prerequisite commits and does not change the runner's graph dispatcher.

Yes. Short prefill tails retain their prefill state-update semantics when speculative decoding is enabled.

- Updated `tests/ut/ops/test_gdn_attn_builder.py` to cover first and stateful speculative-width prompt chunks, asserting that both remain prefill with no speculative acceptance metadata.
- GitHub pre-commit and CPU UT passed on the current head `7352e40f9` (4163 passed, 38 skipped). Device CI awaits a maintainer test label.

### DIFF
--- a/tests/ut/ops/test_gdn_attn_builder.py
+++ b/tests/ut/ops/test_gdn_attn_builder.py
@@ -814,17 +814,15 @@ def test_one_token_prefill_selection_respects_recurrent_state(
 
 
 @pytest.mark.parametrize(
-    ("seq_len", "expected_spec_decodes", "expected_prefills"),
+    "seq_len",
     [
-        pytest.param(4, 0, 1, id="first_chunk_stays_prefill"),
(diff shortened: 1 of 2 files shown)


lijiahang226:
Closing this one deliberately as part of cleaning up my open PR backlog. It has been sitting with merge conflicts and without a maintainer test label for ~10 days, and I no longer plan to land it in this form.

No decision is implied about the underlying bug; if the prefill-tail issue needs fixing later, I will re-open with a rebased branch. The head commit `7352e40f9` stays on my fork branch `glm53-v5-mtp-prefill` for reference.

PR #16303, opened 2026-09-11
[BugFix][Attention] Avoid A2 head-256 paged FIA non-finite output

Big2Wheel:

This PR is the Dense FIA correctness fix split out of #16013 so that the segmented Hybrid KV Cache COW fix and this operator-path workaround can be reviewed and merged independently.

On Ascend 910B3/A2, paged FIA can return non-finite values for a continuation-prefill path with:

- attention head size 256;
- `AscendAttentionState.ChunkedPrefill`;
- an existing paged KV cache represented by a non-null block table.

The failure was reproduced with a Qwen3.6-27B ModelSlim `W8A8_DYNAMIC` checkpoint, but the implementation does not inspect `quant_description`: that field is quantizer-specific metadata rather than a reliable description of the failing FIA execution path.

The problem is not dependent on prefix caching or `prefix_match_unit`. It can be reproduced with prefix caching disabled when a prompt is split by `--max-num-batched-tokens`: the later chunk becomes continuation prefill and reads historical KV through the paged FIA path.

The numerical probe isolated the failure to that FIA invocation. Q/K/V were finite, paged cache contents were finite, and the same effective tensors produced finite output through dense-TND FIA. This PR therefore gathers only the valid paged KV tokens into dense TND and calls FIA without a block table for the affected path.

The fallback is deliberately gated twice:

1. At backend initialization: A2 hardware capability and head size 256.
2. At runtime: `ChunkedPrefill` with a non-null block table.

All other hardware, head sizes, decode calls, and prefill-without-cache calls retain the existing path. No new environment variable or public option is introduced.

The existing paged-to-dense helper is exposed as `get_dense_prefill_kv` because it now serves both the large-head fallback and this attention backend path. Its internal `_gather_paged_kv_to_dense` implementation remains private. The gather supports both NHD and the newly added HND/BNSD KV cache layouts.

Yes. Affected A2 head-size-256 requests no longer produce non-finite attention output and degenerate text during continuation prefill.

The fallback adds a paged-KV gather and temporary dense KV tensors only on affected continuation-prefill calls. It can increase TTFT and temporary HBM use for those requests, but it does not change decode performance or unaffected configurations.

#### Unit coverage

```bash
python -m pytest -q \
  tests/ut/attention/test_attention_v1.py \
  tests/ut/device/test_hardware_profile.py
```

The tests verify that the head-256/chunked-prefill path gathers dense KV, removes the block table before FIA, passes corrected cumulative KV lengths, supports both NHD and HND/BNSD cache layouts, and enables the workaround capability only for A2.

The same core implementation and unit tests previously passed the complete #16013 CI matrix on A2, A3, and 310P. This independent PR is rebased onto the latest `origin/main`, and its CI validates the updated five-file change set.

Local `py_compile` and `git diff --check` pass. The current macOS checkout does not contain the NPU pytest dependencies, so hardware results below are from the Ascend development container.

#### Real-weight A2 reproduction

Hardware: 2 x Ascend 910B3 (A2), TP=2  
Checkpoint: Qwen3.6-27B ModelSlim W8A8_DYNAMIC

With prefix caching disabled and `--max-num-batched-tokens 1296`, an exact 1309-token prompt forces continuation prefill independently of fine-grained prefix matching.

| Build | Serial result | Concurrent result |
| --- | --- | --- |
| Before | 3/3 returned degenerate `!!!!!!!!` output | 1/8 returned degenerate output |
| After | 5/5 identical normal output | 8/8 identical normal output |

Additional probes confirmed that the paged cache was finite and bitwise stable while paged FIA produced non-finite output; dense-TND FIA over the same effective KV remained finite.

After the fix, warm serial smoke-test latency was approximately 1.59-1.72 s and the eight-request concurrent run was approximately 1.88-2.10 s per request. These are correctness smoke measurements rather than a formal performance benchmark.

### DIFF
--- a/vllm_ascend/attention/attention_v1.py
+++ b/vllm_ascend/attention/attention_v1.py
@@ -60,6 +60,7 @@ from vllm_ascend.compilation.acl_graph import (
     update_draft_graph_params_workspaces,
     update_graph_params_workspaces,
 )
+from vllm_ascend.device import utils as device_utils
 from vllm_ascend.device.device_op import DeviceOperator
 from vllm_ascend.device.hardware_profile import HardwareCapability, get_current_hardware_profile
 from vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.attention_fence import record_attention_compute_start
@@ -506,6 +507,13 @@ class AscendAttentionBackendImpl(AttentionImpl):
         self._use_layer_aware_fia_graph_replay = needs_layer_aware_fia_graph_replay()
         self._use_max_workspace_for_fia_graph = self._use_layer_aware_fia_graph_replay
         self.sinks = sinks
+        quant_description = getattr(self.vllm_config.quant_config, "quant_description", None)
+        self._use_dense_kv_for_w8a8_head_256_prefill = (
+            self.head_size == device_utils.FIA_TND_W8A8_PAGED_PREFILL_FALLBACK_HEAD_SIZE
+            and isinstance(quant_description, dict)
+            and "W8A8_DYNAMIC" in quant_description.values()
+            and get_current_hardware_profile().supports(HardwareCapability.FIA_HEAD_256_PAGED_PREFILL_WORKAROUND)
+        )
         self.layerIndex = 0
         # Some mixed-attention models cannot rely on the iteration order of
         # attn_metadata during graph replay. Record the captured layer name only
@@ -1304,9 +1312,10 @@ class AscendAttentionBackendImpl(AttentionImpl):
                 ):
                     self.key_cache, self.value_cache = kv_cache[0], kv_cache[1]
 
-            if self.key_cache is None:
+            if self.key_cache is None or self.value_cache is None:
                 raise RuntimeError(
-                    f"key_cache is None in _get_fia_params for mode {attn_metadata.attn_state}. kv_cache={kv_cache}"
+                    f"KV cache is not fully initialized for mode {attn_metadata.attn_state}: "
+                    f"key_cache={self.key_cache is not None}, value_cache={self.value_cache is not None}"
                 )
 
         if attn_metadata.attn_state == AscendAttentionState.PrefillNoCache:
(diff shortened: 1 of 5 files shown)


vllm_ascend/attention/attention_v1.py
```diff
@@ -1464,6 +1472,27 @@ def forward_fused_infer_attention(
...
                     )
+                if (
+                    self._use_dense_kv_for_w8a8_head_256_prefill
+                    and attn_metadata.attn_state == AscendAttentionState.ChunkedPrefill
+                    and block_table is not None
+                ):
+                    # A2 FIA can return non-finite values for a short continuation
+                    # prefill with ModelSlim W8A8_DYNAMIC and 256-wide heads. The
+                    # dense TND path is numerically stable for the same Q/K/V.
+                    assert self.key_cache is not None and self.value_cache is not None
```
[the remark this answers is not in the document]
Big2Wheel: Good point. Fixed in 4e24897d8. I removed the runtime assert and moved the explicit validation into `_get_fia_params`, before any cached FIA path dereferences the cache. The common guard now checks both `key_cache` and `value_cache`, raises a clear `RuntimeError`, and reports only initialization booleans instead of formatting the potentially large cache tensors. I also added a regression test for an incomplete KV cache.

vllm_ascend/attention/attention_v1.py
```diff
@@ -506,6 +507,13 @@ def __init__(
         self._use_layer_aware_fia_graph_replay = needs_layer_aware_fia_graph_replay()
         self._use_max_workspace_for_fia_graph = self._use_layer_aware_fia_graph_replay
         self.sinks = sinks
+        quant_description = getattr(self.vllm_config.quant_config, "quant_description", None)
```
lijiahang226: Please avoid relying on quant_description here.

Big2Wheel: Agreed. I removed the direct `quant_description` dependency. The workaround is now gated by the A2 hardware capability and head size at initialization, then by `ChunkedPrefill` plus a non-null block table at runtime. This identifies the affected FIA execution path without depending on ModelSlim-specific metadata. While rebasing onto the latest main, I also made the paged-to-dense gather honor the newly added HND/BNSD KV cache layout and added NHD/HND unit coverage.

PR #16216, opened 2026-09-10
[Misc][Main2Main] Main2Main 0911

zhao-stack:
### What this PR does / why we need it?

Advances the verified vLLM main adaptation point from September 9 to the frozen **September 11, 2026, 09:00 HKT** cutoff, retaining v0.28.0 support.

- Old vLLM: `a97dacb7106ee49f39f3d1fc6ae1800ff724e01d`.
- New vLLM: `84030bbe3d74d99bad477a3d2e37a973ccd8865c`, committed 2026-09-11 08:40:10 HKT, latest available at the fixed 09:00 HKT cutoff.
- Release unchanged: **v0.28.0 / `2cf0a6915ce544dc493a0990f2ea38d81601128a`**.
- Dependency #16009 merged on 2026-09-12 at 10:38:14 HKT as `15b0cd4fa8cd62bb72120cda67319185aaf6f521`.
- Rebased onto verified upstream main `3b7cf0fad5f37fb9e55d2622143819081a91e91a`. Current PR head: `18cc17834a727e24cd7a5e68c0019fae53171161`.
- [Full PR diff](https://github.com/vllm-project/vllm-ascend/compare/3b7cf0fad5f37fb9e55d2622143819081a91e91a...18cc17834a727e24cd7a5e68c0019fae53171161): **20 files, +106/-93**, now entirely the 0909-to-0911 increment. Parent commits are part of upstream main and no longer appear in this PR's commit stack.

Only three signed-off upgrade commits remain:

| Commit | Increment |
|---|---|
| `dcd0cbe956e81826646ffe6e47e9505fb8a0dd27` | Pin, scheduler block-state, RoPE, offload and DCP adaptations, with existing dual-version UT adjustments |
| `11fcbc61e14141b1ef3cf6f0cc9b56bfbba0ddbd` | Existing compressed-prefix-cache UT call contract |
| `18cc17834a727e24cd7a5e68c0019fae53171161` | Encoder graph axis-key call compatibility and existing UT |

No workflow changes, release-marker changes, new test functions, added skips, golden/threshold changes or port changes. The parent DSpark guard restoration, slot-mapping UT and subsequent upstream fixes are inherited from main. Existing upstream exclusions, including the parent's release extraction skip, are baseline content and are neither added by this PR nor counted as passing tests.

[...]

PR #16009, opened 2026-09-08
[Misc][Main2Main] Main2main 0909

zhao-stack:
#### Summary

Upgrade the verified vLLM main revision while retaining v0.28.0 compatibility.

| Input | Exact revision |
|---|---|
| Previous vLLM main | `b2f685834a6456197e7033966fdef52a23f1abcd` |
| Target vLLM main | `a97dacb7106ee49f39f3d1fc6ae1800ff724e01d` (September 9 target; unchanged by this revision) |
| Retained release | `v0.28.0` / `2cf0a6915ce544dc493a0990f2ea38d81601128a` |
| Current Ascend PR base | `1b1709b9227ad3e0db2d82b9231c4962919b86f5` |
| Reviewed PR head | `225f7cfdbf1c479798643a7d70a7e2031e5466f2` |

Current PR-wide diff: **61 files, +819/-418**. The numbered inventory below matches the current GitHub changed-file list one-to-one. Removed changes and historical commit-by-commit progress are not part of this inventory.

#### Scope and version handling

- Adapt the exact upstream contracts linked below. Where main and release differ, select the lane with `vllm_version_is("0.28.0")`; share logic where both contracts match.
- The owner also included baseline dual-version compatibility gaps. In particular, #51031's KV/kernel block-size split predates this main upgrade. Newly merged Ascend consumers from [#14068][ascmoon], [#14699][ascextract] and [#15969][ascpcp] require compatibility handling; they are not mislabeled as newly introduced vLLM breaks.
- **No workflow changes remain in this PR.** CPU UT uses the inherited single-main configuration. The existing `main2main` label continues to select main/tag NPU validation.
- Keep existing tests and necessary test adaptations; do not add guard-test files or cases. On release only, skip the three unsupported MRV2 extract-hidden-states cases at the owner's request, rather than asserting an initialization error. V1 on both versions and MRV2 on main remain enabled.
- The existing owner-authorized release SFA PCP precision exclusion remains a temporary exception, not a precision fix or passing result. The rebased Ascend main now inherits [#16306](https://github.com/vllm-project/vllm-ascend/pull/16306), which temporarily skips DSpark PCP. This PR does not modify that test file or add the skip. A green CI with this inherited skip is not an actual DSpark pass; the independent precision investigation/fix remains outside this PR.
- No new patch file is added. This does not add DBO, CUDA Triton attention, NPU ReplaySSM or release MRV2 extract-hidden-states support.

#### File-by-file changes and upstream evidence

[...]

PR #16311, opened 2026-09-11
[BugFix][Attention] Isolate PCP global RoPE runtime buffers

zhao-stack:

Fix RoPE runtime-buffer aliasing between PCP's canonical global KV-update metadata and rank-local attention metadata.

With DSpark verification, short query rows can put both builds on the cached RoPE path even though their positions differ. Each PCP cache-group builder constructs global metadata, then local metadata. Local RoPE proxies are reused through the step's shared metadata, but a later cache group's global build writes the same process-wide runtime buffers again. The local proxy can then contain global-position values, corrupting target hidden states, generated tokens and speculative acceptance.

Each PCP global metadata builder now owns a persistent RoPE buffer map keyed by configuration and group. Each cos/sin pair is allocated once at the full registered capacity; subsequent calls use the existing in-place gather. Addresses remain stable from small warmup/capture batches through larger replay batches. Local/default, uncached prefill and speculative draft-buffer behavior is preserved; no global cache disabling, per-step cloning or version-specific branch is needed.

Buffer lifetime follows the global builder. Extra storage per requested configuration/group per global builder is `2 * max_num_batched_tokens * rotary_dim * element_size`; the global Tensor-position path requests only the default group. Distinct PCP cache-group builders own separate maps. Non-PCP paths allocate no extra buffers.

This independent fix branch starts from upstream main `f4b6bd05f05717e10e4521a78f4b3beba5bb5a99`. It is separate from main2main PR #16009 and changes none of #16009, #15627, #15965 or #16216. The PR-wide diff contains only three production files and two UT files; pins, release tags, skips, thresholds, golden data and e2e sources are unchanged. The shared-buffer structure predates the historical upgrade experiment; there is no evidence that #16009 or `6f3c34a` introduced it.

Correct position encodings and generation/acceptance behavior for affected PCP + DSpark batches. No configuration or public API change.

**Current branch validation — actual pytest logs inspected:**

| Validation | vLLM main `b2f685834a6456197e7033966fdef52a23f1abcd` | release `v0.28.0` / `2cf0a6915ce544dc493a0990f2ea38d81601128a` |
|---|---|---|
| RoPE + DSA unit suites | 56 passed | 56 passed |
| Original four-card PCP DSpark e2e | [1 passed, 525.72 s](https://github.com/vllm-project/vllm-ascend/actions/runs/34554944313/job/103126093187) | [1 passed, 524.02 s](https://github.com/vllm-project/vllm-ascend/actions/runs/34554944313/job/103126093201) |

The e2e run was triggered once with the repository's supported comment syntax:
```
/e2e tests/e2e/pull_request/four_card/context_parallel/test_deepseek_v4.py::test_deepseek_v4_dsa_pcp_dspark
```
The logs confirm the exact pytest node, model, version checkouts, full graph capture and final passing pytest summaries. Original model, prompts, parameters, prefix assertions, acceptance thresholds and 0.03 tolerance remain intact. No probes, controlled admission, extra test plugins or assertion changes were committed. The passing test does not print final aggregate acceptance rates; periodic metric-log rates are not substituted for the assertion's aggregate metrics.

**Commit provenance:** the e2e workflow pins fix commit `c14993202b111e5c496dc4ffa8ce946e226f4fe7` and rebases it on fixed upstream snapshot `fa9f8a7f3d57d28122109dd51a50d91d9f53a805`. Final head `ca169f36aef921203a9857cb15cbf12a4ef695a1` only adds UT dictionary annotations required by mypy. All production files and the original e2e source are byte-identical between these commits; this is not presented as a final-head e2e invocation. The two 56-test suites were rerun after that annotation follow-up, and the remote tested tracked tree was verified against the final signed commit.

The new numerical builder regression fails on an independent, unmodified upstream-base checkout: 12/16 elements differ after global/local builds (maximum absolute difference 1.088031530380249). It passes with the fix. Coverage includes global → local → next global cache-group builds, preservation of both outputs, repeated calls, full-capacity growth, address stability, multiple RoPE configurations/groups, and unchanged uncached/draft paths.

```bash
pytest -q tests/ut/ops/test_rope_dsv4.py \
  tests/ut/attention/test_dsa_v1.py tests/ut/attention/test_dsa_cp.py
```

Final-head repository CI:
- [CPU UT](https://github.com/vllm-project/vllm-ascend/actions/runs/34555534152/job/103128256661): **4190 passed, 38 existing skips**, actual pytest summary inspected.
- [Pre-commit](https://github.com/vllm-project/vllm-ascend/actions/runs/34555534152/job/103127322490): all hooks passed; full production/example/test mypy checks passed for Python 3.10, 3.11 and 3.12.
- Standalone Gitleaks also passed. The initial seven missing test dictionary annotations were fixed before the successful final-head checks.

**Historical causal evidence, distinct from this branch's CI:** an earlier isolated Ascend experiment with vLLM `a97dacb7106ee49f39f3d1fc6ae1800ff724e01d` reproduced third-position acceptance **0.4933095450490633**, matching the failing CI's complete outputs. With fixed diagnostic admission conditions in the same directory, the ownership fix passed, reverting it restored the exact failure, and reapplying it passed with identical tokens/counts to the first fixed run. Two subsequent original uninstrumented main singletons passed; historical release attempts were resource-blocked. Immutable causal archive SHA256: `da75a996eba465e79837c834a5b21d72e9f4b0ae811163593c5eb6ba1f4484f6`. The same Triton disk-cache seed produced both passes and failures, so cache contents are not claimed as the root cause.

The separate local four-card attempt on this branch did **not** complete: main stopped before generation because ModelScope tried to write `.mdl` through a read-only model-cache link; release was blocked before pytest by external device occupancy. Writable private metadata was prepared without downloading/modifying weights, and no external processes were stopped. These local attempts are not counted as passes or precision failures; the completed current-version device results above come from CI.

**Remaining merge gate:** `ci-gate` requires a maintainer to add `ready-precise` or `ready-all` for the repository's broader selected/full suite. The targeted dual-version `/e2e` run does not automatically satisfy that label-controlled gate. No merge is requested by this task.

### DIFF
--- a/tests/ut/attention/test_dsa_v1.py
+++ b/tests/ut/attention/test_dsa_v1.py
@@ -1707,13 +1707,13 @@ def test_pcp_builders_keep_global_rope_separate_from_reused_local_metadata(monke
         )
         AscendDSAMetadataBuilder.build(builder, 0, common, common_ratio_to_sas_metadata=shared)
 
-    addresses = {}
+    addresses: dict[tuple[int | str, str], int] = {}
     for count in (4, 16, 6):
         global_pos, local_pos = torch.arange(count), torch.arange(count - 1) + 5
-        local_cache = {}
+        local_cache: dict[str, Any] = {}
         global_caches = []
         for index, builder in enumerate(builders):
-            global_cache = {}
+            global_cache: dict[str, Any] = {}
             build(builder._global_metadata_builder, global_pos, global_cache)
             global_caches.append(global_cache)
             build(builder, local_pos, local_cache)
--- a/tests/ut/attention/test_dsa_v1.py
+++ b/tests/ut/attention/test_dsa_v1.py
@@ -53,6 +53,7 @@ from vllm_ascend.models.deepseek_v4.indexer import (
     AscendIndexerMetadata,
     IndexerOverlapPlan,
 )
+from vllm_ascend.ops import rope_dsv4
 from vllm_ascend.worker.device_metadata import (
     DeviceMetadataStage,
     DeviceMetadataTask,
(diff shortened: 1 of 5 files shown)


zhao-stack:
/e2e tests/e2e/pull_request/four_card/context_parallel/test_deepseek_v4.py::test_deepseek_v4_dsa_pcp_dspark
[Bot]: e2e command triggered. See [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/34554944313) for details.
[Bot]: e2e command completed successfully.

PR #16317, opened 2026-09-11
[BugFix][KVCache] Allow external lookup below two retention intervals

zouyida2052:

Allow external cache lookup for prompts shorter than two retention intervals. With a retention interval and transfer granularity of 4096, a 5083-token prompt can reuse an existing 4096-token checkpoint. The current early return reports zero without querying the cache.

The production change only removes this seven-line retention-length guard. Existing checks for requests shorter than one transfer unit, actual cache availability, and MTP tail recomputation remain unchanged. Requests previously excluded by the guard can now incur a lookup even when the cache is empty.

The MTP implementation is identical to the upstream base, including its aligned `final_block_start` calculation. MTP boundary tests confirm existing behavior; they are not evidence of another newly fixed MTP bug.

Yes. Requests shorter than two retention intervals can reuse valid external checkpoints instead of being forced to miss. No API or configuration changes.

- **Unit tests:** 50 passed in `tests/ut/distributed/ascend_store/test_pool_scheduler.py`. The retention regression covers MTP enabled/disabled, stored and missing checkpoints, and the existing early return below one transfer unit. Additional boundary coverage confirms the unchanged MTP behavior.
- **Baseline control:** On the upstream base without this fix, the retention lookup regression fails (0 instead of 4096), while the added MTP boundary test already passes.
- **Static checks:** Ruff check, Ruff format check, and `git diff --check` pass for the changed files. The full `bash format.sh ci` hook suite could not run because the runtime lacks `pre-commit`.
- **NPU experiments rerun with upstream MTP logic:** DeepSeek-V4-Flash-w8a8-mtp, TP4 × DP2, MTP=1, retention interval 4096. The runtime uses the upstream MTP block verbatim and removes the retention guard. Five frozen AISBench inputs were replayed with four warmup requests and four serial measured requests per length, one output token per request. Measured external hits/queries and outputs match both the previous implementation and the previously measured local-prefix-only mode.

| AISBench input_len | Actual prompt tokens/request | External hits/queries | Previously measured local prefix hits/queries | Hit rate |
| --- | --- | --- | --- | --- |
| 4096 | 4179 | 16384/16716 | 16384/16716 | 98.01% |
| 5000 | 5083 | 16384/20332 | 16384/20332 | 80.58% |
| 5004 | 5087 | 16384/20348 | 16384/20348 | 80.52% |
| 8192 | 8275 | 32768/33100 | 32768/33100 | 99.00% |
| 9000 | 9083 | 32768/36332 | 32768/36332 | 90.19% |

The chat template adds 83 tokens in this setup. Counts exclude warmup. Direct AISBench commands were also rerun for all five lengths, with four successful requests and zero failures in both warmup and measured stages:

```bash
python3 aisbench_test.py --input_len 5000 --output_len 1 --data_num 4 \
  --concurrency 1 --request_rate 0 --dataset_type prefix_cache \
  --repeat_rate 1.0 --prefix_test --dp 4
```

Token-ID input tests reconfirmed that actual lengths 4096/5004/8192/9000 reuse 0/4096/4096/8192 tokens per request. A separate 5083-token input produced identical 64-token output for an external miss and subsequent hit.

**Validation limitations:** Clean-checkout unit tests use the directory's existing `_mock_deps` stubs and `--confcutdir=tests/ut/distributed/ascend_store`; the runtime's parent conftest has an unavailable `vllm.v1.attention.ops.pcp` dependency. The NPU runtime is the existing `bdab8bde86aad5eba3b362f923a3ff03b2c0e772` deployment with the upstream aligned MTP block and the retention guard removed, retaining its existing memcache backend changes; it is not a full deployment of the latest main. Cold concurrent warmup can differ between local and external caching due to publication timing and DP cache scope. These checks do not establish general model accuracy, lookahead KV safety for different continuations, graph-mode coverage, or statistical performance equivalence. This PR leaves upstream MTP handling unchanged.

### DIFF
--- a/tests/ut/distributed/ascend_store/test_pool_scheduler.py
+++ b/tests/ut/distributed/ascend_store/test_pool_scheduler.py
@@ -149,23 +149,28 @@ class TestKVPoolScheduler(unittest.TestCase):
     @patch("vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.pool_scheduler.LookupKeyClient")
     def test_retention_interval_does_not_skip_external_lookup(self, mock_client_cls):
         """A retained checkpoint can be reused before two retention intervals."""
-        for token_count, lookup_hit in [(5083, 4096), (4083, 0), (8275, 4096)]:
-            with self.subTest(token_count=token_count):
-                mock_client_cls.reset_mock()
-                mock_client_cls.return_value.lookup.return_value = lookup_hit
-                scheduler = KVPoolScheduler(self._make_config(block_size=32), use_layerwise=False)
-                scheduler.retention_interval = 4096
-                request = MagicMock(
-                    prompt_token_ids=list(range(token_count)),
-                    num_tokens=token_count,
-                    request_id="retained-prefix",
-                    block_hashes=[b"h"] * (token_count // 32),
-                )
+        for use_eagle in (False, True):
+            for token_count, lookup_hit in [(5083, 4096), (5083, 0), (4083, 0), (8275, 4096)]:
+                with self.subTest(token_count=token_count, use_eagle=use_eagle):
+                    mock_client_cls.reset_mock()
+                    mock_client_cls.return_value.lookup.return_value = lookup_hit
+                    scheduler = KVPoolScheduler(self._make_config(block_size=4096), use_layerwise=False)
+                    scheduler.retention_interval = 4096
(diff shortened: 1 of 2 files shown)


lijiahang226 (approved):
LGTM.

Pz1116 (approved):

weiguihua2 (approved):

PR #16172, opened 2026-09-09
[Model][Feature] Integrate Model Runner V2 and DSpark Speculative Decoding for Kimi-K3 on Ascend·

qijiajin:
### What this PR does / why we need it?

Integrate Kimi K3 MRV2 DSpark auxiliary-hidden and graph correctness work onto main, currently synced through `5e062cbc85c3c9a1dc6297eb31c4884e9a660e1b`. The original integration was consolidated into a signed-off commit with contributor attribution, followed by a main merge and CI fixes.

- Preserve main's PCP draft-attention configuration context and stable cache-group layer ordering.
- Gather matching token rows before the dense TP MLP and shard after its existing AllReduce.
- Refresh GDN speculative graph state buffers using a backend-local view of inactive padding; preserve shared MLA/FIA metadata.
- Preserve both release and current upstream hidden-state cache layout APIs.
- Include auxiliary-hidden negotiation, draft graph DP metadata and KV regression coverage.

### Does this PR introduce _any_ user-facing change?

Corrects Kimi K3 MRV2 target/speculative graph contracts. **No expert reduction:** the environment switch, expert filtering, router/bias slicing and related documentation/tests have been removed. No layer reduction or checkpoint modifications. The full router and last expert are covered by a regression test.

### How was this patch tested?

- Clean integrated source with vLLM 0.28.0: **219 passed, 9 skipped** across 12 focused test files.
- Includes PCP context/order conflict regression and full expert/router weight forwarding.
- Ruff check/format, codespell, typos and diff-check passed. Markdown formatting applied.
- Full local CI remains incomplete: Windows host lacks /bin/bash for gitleaks, shellcheck and usable python3 hooks.
- Follow-up main merge: 68 focused KV/DSpark tests passed. Optional-config typing fix: 27 DSpark tests passed. Dynamic DFlash replay namespace typing fix: 1 regression test passed.
- CI fixes use the configured vLLM logger and explicit typing for optional configuration and the dynamic test namespace. Remote pre-commit, DCO and Markdown link checks passed on `53fb3081a449c4ed378fa4a38450fd5434d62174`. The DFlash test also handles Python-version-specific AST fields without suppressing checks. All 12 focused test files were rerun after the follow-ups: **219 passed, 9 skipped**. Downstream CPU CI is pending.
- No new NPU service rollout of this PR revision. Running service remains the older `5affa7ac0` snapshot.

[...]

PR #16318, opened 2026-09-11
[Attention][Feature] Support GQA DSpark and MRV2 Speculative Decoding on Ascend

qijiajin:
Depends on #16312.

The branch is linearized in `4408b14ce95be2971ac3cfd91ed197a4fa619d6f` on main snapshot `125924bb2cb5ca3f68368f251c01c705d3a67b10`. Its tree matches the previous PR head merged with that snapshot.

This PR adds the runtime contracts required for Kimi K3 MRV2 to run with a Qwen3 GQA DSpark draft on Ascend.

- Configure the Kimi K3 target to provide **materialized auxiliary hidden states** required by the GQA draft, instead of consuming the raw prefix-sum residual representation.
- Preserve the target QuaRot rotation path while constructing a BF16 GQA draft so the existing draft loader can align FC, embedding and lm_head before upstream decides whether target weights can be shared. QuaRot GQA drafts are rejected if they accidentally share the target embedding or lm_head.
- Keep draft attention layers in KV-cache-group order, validate that all active draft layers are mapped, and require one supported GQA attention backend for draft ACL Graph replay.
- Build speculative-decoding attention metadata with the padded graph query contract and validate `actual_seq_lengths_q` before graph submission.
- Prevent rejected speculative context and NULL/padding query rows from being mapped to real draft cache slots.
- Remove inactive speculative graph suffix rows from the **GDN-local** metadata view so FIA can retain the fixed graph shape without allowing padding requests to advance recurrent GDN state.
- Keep vLLM 0.28 `HiddenStateCacheSpec` reshaping compatible with auxiliary hidden-state caches.
- Add draft ACL Graph contract checks for token/request counts and captured parameter/handle/event cardinality.

This PR intentionally excludes the target Dense-SP and 1P1D fixes already split into #16312. It also excludes reduced-layer/reduced-expert logic, MLA DSpark model changes, deployment scripts and documentation from #16172.

Refs #13745.

Yes. Kimi K3 MRV2 can use a Qwen3 GQA DSpark draft on Ascend with the required auxiliary-hidden-state, QuaRot, cache-state and FULL graph metadata contracts. Unsupported graph/backend combinations now fail early instead of continuing with inconsistent runtime metadata.

#### Focused unit tests

Run on `192.168.13.154` in container `qjj_k3_main_check154`, using vLLM 0.28.0 source from `/home/q00852295/main-k3/vllm-v0.28.0`:

```bash
pytest -q \
  tests/ut/patch/platform/test_patch_speculative_config_dspark.py \
  tests/ut/spec_decode/test_dspark_speculator.py \
  tests/ut/spec_decode/test_dspark_graph_contract.py \
  tests/ut/worker/test_attn_utils_v2.py \
  tests/ut/ops/test_gdn_attn_builder.py
```

Result: **149 passed, 2 skipped**.

The focused coverage includes:

- legacy GQA config normalization to `materialized` auxiliary hidden states;
- target capture-mode configuration before draft loading;
- QuaRot path injection/cleanup and FC projection equivalence;
- rejection of target embedding/lm_head sharing for QuaRot GQA drafts;
- deterministic KV-group layer ordering and single-backend graph contract;
- padded `actual_seq_lengths_q` validation;
- draft graph parameter/handle/event cardinality checks;
- speculative attention state propagation;
- GDN FULL-graph padding refresh while preserving shared metadata.

#### Static / CI checks

For head `4408b14ce95be2971ac3cfd91ed197a4fa619d6f`:

- Local Ruff check and format: passed for all 16 changed Python files.
- Python compilation and `git diff --check`: passed.
- Forced replay onto main snapshot `125924bb2cb5ca3f68368f251c01c705d3a67b10`: passed without conflicts.
- Exact tree equality against merging previous head `1226d4332fbe34174f5311f72d30e703883a374e` with that main snapshot: verified.
- Local `bash format.sh ci` could not run because pre-commit is not installed. CPU/NPU tests were not rerun locally in this history-repair pass.
- [E2E run 34824173421](https://github.com/vllm-project/vllm-ascend/actions/runs/34824173421) completed with failure. In [pre-commit job 103912342865](https://github.com/vllm-project/vllm-ascend/actions/runs/34824173421/job/103912342865), the pre-commit hooks passed, but the separate mypy step failed: `vllm_ascend/xlite/xlite.py:664: "XModelConfig" has no attribute "index_full_mask" [attr-defined]` (Python 3.10, 601 source files checked). CPU UT and downstream tests were skipped. The earlier focused test results are historical.
- The job checked merge `a5848b4f0e647d35f6980ed2f9d973afc0bd1de2` (head `4408b14c` into base `cab37196`) and vLLM `a97dacb7106ee49f39f3d1fc6ae1800ff724e01d`.
- This PR changes neither `xlite.py` nor `xlite/utils.py`; both blobs exactly match current main `d23a1372375b2ecb2a611b8c1eec0c7e6eff8f3d`. Independent [PR #16395 job logs](https://github.com/vllm-project/vllm-ascend/actions/runs/34821759146/job/103904913586) show the same error. This establishes a shared Xlite static attribute-contract blocker, not a DSpark-specific failure. The precise dependency/stub/cache cause remains unverified.
- Maintainer action: inspect the CI-installed Xlite package and type declarations against the required `xlite==0.2.0rc1`, then align the declared/runtime `ModelConfig.index_full_mask` contract in the shared dependency or upstream adapter. If CI environment/type declarations are repaired, rerun failed jobs of run `34824173421`; if the fix is merged into main, rebase this linear branch onto that fix and run a new E2E workflow. No unrelated Xlite source change or type-ignore workaround was added here.

The previous [CPU UT job](https://github.com/vllm-project/vllm-ascend/actions/runs/34813486698/job/103880346289) installed upstream vLLM successfully, then failed before tests at `Rebase on main snapshot`: replaying `a061137c2` conflicted in `tests/ut/worker/test_attn_utils_v2.py` and `vllm_ascend/worker/v2/attn_utils.py`. Linearizing the resolved tree removes the replay of superseded conflicting patches. No CI workflow or additional business logic was changed for this repair. The `ready-precise` label is already present.

#### NPU validation method

The GQA Graph validation is split so Target Graph and Draft Graph failures can be isolated instead of being hidden by the final combined configuration:

| Case | Target | Draft | Purpose |
| --- | --- | --- | --- |
| A | Eager | Eager | Freeze the oracle/baseline |
| B | Graph | Eager | Validate Target verification, materialized aux hidden and GDN graph state |
| C | Eager | Graph | Isolate Draft Graph behavior |
| D | Graph | Graph | Validate the final combined path |

For the first GQA Graph acceptance pass:

- Draft: Qwen3 GQA DSpark, `K=7`, greedy Draft and greedy Target;
- parallelism: single-node TP16/EP, DP1;
- start from the fixed Eager parity fixture, then cover padded graph sizes and dynamic batches;
- confirm the **actual** Target/Draft manager modes and graph descriptors from logs, not only launch arguments;
- compare final token IDs and per-step proposal/accepted progress at the same logical token boundaries;
- cover padding requests, rejected suffixes, cache/block boundaries, prefix cold/warm/reset, request completion and dummy/capture paths;
- verify no inactive/padding request advances GDN state and no rejected context is written to a real Draft cache slot;
- require supported captured sizes to replay without unexpected Eager fallback or post-request ACL/HCCL/Graph errors.

#### Existing GQA Eager baseline data

The following data is the pre-Graph GQA DSpark baseline used for regression comparison; it is **not** claimed as the acceptance result of this Graph PR:

- short-request acceptance recorded in #15807: **26.2%**;
- one GSM8K request: **111 accepted / 161 proposed = 68.94%**;
- fixed **63-token** prompt fixture: **128/128 generated token IDs matched** the same-configuration no-spec Target in Eager mode;
- baseline scope: single input, cold cache, single-node TP16/EP, Eager.

These results show that the previous persistent 0% acceptance issue was no longer present and provide a deterministic oracle for Graph regression. They do not replace the repository precision suite, multi-sample accuracy validation, or full-expert performance/accuracy validation.


vllm_ascend/models/qwen3_dspark.py
```diff
@@ -21,6 +21,16 @@
 )
 
 
+def _get_draft_rotation_path(vllm_config: VllmConfig, config) -> Path | None:
```
lilinsiman, line 24: maybe gather this function into get_rotation_path

vllm_ascend/worker/v2/spec_decode/dflash/aclgraph.py
```diff
@@ -82,18 +83,38 @@ def run_fullgraph(self, desc: BatchExecutionDescriptor) -> torch.Tensor | tuple[
...
+            raise RuntimeError("Draft ACL graph manager has no bound speculator.")
+        if self.update_stream is None:
+            raise RuntimeError("Draft ACL graph manager has no bound update stream.")
+        expected_num_tokens = desc.num_reqs * self.speculator.num_query_per_req
+        if num_tokens != expected_num_tokens:
+            raise RuntimeError(
+                "Draft ACL graph descriptor mismatch: "
+                f"num_tokens={num_tokens}, num_reqs={desc.num_reqs}, "
+                f"num_query_per_req={self.speculator.num_query_per_req}."
+            )
```
slippersss: Why do we need this here?

vllm_ascend/worker/v2/spec_decode/dflash/aclgraph.py
```diff
@@ -122,3 +138,23 @@ def run_fullgraph(self, desc: BatchExecutionDescriptor) -> torch.Tensor | tuple[
...
+            raise RuntimeError(f"Draft ACL graph has no captured parameter bucket for {num_tokens} tokens.")
+        counts = (len(captured_params), len(handles), len(events))
+        if counts[0] == 0:
+            raise RuntimeError(f"Draft ACL graph captured no attention update tasks for {num_tokens} tokens.")
+        if len(set(counts)) != 1:
+            raise RuntimeError(
+                "Draft ACL graph update cardinality mismatch for "
+                f"{num_tokens} tokens: params={counts[0]}, handles={counts[1]}, events={counts[2]}."
+            )
+        return counts[0]
```
slippersss: Return value but never use it? Do we need so much raise?

qijiajin:
Pushed e4eeddac44e241ab863efb122d1c5fed66a6d042: resolve draft capabilities through upstream `get_model_cls()` rather than model-type/architecture heuristics.

- The Ascend Qwen3 DSpark class declares its auxiliary hidden-state format and target QuaRot alignment requirement.
- Legacy normalization continues converting architecture/token/layer fields but no longer injects a duplicate format declaration. Explicit format conflicts remain startup errors.
- Removed the redundant quant-config null check; the shared rotation-path helper already handles it.
- No changes to SP, numerical kernels, cache layout, GDN, or graph updates in this commit. No reduced-layer or reduced-expert code is included.

Validation: 76 targeted tests passed, 2 main-only interface tests skipped under the paired vLLM 0.28.0 environment. Ruff/format and the relevant local checks passed, with platform-dependent checks run separately on Linux.

1P1D smoke/regression used an existing **93-layer, 16-expert experimental deployment**, P eager and D target+draft FULL graph, TP16 on each node, max-model-len=9216, max-num-seqs=4, max-num-batched-tokens=4096, memory utilization=0.95, K=7. Requests went through the proxy: 8 fixed GSM8K questions at concurrency 1 and 4, max_tokens=512, temperature=0, top_p=1, seed=0. Acceptance is the D-side accepted/drafted counter delta.

| Run | Concurrency 1 | Concurrency 4 |
| --- | ---: | ---: |
| Original baseline A0 | 76.58% | 70.84% |
| Refactor B1 | 69.98% | 72.25% |
| Refactor B2 | 73.35% | 72.83% |
| Original restored A1 | 81.71% | 70.32% |
| Original restored A2 | 80.21% | 81.37% |

**Limitations:** This confirms loading, capture and proxy execution, not numerical parity or accuracy. The experimental node baselines include pre-existing changes and are not a clean full-expert PR validation. Same-version repeated outputs were not deterministic: B1/B2 differed for 4/8 serial and 2/8 concurrent requests; A1/A2 differed for 1/8 serial and 4/8 concurrent requests. All 80 requests completed without API errors, but final-answer matches were 0/80 and repetitive abnormal text occurred in both versions. Existing instability prevents attributing cross-version differences solely to this refactor; it also prevents declaring the refactor numerically equivalent. The original service has been restored. Fresh GitHub CI results are still pending.

vllm_ascend/worker/v2/spec_decode/dflash/aclgraph.py
```diff
@@ -82,18 +83,38 @@ def run_fullgraph(self, desc: BatchExecutionDescriptor) -> torch.Tensor | tuple[
         """Override run_fullgraph to update full graph params in run_fullgraph."""
         num_tokens = desc.num_tokens
 
+        if self.speculator is None:
+            raise RuntimeError("Draft ACL graph manager has no bound speculator.")
```
drslark: Redundant checks.
Please remove it.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
```diff
@@ -111,6 +160,22 @@ def set_attn(
...
 
             self.attn_backends = attn_backends
+        if active_layer_names is not None:
+            missing_layers = active_layer_names.difference(attn_backends)
+            if missing_layers:
+                raise RuntimeError(
+                    f"DSpark attention layers were not mapped to KV-cache groups: {sorted(missing_layers)}."
+                )
+
+    def get_draft_graph_backend(self) -> type[AttentionBackend]:
```
drslark: Does vllm has get_draft_graph_backend?

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
```diff
@@ -53,21 +55,66 @@ def load_draft_model(
...
         rotation_path = get_rotation_path(self.vllm_config)
-        if rotation_path is not None and hasattr(model.model, "fc"):
-            rotation_weight = get_rotation_matrix(rotation_path)
-            fc = model.model.fc
-            with torch.no_grad():
-                fc.weight.data.copy_(process_weight(fc.weight.data.cpu(), rotation_weight))
+        injected_rotation = rotation_path is not None and self._is_qwen3_gqa_draft(draft_hf_config)
+        if injected_rotation:
+            draft_hf_config._ascend_target_rotation_path = str(rotation_path)
+        try:
```
drslark: Don't use `try`.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
```diff
@@ -53,21 +55,66 @@ def load_draft_model(
...
+            model = super().load_draft_model(target_model, target_attn_layer_names)
+        finally:
+            if injected_rotation:
+                delattr(draft_hf_config, "_ascend_target_rotation_path")
+
+        if injected_rotation:
```
drslark: rotation must be put in modelding.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
```diff
@@ -156,6 +223,18 @@ def _update_draft_attn_metadata(self, attn_metadata, num_reqs_padded):
             metadata.actual_seq_lengths_q = query_lens_list
         return attn_metadata
 
+    def _validate_draft_attn_metadata(self, attn_metadata, num_reqs_padded):
```
drslark: Why we need this?

vllm_ascend/worker/v2/spec_decode/dflash/aclgraph.py
```diff
@@ -122,3 +138,23 @@ def run_fullgraph(self, desc: BatchExecutionDescriptor) -> torch.Tensor | tuple[
                 draft_attn_metadatas=draft_attn_metadatas,
             )
         return ret
+
+    @staticmethod
+    def _validate_graph_param_cardinality(graph_params: Any, num_tokens: int) -> int:
```
drslark: Why we need this?

drslark (changes requested):

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
```diff
@@ -53,19 +51,22 @@ def load_draft_model(
...
-            rotation_weight = get_rotation_matrix(rotation_path)
-            fc = model.model.fc
-            with torch.no_grad():
-                fc.weight.data.copy_(process_weight(fc.weight.data.cpu(), rotation_weight))
+        draft_hf_config._ascend_target_rotation_path = str(rotation_path) if rotation_path is not None else None
+        try:
+            model = super().load_draft_model(target_model, target_attn_layer_names)
+        finally:
+            delattr(draft_hf_config, "_ascend_target_rotation_path")
+        set_target_model_capture_mode = getattr(model, "set_target_model_capture_mode", None)
```
drslark: What is this?
Why draft affect target's mode?

drslark (changes requested):

vllm_ascend/worker/v2/attn_utils.py
```diff
@@ -1192,18 +1210,21 @@ def build_attn_metadata_wrapper():
...
+    latter must be ``SpecDecoding`` so draft tokens use speculative attention
+    semantics instead of being treated as independent requests. Must run inside
     ``build_attn_metadata_wrapper()``.
     """
     raw = _BUILD_ATTN_METADATA_MODULE.build_attn_metadata  # cache
 
     def build_attn_metadata(*args, **kwargs):
         kwargs["positions"] = positions[:pad]
         kwargs["is_prefilling"] = is_prefilling
+        kwargs["attn_state"] = AscendAttentionState.SpecDecoding
```
drslark: `AscendAttentionState.SpecDecoding` should be deprecated.
Why do you still need it?

drslark (changes requested):

vllm_ascend/worker/v2/spec_decode/dflash/aclgraph.py
qijiajin: Agreed. The replay-time speculator/update-stream checks were removed in the subsequent cleanup; they are no longer present in the current head.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
qijiajin: This was an Ascend-local helper, not an upstream vLLM API. It and the extra single-backend validation were removed in the subsequent cleanup. The current head retains the existing backend selection and the KV-group ordering fix.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
qijiajin: Agreed. The duplicate _validate_draft_attn_metadata() self-check was removed. The actual padded-query metadata update remains, with tests checking that function's output directly.

vllm_ascend/worker/v2/spec_decode/dflash/aclgraph.py
qijiajin: Agreed. _validate_graph_param_cardinality() was removed from this PR in the subsequent cleanup; it was diagnostic hardening rather than a required GQA adaptation.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
qijiajin: This is try/finally, not try/except: loader exceptions are propagated unchanged, while the temporary target-rotation attribute is removed on both success and failure. Removing only the finally cleanup would leave mutated HF configuration after a failed load. I am retaining it until the temporary configuration transport is replaced by an explicit loader interface; this PR does not modify upstream vLLM.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
qijiajin: Agreed that rotation computation belongs in modeling. In the current head, FC weight rotation and fallback embedding/lm_head alignment are performed only by AscendQwen3DSparkForCausalLM.load_weights(). The speculator passes only the target rotation path before loading, because upstream replaces the BF16 draft quant_config with None. It no longer performs post-load FC rotation or substitutes the target quantization config for the draft's.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
qijiajin: Renamed the hook to configure_target_aux_hidden_capture() in 89c86964e and documented its scope. This configures only the auxiliary hidden representation consumed by the GQA draft, not the target's Eager/Graph execution mode. Kimi's raw residual representation differs from the materialized selected-layer inputs expected by this draft. The model-specific hook selects that auxiliary capture behavior; targets without the setter retain their native behavior. The numerical computation is unchanged by this rename.

vllm_ascend/worker/v2/attn_utils.py
qijiajin: Agreed that the new code should not claim this enum is inherently required for all speculative decoding. I corrected that overly strong comment in 89c86964e. I have not removed the assignment yet: the current backend still reads attn_state, including different C8 branches, and a unit test checking enum propagation does not establish numerical equivalence after removal. Deprecation should be validated against the replacement metadata path for draft eager and graph execution before changing routing. This remains open; I am not claiming the concern is resolved.

The updated tree was tested in an isolated directory on 154 with the paired vLLM 0.28.0 source: 151 passed, 2 skipped, and the same 2 pre-existing main-only KVPP failures because upstream_attn_utils.allocate_kv_cache is absent in this environment. Formatting, lint, logger, symbolic-meta, long-function and change-scoped secret checks passed. No service restart or new acceptance-rate measurement was performed in this round.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
drslark: No, `try` is confusing.
It is forbidden.

drslark (changes requested):

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
```diff
@@ -53,19 +51,22 @@ def load_draft_model(
...
-            rotation_weight = get_rotation_matrix(rotation_path)
-            fc = model.model.fc
-            with torch.no_grad():
-                fc.weight.data.copy_(process_weight(fc.weight.data.cpu(), rotation_weight))
+        draft_hf_config._ascend_target_rotation_path = str(rotation_path) if rotation_path is not None else None
+        try:
+            model = super().load_draft_model(target_model, target_attn_layer_names)
+        finally:
+            delattr(draft_hf_config, "_ascend_target_rotation_path")
+        configure_aux_hidden_capture = getattr(model, "configure_target_aux_hidden_capture", None)
```
drslark: Use `hasattr` instead of `getattr`?

drslark (changes requested):

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
qijiajin: Updated to `if hasattr(model, "configure_target_aux_hidden_capture"):` followed by a direct call. Also simplified rotation-path transport: it is retained on the draft HF config, so there is no try/finally or delattr cleanup. Loader failures still propagate. Focused tests on 154: 18 passed. These are unit-test results, not a new acceptance-rate measurement.

vllm_ascend/worker/v2/spec_decode/dflash/aclgraph.py
qijiajin: Already updated.

vllm_ascend/worker/v2/spec_decode/dflash/aclgraph.py
qijiajin: Already updated.

vllm_ascend/models/qwen3_dspark.py
qijiajin: Already updated.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
qijiajin: Already updated.

vllm_ascend/worker/v2/attn_utils.py
qijiajin: Addressed in f271388a59202d9761e8096972d71e598f1cc047. Removed the forced `attn_state=AscendAttentionState.SpecDecoding` from the draft metadata wrapper. It now supplies only positions/is_prefilling and preserves caller/backend state. Tests cover absent, None and explicitly supplied states. This does not remove enum uses belonging to other main-branch paths.

154 NPU A/B: same experimental 93-layer / 16-routed-expert baseline, TP16 DP1 EP, Target FULL_DECODE_ONLY and GQA DSpark draft graph, K=7, capture sizes [16,32], max length 9216, max sequences 4, batched tokens 4096, memory utilization 0.95. Only the state-assignment line differed between A and B. No weight/config files were changed and reduced-expert code is not included in this PR.

Fixed GSM8K rows 2-9, greedy, seed 0, max_tokens=512, cold prefix cache per batch. A ran 2 rounds at concurrency 1 and 4: both rounds had 2286/3143 accepted/drafted (72.73%) at c1 and 2312/2898 (79.78%) at c4. B's first 3 batches matched exactly; the fourth was 2305/2975 (77.48%) with one differing response. An additional 2 rounds on the unchanged B service matched all A counts and all 32 output token sequences. Across 64 B requests, 63/64 token sequences matched the paired A responses. All 96 requests succeeded; graph capture and final health check passed. The isolated concurrent discrepancy remains unexplained, so this is not a claim of unconditional deterministic parity.

Important accuracy limitation: all 96 final answers failed the strict answer check, and repetitive text was present in both A and B. This reduced-expert deployment is a functional comparison, not a model-quality validation; high acceptance must not be read as correct answers. No full-expert, 1P1D, or all-backend/C8 accuracy claim is made here. The A/B deployment also contains pre-existing experimental changes and predates the later loader-cleanup simplification; it isolates the state assignment rather than validating the full latest PR tree end to end.

Updated PR-tree tests on the paired vLLM 0.28 environment: 151 passed, 2 skipped, 2 main-only KVPP cases explicitly deselected because allocate_kv_cache is absent in that version. Lint/format, logger, long-function, symbolic-meta, diff checks and change-scoped secret scan passed.

drslark (approved):
`model_runner.py` and `speculator.py` are clear.
LGTM.

drslark:
Nightly tests of kimi dspark should be added in further pr.

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

qijiajin:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

lijiahang226 (approved):
LGTM

Angazenn (approved):

weiguihua2 (approved):

PR #16328, opened 2026-09-11
[Performance][Attention] Fuse SFA K-path ops and enable PROLOG_V3 fused decode for non-PD serving

tanjiangshan:

GLM5.2 SFA profiling shows several removable ops in the K processing path of every layer. This PR removes them and makes PROLOG_V3 the default fused preprocessing for quantized SFA layers in every deployment:

1. **Per-step int64 slot cast**: `exec_kv` re-cast the shared slot mapping to int64 for `npu_kv_rmsnorm_rope_cache` in every layer, although all layers of a scheduling step receive the same int32 slot tensor. The conversion is now cached on the attention metadata (`_int64_kv_slots`), so one Cast kernel runs per step instead of one per layer (~5us x num_layers per step). The PROLOG_V3 fused preprocess reuses the same cached conversion for its int64 cache indices.

2. **Duplicate indexer GEMM**: the indexer's k path (`forward_k`) and top-k stage (`forward`) both ran the same `wk_weights_proj` GEMM (`[tokens, hidden] x [160, hidden]`) on the same hidden states, once for the indexer K and once for the lightning-indexer weights. `forward_k` now returns the non-K tail of the GEMM output and `forward` reuses it, removing one GEMM plus its slice copy per indexer layer per step (falling back to the GEMM only when the two stages are handed different tensors).

3. **Redundant `.contiguous()` copies**: `npu_rms_norm` returns a contiguous tensor so the copy before the C8 block-quant view was discarded, and `torch.cat` already allocates contiguous outputs for the sparse-attention query concat.

4. **PROLOG_V3 by default, no new switch**: PROLOG_V3 (the `npu_mla_prolog_v3` single fused op covering qkv proj + norm + rope + q up-proj + C8 quantize/pack + direct cache write) was previously gated on `is_kv_consumer`, i.e. only PD-disaggregated decode workers could take it. It is now the default fused preprocessing for quantized SFA layers in every deployment (plain serving, PD KV producers and KV consumers) and serves every attention state: prefill and decode steps both take the fused path (the per-step attention-state fallback to NATIVE is gone; only MLAPO keeps its token-count limit). Switch convergence:
   - `enable_dsa_cp` is the prefill/P-node route selector: it routes to `AscendSFADSACPImpl`, which unconditionally disables fused preprocessing, so the two are mutually exclusive by construction (dsa_cp on => prolog off, dsa_cp off => prolog on);
   - the C8 switches (`enable_sparse_sfa_c8` / `enable_sparse_li_c8`) only select the KV cache layout and are orthogonal to this choice; W8A8Dynamic layers no longer require `enable_sparse_sfa_c8` to take the fused path;
   - unquantized layers keep the NATIVE chain outside KV consumers because the unquantized weight preparation transposes `fused_qkv_a_proj.weight` in place, which the NATIVE fallback still consumes;
   - `dispose_layer` stays gated on `is_kv_consumer` so producers and plain-serving workers keep the fallback weights (the cost is the extra PROLOG_V3 weight copies: memory, not correctness).

Yes, a default behavior change: quantized (W8A8Dynamic / W8A8MXFP8) SFA deployments now take the PROLOG_V3 fused preprocessing for both prefill and decode steps by default, without any additional-config option. Deployments on `enable_dsa_cp` (prefill/P-node CP route) and unquantized (bf16) layers are unaffected. The default trades extra NPU weight memory (the retained qkv_a/q_b fallback weights on producers and plain-serving workers) for kernel savings.

- Unit tests added/updated in `tests/ut/attention/test_sfa_v1.py`:
  - per-step int64 slot caching (passthrough / convert-and-cache / re-convert on new step) and `exec_kv` reusing the cached slots across layers;
  - single `wk_weights_proj` invocation across the indexer k path and top-k stage (plus the fallback recomputation when the tensors differ);
  - PROLOG_V3 routing matrix for the default gate (non-PD W8A8Dynamic with and without C8 -> PROLOG_V3; non-PD MXFP8 -> PROLOG_V3; non-PD unquantized -> NATIVE; KV producer quantized -> PROLOG_V3, unquantized -> NATIVE);
  - weight-disposal guard confirming non-consumer workers keep the fallback weights.

- `uvx ruff==0.14.0 check` and `format --check` on all touched Python files.
- CI cpu-ut green (4431+ tests).

**End-to-end A/B benchmark: DSA-CP route (main) vs PROLOG_V3 route (this PR)** (Atlas 800 A3, 16x 910B, CANN 9.1.0, vllm 0.28.0, GLM-5.2-w4a8c8, DP2xTP8 + EP):

The comparison is between the two decode preprocessing routes, each in its best usable configuration. `enable_dsa_cp` requires SP-MoE and unconditionally disables the fused preprocessing path, so the two options are mutually exclusive by construction; everything else is identical on both sides:

| item | base (main `125924bb2`) | PR (`cb525a869`) |
|---|---|---|
| **decode preprocessing route** | `enable_dsa_cp=true` | PROLOG_V3 route (now the default; measured with the earlier opt-in build of this PR) |
| SP-MoE / sequence parallelism (`VLLM_ASCEND_ENABLE_FLASHCOMM1=1`; auto-enabled by DSA-CP on the base side) | on | on |
| `enable_sparse_sfa_c8` + `enable_sparse_li_c8` | on | on |
| `enable_balance_scheduling`, `enable_fused_mc2=0` | on | on |
| `--enable-expert-parallel` (EP), DP2xTP8 | on | on |
| MTP speculative decoding (`deepseek_mtp`, num_speculative_tokens=3, enforce_eager) | on | on |
| cudagraph `FULL_DECODE_ONLY`, `--quantization ascend` | on | on |
| `multistream_overlap_shared_expert` | off | off (incompatible with FlashComm1 on DP>1, see #16446) |

Both sides verified via serve logs: no "Disabling DSA-CP" / sp-MoE active on the base side, `MlaPrologV3` kernels present in the PR-side profile only. Workload: GSM8K test full 1319 prompts, ais-bench stream mode, concurrency 8, temperature 0.

| metric | base (DSA-CP route) | PR (PROLOG_V3 route) | delta |
|---|---|---|---|
| TPOT avg | 27.9 ms | **22.9 ms** | **-17.9%** |
| TPOT median | 27.8 ms | 22.8 ms | -18.0% |
| E2EL avg | 7,181.6 ms | **5,885.0 ms** | -18.1% |
| TTFT avg | 459.9 ms | 395.8 ms | -13.9% |
| per-request output throughput | 33.73 tok/s | **40.91 tok/s** | +21.3% |
| request throughput (aggregate) | 1.1117 req/s | 1.3564 req/s | +22.0% |
| failed requests | 0/1319 | 0/1319 | = |

Kernel-level verification (rank0 profile of a 500-token decode request):

| kernel | base (DSA-CP) | PR (PROLOG_V3) | note |
|---|---|---|---|
| MlaPrologV3 | 0 | **13,369** | fused decode path taken over |
| InterleaveRope | 28,980 | **1,156** | -96%: NATIVE rope chain leaves decode |
| DynamicBlockQuant | 14,490 | **578** | -96%: c8 quant chain leaves decode |
| ScatterNdUpdate | 21,035 | 6,967 | -67% |
| Slice | 31,556 | 12,211 | -61% count |
| QuantBatchMatmulV3 | 73,652 (1,629 ms) | 44,207 (835 ms) | -40% count / -49% time |
| KvQuantSparseFlashAttention | 14,190 | 13,703 | ~same (attention body) |
| **total decode kernels** | **711,070** | **479,326** | **-32.6%** |

**E2E coverage for the PROLOG_V3 default route**:

- Following #16630 (GLM-5.2 spec-decode PR-level CI tests are removed as nightly-covered), the fused path is exercised by the nightly-covered GLM-5.2 spec-decode e2e suites together with the one-shot hardware verification below; no dedicated per-push full-forward UT is carried in this PR.
- Verified on real hardware (Atlas 800 A3, 8x 910B, CANN 9.1.0, vllm-ascend 0.28.0): the PROLOG_V3-route e2e case PASSED (acceptance length within 3.06 +/- 8%) when run as the eight_card job on branch `perf/glm-kpath-fusion-28` and in the E2E CI run of commit `28e9e11` (all 24 jobs green), confirming the fused route serves plain (non-PD) decode correctly under the full serving stack.
- Note: serving the w4a8c8 checkpoint on torch_npu 2.10 additionally requires a one-word out-of-tree fix in `vllm_ascend/quantization/methods/w4a8/w4a8.py` (`.sum(axis=1)` -> `.sum(dim=1)`; the `axis` kwarg comes from #13713 and torch_npu's `reduce_sum` rejects it; fixed on main by #16413).


lijiahang226:
Thanks for the contribution. Could you provide a performance comparison between the MLA Prolog v3 path and DSACP under prefill / mixed-deployment (hybrid) scenarios? These two features currently conflict with each other.

lijiahang226 (approved):

huamus:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

ZT-AIA (dismissed):

lijiahang226:
Could you please double-check whether we really need this many UTs for this change?

ZT-AIA (approved):

PR #16329, opened 2026-09-11
[BugFix][KV Pool] Don't pin mamba speculative scratch blocks on save

bowgneo:

vLLM PR [#51358](https://github.com/vllm-project/vllm/pull/51358) made the KV cache manager relocate Mamba speculative scratch blocks at chunk boundaries. Relocation is intentionally fail-closed: the scratch block must be exclusively owned (`ref_cnt == 1`) and unhashed before it can be moved.

For hybrid Mamba workloads using speculative decoding and non-layerwise AscendStore saves, `touch_sending_mamba_blocks()` previously pinned every non-null Mamba block until the asynchronous save completed. That included the speculative scratch tail, raising its reference count from 1 to 2. The next chunk-boundary relocation then failed the writable-block assertion and terminated EngineCore.

This PR establishes each cache group's actual `MambaSpec.num_speculative_blocks` as the single source of truth:

- Keeps `_infer_mamba_groups()` as the cache-spec normalization boundary. It unwraps `UniformTypeKVCacheSpecs`, validates the supported `mamba_cache_mode='align'`, and now returns `{group_id: num_speculative_blocks}` instead of only group IDs.
- Passes the same per-group mapping to `RequestTracker`, so its Mamba block-table mirror uses the same scratch-block count as the allocator.
- Excludes only each Mamba group's actual speculative scratch tail from the asynchronous AscendStore pin set. Persisted state blocks remain pinned until the worker acknowledgement.

Using the per-group value is important because the global speculative token count is not always the number of Mamba scratch blocks. For example, RecoverSSM can use speculative decoding while setting `MambaSpec.num_speculative_blocks` to zero. Reading the global value would incorrectly exclude real state blocks from asynchronous-send protection.

`UniformTypeKVCacheSpecs` is safe to represent with its first inner spec here because vLLM only creates a uniform Mamba wrapper when every inner spec is a `MambaSpec` with the same `num_speculative_blocks` and compatible layout properties.

No API or configuration changes.

Hybrid Mamba + speculative decoding workloads using AscendStore no longer pin relocatable scratch blocks during asynchronous saves. The intended persisted-state set is preserved, while zero-scratch groups now keep all real state blocks protected correctly.

### DIFF
--- a/tests/ut/distributed/ascend_store/test_pool_scheduler.py
+++ b/tests/ut/distributed/ascend_store/test_pool_scheduler.py
@@ -20,12 +20,13 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 import torch
-from vllm.v1.kv_cache_interface import FullAttentionSpec, KVCacheGroupSpec, SlidingWindowSpec
+from vllm.v1.kv_cache_interface import FullAttentionSpec, KVCacheGroupSpec, MambaSpec, SlidingWindowSpec
 
 import tests.ut.distributed.ascend_store._mock_deps  # noqa: F401, E402
 from vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.coordinator import AscendStoreCoordinator
 from vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.metadata import (
     LoadSpec,
+    ReqMeta,
     RequestTracker,
 )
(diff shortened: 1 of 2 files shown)


nwpu-zxr (approved):

Pz1116 (approved):

yiz-liu (approved):

yiz-liu:
/revert
[Bot]: revert completed successfully. New PR created: https://github.com/vllm-project/vllm-ascend/pull/17022

PR #16336, opened 2026-09-11
[Ops][Feature] Fused Triton Qwen vision QKV, RoPE, and FIA padding

YujianGuo:

This PR fuses QKV layout extraction, Q/K RoPE and head-dimension padding into one Triton producer for the Qwen2.5-VL/Qwen3.5-VL vision-attention FIA path. It removes the separate PyTorch preprocessing and padding operations before `npu_fused_infer_attention_score`.

- Restricts the specialized path to contiguous BF16 tensors with eight Q/K/V heads and head dimension 72;
- Accepts packed multi-sequence `cu_seqlens` for multi-image, multi-frame video and window-attention inputs;
- Primes and caches host FIA sequence boundaries once per vision forward group, avoiding repeated device-to-host conversion across attention layers;
- Records fused-path hit rate and fallback reasons with periodic statistics logging;
- Launches only the vector cores needed for the token tiles in the fused producer;
- Provides launch A/B knobs for vector-core mode, token tile size, Q/K store mode and output allocation mode;
- Allocates Q/K/V outputs from one contiguous buffer and exposes non-overlapping views to reduce allocator calls;
- Keeps all model integration in `vllm_ascend/patch/worker` instead of changing `vllm/model_executor/models/qwen2_5_vl.py` in the vLLM repository;
- Preserves the existing generic Qwen vision-attention path as the fallback;
- Feeds the fused padded Q/K/V tensors directly to FIA.

Yes. Supported Qwen2.5-VL/Qwen3.5-VL BF16 D72 vision-attention inputs automatically use the fused producer. The path can be disabled with `VLLM_ASCEND_ENABLE_VIT_FUSED_QKV_ROPE_PAD=0`; statistics and launch experiments are controlled by the `VLLM_ASCEND_VIT_*` environment variables added in this PR. Output shapes remain unchanged. Unsupported inputs and ACL graph capture continue to use the existing implementation.

- Tested on Ascend 910B3 with vLLM commit `752a3a504485790a2e8491cacbb35c137339ad34` and vllm-ascend baseline `3cb3f83db5eceeb1c8a8736b2337e167a225c60e`.
- Added unit coverage for the worker-patch short circuit, adapter guards, packed sequence lengths, cache reuse, entry-point priming and fallback statistics.
- Added NPU correctness coverage for token counts `1`, `4`, `41`, `257` and `8160`, packed multi-sequence FIA equivalence and launch-mode combinations.
- Verified padded columns `72:128` are exactly zero.
- The latest head `08e236018ab91ea012414e23125736b7b288e300` contains the packed multi-sequence optimization and its tests. CI validation is running for this new head.

The tests can be reproduced with:

```bash
pytest -q tests/ut/patch/worker/test_patch_qwen2_5_vl.py tests/ut/ops/test_mm_encoder_attention.py -k fused_qkv_rope_pad
pytest -q tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_vision_qkv_rope_pad.py
```

### Baseline vs. fused performance and accuracy

The 2026-09-15 AISBench RealWorldQA A/B run used 765 requests per arm, batch size 64 and max output length 512. Both arms completed 765/765 requests successfully. The only intended difference was `VLLM_ASCEND_ENABLE_VIT_FUSED_QKV_ROPE_PAD`; the fused arm reported 35,000/35,000 fused calls per TP worker with zero fallback, while the baseline arm reported `fallback:env_disabled`.

| Metric | Baseline (fusion off) | Fused (fusion on) | Change |
|---|---:|---:|---:|
| Benchmark duration | 620,967.8643 ms | 609,708.1409 ms | -1.81% |
| Request throughput | 1.2319 req/s | 1.2547 req/s | +1.85% |
| Input-token throughput | 1,784.4225 token/s | 1,817.3761 token/s | +1.85% |
| Output-token throughput | 202.4018 token/s | 206.4217 token/s | +1.99% |
| Total-token throughput | 1,986.8242 token/s | 2,023.7978 token/s | +1.86% |
| E2E latency average | 50,922.9 ms | 49,803.5 ms | -2.20% |
| E2E latency median | 40,224.3 ms | 39,649.4 ms | -1.43% |
| E2E latency P90 | 97,192.7 ms | 92,225.3 ms | -5.11% |
| E2E latency P99 | 162,238.8 ms | 157,892.1 ms | -2.68% |
| Generated output tokens | 125,685 | 125,857 | +0.14% |

### Accuracy

| Arm | Accuracy |
|---|---:|
| Baseline | 72.29% |
| Fused | 74.12% |

These are single-run end-to-end measurements. The fused arm is faster in this run, while the 1.83 percentage-point accuracy difference should not be attributed to the operator without at least three repeated runs and matched sampling; historical data had both arms at 74.25%. Decode batches were concentrated at 56--64 sequences, while the observed ACL graph capture sizes topped out at 24, so graph-size coverage should be expanded in a follow-up experiment.


YujianGuo (commented):
@dcoapp recheck

PR #16343, opened 2026-09-11
[Performance][Communicator] Replace per-layer F.pad with cat of a persistent zero block in MoE prepare

Levi-JQ:

`PrepareAndFinalizeWithMC2.prepare()` (and the identical site in `PrepareAndFinalizeWithAll2All`) pads `hidden_states` and `router_logits` to `padded_num_tokens` with `nn.functional.pad` on **every MoE layer, every step**, before the TP-split that feeds `MoeDistributeDispatchV2` (which requires a uniform token count across ranks). On NPU each `F.pad` lowers to two device kernels (`constant_pad_nd` → MemSet + PadV3) while the padded data is only a few KB in decode, so the four launches are pure overhead: ~44 us per MoE layer (~5.6% of a decode layer's kernel time in a Kimi K3 TP16 profile). With speculative decoding (`num_speculative_tokens=7`, minimum graph tier of 8 tokens), odd batch sizes always pad 8 rows to 16 and hit this path. The padded rows are dead data (dropped via `x_active_mask` in dispatch/combine and sliced off in finalize) — only the shape matters, so the padding can be made cheap.

This PR replaces both pads with `torch.cat([x, zero_block[:pad]], dim=0)` against a cached per-`(width, dtype)` zero block (in `vllm_ascend/ops/fused_moe/moe_utils.py`) that is never written after allocation:

- one ConcatD kernel per tensor instead of MemSet + PadV3 (2 kernels per layer instead of 4);
- value-equivalent to `F.pad` — the zero tail is structural, not maintained by any cross-call bookkeeping;
- cudagraph-replay-safe by construction: the cat output is a fresh tensor per call, so each captured graph owns its output buffer and no mutable state is shared across capture sizes (a copy-into-persistent-buffer scheme breaks its zero-tail invariant when graphs of different sizes alternate at replay, and was rejected for that reason);
- covers eager and any capture size; applies to both MC2 and All2All paths.

No.

- Unit tests: `tests/ut/ops/test_prepare_finalize.py` extended to cover the cat path and the F.pad fallback outside a worker context.

### DIFF
--- a/tests/ut/ops/test_prepare_finalize.py
+++ b/tests/ut/ops/test_prepare_finalize.py
@@ -2,6 +2,7 @@ import unittest
 from unittest.mock import MagicMock, patch
 
 import torch
+from vllm.config import set_current_vllm_config
 from vllm.model_executor.layers.fused_moe import FusedMoEConfig
 from vllm.model_executor.models.utils import sequence_parallel_chunk_impl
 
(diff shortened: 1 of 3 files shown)


weijinqian0 (approved):

weiguihua2 (approved):

ajariley:
/revert
[Bot]: revert command failed: you do not have permission. Only the PR author or users with triage+ permission can trigger /revert.
[Bot]: revert command failed. Please check the [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/35052217332) for details.

PR #16344, opened 2026-09-11
[Feature][Kimi K3 DSPark] Enable TP for context_proj

Levi-JQ:

Switch `context_proj` from `ReplicatedLinear` to `ColumnParallelLinear` (`gather_output=True`) so the projection weight [7168, 35840] is sharded along the output dim (7168) across TP ranks instead of being replicated.

This is not just a memory saving. During decode the context_proj GEMM is memory-bandwidth bound: profiling shows aic_mte2_ratio ~94% vs aic_mac_ratio ~8% (weight-load bound, batch=8). With the weight replicated, every rank re-reads the full ~490MB weight each step and the cluster bandwidth is not distributed. Sharding the output dim cuts the per-rank HBM weight read by TPx and spreads the traffic across ranks. `gather_output=True` reassembles [8, 7168] for `context_norm`, leaving the `combine_hidden_states` caller unchanged. Numerically equivalent to the replicated path; checkpoint layout unchanged.

Idea from: https://github.com/vllm-project/vllm-ascend/pull/15144

before: 710us(matmul)
image-20260828144021394
after: 104us(matmul+allgather)
image-20260831153334539

No.

- Validated end-to-end on a 4-node TP16/DP4 Kimi K3 deployment as part of the K3 v0.26.0rc PR stack: spec-decode acceptance rate flat vs baseline.

### DIFF
--- a/vllm_ascend/models/kimi_k3_dspark.py
+++ b/vllm_ascend/models/kimi_k3_dspark.py
@@ -8,7 +8,7 @@ import torch
 from torch import nn
 from vllm.config import VllmConfig
 from vllm.model_executor.layers.layernorm import RMSNorm
-from vllm.model_executor.layers.linear import ReplicatedLinear
+from vllm.model_executor.layers.linear import ColumnParallelLinear
 from vllm.model_executor.layers.logits_processor import LogitsProcessor
 from vllm.model_executor.layers.vocab_parallel_embedding import (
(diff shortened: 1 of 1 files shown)


drslark (approved):
Clear.
LGTM.

weiguihua2 (approved):

PR #15191, opened 2026-08-28
[Refactor][Attention] Select Ascend CP backends from selector config

pisceskkk:
## What this PR does

- Select GQA, MLA, and SFA context-parallel backend variants from `AttentionSelectorConfig`.
- Keep attention implementation and metadata-builder classes stable across model and KV-cache initialization.
- Preserve DSA backend roles through explicit PCP variants.
- Pass SFA DCP state explicitly through backend resolution.

## Why

Attention backend class factories can be queried at multiple lifecycle phases. Selecting a concrete backend class from selector configuration makes implementation and metadata-builder resolution deterministic, including when multiple configurations coexist in one process.

## Validation

- Targeted Ruff checks for all modified Python files.
- Python byte-compilation for all modified Python files.
- `git diff --check`.

## Test coverage note

Ascend NPU unit tests were not run because a healthy isolated remote session was not available for safe current-tree synchronization.

- vLLM version: v0.27.1
- vLLM main: https://github.com/vllm-project/vllm/commit/ba07e4a48fc951300d97eb506217dd530583dea3

PR #16346, opened 2026-09-11
[Refactor][Attention] Bind Ascend backends at model loading

jiajinzhu2:

Bind SFA, DSA and DCP attention backends when layers are loaded. Draft and target models then keep the correct metadata builder and implementation without reconstructing a draft `VllmConfig`.

For replicated PCP drafts, exclude draft layers from target metadata and build missing replay metadata from the draft model. Existing full-graph metadata still follows the upstream cache-mapping refresh path. The change also reuses initialized DSpark attention groups and resolves DSA cache metadata by layer prefix. Related design: #15191.

N/A


vllm_ascend/platform.py
```diff
@@ -264,19 +268,44 @@ def get_attn_backend_cls(cls, selected_backend, attn_selector_config, num_heads:
...
-                (True, False, True): "vllm_ascend.attention.dsa_v1.AscendDSABackend",
-            }
-            pcp_backend = pcp_backend_map.get(backend_key)
-            if pcp_backend is None:
-                raise NotImplementedError(f"Ascend MRV2 PCP does not support attention backend {backend_key}.")
-            return pcp_backend
-
-        return backend_map[backend_key]
+        use_pcp = attn_selector_config.use_pcp
+        use_dcp = getattr(attn_selector_config, "use_dcp", None)
```
pisceskkk: https://github.com/vllm-project/vllm/blob/v0.29.0/vllm/v1/attention/selector.py#L38

`AttentionSelectorConfig` in v0.29.0 have `use_dcp`, so I think we can remove the following over-defense codes.

vllm_ascend/platform.py
```diff
@@ -264,19 +268,44 @@ def get_attn_backend_cls(cls, selected_backend, attn_selector_config, num_heads:
...
+            # Keep compatibility with selector configs that do not expose DCP.
+            parallel_config = getattr(get_current_vllm_config(), "parallel_config", None)
+            use_dcp = parallel_config is not None and parallel_config.decode_context_parallel_size > 1
+
+        if use_pcp and backend_key not in backend_map:
+            raise NotImplementedError(f"Ascend MRV2 PCP does not support attention backend {backend_key}.")
+        default_backend = backend_map[backend_key]
+        use_dsa_cp = (use_sparse or use_compress) and enable_dsa_cp()
+        if backend_key == (True, True, False):
+            if get_ascend_config().sparse_kv_offload_config.enabled:
```
pisceskkk: We can set up something like `pcp_backend_map` (or just call it `cp_backend_map`), and expand `backend_key` into `cp_backend_key=(*backend_key, use_pcp, use_dcp)`.

Regarding `dsa_cp`, I'd prefer keeping it as a separate conditional branch for now—that way, it'll be much easier to sunset later on.

pisceskkk (commented):
leave some comments, will review the modifications in `vllm_ascend/attention` tmrw.

vllm_ascend/platform.py
jiajinzhu2: Addressed in d59f688b7. `AttentionSelectorConfig.use_dcp` is now read directly; the fallback to `get_current_vllm_config()` and the legacy-config tests were removed.

vllm_ascend/platform.py
jiajinzhu2: Addressed in d59f688b7. Backend selection now uses `cp_backend_key = (*backend_key, use_pcp, use_dcp)` with a `cp_backend_map`, while legacy `dsa_cp` remains in a separate conditional branch.

jiajinzhu2: Follow-up in 5e8928587: legacy `dsa_cp` is driven only by the global `AscendConfig` switch and does not depend on `VllmConfig`, so it no longer has dedicated backend classes. Only PCP/DCP remain represented in `cp_backend_map`; `dsa_cp` stays as a separate compatibility check, and the regular SFA/DSA backends select its builder/impl dynamically.

PR #16349, opened 2026-09-11
[Performance][DCP] Optimize v0.26 DCP decode layouts and fix remap boundaries

foraxe:

Optimize native DCP decode layouts on the pinned `releases/v0.26.0rc` base
`5f12bbf34a530cd6c87463c56404535246af6e94`.

- Keep upstream two-kernel sparse-index remap, with DCP8/interleave128 integer
  boundary and empty-input repairs.
- Pack BF16 output and FP32 LSE as raw bits into one AllToAll, then merge in
  FP32 and convert back to BF16.
- Prepare head-major query gathers and retain native INT8 quantization with
  fused final key/scale stores.
- **New:** expand O/LSE eligibility from `T <= 12` to
  `T <= min(max_num_batched_tokens, 192)`. Query/store limits are unchanged.
  **192 is a manually chosen cap, not a demonstrated kernel limit. Larger
  values have not been tested and retain the upstream fallback.**

The added gain comes from allowing larger eligible calls to use the existing
optimized O/LSE layout, not from increasing the scheduled batch size or
reducing collective count. Both paths already use one AllToAll. No allocator,
memory-placement or experimental VMM peer-exchange change is included.

No new serving API or configuration knob. The existing scheduler token budget
now also bounds optimized O/LSE routing. Unsupported geometry/dtypes retain
upstream fallback. Invalid-shard merge masking preserves upstream nonfinite
handling; integer remap repair remains limited to DCP8/interleave128.

- Fresh focused CPU suite: **95 passed**; routing tests cover scheduler budgets,
  current configuration changes, standalone initialization and fallback.
- A3 direct-helper changing-input graph checks: nine sizes from T13 through
  T192, **1,152 rank-level checks**. Registered-op graph routing/capture checks
  separately cover T48/T192, **128 rank-level checks**.
- Full synthetic decode serving: GLM-5.2 W4A8, 32 A3 NPUs in one superpod,
  DP4/local-DP2/TP8/DCP8/EP32, MTP5 with zero synthetic acceptance,
  zero-filled DecodeBenchConnector KV, full decode graphs, concurrency32,
  50 identical 20K-40K prompts and exactly700 output tokens per request.
  Scheduler budget192; actual vLLM revision `b5e8e2561`.

| Same-image runtime composition | Per-run mean TPOT (ms) | Median of run TPOT (ms) | Median throughput (tok/s) |
|---|---|---:|---:|
| PR15970 default keep-view baseline |114.626,114.192|114.409|221.196|
| Previous combined port, cap12 |114.339,114.006|114.172|221.521|
| Updated combined port, cap192 |105.798,99.110,105.807|105.798|238.331|

All seven fixed-output runs completed **50/50 requests, 35,000 output tokens,
zero recorded errors**, with identical input-length arrays. No repetitions
were discarded. Median-of-runs TPOT improves **7.53% versus PR15970** and
**7.33% versus cap12**; throughput improves **7.75% / 7.59%**, respectively.
The cap12/cap192 manifests differ only in the two O/LSE files. Warmups were
excluded; no audit RPC or profiler ran during included measurements. Imported
source paths/hashes were checked on all32workers before measurement.

Scope: these were image-compatible Python overlays, with PR15970 default
behavior adapted to the image's older config API, not full-worktree installs
or an exact reproduction of a colleague's historical manual merge. Candidate
run variation is larger than baseline variation. These small samples do not
establish confidence intervals, model accuracy, actual shape-frequency counts,
or performance on other workloads.

Known numerical limitation remains: strict CPU-BF16 `atol=rtol=1e-3` checks
fail for both optimized and upstream exchange. No tolerance was relaxed and
no complete numerical acceptance is claimed. The PR's existing review state
is unchanged.

Local publication checks: 95 CPU tests pass. Full `bash format.sh ci` passes
the other hooks, including Gitleaks and ShellCheck, but Actionlint fails when
its macOS ShellCheck subprocesses are terminated; a full CI pass is not claimed.
Raw requests and internal runtime artifacts are not published.
Earlier component results and their boundaries remain documented there.

### DIFF
--- a/tests/ut/attention/a2/test_sfa_dcp_exchange_port.py
+++ b/tests/ut/attention/a2/test_sfa_dcp_exchange_port.py
@@ -1,0 +1,75 @@
+# SPDX-License-Identifier: Apache-2.0
+"""DCP8 exchange comparison; run via torchrun on an isolated eight-NPU group."""
+
+import os
+
+import pytest
+import torch
+import torch.distributed as dist
+
+pytest.importorskip("torch_npu")
+
+from vllm_ascend.ops.triton.sfa_cp import (  # noqa: E402
+    fused_sfa_dcp_lse_combine,
+    pack_sfa_dcp_output_lse,
+    sfa_dcp_a2a_fused_combine,
+)
+
+pytestmark = pytest.mark.skipif(int(os.environ.get("WORLD_SIZE", "1")) != 8, reason="requires isolated DCP8")
+
+
+def test_exchange_matches_upstream_and_reference_with_changed_input_graphs():
+    rank = int(os.environ["RANK"])
+    torch.npu.set_device(int(os.environ["LOCAL_RANK"]))
+    dist.init_process_group("hccl")
+    retained = []
+
+    def inputs(tokens, step):
+        generator = torch.Generator().manual_seed(2026 + step)
+        outputs = torch.randn(8, tokens, 64, 512, generator=generator).to(torch.bfloat16)
+        lses = torch.randn(8, tokens, 64, 1, generator=generator)
+        # All-invalid and mixed-invalid heads cover NaN*0 contamination.
+        lses[:, :, 0] = -torch.inf
+        outputs[:, :, 0] = torch.nan
+        for sender, value in enumerate((torch.nan, torch.inf, -torch.inf)):
+            lses[sender, :, 1] = value
+            outputs[sender, :, 1] = torch.nan
+        return outputs, lses
+
+    for tokens in (1, 6, 12):
+        outputs, lses = inputs(tokens, 0)
+        # Head-major storage exercises a strided token-major view.
+        owner = outputs[rank].transpose(0, 1).contiguous().npu()
+        output = owner.transpose(0, 1)
+        lse = lses[rank].npu()
+
+        def invoke(output=output, lse=lse):
+            return sfa_dcp_a2a_fused_combine(output, lse, 8, 1, dist.group.WORLD)
+
+        for _ in range(5):
+            invoke()
+        torch.npu.synchronize()
+        graph = torch.npu.NPUGraph()
+        with torch.npu.graph(graph):
+            result = invoke()
+        for step in range(8):
+            outputs, lses = inputs(tokens, step)
+            owner.copy_(outputs[rank].transpose(0, 1).contiguous())
+            lse.copy_(lses[rank])
+            # These unchanged upstream helpers bypass the new dispatch guard.
+            send = pack_sfa_dcp_output_lse(output, lse, 8, 1)
+            recv = torch.empty_like(send)
+            dist.all_to_all_single(recv, send)
+            upstream = fused_sfa_dcp_lse_combine(recv, 512, 1)
+            graph.replay()
+            torch.npu.synchronize()
+            finite = torch.isfinite(lses)
+            weights = torch.nan_to_num(torch.softmax(lses.masked_fill(~finite, -torch.inf), dim=0), nan=0.0)
+            expected = (outputs.float().masked_fill(~finite, 0.0) * weights).sum(0)
+            expected = expected[:, rank * 8 : (rank + 1) * 8].to(torch.bfloat16)
+            assert torch.isfinite(result).all()
+            torch.testing.assert_close(result.cpu(), expected, rtol=1e-3, atol=1e-3)
+            torch.testing.assert_close(result, upstream, rtol=1e-3, atol=1e-3)
+        retained.append((graph, result, owner, lse))
+    dist.barrier()
+    dist.destroy_process_group()
--- a/vllm_ascend/ops/triton/sfa_dcp_exchange.py
+++ b/vllm_ascend/ops/triton/sfa_dcp_exchange.py
@@ -1,0 +1,160 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Bit-preserving native-DCP8 output/LSE exchange for small decode batches."""
+
+import torch
+import torch.distributed as dist
+
+from vllm_ascend.ops.triton.sfa_dcp_merge import fused_merge
+
+try:
+    import triton
+    import triton.language as tl
+except ImportError:
+    triton = None
+    tl = None
+
+
+_RANKS = 8
+_HEADS = 64
+_HEAD_DIM = 512
+_OUTPUT_WORDS_PER_PEER = 2048
+_WORDS_PER_PEER = 2056
+_MAX_DECODE_TOKENS = 12
+
+
+if triton is not None:
+
+    @triton.jit
+    def _pack_t1_peer(output_words, lse, send, lse_head_stride: tl.constexpr):
+        peer = tl.program_id(0)
+        word = tl.arange(0, 2048)
+        values = tl.load(output_words + peer * 2048 + word, mask=word < 2048, other=0)
+        tl.store(send + peer * 2056 + word, values, mask=word < 2048)
+        local_head = tl.arange(0, 8)
+        stats = tl.load(
+            lse + (peer * 8 + local_head) * lse_head_stride,
+            mask=local_head < 8,
+            other=0.0,
+        )
+        tl.store(
+            send + peer * 2056 + 2048 + local_head,
+            stats.to(tl.int32, bitcast=True),
+            mask=local_head < 8,
+        )
+
+    @triton.jit
+    def _pack_token_peer(
+        output,
+        lse,
+        send,
+        output_s0: tl.constexpr,
+        output_s1: tl.constexpr,
+        lse_s0: tl.constexpr,
+        lse_s1: tl.constexpr,
+        tokens: tl.constexpr,
+        local_heads: tl.constexpr,
+        head_words: tl.constexpr,
+        stats_tile: tl.constexpr,
+    ):
+        peer = tl.program_id(0)
+        output_words: tl.constexpr = local_heads * head_words
+        peer_words: tl.constexpr = tokens * (output_words + local_heads)
+        word = tl.arange(0, output_words)
+        local_head = word // head_words
+        dim_word = word % head_words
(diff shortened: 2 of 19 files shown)


vllm_ascend/ops/triton/sfa_dcp_exchange.py
```diff
@@ -0,0 +1,162 @@
...
+_OUTPUT_WORDS_PER_PEER = 2048
+_WORDS_PER_PEER = 2056
+# Manually chosen conservative cap, not a demonstrated kernel limit.
+# Inputs above 192 have not been tested and retain the upstream path.
+_MAX_DECODE_TOKENS = 192
+
+
+if triton is not None:
+
+    @triton.jit
```
ZT-AIA, line 29: Please place the single-operator precision test case here. https://github.com/vllm-project/vllm-ascend/tree/main/tests/e2e/nightly/single_node/ops/singlecard_ops/triton

foraxe: Thanks for the review. Working on it.

foraxe:
Moved the single-card operator precision tests in `a78c262be` to the requested directory:

- `tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_sfa_indexer_store_triton.py`
- `tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_sfa_dcp_remap.py`

Both are 100% content-preserving moves; no runtime code, assertions or tolerances changed. Query-gather/O-LSE tests require an eight-NPU group and remain distributed tests, not single-card cases. Both pre-commit stages passed. No new NPU execution is claimed for relocation.

vllm_ascend/ops/triton/sfa_dcp_exchange.py
foraxe: Hi @ZT-AIA , i just moved 'single-operator precision test case' parts. Should others be moved too? 
I noticed the pre-commit errors: 
> Error: vllm_ascend/ops/triton/sfa_indexer_store.py:12: error: Skipping analyzing "triton": module is installed, but missing library stubs or py.typed marker  [import-untyped]

tests/ut/attention/a2/test_sfa_dcp_exchange_port.py
```diff
@@ -0,0 +1,149 @@
...
+        local_rank=rank,
+        backend="hccl",
+        group_name="large_dcp8_regression",
+        use_device_communicator=False,
+    )
+    retained = []
+    for tokens in (48, 192):
+        base = (torch.arange(tokens * 64 * 512).reshape(tokens, 64, 512) % 31 - 15).float() / 32
+        output = (base + rank / 16).to(torch.bfloat16).npu()
+        lse = torch.zeros((tokens, 64, 1), device="npu", dtype=torch.float32)
```
AuroraEmiya, line 120: The new fast-path eligibility extends from `T <= 12` to `T <= 192`, but this regression only covers uniform LSE at T48/T192. That validates packing/layout/graph replay, not general merge numerics. Could we add at least one T>12 non-uniform-LSE case (preferably covering the new boundary, e.g. T13/T191/T192, with mixed finite/-inf ranks) against an independent FP32/FP64 reference? Since the PR already documents the strict random-BF16 limitation, the acceptance tolerance should be stated explicitly.

vllm_ascend/ops/triton/sfa_dcp_exchange.py
```diff
@@ -0,0 +1,162 @@
...
+    def _pack_token_peer(
+        output,
+        lse,
+        send,
+        output_s0: tl.constexpr,
+        output_s1: tl.constexpr,
+        lse_s0: tl.constexpr,
+        lse_s1: tl.constexpr,
+        tokens: tl.constexpr,
```
AuroraEmiya, line 56: `tokens` now ranges up to 192, but it is `tl.constexpr` and also drives `range(tokens)`, so different T values can produce distinct statically specialized/unrolled kernels. Have we measured compile-cache/code-size and first-call behavior for realistic mixed-T decode buckets? Please add a small T sweep, or consider a runtime/`do_not_specialize` form if T does not need to change the static tile structure.

vllm_ascend/ops/triton/sparse_index_remap.py
```diff
@@ -53,18 +54,20 @@ def remap_sparse_indices_fused_kernel(
...
-    # identically on NPU.
-    block_idx = idx // interleave_size
-    owner = block_idx - (block_idx // dcp_size) * dcp_size
-    valid = (idx >= 0) & (owner == dcp_rank)
-
-    if INTERLEAVE_ONE:
-        remapped = idx // dcp_size
+    # This backend's vector divsi loses integer boundaries above 2**24.
+    # Keep the native DCP8/interleave128 path entirely in bit operations.
+    if EXACT_DCP8:
```
AuroraEmiya, line 59: The exact repair is currently limited to DCP8/interleave128, while the generic branch below still uses vector integer division—the operation already shown to lose integer boundaries above `2**24` on this backend. Since the public wrapper still accepts arbitrary `dcp_size`/`interleave_size`, could unsupported configurations fall back to a verified implementation, or could the exact mapping be generalized? Otherwise known-risk arithmetic remains reachable.

vllm_ascend/ops/triton/sfa_dcp_merge.py
```diff
@@ -0,0 +1,157 @@
...
+            head = row % head_count
+            if native_layout:
+                lse_offsets = rank_ids * lse_s0 + head * lse_s1 + token * lse_s2
+            else:
+                lse_offsets = rank_ids * lse_s0 + token * lse_s1 + head * lse_s2
+            raw_lse = tl.load(lse_ptr + lse_offsets).to(tl.float32)
+            finite = (raw_lse == raw_lse) & (raw_lse > float("-inf")) & (raw_lse < float("inf"))
+            safe_lse = tl.where(finite, raw_lse, float("-inf"))
+            max_lse = tl.max(safe_lse, axis=0)
+            shifted = tl.where(finite, safe_lse - max_lse, 0.0)
```
AuroraEmiya, line 85: For an all-invalid rank set, `safe_lse` and `max_lse` are both `-inf`, so this still forms `-inf - (-inf)` before `tl.where` discards the lane. The final result may be zero today, but that relies on masking an invalid intermediate after it is created. Could we derive `has_valid`/a finite `safe_max` so the subtraction itself never produces NaN? That is more robust to backend lowering.

vllm_ascend/ops/triton/sfa_dcp_merge.py
```diff
@@ -0,0 +1,157 @@
...
+        native_layout = True
+    else:
+        _, tokens, heads, head_dim = out_recv.shape
+        native_layout = False
+    result = torch.empty((tokens, heads, head_dim), dtype=torch.float32, device=out_recv.device)
+    # Rank-wise accumulation avoids the failed R×D materialization. D512 now
+    # holds one FP32 accumulator plus one rank vector; retry verifies compiler UB.
+    block_d = min(triton.next_power_of_2(head_dim), 512)
+    tiles = tokens * heads * math.ceil(head_dim / block_d)
+    properties = triton.runtime.driver.active.utils.get_device_properties(out_recv.device.index)
```
AuroraEmiya, line 139: Could this use the repository's centralized VectorCore capability helper (`get_vectorcore_num()` / initialization path) instead of reading driver properties directly? The same direct property access also appears in `sfa_dcp_query.py`. Keeping device probing in one compatibility layer avoids duplicated backend assumptions; the returned core count can still be used as a grid candidate.

AuroraEmiya (commented):
Here are something you might need to notice : )

AuroraEmiya (changes requested):
Thanks for the measurements. The code-size result answers the loop-unrolling concern, but it does not close the review point yet.

For Triton kernels in this project, we use the following parameter convention:

* `tl.constexpr` is reserved for values that are genuinely part of the compile-time kernel configuration or static IR structure, such as fixed tile sizes, static feature/layout switches, or dimensions required by `tl.arange` / `tl.static_range`.
* Compile-time parameters must use uppercase names, e.g. `BLOCK_D`, `RANKS`, `LOCAL_HEADS`, `STATS_TILE`.
* Values that vary with runtime workload, tensor shape/layout, or allocation state should remain normal runtime parameters unless there is measured evidence that specializing them is necessary. Examples include token count, tensor strides, total work items, program count, and cache capacity.
* A shape-derived value should not be marked `tl.constexpr` merely because Triton permits specialization.

With that convention, parameters such as `tokens`, the stride arguments, `total_tiles`, `num_programs`, and similar shape-derived values across these new kernels need to be reviewed. `rows` in `_fused_merge_kernel` also appears to be unused and should be removed rather than becoming another specialization key.

The cold-cache numbers make this more important, not less. The current sweep changes T and therefore creates different specializations. Thirteen pack-kernel cold first calls cost 18.511 s in total. Even after excluding backend initialization, sub-second/second-scale compilation for new shapes is significant compared with seconds-scale model decoding. Also, this measurement covers only the pack kernel; the merge path contains additional shape-derived `tl.constexpr` parameters, so it does not yet establish the cold compilation cost of the complete optimized path.

Therefore, keeping runtime-varying parameters as `tl.constexpr` without a demonstrated need for specialization would not pass review from my side.

Could you please:

1. Audit all new `tl.constexpr` parameters and classify them into true compile-time constants versus runtime-varying values.
2. Rename retained compile-time constants to uppercase.
3. Remove unnecessary/unused constexpr parameters.
4. For any runtime-varying parameter that is intentionally kept as `tl.constexpr`, provide an A/B showing that specialization gives a meaningful steady-state performance benefit over the runtime form.
5. Measure cold first-call/specialization cost for the complete optimized pack → exchange → merge path, not only the pack kernel.

The fact that T192 retains an `scf.for` is useful—it shows there is no token-proportional loop unrolling—but the remaining question is whether T and the other shape-derived values should be specialization keys at all.

PR #16350, opened 2026-09-11
[Performance][DCP] Refresh native decode layouts and remap fixes for main

foraxe:

Continue #16349 on main, refreshed onto
`d052552072254a21560e8c5ebc04b80524c89622` without rewriting the published
branch history. This replaces the old fixed-main-base port and resolves its conflicts.

- Preserve upstream two-kernel sparse-index remap, with DCP8/interleave128
  integer-boundary repairs and an empty-input early return.
- Raw-bit BF16 output/FP32 LSE packing, one AllToAll, FP32 merge and BF16
  conversion for eligible native head-scatter DCP8 calls.
- Bring #16349's O/LSE eligibility expansion to
  `T <= min(max_num_batched_tokens, 192)`. **192 is a manually selected cap,
  not a demonstrated kernel limit. Larger values are untested and fall back.**
- Retain head-major query preparation and native INT8 quantization with fused
  final indexer key/scale writes; their smaller eligibility limits are unchanged.

Current-main integration preserves indexer-owned RoPE metadata and projection
weights reuse. PCP, token-scatter, returned-LSE and deferred-combine modes keep
their upstream implementations. No allocator, memory-placement, SuperMem
serving configuration or experimental VMM peer path is included.

Both exchange implementations already use one AllToAll. This change does not
increase the scheduler's token budget or claim a reduction in collective count.

No new serving API or configuration knob. Main no longer retains `vllm_config`
on `AscendConfig`, so the attention caller explicitly passes its current
scheduler budget through a trailing optional `decode_token_budget` argument
on the registered operator and its FakeTensor implementation. Existing callers
may omit it; standalone helpers retain the manual cap.

Unsupported layouts/dtypes preserve upstream fallback, group validation and
invalid-shard behavior. Query/store limits and numerical thresholds are unchanged.

#### Test placement and checks

Single-card Triton precision tests now live in the requested directory:
`tests/e2e/nightly/single_node/ops/singlecard_ops/triton/`:

- `test_sfa_indexer_store_triton.py`: native INT8 key/scale byte precision and changing ACLGraph slots.
- `test_sfa_dcp_remap.py`: integer-boundary oracle, changing graph inputs and empty-input cases.

Both are byte-identical moves, with unchanged numerical assertions/tolerances.
Query gather and O/LSE exchange tests still require a real eight-NPU DCP group;
they remain under `tests/ut/attention/a2/`, not in the single-card suite.

After relocation: **101 focused CPU tests passed; all-file pre-commit and manual
stages passed**. Runtime code is unchanged. No new NPU rerun is claimed for the
path-only move; the device evidence below names the actual tested revision.

#### Fresh NPU and real-weight integration evidence

- At `222262c17`: DCP8 query gather/unpack changing-input graphs; single-card
  indexer store (48 replays) and remap (27 cases, 192 replays plus 3 empty cases).
- At runtime source **`c1e82ee97510f7805a6eef7663c53e7e3afacfe3`**: registered
  DCP8 O/LSE T48/T192 capture/replay on all eight ranks, including changing inputs
  and 128 rank/shape/replay checks against an exact uniform-LSE oracle.
- Reduced-model HTTP E2E exposed the main-refresh indexer metadata issue:
  `num_decode_tokens` was populated only for PCP, disabling native DCP fused store.
  Fixed in c1e82ee97; 35 indexer unit tests passed, then 9 HTTP requests/576 tokens
  passed with request-time and changed-input optimized-graph evidence.
- **Full 78-layer GLM-5.2 W4A8C8 trained weights**, including MoE, TP8/DCP8/EP8/DP1,
  no MTP/offload: three 1240-token prompts completed normally, generated 64/85/66
  tokens and correctly answered Beijing/Paris/Tokyo with coherent introductions.
  All eight ranks recorded 66 changed-input optimized-graph replays. Source hashes
  matched c1e82ee97. The current follow-up changes tests/docs only.

#### Fresh performance measurements on c1e82ee97

Registered O/LSE: same inputs, ON vs existing fallback OFF, 10 alternating
blocks ×100 graph replays; median of each block's slowest rank:

| T | OFF | ON | Effective replay latency reduction |
|---|---:|---:|---:|
|48|206.69 us|101.21 us|51.03%|
|192|569.19 us|220.02 us|61.35%|

Full trained-weight HTTP: **one completed OFF/ON pair**, 1920 input +128 output
tokens, no MTP/offload/prefix cache. OFF disables only query/store/O-LSE performance
eligibility on the same source; remap correctness fixes remain in both arms.
Both use scheduler budget4096, maxseq192 and KV6GiB. Optimization cap remains192.

| Concurrency | OFF mean TPOT(ms) | ON mean TPOT(ms) | Reduction | OFF / ON output throughput(tok/s) |
|---|---:|---:|---:|---:|
|1|48.74|47.46|2.63%|18.46 /18.88|
|4|62.56|60.05|4.01%|51.65 /53.25|
|192|655.25|635.78|2.97%|157.93 /161.34|

C192 cohort TPOT includes prefill/batch-filling interference. Within the common
interval after all192 first tokens arrived and before any request's last token,
median client inter-token latency was **173.51 ->144.80ms (-16.54%)**.
The windows were5.90/4.92s, extracted identically from raw SSE; this is not an
independent repeat, isolated kernel time or overall-request TPOT reduction.

Each rank verified34 actual unpadded T192 optimized replays in the ON measurement.
All197 measured requests per arm completed with128 output tokens, normal length
termination and zero preemptions. All197 paired token sequences matched.
The user requested stopping after one pair; later incomplete work is excluded.

#### Evidence limits

- 192 is manually selected; larger values are untested. Query/store retain their
  smaller bounds; the optimized path verified at T192 is O/LSE.
- Uniform-LSE graph checks do not resolve historical strict random CPU-BF16
  `atol=rtol=1e-3` failures in both compared exchanges. Thresholds remain unchanged;
  finite nonuniform-LSE timing inputs do not establish general precision acceptance.
- Serving used an isolated empty-target vLLM build at verified pin84030bbe,
  Torch2.10.0/torch-npu2.10.0.post4, Triton-Ascend3.2.2, CANN9.1.0 and image native
  libraries. FastAPI0.133.0/Starlette1.0.1 satisfy vLLM but conflict with Ascend's
  existing declaration; full pip check remains non-green. Not default-install acceptance.
- Single-pair HTTP results do not establish cross-run variance or confidence.
  No broad accuracy, 1M or SuperMem ON/OFF performance claim.
- #16349's synthetic release-image TPOT evidence is **reused release evidence**,
  not these main measurements. Its setup and numbers remain documented in `DCP_PORT.md`.

Evidence threads: [NPU](https://github.com/vllm-project/vllm-ascend/pull/16350#issuecomment-5759854639),
[reduced E2E](https://github.com/vllm-project/vllm-ascend/pull/16350#issuecomment-5763308847),
[full weights](https://github.com/vllm-project/vllm-ascend/pull/16350#issuecomment-5764536283),
[performance](https://github.com/vllm-project/vllm-ascend/pull/16350#issuecomment-5770334097).

### DIFF
--- a/tests/ut/attention/a2/test_sfa_dcp_exchange_port.py
+++ b/tests/ut/attention/a2/test_sfa_dcp_exchange_port.py
@@ -1,0 +1,75 @@
+# SPDX-License-Identifier: Apache-2.0
+"""DCP8 exchange comparison; run via torchrun on an isolated eight-NPU group."""
+
+import os
+
+import pytest
+import torch
+import torch.distributed as dist
+
+pytest.importorskip("torch_npu")
+
+from vllm_ascend.ops.triton.sfa_cp import (  # noqa: E402
+    fused_sfa_dcp_lse_combine,
+    pack_sfa_dcp_output_lse,
+    sfa_dcp_a2a_fused_combine,
+)
+
+pytestmark = pytest.mark.skipif(int(os.environ.get("WORLD_SIZE", "1")) != 8, reason="requires isolated DCP8")
+
+
+def test_exchange_matches_upstream_and_reference_with_changed_input_graphs():
+    rank = int(os.environ["RANK"])
+    torch.npu.set_device(int(os.environ["LOCAL_RANK"]))
+    dist.init_process_group("hccl")
+    retained = []
+
+    def inputs(tokens, step):
+        generator = torch.Generator().manual_seed(2026 + step)
+        outputs = torch.randn(8, tokens, 64, 512, generator=generator).to(torch.bfloat16)
+        lses = torch.randn(8, tokens, 64, 1, generator=generator)
+        # All-invalid and mixed-invalid heads cover NaN*0 contamination.
+        lses[:, :, 0] = -torch.inf
+        outputs[:, :, 0] = torch.nan
+        for sender, value in enumerate((torch.nan, torch.inf, -torch.inf)):
+            lses[sender, :, 1] = value
+            outputs[sender, :, 1] = torch.nan
+        return outputs, lses
+
+    for tokens in (1, 6, 12):
+        outputs, lses = inputs(tokens, 0)
+        # Head-major storage exercises a strided token-major view.
+        owner = outputs[rank].transpose(0, 1).contiguous().npu()
+        output = owner.transpose(0, 1)
+        lse = lses[rank].npu()
+
+        def invoke(output=output, lse=lse):
+            return sfa_dcp_a2a_fused_combine(output, lse, 8, 1, dist.group.WORLD)
+
+        for _ in range(5):
+            invoke()
+        torch.npu.synchronize()
+        graph = torch.npu.NPUGraph()
+        with torch.npu.graph(graph):
+            result = invoke()
+        for step in range(8):
+            outputs, lses = inputs(tokens, step)
+            owner.copy_(outputs[rank].transpose(0, 1).contiguous())
+            lse.copy_(lses[rank])
+            # These unchanged upstream helpers bypass the new dispatch guard.
+            send = pack_sfa_dcp_output_lse(output, lse, 8, 1)
+            recv = torch.empty_like(send)
+            dist.all_to_all_single(recv, send)
+            upstream = fused_sfa_dcp_lse_combine(recv, 512, 1)
+            graph.replay()
+            torch.npu.synchronize()
+            finite = torch.isfinite(lses)
+            weights = torch.nan_to_num(torch.softmax(lses.masked_fill(~finite, -torch.inf), dim=0), nan=0.0)
+            expected = (outputs.float().masked_fill(~finite, 0.0) * weights).sum(0)
+            expected = expected[:, rank * 8 : (rank + 1) * 8].to(torch.bfloat16)
+            assert torch.isfinite(result).all()
+            torch.testing.assert_close(result.cpu(), expected, rtol=1e-3, atol=1e-3)
+            torch.testing.assert_close(result, upstream, rtol=1e-3, atol=1e-3)
+        retained.append((graph, result, owner, lse))
+    dist.barrier()
+    dist.destroy_process_group()
--- a/vllm_ascend/ops/triton/sfa_dcp_exchange.py
+++ b/vllm_ascend/ops/triton/sfa_dcp_exchange.py
@@ -1,0 +1,160 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Bit-preserving native-DCP8 output/LSE exchange for small decode batches."""
+
+import torch
+import torch.distributed as dist
+
+from vllm_ascend.ops.triton.sfa_dcp_merge import fused_merge
+
+try:
+    import triton
+    import triton.language as tl
+except ImportError:
+    triton = None
+    tl = None
+
+
+_RANKS = 8
+_HEADS = 64
+_HEAD_DIM = 512
+_OUTPUT_WORDS_PER_PEER = 2048
+_WORDS_PER_PEER = 2056
+_MAX_DECODE_TOKENS = 12
+
+
+if triton is not None:
+
+    @triton.jit
+    def _pack_t1_peer(output_words, lse, send, lse_head_stride: tl.constexpr):
+        peer = tl.program_id(0)
+        word = tl.arange(0, 2048)
+        values = tl.load(output_words + peer * 2048 + word, mask=word < 2048, other=0)
+        tl.store(send + peer * 2056 + word, values, mask=word < 2048)
+        local_head = tl.arange(0, 8)
+        stats = tl.load(
+            lse + (peer * 8 + local_head) * lse_head_stride,
+            mask=local_head < 8,
+            other=0.0,
+        )
+        tl.store(
+            send + peer * 2056 + 2048 + local_head,
+            stats.to(tl.int32, bitcast=True),
+            mask=local_head < 8,
+        )
+
+    @triton.jit
+    def _pack_token_peer(
+        output,
+        lse,
+        send,
+        output_s0: tl.constexpr,
+        output_s1: tl.constexpr,
+        lse_s0: tl.constexpr,
+        lse_s1: tl.constexpr,
+        tokens: tl.constexpr,
+        local_heads: tl.constexpr,
+        head_words: tl.constexpr,
+        stats_tile: tl.constexpr,
+    ):
+        peer = tl.program_id(0)
+        output_words: tl.constexpr = local_heads * head_words
+        peer_words: tl.constexpr = tokens * (output_words + local_heads)
+        word = tl.arange(0, output_words)
+        local_head = word // head_words
+        dim_word = word % head_words
+        peer_base = peer * peer_words
+        for token in range(tokens):
+            values = tl.load(output + token * output_s0 + (peer * local_heads + local_head) * output_s1 + dim_word)
+            tl.store(send + peer_base + token * output_words + word, values)
+        index = tl.arange(0, stats_tile)
+        token, head = index // local_heads, index % local_heads
+        stats = tl.load(
+            lse + token * lse_s0 + (peer * local_heads + head) * lse_s1,
+            mask=index < tokens * local_heads,
+            other=0.0,
+        )
+        tl.store(
+            send + peer_base + tokens * output_words + index,
+            stats.to(tl.int32, bitcast=True),
+            mask=index < tokens * local_heads,
+        )
+
+
+def can_exchange(output: torch.Tensor, lse: torch.Tensor) -> bool:
+    # These properties are uniform within a valid DCP group. Do not choose
+    # different collective counts based on rank-local allocation/stride state.
+    return (
+        triton is not None
+        and output.device.type == "npu"
+        and output.dtype == torch.bfloat16
+        and output.ndim == 3
+        and 0 < output.shape[0] <= _MAX_DECODE_TOKENS
+        and tuple(output.shape[1:]) == (_HEADS, _HEAD_DIM)
+        and lse.device == output.device
+        and lse.dtype == torch.float32
+        and tuple(lse.shape) == (output.shape[0], _HEADS, 1)
+    )
+
+
+def _exchange_tokens(output: torch.Tensor, lse: torch.Tensor, group) -> torch.Tensor:
+    if not can_exchange(output, lse) or dist.get_world_size(group) != _RANKS:
+        raise ValueError("Unsupported native-DCP8 output/LSE exchange")
+    tokens = output.shape[0]
+    local_heads = _HEADS // _RANKS
+    # Stride-dependent normalization stays inside the one-collective path.
+    if output.stride(-1) != 1 or any(stride % 2 for stride in output.stride()[:2]):
+        output = output.contiguous()
+    if output.storage_offset() % 2:
+        output = output.clone()
+    words = output.view(torch.int32)
+    peer_words = tokens * _WORDS_PER_PEER
+    send = torch.empty((_RANKS, peer_words), dtype=torch.int32, device=output.device)
+    _pack_token_peer[(_RANKS,)](
+        words,
+        lse,
+        send,
+        *words.stride()[:2],
+        *lse.stride()[:2],
+        tokens,
+        local_heads,
+        _HEAD_DIM // 2,
+        triton.next_power_of_2(tokens * local_heads),
+    )
+    recv = torch.empty_like(send)
+    dist.all_to_all_single(recv, send, group=group)
+    # Final receiver layout is token-major: no transpose or receive-side copy.
+    parts = recv.view(output.dtype).as_strided(
+        (_RANKS, tokens, local_heads, _HEAD_DIM),
+        (peer_words * 2, local_heads * _HEAD_DIM, _HEAD_DIM, 1),
+    )
+    stats = recv.view(torch.float32).as_strided(
+        (_RANKS, tokens, local_heads),
+        (peer_words, local_heads, 1),
+        storage_offset=tokens * _OUTPUT_WORDS_PER_PEER,
+    )
+    return fused_merge(parts, stats, token_dim=1)
+
+
+def exchange(output: torch.Tensor, lse: torch.Tensor, group) -> torch.Tensor:
+    """Exchange eight head shards, retaining FP32 LSE bits and merge math."""
+    if output.shape[0] != 1:
+        return _exchange_tokens(output, lse, group)
+    output = output.contiguous()
+    if output.storage_offset() % 2:
+        output = output.clone()
+    lse = lse.contiguous()
+    words = output.view(torch.int32)
+    send = torch.empty((_RANKS, _WORDS_PER_PEER), dtype=torch.int32, device=output.device)
+    _pack_t1_peer[(_RANKS,)](words, lse, send, lse.stride(1))
+    recv = torch.empty_like(send)
+    dist.all_to_all_single(recv, send, group=group)
+    parts = recv.view(output.dtype).as_strided(
+        (_RANKS, _HEADS // _RANKS, 1, _HEAD_DIM),
+        (_WORDS_PER_PEER * 2, _HEAD_DIM, _HEAD_DIM, 1),
+    )
+    stats = recv.view(torch.float32).as_strided(
+        (_RANKS, _HEADS // _RANKS, 1),
+        (_WORDS_PER_PEER, 1, 1),
+        storage_offset=_OUTPUT_WORDS_PER_PEER,
+    )
+    return fused_merge(parts, stats, token_dim=2)
--- a/vllm_ascend/ops/triton/sfa_dcp_merge.py
+++ b/vllm_ascend/ops/triton/sfa_dcp_merge.py
@@ -1,0 +1,157 @@
+# SPDX-License-Identifier: Apache-2.0
+"""FP32 local merge for the small-batch DCP exchange."""
+
+from __future__ import annotations
+
+import math
+
+import torch
+
+try:
+    import triton
+    import triton.language as tl
+    from triton.language.extra.cann import extension
+
+    get_element = extension.get_element
+except ImportError:  # CPU-only test environments retain exact torch semantics.
+    triton = None
+    tl = None
+    get_element = None
+
+
+_SUPPORTED_RANKS = (2, 8)
+_SUPPORTED_DTYPES = (torch.bfloat16, torch.float16, torch.float32)
+
+
+def torch_merge(out_recv: torch.Tensor, lse_recv: torch.Tensor, token_dim: int) -> torch.Tensor:
+    """Reference merge with upstream invalid-rank masking semantics."""
+    if out_recv.ndim != 4 or lse_recv.ndim != 3 or out_recv.shape[:3] != lse_recv.shape:
+        raise RuntimeError(
+            "DCP output merge expects matching rank/token/head dimensions, "
+            f"got {tuple(out_recv.shape)} and {tuple(lse_recv.shape)}."
+        )
+    if token_dim not in (1, 2):
+        raise RuntimeError(f"DCP output merge token_dim must be 1 or 2, got {token_dim}.")
+    finite = torch.isfinite(lse_recv)
+    lse = lse_recv.masked_fill(~finite, float("-inf"))
+    weights = torch.nan_to_num(torch.softmax(lse, dim=0), nan=0.0)
+    values = out_recv.to(lse.dtype).masked_fill(~finite.unsqueeze(-1), 0.0)
+    output = (values * weights.unsqueeze(-1)).sum(dim=0)
+    return output.movedim(token_dim - 1, 0).contiguous()
+
+
+if triton is not None:
+
+    @triton.jit
+    def _fused_merge_kernel(
+        out_ptr,
+        lse_ptr,
+        result_ptr,
+        out_s0: tl.constexpr,
+        out_s1: tl.constexpr,
+        out_s2: tl.constexpr,
+        out_s3: tl.constexpr,
+        lse_s0: tl.constexpr,
+        lse_s1: tl.constexpr,
+        lse_s2: tl.constexpr,
+        result_s0: tl.constexpr,
+        result_s1: tl.constexpr,
+        result_s2: tl.constexpr,
+        rows: tl.constexpr,
+        head_count: tl.constexpr,
+        head_dim: tl.constexpr,
+        total_tiles: tl.constexpr,
+        num_programs: tl.constexpr,
+        ranks: tl.constexpr,
+        native_layout: tl.constexpr,
+        block_d: tl.constexpr,
+    ):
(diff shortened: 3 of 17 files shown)


Tflowers-0129:
This PR changes several NPU-native DCP8 decode paths, including the registered O/LSE exchange up to T=192, but the current main-refresh commit has not been executed on NPU. Please provide at least one fresh DCP8 NPU run on the current head, especially T=48/T=192 with ACLGraph capture/replay and changing inputs. Reused #16349 results are useful performance evidence, but they do not validate this refreshed integration.

foraxe:
Fresh NPU validation is now complete on the **unchanged current head
`222262c175cbbcafb45b8429d89b28f0aa650eb7`**, not reused from #16349.

The current-head test
`test_large_batch_registered_graph_matches_exact_uniform_merge` passed on
all **8 NPU ranks**:

- **T48 and T192**,5 warmups per shape, ACLGraph capture and8 changing-input replays.
- Actual `torch.ops.vllm.sfa_dcp_a2a_fused`, with `decode_token_budget=192`.
- The test verifies capture enters the optimized exchange rather than fallback.
- **128 rank × shape × replay checks**, with BF16 raw-bit equality to the exact oracle.
- Both O and LSE buffers are updated before replay. LSE values remain equal across
  ranks in this test, so weights are uniform; this is not arbitrary nonuniform-LSE coverage.

Additional unchanged current-head device tests also passed:

| Test | Execution | Result |
|---|---|---|
| Q gather/unpack |8 ranks; T2/6/12;16 changing-input replays each|8 rank-local pytest passes,384 replay checks|
| Native quantized indexer store |1 NPU; rows1/6/12; changing keys/scales/slots and sentinels|1 pytest pass,48 replay checks|
| Integer remap |1 NPU;8 logical-rank cases ×3 row counts, plus empty inputs|27 pytest passes,192 replays +3 empty-input cases|

Runtime: A3,8 exclusively allocated dies; Python3.12.13, Torch2.10.0+cpu with
torch-npu2.10.0.post4, Triton-Ascend3.2.2, CANN9.1.0, driver25.5.1.1,
vLLM distribution metadata `0.28.0+empty`. The full PR source tree was used
directly. Per-rank imported runtime/test Git blobs and SHA256 values were
checked against the current head; no runtime/test source substitutions or
numerical-threshold changes were used.

The image-generated `_build_info.py` contains only `__device_type__ = 'A3'`
and was separately path/hash-verified. An initial audit-only rejection of this
generated metadata is retained as an invalid audit attempt; the successful
results above are from a subsequent fresh run after correcting only the audit.

Core reproduction from this commit:

```bash
OMP_NUM_THREADS=1 HCCL_OP_EXPANSION_MODE=AIV \
HCCL_CONNECT_TIMEOUT=120 HCCL_EXEC_TIMEOUT=120 \
python -m torch.distributed.run --standalone --nproc-per-node=8 \
  -m pytest --noconftest -s -q \
  tests/ut/attention/a2/test_sfa_dcp_exchange_port.py::test_large_batch_registered_graph_matches_exact_uniform_merge
```

All four valid test launches exited0; every allocated device was healthy and
empty after each launch. Registered-test log SHA256:
`e8d6bafd90830d01d2986d7232a0370c2fe9697ee9b47e3e6028aaab103eb763`.
Current operator source SHA256 (`vllm_ascend/ops/triton/sfa_cp.py`):
`eeef09e093f36ce19272ee0c586f14feb550b8798d7690cd6c4aff10c48ebc14`.

This closes the requested fresh registered-op/DCP8 ACLGraph execution gap.
It does **not** claim fresh full-model serving/performance validation or
complete numerical acceptance. The existing strict random-BF16 limitation
remains documented; the exact uniform oracle does not replace that test.

foraxe:
Fresh reduced-model HTTP E2E now passes on the source committed as
`c1e82ee97510f7805a6eef7663c53e7e3afacfe3`.

The E2E caught a real main-refresh integration bug: the indexer metadata builder
populated `num_decode_tokens` only for PCP, leaving native DCP at zero and silently
excluding the fused store. This commit extends the existing boundary calculation
to DCP, with mixed/prefill and pure-decode/graph-capture regressions. No kernel
numerical implementation changed.

- A3, TP8/DCP8/PCP1, two dense GLM-DSA layers, random BF16 weights, int8 indexer KV.
- Real HTTP prefill/decode with computed KV; no MTP, KV connector, offload or prefix cache.
- Input lengths 1033–1057; cohorts C1, C4, then C4 with different token IDs.
- 9/9 requests completed, 64 tokens each (576 total), normal length termination,
  nonempty outputs and finite logprobs.
- FULL_DECODE_ONLY capture buckets 1/2/4/8. Every rank recorded 116 request-time
  replays of graphs capturing all four optimized paths (query/store/remap/O-LSE),
  including 58 during the changed-input cohort. Observer wrappers delegated
  unchanged tensors; 12 loaded module hashes per rank match the final commit.
- Also reran the registered T48/T192 changing-input ACLGraph test on the fixed
  source: all 8 ranks passed (128 rank/shape/replay checks, uniform-LSE oracle).
- Indexer unit tests: 35 passed. Full pre-commit and manual stages passed.
- Allocated devices healthy and empty after execution; unrelated serving untouched.

Runtime caveat: isolated venv uses this branch's verified vLLM pin
`84030bbe3d74d99bad477a3d2e37a973ccd8865c` (empty-target build), FastAPI0.133.0
and Starlette1.0.1, plus the image's existing custom-op vendor library.
The FastAPI requirements of pinned vLLM and Ascend metadata conflict, so this is
**not** a default-install/dependency-resolution acceptance; full pip check remains
non-green. No global runtime or PR dependency declaration was changed.

Scope: random-weight integration smoke, **not** trained-model answer quality,
MoE, full production model, SuperMem/1M performance, or a throughput measurement.
T48/T192 coverage is the separate distributed operator test, not HTTP batch sizes.
The 192 cap remains manually selected; larger values have not been tested.

foraxe:
Additional **full real-weight GLM-5.2 HTTP E2E PASS** on unchanged head
`c1e82ee97510f7805a6eef7663c53e7e3afacfe3`.

- Full 78-layer GLM-5.2 W4A8C8 checkpoint (~390.56 GiB), including MoE;
  TP8/DCP8/EP8/DP1, PCP1. No reduced layers or dummy weights.
- Real prefill/KV/decode; no MTP, offload, KV connector or prefix caching.
- One request followed by two concurrent different prompts, each1240 input tokens.
  All3 completed normally (`stop`), generating64/85/66 tokens (215 total).
  Responses correctly named Beijing, Paris and Tokyo and gave coherent introductions.
- FULL_DECODE_ONLY buckets1/2/4. All8 ranks independently recorded66 changed-input
  replays of graphs capturing query gather, fused indexer store, sparse remap and
  O/LSE exchange. Loaded module hashes match the current commit.
- All8 allocated dies healthy and empty after the run; unrelated services untouched.

Same isolated runtime/dependency caveat as the previous smoke: verified vLLM
pin84030bbe, image native libraries, FastAPI0.133.0/Starlette1.0.1; this does not
resolve the existing dependency declaration conflict.

This closes the full trained-weight **small-concurrency integration smoke** gap,
not broad accuracy,1M context, SuperMem ON/OFF, performance comparison or
nonuniform-LSE strict numerical acceptance. No test beyondT192 is claimed.
An initial1001-token prompt failed the harness's >1024-token shard-coverage gate;
it is excluded. The successful run used longer prompts without changing runtime code.

foraxe:
Performance follow-up on unchanged `c1e82ee97510f7805a6eef7663c53e7e3afacfe3`:

**Registered DCP8 O/LSE graph replay**, same inputs, optimization ON vs existing
fallback OFF, 10 alternating blocks ×100 replays, median of per-block max-rank timings:

| T | OFF | ON | Latency reduction |
|---|---:|---:|---:|
|48|206.69 µs|101.21 µs|51.03%|
|192|569.19 µs|220.02 µs|61.35%|

These are effective graph-replay intervals, not isolated kernel timing. Wall-time
results agree. Both arms passed changing-input uniform-LSE exact checks; timing
used finite nonuniform LSE with finite-output checks, not general strict numerical acceptance.

**Full 78-layer GLM-5.2 W4A8C8 HTTP**, one completed OFF/ON pair, TP8/DCP8/EP8,
no MTP/offload/prefix cache, 1920 input +128 output tokens:

| Concurrency | OFF mean TPOT | ON mean TPOT | Reduction | OFF / ON output throughput |
|---|---:|---:|---:|---:|
|1|48.74 ms|47.46 ms|2.63%|18.46 /18.88 tok/s|
|4|62.56 ms|60.05 ms|4.01%|51.65 /53.25 tok/s|
|192|655.25 ms|635.78 ms|2.97%|157.93 /161.34 tok/s|

The C192 cohort includes batch filling/prefill interference. Within the common
interval after all192 first tokens arrived and before any request's final token,
median client inter-token latency was **173.51 →144.80 ms (-16.54%)**.
These windows were5.90/4.92 seconds, extracted using the same rule from raw SSE;
this is not an independent repeat or an overall-request TPOT claim.

All197 measured requests per arm completed with128 token IDs and normal length
termination; zero preemptions. All197 paired output-token sequences matched.
Each of8 ranks confirmed34 actual unpadded T192 optimized replays in the ON C192
measurement. Query/store optimizations retain their smaller bounds; the T192
optimized path verified here is O/LSE. Loaded source hashes match this commit.

OFF disables only query/store/O-LSE performance eligibility on the same head;
remap correctness fixes remain in both arms. Shared launch settings include
scheduler budget4096, maxseq192, KV6GiB; the optimization cap remains192.
The previously disclosed isolated-runtime dependency caveat still applies.

The user requested stopping after one pair; subsequent incomplete work is excluded.
No cross-run variance/confidence claim, no performance extrapolation to1M or
SuperMem, and no testing aboveT192. Test resources were released; existing services untouched.

foraxe:
Moved the single-card operator precision tests in `c544a568b` to the requested directory:

- `tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_sfa_indexer_store_triton.py`
- `tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_sfa_dcp_remap.py`

Both are 100% content-preserving moves; no runtime code, assertions or tolerances changed. Query-gather/O-LSE tests require an eight-NPU group and remain distributed tests, not single-card cases. Both pre-commit stages passed. No new NPU execution is claimed for relocation.

Also consolidated the PR description and `DCP_PORT.md` with the actual real-weight E2E, T192 operator and single-pair HTTP results, replacing stale not-yet-tested wording. The tested runtime revision remains c1e82ee97; this follow-up changes tests/docs only. Fresh focused CPU suite:101 passed.

foraxe:
Hi @Tflowers-0129, this DCP optimization has been tested on A3 env, and still have similar perf optimization and correct outputs. 
cc @pisceskkk

AuroraEmiya (commented):
Project-guideline review: comments 2–6 are attached inline at the relevant code/test locations.

tests/ut/attention/a2/test_sfa_dcp_exchange_port.py
```diff
@@ -0,0 +1,149 @@
...
+        local_rank=rank,
+        backend="hccl",
+        group_name="large_dcp8_regression",
+        use_device_communicator=False,
+    )
+    retained = []
+    for tokens in (48, 192):
+        base = (torch.arange(tokens * 64 * 512).reshape(tokens, 64, 512) % 31 - 15).float() / 32
+        output = (base + rank / 16).to(torch.bfloat16).npu()
+        lse = torch.zeros((tokens, 64, 1), device="npu", dtype=torch.float32)
```
AuroraEmiya, line 120: The new fast-path eligibility extends from `T <= 12` to `T <= 192`, but this regression only covers uniform LSE at T48/T192. That validates packing/layout/graph replay, not general merge numerics. Could we add at least one T>12 non-uniform-LSE case (preferably covering the new boundary, e.g. T13/T191/T192, with mixed finite/-inf ranks) against an independent FP32/FP64 reference? Since the PR already documents the strict random-BF16 limitation, the acceptance tolerance should be stated explicitly.

vllm_ascend/ops/triton/sfa_dcp_exchange.py
```diff
@@ -0,0 +1,162 @@
...
+    @triton.jit
+    def _pack_token_peer(
+        output,
+        lse,
+        send,
+        output_s0: tl.constexpr,
+        output_s1: tl.constexpr,
+        lse_s0: tl.constexpr,
+        lse_s1: tl.constexpr,
+        tokens: tl.constexpr,
```
AuroraEmiya, line 56: `tokens` now ranges up to 192, but it is `tl.constexpr` and also drives `range(tokens)`, so different T values can produce distinct statically specialized/unrolled kernels. Have we measured compile-cache/code-size and first-call behavior for realistic mixed-T decode buckets? Please add a small T sweep, or consider a runtime/`do_not_specialize` form if T does not need to change the static tile structure.

vllm_ascend/ops/triton/sparse_index_remap.py
```diff
@@ -53,18 +54,20 @@ def remap_sparse_indices_fused_kernel(
...
-    # identically on NPU.
-    block_idx = idx // interleave_size
-    owner = block_idx - (block_idx // dcp_size) * dcp_size
-    valid = (idx >= 0) & (owner == dcp_rank)
-
-    if INTERLEAVE_ONE:
-        remapped = idx // dcp_size
+    # This backend's vector divsi loses integer boundaries above 2**24.
+    # Keep the native DCP8/interleave128 path entirely in bit operations.
+    if EXACT_DCP8:
```
AuroraEmiya: The exact repair is currently limited to DCP8/interleave128, while the generic branch below still uses vector integer division—the operation already shown to lose integer boundaries above `2**24` on this backend. Since the public wrapper still accepts arbitrary `dcp_size`/`interleave_size`, could unsupported configurations fall back to a verified implementation, or could the exact mapping be generalized? Otherwise known-risk arithmetic remains reachable.

vllm_ascend/ops/triton/sfa_dcp_merge.py
```diff
@@ -0,0 +1,157 @@
...
+            head = row % head_count
+            if native_layout:
+                lse_offsets = rank_ids * lse_s0 + head * lse_s1 + token * lse_s2
+            else:
+                lse_offsets = rank_ids * lse_s0 + token * lse_s1 + head * lse_s2
+            raw_lse = tl.load(lse_ptr + lse_offsets).to(tl.float32)
+            finite = (raw_lse == raw_lse) & (raw_lse > float("-inf")) & (raw_lse < float("inf"))
+            safe_lse = tl.where(finite, raw_lse, float("-inf"))
+            max_lse = tl.max(safe_lse, axis=0)
+            shifted = tl.where(finite, safe_lse - max_lse, 0.0)
```
AuroraEmiya: For an all-invalid rank set, `safe_lse` and `max_lse` are both `-inf`, so this still forms `-inf - (-inf)` before `tl.where` discards the lane. The final result may be zero today, but that relies on masking an invalid intermediate after it is created. Could we derive `has_valid`/a finite `safe_max` so the subtraction itself never produces NaN? That is more robust to backend lowering.

vllm_ascend/ops/triton/sfa_dcp_merge.py
```diff
@@ -0,0 +1,157 @@
...
+        native_layout = True
+    else:
+        _, tokens, heads, head_dim = out_recv.shape
+        native_layout = False
+    result = torch.empty((tokens, heads, head_dim), dtype=torch.float32, device=out_recv.device)
+    # Rank-wise accumulation avoids the failed R×D materialization. D512 now
+    # holds one FP32 accumulator plus one rank vector; retry verifies compiler UB.
+    block_d = min(triton.next_power_of_2(head_dim), 512)
+    tiles = tokens * heads * math.ceil(head_dim / block_d)
+    properties = triton.runtime.driver.active.utils.get_device_properties(out_recv.device.index)
```
AuroraEmiya: Could this use the repository's centralized VectorCore capability helper (`get_vectorcore_num()` / initialization path) instead of reading driver properties directly? The same direct property access also appears in `sfa_dcp_query.py`. Keeping device probing in one compatibility layer avoids duplicated backend assumptions; the returned core count can still be used as a grid candidate.

foraxe:
Hi @AuroraEmiya, I am working on those review opinions and will provide feedback before noon. 
Is there any opinon on the optimization direction?

AuroraEmiya:
> Hi @AuroraEmiya, I am working on those review opinions and will provide feedback before noon. Is there any opinon on the optimization direction?

Performance should come before your ability to function correctly and run safely. : )

tests/ut/attention/a2/test_sfa_dcp_exchange_port.py
foraxe: Addressed in 3fd321e13. Added the actual registered DCP8 graph test at T13/T191/T192 with changing O/LSE, nonuniform and extreme weights, all-invalid/mixed-invalid ranks, and NaN/+Inf LSE across every receiver shard. Fresh A3 execution passed all8 ranks (96 rank/shape/replay cases); the FP32 result is separately checked against independent CPU FP64 reduction before checking BF16 output.

The test states its per-element budget explicitly: `1e-6 + 8*eps32*sum(abs(weighted contributions))` for FP32, plus half a local BF16 ULP for the registered output. This is an acceptance budget, not a proved bound on backend exp. The old strict random-BF16 test is unchanged; I am not claiming its historical limitation is resolved by the new test. Runtime/test byte hashes match the commit.

vllm_ascend/ops/triton/sfa_dcp_exchange.py
foraxe: Measured on A3 and documented in 3fd321e13 / DCP_PORT.md. T1/2/6/12/13/16/32/48/64/96/128/191/192, empty dedicated cache then a new process reusing it: thirteen cold first calls total18.511s; disk-cache first calls total0.169s. Cache growth531350 bytes including IR/metadata; thirteen npubin artifacts total41024 bytes, each3000–3400 bytes. T192 is3192 bytes and its TTIR retains `scf.for`. All pack outputs passed bitwise checks.

This measures synchronized host first-call compilation/loading/launch, not kernel latency; the first cold call also includes Triton backend initialization. Specialization exists, but this sweep does not show token-proportional unrolled code growth. I retained the unchanged pack kernel rather than introducing a dynamic-T rewrite without measured benefit. This is13 shapes, not an exhaustive all-T cache/latency guarantee.

vllm_ascend/ops/triton/sparse_index_remap.py
foraxe: Addressed in 3fd321e13. I reproduced wrong results on the existing generic branch for DCP4/128, DCP8/64, DCP2/1 and DCP3/127 with int32-range indices above2**24. The public wrapper now routes non-DCP8/128 configurations to exact INT64 Torch quotient/remainder and order-preserving compaction; the unsafe generic Triton arithmetic is removed. DCP8/128 keeps the original bitwise fast path.

Fresh NPU tests passed:35 single-card cases, including both int32/int64, every logical rank in those configurations, noncontiguous tensors, arbitrary leading dimensions, all-invalid inputs and102 changed-input fallback graph replays. CPU tests use an independent Python-integer oracle. Integer sort can run on AICPU; this intentionally favors correctness and does not claim a generic-path performance improvement.

vllm_ascend/ops/triton/sfa_dcp_merge.py
foraxe: Addressed in 3fd321e13: all-invalid rows select a finite `safe_max=0` before subtraction, eliminating `-inf - (-inf)` even in a subsequently discarded lane. Existing invalid-rank output masking and zero denominator handling remain. Fresh eight-rank T13/T191/T192 nonuniform graph tests passed, including exact-zero all-invalid outputs and NaN/+Inf/-Inf invalid ranks in every receiver shard; the existing uniform T48/T192 regression also passed.

vllm_ascend/ops/triton/sfa_dcp_merge.py
foraxe: Addressed in 3fd321e13. Both merge and query helpers now use `init_device_properties_triton()` plus `get_vectorcore_num()`, including standalone use before worker initialization. They retain their grid caps and existing one-device-per-worker convention. Fresh DCP8 query graph regression and both registered exchange graph tests passed on all8 ranks;113 focused CPU tests also passed.

foraxe:
CI status on 5db959926: the pre-commit job, including mypy, now passes (job 107497888247). The CPU-UT job did not execute tests: its automatic rebase on main failed while replaying the historical f8186d340 commit, with a content conflict in vllm_ascend/attention/indexer.py (job 107498909654). This is separate from the completed fresh NPU validation and the 113 passing focused CPU tests. I have not bypassed CI or rewritten the published branch history. Logs: https://github.com/vllm-project/vllm-ascend/actions/runs/35957186465/job/107498909654

AuroraEmiya (changes requested):
For Triton kernels in this project, we use the following parameter convention:

* `tl.constexpr` is reserved for values that are genuinely part of the compile-time kernel configuration or static IR structure, such as fixed tile sizes, static feature/layout switches, or dimensions required by `tl.arange` / `tl.static_range`.
* Compile-time parameters must use uppercase names, e.g. `BLOCK_D`, `RANKS`, `LOCAL_HEADS`, `STATS_TILE`.
* Values that vary with runtime workload, tensor shape/layout, or allocation state should remain normal runtime parameters unless there is measured evidence that specializing them is necessary. Examples include token count, tensor strides, total work items, program count, and cache capacity.
* A shape-derived value should not be marked `tl.constexpr` merely because Triton permits specialization.

With that convention, parameters such as `tokens`, the stride arguments, `total_tiles`, `num_programs`, and similar shape-derived values across these new kernels need to be reviewed. `rows` in `_fused_merge_kernel` also appears to be unused and should be removed rather than becoming another specialization key.

The cold-cache numbers make this more important, not less. The current sweep changes T and therefore creates different specializations. Thirteen pack-kernel cold first calls cost 18.511 s in total. Even after excluding backend initialization, sub-second/second-scale compilation for new shapes is significant compared with seconds-scale model decoding. Also, this measurement covers only the pack kernel; the merge path contains additional shape-derived `tl.constexpr` parameters, so it does not yet establish the cold compilation cost of the complete optimized path.

Therefore, keeping runtime-varying parameters as `tl.constexpr` without a demonstrated need for specialization would not pass review from my side.

Could you please:

1. Audit all new `tl.constexpr` parameters and classify them into true compile-time constants versus runtime-varying values.
2. Rename retained compile-time constants to uppercase.
3. Remove unnecessary/unused constexpr parameters.
4. For any runtime-varying parameter that is intentionally kept as `tl.constexpr`, provide an A/B showing that specialization gives a meaningful steady-state performance benefit over the runtime form.
5. Measure cold first-call/specialization cost for the complete optimized pack → exchange → merge path, not only the pack kernel.

The fact that T192 retains an `scf.for` is useful—it shows there is no token-proportional loop unrolling—but the remaining question is whether T and the other shape-derived values should be specialization keys at all.

PR #16362, opened 2026-09-11
[Feature] MLA overlays the DCP and speculation support the chunkprefill tiling solution

weiguihua2:
The chunkprefill tiling solution is supported when the MLA overlays the DCP and speculation.

The DCP algorithm enables the KV cache to be stored on different cards. When DCP is used together with MTP, a heterogeneous mask is required for attention calculation. However, the 96-head FIA operator of Ascend A5 does not support the transmission of heterogeneous masks. Therefore, the solution needs to be modified as follows: The attention calculation is divided into two steps. In the first step, the Q and historical KV are calculated, and the mask does not need to be transmitted. In the second step, the current KV is calculated, and the KV is complete. In this case, only the lower triangular mask needs to be transmitted. Then, the attention output of the two steps is updated.

no


vllm_ascend/worker/dcp_utils.py
```diff
@@ -706,7 +696,7 @@ def generate_dcp_metadata(
             max_query_len=(int(query_lens_cpu[:num_reqs].max().item()) if num_reqs else 0),
         )
 
-        if self.speculative_config and not self.use_sparse:
+        if self.speculative_config and not self.use_sparse and not self.vllm_config.model_config.use_mla:
```
pisceskkk, line 699: We can only use `not self.vllm_config.model_config.use_mla` to guard both DSv3 and DSv3.2/GLM5.2, so just replcae `use_sparse` with `use_mla`.

pisceskkk (approved):
LGTM, leave a small nit.

zhenwenqi2024 (approved):

Ronald1995 (approved):
LGTM

vllm_ascend/worker/dcp_utils.py
pisceskkk: we can simplify this code in a follow-up PR.

weiguihua2: OK

ZT-AIA (approved):

PR #16363, opened 2026-09-11
[Feature][Spec Decode] Add FLy verification for MRV1

Dunnn-Smith:

This PR adds an initial Ascend implementation of FLy verification for speculative decoding.

With standard greedy verification, decoding stops at the first draft token that does not match the target argmax. FLy may keep this token when the target entropy is high and the following draft tokens match the target results within a lookahead window.

The goal is to accept more draft tokens and improve the average accepted length and throughput.

The current implementation:

* focuses on the MRV1 greedy verification path;
* adds configurable entropy top-k, entropy threshold, and window size;
* supports both the Triton path and the PyTorch fallback;
* includes the corresponding Triton warmup changes;
* is disabled by default.

This work is based on the [FLy paper](https://arxiv.org/abs/2511.22972) and refers to [vLLM PR #53987](https://github.com/vllm-project/vllm/pull/53987), with a separate implementation for the Ascend rejection sampler.

Yes, and it is opt-in only.

The following options are added under `rejection_sampler_config`:

* `enable_fly_verify`
* `fly_entropy_top_k`
* `fly_entropy_threshold`
* `fly_window_size`

FLy is disabled by default, so the existing behavior is unchanged when it is not enabled.

**Experimental setup.** Experiments were conducted on 8 Ascend 910B NPUs using the `dsv4-flash-0731-w4a8` model with TP=8, DP=1, and DSpark with `spec_token=7`. For all FLy configurations, `fly_entropy_top_k=3` and `fly_entropy_threshold=0.3` were fixed. Throughput was evaluated on GSM8K using AISBench with 200 requests and a maximum concurrency of 6. 

#### Throughput

| Configuration | `fly_window_size` | Output Throughput (token/s) | Improvement over Baseline |
|---|---:|---:|---:|
| Baseline | – | 183.3752 | – |
| FLy | 4 | 188.0102 | +2.53% |
| FLy | 5 | 188.9307 | +3.03% |
| FLy | 6 | 189.0912 | +3.12% |

#### Accuracy

Accuracy was evaluated on 2,000 GSM8K cases sampled using the same random seed.

| Configuration | `fly_window_size` | Accuracy | Difference from Baseline |
|---|---:|---:|---:|
| Baseline | – | 95.60% | – |
| FLy | 6 | 95.55% | −0.05 percentage points |


drslark:
Wait https://github.com/vllm-project/vllm/pull/53987 to merge.

PR #16390, opened 2026-09-12
feat(mla): Support first-axis non-contiguous MLA KV Cache and enable it by default

HackClawMxw:
Refactor the MLA KV Cache storage layout on NPU to support a component-major organization with a non-contiguous first axis, and enable it by default on ModelRunner V1.

Core changes:
- Change the physical MLA KV Cache layout from token-major to component-major. Within each component slot, all nope data is stored first, followed by all rope data. Both parts are constructed from the shared raw storage via as_strided, while the call arguments remain unchanged when passing the cache to the attention operator.
- Support models where the manager block size differs from the kernel block size. A manager page is automatically split into multiple kernel slots according to the ratio r. DeepSeek V3 corresponds to r=1, while Kimi K3 corresponds to r=3 or r=6 depending on the TP configuration.
- Integrate allocation and reshape logic into the normal _allocate_kv_cache_tensors and _reshape_kv_cache_tensors paths.
- Explicitly declare LBNHC as the supported KV Cache layout in the Ascend MLA backend, ensuring correct layout negotiation for hybrid models.
- Validate that the final layout is LBNHC before cache allocation and before component-view construction, failing early on conflicts.
- For MLA specs that cannot be represented as pure nope/rope components, such as compressed, quantized, sparse, or opaque formats, automatically fall back to the existing Ascend legacy path to keep all MLA models runnable.
- Adjust the COW prefix-cache copy logic to perform structural conversion, consistently recognize component pairs, and preserve the component-major layout.
- Update cache zeroing logic to recognize component pairs and zero the entire manager page.
- ModelRunner V2 uses its own implementation and does not apply the component path.

### DIFF
--- a/vllm_ascend/attention/mla_v1.py
+++ b/vllm_ascend/attention/mla_v1.py
@@ -56,6 +56,7 @@ from vllm_ascend.utils import (
     vllm_version_is,
     weak_ref_tensors,
 )
+from vllm_ascend.worker.mla_component_cache_v1 import is_mla_component_pair
 from vllm_ascend.worker.npu_input_batch import NPUInputBatch
 
 if vllm_version_is("0.28.0"):
@@ -104,7 +105,13 @@ class AscendMLABackend(AttentionBackend):
         return num_blocks, block_size, num_kv_heads, head_size
 
     @classmethod
-    def supported_kv_cache_layouts(cls) -> tuple[KVCacheLayout, ...]:
+    def supported_kv_cache_layouts(cls) -> tuple[KVCacheLayout, ...] | None:
+        try:
+            vllm_config = get_current_vllm_config()
+        except AssertionError:
+            vllm_config = None
+        if vllm_config is not None and getattr(vllm_config, "use_v2_model_runner", None) is True:
+            return None
         return (KVCacheLayout.LBNHC,)
 
     @classmethod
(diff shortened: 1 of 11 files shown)


vllm_ascend/attention/mla_v1.py
chenlisi, line 1: 去掉nz的判断，算子重构需要保障功能不丢失

chenlisi: spec没有定制，不需要实现customize接口；默认的layout就是LBNHC，无需新增接口

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -4990,7 +4951,7 @@ def _reshape_kv_cache_tensors(
         self,
         kv_cache_config: KVCacheConfig,
         kv_cache_raw_tensors: dict[str, torch.Tensor],
-    ) -> dict[str, torch.Tensor]:
+    ) -> dict[str, Any]:
```
chenlisi: 不要修改接口返回值

vllm_ascend/worker/model_runner_v1.py
chenlisi: 不要改变接口返回值

chenlisi: 不要改动接口返回值

chenlisi: 此段逻辑要删除

PR #16412, opened 2026-09-12
[Feature][KV Transfer] Support pipeline-parallel KV offloading

luo-rabbit:
Add PP-aware SFA completion tracking and handshake, isolate MemCache keys by PP rank, and correct stage-local layerwise reuse mapping.

  This PR adds pipeline-parallel support for KV offloading. It aggregates SFA transfer completion across all PP stages, introduces a PP-aware metadata handshake, isolates MemCache keys by PP rank, and consistently translates global layer IDs to stage-local indices for layerwise reuse and multiple KV cache groups. These changes prevent premature request completion, cross-stage cache-key collisions, and valid layers being excluded on nonzero PP stages.

  Yes. Pipeline-parallel deployments can now use KV offloading through the affected SFA RD2H and MemCache paths. No new user configuration or public API is introduced.

  ## How was this patch tested?

  Added and updated unit tests covering PP-aware completion aggregation and topology validation, metadata handshake compatibility, MemCache key isolation, stage-local layer reuse, and multiple KV cache groups on a nonzero PP stage. The targeted PP layer-index regression tests pass, and the complete KV-offloading patch was also validated in the PP validation environment.


luo-rabbit:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

luo-rabbit:
/rerun
[Bot]: rerun completed.

**Failed:**

nwpu-zxr (approved):

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
```diff
@@ -402,7 +402,7 @@ def _init_layerwise_config(self) -> None:
                 for layer_name in group_spec.layer_names
             }
             if physical_layers:
-                effective_num_layers = max(self.num_layers, max(physical_layers) + 1)
+                effective_num_layers = max(self.num_layers, len(physical_layers))
```
ader47, line 409: For configurations with multiple KV cache groups, changing `num_layers` to the stage-local layer count makes it inconsistent with the downstream global-layer-index checks. For example, with 8 layers evenly split across PP=2, the second stage has `num_layers=4` and global layer IDs `{4, 5, 6, 7}`. The `physical_layer >= self.num_layers` check below then excludes all four layers from the layer-to-group mapping. `_infer_cache_group_metadata()` applies the same filtering and produces empty cache-address lists.

Could we consistently map global layer IDs to stage-local indices when building both the layer-to-group mapping and cache metadata? Please also add a regression test covering a nonzero PP stage with multiple KV cache groups.

ader47 (approved):

Ronald1995 (approved):
LGTM

vllm_ascend/distributed/kv_transfer/kv_pool/ascend_store/pool_worker.py
luo-rabbit: Thanks for catching this. I've updated both the layer-to-group mapping and cache metadata construction to consistently translate global layer IDs into stage-local indices, and added a regression test covering a nonzero PP stage with multiple KV cache groups. The mapping remains an identity mapping for PP=1, so existing behavior is unchanged.

ader47 (approved):

Tflowers-0129 (approved):

PR #15956, opened 2026-09-07
[Ops][Feature] Support Ascend 950 and upgrade batch invariant ops to 2.0.0

wangx700:
### What this PR does / why we need it?

ref to https://github.com/vllm-project/vllm-ascend/issues/10072

Upgrade the batch invariance operator run package and PyTorch extension archive to 2.0.0. Map ascend950 variants to the 950 package and enable operator installation during both Ubuntu and openEuler A5 image builds.

Document Ascend 950 installation and graph support. Remove the forced PIECEWISE configuration from the online/offline examples and the TP4 batch invariance test, preserving its capture sizes and updating the graph coverage annotation.

Restrict the custom reduce-sum operator to FP16, FP32, and BF16. Other dtypes fall back to the saved native torch.sum implementation, avoiding recursive calls through the patched Tensor.sum. Add parameterized regression tests for supported and unsupported NPU dtypes.

### Does this PR introduce _any_ user-facing change?
A5 source-built images install the batch invariance operators. Users still set VLLM_BATCH_INVARIANT=1 at inference time. Examples and the TP4 consistency test use the default graph configuration.

### How was this patch tested?
- pytest -sv tests/e2e/pull_request/four_card/rlhf/consistency/test_batch_invariant_tp4.py
```
======================================================================= warnings summary ========================================================================
../../../../../usr/local/python3.11.10/lib/python3.11/site-packages/torch/jit/_script.py:362: 14 warnings
  /usr/local/python3.11.10/lib/python3.11/site-packages/torch/jit/_script.py:362: DeprecationWarning: `torch.jit.script_method` is deprecated. Please switch to `torch.compile` or `torch.export`.
    warnings.warn(

tests/e2e/pull_request/four_card/rlhf/consistency/test_batch_invariant_tp4.py:111
  /mnt/share/w00899129/vllm_code_0908/vllm-ascend/tests/e2e/pull_request/four_card/rlhf/consistency/test_batch_invariant_tp4.py:111: PytestUnknownMarkWarning: Unknown pytest.mark.model - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.model(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================================================== 1 passed, 15 warnings in 94.14s (0:01:34) ===========================================================
[...]

PR #16413, opened 2026-09-12
[BugFix] Fix batch-invariant reduce_sum crash on non-last-dim reductions

EliasKaslan:

`vllm_ascend/batch_invariant.py:reduce_sum` unconditionally forwards every NPU
`torch.sum(x, dim, keepdim)` (for supported dtypes) to
`torch.ops.batch_invariant_ops.npu_reduce_sum_batch_invariant`. But the
`aclnnReduceSumBatchInvariant` kernel **only supports reducing the
last dimension**; for any other `dim` it raises:

```
AclNN_Parameter_Error(EZ1001): Provided dim only support last dim
```

Because `enable_batch_invariant_mode()` monkey-patches `torch.sum` /
`torch.Tensor.sum` globally, **any** non-last-dim `.sum()` on NPU kills all TP
workers.

To fix it, this PR only takes the batch-invariant path for a **genuine
last-dim reduction**:

1. The last dim can only be spelled as `-1` or `x.dim() - 1`, so the guard is
   simply `dim == -1 or dim == x.dim() - 1`, and the caller's `dim` is
   forwarded to the kernel unchanged. No dimension arithmetic is needed,
   and equality comparison is safe for every value of `dim`: a tuple, for
   example, simply fails both comparisons and falls back instead of raising.
   The kernel is therefore used only when all of the following hold: NPU
   tensor, a genuine last-dim reduction, and dtype in `{fp16, fp32, bf16}`.
   Forwarding `dim` unchanged means every previously-working last-dim call keeps
   the exact behavior it had before this PR, and the existing unit test stays
   valid.
2. Everything else - non-last-dim, tuple `dim`, full reduction (`dim is None`
   on >1-D), CPU tensors, unsupported dtypes - falls back to the **saved
   native `torch_sum` reference** captured at module import, so the fallback
   cannot recurse into the patched entry points.

Every input that now falls back previously crashed, so the change strictly
widens the set of inputs that run; there is no working behavior being changed.

Why this does not affect batch invariance?

1. **The invariance-critical reductions are all last-dim, and they are
   unchanged.** The reductions whose results must be batch-invariant (softmax
   denominators, RMSNorm variance - reductions over the hidden/vocab
   dimension) are last-dim reductions on `[num_tokens, hidden]` tensors.
   Native kernels pick their reduction strategy from the total tensor size,
   so the per-row accumulation order can differ between BS=1 and BS=N - that
   is exactly what `npu_reduce_sum_batch_invariant` prevents with a
   row-count-independent accumulation order. These calls still take the
   batch-invariant kernel; their behavior is byte-for-byte identical to
   before.

2. **The newly-fallback paths have no batch-shape-dependent input.** The
   non-last-dim sums that now fall back (e.g. `embeds.sum(dim=0)` in
   multimodal position-embedding preprocessing) are per-request computations:
   the input tensor is built from one request's own patches, so its shape and
   values are identical in a BS=1 run and in a BS=N run. Native `torch.sum`
   is deterministic for a given tensor (same shape ? same kernel and tiling ?
   same accumulation order), so identical input yields bitwise-identical
   output across batch sizes. Invariance is preserved.

3. **No regression is possible.** Every path that now falls back previously
   raised EZ1001 (non-last-dim int dims) or a schema type error (tuple dims).
   There was no working behavior to preserve.

4. **It follows the established fallback design.** #15956 introduced the same
   pattern for the dtype axis: unsupported dtypes "fall back to the saved
   native torch.sum implementation, avoiding recursive calls through the
   patched Tensor.sum". This PR extends that pattern from dtype to dim.

No

- 10-path NPU unit test on Ascend910 (dim=0 fallback, dim=2 / dim=-1 equivalence
  vs native, keepdim, 1-D `.sum()`, tuple dim, all-dims reduction, fp16,
  functional `torch.sum`, fp32) - all pass.
- End-to-end batch-invariance logprobs test (BS=1 vs BS=N bitwise) runs through
  `profile_run` and generation with no EZ1001.

https://github.com/vllm-project/vllm/commit/b2f685834a6456197e7033966fdef52a23f1abcd

---------

Signed-off-by: EliasKaslan <3248569436@qq.com>


vllm_ascend/batch_invariant.py
```diff
@@ -70,9 +70,19 @@ def reduce_sum(x: torch.Tensor, dim: int | None = None, keepdim: bool = False) -
...
-    # CPU tensors and unsupported dtypes/dimensions use the saved native torch.sum.
+    # aclnnReduceSumBatchInvariant only supports reducing the last dimension and
+    # raises AclNN_Parameter_Error(EZ1001, "Provided dim only support last dim")
+    # for any other dim. Normalize negative dims and only take the batch-invariant
+    # path for a genuine last-dim reduction; everything else (non-last-dim, tuple
+    # dim, full reduction when dim is None, CPU tensors, unsupported dtypes) falls
+    # back to the saved native torch.sum.
+    if (
+        x.device.type == "npu"
+        and isinstance(dim, int)
```
EliasKaslan: 可以直接删掉

linfeng-yuan (dismissed):
Please remove ` and isinstance(dim, int)`

vllm_ascend/batch_invariant.py
lcfenglinwan: can remove this line

EliasKaslan:
> Please remove ` and isinstance(dim, int)`

done

linfeng-yuan (approved):

PR #16430, opened 2026-09-12
[BugFix][Sampling] Fix greedy token IDs in mixed batches

### LINKED ISSUE #16396: [Bug]: enable_reduce_sample returns incorrect vocabulary IDs for greedy requests in mixed batches
### Your current environment

The issue was reproduced with vLLM-Ascend built from source at `fb00c6b`, together with vLLM at the CI-pinned commit `b2f6858`. Key output from `python collect_env.py` in the same test environment:

```text
OS: Ubuntu 22.04.5 LTS (aarch64)
Python: 3.12.13
PyTorch: 2.10.0+cpu
torch-npu: 2.10.0.post4
CANN: 9.1.0
npu-smi: 25.5.2
NPU: Ascend910; 1 and 2 NPUs used to test TP1 and TP2, respectively
```

### 🐛 Describe the bug

With `enable_reduce_sample` enabled during speculative decoding, a batch containing both greedy-sampling and random-sampling requests can cause the greedy requests to return positions in the reordered candidate tensor instead of the corresponding vocabulary token IDs. This produces incorrect output tokens and also affects draft-token acceptance and rejection.

A component-level reproducer uses two requests with entirely finite target logits. Each row is:

```python
[-8., -3., -2., -1., 0., 9., 1., 2.]
```

The maximum corresponds to vocabulary token ID **5**. The first request has temperature `0`; the second has temperature `0` or `0.7`. Each request contains one draft token, with no Top-k/Top-p filtering. The bonus token is fixed to ID `7` so that we can observe whether it is appended; `-1` denotes an output placeholder after rejection.

Results for the first request at TP1:

| Batch temperatures | Draft token ID | Expected output | Actual output |
|---|---:|---|---|
| `[0, 0]` | 0 | `[5, -1]` | `[5, -1]` |
| `[0, 0.7]` | 0 | `[5, -1]` | `[0, 7]` |
| `[0, 0]` | 5 | `[5, 7]` | `[5, 7]` |
| `[0, 0.7]` | 5 | `[5, 7]` | `[0, -1]` |

The same greedy request incorrectly accepts draft token ID `0` and incorrectly rejects draft token ID `5` solely because it is batched with a random-sampling request.

The reproducer and instructions are packaged in `native-reproducer.zip` . The reproducer uses synthetic logits and requires neither model weights nor a service endpoint. Run it in the source environment above without applying the fix:

```bash
TORCH_DEVICE_BACKEND_AUTOLOAD=0 python -m torch.distributed.run \
  --standalone --nnodes=1 --nproc_per_node=1 native_sampling_probe.py \
  --variant before --output-dir results/before-tp1
```

Of the 16 TP1 test cases, all 8 pure-greedy controls pass and all 8 mixed-batch cases fail. Changing `--nproc_per_node` to `2` and using a different output directory reproduces the same type of error at TP2. The JSON reports contain the expected and actual IDs for each case. The script fixes the bonus token and random inputs while executing the actual NPU rejection-sampling kernels and HCCL.

The relevant code is in `vllm_ascend/sample/rejection_sampler.py`. An all-greedy batch skips candidate reordering, whereas a mixed batch passes through `apply_sampling_constraints`, which returns reordered scores and vocabulary indices. The subsequent greedy branch calls `greedy_sample(target_logits)` and interprets column positions using the original vocabulary sharding, without applying the candidate-to-vocabulary mapping.

Related design discussion: #9248. A local fix and regression tests are available; I plan to submit them in a separate PR.

[native-reproducer.zip](https://github.com/user-attachments/files/32138241/native-reproducer.zip)

KeyanHu-git:

Fixes #16396.

With `enable_reduce_sample=True`, the rejection sampler passes reordered logits to `greedy_sample` when greedy and random requests share a speculative batch. `greedy_sample` assumes the columns still follow the local vocabulary order. For logits `[-8, -3, -2, -1, 0, 9, 1, 2]`, the mixed TP1 path returns candidate position 0 instead of vocabulary ID 5. It incorrectly accepts draft token 0 and rejects draft token 5.

Compute greedy IDs after logits processors and before `apply_sampling_constraints`, then pass them to `rejection_sample`. This keeps processor masks effective and preserves the original argmax tie-break before top-k truncation. The two greedy all-gathers now run before candidate gathering.

Fixes incorrect tokens and draft acceptance for greedy requests batched with random requests. No serving API or configuration changes.

Tested on Ascend `9d9b9ed` with its paired vLLM `a97dacb`:

```bash
TORCH_DEVICE_BACKEND_AUTOLOAD=0 python -m pytest -sv tests/ut/sample/
```

Applying only the test changes gives **32 passed, 8 failed**. With the fix, **40 passed**. The 16 new cases cover pure-greedy/mixed batches, masks, ties, acceptance/rejection, and a random row whose expected result differs from greedy sampling.

Native TP1 and TP2 checks passed all 16 cases per rank; the original code failed the 8 mixed cases. These use real HCCL and rejection kernels with fixed bonus and random inputs. Python 3.12, Torch 2.10.0+cpu, torch-npu 2.10.0.post4 and triton-ascend 3.2.2. The test environment had pre-existing FastAPI and Transformers dependency conflicts.

Applicable pre-commit hooks passed for both changed Python files, including Typos and Gitleaks.

Greedy argmax now scans the local vocabulary shard. Measured target preprocessing time with vocabulary 131072, draft length 3, batch 256 and top-k=64:

| Target preprocessing | Before | After |
|---|---:|---:|
| TP1 | 4.9284 ms | 5.2203 ms |
| TP2 | 2.6743 ms | 2.7807 ms |

Timings cover target preprocessing (cloning, constraints and greedy reduction) only. Values are medians from 20 alternating AB/BA groups of 10 iterations after warmup.

AI assistance: Codex was used for the patch, tests and PR description.

### DIFF
--- a/tests/ut/sample/test_rejection_sampler.py
+++ b/tests/ut/sample/test_rejection_sampler.py
@@ -24,6 +24,7 @@ from vllm_ascend.sample.rejection_sampler import (
     sample_recovered_tokens_blockwise_pytorch,
     sample_recovered_tokens_pytorch,
 )
+from vllm_ascend.sample.sampler import _apply_top_k_top_p_pytorch
 
 # Global constants
 PLACEHOLDER_TOKEN_ID = -1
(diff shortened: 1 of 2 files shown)


KeyanHu-git:
@realliujiaxu Could you add the `ready-precise` label to this PR?

The ci-gate job passed pre-commit and cpu-ut, but precision tests were skipped because this PR changes source code and no test-selection label is set.

PR #16455, opened 2026-09-13
[Bugfix][MoE] Skip compiled graphs when MoE comm contract mismatches

danziheng1024:

The V1 compiled MoE graph bakes in the routed-output reduction contract that is selected at `mc2_tokens_capacity` (decode-sized, i.e. MC2/FUSED_MC2). However, at runtime large prefill batches may dynamically select a different MoE comm method (e.g. `ALLGATHER`) whose routed-output reduction contract differs. When such a call replays the compiled graph, the `_fused_output_is_reduced` flag baked into the graph no longer matches the runtime method, causing `_maybe_reduce_final_output` to either apply a second TP all-reduce (double-counting contributions) or skip a required one.

This PR makes the reduction contract consistent between compiled graphs and runtime execution:

- Add `moe_output_is_reduced()`, `get_compiled_moe_comm_type()` and `moe_comm_contract_mismatch()` helpers in `ascend_forward_context.py` to describe and compare the routed-output reduction contract of each MoE comm method (ALLTOALL/MC2/FUSED_MC2, or ALLGATHER under sequence parallelism).
- While `torch.compiler.is_compiling()`, resolve the contract from `compiled_moe_comm_type` (derived from `mc2_tokens_capacity`) so the decode/capture contract is baked into the outer graph. Profile runs still compile at `max_num_batched_tokens` where the runtime method can be ALLGATHER; the profile output is discarded, and real calls with a different contract bypass this graph (see below).
- In `set_ascend_forward_context()`, set `skip_compiled` for real (non-profile) calls whose runtime MoE method disagrees with the compiled contract, so those calls (e.g. large prefill batches) execute eagerly instead of replaying the mismatched compiled graph.
- Register `compiled_moe_comm_type` in `_ExtraForwardContextProxy.extra_attrs` so it is accessible from CustomOps via `_EXTRA_CTX`.

No. The change only fixes an internal correctness issue (wrong/duplicated TP all-reduce on the MoE routed output for large prefills under compiled graphs). No API, interface or behavior changes are introduced for users.

- Existing MoE / aclgraph CI cases pass (no behavior change for matched-contract paths).
- Manually verified on a TP>1 deployment with a MoE model where decode uses MC2 while large prefills fall back to ALLGATHER: outputs of mixed prefill/decode workloads match eager-mode golden results, and contract-mismatched prefills now run eagerly (skip_compiled) instead of replaying the compiled graph.


danziheng1024:
Closing in favor of #16459, which fixes the same bug at the root instead of working around it.

The approach here (bake the capacity-sized contract into the compiled graph and skip compiled execution on mismatch) stabilized correctness but kept two sources of truth for the reduction contract. #16459 unifies the contract across all comm methods (ALLGATHER finalize now reduces the routed output, like MC2/ALLTOALL already did), so `_fused_output_is_reduced` depends only on static layer config and the compiled graph can no longer disagree with the runtime -- no skip machinery needed.

PR #16458, opened 2026-09-13
[Feature][DSA] Support MXFP4 (C4) quantization for sparse lightning indexer

danziheng1024:

This PR adds MXFP4 (C4) quantization support for the DSA sparse lightning indexer (LI), complementing the existing C8 (INT8_DYNAMIC / W8A8_MXFP8) path, so that W4A4_MXFP4 quantized models can also store and select the indexer cache in C4 format on A5.

- Add `enable_sparse_li_c4` to `AscendConfig`, mutually exclusive with `enable_sparse_li_c8`; generalize the C8-only layer parsing into `_parse_sparse_li_layers_from_quant_config()` with per-mode valid quant types (`W4A4_MXFP4` for C4) and add `is_sparse_li_c4_layer()`
- Unify q/k LI quantization in `AscendSFAIndexerBackend._quantize_li_tensor()`: the C4 path uses `torch_npu.npu_dynamic_mx_quant` to `float4_e2m1fn_x2` with `float8_e8m0fnu` scales; scale cache writes go through `uint8` views since `npu_scatter_nd_update_` requires uint8 for non-standard dtypes
- Replace the `cache_sparse_li_c8` bool in `AscendSFAIndexerCacheSpec` with a generic `li_quant_mode` field (`""` / `"cache_sparse_li_c8"` / `"cache_sparse_li_c4"`)
- Implement the C4 lightning indexer path on A5 via `torch.ops._C_ascend.npu_quant_lightning_indexer_v2(_metadata)`; non-A5 devices raise a clear `RuntimeError`
- Wire C4 cache specs in both model runner v1 and v2: halved `head_size` for packed FP4 storage and MX block-scale layout (`head_size * 2 // 64 x 2`)

Yes. A new Ascend additional-config option `enable_sparse_li_c4` (default `False`) is added. Existing behavior is unchanged unless it is explicitly enabled.

Served a W4A4 MXFP4 quantized GLM-5.2 checkpoint on an Atlas A5 x8 node with model runner v2, TP=8, EP, and deepseek MTP speculative decoding:

```bash
vllm serve /path/to/GLM-5.2-W4A4-mxfp4 \
  --tensor-parallel-size 8 \
  --enable-expert-parallel \
  --quantization ascend \
  --max-model-len 70000 \
  --speculative-config '{"num_speculative_tokens": 3, "method": "deepseek_mtp", "enforce_eager": true}' \
  --additional-config '{"enable_sparse_sfa_c8": true, "enable_sparse_li_c4": true}'
```

`ruff check` and `ruff format --check` (ruff v0.14.0, same as pre-commit) pass on all changed files.


vllm_ascend/ascend_config.py
```diff
@@ -770,8 +713,14 @@ def derive_and_validate(self, vllm_config: VllmConfig) -> AscendConfig:
...
             self._sparse_li_c8_layer_names,
-        ) = self._parse_sparse_li_c8_layers_from_quant_config(quant_config)
-        self._sparse_li_c8_layer_filter_enabled = self._has_sparse_li_c8_layer_config(quant_config)
+        ) = self._parse_sparse_li_layers_from_quant_config(
+            quant_config, ("INT8_DYNAMIC", "W8A8_MXFP8"))
+        (
+            self._sparse_li_c4_layer_ids,
+            self._sparse_li_c4_layer_names,
+        ) = self._parse_sparse_li_layers_from_quant_config(
+            quant_config, ("W4A4_MXFP4",))
```
lijiahang226, line 722: Could you please check whether W4A4_MXFP4 should be used here? In the current MXFP quantization scheme, attention quantization is always fixed to MXFP8.

PR #16465, opened 2026-09-13
[Attention][BugFix] Remove the blocking seq_lens D2H on DSpark/DFlash draft builds (#16271)

### LINKED ISSUE #16271: Qwen3/Qwen3.5 Dspark host下发存在很大的流同步，阻塞下发流水。导致device 侧的Dspark 存在很多空泡
### 📚 The doc issue

Image attention v1第342行seq_lens.tolist()会阻塞下发流水，Dspark下这个seq_lens本身在设备上。to_list会转到cpu上。会阻塞流水。后面的attention的menta又需要这个cpu上的seq_len。

### Suggest a potential alternative/fix

能不能attention 的menta 支持设备上的seq_lens

Liuchenbing-2026:

Fixes #16271.

Under parallel drafting (DFlash / DSpark), `AscendAttentionMetadataBuilder.build`
obtains the Python `seq_lens_list` it hands the attention op from
`seq_lens.tolist()` on an **NPU tensor**, which copies device→host and blocks the
host until the compute stream drains. Every attention-metadata build therefore
serialises dispatch against the device instead of running ahead of it. Model
runner v2 is affected too - `worker/v2/attn_utils.build_attn_metadata` calls the
same builder.

Measured on A2 (8x910B4, cards 4-7, TP4+EP, Qwen3.6-35B-A3B /
`Qwen3_5MoeForConditionalGeneration` plus its DSpark speculator, MRV2, vLLM
0.28.0): 20000 of 20000 metadata builds went through a device→host copy,
118.5 s of blocking in 787.6 s of wall time.

The exactness has to be carried explicitly, so this PR adds two fields to
`AscendCommonAttentionMetadata` and splits the fix in three steps, each one
behind an independent opt-in:

**1. Target builds stop paying the copy.** `seq_lens_cpu_is_exact` is set by
`worker/v2/attn_utils.build_attn_metadata` from `seq_lens_np is not None` - only
a *target* build has a genuinely exact host mirror. 75.8 % of the copies go
away (7566 of every 10000 builds), and every host-derived list was compared
element-wise against the device tensor: 30264 comparisons across 4 ranks, zero
differences. Median ITL at concurrency 8: 75.40 ± 0.62 ms → 74.54 ± 0.40 ms
(-1.1 %, p = 0.018).

**2. Draft builds, opt-in approximation.** The remaining 24.2 % are all draft
builds, and they carry almost all of the cost, because their `.tolist()` lands
after the target forward has been enqueued. The host *cannot* make that value
exact under MRV2 - `sampled_token_ids` is deliberately left `None` and the
output D2H is scheduled to overlap `speculator.propose()`, so waiting for the
rejection count would cancel the overlap MRV2 exists to create. But the draft
does not need an exact value: the length only feeds `actual_seq_lengths_kv`,
i.e. how far back the draft reads, while the `slot_mapping` that decides what is
*written* is computed separately. Speculative sampling uses the same draft
distribution in the acceptance test and in the recovery distribution, so the
emitted distribution is the target's for *any* draft. Feeding the draft a
slightly-too-long KV extent costs draft *quality*, nothing else.
`VLLM_ASCEND_DSPARK_APPROX_DRAFT_KV=1` publishes `seq_lens_cpu_upper_bound` as
the draft's host mirror, marked by `seq_lens_cpu_is_approximate` (kept separate
from `seq_lens_cpu_is_exact` on purpose, so that flag keeps meaning "equals the
device tensor"). At concurrency 8: median ITL 75.4 → 50.0 ms (-34 %), output
throughput 136.0 → 179.8 tok/s (+31 %, already net of the extra decode steps),
at a cost of 6.4 % acceptance length.

**3. Recover the acceptance, still without the sync.**
`VLLM_ASCEND_DSPARK_DRAFT_KV_DEVICE_MASK=1` (requires flag 2) builds the
attention mask **on the device** from `seq_lens`. The mask hides the `[L, U)`
tail - the draft KV the previous step rolled back - which is the *entire*
mechanism by which the approximation cost anything. This is possible because
`atten_mask` is an ordinary device tensor and is not declared `ValueDepend` in
the operator's IR, unlike `actual_seq_lengths_kv`; the host therefore never
learns `L`. One structural obstacle is template selection rather than the
interface: `IsUsingFAI()` routes TND to the split-fuse template, whose
`CheckFAIMask` only accepts sparse_mode 3/4 with a 2048x2048 mask, so a
per-request mask is impossible there. Since every request in a parallel-drafting
draft build contributes exactly `num_speculative_tokens + 1` query tokens, TND
`[T, N, D]` is the same buffer as BSND `[B, S, N, D]` - relabelling it (a
`view`, no copy) reaches the general template, which accepts sparse_mode 0 with
`[B or 1, >=Q_S, >=KV_S]`. Result at concurrency 8, six arms in a palindrome
`base c0 mask mask c0 base`:

| Arm | Throughput | Median ITL | Pooled accept_len |
|---|---|---|---|
| base | 131.74 tok/s | 75.35 ms | 1.6117 |
| flag 2 only | 180.47 (+37.0 %) | 50.12 (-33.5 %) | 1.5311 (-5.00 %) |
| flags 2+3 | 189.09 (+43.5 %) | 50.02 (-33.6 %) | 1.6412 (+1.83 %) |

All six arms emit byte-identical text (`HCCL_DETERMINISTIC=true`, greedy), and
the acceptance is recovered **in full**: the flags-2+3 instances (1.6277 /
1.6547) sit entirely above the bound-only ones (1.5149 / 1.5472) and overlap
base's range (1.5893 / 1.6340).

#### Limits worth stating

- The byte-identical result is a greedy (temperature 0) one, where verification
  degenerates to an exact-match test; above temperature 0 the guarantee is
  distributional, not token-identical.
- Acceptance figures are one model / one speculator / 8 speculative tokens / one
  prompt set, and should not be assumed to transfer.
- Two instances per arm is not much: the direction is consistent and the arms do
  not overlap, but 2v2 gives no significance.
- The random dataset sits near the acceptance floor (ratio ~0.07), so the gap on
  a healthier workload is untested; concurrency above 8 is untested, which is
  exactly where the +50.8 % operator cost at n=64 would show if it is going to.
- The kernel cost of the general template is level without the sliding window
  (-2.7 % / +6.1 % / -1.1 % at n=8/32/64) and +3.1 % at n=8 / +50.8 % at n=64
  with it; mask construction is ~153-170 µs, once per step across all layers.
  Build it with a comparison, not `index_select` - gather on this hardware is
  priced by the *source* size, so selecting a few rows out of a 2048x2048
  triangle costs 444 µs.
- If flags 2+3 are not enabled, the fall-through path is byte-for-byte today's
  behaviour. Both flags default to 0.

### Does this PR introduce any _user-facing_ change?

Yes, two new environment variables, both default off:

- `VLLM_ASCEND_DSPARK_APPROX_DRAFT_KV` (0/1, default 0)
- `VLLM_ASCEND_DSPARK_DRAFT_KV_DEVICE_MASK` (0/1, default 0, requires the above)

No API, interface or default-behaviour change. With both flags at their default
the code takes exactly the same path as before.

**Unit tests.** `build_draft_tail_mask` is checked directly - per-request
lengths, the rolled-back tail hidden for every query token, and the `W+1`
sliding-window boundary - plus the fall-through conditions one by one (non-draft
build, learnable sink, non-causal, missing block table, ragged batch, token
count mismatch). Verified they fail without the change (10 failed) and pass with
it (10 passed); the whole `tests/ut/attention/test_attention_v1.py` is 43
passed.

One boundary is worth calling out because it fails quietly rather than loudly:
band mode with `pre_tokens=W` keeps `W+1` positions, `[last - W, last]`, closed
at both ends. Sweeping the offset against the operator itself gives 0.005848
relative error at `W+1` and 0.1565 at `W` - 27x the bf16 floor, and nowhere near
large enough to crash anything.

**Measurements.** All end-to-end numbers above are from A2 (8x910B4, cards 4-7,
TP4+EP, Qwen3.6-35B-A3B + DSpark speculator, MRV2, vLLM 0.28.0) with
`HCCL_DETERMINISTIC=true`. Operator-level correctness was checked against a
plain-torch reference computed from the true lengths:

| | exact lengths | flags 2+3 | bound only |
|---|---|---|---|
| n=8 | 0.003493 | 0.003493 | 0.1914 |
| n=64 | 0.002919 | 0.002919 | 0.2749 |

**Note on this PR's contents.** The full investigation - report, serve/bench
harness, operator-level and end-to-end probes, raw A/B logs, and the exact vLLM
and vllm-ascend commits it was measured on - lives on the branch
`Liuchenbing-2026:feat/16271-draft-kv-device-tail-mask` under
`issue-16271-repro/`. It is deliberately **not** included here to keep the diff
to the code and its tests; happy to attach any part of it on request.


vllm_ascend/attention/attention_v1.py
```diff
@@ -316,20 +337,42 @@ def build(
...
-            seq_lens = common_attn_metadata._seq_lens_cpu[:num_reqs]
+            seq_lens_cpu_mirror = common_attn_metadata._seq_lens_cpu[:num_reqs]
         elif common_attn_metadata.seq_lens_cpu is not None:
-            seq_lens = common_attn_metadata.seq_lens_cpu[:num_reqs]
+            seq_lens_cpu_mirror = common_attn_metadata.seq_lens_cpu[:num_reqs]
         else:
-            seq_lens = common_attn_metadata.seq_lens[:num_reqs].to("cpu")
+            seq_lens_cpu_mirror = common_attn_metadata.seq_lens[:num_reqs].to("cpu")
+        seq_lens = seq_lens_cpu_mirror
+        # Whether ``seq_lens_cpu_mirror`` still describes whatever ``seq_lens``
```
SkychenLee: why here you change the name of seq_lens to be seq_lens_cpu_mirror?  in the end you use seq_lens right?

vllm_ascend/attention/attention_v1.py
```diff
@@ -830,6 +1474,84 @@ def _get_fia_params(self, key: torch.Tensor, value: torch.Tensor, attn_metadata:
...
+            )
+            attn_metadata.draft_tail_mask_cache[cache_key] = mask
+
+        attn_output, _ = torch_npu.npu_fused_infer_attention_score(
+            query=query.view(num_reqs, query_len, self.num_heads, self.head_size),
+            key=key,
+            value=value,
+            atten_mask=mask,
+            block_table=block_table,
+            input_layout="BSND",
```
drslark, line 1542: In future, query will always be "TND".

vllm_ascend/attention/attention_v1.py
```diff
@@ -393,6 +452,9 @@ def build(
             seq_lens_list=seq_lens_list,
             max_query_len=common_attn_metadata.max_query_len,
             actual_seq_lengths_q=actual_seq_lengths_q,
+            draft_kv_upper_bound=draft_kv_upper_bound,
```
drslark, line 455: I think the root cause is host tiling.
This fix is too complex to maintain.

PR #16471, opened 2026-09-14
[BugFix] Synchronize valid token count before AIV MTP forward

Yaphets24:

When speculative decoding is used with `HCCL_OP_EXPANSION_MODE=AIV`, the asynchronous device-to-host copy of `valid_sampled_token_count` can overlap the following MTP draft forward. On Ascend TP16 expert-parallel deployments, this race can surface asynchronously as failures in AIV/HCCL work or `aclnnMoeGatingTopK`.

Synchronize the existing `valid_sampled_token_count_event` immediately before the MTP proposal, and only when AIV expansion mode is enabled. This waits for the specific small D2H copy instead of introducing a device-wide synchronization.

The existing pinned CPU metadata lifetime fix is already present on `main`; this PR only adds the synchronization that was required to eliminate the remaining race.

No API or numerical behavior changes. It prevents intermittent failures for AIV + speculative decoding. Non-AIV execution is unchanged.

Hardware reproduction on a 16-NPU Ascend 800I A3 node:

- DeepSeek-V3.1 Terminus W8A8
- TP=16, DP=1, expert parallel enabled
- async scheduling enabled
- DeepSeek MTP with 3 speculative tokens
- `HCCL_OP_EXPANSION_MODE=AIV`
- 6,000-token GSM8K prompts, concurrency 1
- 4 requests per round, 2 rounds per cold start

The final patch passed 3 independent cold starts and 6 AISBench rounds: 24/24 requests succeeded, with no server-side ERROR. Observed average TPOT was 75.39-79.73 ms.

Diagnostic comparison:

- retaining pinned H2D metadata sources alone still reproduced the failure
- synchronizing only the current stream still reproduced the failure
- a device-wide synchronization avoided the failure but was unnecessarily broad
- synchronizing this specific event avoided the failure across all repeated runs

Also checked with:

- `python -m py_compile vllm_ascend/worker/model_runner_v1.py`
- `git diff --check`

A regular CPU/mock unit test was not added because the failure requires real NPU side-stream progress and AIV collective timing; mocking the event order would only restate the conditional.


Yaphets24:
Closing after an additional isolation test disproved the proposed root cause. With DeepSeek MTP disabled while keeping AIV, TP16, EP, async scheduling, and the same workload unchanged, the failure reproduced immediately (3/4 requests returned HTTP 500). The device log again reports an AIV AllReduce SDMA exception with invalid GM access/cross-device timeout. Therefore the MTP-boundary event synchronization only changes timing and is not a complete fix. I will continue narrowing the shared AIV path before submitting a replacement.

PR #16475, opened 2026-09-14
[BugFix][KVCache] Isolate vLLM config to fix KV cache memory calculation

Lethobenthos20:

This PR fixes KV cache configuration handling when creating the attention backend by isolating the cache configuration with `replace`:

```python
cache_config=replace(self.vllm_config.cache_config),
```
The attention backend needs access to the vLLM configuration to select the appropriate backend. During backend creation, the cache configuration may be adjusted, including block_size. Without copying the configuration first, these changes can overwrite the original vllm_config.cache_config.
This causes the configured block_size to be changed unexpectedly, which results in incorrect KV cache memory-usage calculations and may lead to KV cache out-of-memory errors. V1 already isolates the cache configuration in this way; this change aligns the corresponding code path with that behavior.
For draft-model execution, the attention backend must also be created inside the set_current_vllm_config context so that it can read the correct vLLM configuration and select the appropriate backend.
However, isolation is implemented here under V1 mode.
```
vllm/v1/spec_decode/llm_base_proposer.py:1274-1310

    spec_cfg = self.speculative_config
    base = self.vllm_config

    ...

    if spec_cfg.kv_cache_dtype is not None:
        base = replace(
            base,
            cache_config=replace(
                base.cache_config,
                cache_dtype=spec_cfg.kv_cache_dtype,
            ),
        )

    return base
```

No. This is an internal configuration-isolation fix and does not introduce any API or interface changes.

The patch was validated using the CI configuration and test case associated with the original failure:
https://github.com/vllm-project/vllm-ascend/actions/runs/33945552567/job/101254066304
```
ValueError: To serve at least one request with the model's max seq len (196608),
(78.6 GiB KV cache is needed, which is larger than the available KV cache memory
(35.72 GiB). Based on the available memory, the estimated maximum model length is
88448.
```

Additional validation was performed on the following models and configurations:
- Qwen3.5-122B-A10B-W8A8
- DeepSeek V4.1


Ronald1995 (approved):
LGTM

weijinqian0 (approved):

PR #16477, opened 2026-09-14
[BugFix] Fix SP/PCP token layout in MoE AllGather

wzx0726:

Fix sequence-parallel MoE ALLGATHER with PCP enabled. SP length metadata describes DP x SP, while the EP group spans DP x PCP x TP. Discarding the shorter metadata selects a DP-only reshape that cannot represent the gathered tensor.

Extend the existing `_get_ep_local_sizes` helper to expand lengths in DP/PCP/TP order when the topology matches. When the layout cannot be expanded, retain the existing `None` return and caller fallback instead of introducing a new exception. Existing gather, reduce-scatter and fake implementations share the mapping. Remove the redundant PCP collectives from the EP sequence-parallel prepare/finalize path. The non-SP path is retained. No new persistent state, flags or public interfaces are introduced.

Fixes a shape error in the SP+PCP MoE ALLGATHER path and removes duplicate communication. No configuration changes.

Added regression tests for multiple DP/PCP/SP combinations, uneven token counts, token ordering, gather/reduce-scatter round trips, fake shapes, fallback for unmatched topology and absence of redundant PCP collectives.

The latest compatibility update changes the unmatched-layout regression test to assert `None`. `git diff --check` and Python AST syntax checks passed for the update, and both changed-file diffs were reviewed. `bash format.sh ci` could not run its checks because pre-commit is not installed in the editing environment. Runtime tests have not been rerun on this exact PR head; targeted unit tests and NPU validation remain pending.


jiaqi-lee (approved):

realliujiaxu (dismissed):
LGTM

wzx0726:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

linfeng-yuan:
@weiguihua2 PTAL

vllm_ascend/ops/fused_moe/prepare_finalize.py
```diff
@@ -425,26 +425,6 @@ def _prepare_with_ep_group(
         if pertoken_scale is not None:
             pertoken_scale = torch.ops.vllm.maybe_all_gather_and_maybe_unpad(pertoken_scale)
 
-        if self.moe_config.pcp_size > 1:
```
lcfenglinwan: check if can removed

vllm_ascend/ops/register_custom_ops.py
```diff
@@ -24,8 +25,21 @@ def _get_ep_local_sizes(dp_metadata, ep_group) -> list[int] | None:
     except (AssertionError, AttributeError):
         return None
 
-    if local_sizes is None or len(local_sizes) != ep_group.world_size:
+    if local_sizes is None:
         return None
+    if len(local_sizes) != ep_group.world_size:
+        pcp_size = get_pcp_group().world_size
+        dp_size = get_dp_group().world_size
+        if len(local_sizes) * pcp_size != ep_group.world_size or len(local_sizes) % dp_size:
```
weiguihua2, line 33: Should this side report an error or return None? Reporting an error is inconsistent with previous behavior.

weiguihua2 (approved):

vllm_ascend/ops/register_custom_ops.py
wzx0726: The path has been changed to the original None path.

wzx0726:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

linfeng-yuan (approved):

PR #16485, opened 2026-09-14
[BugFix][Model] Fix DeepSeek V4 load-format=dummy profile run failures

Debonex:

This PR fixes two failures when serving DeepSeek V4 with --load-format=dummy (profile run):

1. Hash MoE routing table (tid2eid) now uses torch.rand(...).topk(...) instead of a torch.zeros/torch.randint-style initialization. topk over random scores guarantees each token's top-k expert ids are unique within a row. The MC2 dispatch/combine kernels assume a token never routes to the same expert twice; duplicate ids sporadically fail them during the dummy profile run. topk avoids duplicates entirely.

2. wo_a weight 2D→3D reshape is moved from AscendColumnParallelLinear.weight_loader to AscendUnquantizedLinearMethod.process_weights_after_loading.
Before this change, the reshape only ran in weight_loader, so the two load paths diverged:
- load-format=auto: 
    - init param (2D)
    - weight_loader reshapes 2D→3D
    - process_weights_after_loading no-op
    - capture: gmm receives 3D and runs normally.
- load-format=dummy: 
    - init param (2D)
    - weight_loader never runs
    - process_weights_after_loading no-op
    - capture: gmm receives 2D and errors.

  After moving the reshape into process_weights_after_loading, both paths complete the 2D→3D conversion:
The shared condition is extracted into _should_reshape_wo_a_to_3d, reused by both weight_loader and process_weights_after_loading.

No


Debonex:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/models/deepseek_v4/model.py
```diff
@@ -339,14 +339,23 @@ def __init__(
...
             )
         if self.hash:
-            # Use zeros instead of empty to avoid garbage values causing
-            # invalid memory access in dummy mode (--load-format="dummy")
-            self.gate.tid2eid = nn.Parameter(
-                torch.zeros(
+            # Use `torch.rand(...).topk(...)` instead of `torch.randint` so
+            # each token's top-k expert ids are unique: MC2 dispatch/combine
+            # fails in the dummy profile run if a token repeats an expert id.
+            token_to_expert = (
```
Ronald1995: [P1] Avoid allocating a `[vocab_size, n_routed_experts]` tensor just to build the dummy routing table.

The only requirement here is that each row contains `num_experts_per_tok` distinct, valid expert IDs. `rand(...).topk(...)` makes the initialization O(vocab_size * n_routed_experts) in both memory and compute, even though the resulting table is only O(vocab_size * top_k).

For DeepSeek-V4-Flash this creates a temporary `129280 x 256` FP32 tensor (~126 MiB), and for V4-Pro it is `129280 x 384` (~189 MiB), before accounting for the `topk` values/indices. This also runs during normal checkpoint loading, where the generated `tid2eid` will subsequently be overwritten by the checkpoint.

Could we generate the dummy mapping directly in O(vocab_size * top_k), for example by assigning consecutive expert IDs modulo `n_routed_experts` for each token? That would still guarantee unique IDs per row, provide a balanced routing distribution, and avoid depending on global RNG state.

Please also add a regression test asserting that every dummy `tid2eid` row contains exactly `num_experts_per_tok` unique IDs.

Ronald1995 (changes requested):

Ronald1995 (approved):
LGTM

Ronald1995 (approved):

kunpengW-code (approved):

PR #16501, opened 2026-09-14
[Feature][Attention] Add FA3 attention backend

CXY-Katrina:

- Add an FA3 inference backend using `flash_attn_with_kvcache` and device-side GQA tiling.
- Automatically select it for supported Qwen3 dense/MoE decoder self-attention on A2/A3.
- Use the paged FA3 path for prefill, decode, chunked prefill, prefix cache, and mixed P/D batches.
- Share scheduler metadata across compatible layers and use `num_splits=0` for automatic splitting.
- Preserve stable graph buffers while excluding padded requests from FlashDecode selection.
- Fall back to the existing FIA path for missing operators, unsupported configurations, 310P/A5, and C8.

Supported Qwen3 models use FA3 automatically when `flash_attn_npu_3` is installed; no new user option is introduced.

- Backend/platform/config unit tests: **493 passed, 1 skipped**.
- NPU backend tests: **28 passed**.
- Historical accuracy: Qwen3-235B GPQA **143/198 = 72.22**, consistent with the **72.12** baseline (`top_p=0.95`, `top_k=20`, `temperature=0.6`).
- Historical same-workload performance (Qwen3-235B, bs=1, 3500→8): FA3 TTFT **708.3 ms** vs FIA **702.3 ms**, and output throughput **8.51** vs **8.81 token/s**; end-to-end performance is broadly on par with FIA.

### DIFF
--- a/vllm_ascend/attention/flash_attention_v3.py
+++ b/vllm_ascend/attention/flash_attention_v3.py
@@ -1,0 +1,273 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+
+"""Dense Ascend attention using flash-attention-npu's device-side FA3 tiling."""
+
+from dataclasses import dataclass, field
+
+import torch
+from flash_attn_npu_3 import flash_attn_varlen_func, flash_attn_with_kvcache, get_scheduler_metadata
+from vllm.config import VllmConfig
+from vllm.v1.attention.backend import AttentionCGSupport, AttentionMetadataBuilder, AttentionType
+from vllm.v1.kv_cache_interface import AttentionSpec
+
+from vllm_ascend.ascend_forward_context import _EXTRA_CTX
+from vllm_ascend.attention.attention_v1 import (
+    AscendAttentionBackend,
+    AscendAttentionBackendImpl,
+    AscendAttentionState,
+    AscendMetadata,
+)
+from vllm_ascend.attention.utils import AscendCommonAttentionMetadata
+from vllm_ascend.distributed.kv_transfer.kv_pool.ascend_store.attention_fence import record_attention_compute_start
+
+
+class AscendFlashAttentionBackend(AscendAttentionBackend):
+    @staticmethod
+    def get_impl_cls() -> type["AscendFlashAttentionImpl"]:
+        return AscendFlashAttentionImpl
+
+    @staticmethod
+    def get_builder_cls() -> type["AscendFlashAttentionMetadataBuilder"]:
+        return AscendFlashAttentionMetadataBuilder
+
+
+@dataclass
+class AscendFlashAttentionMetadata(AscendMetadata):
+    # Shared only within this execution, keyed by the operator's static parameters.
+    scheduler_metadata: dict[tuple, torch.Tensor] = field(default_factory=dict)
+
+
+class AscendFlashAttentionMetadataBuilder(AttentionMetadataBuilder[AscendFlashAttentionMetadata]):
+    # A single TND invocation handles mixed query lengths, including verification
+    # of multiple speculative tokens. No decode-first batch reordering is needed.
+    reorder_batch_threshold = None
+
+    def __init__(
+        self,
+        kv_cache_spec: AttentionSpec,
+        layer_names: list[str],
+        vllm_config: VllmConfig,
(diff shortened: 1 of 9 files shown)


vllm_ascend/attention/flash_attention_v3.py
```diff
@@ -0,0 +1,278 @@
...
+        )
+        if source_key not in self.graph_buffers:
+            self.graph_buffers[source_key] = (
+                torch.empty(self.max_num_reqs + 1, dtype=torch.int32, device=self.device),
+                torch.empty(self.max_num_reqs, dtype=torch.int32, device=self.device),
+                torch.empty(
+                    (self.max_num_reqs, common.block_table_tensor.shape[1]), dtype=torch.int32, device=self.device
+                ),
+            )
+        query_start_loc, seq_lens, block_tables = self.graph_buffers[source_key]
```
yiz-liu, line 112: Shouldn't the buffers of these three tensors be fixed? Why do we need another self.graph_buffers?

vllm_ascend/attention/flash_attention_v3.py
```diff
@@ -0,0 +1,278 @@
...
+        # The runner may add a dummy FIA request beyond the seq_lens/block-table
+        # capacity. FA3 needs no such request: zero query lengths skip padding.
+        query_start_loc.fill_(common.num_actual_tokens)
+        query_start_loc[: num_reqs + 1].copy_(common.query_start_loc[: num_reqs + 1])
+        query_start_loc.clamp_(min=0, max=common.num_actual_tokens)
```
yiz-liu, line 117: So we can remove these after FIA is deprecated?

vllm_ascend/attention/flash_attention_v3.py
```diff
@@ -0,0 +1,278 @@
...
+            scheduler_metadata=scheduler_metadata,
+        )
+
+    def build_for_graph_capture(
+        self,
+        common_attn_metadata: AscendCommonAttentionMetadata,
+        attn_state: AscendAttentionState = AscendAttentionState.DecodeOnly,
+    ) -> AscendFlashAttentionMetadata:
+        metadata = self.build(0, common_attn_metadata)
+        metadata.attn_state = attn_state
```
yiz-liu, line 179: Is this attn_state still needed?

yiz-liu (dismissed):

vllm_ascend/attention/flash_attention_v3.py
CXY-Katrina: Yes. The runner-owned `query_start_loc`, `seq_lens`, and `block_table_tensor` buffers are already graph-stable. I removed the extra `self.graph_buffers` allocation/copy layer and now use those buffers directly. Only the FA3 scheduler metadata keeps dedicated stable storage because its tensor address is also captured by the graph.

vllm_ascend/attention/flash_attention_v3.py
CXY-Katrina: Updated. The FA3 builder now uses the runner-owned buffers directly and limits scheduler reads with `active_reqs`, so it no longer copies or sanitizes the FIA dummy request. The FIA-specific handling and comment have been removed.

vllm_ascend/attention/flash_attention_v3.py
CXY-Katrina: FA3 does not need `attn_state` because all supported scheduling states use the same paged `flash_attn_with_kvcache` path. I removed it from `AscendFlashAttentionMetadata` and stopped assigning it during graph capture. The optional method argument remains only for compatibility with the speculative proposer caller/interface and is intentionally ignored.

CXY-Katrina:
/nightly Qwen3-32B-QuaRot
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/35717111357)

CXY-Katrina:
/nightly DeepSeek-V4-Flash-W8A8-A3
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/35718096754)

CXY-Katrina:
/nightly Qwen3-32B-QuaRot
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/35720902336)

PR #16515, opened 2026-09-14
[v0.27.1rc][Performance] Optimize MiniMax-M3 A5 AscendC index score decode

HaoxinZong:

This PR enables the optimized AscendC `MsaIndexScore` path for MiniMax-M3 A5 decode on `releases/v0.27.1rc`.

It contains two related changes:

- selectively ports the missing `kvChunks` short-M/long-KV optimization and A2/A5 fixes from ops-transformer PR 11754 while preserving the vLLM-specific adapter;
- scores the complete replicated index-K block table on A5, avoiding the TP block-splitting collectives that consumed the AscendC kernel gain;
- applies `init_blocks` and `local_blocks` in the AscendC epilogue for the full-table path;
- reuses the existing A5 TopK postprocess kernel to produce the final indices and selection counts.

A2/A3 keep the existing TP-sharded decode path. The TP-sharded path also keeps candidate forcing in its TopK stage because the operator attributes do not carry a global block offset.

Matched Ascend 950 profiling used MiniMax-M3, TP4 x DP2, decode-only requests, and the maximum TP-rank duration within each synchronous DP step. The complete layer path is measured after `ScatterPaCache` through the final TopK output.

| Input / output | Original A5 Triton | Optimized AscendC | Change | Speedup |
|---|---:|---:|---:|---:|
| 16K / 1K | 22.802 us/layer | 20.579 us/layer | -9.75% | 1.108x |
| 128K / 1K | 57.011 us/layer | 26.735 us/layer | -53.11% | 2.132x |

At 128K, this saves 30.276 us per layer, or about 1.726 ms across the 57 sparse-attention layers in one model step. These are IndexScore layer-path device-time observations, not end-to-end latency claims.

Yes. MiniMax-M3 uses AscendC `MsaIndexScore` for both prefill and decode on A5, with the optimized full-table decode path.

- `python -m pytest -q tests/ut/models/minimax_m3/test_msa_m3.py` — 64 passed on the release A5 container
- `python -m py_compile` for the modified model, NPU helper, and unit-test files
- `git diff --check`
- Ascend 950 `MsaIndexScore` package build for the operator-sync commit
- A5 service profiling with two 16K/1K and two 128K/1K requests; all requests returned HTTP 200 with exact prompt/completion token counts, and all eight DP2 x TP4 rank traces were generated

The service used for profiling was stopped after trace collection.


HaoxinZong:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #16189, opened 2026-09-10
[BugFix] Precast MoE gate weight_fp32 to avoid aclop Cast

LQDLove:
### What this PR does / why we need it?

Nightly `qwen3-vl-235b-a22b-instruct-w8a8` (A3, `FULL_DECODE_ONLY` ACLGraph) fails because internal-router MoE recomputes logits with:

`gate.weight.to(torch.float32)`

That emits a per-forward `aclop Cast`. DeepSeek-V4 and 310P already avoid this by setting `gate.precast_fp32_weight = True` so `AscendUnquantizedLinearMethod.process_weights_after_loading` materializes `weight_fp32` at load time.

This PR applies the same precast in `AscendMoERunner.__init__` for all Ascend platforms, and routes the hot path through `_gate_weight_fp32()` so ACLGraph capture no longer depends on a live Cast.

### Does this PR introduce _any_ user-facing change?

No.

### How was this patch tested?

- Code-path review against 310P `AscendMoERunner310` and DeepSeek-V4 `precast_fp32_weight`.
- Nightly should re-run `Qwen3-VL-235B-A22B-Instruct-W8A8`.

Made with [Cursor](https://cursor.com)

- vLLM main: https://github.com/vllm-project/vllm/commit/a97dacb7106ee49f39f3d1fc6ae1800ff724e01d

PR #16521, opened 2026-09-14
[BugFix] Fallback to gate forward when gate lacks weight_fp32

LQDLove:

In `AscendMoERunner`, the shared-expert router path assumed `gate.weight_fp32` always exists (introduced by #16189). For gate implementations that do not expose the precast `weight_fp32` attribute, this raises an `AttributeError`. This PR falls back to invoking the gate module directly and extracting `router_logits` from its output, so the fused MoE path works for all gate implementations.

No.

- `ruff check` and `ruff format --check` pass on the modified file.
- Existing fused MoE unit tests.


LQDLove:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

weijinqian0 (approved):

PR #16347, opened 2026-09-11
[Feature][Spec Decode] Support Kimi K3 MLA DSpark on MRV2

kk-ss1999:
### What this PR does / why we need it?

Support Kimi K3 MLA DSpark eager execution and ACLGraph on Model Runner V2.

Keep MLA graph support in the shared `AscendDSparkSpeculator`, following the shared-speculator approach in #13027. Remove `dspark/mla.py` and its separate class routing. Select dense MLA metadata internally by draft capability: update query lengths inside `.decode` for dense MLA and preserve the existing GQA, compressed MLA, and SFA paths. Capture and replay retain the dense MLA speculative-decode state and padded query boundaries. Remove the subclass-only prefill hook and custom metadata exceptions.

Reuse upstream Kimi AttnRes auxiliary capture, which already excludes the 3-D residual bank. The duplicate Ascend raw-capture helper has been removed; model-owned raw-mode configuration and validation remain.

Depends on:

- #16312: Kimi K3 MRV2 target graph and PD support.
- #16318: Shared MRV2 DSpark runtime and GQA DSpark support.

### Does this PR introduce _any_ user-facing change?

Adds Kimi K3 MLA DSpark support on top of these dependencies. No new flags or changes to GQA behavior.

### How was this patch tested?

- Shared runtime (`c37408a4d`): **29 isolated CPU regression cases passed**, executing the shared initialization, capture/replay methods, capability predicates, and repository test assertions with heavy imports/upstream construction stubbed. Covers dense MLA, GQA, actual DSA/SFA metadata types, compression fields, kpool, target/draft independence, padded query lengths, prefill flags, empty metadata, and capture restoration after an exception.
- Raw auxiliary capture (`6dc953122`): **12 isolated CPU regression cases passed** against the upstream capture methods from vLLM `b2f685834` and `a97dacb71`. Covers selected/unselected boundaries and absent/3-D residuals; both restored forward call sites were checked.
- Ruff 0.14.0 check/format and `git diff --check`: passed.
- Integration used for the full-weight results below: Ascend main `fb4c9bdf2` + #16318 `dee63c9a9` (includes #16312) + this PR at `121f9faf`; vLLM `a97dacb71`.
- Existing dependency-baseline regression failures: `test_quarot_loaded_weights_survive_upstream_sharing` and `test_normalize_preserves_distinct_mamba_layouts`.

[...]

PR #16527, opened 2026-09-14
[CI] main2main vllm 0914 e82794f

zhangxinyuehfad:


vllm_ascend/worker/v2/pp_utils.py
```diff
@@ -25,13 +23,13 @@
 
 
 def use_legacy_spec_pp() -> bool:
```
LostFox11: 后续会将整个旧方案日落下架

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -325,16 +327,15 @@ def execute_model(
...
-            )
+        output = super().execute_model(
+            scheduler_output,
+            intermediate_tensors=intermediate_tensors,
+            dummy_run=dummy_run,
+            skip_attn_for_dummy_run=skip_attn_for_dummy_run,
+            is_profile=is_profile,
+            context_len=context_len,
+            valid_dummy_state_slots=valid_dummy_state_slots,
+        )
```
li1how, line 338: https://github.com/vllm-project/vllm-ascend/pull/17174 has been merged, please rebase it.

vllm_ascend/worker/v2/pcp_manager.py
```diff
@@ -251,26 +253,15 @@ def _full_decode_requests_are_token_sized(self, global_batch: AscendInputBatch)
...
-            sum(
-                segment.num_tokens
-                for segment in self._get_rank_segments(
-                    rank, num_scheduled_tokens, num_computed_tokens, is_prefilling, query_start_loc
-                )
-            )
-            for rank in range(self.pcp_world_size)
-        )
+        # Upstream #53515 provides get_num_tokens_for_dispatch; the 0.28.0-only
+        # Ascend partition workaround is dropped.
```
li1how, line 257: This function can be deleted directly.

vllm_ascend/models/deepseek_v41/vl_model.py
```diff
@@ -26,6 +29,8 @@
 from .model import AscendDeepseekV41LLMForCausalLM
 from .vision import DeepseekV41Aligner, DeepseekV41ViT
 
+IMAGE_PAD_ID: int | None = None
```
zhangxinyuehfad: !!!

vllm_ascend/lora/punica_npu.py
```diff
@@ -581,12 +581,54 @@ def add_lora_logits(
         x = x.view(-1, x.shape[-1])
         r = lora_b_stacked.size(-1)
 
+        indices = torch.narrow(self._sampler_indices, 0, 0, x.size(0))
```
zhangxinyuehfad: !

vllm_ascend/attention/mla_v1.py
```diff
@@ -1654,14 +1654,23 @@ def _forward_decode(
...
+                ]
+                and self.speculative_config is not None
+            )
+            # vLLM main (#56181) restructured the draft decode metadata flow;
+            # the draft decode graph capture then records a non-TND layout,
+            # while replay still passes cumulative actual_seq_lengths_q. Force
+            # TND for the draft (its metadata uses cumulative lengths). Use the
+            # forward-context flag, not self.is_draft_model: the draft MLA impl
+            # shares the target's vllm_config, so runner_type is "generate".
+            or _EXTRA_CTX.is_draft_model
```
zhangxinyuehfad: !

vllm_ascend/ops/mla.py
```diff
@@ -186,6 +186,12 @@ def __init__(
             ascend_indexer = IndexerWrapper(mla_modules.indexer, self.qk_rope_head_dim)
         else:
             ascend_indexer = None
+        # vLLM main (#56157) gates PCP+DCP on MLAAttention.supports_pcp_dcp
+        # (default False) and the guard runs inside __init__, so the flag must
+        # be visible on the class before construction. Ascend implements PCP+DCP
+        # inside its attention impls (AscendSFAPCPDCPImpl / AscendMlaDCPImpl) and
+        # forwards through a custom op, so opt in on main.
+        MLAAttention.supports_pcp_dcp = True
```
zhangxinyuehfad: 后端默认打开

vllm_ascend/attention/mla_v1.py
drslark, line 1673: What is this for?

vllm_ascend/ops/rotary_embedding.py
```diff
@@ -336,15 +336,19 @@ def __init__(
         beta_fast: int = 32,
         beta_slow: int = 1,
         apply_yarn_scaling: bool = True,
+        mscale: float | None = None,
+        mscale_all_dim: float | None = None,
+        attention_factor: float | None = None,
```
zhangxinyuehfad: 确认参数

vllm_ascend/patch/worker/patch_v2/patch_attn_utils.py
```diff
@@ -21,4 +24,21 @@ def _get_ascend_sfa_indexer_backend(_self):
...
+vllm.v1.worker.gpu.attn_utils.bind_kv_cache_to_layers = bind_kv_cache_to_layers
+
+# vLLM main (#53781) also builds self.kv_caches by filtering
+# cache.device, assuming single-tensor allocations; Ascend allocates
+# per-layer (k, v) tuples. Expose the first tensor for that filter.
+# The binding inside init_kv_cache still uses the raw allocations.
+_orig_init_kv_cache = vllm.v1.worker.gpu.model_runner.init_kv_cache
+
+
+def _ascend_init_kv_cache(*args, **kwargs):
```
wangxiyuan, line 38: note

vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
```diff
@@ -489,18 +490,117 @@ def _build_draft_attn_metadata(  # type: ignore[misc]
...
+                )
+        if attn_metadata is not None:
+            # Ascend-specific: force DecodeOnly attention state for the draft model.
+            for metadata in attn_metadata.values():
+                if metadata is None:
+                    continue
+                metadata.attn_state = AscendAttentionState.DecodeOnly
+        return attn_metadata
+
+    def _build_uniform_attn_metadata(  # type: ignore[misc]
```
drslark, line 529: `_build_uniform_attn_metadata` should be in DraftModelSpeculator's speculator.py.

vllm_ascend/patch/worker/patch_v2/patch_uva.py
```diff
@@ -145,9 +145,7 @@ def cpu(self):
     def np(self):
         return self._np if is_uva_available() else MonitoredNumPyArray(self._np, self._mark_cpu_modified)
 
-    @property
-    def uva(self):
-        """Get the device data of the buffer."""
+    def _sync_dirty_rows(self) -> None:
```
wangxiyuan, line 148: note

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -85,7 +85,9 @@ def combine_hidden_states(self, aux_hidden_states: torch.Tensor) -> torch.Tensor
     DSparkDeepseekV4ForCausalLM,
 )
 
-if not vllm_version_is("0.29.0"):
+# DeepSeek V4.1 is skipped on Triton-less builds such as 310P (newer main's
+# sparse_mqa_logits calls tl.constexpr at module scope, crashing the import).
+if HAS_TRITON:
```
wangxiyuan, line 90: note

vllm_ascend/spec_decode/llm_base_proposer.py
```diff
@@ -388,10 +384,12 @@ def load_model(self, model: nn.Module) -> None:
 
         if self.supports_mm_inputs:
             # Match upstream: a multimodal target can use a text-only drafter.
-            try:
-                dummy_input_ids = torch.tensor([[1]], device=self.input_ids.device)
-                self.model.embed_input_ids(dummy_input_ids, multimodal_embeddings=None)
-            except (NotImplementedError, AttributeError, TypeError):
+            if not _draft_embed_accepts_mm(getattr(self.model, "embed_input_ids", None)):
```
wangxiyuan, line 387: note

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
```diff
@@ -140,29 +148,6 @@ def build_draft_attn_metadatas(self, num_reqs_padded, seq_lens_cpu_upper_bound):
 
         return [self._update_draft_attn_metadata(attn_metadata, num_reqs_padded)]
 
-    def _build_draft_attn_metadata(self, *, num_reqs_padded, **kwargs):
```
drslark, line 143: I think we should extend vllm's `_build_uniform_attn_metadata` here.

voidvelocity (approved):
LoRA modification lgtm

nwpu-zxr (approved):

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -238,10 +233,17 @@ def _restore_replicated_draft_target_states(self) -> None:
...
 
+        # vLLM main captures draft_hidden_states before maybe_restore_pcp_for_sampling,
+        # so a replicated draft would read the PCP-local target output. Restore
+        # it to the global layout up front.
+        if state.hidden_states is not None:
+            state = state._replace(hidden_states=pcp_manager.restore_hidden_states(state.hidden_states))
+
         aux_hidden_states = state.aux_hidden_states
         if aux_hidden_states:
             restored_aux_hidden_states = pcp_manager.restore_hidden_states(torch.cat(aux_hidden_states, dim=-1))
```
weiguihua2, line 244: 上游v0.30.0 sample_tokens方法里面新增恢复了auxiliary hidden states，ascend vllm中也有这段，重复了

vllm_ascend/lora/punica_npu.py
```diff
@@ -581,12 +581,54 @@ def add_lora_logits(
...
+        if y.dtype in (torch.half, torch.bfloat16):
+            self.bgmv_expand(buffer, lora_b_stacked, y, indices, add_inputs=True)
+        else:
+            # Ascend bgmv_expand only writes half/bf16 outputs. Compute
+            # the LoRA delta into a matching workspace and merge it back
+            # so fp32 classification outputs (classifier LoRA, #53555)
+            # work on vLLM main.
+            y_half = y.to(torch.bfloat16)
+            self.bgmv_expand(buffer, lora_b_stacked, y_half, indices, add_inputs=True)
+            y.copy_(y_half.to(y.dtype))
```
Tflowers-0129, line 607: 这里精度强制转换又转回去，且出现了bf16硬编码，而且这个分支是不是在某些情况一定走到

op_dtype = lora_b_stacked.dtype
delta = torch.zeros(
    y.shape,
    dtype=op_dtype,
    device=y.device,
)
self.bgmv_expand(
    buffer,
    lora_b_stacked,
    delta,
    indices,
    add_inputs=True,  # delta 初始为 0
)
y.add_(delta.to(y.dtype))

vllm_ascend/patch/worker/patch_v2/patch_attn_utils.py
```diff
@@ -21,4 +24,21 @@ def _get_ascend_sfa_indexer_backend(_self):
...
+    d = _orig_init_kv_cache(*args, **kwargs)
+    return {name: (v[0] if isinstance(v, (tuple, list)) and v else v) for name, v in d.items()}
```
Tflowers-0129, line 40: i think this change need explain

vllm_ascend/attention/mla_v1.py
linfeng-yuan, line 1673: Please make sure why we need this line. Without #56181, how drafter model with mla attention execute here?

vllm_ascend/worker/v2/model_runner.py
zhangxinyuehfad: done

vllm_ascend/worker/v2/pcp_manager.py
zhangxinyuehfad: deleted

vllm_ascend/attention/mla_v1.py
zhangxinyuehfad: Without #56181 (v0.29.0): the drafter was covered implicitly. Draft decode metadata was built
only via _build_draft_attn_metadata, whose Ascend override forces attn_state = DecodeOnly on
every layer (this forcing predates #56181), and self.speculative_config is not None since the
draft impl shares the target's vllm_config — so the first branch above already selected TND
for the drafter; no extra term was needed.

After #56181: upstream renamed/split the builder into _build_uniform_attn_metadata /
_build_attn_metadata and the draft decode loop calls the new names directly, bypassing the
legacy override where the DecodeOnly forcing lived. attn_state is an Ascend-specific field
(defaults to None), so during draft decode graph capture it was None, the first branch did
not match, and the captured graph baked a non-TND layout while replay feeds cumulative
actual_seq_lengths_q — FIA rejects that (actualSeqLengthsQ > Q_S). Hence
`or _EXTRA_CTX.is_draft_model`: the per-forward drafter flag (set by every proposer via
set_ascend_forward_context and during draft ACL-graph capture), making the TND choice
independent of metadata attn_state.

Note there is now overlap: 1bf52e8cb re-hooked the DecodeOnly forcing onto the new builders,
so the draft metadata carries DecodeOnly again. I kept the explicit flag as the structural
guarantee — layout must not depend on every metadata builder setting attn_state, and it is
the same signal update_graph_params uses to identify draft forwards. Happy to drop it if you
prefer relying on attn_state alone.

vllm_ascend/lora/punica_npu.py
zhangxinyuehfad: It's better, changed it .

vllm_ascend/worker/v2/spec_decode/dflash/speculator.py
```diff
@@ -25,10 +26,17 @@ class AscendDFlashSpeculator(DFlashSpeculator):
...
         with build_attn_metadata_wrapper():
-            attn_metadata = self._build_draft_attn_metadata(
+            # vLLM main (#56181) replaced _build_draft_attn_metadata with
+            # _build_uniform_attn_metadata (BatchExecutionDescriptor).
+            batch_desc = BatchExecutionDescriptor(
+                cg_mode=CUDAGraphMode.FULL,
+                num_tokens=num_tokens_padded,
+                num_reqs=num_reqs_padded,
+            )
+            attn_metadata = self._build_uniform_attn_metadata(
```
drslark, line 36: I reviewed it.
I think it is right for dflash.

zhangxinyuehfad: thanks

vllm_ascend/worker/v2/model_runner.py
zhangxinyuehfad: Confirmed, removed — upstream #56107 already restores aux_hidden_states in sample_tokens.
hidden_states pre-restore kept (upstream captures draft_hidden_states earlier). UT updated.

vllm_ascend/attention/mla_v1.py
zhangxinyuehfad: Without #56181 (v0.29.0): the drafter was covered implicitly. Draft decode metadata was built
only via _build_draft_attn_metadata, whose Ascend override forces attn_state = DecodeOnly on
every layer (this forcing predates #56181), and self.speculative_config is not None since the
draft impl shares the target's vllm_config — so the first branch above already selected TND
for the drafter; no extra term was needed.

After #56181: upstream renamed/split the builder into _build_uniform_attn_metadata /
_build_attn_metadata and the draft decode loop calls the new names directly, bypassing the
legacy override where the DecodeOnly forcing lived. attn_state is an Ascend-specific field
(defaults to None), so during draft decode graph capture it was None, the first branch did
not match, and the captured graph baked a non-TND layout while replay feeds cumulative
actual_seq_lengths_q — FIA rejects that (actualSeqLengthsQ > Q_S). Hence
or _EXTRA_CTX.is_draft_model: the per-forward drafter flag (set by every proposer via
set_ascend_forward_context and during draft ACL-graph capture), making the TND choice
independent of metadata attn_state.

Note there is now overlap: 1bf52e8cb re-hooked the DecodeOnly forcing onto the new builders,
so the draft metadata carries DecodeOnly again. I kept the explicit flag as the structural
guarantee — layout must not depend on every metadata builder setting attn_state, and it is
the same signal update_graph_params uses to identify draft forwards. Happy to drop it if you
prefer relying on attn_state alone.

vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
drslark, line 476: Is `_build_draft_attn_metadata` only in 0.29.0?

vllm_ascend/patch/worker/patch_v2/patch_attn_utils.py
zhangxinyuehfad: Ascend allocates per-layer (k, v) tuples, while upstream model_runner filters self.kv_caches with cache.device, assuming each value is a single tensor (a tuple has no .device and would error). This just unwraps the tuple to its first tensor to satisfy that filter; the real (k, v) pairs are already bound to the layers via bind_kv_cache_to_layers inside init_kv_cache, so semantics are unchanged.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
```diff
@@ -140,29 +148,6 @@ def build_draft_attn_metadatas(self, num_reqs_padded, seq_lens_cpu_upper_bound):
...
-        if self.attn_architecture not in ("GQA", "MLA"):
-            return super()._build_draft_attn_metadata(num_reqs_padded=num_reqs_padded, **kwargs)
-
-        # This kwargs["num_tokens_padded"] is only useful in eager/PIECEWISE.
-        # TODO: Replace this temporary padding workaround with upstream #56181's
-        # actual-token metadata and MLA input slicing for non-FULL execution.
-        num_tokens_padded = kwargs["num_tokens_padded"]
```
drslark, line 150: I think you should wrap `_build_draft_attn_metadata` from vllm.
It avoided a bug.

drslark: This bug is solved in https://github.com/vllm-project/vllm-ascend/pull/16347.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
zhangxinyuehfad: Agreed. I've moved the actual_seq_lengths_q fix into an override of _build_uniform_attn_metadata instead of post-processing in build_draft_attn_metadatas, matching the autoregressive speculator's pattern.

vllm_ascend/worker/v2/aclgraph_utils.py
```diff
@@ -135,14 +136,18 @@ def __init__(  # type: ignore[misc]
...
         super().__init__(
             vllm_config,
             device,
             cudagraph_mode,
             decode_query_len,
             lora_capture_cases=lora_capture_cases,
             varlen_decode=varlen_decode,
+            ubatch_runner=ubatch_runner,
         )
         self.breakable_cg_runner: BreakableACLGraphWrapper | None = None
```
weiguihua2, line 163: _prepare_pcp_inputs_to_capture capture的时候也要执行maybe_prepare_dcp_local_seq_lens，上游有这段逻辑

vllm_ascend/worker/v2/spec_decode/autoregressive/speculator.py
zhangxinyuehfad: It's an internal Ascend adapter, still used on vLLM 0.30.0: our callers use the old padded-count args and it forwards to the new _build_uniform_attn_metadata/_build_attn_metadata hooks.

vllm_ascend/worker/v2/aclgraph_utils.py
zhangxinyuehfad: I will add it.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
zhangxinyuehfad: ok

vllm_ascend/attention/mla_v1.py
zhangxinyuehfad: ditto

vllm_ascend/patch/worker/patch_v2/patch_attn_utils.py
zhangxinyuehfad: ditto

vllm_ascend/patch/worker/patch_v2/patch_uva.py
zhangxinyuehfad: Ascend NPU can't expose a UVA view of pinned CPU memory (unless PYTORCH_NPU_ALLOC_CONF=pinned_mem_register:True). This wrapper keeps a pinned CPU copy + a device copy and syncs only dirty rows on uva(): writes via the CPU/NumPy view are tracked by row; _sync_dirty_rows uses a contiguous non-blocking copy_ fast path for the common dst[:n]=x case and a re-pinned indexed copy for sparse writes. uva is a method (not property) to match upstream #56908.

vllm_ascend/spec_decode/llm_base_proposer.py
zhangxinyuehfad: This replaces the runtime probe (call embed_input_ids(dummy, multimodal_embeddings=None) in a try/except) with signature introspection via _draft_embed_accepts_mm. Drafts sharing the target embedding (e.g. K3 DSpark on a kv_consumer) have no own embed_tokens before the sharing step, so the runtime probe trips an uncaught AssertionError; introspection avoids calling the embed at load time. The result gates mm-kwarg forwarding in load_model and both draft-forward paths.

vllm_ascend/spec_decode/llm_base_proposer.py
zhangxinyuehfad: Removed the vllm_version_is("0.29.0") check because the branch now only uses vLLM v0.30.0, so it's always False — dead code. if HAS_TRITON: is the real guard now.

vllm_ascend/attention/mla_v1.py
zhangxinyuehfad: ditto

li1how (approved):

vllm_ascend/worker/v2/pcp_manager.py
```diff
@@ -384,6 +403,22 @@ def restore_hidden_states(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
+    ) -> tuple[torch.Tensor, InputBatch]:
+        """Return the global batch and, when already global, skip re-gathering.
+
+        On vLLM main the Ascend runner restores the target hidden states to the
+        global PCP layout before sampling (draft_hidden_states is captured
+        before the upstream restore), so a second all-gather here would reorder
+        them. Detect the already-restored layout by its padded length.
+        """
+        assert self._global_batch is not None
+        if hidden_states.shape[0] == self._global_batch.num_tokens_after_padding:
```
weiguihua2: 用这个长度判断是不是不太准，因为piecewise下可能存在全长是8，切完每张卡上是4，然后pad到8，这时候就不会restore了

vllm_ascend/models/deepseek_v41/vl_model.py
```diff
@@ -197,9 +197,7 @@ def embed_input_ids(
             _merge_multimodal_embeddings,
         )
 
-        # The leading alignment row is not an image-feature position. It uses
-        # the checkpoint's ordinary image-token embedding instead.
-        embedding_ids = input_ids.masked_fill(input_ids == IMAGE_PAD_ID, IMAGE_SENTINEL_BASE_ID)
+        embedding_ids = input_ids
```
wangxiyuan, line 200: nit: this `embedding_ids` is useless. Just use `input_ids` is enough

wangxiyuan (dismissed):

Angazenn (approved):

vllm_ascend/attention/mla_v1.py
linfeng-yuan: @drslark As @zhangxinyuehfad 's reply, vLLM here capture a non TND layout and we force treat it as TND here. But here the attn_metadata.attn_state does not belong to expected one and here is a work around fix. Is it OK? Please make sure draft model always execute this branch.

vllm_ascend/worker/v2/pcp_manager.py
zhangxinyuehfad: @weiguihua2 Good catch — the length check was unreliable and is gone. The runner now sets an explicit `_sampling_hidden_restored` marker after it actually restores, `restore_for_sampling` consumes it, and it is reset every `sample_tokens` step and on batch re-partition, so the piecewise "same length, different layout" case no longer matters.

vllm_ascend/worker/v2/spec_decode/dspark/speculator.py
```diff
@@ -203,23 +245,67 @@ def _build_draft_attn_metadata(self, *, num_reqs_padded, **kwargs):
                 parallel_config=self.attn_vllm_config.parallel_config,
             ),
         ):
-            attn_metadata = super()._build_draft_attn_metadata(num_reqs_padded=num_reqs_padded, **kwargs)
+            if query_start_loc_np is not None:
```
drslark, line 248: What is this branch?
Do we only need `_build_uniform_attn_metadata`?

vllm_ascend/worker/v2/pcp_manager.py
```diff
@@ -262,6 +276,9 @@ def partition_batch(
             local_batch = super().partition_batch(
                 global_batch,
                 padded_num_tokens=padded_num_tokens,
+                # vLLM main (#53867) pads request metadata for FULL graphs via
+                # padded_num_reqs.
+                padded_num_reqs=padded_num_reqs,
```
li1how: 这个入参删了吧，由下面来做padd

zhangxinyuehfad: done

Tflowers-0129 (approved):

linfeng-yuan (approved):

drslark (approved):

PR #16530, opened 2026-09-14
[Test] Set DeepSeek V4 prefill token budget to 8192

sugm0521:

This PR changes `--max-num-batched-tokens` from `4096` to `8192` for both prefill nodes in the DeepSeek-V4-Pro W4A8 prefix-cache PD Nightly-A3 case.

The decode nodes remain at `120`, and all other service and benchmark parameters are unchanged. This provides the candidate configuration for a controlled comparison against the current `4096` baseline when investigating the V1 runner output-token throughput regression.

No. This only changes a Nightly-A3 test configuration.

- Parsed the YAML configuration successfully.
- Verified the four node token budgets are `8192`, `8192`, `120`, and `120` for nodes 0 through 3.
- Verified the diff changes only the two prefill values and passes `git diff --check`.
- The target four-node NPU Nightly test has not been run locally and should be triggered through `/nightly` for performance evidence.


CuixyHW:
/nightly DeepSeek-V4-Pro-w4a8-prefix-cache-PD
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/34840359738)

PR #16541, opened 2026-09-14
[Perf] Include torch_npu version in StartPlan fingerprint

yanganYNU:
## Background

IR20260814000568 cold start / reschedule warmup. vLLM StartPlan skips memory-profiling math on a fingerprint hit. The upstream key includes `torch.version.cuda`, which is empty on Ascend, so `torch_npu` / CANN changes are invisible. A stale KV-cache budget can then be applied after a runtime bump.

## Change

Wrap `compute_plan_fingerprint` in `vllm_ascend.worker.startup_plan` to mix `torch_npu.__version__` into the 16-char digest. `maybe_apply_startup_plan` / `maybe_save_startup_plan` stay upstream; NPUWorker imports them through the wrapper so the patch is installed.

Existing StartPlan files miss once and are rewritten. This does not skip `profile_run()`; comm-buffer preallocation still happens.

## Test plan

- Import `vllm_ascend.worker.worker` and confirm `vllm.v1.worker.startup_plan.compute_plan_fingerprint` is the wrapped function.
- With `VLLM_ENABLE_STARTUP_PLAN=1`, first boot writes a plan; second boot with the same torch_npu hits; bumping torch_npu (or mocking the version) misses and re-profiles.


yanganYNU:
Scope changed: only cold-start (engine process init). Warmup (memory profiling / graph reuse) is out of scope. Closing this PR.

PR #16548, opened 2026-09-14
[Performance] Speed up startup for DeepSeek V4

### LINKED ISSUE #16552: [Feature]: 大EP启动时长优化
PD 分离大 EP 下，DeepSeek V4 Flash / Pro、GLM 等模型，以及 2 机 / 4 机 / 8 机等场景，首次拉起都偏长。一个 Prefill 或 Decode pod 会同时拉多路 `vllm serve`，从进程起来到可以服务经常要数分钟。

偏长、也较有优化空间的阶段：

- 构图：量化配置扫描、RMSNorm、MTP 再构图
- 权重加载：多卡并行读盘；加载线程不合适时 I/O 会变成墙钟
- 进程拉起：同一 pod 多路重复 import / 初始化
- 首次 warmup：RMS / Triton JIT、NZ 格式转换

MindIE-Motor 改进程拉起和加载默认，vLLM-Ascend 改构图和 warmup。两个仓对应同一问题。

Related:
- MindIE-Motor issue https://gitcode.com/Ascend/MindIE-Motor/issues/621
- MindIE-Motor https://gitcode.com/Ascend/MindIE-Motor/merge_requests/993
- vLLM-Ascend main https://github.com/vllm-project/vllm-ascend/pull/16548
- vLLM-Ascend 0.26 https://github.com/vllm-project/vllm-ascend/pull/16860
- 拉起时间分析 https://gitcode.com/mindie_yangan/Scripts-LLM/blob/main/lep_speed/startup_timing.md
- 冷启动优化方案 https://gitcode.com/mindie_yangan/Scripts-LLM/blob/main/lep_speed/coldstart_plan.md

yanganYNU:
ISSUE: https://github.com/vllm-project/vllm-ascend/issues/16552
PR: https://github.com/vllm-project/vllm-ascend/pull/16548
Branch: `coldstart` (fork `yanganYNU/vllm-ascend`)
Base: `main` (rebuilt on latest `upstream/main`)

Related:
- 0.27 backport: https://github.com/vllm-project/vllm-ascend/pull/17130
- 拉起时间分析 https://gitcode.com/mindie_yangan/Scripts-LLM/blob/main/lep_speed/startup_timing.md
- 冷启动优化方案 https://gitcode.com/mindie_yangan/Scripts-LLM/blob/main/lep_speed/coldstart_plan.md

No Motor-side change is needed for this PR. The two `MOTOR_*` switches below
are read by vllm-ascend, not by Motor.

## **1. 背景**

Fixes #16552

Large-EP PD (Flash / Pro / GLM and similar). Construct used to scan
~820k quant keys per RMSNorm; warmup JIT'd 16 RMS `BLOCK_M` values;
first NZ format-cast and Triton compile sat after I/O.

## **2. 设计**

Same Flash-best optimizations as the 0.26 PR, ported to the main package
layout. `is_layer_skipped_ascend` is already O(1) on main (#16051), so that
index is not in this PR. Main already maps a missing
`quant_description[prefix + ".weight"]` to unquantized, so the Pro MTP
`mtp.0.head.weight` KeyError on 0.26 does not apply here.

Dropped the old CPU-then-NPU expert path (it hit `Expected NPU tensor` on
Prefill dummy_run). Kept main's current hash-router `tid2eid` table.

Opt-in env, both default off:
- `MOTOR_COLDSTART_NZ_WARM=thread`
- `MOTOR_WARMUP_EARLY_KERNEL=thread`

No loader default is changed. The netloader disk fallback used to drop the
whole `model_loader_extra_config`, which silently disabled the user's
multithread load; it now carries the keys `DefaultModelLoader` accepts and
drops only the netloader-only ones. Thread-count defaults stay with vLLM
(review feedback from @nu11ptr7).

### 3.1 修改清单

业务代码:

- `vllm_ascend/ops/layernorm.py`
- `vllm_ascend/ops/triton/rms_norm.py`
- `vllm_ascend/model_executor/warmup/rms_triton_warmup.py`
- `vllm_ascend/model_executor/warmup/rejection_sampler_triton_warmup.py`
- `vllm_ascend/model_executor/warmup/early_kernel_warmup.py`
- `vllm_ascend/quantization/methods/w8a8/w8a8_dynamic.py`
- `vllm_ascend/models/deepseek_v4/model.py`
- `vllm_ascend/models/deepseek_v4/mtp.py`
- `vllm_ascend/eplb/core/eplb_utils.py`
- `vllm_ascend/model_loader/netloader/utils.py`
- `vllm_ascend/model_loader/netloader/netloader.py`

DT:

- `tests/ut/model_executor/warmup/test_early_kernel_warmup.py` (new)
- `tests/ut/ops/test_rms_block_sizing.py` (new)
- `tests/ut/model_executor/warmup/test_rms_triton_warmup.py`
- `tests/ut/model_executor/test_deepseek_v4.py`
- `tests/ut/model_loader/netloader/test_netloader_utils.py`
- `tests/ut/eplb/core/test_eplb_utils.py`
- `tests/ut/ops/test_layernorm.py`
- `tests/ut/quantization/methods/w8a8/test_w8a8_dynamic.py`

### 3.2 DT

| 用例 | 覆盖 |
| --- | --- |
| `test_early_kernel_warmup.py` | env gating, thread is daemon and started once, `get_current_vllm_config() is None` does not latch the state, failures never propagate, `join` no-ops without a live thread |
| `test_rms_block_sizing.py` | `_rms_block_m` floors to `{1,2,4,8,16}`, is capped at 16, never returns 0 |
| `test_rms_triton_warmup.py` | warmup compiles exactly the five power-of-two keys, `assume_used=True` skips the `attn_groups` probe, the early thread is joined first |
| `test_deepseek_v4.py` | both construct hooks are opt-in and never break construct when they fail |
| `test_netloader_utils.py` | disk fallback keeps `DefaultModelLoader` keys, drops `SOURCE`/`LISTEN_PORT`, adds no defaults, preserves an explicit opt-out |
| `test_eplb_utils.py` | linear expert map is built once per `(ep_size, ep_rank, n_experts)`, each layer gets its own copy, a `None` map is passed through |
| `test_layernorm.py` | `quant_description` is scanned once, not once per RMSNorm |
| `test_w8a8_dynamic.py` | NZ warmup runs once, swallows failures, starts at most one daemon thread |

### 3.3 pre-commit

No config change. `ruff check` / `ruff format` / `codespell` / `markdownlint`
/ `gitleaks` clean on the branch.

## **4. 自验证**

Ran on the 0.27.1rc image
`mindie-motor:3.2.0b3-vllm_ascendnightly-releases-0.27.1rc-a2-openeuler24.03-py3.12-aarch64`,
image Motor 3.2.0b3, no wheel. Overlay is the 0.27 layout of this change set
plus `modelslim_config.py` (that file is not in this PR: main no longer has
`is_layer_skipped_ascend`). Same Motor both sides. Triton cache hot both
sides. HTTP 200.

A = first `Available plugins` to the last `Model runner load_model total
time`; B = `init engine (profile, create kv cache, warmup model)`. A+B is
taken on the slowest decode pod.

Flash W8A8, A2 4-node, one P and one D instance of two pods each, dp16
(`17-17-06` vs `18-29-36`):

| | baseline | this change set |
| --- | --- | --- |
| slowest decode A+B | 318 s | **206 s** |
| `load_model total` | 151~161 s | **58~68 s** |
| construct | 77 s | **3~4 s** |

Pro W4A8, A2 8-node, one P and one D instance of four pods each
(`18-58-22` vs `19-26-10`):

| | baseline | this change set |
| --- | --- | --- |
| slowest decode A+B | 811 s | **461 s** |
| `load_model total` | 463~529 s | **135~161 s** |
| construct | 308~318 s | **4~5 s** |

Weight I/O is unchanged (Pro 83~98 s both sides). Warmup did not move: both
runs reuse a hot Triton cache, the only B-stage item in this change set is
the Triton kernel warmup and it was already 0.33~0.47 s on the baseline, and
the bulk of B is `torch.compile` plus graph capture (Pro 43 s + 59 s, equal on
both sides), which this PR does not touch. Construct on 0.27.1rc still needs
the `modelslim_config.py` index; that lands in the 0.27 backport, not here.


vllm_ascend/model_loader/netloader/utils.py
```diff
@@ -16,10 +16,71 @@
 
 import os
 import socket
+from typing import Any
 
 import regex as re
 from vllm.logger import logger
 
+# Match DeepSeek V4 Pro/Flash tutorials. DefaultModelLoader.DEFAULT_NUM_THREADS is 8.
+DEFAULT_MULTITHREAD_LOAD_THREADS = 128
```
nu11ptr7, line 25: Default-enabling multithread load with `num_threads=128` (upstream defaults to 8) applies to every local safetensors load, not just the DeepSeek V4 tutorials quoted here. With many ranks starting at once that multiplies disk/CPU pressure and changes behaviour for all models - worth a more conservative default or making the 128 explicit per model/config.

yanganYNU:
Rebuilt on latest `main` (force-pushed `coldstart`). Same Flash-best optimizations as the 0.26 PR https://github.com/vllm-project/vllm-ascend/pull/16860. Dropped the CPU-then-NPU expert path.

vllm_ascend/model_loader/netloader/utils.py
yanganYNU: Dropped the 128 default entirely - thanks, you were right that it
changed behaviour for every local safetensors load.

`apply_default_multithread_weight_load` and its `platform.py` call site are
gone, so this PR no longer sets `enable_multithread_load` or `num_threads` for
anyone. The only thing left in this file is the disk-fallback fix: the netloader
fallback reset `model_loader_extra_config` to `{}`, which silently dropped a
multithread-load setting the user had configured. It now keeps the keys
`DefaultModelLoader` accepts and drops only the netloader-only ones
(`SOURCE`/`LISTEN_PORT`). Defaults stay with vLLM.

Covered by `tests/ut/model_loader/netloader/test_netloader_utils.py`.

yanganYNU:
Pushed an update addressing the review and CI:

- **@nu11ptr7's point on `num_threads=128`**: dropped. The PR no longer sets
  any loader default. `apply_default_multithread_weight_load` is gone and the
  `platform.py` hook with it, so no behaviour changes for models that do not
  configure it. What remains in `netloader/utils.py` is only the disk-fallback
  bug fix: the fallback used to reset `model_loader_extra_config` to `{}`,
  which silently dropped a user-configured multithread load; it now keeps the
  keys `DefaultModelLoader` accepts and drops only `SOURCE`/`LISTEN_PORT`.
  Thread-count defaults stay with vLLM.
- **`early_kernel_warmup.start_early_kernel_warmup`**: `_STATE["started"]` was
  set before the `get_current_vllm_config() is None` check, so a call made
  before the config context would permanently disable the warmup. Moved the
  flag after the check, with a regression test.
- **`platform.py` removed from the PR.** Its only hunk keyed off
  `MOTOR_VLLM_FORK_PARENT`, which no upstream code sets. The corresponding
  Motor-side change has been withdrawn, so the branch is now pure vllm-ascend.
- **pre-commit**: fixed (`ruff` wanted an import-group blank line in
  `early_kernel_warmup.py`). `ruff check`, `ruff format`, `codespell` and the
  repo's local hooks are clean on the full tree.
- **DCO**: commit is signed off and squashed to one.
- **UT added**: 8 files, covering the early-warmup module, `BLOCK_M`
  flooring, the EPLB map cache, the `norm.bias` cache, the NZ warm thread and
  the loader fallback. `tests/ut/model_executor/warmup/test_rms_triton_warmup.py`
  also had a stale assertion (16 compiles) that this change set makes wrong; it
  now asserts the five power-of-two keys.

Self-verification numbers in the description are unchanged: they were measured
with these same optimizations and without any loader-default change.

yanganYNU:
Rebased onto latest `main` and pushed. Status of this PR now:

| check | result |
| --- | --- |
| DCO | success |
| pre-commit (lint / mypy / gitleaks) | success |
| cpu-ut | success - 5713 passed |
| ci-gate | blocked on a label only |

`ci-gate` says: *"This PR touches source code, so precision/full tests are
required before merge. A maintainer adds exactly one label - `ready-precise`
/ `ready-all` / `main2main`."* Could a maintainer add `ready-precise`? I do not
have label permissions.

Two notes on this push:

- `tests/ut/test_ascend_forward_context.py` - the `SimpleNamespace` model stub
  needed `_needs_mtp_hidden_states=True` because the MTP hidden buffer is now
  allocated lazily (it used to be allocated in `__init__` even on ranks that
  never use it).
- `vllm_ascend/xlite/xlite.py` - one line, unrelated to this feature.
  `main` currently fails mypy with
  `"XModelConfig" has no attribute "index_full_mask"` (introduced by #16669,
  still unfixed), which fails `pre-commit` for every PR rebased on it. Used
  the same `getattr` form already present a few lines above in that file.
  Happy to drop it if you prefer a separate PR.

vllm_ascend/models/deepseek_v4/mtp.py
```diff
@@ -194,10 +195,17 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
                 )
             }
         )
-        self.embed_tokens = VocabParallelEmbedding(
-            config.vocab_size,
-            config.hidden_size,
-        )
+        # PP=1 MTP rebinds embed_tokens from the target after load. Skip a
+        # second vocab table during draft construct.
+        use_compress = hasattr(vllm_config.model_config.hf_config, "compress_ratios")
```
Angazenn: deepseek v4 should have `use_compress == True`. You don't need to add duplicated conditions.

drslark: Others are okay.

drslark (approved):
I reviewed mtp.
LGTM.

vllm_ascend/quantization/methods/w8a8/w8a8_dynamic.py
```diff
@@ -45,6 +45,51 @@
 )
 from ..registry import register_scheme
 
+_NZ_WARMED = False
+_NZ_THREAD_STARTED = False
+
+
+def warm_nz_format_cast(tag: str) -> None:
```
kunpengW-code: This logic is actually unrelated to `w8a8_dynamic.py`. Could we move it into utils instead?

vllm_ascend/models/deepseek_v4/mtp.py
```diff
@@ -194,10 +195,17 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
...
+        self._share_target_embed = get_pp_group().world_size == 1 and not use_compress
+        if self._share_target_embed:
```
drslark: What about `get_pp_group().world_size > 1`?
Is this a scenario considered by this PR?

vllm_ascend/models/deepseek_v4/mtp.py
yanganYNU: Leaving this as-is. `_maybe_share_embeddings` already shares the MTP embed only when PP=1 and the config has no `compress_ratios`. This guard copies that condition so construct does not drop a table the proposer will not rebind.

DeepSeek V4 configs have `compress_ratios`, so `use_compress` is true and this branch does not run: the draft still builds its own `VocabParallelEmbedding`. Removing the condition would skip that table on DeepSeek V4 and break weight load. The measured path is unchanged.

vllm_ascend/quantization/methods/w8a8/w8a8_dynamic.py
yanganYNU: Leaving this in `w8a8_dynamic.py`. The two helpers only pay the first `npu_format_cast` lazy init; they do not change the W8A8 scheme. Moving them to utils would be a relocation with the same behavior, so the diff stays where the cast it warms already lives.

vllm_ascend/models/deepseek_v4/mtp.py
yanganYNU: Yes, PP>1 is covered and needs no change. `world_size > 1` keeps `VocabParallelEmbedding`. `_maybe_share_embeddings` only shares when `world_size == 1`, because with PP>1 the draft embedding and the target embedding are not on the same device. The scenes for this PR are PP=1.

vllm_ascend/model_executor/warmup/early_kernel_warmup.py
```diff
@@ -0,0 +1,134 @@
...
+def join_early_kernel_warmup(where: str) -> None:
+    del where
+    thread = _STATE.get("thread")
+    if thread is None:
+        return
+    if thread is threading.current_thread():
+        return
+    if not thread.is_alive():
+        return
+    thread.join(timeout=_JOIN_TIMEOUT_S)
```
yiz-liu, line 137: The background warmup can outlast weight loading and overlap with memory profiling or KV cache allocation. Since warmup allocates NPU tensors and launches kernels, this overlap can distort the memory estimate used to size the KV cache and increase peak memory during allocation.

vllm_ascend/model_executor/warmup/early_kernel_warmup.py
```diff
@@ -0,0 +1,134 @@
+# Copyright (c) 2025 Huawei Technologies Co., Ltd. All Rights Reserved.
+"""Compile Triton warmup kernels while weight I/O is already on the critical path.
+
+Rejection-sampler / penalty / RMS JIT is host work. Starting it after model
+construct overlaps the compile with safetensors load. Regular kernel_warmup
+then joins this thread and hits the Triton cache.
+
+Enable with MOTOR_WARMUP_EARLY_KERNEL=thread. Imports that the thread needs
+are done on the main thread first so a concurrent import cannot walk
+sys.modules mid-registration.
```
yiz-liu: We should not introduce an environment variable for this feature. Maybe expose this through additional_config, defaulting to false.

vllm_ascend/models/deepseek_v4/mtp.py
yanganYNU: Done in 4621d926c. You were right: the startup logs never print "Sharing target model embedding weights" on DeepSeek V4, so this branch never ran. `mtp.py` is back to upstream.

vllm_ascend/models/deepseek_v4/mtp.py
yanganYNU: `mtp.py` is back to upstream in 4621d926c, so PP>1 and PP=1 both take the existing proposer path unchanged.

vllm_ascend/model_executor/warmup/early_kernel_warmup.py
yanganYNU: Done in 4621d926c. `NPUWorker.load_model` now joins both prewarm threads (this one and the NZ format-cast warm thread) before it returns, so they finish before memory profiling and KV cache allocation.

vllm_ascend/model_executor/warmup/early_kernel_warmup.py
yanganYNU: Done in 4621d926c. Both env vars are gone. The switches are now `additional_config` fields, default false: `enable_early_kernel_warmup` and `enable_early_nz_warmup`.

vllm_ascend/quantization/methods/w8a8/w8a8_dynamic.py
yanganYNU: Changed my mind after a closer look: moved in 2526b20cf. This NZ warmup is a startup acceleration and has nothing to do with the W8A8 scheme (float and W4A8 models cast to NZ too), so it now lives in `vllm_ascend/model_executor/warmup/nz_warmup.py` next to the other startup warmups. `w8a8_dynamic.py` and its tests are back to upstream.

kunpengW-code: This logic is independent of quantization schemes. It would be better to extract it as a standalone `warm_up` utility method in `vllm_ascend/utils.py`.

vllm_ascend/ascend_config.py
```diff
@@ -508,6 +501,10 @@ class AscendConfig:
     msmonitor_use_daemon: bool = False
     enable_transpose_kv_cache_by_block: bool = True
     weight_nz_mode: int = 1
+    # Startup prewarm threads that overlap weight I/O. Both are joined before
+    # memory profiling so they never touch the KV cache budget.
+    enable_early_kernel_warmup: bool = False
```
kunpengW-code: It would be better to maintain this in a config, similar to `ascend_compilation_config`. Also, please update `docs/source/user_guide/configuration/additional_config.md`.

vllm_ascend/quantization/methods/w8a8/w8a8_dynamic.py
yanganYNU: Done in 9ce81b9b5. `warm_nz_format_cast` / `start_nz_warm_thread` / `join_nz_warm_thread` are now standalone utilities in `vllm_ascend/utils.py`, next to `maybe_trans_nz`.

vllm_ascend/ascend_config.py
yanganYNU: Done in 9ce81b9b5. Both switches moved into a new `ascend_warmup_config` sub-config (`AscendWarmupConfig`), and `additional_config.md` documents it.

vllm_ascend/quantization/methods/w8a8/w8a8_dynamic.py
kunpengW-code: Sorry for the delayed reply. Placing it under `vllm_ascend/model_executor/warmup/nz_warmup.py`, alongside other startup warmup routines, would be a great fit.

yanganYNU: Done in 86def8ed2. Moved back to `vllm_ascend/model_executor/warmup/nz_warmup.py`; `vllm_ascend/utils.py` is untouched.

vllm_ascend/models/deepseek_v4/model.py
```diff
@@ -107,6 +107,32 @@
...
+        from vllm_ascend.model_executor.warmup.nz_warmup import start_nz_warm_thread
+
+        start_nz_warm_thread("thread")
+    except Exception:
+        from vllm.logger import logger
+
+        logger.warning("NZ format-cast warmup thread skipped", exc_info=True)
+
+
+def _maybe_start_early_kernel_warmup() -> None:
```
weijinqian0: need remove this.

vllm_ascend/models/deepseek_v4/model.py
```diff
@@ -838,6 +864,8 @@ def __init__(self, *, vllm_config: VllmConfig, prefix: str = ""):
             lambda prefix: DeepseekV4DecoderLayer(vllm_config, prefix, topk_indices_buffer=topk_indices_buffer),
             prefix=f"{prefix}.layers",
         )
+        _maybe_start_nz_warm_thread()
+        _maybe_start_early_kernel_warmup()
```
weijinqian0, line 841: need change other position.

vllm_ascend/models/deepseek_v4/model.py
yanganYNU: Done in f1e12d6ae.

vllm_ascend/models/deepseek_v4/model.py
yanganYNU: Done in f1e12d6ae.

weijinqian0 (dismissed):

weijinqian0 (approved):

PR #16549, opened 2026-09-14
[Misc][BugFix] Fix speculative config MTP model types literal definition

KlyzhenkoVadim:
This pull request attempts to modify how `speculative_config.MTPModelTypes` is dynamically updated by wrapping unpacked arguments in parentheses. However, doing so in Python 3.10 creates a single tuple argument for `Literal`, meaning it only accepts the tuple itself rather than its individual elements. To resolve this, the review suggests using `Literal.__getitem__` to properly construct the literal type across Python versions.

No.

No testing details were provided in the patch.


slippersss:
LGTM. I have tried it manually and erros indeed exist. Since we claim to support python3.10, we should fix it.

```
Python 3.10.16 (main, Sep 16 2026, 11:25:47) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from typing import Literal, get_args
>>> before = Literal["a", "b", "c"]
>>> after = Literal[*get_args(before), "more"]
  File "<stdin>", line 1
    after = Literal[*get_args(before), "more"]
                    ^
SyntaxError: invalid syntax
>>> after = Literal[(*get_args(before), "more")]
>>> after
typing.Literal['a', 'b', 'c', 'more']
```

slippersss (approved):

kunpengW-code:
pre-commit check failed. Please rebase onto the latest upstream code.

kunpengW-code (approved):

PR #16550, opened 2026-09-14
[BugFix][MoE] Keep shared and final maybe_all_reduce decisions runtime-aware

drslark:

This PR reverts #16495 and replaces the complete MoE forward wrapper with runtime-aware reduction at three points: routed output before a transform or shared DP addition, shared TP output before addition, and final output after addition.

```mermaid
flowchart TD
    R["Routed output"] --> R1["1. Maybe reduce routed early<br/>Required by transform or shared DP?<br/>If yes, reduce unless already reduced"]
    R1 --> T["Routed transform, if any"]
    S["Shared output"] --> S1["2. Maybe reduce shared<br/>Shared TP + routed already reduced?<br/>If yes, reduce shared"]
    T --> A["Add shared and routed, if shared exists"]
    S1 --> A
    A --> F["3. Maybe reduce final output<br/>No SP + no early reduction<br/>+ routed not already reduced?<br/>If yes, reduce combined output"]
    F --> O["Return"]
```

#### Reduction conditions

| Step | When to check | When to all-reduce |
|---|---|---|
| **1. Routed early** | Shared DP only, **or** a routed transform exists with SP disabled and (TP > 1 or EP > 1) | Routed output is not already reduced |
| **2. Shared** | Shared output exists and uses `TENSOR_PARALLEL` | Routed output was reduced early or already reduced by communication |
| **3. Final** | SP is disabled and early routed reduction is not required | Routed output is not already reduced; reduce after adding shared and routed outputs |

#### Runtime check

| Communication path | Routed already reduced? |
|---|---|
| MC2 / ALLTOALL / FUSED_MC2 | Yes |
| ALLGATHER + SP | Yes |
| ALLGATHER without SP | No |

#### Static vs. runtime decisions

| Location | Decision | Evaluated at |
|---|---|---|
| The three `_maybe_reduce_*` runner methods | Layer configuration: shared layout, SP, transform presence, TP / EP sizes | FX tracing; fixed for the layer |
| Custom op real implementations | Whether routed output is already reduced, using the current communication type | Runtime |
| `fake_impl` | Output metadata only; no communication decision or collective | FX tracing |

The runner methods select reduction points using static conditions. Communication-dependent decisions stay inside the custom ops and are evaluated by their real implementations at runtime. Both ops receive `layer_name` and read the layer's MoE configuration internally.

Fixes incorrect MoE output when tracing and replay use different communication types. No configuration changes are required.

Validated the runtime reduction patch from `eaca0c24d` on the native 20260914 Ascend image. Only the reduction changes in `fused_moe.py` and `register_custom_ops.py` were applied; the rest of the image was retained. This does not validate the entire PR branch against current main.

| Test | Result |
|---|---|
| AIS Bench GSM8K, first 20 test samples, concurrency **1** | **20/20 correct, 100.00% accuracy** |
| AIS Bench GSM8K, the same 20 samples, concurrency **4** | **18/20 correct, 90.00% accuracy** |
| Review of all 20 complete responses at concurrency 4 | No looping repetition, garbled text or visibly truncated answers observed |
| Three original arithmetic reproducer prompts | **3/3 correct**: 391, 16, 150 |
| Graph-replay regression cases extracted from the PR | **10 passed**, covering ALLGATHER, MC2, ALLTOALL and FUSED_MC2 |

The two incorrect responses at concurrency 4 contain complete reasoning and final answers:

| Sample (1-based) | Predicted | Reference | Observed reasoning error |
|---|---|---|---|
| 3 | 25000 | 70000 | Applies the house-value increase percentage to the repair cost |
| 13 | 12 | 13 | Chooses the break-even year instead of the first profitable year |

The concurrency-dependent accuracy difference has not been isolated to a root cause. The absence of looping or garbled output does not establish batch-invariant accuracy.

Graph mode was enabled: `FULL_DECODE_ONLY`.

Only AIS Bench concurrency changed between the two runs; the dataset, generation settings and running service were retained. The full PR test suite and full lint workflow have not been rerun on the latest revision.


weijinqian0 (dismissed):

wxsIcey (approved):

drslark:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

weijinqian0 (approved):

Issue #6026, opened 2026-01-20
[RFC]: Tracking Remaining Work for Encode–Prefill–Decode Disaggregation


Shirley125:
### Motivation.

Motivation for Encode–Prefill–Decode Disaggregation refers to https://github.com/vllm-project/vllm/pull/25233
This RFC focuses on tracking incomplete or unresolved items in the current solution.

### Proposed Change.

P0:
- [ ]  Add EPD end-to-end (E2E) test cases
- [ ]  Provide EPD proxy example
- [ ] Optimize encode cache lookup timing to reduce TPOT：https://github.com/vllm-project/vllm/pull/32585

P1:
- [ ] Optimize encoder instance weight loading
- [ ] Enable cross-node EPD support by using Mooncake as the storage backend

zengchuang-hw:
referring to https://github.com/vllm-project/vllm/pull/30242, models except qwen-vl should make some modifications to achieve encoder skip loading language model

PR #16555, opened 2026-09-15
[Feature][EC Connector] Add Mooncake EC transfer support for ModelRunner V1

lirx-pd:
Add an Ascend adapter for the upstream `ECMooncakeConnector` so V1 workers can transfer multimodal encoder outputs between encoder and prefill/decode instances.
The adapter reuses the upstream scheduler, reservation state machine, and worker lifecycle, with NPU-specific memory and synchronization handling.

Related to the EPD roadmap in issue #6026 which tracks Mooncake-based cross-node encoder-cache sharing among other work. This PR adapts the `ECMooncakeConnector` TransferEngine data plane for Ascend.

- Register the Ascend implementation under the existing `ECMooncakeConnector`
  name through the Ascend plugin entry point.
- Reuse `MooncakeTransfer` with the process-local Ascend TransferEngine. Allocate
  registered staging and receive buffers with 2 MiB alignment, and reject direct
  registration of arbitrary source tensor views when staging is unavailable.
- Use NPU events to track source readiness and delay receive-buffer reuse until
  device work completes.
- Start EC worker services after model loading and shut them down with the worker.
- Return EC completion information from V1 encoder-only execution and steps that
  skip model forward. Previously, these return paths could omit the information
  the scheduler needs to advance asynchronous transfers.

The change also allocates the MoE load counter directly on NPU, removing the intermediate CPU allocation and device copy. V2 runner support is outside this PR.

The V1 runner changes belong in Ascend's existing `execute_model()` override:
that method owns the encoder-only and no-forward return paths where EC output must be collected. Patching the upstream runner would not update these overridden paths; replacing the Ascend method through a runtime patch would add indirection to code already maintained here. Both no-forward exits share one helper, which preserves KV output and invokes the existing EC lifecycle when metadata is present. The connector continues to own transfer scheduling and memory management. Worker shutdown follows the upstream KV-then-EC order.

Yes. On Ascend, `ec_connector="ECMooncakeConnector"` selects the NPU adapter. The adapter defaults to the `ascend` transport protocol and maps the default CUDA buffer setting to NPU. Explicit non-Ascend transport protocols are rejected.

Transfers use the configured `ec_buffer_size`. If the producer cannot stage a batch in its registered pool, it reports a transfer failure instead of falling back to registering the original tensor addresses.

Added `tests/ut/distributed/ec_transfer/test_mooncake_adapter.py` and ran it in the existing Ascend container after synchronizing the changes to the editable runtime
checkout:

```bash
python -m pytest -q tests/ut/distributed/ec_transfer/test_mooncake_adapter.py
```

Result: **12 passed**. Tests cover connector registration, protocol validation, shared-engine reuse, write failures, failed-unregistration ownership and cleanup, direct-source rejection, Producer/Consumer worker initialization and dispatcher shutdown, and V1 no-forward EC output handling. TransferEngine calls are mocked.

The following checks passed locally:

```bash
bash format.sh
bash format.sh ci
git diff --check
```

The format checks used ARM64 Gitleaks and ShellCheck, a writable Ruff cache, and initialized pre-commit environments with compatible Go and Node versions. Python syntax checks also passed. These are local checks, not hosted CI results.

#### Real-NPU TP2 validation

The TP2 consumer regression passed on three Ascend 910B1 NPUs, using a real Mooncake sender and two HCCL consumer ranks. Each rank handled six reservation callbacks on non-main control threads with the correct receive-pool device bound, including `npu:1` on rank 1. Tensor equality, cancellation, buffer reuse, pool reclamation, and cleanup checks passed.

An EPD correctness test compared a TP2 monolithic baseline with one encoder plus a TP2 consumer on the same PD devices: Qwen3.5-9B, MRV1, BF16, eager, CANN 9.1.0, and torch/torch_npu 2.10.0. Eight serial requests contained one or three 224×224 images (16 images total), with greedy decoding and 32 output tokens each. All requests completed without EC transfer errors.

- **Transfer correctness:** all 16 source embeddings matched both TP ranks'
  received tensors bit-for-bit in two diagnostic replays, including one with
  reads deferred until after sending and consumer prefill. Staging tensors also
  matched in the full-capture replay; no transfer corruption was observed.
- **Baseline comparison:** the monolithic baseline batches two images together
  on one vision rank, while EPD encodes images individually. With the baseline
  changed only to encode images individually, all 16 image embeddings, merged
  inputs on both ranks, all 80 captured logits tensors (five steps per request
  per rank), and **8/8 complete output token sequences matched bit-for-bit**.

The original batched-baseline comparison matched 6/8 outputs; both diagnostic replays matched 7/8. The reproducible mismatch was traced to batch-dependent visual numerical differences affecting greedy token selection, independently of Mooncake. 

#### End-to-end benchmarks

Qwen3.5-35B-A3B (W8A8_DYNAMIC, BF16 activations), 4 × Ascend 910B (64 GiB), CANN 9.1.0, torch/torch_npu 2.10.0, vLLM `a97dacb7106e`. Each setup uses four NPUs: DP4EP4 or 1 Encoder + 3 PD for both EPD connectors.
Client concurrency is 128; output is fixed at 64 tokens. Eager execution, no MTP, prefix caching, or profiling. DP4 and Encoder use OMP8; PD uses OMP1.

| Workload | Images per request | Input tokens | Warmup / measured requests per setup |
|---|---|---:|---:|
| Many small images | 16 × 224×224 | 4096 | 128 / 1024 |
| 3 large images | 3 × 1088×1120 | 4096 | 256 / 2048 |
| 6 large images | 6 × 1088×1120 | 7666 | 256 / 2048 |

Identical inputs are used across setups within each workload. The large-image workloads keep 526 non-visual input tokens; six images double the visual tokens. Latency columns below are means. Parentheses show changes relative to Example:
higher throughput and lower latency are better.

| Workload | Setup | Throughput (req/s) | TTFT (s) | TPOT (ms) | E2E (s) |
|---|---|---:|---:|---:|---:|
| Many small images | DP4EP4 | 3.483 | 9.383 | 432.3 | 36.619 |
| Many small images | Example EPD | 5.897 | 3.938 | 267.6 | 20.800 |
| Many small images | Mooncake EPD | 6.035 (+2.4%) | 3.472 (-11.8%) | 269.1 (+0.5%) | 20.423 (-1.8%) |
| 3 large images | DP4EP4 | 2.713 | 6.214 | 641.7 | 46.642 |
| 3 large images | Example EPD | 4.203 | 14.236 | 245.6 | 29.708 |
| 3 large images | Mooncake EPD | 6.010 (+43.0%) | 3.996 (-71.9%) | 264.0 (+7.5%) | 20.626 (-30.6%) |
| 6 large images | DP4EP4 | 1.958 | 15.188 | 794.1 | 65.213 |
| 6 large images | Example EPD | 2.132 | 43.529 | 234.7 | 58.317 |
| 6 large images | Mooncake EPD | 3.885 (+82.2%) | 14.113 (-67.6%) | 288.1 (+22.7%) | 32.261 (-44.7%) |

All 15,360 measured requests succeeded with the expected input/output token counts. These are single-run, end-to-end comparisons using each connector's corresponding proxy; Example uses a local shared filesystem. They do not isolate transport cost. Mooncake's largest throughput gain over Example is on six large images (+82.2%), with 22.7% higher TPOT; its small-image throughput gain needs repeated-run validation.


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -680,6 +680,21 @@ def reset_encoder_cache(self) -> None:
...
+            else EMPTY_MODEL_RUNNER_OUTPUT
+        )
+        if not has_ec_transfer() or scheduler_output.ec_connector_metadata is None:
+            return output
+        with self.maybe_get_ec_connector_output(
+            scheduler_output,
+            encoder_cache=self.encoder_cache,
+        ) as ec_connector_output:
+            pass
+        return ModelRunnerOutput.with_ec_conn_output(output, ec_connector_output)
```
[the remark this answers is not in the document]
lirx-pd: The EC context manager yields an initially empty output object and populates its completion fields on exit through `finally`. Calling `with_ec_conn_output()` inside the `with` block would evaluate its emptiness check before those fields are populated, causing the EC output to be dropped. The call is therefore intentionally placed after the with block.

vllm_ascend/__init__.py
```diff
@@ -74,9 +74,11 @@ def register():
 def register_connector():
     _ensure_global_patch()
 
+    from vllm_ascend.distributed.ec_transfer import register_connector as register_ec_connector
     from vllm_ascend.distributed.kv_transfer import register_connector
     from vllm_ascend.distributed.weight_transfer import register_engine
 
+    register_ec_connector()
     register_connector()
     register_engine()
```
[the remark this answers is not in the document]
lirx-pd: fix this problem

vllm_ascend/ops/fused_moe/routed_experts.py
```diff
@@ -471,7 +471,7 @@ def init_eplb(self, n_shared_experts):
         self.multi_stage = False
         self.load_counter = None
         self.num_iter = None
-        self.moe_load = torch.zeros(local_num_experts, dtype=torch.int64).npu()
```
hsliuustc0106, line 474: is this change necessary?

vllm_ascend/worker/model_runner_v1.py
```diff
@@ -680,6 +680,21 @@ def reset_encoder_cache(self) -> None:
         self.cached.clear()
         self._pending_encoder_cache_copies.clear()
 
+    def _no_forward_output(self, scheduler_output: "SchedulerOutput") -> ModelRunnerOutput:
```
hsliuustc0106, line 683: does this feature only work in MRV1?

vllm_ascend/ops/fused_moe/routed_experts.py
lirx-pd: This change is required for encoder-only initialization. With --mm-encoder-only, the language model is constructed under torch.device("meta"). The original torch.zeros(...).npu() therefore attempts to copy a meta tensor to NPU, which fails because it has no backing data. Explicitly setting device="npu" avoids that copy. I observed this failure when starting the Qwen3.5-35B encoder.

vllm_ascend/worker/model_runner_v1.py
lirx-pd: This PR focuses on MRV1, where the vLLM-Ascend EPD integration has been validated. Upstream vLLM already provides EC support for MRV2, but the Example EC connector’s end-to-end EPD path is not yet supported with MRV2 in vLLM-Ascend. My earlier MRV2 implementation required some changes to Ascend’s V2 runner beyond the connector itself. I propose handling MRV2 support for both Example and Mooncake ec connector in a follow-up PR so those changes can be reviewed and validated separately.

Tflowers-0129:
Please provide a real NPU consumer test with TP>1, so the reservation callback executes on non-main control threads for rank>0. Correct per-thread NPU device binding is critical for the receive-pool path.

lirx-pd:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

vllm_ascend/distributed/ec_transfer/mooncake.py
```diff
@@ -0,0 +1,175 @@
...
+        self._free = [(0, self._capacity)]
+
+
+class _AscendProducerMemoryPool(ProducerMemoryPool):
+    def __init__(self, capacity: int, transfer: AscendMooncakeTransfer) -> None:
+        super().__init__(capacity, transfer)
+        self._allocator = _AscendContiguousAllocator(capacity)
+
+    def stage(self, tensors: list[torch.Tensor]) -> StagedSources:
+        staged = super().stage(tensors)
```
Yuli-yx, line 103: Could we ensure the NPU staging copy completes before passing its pointer to Mooncake? A cross-stream stale read was reproducible, though Mooncake E2E remains unverified.

lirx-pd: Add NPU staging copies synchronization before Mooncake transfer

lirx-pd:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #16562, opened 2026-09-15
[BugFix]Fix MRV2 Mamba block-table capacity for dspark,eagle3,dflash RecurrentGatedDeltaRule MTE OOM, when running with vLLM 0.28. without enable prefix cache

SkychenLee:

Fix MRV2 Mamba block-table capacity for EAGLE3 when running with vLLM 0.28. With prefix caching disabled and three speculative state blocks, a homogeneous Mamba group wrapped in `UniformTypeKVCacheSpecs` receives only one block-table column instead of four. The upstream 0.28 initializer checks `MambaSpec` directly, so the wrapper hides the additional speculative slots and can cause out-of-bounds state-index reads in GDN.

Before calling the parent KV-cache initializer, use `vllm_version_is("0.28.0")` to restrict this workaround to the vLLM 0.28.0 compatibility path. For that path, unwrap groups whose layer specifications are all identical `MambaSpec` instances, preserving group order, the other group fields, and allocation descriptors. Other versions pass the original configuration directly to the upstream initializer. Adapt the existing Mamba unit tests to call the unwrap helper directly.

This is a compatibility workaround for vLLM 0.28, which the Ascend main branch still supports. Upstream vLLM 0.29 already handles wrapped specifications through cache-type-specific block-table sizing, including the changes in [vllm#50823](https://github.com/vllm-project/vllm/pull/50823) and [vllm#50493](https://github.com/vllm-project/vllm/pull/50493). A TODO marks this workaround for removal once vLLM 0.28 support is dropped.

Affected dspark,eagle3,dflashMRV2 configurations reserve all speculative Mamba state slots and can run with draft graph execution enabled and prefix caching disabled, avoiding the observed GDN illegal-address failure. No new configuration options are required.


vllm_ascend/worker/v2/model_runner.py
```diff
@@ -247,6 +240,9 @@ def sample_tokens(self, grammar_output):
         return output
 
     def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
+        # TODO: Remove this vLLM 0.28 workaround once support for 0.28 is dropped.
+        # vLLM 0.29 already fixes wrapped Mamba block-table sizing upstream.
+        kv_cache_config = unwrap_mamba_kv_cache_groups(kv_cache_config)
```
drslark: Could you add `vllm_version_is("0.28.0")`?

drslark (changes requested):

drslark (approved):
LGTM.

vllm_ascend/worker/v2/model_runner.py
SkychenLee: sure, already add

vllm_ascend/worker/v2/model_runner.py
```diff
@@ -250,9 +247,10 @@ def sample_tokens(self, grammar_output):
         return output
 
     def initialize_kv_cache(self, kv_cache_config: KVCacheConfig) -> None:
-        # vLLM 0.28 model-state code recognizes bare MambaSpec groups, while
-        # Ascend's hybrid cache planner retains per-layer uniform wrappers.
-        kv_cache_config = normalize_mamba_kv_cache_config(kv_cache_config)
+        # TODO: Remove this vLLM 0.28 workaround once support for 0.28 is dropped.
+        # vLLM 0.29 already fixes wrapped Mamba block-table sizing upstream.
+        if vllm_version_is("0.28.0"):
```
nu11ptr7, line 252: `vllm_version_is("0.28.0")` reads like an exact-version match, so 0.28.1/0.28.x would skip the Mamba unwrap workaround and hit the block-table overflow this PR fixes. Worth gating on the minor version (or `<0.29`) so patch releases keep the workaround.

SkychenLee: 0.28.1/0.28.x will  not include in the version

wxsIcey (approved):

SkychenLee:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

SkychenLee:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

linfeng-yuan (approved):

PR #16564, opened 2026-09-15
[BugFix][KV Transfer] Merge overlapping Mooncake registration regions

iKeybot-code:

Hybrid KV cache descriptors can refer to overlapping ranges when shared descriptor buffers and private layer storages resolve into the same backing allocation. The Mooncake transfer engine rejects registering those ranges independently with `Transfer Engine does not support overlapped memory region`, so V2 workers fail during KV cache registration.

This change:

- sorts and coalesces strictly overlapping registration ranges before calling Mooncake;
- preserves adjacent non-overlapping ranges as separate registrations;
- validates that pointer and length lists have matching sizes;
- adds a regression test covering overlapping and independent ranges.

No API or configuration change. It allows Mooncake P/D workers with hybrid KV cache layouts to start instead of failing memory registration.

- `python -m ruff check vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py tests/ut/kv_offload/test_mooncake_connector.py`
- `python -m ruff format --check vllm_ascend/distributed/kv_transfer/kv_p2p/mooncake_connector.py tests/ut/kv_offload/test_mooncake_connector.py`
- `python -m pytest -q tests/ut/kv_offload/test_mooncake_connector.py -k registered_hybrid_buffer --disable-warnings`: `4 passed, 123 deselected` in the Linux Ascend runtime.
- Ascend 950DT manual P/D validation with `VLLM_USE_V2_MODEL_RUNNER=1`, TP2 producer and TP2 consumer: both workers registered the coalesced regions, both TP ranks completed KV transfer, the request returned HTTP 200, and the decoder remained healthy.


nwpu-zxr (approved):

nwpu-zxr:
LGTM.

iKeybot-code:
/rerun

[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

iKeybot-code:
**Background of the failure (supplementary context)**

Triggered on Ascend PD disaggregation with `VLLM_USE_V2_MODEL_RUNNER=1` + a hybrid GDN/Mamba+Attention model (Qwen3.5 class). The combination is what exposes the bug:

- **Hybrid KV layout**: GDN state and FullAttention k/v views live as padded views inside per-layer 2 MiB-aligned private allocations, so multiple logical views map to **overlapping underlying storage ranges** (different groups can even use different paddings).
- **V2 descriptor path**: since vLLM #51718, one `KVCacheTensor` descriptor (`size` + `shared_by`/`layers`) covers multiple layers sharing an allocation. When the storage-base recovery falls back to per-layer private storages, each view's `[ptr, ptr+len)` region is handed to Mooncake **as-is**.
- **Mooncake constraint**: the transfer engine rejects overlapping registrations (`Transfer Engine does not support overlapped memory region`), so D fails at init with `Mooncake memory registration failed`.

None of the three conditions alone triggers it: pure-attention models register disjoint contiguous buffers, V1 layout is 1:1 per layer, and non-PD paths never batch-register. The old registration code implicitly assumed "one KV tensor == one disjoint memory range", which silently broke once all three conditions coexisted.

The fix sorts and coalesces strictly-overlapping `[ptr, ptr+length)` ranges before registration (adjacent-but-non-overlapping ranges stay separate), which is semantics-preserving for Mooncake. Verified on the failing deployment: D now logs `Coalesced 4 overlapping Mooncake KV registration ranges into 1 ranges` and real KV transfer succeeds on both TP ranks.

kunpengW-code (dismissed):

vllm_ascend/xlite/xlite.py
```diff
@@ -661,7 +661,7 @@ def extract_kv_cache(self, kv_caches: list[tuple[torch.Tensor, ...]], /) -> list
         mla_caches: list[tuple[torch.Tensor, ...]] = []
 
         idx = 0
-        index_mask = self.xlite_config.index_full_mask or [True] * self.xlite_config.n_layers
+        index_mask = getattr(self.xlite_config, "index_full_mask", None) or [True] * self.xlite_config.n_layers
```
kunpengW-code: CI has been fixed. Just rebase and revert this change.

iKeybot-code: Done

kunpengW-code (approved):

PR #16569, opened 2026-09-15
[Refactor][Patch] Remove obsolete MiniMax-M2 config patches

AceCoder0:
## Description

Remove `vllm_ascend/patch/platform/patch_minimax_m2_config.py` entirely. Both hooks it installed are obsolete on current main:

### 1. `_verify_quantization` (force `quantization=None` for MiniMax-M2 fp8 on NPU)

The original premise no longer holds:

- fp8 is now in the Ascend platform's `supported_quantization`, and native block-wise fp8 serving exists via `AscendFp8Config` (`vllm_ascend/quantization/configs/fp8_config.py`). The official MiniMax-M2.5 checkpoint format (`weight_block_size: [128,128]`, `activation_scheme: dynamic`, e4m3) satisfies its constraints, so validation would pass without this patch.
- The companion worker-side load-time fp8→bf16 dequant (`MiniMaxM2Model.load_weights` wrap) was removed in #10906. With it gone, forcing `quantization=None` now routes the fp8 checkpoint through the unquantized loader: `weight_scale_inv` is dropped and raw fp8 values are cast into bf16 params, silently producing wrong weights. The patch's effect went from "fallback that works" to "silent corruption path".
- No CI test covers the official fp8 checkpoint (`MiniMax/MiniMax-M2.5`); all MiniMax e2e configs use the w8a8-QuaRot checkpoint with `--quantization ascend`, which never triggers this hook (`quant_method != fp8`). Removing it does not affect the documented/supported serving path.

### 2. `_verify_cuda_graph` (default `HCCL_OP_EXPANSION_MODE=AIV`)

Every supported launch path already sets the variable explicitly:

- `docs/source/tutorials/models/MiniMax-M2.md` — all 4 launch scripts `export HCCL_OP_EXPANSION_MODE="AIV"`
- `tests/e2e/nightly/.../MiniMax-M2.5-*-A2/A3.yaml` — `envs: HCCL_OP_EXPANSION_MODE: "AIV"`

so the model-specific env side effect inside config verification is redundant. Docs/scripts remain the source of truth for this knob.

### Cleanup

Trim the now-stale registry comments in `vllm_ascend/patch/__init__.py`:

- section 13 (the removed file; its items 3/4 described hooks already trimmed by #10906)
- section 10 items for `MiniMaxM2MoE.forward` and `MiniMaxM2Model.load_weights`, which no longer exist in `worker/patch_minimax_m2.py` after #10906 (only the fused attention `forward` remains, now item 1)

## Test Plan

- [x] `ruff check` / `ruff format --check` pass on the touched files
- [x] `python -m compileall` on both modified `__init__.py`
- [x] `grep` confirms no remaining references to `patch_minimax_m2_config` in code/tests/docs
- [ ] CI (maintainer-triggered)

## Notes

Follow-up to #10906 ("Trim upstreamed MiniMax-M2 hooks"), same direction: MiniMax-M2's supported path on NPU is the w8a8-QuaRot checkpoint + `--quantization ascend`; native fp8 (if desired for the official checkpoint) should go through `AscendFp8Config`, which requires a separate accuracy validation on A3 and is out of scope here.


AceCoder0:
The cpu-ut failure (`test_initialize_kv_cache_installs_aclgraph_factory_and_pcp`: `'str' object has no attribute 'kv_cache_groups'`) is a pre-existing main breakage from #16318 (`normalize_mamba_kv_cache_config` added to `NPUModelRunner.initialize_kv_cache`), reproducible in other open PRs' cpu-ut as well. Added a one-line test fix as a separate commit here to get CI green; happy to split it into its own PR if maintainers prefer.

AceCoder0:
Update on the run-selected-tests failures from the previous run:

- a2-1 part 3/5: `test_qwen3_0.6B_basic.py` cannot be collected by pytest because of the dot in the filename (ModuleNotFoundError on `test_qwen3_0`) — a pre-existing infra issue that only bites when the recommender selects the file explicitly. Fixed by renaming to `test_qwen3_0_6B_basic.py` (same convention as `one_card/test_qwen3_0_6b.py`) + updating the two workflow scripts referencing it.
- a3-4 part 2/4 (dsv3_2_sfa_pcp accuracy) and part 4/4 (test_kvpp_pp, DeepSeek-V3.2-W8A8 + `--quantization ascend`): unrelated to this PR's diff (MiniMax patch removal); these tests use DeepSeek W8A8 and don't touch the deleted code. Cross-checking against another PR's run on the same partitions to confirm they're main-side/flaky.

AceCoder0:
Latest run: all partitions green except `a2-1 card-(part 2-5)`, which failed during **git checkout** (runner network timeout: `read tcp ... operation timed out`) before any test executed — pure infra flake. The rerun requires admin rights; could a maintainer please rerun the failed jobs (or merge as-is if that's the workflow)? cc @QwertyJack

AceCoder0:
CI is fully green now (cpu-ut, all run-selected-tests partitions incl. a2/a3/310p/upstream, ci-gate pass). The glm5_2 acceptance flake did not reproduce this round. Ready for review/merge.

PR #16570, opened 2026-09-15
[BugFix][Platform] Re-raise fresh EngineDeadError in GLM usage wrappers

### LINKED ISSUE #16567: [Bug]:[main] Per-request EngineDeadError tracebacks accumulate stack frames and flood logs after a worker crash
### Your current environment

| Item | Value |
| :--- | :--- |
| vllm-ascend | main / v0.26.0rc (incident log from an internal build; relevant frame line numbers match main) |
| vllm | v1 engine, Python 3.11 |
| Model | any model — the mechanism is model-agnostic (incident observed on a GLM-family MoE, deepseek_v2-based, MLA) |
| Hardware / topology | Ascend NPU, PD disaggregated prefill (TP16×DP2) / decode (TP4×DP8), Mooncake KV transfer, MTP speculative decoding |

Note: this is a pure Python-level logging behavior issue, independent of hardware/CANN versions. Environment details are from an internal deployment and partially omitted.

### 🐛 Describe the bug

When a TP worker process crashes during serving, every in-flight request's stream generator logs a full `EngineDeadError` traceback via `logger.exception("Error in chat completion stream generator.")` (vllm `serving.py`). Worse, **the same exception instance is re-raised once per request, and each re-raise appends another round of stack frames to the exception's `__traceback__`** — so the logged tracebacks grow linearly with the number of failed requests.

This is **not specific to GLM models**: the shared-instance re-raise lives in vllm core (`vllm/v1/engine/output_processor.py` `get()`/`get_nowait()` → `raise output`), so any model on any backend hits the same storm when a worker dies with streaming requests in flight. The vllm-ascend GLM reasoning-usage wrapper (`releases/v0.26.0rc` only) just adds one more frame to every round and makes the storm ~33% larger for GLM deployments; the accumulation itself happens without it (our incident log shows 3 of the 4 per-round frames are pure vllm core).

Numbers from our incident (worker died from an NPU-side `NotImplementedError`; 93 streaming requests were in flight):

- 93 `Error in chat completion stream generator.` tracebacks were logged in ~0.5 s
- traceback length **strictly grows with the request index**: first traceback 233 lines, last one 845 lines, max 1,359 lines
- the k-th traceback contains exactly **k rounds** of the repeated frame chain (the 93rd traceback contains 93 rounds)
- total: **41,340 ERROR lines (7.4 MB) in a 10-minute log window**, vs. ~300 lines for the actual root-cause worker traceback — the root cause is completely buried

### Mechanism

1. Worker crashes → `EngineCoreClient.get_output_async()` raises `EngineDeadError` → `output_handler` in `async_llm.py` puts the **same exception object** into every active request's output queue.
2. Each request's `OutputProcessor.get()` does `raise output` (`vllm/v1/engine/output_processor.py`, `get()` / `get_nowait()`) — re-raising the **same** exception object, which prepends this request's call stack onto the shared `__traceback__`.
3. The exception propagates through `AsyncLLM.generate()` (`except EngineDeadError: ... raise`), through vllm-ascend's `patch_glm_reasoning_usage_accounting._tracked_stream_results` wrapper (adds one more frame per round), up to `chat_completion_stream_generator`.
4. vllm `serving.py`'s `except Exception: logger.exception(...)` prints the full accumulated traceback — once per request.

So the k-th failed request prints k rounds of:

```
File ".../vllm/entrypoints/openai/chat_completion/serving.py", line 522, in chat_completion_stream_generator
    async for res in result_generator:
File ".../vllm_ascend/patch/platform/patch_glm_reasoning_usage_accounting.py", line 169, in _tracked_stream_results
    async for res in result_generator:
File ".../vllm/v1/engine/async_llm.py", line 619, in generate
    out = q.get_nowait() or await q.get()
File ".../vllm/v1/engine/output_processor.py", line 86, in get
    raise output
```

and only the final (oldest) round ends at the genuine first-raise site:

```
File ".../vllm/v1/engine/async_llm.py", line 700, in output_handler
    outputs = await engine_core.get_output_async()
File ".../vllm/v1/engine/core_client.py", line 1062, in get_output_async
    raise self._format_exception(outputs) from None
vllm.v1.engine.exceptions.EngineDeadError: EngineCore encountered an issue. See stack trace (above) for the root cause.
```

Verification from the log: the 93 logged tracebacks pair 1:1 with the 93 `Request ... failed (engine dead).` INFO lines; the per-traceback count of `chat_completion_stream_generator` frames goes 25, 26, ... up to 93 — i.e. frame accumulation is per-request, not per-request-batch.

### To Reproduce

```bash
# 1. serve any model (GLM is NOT required — the mechanism is in vllm core and
#    is model-agnostic; the GLM wrapper only adds one frame per round)
vllm serve <model_path> --tensor-parallel-size 4 --port 8000

# 2. start a batch of concurrent streaming requests with a long max_tokens
for i in $(seq 1 50); do
  curl -N -s http://localhost:8000/v1/chat/completions \
    -d '{"model":"auto","messages":[{"role":"user","content":"hi"}],"max_tokens":2000,"stream":true}' &
done

# 3. while requests are in flight, kill ONE TP worker subprocess (ps / npu-smi to find it)
kill -9 <worker_pid>

# 4. observe the api-server log: N "Error in chat completion stream generator."
#    tracebacks, each longer than the previous one
```

### Expected Behavior

- `EngineDeadError` is a global engine state, not a per-request error: its full stack trace should be logged **once**. Note vllm already logs a short per-request INFO line (`Request %s failed (engine dead).` in `async_llm.py`), so the per-request full traceback in the serving layer is redundant.
- The traceback must not accumulate frames across requests — re-raising a shared exception object into N request generators produces O(N²) log volume in total.
- Requests arriving after the engine is known dead should fail fast with a short error message instead of joining the generator chain and printing yet another traceback.

### Suggested fix directions

The root behavior lives in vllm core; the vllm-ascend patch is an amplifier:

- vllm core, `output_processor.py` `get()`/`get_nowait()`: when `output` is an `Exception`, avoid re-raising the shared instance as-is (e.g. raise a copy without the accumulated `__traceback__`, so each request logs only its own round).
- vllm core, `serving.py` `chat_completion_stream_generator`: special-case `EngineDeadError` — log the full stack on first occurrence, then only short per-request lines (the INFO line already exists).
- vllm-ascend, `patch_glm_reasoning_usage_accounting.py`: `_tracked_stream_results`/`_tracked_full_results` are transparent pass-through wrappers (no exception handling), but they add one async-generator frame to every round of the accumulated tracebacks (+33% volume here, 4 frames per round instead of 3) and put this file name into every round, which is misleading during incident triage. This amplification is fixed in #16570 (the wrapper now re-raises a fresh `EngineDeadError` per request); the core-side accumulation itself still needs one of the two vllm fixes above.

### Additional Context

- The real root cause in our incident was a worker-side exception (~300 lines of log). It took manual dedup of 41,340 ERROR lines to find it — the log flood actively hides the root cause during incident response.
- Same behavior applies to `chat_completion_full_generator` (non-streaming path goes through `_tracked_full_results`).

linnea-lin-00638949:

Fixes #16567

When the engine dies, `OutputProcessor.propagate_error()` hands **one shared `EngineDeadError` instance** to every in-flight request. Each request that propagates that same instance up its generator chain prepends its own frames onto the shared `__traceback__`, so the per-request `Error in chat completion stream generator` traceback printed by the serving layer grows by one round of frames per failed request, long after the root cause has already been logged once by `AsyncLLM.output_handler`.

The GLM reasoning-usage wrappers (`_tracked_stream_results` / `_tracked_full_results`, introduced in #13604) sit on this propagation path and forward the shared instance, which makes every failed request print an increasingly large duplicated stack. In the production incident behind #16567, 93 in-flight requests produced 93 stacked tracebacks growing from 233 to 845 lines (41340 ERROR lines in ~40 seconds), burying the actual root cause (a worker crash) in log noise.

This PR catches `EngineDeadError` in both wrappers and re-raises a **fresh** instance with the context suppressed (`raise EngineDeadError() from None`):

- each failed request now logs only its own constant-size traceback;
- the root cause is still logged exactly once by `AsyncLLM.output_handler` (unchanged vllm behavior), and the standard `EngineDeadError` message still points readers to the stack trace above;
- non-`EngineDeadError` exceptions keep propagating untouched.

No API/behavior change. The only observable difference is log output: after a worker/engine crash, each failed GLM request logs one short `EngineDeadError` traceback instead of an ever-growing duplicated one.

Added three unit tests in `tests/ut/patch/platform/test_patch_glm_reasoning_usage_accounting.py`:

- `test_tracked_stream_results_reraise_fresh_engine_dead_error` and `test_tracked_full_results_reraise_fresh_engine_dead_error`: consume a wrapper-wrapped generator that raises a shared `EngineDeadError` for 3 simulated requests; assert each request raises a *fresh* `EngineDeadError` (not the shared instance), with the context suppressed, and that the traceback depth stays constant across requests.
- `test_shared_engine_dead_error_traceback_accumulates_without_wrapper`: control group proving the mechanism — consuming the raw shared instance directly makes its traceback depth grow strictly per request (17 → 22 → 27 frames in a 3-request simulation), which is what floods the server log without this fix.

The same scenario was also verified end-to-end with a standalone script (no NPU required) reproducing the wrapper logic: fixed wrappers keep a constant depth of 6 frames per request while the unwrapped control accumulates +5 frames per request.


vllm_ascend/patch/platform/patch_glm_reasoning_usage_accounting.py
```diff
@@ -162,24 +163,43 @@ def _create_stream_usage_state(
     return _StreamUsageState(counters=counters)
 
 
+def _raise_fresh_engine_dead() -> NoReturn:
```
nu11ptr7, line 166: Raising a fresh `EngineDeadError()` avoids the ever-growing traceback, but it also drops the original exception message - per-request logs now only carry the copy printed by `AsyncLLM.output_handler`. Worth passing the original text through (`EngineDeadError(str(error)) from None`) so the failing request still explains why the engine died.

PR #16579, opened 2026-09-15
[MRV2][BugFix][Kernel] Support apply_thinking_budget on Ascend

Liamup777:

Support the Model Runner V2 thinking-budget Triton path on Ascend.

The upstream implementation currently exposes two Triton-Ascend compiler compatibility issues:

- `_load_effective_token` uses pointer arguments and branch-local returns when called from dynamic `tl.range`, which can fail during TTIR-to-Linalg lowering.
- `_update_committed_marker_cache_kernel` uses chained runtime boolean expressions unsupported by affected Triton-Ascend versions.

This patch adds Ascend-compatible implementations under `vllm_ascend/ops/triton/v2/sample/` and rebinds the two upstream functions. The upstream `ThinkingBudgetState`, `apply_thinking_budget`, and `_thinking_budget_kernel` remain unchanged.

The `_load_effective_token` workaround can be removed after the new Q4 Triton-Ascend release is available. The marker-cache workaround can be removed when Triton-Ascend 3.6.0 is the minimum supported version.

No

Passed:

- `bash format.sh`
- `SHELLCHECK_OPTS="--exclude=SC2046,SC2006,SC2086" pre-commit run --all-files --hook-stage manual --show-diff-on-failure`
- Python compile validation
- `git diff --check`
- `pytest -sv tests/e2e/pull_request/one_card/test_thinking_budget.py`


vllm_ascend/patch/worker/patch_v2/patch_triton.py
```diff
@@ -2,7 +2,16 @@
...
-from vllm.v1.worker.gpu.sample import bad_words, gumbel, logprob, penalties, prompt_logprob, sampler, states
+from vllm.v1.worker.gpu.sample import (
+    bad_words,
+    gumbel,
+    logprob,
+    penalties,
+    prompt_logprob,
+    sampler,
+    states,
+    thinking_budget,
```
nu11ptr7, line 13: `thinking_budget` is a new upstream sample module and is now imported unconditionally in this patch chain. On vLLM versions that don't ship it (the 0.26/0.27 branches this repo still supports), the import fails and takes the whole platform patch import down. Suggest guarding it with try/except or a version check like the other optional patches.

Liamup777: Thanks for the review. This PR targets the `main` branch, which corresponds to upstream vLLM v0.28+. The v0.26 and v0.27 compatibility lines have already been released as tags and do not consume new changes from `main`. Therefore, the `thinking_budget` module is expected to exist for the supported upstream version on this branch, and we will keep the import unconditional rather than add a compatibility guard for the older tagged releases.

AuroraEmiya (approved):
LGTM

ZT-AIA (approved):

linfeng-yuan (approved):

PR #16580, opened 2026-09-15
[BugFix][ModelRunner] Prevent AIV event reuse after long prefill

Yaphets24:
> **Status: draft — issue still reproduces.**\n>\n> A later cold-start rerun without MTP reproduced 
tEventCreateExWithFlag on the 9-token tail after a 6000-token prefill chunk. The current batch-boundary synchronization changes timing but is not a complete fix. Investigation is continuing.\n\n### What this PR does\n\nExperimental synchronization after target and optional draft prefill forwards when HCCL AIV is enabled.\n\n### Current verification\n\nTP16, AIV enabled, async scheduling, 6000+tail chunked prefill, concurrency 1:\n\n- Earlier runs passed with and without MTP.\n- A later cold restart without MTP failed on the first request.\n- Therefore this patch is not considered a fix and the PR remains Draft.


vllm_ascend/worker/model_runner_v1.py
```diff
@@ -2501,6 +2502,15 @@ def execute_model(
...
                 num_tokens_padded, input_ids, positions, intermediate_tensors, inputs_embeds, **model_kwargs
             )
+            is_prefill_batch = bool(torch.any(
+                self.input_batch.num_computed_tokens_cpu_tensor[:num_reqs]
+                < self.input_batch.num_prompt_tokens_cpu_tensor[:num_reqs]
+            ))
+            # AIV all-reduce is asynchronous with respect to the host. Drain
+            # every prefill forward at the batch boundary, including a short
+            # final chunk. Decode remains asynchronous and pays no sync cost.
+            if is_prefill_batch and os.getenv("HCCL_OP_EXPANSION_MODE") == "AIV":
```
nu11ptr7: Relying on `os.getenv("HCCL_OP_EXPANSION_MODE") == "AIV"` is brittle: any other spelling (lowercase, stray whitespace, a wrapper value) silently skips the drain and the AIV event-reuse bug returns. Suggest normalizing + caching the env read at init (or querying the platform capability) so the sync can't be missed.

Yaphets24:
Superseded by #16635, which contains only the validated shared-output functional all-reduce fix.

PR #16582, opened 2026-09-15
[BugFix][MoE] Fix moe activaton selection

earthmanylf:

Fix moe activation selection. MoECommMethod is a singleton which holds moe_config, and  this moe_config always holds the information of the last layer, so if the model has multiple activation methods, the moe activation selection will be wrong. We can just use upstream layer.activation to determine the moe activation method.

No

Step3.5-mxfp8 aime26 acc: 100%


weijinqian0 (approved):

linfeng-yuan (approved):

lcfenglinwan (approved):

earthmanylf:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

PR #16243, opened 2026-09-10
[Feature][Ops] Add Triton KeyPool compression and pooled indexing

lijiahang226:
### What this PR does / why we need it?

Refs #15665

Add Triton KeyPool state compression, paged cache writes, and pooled-key selection for GLM-5.3-Flash. Allow `IndexerWrapper` to select a model-provided backend while preserving checkpoint parameter names. Both operators reuse the shared `next_power_of_2` helper.

Include single-operator accuracy tests and documentation covering formulas, parameters, layout constraints, graph replay, and test commands. Model-specific backend integration is handled separately.

### Does this PR introduce _any_ user-facing change?

Yes. Provides the Triton pooled-indexing operators and a backend-selection hook. This PR does not enable the model backend by itself or change cache allocation and grouping.

### How was this patch tested?

Validated on Atlas A3 with PyTorch 2.10.0, torch-npu 2.10.0.post4, and Triton-Ascend 3.2.0:

- 10 tests passed in `tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_glm5next_kpool_triton.py`: CPU-reference accuracy, historical windows, rollback, paging, noncontiguous storage, invalid slots, graph padding, empty inputs, and eager/graph replay.
- 9 tests passed in `tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_glm5next_pool_key_indexer_triton.py`: multiple requests, pool capacities, paged caches, token chunking, causal tails, and eager/graph replay with changing inputs.
- 16 tests passed across `tests/ut/models/test_glm5next_indexer_backend.py` and `tests/ut/ops/test_mla.py`.

The CI follow-up adds a type annotation to the test reference data without changing test behavior. Hardware coverage is limited to Atlas A3; these tests establish operator accuracy, not model-level throughput.

- vLLM main: https://github.com/vllm-project/vllm/commit/a97dacb7106ee49f39f3d1fc6ae1800ff724e01d

PR #16584, opened 2026-09-15
[BugFix][Ops] Keep pooled indexer score masks finite

lijiahang226:

Follow-up to #16243. The pooled indexer leaves invisible pool sub-tiles at the score buffer's initial `-inf` value and also writes `-inf` to invalid lanes of partially visible sub-tiles. These values reach the NPU `torch.topk` input.

Use one finite FP32 minimum sentinel for initialization, Triton masking, and invalid top-k result detection. This preserves the visible-score calculation and dynamic sub-tile skipping without adding an inference-path kernel launch or host/device synchronization.

Add eager and graph regression coverage that checks the actual top-k input buffers, selected token indices, history padding, and causal tail.

Fixes mask-generated non-finite inputs to pooled-key selection. No public API or configuration changes. Normal finite score ordering is preserved; NaN/Inf from valid inputs or arithmetic overflow is outside this fix.

- **A3 operator validation: 11 passed**, with CANN 9.1.0, PyTorch 2.10.0, torch-npu 2.10.0.post4, and Triton-Ascend 3.2.0. The exact PR operator source was loaded into an existing NPU environment and the following repository test file was executed with pytest:
  ```bash
  pytest -sv tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_glm5next_pool_key_indexer_triton.py
  ```
- The new eager regression fails on the original operator at the finite-input assertion **before top-k dispatch**. This reproduces the unwanted `-inf` input, without relying on a backend crash.
- New cases use BF16 `[T, 32, 128]` queries, pool size 4, and top-k 2048. They cover visible pool counts 0, 1, 127, 128, 129, 2047, 2048, 2049, and 2050; distinct negative scores; token chunking; and shrinking visible lengths during graph replay. Expected output indices and padding match exactly.
- Existing tests cover paged/noncontiguous caches, mixed request lengths, graph padding, empty queries, and tail-only output.
- Two all-visible graph replay comparisons produced identical outputs and showed no latency regression: 305.16 -> 300.70 us for 8 tokens / 2050 pools, and 343.30 -> 340.99 us for 16 tokens / 8192 pools (median of 7 rounds, 30 replays per round). These are operator measurements, not full-model throughput claims.
- Local Ruff, formatting, codespell, typos, and Markdown checks passed. `bash format.sh ci` was attempted; the full run is not green because some shell hooks require `/bin/bash` and `shellcheck`, which are unavailable in the local Windows environment. Those hooks need Linux CI.
- Validation scope is operator correctness and graph replay on A3. Full-model accuracy, A2, and A5 were not tested.


lijiahang226:
Closing this one deliberately as part of cleaning up my open PR backlog.

For the record, the red CI here was not caused by this change: `cpu-ut` failed on `tests/ut/worker/test_model_runner_v2.py::test_initialize_kv_cache_installs_aclgraph_factory_and_pcp` with `AttributeError: 'str' object has no attribute 'kv_cache_groups'` (4707 passed, 1 failed), and `ci-gate` only failed because it propagates the `cpu-ut` result. That looks like an unrelated flaky main2main test rather than anything about the pooled indexer mask.

The real blocker was that the branch had accumulated merge conflicts and stayed unlabelled/unreviewed for ~10 days, so I am dropping it rather than keeping a stale branch open. The head commit `eef3ca67a` stays on my fork branch `fix-glm5next-score-mask` if the finite-sentinel fix is needed again.

PR #16508, opened 2026-09-14
[Test][CI] Split MRv2 basic E2E tests for parallel scheduling

Liamup777:
### What this PR does / why we need it?

The ModelRunner V2 basic E2E test file contains 15 parameterized cases. Since CI schedules tests at file granularity, all cases in the combined file run serially on one NPU runner.

Split the combined test into six model- and feature-specific files so CI can schedule them independently across the available one-card partitions.

This change preserves all existing test parameters and assertions, including the DSpark adaptive verification coverage added upstream. It updates the main2main test targets and estimated timing entries, refreshes the generated E2E coverage report, and corrects the `egale` spelling in the EAGLE test name.

### Does this PR introduce _any_ user-facing change?

No. This only changes E2E test organization and CI scheduling metadata.

### How was this patch tested?

Passed:

- `bash format.sh`
- `bash format.sh ci`
- `SHELLCHECK_OPTS="--exclude=SC2046,SC2006,SC2086" pre-commit run --all-files --hook-stage manual --show-diff-on-failure`
- `UV_PROJECT_ENVIRONMENT=/tmp/pr16508-coverage-venv uv run --with regex --with pyyaml python3 .github/workflows/scripts/coverage.py`
- Static validation that all 15 parameterized cases are preserved
- Validation that all main2main targets and coverage report paths resolve
- Repository CI configuration validation
- Rebase and conflict-resolution check against the latest `upstream/main`

NPU E2E validation has not completed successfully on A3. Initial attempts failed during collection due to an incompatible vLLM/vLLM-Ascend environment; a later DFlash attempt stopped producing logs during model loading, with the cause still unconfirmed. Full NPU validation and measured CI duration improvement remain pending.

- vLLM main: https://github.com/vllm-project/vllm/commit/84030bbe3d74d99bad477a3d2e37a973ccd8865c

PR #16587, opened 2026-09-15
[Test][CI] Fix Qwen3 basic test module name

Liamup777:

The Qwen3 basic E2E module introduced in #16508 has a decimal point in its filename. Rename `test_qwen3_0.6B_basic.py` to `test_qwen3_0_6B_basic.py` so the module basename is a valid Python identifier. Update all five references in `estimated_times.yaml`, `main2main_tests.json`, and `tests/e2e/coverage.html`.

This branch is currently stacked on #16590 and includes its original commit `3145275867c0208ca6a12ad40677c0e10fa2b7dd` to unblock CPU CI. After KV cache normalization was added, `test_initialize_kv_cache_installs_aclgraph_factory_and_pcp` still passed the string `"kv"`, causing `AttributeError: 'str' object has no attribute 'kv_cache_groups'`. The fix from #16590 supplies a minimal valid `KVCacheConfig` and updates the equality assertion while preserving the existing ACLGraph, KVPP, PCP, and routed-expert assertions. Please merge #16590 first; the filename change is the scope of this PR on top of that dependency.

No. The renamed E2E test contents are byte-for-byte unchanged. The dependency from #16590 only updates a unit-test fixture; no production code or runtime behavior changes.

- Verified that the unit-test file exactly matches #16590 and the rename/reference changes exactly match the previously submitted revision.
- Compared the renamed E2E test with the original file: identical contents.
- `UV_PROJECT_ENVIRONMENT=/tmp/pr16508-coverage-venv uv run --with regex --with pyyaml python3 .github/workflows/scripts/coverage.py`: passed, 127/127 E2E files covered.
- `bash format.sh`: passed.
- `bash format.sh ci`: passed.
- `SHELLCHECK_OPTS="--exclude=SC2046,SC2006,SC2086" pre-commit run --all-files --hook-stage manual --show-diff-on-failure`: passed.
- `git diff --check`: passed.

CPU UT was not run locally because this macOS environment lacks torch and vLLM. NPU E2E tests were not run locally because NPU hardware is unavailable. CI results for the updated head are pending.


Liamup777:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

linfeng-yuan (approved):

PR #16590, opened 2026-09-15
[Test][Misc] Update test_model_runner_v2 to use KVCacheConfig

qijiajin:

`NPUModelRunner.initialize_kv_cache()` now inspects/normalizes the incoming `KVCacheConfig`. The existing unit test passed the string `"kv"` as a simplified fixture, which no longer satisfies the real method contract and causes the CPU UT to fail before the test reaches its original ACLGraph/KVPP/PCP wiring assertions.

This PR updates the test fixture only:

- imports `KVCacheConfig`;
- constructs a minimal valid `KVCacheConfig` with empty cache groups;
- passes that config to `initialize_kv_cache()`;
- checks equality with the config received by the mocked parent method.

No production code or runtime behavior is changed.

The change is limited to `tests/ut/worker/test_model_runner_v2.py` and preserves the existing assertions for ACLGraph factory installation, KVPP creation, PCP/model-state wiring, and routed-expert capturer initialization.


linfeng-yuan (approved):

huamus:
/rerun
[Bot]: rerun command failed: you do not have permission. Only the PR author or users with triage+ permission can trigger /rerun.

huamus:
Hi @qijiajin, thanks for the quick fix! This PR unblocks the cpu-ut regression introduced by #16318 — our #16328 is hitting the same `test_initialize_kv_cache_installs_aclgraph_factory_and_pcp` failure (`'str' object has no attribute 'kv_cache_groups'`) after rebasing onto the latest main snapshot.

The only red check left here is `run-selected-tests / a2-1 card-(part 2-5)`, which fails during collection of `test_qwen3_0.6B_basic.py` — that module-name issue was already fixed on main by #16587 (`fbb75a4`). Could you `/rerun` the failed a2-1 job (or push an empty update to re-trigger CI)? It should pass against the current main snapshot and clear the way for merging. Thanks!

PR #16599, opened 2026-09-15
[CI] Fix precise test file selection logic

MrZ20:
The precise test selector previously used the regex `test_\w+\.py`.

This pattern did not recognize `test_qwen3_0.6B_basic.py` because `\w` does not match the dot in the filename. As a result, the newly added test was not selected.

PR #16587 fixed the immediate naming issue by renaming the file to use underscores.

This PR provides a follow-up fix to prevent similar problems:

- Match test filenames using the `test_*.py` pattern.
- Allow any characters between `test_` and `.py`.
- Keep the selection scope limited to `tests/e2e/pull_request/` and `tests/ut/`.
- Apply the same rule to added, modified, renamed, and deleted test files.


MrZ20:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

wangxiyuan (approved):

PR #16611, opened 2026-09-15
e2e: change Kimi-K2.5 perf case to concurrency 4 with 16 prompts

xiayingqing:
**What does this PR do?**

Changes the Kimi-K2.5-W4A8 nightly perf case from 2 concurrent requests / 8 prompts to 4 concurrent requests / 16 prompts (num_prompts: 8 -> 16, batch_size: 2 -> 4).

**Why?**

With batch_size=2 < DP=4, the DP load balancer tie-break rotation (vllm #47420) interacts with the shared 131K prefix cache: requests get spread across 4 engine replicas instead of sticking to 2, forcing redundant full-length prefills of the shared prefix and making the throughput metric sensitive to the tie-break policy. At concurrency 4 (= DP), all replicas are naturally utilized regardless of tie-break behavior, making the metric more stable while still exercising the prefix-cache-heavy path.

Validation data: at concurrency 4, throughput rises from ~28.5 tok/s (conc=2) to ~50.9 tok/s on the same server setup, and the A/B spread between load-balancer tie-break variants collapses to noise level.

**Baseline**: intentionally left unchanged in this PR (34.2463 on main / 1 on v0.23); re-baselining will follow after this change lands.


wenjun91:
/nightly  kimi-k2.5
nightly command triggered.
- Nightly-A3: [workflow run](https://github.com/vllm-project/vllm-ascend/actions/runs/34965307693)

PR #16617, opened 2026-09-15
[Test] Disable compile cache for Qwen3.6 27B nightly

q664171689:

Disables the vLLM compile cache for the Qwen3.6-27B A5 nightly single-node configuration.

No.

## How was this patch tested?

Validated the YAML configuration parses successfully and that `VLLM_DISABLE_COMPILE_CACHE` resolves to `"1"`.


kitaikuyo421esfj-hub (approved):
去掉缓存

PR #16635, opened 2026-09-15
[BugFix][Distributed] Preserve AIV all-reduce buffers across async phases

Yaphets24:

Small AIV all-reduce reads peer buffers asynchronously. A rank can finish locally and reclaim its buffer while a slower peer is still reading it, causing an invalid GM address.

Use independent storage for AIV messages up to 2 MiB and track pending outputs across communicator instances, because the model runner and collective path may hold different wrappers. Release storage only after local completion plus a TP-group barrier. Target AIV work is also retired before MTP draft execution. Larger collectives and non-AIV modes keep the existing path.

No.

- Communicator unit tests: 6 passed.
- TP16, AIV, async scheduling, cold start, no MTP: 4/4 requests passed.
- TP16, AIV, async scheduling, cold start, MTP: 4/4 requests passed.
- No AIV invalid-GM or HCCL errors in either run.


Yaphets24:
Two additional no-MTP cold-start reruns both reproduced the failure on the first request. Both failed in aclnnMoeGatingTopK with EZ1008 tiling execution errors and left the worker group hung. This shows the functional shared-output all-reduce change is not a complete fix; returning the PR to Draft.

PR #16656, opened 2026-09-16
[BugFix][SFA] Fix padded-index LSE and empty shards for A5 DCP

recky-c:

A5 SFA DCP decode can select fewer keys on a rank than its local KV length. The sparse index tensor then contains a valid prefix followed by `-1` padding. Using the KV length as the softmax extent includes invalid entries and produces incorrect per-rank normalization for the DCP merge.

This patch ports the focused fixes from our standalone SFA operator experiment into the existing custom operator:

- Bound sparse-mode-0 LSE computation by the actual valid index prefix.
- Initialize skipped queries to zero output and zero softmax max/sum (reconstructed LSE is negative infinity), including queries with a nonempty cache but no selected local keys.
- Initialize `n2Size` before output initialization.
- Make prefill KV and RoPE views contiguous after the packed DCP gather.
- Advertise `SFA_DCP_REPLICATED_INDEXER` on A5 while retaining the platform capability check.

The current custom operator already accepts PA_BSND with `return_softmax_lse`, so no host-tiling change is needed on this base. No Mooncake changes or bulk operator-source refresh are included.

Enables the A5 SFA DCP capability and fixes its BF16 KV-cache path. The validated target is full GLM5.2 with W4A8 weights and **both SFA C8 and indexer C8 disabled**, TP8/DCP8/EP, eager execution, no MTP or PD disaggregation. This PR does not claim C8 support.

The full-model experiment also applied #16325 (tested revision `ec6e301d183c5da1023652744b4e9d6395172b86`) for indexer metadata. Those changes are not included here; the newer revision of that PR has not been validated as part of this work.

- Historical experiment: the equivalent numerical fixes in standalone ops-transformer, together with contiguous prefill inputs, the capability bypass and the above #16325 revision, passed all six full 78-layer GLM5.2 functional cases on A5 (arithmetic, factual answer, translation, Python generation, long-context retrieval and repeated arithmetic).
- This is a minimal port to the custom operator on main `4c5ee3320`, not the exact operator build used in that experiment. **The port has not yet been recompiled or rerun on NPU.** Historical results are not evidence of a completed regression for this exact commit.
- Added an A5 operator regression comparing output and LSE against a PyTorch reference for 0, 1, 127, 128, 129 and 257 selected keys, including a nonempty cache with no selected keys.
- Added a prefill regression using non-contiguous packed-cache split views; updated the hardware capability expectation.
- Ruff 0.14.0 check/format, Python syntax parsing and `git diff --check` passed. The repository-wide `bash format.sh ci` could not run because pre-commit is not installed in this local environment. New unit/NPU tests have not been executed in this PR preparation.

Performance, graph mode, MTP, C8 and PD combinations remain outside the completed validation scope.


zhenwenqi2024 (approved):

paddy-admin (approved):
the new `.contiguous()` shouldn't impact perf for A3 since it's overlapping with indexer processing, still some headroom left.

ningjingbengxiaohai (dismissed):

ZT-AIA (approved):
Make this modification and subsequently add it to the operator description document.

PR #16661, opened 2026-09-16
[BugFix][Config] Strip upstream-injected GDN/KDA prefill backend keys on Ascend

lizy124:
﻿What this PR does / why we need it?

Upstream EngineArgs injects --gdn-prefill-backend / --kda-prefill-backend into additional_config. The generic GDN/KDA layers (qwen_gdn_linear_attn / kimi_gdn_linear_attn) consume these keys on all platforms, but on Ascend only the triton path is available (CUDA-only flashinfer / cutedsl / flashkda have no kernel). The injected keys hit the strict extra="forbid" AscendConfig and abort startup with an unexpected_keyword_argument validation error, even when the user selects the triton path that Ascend actually supports.

Strip the injected keys in init_ascend_config (added to _NON_USER_INPUT_KEYS, same mechanism as the already-injected scheduler_config) and warn only when the user requested an unsupported CUDA-only value; triton/auto are stripped silently since they match the default triton path.

Yes. On Ascend, passing --gdn-prefill-backend / --kda-prefill-backend previously crashed startup with pydantic unexpected_keyword_argument; triton/auto values now behave like the default, CUDA-only values warn and fall back to triton. Other config typo detection stays strict.

How was this patch tested?

Added 3 unit tests in tests/ut/test_ascend_config.py: CUDA-only values warn and build succeeds, triton/auto stripped silently with no warning, keys absent emits no warning. Verified locally with ruff 0.14.0 check/format and Python AST compile; will also run pytest -sv tests/ut/test_ascend_config.py in an Ascend container.


vllm_ascend/ascend_config.py
```diff
@@ -1536,6 +1536,26 @@ def init_ascend_config(vllm_config):
...
+    # only the triton path is available: the FLA Triton kernels run on Ascend
+    # via triton-ascend (the CUDA triton package is replaced in Ascend images).
+    # CUDA-only values (flashinfer/cutedsl for GDN, flashkda for KDA) have no
+    # kernel on Ascend. Strip the keys here so extra="forbid" does not reject
+    # them as typos, and warn only when the user requested an unsupported value.
+    _TRITON_COMPATIBLE_VALUES = ("auto", "triton")
+    for _prefill_key in ("gdn_prefill_backend", "kda_prefill_backend"):
+        _prefill_value = additional_config.get(_prefill_key)
+        if _prefill_value is not None and str(_prefill_value).strip().lower() not in _TRITON_COMPATIBLE_VALUES:
+            logger.warning(
```
kunpengW-code: Please replace it with `warning_once`.

lizy124: Done in 229a661: switched to `logger.warning_once` and updated the UT assertions to `warning_once.call_args_list` accordingly.

kunpengW-code (approved):

PR #16664, opened 2026-09-16
[BugFix][Attention] Fix C8 slot mappings under PCP and DCP

recky-c:

Fixes two C8 cache-write slot mapping issues exposed by PCP and PD disaggregation:

1. PCP can supply an int64 indexer `slot_mapping`, while the native `StoreKvBlockMetadata` kernel reads it as `int32_t*`. Convert only the operator argument to int32, preserving the original metadata.
2. In the combined PCP+DCP SFA path, PCP gathers a full token view before the deferred C8 cache write, but the inherited DCP helper truncated the slot mapping to the local token count. Override the helper in `AscendSFAPCPDCPImpl` so the gathered KV rows use the complete DCP slot mapping. The ordinary DCP path keeps its existing local slice.

The tests extend the existing indexer dtype coverage and verify that the combined PCP+DCP implementation keeps the complete SFA slot mapping.

Fixes incorrect indexer and SFA C8 cache writes that can produce incorrect completions when PCP, DCP, and PD disaggregation are combined. No configuration changes.

- Python compilation and `git diff --check` passed.
- The indexer module tests passed 5/5 in the existing Ascend validation environment.
- Hardware validation used main `28acc786` + #16325 (`f56cb5b2`) + #16492 (`037f5143`), paired with vLLM `84030bbe`, GLM-5.2 W4A8C8, MRV2, EP, async scheduling, FULL_DECODE_ONLY, MooncakeConnectorV1, and P TP8 x PCP2 -> D TP8.
- A 48-sample probe confirmed the SFA failure: gathered KV had 4096 rows and the DCP context had 4096 slots, but only 2048 slots were passed to the C8 write. The typical long-request sample had 128 incorrect valid rows.
- After the fixes, P DCP16 -> D DCP1 and P DCP16 -> D DCP8 each passed 28/28 requests with both SFA C8 and indexer C8 enabled. Each run transferred 87,767 external KV tokens. The isolated SFA C8 and combined-C8 boundary suites also passed 3/3.

The hardware results apply to the stated combined checkout and topology.


vllm_ascend/attention/indexer.py
```diff
@@ -519,7 +519,8 @@ def build(
 
         if get_ascend_config().c8_reshape_optim_enabled:
             torch.ops._C_ascend.store_kv_block_metadata(
-                slot_mapping,
+                # The native metadata kernel reads slot indices as int32.
+                slot_mapping.to(torch.int32),
```
kunpengW-code, line 523: Not recommended to make changes in common code paths. Instead, add type conversion at the points where PCP introduces differences.

PR #16682, opened 2026-09-16
[BugFix][Multimodal] Port upstream mm_encoder_only early-exit in dummy runs

lizy124:

With --mm-encoder-only, vLLM keeps LM weights on the meta device and only materializes the multimodal encoder. The Ascend model runner still executes the LM forward in its compile-time dummy runs, so an npu:0 activation meets a meta parameter inside the npu_gemma_rms_norm FakeTensor dispatch and startup always fails with Unhandled FakeTensor Device Propagation. Upstream gpu_model_runner.py already skips its dummy runs for this mode (_dummy_run / _dummy_sampler_run / _dummy_pooler_run); this fork did not carry those early exits over. Port the same early exits into the two dummy runs that exist here (+11 lines, identical to upstream). _dummy_pooler_run does not exist in this fork, so no change is needed there.

Yes. On Ascend, --mm-encoder-only previously crashed at startup (Engine core initialization failed); it now starts normally. mm_encoder_only defaults to False, so all other paths are unchanged.

How was this patch tested?

Verified on Ascend hardware (8x910, container, vllm 0.28.0):
- Before: vllm serve with --mm-encoder-only always crashed in _dummy_run with TorchRuntimeError: Unhandled FakeTensor Device Propagation for npu_gemma_rms_norm (npu:0 vs meta).
- After: the encoder-only server reaches Application startup complete and /health returns 200.
- Regression: baseline serve without the flag starts normally and a real completion request returns the same output as before the patch.
- Local: python -m py_compile and ruff 0.14.0 check/format pass.


zouzy5137 (approved):
LGTM

lizy124:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

kunpengW-code (approved):

wangxiyuan (approved):

PR #16687, opened 2026-09-16
[BugFix][Ops] Remove hardcoded quant_mode check in QLI V2 adapter

ZT-AIA:
### What

Remove the hardcoded `quant_mode == 2` and INT8/FP16 dtype checks from `QuantLightningIndexerCandidate` in `quant_lightning_indexer_v2_torch_adpt.h`.

### Why

- The Python side dispatches the quantization mode per hardware via `DeviceOperator.get_dsa_indexer_quant_mode()`: non-A5 returns `2` (INT8 query/key + FP16 scales/weights), while Ascend 950 (A5) returns `1` (FP8 e4m3 query/key + FP32 scales/weights).
- The aclnn kernel itself supports both modes on 950 and validates unsupported combinations internally, so the adapter-level check only blocked the legitimate FP8 path on A5 with a hard error, while duplicating (and drifting from) the kernel's own constraint table.
- The mode and dtypes always come from the same device operator, so they are consistent by construction; passing an invalid combination is already rejected by the two-stage aclnn call.

### Test

- Built and validated the QLI V2 path on Ascend hardware with both quant modes.


ZT-AIA:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

linfeng-yuan (approved):

Issue #16049, opened 2026-09-08
[Bug]: off-stride `max_num_seqs` silently disables ACL graph capture for the largest batches (~7.5x slower decode)


MirkoDeVita98:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
Collecting environment information...
PyTorch version: 2.10.0+cpu
Is debug build: False

OS: Ubuntu 22.04.5 LTS (aarch64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
Clang version: 15.0.7
CMake version: version 4.4.0
Libc version: glibc-2.35

Python version: 3.12.13 (main, Jul  7 2026, 16:29:36) [GCC 11.4.0] (64-bit runtime)
Python platform: Linux-6.8.0-101-generic-aarch64-with-glibc2.35

CPU:
Architecture:                            aarch64
CPU op-mode(s):                          64-bit
Byte Order:                              Little Endian
CPU(s):                                  192
On-line CPU(s) list:                     0-191
Vendor ID:                               HiSilicon
Model name:                              Kunpeng-920
Model:                                   0
Thread(s) per core:                      1
Core(s) per cluster:                     48
Socket(s):                               -
Cluster(s):                              4
Stepping:                                0x1
BogoMIPS:                                200.00
Flags:                                   fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma dcpop asimddp asimdfhm ssbs
L1d cache:                               12 MiB (192 instances)
L1i cache:                               12 MiB (192 instances)
L2 cache:                                96 MiB (192 instances)
L3 cache:                                192 MiB (8 instances)
NUMA node(s):                            8
NUMA node0 CPU(s):                       0-23
NUMA node1 CPU(s):                       24-47
NUMA node2 CPU(s):                       48-71
NUMA node3 CPU(s):                       72-95
NUMA node4 CPU(s):                       96-119
NUMA node5 CPU(s):                       120-143
NUMA node6 CPU(s):                       144-167
NUMA node7 CPU(s):                       168-191
Vulnerability Gather data sampling:      Not affected
Vulnerability Indirect target selection: Not affected
Vulnerability Itlb multihit:             Not affected
Vulnerability L1tf:                      Not affected
Vulnerability Mds:                       Not affected
Vulnerability Meltdown:                  Not affected
[...]

PR #16692, opened 2026-09-16
[BugFix] Pad `max_cudagraph_capture_size` to a multiple of 8/16 to avoid eager fallback.

zhiyu-wa:
Pad `max_cudagraph_capture_size` to a multiple of 8 or 16 to avoid eager fallback.

Details: https://github.com/vllm-project/vllm-ascend/issues/16049

No.
Tested locally.


MirkoDeVita98:
Tested on 8×910B2: Qwen3-30B-A3B, TP=4, max_num_seqs=12, batch 12, ignore_eos,
best of 3.

  stock                 [1, 2, 4, 8]       73.6 tok/s   162.98 ms/step
  this PR               [1, 2, 4, 8, 16]   74.7 tok/s   160.72 ms/step
  capturing 12 exactly  [1, 2, 4, 8, 12]  555.9 tok/s    21.59 ms/step

The list gains 16, but decode still runs eager.

In FULL_DECODE_ONLY a uniform decode batch is one token per request, so
num_tokens == num_reqs <= max_num_seqs. Nothing can fill a size-16 graph when
max_num_seqs is 12. Same list with max_num_seqs raised to 16 gives 500.2 tok/s —
the graph is fine, it's unreachable at 12.

So this can't be fixed inside _get_default_max_cudagraph_capture_size: it returns
the ceiling, and upstream's grid can't land on an off-grid one. The size has to
be added to the list. One option, in apply_config_platform_defaults:

    sizes = [s for s in (1, 2, 4) if s <= max_capture_size]
    sizes += list(range(8, min(max_capture_size + 1, 256), 8))
    sizes += list(range(256, max_capture_size + 1, 16))
    compilation_config.cudagraph_capture_sizes = sorted({*sizes, max_capture_size})

Leave max_cudagraph_capture_size unset so upstream derives it from the list —
otherwise SP filtering that drops the top size raises instead of truncating. And
skip this when performance_mode == "interactivity", where upstream builds a
contiguous 1..min(max, 32) list.

Separately, `> 4` leaves a ceiling of 3 unpadded, and 3 isn't on the [1, 2, 4]
grid, so the list stops at 2. Hit by max_num_seqs=3, or max_num_seqs=1 with 2
speculative tokens.

Issue #16674, opened 2026-09-16
[Bug]: split_qkv_rmsnorm_rope returns wrong, non-deterministic Q/K due to a racy RoPE position read


Celeste-jq:
### Your current environment

vLLM Ascend:    0.20.2rc2.dev0+g367b8e62d   (also verified present in main @ 10b4fb2a9)
CANN:           9.1.0
dtype:          bfloat16
head_dim:       128
rope_dim:       128 (full rope; partial rope not tested)

### 🐛 Describe the bug

`split_qkv_rmsnorm_rope` (the `vllm::qkv_rmsnorm_rope` fused op) returns **wrong and run-to-run non-deterministic** Q and K above a shape-dependent `num_tokens` threshold. V is always correct.

This is not a tolerance issue. Measured against a fp32 reference, `max|err|` reaches **~9.4** while the op's own test allows `atol=5e-2`, and it fails on **every one of 6 consecutive calls** with identical inputs, producing different results each time.

Present since the op was added in #4711 (Dec 2025), still in `main` as of `10b4fb2a9`.

### Root cause

In `vllm_ascend/ops/triton/linearnorm/split_qkv_rmsnorm_rope.py`, the RoPE position is staged into UB as a tile, then read back one element at a time as a scalar:

```python
x = tl.load(
    positions_gm_ptr + pos_indices + pos_offset, mask=(pos_indices + pos_offset) < input_batch_offset_end
)
...
for i in tl.range(batch_size_per_iter_per_vec):
    pos = get_element(x, (i,))                      # <-- reads the UB tile
    values_tmp3 = insert_slice(
        values_tmp3.reshape(batch_size_per_iter_per_vec, ROPE_DIM),
        tl.load(pos * ROPE_DIM + cos_sin_cache_offset[:, None]).reshape(1, ROPE_DIM),
        ...
    )
```

`get_element` reads the scalar out of the UB buffer that the *same* iteration's `tl.load` is still filling. Nothing orders the MTE transfer before the scalar read, so from the second outer iteration onward the scalar unit can still observe the **previous** iteration's position.

The result is that `cos`/`sin` are fetched for the wrong position and the entire token row — every Q head and every K head — is rotated by the wrong angle. V is unaffected because it skips RoPE, which matches the observed signature exactly.

This also explains the shape dependence. `x` is reloaded once per *outer* iteration, so the race needs `iter_num_per_vec > 1`. Since
[...]

PR #16699, opened 2026-09-16
[BugFix][Ops] Fix racy RoPE position read in `split_qkv_rmsnorm_rope`

Celeste-jq:
### What this PR does
Fixes issue #16674 
`split_qkv_rmsnorm_rope` produced wrong, run-to-run non-deterministic Q and K above a shape-dependent `num_tokens` threshold — `max|err|` up to **9.4** against the op's own `atol=5e-2`, failing on **6/6** consecutive calls with identical inputs. V was always correct.

### Root cause

The RoPE position was staged into UB as a tile, then read back one element at a time as a scalar:

```python
x = tl.load(positions_gm_ptr + pos_indices + pos_offset, mask=...)
...
pos = get_element(x, (i,))
```

`get_element` reads the scalar from the UB buffer that the same iteration's `tl.load` is still filling, with nothing ordering the MTE transfer first. From the second outer iteration onward the scalar unit can still observe the previous
iteration's position, so `cos`/`sin` are fetched for the wrong position and the whole row — every Q head and every K head — is rotated by the wrong angle. V skips RoPE, which is why only Q/K were affected.

Because `x` is reloaded once per outer iteration, the race requires `iter_num_per_vec > 1`. `batch_size_per_iter_per_vec` is sized from the UB budget, so a larger `q_hidden_size + kv_hidden_size` yields more iterations and failure at a smaller `num_tokens` — `(12, 1)` first fails at 1024, `(64, 4)` at 257. `num_tokens` 1 and 16 pass only because the single iteration reads a freshly-zeroed buffer.

### The fix

Read the position straight from global memory as a scalar. The UB round-trip, and with it the race, is gone.

```diff
-from vllm_ascend.ops.triton.triton_utils import extract_slice, get_element, get_vectorcore_num, insert_slice
+from vllm_ascend.ops.triton.triton_utils import extract_slice, get_vectorcore_num, insert_slice
@@
-    pos_indices = input_batch_offset + tl.arange(0, batch_size_per_iter_per_vec)
     output_q_nblk_idx = tl.arange(0, q_hidden_size)
     output_q_nmask = output_q_nblk_idx < q_hidden_size
     output_kv_nblk_idx = tl.arange(0, kv_hidden_size)
     output_kv_nmask = output_kv_nblk_idx < kv_hidden_size
     sin_cos_range = tl.arange(0, ROPE_DIM)
     cos_sin_cache_offset = cos_sin_cache_gm_ptr + sin_cos_range
+    last_row = input_batch_offset_end - 1
 
     for iter in tl.range(iter_num_per_vec):
         pos_offset = iter * batch_size_per_iter_per_vec
-        x = tl.load(
-            positions_gm_ptr + pos_indices + pos_offset, mask=(pos_indices + pos_offset) < input_batch_offset_end
-        )
         mmask = (mblk_idx + pos_offset) < input_batch_offset_end
@@
         for i in tl.range(batch_size_per_iter_per_vec):
-            pos = get_element(x, (i,))
+            pos = tl.load(positions_gm_ptr + min(input_batch_offset + pos_offset + i, last_row))
             values_tmp3 = insert_slice(
```

Three hunks total; `get_element` becomes unused in this file and is dropped from the import.

**On the clamp.** `last_row = input_batch_offset_end - 1` is always `>= 0`: for a core with no work, `input_batch_offset_end == batch_size >= 1`. Rows past the end read the last valid position instead of going out of bounds, and their outputs are discarded by `mmask` further down. This preserves the previous behaviour,
where the masked `tl.load` yielded `0` and those rows harmlessly read `cos_sin_cache[0]`.

**Why not a vectorized gather.** Replacing the inner loop with `tl.load(cos_sin_cache_gm_ptr + x[:, None] * ROPE_DIM + sin_cos_range[None, :])` would also fix it and be tidier, but limited vector-indexed gather support on
this backend is the likely reason the scalar loop exists at all. This patch keeps the existing structure and changes only the racy read.

### Validation

Ascend NPU, bf16, `head_dim=128`, full RoPE. Shapes and tolerance are taken from the op's own nightly test (`NUM_QKV_HEADS`, `NUM_TOKENS`, `DEFAULT_ATOL = 5e-2`), plus `num_tokens=257` from the PR gate.

**1. Non-determinism is gone.** Bitwise comparison of 24 repeated calls with identical inputs:

| num_tokens | (12,1) before | (12,1) after | (64,4) before | (64,4) after |
|---:|:---:|:---:|:---:|:---:|
| 1 | stable | stable | stable | stable |
| 16 | stable | stable | stable | stable |
| 257 | stable | stable | **UNSTABLE** | stable |
| 1024 | **UNSTABLE** | stable | **UNSTABLE** | stable |
| 10240 | **UNSTABLE** | stable | **UNSTABLE** | stable |

**2. Results are correct, not merely stable.** Per-row comparison against a fp32 reference at `atol=5e-2`, plus a count of rows matching the *preceding* token's position (the original failure mode):

`(12, 1)`:

| num_tokens | before max\|err\| Q / K | after max\|err\| Q / K | off-by-one rows after |
|---:|---:|---:|:---:|
| 1 | 0.0119 / 0.0077 | 0.0119 / 0.0077 | 0 |
| 16 | 0.0156 / 0.0143 | 0.0156 / 0.0143 | 0 |
| 257 | 0.0300 / 0.0283 | 0.0300 / 0.0283 | 0 |
| 1024 | **7.8431 / 7.6423** | **0.0310 / 0.0309** | 0 |

`(64, 4)`:

| num_tokens | before max\|err\| Q / K | after max\|err\| Q / K | off-by-one rows after |
|---:|---:|---:|:---:|
| 1 | 0.0148 / 0.0077 | 0.0148 / 0.0077 | 0 |
| 16 | 0.0306 / 0.0146 | 0.0306 / 0.0146 | 0 |
| 257 | **8.0298 / 6.8400** | **0.0311 / 0.0156** | 0 |
| 1024 | **9.3595 / 6.4640** | **0.0293 / 0.0306** | 0 |

Two things to note: on shapes that already passed, before and after are **identical**, so the patch does not perturb correct results; and residual error on the fixed shapes is ordinary bf16 rounding, comfortably inside `atol=5e-2`.

**3. No performance regression.** Measured over 2 repeats × 300 iterations (200 warmup) per cell, all 8 shape/config cells came out at or below baseline. I am deliberately not quoting a speedup: in this eager micro-benchmark
`num_tokens=1` and `num_tokens=1024` both land around 47 us, so it is dominated by dispatch overhead rather than kernel time, and run-to-run spread (~10%) is comparable to the effect. The patch removes one UB transfer per outer iteration, so a regression would be surprising, but a proper device-side kernel-time measurement would be needed to quantify any gain.

### Also: schedule the test that would have caught this

`tests/e2e/nightly/.../triton/test_split_qkv_rmsnorm_rope.py` already has a correct fp32 gold reference and covers the failing shapes. It has simply never been run — `git grep -c 'split_qkv_rmsnorm_rope' -- .github/` returns `0` on the
base branch.

Without this, the same class of bug regresses silently:

```diff
       - name: rejection-sample
         os: linux-aarch64-a2b3-1
         tests: tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_rejection_sample.py
+      - name: split-qkv-rmsnorm-rope
+        os: linux-aarch64-a2b3-1
+        tests: tests/e2e/nightly/single_node/ops/singlecard_ops/triton/test_split_qkv_rmsnorm_rope.py
```

Per the header comment in `nightly_config.yaml`, this can be exercised on this PR without merging via `/nightly split-qkv-rmsnorm-rope`. Happy to add the matching `a3:` entry, or to split the CI change into a separate PR if preferred.

Separately, the PR gate `test_graphex_qknorm_rope_fusion.py` runs at `num_tokens=257` — a failing shape for `(64, 4)` — but only asserts `expect_fused=True`. Adding a numeric assertion there would turn it into a real guard. Left out of this PR to keep the change focused.


Celeste-jq:
Hi @zzzzwwjj @realliujiaxu, could one of you please add the `ready-precise` label to trigger the selected tests for this source change? The existing `split_qkv_rmsnorm_rope` numerical test has also been added to the A2 nightly matrix in this PR. Thanks!

Celeste-jq:
@wangxiyuan @Yikun PTLA

Issue #16322, opened 2026-09-11
[RFC]: Make live weight updates transactional and graph-safe


Ronald1995:
## Background

vLLM Ascend is used as the rollout/inference backend by RL systems such as verl and slime. These systems update model weights repeatedly while the engine remains alive, often with ACLGraph enabled and without a sleep/wake cycle.

Today, **process_weights_after_loading()** is responsible for several different jobs: checkpoint-to-runtime layout conversion, quantization/repacking, derived-weight construction, backend buffer preparation, and memory disposal. A cold-load finalizer is therefore implicitly reused as a hot-reload mechanism, although many implementations are not idempotent and allocate new storage.

The current failures and audit findings share three missing invariants:

1. **Representation correctness before load:** every destination must be restored to the checkpoint/model representation understood by its weight loader.
2. **Value freshness after load:** every derived or runtime-only tensor must be refreshed from the new source weights in dependency order.
3. **Storage identity before graph replay:** graph-visible persistent tensors and backend bindings must continue to reference the storage captured by ACLGraph.

The observable risk patterns include:

- reassignment after .to(), .contiguous(), .clone(), or npu_format_cast(), leaving ACLGraph with stale pointers;
- checkpoint/runtime layout mismatch on the next reload (transpose, NZ conversion, reshape, flatten, or repack);
- stale derived weights such as weight_fp32, W_UK_T/W_UV, composed projections, scales, and biases;
- destructive source disposal or deletion, leaving no valid destination for the next checkpoint;
- non-idempotent PWAL tied to wake_up() even though asynchronous RL may never sleep;
- runtime-owned plain tensors, expert lists, static buffers, workspaces, and offloader/kernel aliases that generic layerwise reload cannot see;
- different semantics across HCCL and NPU IPC.

There are also two immediate gaps in the current NPU IPC path: start_weight_update() / finish_weight_update() are no-ops, and the packed receive branch builds weights but does not reach model.load_weights(weights).

Related work has already demonstrated parts of the solution: #13905 preserves MXFP8 runtime storage, #15666 addresses stale weight_fp32, #15463 reports destructive SFA source disposal, #5915 shows why restore cannot depend on wake-up, #15415 shows alias rebinding after NZ conversion, RFC #2851 proposes centralized model-loader behavior, and RFC #12073 proposes broader RL lifecycle coverage. This RFC connects those fixes under one reload contract instead of adding unrelated per-model exceptions.

## Goals

[...]

Issue #15463, opened 2026-09-01
[Bug]: `dispose_layer(kv_b_proj)` breaks live weight synchronization for SFA models


lxb007981:
### Your current environment

<details>
<summary>The output of `python collect_env.py`</summary>

```text
Collecting environment information...
PyTorch version: 2.10.0+cpu
Is debug build: False

OS: Ubuntu 22.04.5 LTS (aarch64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
Clang version: Could not collect
CMake version: version 4.4.2
Libc version: glibc-2.35

Python version: 3.12.13 (main, Aug  3 2026, 02:40:28) [GCC 11.4.0] (64-bit runtime)
Python platform: Linux-5.10.0-182.0.0.95.r2220_157.hce2.aarch64-aarch64-with-glibc2.35

CPU:
Architecture:                         aarch64
CPU op-mode(s):                       64-bit
Byte Order:                           Little Endian
CPU(s):                               192
On-line CPU(s) list:                  0-191
Vendor ID:                            HiSilicon
BIOS Vendor ID:                       HiSilicon
BIOS Model name:                      Kunpeng 920 5250Y
Model:                                0
Thread(s) per core:                   1
Core(s) per socket:                   48
Socket(s):                            4
Stepping:                             0x0
Frequency boost:                      disabled
CPU max MHz:                          2000.0000
CPU min MHz:                          1200.0000
BogoMIPS:                             200.00
Flags:                                fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 sm3 sm4 asimddp sha512 sve asimdfhm dit uscat ilrcpc flagm ssbs sb paca pacg dcpodp sve2 sveaes svepmull svebitperm svesha3 svesm4 flagm2 frint svei8mm svebf16 i8mm bf16 dgh rng ecv
L1d cache:                            12 MiB (192 instances)
L1i cache:                            24 MiB (192 instances)
L2 cache:                             192 MiB (192 instances)
L3 cache:                             336 MiB (8 instances)
NUMA node(s):                         8
NUMA node0 CPU(s):                    0-23
NUMA node1 CPU(s):                    24-47
NUMA node2 CPU(s):                    48-71
NUMA node3 CPU(s):                    72-95
NUMA node4 CPU(s):                    96-119
[...]

Issue #2851, opened 2025-09-10
[RFC]: Add NPUModelLoader


momo609:
### Motivation.

Motivation.
In the current vllm-ascend implementation, each layer requires reimplementing the process_weights_after_loading function to apply additional weight processing, resulting in scattered code. In RL scenarios, where inference weights need to be updated after each training iteration, the existing code structure requires overriding multiple weight-loader func to support dynamic weight updates. To improve the extensibility of weight-loader func, a new NPUModelLoader class is introduced, dedicated to centralizing and streamlining weightloader operations.

### Proposed Change.

Add a new NPUModelLoader class

### Feedback Period.

No response

### CC List.

No response

### Any Other Things.

Temporary Solution Code: https://github.com/vllm-project/vllm-ascend/pull/2756 @momo609

Nagisa125:
Hello, how is the current progress? Has the issue been resolved?

PR #16717, opened 2026-09-16
[BugFix][Attention] Refresh MLA/SFA runtime weights in place on reload

Ronald1995:

`AscendMLAImpl` and `AscendSFAImpl` absorb `kv_b_proj` into `W_UV` / `W_UK_T` in
`process_weights_after_loading`. In graph + RL scenario the graph is captured
once, so those two tensor addresses are baked into the captured graph and a
later weight-update reload must refresh the tensors **in place** instead of
rebinding the attributes.

That was hand-rolled (#7473) as:

```python
if not hasattr(self, "W_UV"):
    self.W_UV = W_UV.transpose(0, 1).contiguous()
    self.W_UK_T = W_UK.permute(1, 2, 0).contiguous()
else:
    self.W_UV.copy_(W_UV.transpose(0, 1).contiguous())
    self.W_UK_T.copy_(W_UK.permute(1, 2, 0).contiguous())
```

The `copy_` on the reload path is unconditional, so it raises as soon as the
derived value stops being byte-compatible with the stored one (shape, dtype or
device changed). It also duplicates a contract that upstream vLLM already owns:
`MLAAttention` refreshes exactly these two tensors with
`replace_parameter(self, "W_UV", ..., prefer_copy=True)`.

This PR switches both implementations to that same call. It does not change the
primary intent of #7473 (the address still stays stable across reloads) but it
gets the mismatch handling for free:

* compatible new value -> copied into the existing storage, `data_ptr()` unchanged;
* incompatible new value -> the attribute is rebound instead of raising.

`AscendMLAImpl` / `AscendSFAImpl` are plain `MLAAttentionImpl` objects rather
than `nn.Module`s, but that is not a blocker: `replace_parameter` only needs
`getattr`/`setattr` plus an optional `weight_loader` passthrough, and upstream
calls it on the very same kind of object. The attributes stay plain Python
attributes, so this does **not** register the derived tensors as module
parameters — which is deliberate, since registering them would break the
layerwise load accounting of vLLM's reload path.

### Relation to RFC #16322 (live weight updates transactional and graph-safe)

**Related to #16322** — this is a partial implementation of **Phase 1 — Graph-stable derived state**, item 2 ("Migrate `weight_fp32` (#15666) and SFA `W_UK_T`/`W_UV` first"). It does not close the RFC.

It implements state-ownership rule **§2 — Derived graph-visible state**. `W_UV` / `W_UK_T` are derived, graph-visible tensors, and the previous code was exactly the "model-specific `hasattr` + `copy_` variant" the RFC asks to replace:

| RFC §2 requirement | this PR |
| --- | --- |
| cold load: allocate the final runtime representation once | first call allocates and registers `W_UV`/`W_UK_T` |
| reload: derive into a temporary, then `copy_` into the existing storage | `prefer_copy=True` does the derive-then-copy, `data_ptr()` unchanged |

It also converges on upstream vLLM's own contract for these same two tensors (`MLAAttention`, same `replace_parameter(..., prefer_copy=True)`), so it is not a new Ascend-specific mechanism.

**Explicitly not covered here** (still open in the RFC):

* the RFC's *"one shared helper for this policy instead of model-specific `hasattr` + `copy_` variants"* — this converges the two MLA/SFA sites but adds no Ascend-owned helper, and does not validate **format** (e.g. the NZ rebind at `sfa_v1.py`'s `maybe_trans_nz(self.W_UK_T)`). Whether that helper should be owned centrally is RFC open question 1 (#2851), so it is left alone;
* the RFC's *"if existing storage is incompatible after graph capture, raise an actionable error rather than silently replacing it"* — `prefer_copy=True` instead **silently rebinds** on an incompatible shape/dtype/device. That is a live decision on RFC open question 2 (raise vs. explicit ACLGraph recapture). If consensus lands on *always raise*, this call site needs a fail-closed wrapper;
* Phase 1 item 3 (audit `310P weight_nz` and other composed projections). Candidates found while doing this work: the `maybe_trans_nz` rebinds at `mla_v1.py` / `sfa_v1.py`, and the `replace_parameter` calls missing `prefer_copy` in `quantization/methods/w8a8/fp8_block.py` and `ops/fused_moe/routed_experts.py`;
* Phase 1 item 4's "`data_ptr()` identity across **at least three** reloads after graph capture" — the tests here cover two calls / one reload, with `torch_npu` mocked;
* Phase 2 item 1 (destructive SFA `kv_b_proj` disposal, #15463). That is the complementary half: it keeps the *source* parameter alive, this keeps the *derived* storage identity. Both touch `AscendSFAImpl.process_weights_after_loading`, so one will need a trivial rebase over the other.

The RFC's full validation matrix (`cold A → capture → reload B → reload C → compare vs fresh cold C`, HCCL/NPU-IPC, packed/unpacked, eager/ACLGraph) was **not** run here; end-to-end NPU ACLGraph replay verification is left to CI and follow-up. See the companion RFC comment: https://github.com/vllm-project/vllm-ascend/issues/16322#issuecomment-5698320015

No. The tensor values, their layout and their addresses across reloads are
unchanged; only the failure mode for an incompatible reload changes from an
exception in `copy_` to a rebind.

Unit tests added (they run on CPU-only runners, `torch_npu` calls are mocked):

* `tests/ut/attention/test_mla_v1.py::TestAscendMLAImpl::test_process_weights_after_loading_keeps_runtime_weight_address`
* `tests/ut/attention/test_sfa_v1.py::TestAscendSFAImpl::test_process_weights_after_loading_keeps_runtime_weight_address`

Each test calls `process_weights_after_loading` twice, mutates `kv_b_proj.weight`
in between, and asserts that:

1. `W_UV.data_ptr()` / `W_UK_T.data_ptr()` are unchanged after the reload;
2. the buffers actually hold the re-derived `kv_b_proj` split (values follow the
   new weight).

```bash
pytest -sv tests/ut/attention/test_mla_v1.py -k keeps_runtime_weight_address
pytest -sv tests/ut/attention/test_sfa_v1.py -k keeps_runtime_weight_address
```

The existing `process_weights_after_loading` tests in both files are unaffected.

Not done: end-to-end verification on NPU hardware (RL weight-update graph replay)
was not possible in this environment — please rely on the PR CI for that.


Annarine (approved):

vllm_ascend/attention/mla_v1.py
```diff
@@ -786,6 +787,9 @@ class AscendMLAImpl(MLAAttentionImpl):
     understand this class
     """
 
+    W_UV: torch.Tensor
+    W_UK_T: torch.Tensor
```
lcfenglinwan, line 808: this can be removed

vllm_ascend/attention/sfa_v1.py
```diff
@@ -658,6 +659,9 @@ class AscendSFAImpl(MLAAttentionImpl):
     understand this class
     """
 
+    W_UV: torch.Tensor
+    W_UK_T: torch.Tensor
```
lcfenglinwan, line 750: same as above

Ronald1995:
## Verification on real Ascend 910B — rebase onto latest `main` + review follow-up

Rebased onto `main` `5591facb3` (clean, no conflicts) and pushed as `b43226c59`. Two changes on top of the original patch:

- dropped the `W_UV: torch.Tensor` / `W_UK_T: torch.Tensor` class annotations as asked above (they were introduced by this PR, so removing them is a no-op return to the pre-existing state — `replace_parameter` only needs `getattr`/`setattr`);
- added `test_process_weights_after_loading_rebinds_incompatible_value` to `tests/ut/attention/test_mla_v1.py` and `tests/ut/attention/test_sfa_v1.py`, because the two address tests as pushed cannot observe this PR's behaviour change (see §2).

### 1. Equivalence with the hand-rolled `hasattr`/`copy_`: holds on every reachable path

`old.shape/dtype/device == new` *is* `replace_parameter`'s `prefer_copy` predicate, so each path can be compared exactly:

| path | old | new | verdict |
| --- | --- | --- | --- |
| first call (no attribute) | bare `Tensor` backed by the freshly derived `.contiguous()` | `nn.Parameter(..., requires_grad=False)` backed by **the same** storage | same shape/dtype/device/`data_ptr()`; only the Python type differs |
| later call, compatible value — **the only reload a serving session can produce** | in-place `copy_`, `data_ptr()` kept, values refreshed | `prefer_copy` branch: in-place `copy_`, `data_ptr()` kept, values refreshed | **bit-for-bit identical** |
| later call, incompatible value | `RuntimeError` from `copy_` (or a silent broadcast when the shapes happen to broadcast) | rebinds to a new `nn.Parameter` | **deliberately different** — the stated purpose of the PR |

The incompatible path is unreachable in practice: the derived shapes are a pure function of `kv_lora_rank`, `num_heads`/`local_num_heads`, `qk_nope_head_dim`, `v_head_dim`, `act_dtype` and device — all fixed for the lifetime of the impl — and every reload splits the same `kv_b_proj`. So the compatible branch is always taken.

The `nn.Parameter` difference at cold load is not a concern. `AscendMLAImpl`/`AscendSFAImpl` derive from `MLAAttentionImpl`, which is **not** an `nn.Module` (verified), so nothing lands in `_parameters`; `state_dict`, `named_parameters()` and the layerwise load accounting are untouched. `Parameter(requires_grad=False)` is a `Tensor` subclass sharing the original storage, so bmm / `npu_transpose_batchmatmul` / `maybe_trans_nz` / `F.pad` are unaffected. Upstream `MLAAttention` ends up with a `Parameter` in the same place, so this also converges the runtime representation.

Pre-existing and unchanged here: `maybe_trans_nz(self.W_UK_T)` (`mla_v1.py:1110`, `sfa_v1.py:882`, NATIVE branch) rebinds `W_UK_T` when NZ applies, which would defeat the stable-address guarantee. Both this e2e and RL run with `weight_nz_mode=0`, so it does not trigger; it belongs to RFC #16322 Phase 1 item 3.

### 2. The unit tests as pushed could not see this change

`test_process_weights_after_loading_keeps_runtime_weight_address` asserts that `data_ptr()` is unchanged and that the buffers hold the re-derived split. **The old implementation satisfies both** (its `else` branch `copy_`s in place), so the test passes before and after the change and cannot fail if the switch to `replace_parameter` is reverted.

Mutation check — implementation reverted to the base commit, tests kept:

| test | old impl | new impl |
| --- | --- | --- |
| `..._keeps_runtime_weight_address` (MLA) | PASS | PASS |
| `..._rebinds_incompatible_value` (MLA) | FAIL — `RuntimeError: The size of tensor a (256) must match the size of tensor b (128) at non-singleton dimension 0` @ `mla_v1.py:1068` | PASS |
| `..._keeps_runtime_weight_address` (SFA) | PASS | PASS |
| `..._rebinds_incompatible_value` (SFA) | FAIL — same @ `sfa_v1.py:934` | PASS |

The new test is therefore the only one that pins the actual behavioural delta; the address test remains a regression guard for #7473's contract.

### 3. Unit tests + lint (910B container)

```
pytest -q tests/ut/attention/test_mla_v1.py -k process_weights_after_loading   -> 8 passed
pytest -q tests/ut/attention/test_sfa_v1.py -k process_weights_after_loading   -> 3 passed
ruff check / ruff format --check (all 4 files)                                 -> clean
```

### 4. End-to-end: the GLM case of the NPU IPC weight-transfer transaction

`tests/e2e/pull_request/one_card/rlhf/state_transitions/test_npu_ipc_weight_transfer.py`, case `glm-5.1-sfa-derived-kv` (`zai-org/GLM-5.1`, 4 layers / 8 experts), both packed modes. This is exactly the target scenario: `cudagraph_mode=FULL_DECODE_ONLY` with `cudagraph_capture_sizes=[1]` and `enforce_eager=False` (graph captured once), `weight_nz_mode=0` (so `maybe_trans_nz` is the identity and the stable address genuinely matters), and the case's own comment already lists SFA's `W_UK_T`/`W_UV` among the derived state under test.

```
[unpacked-glm-5.1-sfa-derived-kv] PASSED
[packed-glm-5.1-sfa-derived-kv]   PASSED
2 passed in 162.19s
```

Each lane completed both rounds of the transaction — `POST /start_weight_update` → `/update_weights` → `/finish_weight_update`, all `200 OK` — with the reference server loading the identical payload at normal startup, and the second round exercising the layerwise reload after graph capture.

**No regression:** the same two lanes were then run with the implementation reverted to the base commit — `2 passed in 168.42s`. Both implementations pass.

**Caveat, stated plainly:** this e2e reload keeps the shapes identical, i.e. it exercises the *compatible* branch, which the old `copy_` also handled. It therefore demonstrates equivalence / no regression on the real graph + RL path; it does **not** demonstrate that the fix is required. Only the `rebinds_incompatible_value` unit tests cover the delta.

Two environment notes for anyone reproducing this (unrelated to the patch):

1. the container has no egress to huggingface.co, so `HF_HUB_OFFLINE=1` must be exported, otherwise the reference server dies on `httpx.ConnectError: [Errno 101] Network is unreachable` while resolving the tokenizer;
2. NPU IPC requires trainer and worker on the *same physical* NPU. This test hands the server `ASCEND_RT_VISIBLE_DEVICES=<VLLM_RL_TEST_DEVICE_INDEX>` as an absolute chip id while also calling `torch.npu.set_device(<same index>)` in the pytest process, so that process must see logical == physical. With an externally restricted `ASCEND_RT_VISIBLE_DEVICES`, the trainer and the server land on different chips and the transfer fails with `IPC handle not found for NPU UUID ...-0`.

### 5. One inaccuracy in the PR description

The description says "upstream calls it on the very same kind of object". Upstream calls `replace_parameter(..., prefer_copy=True)` from `MLAAttention.process_weights_after_loading`, and `MLAAttention` **is** an `nn.Module`; there is no `replace_parameter` call site anywhere under `vllm/v1/attention/`, i.e. no upstream MLA *impl* uses it. The substantive half of the argument — that the helper only needs `getattr`/`setattr` plus the optional `weight_loader` passthrough, so a non-`Module` is fine — does hold, and is what the tests above confirm. Worth rewording that sentence.

vllm_ascend/attention/mla_v1.py
Ronald1995: Sorry, I have to put these back — `replace_parameter` is `setattr`-based, so static analysis cannot see the assignment and the `mypy` step of `pre-commit` fails without the declarations:

```
vllm_ascend/attention/sfa_v1.py:958: error: Cannot determine type of "W_UK_T"  [has-type]
vllm_ascend/attention/mla_v1.py:1022: error: Cannot determine type of "W_UK_T"  [has-type]
vllm_ascend/attention/mla_v1.py:1082: error: Cannot determine type of "W_UK_T"  [has-type]
vllm_ascend/attention/mla_v1.py:1112: error: Cannot determine type of "W_UK_T"  [has-type]
Found 4 errors in 2 files (checked 643 source files)
```

That is exactly what the `pre-commit` job reported after I removed them. `W_UK_T: torch.Tensor` is strictly required; I kept `W_UV: torch.Tensor` alongside it because both attributes are injected the same way and dropping only one would leave the `torch.bmm` / `npu_transpose_batchmatmul` consumers of `W_UV` unchecked.

Pushed with a short comment above the declarations recording why they cannot be deleted. Re-verified locally with `mypy --follow-imports skip --check-untyped-defs --python-version {3.10,3.11,3.12}`: zero `has-type` errors.

vllm_ascend/attention/sfa_v1.py
Ronald1995: Same as the other thread — these have to stay. `replace_parameter` writes the attribute with `setattr`, so without the declaration `mypy` in `pre-commit` reports `vllm_ascend/attention/sfa_v1.py:958: error: Cannot determine type of "W_UK_T"  [has-type]`. I added a comment above them explaining why, so they do not get removed again.

PR #16718, opened 2026-09-16
[Feature][CI] Add persistent incremental csrc build cache

AuroraEmiya:

vLLM Ascend currently has an exact csrc artifact cache. It is very fast on an exact match, but any tracked csrc change invalidates the whole artifact and forces the native source build to start from scratch.

This PR adds a persistent, fine-grained L1 cache for csrc compilation while keeping the existing exact artifact cache as L0:

```text
L0 exact final artifact
        |
        | miss
        v
L1 persistent per-action cache
        |
        v
ordinary source build
```

L0 remains the only cache layer that can skip the whole native build. On an L0 miss, L1 restores a compatible set of candidate entries, then the normal source build runs. Each compiler action independently recomputes its identity and either reuses a verified entry or executes the original compiler command.

The source build therefore remains authoritative: an L1 miss, unavailable remote cache, or unsupported source snapshot falls back to the ordinary build.

### Cache identity and correctness

Each cached action is keyed by three semantic identities:

```text
prepared_input_hash
recipe_hash
compiler_environment_hash
```

The local cache engine in `csrc/scripts/build_cache.py` validates the action key, schema, artifact model, manifest, and artifact contents before accepting a HIT.

Prepared inputs cover generated operator sources, shared kernel sources, recipe files, and shared compatibility headers such as `cann_compat.h`. Explicit checkout/build roots may be normalized for UTF-8 text so equivalent workspaces can share entries, while binary content, external paths, symlink identity/targets, and semantic file contents remain sensitive.

The cache also preserves the build's concurrency and publication invariants:

- atomic local entry save;
- entry/action/publish locks with separate failure semantics;
- private staging for custom operators;
- artifact ownership tracking and collision detection;
- safe manifest/path validation before restored artifacts are used.

The persistent outer snapshot is only a transport/candidate-selection layer. It is not the correctness authority; inner action validation always decides whether an entry may be reused.

See the [design document](docs/source/developer_guide/Design_Documents/persistent_csrc_build_cache.md) for the full contract.

### Persistent L1 transport

CI persists the local action cache through `runs-on/cache` using the existing Huawei OBS S3-compatible backend.

`prepare_csrc_l1_restore.py` is the shared key-preparation helper. It:

- detects whether the tested source contains the cache engine;
- derives the tracked-csrc identity;
- builds the compatibility/snapshot keys; and
- exports the local cache and telemetry paths.

Persistent callers use workflow-owned same-repository actions:

```text
$/.github/actions/csrc-l1-restore
$/.github/actions/csrc-l1-save
```

This intentionally separates workflow implementation from the source being tested. A caller may build a PR, fork, external repository, or historical ref without requiring that source snapshot to carry the current CI action implementation.

The restore action resets the `.updated` marker. The local engine creates that marker only when a new/replaced entry is successfully saved. Trusted writers therefore publish an updated shared snapshot only when the local L1 actually changed; an all-HIT build does not republish the same cache.

Historical source that predates `csrc/scripts/build_cache.py` is explicitly unsupported by L1:

```text
supported=false
→ skip L1 export/save
→ continue the original wheel/image/source build unchanged
```

No cache engine is injected or backported into old source snapshots.

### Trust and failure model

Shared L1 publication is limited to trusted producer paths, including the central csrc producer, image builds, and release wheel builds. PR-triggered nightly, selected tests, doctest/source consumers, and other consumer paths restore only.

The failure boundary is intentional:

```text
cache availability / performance failure
→ fail open to ordinary source compilation

integrity / correctness violation
→ fail safely
```

Examples of fail-open cases include remote MISS, unavailable OBS transport, missing writer credentials, unsupported historical source, or an unavailable optional snapshot.

Unsafe manifests, artifact ownership collisions, or synchronization failures that would make publication correctness ambiguous are not treated as cache misses.

No model-serving API or runtime behavior changes.

This PR only changes native csrc build/CI caching and related diagnostics.

The implementation was validated at three levels: the local cache engine, isolated CI workflow/component paths in the CI development repository, and formal vLLM Ascend CI.

#### 1. Cache engine and static correctness

- [CI-dev cache review validation — run 35215229554](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35215229554): cache-engine/concurrency tests, targeted mypy, production workflow YAML parsing, credential-less coordinator contract, and whitespace checks all passed.
- The cache suite covers content-addressed HIT/MISS behavior, cross-root reuse, semantic invalidation, schema/artifact validation, unsafe manifest/path rejection, symlink semantics, atomic save, and concurrency/ownership behavior.

The CI-dev repository is an isolated validation repository; access may be required. These runs validate production-equivalent snapshots/components and are not claimed to be final-head GitHub checks.

#### 2. Real source-build cold/warm reuse

Two real source-build environments were exercised through the persistent L1 workflow.

- `SOC_VERSION=ascend910_9391`
  - [cold run 34972671558](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/34972671558): 1 third-party + 248 custom-operator MISSes, followed by successful local saves.
  - [warm run 34974769880](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/34974769880): 1 third-party + 248 custom-operator HITs.
- `SOC_VERSION=ascend910b1`
  - [cold run 34972941408](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/34972941408): 1 third-party + 233 custom-operator MISSes, followed by successful local saves.
  - [warm run 34978189200](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/34978189200): 1 third-party + 233 custom-operator HITs.

These runs exercise real CMake/source compilation rather than synthetic cache-only actions.

#### 3. CI workflow/component integration

Representative successful CI-dev workflow validations:

- [central producer — run 35044413981](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35044413981): restore L1, install/build vllm-ascend, verify csrc artifacts, summarize the incremental cache, save L1, and save the exact csrc artifact.
- [A2 wheel path — run 35044420071](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35044420071): restore isolated wheel L1, build the wheel image, export the updated cache from the container, and save the snapshot.
- [A2 image path — run 35045094174](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35045094174): restore image-build L1, build a real Docker image, export the updated container cache, and save it.
- [doctest/source-install path — run 35084117212](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35084117212): prepare/restore L1, run the real source-install doctest path, collect cache telemetry, inspect the cache, and save the updated snapshot.

These component runs are intentionally listed here because they exercise the production workflow shapes independently, including the host/container cache-export boundary.

#### 4. Persistent OBS transport and compatibility fallback

- [TEST OBS transport validation — run 35215229511](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35215229511):
  - a producer created a real cold action entry and published an L1 snapshot;
  - a same-csrc consumer restored the snapshot and obtained an inner HIT;
  - a compatible historical-csrc consumer restored through the compatibility fallback and obtained an inner HIT; and
  - an incompatible environment remained cold.

This validates the persistent restore hierarchy and the separation between outer snapshot compatibility and inner action correctness.

#### 5. Formal vLLM Ascend CI

- [formal E2E run 35493648769](https://github.com/vllm-project/vllm-ascend/actions/runs/35493648769) exercised `$/.github/actions/csrc-l1-restore` on the repository's Kubernetes-backed selected-test runner topology. The same-repository action resolved and executed successfully, the incremental restore step completed, and downstream vllm-ascend source installation remained healthy.

This is important because the workflow-owned `$/` action reference is the production path used to avoid coupling action resolution to the tested checkout/container workspace.

The final historical-source fail-open behavior is also guarded at workflow level: when key preparation reports `supported=false`, Docker wheel/image builds remain reachable while L1 export and save steps are skipped.

### vLLM reference


AuroraEmiya:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- PR Create

**Failed:**
- Doc Test: still in progress, retry /rerun after completion

AuroraEmiya:
[code-review] High — OBS transport contract is inconsistent across the new L1 paths.

`runs-on/cache@v5` is used for L1 restore/save throughout the PR, but the Huawei OBS backend is only explicitly configured in the release-wheel jobs (`RUNS_ON_S3_BUCKET_CACHE`, `RUNS_ON_S3_BUCKET_ENDPOINT`, region, and `HW_OBS_AK/HW_OBS_SK`). The central producer (`_build_csrc_cache.yaml`), selected-tests fallback, image L1 path, and other composite-action consumers do not show the same OBS backend contract.

Why this matters: generic `runs-on/cache` persistence and OBS-backed persistence are not the same guarantee. Unless the relevant runner classes inject exactly the same OBS backend implicitly, L1 snapshots can be fragmented across different backends, and existing cold/warm validation does not prove that production L1 is actually stored in OBS. This is a production transport/integration issue, not a cache-engine correctness issue.

Please audit every L0/L1 `runs-on/cache` restore/save site and record its effective backend, credential source, trust context, and fork behavior. Then define one minimal transport contract: trusted upstream workflows may restore+save to the shared OBS namespace; untrusted/fork PR contexts should be restore-only (do not rely on a shared save merely failing). Keep OBS configuration at the workflow/job transport layer rather than in `build_cache.py` or the composite cache semantics.

Smallest recommended validation: add a TEST-ONLY OBS transport smoke that writes a unique isolated key from one trusted job and restores it from a second independent job/runner, then verify a read-only consumer path separately if the infrastructure allows it. No A2/A3/310P full matrix rerun or cache-engine change should be required.

AuroraEmiya:
[code-review] Medium — CI/review gate, not a product failure: the current `E2E` run is red only because the PR changes `src`/`ci_pipeline` paths but has no run-enabling test label.

Run: https://github.com/vllm-project/vllm-ascend/actions/runs/35101652558

`pre-commit` and `cpu-ut` both passed. The failing `ci-gate` reports `select-tests: skipped`, with `main2main=false`, `ready-precise=false`, `ready-all=false`, and explicitly requires a maintainer to add exactly one of `ready-precise` (recommended), `ready-all`, or `main2main` before merge. Current PR labels are only `documentation`, `ci/build`, and `module:tests`.

Why this matters: `/rerun` alone will reproduce the same gate failure until a run-enabling label is added; this should not be debugged as a cache regression.

Smallest next action: ask a maintainer to add `ready-precise` unless broader coverage is intentionally desired, then inspect the resulting selected-test jobs separately for any real cache/workflow failures.

AuroraEmiya:
[code-review] High — base drift now touches the same production workflow this PR modifies: `.github/workflows/schedule_main2main.yaml`.

`main` advanced from the PR base `1c15fc1a...` to `7949403b...` via #16724, which switches Main2Main submission from `vllm-ascend-ci/vllm-ascend` + `PAT_TOKEN` / `main2main-bot` to `dev-submitter/vllm-ascend` + `DEV_SUBMITTER_TOKEN` / `dev-submitter`. The PR head is still `75d3d8a75...` and its `schedule_main2main.yaml` still contains the old `HEAD_FORK`, `PAT_TOKEN`, and bot identity while also adding L1 restore/save and helper-checkout logic to that same workflow.

Why this matters: this is a relevant upstream change in a credential/trust-sensitive production caller. GitHub currently reports the PR mergeable, so this is not a textual-conflict report; the risk is that the reviewed/tested head is no longer the production merge result, and the Main2Main L1 transport/fork behavior has not been audited against the new `dev-submitter` trust model. This is a product/integration review issue, not an infra flake.

Smallest next action: rebase/replay the three PR commits onto current `main` (`7949403b...`), verify the merged Main2Main workflow retains the new `dev-submitter` credentials/HEAD_FORK semantics plus the L1 changes, then rerun only the low-cost workflow gates (`pre-commit`/actionlint/YAML/static checks and diff scan). No A2/A3/310P matrix rerun is implied unless the rebase changes cache/build semantics.

AuroraEmiya:
[code-review] High — new base drift now also overlaps the image-build workflow modified by this PR: `.github/workflows/_schedule_image_build.yaml`.

Since the previous review, `main` advanced to `52ce43b15172fce79297ae4f20612fec3a476f3f`. Commit `3d901a5557617475de0264f6be5685bc941f923e` / #15611 changes the same reusable image workflow this PR augments with L1 restore/export/save. Upstream adds the `SYNC_TOOLS_TOKEN` secret interface and in-workflow quay sync dispatch, and also changes the Docker build-args block (`RUSTUP_UPDATE_ROOT` plus `CRATES_IO_INDEX`). The PR head is still `75d3d8a75...` on base `1c15fc1a...`, so its image-L1 logic has only been reviewed/validated against the pre-#15611 image workflow.

Why this matters: this is a product/integration review issue, not an infra flake. Even if GitHub can merge the text cleanly, the effective production image job after merge combines two independently changed control paths: the new SWR/sync credential flow and this PR's Docker L1 restore/export/save path. In particular, the caller/reusable-workflow secret contract and the `Build and push`/post-build ordering need to be checked together.

Smallest next action: rebase/replay onto current `main` (`52ce43b...`), preserve #15611's `SYNC_TOOLS_TOKEN`/SWR/sync semantics and its updated build args, then re-audit only `_schedule_image_build.yaml` + its callers and run actionlint/YAML/pre-commit/static diff checks. No A2/A3/310P matrix rerun is implied unless the rebase changes cache/build semantics.

AuroraEmiya:
[code-review] High — L1 snapshot key construction is split across incompatible producer/consumer implementations, which fragments the persistent cache namespace even when build environments are compatible.

Exact locations: `.github/workflows/_build_csrc_cache.yaml` (`Build incremental cache key`) builds a literal prefix from `TARGET_ID/IMAGE_TAG/SOC_VERSION`, while `.github/actions/csrc-l1-restore/action.yaml` hashes a caller-provided `compatibility-id`. Consumers then use workflow-role-specific IDs such as `selected-*`, `upstream-*`, `doctest-source-*`, `image-*`, and `main2main-*`. These prefixes cannot match the central producer's prefix, so a snapshot published by the producer is not discoverable by those consumers even if the inner action cache would validate it safely. With PR/external-source runners being OBS read-only, this becomes a functional integration problem: those jobs cannot seed their own shared namespace and may miss the trusted writer's snapshots entirely.

Classification: product/integration + maintainability.

Smallest next action: make one helper/action own persistent L1 key construction and derive the compatibility descriptor from actual build compatibility (schema + architecture/SOC + image/toolchain), not workflow role. Reuse that implementation in the central producer and consumers, then validate one producer-save -> independent consumer-restore path. Simplification opportunity: this should also delete the duplicate inline schema/key Python+shell block in `_build_csrc_cache.yaml` (roughly 40–50 lines) instead of adding another mapping layer.

Related lifecycle cleanup: `.github/actions/csrc-l1-save/action.yaml` currently saves any non-empty restored directory under a new unique primary key, so a trusted all-HIT run can republish an effectively unchanged snapshot. Once the key contract is centralized, gate remote publication on the build having produced at least one new `SAVED` entry; transport save remains best-effort.

AuroraEmiya:
[code-review] High — the new unified snapshot key can now merge different **actual compiler OS/image environments**, and the image-build path fingerprints the outer runner container rather than the Docker image that performs the csrc build.

Exact locations: `.github/actions/csrc-l1-restore/action.yaml` builds `compat_digest` from architecture + SOC + `/usr/local/Ascend/.../ascend_toolkit_install.info` whenever that metadata exists; `_schedule_image_build.yaml` invokes this action in its outer CANN Ubuntu container, while compilation happens later inside `docker/build-push-action` and may use a different Dockerfile/base OS. `csrc_cache_targets.json` also has Ubuntu/openEuler target pairs with the same architecture/SOC/CANN version, so those producers can land in the same persistent prefix when their CANN metadata matches. The inner `host-cxx` identity in `csrc/scripts/build_cache.py::_hash_compiler_environment()` records only `platform.system()`/machine plus compiler version strings; it does not fingerprint distro/libc/sysroot/system headers, so a restored third-party entry (e.g. protobuf/protoc) is not guaranteed to MISS across Ubuntu/openEuler when tool versions happen to match.

Why this matters: the previous namespace fragmentation is fixed, but the new compatibility domain is now broader than the proven semantic identity. This can move from a performance issue to false-HIT/ABI risk for host-built third-party artifacts. Classification: product/correctness + maintainability.

Smallest next action: make the persistent descriptor represent the **actual build environment**. For nested Docker builds, use the exact compiler/base image identity (and make it take precedence over outer-container CANN metadata); for normal container builds, include an OS/base-image discriminator unless the inner `host-cxx` environment hash is strengthened to cover distro/libc/sysroot. Add one Ubuntu-vs-openEuler key/inner-identity regression test before claiming cross-image compatibility.

Simplification opportunity: move the compatibility descriptor logic out of the now ~200-line composite YAML into one small tested helper/script; that makes the outer-vs-inner environment choice explicit and removes the large inline Python alias/fingerprint block instead of adding more YAML conditionals.

AuroraEmiya:
[code-review] High — NEW failure-containment regression after the latest simplification, plus one CI blocker and one leftover trust/plumbing cleanup.

1. **High — `.github/actions/csrc-l1-restore/action.yaml` / `csrc/scripts/build_cache.py::snapshot-key`**: `Prepare L1 key` runs under `set -euo pipefail`, and `main()` dispatches `snapshot-key` outside the cache engine's fail-open handling. `_snapshot_compatibility()` can raise on missing/unsupported architecture/SOC/toolchain metadata, and `_sha256_file()` can raise if `ascend_toolkit_install.info` becomes unreadable. Any of those now abort the composite action before ordinary compilation, even though L1 restore is an optional optimization. This is distinct from false-HIT safety: an unusable identity should mean **no restore / ordinary build**, not a weaker identity and not a failed build. Classification: product/integration + failure containment. Smallest next action: make key-preparation failure explicitly return `supported=false` (or equivalent restore-skip) with a warning, while keeping invalid/unreadable identity data ineligible for HIT; add one regression where snapshot-key cannot fingerprint the toolchain and the build path still proceeds uncached.

2. **Medium — current E2E run `35180879765`, `pre-commit` -> `Run mypy`**: the new shared test helper refactor is the only current CI blocker. Mypy reports 4 errors in `tests/ut/_tools/build_cache_test_utils.py:72` and `tests/ut/_tools/test_build_cache.py:141,1596`; lint/actionlint/shellcheck all pass, and downstream tests are skipped because pre-commit failed. Classification: review-only/CI. Smallest next action: type the mixed optional custom-option mapping (or avoid the heterogeneous dict), give `custom_options` a precise TypedDict/explicit kwargs shape, and narrow `shutil.which("cmake")` before constructing the subprocess argv; rerun only pre-commit/mypy first.

3. **Low — maintainability / least privilege**: after consumer-side L1 saves were removed, write-secret plumbing remains in restore-only workflows. `_selected_tests.yaml` still declares `HW_OBS_AK/HW_OBS_SK`, while `schedule_e2e_test.yaml` and `schedule_test_coverage.yaml` still pass them; `_e2e_nightly_single_node.yaml`, `_560t.yaml`, and `_models.yaml` also retain unused HW_OBS secret declarations. These no longer contribute to cache publication and unnecessarily widen the reusable-workflow secret interface. Smallest next action: delete the unused declarations/pass-throughs and keep write credentials only on actual writers (`_ensure/_build`, image, wheel as applicable). Simplification opportunity: this removes roughly 15–20 lines of dead YAML and makes the read-only consumer trust boundary visible from the workflow signatures.

AuroraEmiya:
[code-review] High/Medium — two NEW OBS/trust findings after the latest fixes.

1. **High — `.github/workflows/schedule_image_build_and_push.yaml` → `_schedule_image_build.yaml`** — **classification: product/trust-boundary.** The PR-triggered `image-build` path still passes `HW_OBS_AK/HW_OBS_SK` into the reusable image workflow, and `Save incremental cache for image build` is gated only by L0 miss; it is not gated on `github.event_name != 'pull_request'`. Fork PRs lose secrets automatically, but a same-repository PR can receive repository secrets, so an unmerged PR image build can publish into the shared production L1 namespace. This also contradicts the PR description's claim that PR callers do not receive write secrets. Smallest next action: make PR image builds restore-only by withholding `HW_OBS_*` for `pull_request` or adding an explicit non-PR/trusted-writer condition to the L1 save step; keep push/tag/approved dispatch writers unchanged.

2. **Medium — PR description / validation evidence** — **classification: review-only.** The OBS evidence is now stale: validation-fork run `35197429895` (`PR16718 OBS transport validation`) succeeded with a producer save and an independent consumer restore using the CI fork's dedicated Huawei OBS bucket/endpoint and credentials, via the production restore/save actions with only the TEST backend substituted. The description still reports the dedicated OBS round-trip as `BLOCKED-BY-INFRA` because the fork lacks production `HW_OBS_AK/HW_OBS_SK`. Smallest next action: update the evidence wording to `TEST OBS transport write/independent restore = VALIDATED`, while separately stating that the production `ascend-ci-cache-hk` credentials/IAM were not runtime-tested. No production-code change is needed for this item.

AuroraEmiya:
[code-review] Medium — the new explicit OBS write boundary leaves the PR cache coordinator doing full CPU builds that can no longer publish anything.

**Exact path:** `pr_test.yaml::prepare-csrc-cache` → `.github/workflows/_ensure_csrc_cache.yaml::build-missing-caches` → `.github/workflows/_build_csrc_cache.yaml`.

`pr_test.yaml` invokes `_ensure_csrc_cache.yaml` without passing `HW_OBS_AK/HW_OBS_SK`. This PR changed `_build_csrc_cache.yaml` so both shared publications are credential-gated: `csrc-l1-save` skips without the two inputs, and the final exact L0 `runs-on/cache/save` now has an `if` requiring both credentials. `_ensure_csrc_cache.yaml` still launches the missing-target CPU build matrix anyway. Therefore on a PR L0 miss it can spend the full source-build cost, verify the artifacts, then publish neither L1 nor L0; the later NPU selected-test job sees the same L0 miss and compiles again. The standalone trusted producer does pass the secrets, so this is specifically a credential-less coordinator regression introduced by the trust-boundary change.

**Why it matters:** this adds potentially large duplicate CI work on exactly the cold/miss path the cache design is meant to accelerate, while producing no reusable state. **Classification:** product/CI-performance + maintainability (not a correctness/security failure).

**Smallest next action:** keep PR consumers restore-only and make `_ensure_csrc_cache.yaml` lookup-only when writer credentials are absent; do not launch `build-missing-caches` in that case. Missing PR caches should fall through directly to the existing NPU source build, while cache seeding remains in `push_build_csrc_cache.yaml`/other trusted writers. **Simplification opportunity:** deleting/skipping the credential-less CPU build path removes an entire pointless matrix execution rather than adding more secret plumbing or passing production write credentials into PR workflows.

csrc/scripts/build_cache.py
```diff
@@ -0,0 +1,1725 @@
...
+    after = _snapshot(root)
+    selected = {relative for relative, signature in after.items() if before.get(relative) != signature}
+    return sorted(_expand_internal_symlink_targets(root, selected))
+
+
+def _safe_component(value: str) -> str:
+    safe = "".join(character if character.isalnum() or character in "._-" else "_" for character in value)
+    if not safe:
+        raise ValueError(f"invalid empty cache path component from {value!r}")
+    return safe
```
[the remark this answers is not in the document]
AuroraEmiya: Fixed in [f609248](https://github.com/vllm-project/vllm-ascend/commit/f609248acad333dbee2d0324b56490ff2aaa3752): _safe_component now explicitly rejects both dot segments, as well as empty sanitized components. Regression: test_cache_path_components_reject_dot_segments covers this traversal boundary.

Current verification: [cache-engine + concurrency suite, 41 passed](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35215229554). This is a private validation repository (access may be required); validation SHA 79612d988f6a41225e98c927a94f6a21882eac0a has production files identical to PR head 6101f59ea291e06ac6d0dfb1ca36b48cb271b2cd.

csrc/scripts/build_cache.py
```diff
@@ -0,0 +1,1725 @@
...
+        ):
+            relative = child.relative_to(path).as_posix()
+            if _is_excluded(relative, excludes):
+                continue
+
+            logical_path = f"{label}/{relative}"
+            if child.is_symlink():
+                target = os.readlink(child)
+                records.append((logical_path, f"symlink:{target}"))
+                manifest.append({"path": logical_path, "kind": "symlink", "target": target})
```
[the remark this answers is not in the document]
AuroraEmiya: Fixed in [f609248](https://github.com/vllm-project/vllm-ascend/commit/f609248acad333dbee2d0324b56490ff2aaa3752): prepared-input symlinks retain their link identity and also hash resolved target content, including nested directory targets. Resolution is strict: unreadable semantic inputs are not silently omitted. Regressions: test_nested_prepared_symlink_tracks_target_content and test_top_level_prepared_input_symlink_tracks_link_and_target_content verify target-content invalidation.

Current verification: [cache-engine + concurrency suite, 41 passed](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35215229554). This is a private validation repository (access may be required); validation SHA 79612d988f6a41225e98c927a94f6a21882eac0a has production files identical to PR head 6101f59ea291e06ac6d0dfb1ca36b48cb271b2cd.

csrc/scripts/build_cache.py
```diff
@@ -0,0 +1,1725 @@
...
+            normalized = _normalize_text(text, normalize_paths)
+            return _sha256_bytes(normalized.encode("utf-8"))
+    return _sha256_bytes(data)
+
+
+def _git_tracked_files(source_dir: Path, repo_root: Path) -> list[Path] | None:
+    try:
+        relative = source_dir.resolve().relative_to(repo_root.resolve())
+    except ValueError:
+        return None
```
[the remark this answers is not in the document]
AuroraEmiya: Fixed in [f609248](https://github.com/vllm-project/vllm-ascend/commit/f609248acad333dbee2d0324b56490ff2aaa3752): _git_tracked_files catches both OSError and ValueError around path resolution and returns None, preserving the existing non-Git traversal fallback. Verification is the full suite linked below; this specific injected OSError branch does not have a separate dedicated regression test.

Current verification: [cache-engine + concurrency suite, 41 passed](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35215229554). This is a private validation repository (access may be required); validation SHA 79612d988f6a41225e98c927a94f6a21882eac0a has production files identical to PR head 6101f59ea291e06ac6d0dfb1ca36b48cb271b2cd.

csrc/scripts/build_cache.py
```diff
@@ -0,0 +1,1725 @@
...
+        environment_files = list(environment_files) + _candidate_cann_metadata_files()
+    else:
+        raise ValueError(f"unknown environment profile: {profile}")
+
+    seen_tools: set[str] = set()
+    for index, tool in enumerate(list(environment_tools) + default_tools):
+        resolved = _resolve_tool(tool)
+        if resolved is None:
+            continue
+        canonical = str(Path(resolved).resolve())
```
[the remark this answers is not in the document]
AuroraEmiya: Fixed in [f609248](https://github.com/vllm-project/vllm-ascend/commit/f609248acad333dbee2d0324b56490ff2aaa3752): tool canonicalization catches OSError and falls back to the absolute path. The tool/version fingerprinting continues rather than dropping the tool from identity. Verification is the full suite linked below; there is no separate fault-injection regression for this resolution branch.

Current verification: [cache-engine + concurrency suite, 41 passed](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35215229554). This is a private validation repository (access may be required); validation SHA 79612d988f6a41225e98c927a94f6a21882eac0a has production files identical to PR head 6101f59ea291e06ac6d0dfb1ca36b48cb271b2cd.

csrc/scripts/build_cache.py
```diff
@@ -0,0 +1,1725 @@
...
+                "version": output,
+            }
+        )
+
+    seen_files: set[str] = set()
+    for raw_path in environment_files:
+        path = raw_path.expanduser()
+        if not path.is_file():
+            continue
+        digest = _sha256_file(path)
```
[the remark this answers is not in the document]
AuroraEmiya: Addressed with a correctness-preserving alternative in [f609248](https://github.com/vllm-project/vllm-ascend/commit/f609248acad333dbee2d0324b56490ff2aaa3752). We must not skip unreadable compiler metadata: doing so could reuse an entry with an incomplete environment identity. Explicit missing metadata is rejected, and an OSError while hashing is reported as a contextual RuntimeError instead of silently constructing a weaker key. Regression test_explicit_environment_metadata_cannot_be_silently_dropped verifies that required metadata cannot disappear from identity. The suggested skip-and-continue behavior was deliberately not adopted.

Current verification: [cache-engine + concurrency suite, 41 passed](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35215229554). This is a private validation repository (access may be required); validation SHA 79612d988f6a41225e98c927a94f6a21882eac0a has production files identical to PR head 6101f59ea291e06ac6d0dfb1ca36b48cb271b2cd.

csrc/cmake/build_cache.cmake
```diff
@@ -0,0 +1,207 @@
...
+        ${VLLM_ASCEND_BUILD_CACHE_DIR}
+        --domain
+        ${CACHE_DOMAIN}
+        --unit
+        ${CACHE_UNIT}
+        --output-dir
+        ${CACHE_OUTPUT_DIR}
+        --environment-profile
+        ${CACHE_ENVIRONMENT_PROFILE}
+    )
```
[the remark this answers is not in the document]
AuroraEmiya: Fixed in [f609248](https://github.com/vllm-project/vllm-ascend/commit/f609248acad333dbee2d0324b56490ff2aaa3752): scalar paths/arguments in the CMake cache adapter are quoted and list iteration preserves argument boundaries. Follow-up commit 9e97b1354 fixes the regression fixture's inherited CMake cache variable. test_cmake_adapter_preserves_arguments_with_spaces passes in the current suite and exercises paths containing spaces.

Current verification: [cache-engine + concurrency suite, 41 passed](https://github.com/vllm-ascend-ci-dev/vllm-ascend/actions/runs/35215229554). This is a private validation repository (access may be required); validation SHA 79612d988f6a41225e98c927a94f6a21882eac0a has production files identical to PR head 6101f59ea291e06ac6d0dfb1ca36b48cb271b2cd.

AuroraEmiya:
[code-review] High — NEW regression in `1f0d4e186a40239b884c2abfb6351018dee8f6f5`: scheduled upstream E2E now relies solely on the canonical exact L0, but the same commit removed the OBS backend selection from the workflow that performs the remaining direct L0 restores.

**Exact location:** `.github/workflows/schedule_e2e_upstream_test.yaml`. The commit correctly removes the three direct L1 fallback restores, but it also deletes the workflow-level `RUNS_ON_S3_BUCKET_CACHE`, endpoint, path-style, region, and runner-name env block. The three `runs-on/cache/restore@v5` L0 steps remain and use `vllm-ascend-build-v2-*` keys. `_build_csrc_cache.yaml` publishes those canonical L0 artifacts with the Huawei OBS backend explicitly selected, so producer and consumer no longer have an explicit guarantee that they are using the same backend.

**Why it matters:** after the L1 fallback removal, this L0 lookup is the only cache fast path for scheduled upstream E2E. If `runs-on/cache` falls back to its default backend here, a valid OBS-produced L0 is invisible, every job takes the NPU source-build fallback, and the updated design-doc statement that scheduled upstream E2E consumes the canonical producer's exact L0 is not operationally true. **Classification:** product/integration + CI-performance.

**Smallest next action:** restore the six OBS backend-selection env vars once at workflow scope (or equivalent job scope) for these restore-only consumers; do not pass write credentials. Then verify the effective backend/key contract statically or with one lookup. **Simplification opportunity:** one workflow-level backend block covers all three remaining L0 restores; no new helper/action or L1 path is needed.

AuroraEmiya:
[code-review] **High — formal selected-NPU fallback is broken before ordinary compilation can run**

- **Exact location:** `.github/workflows/_selected_tests.yaml`, step `Restore incremental cache for NPU fallback` (`uses: ./.github/actions/csrc-l1-restore`). In formal E2E run `35305968013`, the exact L0 restore correctly used `ascend-ci-cache-hk` and missed, but every selected NPU job that reached the L1 fallback failed with: `Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under .../.github/actions/csrc-l1-restore`. At least 16 NPU matrix jobs stop at this same step, so `Install vllm-project/vllm-ascend with device` is skipped instead of doing the intended source-build fallback.
- **Why it matters:** this violates the failure-containment contract: an optional L1 optimization is currently able to abort selected-test jobs on an L0 miss. The same run shows the PR rebases all 13 commits cleanly onto fixed base `21bfcb10...`, so this is not a merge-conflict/base-drift failure. `pre-commit`, mypy and CPU UT are green; the current blocking failure is the local-action integration path.
- **Classification:** product / CI integration.
- **Smallest next action:** make the L1 helper independent of the source tree that is checked out/rebased. The repository already uses this pattern in `_ensure_csrc_cache.yaml`: sparse-checkout current workflow helpers into a stable helper directory, then invoke the composite action from that directory while pointing `source-root` at the rebased source tree. Alternatively, if the local action itself cannot be made available, skip L1 and continue directly to the existing NPU source build rather than failing the job.

**Review-only evidence update:** the PR description still says formal selected NPU tests are `BLOCKED-BY-REVIEW-LABEL`. That is now stale: run `35305968013` actually dispatched selected tests and exposed the failure above. After fixing/re-running, update the description to the current formal result rather than the old skipped run.

AuroraEmiya:
[code-review] **High — `70ad1cd` fixes the rebase/source-tree diagnosis, but the selected-test L1 fallback is still unusable on the NPU container runners.**

- **Exact location:** `.github/workflows/_selected_tests.yaml`, the new `Checkout cache workflow helpers` step and the following `Restore incremental cache for NPU fallback` local action. In formal E2E run `35312835137`, the helper checkout succeeds and checks out PR merge commit `d429d0a...` under `/__w/vllm-ascend/vllm-ascend/.csrc-cache-workflow`; the exact L0 restore then cleanly misses on OBS. But the next `uses: ./.csrc-cache-workflow/.github/actions/csrc-l1-restore` is resolved by the runner under `/home/runner/_work/vllm-ascend/vllm-ascend/.csrc-cache-workflow/...` and fails with `Can't find 'action.yml', 'action.yaml' or 'Dockerfile'`. All selected NPU jobs that reach this step fail before `Install vllm-project/vllm-ascend with device`, so the ordinary source-build fallback is still never reached.
- **Why it matters:** this is stronger evidence than the previous failure: the helper is no longer lost by the source rebase; the checkout itself succeeds. The remaining failure is the runner/container workspace boundary for a local composite action. An optional L1 optimization is therefore still a hard dependency for selected tests, violating the intended fail-open containment contract.
- **Classification:** product / CI integration + failure containment.
- **Smallest next action:** stop trying to make this selected-test path depend on a local `uses: ./...` action. The minimum rollback-safe fix is to skip L1 after an L0 miss and continue directly to the existing NPU source build. If selected-test L1 is retained, invoke only the minimal key-prep/restore logic as ordinary job steps inside the container (prefer a small testable repository script for key preparation plus `runs-on/cache/restore@v5`) rather than adding another checkout/copy/sync layer.
- **Simplification opportunity:** the new helper checkout plus local-action indirection can be removed from `_selected_tests.yaml` (roughly 12–15 lines) instead of adding host/container synchronization glue. Keep the composite action only in runner paths where local actions are actually visible.

Current base drift does not add a cache/workflow conflict: `main` is two commits ahead of the fixed E2E base `3caf228d...`, and those changes are in docs/tests/model/runtime files rather than the csrc-cache workflow implementation.

AuroraEmiya:
[code-review] **Medium — `b7d059c` fixes selected-test failure containment by removing L1 restore, but this is now a material scope reduction that the PR description still overstates.**

- **Exact location:** `.github/workflows/_selected_tests.yaml` and `docs/source/developer_guide/Design_Documents/persistent_csrc_build_cache.md` at `b7d059c`, versus the PR description’s generic `L0 miss -> persistent L1 restore -> normal source build` flow and statement that production integrations cover `selected tests`.
- **New evidence:** `b7d059c` deletes the selected-test `csrc-l1-restore` path entirely. Formal E2E run `35316392719` now confirms the intended containment fix: selected NPU jobs pass the exact L0 restore step, then successfully perform ordinary source installation and run tests. For L0-miss PRs, however, selected tests no longer consume persistent L1 at all, so a small csrc edit still takes the full NPU compile path. The design document now explicitly records selected tests as the runner/container-boundary exception, but the top-level PR description does not.
- **Why it matters:** reviewers evaluating the main performance contract can currently read the PR body as claiming `L0 miss -> L1 reuse` for selected tests, while the production workflow intentionally does `L0 miss -> full source build`. That is a meaningful difference for the primary PR csrc-change use case, even though correctness/fail-open behavior is now improved.
- **Classification:** product-performance / review-only.
- **Smallest next action:** update the PR description to explicitly carve selected tests out of persistent-L1 integration and state that they are currently L0-only with direct source fallback. Do not reintroduce selected-test L1 in this PR unless the runner host/container boundary is solved with a path that preserves fail-open behavior.

AuroraEmiya:
[code-review] **High — the selected-test runner/container fix is incomplete: the same local-composite dependency still exists in the PR-triggered nightly source-build paths.**

- **Exact locations:** `.github/workflows/_e2e_nightly_single_node.yaml`, `.github/workflows/_e2e_nightly_single_node_560t.yaml`, and `.github/workflows/_e2e_nightly_single_node_models.yaml`, steps `Checkout incremental-cache workflow helpers` → `Restore incremental cache for source checkout`. All three jobs run inside `container: image: ...`, sparse-checkout `.github/actions/csrc-l1-restore`, then invoke it with `uses: ./.incremental-cache-workflow/.github/actions/csrc-l1-restore` whenever `request_id != ''`.
- **Why it matters:** formal selected-NPU run `35312835137` already demonstrated that this exact helper-checkout + local-`uses` pattern can resolve the checkout inside the container workspace while the runner resolves the local action from its host workspace, failing before ordinary source installation. `b7d059c` removed that dependency from `_selected_tests.yaml`, but these three production reusable workflows still retain it. If their NPU runner topology has the same workspace mapping, a PR `/nightly`-style source run will abort at optional L1 restore instead of falling through to `uv pip install -e .`. The PR body also notes these scheduled/reusable callers were not runtime-dispatched in the validation fork, so there is no contrary runtime evidence for this path. **Classification: product / CI integration + failure containment.**
- **Smallest next action:** apply the same rollback-safe treatment used for selected tests: remove the local L1 restore from these job-container PR source paths and let them source-build on a miss. Reintroduce L1 only via ordinary in-container steps (key-prep script + `runs-on/cache/restore@v5`) once that path is validated on these runners.
- **Simplification opportunity:** deleting the duplicated helper-checkout/local-action blocks removes roughly 45–50 lines across the three workflows and eliminates one unvalidated runner/container coupling instead of adding more copy/sync glue.

AuroraEmiya:
[code-review] **High/Medium — two NEW safety/failure-containment gaps remain at current head `b7d059c`.**

1. **High — `csrc/scripts/build_cache.py::_validate_entry/_validate_artifact/_restore_entry` — classification: product/correctness + trust-boundary.** Restored entry manifests are treated as trusted path records. `artifact["path"]` is joined directly onto both the cached `artifact_root` and the live `output_dir`/publish directory with no artifact-record schema/type validation and no containment check. A malformed record can raise `KeyError`/`TypeError` outside the current `except OSError` and hard-fail the build; a poisoned record such as an absolute path or `../../...` can escape the intended cache/output roots during validation or restore. This matters because L1 snapshots come from shared OBS and compatibility fallback deliberately consumes snapshots produced by other jobs: cache validation must turn corrupted/poisoned remote state into a MISS, never let manifest data select arbitrary host paths. **Smallest next action:** add one small artifact-record validator used by both validation and restore: require a dict with a non-empty normalized relative path, reject absolute/`.`/`..` components, validate `kind`/target types, and reject the whole entry as a MISS on any malformed record; additionally assert source/destination containment before publication. Add regressions for `../../`, absolute path, missing `path`, and wrong-shaped `artifacts`.

2. **Medium — `.github/actions/csrc-l1-restore/action.yaml` → `.github/workflows/_build_csrc_cache.yaml::Install vllm-ascend` — classification: product/CI integration + failure containment.** The restore action now correctly returns `supported=false` when the engine/key fingerprint is unavailable, but it only exports `VLLM_ASCEND_BUILD_CACHE_EVENT_LOG` to `GITHUB_ENV` after successful key preparation. `_build_csrc_cache.yaml` then unconditionally executes `rm -f "$VLLM_ASCEND_BUILD_CACHE_EVENT_LOG"` and `touch "$VLLM_ASCEND_BUILD_CACHE_EVENT_LOG"` before `uv pip install -e .`. On the intended `supported=false` path that variable is unset, so `touch ""` fails under the job's `bash -e` shell before ordinary compilation starts. The metadata-read-failure/historical-source fail-open contract is therefore not actually preserved in this central producer path. **Smallest next action:** define/export the event-log path before every early `supported=false` return (or set it once at job/workflow env scope), then add one integrated regression where snapshot-key preparation fails and the source build still runs. This can be fixed by moving the common env export earlier; no additional conditional layer is needed.

.github/actions/csrc-l1-restore/action.yaml
```diff
@@ -0,0 +1,139 @@
...
+      shell: bash
+      env:
+        CACHE_DIR: ${{ inputs.cache-dir }}
+        CACHE_ARCHITECTURE: ${{ inputs.architecture }}
+        CACHE_SOC_VERSION: ${{ inputs.soc-version }}
+        CACHE_TOOLCHAIN_IMAGE: ${{ inputs.toolchain-image }}
+        SOURCE_ROOT: ${{ inputs.source-root }}
+        CALLER_CSRC_HASH: ${{ inputs.csrc-hash }}
+      run: |
+        set -euo pipefail
```
AuroraEmiya: What‘s this command for?

.github/actions/csrc-l1-restore/action.yaml
```diff
@@ -0,0 +1,139 @@
...
+        RUNS_ON_RUNNER_NAME: ""
+      with:
+        path: ${{ inputs.cache-dir }}
+        key: ${{ steps.key.outputs.primary_key }}
+        restore-keys: |
+          ${{ steps.key.outputs.same_csrc_prefix }}
+          ${{ steps.key.outputs.compat_prefix }}
+
+    - name: Reset L1 update marker
+      if: steps.key.outputs.supported == 'true'
```
AuroraEmiya, line 80: What's this action for?

.github/actions/csrc-l1-restore/action.yaml
```diff
@@ -0,0 +1,139 @@
...
+  csrc-hash:
+    value: ${{ steps.key.outputs.csrc_hash }}
+  cache-hit:
+    value: ${{ steps.restore.outputs.cache-hit }}
+  cache-matched-key:
+    value: ${{ steps.restore.outputs.cache-matched-key }}
+
+runs:
+  using: composite
+  steps:
```
AuroraEmiya, line 39: What 's this action for

.github/workflows/scripts/csrc_cache_targets.json
```diff
@@ -94,5 +94,13 @@
     "os": "openeuler",
     "image": "swr.cn-southwest-2.myhuaweicloud.com/base_image/ascend-ci/cann:9.1.0-310p-openeuler24.03-py3.12",
     "soc_version": "ascend310p1"
+  },
+  "a5-arm64-ubuntu": {
+    "device": "a5",
```
AuroraEmiya: Why add this config？

.github/actions/csrc-l1-restore/action.yaml
AuroraEmiya, line 51: `set -euo pipefail` enables strict Bash behavior for this small wrapper step: `-e` stops on an unexpected command failure, `-u` treats an unset variable as an error, and `pipefail` propagates failures inside pipelines. Expected cases where persistent L1 cannot be used are handled inside `prepare_csrc_l1_restore.py` and return `supported=false`, so this strict mode is only intended to prevent silent wrapper/setup errors.

.github/actions/csrc-l1-restore/action.yaml
AuroraEmiya: This step clears the local L1 `\.updated` dirty marker after restoring a remote snapshot. `build_cache.py` touches `.updated` only when the current build creates or replaces a local L1 entry, and `csrc-l1-save` publishes a shared snapshot only when that marker exists. Clearing it after restore means an all-HIT build does not re-upload an unchanged snapshot; if any action later MISSes and saves a new entry, the marker is recreated.

.github/actions/csrc-l1-restore/action.yaml
AuroraEmiya: This composite action is the persistent-L1 transport/orchestration layer, not the per-action cache engine. It prepares the compatible snapshot key, restores a candidate L1 snapshot from the shared OBS backend through `runs-on/cache`, exports the local cache directory for the subsequent normal source build, and resets the dirty marker. The actual correctness decision for each compiler action still happens later in `csrc/scripts/build_cache.py`, which recomputes the semantic action key and validates artifacts before declaring a HIT. In the latest `f9198add`, key preparation was moved into `prepare_csrc_l1_restore.py` so selected NPU tests can reuse the same logic without depending on a local composite action across the container/runner workspace boundary.

AuroraEmiya:
[code-review] **Medium — `f9198add` fixes the container/local-action failure, but the design document now contradicts the final selected/nightly L1 behavior.**

- **Exact locations:** `docs/source/developer_guide/Design_Documents/persistent_csrc_build_cache.md` (`Architecture` and `Persistent transport and trust`) versus `.github/workflows/_selected_tests.yaml`, `_e2e_nightly_single_node.yaml`, `_e2e_nightly_single_node_560t.yaml`, and `_e2e_nightly_single_node_models.yaml`. The doc still says composite actions own snapshot keys/OBS transport and that selected tests do `L0 miss -> ordinary source build` without persistent L1. At current head, those container jobs instead call `prepare_csrc_l1_restore.py` directly and invoke `runs-on/cache/restore@v5` from workflow YAML; selected tests now restore L1 after an L0 miss.
- **New runtime evidence:** formal E2E run `35484861051` has A2/A3/310P selected jobs successfully completing `Prepare incremental cache key`, `Restore incremental cache`, and marker reset before entering the source install, so this is the actual production path rather than dead YAML.
- **Why it matters:** the design doc is the stated trust/failure-boundary contract. In its current form it gives reviewers and future maintainers the wrong selected-test performance scope and the wrong ownership model for key preparation/transport, making it easy to undo the direct in-container exception or audit the wrong trust surface.
- **Classification:** review-only / design-contract maintainability.
- **Smallest next action:** update the doc only: make `prepare_csrc_l1_restore.py` the shared key-preparation contract, describe the composite action as the normal adapter where local actions are usable, document direct in-container `runs-on/cache/restore` for selected/nightly jobs, and state that selected tests are restore-only L1 consumers after L0 miss. No production-code change or additional device rerun is required for this finding.

AuroraEmiya:
[code-review] **High — NEW merge blocker/base drift now overlaps the cache workflow callers, and the latest four PR commits are net-zero review churn.**

- **Exact workflows/base changes:** GitHub now reports PR #16718 as `mergeable=false`, and the bot posted the new conflict notice after head `75e998f`. Since the previous reviewed state, `main` merged `feb6bd9` / #16965, which changes `.github/workflows/schedule_weekly_test_a2.yaml`, and `4be858b` / #16499, which changes `.github/workflows/schedule_e2e_test.yaml` and `.github/workflows/schedule_nightly_test_a2.yaml`; all three files are also modified by this PR. #16965 removes the native weekly cron, while #16499 adds the new daytime/daily dispatch path, changes PR-nightly `request_id` semantics, and adds nightly `run-name` behavior. Current `main` is `44f9bf5`; that latest commit itself states workflows are unchanged.
- **Why it matters:** these are production orchestration surfaces for the cache-enabled nightly/E2E paths. The eventual conflict resolution must preserve the new upstream trigger/dispatch semantics **and** this PR's cache caller inputs (`soc_version`, csrc-cache preparation/OBS secret contract). A mechanical resolution can silently drop daily/weekly scheduling behavior or the cache integration even though the cache engine itself is unchanged. The current head also has only the lightweight DCO/PR-create/main checks; there is no new E2E/pre-commit evidence for the post-conflict merge result. **Classification: infra/product workflow integration + review-only.**
- **Smallest next action:** rebase `75e998f` onto current `main`, reconcile these workflow files explicitly, then rerun YAML/actionlint/pre-commit plus the affected workflow admission/dispatch checks. No cache-engine or full A2/A3/310P rerun is needed if the rebase does not change csrc/cache semantics.
- **Simplification opportunity:** the four commits after `f9198add` (`b0eb7842` → `54f92186` and `d0389a57` → `75e998f`) are two experiment/revert pairs; the compare has `files: []`, so the current tree is identical to `f9198add`. Drop/squash those net-zero commits during the rebase to reduce reviewer burden without changing behavior.

AuroraEmiya:
[code-review] **Medium — the new self-repository action path is now runtime-proven, so the three nightly workflows can delete their duplicated L1 transport implementation.**

- **Exact locations:** current head `ba21533c` uses `uses: $/.github/actions/csrc-l1-restore` in `.github/workflows/_selected_tests.yaml`, while `.github/workflows/_e2e_nightly_single_node.yaml`, `_e2e_nightly_single_node_560t.yaml`, and `_e2e_nightly_single_node_models.yaml` still each carry `Checkout incremental-cache workflow helpers` → key-prep Python → `runs-on/cache/restore@v5` + repeated OBS env/restore-prefix wiring → marker reset.
- **New evidence / why it matters:** formal E2E run `35493648769` on this head has already executed `Restore incremental cache` successfully through the `$/` composite on multiple 310P/A2/A3 Kubernetes-backed selected-test jobs, and several then completed the normal source install. This is materially different from the earlier failing workspace-relative `./` action path: the self-repository reference resolves the action independently of the container workspace. Keeping the nightly copies now duplicates the correctness-sensitive snapshot-key, OBS backend, fail-open, and dirty-marker contract in four places, making future schema/backend changes easy to drift. **Classification: maintainability + infra correctness-contract.**
- **Smallest next action:** replace each nightly helper/key/restore/reset block with the same `$/ .github/actions/csrc-l1-restore` call (without the space after `$/`) using `source-root: ./temp-vllm-ascend`, `architecture: ${{ runner.arch }}`, `soc-version: ${{ inputs.soc_version }}`, and the existing cache dir. Preserve the composite's internal `continue-on-error` behavior. The repository's actionlint `v1.7.7` predates GitHub's `$/` syntax, so keep the temporary lint exception narrowly scoped until actionlint catches up.
- **Simplification opportunity:** this removes roughly **40–45 lines per nightly workflow (about 120–135 lines total)**: three sparse helper checkouts, three key-prep blocks, three copies of OBS env/key restore wiring, and three marker-reset steps, while centralizing the cache contract in the action that is now actually exercised on the target runner topology.

AuroraEmiya:
/nightly msa-index-score-a3

AuroraEmiya:
/nightly msa-index-score-a3

AuroraEmiya:
[code-review] **High — historical-source fail-open is still broken in the Docker export path.**

- **Exact locations:** `.github/workflows/schedule_release_code_and_wheel.yml` (all three `Export incremental cache from wheel image` steps) and `.github/workflows/_schedule_image_build.yaml::Export incremental cache from image`, via `.github/workflows/scripts/export_csrc_l1_from_image.sh`. `csrc-l1-restore` intentionally returns `supported=false` when the checked-out source predates `csrc/scripts/build_cache.py`; the release `workflow_dispatch` explicitly offers historical tags such as `v0.23.0`, and that tag has no cache engine / `/workspace/vllm-ascend/build_cache`. The ordinary wheel Docker build can therefore succeed, but the exporter still runs unconditionally and its strict `docker cp ".../build_cache/."` fails, aborting before wheel repair/archive/release. `_schedule_image_build.yaml` has the same assumption on an L0 miss.
- **Why it matters:** optional L1 is not actually fail-open for the historical-source case the new helper explicitly supports; a cache-unsupported source can become a hard release/image-build failure after ordinary compilation succeeded. **Classification: product/release-CI + failure containment.**
- **Smallest next action:** gate Docker export (and preferably save) on the existing `${{ steps.wheel-l1.outputs.supported == 'true' }}` / `${{ steps.image-l1.outputs.supported == 'true' }}` output. Reuse that signal; no new fallback/helper is needed. Add one cheap historical-tag/no-cache-dir regression around the exporter path; no NPU matrix rerun is required.

AuroraEmiya:
Superseded by #17006, which carries the validated final tree as five clean semantic commits rebased onto current main. This PR remains the iterative integration and runtime-evidence record; its final validation head is bde5d851375b32ec6fb64e001d9c6124687cfa5c.

AuroraEmiya:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- Doc Test
- E2E

Tflowers-0129:
整体思路没问题，可不可以拆分PR，一步一步来呢

ZT-AIA (approved):

zhangxinyuehfad:
/rerun
[Bot]: rerun completed.

**Rerun (failed jobs only):**
- E2E

AuroraEmiya:
/rerun
[Bot]: rerun completed.

**Failed:**

MrZ20:
https://github.com/vllm-project/vllm-ascend/actions/runs/35944630061/job/107523696375#step:11:27

The following error occurs on 310p and a3 jobs. Please determine whether it affects existing functions：
```bash
Error listing objects with prefix vllm-ascend-build-v2-ARM64-9.1.0-a3-ubuntu22.04-py3.12-a70dec09f20eb6702ada746dbf7f01f2bd83d2bc2043bbc51b1b567bad8b1a7b in bucket ascend-ci-cache-hk: CredentialsProviderError: Could not load credentials from any providers
```

csrc/scripts/build_cache.py
```diff
@@ -0,0 +1,1746 @@
...
+        }
+        _atomic_write_json(owners_path, owners)
+        _atomic_write_json(action_state_path, action_state)
+
+
+def _run_build_command(
+    args: argparse.Namespace,
+    command: Sequence[str],
+) -> tuple[int, float]:
+    environment = _parse_set_env(args.set_env)
```
MrZ20: Can we confirm this scenario: the `--set-env` environment variables get passed to the compiler and change how compilation works, but they aren’t included in `recipe_hash` when calculating the cache key. So if `set_env` changes (like toggling `TILINGKEY_PAR_COMPILE` between `0` and `1`), the cache key stays the same, causing a wrong cache hit and reusing old build outputs.

.github/workflows/_ensure_csrc_cache.yaml
```diff
@@ -251,12 +292,17 @@ jobs:
   build-missing-caches:
     name: Build csrc cache (${{ matrix.target.id }})
     needs: prepare-cache-checks
-    if: needs.prepare-cache-checks.outputs.has_missing == 'true'
+    if: >-
+      needs.prepare-cache-checks.outputs.has_missing == 'true' &&
+      needs.prepare-cache-checks.outputs.has_write_credentials == 'true'
```
MrZ20: For the build process, I would prefer incremental compilation to run on the CPU instead of the NPU. 
If this is truly not achievable, the code for the current CPU build needs to be removed, because the current logic is skipped and will never be reached.
```
CPU job:
Restore L1
HIT xxx
MISS a small number
Save L1
Save L0

Then NPU job:
L0 HIT
csrc cache hit: skip kernel compilation
```

PR #16721, opened 2026-09-16
[CI] Pin DeepSeek-V4 Flash GPQA nightly to low reasoning effort

GDzhu01:
## Summary

- Pass the optional `reasoning_effort` benchmark setting through to AISBench chat request generation.
- Set `reasoning_effort: low` for the DeepSeek-V4 Flash W8A8 A3 nightly GPQA case. Keep `thinking: true` and the performance case unchanged.
- Add a focused test for the generated request config, including unchanged behavior when the setting is absent.

## Context

The [failing nightly job](https://github.com/vllm-project/vllm-ascend/actions/runs/34497673764/job/103072819176#logs) evaluated this case without an explicit reasoning effort. The DeepSeek-V4 default thinking strength changed upstream, so this pins the intended GPQA protocol rather than relying on that default.

An earlier manual GPQA Diamond run using `reasoning_effort="low"` and `chat_template_kwargs={"thinking": True}` scored 88.38% (175/198) on the e5118d15/vLLM 0.28.0 setup. That is supporting evidence, not a validation of this PR's current-main nightly run.

## Validation

- Parsed the modified YAML and Python files; confirmed only the GPQA benchmark selects `low`.
- `git diff --check` passed.
- The new unit test and full A3 nightly GPQA run are pending CI/runner execution.


tools/aisbench.py
```diff
@@ -175,6 +176,12 @@ def _init_request_conf(self):
         if self.thinking:
             field_thinking = 'chat_template_kwargs={"thinking": True}'
             content = re.sub(r"ignore_eos.*", f"ignore_eos=False,\n            {field_thinking},", content)
+        if self.reasoning_effort is not None:
+            content = re.sub(
+                r"ignore_eos.*",
+                f'ignore_eos=False,\n            reasoning_effort="{self.reasoning_effort}",',
```
GDzhu01, line 182: [Security] reasoning_effort comes from YAML and is interpolated directly into generated Python source. A value containing a quote, backslash, or newline can make the request config invalid; if this file is imported or executed, it can also inject Python into CI. Please validate against the supported effort values and serialize the literal safely (for example, repr after validation) instead of embedding raw text. A regression case with quoted and newline values would lock this down.

tools/aisbench.py
```diff
@@ -175,6 +176,12 @@ def _init_request_conf(self):
         if self.thinking:
             field_thinking = 'chat_template_kwargs={"thinking": True}'
             content = re.sub(r"ignore_eos.*", f"ignore_eos=False,\n            {field_thinking},", content)
+        if self.reasoning_effort is not None:
```
GDzhu01, line 179: [Maintainability / Readability] Every optional sampling field rewrites the same ignore_eos line, so the resulting config depends on mutation order and adding another field requires another regex copy. This is already difficult to audit when thinking and reasoning_effort are both enabled. Please collect the enabled request arguments and render or insert them once (or edit a structured config) so ordering and comma handling live in one place.

weijinqian0 (approved):

